PR #736, opened 2016-01-31
Accept rename to same name

### LINKED ISSUE #724: Behavior of ds.rename when old and new name are the same
Before #648, passing `Dataset.rename` a `name_dict` whose keys and values were identical did not raise an exception.  Now, however, it raises a ValueError:
```python
arr = xr.DataArray(range(2), name='arrname')
ds = arr.to_dataset()
ds.rename({'dim_0':'dim_0'})
<ipython-input-15-a5b851f9fb10> in <module>()
----> 1 ds.rename({'dim_0':'dim_0'})

/Users/spencerahill/anaconda/lib/python2.7/site-packages/xarray/core/dataset.pyc in rename(self, name_dict, inplace)
   1245                                  "variable in this dataset" % k)
   1246             if v in self:
-> 1247                 raise ValueError('the new name %r already exists' % v)
   1248
   1249         variables = OrderedDict()

ValueError: the new name 'dim_0' already exists
```

This is easy enough to handle with a try/except clause in my own code, but it would be nice (for me at least) for `rename` to not raise anything for these cases.  Since the result is simply the original Dataset unchanged, this could be implemented by replacing line 1246 with `if v in self and k != v:`

The use case is that we have data coming from multiple sources, often with differing internal names for coordinates and variables, and we're programmatically forcing them to have consistent names via `rename`.  Sometimes, a piece of data already has the name that we are after, hence the non-rename `rename`.  Thanks!

spencerahill:
Closes #724

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1223,7 +1223,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         ----------
         name_dict : dict-like
             Dictionary whose keys are current variable or dimension names and
-            whose values are new names.
+            whose values are new names (although the new name is permitted
+            to be the same as the old name).
         inplace : bool, optional
             If True, rename variables and dimensions in-place. Otherwise,
             return a new dataset object.
@@ -1243,7 +1244,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
             if k not in self:
                 raise ValueError("cannot rename %r because it is not a "
                                  "variable in this dataset" % k)
-            if v in self:
+            if v in self and k != v:
                 raise ValueError('the new name %r already exists' % v)
 
         variables = OrderedDict()
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1223,8 +1223,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         ----------
         name_dict : dict-like
             Dictionary whose keys are current variable or dimension names and
-            whose values are new names (although the new name is permitted
-            to be the same as the old name).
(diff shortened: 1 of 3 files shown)


xarray/core/dataset.py
```diff
@@ -1223,7 +1223,8 @@ def rename(self, name_dict, inplace=False):
         ----------
         name_dict : dict-like
             Dictionary whose keys are current variable or dimension names and
-            whose values are new names.
+            whose values are new names (although the new name is permitted
```
shoyer: Maybe it would be better to simple state "are desired names". It seems pretty logical that the same names would be OK.

xarray/test/test_dataset.py
```diff
@@ -1101,6 +1101,13 @@ def test_rename(self):
         with self.assertRaises(UnexpectedDataAccess):
             renamed['renamed_var1'].values
 
+    def test_rename_same_name(self):
+        data = create_test_data()
+        newnames = {'var1': 'var1', 'dim2': 'dim2'}
+        renamed = data.rename(newnames)
+        self.assertTrue('var1' in renamed)
```
shoyer: use `self.assertDatasetIdentical(renamed, data)`

shoyer:
Could you please add a brief note to what's new for 0.7.1? This should probably go under "enhancements" or "bug fixes". You'll need to create a new section `whats-new.rst` since this is the first fix -- don't sweat the details very much, I'll go over them before the release :).

spencerahill:
Sure, will do all of these.

shoyer:
Thanks!

spencerahill:
No problem.  Sorry for the extra commit -- I tried to squash them together, but I'm new to rebase and apparently didn't do it right.  Thanks for helping me incorporate this!

Issue #739, opened 2016-02-02
Numerical operations with empty DataArrays fail


MaximilianR:
We're running into this a fair amount at the moment, and so constantly having to check whether DataArrays are empty before multiplying them together:

```python
In [6]: a=xr.DataArray([[]])

In [7]: b=xr.DataArray([[1,2,3]])

In [8]: a*b
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-8-8201c27d19b7> in <module>()
----> 1 a*b

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/dataarray.py in func(self, other)
   1182                 if empty_indexes:
   1183                     raise ValueError('no overlapping labels for some '
-> 1184                                      'dimensions: %s' % empty_indexes)
   1185             other_variable = getattr(other, 'variable', other)
   1186             other_coords = getattr(other, 'coords', None)

ValueError: no overlapping labels for some dimensions: ['dim_1']

```

Is the failure a useful one? If it's taking an inner join, isn't it OK to return an empty result if there are no matching labels?

FWIW Pandas & Numpy both allow this:

```python
In [12]: pd.Series()*pd.Series()
Out[12]: Series([], dtype: float64)

In [13]: pd.DataFrame()*pd.DataFrame()
Out[13]: 
Empty DataFrame
Columns: []
Index: []

In [16]: np.asarray([])*np.asarray([])
Out[16]: array([], dtype=float64)
```

shoyer:
I would be OK with changing this. This would be pretty straightforward and almost entirely backwards compatible.

My original thought was that was that raising an error would more immediately identify cases where the user didn't have compatible coordinates to both arrays, because in many (most?) cases this does indicate an error in the original code. Given that concern, we might consider switching the error to only occur if both of the original dimensions were *not* size 0. On the other hand, making things more complex is usually not the solution :).

PR #741, opened 2016-02-02
Allow empty result of numerical operations between DataArrays

MaximilianR:
Resolves https://github.com/pydata/xarray/issues/739

@shoyer let me know your thoughts and I'll add what's new note

### DIFF
--- a/xarray/test/test_dataset.py
+++ b/xarray/test/test_dataset.py
@@ -2182,8 +2182,10 @@ class TestDataset(TestCase):
         actual = ds + subset
         self.assertDatasetIdentical(expected, actual)
 
-        with self.assertRaisesRegexp(ValueError, 'no overlapping labels'):
-            ds.isel(x=slice(1)) + ds.isel(x=slice(1, None))
+
+        actual = ds.isel(x=slice(1)) + ds.isel(x=slice(1, None))
+        expected = ds.drop(ds.x, dim='x')
+        self.assertDatasetEqual(actual, expected)
 
         actual = ds + ds[['bar']]
         expected = (2 * ds[['bar']]).merge(ds.coords)
@@ -2191,6 +2193,7 @@ class TestDataset(TestCase):
 
         with self.assertRaisesRegexp(ValueError, 'no overlapping data'):
             ds + Dataset()
+
 
         with self.assertRaisesRegexp(ValueError, 'no overlapping data'):
             Dataset() + Dataset()
--- a/xarray/test/test_dataset.py
+++ b/xarray/test/test_dataset.py
@@ -1620,7 +1620,7 @@ class TestDataset(TestCase):
 
         ds = Dataset({'x': ('time', np.arange(100)),
                       'time': pd.date_range('2000-01-01', periods=100)})
-        with self.assertRaisesRegexp(ValueError, 'no overlapping labels'):
+        with self.assertRaisesRegexp(ValueError, 'incompat.* grouped binary'):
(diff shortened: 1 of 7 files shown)


MaximilianR:
Not sure if I'm doing something foolish - I had some issues running the tests on my machine because of the `from .version import version as __version__`. Let me know if there's a quick fix beyond me commenting that line out.

MaximilianR:
Forgive the python question - please disregard unless you know the answer off hand - do you know if there any way of monkey patching this into our code? I did that for the 0.7 fixes we added, but I'm not sure that's possible here: `_binary_op` gets called many times upon `import xarray`, and so I think the prior `_binary_op` is already 'embedded' in all the functions before we have a chance to monkey patch.

spencerahill:
@MaximilianR FWIW I had problems with `.version` recently when I cloned from the main repo...had to copy the `version.py` from another xarray install.  Maybe has something to do with .gitignore: xarray/version.py is included there.

shoyer:
Version.py gets created when you install xarray, e.g., when you run `pip install -e .` after running git clone.

spencerahill:
@shoyer thanks!  Good to know.

shoyer:
For consistency, we should also fix arithmetic with Dataset (allow no overlapping labels and no overlapping variables) and GroupBy objects:
https://github.com/pydata/xarray/blob/v0.7.0/xarray/core/dataset.py#L1933
https://github.com/pydata/xarray/blob/v0.7.0/xarray/core/dataset.py#L1975
https://github.com/pydata/xarray/blob/v0.7.0/xarray/core/groupby.py#L220

MaximilianR:
Super - those are now updated @shoyer

xarray/test/test_dataset.py
```diff
@@ -1868,8 +1868,9 @@ def test_fillna(self):
         expected = Dataset({'a': ('x', [-1, 1, -1, 3]), 'b': np.nan})
         self.assertDatasetIdentical(expected, actual)
 
-        with self.assertRaisesRegexp(ValueError, 'no overlapping'):
-            ds.fillna({'x': 0})
+        actual=ds.fillna({'x': 0})
+        expected=ds
+        self.assertDatasetEqual(actual, expected)
```
MaximilianR: I think these may now be *too* liberal. Here, there is no variable called `x`. I think this _should_ raise - it seems fairly different to multiply empty arrays vs supplying variables that don't exist.

I'll ponder further, let me know if you have any thoughts on this in the meantime

MaximilianR:
@shoyer I think this is a decent balance:
- Allow empty dimensions
- But raise if non-existent dimensions are supplied

I think the code and tests now reflect that. Let me know your thoughts / comments

MaximilianR:
This is my last PR that's ready to go @shoyer

shoyer:
Still thinking about the fillna behavior... I agree, it's weird that `ds.fillna({'missing_variable': 0})` works without raising an error.

I would say this is consistent with in-place arithmetic on Datasets (i.e., left alignment of variable names), but we actually don't allow that with different variables currently... probably because I wanted to avoid this headache :). I think I'll add a special case in to ensure that we raise an error if there are unexpected variables in the fillna argument.

shoyer:
OK, going to merge this...

shoyer:
I'm probably going to call this an "enhancement" rather than a bug fix, but otherwise this looks good... I'll fix up the what's new page shortly.

MaximilianR:
Cheers!

PR #743, opened 2016-02-02
BUG: reindexing an array with complex values should keep dtype=complex

### LINKED ISSUE #738: reindex_like with tolerance changes the type of returned DataArray
I am really excited about all of the new changes to xarray! reindexing with tolerance has great potential to clean up some sort of nasty stuff that I use a lot but have tucked away into a corner of my code. I am not sure I am 100% comfortable with it at the moment, however, because it seems to be changing the types of my data in unexpected ways.

Bei:

```
test = xr.DataArray([1,2,3], coords=[[0.1, 0.2, 0.3]])
test_i = xr.DataArray([5,6,7,8], coords=[[0.21, 0.31, 0.41, 0.51]])

test.dtype
test.reindex_like(test_i, method='pad', tolerance=0.2).dtype
```

the reindexed output has changed from int64 to float64!

Even more concerning is the complex example:

```
test = xr.DataArray(np.array([1,2,3], dtype=np.complex), coords=[[0.1, 0.2, 0.3]])

test.dtype
test.reindex_like(test_i, method='pad', tolerance=0.2).dtype
```

Conversion from complex to object? Oh no. 

Hope this isn't anything too big, let me know if I can help.
Thank

oceanEngineer:
Fixes #738 

I have added a simple behaviour for complex numbers in _maybe_promote. This fix means that  _maybe_promote does indeed return (dtype, np.nan + 0j) for complex dtypes, and the complex -> object conversion shown in #738 does not occur.

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -418,6 +418,8 @@ def _maybe_promote(dtype):
         # convert to floating point so NaN is valid
         dtype = float
         fill_value = np.nan
+    elif np.issubdtype(dtype, complex):
+        fill_value = np.nan + np.nan * 1j
     elif np.issubdtype(dtype, np.datetime64):
         fill_value = np.datetime64('NaT')
     elif np.issubdtype(dtype, np.timedelta64):
--- a/xarray/test/test_dataarray.py
+++ b/xarray/test/test_dataarray.py
@@ -635,6 +635,13 @@ class TestDataArray(TestCase):
         actual = expected.reindex(time=time2)
         self.assertDataArrayIdentical(actual, expected)
 
+        # regression test for #736, reindex can not change complex nums dtype
+        x = np.array([1, 2, 3], dtype=np.complex)
+        x = DataArray(x, coords=[[0.1, 0.2, 0.3]])
+        y = DataArray([2, 5, 6, 7, 8], coords=[[-1.1, 0.21, 0.31, 0.41, 0.51]])
+        re_dtype = x.reindex_like(y, method='pad').dtype
+        self.assertEqual(x.dtype, re_dtype)
+
     def test_reindex_method(self):
         x = DataArray([10, 20], dims='y')
         y = [-0.1, 0.5, 1.1]
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -19,6 +19,7 @@ Enhancements
 ------------
 
 - ``Dataset.rename`` and ``DataArray.rename`` support the old and new names being the same.  This had been supported prior to v0.7.0 but was broken in 0.7.0.
+- ``DataArray.reindex_like`` now maintains the dtype of complex numbers when reindexing leads to na values.
 
 .. _whats-new.0.7.0:
 


xarray/core/common.py
```diff
@@ -418,6 +418,9 @@ def _maybe_promote(dtype):
         # convert to floating point so NaN is valid
         dtype = float
         fill_value = np.nan
+    elif np.issubdtype(dtype, complex):
+        #explicit cast of np.nan is probibly not necassary
+        fill_value = np.complex(np.nan)
```
shoyer: Actually, let's use: `np.nan + np.nan * 1j`. Otherwise, the NaN is only done for real part of the number:
```
In [19]: complex(np.nan)
Out[19]: (nan+0j)
```

I'm not sure this actually matters but it can't hurt

shoyer:
This also needs a test. I would recommend adding a regression test in `test_reindex` (https://github.com/pydata/xarray/blob/v0.7.0/xarray/test/test_dataarray.py#L622) to verify that complex valued arrays keep a complex dtype when indexed. You can use the example you provided for the github issue for this.

shoyer:
Could you also please add a brief change note item to `whats-new.rst` under 0.7.1?

oceanEngineer:
tolerance keyword is not necessary to raise this error, simply raising na from reindex operation is enough. I have changed test to reflect this, which also removes the need for pandas version checking. 

new test code is:
```
x = np.array([1, 2, 3], dtype=np.complex)
x = xr.DataArray(x, coords=[[0.1, 0.2, 0.3]])
y = xr.DataArray([2, 5, 6, 7, 8], coords=[[-1.1, 0.21, 0.31, 0.41, 0.51]])
x.reindex_like(y, method='pad')
```

xarray/test/test_dataarray.py
```diff
@@ -635,6 +635,13 @@ def test_reindex(self):
         actual = expected.reindex(time=time2)
         self.assertDataArrayIdentical(actual, expected)
 
+        #regression test for #736, reindex should not change complex nums dtype
```
shoyer: per pep8, add a space after the `#` symbol

xarray/core/common.py
```diff
@@ -418,6 +418,9 @@ def _maybe_promote(dtype):
         # convert to floating point so NaN is valid
         dtype = float
         fill_value = np.nan
+    elif np.issubdtype(dtype, complex):
+        #explicit cast of np.nan is probibly not necassary
```
shoyer: This comment is no-longer relevant

shoyer:
Looks good to me, assuming tests pass. Could you please squash your commits into one?
http://pandas.pydata.org/pandas-docs/stable/contributing.html#combining-commits

oceanEngineer:
Sorry about all the confusion squashing the commits. Hope I didn't stomp on anyone else's work. Thanks for bearing with me, and for the help along the way.

oceanEngineer:
Hope that does it. Thanks for your help.

shoyer:
Looks great -- thanks!

Next time, it's better to give your commits are more descriptive title, and keep GH issue numbers to the commit comments.

PR #744, opened 2016-02-03
Better error message when version.py is missing

shoyer:
xref: discussion in https://github.com/pydata/xarray/pull/741

### DIFF
--- a/xarray/__init__.py
+++ b/xarray/__init__.py
@@ -8,6 +8,12 @@ from .core.options import set_options
 from .backends.api import open_dataset, open_mfdataset, save_mfdataset
 from .conventions import decode_cf
 
-from .version import version as __version__
+try:
+    from .version import version as __version__
+except ImportError:  # pragma: no cover
+    raise ImportError('xarray not properly installed. If you are running from '
+                      'from the source directory, please instead create a new '
+                      'virtual environment (using conda or virtualenv) and '
+                      'then install it in-place by running: pip install -e .')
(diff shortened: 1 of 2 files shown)


MaximilianR:
I'm fairly ignorant when it comes to these build / setup issues.
Generally, from my perspective, I don't want to install master on my system (it's important that has 0.7.0), I just want to use it in its cloned directory. Is there an equivalent command to pandas' `python setup.py build_ext --inplace` that would work for xarray?

shoyer:
I recommend using virtual environments for this purpose, either with pip or conda. I'll add a note to the error message.

This versioning system used to be how pandas did things, but then it switched to versioneer ~6 month ago. That might be worth a shot instead. I think it falls back more gracefully in such cases.

xarray/__init__.py
```diff
@@ -8,6 +8,12 @@
 from .backends.api import open_dataset, open_mfdataset, save_mfdataset
 from .conventions import decode_cf
 
-from .version import version as __version__
+try:
+    from .version import version as __version__
+except ImportError:  # pragma: no cover
+    raise ImportError('xarray not properly installed. If you are running from '
+                      'from the source directory, please instead create a new '
+                      'virtual environment (using conda or virtualenv) and '
```
fmaussion: A bit too late, but there are two "from"s

shoyer: thanks, will fix

Issue #745, opened 2016-02-04
Transposing a Dataset causes PeriodIndex to lose its type 


MaximilianR:
Note the different types in the final two outputs

```python

periods = pd.period_range(start='2000', freq='B', periods=6000)
np_points = np.random.rand(6000,20)

period_array = xr.DataArray(np_points)
period_array['dim_0']=periods
period_array

Out[87]:
<xarray.DataArray (dim_0: 6000, dim_1: 20)>
array([[ 0.36453381,  0.65939328,  0.65642922, ...,  0.66950028,
         0.03690508,  0.85428786],
       [ 0.06142194,  0.6391667 ,  0.93972185, ...,  0.26272683,
         0.17446443,  0.05473016],
       [ 0.06888458,  0.88798184,  0.7004805 , ...,  0.54081794,
         0.11690242,  0.71239621],
       ..., 
       [ 0.46578244,  0.47498626,  0.11854992, ...,  0.73731368,
         0.44784859,  0.24722402],
       [ 0.02694025,  0.26113875,  0.27635559, ...,  0.6397514 ,
         0.94297744,  0.50903873],
       [ 0.2302912 ,  0.5255501 ,  0.98877204, ...,  0.51659326,
         0.5516555 ,  0.10720623]])
Coordinates:
  * dim_0    (dim_0) object 2000-01-03 2000-01-04 2000-01-05 2000-01-06 ...
  * dim_1    (dim_1) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19


period_array.dim_0.to_index()
Out[93]:
PeriodIndex(['2000-01-03', '2000-01-04', '2000-01-05', '2000-01-06',
             '2000-01-07', '2000-01-10', '2000-01-11', '2000-01-12',
             '2000-01-13', '2000-01-14',
             ...
             '2022-12-19', '2022-12-20', '2022-12-21', '2022-12-22',
             '2022-12-23', '2022-12-26', '2022-12-27', '2022-12-28',
             '2022-12-29', '2022-12-30'],
            dtype='int64', name=u'dim_0', length=6000, freq='B')
In [95]:


period_array.transpose('dim_0').dim_0.to_index()
Out[95]:
Index([2000-01-03, 2000-01-04, 2000-01-05, 2000-01-06, 2000-01-07, 2000-01-10,
       2000-01-11, 2000-01-12, 2000-01-13, 2000-01-14,
       ...
[...]

PR #746, opened 2016-02-04
BUG: Don't transpose variables of one dimension

MaximilianR:
Resolves https://github.com/pydata/xarray/issues/745

A bit of a 'convenient hack'. Let me know if you think there's a better way to do this

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -648,6 +648,8 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
         if len(dims) == 0:
             dims = self.dims[::-1]
         axes = self.get_axis_num(dims)
+        if len(dims) < 2: # no need to transpose if only one dimension
+            return self.copy(deep=False)
         data = ops.transpose(self.data, axes)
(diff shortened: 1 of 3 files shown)


xarray/core/variable.py
```diff
@@ -647,6 +647,9 @@ def transpose(self, *dims):
         """
         if len(dims) == 0:
             dims = self.dims[::-1]
+        # no need to transpose if only one dimension
+        if len(dims) == 1:
+            return self
```
shoyer: It's OK to keep the data the same, but we shouldn't return the same `Variable` object. I would use `self.copy(deep=False)` instead. Also, let's check *after* we verify that `dims` is valid (e.g., move the conditional one line below). Probably a little safer to use `if self.ndim < 2` as the check instead, too.

MaximilianR:
Great, @shoyer, updated

MaximilianR:
Failing on this: https://github.com/pydata/pandas/issues/12235, will push again when resolved

MaximilianR:
@shoyer green

PR #749, opened 2016-02-08
Retain label type in .to_dataset

MaximilianR:
Not sure if this was a deliberate choice (maybe so integer indexing wasn't disrupted?). This allows objects as dataset keys / labels.

Let me know your thoughts and I'll add a what's new.

(I know I have a few PRs dangling - apologies - will go back and clean them up soon)

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -26,7 +26,9 @@ Bug fixes
 ~~~~~~~~~
 
 - Single dimension variables no longer transpose as part of a broader ``.transpose``. This behavior
-  was causing ``pandas.PeriodIndex`` dimensions to lose their type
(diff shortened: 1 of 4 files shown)


jhamman:
@shoyer - I don't remember which issue it was but I seem to remember implementing this on purpose. I think we were having trouble serializing and round-tripping with non string type labels.  Ring a bell?

shoyer:
I recall discussing this and coming to the conclusion that any hashable object should be valid as dataset keys. Obviously, you need strings if you want to use standard serialization backends, but otherwise it's better to keep things flexible.

shoyer:
@MaximilianR could please add a quick what's new note? otherwise this looks good to me

MaximilianR:
Added @shoyer

Issue #740, opened 2016-02-02
Support pandas.Series in the Dataset constructor


shoyer:
Follow up on #677

cc @MaximilianR

MaximilianR:
I think this is why it doesn't currently work: http://stackoverflow.com/questions/35442924/how-does-pythons-dict-constructor-handle-mappings

Tangentially related to: https://github.com/pydata/pandas/issues/12056 - i.e. from the pandas side, should Series be Mappings?

We could add a separate path for `Series` in the dataset constructor, or even monkey patch `ChainMap`. Neither seems that elegant

shoyer:
I would be perfectly happy with adding something like this to the Dataset constructor:
```
if isinstance(data_vars, pd.Series):
    data_vars = dict(data_vars)
```

PR #764, opened 2016-02-17
Coerce Series to dict in Dataset constructor

MaximilianR:
Resolves https://github.com/pydata/xarray/issues/740

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -253,6 +253,11 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         """
         _assert_empty([k for k in vars if k in coords],
                       'redundant variables and coordinates: %s')
+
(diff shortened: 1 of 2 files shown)


xarray/core/dataset.py
```diff
@@ -253,6 +253,11 @@ def _set_init_vars_and_dims(self, vars, coords, compat):
         """
         _assert_empty([k for k in vars if k in coords],
                       'redundant variables and coordinates: %s')
+
+        # Need to coerce Series to a dict as it doesn't implement __iter__ like a mapping
+        # GH470
+        vars = dict(vars) if isinstance(vars, pd.Series) else vars
```
shoyer: Can we do the two-liner instead? It's an extra line but less text -- and is clearer IMO

shoyer: Eh... this is good enough.

MaximilianR:
Would have been happy to do the two line version - agree it's clearer
Cheers for merging though

PR #668, opened 2015-12-02
Feature/rolling

jhamman:
This is an initial take at the rolling aggregation object and methods in xray.  This PR implements:

- A new `Rolling` class available in `DataArray` objects:
    
    ```Python
    rolling_obj = da.rolling(time=7)
   ```
- `bottleneck.move_*` functions are wrapped and are available in the following manor:

    ```Python
    rolling_obj.mean()
    ```
- generic reduce method

    ```Python
    from numpy import nanpercentile
    rolling_obj.reduce(nanpercentile, q=5)
    ```
- iterating through the rolling object:

    ```Python
     for label, da_window in rolling_obj:
         # da_window is a view of da
    ````

TODO:

- [x] Documentation
- [x] Cleanup `_setup_windows`
- [x] Inject bottleneck methods in some generic way
- [x] ~~Possibly create a `Window` object to hold windowed `DataArray`s~~

closes #641 #130
xref pydata/pandas#11603, pydata/pandas#11704
cc @shoyer @bartnijssen

PR #772, opened 2016-02-20
update rolling doc string

jhamman:
minor update of `rolling` doc string.  Missed this update after @shoyer's last review.

xref: #668

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -343,17 +343,21 @@ class BaseDataObject(AttrAccessMixin):
             group = self[group]
         return self.groupby_cls(self, group, squeeze=squeeze)
 
-    def rolling(self, min_periods=None, center=False, **kwarg):
-        """Returns a Rolling object for performing moving window operations.
+    def rolling(self, min_periods=None, center=False, **windows):
+        """
+        Moving window object.
 
         Parameters
         ----------
+        obj : Dataset or DataArray
+            Object to window.
         min_periods : int, default None
             Minimum number of observations in window required to have a value
-            (otherwise result is NA).
+            (otherwise result is NA). The default, None, is equivalent to
+            setting min_periods equal to the size of the window.
         center : boolean, default False
             Set the labels at the center of the window.
-        kwarg : dim=window
+        **windows : dim=window
             dim : str
                 Name of the dimension to create the rolling iterator
                 along (e.g., `time`).
@@ -366,7 +370,7 @@ class BaseDataObject(AttrAccessMixin):
         """
 
         return self.rolling_cls(self, min_periods=min_periods,
-                                center=center, **kwarg)
+                                center=center, **windows)
 
     def resample(self, freq, dim, how='mean', skipna=None, closed=None,
                  label=None, base=0):


xarray/core/common.py
```diff
@@ -343,17 +343,21 @@ def groupby(self, group, squeeze=True):
...
 
-    def rolling(self, min_periods=None, center=False, **kwarg):
-        """Returns a Rolling object for performing moving window operations.
+    def rolling(self, min_periods=None, center=False, **windows):
+        """
+        Moving window object.
 
         Parameters
         ----------
+        obj : Dataset or DataArray
```
shoyer: We usually don't include the object itself as a parameter on method docstrings.

jhamman: rrrrrrrr... I didn't mean to copy this line. This is needed in the `Rolling.__init__` docstring so that's how it got here.  I'll issue another PR to fix this.

xarray/core/common.py
```diff
@@ -343,17 +343,21 @@ def groupby(self, group, squeeze=True):
...
+            Object to window.
         min_periods : int, default None
             Minimum number of observations in window required to have a value
-            (otherwise result is NA).
+            (otherwise result is NA). The default, None, is equivalent to
+            setting min_periods equal to the size of the window.
```
jhamman: I think bottleneck and pandas differ on how they handle the `min_periods` argument.  

Bottleneck:

>min_count: {int, None}, optional :
>
>    If the number of non-NaN values in a window is less than min_count, then a value of NaN is assigned to the window. By default min_count is None, which is equivalent to setting min_count equal to window.

Pandas doesn't say in its doc string: 

>min_periods : int, default None
>
>    Minimum number of observations in window required to have a value (otherwise result is NA).

So, comparing their behavior, we see they both set `min_periods` to the size of the window.

```Python
In [1]: import pandas as pd
In [2]: s = pd.Series(range(8))    
In [3]: pd.rolling_mean(s, 3)
Out[3]: 
0   NaN
1   NaN
2     1
3     2
4     3
5     4
6     5
7     6
dtype: float64
In [4]: import bottleneck as bn
In [6]: bn.move_mean(s, 3)
Out[6]: array([ nan,  nan,   1.,   2.,   3.,   4.,   5.,   6.])

Comment (shoyer) [inline xarray/core/common.py]: Something seems to be out of sync for NaN handling, though:
```
In [27]: d = xr.DataArray([0, np.nan, 1, 2, np.nan, 3, 4, 5, np.nan, 6, 7], dims='x')

In [28]: d.rolling(x=2).mean()
Out[28]:
<xarray.DataArray (x: 11)>
array([ nan,  0. ,  1. ,  1.5,  2. ,  3. ,  3.5,  4.5,  5. ,  6. ,  6.5])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9 10

# using the pandas RC for v0.18
In [29]: d.to_series().rolling(2).mean().to_xarray()
Out[29]:
<xarray.DataArray (x: 11)>
array([ nan,  nan,  nan,  1.5,  nan,  nan,  3.5,  4.5,  nan,  nan,  6.5])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9 10
```

Comment (jhamman) [inline xarray/core/common.py]: hmmm, you must not have bottleneck in your environment because I get:

```Python
In [6]: d.rolling(x=2).mean()
Out[6]: 
<xarray.DataArray (x: 11)>
array([ nan,  nan,  nan,  1.5,  nan,  nan,  3.5,  4.5,  nan,  nan,  6.5])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9 10
```

I suppose we should open an issue on this. I guess we need to use the non-nan safe numpy methods in `Rolling.reduce` to get the same behavior. We'll have to come up with a solution to get this to work in a vectorized manor.

shoyer: Yep, no bottleneck on my work machine.

shoyer: made a new issue :https://github.com/pydata/xarray/issues/776

PR #773, opened 2016-02-20
IDEs resolves ops

### LINKED ISSUE #771: PyCharm warnings on numeric operations
Currently PyCharm warns if a numeric operation is performed on an xarray object. We could suppress these by:
- Adding `__add__ = NotImplemented` onto the `BaseDataObject` (and all the other ops from `CMP_BINARY_OPS` & `NUM_BINARY_OPS`), so IDEs know about them
- Adding a set of skeletons here: https://github.com/JetBrains/python-skeletons. I haven't tried this, and the readme looks out of date. But it does seem to work for numpy objects
- Wait for IDEs to be able to recognize the monkey patching that is done here (not my preference)

MaximilianR:
resolves #771

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -3,7 +3,7 @@ import pandas as pd
 
 from .pycompat import basestring, iteritems, suppress, dask_array_type
 from . import formatting
-from .utils import SortedKeysDict
+from .utils import SortedKeysDict, not_implemented
 
 
 class ImplementsArrayReduce(object):
@@ -476,6 +476,12 @@ class BaseDataObject(AttrAccessMixin):
           * x        (x) int64 0 1 2 3 4
         """
         return self._where(cond)
+
+    # this has no runtime function - these are listed so IDEs know these methods
+    # are defined and don't warn on these operations
+    __lt__ = __le__ =__ge__ = __gt__ = __add__ = __sub__ = __mul__ = \
+    __truediv__ = __floordiv__ = __mod__ = __pow__ = __and__  = __xor__ = \
+    __or__ = __div__ = __eq__ = __ne__ = not_implemented
 
 
 def squeeze(xarray_obj, dims, dim=None):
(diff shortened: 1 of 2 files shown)


xarray/core/common.py
```diff
@@ -477,6 +477,10 @@ def where(self, cond):
         """
         return self._where(cond)
 
+    __lt__ = __le__ =__ge__ = __gt__ = __add__ = __sub__ = __mul__ = \
+    __truediv__ = __floordiv__ = __mod__ = __pow__ = __and__  = __xor__ = \
+    __or__ = __div__ = not_implemented
```
MaximilianR: Parentheses for line continuation didn't seem to work here

xarray/core/utils.py
```diff
@@ -431,3 +431,7 @@ def hashable(v):
     except TypeError:
         return False
     return True
```
MaximilianR: FWIW all tests pass if this function is replaced with `isinstance(v, Hashable)` (but I don't think worth changing given the risks, even if it seems to work)

xarray/core/utils.py
```diff
@@ -431,3 +431,7 @@ def hashable(v):
...
+
+
+def not_implemented(*args, **kwargs):
+    return NotImplemented
```
shoyer: missing a new line here at the end of the file

xarray/core/common.py
```diff
@@ -477,6 +477,10 @@ def where(self, cond):
         """
         return self._where(cond)
 
+    __lt__ = __le__ =__ge__ = __gt__ = __add__ = __sub__ = __mul__ = \
```
shoyer: If we can't add tests, let's at least add a comment noting why we added this (otherwise it's likely to do away when someone refactors)

xarray/core/utils.py
MaximilianR: fixed (I think)

MaximilianR:
done @shoyer

xarray/core/utils.py
shoyer: yep -- you can notice the red marker at the end of the line if the final new line is missing

PR #777, opened 2016-02-20
minor refactor of Rolling.reduce to handle nans

### LINKED ISSUE #776: The result of rolling aggregations in the presence of NaNs depends changes if bottleneck is installed
Without bottleneck, we are inconsistent with pandas:
```
In [27]: d = xr.DataArray([0, np.nan, 1, 2, np.nan, 3, 4, 5, np.nan, 6, 7], dims='x')

In [28]: d.rolling(x=2).mean()
Out[28]:
<xarray.DataArray (x: 11)>
array([ nan,  0. ,  1. ,  1.5,  2. ,  3. ,  3.5,  4.5,  5. ,  6. ,  6.5])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9 10

# using the pandas RC for v0.18
In [29]: d.to_series().rolling(2).mean().to_xarray()
Out[29]:
<xarray.DataArray (x: 11)>
array([ nan,  nan,  nan,  1.5,  nan,  nan,  3.5,  4.5,  nan,  nan,  6.5])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9 10
```
xref https://github.com/pydata/xarray/pull/772#discussion_r53556663

jhamman:
fixes #776 

~~Still need to add a regression test~~

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -475,7 +475,11 @@ def inject_all_ops_and_reduce_methods(cls, priority=50, array_only=True):
 def inject_bottleneck_rolling_methods(cls):
     # standard numpy reduce methods
     for name in NAN_REDUCE_METHODS:
-        f = getattr(np, name)
+        try:
+            f = getattr(np, 'nan' + name)
+        except AttributeError:
+            # some versions of numpy dont have all the nan-safe methods
+            f = getattr(np, name)
         func = cls._reduce_method(f)
         func.__name__ = name
         func.__doc__ = 'todo'
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -475,11 +475,14 @@ def inject_all_ops_and_reduce_methods(cls, priority=50, array_only=True):
 def inject_bottleneck_rolling_methods(cls):
     # standard numpy reduce methods
     for name in NAN_REDUCE_METHODS:
+        func_name = 'nan' + name
         try:
-            f = getattr(np, 'nan' + name)
+            f = getattr(np, func_name)
         except AttributeError:
-            # some versions of numpy dont have all the nan-safe methods
-            f = getattr(np, name)
+            if func_name == 'nanprod':
+                f = npcompat.nanprod
+            else:
+                raise AttributeError
         func = cls._reduce_method(f)
         func.__name__ = name
         func.__doc__ = 'todo'
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -259,6 +259,21 @@ _REDUCE_DOCSTRING_TEMPLATE = \
             indicated dimension(s) removed.
         """
 
+_ROLLING_REDUCE_DOCSTRING_TEMPLATE = \
+        """Reduce this DataArrayRolling's data windows by applying `{name}`
+        along its dimension.
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Additional keyword arguments passed on to `{name}`.
+
+        Returns
+        -------
+        reduced : DataArray
+            New DataArray object with `{name}` applied along its rolling dimnension.
+        """
+
 
 def count(data, axis=None):
     """Count the number of non-NA in this array along the given axis or axes
@@ -474,18 +489,11 @@ def inject_all_ops_and_reduce_methods(cls, priority=50, array_only=True):
 
 def inject_bottleneck_rolling_methods(cls):
     # standard numpy reduce methods
-    for name in NAN_REDUCE_METHODS:
-        func_name = 'nan' + name
-        try:
-            f = getattr(np, func_name)
-        except AttributeError:
-            if func_name == 'nanprod':
-                f = npcompat.nanprod
-            else:
-                raise AttributeError
+    methods = [(name, globals()[name]) for name in NAN_REDUCE_METHODS]
+    for name, f in methods:
         func = cls._reduce_method(f)
         func.__name__ = name
-        func.__doc__ = 'todo'
+        func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(name=func.__name__)
         setattr(cls, name, func)
 
     # bottleneck rolling methods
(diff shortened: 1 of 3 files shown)


jhamman:
@shoyer and others, this is ready for review...

xarray/core/rolling.py
```diff
@@ -162,9 +167,13 @@ def reduce(self, func, **kwargs):
             dims.remove(dims[self._axis_num])
             windows = [DataArray(window, dims=dims) for window in windows]
```
shoyer: Just a note -- I probably should have caught this earlier -- but it might be cleaner to use the `reduce` method on each window instead, e.g.,
```
windows = [window.reduce(func, dims=self.dims, **kwargs) for _, window in self]
```

That would let you skip this logic to recreate the DataArrays -- and would also generalize immediately to Dataset.

shoyer:
LGTM.

I have one comment on your old code, but you can handle that in another PR if you want to test it out.

xarray/core/rolling.py
jhamman: Good call.  That was much easier.

jhamman:
Thanks @shoyer - I'll merge on green.

jhamman:
Turns out using `DataArray.reduce` highlighted some remaining difficulties. I'll illustrate below but 1470b89 should have resolved everything 

Below I show the pre-1470b89 behavior of `pandas`, `xarray`, and `bottleneck` side-by-side on a 1-D array with nans in it.

```Python
In [1]: 
import numpy as np
import pandas as pd
import xarray as xr
try:
    import bottleneck as bn
except:
    bn = False

In [2]:
a = np.array([0, np.nan, 1, 2, np.nan, 3, 4, 5, np.nan, 6, 7])

sa = pd.Series(a)
da = xr.DataArray(a)

min_periods, window= 1, 2

In [3]: 
print('pandas rolling_sum  ', pd.rolling_sum(sa, window, min_periods=min_periods).values)
print('pandas rolling_apply', pd.rolling_apply(sa, window, np.nansum, min_periods=min_periods).values)
print('xarray rolling.sum  ', da.rolling(min_periods=min_periods, dim_0=window).sum().values)
print('xarray rolling.apply', da.rolling(min_periods=min_periods, dim_0=window).reduce(np.nansum).values)
if bn:
    print('bottleneck move_sum ', bn.move_sum(a, window, min_count=min_periods))

pandas rolling_sum   [  0.   0.   1.   3.   2.   3.   7.   9.   5.   6.  13.]
pandas rolling_apply [  0.   0.   1.   3.   2.   3.   7.   9.   5.   6.  13.]
xarray rolling.sum   [  0.   0.   1.   3.   2.   3.   7.   9.   5.   6.  13.]
xarray rolling.apply [  0.   0.   1.   3.   2.   3.   7.   9.   5.   6.  13.]
bottleneck move_sum  [  0.   0.   1.   3.   2.   3.   7.   9.   5.   6.  13.]
```
Great, its nice when we can all agree!

Now, let's move to an environment that doesn't have `bottleneck` and run the same set of commands:

```Python
In [3]: 
print('pandas rolling_sum  ', pd.rolling_sum(sa, window, min_periods=min_periods).values)
print('pandas rolling_apply', pd.rolling_apply(sa, window, np.nansum, min_periods=min_periods).values)
print('xarray rolling.sum  ', da.rolling(min_periods=min_periods, dim_0=window).sum().values)
print('xarray rolling.apply', da.rolling(min_periods=min_periods, dim_0=window).reduce(np.nansum).values)
if bn:
    print('bottleneck move_sum ', bn.move_sum(a, window, min_count=min_periods))

('pandas rolling_sum  ', array([  0.,   0.,   1.,   3.,   2.,   3.,   7.,   9.,   5.,   6.,  13.]))
('pandas rolling_apply', array([  0.,   0.,   1.,   3.,   2.,   3.,   7.,   9.,   5.,   6.,  13.]))
('xarray rolling.sum  ', array([  0.,  nan,  nan,   3.,  nan,  nan,   7.,   9.,  nan,  nan,  13.]))
('xarray rolling.apply', array([  0.,   0.,   1.,   3.,   2.,   3.,   7.,   9.,   5.,   6.,  13.]))
```
So that isn't what we wanted to see. It turns out that we were not injecting the `nan`-save numpy methods so without bottleneck, we were falling back to plain-old `np.sum`.

xarray/core/ops.py
```diff
@@ -475,7 +475,11 @@ def inject_all_ops_and_reduce_methods(cls, priority=50, array_only=True):
 def inject_bottleneck_rolling_methods(cls):
     # standard numpy reduce methods
     for name in NAN_REDUCE_METHODS:
-        f = getattr(np, name)
+        try:
+            f = getattr(np, 'nan' + name)
+        except AttributeError:
+            # some versions of numpy dont have all the nan-safe methods
+            f = getattr(np, name)
```
jhamman: @shoyer - do you have any problem with this bit of new code?  You've seen everything else.

shoyer: Which versions of numpy are missing these methods? in general I would rather give errors than silently yield inconsistent results.

jhamman: `np.nanprod` was added in 1.10.0.  That was the one that was giving me problems.  `prod` is listed in `NAN_REDUCE_METHODS` above.

shoyer: We have a back port of nanprod in npcompat.py

jhamman: I didn't know that.  Updated my PR accordingly.

xarray/core/ops.py
```diff
@@ -475,7 +475,14 @@ def inject_all_ops_and_reduce_methods(cls, priority=50, array_only=True):
 def inject_bottleneck_rolling_methods(cls):
     # standard numpy reduce methods
     for name in NAN_REDUCE_METHODS:
-        f = getattr(np, name)
+        func_name = 'nan' + name
```
shoyer: Can we reuse these functions instead of writing this all again?
https://github.com/jhamman/xarray/blob/rolling_with_nans/xarray/core/ops.py#L293
https://github.com/jhamman/xarray/blob/rolling_with_nans/xarray/core/ops.py#L389

Yes, I know it's heavy on the magic :)

jhamman: done.

jhamman:
@shoyer - I went ahead and used the `_create_nan_agg_method` version of the nan-safe methods.  I also added a simple docstring to the rolling reduce methods.  I think we're good to go here.

shoyer:
Yep, agreed! Feel free to merge.

PR #778, opened 2016-02-29
fix value error in data-structures doc

fmaussion:
This should fix the error in the docs here: http://xarray.pydata.org/en/latest/data-structures.html

It is a detail, but it occurred to me that the examples of this section don't use the (space, time) dimension order (e.g. `dims=['time', 'space']`). Most of the climate datasets I know of are using the numpy (lat, lon, time) dim order. Is it intentional?


shoyer:
Thanks for catching this!

There's no particular reason why I wrote the docs with (time, space). I would be equally happy with (space, time) and the later does make a bit more sense...

PR #782, opened 2016-03-03
dont infer interval breaks in pcolormesh when ax is a cartopy axis

### LINKED ISSUE #781: Don't infer x/y coordinates interval breaks for cartopy plot axes
The `DataArray.plot.pcolormesh()` method modifies the x/y coordinates of its plots.  I'm finding that, at least for custom `cartopy` projections, the offset applied [here](https://github.com/pydata/xarray/blob/master/xarray/plot/plot.py#L543-L544) causes some real issues downstream.  

@clarkfitzg - Do you see any problem with treating the x/y offset in the [same way](https://github.com/pydata/xarray/blob/master/xarray/plot/plot.py#L550-L553) as the axis limits?

jhamman:
This PR chooses not to infer interval breaks in pcolormesh if the axis is a Cartopy axis. Below is an example of a plot using the new behavior - compare to https://github.com/pydata/xarray/issues/781#issuecomment-191595449.
xarray_new

I'll come up with a few tests here.

closes #781.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -53,6 +53,11 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
+- Fixed an issue where plots using pcolormesh and Cartopy axes were being distorted
+  by the inference of the axis interval breaks. This change chooses not to modify
+  the coordinate variables when the axes have the attribute ``projection``, allowing
+  Cartopy to handle the extent of pcolormesh plots (:issue:`781`).
+
 .. _whats-new.0.7.1:
 
 v0.7.1 (16 February 2016)
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -540,8 +540,10 @@ def pcolormesh(x, y, z, ax, **kwargs):
 
     Wraps matplotlib.pyplot.pcolormesh
     """
-    x = _infer_interval_breaks(x)
-    y = _infer_interval_breaks(y)
+
+    if not hasattr(ax, 'projection'):
+        x = _infer_interval_breaks(x)
+        y = _infer_interval_breaks(y)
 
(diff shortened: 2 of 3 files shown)


ocefpaf:
Bare in mind that some netCDF4 files do have the bounds data and those should be used as "breaks" when available.

BTW I'd rather not have the `_infer_interval_breaks` at all, not sure if assuming linear at the mid points are generic enough.

jhamman:
> BTW I'd rather not have the _infer_interval_breaks at all.

Me too, but I don't mind doing it when we're not projecting the data.

ocefpaf:
What if there are bounds in the file and the data is regularly spaced? I consider the current behavior a guess, and guessing should be an active user choic, not automatic.

jhamman:
Short of adding cartopy to our continuous integration setup, does anyone have any ideas on how to test this change more robustly than I have shown here? I want to avoid testing the behavior of the either `matplotlib` or `Cartopy` and since in this case, we're choosing not to modify their behavior, I'm not sure we need to test further.

jhamman:
any more comments here?

shoyer:
This PR looks good to me.

Regarding the larger issue of not inferring bounds, I guess I would be on board with that change, too. The original idea was to make pcolormesh and imshow consistent (other than the underlying plotting engine) and to always plot that last row/column of data, but maybe it's better to force users to be explicit. At the very least, it is clearly a bad idea to adjust bounds if they are not equally spaced (but I agree that guessing is bad default behavior).

jhamman:
So do we want to remove `_infer_interval_breaks` all together in this PR?  I'm leaning toward merging this as is and following up with another PR to handle the unevenly spaced coordinates in the future.

shoyer:
Works for me

Issue #788, opened 2016-03-10
Composite coordinate plot only works in one orientation


jamesp:
When I create a composite coordinate i.e. `d.coords['c'] = d.a * d.b` and then try and plot a 2d plot of the data, it only works when the composite index is on the y axis.  When on the x-axis it fails with 

`TypeError: Dimensions of C (3, 5) are incompatible with X (4) and/or Y (6)`

ipynb notebook (and pdf version of same notebook) attached that recreate the error

[xarray composite coord plot.pdf](https://github.com/pydata/xarray/files/167044/xarray.composite.coord.plot.pdf)
[xarray composite coord plot.ipynb.zip](https://github.com/pydata/xarray/files/167045/xarray.composite.coord.plot.ipynb.zip)

fmaussion:
Just out of curiosity, what is the use case here? In your example the coordinate `rtime` has a shape `(time: 3, r: 3)` ant it is not linked to any dimension of the variable that you are plotting. What do you expect the plot to do?

jamesp:
my use case is that `rtime` is kind of like normalised time. `rtime` is implicitly linked to temperature via `time` and `r`, no?

I'm running many simulations with varying rotation period `r`.  `time` is just clock time in seconds; what I want to plot is time in terms of revolutions i.e. time/r.

It works just fine with `rtime` along the y-axis, but I would like it along the x and there it fails as the length of the coordinate seems to have changed.

fmaussion:
I'm not an xarray developer so use my answer with caution, but: 

> my use case is that rtime is kind of like normalised time. rtime is implicitly linked to temperature via > time and r, no?

It is not "linked" in the sense that this new coordinate has no special meaning for indexing and plotting methods (they don't have the `*`, as explained in the [documentation](http://xarray.pydata.org/en/stable/data-structures.html#coordinates)). 

> It works just fine with rtime along the y-axis, but I would like it along the x and there it fails as the 
> length of the coordinate seems to have changed.

It works, but by "chance". The variable you are plotting has dimensions `(time: 3, x: 5)`. If you stick with the original dimensions, xarray will now what to do, i.e.: `.plot(x='time', y='x')` will make a plot rotated by 90°.

When you do `.plot(x='x', y='rtime')`, xarray doesn't know which dimension is meant by `rtime` and falls back to the default behavior, which is to plot the first dimension along the y axis. However, it does use your provided coordinates to scale the axis (this works because `time` and `rtime` share one dimension length). 

[...]

PR #794, opened 2016-03-12
fix: transpose non-linked coordinates in 2d plots

fmaussion:
Should fix https://github.com/pydata/xarray/issues/788


fmaussion:
(I don't think my PR has something to do with the py2.6 test failure)

jhamman:
you are right.  The failure is on master as well and seems to be coming from a dask change.

shoyer:
This looks great -- thanks!

Could you please add a brief note on the bug fix to the What's New page?

fmaussion:
I added something to the What's New page and squashed the commits. I didn't really know how to describe the problem, I hope it's ok like this. Don't hesitate to correct me if needed!

shoyer:
Looks good to me -- thanks!

Issue #793, opened 2016-03-11
dask.async.RuntimeError: NetCDF: HDF error on xarray to_netcdf


pwolfram:
Dask appears to be failing on serialization following a ds.to_netcdef() via a NETCDF: HDF error.  
Excerpted error below:
```
Traceback (most recent call last):
  File "reduce_dispersion_file.py", line 40, in <module>
    if __name__ == "__main__":
  File "reduce_dispersion_file.py", line 36, in reduce_dispersion_file
    with timeit_context('output to disk'):
  File "/users/pwolfram/envs/LIGHT_analysis/lib/python2.7/site-packages/xarray/core/dataset.py", line 791, in to_netcdf
    engine=engine, encoding=encoding)
  File "/users/pwolfram/envs/LIGHT_analysis/lib/python2.7/site-packages/xarray/backends/api.py", line 356, in to_netcdf
    dataset.dump_to_store(store, sync=sync, encoding=encoding)
  File "/users/pwolfram/envs/LIGHT_analysis/lib/python2.7/site-packages/xarray/core/dataset.py", line 739, in dump_to_store
    store.sync()
  File "/users/pwolfram/envs/LIGHT_analysis/lib/python2.7/site-packages/xarray/backends/netCDF4_.py", line 283, in sync
    super(NetCDF4DataStore, self).sync()
  File "/users/pwolfram/envs/LIGHT_analysis/lib/python2.7/site-packages/xarray/backends/common.py", line 186, in sync
    self.writer.sync()
  File "/users/pwolfram/envs/LIGHT_analysis/lib/python2.7/site-packages/xarray/backends/common.py", line 165, in sync
    da.store(self.sources, self.targets)
  File "/users/pwolfram/lib/python2.7/site-packages/dask/array/core.py", line 712, in store
    Array._get(dsk, keys, **kwargs)
  File "/users/pwolfram/lib/python2.7/site-packages/dask/base.py", line 43, in _get
    return get(dsk2, keys, **kwargs)
  File "/users/pwolfram/lib/python2.7/site-packages/dask/threaded.py", line 57, in get 
    **kwargs)
  File "/users/pwolfram/lib/python2.7/site-packages/dask/async.py", line 481, in get_async
    raise(remote_exception(res, tb))
dask.async.RuntimeError: NetCDF: HDF error

Traceback
---------
  File "/users/pwolfram/lib/python2.7/site-packages/dask/async.py", line 264, in execute_task
    result = _execute_task(task, data)
[...]

PR #800, opened 2016-03-22
Adds dask lock capability for backend writes

pwolfram:
This fixes an error on an asynchronous write for `to_netcdf`
resulting in an `dask.async.RuntimeError: NetCDF: HDF error`

Resolves issue https://github.com/pydata/xarray/issues/793
following dask improvement at https://github.com/dask/dask/pull/1053
following advice of @shoyer.

### DIFF
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -3,6 +3,7 @@ import itertools
 import logging
 import time
 import traceback
+import threading
 from collections import Mapping
 
 from ..conventions import cf_encoder
@@ -162,7 +163,11 @@ class ArrayWriter(object):
     def sync(self):
         if self.sources:
             import dask.array as da
-            da.store(self.sources, self.targets)
+            import dask
+            if dask.__version__ > '0.8.1':
+                da.store(self.sources, self.targets, lock=threading.Lock())
+            else:
+                da.store(self.sources, self.targets)
             self.sources = []
             self.targets = []
 


xarray/backends/common.py
```diff
@@ -161,8 +161,12 @@ def add(self, source, target):
 
     def sync(self):
         if self.sources:
-            import dask.array as da
-            da.store(self.sources, self.targets)
+            import dask, dask.array as da
```
shoyer: Please keep each import to its own line

xarray/backends/common.py
```diff
@@ -161,8 +161,12 @@ def add(self, source, target):
...
+            import threading
```
shoyer: We can move `import threading` to the top of the file -- we only have to import dask here because it isn't a required dependency

shoyer:
This is going to be hard to test, but honestly I don't think that's essential given the simplicity of this code -- the complexity is buried upstream in dask, which does have comprehensive tests.

If you do want to write a test for this, you could mock out a fake DataStore similar to what we wrote for the dask tests and turn on multi-threading. But I think that's probably overkill here.

pwolfram:
Thanks @shoyer for the review.  I agree regarding this being simple enough to not require testing so this PR is probably ready to go.

shoyer:
Thanks @pwolfram !

Issue #767, opened 2016-02-17
MultiIndex and data selection


benbovy:
[Edited for more clarity]

First of all, I find the MultiIndex very useful and I'm looking forward to see the TODOs in #719  implemented in the next releases, especially the three first ones in the list!

Apart from these issues, I think that some other aspects may be improved, notably regarding data selection. Or maybe I've not correctly understood how to deal with multi-index and data selection...

To illustrate this, I use some fake spectral data with two discontinuous bands of different length / resolution:

```
In [1]: import pandas as pd

In [2]: import xarray as xr

In [3]: band = np.array(['foo', 'foo', 'bar', 'bar', 'bar'])

In [4]: wavenumber = np.array([4050.2, 4050.3, 4100.1, 4100.3, 4100.5])

In [5]: spectrum = np.array([1.7e-4, 1.4e-4, 1.2e-4, 1.0e-4, 8.5e-5])

In [6]: s = pd.Series(spectrum, index=[band, wavenumber])

In [7]: s.index.names = ('band', 'wavenumber')

In [8]: da = xr.DataArray(s, dims='band_wavenumber')

In [9]: da
Out[9]:
<xarray.DataArray (band_wavenumber: 5)>
array([  1.70000000e-04,   1.40000000e-04,   1.20000000e-04,
         1.00000000e-04,   8.50000000e-05])
Coordinates:
  * band_wavenumber  (band_wavenumber) object ('foo', 4050.2) ...
```

I extract the band 'bar' using `sel`:

```
In [10]: da_bar = da.sel(band_wavenumber='bar')

In [11]: da_bar
Out[11]:
<xarray.DataArray (band_wavenumber: 3)>
array([  1.20000000e-04,   1.00000000e-04,   8.50000000e-05])
Coordinates:
  * band_wavenumber  (band_wavenumber) object ('bar', 4100.1) ...
```

It selects the data the way I want, although using the dimension name is confusing in this case. It would be nice if we can also use the `MultiIndex` names as arguments of the `sel` method, even though I don't know if it is easy to implement.

Futhermore, `da_bar` still has the 'band_wavenumber' dimension and the 'band' index-level, but it is not very useful anymore. Ideally, I'd rather like to obtain a `DataArray` object with a 'wavenumber' dimension / coordinate and the 'bar' band name dropped from the multi-index, i.e., something would require automatic index-level removal and/or automatic unstack when selecting data.

[...]

Issue #719, opened 2016-01-17
Follow-ups on MultIndex support


shoyer:
xref #702
- [ ] Serialization to NetCDF
- [x] Better repr, showing level names/dtypes?
- [x] Indexing a scalar at a particular level should drop that level from the MultiIndex (#767)
- [x] Make levels accessible as coordinate variables (e.g., `ds['time']` can pull out the `'time'` level of a multi-index)
- [x] Support indexing with levels, e.g., `ds.sel(time='2000-01')`.
- [x] ~~Make `isel_points`/`sel_points` return objects with a MultiIndex? (probably after the previous TODO, so we can preserve basic backwards compatibility)~~ (differed until we figure out #974)
- [x] Add `set_index`/`reset_index`/`swaplevel` to make it easier to create and manipulate multi-indexes

benbovy:
Thinking about serialization, a possible solution would be splitting the multi-index into separate coordinates for the same dimension, then assign some specific attributes (e.g., `xarray_idx_name`, `xarray_idx_level`) to each of these coordinates so that it is possible to further rebuild the multi-index.

More generally, a couple of methods acting on coordinates only can be a complement to stack/unstack methods (acting on both coordinates and dimensions). Merge/split xarray coordinates into/from a multi-index seem very straightforward to implement.

Similarly, a better repr would represent levels as sub-coordinates: 

```
<xarray.DataArray (dim_0: 3)>
array([0, 1, 2])
Coordinates:
  * dim_0    (dim_0) object
        level_0 (0) object 'foo' 'foo' 'bar'
        level_1 (1) int64 0 1 2
```

Any thoughts on this (I can start PRs)?

shoyer:
> Thinking about serialization, a possible solution would be splitting the multi-index into separate coordinates for the same dimension, then assign some specific attributes (e.g., xarray_idx_name, xarray_idx_level) to each of these coordinates so that it is possible to further rebuild the multi-index.

I think `index_level` would suffice for the serialization, unless we're interested in supporting serializing multiple multi-indexes along a dimension. I don't see much use for that, so I think it's fine to only support serializing mulit-indexes that are actually used as an index.

> More generally, a couple of methods acting on coordinates only can be a complement to stack/unstack methods (acting on both coordinates and dimensions). Merge/split xarray coordinates into/from a multi-index seem very straightforward to implement.

[...]

PR #802, opened 2016-03-24
Multi-index indexing

benbovy:
Follows #767.

This is incomplete (it still needs some tests and documentation updates), but it is working for both `Dataset` and `DataArray` objects. I also don't know if it is fully compatible with lazy indexing (Dask). 

Using the example from #767:

```
In [4]: da.sel(band_wavenumber={'band': 'foo'})
Out[4]:
<xarray.DataArray (wavenumber: 2)>
array([ 0.00017,  0.00014])
Coordinates:
  * wavenumber  (wavenumber) float64 4.05e+03 4.05e+03
```

As shown in this example, similarily to pandas, it automatically renames the dimension and assigns a new coordinate when the selection doesn't return a `pd.MultiIndex` (here it returns a `pd.FloatIndex`).

In some cases this behavior may be unwanted (??), so I added a `drop_level` keyword argument (if `False` it keeps the multi-index and doesn't change the dimension/coordinate names):

```
In [5]: da.sel(band_wavenumber={'band': 'foo'}, drop_level=False)
Out[5]:
<xarray.DataArray (band_wavenumber: 2)>
array([ 0.00017,  0.00014])
Coordinates:
  * band_wavenumber  (band_wavenumber) object ('foo', 4050.2) ('foo', 4050.3)
```

Note that it also works with `DataArray.loc`, but (for now) in that case it always returns the multi-index:

```
In [6]: da.loc[{'band_wavenumber': {'band': 'foo'}}]
Out[6]:
<xarray.DataArray (band_wavenumber: 2)>
array([ 0.00017,  0.00014])
Coordinates:
  * band_wavenumber  (band_wavenumber) object ('foo', 4050.2) ('foo', 4050.3)
```

This is however inconsistent with `Dataset.sel` and `Dataset.loc` that both apply `drop_level=True` by default, due to their different implementation. Two solutions: (1) make `DataArray.loc` apply drop_level by default, or (2) use `drop_level=False` by default everywhere.

### DIFF
--- a/doc/data-structures.rst
+++ b/doc/data-structures.rst
@@ -115,11 +115,9 @@ If you create a ``DataArray`` by supplying a pandas
     df
     xr.DataArray(df)
 
-xarray does not (yet!) support labeling coordinate values with a
-:py:class:`pandas.MultiIndex` (see :issue:`164`).
-However, the alternate ``from_series`` constructor will automatically unpack
-any hierarchical indexes it encounters by expanding the series into a
-multi-dimensional array, as described in :doc:`pandas`.
+Xarray supports labeling coordinate values with a :py:class:`pandas.MultiIndex`.
+While it handles multi-indexes with unnamed levels, it is recommended that you
+explicitly set the names of the levels.
 
 DataArray properties
 ~~~~~~~~~~~~~~~~~~~~
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -316,14 +316,11 @@ pandas multi-index ; it automatically renames the dimension and replaces the
 coordinate when a single index is returned (level drop).
 
 Like pandas, it is also possible to use tuples of tuples, lists or slices
-(in that case xarray always returns the full multi-index):
+(for now xarray always returns the full multi-index in that case):
 
 .. ipython:: python
 
    da_midx.sel(x=(list('ab'), [0]))
-
-Indexing with dictionaries uses the ``MultiIndex.get_loc_level`` pandas method
-while indexing with nested tuples uses the ``MultiIndex.get_locs`` method.
 
 Multi-dimensional indexing
 --------------------------
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -315,6 +315,16 @@ As shown in the last example above, xarray handles partial selection on
 pandas multi-index ; it automatically renames the dimension and replaces the
 coordinate when a single index is returned (level drop).
 
+Like pandas, it is also possible to use tuples of tuples, lists or slices
+(in that case xarray always returns the full multi-index):
+
+.. ipython:: python
+
+   da_midx.sel(x=(list('ab'), [0]))
+
+Indexing with dictionaries uses the ``MultiIndex.get_loc_level`` pandas method
+while indexing with nested tuples uses the ``MultiIndex.get_locs`` method.
+
 Multi-dimensional indexing
 --------------------------
 
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -294,6 +294,27 @@ elements that are fully masked:
 
     arr2.where(arr2.y < 2, drop=True)
 
+.. _multi-index indexing:
+
+Multi-index indexing
+--------------------
+
+The ``loc`` and ``sel`` methods of ``Dataset`` and ``DataArray`` both accept
+dictionaries for indexing on multi-index dimensions:
+
+.. ipython:: python
+
+    idx = pd.MultiIndex.from_product([list('abc'), [0, 1]],
+                                     names=('one', 'two'))
+    da_midx = xr.DataArray(range(6), [('x', idx)])
+    da_midx
+    da_midx.sel(x={'one': 'a', 'two': 0})
+    da_midx.loc[{'one': 'a'}, ...]
+
+As shown in the last example above, xarray handles partial selection on
+pandas multi-index ; it automatically renames the dimension and replaces the
+coordinate when a single index is returned (level drop).
+
 Multi-dimensional indexing
 --------------------------
 
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -299,10 +299,10 @@ elements that are fully masked:
 Multi-level indexing
 --------------------
 
-Just like pandas, advanced indexing on multi-level indexes is also possible with
+Just like pandas, advanced indexing on multi-level indexes is possible with
 ``loc`` and ``sel``. You can slice a multi-index by providing multiple indexers,
 i.e., a tuple of slices, labels, list of labels, or any selector allowed by
-pandas (see :doc:`pandas`):
+pandas:
 
 .. ipython:: python
 
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -299,54 +299,45 @@ elements that are fully masked:
 Multi-level indexing
 --------------------
 
-The ``loc`` and ``sel`` methods of ``Dataset`` and ``DataArray`` both accept
-dictionaries for label-based indexing on multi-index dimensions:
+Just like pandas, advanced indexing on multi-level indexes is also possible with
+``loc`` and ``sel``. You can slice a multi-index by providing multiple indexers,
+i.e., a tuple of slices, labels, list of labels, or any selector allowed by
+pandas (see :doc:`pandas`):
 
 .. ipython:: python
 
-    idx = pd.MultiIndex.from_product([list('abc'), [0, 1]],
-                                     names=('one', 'two'))
-    da_midx = xr.DataArray(np.random.rand(6, 3),
-                           [('x', idx), ('y', range(3))])
-    da_midx
-    da_midx.sel(x={'one': 'a', 'two': 0})
-    da_midx.loc[{'one': 'a'}, ...]
+    midx = pd.MultiIndex.from_product([list('abc'), [0, 1]],
+                                      names=('one', 'two'))
+    mda = xr.DataArray(np.random.rand(6, 3),
+                       [('x', midx), ('y', range(3))])
+    mda
+    mda.sel(x=(list('ab'), [0]))
 
-As shown in the last example above, xarray handles partial selection on
-pandas multi-index ; it automatically renames the dimension and replaces the
-coordinate when a single index is returned (level drop).
-
-Like pandas, it is also possible to slice a multi-indexed dimension by providing
-a tuple of multiple indexers (i.e., slices, labels, list of labels, or any
-selector allowed by pandas). Note that for now xarray doesn't fully handle
-partial selection in that case (no level drop is done):
+You can also select multiple elements by providing a list of labels or tuples or
+a slice of tuples:
 
 .. ipython:: python
 
-   da_midx.sel(x=(list('ab'), [0]))
+   mda.sel(x=[('a', 0), ('b', 1)])
 
-Lists or slices of tuples can be used to select several combinations of
-multi-index labels:
-
-.. ipython:: python
-
-   da_midx.sel(x=[('a', 0), ('b', 1)])
-
-A single, flat tuple can be used to select a given combination of
-multi-index labels:
+Additionally, xarray supports dictionaries:
 
 .. ipython:: python
 
-   da_midx.sel(x=('a', 0))
+   mda.sel(x={'one': 'a', 'two': 0})
+   mda.loc[{'one': 'a'}, ...]
 
-Unlike pandas, xarray can't make the distinction between index levels and
+Like pandas, xarray handles partial selection on multi-index (level drop).
+As shown in the last example above, it also renames the dimension / coordinate
+when the multi-index is reduced to a single index.
+
+Unlike pandas, xarray does not guess whether you provide index levels or
 dimensions when using ``loc`` in some ambiguous cases. For example, for
-``da_midx.loc[{'one': 'a', 'two': 0}]`` and ``da_midx.loc['a', 0]`` xarray
+``mda.loc[{'one': 'a', 'two': 0}]`` and ``mda.loc['a', 0]`` xarray
 always interprets ('one', 'two') and ('a', 0) as the names and
 labels of the 1st and 2nd dimension, respectively. You must specify all
 dimensions or use the ellipsis in the ``loc`` specifier, e.g. in the example
-above, ``da_midx.loc[{'one': 'a', 'two': 0}, :]`` or
-``da_midx.loc[('a', 0), ...]``.
+above, ``mda.loc[{'one': 'a', 'two': 0}, :]`` or ``mda.loc[('a', 0), ...]``.
 
 Multi-dimensional indexing
 --------------------------
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -294,19 +294,20 @@ elements that are fully masked:
 
     arr2.where(arr2.y < 2, drop=True)
 
-.. _multi-index indexing:
+.. _multi-level indexing:
 
-Multi-index indexing
+Multi-level indexing
 --------------------
 
 The ``loc`` and ``sel`` methods of ``Dataset`` and ``DataArray`` both accept
-dictionaries for indexing on multi-index dimensions:
+dictionaries for label-based indexing on multi-index dimensions:
 
 .. ipython:: python
 
     idx = pd.MultiIndex.from_product([list('abc'), [0, 1]],
                                      names=('one', 'two'))
-    da_midx = xr.DataArray(range(6), [('x', idx)])
+    da_midx = xr.DataArray(np.random.rand(6, 3),
+                           [('x', idx), ('y', range(3))])
     da_midx
     da_midx.sel(x={'one': 'a', 'two': 0})
     da_midx.loc[{'one': 'a'}, ...]
@@ -315,12 +316,37 @@ As shown in the last example above, xarray handles partial selection on
 pandas multi-index ; it automatically renames the dimension and replaces the
 coordinate when a single index is returned (level drop).
 
-Like pandas, it is also possible to use tuples of tuples, lists or slices
-(for now xarray always returns the full multi-index in that case):
+Like pandas, it is also possible to slice a multi-indexed dimension by providing
+a tuple of multiple indexers (i.e., slices, labels, list of labels, or any
+selector allowed by pandas). Note that for now xarray doesn't fully handle
+partial selection in that case (no level drop is done):
 
 .. ipython:: python
 
    da_midx.sel(x=(list('ab'), [0]))
+
+Lists or slices of tuples can be used to select several combinations of
+multi-index labels:
+
+.. ipython:: python
+
+   da_midx.sel(x=[('a', 0), ('b', 1)])
+
+A single, flat tuple can be used to select a given combination of
+multi-index labels:
+
+.. ipython:: python
+
+   da_midx.sel(x=('a', 0))
+
+Unlike pandas, xarray can't make the distinction between index levels and
+dimensions when using ``loc`` in some ambiguous cases. For example, for
+``da_midx.loc[{'one': 'a', 'two': 0}]`` and ``da_midx.loc['a', 0]`` xarray
+always interprets ('one', 'two') and ('a', 0) as the names and
+labels of the 1st and 2nd dimension, respectively. You must specify all
+dimensions or use the ellipsis in the ``loc`` specifier, e.g. in the example
+above, ``da_midx.loc[{'one': 'a', 'two': 0}, :]`` or
+``da_midx.loc[('a', 0), ...]``.
 
 Multi-dimensional indexing
 --------------------------
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -26,6 +26,9 @@ Breaking changes
 ~~~~~~~~~~~~~~~~
 
 - Dropped support for Python 2.6 (:issue:`855`).
+- Indexing on multi-index now drop levels, which is consitent with pandas.
+  It also changes the name of the dimension / coordinate when the multi-index is
+  reduced to a single index.
 
 Enhancements
 ~~~~~~~~~~~~
@@ -39,9 +42,11 @@ Enhancements
   attributes are retained in the resampled object. By
   `Jeremy McGibbon <https://github.com/mcgibbon>`_.
 
-- DataArray and Dataset methods :py:meth:`sel` and :py:meth:`loc` now
-  accept dictionaries or nested tuples for indexing on multi-index dimensions.
-  By `Benoit Bovy <https://github.com/benbovy>`_.
+- Better multi-index support in DataArray and Dataset :py:meth:`sel` and
+  :py:meth:`loc` methods, which now behave more closely to pandas and which
+  also accept dictionaries for indexing based on given level names and labels
+  (see :ref:`multi-level indexing`). By
+  `Benoit Bovy <https://github.com/benbovy>`_.
 
 - New (experimental) decorators :py:func:`~xarray.register_dataset_accessor` and
   :py:func:`~xarray.register_dataarray_accessor` for registering custom xarray
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -39,10 +39,14 @@ Enhancements
   attributes are retained in the resampled object. By
   `Jeremy McGibbon <https://github.com/mcgibbon>`_.
 
+- DataArray and Dataset methods :py:meth:`sel` and :py:meth:`loc` now
+  accept dictionaries or nested tuples for indexing on multi-index dimensions.
+  By `Benoit Bovy <https://github.com/benbovy>`_.
+
 - New (experimental) decorators :py:func:`~xarray.register_dataset_accessor` and
   :py:func:`~xarray.register_dataarray_accessor` for registering custom xarray
   extensions without subclassing. They are described in the new documentation
-  page on :ref:`internals`. By `Stephan Hoyer <https://github.com/shoyer>`
+  page on :ref:`internals`. By `Stephan Hoyer <https://github.com/shoyer>`_.
 
 - Round trip boolean datatypes. Previously, writing boolean datatypes to netCDF
   formats would raise an error since netCDF does not have a `bool` datatype.
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -250,7 +250,7 @@ class DataArray(AbstractArray, BaseDataObject):
         # switch from dimension to level names, if necessary
         dim_names = {}
         for dim, idx in indexes.items():
-            if idx.name != dim:
+            if not isinstance(idx, pd.MultiIndex) and idx.name != dim:
                 dim_names[dim] = idx.name
         if dim_names:
             obj = obj.rename(dim_names)
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -88,7 +88,9 @@ class _LocIndexer(object):
     def _remap_key(self, key):
         def lookup_positions(dim, labels):
             index = self.data_array.indexes[dim]
-            return indexing.convert_label_indexer(index, labels)
+            indexer, new_idx = indexing.convert_label_indexer(index, labels)
+            # don't replace indexes in this case.
+            return indexer
 
         if utils.is_dict_like(key):
             return dict((dim, lookup_positions(dim, labels))
@@ -243,6 +245,14 @@ class DataArray(AbstractArray, BaseDataObject):
             coords = OrderedDict((k, v) for k, v in self._coords.items()
                                  if set(v.dims) <= allowed_dims)
         return self._replace(variable, coords, name)
+
+    def _replace_indexes(self, indexes):
+        obj = self
+        for dim, idx in iteritems(indexes):
+            obj = obj.rename({dim: idx.name})
+            new_coord = Coordinate(idx.name, idx)
+            obj = obj._replace(coords={idx.name: new_coord})
+        return obj
 
     __this_array = _ThisArray()
 
@@ -590,7 +600,7 @@ class DataArray(AbstractArray, BaseDataObject):
         ds = self._to_temp_dataset().isel(**indexers)
         return self._from_temp_dataset(ds)
 
-    def sel(self, method=None, tolerance=None, **indexers):
+    def sel(self, method=None, tolerance=None, drop_levels=True,  **indexers):
         """Return a new DataArray whose dataset is given by selecting
         index labels along the specified dimension(s).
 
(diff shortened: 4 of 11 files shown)


shoyer:
This looks very nice! I would probably opt for making `drop_level=True` the default for `.loc` -- it's hard to see when that wouldn't be desirable. In fact, I'm not even sure that it's worth having the option to set `drop_level=False`.

benbovy:
I followed your suggestions.

Two more comments (not critical issues I think) :

- Multi-index level drop only works when a dict is provided for a dimension, i.e., `da.sel(band_wavenumber='foo')` still returns the full `pandas.MultiIndex`.  This would require for `indexing.convert_label_indexer` to return an updated index also when `label` is not dict-like, which is not that straightforward I think.
- There is a potential conflict when providing a dictionary to `DataArray.loc`, it may be interpreted either as a mapping of dimensions / labels or a mapping of index levels / labels for the 1st dimension. For now, the second option is not handled, e.g., `da.loc[{'band': 'foo'}]` raises a `KeyError`.

In summary, `da.loc[{'band_wavenumber': {'band': 'foo'}}]` returns a DataArray with updated index and coordinate / dimension name, `da.loc['foo']` returns a DataArray with the full multi-index, and `da.loc[{'band': 'foo'}]` raises a `KeyError`. Three different results for the same operation, although I think that the first one is more explicit and better.

shoyer:
> Multi-index level drop only works when a dict is provided for a dimension, i.e., da.sel(band_wavenumber='foo') still returns the full pandas.MultiIndex. This would require for indexing.convert_label_indexer to return an updated index also when label is not dict-like, which is not that straightforward I think.

Indeed. This would require an another data structure somewhere keeping track of level names -- and ideally also ensuring that they are always unique (like dimensions). This seems fine to me for now.

> There is a potential conflict when providing a dictionary to DataArray.loc, it may be interpreted either as a mapping of dimensions / labels or a mapping of index levels / labels for the 1st dimension. For now, the second option is not handled, e.g., da.loc[{'band': 'foo'}] raises a KeyError.

I agree -- better to require the user to be explicit. I also don't see many use cases for specifying the coordinate value and level name but not the dimension name. What happens if you type `da.loc[{'band': 'foo'}, :]` or `da.loc[{'band': 'foo'}, ...]`? In principle these are not ambiguous, but I could only guess at what happens without looking at the code.

benbovy:
After refactoring a bit, `da.loc[{'band': 'foo'}, :]` and `da.loc[{'band': 'foo'}, ...]` now work as expected. 

`da.loc[{'band': 'foo'}]` still raises a `KeyError`. Although it is possible to catch this exception, it still remains ambiguous so it is probably better to do nothing.

shoyer:
Dictionaries not hashable, so we might be able to detect this case by
calling hash() and raising a KeyError with alternative suggestions. Just an
idea -- not sure if there's a clean way to implement that.

benbovy:
Unless you see any other issues, I think that this feature doesn't need more development for now.

I'll be back next week to finish this PR (write some tests and doc).

tippetts:
This will be a great feature. I for one am really looking forward to using it.

Will this work also allow saving to/reading from hdf5 and netcdf files with a MultiIndex? If not, can you give a sketch outline of the approach you (Stephan or Benoit) would take? I assume it would involve saving the information about the MultiIndex structure in some transformed way that fits into an hdf5 file, then reconstructing it on the read. I might need to hack together something for that before MultiIndex serialization makes it into xarray, but I'd like to make sure I don't veer too far off from the real solution that will ultimately come out.

benbovy:
I hope to have some time next week to work again on this PR.

@tippetts  You can see in #719  a few comments about saving/reading xarray data objects with multi-index to/from netCDF. I also looking forward to see this feature implemented - actually I need it for another project that uses xarray - so maybe I'll find some time in the next couple of weeks to start a new PR on this.

benbovy:
I finally managed to add some tests and docs.

Two more comments:

- Index replacement: if the multi-index has unnamed levels, the chosen name for the replaced index (if any) is `<old_dim>_unnamed_level` (`old_dim` is the name of the replaced multi-index dimension).
Not sure whether or not it is really necessary to track the index level number. Single indexes returned by `get_loc_level` don't keep that information. Anyway, it is always better to set explicit names for the index levels.
- Indexing on multi-index now also works with nested tuples (#850).

@shoyer I think that it is ready for review. I successfully run the tests on my local branch. Currently, CI tests seem broken for some reason I don't know.

xarray/core/dataarray.py
```diff
@@ -244,6 +240,16 @@ def _replace_maybe_drop_dims(self, variable, name=__default):
                                  if set(v.dims) <= allowed_dims)
         return self._replace(variable, coords, name)
 
+    def _replace_indexes(self, indexes):
+        obj = self
+        for dim, idx in iteritems(indexes):
+            if idx.name is None:
+                idx.name = dim + "_unnamed_level"
```
shoyer: Maybe should handle this instead by requiring names for all levels on multi-indexes used by xarray? Or we could pick defaults (e.g., `level_0`, `level_1`, etc.). I worry that this choice is likely to give you multiple levels with the same name "unnamed_level", though I suppose this isn't so bad given that the unnamed level stuff is mostly there as a soft reminder to add names.

xarray/core/dataarray.py
```diff
@@ -244,6 +240,16 @@ def _replace_maybe_drop_dims(self, variable, name=__default):
...
+            obj = obj.rename({dim: idx.name})
+            new_coord = Coordinate(idx.name, idx)
+            obj = obj._replace(coords={idx.name: new_coord})
```
shoyer: It's better to accumulate all the new coordinates in a dictionary, and then call `_replace` once outside the loop.

doc/indexing.rst
```diff
@@ -294,6 +294,37 @@ elements that are fully masked:
...
+    da_midx
+    da_midx.sel(x={'one': 'a', 'two': 0})
+    da_midx.loc[{'one': 'a'}, ...]
+
+As shown in the last example above, xarray handles partial selection on
+pandas multi-index ; it automatically renames the dimension and replaces the
+coordinate when a single index is returned (level drop).
+
+Like pandas, it is also possible to use tuples of tuples, lists or slices
+(in that case xarray always returns the full multi-index):
```
shoyer: Is this the preferred behavior or just what you could get to work? i.e., will we want to change it in the future?

benbovy: Just what I could get to work...

xarray/core/dataarray.py
benbovy: IMO `level_0`, `level_1`, etc. is the best option. I finally found a way to handle this: set default levels names on a copy of the MultIndex before calling `convert_label_indexer`.

benbovy:
Thanks for the review @shoyer !

I made some updates according to your comments.

A possible remaining issue is that using a tuple of tuples, lists or slices exactly behaves like pandas (it calls `get_locs`) while a nested list can still be used to select couples of multi-index labels (it calls `get_indexer`). Should we allow only one of two behaviors, or keep both (and mention it in the docs)?

shoyer:
> A possible remaining issue is that using a tuple of tuples, lists or slices exactly behaves like pandas (it calls get_locs) while a nested list can still be used to select couples of multi-index labels (it calls get_indexer). Should we allow only one of two behaviors, or keep both (and mention it in the docs)?

Can you give a quick example of what using the nested list look like?

benbovy:
```
>>> index = pd.MultiIndex.from_product((list('abc'), range(3)))
>>> index.names = ('one', 'two')
>>> data = xr.DataArray(range(9), [('x', index)])
>>> data
<xarray.DataArray (x: 9)>
array([0, 1, 2, 3, 4, 5, 6, 7, 8])
Coordinates:
  * x        (x) object ('a', 0) ('a', 1) ('a', 2) ('b', 0) ('b', 1) ...
```

Example using a tuple of scalar/tuple/list/slice labels for each level:

```
>>> data.sel(x=('a', (0, 1)))
<xarray.DataArray (x: 2)>
array([0, 1])
Coordinates:
  * x        (x) object ('a', 0) ('a', 1)
```

Example using a list of several combinations of multi-index labels:  

```
>>> data.sel(x=[('a', 0), ('b', 1), ('c', 1)])
<xarray.DataArray (x: 3)>
array([0, 4, 7])
Coordinates:
  * x        (x) object ('a', 0) ('b', 1) ('c', 1)
```

All the following examples raise an error:

```
>>> data.sel(x=(('a', 0), ('b', 1), ('c', 1)))
>>> data.sel(x=['a', (0, 1)])
>>> data.sel(x=[['a', 0], ['b', 1], ['c', 1]])
```

Note that after verification, pandas has the same behavior for all of the examples shown here.

shoyer:
Yeah, that behavior looks pretty good to me -- there are limits on how flexible you can make things :).

We are fortunate that we have `.sel`, which lets us side-step some tricky ambiguous cases. Still, how do we handle `data.loc['a', 0]` if `data` is 2-dimensional? Pandas uses some terrible fall-back logic to allow the `0` to index either the second level or the second dimension, depending upon which works.

benbovy:
That behavior looks good enough to me too! It is indeed already quite flexible :-).

Do you think that we really need to handle `data.loc['a', 0]` if `data` is 2 or higher dimensional? We would also have to handle the case `data.loc[{'one': 'a', 'two': 0}]`...
Having to explicitly write `data.loc[('a', 0), :]` or `data.loc[('a', 0), ...]` seems fine to me (at least for now).

shoyer:
I don't think we need to handle it -- I'm definitely supportive of only supporting unambiguous syntax. Let's just note the difference from pandas.

benbovy:
Noted the difference in documentation.

Hope that the whole "multi-level indexing" section is clear enough, not sure if it's proper English.

doc/indexing.rst
```diff
@@ -294,6 +294,60 @@ elements that are fully masked:
...
+    idx = pd.MultiIndex.from_product([list('abc'), [0, 1]],
+                                     names=('one', 'two'))
+    da_midx = xr.DataArray(np.random.rand(6, 3),
+                           [('x', idx), ('y', range(3))])
+    da_midx
+    da_midx.sel(x={'one': 'a', 'two': 0})
+    da_midx.loc[{'one': 'a'}, ...]
+
+As shown in the last example above, xarray handles partial selection on
+pandas multi-index ; it automatically renames the dimension and replaces the
```
shoyer: nit: I would use a new sentence instead of the semicolon.

doc/indexing.rst
```diff
@@ -294,6 +294,60 @@ elements that are fully masked:
...
+   da_midx.sel(x=[('a', 0), ('b', 1)])
+
+A single, flat tuple can be used to select a given combination of
+multi-index labels:
+
+.. ipython:: python
+
+   da_midx.sel(x=('a', 0))
+
+Unlike pandas, xarray can't make the distinction between index levels and
```
shoyer: Instead of "can't make the distinction", let's say "does not guess".

doc/whats-new.rst
```diff
@@ -39,10 +39,14 @@ Enhancements
   attributes are retained in the resampled object. By
   `Jeremy McGibbon <https://github.com/mcgibbon>`_.
 
+- DataArray and Dataset methods :py:meth:`sel` and :py:meth:`loc` now
+  accept dictionaries or nested tuples for indexing on multi-index dimensions.
+  By `Benoit Bovy <https://github.com/benbovy>`_.
```
shoyer: Can you also please add a note about the changed behavior (we now drop levels, which is consistent with pandas) in the "Breaking changes" section above?

Also, add a reference here to the documentation section you added:

    :ref:`multi-level indexing`

xarray/core/dataarray.py
```diff
@@ -86,24 +86,22 @@ def __init__(self, data_array):
...
-            return tuple(lookup_positions(dim, labels) for dim, labels
-                         in zip(self.data_array.dims, key))
+            labels = indexing.expanded_indexer(key, self.data_array.ndim)
+            key = dict(zip(self.data_array.dims, labels))
+        return indexing.remap_label_indexers(self.data_array, key)
 
     def __getitem__(self, key):
-        return self.data_array[self._remap_key(key)]
+        pos_indexers, new_indexes = self._remap_key(key)
+        ds = self.data_array[pos_indexers]._to_temp_dataset()
```
shoyer: Can we avoid creating the temporary dataset here? `_to_temp_dataset()`/`_from_temp_dataset` is a little expensive (and also a bit of a hack).

xarray/core/dataset.py
```diff
@@ -419,6 +419,18 @@ def _replace_vars_and_dims(self, variables, coord_names=None,
             obj = self._construct_direct(variables, coord_names, dims, attrs)
         return obj
 
+    def _replace_indexes(self, indexes):
+        variables = OrderedDict()
+        for k, v in iteritems(self._variables):
+            if k in indexes.keys():
+                idx = indexes[k]
+                variables[k] = Coordinate(idx.name, idx)
```
shoyer: If `idx.name != k` above, then this could be constructing an invalid dataset.

I think we should create `Coordinate(k, idx)` and then remap back to the original names below, if necessary.

xarray/core/dataset.py
```diff
@@ -419,6 +419,18 @@ def _replace_vars_and_dims(self, variables, coord_names=None,
...
+            else:
+                variables[k] = v
+        obj = self._replace_vars_and_dims(variables)
+        dim_names = {dim: idx.name for dim, idx in iteritems(indexes)}
+        return obj.rename(dim_names)
```
shoyer: Can we make the rename only done if necessary? I think this can be kind of expensive.

Putting things together:
```python
variables = self._variables.copy()
for name, idx in indexes.items():
   variables[name] = Coordinate(name, idx)
obj = self._replace_vars_and_dims(variables)

# switch from dimension to level names, if necessary
dim_names = {}
for dim, idx in indexes.items():
    if idx.name != dim:
        dim_names[dim] = idx.name
if dim_names:
    obj = obj.rename(dim_name)
```

xarray/core/dataarray.py
benbovy: Yeah I agree this is not very nice. I did this to avoid duplicating the `_replace_indexes` method, but I agree that it may be actually the least bad option.

xarray/core/indexing.py
```diff
@@ -177,18 +206,36 @@ def convert_label_indexer(index, label, index_name='', method=None,
...
-                                            dim, method, tolerance))
-                for dim, label in iteritems(indexers))
+
+    pos_indexers, new_indexes = {}, {}
+    for dim, label in iteritems(indexers):
+        index = data_obj[dim].to_index()
+
+        if isinstance(index, pd.MultiIndex):
+            # set default names for multi-index unnamed levels so that
+            # we can safely rename dimension / coordinate later
```
shoyer: This looks great! We might also consider moving this logic to [around this line](https://github.com/pydata/xarray/blob/2448b5b3fa3647419df8f14d60f1c32b9d002af7/xarray/core/variable.py#L107) of `core.variable.as_compatible_data`. All data passed to xarray objects goes through this method.

xarray/core/dataset.py
```diff
@@ -1114,8 +1130,9 @@ def sel_points(self, dim='points', method=None, tolerance=None,
         Dataset.isel_points
         DataArray.sel_points
         """
-        pos_indexers = indexing.remap_label_indexers(
-            self, indexers, method=method, tolerance=tolerance)
+        pos_indexers, new_indexes = indexing.remap_label_indexers(
```
shoyer: if we ignore `new_indexes` here (which is fine, certainly for now), use the variable name `_` instead to indicate that it's unused.

xarray/core/indexing.py
```diff
@@ -135,11 +135,27 @@ def _asarray_tuplesafe(values):
     return result
 
 
+def _is_nested_tuple(tup, index):
+    """Check for a compatible nested tuple and multiindex (taken from
+    pandas.core.indexing.is_nested_tuple).
```
shoyer: I'm still trying to wrap my head around exactly what this check does :).

xarray/core/dataset.py
```diff
@@ -972,8 +986,10 @@ def sel(self, method=None, tolerance=None, **indexers):
         Dataset.isel_points
         DataArray.sel
         """
-        return self.isel(**indexing.remap_label_indexers(
-            self, indexers, method=method, tolerance=tolerance))
+        pos_indexers, new_indexes = indexing.remap_label_indexers(
+            self, indexers, method=method, tolerance=tolerance
+        )
+        return self.isel(**pos_indexers)._replace_indexes(new_indexes)
```
shoyer: Does this handle the case where `new_indexes` is None?

shoyer: Nevermind, that can't happen.

xarray/core/dataset.py
benbovy: Seems much nicer!

What about

```
if not len(indexes):
    return self
```

at the top of the function, given that in many use cases indexes will be empty?  (I don't know if `self._replace_vars_and_dims` is expensive)

xarray/core/indexing.py
```diff
@@ -177,18 +206,36 @@ def convert_label_indexer(index, label, index_name='', method=None,
...
+            # we can safely rename dimension / coordinate later
+            valid_level_names = [name or '{}_level_{}'.format(dim, i)
+                                 for i, name in enumerate(index.names)]
+            index = index.copy()
+            index.names = valid_level_names
+
+        idxr, new_idx = convert_label_indexer(index, label,
+                                              dim, method, tolerance)
+        pos_indexers[dim] = idxr
+        if new_idx is not None and not isinstance(new_idx, pd.MultiIndex):
```
shoyer: What should happen if `new_idx` *is* a MultiIndex? Right not it looks like it gets thrown away?

shoyer: We definitely need to add a test for this situation (e.g., 3 level index -> 2 level index).

xarray/core/indexing.py
benbovy: So I'm ! I've just stolen this from pandas without much modification :).

shoyer: This is such a weird function the way it's currently written. Why not make this:
```python
def _is_nested_tuple(possible_tuple):
    return (isinstance(possible_tuple, tuple)
            and any(isinstance(value, (tuple, list, slice)
                    for value in possible_tuple))
```

The `isinstance(index, pd.MultiIndex)` should be outside the function, since it's totally unrelated.

xarray/core/indexing.py
```diff
@@ -166,6 +184,17 @@ def convert_label_indexer(index, label, index_name='', method=None,
...
+    elif is_dict_like(label):
+        if not isinstance(index, pd.MultiIndex):
+            raise ValueError('cannot use a dict-like object for selection on a '
+                             'dimension that does not have a MultiIndex')
+        indexer, new_index = index.get_loc_level(tuple(label.values()),
+                                                 level=tuple(label.keys()))
+
+    elif _is_nested_tuple(label, index):
+        indexer = index.get_locs(label)
+
```
shoyer: I think we could reproduce what pandas does in terms of collapsing tuple levels if we call `get_loc_level` for non-nested tuples:

```python
# untested!

elif isinstance(label, tuple) and isinstance(index, pd.MultiIndex):
    if _is_nested_tuple(label):
        indexer = index.get_locs(label)
    else:
        indexer, new_index = index.get_loc_level(label, level=range(len(label)))

```

shoyer:
Just went through and gave another full review -- this is looking quite nice, just a few more things to clean up!

xarray/core/dataset.py
shoyer: Yes, that's even better!

`self._replace_vars_and_dims` is pretty cheap (it skips all validation) but not free.

xarray/core/indexing.py
benbovy: Yes it works!

However, using non-nested tuples here consists of selecting single elements and raises the question of how we handle returned scalar values. In that specific case we should drop the dimension but keep the 0-d (multi-level) coordinate so that `data.sel(x=('a', 0))` is equivalent to `data.isel(x=0)` in the example above.

More generally, I think we definitely need to carefully address level drop in all cases.

shoyer: Doing some tests, it seems like `get_loc_level` properly drops levels from a MultiIndex, but doesn't handle dropping the level entirely. For that, we need to use `get_loc`.

Good:
```
In [11]: idx = pd.MultiIndex.from_product([['a', 'b'], [1, 2], [-1, -2]])

In [12]: idx.get_loc_level(('a', 1), [0, 1])
Out[12]:
(array([ True,  True, False, False, False, False, False, False], dtype=bool),
 Int64Index([-1, -2], dtype='int64'))

In [13]: idx.get_loc_level(('a', -1), [0, 2])
Out[13]:
(array([ True, False,  True, False, False, False, False, False], dtype=bool),
 Int64Index([1, 2], dtype='int64'))

In [15]: idx.get_loc_level(('a',), [0])
Out[15]:
(array([ True,  True,  True,  True, False, False, False, False], dtype=bool),
 MultiIndex(levels=[[1, 2], [-2, -1]],
            labels=[[0, 0, 1, 1], [1, 0, 1, 0]]))

In [16]: idx.get_loc_level((1,), [1])
Out[16]:
(array([ True,  True, False, False,  True,  True, False, False], dtype=bool),
 MultiIndex(levels=[['a', 'b'], [-2, -1]],
            labels=[[0, 0, 1, 1], [1, 0, 1, 0]]))
```

Bad:
```
In [14]: idx.get_loc_level(('a', 1, -1), [0, 1, 2])
Out[14]:
(array([ True, False, False, False, False, False, False, False], dtype=bool),
 MultiIndex(levels=[['a', 'b'], [1, 2], [-2, -1]],
            labels=[[0], [0], [1]]))
```

Good:
```
In [17]: idx.get_loc(('a', 1, -1))
Out[17]: 0
```

shoyer: So I guess we need to check the length of the tuple (probably also in the `dict` path above):
```python
elif isinstance(label, tuple) and isinstance(index, pd.MultiIndex):
    if _is_nested_tuple(label):
        indexer = index.get_locs(label)
    elif len(label) == index.nlevels:
        indexer = index.get_loc(label)
    else:
        indexer, new_index = index.get_loc_level(label, level=range(len(label)))
```

benbovy: I thinking about something like this:

```python
def _maybe_drop_levels(index):
    drop_levels = [i for i, lab in enumerate(index.labels)
                   if not np.ptp(lab.values())]
    if len(drop_levels) < len(index.labels):
        return index.droplevel()
    else:
        return index

def convert_label_indexer(...):
    
    # ...

    if isinstance(new_index, pd.MultiIndex):
        new_index = _maybe_drop_levels(new_index)

    return indexer, new_index
```

xarray/core/indexing.py
benbovy: Yes multi-indexes are not updated, but maybe we should do so (see my comment below on level drop).

xarray/core/indexing.py
shoyer: The advantage of doing something like my proposed logic (which I think is similar to what pandas does) is that whether a level is dropped depends only on the indexer type and the number of multi-index levels, as opposed to dropping levels in a way that depends also on the particular values in the indexer and index. Code that depends only on type information rather than values is generally easier to understand and less error prone.

shoyer: also, `np.ptp` only works on numbers, not strings.

benbovy: Yep, I used `np.ptp` here as pandas `MultiIndex.labels` elements are integers by definition.

Anyway, I get your logic. It is also much more efficient!

xarray/core/indexing.py
benbovy: Shouldn't be better to move this logic to `indexing.PandasIndexAdapter`, wrap the `names` (and/or `name`) property, and return a `PandasIndexAdapter` object from `convert_label_indexer`?

This is because I worry about implicit copy or renaming.

The problem would be to set default level names that are unique across dimensions, but maybe we can pass the variable name in the `PandasIndexAdapter` constructor (as an additional `name` kwarg).

benbovy: Or maybe moving this [here](https://github.com/pydata/xarray/blob/2448b5b3fa3647419df8f14d60f1c32b9d002af7/xarray/core/variable.py#L1093) (`to_index()`) would be simpler...

shoyer: `PandasIndexAdapter` is really for handling aspects of pandas indexes that we can't change by converting into a new index. For example, PeriodIndex claims to have `dtype=int`, when it's really a sort of object dtype.

`pandas.Index` is actually immutable, so we don't need to worry about changing the data. Calling `.copy()` just copies the metadata and reuses the same data. That said, calling `.set_names()` (which returns a new MultiIndex) is probably a more obvious way of handling this.

It does looks like we already some similar logic in `to_index()`, so that could be a reasonable place to put this, too.

benbovy: Compared to `as_compatible_data`, moving to `to_index()` would create a copy of the multi-index (almost free I guess) each time it is called but it would also allow tracking level name changes on the original `MultiIndex`:

```python
idx = pd.MultiIndex.from_product([['a', 'b'], [1, 2], [-1, -2]])
>>> y = xr.DataArray(np.random.rand(2 * 2 * 2), [('x', idx)])
>>> y.x.to_index()
MultiIndex(levels=[['a', 'b'], [1, 2], [-2, -1]],
           labels=[[0, 0, 0, 0, 1, 1, 1, 1], [0, 0, 1, 1, 0, 0, 1, 1], [1, 0, 1, 0, 1, 0, 1, 0]],
           names=['x_level_0', 'x_level_1', 'x_level_2'])
>>> idx.names = ('one', 'two', 'three')
>>> y.x.to_index()
MultiIndex(levels=[['a', 'b'], [1, 2], [-2, -1]],
           labels=[[0, 0, 0, 0, 1, 1, 1, 1], [0, 0, 1, 1, 0, 0, 1, 1], [1, 0, 1, 0, 1, 0, 1, 0]],
           names=['one', 'two', 'three'])
```

and we have also direct access to the dimension/coordinate name so that we can set default level names that are unique (as shown above).

shoyer: > and we have also direct access to the dimension/coordinate name so that we can set default level names that are unique (as shown above).

True, but unless we allow directly accessing levels as variables, `level_0` on `dim_0` is an unambiguous as `dim_0_level_0`.

benbovy:
Thanks for your second review @shoyer ! It actually helped me to better understand the indexing logic of pandas (never too late)!

I made some updates according to your comments.

I think we're getting closer to a working feature!

doc/data-structures.rst
```diff
@@ -115,11 +115,9 @@ If you create a ``DataArray`` by supplying a pandas
     df
     xr.DataArray(df)
 
-xarray does not (yet!) support labeling coordinate values with a
-:py:class:`pandas.MultiIndex` (see :issue:`164`).
-However, the alternate ``from_series`` constructor will automatically unpack
-any hierarchical indexes it encounters by expanding the series into a
-multi-dimensional array, as described in :doc:`pandas`.
+Xarray supports labeling coordinate values with a :py:class:`pandas.MultiIndex`.
```
shoyer: It might make sense simply to drop this paragraph instead -- do we really need to explicitly call out MultiIndex if it's supported?

xarray/core/dataarray.py
```diff
@@ -244,6 +239,23 @@ def _replace_maybe_drop_dims(self, variable, name=__default):
                                  if set(v.dims) <= allowed_dims)
         return self._replace(variable, coords, name)
 
+    def _replace_indexes(self, indexes):
+        if not len(indexes):
```
shoyer: just use `if not indexes:` (it's more idiomatic)

xarray/core/variable.py
```diff
@@ -1096,7 +1096,13 @@ def to_index(self):
         # basically free as pandas.Index objects are immutable
         assert self.ndim == 1
         index = self._data_cached().array
-        if not isinstance(index, pd.MultiIndex):
+        if isinstance(index, pd.MultiIndex):
+            # set default names for multi-index unnamed levels so that
+            # we can safely rename dimension / coordinate later
+            valid_level_names = [name or '{}_level_{}'.format(self.name, i)
```
shoyer: I would lean towards defaulting to simply "level_0",  etc., but I don't feel too strongly here.

shoyer:
I did a little bit of testing. Here is one case I found where things don't work as I expected:
```
In [32]: idx = pd.MultiIndex.from_product([['a', 'b'], [0, 1, 2]], names=['lev0', 'lev1'])

In [33]: data = xr.DataArray(range(6), [('x', idx)])

In [34]: data.sel(x=('a',))
Out[34]:
<xarray.DataArray (lev1: 3)>
array([0, 1, 2])
Coordinates:
  * lev1     (lev1) int64 0 1 2

In [35]: data.sel(x='a')
Out[35]:
<xarray.DataArray (x: 3)>
array([0, 1, 2])
Coordinates:
  * x        (x) object ('a', 0) ('a', 1) ('a', 2)
```

I would expect an index drop in the last case, too. I guess we need to check for scalars.

xarray/core/indexing.py
```diff
@@ -166,6 +175,28 @@ def convert_label_indexer(index, label, index_name='', method=None,
...
+        elif len(label) == index.nlevels:
+            indexer = index.get_loc(label)
+        else:
+            indexer, new_index = index.get_loc_level(
+                label, level=list(range(len(label)))
+            )
+
     else:
         label = _asarray_tuplesafe(label)
         if label.ndim == 0:
```
shoyer: this is where scalars end up -- probably need to add a clause here to handle MultiIndex

doc/data-structures.rst
benbovy: No I don't think we need it. However, it might be good to put a sentence somewhere in the docs to recommend users to set names for multi-index levels before creating data arrays or datasets. What do you think?

xarray/core/variable.py
benbovy: Yes it makes sense and it is more compact. However, how do we handle cases where two or more multi-index dimensions both collapse to `level_0` after indexing? Maybe such cases are rather edge cases so we don't need to care too much about this? Also a dimension / coordinate renamed to `level_0` is not very informative. Maybe we can include the old dimension name when renaming in `_replace_indexes`? Or, if we recommend xarray users to set explicit names for multi-index levels, we can probably leave this as-is...

I let you decide.

tippetts:
Mind if I ask if this will get merged into master? It looks like a lot of work went into the pull request, and the discussion + passed checks lead me to believe it could be close to going in. Is there anything a third party can do to push it across the finish line?

shoyer:
@tippetts Thanks for checking in. I actually hadn't noticed the last round of changes here, but I think this is good to go in now! Big thanks to @benbovy!!!

benbovy:
Thanks again @shoyer for your reviews and comments!

Merging this PR broke the tests added in #881. I fixed that in #903.

PR #806, opened 2016-03-28
Decorators for registering custom accessors in xarray

### LINKED ISSUE #706: Subclassing Dataset and DataArray
Hi guys,

I have started writing some SpecArray class which inherits from DataArray and defines some methods useful for dealing with wave spectra, such as calculating spectral wave statistics like significant wave height, peak wave period, etc, interpolating, splitting, and performing some other tasks. I'd like to ask please if:

- Is this something you guys would maybe be interested to add to your library?
- Is there a simple way to ensure the methods I am defining are preserved when creating a Dataset out of this SpecArray object? currently I can create / add to a Dataset using this new object, but all new methods get lost by doing that.

Thanks,
Rafael

shoyer:
Fixes #706

New (experimental) decorators `xarray.register_dataset_accessor` and `xarray.register_dataarray_accessor` for registering custom xarray extensions without subclassing. They are described in the new documentation page on `internals`.

CC @rafa-guedes @rabernat @fmaussion @khaeru @ajdawson -- as people who might use such an interface, it would be great to get some feedback about how well this would work for you.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -411,22 +411,6 @@ DataArray methods
    DataArray.load
    DataArray.chunk
 
-Backends (experimental)
------------------------
-
-These backends provide a low-level interface for lazily loading data from
-external file-formats or protocols, and can be manually invoked to create
-arguments for the ``from_store`` and ``dump_to_store`` Dataset methods.
-
-.. autosummary::
-   :toctree: generated/
-
-   backends.NetCDF4DataStore
-   backends.H5NetCDFStore
-   backends.PydapDataStore
-   backends.ScipyDataStore
-
-
 Plotting
 ========
 
@@ -441,3 +425,26 @@ Plotting
    plot.line
    plot.pcolormesh
    plot.FacetGrid
+
+Advanced API
+============
+
+.. autosummary::
+   :toctree: generated/
+
+   Variable
+   Coordinate
+   register_dataset_accessor
+   register_dataarray_accessor
+
+These backends provide a low-level interface for lazily loading data from
+external file-formats or protocols, and can be manually invoked to create
+arguments for the ``from_store`` and ``dump_to_store`` Dataset methods:
+
+.. autosummary::
+   :toctree: generated/
+
+   backends.NetCDF4DataStore
+   backends.H5NetCDFStore
+   backends.PydapDataStore
+   backends.ScipyDataStore
--- a/doc/internals.rst
+++ b/doc/internals.rst
@@ -78,7 +78,7 @@ xarray:
 
 .. literalinclude:: examples/_code/accessor_example.py
 
-This achieves the same result as if the ``Dataset`` class had a property
+This achieves the same result as if the ``Dataset`` class had a cached property
 defined that returns an instance of your class:
 
 .. python::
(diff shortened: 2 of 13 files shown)


rabernat:
:+1: very cool

dopplershift:
:+1: Very nice.

khaeru:
Of the two different projects I'm working (sporadically) on that both subclass Dataset, it seems like one (pyGDX) should more properly be a backend, while the other could work as an accessor. This code looks good!

Just to be clear—`xarray_obj` is passed to the `__init__()` method of an accessor. Will this happen before, or after `Dataset.__init__()`/`DataArray.__init__()` is invoked?

doc/api.rst
```diff
@@ -442,3 +426,26 @@ Plotting
    plot.line
    plot.pcolormesh
    plot.FacetGrid
+
+Advanced API
+============
+
+.. autosummary::
+   :toctree: generate/
```
fmaussion: generat

fmaussion:
Yes, this looks very good, thanks! 

@khaeru : I had the same question and just tested it:

```python
import xarray as xr
import numpy as np

@xr.register_dataset_accessor('geo')
class GeoAccessor(object):
    def __init__(self, xarray_obj):
        print('In constructor')
        self._obj = xarray_obj

    @property
    def center(self):
        # return the geographic center point of this dataset
        lon = self._obj.latitude
        lat = self._obj.longitude
        return (float(lon.mean()), float(lat.mean()))

    def plot(self):
        print('plot data on a map')  # e.g., using Cartopy

# Nothing is printed
ds = xr.Dataset({'longitude': np.linspace(0, 10),
                  'latitude': np.linspace(0, 20)})

# out: In constructor
out = ds.geo.center

# out: In constructor
#      plot data on a map
ds.geo.plot()
```
So the constructor is called only if the property is accessed, which is good. the drawback is that a new object is instantiated at each access.

fmaussion:
Oh, I realize that this doesn't answer your question though :flushed:

shoyer:
> Just to be clear—xarray_obj is passed to the __init__() method of an accessor. Will this happen before, or after Dataset.__init__()/DataArray.__init__() is invoked?

This will happen after Dataset.__init__/DataArray.__init__. For cases where you want to do custom initialization, the suggestion (which I should add) is to simply write your own function to use in place of `xarray.open_dataset`.

It's probably worth noting in the documentation that this is equivalent to just adding a property, e.g.,
```python
class Dataset:
    ...
    @property
    def geo(self)
        return GeoAccessor(self)
```

fmaussion:
Thanks for the clarification. Would it make sense to make a `@lazy_property` out of it in oder to avoid multiple builds?

khaeru:
@fmaussion that's still helpful, thanks.

> For cases where you want to do custom initialization, the suggestion (which I should add) is to simply write your own function to use in place of `xarray.open_dataset`.

Now that I think of it, it should also be possible to use some other in logic in `__init__()`—or even as a kludge store something like `xarray_obj.attrs['_geoaccessor_state']`—to determine whether the object is already, or needs to be, "initialized" (whatever that happens to mean for each accessor).

rafa-guedes:
:+1: nice one

shoyer:
Finally got around to updating this.

I added caching of these properties per @fmaussion's suggestion. Caching is a little error prone (I wrote a custom descriptor), but I think test cases should cover it pretty well.

jhamman:
This looks good to me.  Anyone else have any final comments?

doc/internals.rst
```diff
@@ -0,0 +1,138 @@
...
+.. ipython:: python
+
+    ds = xr.Dataset({'longitude': np.linspace(0, 10),
+                     'latitude': np.linspace(0, 20)})
+    ds.geo.center
+    ds.geo.plot()
+
+The intent here is that libraries that extend xarray could add such an accessor
+to implement subclass specific functionality rather than using actual subclasses
+rather patching in a large number of domain specific methods.
```
fmaussion: Something is wrong in this sentence

fmaussion:
Apart for the doc typo I mentioned, this looks very good, thanks!

doc/internals.rst
shoyer: fixed

Issue #807, opened 2016-03-28
cumprod returns errors


pwolfram:
The xarray implementation of `cumprod` returns an assertion error, presumably because of bottleneck, e.g., https://github.com/pydata/xarray/blob/master/xarray/core/ops.py#L333.  The error is 

```
└─▪ ./test_cumprod.py 
[ 0.8841785   0.54181236  0.29075258  0.28883015  0.1137352   0.09909713
  0.03570122  0.0304542   0.01578143  0.01496195  0.01442681  0.00980845]
Traceback (most recent call last):
  File "./test_cumprod.py", line 13, in <module>
    foo.cumprod()
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/common.py", line 16, in wrapped_func
    skipna=skipna, allow_lazy=True, **kwargs)
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/dataarray.py", line 991, in reduce
    var = self.variable.reduce(func, dim, axis, keep_attrs, **kwargs)
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/variable.py", line 871, in reduce
    axis=axis, **kwargs)
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/ops.py", line 346, in f
    assert using_numpy_nan_func
AssertionError

```

If bottleneck is uninstalled then a value error is returned:
```
└─▪ ./test_cumprod.py 
[  2.99508768e-01   2.80142920e-01   1.56389242e-01   1.10791301e-01
   4.58372649e-02   4.10865622e-02   9.91362500e-03   6.76033435e-03
   3.83574249e-03   9.54972340e-04   1.56846616e-04   6.44088547e-05]
Traceback (most recent call last):
  File "./test_cumprod.py", line 13, in <module>
    foo.cumprod()
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/common.py", line 16, in wrapped_func
    skipna=skipna, allow_lazy=True, **kwargs)
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/dataarray.py", line 991, in reduce
    var = self.variable.reduce(func, dim, axis, keep_attrs, **kwargs)
  File "/Users/pwolfram/anaconda/lib/python2.7/site-packages/xarray/core/variable.py", line 880, in reduce
[...]

Issue #791, opened 2016-03-11
Adding cumsum / cumprod reduction operators


pwolfram:
It would be useful to have the cumsum / cumprod reduction operator for DataArray and Dataset, analagous to http://xarray.pydata.org/en/stable/generated/xarray.DataArray.sum.html?highlight=sum#xarray.DataArray.sum
and http://xarray.pydata.org/en/stable/generated/xarray.Dataset.sum.html?highlight=sum#xarray.Dataset.sum

I notice this is on the TODO at https://github.com/pydata/xarray/blob/master/xarray/core/ops.py#L54 and am assuming there is something subtle here about the implementation.  I believe the issue was probably with dask, but the issue / PR at https://github.com/dask/dask/issues/923 & https://github.com/dask/dask/pull/925 may have removed the roadblock.

pwolfram:
Advice on the feasibility of adding cumsum / cumprod appreciated.

jhamman:
I don't think this should be too difficult now that cumulative reductions are available in dask and numpy. I think you'll just have to add the `cumsum` and `cumprod` names to the list of `NAN_REDUCE_METHODS`.  You may also need modify some of the logic in `_create_nan_agg_method` so it doesn't prepend a `nan` in-front of these two methods.

pwolfram:
It may be more involved, e.g., there appears to be a custom nanprod called by https://github.com/pydata/xarray/blob/master/xarray/core/ops.py#L374-L383 at https://github.com/pydata/xarray/blob/master/xarray/core/npcompat.py#L186-L253.  I could inject logic to remove the `nan` prepending for these functions at https://github.com/pydata/xarray/blob/master/xarray/core/ops.py#L310-L352 but this may be somewhat of a hack...

pwolfram:
You are correct that it shouldn't be too difficult.

pwolfram:
Note, numpy doesn't provide `cumsum` or `cumprod` that are `nan` compatible.  I just checked https://github.com/numpy/numpy/blob/master/numpy/lib/nanfunctions.py#L4-L17 so injecting logic may be the way to go.

shoyer:
`cumsum`/`cumprod` will need a slightly different (simpler) interface than the other reduce methods, because unlike other aggregation functions they don't remove a dimension (the result has the same size as the input). Also, as you point out, NumPy doesn't have a nan-skipping version of these functions.

[...]

PR #812, opened 2016-03-31
Adds cummulative operators to API 

pwolfram:
This PR will add cumsum and cumprod as discussed in https://github.com/pydata/xarray/issues/791 as well ensuring `cumprod` works for the API, resolving issues discussed at https://github.com/pydata/xarray/issues/807.  

TO DO (dependencies)

* [x] Add `nancumprod` and `nancumsum` to numpy (https://github.com/numpy/numpy/pull/7421)
* [x] Add `nancumprod` and `nancumsum` to dask (https://github.com/dask/dask/pull/1077)

This PR extends infrastructure to support `cumsum` and `cumprod` (https://github.com/pydata/xarray/issues/791).

References:
 * https://github.com/numpy/numpy/pull/7421

cc @shoyer, @jhamman

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -73,6 +73,10 @@ Enhancements
   which to concatenate.
   By `Stephan Hoyer <https://github.com/shoyer>`_.
 
+- Adds DataArray and Dataset methods :py:meth:`~xarray.DataArray.cumsum` and
+  :py:meth:`~xarray.DataArray.cumprod`.  By `Phillip J. Wolfram
+  <https://github.com/pwolfram>`_.
+
 Bug fixes
 ~~~~~~~~~
 - ``groupby_bins`` now restores empty bins by default (:issue:`1019`).
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -45,10 +45,11 @@ PANDAS_UNARY_FUNCTIONS = ['isnull', 'notnull']
 REDUCE_METHODS = ['all', 'any']
 NAN_REDUCE_METHODS = ['argmax', 'argmin', 'max', 'min', 'mean', 'prod', 'sum',
                       'std', 'var', 'median']
+NAN_CUM_METHODS = ['cumsum', 'cumprod']
 BOTTLENECK_ROLLING_METHODS = {'move_sum': 'sum', 'move_mean': 'mean',
                               'move_std': 'std', 'move_min': 'min',
                               'move_max': 'max'}
-# TODO: wrap cumprod/cumsum, take, dot, sort
+# TODO: wrap take, dot, sort
 
 
 def _dask_or_eager_func(name, eager_module=np, list_of_args=False,
@@ -201,6 +202,30 @@ def _func_slash_method_wrapper(f, name=None):
     func.__doc__ = f.__doc__
     return func
 
+_CUM_DOCSTRING_TEMPLATE = \
+        """Apply `{name}` along some dimension of {cls}.
+
+        Parameters
+        ----------
+        {extra_args}
+        skipna : bool, optional
+            If True, skip missing values (as marked by NaN). By default, only
+            skips missing values for float dtypes; other dtypes either do not
+            have a sentinel missing value (int) or skipna=True has not been
+            implemented (object, datetime64 or timedelta64).
+        keep_attrs : bool, optional
+            If True, the attributes (`attrs`) will be copied from the original
+            object to the new one.  If False (default), the new object will be
+            returned without attributes.
+        **kwargs : dict
+            Additional keyword arguments passed on to `{name}`.
+
+        Returns
+        -------
+        cumvalue : {cls}
+            New {cls} object with `{name}` applied to its data along the
+            indicated dimension.
+        """
 
 _REDUCE_DOCSTRING_TEMPLATE = \
         """Reduce this {cls}'s data by applying `{name}` along some
@@ -274,7 +299,9 @@ def _ignore_warnings_if(condition):
         yield
 
 
-def _create_nan_agg_method(name, numeric_only=False, coerce_strings=False):
+def _create_nan_agg_method(name, numeric_only=False, np_compat=False,
+                           no_bottleneck=False, coerce_strings=False,
+                           keep_dims=False):
     def f(values, axis=None, skipna=None, **kwargs):
         # ignore keyword args inserted by np.mean and other numpy aggregators
         # automatically:
@@ -292,14 +319,17 @@ def _create_nan_agg_method(name, numeric_only=False, coerce_strings=False):
                     'skipna=True not yet implemented for %s with dtype %s'
                     % (name, values.dtype))
             nanname = 'nan' + name
-            if isinstance(axis, tuple) or not values.dtype.isnative:
+            if isinstance(axis, tuple) or not values.dtype.isnative or no_bottleneck:
                 # bottleneck can't handle multiple axis arguments or non-native
                 # endianness
-                eager_module = np
+                if np_compat:
+                    eager_module = npcompat
+                else:
+                    eager_module = np
             else:
                 eager_module = bn
             func = _dask_or_eager_func(nanname, eager_module)
-            using_numpy_nan_func = eager_module is np
+            using_numpy_nan_func = eager_module is np or eager_module is npcompat
         else:
             func = _dask_or_eager_func(name)
             using_numpy_nan_func = False
@@ -312,10 +342,12 @@ def _create_nan_agg_method(name, numeric_only=False, coerce_strings=False):
                 else:
                     assert using_numpy_nan_func
                     msg = ('%s is not available with skipna=False with the '
-                           'installed version of numpy; upgrade to numpy 1.9 '
+                           'installed version of numpy; upgrade to numpy 1.12 '
                            'or newer to use skipna=True or skipna=None' % name)
                 raise NotImplementedError(msg)
     f.numeric_only = numeric_only
+    f.keep_dims = keep_dims
+    f.__name__ = name
     return f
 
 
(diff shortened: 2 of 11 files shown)


pwolfram:
@mrocklin and @shoyer, will we need to modify the definitions for `nanprod`, `nancumsum`, and `nancumprod` in dask for this to work in multi-threaded mode?  I took a quick look and it appears they are defined https://github.com/dask/dask/blob/d82cf2ac3fa3a61912b7934afe7b2fe9e14cc4ff/dask/array/__init__.py#L17-L22 so I'm assuming xarray/dask should just work once the issues on the xarray end are resolved but just wanted to double check to make sure this is the case.

shoyer:
@pwolfram dask will also definitely need `nancumsum`/`nancumprod` functions. These should be quite straightforward to add though given the [existing infrastructure](https://github.com/dask/dask/blob/master/dask/array/reductions.py).

pwolfram:
@shoyer see https://github.com/dask/dask/pull/1077 for the `nancumsum` and `nancumprod` dask PR

pwolfram:
@shoyer, it looks like I'll need to have a `_reduce_method`-like method that doesn't reduce the dimensions, e.g., https://github.com/pydata/xarray/blob/master/xarray/core/common.py#L11.  I haven't been able to get the current branch to work properly (it returns a numpy array vs xarray datatype) and seem to be missing something.  Am I on the right track that I need to add a new abstract method?  This seems overly complicated but I haven't been able to get this to work cleanly otherwise, e.g., trying things like

```python
    f = _func_slash_method_wrapper(method, name)
    setattr(cls, name, cls._binary_op(f))
```

Some advice or help to get me out of my naivete would be greatly appreciated.  Thanks!

jhamman:
@pwolfram - how's this going?  Are we still stuck on the above question?

pwolfram:
@jhamman, I need to get back to this when I can make the time but will let you know if I have more trouble.  Thanks for following up.

pwolfram:
@shoyer and @jhamman here is the general use of the cumsum and cumprod operators.  Note, I probably need to have some type of error checking for the dask version (e.g., we require a version of dask with `nancumsum`, `nanprod`, and `nancumprod`).  What is the standard way to do this?

pwolfram:
@shoyer, @jhamman, and @MaximilianR this should be ready for a preliminary review because it works.  The key thing missing is a check for dask version and potentially more testing.  Thoughts on these issues are greatly appreciated.

xarray/core/variable.py
```diff
@@ -893,16 +893,27 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=False,
         if dim is not None and axis is not None:
             raise ValueError("cannot supply both 'axis' and 'dim' arguments")
 
+        if 'cum' in func.__name__:
```
shoyer: Can we put these error checks in `ops.cumsum` instead? It's poor separation of concerns to do these checks over here.

xarray/core/variable.py
```diff
@@ -893,16 +893,27 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=False,
...
         data = func(self.data if allow_lazy else self.values,
                     axis=axis, **kwargs)
 
         removed_axes = (range(self.ndim) if axis is None
                         else np.atleast_1d(axis) % self.ndim)
-        dims = [dim for n, dim in enumerate(self.dims)
+        dims = [adim for n, adim in enumerate(self.dims)
                 if n not in removed_axes]
 
+        if 'cum' in func.__name__:
```
shoyer: Same as above -- let's keep `Variable.reduce` unmodified for now. I think we'll want to actually implement this using `xarray.apply`, anyways.

xarray/core/ops.py
```diff
@@ -344,11 +349,30 @@ def prod(values, axis=None, skipna=None, **kwargs):
             raise NotImplementedError(
                 'skipna=True not yet implemented for prod with dtype %s'
                 % values.dtype)
-        _fail_on_dask_array_input_skipna(values)
-        return npcompat.nanprod(values, axis=axis, **kwargs)
+        return _nanprod(values, axis=axis, **kwargs)
     return _prod(values, axis=axis, **kwargs)
 prod.numeric_only = True
 
+def cumsum(values, axis=None, skipna=None, **kwargs):
```
shoyer: Can you make a wrapper function that generates these functions, instead of repeating the logic three times for `prod`/`cumsum`/`cumprod`?

pwolfram: Better than that-- I extended functionality to `_create_nan_agg_method` to generalize, which removes quite a few lines of code.  See `d1c1077`.

xarray/core/ops.py
```diff
@@ -274,7 +274,8 @@ def _ignore_warnings_if(condition):
         yield
 
 
-def _create_nan_agg_method(name, numeric_only=False, coerce_strings=False):
+def _create_nan_agg_method(name, numeric_only=False, np_compat=False, 
+        no_bottleneck=False, coerce_strings=False):
```
shoyer: keep indentation aligned with `(` per PEP8

xarray/core/ops.py
```diff
@@ -292,14 +293,17 @@ def f(values, axis=None, skipna=None, **kwargs):
...
-                eager_module = np
+                if np_compat:
+                    eager_module = npcompat
+                else:
+                    eager_module = np
             else:
                 eager_module = bn
             func = _dask_or_eager_func(nanname, eager_module)
-            using_numpy_nan_func = eager_module is np
+            using_numpy_nan_func = (eager_module is np) or (eager_module is npcompat)
```
shoyer: just a nit: `is` has higher precedence than `or`, so you don't need parentheses

xarray/core/variable.py
shoyer: Actually, I guess it's OK to keep this -- it is all we need to actually implement `cumsum`, after all, but let's do the shape check unilaterally, without looking at the function name.

xarray/core/variable.py
```diff
@@ -893,16 +893,27 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=False,
...
+            def safe_shape(val):
+                return val.shape if type(val) is np.ndarray else ()
```
shoyer: this should probably be just `getattr(val, 'shape', ())` (dask arrays have shape defined, too)

xarray/core/ops.py
shoyer: Nice

xarray/core/variable.py
pwolfram: I don't think so unless there is a clever way to do this that I'm missing.  It looks like I'd need to have something like a new _partial_reduce_method that is for cumsum, cumprod, prod, etc.  However, this will require additional code lines beyond the simplistic approach I've taken.  But this may be necessary to get this to work in general in a clean way.  For instance, I don't think the existing code works properly with dataset yet.

pwolfram: @shoyer, this obviously changes the comments below because we need this to work for both dataset and dataarray and I need to verify the existing implementation does just that.

xarray/core/variable.py
pwolfram: @shoyer, I had originally implemented it that way, but this introduced a bug, I think related to use of prod if I recall correctly.

xarray/core/variable.py
pwolfram: @shoyer, I double-checked.  This appears to work for both dataarray and dataset but it would be good to have some tests to ensure this functionality works in the future.

xarray/core/variable.py
pwolfram: Thanks @shoyer!

xarray/core/ops.py
pwolfram: Thanks @shoyer!

xarray/core/variable.py
pwolfram: I've added some tests and will push them soon.

xarray/core/variable.py
pwolfram: Thankfully, it appears to work without issue on re-examination.

xarray/core/variable.py
pwolfram: Pushed.

pwolfram: I didn't individual test dataarray and dataset functionality but assume it is inherited by variable.  Hence, I've tested variable methods for cumsum and cumprod for now.

xarray/core/variable.py
```diff
@@ -893,16 +893,27 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=False,
...
         removed_axes = (range(self.ndim) if axis is None
                         else np.atleast_1d(axis) % self.ndim)
-        dims = [dim for n, dim in enumerate(self.dims)
+        dims = [adim for n, adim in enumerate(self.dims)
                 if n not in removed_axes]
 
+        def safe_shape(val):
+            return getattr(val, 'shape', ())
+
+        if safe_shape(data) == safe_shape(self.data):
```
shoyer: This should be done alternatively to calculating `dims` a few lines above, e.g.,

    if getattr(data, 'shape', ()) == self.shape:
        dims = self.dims
    else:
        removed_axes = ...
        dims = [adim for n, admin in enumerate(self.dims) ...]

Otherwise we calculate those dimensions just to throw them away

pwolfram:
@shoyer, this should be ready for another review.  I also tested it with this somewhat hacky code at https://gist.github.com/a329d441fe99ae342a34b1a374650138.  It may be good to get some type of test like this into the test suite.  However, the correct location for testing these methods, in general, is not transparent to me.  It doesn't look like we broadly check reduction operations with nans, e.g., `prod` outside the `test_variable.py` file.  I have made additions here but broader testing may be useful.

xarray/core/variable.py
shoyer: I still don't like this approach. It feels very fragile.

A slightly saner approach would another function attribute like `numeric_only` that we use in `ops.py`. Then this check could be: `keep_dims = getattr(func, 'keep_dims', False)`.

shoyer:
We don't need to verify that every value is exactly as expected for Dataset/DataArray, but we should verify the general API (e.g., do at least one of `cumsum`/`cumprod` and make sure the result has the right dimensions and errors when it should). See the various `test_reduce` methods in `test_dataset.py` for examples.

xarray/core/variable.py
pwolfram: Removed in favor of use of `keep_dims` approach

xarray/core/variable.py
pwolfram: Modified as you discussed below

xarray/core/variable.py
pwolfram: Agreed

pwolfram:
@shoyer, I think this fixes the concerns you raised including the testing.  Thanks for all the tips!

shoyer:
Can you add a basic sanity check for `DataArray.cumsum`?

Others I think this just needs docs (on the What's New and API pages)

pwolfram:
@shoyer, is this what you were thinking?

doc/whats-new.rst
```diff
@@ -62,6 +62,9 @@ By `Robin Wilson <https://github.com/robintw>`_.
   overlapping (:issue:`835`) coordinates as long as any present data agrees.
   By `Johnnie Gray <https://github.com/jcmgray>`_.
 
+- Adds DataArray and Dataset methods :py:meth:`cumsum` and :py:meth:`cumprod`.
```
shoyer: To make the links work, use, e.g.,

   :py:meth:`~DataArray.cumsum`

pwolfram: Thanks @shoyer, this is fixed now and the link should now work (had to do some minor refractoring).  However, a search for `cumsum` does not return the DataArray and Dataset results for my local test, which is very strange.

shoyer:
Thanks! Let's see how the docs look at http://xarray.pydata.org/en/latest/whats-new.html in a few minutes after the doc build completes

Issue #621, opened 2015-10-12
An iris cube converter?


DamienIrving:
Would it be possible to include a function to convert from an `xray.DataArray` to an `iris.cube`? 

For weather/climate/ocean science, xray is by far the best library for general data processing, while [iris and cartopy](http://scitools.org.uk/) are the best for geospatial plotting. At the moment I usually write the results of my xray data processing out to a netCDF file, then read that file in with iris for plotting. I'm assuming many other people do the same, so a `.to_cube()` function might allow people to skip that extra input/output step.

shoyer:
Yes, this would certainly be a welcome addition (see #37). If you're interested in working on this, #236 is reasonable template.

Issue #37, opened 2014-03-03
Functions for converting DataArrays to and from iris.Cubes


shoyer:

Zac-HD:
Closed by #1750?

PR #814, opened 2016-04-01
xarray to and from iris

nparley:
Added a first go at a converter from xarray.DataArrays objects to Iris.cube.Cube objects. Uses the same template as the cdms2 conversion.


nparley:
Reference https://github.com/pydata/xarray/issues/621 and https://github.com/pydata/xarray/issues/37. Uses template from https://github.com/pydata/xarray/pull/236.

nparley:
`docker pull nparley/xarray_test` is a python 2.7 environment with Iris etc. which mirrors the travis `CONDA_ENV=py27-cdat+iris+pynio`. 

It doesn't do everything, but it's a good first go. Here is an example:

```python 
>>> import xarray
>>> import iris
>>> nc_file = 'E1_north_america.nc'
>>> cube = iris.load_cube(nc_file)
>>> darray = xarray.open_dataset(nc_file)['air_temperature']
>>> print(darray)
<xarray.DataArray 'air_temperature' (time: 240, latitude: 37, longitude: 49)>
[435120 values with dtype=float32]
Coordinates:
  * time                     (time) datetime64[ns] 1860-06-01 1861-06-01 ...
  * latitude                 (latitude) float32 15.0 16.25 17.5 18.75 20.0 ...
  * longitude                (longitude) float32 225.0 226.875 228.75 ...
    forecast_period          (time) timedelta64[ns] 449 days 18:00:00 ...
    forecast_reference_time  datetime64[ns] 1859-09-01T06:00:00
    height                   float64 1.5
Attributes:
    standard_name: air_temperature
    units: K
    Model scenario: E1
    ukmo__um_stash_source: m01s03i236
    source: Data from Met Office Unified Model 6.05
    cell_methods: time: mean (interval: 6 hour)
>>> print(xarray.DataArray.from_iris(cube))
<xarray.DataArray u'air_temperature' (time: 240, latitude: 37, longitude: 49)>
[435120 values with dtype=float32]
Coordinates:
  * time                     (time) datetime64[ns] 1860-06-01 1861-06-01 ...
  * latitude                 (latitude) float32 15.0 16.25 17.5 18.75 20.0 ...
  * longitude                (longitude) float32 225.0 226.875 228.75 ...
    forecast_reference_time  datetime64[ns] 1859-09-01T06:00:00
    height                   float64 1.5
    forecast_period          (time) timedelta64[ns] 449 days 18:00:00 ...
Attributes:
    Model scenario: E1
    STASH: m01s03i236
    Conventions: CF-1.5
    source: Data from Met Office Unified Model 6.05
    standard_name: air_temperature
    units: K
>>> print(cube)
air_temperature / (K)               (time: 240; latitude: 37; longitude: 49)
     Dimension coordinates:
          time                           x              -              -
          latitude                       -              x              -
          longitude                      -              -              x
     Auxiliary coordinates:
          forecast_period                x              -              -
     Scalar coordinates:
          forecast_reference_time: 1859-09-01 06:00:00
          height: 1.5 m
     Attributes:
          Conventions: CF-1.5
          Model scenario: E1
          STASH: m01s03i236
          source: Data from Met Office Unified Model 6.05
     Cell methods:
          mean: time (6 hour
>>> print(darray.to_iris())
air_temperature / (K)               (time: 240; latitude: 37; longitude: 49)
     Dimension coordinates:
          time                           x              -              -
          latitude                       -              x              -
          longitude                      -              -              x
     Auxiliary coordinates:
          forecast_period                x              -              -
     Scalar coordinates:
          forecast_reference_time: 1859-09-01 06:00:00
          height: 1.5 m
     Attributes:
          Model scenario: E1
          source: Data from Met Office Unified Model 6.05
          ukmo__um_stash_source: m01s03i236
```

xarray/test/test_dataarray.py
```diff
@@ -1639,6 +1639,50 @@ def test_to_and_from_cdms2(self):
...
+        original = DataArray(np.arange(6).reshape(2, 3), coord_dict,
+                             name='Temperature', attrs={'baz': 123,
+                                                        'units': 'Kelvin',
+                                                        'standard_name':
+                                                            'fire_temperature',
+                                                        'long_name':
+                                                            'Fire Temperature'},
+                             dims=('distance', 'time'))
+
+        expected_coords = [Coordinate('distance', [-2, 2]),
```
shoyer: Maybe better to just pull these out of `original.coords` instead?

xarray/test/test_dataarray.py
```diff
@@ -1639,6 +1639,50 @@ def test_to_and_from_cdms2(self):
...
+        self.assertEqual(actual.var_name, original.name)
+        self.assertItemsEqual([d.var_name for d in actual.dim_coords],
+                              original.dims)
+
+        for coord, expected_coord in zip((actual.coords()), expected_coords):
+            self.assertEqual(coord.var_name, expected_coord.name)
+            self.assertArrayEqual(coord.points, expected_coord.values)
+            self.assertEqual(actual.coord_dims(coord),
+                             original.get_axis_num
+                             (original.coords[coord.var_name].dims))
```
shoyer: Maybe add a check for coordinate attributes?

Might be worth checking:
- that units get converted properly to/from attributes into cf_units
- that random attributes are preserved as attributes

xarray/convert.py
```diff
@@ -49,3 +58,81 @@ def set_cdms2_attrs(var, attrs):
...
+        return _args
+
+    dim_coords = []
+    aux_coords = []
+
+    for coord_name in dataarray.coords:
+        coord = encode(dataarray.coords[coord_name])
+        coord_args = get_args(coord.attrs)
+        coord_args['var_name'] = coord_name
+        iris_coord = iris.coords.DimCoord(coord.values, **coord_args)
```
shoyer: You are probably going to need to use `AuxCoord` for some coordinates -- maybe anything that doesn't match a dimension name?  In principle, `DimCoord` instances [must be 1d, numeric and strictly monotonic](http://scitools.org.uk/iris/docs/latest/iris/iris/coords.html#iris.coords.DimCoord). It does seem like this works for your test examples, though, so I'm not sure what's going on.

xarray/convert.py
```diff
@@ -49,3 +58,81 @@ def set_cdms2_attrs(var, attrs):
...
+        attrs.update(_obj.attributes)
+        return dict((k, v) for k, v in attrs.items() if v is not None)
+
+    name = cube.var_name
+    dims = [dim.var_name for dim in cube.dim_coords]
+    coords = OrderedDict()
+
+    for coord in cube.coords():
+        coord_attrs = get_attr(coord)
+        coord_dims = [cube.coords()[i].var_name for i in cube.coord_dims(coord)]
```
shoyer: it's probably a better idea to calculate `cube.coords()` once.

shoyer:
This is awesome!

Let's try to add some documentation -- at a mention in What's New.

CC @pelson

xarray/convert.py
```diff
@@ -3,24 +3,36 @@
 import numpy as np
 
 from .core.dataarray import DataArray
+from .core.pycompat import OrderedDict
 from .conventions import (
     maybe_encode_timedelta, maybe_encode_datetime, decode_cf)
 
-ignored_attrs = set(['name', 'tileIndex'])
+cdms2_ignored_attrs = {'name', 'tileIndex'}
```
shoyer: Currently we support Python 2.6, so set literals will cause test failures.

I think we can change this for the next major release of xarray though -- I just sent out a mailing list post about this.

xarray/convert.py
nparley: Yes I had that thought too. I started off with the code only working for dimensional coordinates and then extended to auxiliary later. It's because `add_aux_coord` can accept `DimCoord` or `AuxCoord` and my test is only 1D. I think it'll probably fail if the coord is 2D with the way it's currently written. I will check.

xarray/convert.py
nparley: I can change these, I wasn't doing dictionary comprehension with the new syntax for this reason but forgot about sets. It would not surprise me if quite a few scientist were still using python 2.6. Why did this not get picked up in the travis py26 tests?

xarray/convert.py
nparley: Yes I should actually be able to use `dims` as a lookup here, which is a list of the dimension names.

pelson:
Thanks for pinging @shoyer. I'll also ping @rhattersley, and one of us will take a good look in the next few days. :+1:

nparley:
There is a problem when it comes to bounds. Xarray adds the bounds as another DataArray to the DataSet, i.e.: 
```
Data variables:
    votemper      (dim0, dim1) float64 nan nan nan nan nan nan nan nan nan ...
    deptht_bnds   (bnds) float64 0.0 10.0
```
I can get the name of this variable from the Coord attribute in the DataArray for another variable, i.e. `ds['votemper'].coords['deptht'].bounds` but can't get the bound array from the DataArray? is that correct?

shoyer:
There's no good solution for representing bounds in xarray -- we can put these as coordinates on Dataset objects, but they aren't supported on DataArray objects. I had some plans to add an `IntervalIndex` type to pandas (https://github.com/pydata/pandas/pull/8707) which would let us support bounds, but honestly I'm not sure if/when I'll finish that.

nparley:
Ok, bounds can just be something to add to the converter if / when it's supported.

nparley:
Latest commit fixes bug if the Iris cube does not have any dimension coordinates. What do people think about keeping or dropping the automatic dimension coordinates that xarray creates? And if dropping should be done should all coordinates named `dim{x}` be dropped? Or are we happy with just leaving them there?

For example at the moment:

```python

>>> print(cube)
sea_water_potential_temperature / (degC) (-- : 148; -- : 180)
     Auxiliary coordinates:
          latitude                           x         x
          longitude                          x         x
     Scalar coordinates:
          depth: 4.99994 m, bound=(0.0, 10.0) m
          time:    1-01-01 12:00:00
     Attributes:
          Conventions: CF-1.5
     Cell methods:
          mean: time_counter
>>> print(xarray.DataArray.from_iris(cube).to_iris())
sea_water_potential_temperature / (degC) (dim0: 148; dim1: 180)
     Dimension coordinates:
          dim0                                x          -
          dim1                                -          x
     Auxiliary coordinates:
          latitude                            x          x
          longitude                           x          x
     Scalar coordinates:
          depth: 4.99993801117 m
          time:    1-01-01 12:00:00
     Attributes:
          Conventions: CF-1.5

```

rhattersley:
> What do people think about keeping or dropping the automatic dimension coordinates that xarray creates?

Is there any way to know for sure whether a coordinate has been automatically created? Or is the name the only clue?

rhattersley:
> Uses the same template as the cdms2 conversion.

If this is the way to do things then does [this comment](https://github.com/pydata/xarray/blob/2ee2c653946cc12e80e640144835044f08cae8fc/xarray/core/variable.py#L36) need deleting?

nparley:
The comment was just to mention that I had based my code on the code from cdms2 from #236 as suggested by @shoyer. I guess this comment would not be needed in the final merge.

shoyer:
> Is there any way to know for sure whether a coordinate has been automatically created? Or is the name the only clue?

I'm afraid the name is the only clue. My inclination would be not to do anything special here to "un-name" default coordinates, assuming Iris has a straightforward way to get rid of them. Dimensions like `dim0` come up quickly in demo code but rarely in practice -- it's mostly a signal to the user that the should pick a meaningful name.

> If this is the way to do things then does this comment need deleting?

Yes, that comment is out of date. I'm currently refactoring that code though, so let me take care care of it.

There's a separate argument about whether or not xarray should automatically convert cubes when passed into the `xarray.DataArray` constructor. We do this for pandas objects, but not any other labelled array types, currently.

shoyer:
One other thing that I noticed from the latest example -- it would be nice to covert cell methods from Cubes into string attributes on DataArrays.

nparley:
OK the latest commit tries to do the cell methods:

```python
>>> print(cube)
sea_water_potential_temperature / (degC) (-- : 148; -- : 180)
     Auxiliary coordinates:
          latitude                           x         x
          longitude                          x         x
     Scalar coordinates:
          depth: 4.99994 m, bound=(0.0, 10.0) m
          time:    1-01-01 12:00:00
     Attributes:
          Conventions: CF-1.5
     Cell methods:
          mean: time_counter
>>> print(xarray.DataArray.from_iris(cube))
<xarray.DataArray u'votemper' (dim0: 148, dim1: 180)>
array([[ nan,  nan,  nan, ...,  nan,  nan,  nan],
       [ nan,  nan,  nan, ...,  nan,  nan,  nan],
       [ nan,  nan,  nan, ...,  nan,  nan,  nan],
       ..., 
       [ nan,  nan,  nan, ...,  nan,  nan,  nan],
       [ nan,  nan,  nan, ...,  nan,  nan,  nan],
       [ nan,  nan,  nan, ...,  nan,  nan,  nan]], dtype=float32)
Coordinates:
    deptht        float64 5.0
    time_counter  object    1-01-01 12:00:00
    nav_lat       (dim0, dim1) float32 -78.1906 -78.1906 -78.1906 -78.1906 ...
    nav_lon       (dim0, dim1) float32 80.0 82.0 83.9999 85.9999 87.9999 ...
  * dim0          (dim0) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 ...
  * dim1          (dim1) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 ...
Attributes:
    units: degC
    long_name: Temperature
    standard_name: sea_water_potential_temperature
    cell_methods: time_counter: mean
    Conventions: CF-1.5
>>> print(xarray.DataArray.from_iris(cube).to_iris())
sea_water_potential_temperature / (degC) (dim0: 148; dim1: 180)
     Dimension coordinates:
          dim0                                x          -
          dim1                                -          x
     Auxiliary coordinates:
          latitude                            x          x
          longitude                           x          x
     Scalar coordinates:
          depth: 4.99993801117 m
          time:    1-01-01 12:00:00
     Attributes:
          Conventions: CF-1.5
     Cell methods:
          mean: time_counter
```

xarray/convert.py
```diff
@@ -49,3 +62,142 @@ def set_cdms2_attrs(var, attrs):
...
+        _cell_method_words = [w.strip() for w in cell_methods_str.split(':')]
+        cm = {'coords': [], 'method': '', 'interval': [], 'comment': []}
+        skip = False
+        for i, word in enumerate(_cell_method_words):
+            # If this value is a comment or an interval don't read
+            if skip:
+                skip = False
+                continue
+            # If this word is an axis
+            if word not in cell_methods_strings | set(['interval', 'comment']):
```
shoyer: can you give an example of what you're trying to catch here? this looks very complicated to me -- maybe there is an more straightforward way to do this

nparley: Yes there may, but it is actually a little complicated to parse these strings. The format is:

"coord1: [coord2:] method: [comment: a comment] [interval: a interval]"

I.e. here are some examples:

```python
>>> get_cell_methods('dim1: dim2: mean: comment: A comment: interval: A interval')
[CellMethod(method='mean', coord_names=('dim1', 'dim2'), intervals=('A interval',), comments=('A comment',))]
>>> get_cell_methods('dim1: dim2: mean: dim2: sum:')
[CellMethod(method='mean', coord_names=('dim1', 'dim2'), intervals=(), comments=()), CellMethod(method='sum', coord_names=('dim2',), intervals=(), comments=())]
>>> get_cell_methods('dim1: mean: dim2: dim3: sum: comment: A comment')
[CellMethod(method='mean', coord_names=('dim1',),  intervals=(), comments=()), CellMethod(method='sum', coord_names=('dim2', 'dim3'), intervals=(), comments=('A comment',))]
```

So that line is looking to see if the word is a method or "comment" or "interval", if not it's a coord and so checks to see if it's a new cell method or another coord for the current one.

shoyer: Is there something like `iris.coords.CellMethod.from_string` which handles this logic? It might be a happier path to expose such a method from Iris (which already is using this sort of logic internally) rather than incorporating it into xarray.

pelson: I haven't looked closely enough, but maybe https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/netcdf.py#L1422

shoyer: @pelson that's going the other direction: iris coordinate -> string (which I think is easier to get right)

nparley: @shoyer I had a look before I wrote that code but could not see one. Is there one in the netcdf library?

pelson: Looks like the code is horrifically obfuscated in Pyke rules: https://github.com/SciTools/iris/blob/master/lib/iris/fileformats/_pyke_rules/fc_rules_cf.krb#L1556-L1619

I'd be completely happy if this were moved into a more useful place. I *think* at the moment, it is reachable through ```iris.fileformats._pyke_rules.compiled_krb.fc_rules_cf_fc```, but I haven't tested it, and can't guarantee its stability there.

nparley:
OK this update uses ```iris.fileformats._pyke_rules.compiled_krb.fc_rules_cf_fc``` as @pelson mentioned in the comment above. I also added a cell method to the test

xarray/convert.py
```diff
@@ -49,3 +62,108 @@ def set_cdms2_attrs(var, attrs):
...
     return cdms2_var
+
+
+# TODO: Add converting bounds from xarray to Iris and back
+def to_iris(dataarray):
+    """Convert a DataArray into a Iris Cube
+    """
+    # Iris not a hard dependency
+    import iris
+    import iris.fileformats._pyke_rules.compiled_krb.fc_rules_cf_fc \
```
shoyer: Can we add some sort of guard so this doesn't fail if/when Iris cleans up it's internals and moves this function? We've been bitten by using private APIs before.

Ideally, @pelson would add something exposing this to the public API for the next release, and we could use something like:
```python
try:
    from iris.fileformats import parse_cell_methods
except ImportError:
    # prior to Iris 2.0 (or whatever)
    from iris.fileformats._pyke_rules.compile_krb.fc_rules_cf_fc \
        import _parse_cell_methods as parse_cell_methods
```

nparley: Yes I can add that. @pelson how do you think this might be exposed by Iris?

rhattersley: I'll put up a PR for `iris.fileformats.netcdf.parse_cell_methods`

rhattersley: See SciTools/iris#1981

nparley:
@shoyer @pelson @rhattersley Is there anything you think could be added to this PR or is this ok for now?

xarray/test/test_dataarray.py
```diff
@@ -1639,6 +1640,60 @@ def test_to_and_from_cdms2(self):
         roundtripped = DataArray.from_cdms2(actual)
         self.assertDataArrayIdentical(original, roundtripped)
 
+    def test_to_and_from_iris(self):
+        try:
+            import iris
+            import cf_units
+        except ImportError:
+            raise unittest.SkipTest('iris not installed')
```
jhamman: would it be better to use a `@requires_iris` decorator?

jhamman:
This PR needs "what's new" and "api" entries in the docs.

xarray/convert.py
```diff
@@ -31,12 +47,9 @@ def to_cdms2(dataarray):
     # we don't want cdms2 to be a hard dependency
     import cdms2
 
-    def encode(var):
-        return maybe_encode_timedelta(maybe_encode_datetime(var.variable))
-
-    def set_cdms2_attrs(var, attrs):
+    def set_cdms2_attrs(_var, attrs):
```
shoyer: is there a reason you needed to change this variable name?

xarray/convert.py
```diff
@@ -49,3 +62,111 @@ def set_cdms2_attrs(var, attrs):
...
+            import _parse_cell_methods as parse_cell_methods
+
+    # iris.unit is deprecated in Iris v1.9
+    import cf_units
+
+    def check_attrs(attrs, keys):
+        return dict((k, v) for k, v in attrs.items() if k in keys)
+
+    def get_args(attrs):
+        _args = {'attributes': filter_attrs(attrs, iris_forbidden_keys)}
```
shoyer: no need to preface variables inside a function with an underscore -- they are already scoped to only the internal helper function.

xarray/convert.py
```diff
@@ -49,3 +62,111 @@ def set_cdms2_attrs(var, attrs):
...
+            iris_coord = iris.coords.AuxCoord(coord.values, **coord_args)
+            aux_coords.append((iris_coord, axis))
+
+    args = get_args(dataarray.attrs)
+    args['var_name'] = dataarray.name
+    args['dim_coords_and_dims'] = dim_coords
+    args['aux_coords_and_dims'] = aux_coords
+    if 'cell_methods' in dataarray.attrs:
+        args['cell_methods'] = \
+            parse_cell_methods(dataarray.name, dataarray.attrs['cell_methods'])
```
shoyer: Avoid line breaks with `\` is possible, per PEP8. I would prefer:
```
        args['cell_methods'] = parse_cell_methods(
            dataarray.name, dataarray.attrs['cell_methods'])
 ```

**Comment by shoyer** [comment]
@jhamman and I have a few minor suggestions, but otherwise this looks good!

File: xarray/convert.py (hunk 9 above)
Comment (nparley) [inline xarray/convert.py]: I don't like variable shadowing so I prefer to add the underscore, it's because I have had to deal with some really badly written academic code. Here it's clear enough so can change it back.

File: xarray/convert.py (hunk 10 above)
Comment (nparley) [inline xarray/convert.py]: It was because there was an outer args and I wanted to make it clear this was not that. I have seen people write helper functions and make use of outer scope variables. Again maybe it's clear enough here.

Comment (shoyer) [inline xarray/convert.py]: Also option is to make some private helper functions (preface the names
with _) in this module instead of nesting them inside the convert function.
That can often be significantly cleaner.

Comment (nparley) [inline xarray/convert.py]: Yes, that's a good idea. I will add those changes.

**Comment by rhattersley** [comment]
There's been a question over whether `parse_cell_methods` should really have the `cf_var_name` argument. It's only used if it raises a warning, and even then it's only used to help make the message more user friendly. In itself that doesn't seem like a big deal, but it did make me wonder if you'd rather if it didn't raise the warning _at all_.?

**Comment by shoyer** [comment]
@rhattersley I think this function is cleaner without the `cf_var_name` argument, so I would probably remove it. That said, I usually try to avoid issuing warnings *at all*.

**Comment by nparley** [comment]
I have merged upstream and it seems to be failing but I don't think it's related to this PR, as a clean clone of master branch is also failing https://travis-ci.org/nparley/xarray/jobs/132872124

**Comment by shoyer** [comment]
@nparley Let's see if https://github.com/pydata/xarray/pull/860 works. I think that will probably solve the CI issue (at least on Travis-CI).

**Comment by shoyer** [comment]
OK, I was not paying attention here with #860 -- I assumed that the build problem was with the build including Iris, but it was actually the other build that started spontaneously failing.

I'm afraid I already merged that PR, so could you please try rebasing again? Sorry for the hassle! On the plus side, I think using conda-forge will ultimately work better than using all the separate channels.

File: xarray/convert.py [hunk 12]
```diff
@@ -49,3 +64,121 @@ def set_cdms2_attrs(var, attrs):
...
+        cell_methods.append(names + cell_method.method + extra)
+    return ' '.join(cell_methods)
+
+
+def from_iris(cube):
+    """ Convert a Iris cube into an DataArray
+    """
+    name = cube.var_name
+    dims = [dim.var_name for dim in cube.dim_coords]
+    if not dims:
```
shoyer: I don't think this logic is quite right. If I remember correctly, a cube can have any number of dimensions without coordinates. This only works if the cube has no dimension coords, but doesn't cover the mixed case.

xarray/convert.py
```diff
@@ -49,3 +64,121 @@ def set_cdms2_attrs(var, attrs):
...
+        dims = ["dim{}".format(i) for i in range(cube.data.ndim)]
+    coords = OrderedDict()
+
+    for coord in cube.coords():
+        coord_attrs = _iris_obj_to_attrs(coord)
+        coord_dims = [dims[i] for i in cube.coord_dims(coord)]
```
shoyer: Is it possible to have coord_dims(coord) return a tuple that does not have length equal to it's number of dimensions?

xarray/convert.py
shoyer: Is there a sane/standard way to figure out which dimensions on the cube are missing from dim_coords, given that [it may not have length equal to `cube.ndim`](http://scitools.org.uk/iris/docs/v1.9.2/iris/iris/cube.html#iris.cube.Cube.dim_coords)?

nparley: Good spot, I'll investigate

xarray/convert.py
nparley: coord_dims(coord): 

> Returns a tuple of the data dimensions relevant to the given coordinate.

So I think it's ok.

shoyer:
@pelson @rhattersley any tips on handling the edge cases of cubes with some, but not all missing dim_coords?

@nparley I would hate for this to get lost. I'm OK merging this before it covers every edge case -- let's just document the limitations and leave that for later cleanup.

nparley:
@shoyer I have not forgotten about this PR, I just been working a lot on a pandas PR updating the ci and it's been taking up a lot of my spare time.

shoyer:
No worries, no rush here from my end :).

pelson:
> @pelson @rhattersley any tips on handling the edge cases of cubes with some, but not all missing dim_coords?

I might consider something like (untested):

```
for dim in xrange(cube.ndim):
    try:
        dim_coord = cube.coord(dim_coords=True, dimensions=(dim,))
    except iris.exceptions.CoordinateNotFoundError:
        index_coord = range(cube.shape[dim])
```

nparley:
This adds the extra code that @pelson suggested but it would be nice to create a little unit test with a mixed Dimension and Auxiliary coordinates. Is there an easy way to create a simple cube in Iris that could test this?

xarray/convert.py
```diff
@@ -49,3 +64,131 @@ def set_cdms2_attrs(var, attrs):
...
+    import iris.exceptions
+    name = cube.var_name
+    dims = []
+    for dim in xrange(cube.ndim):
+        try:
+            dim_coord = cube.coord(dim_coords=True, dimensions=(dim,))
+            dims.append(dim_coord.var_name)
+        except iris.exceptions.CoordinateNotFoundError:
+            index_coord = range(cube.shape[dim])
+            dims.append("dim{}".format(index_coord))
```
shoyer: Use `dim_{}` to be consistency with our normal rule for default dimensions

xarray/convert.py
```diff
@@ -49,3 +64,131 @@ def set_cdms2_attrs(var, attrs):
...
+    return ' '.join(cell_methods)
+
+
+def from_iris(cube):
+    """ Convert a Iris cube into an DataArray
+    """
+    import iris.exceptions
+    name = cube.var_name
+    dims = []
+    for dim in xrange(cube.ndim):
```
shoyer: I would use something other `dim` here (maybe `dim_index`) to clarify that it's an integer

xarray/convert.py
shoyer: This should actually be something like `dims.append('dim_{}'.format(dim))` -- `index_coord` is currently the default coordinate labels, not the dimension name.

xarray/convert.py
```diff
@@ -49,3 +64,131 @@ def set_cdms2_attrs(var, attrs):
...
+            extra = ' ({})'.format(extra)
+        cell_methods.append(names + cell_method.method + extra)
+    return ' '.join(cell_methods)
+
+
+def from_iris(cube):
+    """ Convert a Iris cube into an DataArray
+    """
+    import iris.exceptions
+    name = cube.var_name
```
shoyer: Is it possible for a cube not to have variable name?

xarray/convert.py
```diff
@@ -49,3 +64,131 @@ def set_cdms2_attrs(var, attrs):
...
+            coords[coord.var_name] = ((),
+                                      np.asscalar(coord.points), coord_attrs)
+
+    array_attrs = _iris_obj_to_attrs(cube)
+    cell_methods = _iris_cell_methods_to_str(cube.cell_methods)
+    if cell_methods:
+        array_attrs['cell_methods'] = cell_methods
+    dataarray = DataArray(cube.data, coords=coords, name=name,
+                          attrs=array_attrs, dims=dims)
+    return decode_cf(dataarray.to_dataset())[dataarray.name]
```
shoyer: Note that `to_dataset()` will fail if you don't have a `name` (other than `None`) on the DataArray.

xarray/convert.py
nparley: I don't think so. A cube has to be associated with a data variable I think, but maybe need to tag in the Met people on this question.

nparley: A quick search in the Iris source gives this:

```python
    @property
    def var_name(self):
        """The CF variable name for the object."""
        return self._var_name

    @var_name.setter
    def var_name(self, name):
        if name is not None:
            if not name:
                raise ValueError('An empty string is not a valid CF variable '
                                 'name.')
            elif set(name).intersection(string.whitespace):
                raise ValueError('{!r} is not a valid CF variable name because'
                                 ' it contains whitespace.'.format(name))
        self._var_name = name
```

So would conclude that's it's not an empty string or None

xarray/convert.py
nparley: From above I think name could be None but will not else be empty. Is this ok?

shoyer: Unfortunately not -- you need to provide a name other than `None` to make this work.

Let's use the private `_to_temp_dataset()` and `_from_temp_dataset()` methods from `DataArray` that already handle this:
```
decoded_ds = decode_cf(dataarray._to_temp_dataset())
return dataarray._from_temp_dataset(decoded_ds)
```

Or once I merge #959:
```
decoded_ds = decode_cf(dataarray._to_temp_dataset())
return DataArray._from_temp_dataset(decoded_ds)
```

xarray/convert.py
```diff
@@ -49,3 +64,130 @@ def set_cdms2_attrs(var, attrs):
...
+    # dims = [dim.var_name for dim in cube.dim_coords]
+    # if not dims:
+    #     dims = ["dim{}".format(i) for i in range(cube.data.ndim)]
+    coords = OrderedDict()
+
+    for coord in cube.coords():
+        coord_attrs = _iris_obj_to_attrs(coord)
+        coord_dims = [dims[i] for i in cube.coord_dims(coord)]
+        if coord_dims:
+            coords[coord.var_name] = (coord_dims, coord.points, coord_attrs)
```
shoyer: I think it's also possible to `coord.var_name` be None. We should probably raise an error in that case.

nparley: Is there an xarray exception that might make sense for the error? Or just `nameerror`?

shoyer: We usually just raise ValueError for cases like this where we cannot handle
the input.

pelson:
This is looking pretty comprehensive to me. Could maybe do with a couple of tests perhaps, but otherwise 👍 .

shoyer:
> This is looking pretty comprehensive to me. Could maybe do with a couple of tests perhaps, but otherwise :+1: .

Indeed, merging this is really just blocked on tests right now. We need at least *something* so we can verify that this basically works.

nparley:
This did used to have some tests, but they seem to have got lost somewhere along the way. I'll add the old tests back in, which would be a start.

xarray/test/test_dataarray.py
```diff
@@ -1816,60 +1816,6 @@ def test_to_and_from_cdms2(self):
         roundtripped = DataArray.from_cdms2(actual)
         self.assertDataArrayIdentical(original, roundtripped)
 
-    def test_to_and_from_iris(self):
```
DWesl: Was this overlooked in a big merge?

DWesl:
If you're still looking for the old tests, it looks like they disappeared in the last merge commit, [f48de5](https://github.com/pydata/xarray/pull/814/commits/f48de5a3ea916b80a7bfc070c3c1c3549e931189#diff-48ab4ba033ad06981f566a2e2f561f5aL1819).

shoyer:
Yes, we would still love to get this in, though it might need a few other
changes to work now. Could you open a new PR with the tests added back in?
(Assuming Neil is not interested in picking this back up)

nparley:
Sorry guys I don't have time to do anything on this (or any open source stuff) at the moment.

shoyer:
No worries! The great thing about GitHub is that it's pretty easy for
someone else to pick up your work :)

duncanwp:
I'm happy to pick this up - I'm not sure how best to pick up someone else's pull request though. Am I best off forking nparley:master then resubmitting, or creating a new one and merging in the changes from nparley:master?

nparley:
I think also I could make you a contributor or transfer the branch to someone. So that might be an option too.

shoyer:
I think either way works. Probably easiest to fork this branch and then
merge in the latest from master.

pelson:
Closed by #1750?

fmaussion:
Indeed. Thanks to all people involved!

Issue #816, opened 2016-04-04
opendap and gzipped files


swnesbitt:
Found an issue with opening opendap files that are remotely gzipped - appears that the scipy netCDF backend is used to read gzipped netCDF, but that backend can't do openDAP apparently?  Works with local files.

``` ipython
print(ncfiles[0])
ascat=xr.open_dataset(ncfiles[0])

http://opendap.jpl.nasa.gov/opendap/OceanWinds/ascat/preview/L2/metop_a/12km/2013/027/ascat_20130127_004801_metopa_32553_eps_o_125_2101_ovw.l2.nc.gz
---------------------------------------------------------------------------
IOError                                   Traceback (most recent call last)
<ipython-input-35-c5425b003df3> in <module>()
      1 print(ncfiles[0])
----> 2 ascat=xr.open_dataset(ncfiles[0])

/data/keeling/a/snesbitt/anaconda2/lib/python2.7/site-packages/xarray/backends/api.pyc in open_dataset(filename_or_obj, group, decode_cf, mask_and_scale, decode_times, concat_characters, decode_coords, engine, chunks, lock, drop_variables)
    197                                  'supported on Python 2.6')
    198             try:
--> 199                 store = backends.ScipyDataStore(gzip.open(filename_or_obj))
    200             except TypeError as e:
    201                 # TODO: gzipped loading only works with NetCDF3 files.

/data/keeling/a/snesbitt/anaconda2/lib/python2.7/gzip.pyc in open(filename, mode, compresslevel)
     32 
     33     """
---> 34     return GzipFile(filename, mode, compresslevel)
     35 
     36 class GzipFile(io.BufferedIOBase):

/data/keeling/a/snesbitt/anaconda2/lib/python2.7/gzip.pyc in __init__(self, filename, mode, compresslevel, fileobj, mtime)
     92             mode += 'b'
     93         if fileobj is None:
---> 94             fileobj = self.myfileobj = __builtin__.open(filename, mode or 'rb')
     95         if filename is None:
     96             # Issue #13781: os.fdopen() creates a fileobj with a bogus name

IOError: [Errno 2] No such file or directory: 'http://opendap.jpl.nasa.gov/opendap/OceanWinds/ascat/preview/L2/metop_a/12km/2013/027/ascat_20130127_004801_metopa_32553_eps_o_125_2101_ovw.l2.nc.gz'
```

dopplershift:
[...]

PR #817, opened 2016-04-05
	modified:   xarray/backends/api.py

swnesbitt:
This pull request addresses issue #816, allowing the user to open remote openDAP files that are gzipped using the netCDF4 openDAP-enabled backend. Thanks to @dopplershift and @shoyer for the quick guidance!

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -195,7 +195,8 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
             else:
                 if engine is not None and engine != 'scipy':
                    raise ValueError('can only read gzipped netCDF files with '
-                                     "default engine or engine='scipy'")
+                                     "default engine, engine='pydap' or "
+                                     "engine='scipy'")
                 # if the string ends with .gz, then gunzip and open as netcdf file
             if sys.version_info[:2] < (2, 7):
                 raise ValueError('reading a gzipped netCDF not '
@@ -203,6 +204,8 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
             try:
                 if engine == 'netcdf4':
                     store = backends.NetCDF4DataStore(filename_or_obj, group=group)
+                elif engine == 'pydap':
+                    store = backends.PydapDataStore(filename_or_obj)
                 else:
                     store = backends.ScipyDataStore(gzip.open(filename_or_obj))
             except TypeError as e:
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -189,14 +189,13 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     elif isinstance(filename_or_obj, basestring):
         if filename_or_obj.endswith('.gz'):
             if is_remote_uri(filename_or_obj):
-                if engine !='netcdf4':
+                if engine !='netcdf4' and engine !='pydap':
                    raise ValueError('can only read remote gzipped netCDF files '
-                                     "with engine='netcdf4'")
+                                     "with engine='netcdf4' or 'pydap'")
             else:
                 if engine is not None and engine != 'scipy':
                    raise ValueError('can only read gzipped netCDF files with '
-                                     "default engine, engine='pydap' or "
-                                     "engine='scipy'")
+                                     "default engine, engine='scipy'")
                 # if the string ends with .gz, then gunzip and open as netcdf file
             if sys.version_info[:2] < (2, 7):
                 raise ValueError('reading a gzipped netCDF not '
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -188,15 +188,23 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         store = filename_or_obj
     elif isinstance(filename_or_obj, basestring):
         if filename_or_obj.endswith('.gz'):
-            if engine is not None and engine != 'scipy':
-                raise ValueError('can only read gzipped netCDF files with '
-                                 "default engine or engine='scipy'")
-            # if the string ends with .gz, then gunzip and open as netcdf file
+            if is_remote_uri(filename_or_obj):
+                if engine !='netcdf4':
+                   raise ValueError('can only read remote gzipped netCDF files '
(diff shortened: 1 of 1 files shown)


xarray/backends/api.py
```diff
@@ -188,15 +188,23 @@ def maybe_decode_store(store, lock=False):
...
     elif isinstance(filename_or_obj, basestring):
         if filename_or_obj.endswith('.gz'):
-            if engine is not None and engine != 'scipy':
-                raise ValueError('can only read gzipped netCDF files with '
-                                 "default engine or engine='scipy'")
-            # if the string ends with .gz, then gunzip and open as netcdf file
+            if is_remote_uri(filename_or_obj):
+                if engine !='netcdf4':
+                   raise ValueError('can only read remote gzipped netCDF files '
+                                     "with engine='netcdf4'")
```
shoyer: this is actually not quite right -- pydap would also work.

shoyer:
I think this is probably correct, but the heuristics here are starting to get convoluted enough that I worry about test coverage. Is there any way we can test this? Maybe try to pull the gzip logic into a helper function (an extended variant of `_get_default_engine`) that we could test?

swnesbitt:
Certainly can take a crack at this.  This is working for my application, but would be good to clean up the code to simplify the cases as you suggest.

Regarding testing, it would be good to have a test module that includes grabbing files for local access and accessing openDAP datasets (from UNIDATA?) to ensure that the installed backends are all working as expected.

xarray/backends/api.py
swnesbitt: didn't deal with the mess, but added `pydap` to the possibilities.

shoyer:
We do have existing tests for backends: https://github.com/pydata/xarray/blob/master/xarray/test/test_backends.py

This includes a test accessing an OpenDAP dataset from the OpenDAP test server (via pydap, at the end). But in my experience, there servers are somewhat unreliable (maybe available 90%), so we don't require that test to pass for the build to pass. Also, even in the best case scenario network access is slow. So it would be nice to modularize this enough that our logic is testable without actually using opendap.

jhamman:
> Regarding testing, it would be good to have a test module that includes grabbing files for local access and accessing openDAP datasets (from UNIDATA?) to ensure that the installed backends are all working as expected.

I'd be a little hesitant to enforce that we get access to opendap (or other remote datasets) on travis.  I've seen this come back to bite us in the past.  We can allow tests to fail on travis and just print a warning via pytest if it is something we really want to see tested in that way.

shoyer:
Currently the way we handle this is that the only test that accessing
remote resources is the pydap test, which only runs in one build (not
required to pass).

swnesbitt:
OK - understood.  Incidentally, I didn't include pydap because it doesn't seem to be python 3.* compatible.  I tested it on 2.7 for my application and it seems to work.

jhamman:
@shoyer - are we okay with the final logic here?

@swnesbitt - can we get an entry in the what's new docs?

xarray/backends/api.py
```diff
@@ -188,15 +188,25 @@ def maybe_decode_store(store, lock=False):
...
+                                     "with engine='netcdf4' or 'pydap'")
+            else:
+                if engine is not None and engine != 'scipy':
+                   raise ValueError('can only read gzipped netCDF files with '
+                                     "default engine, engine='scipy'")
+                # if the string ends with .gz, then gunzip and open as netcdf file
             if sys.version_info[:2] < (2, 7):
                 raise ValueError('reading a gzipped netCDF not '
                                  'supported on Python 2.6')
             try:
```
shoyer: This `try`/`except` clause needs to only wrap the call to `ScipyDataStore` -- otherwise it's catching too many potentially wrong conditions.

xarray/backends/api.py
```diff
@@ -188,15 +188,25 @@ def maybe_decode_store(store, lock=False):
         store = filename_or_obj
     elif isinstance(filename_or_obj, basestring):
         if filename_or_obj.endswith('.gz'):
-            if engine is not None and engine != 'scipy':
-                raise ValueError('can only read gzipped netCDF files with '
-                                 "default engine or engine='scipy'")
-            # if the string ends with .gz, then gunzip and open as netcdf file
+            if is_remote_uri(filename_or_obj):
+                if engine !='netcdf4' and engine !='pydap':
```
shoyer: pep8 nit: you're missing a space after `!=`.

shoyer:
I am reluctant to merge this without having any way to test the logic. Without automated tests this issue is likely to recur.

That said, I suppose we could leave the refactoring for a TODO. Let's add a note on that and also one minimal test to verify that we raise an error if you try to use `engine='netcdf4'` with a path to local gzipped file.

tsaoyu:
How is this issue progressed so far?  I am running into the same problem now and I might able to offer some help in testing.

dopplershift:
Regarding the testing issue, another option is to use something like vcrpy to record and playback http responses for opendap requests. I've had good luck with that for Siphon.

tsaoyu:
I am trying to understand the logic by looking through the comments/discussion on this commit.
After two years of development, there are quite some changes on the API itself e.g. the logic on the gzipped file handling seems already moved from `api.py`
to `scipy_.py`. So I think it would be better to start a new commit from the latest version. 

To bring the concept on the same table, there are 3 backends that can be used to handle gzipped netCDF file namely `scipy`, `netcdf4` and `pydap`. `scipy` backend only support netCDF3 so far and that is the only implemented method to handle gzipped netCDF3 file. `netcdf4` and `pydap` should able to deal with netCDF4 file given the file is first opened by `gzip.open` (as it was implement in `scipy_.py`). 

So what we need to do is to make both `netCDF4_.py` and `pydap_.py` aware of the file is end with `.gz` and first open it using `gzip.open`. Following tests also need to be done to verify the implementation of the logic. I have looked at vcrpy and it seems a nice idea to speed up tests with HTTP stuff involved. I must confess I only use xarray with local file so it may takes quite a while to understand the issue.

jhamman:
Closing this as stale and out of date with our current backends. @swnesbitt (or others) - feel free to open a new PR if you feel there is more to do here.

PR #823, opened 2016-04-14
Remove non-existing empty_like from Sphinx autosummary

cdeil:
I noticed that `empty_like` is listed in the top-level function API:
http://xarray.pydata.org/en/latest/api.html#top-level-functions

But I think it doesn't exist in xarray.
This PR removes the entry from the automodsummary list.

Also: I was the 100th person to fork xarray a minute ago.
What's my prize? I could use a better :car:  or a :house_with_garden: .

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -17,7 +17,6 @@ Top-level functions
    align
    broadcast
    concat
-   empty_like
    set_options
 
 Dataset


jhamman:
Thanks.  This snuck through in #668.  FWIW, there is a private method `xarray.core.common._full_like`.

Issue #828, opened 2016-04-16
Attributes are currently kept when arrays are resampled, and not when datasets are resampled


mcgibbon:
Because line 323 of groupby.py copies attributes from a DataArray to its resampling output (it shouldn't), attributes are kept in many cases when DataArrays are resampled (and not kept for similar cases when Datasets are resampled).

shoyer:
I agree, this is a bug. We should make this consist and follow the `keep_attrs` argument.

mcgibbon:
@shoyer I've corrected this in a PR I'll submit shortly.

shoyer:
Awesome, thanks!

shoyer:
Fixed by #829

PR #829, opened 2016-04-16
keep_attrs for Dataset.resample and DataArray.resample

### LINKED ISSUE #825: keep_attrs for Dataset.resample and DataArray.resample
Currently there is no option for preserving attributes when resampling a Dataset or DataArray. Could there be a keep_attrs keyword argument for these methods?

mcgibbon:
Closes #825 and #828

This update might break some scripts, because of bugs that existed in the code before. resample was inconsistent as to when it would and would not keep attributes. In particular, Dataset resampling usually threw out attributes, but not when how='first' or how='last', and DataArray preserved attributes for everything I tested ('first', 'last', and 'mean').

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -22,9 +22,25 @@ This release includes
 
 Enhancements
 ~~~~~~~~~~~~
+
 - DataArray and Dataset method :py:meth:`where` now supports a ``drop=True``
   option that clips coordinate elements that are fully masked.  By
   `Phillip J. Wolfram <https://github.com/pwolfram>`_.
+
+- DataArray and Dataset method :py:meth:`resample` now supports the 
+  ``keep_attrs=False`` option that determines whether variable and dataset
+  attributes are retained in the resampled object. By
+  `Jeremy McGibbon <https://github.com/mcgibbon>`_.
+
+Bug fixes
+~~~~~~~~~
+
+- Attributes were being retained by default for some resampling
+  operations when they should not. With the ``keep_attrs=False`` option, they
+  will no longer be retained by default. This may be backwards-incompatible
+  with some scripts, but the attributes may be kept by adding the
+  ``keep_attrs=True`` option. By
+  `Jeremy McGibbon <https://github.com/mcgibbon>`_.
 
 .. _whats-new.0.7.2:
 
--- a/xarray/test/test_dataarray.py
+++ b/xarray/test/test_dataarray.py
@@ -1413,17 +1413,12 @@ class TestDataArray(TestCase):
         array.attrs['meta'] = 'data'
 
         resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=True)
-        actual = resampled_array.meta
-        expected = 'data'
+        actual = resampled_array.attrs
+        expected = array.attrs
         self.assertEqual(expected, actual)
 
         resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=False)
-        try:
-            resampled_array.meta
-        except AttributeError:
-            pass
-        else:
-            self.fail('metadata should be discarded when keep_attrs=False')
+        assert resampled_array.attrs == {}
 
     def test_resample_mean_keep_attrs(self):
         times = pd.date_range('2000-01-01', freq='6H', periods=10)
@@ -1431,17 +1426,12 @@ class TestDataArray(TestCase):
         array.attrs['meta'] = 'data'
 
         resampled_array = array.resample('1D', dim='time', how='mean', keep_attrs=True)
-        actual = resampled_array.meta
-        expected = 'data'
+        actual = resampled_array.attrs
+        expected = array.attrs
         self.assertEqual(expected, actual)
 
         resampled_array = array.resample('1D', dim='time', how='mean', keep_attrs=False)
-        try:
-            resampled_array.meta
-        except AttributeError:
-            pass
-        else:
-            self.fail('metadata should be discarded when keep_attrs=False')
+        assert resampled_array.attrs == {}
 
     def test_resample_skipna(self):
         times = pd.date_range('2000-01-01', freq='6H', periods=10)
--- a/xarray/test/test_dataarray.py
+++ b/xarray/test/test_dataarray.py
@@ -1407,6 +1407,42 @@ class TestDataArray(TestCase):
                              name='time')
         self.assertDataArrayIdentical(expected, actual)
 
+    def test_resample_first_keep_attrs(self):
+        times = pd.date_range('2000-01-01', freq='6H', periods=10)
+        array = DataArray(np.arange(10), [('time', times)])
+        array.attrs['meta'] = 'data'
+
+        resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=True)
+        actual = resampled_array.meta
+        expected = 'data'
+        self.assertEqual(expected, actual)
+
+        resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=False)
+        try:
+            resampled_array.meta
+        except AttributeError:
+            pass
+        else:
+            self.fail('metadata should be discarded when keep_attrs=False')
+
+    def test_resample_mean_keep_attrs(self):
+        times = pd.date_range('2000-01-01', freq='6H', periods=10)
+        array = DataArray(np.arange(10), [('time', times)])
+        array.attrs['meta'] = 'data'
+
+        resampled_array = array.resample('1D', dim='time', how='mean', keep_attrs=True)
+        actual = resampled_array.meta
+        expected = 'data'
+        self.assertEqual(expected, actual)
+
+        resampled_array = array.resample('1D', dim='time', how='mean', keep_attrs=False)
+        try:
+            resampled_array.meta
+        except AttributeError:
+            pass
+        else:
+            self.fail('metadata should be discarded when keep_attrs=False')
+
     def test_resample_skipna(self):
         times = pd.date_range('2000-01-01', freq='6H', periods=10)
         array = DataArray(np.ones(10), [('time', times)])
--- a/xarray/test/test_dataset.py
+++ b/xarray/test/test_dataset.py
@@ -1647,7 +1647,7 @@ class TestDataset(TestCase):
                       'bar': ('time', np.random.randn(10), {'meta': 'data'}),
                       'time': times})
 
-        actual = ds.resample('1D', dim='time', how='first')
+        actual = ds.resample('1D', dim='time', how='first', keep_attrs=True)
         expected = ds.isel(time=[0, 4, 8])
         self.assertDatasetIdentical(expected, actual)
 
@@ -1663,10 +1663,15 @@ class TestDataset(TestCase):
         ds = Dataset({'foo': (['time', 'x', 'y'], np.random.randn(10, 5, 3)),
                       'bar': ('time', np.random.randn(10), {'meta': 'data'}),
                       'time': times})
+        ds.attrs['dsmeta'] = 'dsdata'
 
         resampled_ds = ds.resample('1D', dim='time', how='mean', keep_attrs=True)
         actual = resampled_ds['bar'].meta
         expected = 'data'
+        self.assertEqual(expected, actual)
+
+        actual = resampled_ds.dsmeta
+        expected = 'dsdata'
         self.assertEqual(expected, actual)
 
     def test_resample_by_mean_discarding_attrs(self):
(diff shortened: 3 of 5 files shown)


xarray/test/test_dataarray.py
```diff
@@ -1407,6 +1407,42 @@ def test_resample_first(self):
...
+        array.attrs['meta'] = 'data'
+
+        resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=True)
+        actual = resampled_array.meta
+        expected = 'data'
+        self.assertEqual(expected, actual)
+
+        resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=False)
+        try:
+            resampled_array.meta
```
shoyer: Instead, use something like: `assert 'meta' not in resampled_array.attrs`

shoyer: Another option, slightly better still: `assert resampled_array.attrs == {}`

xarray/test/test_dataarray.py
```diff
@@ -1407,6 +1407,42 @@ def test_resample_first(self):
...
 
+    def test_resample_first_keep_attrs(self):
+        times = pd.date_range('2000-01-01', freq='6H', periods=10)
+        array = DataArray(np.arange(10), [('time', times)])
+        array.attrs['meta'] = 'data'
+
+        resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=True)
+        actual = resampled_array.meta
+        expected = 'data'
+        self.assertEqual(expected, actual)
```
shoyer: use `assert expected.attrs == array.attrs` or `self.assertEqual(expected.attrs, array.attrs)`

xarray/test/test_dataarray.py
```diff
@@ -1407,6 +1407,42 @@ def test_resample_first(self):
                              name='time')
         self.assertDataArrayIdentical(expected, actual)
 
+    def test_resample_first_keep_attrs(self):
+        times = pd.date_range('2000-01-01', freq='6H', periods=10)
+        array = DataArray(np.arange(10), [('time', times)])
+        array.attrs['meta'] = 'data'
+
+        resampled_array = array.resample('1D', dim='time', how='first', keep_attrs=True)
```
shoyer: Instead of writing this whole method twice, why don't you use a loop? e.g.,
```
for how  in ['first', 'mean']:
   ...
```

xarray/test/test_dataset.py
```diff
@@ -1658,6 +1658,59 @@ def test_resample_and_first(self):
             actual = ds.resample('3H', 'time', how=how)
             self.assertDatasetEqual(expected, actual)
 
+    def test_resample_by_mean_with_keep_attrs(self):
```
shoyer: I have the same thoughts on these methods as the other -- you can consolidate them into a single test with a loop over `how`

shoyer:
Thanks for diving into this!

I think you can write the tests a little bit more succinctly, but otherwise this looks good. Also, please add a note to the "What's New" giving yourself credit.

MaximilianR:
I know I'm late to the party here. Are there many cases where `keep_attrs` would be expected to be `False`? I would have thought that it's generally expected that `attrs` would stick around when doing operations, either as a default or without an option to drop them. (Of course, it's easy to drop them if a user explicitly sets them to `{}` in a separate operation)

mcgibbon:
The idea is that if a dataset has an attribute, it is making a claim about that data. xarray can't guarantee that claims the attributes make about the data remain valid after operating on that data, so it shouldn't retain those attributes unless the user says it can.

I may have ceilometer data that tells me whether a cloud base is detected at any point in time, with an attribute saying that 0 means no cloud detected, another attribute saying that 1 means cloud detected, and another saying that nan means some kind of error. If I resample that data using a mean, those attributes are no longer valid.

Or my Dataset may have an attribute saying that it was output by a certain instrument. If I save that Dataset after doing some analysis, it may give the impression to someone reading the netCDF that they're reading unprocessed instrument data, when they aren't.

Or I may want the hourly variance of a dataset, and do dataset.resample('1H', how='var'). In this case, the units are no longer valid.

It may seem like these are edge cases, but it's better to make no claims most of the time than to make bad claims some of the time.

MaximilianR:
@mcgibbon That makes sense to me

xarray/test/test_dataarray.py
mcgibbon: There are a couple reasons not to - it may be unclear which "how" caused the test to fail (though that can usually be fixed with good error messages), and it will often be unclear whether multiple "how"s would cause the test to fail. For "first" and "mean" in particular, different bugs may cause different ones (or both) to fail.

I'll consider this though as I clean the tests up.

shoyer: That's a fair point -- this is always a tradeoff, especially with tests.

mcgibbon:
@shoyer I've done the clean-ups you suggested, apart from for-looping tests for the reasons I mentioned in the line note. I hope my "what's new" additions are appropriate.

doc/whats-new.rst
```diff
@@ -22,10 +22,26 @@ This release includes
...
+- DataArray and Dataset method :py:meth:`resample` now supports the 
+  ``keep_attrs=False`` option that determines whether variable and dataset
+  attributes are retained in the resampled object. By
+  `Jeremy McGibbon <https://github.com/mcgibbon>`_.
+
+Bug fixes
+~~~~~~~~~
+
+- Attributes were being retained by default for some resampling
+  operations when they should not. With the ``keep_attrs=False`` option, they
```
shoyer: can we call out `first` and `last` as these specific methods?

mcgibbon: Not really. It was also a problem with "mean" and I don't know what other operations on DataArray objects, probably all of the numpy operations. AFAIK for Dataset the only problems were "first" and "last", but I only tested "first" and "mean" - it was a problem for "first" and not for "mean".

mcgibbon: If you'd like another message that takes that into account, let me know and I can put it in. It just gets verbose quickly to explain the difference in behavior for DataArray and Dataset objects, with different "how" options (do you call those different resampling operations?), and the fact that not all "how"s were tested.

mcgibbon:
Appveyor build failed for some reason when trying to set up Miniconda on Windows 32-bit with Python 2.7. The 64-bit build of Python 3.4 passed.

> Downloading Miniconda-3.7.3-Windows-x86.exe from http://repo.continuum.io/miniconda/Miniconda-3.7.3-Windows-x86.exe
> File saved at C:\projects\xray\Miniconda-3.7.3-Windows-x86.exe 
> Installing C:\projects\xray\Miniconda-3.7.3-Windows-x86.exe to C:\Python27-conda32 
> C:\projects\xray\Miniconda-3.7.3-Windows-x86.exe /S /D=C:\Python27-conda32 
> Start-Process : This command cannot be run due to the error: The specified 
> executable is not a valid application for this OS platform..

xarray/test/test_dataarray.py
jhamman: Wait, I thought the whole point here was that they the different methods (e.g. `mean`, `first`, `max`, ...) are supposed to act in the same way.  I think this is a great place to loop through a series of options:

```Python
for keep_attrs in [True, False]:
    for how in ['mean', 'max', 'min', 'std', 'first', 'last']:
        resampled_array = array.resample('1D', dim='time', how=how, keep_attrs=keep_attrs)

        if keep_attrs:
            # check that attrs are present
        else:
            # check that attrs are not present
```

xarray/test/test_dataarray.py
```diff
@@ -1407,6 +1407,32 @@ def test_resample_first(self):
                              name='time')
         self.assertDataArrayIdentical(expected, actual)
 
+    def test_resample_first_keep_attrs(self):
+        times = pd.date_range('2000-01-01', freq='6H', periods=10)
+        array = DataArray(np.arange(10), [('time', times)])
+        array.attrs['meta'] = 'data'
```
jhamman: you may consider using the fixtures we create [here](https://github.com/mcgibbon/xarray/blob/master/xarray/test/test_dataarray.py#L17-L22).  You could get away with:

```Python
array = self.dv
array.attrs = self.attrs

# later on...
assert array.attrs == self.attrs
```

xarray/test/test_dataarray.py
mcgibbon: In theory yes, but "first" and "last" are treated differently from the rest, so I tested "first" and one random other option ("mean"). If I ran the test you just wrote, and it failed for how="mean", keep_attrs=True, I wouldn't know anything about any of the other cases until I fixed the code for that case so that it could run the other cases. It's best practice to split different cases you want to test into separate tests.

It might make sense to have one test for ("first", "last") and another for the rest, but it shouldn't be necessary to test the other cases (and you'd never get a fully exhaustive list). If you were to add any new case it should probably be a lambda function.

xarray/test/test_dataarray.py
mcgibbon: I was copying the style of the existing test_resample methods. I think a new array/dataset is created for that method because the fixtures don't include a time axis, which is important for resampling. 

Code duplication in tests isn't something to be feared, though. [This person on stack overflow](http://stackoverflow.com/a/3436062) summarizes how I feel about code duplication in tests pretty well.

xarray/test/test_dataarray.py
jhamman: I get what you are saying but I disagree with the premise. Remember, the reason you are writing this test in the first place was that  `how = {first, last}` were behaving differently. So the point here should be to make sure that they don't do that anymore. It is because they are treated differently that we should test them together. The way I see it is that this is a single case.

In practice, we make multiple assertions in single unit test all the time (browse the xarray code base for plenty of examples). If the first assertion fails in a test, it doesn't continue and that isn't a problem because the test failed.

Lastly, no, we'll never get an exhaustive list but that shouldn't stop us from thinking about the known edge cases. The known edge case in question is that `how = {first, mean}` treat the attributes differently.  If we are concerned about the `resample.reduce` treatment of attributes, maybe add a `how=np.mean` along with the others.

xarray/test/test_dataarray.py
jhamman: Readability still counts in test code. I find unit tests that copy all but a line or two from a related test difficult to distinguish between (e.g. `test_resample_first_keep_attrs` vs. `test_resample_mean_keep_attrs`).

  I don't entirely buy the SO post's perspective but he does contradict your point to some extent:

> If you can factor something in test code and you're sure it's ok, then do it.

mcgibbon: Yes, readability does count, and simple code is more readable. When I look at the line "assert array.attrs=self.attrs", my first thought is "why am I comparing an array to a Unittest.TestCase?". I have to go somewhere else in the code (to the setup function) to figure out what's going on. There are similar questions like "what is self.dv and how is it different from self.v" that would be more readable if the code was not factored out. I'm questioning "is it really OK to factor this?". We're talking about a very small difference in readability, if any, so if you like elegance I would see why you'd rather have it factored. And it's not that bad when the variables that are set up explain themselves.

You can immediately distinguish between test_resample_first_keep_attrs and test_resample_mean_keep_attrs by their name. If you can't, I should make the name clearer or add a docstring.

For this particular case, since you want the initialization code factored out I'll edit the fixtures to have a time axis, and then use them. I don't think it's a good idea to have two sets of fixtures (one with and one without a time axis). If adding a time axis breaks other tests I'll keep it how it is now.

xarray/test/test_dataarray.py
mcgibbon: The issue wasn't that how=first behaves differently from how=mean, the problem was that how=first and how=mean were both failing, but the causes of the failures were different because the code that handles them is different. Because the failure of one doesn't tell you whether the other will fail, they're two different tests.

It makes sense to have a test that loops over how=('first', 'last') and a test that loops over ('mean', 'max', 'min', 'std'), but not a test that loops over the union. And you can have multiple assertions in one test (so long as they're all testing the same thing), but not multiple tests in one test function. If you want to know when they fail independently of each other they should be split into multiple tests.

I treated the known edge case by having a test for how=first and a test for how=mean. I'm not really concerned about the treatment of other functions based on what I've seen of the code, but if you want I can add some more functions in the form of a for loop for completeness. I should at least add a case for a function argument like np.mean.

shoyer:
@mcgibbon thanks.

I agree that there is a careful balance when writing test methods between maintainability and readability.

I do think fixtures that can be helpful, but we definitely do over-use them right now with our giant test cases -- those should really be cleaned up and split up.  I just haven't had the energy to get around to that.

PR #837, opened 2016-04-27
Fixed a bug that showed when trying to establish equivalence of two a…

maciekswat:
…rrays that are numpy.recarrays

The following code WILL run with the fix but will crash when running current xarray:

```Python
import numpy as np
import xarray as xr

p_data = np.array([('John', 180), ('Stacy', 150), ('Dick',200)], dtype=[('name', '|S256'), ('height', int)])
p_data_1 = np.array([('John', 180), ('Stacy', 150), ('Dick',200)], dtype=[('name', '|S256'), ('height', int)])

weights_0 = xr.DataArray([80,56,120], dims=['participant'], coords={'participant':p_data})
weights_1 = xr.DataArray([81,52,115], dims=['participant'], coords={'participant':p_data_1})

print weights_1-weights_0
```


shoyer:
Thanks for looking into this!

Just to summarize, the issue is that `isnull` doesn't work on structured dtypes:
```
>>> pd.isnull(p_data)
TypeError: Not implemented for this type
```

The problem with your current patch is that it will actually compare all array elements twice. Because checks for equality are done quite often in xarray, this will have prohibitive performance cost.

So, we'll either need to check for structured dtypes, or catch `TypeError` exceptions and try doing the comparison without checking for nulls.

Also, we need tests, to verify that the bug is fixed.

maciekswat:
I have rewrote array_equiv fcn from ops.py to use single comparison (the one which tests from null is inside  try/except clause). Also added test case to test_DataArray.py

shoyer:
@maciekswat awesome -- thanks! 

can you please run your code through a PEP8 checker like pep8 or pyflakes? some of your spacing is off.

also, please add a note to "What's new" giving yourself credit for the bug fix :).

xarray/core/ops.py
```diff
@@ -187,7 +187,18 @@ def array_equiv(arr1, arr2):
...
+    try:
+
+        flag_array |= (isnull(arr1) & isnull(arr2))
+
+    except TypeError:
+        pass
+
+    return bool(flag_array.all())
+
+    # return bool(((arr1 == arr2) | (isnull(arr1) & isnull(arr2))).all())
```
shoyer: remove commented out lines

maciekswat:
just fixed the formatting according to PEP8 guidelines

xarray/test/test_dataarray.py
```diff
@@ -8,7 +8,10 @@
                     Coordinate, Variable)
 from xarray.core.pycompat import iteritems, OrderedDict
 from xarray.core.common import _full_like
-from . import (TestCase, ReturnItem, source_ndarray, unittest, requires_dask,
+# from . import (TestCase, ReturnItem, source_ndarray, unittest, requires_dask,
+#                requires_bottleneck)
+
```
jhamman: remove this commented out code.

xarray/test/test_dataarray.py
```diff
@@ -63,6 +66,67 @@ def test_data_property(self):
...
+        weights_1 = DataArray([81, 52, 115], dims=['participant'],
+                              coords={'participant': p_data_1})
+
+        actual = weights_1 - weights_0
+
+        expected = DataArray([1, -5], dims=['participant'],
+                             coords={'participant': p_data_2})
+
+        self.assertDataArrayIdentical(actual, expected)
+
```
jhamman: I think the tests you've written are overly verbose.  I don't think you need to recreate `p_data*` and `weights*` for each of your tests.  The way it is written now, I'm not sure what the differences between your three tests are (at least test1 and test2).

maciekswat:
I simplified the test - please feel free to do further edits, also added doc string explaining what the test checks for - essentially it checks if DataArray subtraction behaves correctly when dimension is a structured numpy array.

xarray/core/ops.py
```diff
@@ -187,7 +187,16 @@ def array_equiv(arr1, arr2):
...
     if arr1.shape != arr2.shape:
         return False
-    return bool(((arr1 == arr2) | (isnull(arr1) & isnull(arr2))).all())
+
+    flag_array = (arr1 == arr2)
+    try:
+
+        flag_array |= (isnull(arr1) & isnull(arr2))
+
+    except TypeError:
```
shoyer: could you add a brief comment explaining why we allow `TypeError` here? a reference to this issue (e.g., `GH837`)` could be nice, too.

maciekswat:
Added a comment explaining why we need to catch typeError in array_equiv

jhamman:
@maciekswat - this is almost there. @shoyer and I each have comments remaining that should be addressed and then we need a bug fix entry in "what's new".  Thanks!

maciekswat:
Added bugfix entry in what's new

shoyer:
This looks good to me. The test could probably still be made more succinct but this covers the essentials. I like that it provides an integration test for using rec arrays.

@jhamman any more comments before I merge?

shoyer:
Merged via https://github.com/pydata/xarray/commit/0d0ae9d3e766c3af9dd98383ab3b33dfea9494dc

Issue #567, opened 2015-09-09
Best way to find data variables by standard_name


rsignell-usgs:
Is there a way to return the data variables that match a specified `standard_name`?

I came up with this, but maybe the functionality already exists or there is a better way.

```
def get_std_name_vars(ds,std_name):
    return {k: v for k, v in ds.data_vars.iteritems() if 'standard_name' in v.attrs.keys() and std_name in v.standard_name}
```
as in this example: 
http://nbviewer.ipython.org/gist/rsignell-usgs/5b263906e92ce47bf05e

shoyer:
I would probably make this return an xray.Dataset object instead of a plain dict, but otherwise this looks about right:
```python
def get_std_name_vars(ds,std_name):
    return ds[[k for k, v in ds.data_vars.iteritems() if 'standard_name' in v.attrs and std_name in v.standard_name]]
```

I don't think there's a better way to do this currently. Hypothetically, we could add this to a CF-specific API for xray, e.g., as discussed in #461.

rsignell-usgs:
I was thinking that the data variables that matched a specified `standard_name` would be a subset of the variables in the `data_vars` object.

ocefpaf:
How about the function below?  (Stolen from [pyaxiom](https://github.com/axiom-data-science/pyaxiom/blob/master/pyaxiom/netcdf/dataset.py#L10-L47).)

```python
def get_variables_by_attributes(ds, strict=False, **kwargs):
    """
    Returns variables that match specific conditions.
    * Can pass in key=value parameters and variables are returned that
      contain all of the matches.
      ex.:
          vs = nc.get_variables_by_attributes(axis='X')
    * Can pass in key=callable parameter and if the callable returns
      True.  The callable should accept a single parameter, the attribute
      value.  None is returned as the attribute valuewhen the attribute
      does not exist on the variable.
      ex.:
          # Get Axis variables
          vs = nc.get_variables_by_attributes(axis=lambda v: v in ['X', 'Y', 'Z', 'T'])
          # Get variable that don't have a "axis" attribute
          vs = nc.get_variables_by_attributes(axis=lambda v: v is None)
          # Get variable that have a "grid_mapping" attribute
          vs = nc.get_variables_by_attributes(axis=lambda v: v is not None)
    
    * strict : True/False
        If True will return only 1 variable if only one is found. Default is False.
    """
    vs = []

    has_value_flag  = False
    for vname, var in ds.iteritems():
        for k, v in kwargs.items():
[...]

PR #844, opened 2016-05-04
Add a filter_by_attrs method to Dataset

ocefpaf:
This PR adds the `get_variables_by_attributes` method similar to the one in `netCDF4-python` and netcdf-java libraries. It is useful to filter a Dataset to known/expected attributes.

@shoyer I don't really like the docs nor the change log entry I created. I will look at them again tomorrow with fresh eyes to see if I can improve them.

Closes https://github.com/pydata/xarray/issues/567

xref: https://github.com/pydata/xarray/issues/567#issuecomment-216947679

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2224,6 +2224,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
 
         Examples
         --------
+        # "Create an example dataset:
         >>> import numpy as np
         >>> import pandas as pd
         >>> import xarray as xr
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2250,25 +2250,13 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
             *empty*
 
         """
-        vs = []
-
-        has_value_flag  = False
-        for vname, data_array in self.items():
-            for k, v in kwargs.items():
-                if callable(v):
-                    has_value_flag = v(getattr(data_array, k, None))
-                    if has_value_flag is False:
-                        break
-                elif hasattr(data_array, k) and getattr(data_array, k) == v:
-                    has_value_flag = True
-                else:
-                    has_value_flag = False
-                    break
-
-            if has_value_flag is True:
-                vs.append(vname)
-        if len(self) < len(vs) > 1:
-            vs = [vs]
-        return self[vs]
+        selection = []
+        for var_name, variable in self.items():
+            for attr_name, pattern in kwargs.items():
+                attr_value = variable.attrs.get(attr_name)
+                if ((callable(pattern) and pattern(attr_value))
+                        or attr_value == pattern):
+                    selection.append(var_name)
+        return self[selection]
 
 ops.inject_all_ops_and_reduce_methods(Dataset, array_only=False)
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2200,6 +2200,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
 
     def get_variables_by_attributes(self, **kwargs):
         """Returns a ``Dataset`` with variables that match specific conditions.
+
         Can pass in ``key=value ``or ``key=callable``.  Variables are returned
         that contain all of the matches or callable returns True.  If using a
         callable note that it should accept a single parameter only,
@@ -2209,13 +2210,17 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         ----------
         **kwargs : key=value
             key : str
-                Filter variables based on a string attribute.
-            value : str or callable
-                Filter variables based on callable False/True answer.
+                Attribute name.
+            value : callable or obj
+                If value is a callable, it should return a boolean in the form
+                of bool = func(attr) where attr is da.attrs[key].
+                Otherwise, value will be compared to the each
+                DataArray's attrs[key].
 
         Returns
         -------
         new : Dataset
+            New dataset with variables filtered by attribute.
 
         Examples
         --------
@@ -2226,15 +2231,20 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         >>> precip = 10 * np.random.rand(2, 2, 3)
         >>> lon = [[-99.83, -99.32], [-99.79, -99.23]]
         >>> lat = [[42.25, 42.21], [42.63, 42.59]]
-
-        >>> ds = xr.Dataset({'temperature': (['x', 'y', 'time'],  temp, dict(standard_name='air_potential_temperature')),
-        ...                  'precipitation': (['x', 'y', 'time'], precip, dict(standard_name='convective_precipitation_flux'))},
-        ...                 coords={'lon': (['x', 'y'], lon, dict(axis='X')),
-        ...                         'lat': (['x', 'y'], lat, dict(axis='Y')),
-        ...                         'time': pd.date_range('2014-09-06', periods=3),
-        ...                         'reference_time': pd.Timestamp('2014-09-05')})
+        >>> dims = ['x', 'y', 'time']
+        >>> temp_attr = dict(standard_name='air_potential_temperature')
+        >>> precip_attr = dict(standard_name='convective_precipitation_flux')
+        >>> ds = xr.Dataset({
+        ...         'temperature': (dims,  temp, temp_attr),
+        ...         'precipitation': (dims, precip, precip_attr)},
+        ...                 coords={
+        ...         'lon': (['x', 'y'], lon),
+        ...         'lat': (['x', 'y'], lat),
+        ...         'time': pd.date_range('2014-09-06', periods=3),
+        ...         'reference_time': pd.Timestamp('2014-09-05')})
         >>> # Get variables matching a specific standard_name.
-        >>> ds.get_variables_by_attributes(standard_name='convective_precipitation_flux')
+        >>> ds.get_variables_by_attributes(
+        ...     standard_name='convective_precipitation_flux')
         <xarray.Dataset>
         Dimensions:         (time: 3, x: 2, y: 2)
         Coordinates:
@@ -2247,7 +2257,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         Data variables:
             precipitation   (x, y, time) float64 4.178 2.307 6.041 6.046 0.06648 ...
         >>> # Get all variables that have a standard_name attribute.
-        >>> ds.get_variables_by_attributes(standard_name=lambda v: v is not None)
+        >>> standard_name = lambda v: v is not None
+        >>> ds.get_variables_by_attributes(standard_name=standard_name)
         <xarray.Dataset>
         Dimensions:         (time: 3, x: 2, y: 2)
         Coordinates:
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2203,8 +2203,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         Can pass in ``key=value ``or ``key=callable``.  Variables are returned
         that contain all of the matches or callable returns True.  If using a
         callable note that it should accept a single parameter only,
-        the attribute value.  None is given as the attribute value when the
-        attribute does not exist on the variable.
+        the attribute value.
 
         Examples
         --------
@@ -2258,9 +2257,6 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
                 if ((callable(pattern) and pattern(attr_value))
                         or attr_value == pattern):
                     selection.append(var_name)
-        if selection:
-            return self[selection]
-        else:
-            return None
+        return self[selection]
 
 ops.inject_all_ops_and_reduce_methods(Dataset, array_only=False)
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2248,10 +2248,10 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         <xarray.Dataset>
         Dimensions:         (time: 3, x: 2, y: 2)
         Coordinates:
-        * x               (x) int64 0 1
-        * time            (time) datetime64[ns] 2014-09-06 2014-09-07 2014-09-08
+          * x               (x) int64 0 1
+          * time            (time) datetime64[ns] 2014-09-06 2014-09-07 2014-09-08
             lat             (x, y) float64 42.25 42.21 42.63 42.59
-        * y               (y) int64 0 1
+          * y               (y) int64 0 1
             reference_time  datetime64[ns] 2014-09-05
             lon             (x, y) float64 -99.83 -99.32 -99.79 -99.23
         Data variables:
@@ -2264,9 +2264,9 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
         Coordinates:
             lon             (x, y) float64 -99.83 -99.32 -99.79 -99.23
             lat             (x, y) float64 42.25 42.21 42.63 42.59
-        * x               (x) int64 0 1
-        * y               (y) int64 0 1
-        * time            (time) datetime64[ns] 2014-09-06 2014-09-07 2014-09-08
+          * x               (x) int64 0 1
+          * y               (y) int64 0 1
+          * time            (time) datetime64[ns] 2014-09-06 2014-09-07 2014-09-08
             reference_time  datetime64[ns] 2014-09-05
         Data variables:
             temperature     (x, y, time) float64 25.86 20.82 6.954 23.13 10.25 11.68 ...
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2198,5 +2198,77 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject):
     def imag(self):
         return self._unary_op(lambda x: x.imag, keep_attrs=True)(self)
 
+    def get_variables_by_attributes(self, **kwargs):
+        """Returns a ``Dataset`` with variables that match specific conditions.
+        Can pass in ``key=value ``or ``key=callable``.  Variables are returned
+        that contain all of the matches or callable returns True.  If using a
+        callable note that it should accept a single parameter only,
+        the attribute value.  None is given as the attribute value when the
+        attribute does not exist on the variable.
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> import pandas as pd
+        >>> import xarray as xr
+        >>> temp = 15 + 8 * np.random.randn(2, 2, 3)
+        >>> precip = 10 * np.random.rand(2, 2, 3)
+        >>> lon = [[-99.83, -99.32], [-99.79, -99.23]]
+        >>> lat = [[42.25, 42.21], [42.63, 42.59]]
+
+        >>> ds = xr.Dataset({'temperature': (['x', 'y', 'time'],  temp, dict(standard_name='air_potential_temperature')),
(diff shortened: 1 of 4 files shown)


xarray/core/dataset.py
```diff
@@ -2179,5 +2179,77 @@ def real(self):
...
+                        break
+                elif hasattr(data_array, k) and getattr(data_array, k) == v:
+                    has_value_flag = True
+                else:
+                    has_value_flag = False
+                    break
+
+            if has_value_flag is True:
+                vs.append(vname)
+        if len(self) < len(vs) > 1:
```
ocefpaf: This was my trick to always return a `Dataset` even when it finds one, all, a subset, or none variables. I am not familiar with `xarray` internals, so maybe there is a better way.

xarray/test/test_dataset.py
```diff
@@ -2494,3 +2494,45 @@ def test_setattr_raises(self):
             ds.foo = 2
         with self.assertRaisesRegexp(AttributeError, 'cannot set attr'):
             ds.other = 2
+
+    def test_get_variables_by_attributes(self):
```
ocefpaf: I am not very creative with the tests. Basically just example cases and not really unittests.  Let me know if that is OK.

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,77 @@ def real(self):
...
+        * y               (y) int64 0 1
+            reference_time  datetime64[ns] 2014-09-05
+        Data variables:
+            *empty*
+
+        """
+        vs = []
+
+        has_value_flag  = False
+        for vname, data_array in self.items():
```
shoyer: Can we rewrite this loop not to use `break` for control flow? Something like this is much more comprehensible to me:
```
selection = []
for var_name, variable in self.data_vars.items():
    for attr_name, pattern in kwargs.item():
        attr_value = array.attrs.get(attr_name)
        if (callable(pattern) and pattern(attr_value)
                or attr_value == pattern):
            selection.append(var_name)
```

xarray/core/dataset.py
shoyer: This should not be necessary -- `vs` (or `selection`) in my example will always be a list.

xarray/test/test_dataset.py
shoyer: This is fine. But because these variables don't depend on the shape of the data, can you use simpler data? Maybe all zeros?

shoyer:
An important design question: should this filter both data variables and coordinates or only data variables? My thought is that it's only worth filtering data variables -- filtering out unmatched coordinates is not very useful.

xarray/core/dataset.py
ocefpaf: Sure. That does read much better than the original.

xarray/core/dataset.py
ocefpaf: Yep. I am happy to see my ugly workaround go away :smile:

xarray/test/test_dataset.py
ocefpaf: OK. That was a terrible copy-and-paste from a notebook example/docs/tests. It deserves more thinking and simplification.

ocefpaf:
> An important design question: should this filter both data variables and coordinates or only data variables? My thought is that it's only worth filtering data variables -- filtering out unmatched coordinates is not very useful.

I understand that returning coordinates without a data variable associated to them seems weird to the high level model of `xarray.Dataset` , but I disagree that it is not very useful. In fact that is the most common operation I do: find coordinates based on `axis`, `formula_terms`, etc and construct common grids for plotting and/or building the `z` coords from non-dimension coords.

ocefpaf:
> should this filter both data variables and coordinates or only data variables?

I thought a little bit more about this and now I am on the fence. The pros of filtering only data variables are a nice and clean Dataset object, and overall consistency with the high level xarray model. The cons are that we cannot do the filtering on the coords (obviously), but most of the time that we need to do that we go to a lower level object like `netCDF4.Dataset`. However, it would be nice if both `xarray` and netCDF4 behaved the same way...

I am 51% with filtering both (and the current implementation does that) but I will leave the final decision to you.

ocefpaf:
After discussing with my CF guru (@rsignell-usgs) and the original author of the `get_variables_by_attributes` (@kwilcox) I jumped the fence and I am OK leaving the xarray model pure and filtering only the data variables.

kwilcox:
@ocefpaf I agree that xarray model consistent should take precedence. I also use this method to (mostly) pull out coordinate variables and can continue to use netCDF4.Dataset to do that.

shoyer:
Well, if you guys are mostly excited about using this for coordinate
variables, another consistent choice would be to return a list of matching
DataArrays. But if we want to return a Dataset, we should only do data
variables, because it's weird to lose all the describing coordinates when
you match, e.g., standard_name="air_temperature".

ocefpaf:
> Well, if you guys are mostly excited about using this for coordinate
variables, another consistent choice would be to return a list of matching
DataArrays. But if we want to return a Dataset, we should only do data
variables, because it's weird to lose all the describing coordinates when
you match, e.g., standard_name="air_temperature".

We agree with you and I prefer to return a Dataset.

Right now we always go to netCDF4-python to do this low-level CF interpretation stuff. If we start using `xarray` for that we should improve xarray, to take advantage of the conventions (CF/SGRID/UGRID), instead of using `xarray` to just find the coords data. (For example: create the `z` coordinates from non-dimension coordinates and add that to the Dataset coords automatically.)

rsignell-usgs:
It also seems consistent to me to return a Dataset.

ocefpaf:
@shoyer this is ready for another round of review.

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,69 @@ def real(self):
...
+        for var_name, variable in self.data_vars.items():
+            for attr_name, pattern in kwargs.items():
+                attr_value = variable.attrs.get(attr_name)
+                if ((callable(pattern) and pattern(attr_value))
+                        or attr_value == pattern):
+                    selection.append(var_name)
+        if selection:
+            return self[selection]
+        else:
+            return None
```
MaximilianR: Do we want really want to return different types depending on the results? I'd generally vote to just return `self[selection]`, and if selection is empty, then it returns an empty ds

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,69 @@ def real(self):
...
+            reference_time  datetime64[ns] 2014-09-05
+        Data variables:
+            temperature     (x, y, time) float64 25.86 20.82 6.954 23.13 10.25 11.68 ...
+            precipitation   (x, y, time) float64 5.702 0.9422 2.075 1.178 3.284 ...
+
+        """
+        selection = []
+        for var_name, variable in self.data_vars.items():
+            for attr_name, pattern in kwargs.items():
+                attr_value = variable.attrs.get(attr_name)
```
MaximilianR: Should this raise if there's no attribute by that name? I think currently this requires every variable to have an attribute in order to search. 
Changing this to `variable.attrs.get(attr_name, None)` (or check if it exists first if we want to search for `None` values) will be a bit less strict, I think

xarray/core/dataset.py
ocefpaf: Make sense. Adding the changes now.

xarray/core/dataset.py
ocefpaf: No need. `None` is already the default there.

xarray/test/test_dataset.py
```diff
@@ -2494,3 +2494,36 @@ def test_setattr_raises(self):
...
+        ds = Dataset({'temperature_0': (['t'],  [0], dict(standard_name='air_potential_temperature', height='0 m')),
+                      'temperature_10': (['t'],  [0], dict(standard_name='air_potential_temperature', height='10 m')),
+                      'precipitation': (['t'], [0], dict(standard_name='convective_precipitation_flux'))},
+                    coords={'time': (['t'], [0], dict(axis='T'))})
+
+        # Test return empty Dataset.
+        ds.get_variables_by_attributes(standard_name='invalid_standard_name')
+        new_ds = ds.get_variables_by_attributes(standard_name='invalid_standard_name')
+        self.assertFalse(bool(new_ds))
```
ocefpaf: @MaximilianR is this OK to test for am empty `Dataset`? Pandas has the `.empty` property, but I could not find am equivalent on `xarray`.

xarray/core/dataset.py
MaximilianR: Great. My mistake

xarray/test/test_dataset.py
MaximilianR: I think so!

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,65 @@ def real(self):
     def imag(self):
         return self._unary_op(lambda x: x.imag, keep_attrs=True)(self)
 
+    def get_variables_by_attributes(self, **kwargs):
+        """Returns a ``Dataset`` with variables that match specific conditions.
+        Can pass in ``key=value ``or ``key=callable``.  Variables are returned
+        that contain all of the matches or callable returns True.  If using a
+        callable note that it should accept a single parameter only,
+        the attribute value.
+
```
jhamman: I'd like to see a more formal docstring here, more like: 

```Python

    def get_variables_by_attributes(self, **kwargs):
        """Summary...

        Parameters
        ----------
        **kwargs : key=value
            key : str
                Description...
            value : str or callable
                Description...
        Returns
        -------
        new : Dataset
       
        Examples
        --------
        ...
    """"
```
This is similar to what we did in `da.rolling`: https://github.com/pydata/xarray/blob/master/xarray/core/rolling.py#L36-L41

```

File: xarray/test/test_dataset.py [hunk 8]
```diff
@@ -2494,3 +2494,36 @@ def test_setattr_raises(self):
             ds.foo = 2
         with self.assertRaisesRegexp(AttributeError, 'cannot set attr'):
             ds.other = 2
+
+    def test_get_variables_by_attributes(self):
+        ds = Dataset({'temperature_0': (['t'],  [0], dict(standard_name='air_potential_temperature', height='0 m')),
+                      'temperature_10': (['t'],  [0], dict(standard_name='air_potential_temperature', height='10 m')),
+                      'precipitation': (['t'], [0], dict(standard_name='convective_precipitation_flux'))},
+                    coords={'time': (['t'], [0], dict(axis='T'))})
```
jhamman: nit: can you fix the styling here? These lines are pretty long and there are some extra spaces before the `[0]`s.

xarray/test/test_dataset.py
```diff
@@ -2494,3 +2494,36 @@ def test_setattr_raises(self):
...
+        ds.get_variables_by_attributes(standard_name='invalid_standard_name')
+        new_ds = ds.get_variables_by_attributes(standard_name='invalid_standard_name')
+        self.assertFalse(bool(new_ds))
+
+        # Test return one Dataset.
+        new_ds = ds.get_variables_by_attributes(standard_name='convective_precipitation_flux')
+        self.assertEqual(new_ds['precipitation'].standard_name, 'convective_precipitation_flux')
+        self.assertDatasetEqual(new_ds['precipitation'], ds['precipitation'])
+
+        # Test return more than one Dataset.
```
jhamman: `# Test return Dataset with more than one DataArray`

xarray/test/test_dataset.py
```diff
@@ -2494,3 +2494,36 @@ def test_setattr_raises(self):
...
+                      'temperature_10': (['t'],  [0], dict(standard_name='air_potential_temperature', height='10 m')),
+                      'precipitation': (['t'], [0], dict(standard_name='convective_precipitation_flux'))},
+                    coords={'time': (['t'], [0], dict(axis='T'))})
+
+        # Test return empty Dataset.
+        ds.get_variables_by_attributes(standard_name='invalid_standard_name')
+        new_ds = ds.get_variables_by_attributes(standard_name='invalid_standard_name')
+        self.assertFalse(bool(new_ds))
+
+        # Test return one Dataset.
```
jhamman: Dataset --> DataArray

xarray/core/dataset.py
ocefpaf: @jhamman can you take a second look? (Docs is not my forte even in my first language :stuck_out_tongue_winking_eye:)

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,77 @@ def real(self):
...
+        **kwargs : key=value
+            key : str
+                Filter variables based on a string attribute.
+            value : str or callable
+                Filter variables based on callable False/True answer.
+
+        Returns
+        -------
+        new : Dataset
+
```
jhamman: I'll suggest a few more modifications:

```Python
    """
    ...
        Parameters
        ----------
        **kwargs : key=value
            key : str
                Attribute name.
            value : callable or obj
                If value is a callable, it should return a boolean in the form of bool = func(attr) where attr is da.attrs[key].  
                Otherwise, value will be compared to the each DataArray's attrs[key].

        Returns
        -------
        new : Dataset
            New dataset with variables filtered by attribute.

    ...
    """
```

I think the `value` section could use a bit more thought.

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,77 @@ def real(self):
     def imag(self):
         return self._unary_op(lambda x: x.imag, keep_attrs=True)(self)
 
+    def get_variables_by_attributes(self, **kwargs):
+        """Returns a ``Dataset`` with variables that match specific conditions.
+        Can pass in ``key=value ``or ``key=callable``.  Variables are returned
```
shoyer: add an extra line break after the first line.

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,77 @@ def real(self):
...
+        --------
+        >>> import numpy as np
+        >>> import pandas as pd
+        >>> import xarray as xr
+        >>> temp = 15 + 8 * np.random.randn(2, 2, 3)
+        >>> precip = 10 * np.random.rand(2, 2, 3)
+        >>> lon = [[-99.83, -99.32], [-99.79, -99.23]]
+        >>> lat = [[42.25, 42.21], [42.63, 42.59]]
+
+        >>> ds = xr.Dataset({'temperature': (['x', 'y', 'time'],  temp, dict(standard_name='air_potential_temperature')),
```
shoyer: This example needs to fit within the 80 character line limit :)

xarray/test/test_dataset.py
shoyer: I would test `bool(new_ds.data_vars)` rather than `bool(new_ds)`. The later works here only because you have no scalar coordinates.

ocefpaf: Sure. Makes perfect sense. I don't have the `Dataset` and `DataArray` concept in my blood stream yet, but I am getting there :grinning:

ocefpaf:
@jhamman and @shoyer if the tests passes this is ready for another round of review. (Let me know if I should squash the previous ones to make it easier to review.)

shoyer:
@ocefpaf We can squash on merge now in the web interface, so that makes things easier.

The appveyor build failure looks unrelated to this change -- something about conda dependencies.

ocefpaf:
> The appveyor build failure looks unrelated to this change -- something about conda dependencies.

I am experiencing that in other projects. It is actually a bad download of `miniconda` and powershell makes the error message unusable. Re-starting should fix it. (BTW you can miniconda pre-installed on AppVeyor and that download is not needed.)

ocefpaf:
Rebased and ready for another round of reviews :wink:

jhamman:
This looks good to me.  Do we want to add this to the api page in the docs?

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,88 @@ def real(self):
...
+        ...         'lat': (['x', 'y'], lat),
+        ...         'time': pd.date_range('2014-09-06', periods=3),
+        ...         'reference_time': pd.Timestamp('2014-09-05')})
+        >>> # Get variables matching a specific standard_name.
+        >>> ds.get_variables_by_attributes(
+        ...     standard_name='convective_precipitation_flux')
+        <xarray.Dataset>
+        Dimensions:         (time: 3, x: 2, y: 2)
+        Coordinates:
+        * x               (x) int64 0 1
```
shoyer: the indentation on this line (and the similar ones below) is off

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,88 @@ def real(self):
...
+        * time            (time) datetime64[ns] 2014-09-06 2014-09-07 2014-09-08
+            lat             (x, y) float64 42.25 42.21 42.63 42.59
```
shoyer: or maybe this is the line with the bad indentation. either way please fix :)

xarray/core/dataset.py
```diff
@@ -2179,5 +2179,88 @@ def real(self):
...
+                DataArray's attrs[key].
+
+        Returns
+        -------
+        new : Dataset
+            New dataset with variables filtered by attribute.
+
+        Examples
+        --------
+        >>> import numpy as np
```
shoyer: Insert something above here like "Create an example dataset:"

shoyer:
Agreed -- this is almost ready. Please add to the API docs (`api.rst`) and do the docstring fixes.

xarray/core/dataset.py
ocefpaf: I believe I got it correct now. Let me know if this is OK.

ocefpaf:
> Agreed -- this is almost ready. Please add to the API docs (api.rst) and do the docstring fixes.

Not sure if I did this right :grimacing:

ocefpaf:
Is there still an interested in this or should I close in light of https://github.com/pydata/xarray/issues/883?

shoyer:
@ocefpaf Good point. It probably is indeed better to write be able to write something generic like `ds.filter(lambda x: x.attrs['standard_name'] == 'convective_precipitation_flux')` rather the specialized to attributes `ds.get_variables_by_attributes(standard_name='convective_precipitation_flux')`.

Another option would be simply renaming this to `filter_attrs`, which is more succinct.

ocefpaf:
I fine with whatever you decide but here are my two cents:

- `get_variables_by_attributes` is the same name of this method in `netcd4` and some java `netcdf` libraries. So I'd rather not have a specialized version for attributes than adding it with name with a different name.
- I see the elegance in `s.filter(lambda x: x.attrs['standard_name'] == 'convective_precipitation_flux')` and I like it a lot! But the specialized version for attributes is more compact to write and, at least in my field, filtering by attributes is more common making this version more convenient.

Feel free to close this if you don't think it is worth adding the specialized version. Or let me know if you want to rename it.

shoyer:
I'm OK with the specialized `get_variables_by_attributes`. It doesn't need to be mutually exclusive with the more generic option. I do think the name is a mouthful, though :).

ocefpaf:
>  I do think the name is a mouthful, though :).

I agree that is a mouthful but then again, I was trying to be consistent with the existing versions. However, I don't really have a strong opinion on the matter and I am fine with whatever you decide. Should I renamed then? `filter_attrs`? (Sounds weird though b/c you are not filtering the `attrs`, but the variables based on the `attrs` :expressionless:)

shoyer:
Assuming we do add the `filter` method, maybe `filter_by_attrs`?

It's nice to use names with the same prefix because they appear next to each other in auto-complete.

ocefpaf:
> Assuming we do add the `filter` method, maybe `filter_by_attrs`?

Makes sense. I will modify this soon.

ocefpaf:
Done. Not sure why only `AppVeyor` started :confused:

shoyer:
Anyone else have an opinion on the name? I'd like to merge this shortly before releasing v0.8.0 tomorrow.

fmaussion:
:+1:  for `filter_by_attrs` , and I find it good to have such convenience function in addition to the general ones like https://github.com/pydata/xarray/issues/916 : newcomers will find their way through xarray more easily

shoyer:
OK, in it goes. Thanks @ocefpaf !

PR #401, opened 2015-04-24
Handle bool in NetCDF4 conversion

khaeru:
I am working on some code that creates `xray.Datasets` with a 'bool' dtype.

Trying to call `Dataset.to_netcdf()` on this code causes `_nc4_values_and_dtype()` to raise a `ValueError`, so I added these few lines to force the storage of these variables as 1-byte integers.

Perhaps it should be 'u1' instead; I can change that if need be.

PR #849, opened 2016-05-17
roundtrip boolean datatype

jhamman:
This PR allows boolean datatypes to be faithfully roundtripped.  My approach follows @shoyer's suggestions in https://github.com/pydata/xarray/pull/401#issuecomment-96085330.

closes #401 
cc @klapo, @khaeru

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -37,6 +37,11 @@ Enhancements
   extensions without subclassing. They are described in the new documentation
   page on :ref:`internals`. By `Stephan Hoyer <https://github.com/shoyer>`
 
+- Round trip boolean datatypes. Previously, writing boolean datatypes to netCDF
+  formats would raise an error since netCDF does not have a `bool` datatype.
+  This feature reads/writes a `dtype` attribute to boolean variables in netCDF
+  files. By `Joe Hamman <https://github.com/jhamman>`_.
+
 Bug fixes
 ~~~~~~~~~
 
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -11,8 +11,7 @@ from pandas.tslib import OutOfBoundsDatetime
 from .core import indexing, ops, utils
 from .core.formatting import format_timestamp, first_n_items
 from .core.variable import as_variable, Variable
-from .core.pycompat import (iteritems, bytes_type, unicode_type, OrderedDict,
-                            PY3, basestring)
+from .core.pycompat import iteritems, OrderedDict, PY3, basestring
 
 
 # standard calendars recognized by netcdftime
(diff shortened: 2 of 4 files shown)


xarray/conventions.py
```diff
@@ -776,6 +815,10 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
     else:
         encoding['dtype'] = original_dtype
 
+    if 'dtype' in attributes and attributes['dtype'] == 'bool':
+        del attributes['dtype']
```
jhamman: @shoyer - do you think it makes since to throw this attribute away? The roundtrip fails in its current state if I move it to encoding or leave it where it is.

doc/whats-new.rst
```diff
@@ -37,6 +37,11 @@ Enhancements
   extensions without subclassing. They are described in the new documentation
   page on :ref:`internals`. By `Stephan Hoyer <https://github.com/shoyer>`
 
+- Round trip boolean datatypes. Previously, writing boolean datatypes to netCDF
+  formats would raise an error since netCDF does not have a `bool` datatype.
+  This feature reads/writes a `dtype` attribute to boolean variables in netCDF
+  files. By `Joe Hamman <https://github.com/jhamman>`_.
+
```
jhamman: Do we also need to add some documentation of this feature to `io.rst`?

jhamman:
Note that the failing test appears to be a anaconda issue on [Appveyor](https://ci.appveyor.com/project/shoyer/xray/build/1.0.1078/job/4gosksjqt4utb6c8#L409) and is not related to this PR.

xarray/conventions.py
```diff
@@ -602,6 +629,17 @@ def maybe_encode_dtype(var, name=None):
     return var
 
 
+def maybe_encode_bools(var, needs_copy=True):
+    if ((var.dtype == np.bool) and
+            ('dtype' not in var.encoding) and ('dtype' not in var.attrs)):
+        dims, data, attrs, encoding = _var_as_tuple(var)
+        attrs['dtype'] = 'bool'
+        data = data.astype(dtype='i1', copy=needs_copy)
```
shoyer: It looks like casting to `i1` always makes a copy, even though I don't think it's strictly necessary. So let's just drop the needs_copy business here and always copy:
```
In [54]: x
Out[54]: array([ True, False,  True], dtype=bool)

In [55]: y
Out[55]: array([1, 0, 1], dtype=int8)

In [56]: x[0] = False

In [57]: y
Out[57]: array([1, 0, 1], dtype=int8)
```

xarray/conventions.py
shoyer: Yes, throw it away.

shoyer:
One annoying potential inconsistency here is that it *is* possible to write boolean dtypes using h5py (which uses HDF5 enums), which is different from how we do things here.

jhamman:
> One annoying potential inconsistency here is that it is possible to write boolean dtypes using h5py (which uses HDF5 enums), which is different from how we do things here.

okay, so do we want/need to insert some logic into the `h5py` backend that handles the boolean type differently. I suppose this only matters for the write, reading from boolean (enum) types should be unaffected by this change.

shoyer:
I think it's fine to write all booleans data using this encoding. It's pretty obvious how it works, and it's compatible with all versions of netCDf.

Unfortunately, netCDF-C libraries can't read variables with the enum dtype created by h5py for boolean data (https://github.com/Unidata/netcdf-c/issues/267), so I don't think we should encourage that option.

shoyer:
I think this is good to merge. Appveyor is definitely doing something strange here with the Python 3 build.

PR #852, opened 2016-05-21
BUG: Stringify DS attrs in dir

MaximilianR:
Previously:

```python
In [21]: ds
Out[21]: 
<xarray.Dataset>
Dimensions:  (abc: 3)
Coordinates:
  * abc      (abc) int64 0 1 2
Data variables:
    0        (abc) float64 0.3168 0.1132 0.6664
    1        (abc) float64 0.9799 0.9663 0.01084
    2        (abc) float64 0.9171 0.6848 0.5028


In [22]: dir(ds)
<ipython-input-22-fd55ce5c03db> in <module>()
----> 1 dir(ds)

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/common.py in __dir__(self)
    157         extra_attrs = [item for sublist in self._attr_sources
    158                        for item in sublist]
--> 159         return sorted(set(dir(type(self)) + extra_attrs))
    160 
    161 

TypeError: unorderable types: numpy.ndarray() > str()
```

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -26,6 +26,10 @@ Bug fixes
 
 - Fix to ensure xarray works with h5netcdf v0.3.0 for arrays with ``dtype=str``
   (:issue:`953`). By `Stephan Hoyer <https://github.com/shoyer>`_.
+- ``Dataset.__dir__()`` (i.e. the method python calls to get autocomplete options) failed
+  if one of the ``Dataset``'s keys was not a string (:issue:`852`). By
+  `Maximilian Roos <https://github.com/maximilianr>`_.
+
 
 .. _whats-new.0.8.1:
 
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pandas as pd
 
-from .pycompat import basestring, iteritems, suppress, dask_array_type
+from .pycompat import basestring, iteritems, suppress, dask_array_type, bytes_type
 from . import formatting
 from .utils import SortedKeysDict, not_implemented
 
@@ -211,8 +211,9 @@ class AttrAccessMixin(object):
         """Provide method name lookup and completion. Only provide 'public'
         methods.
         """
-        extra_attrs = [item for sublist in self._attr_sources
-                       for item in sublist]
+        extra_attrs = [
+            item for sublist in self._attr_sources for item in sublist
+            if isinstance(item, (basestring))]
         return sorted(set(dir(type(self)) + extra_attrs))
 
 
--- a/xarray/test/test_dataset.py
+++ b/xarray/test/test_dataset.py
@@ -12,6 +12,7 @@ except ImportError:
 
 import numpy as np
 import pandas as pd
+import pytest
 
 from xarray import (align, broadcast, concat, merge, conventions, backends,
                     Dataset, DataArray, Variable, Coordinate, auto_combine,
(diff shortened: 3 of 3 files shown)


MaximilianR:
AppVeyor break is an existing issue I think?

xarray/test/test_dataset.py
```diff
@@ -323,6 +323,16 @@ def test_attr_access(self):
         self.assertEqual(ds.attrs['tmin'], -999)
         self.assertDataArrayIdentical(ds.tmin, ds['tmin'])
 
+    def test_dir(self):
+        ds = create_test_data()
+        # add a numbered key to ensure this doesn't break dir
+        ds[5] = list(ds.keys())[0]
+
+        some_expected_attrs = {'pipe', 'mean', 'isnull', 'var1',
+                               'dim1', 'numbers'}
```
jhamman: this breaks the Python 2.6 test on travis.  Use a list or a tuple.

MaximilianR: We still support 2.6? How about dropping that...

shoyer: The plan (per mailing list discussino) is to drop Python 2.6 support for the next major release. That should probably be the next release, but we'll need to cut it out of the Travis-CI setup first.

xarray/test/test_dataset.py
```diff
@@ -323,6 +323,16 @@ def test_attr_access(self):
         self.assertEqual(ds.attrs['tmin'], -999)
         self.assertDataArrayIdentical(ds.tmin, ds['tmin'])
 
+    def test_dir(self):
+        ds = create_test_data()
+        # add a numbered key to ensure this doesn't break dir
+        ds[5] = list(ds.keys())[0]
```
shoyer: Can you simply create the key directly, e.g., `ds[5] = 'foo'`? It's not clear to me why you're using indexing here.

shoyer:
this looks good, once you fix a few quick nits

xarray/test/test_dataset.py
jhamman: see #855 and #856 for issue/PR on removing Python 2.6 support.

shoyer:
@MaximilianR Can you rebase this? Otherwise this looks good to go to me.

doc/whats-new.rst
```diff
@@ -51,6 +51,10 @@ Bug fixes
    This fixes issue :issue:`665`.
    `Filipe Fernandes <https://github.com/ocefpaf>`_.
 
+- ``Dataset.__dir__()`` (the method python calls to get autocomplete options) failed
+  if one of the ``DataSet``'s keys was not a string (:issue:`852`).
```
shoyer: don't forget to give yourself credit :)

MaximilianR:
I've been remiss about the final fixes here. Will need a couple of weeks on other commitments but I will get to this

MaximilianR:
@shoyer @jhamman updated, but travis queue looks stopped?

doc/whats-new.rst
```diff
@@ -24,6 +24,11 @@ Bug fixes
 - Fix regression in v0.8.0 that broke assignment to Datasets with non-unique
   indexes (:issue:`943`).
 
+- ``Dataset.__dir__()`` (i.e. the method python calls to get autocomplete options) failed
+  if one of the ``DataSet``'s keys was not a string (:issue:`852`). By
```
jhamman: `DataSet` --> `Dataset`

MaximilianR:
Rebased @shoyer @jhamman

shoyer:
Looking at this again -- maybe better just to drop non-strings from `__dir__`?

For example, if you put `None` is an `xr.Dataset`, this would give you autocomplete for `ds.None`, which won't actually work when it's mapped to `ds['None']`.

MaximilianR:
> Looking at this again -- maybe better just to drop non-strings from __dir__?

Yes!

xarray/core/common.py
```diff
@@ -212,7 +212,7 @@ def __dir__(self):
         methods.
         """
         extra_attrs = [item for sublist in self._attr_sources
-                       for item in sublist]
+                       for item in sublist if isinstance(item, str)]
```
shoyer: This should be `pycompat.basestring` -- otherwise I think this would break for unicode variable names.

MaximilianR: Great. Not putting `bytes` there as I get `E       TypeError: unorderable types: bytes() < str()`. Let me know if that's not OK though

shoyer: basestring is a super class of both bytes and str on Python 2. On Python 3, it's only str.

shoyer:
Merged in 15b4291bd9e02440a5bd5346b55a8f6bee939690

doc/whats-new.rst
```diff
@@ -26,6 +26,10 @@ Bug fixes
 
 - Fix to ensure xarray works with h5netcdf v0.3.0 for arrays with ``dtype=str``
   (:issue:`953`). By `Stephan Hoyer <https://github.com/shoyer>`_.
+- ``Dataset.__dir__()`` (i.e. the method python calls to get autocomplete options) failed
+  if one of the ``Dataset``'s keys was not a string (:issue:`852`). By
```
shoyer: Just a note, in restructured text this breaks:

    ``Dataset``'s

You need whitespace or punctuation after the backticks.

Issue #847, opened 2016-05-13
Dataset constructor fails if values are objects


MaximilianR:
For an item to be a value in a Dataset, it either needs to be array-like, or pass `lib.is_scalar`. 

I think this is probably too strict - I'd propose anything can be a value, and it only gets treated as an array if it looks like one - thoughts?

Here's the code that checks the values: https://github.com/pydata/xarray/blob/master/xarray/core/variable.py#L52

```python
In [13]: class B(object):
    pass
   ....: 

In [14]: xr.Dataset({'a':B()})
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-b2eeb6de19ea> in <module>()
----> 1 xr.Dataset({'a':B()})

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/dataset.py in __init__(self, data_vars, coords, attrs, compat, **kwargs)
    207             coords = set()
    208         if data_vars is not None or coords is not None:
--> 209             self._set_init_vars_and_dims(data_vars, coords, compat)
    210         if attrs is not None:
    211             self.attrs = attrs

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/dataset.py in _set_init_vars_and_dims(self, vars, coords, compat)
    258         aligned = align_variables(variables)
    259         new_variables, new_coord_names = expand_variables(aligned,
--> 260                                                           compat=compat)
    261 
    262         new_coord_names.update(coords)

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/merge.py in expand_variables(raw_variables, old_variables, compat)
     75                     add_variable(dim, coord.variable)
     76             var = var.variable
---> 77         add_variable(name, var)
     78 
     79     return new_variables, new_coord_names

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/merge.py in add_variable(name, var)
     55 
     56     def add_variable(name, var):
---> 57         var = _as_dataset_variable(name, var)
     58         if name not in variables:
     59             variables[name] = var

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/merge.py in _as_dataset_variable(name, var)
      9     """
     10     try:
---> 11         var = as_variable(var, key=name)
     12     except TypeError:
[...]

PR #854, opened 2016-05-21
BUG: Allow arbitrary objects to be Dataset values

MaximilianR:
Closes https://github.com/pydata/xarray/issues/847

The method here might be a bit sweeping. Let me know your thoughts


MaximilianR:
@shoyer do you recognize this failure? I don't think it's related to this PR (it's also on my other outstanding PR), but not sure?

shoyer:
Yeah, definitely unrelated. I think this is a bug in my h5netcdf project -- I just issued a new release there (should have tested against xarray first!).

MaximilianR:
OK, in which case - let me know feedback here!

MaximilianR:
@shoyer tests pass! Feedback welcome!

shoyer:
This is remarkably simple. I'm pleasantly surprised that it works!

PR #856, opened 2016-05-23
drop python 2.6 support

### LINKED ISSUE #855: drop support for Python 2.6
@shoyer polled the xarray users list about dropping Python 2.6 from the supported versions of Python for xarray.  There were no complaints so it looks like we are moving forward on this at the next major release (0.8).  

xref: https://groups.google.com/forum/#!searchin/xarray/2.6/xarray/JVIUiIhEW_8/qBjxmestCQAJ

jhamman:
closes #855.

There is still a code base of work-arounds for Python 2.6 (examples below).  I'm not sure if we want to remove those now or just stop testing with 2.6 in this PR.  Thoughts from @shoyer and @MaximilianR?
 - [tests](https://github.com/pydata/xarray/blob/master/xarray/test/__init__.py#L110-L120)
 - [backends](https://github.com/pydata/xarray/blob/master/xarray/backends/api.py#L196-L197)


shoyer:
I'm OK either way with removing the 2.6 compatibility cludges. Probably best to remove any obvious ones you notice now -- more subtle things (like dict/set comprehensions) can wait until we refactor the relevant code.

jhamman:
@shoyer, I think this is good to go. I've removed a few 2.6 cludges but avoided refactoring any code.

shoyer:
Looks good to me!

PR #857, opened 2016-05-23
xarray.merge function and major refactor for merge logic

### LINKED ISSUE #417: Make merge a function as well as a method
Roughly speaking, this should look like:

```python
from functools import reduce

def merge(datasets, **kwargs):
    return reduce(lambda x, y: x.merge(y, **kwargs), datasets)
```

We use already use this in `auto_combine`.

Including it would make for a better user experience that requiring users to write it themselves.

shoyer:
Fixes #417

New top level :py:func:`merge` function allows for combining variables from any number of ``Dataset`` and/or ``DataArray`` variables.

Example usage:

    >>> arrays = [xr.DataArray(n, name='var%d' % n) for n in range(5)]
    >>> xr.merge(arrays)
    <xarray.Dataset>
    Dimensions:  ()
    Coordinates:
        *empty*
    Data variables:
        var0     int64 0
        var1     int64 1
        var2     int64 2
        var3     int64 3
        var4     int64 4

The internal refactoring also lays the ground work for supporting ufunc-like functions that merge three or more arguments, such as the full form of `where`.

Notes to any potential reviewers:

This doesn't have as much test coverage as I would like, but I did get all the existing functionality working. I'll see if I can add in some more tests for the internal code in `merge.py`. Also, more docstrings is probably a good idea.

Either way, I'm a bit nervous that I haven't broken things, because frankly I still don't understand exactly how the old code worked (and yes, I wrote it!). So we should probably make a release candidate for xarray v0.8 to avoid the need for a bug fix release shortly afterwards.

cc @jhamman @MaximilianR


doc/combining.rst
```diff
@@ -91,9 +92,22 @@ coordinates:
...
+
+    @verbatim
+    In [1]: xr.merge([ds, ds + 1])
+    MergeError: conflicting value for variable foo:
+    first value: <xarray.Variable (x: 2, y: 3)>
+    array([[ 0.4691123 , -0.28286334, -1.5090585 ],
+           [-1.13563237,  1.21211203, -0.17321465]])
+    second value: <xarray.Variable (x: 2, y: 3)>
+    array([[ 1.4691123 ,  0.71713666, -0.5090585 ],
+           [-0.13563237,  2.21211203,  0.82678535]])
```
jhamman: I'm a little confused by this `MergeError` example. Is the error raised because this merge is trying to merge two datasets that have the same variables which have the same coordinates?

xarray/core/alignment.py
```diff
@@ -92,36 +108,49 @@ def align(*objects, **kwargs):
...
+    aligned = partial_align(*targets, join=join, copy=copy, indexes=indexes)
+
+    for key, aligned_obj in zip(keys, aligned):
+        if isinstance(key, tuple):
+            n, k = key
+            out[n][k] = aligned_obj
+        else:
+            out[key] = aligned_obj
+
+    assert all(arg is not None for arg in out)
```
jhamman: Can you raise an `Exception` here or at a minimum add a message to be printed by the `AssertionError`.

xarray/core/combine.py
```diff
@@ -3,6 +3,8 @@
 import pandas as pd
 
 from . import utils
+from .alignment import partial_align
```
jhamman: this isn't used in `combine.py`.

xarray/core/coordinates.py
```diff
@@ -1,40 +1,12 @@
-from collections import Mapping
+from collections import Mapping, defaultdict
```
jhamman: `defaultdict` imported but unused.

xarray/core/coordinates.py
```diff
@@ -1,40 +1,12 @@
...
 from contextlib import contextmanager
 import pandas as pd
 
 from . import formatting
-from .merge import merge_dataarray_coords
+from .utils import Frozen
+from .merge import merge_core, merge_variables, merge_coords_only, align_and_merge_coords
```
jhamman: `merge_core` and `merge_variables` imported but unused.

xarray/core/coordinates.py
```diff
@@ -74,54 +57,49 @@ def to_index(self, ordered_dims=None):
...
-                    if k not in other.dims:
-                        other_conflicts.add(k)
-                elif _dim_shape(self_var) != _dim_shape(other_var):
-                    promote_dims[k] = _common_shape(self_var, other_var)
-                    self_conflicts.add(k)
-        return self_conflicts, other_conflicts, promote_dims
+    def update(self, other):
+        other_vars = getattr(other, 'variables', other)
+        coords = align_and_merge_coords([self.variables, other_vars],
+                                        # priority_vars=other_vars,
```
jhamman: probably remove this commented out line.

xarray/core/dataarray.py
```diff
@@ -39,11 +39,15 @@ def _infer_coords_and_dims(shape, coords, dims):
...
                 dims = list(coords.keys())
             else:
-                for n, (dim, coord) in enumerate(zip(dims, coords)):
-                    coord = as_variable(coord, key=dim).to_coord()
-                    dims[n] = coord.name
+                for n, coord in enumerate(coords):
+                    try:
+                        coord = as_variable(coord)
+                        dims[n], = coord.dims
+                    except (TypeError, ValueError):
```
jhamman: add a note why these errors would be raised.  Something like:

```Python
except (TypeError, ValueError):
    # TypeError: raised by as_variable if coord is not a coordinate
    # ValueError: raised if coord is the wrong size
    pass
```

xarray/core/dataset.py
```diff
@@ -14,13 +14,13 @@
 from . import alignment
 from . import formatting
 from .. import conventions
-from .alignment import align, align_variables
+from .alignment import align  #, align_valid_dict
```
jhamman: lets remove this comment

xarray/core/dataset.py
```diff
@@ -14,13 +14,13 @@
...
 from .coordinates import DatasetCoordinates, Indexes
 from .common import ImplementsDatasetReduce, BaseDataObject
-from .merge import merge_datasets, expand_variables
+from .merge import (dataset_update_method, dataset_merge_method,
+                    merge_data_and_coords)
 from .utils import Frozen, SortedKeysDict, ChainMap, maybe_wrap_array, hashable
-from .variable import (as_variable, Variable, Coordinate, broadcast_variables,
-                       default_index_coordinate)
+from .variable import (as_variable, Variable, Coordinate, broadcast_variables)
```
jhamman: `as_variable` is also not used here.

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
+
+        if compat == 'broadcast_equals':
+            dim_lengths = broadcast_dimension_size(variables)
+            out = out.expand_dims(dim_lengths)
+
+        for var in variables[1:]:
+            if not getattr(out, compat)(var):
+                raise MergeError('conflicting value for variable %s:\n'
+                                 'first value: %r\nsecond value: %r'
+                                 % (name, out, var))
```
jhamman: maybe flesh this `Error` out a bit. Its not clear to me what this is saying.

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
-from .alignment import align, partial_align, align_variables
-from .utils import ChainMap
-from .variable import as_variable
+from collections import defaultdict, deque
+
+import pandas as pd
+
+from .alignment import align, deep_align
+from .utils import ChainMap, Frozen
+from .variable import as_variable, Coordinate, Variable, default_index_coordinate
```
jhamman: The following are imported but are not used: `defaultdict, deque, align, ChainMap, Coordinate, Variable iteritems`

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
 from .pycompat import (basestring, iteritems, OrderedDict)
 
 
-def _as_dataset_variable(name, var):
-    """Prepare a variable for adding it to a Dataset
+def broadcast_dimension_size(variables):
+    # type: (List[Variable],) -> Variable
```
jhamman: out of curiosity, what's with the type comment at the start of your new functions?

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
+    def __missing__(self, key):
+        self[key] = default = self.default_factory()
+        return default
+
+
+def get_top_priority_variable(priority_arg=None):
+    if priority_arg is not None:
+        # one of these arguments (e.g., the first for in-place
+        # arithmetic or the second for Dataset.update) takes priority
+        priority_variables = list_of_variables_dicts[priority_arg]
```
jhamman: where does `get_top_priority_variable` come from?

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
+        ):
+    # type: (...) -> Dict[Any, Variable]
+    """Merge dicts of variables, while resolving conflicts appropriately.
+    """
+    if priority_vars is None:
+        # one of these arguments (e.g., the first for in-place
+        # arithmetic or the second for Dataset.update) takes priority
+        priority_vars = {}
+
+    assert compat in _VALID_COMPAT
```
jhamman: consider adding a error message here.

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
+
+
+def merge_variables(
+        list_of_variables_dicts,  # type: List[Mapping[Any, Variable]]
+        priority_vars=None,       # type: Optional[int]
+        compat='minimal',         # type: str
+        ):
+    # type: (...) -> Dict[Any, Variable]
+    """Merge dicts of variables, while resolving conflicts appropriately.
+    """
```
jhamman: I know this is private api but maybe just do a formal docstring?

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
+def get_top_priority_variable(priority_arg=None):
+    if priority_arg is not None:
+        # one of these arguments (e.g., the first for in-place
+        # arithmetic or the second for Dataset.update) takes priority
+        priority_variables = list_of_variables_dicts[priority_arg]
+    else:
+        priority_variables = {}
+
+
+
```
jhamman: remove extra blank line

xarray/core/merge.py
```diff
@@ -1,173 +1,368 @@
...
 
-def _merge_dataset_with_dataset(self, other, overwrite_vars, compat, join):
-    aligned_self, other = align(self, other, join=join, copy=False)
 
-    replace_vars, new_vars, new_coord_names = _merge_expand(
-        aligned_self._variables, other._variables, overwrite_vars, compat)
-    new_coord_names.update(other._coord_names)
 
-    return replace_vars, new_vars, new_coord_names
+PANDAS_TYPES = (pd.Series, pd.DataFrame, pd.Panel)
```
jhamman: put this at the top of the module

shoyer:
@jhamman thanks for catching my lint errors -- that's a little embarrassing! My linter seems to have disabled itself, which I need to fix.

jhamman:
@shoyer - I gave this a once over. A few minor comments in addition to my comments above, 

- I didn't look at the tests in depth so whoever looks at this next, that would be a place to start. I also glazed over toward the end of `merge.py` so that needs another look.
- There is a lot of private functions/methods involved in this.  That may be how it has to be but it seems a bit messy at times.

doc/combining.rst
shoyer: Yes, that's what this is attempting to indicate. Let me know if this is clearer....

xarray/core/alignment.py
shoyer: I clarified what this assert is about.

xarray/core/combine.py
shoyer: fixed

xarray/core/coordinates.py
shoyer: fixed

xarray/core/dataset.py
shoyer: done

xarray/core/dataset.py
shoyer: fixed

xarray/core/merge.py
shoyer: oops, removed

xarray/core/merge.py
shoyer: fixed

xarray/core/merge.py
shoyer: fixed

xarray/core/coordinates.py
shoyer: fixed

xarray/core/coordinates.py
shoyer: done

xarray/core/merge.py
shoyer: fixed

xarray/core/merge.py
shoyer: is this better?

xarray/core/merge.py
shoyer: This is Python 3.5 style type annotations (for use with mypy)

xarray/core/dataarray.py
shoyer: good catch -- I was able to remove this and restore the original.

xarray/core/merge.py
shoyer: done

shoyer:
@jhamman another look?

shoyer:
OK, I want to get this in for v0.8.0 (and make that release soon), so I'm merging this. I think it's in significantly better shape that before, though it's certainly still more complex than ideal.

PR #867, opened 2016-06-03
Fix issue with expand_dims on 0d arrays of tuples

### LINKED ISSUE #841: Unable to `squeeze` array of lists
Using python 2.7.10, xarray 0.7.2

```python
listarray = numpy.empty((1,), dtype=object)
listarray[0] = [1,2,3]  # note list type

ds = xarray.Dataset({'v2':(('b',), listarray)})
ds.squeeze()
```
this generates
`ValueError: dimensions () must have the same length as the number of data dimensions, ndim=1`

shoyer:
This was originally reported on the mailing list, because it broke groupby on stacked dimensions (with a multi-index):
https://groups.google.com/forum/#!topic/xarray/fz7HHgpgwk0

Unfortunately, the fix requires a backwards compatibility break, because it
changes how 0d object arrays are handled:

```
In [4]: xr.Variable([], object()).values
Out[4]: array(<object object at 0x10072e2e0>, dtype=object)
```

Previously, we just returned the original object, which was consistent with NumPy. But I think this new behavior is actually more consistent overall, and required far fewer hacks to get `expand_dims` working properly.

fixes #841


shoyer:
Just added a fix for indexing or squeezing object arrays of lists

shoyer:
Added a couple of regression tests and what's new note.  I think this should be good to go, now.

Issue #866, opened 2016-06-02
Drawing only one contour


JoyMonteiro:
Hello,

I was trying to draw only a single contour by passing levels=[0], and nothing gets
plotted.

I checked utils.py, and the logic used to calculate **n_colors** in **_build_discrete_cmap**
gives **n_colors**=0, since it will first set **extend** to 'neither', and so **ext_n** = 0, and 
n_colors = len(levels) + ext_n - 1

I'm not sure, but this might be the issue.

Another issue, which might be unrelated, is when I'm trying to draw two contours. It plots only
one contour, and it will plot two contours only if I set  **norm**=None.

Any suggestions?

TIA,
Joy

fmaussion:
Here's a mwe:

```python
import numpy as np
import matplotlib.pyplot as plt
import xarray as xr

x, y = np.meshgrid(np.arange(12), np.arange(12)) 
z = xr.DataArray(x**2 + y**2)
z.plot.contour(levels=[10], add_colorbar=False)
```
which indeed doesn't plot anything.

Note that without `add_colorbar=False`, it crashes with:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mowglie/.pyvirtualenvs/oggm_pycharm/lib/python3.4/site-packages/xarray/plot/plot.py", line 452, in plotmethod
    return newplotfunc(**allargs)
  File "/home/mowglie/.pyvirtualenvs/oggm_pycharm/lib/python3.4/site-packages/xarray/plot/plot.py", line 426, in newplotfunc
    cbar = plt.colorbar(primitive, ax=ax, extend=cmap_params['extend'])
  File "/home/mowglie/.pyvirtualenvs/oggm_pycharm/lib/python3.4/site-packages/matplotlib/pyplot.py", line 2237, in colorbar
    ret = gcf().colorbar(mappable, cax = cax, ax=ax, **kw)
  File "/home/mowglie/.pyvirtualenvs/oggm_pycharm/lib/python3.4/site-packages/matplotlib/figure.py", line 1595, in colorbar
    cb = cbar.colorbar_factory(cax, mappable, **kw)
  File "/home/mowglie/.pyvirtualenvs/oggm_pycharm/lib/python3.4/site-packages/matplotlib/colorbar.py", line 1330, in colorbar_factory
    cb = Colorbar(cax, mappable, **kwargs)
[...]

PR #868, opened 2016-06-03
BUG: contour plot levels

fmaussion:
This should fix https://github.com/pydata/xarray/issues/866


xarray/plot/plot.py
```diff
@@ -387,6 +387,18 @@ def newplotfunc(darray, x=None, y=None, ax=None, row=None, col=None,
...
+        if plotfunc.__name__ == 'contour':
+            # Contouring data is equivalent to extend upwards
+            extend = extend or 'max'
+
+            # A colorbar with one level is not possible with mpl
+            try:
+                if len(levels) < 2:
+                    add_colorbar = False
+            except TypeError:
+                if levels < 2:
```
fmaussion: This is a huge corner case, but well...

xarray/plot/plot.py
```diff
@@ -384,8 +384,16 @@ def newplotfunc(darray, x=None, y=None, ax=None, row=None, col=None,
 
         _ensure_plottable(xval, yval)
 
-        if 'contour' in plotfunc.__name__ and levels is None:
-            levels = 7  # this is the matplotlib default
+        if 'contour' in plotfunc.__name__:
+            if levels is None:
+                levels = 7  # this is the matplotlib default
+            # A colorbar with one level is not possible with mpl
+            try:
```
shoyer: Can we do a type check instead? e.g.,
```
if (isinstance(levels, int) and levels < 2) or len(levels) < 2:
    add_colorbar = False
```

shoyer: We could also change the default to `add_colorbar=None`, and only apply this logic if `add_colorbar` wasn't set directly.

shoyer:
I wonder -- would it maybe make more sense to simply *not* to add a colorbar by default? It's pretty easy to toggle if desired.

fmaussion:
@shoyer yes, this was my original idea too: `contour` plots usually don't need a colorbar. I changed the default to false for `contour` plots only, and removed the convoluted logic for when the user gives only one level: this is a matplotlib problem after all, and if the user sets `levels=1` and `add_colorbar=True` it really isn't xarray's problem anymore.

I think it's better now, let me know what you think.

xarray/plot/utils.py
```diff
@@ -41,7 +41,7 @@ def _build_discrete_cmap(cmap, levels, extend, filled):
 
     if not filled:
         # non-filled contour plots
-        extend = 'neither'
+        extend = 'max'
```
shoyer: I think I missed the rationale for this one -- why should this change?

fmaussion: This is the actual bug fix for https://github.com/pydata/xarray/issues/866 which motivated this PR ;-).

xarray wasn't plotting contours correctly before this change, see the following code:

```python
import numpy as np
import xarray as xr

x, y = np.meshgrid(np.arange(12), np.arange(12))
z = xr.DataArray(np.sqrt(x**2 + y**2))
z.plot.contour(levels=[2, 4, 6, 8], add_colorbar=False, colors='k')
plt.show()
```
figure_1

With this fix everything is fine now:

figure_1-1

Intuitively, it also makes sense: the last contour does mark all data beyond the last level, hence `extend = 'max'`

shoyer: Got it, thanks!

Issue #752, opened 2016-02-09
Passing colorbar axis argument in plot statement 


dgergel:
I am wondering if it would be possible to add a feature similar to the `seaborn` `heatmap` plotting feature where `cbar_ax=cbar_ax` could be passed in as an argument for da.plot(), such that I could create a separate colorbar axis and use that axis for the `add_colorbar` option.

shoyer:
Yes, contributions to add this feature would be very welcome!

fmaussion:
I just opened a PR for this: https://github.com/pydata/xarray/pull/872

jhamman:
close via #872.

PR #872, opened 2016-06-07
ENH: more control on colorbar

fmaussion:
Addresses https://github.com/pydata/xarray/issues/752 and allows to pass kwargs  to colorbar. For example, it is now possible to do:

```python
import numpy as np
import xarray as xr

x, y = np.meshgrid(np.arange(12), np.arange(12))
z = xr.DataArray(np.sqrt(x**2 + y**2))
ds = z.to_dataset(name='z')

fig, ((ax1, ax2), (ax3, ax4)) = plt.subplots(2, 2, figsize=(12, 12))
ds.z.plot.contourf(ax=ax1)
ds.z.plot.contourf(ax=ax2, cbar_kwargs={'orientation':'horizontal', 'label':'MyLabel'})
ds.z.plot.contourf(ax=ax3, cbar_ax=ax4, cbar_kwargs={'orientation':'horizontal',
                                                     'label':'Funny Cbar',
                                                     'drawedges':True})
plt.tight_layout()
plt.show()
```

test_cbar

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -342,7 +342,7 @@ def _plot2d(plotfunc):
                     add_colorbar=True, add_labels=True, vmin=None, vmax=None,
                     cmap=None, center=None, robust=False, extend=None,
                     levels=None, colors=None, subplot_kws=None,
-                    cbar_ax=None, cbar_kwargs={}, **kwargs):
+                    cbar_ax=None, cbar_kwargs=None, **kwargs):
         # All 2d plots in xarray share this function signature.
         # Method signature below should be consistent.
 
@@ -428,7 +428,7 @@ def _plot2d(plotfunc):
             ax.set_title(darray._title_for_slice())
 
         if add_colorbar:
-            cbar_kwargs = dict(cbar_kwargs)
+            cbar_kwargs = {} if cbar_kwargs is None else dict(cbar_kwargs)
             cbar_kwargs.setdefault('extend', cmap_params['extend'])
             if cbar_ax is None:
                 cbar_kwargs.setdefault('ax', ax)
@@ -437,6 +437,10 @@ def _plot2d(plotfunc):
             cbar = plt.colorbar(primitive, **cbar_kwargs)
             if darray.name and add_labels and 'label' not in cbar_kwargs:
                 cbar.set_label(darray.name, rotation=90)
+        elif cbar_ax is not None or cbar_kwargs is not None:
+            # inform the user about keywords which aren't used
+            raise ValueError("cbar_ax and cbar_kwargs can't be used with "
+                             "add_colorbar=False.")
 
         _update_axes_limits(ax, xincrease, yincrease)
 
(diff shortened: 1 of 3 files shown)


shoyer:
This looks great -- it is possible to add a test for some of this functionality? Take a look at the existing plotting tests and see if there's anything you can do easily.

xarray/plot/plot.py
```diff
@@ -423,8 +428,14 @@ def newplotfunc(darray, x=None, y=None, ax=None, row=None, col=None,
...
-            cbar = plt.colorbar(primitive, ax=ax, extend=cmap_params['extend'])
-            if darray.name and add_labels:
+            cbar_kwargs = dict(cbar_kwargs)
+            cbar_kwargs.setdefault('extend', cmap_params['extend'])
+            if cbar_ax is None:
+                cbar_kwargs.setdefault('ax', ax)
+            else:
+                cbar_kwargs.setdefault('cax', cbar_ax)
+            cbar = plt.colorbar(primitive, **cbar_kwargs)
+            if darray.name and add_labels and 'label' not in cbar_kwargs:
```
jhamman: one corner case that may require checking is when `add_colorbar = False` and `cbar_ax` or `cbar_kwargs` is specified. It might be worth checking and just raising an error.  Unless you can think of a time when you want to be able to specify arguments that aren't used.

xarray/plot/plot.py
```diff
@@ -437,7 +448,8 @@ def plotmethod(_PlotMethods_obj, x=None, y=None, ax=None, row=None,
                    col=None, col_wrap=None, xincrease=True, yincrease=True,
                    add_colorbar=True, add_labels=True, vmin=None, vmax=None,
                    cmap=None, colors=None, center=None, robust=False,
-                   extend=None, levels=None, subplot_kws=None, **kwargs):
+                   extend=None, levels=None, subplot_kws=None,
+                   cbar_ax=None, cbar_kwargs={}, **kwargs):
```
jhamman: I'd prefer `cbar_kwargs=None`, matches the `subplot_kws` style and it is generally good not to put mutable objects as kwarg defaults.

jhamman:
This is great.  I had two small comments and agree that it could use some tests.

fmaussion:
@shoyer I added some tests and and implemented @jhamman 's comments. I am not a big fan of raising an error when keywords are set and not used, but I don't have a use case for this either so I implemented it as requested ;)

fmaussion:
Right, I always forget about this one. Let me know if I should rebase, too.

jhamman:
No need to rebase.  We can squash on the github side now.

fmaussion:
Yes, ok. 

[OT]: Some people on the mpl said it was a dangerous feature of github but I didn't understand why. I guess that for large PRs it might be useful to keep the intermediate commits, but for the majority of the cases it makes the git history much clearer I think.

shoyer:
Yeah, I'm a big fan of squash-on-merge. It's much easier than forcing contributors to master `git rebase`.

shoyer:
Thanks @fmaussion !

rvalenzuelar:
Could you give an example on how to use this functionality when using faceting plots? For example, I have a 4 column figure and I would like to have the colorbar on top.

PR #875, opened 2016-06-08
Correctly concatenate pandas.Index objects

shoyer:
We now have a dedicated method, Coordinate.concat that will correctly
concatenate pandas.Index objects, even if they can't be properly expressed as
NumPy arrays (e.g., PeriodIndex and MultiIndex).

As part of this change, I removed and replaced the internal `interleaved_concat`
routine. It turns out we can do this with an inverse permutation instead, which
results in much simpler and cleaner code. In particular, we no longer need a
special path to support dask.array.

This should help with #818.

### DIFF
--- a/xarray/core/groupby.py
+++ b/xarray/core/groupby.py
@@ -2,6 +2,7 @@ import functools
 import numpy as np
 import pandas as pd
 
+from . import nputils
 from . import ops
 from .combine import concat
 from .common import (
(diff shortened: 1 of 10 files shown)


xarray/core/groupby.py
```diff
@@ -66,6 +67,52 @@ def _dummy_copy(xarray_obj):
         raise AssertionError
     return res
 
+def _is_one_or_none(obj):
+    return obj == 1 or obj is None
```
jhamman: do you want to treat `True` as `1`?  If not, this should be `obj is 1 or obj is None`.

MaximilianR: Interesting! I never knew that. 
I always thought ints & strings should use `==`, but for -1 to 256 `is` is fine: http://stackoverflow.com/questions/2239737/is-it-better-to-use-is-or-for-number-comparison-in-python

fmaussion: But this depends on the interpreter, then. CPython might work, but others might not I guess?

shoyer: These slice objects are generated internally by groupby, so I think it's pretty far fetched to get `step is True`. Using `==` feels more idiomatic, anyways.

xarray/core/variable.py
```diff
@@ -1163,3 +1208,41 @@ def _broadcast_compat_data(self, other):
...
+        arrays to concatenate are skipped.
+
+    Returns
+    -------
+    stacked : Variable
+        Concatenated Variable formed by stacking all the supplied variables
+        along the given dimension.
+    """
+    variables = list(variables)
+    print(variables)
```
MaximilianR: guessing this is left over from debugging

shoyer: fixed, thanks

PR #879, opened 2016-06-11
Multi-index repr

benbovy:
Another item of #719.

An example:

```python
>>> index = pd.MultiIndex.from_product((list('ab'), range(10)))
>>> index.names= ('a_long_level_name', 'level_1')
>>> data = xr.DataArray(range(20), [('x', index)])
>>> data
<xarray.DataArray (x: 20)>
array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,
       17, 18, 19])
Coordinates:
  * x                    (x) object MultiIndex
    - a_long_level_name  object 'a' 'a' 'a' 'a' 'a' 'a' 'a' 'a' 'a' 'a' 'b' ...
    - level_1            int64 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9
```

To be consistent with the displayed coordinates and/or data variables, it displays the actual used level values. Using the `pandas.MultiIndex.get_level_values` method would be expensive for big indexes, so I re-implemented it in xarray so that we can truncate the computation to the first *x* values, which is very cheap.

It still needs testing.

Maybe it would be nice to align the level values.

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -6,6 +6,7 @@ import pandas as pd
 
 from .options import OPTIONS
 from .pycompat import iteritems, unicode_type, bytes_type, dask_array_type
+from .indexing import PandasIndexAdapter
 
 
 def pretty_print(x, numchars):
@@ -122,13 +123,17 @@ def format_items(x):
     return formatted
 
 
+def _get_max_items(max_width):
+    # every item will take up at least two characters, but we always want to
+    # print at least one item
+    return max(int(np.ceil(max_width / 2.0)), 1)
+
+
 def format_array_flat(items_ndarray, max_width):
     """Return a formatted string for as many items in the flattened version of
     items_ndarray that will fit within max_width characters
     """
-    # every item will take up at least two characters, but we always want to
-    # print at least one item
-    max_possibly_relevant = max(int(np.ceil(max_width / 2.0)), 1)
+    max_possibly_relevant = _get_max_items(max_width)
     relevant_items = first_n_items(items_ndarray, max_possibly_relevant)
     pprint_items = format_items(relevant_items)
 
@@ -145,27 +150,36 @@ def format_array_flat(items_ndarray, max_width):
     return pprint_str
 
 
-def _get_level_values(index, level_num, max_size=50):
-    """Similar to `pd.MultiIndex.get_level_values` but
-    here truncated to `max_size`.
+def _format_multiindex_levels(index, col_width, max_width):
+    """Return a formatted string for multi-index levels with
+    their name, dtype and (first) actual used values.
     """
-    unique = index.levels[level_num]
-    labels = index.labels[level_num]
-    size = min(max_size, labels.size)
-    filled = pd.core.algorithms.take_1d(unique.values, labels[:size],
-                                        fill_value=unique._na_value)
-    _simple_new = unique._simple_new
-    values = _simple_new(filled, name=index.names[level_num],
-                         freq=getattr(unique, 'freq', None),
-                         tz=getattr(unique, 'tz', None))
-    return values
+    level_str_lines = []
+    for i in range(index.nlevels):
+        idx = index.get_level_values(i)
+        first_col = pretty_print('- %s ' % idx.name, col_width)
+        front_str = first_col + ('%s ' % idx.dtype)
+        values_str = format_array_flat(idx, max_width - len(front_str))
+        level_str_lines.append(front_str + values_str)
+
+    return '\n'.join(level_str_lines)
 
 
-def _summarize_index_level(name, index, col_width, max_width):
-    first_col = pretty_print('    - %s ' % name, col_width)
-    front_str = first_col + ('%s ' % index.dtype)
-    values_str = format_array_flat(index, max_width - len(front_str))
-    return front_str + values_str
+def _maybe_summarize_multiindex(var, col_width, max_width):
+    if isinstance(var.variable._data, PandasIndexAdapter):
+        indent = 4
+        max_width -= indent
+
+        # get first relevant items before convert to a pandas.Index
+        # so that we avoid loading the whole data
+        index = var[:_get_max_items(max_width)].to_index()
+
+        if isinstance(index, pd.MultiIndex):
+            index_str = _format_multiindex_levels(index, col_width, max_width)
+            return wrap_indent(index_str, length=indent,
+                               start='MultiIndex\n' + ' ' * indent)
+
+    return ''
 
 
 def _summarize_var_or_coord(name, var, col_width, show_values=True,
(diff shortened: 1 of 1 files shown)


xarray/core/formatting.py
```diff
@@ -145,15 +145,50 @@ def format_array_flat(items_ndarray, max_width):
...
 
 
+def _get_level_values(index, level_num, max_size=50):
+    """Similar to `pd.MultiIndex.get_level_values` but
+    here truncated to `max_size`.
+    """
+    unique = index.levels[level_num]
+    labels = index.labels[level_num]
+    size = min(max_size, labels.size)
+    filled = pd.core.algorithms.take_1d(unique.values, labels[:size],
```
shoyer: We need to figure out how to implement this only using public API (nothing prefaced with an underscore). Otherwise, pandas will almost certainly break us in a future release.

shoyer: I would suggest simply using `index.levels[level_num][:max_size]`

xarray/core/formatting.py
```diff
@@ -197,7 +232,14 @@ def summarize_attr(key, value, col_width=None):
...
 
 def _calculate_col_width(mapping):
-    max_name_length = max(len(str(k)) for k in mapping) if mapping else 0
+    names = []
+    for k, v in mapping.items():
+        names.append(k)
+        if hasattr(v, 'to_index'):
+            index = v.to_index()
+            if isinstance(index, pd.MultiIndex):
+                names += ['  %s' % n for n in index.names]
```
shoyer: probably better to use `'- %s'` here just to be make things more explicit.

xarray/core/formatting.py
```diff
@@ -145,15 +145,50 @@ def format_array_flat(items_ndarray, max_width):
...
+
+
 def _summarize_var_or_coord(name, var, col_width, show_values=True,
                             marker=' ', max_width=None):
     if max_width is None:
         max_width = OPTIONS['display_width']
     first_col = pretty_print('  %s %s ' % (marker, name), col_width)
     dims_str = '(%s) ' % ', '.join(map(str, var.dims)) if var.dims else ''
-    front_str = first_col + dims_str + ('%s ' % var.dtype)
+    index = var.to_index()
```
shoyer: We can use `isinstance(var, Coordinate)` here. I don't think it's safe to try to convert every variable into an index (especially not for higher dimensional data).

xarray/core/formatting.py
MaximilianR: or `.join`

xarray/core/formatting.py
benbovy: In some cases it might make sense to just use `index.levels[level_num][:max_size]` to show the (first) unique values for each level.

But in other cases I find this

```python
>>> data.isel(x=range(3))
<xarray.DataArray (x: 3)>
array([0, 1, 2])
Coordinates:
  * x                    (x) object MultiIndex
    - a_long_level_name  object 'a' 'a' 'a'
    - level_1            int64 0 1 2
```

much better than this:
```python
>>> data.isel(x=range(3))
<xarray.DataArray (x: 3)>
array([0, 1, 2])
Coordinates:
  * x                    (x) object MultiIndex
    - a_long_level_name  object 'a' 'b'
    - level_1            int64 0 1 2 3 4 5 6 7 8 9
```

What about just returning `pd.core.algorithms.take_1d(unique.values, labels[:max_size])` or even `np.take(unique.values, labels[:max_size])`?

xarray/core/formatting.py
benbovy: Can we use `var.ndim == 1`? importing `Coordinate` in `formatting` gives sort of circular import, and `DataArray` or `Dataset` 1d coordinates can actually be `Variable` objects.

xarray/core/formatting.py
shoyer: Yes, that's a good point. It's definitely more useful to display the particular index levels that are actually used at those points.

In that case I would use `unique[labels[:max_size]]`. It's better to avoid the NumPy methods like `np.take` on pandas.Index objects because they don't always always preserve dtypes properly.

xarray/core/formatting.py
```diff
@@ -145,15 +145,50 @@ def format_array_flat(items_ndarray, max_width):
...
+    multiindex_str = 'MultiIndex\n' if isinstance(index, pd.MultiIndex) else ''
+    front_str = first_col + dims_str + ('%s %s' % (var.dtype, multiindex_str))
     if show_values:
-        values_str = format_array_flat(var, max_width - len(front_str))
+        values_width = max_width - len(front_str)
+        if isinstance(index, pd.MultiIndex):
+            # get actual level values for at most the first `value_width` items
```
shoyer: put this in another private method, `_format_multiindex`

xarray/core/formatting.py
shoyer: Yeah, that seems OK.

Probably better to first do the indexing up to `max_size`, though. Otherwise this could load a lot more data into memory than necessary when used with dask.

shoyer: possibly better to refactor this into `_format_array_flat`

benbovy:
I ended up refactoring most of the implementation.

It looks cleaner I think, although I'm still not fully satisfied with this version. It isn't as straightforward to implement as I thought, especially for checking (and getting) MultiIndex and for dealing with `col_width` and `max_width`.

xarray/core/formatting.py
```diff
@@ -145,6 +150,38 @@ def format_array_flat(items_ndarray, max_width):
...
+        first_col = pretty_print('- %s ' % idx.name, col_width)
+        front_str = first_col + ('%s ' % idx.dtype)
+        values_str = format_array_flat(idx, max_width - len(front_str))
+        level_str_lines.append(front_str + values_str)
+
+    return '\n'.join(level_str_lines)
+
+
+def _maybe_summarize_multiindex(var, col_width, max_width):
+    if isinstance(var.variable._data, PandasIndexAdapter):
```
shoyer: I think `isinstance(var, Coordinate)` would be a safer check.

xarray/core/formatting.py
```diff
@@ -197,7 +236,15 @@ def summarize_attr(key, value, col_width=None):
 
 
 def _calculate_col_width(mapping):
-    max_name_length = max(len(str(k)) for k in mapping) if mapping else 0
+    names = []
+    for k, v in mapping.items():
+        names.append(k)
+        if (hasattr(v, 'variable')
+            and isinstance(v.variable._data, PandasIndexAdapter)):
```
shoyer: again, this seems a little fragile.

benbovy:
Closing this. See #947.

PR #881, opened 2016-06-13
Fix variable copy with multi-index 

### LINKED ISSUE #769: BUG: .copy() converts MultiIndex into an base Index
Using the example from https://github.com/pydata/xarray/issues/767:
```
In [15]: da.indexes
Out[15]:
band_wavenumber: MultiIndex(levels=[['bar', 'foo'], [4050.2, 4050.3, 4100.1, 4100.3, 4100.5]],
                            labels=[[1, 1, 0, 0, 0], [0, 1, 2, 3, 4]],
                            names=['band', 'wavenumber'])

In [16]: da.copy().indexes
Out[16]:
band_wavenumber: Index([('foo', 4050.2), ('foo', 4050.3), ('bar', 4100.1), ('bar', 4100.3),
                        ('bar', 4100.5)],
                       dtype='object', name='band_wavenumber')
```

benbovy:
Fixes #769.

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -410,7 +410,11 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
         If `deep=True`, the data array is loaded into memory and copied onto
         the new object. Dimensions, attributes and encodings are always copied.
         """
-        if deep and not isinstance(self.data, dask_array_type):
+        if isinstance(self._data, PandasIndexAdapter):
+            # there is no need to copy the index values here even if deep=True
+            # since pandas.Index objects are immutable
+            data = PandasIndexAdapter(self) if deep else self._data
+        elif deep and not isinstance(self.data, dask_array_type):
             # dask arrays don't have a copy method
             # https://github.com/blaze/dask/issues/911
             data = self.data.copy()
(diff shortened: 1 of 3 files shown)


xarray/core/variable.py
```diff
@@ -410,7 +410,11 @@ def copy(self, deep=True):
         If `deep=True`, the data array is loaded into memory and copied onto
         the new object. Dimensions, attributes and encodings are always copied.
         """
-        if deep and not isinstance(self.data, dask_array_type):
+        if isinstance(self._data, PandasIndexAdapter):
+            # there is no need to copy the index values here even if deep=True
+            # since pandas.Index objects are immutable
+            data = PandasIndexAdapter(self) if deep else self._data
```
shoyer: Why do we need these two different code paths? Why not always return `self._data` in this case?

benbovy: I saw this in `Coordinate.copy()` so I thought there might have some wrapper data to copy in `PandasIndexAdapter` but I admit I didn't checked it.

Indeed, now that I see why we have this in `Coordinate.copy()` I don't see any reason to have these two different code paths here.

shoyer:
Thanks!

Issue #880, opened 2016-06-11
Possibility to provide default map background/ elements to plot


cwerner:
Hi.

Really enjoying xarray! Converting all my scripts and on the quest to simplify my plotting.

For my current project I need loads of maps and generally add a land/ ocean layer and country boundaries as well as an default extent to the plot axes...

I use for instance a extra layer like this:

```
import cartopy.crs as cars
import cartopy.feature as feature

# default map backgrounds    
countries = cfeature.NaturalEarthFeature(
    category='cultural',
    name='admin_0_countries',
    scale='50m',
    facecolor='none')

```

An example from a class I work on right now:

```

    def plot_monthly(self, figsize=(12,10)):
        _unit = 'n.d.'
        _name = 'var' 
        if 'units' in self._dataArray.attrs.keys():
            _unit = self._dataArray.attrs['units']


        self._fig, self._axes = plt.subplots(ncols=4, nrows=3, \
                                             figsize=figsize, subplot_kw={'projection': ccrs.PlateCarree()})

        _min = self._dataArray.min()
        _max = self._dataArray.max()

        print self._dataArray

        for i, ax in enumerate(self._axes.flatten()):
            ax.set_extent( self._extent )
            ax.add_feature(self._countries, edgecolor='gray')     
            ax.add_feature(self._ocean)                           
            ax.coastlines(resolution='50m')
            ax.set_title(self._monthNames[i+1])
            _plot = xr.plot.plot(self._dataArray.sel(month=i+1), ax=ax, \
                            vmin=_min, vmax=_max, \
                            transform=ccrs.PlateCarree(), \
                            add_labels=False,
                            robust=True, add_colorbar=False)


        self._fig.subplots_adjust(right=0.8)
        _cbar_ax = self._fig.add_axes([0.85, 0.15, 0.05, 0.7])

        _var_str = "%s [%s]" % (_name, _unit)
        self._fig.colorbar(_plot, cax=_cbar_ax, label=_var_str)
```

This is a bit clumsy at the moment - I basically define the axes for each subplot (say 1..12 month), loop over them, select the axis and add them with ax.set_extent(), ax.add_feature() etc...

This works, but I'd rather use the plot function within DataArray or DataSet... Also thinking about using FacetGrid instead of doing it manually.

I thought about using the new decorators to patch this onto a custom plot directive but my Python is not really advanced enough for decorators and the inner functioning of array it seems...

```
@xr.register_dataarray_accessor('map')
class MapDecorator(object):
[...]

Issue #657, opened 2015-11-15
Plotting on map projection much slower on v0.6.1 than 0.6.0


fmaussion:
The following code snippet produces an average of ERA-Interim temperature on a map:

```python
import matplotlib.pyplot as plt 
import xray
import cartopy.crs as ccrs 
import time

netcdf = xray.open_dataset('ERA-Int-Monthly-2mTemp.nc')
t2_avg = netcdf.t2m.mean(dim='time')

start_time = time.time()
ax = plt.axes(projection=ccrs.Robinson())
if xray.__version__ == '0.6.0':
    t2_avg.plot(ax=ax, origin='upper', aspect='equal', transform=ccrs.PlateCarree()) 
else:
    t2_avg.plot(ax=ax, transform=ccrs.PlateCarree()) 
ax.coastlines()
plt.savefig('t_xray.png')
print("xray V{}: {:.2f} s".format(xray.__version__, time.time() - start_time))
```

![t](https://cloud.githubusercontent.com/assets/10050469/11169673/62804fb0-8bbe-11e5-9324-105a5eb20e08.png)

I've been very careful to check that my environments are exact same (mpl 1.4.3, numpy 1.10.1, cartopy 0.13.0). 

See the output for V0.6.0 and 0.6.1 (output from the latest master is similar to 0.6.1):

**0.6.0**:
```python
python test_xray.py
/home/mowglie/.bin/conda/envs/climate/lib/python3.4/site-packages/matplotlib/collections.py:590: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if self._edgecolors == str('face'):
xray V0.6.0: 3.21 s
```

**0.6.1**:
```python
 python test_xray.py 
/home/mowglie/.bin/conda/envs/upclim/lib/python3.4/site-packages/numpy/lib/shape_base.py:431: FutureWarning: in the future np.array_split will retain the shape of arrays with a zero size, instead of replacing them by `array([])`, which always has a shape of (0,).
  FutureWarning)
/home/mowglie/.bin/conda/envs/upclim/lib/python3.4/site-packages/matplotlib/collections.py:590: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if self._edgecolors == str('face'):
xray V0.6.1: 28.52 s
```
The first warning seems related to recent numpy. Note that a second warning appeared with xray V0.6.1. 

[...]

PR #882, opened 2016-06-13
add geocolormesh

mathause:
Addresses #880 (partially): Automatically create a cartopy GeoAxis for easier plotting of map data. Works for single plots and for FacetGrids. If a normal pyplot axes is passed, defaults to pcolormesh.

Example:

    import xarray as xr
    t = xr.tutorial.load_dataset('air_temperature')

    # simple plot
    t.isel(time=0).air.plot.geocolormesh()
    plt.show()

    # facet grid, no coastlines
    f = t.isel(time=slice(0, 3)).air.plot.geocolormesh(col='time', coastlines=False)

    # higher resolution coastlines
    for ax in f.axes.flatten():
        ax.coastlines(resolution='50m')

    plt.show()

    # if a matplotlib axis is passed, defaults to pcolormesh
    ax = plt.axes()
    t.isel(time=0).air.plot.geocolormesh(ax=ax)
    plt.show()

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -13,6 +13,20 @@ What's New
     import xarray as xr
     np.random.seed(123456)
 
+.. _whats-new.0.9.0:
+
+v0.9.0 (unreleased)
+-------------------
+
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+Enhancements
+~~~~~~~~~~~~
+
+Bug fixes
+~~~~~~~~~
+
 .. _whats-new.0.8.2:
 
 v0.8.2 (18 August 2016)


mathause:
These are the plots from above.

geocolormap1
geocolormap2
geocolormap3

shoyer:
I was trying to avoid getting too much geo-specific functionality in xarray, but this does nicely extend the existing plotting methods, so I think this is a reasonable addition. Anyone else have opinions on this one?

spencerahill:
@shoyer maybe just a name change then?  cartopy obviously is inherently geo-specific.  Is it possible that other fields would find the utilities in geocolormesh useful?  Or could the geocolormesh functionality be refactored into the existing plotting framework?  (Just thinking out loud...not familiar with the plotting codebase at all)

fmaussion:
I also agree that this is a not to intrusive and useful "geo-specific" addition. Without looking at the implementation, I also wonder if it would be possible to add the same functionality to the existing framework without adding a specific function? The background is that there are other ways than pcolormesh to plot on a map (see also https://github.com/pydata/xarray/issues/657)

shoyer:
I'm pretty sure it's *possible* to add cartopy axes to a FacetGrid using the `subplot_kwds` argument already. Unfortunately, I don't have any examples handy since I wrote that code at my last job.

What this code does is add the `projection=ccrs.PlateCarree()` automatically, along with a few other geo specific niceties. I made that suggestion for cartopy itself (https://github.com/SciTools/cartopy/issues/645) but it hasn't been acted on yet.

I agree that this is a good fit for xarray because the plotting API is difficult to extend externally. One nice way to present this to users could be to make a `geoplot` accessor that works like the current `plot` accessor but adds these geo specific options. Then you could write, e.g., `data.geoplot.pcolormesh()`.

mathause:
Yes is is possible to say `subplot_kws=dict(projection=ccrs.PlateCaree()`. For a single plot it is not possible to specify the `projection` (as `ax=plt.gca()`) in the current implementation. Thus one has to create and pass a geoaxes explicitly.

I like the idea of doing `data.geoplot.pcolormesh()`
- We could keep the standard name of the plotting function without the need to make up our own.
- It is possibly easier to extend it to contour and contourf.
- We can possibly offer more control over specific cartopy features (although adding coasts is possibly the one thing that we will be able to agree on).

Some questions:
- Would it also offer `data.geoplot.line` ect.? There is no point but this would be consistent might be asked for. 
- We would need to think hard about how to extend the current implementation without needing to duplicate it.

fmaussion:
> We would need to think hard about how to extend the current implementation without needing to duplicate it.

Do you think it would be enough to expose the `@_plot2d` decorator?

mathause:
What *may* be possible: extend `@_plot2d` with a `projection=None` and a `transform=None` keyword (and the handling thereof)*. In `geoplot` we import `@_plot2d` set the defaults to `projection=ccrs.PlateCaree()` and `transform=ccrs.PlateCaree()`, possibly intercept `coastlines` from `kwargs`. I don't know if it is possible to update the defaults without explicitly re-writing the call signature of the function (which may change (but this could possibly be tested)).

\* This may conflict with the current `ax = plt.gca()`.

PR #886, opened 2016-06-16
Fix drop docstring

### LINKED ISSUE #885: Docstring for Dataset.drop needs revision
The docstring for [Dataset.drop](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.drop.html#xarray.Dataset.drop) indicates that the first argument, labels, should be a string indicating names of variables or index labels to drop. I'm not sure, but I'm guessing it either takes in several strings (in which case the docstring should say *labels), or it takes in either a string or an iterable (in which case the argument type should reflect this).

shoyer:
Fixes #885

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -898,8 +898,8 @@ class DataArray(AbstractArray, BaseDataObject):
 
         Parameters
         ----------
-        labels : str
-            Names of coordinate variables or index labels to drop.
+        labels : scalar or list of scalars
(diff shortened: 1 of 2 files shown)


mcgibbon:
Do you really want the docstring to say "scalar or list of scalars" and not "str or list of str"?

shoyer:
Yes, it's possible to have non-string index labels, or even variable names
(though the later isn't recommended)

Issue #862, opened 2016-05-30
decode_cf not concatenating string arrays


mogismog:
**TL;DR**: `xarray.conventions.decode_cf()` doesn't seem to want to concatenate string arrays after opening up dataset with `decode_cf=False`.

**OS**: Tried on both OS X 11.10 and 11.11
**xarray version**: 0.7.2 installed via conda
**Python version**: 2.7.11

Hey all,

I'm not sure if this is a bug or the intended behavior, but running `xarray.conventions.decode_cf` doesn't seem to concatenate 2D string arrays as promised given certain circumstances.

Specifically,  MADIS netCDF files have `_FillValue`/`missing_value` conflicts. When opening up the file in `xarray`, the exception gives this suggestion:

`ValueError: ('Discovered conflicting _FillValue and missing_value.  Considering opening the offending dataset using decode_cf=False, corrected the attributes', 'and decoding explicitly using xarray.conventions.decode_cf(ds)')`

Doing this, though, doesn't result in 2D string arrays being concatenated:

```
In [50]: import xarray as xr

In [51]: fname = '20160518_1200'

In [52]: ds = xr.open_dataset(fname, decode_cf=False)

In [53]: ds.stationId
Out[53]:
<xarray.DataArray 'stationId' (recNum: 126154, maxStaIdLen: 6)>
[756924 values with dtype=|S1]
Coordinates:
  * maxStaIdLen  (maxStaIdLen) int64 0 1 2 3 4 5
  * recNum       (recNum) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 ...
Attributes:
    long_name: alphanumeric station Id
    reference: station table

In [54]: for _, v in ds.variables.iteritems():
    _fix_fillval_conflict(v) # You can find this function in the linked gist
   ....:

In [55]: decoded_ds = xr.conventions.decode_cf(ds, concat_characters=True)

In [56]: decoded_ds.stationId
Out[56]:
<xarray.DataArray 'stationId' (recNum: 126154, maxStaIdLen: 6)>
[756924 values with dtype=|S1]
Coordinates:
  * maxStaIdLen  (maxStaIdLen) int64 0 1 2 3 4 5
  * recNum       (recNum) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 ...
Attributes:
    long_name: alphanumeric station Id
    reference: station table
```

[...]

PR #889, opened 2016-06-21
Add concat_dimensions kwarg to decode_cf

mogismog:
Addressing #862 (and maybe others), this adds a new keyword argument `concat_dimensions` to `decode_cf`. This allows the user to explicitly specify which dimensions to concatenate over for string arrays. The dimensions that are concatenated over are left in the resulting Dataset. Also added tests!

### DIFF
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -824,7 +824,8 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
 
 def decode_cf_variables(variables, attributes, concat_characters=True,
                         mask_and_scale=True, decode_times=True,
-                        decode_coords=True, drop_variables=None):
+                        decode_coords=True, drop_variables=None,
+                        concat_dimensions=None):
     """
     Decode a several CF encoded variables.
 
@@ -837,11 +838,12 @@ def decode_cf_variables(variables, attributes, concat_characters=True,
 
     def stackable(dim):
         # figure out if a dimension can be concatenated over
-        if dim in variables:
+        if dim in variables and dim not in concat_dimensions:
             return False
         for v in dimensions_used_by[dim]:
             if v.dtype.kind != 'S' or dim != v.dims[-1]:
-                return False
+                if dim not in concat_dimensions:
+                    return False
         return True
 
     coord_names = set()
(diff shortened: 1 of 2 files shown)


xarray/conventions.py
```diff
@@ -837,11 +838,12 @@ def decode_cf_variables(variables, attributes, concat_characters=True,
 
     def stackable(dim):
         # figure out if a dimension can be concatenated over
-        if dim in variables:
+        if dim in variables and dim not in concat_dimensions:
```
shoyer: It probably makes more sense to simply use `concat_dimensions` instead of the heuristics here if it is provided? E.g., `if concat_dimensions is not None: return dim in stack_dimensions`?

mogismog: Apologies for the delayed response, @shoyer...

I guess my concern was making sure this still worked as expected when opening up netCDF files (e.g. it goes through all the checks and concatenates as it should, etc.), so I didn't want to change _too_ much, but I can see how that looks a bit sloppy.

Gave it a bit of thought, maybe something like this would work:

```
    def stackable(dim):
        # figure out if a dimension can be concatenated over
        # concat_dimensions is turned into a set (like drop_variables) before
        # this function is used.
        if concat_dimensions:
            return dim in concat_dimensions
        else:
            if dim in variables:
                return False
            for v in dimensions_used_by[dim]:
                if v.dtype.kind != 'S' or dim != v.dims[-1]:
                    return False
        return True
```

...or maybe I'm misunderstanding what you're suggesting.

Also: I wonder if `concat_dimensions` should be added as a kwarg in `open_dataset`? If the user knows the dimension names to concatenate before hand, I could see that functionality being useful.

shoyer: Yes, this logic looks right to me.

Adding it to open_dataset also seems OK, but let's pick a more descriptive name that makes it clear it applies to character/string arrays, e.g., `concat_char_dimensions`?

jhamman:
@mogismog - is this something you want to try and finish up? It looks like we're pretty close although we'll need to add a note in the Whats New page and sort out the merge conflict.

mogismog:
@jhamman Sorry for the delayed response (and the even _more_ delayed PR)! I'd love to finish this up, apologies for having this fall to the wayside.

Lemme look at it a bit after work and see how much work it would take to resolve the merge conflicts, though it seems like the issue is only in the `test_conventions.py` file, so it may not be that difficult to resolve.

jhamman:
This is now quite out of date. Closing as part of a housekeeping effort.

Issue #887, opened 2016-06-19
Perf: use Scipy engine by default for netcdf3?


mangecoeur:
Not really a bug, but I'm finding that the scipy backend is considerably faster than the netCDF backend for netCDF 3 files (using dataset: http://rda.ucar.edu/datasets/ds093.1/). Using Anaconda python with MKL. Not sure if this is always faster, but if it is perhaps xarray should default to scipy backend for netCDF 3 files?

shoyer:
Yes, I have found this to be true in my experience, as well. The scipy backend also works more smoothly when using multiple threads with dask, because it releases the GIL.

I would be happy to accept a patch that adds this as the default behavior. We would need to detect netCDF3 vs netCDF4 files automatically when doing file reading by looking at the first few bytes from the file. See these docs for file format specs, which include the necessary details:
https://www.hdfgroup.org/HDF5/doc/H5.format.html#Superblock
http://www.unidata.ucar.edu/software/netcdf/docs/file_format_specifications.html

PR #895, opened 2016-07-11
Tweaks for opening datasets

tsupinie:
I tweaked the open_dataset() and open_mfdataset() functions for better performance with the PyNIO engine. 
1. I use a lot of HDF4 files with what I guess I'll call "malformed names," where the file name does not end with ".hdf". PyNIO is able to figure out the format of the file, but it prints an annoying warning. As far as I can tell, the only way to shut off the warning is to tell it the format of the file, which I've added a `format` option for.
2. I added an option called `only_variables` which specifies which variables to load from the dataset in the event you don't want to load all variables. Say, for example, I have a dataset with 47 variables in it, but I only need 3 of them. If the data are not cached, then only loading the 3 I need cuts the I/O time in half. If they _are_ cached, then loading only the 3 takes 20% of the time to load the full dataset. The `only_variables` option behaves pretty similarly to `drop_variables`. The default is to load all variables.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -81,7 +81,7 @@ def _validate_dataset_names(dataset):
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=True, decode_times=True,
                  concat_characters=True, decode_coords=True, engine=None,
-                 chunks=None, lock=None, drop_variables=None):
+                 chunks=None, lock=None, drop_variables=None, format=''):
     """Load and decode a dataset from a file or file-like object.
 
     Parameters
@@ -135,6 +135,10 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         A variable or list of variables to exclude from being parsed from the
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
+    format: string, optional
+        The format of the file to open (PyNIO engine only). This may be useful 
+        for files with malformed names. Acceptable values are those formats 
+        handled by PyNIO; default is to let it autodetect the format.
 
     Returns
     -------
@@ -214,7 +218,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
             elif engine == 'h5netcdf':
                 store = backends.H5NetCDFStore(filename_or_obj, group=group)
             elif engine == 'pynio':
-                store = backends.NioDataStore(filename_or_obj)
+                store = backends.NioDataStore(filename_or_obj, format=format)
             else:
                 raise ValueError('unrecognized engine for open_dataset: %r'
                                  % engine)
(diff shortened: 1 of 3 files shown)


mathause:
I like (2) - I once needed this functionality and had to infer the `drop_vars`.

````
def get_drop_vars(variables, ds):

    if isinstance(variables, basestring):
        variables = [variables]
    else:
        variables = list(variables)

    # all variables
    drop_var = set(ds.variables.keys())

    # do not drop: coordinates, variables and time bounds
    keep_var = ds.coords.keys() + variables + ['time_bounds', 'time_bnds']

    drop_var.difference_update(keep_var)

    return drop_var`

**Comment by fmaussion** [comment]
Maybe `keep_variables` would be a more appropriate name? ('keep' vs 'drop').

File: xarray/backends/api.py [hunk 1]
```diff
@@ -81,7 +81,8 @@ def check_name(name):
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=True, decode_times=True,
                  concat_characters=True, decode_coords=True, engine=None,
-                 chunks=None, lock=None, drop_variables=None):
+                 chunks=None, lock=None, drop_variables=None, 
+                 only_variables=None, format=''):
```
shoyer: The default should be `format=None`.

xarray/backends/api.py
```diff
@@ -135,6 +136,14 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         A variable or list of variables to exclude from being parsed from the
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
+    only_variables: string or iterable, optional
+        A variable or list of variables to load from the dataset. This is
+        useful if you don't need all the variables in the file and don't want
+        to spend time loading them. Default is to load all variables.
+    format: string, optional
+        The format of the file to open (PyNIO engine only). This may be useful
```
shoyer: I would simply say "currently only used by the PyNIO engine". We might eventually use this later, e.g., for selecting the fastest engine to use when open netCDF3 files (https://github.com/pydata/xarray/issues/887).

shoyer:
This looks like nice functionality!

The main thing it needs is tests. Let me know if you have any questions about how to go about writing those.

I also like the name `keep_variables` better than than `only_variables`.

jhamman:
@tsupinie - are you interested in finishing this up? Beyond a few changes requested by @shoyer, we just need some test coverage and a whatsnew entry.

max-sixty:
Would anyone like to take this up? I was going to close as stale, but it's fairly close!

mathause:
Once #2732 is merged (#2380) the first issue from the list is resolved. Id' still be interested to get No 2 in.Maybe it would be better to open a new PR?

mathause:
I am closing this as stale. Half of this OR is implemented (#2732) the other half is  #1754.

PR #905, opened 2016-07-19
Fix incorrect test for dask version

### LINKED ISSUE #891: BUG: Test for dask version erroneously fails when calling xr.mfdataset()
I have latest version of dask installed but get the following exception when using xr.mfdataset():
'ImportError: xarray requires dask version 0.6 or newer'

Problem appears to be due to the fact that the latest version is 0.10 (i.e., dask.__version__ =  '0.10.0') but exception is raised since 0.1 < 0.6

shoyer:
Fixes #891

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -162,7 +162,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
                 from dask.base import tokenize
             except ImportError:
                 import dask  # raise the usual error if dask is entirely missing
-                if dask.__version__ < '0.6':
+                if StrictVersion(dask.__version__) < StrictVersion('0.6'):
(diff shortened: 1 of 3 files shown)


xarray/backends/api.py
```diff
@@ -162,7 +162,7 @@ def maybe_decode_store(store, lock=False):
                 from dask.base import tokenize
             except ImportError:
                 import dask  # raise the usual error if dask is entirely missing
-                if dask.__version__ < '0.6':
+                if StrictVersion(dask.__version__) < StrictVersion('0.6'):
```
jhamman: Do you need to import `StrictVersion`  c. line 13?

shoyer: Why yes, we do need that!

shoyer:
Thanks for reviewing @jhamman

Issue #432, opened 2015-06-13
Tools for converting between xray.Dataset and nested dictionaries/JSON


shoyer:
This came up in discussion with @freeman-lab -- xray does not have direct support for converting datasets to or from nested dictionaries (i.e., as could be serialized in JSON).

This is quite straightforward to implement oneself, of course, but there's something to be said for making this more obvious. I'm thinking of a serialization format that looks something like this:
```
{
    'variables': {
        'temperature': {
            'dimensions': ['x'],
            'data': [1, 2, 3],
            'attributes': {}
            }
        ...
    }
    'attributes': {
        'title': 'My example dataset',
        ...
    }
}
```

The solution here would be to either:

1. add a few examples to the IO documentation of how to roll this one-self, or
2. create a few helper methods/functions to make this even easier: `xray.Dataset.to_dict`/`xray.read_dict`.

jsignell:
Is this still of interest? I was thinking it would look something like this: 

```
d = {'coordinates': {}, 'variables': {}, 'attributes': {}}

d['attributes'].update(dict(self.attrs))

for k in self.coords:
    d['coordinates'].update({k: {'data': self[k].data,
                                 'dimensions': list(self[k].dims),
                                 'attributes': dict(self[k].attrs)}})
if hasattr(self, 'data_vars'):
    for k in self.data_vars:
        d['variables'].update({k: {'data': self[k].data,
                                 'dimensions': list(self[k].dims),
                                 'attributes': dict(self[k].attrs)}})
else:
    d['variables'].update({'data': self.data,
                           'dimensions': list(self.dims),
                           'attributes': dict(self.attrs)})
```

shoyer:
Yes, I think this is still of interest, though of course the devil is in the details.

1. Do we make this look closer to the `xarray.Dataset` data model (`coords`, `data_vars`, `attrs`, `dims`) or netCDF (`variables`, `attributes`, `dimensions`)?
2. If the later -- do we go so far as to encode all data types (e.g., dates and times) according to CF conventions?
3. Do we save data in the form of nested lists or in a numpy array?
4. Do we output directly output to JSON or just a `dict`?
5. Do we include `dims` or `dimensions` (providing dimension sizes) as a top level field/check?
6. How does the format differ for `xarray.DataArray`? Do we even bother with `DataArray`?

My inclinations:
1. Mirror `xarray.Dataset`
2. NA
3. Use nested lists of native Python types, e.g., generated with numpy's `.tolist()` method.
[...]

PR #917, opened 2016-07-22
added to_dict function for xarray objects

jsignell:
After the conversation #432

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -25,7 +25,7 @@ module:
 
 .. ipython:: python
 
-    import pickle as pickle
+    import pickle
 
     ds = xr.Dataset({'foo': (('x', 'y'), np.random.rand(4, 5))},
                     coords={'x': [10, 20, 30, 40],
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -25,7 +25,7 @@ module:
 
 .. ipython:: python
 
-    import cPickle as pickle
+    import pickle as pickle
 
     ds = xr.Dataset({'foo': (('x', 'y'), np.random.rand(4, 5))},
                     coords={'x': [10, 20, 30, 40],
@@ -50,6 +50,8 @@ and lets you use xarray objects with Python modules like
    remains unchanged. Because the internal design of xarray is still being
    refined, we make no guarantees (at this point) that objects pickled with
    this version of xarray will work in future versions.
+
+.. _dictionary io:
 
 Dictionary
 ----------
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -51,6 +51,32 @@ and lets you use xarray objects with Python modules like
    refined, we make no guarantees (at this point) that objects pickled with
    this version of xarray will work in future versions.
 
+Dictionary
+----------
+
+Serializing an xarray object to a Python dictionary is also simple.
+
+We can convert a ``Dataset`` (or a ``DataArray``) to a dict using
+:py:meth:`~xarray.Dataset.to_dict`:
+
+.. ipython:: python
+
+    d = ds.to_dict()
+    d
+
+We can create a new xarray object from a dict using
+:py:meth:`~xarray.Dataset.from_dict`:
+
+.. ipython:: python
+
+    ds_dict = xr.Dataset.from_dict(d)
+    ds_dict
+
+Dictionary support allows for flexible use of xarray objects. It doesn't
+require external libraries and dicts can easily be pickled, or converted to
+json, or geojson. All the values are converted to lists, so dicts might
+be quite large.
+
 netCDF
 ------
 
@@ -58,7 +84,7 @@ Currently, the only disk based serialization format that xarray directly support
 is `netCDF`__. netCDF is a file format for fully self-described datasets that
 is widely used in the geosciences and supported on almost all platforms. We use
 netCDF because xarray was based on the netCDF data model, so netCDF files on disk
-directly correspond to :py:class:`~xarray.Dataset` objects. Recent versions
+directly correspond to :py:class:`~xarray.Dataset` objects. Recent versions of
 netCDF are based on the even more widely used HDF5 file-format.
 
 __ http://www.unidata.ucar.edu/software/netcdf/
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -20,9 +20,12 @@ v0.8.2 (unreleased)
 
 Enhancements
 ~~~~~~~~~~~~
-
 - New documentation on :ref:`panel transition`. By
   `Maximilian Roos <https://github.com/MaximilianR>`_.
+- New ``Dataset`` and ``DataArray`` methods :py:meth:`to_dict` and
+  :py:meth:`from_dict` to allow easy conversion between dictionaries and
+  xarray objects (:issue:`432`). See :ref:`io` for more details. By
+  `Julia Signell <https://github.com/jsignell>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,8 +24,8 @@ Enhancements
   `Maximilian Roos <https://github.com/MaximilianR>`_.
 - New ``Dataset`` and ``DataArray`` methods :py:meth:`to_dict` and
   :py:meth:`from_dict` to allow easy conversion between dictionaries and
-  xarray objects (:issue:`432`). See :ref:`io` for more details. By
-  `Julia Signell <https://github.com/jsignell>`_.
+  xarray objects (:issue:`432`). See :ref:`dictionary IO <dictionary io>`
+  for more details. By `Julia Signell <https://github.com/jsignell>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -577,6 +577,25 @@ class BaseDataObject(AttrAccessMixin):
 
         return outobj._where(outcond)
 
+    def to_dict(self):
+        """Convert xarray object to a dictionary following xarray naming conventions.
+        """
+        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': list(self.dims)}
+
+        for k in self.coords:
+            d['coords'].update({k: {'data': self[k].values.tolist(),
+                                    'dims': list(self[k].dims),
+                                    'attrs': dict(self[k].attrs)}})
+        if hasattr(self, 'data_vars'):
+            d.update({'data_vars': {}})
+            for k in self.data_vars:
+                d['data_vars'].update({k: {'data': self[k].values.tolist(),
+                                           'dims': list(self[k].dims),
+                                           'attrs': dict(self[k].attrs)}})
+        else:
+            d.update({'data': self.values.tolist(),
+                      'name': self.name})
+        return d
 
     # this has no runtime function - these are listed so IDEs know these methods
     # are defined and don't warn on these operations
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -577,26 +577,6 @@ class BaseDataObject(AttrAccessMixin):
 
         return outobj._where(outcond)
 
-    def to_dict(self):
-        """Convert xarray object to a dictionary following xarray naming conventions.
-        """
-        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': list(self.dims)}
-
-        for k in self.coords:
-            d['coords'].update({k: {'data': self[k].values.tolist(),
-                                    'dims': list(self[k].dims),
-                                    'attrs': dict(self[k].attrs)}})
-        if hasattr(self, 'data_vars'):
-            d.update({'data_vars': {}})
-            for k in self.data_vars:
-                d['data_vars'].update({k: {'data': self[k].values.tolist(),
-                                           'dims': list(self[k].dims),
-                                           'attrs': dict(self[k].attrs)}})
-        else:
-            d.update({'data': self.values.tolist(),
-                      'name': self.name})
-        return d
-
     # this has no runtime function - these are listed so IDEs know these methods
     # are defined and don't warn on these operations
     __lt__ = __le__ =__ge__ = __gt__ = __add__ = __sub__ = __mul__ = \
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1086,7 +1086,11 @@ class DataArray(AbstractArray, BaseDataObject):
         return np.ma.MaskedArray(data=self.values, mask=isnull, copy=copy)
 
     def to_dict(self):
-        """Convert this xarray.DataArray into a dictionary following xarray naming conventions.
+        """
+        Convert this xarray.DataArray into a dictionary following xarray
+        naming conventions.
+
+        Useful for coverting to json.
         """
         d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': self.dims}
 
@@ -1095,10 +1099,11 @@ class DataArray(AbstractArray, BaseDataObject):
                 val = val.astype('datetime64[us]')
             elif np.issubdtype(val.dtype, np.timedelta64):
                 val = val.astype('timedelta64[us]')
-            return(val)
+            return val
 
         for k in self.coords:
-            d['coords'].update({k: {'data': time_check(self[k].values).tolist(),
+            data = time_check(self[k].values).tolist()
+            d['coords'].update({k: {'data': data,
                                     'dims': self[k].dims,
                                     'attrs': dict(self[k].attrs)}})
 
@@ -1111,33 +1116,52 @@ class DataArray(AbstractArray, BaseDataObject):
         """
         Convert a dictionary into an xarray.DataArray
 
+        Input dict can take several forms:
+
+        d = {'dims': ('t'), 'data': x}
+
+        d = {'coords': {'t': {'dims': 't', 'data': t,
+                              'attrs': {'units':'s'}}},
+             'attrs': {'title': 'air temperature'},
+             'dims': 't',
+             'data': x,
+             'name': 'a'}
+
+        where 't' is the name of the dimesion, 'a' is the name of the array,
+        and  x and t are lists, numpy.arrays, or pandas objects.
+
         Parameters
         ----------
         d : dict, with a minimum structure of {'dims': [..], 'data': [..]}
-            or optionally {'dims': [..], 
-                           'data': [..], 
+            or optionally {'dims': [..],
+                           'data': [..],
                            'coords': {'dims': [..], 'data': [..]}}
 
         Returns
         -------
         obj : xarray.DataArray
 
+        See also
+        --------
+        xarray.Dataset.from_dict
         """
         coords = None
-        if 'coords' in d.keys():
+        if 'coords' in d:
             try:
-                coords = OrderedDict([(var[0], (var[1]['dims'], 
-                                                var[1]['data'], 
-                                                var[1].get('attrs'))) for var in d['coords'].items()])
+                coords = OrderedDict([(k, (v['dims'],
+                                           v['data'],
+                                           v.get('attrs')))
+                                      for k, v in d['coords'].items()])
             except KeyError as e:
                 raise KeyError(
-                    'cannot convert dict when coords are missing {dims_data}'.format(
-                        dims_data=str(e.args[0])))
+                    'cannot convert dict when coords are missing '
+                    '{dims_data}'.format(dims_data=str(e.args[0])))
         try:
-            obj = cls(d['data'], coords, d.get('dims'), d.get('name'), d.get('attrs'))
+            obj = cls(d['data'], coords, d.get('dims'), d.get('name'),
+                      d.get('attrs'))
         except KeyError:
             raise KeyError('cannot convert dict with missing data')
-        return(obj)
+        return obj
 
     @classmethod
     def from_series(cls, series):
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1085,6 +1085,20 @@ class DataArray(AbstractArray, BaseDataObject):
         isnull = pd.isnull(self.values)
         return np.ma.MaskedArray(data=self.values, mask=isnull, copy=copy)
 
+    def to_dict(self):
+        """Convert this array to a dictionary following xarray naming conventions.
+        """
+        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': list(self.dims)}
+
+        for k in self.coords:
+            d['coords'].update({k: {'data': self[k].values.tolist(),
+                                    'dims': list(self[k].dims),
+                                    'attrs': dict(self[k].attrs)}})
+
+        d.update({'data': self.values.tolist(),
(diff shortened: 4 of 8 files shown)


xarray/core/common.py
```diff
@@ -577,6 +577,25 @@ def where(self, cond, other=None, drop=False):
...
+    def to_dict(self):
+        """Convert xarray object to a dictionary following xarray naming conventions.
+        """
+        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': list(self.dims)}
+
+        for k in self.coords:
+            d['coords'].update({k: {'data': self[k].values.tolist(),
+                                    'dims': list(self[k].dims),
+                                    'attrs': dict(self[k].attrs)}})
+        if hasattr(self, 'data_vars'):
```
shoyer: These sorts of checks are best avoided, if possible. I would simply write two implementations of `to_dict`, one on `Dataset` and one on `DataArray`.

jsignell: That makes sense. I changed it below.

shoyer:
This looks like a great start, but to be really useful it needs a couple other things:

-[ ] corresponding `Dataset.from_dict` and `DataArray.from_dict` class methods for creating xarray datasets from the output of `to_dict`
-[ ] unit tests verifying that these methods output the expected thing, and that a dataset can be faithfully round-tripped through `to_dict`/`from_dict`
-[ ] a section in the "Serialization and IO" describing how to use these methods

Please ask if you need guidance on where to start for any of these! This will be a very welcome addition to xarray :).

jsignell:
Ok, I wrote a `Dataset.from_dict` class method. I imagine the `DataArray` one will look pretty similar, so I just wanted to see what you think. The main issue that I see is that time doesn't round trip. I wasn't sure if the user should need to set a `parse_date` flag with the `dim` name or something or whether the function should try to convert to time any `dim` with the string 'time' in it? 
```
    @classmethod
    def from_dict(cls, d):
        """
        Convert a dictionary into an xarray.Dataset.
        """
        obj = cls()
        coords=OrderedDict([(k, d['coords'][k]) for k in d['dims']])
        for dim, dim_d in coords.items():
            obj[dim] = (dim_d['dims'], dim_d['data'], dim_d['attrs'])
            print(dim, dim_d['data'][0])
        for var, var_d in d['data_vars'].items():
            obj[var] = (var_d['dims'], var_d['data'], var_d['attrs'])
        obj.attrs.update(d['attrs'])

        return(obj)
```
Regarding unit tests, I haven't ever written one before but I would be happy to try (I know they are an important part of good development).

jsignell:
Ok, I have committed a first pass at unit tests. I purposefully wrote a failing time test.

shoyer:
I think the datetime issue is a numpy bug related to https://github.com/numpy/numpy/issues/7619

We can work around this by casting to `datetime64[ns]`:
```
(Pdb) ds['t'].values.tolist()
[1356998400000000000, 1357084800000000000, 1357171200000000000, 1357257600000000000, 1357344000000000000, 1357430400000000000, 1357516800000000000, 1357603200000000000, 1357689600000000000, 1357776000000000000]
(Pdb) ds['t'].values.astype('datetime64[us]').tolist()
[datetime.datetime(2013, 1, 1, 0, 0), datetime.datetime(2013, 1, 2, 0, 0), datetime.datetime(2013, 1, 3, 0, 0), datetime.datetime(2013, 1, 4, 0, 0), datetime.datetime(2013, 1, 5, 0, 0), datetime.datetime(2013, 1, 6, 0, 0), datetime.datetime(2013, 1, 7, 0, 0), datetime.datetime(2013, 1, 8, 0, 0), datetime.datetime(2013, 1, 9, 0, 0), datetime.datetime(2013, 1, 10, 0, 0)]
```
The same issue holds for timedelta:
```
(Pdb) (ds['t'] - ds['t'][0]).values.tolist()
[0, 86400000000000, 172800000000000, 259200000000000, 345600000000000, 432000000000000, 518400000000000, 604800000000000, 691200000000000, 777600000000000]
(Pdb) (ds['t'] - ds['t'][0]).values.astype('timedelta64[us]').tolist()
[datetime.timedelta(0), datetime.timedelta(1), datetime.timedelta(2), datetime.timedelta(3), datetime.timedelta(4), datetime.timedelta(5), datetime.timedelta(6), datetime.timedelta(7), datetime.timedelta(8), datetime.timedelta(9)]
```

We can work around these by checking for timedelta64 or datetime64 dtypes (use `np.issubdtype(values.dtype, np.datetime64)`, and using astype to convert to `datetime64[us]`/`timedelta64[us]`.

xarray/test/test_dataset.py
```diff
@@ -1849,6 +1849,75 @@ def test_convert_dataframe_with_many_types_and_multiindex(self):
...
+        expected = {
+                    'coords': {
+                               't':{
+                                    'dims': ['t'],
+                                    'data': t,
+                                    'attrs': {}
+                                    }
+                              },
+                    'attrs': {}, 
+                    'dims': ['t'],
```
shoyer: Maybe save `dims` on a dataset as a dict instead?

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,47 @@ def from_dataframe(cls, dataframe):
...
+        obj = cls()
+
+        dims=OrderedDict([(k, d['coords'][k]) for k in d['dims']])
+        for dim, dim_d in dims.items():
+            obj[dim] = (dim_d['dims'], dim_d['data'], dim_d['attrs'])
+
+        for var, var_d in d['data_vars'].items():
+            obj[var] = (var_d['dims'], var_d['data'], var_d['attrs'])
+
+        # what it coords aren't dims?
```
shoyer: A slightly cleaner way to solve this is to loop over `d['coords'].items()` instead of looping of `dims.items()` separately.

xarray/test/test_dataset.py
jsignell: What is the value of that? For `DataArray` it will have to be a list or an `OrderedDict` anyways so that the shape of the data matches the shape of the dims.

shoyer: It's true -- `dims` is inconsistent between `Dataset` and `DataArray`. But short of switching `dims` on `DataArray` to an OrderedDict (maybe not a bad idea, but a separate discussion), I think the serialization format should be consistent with xarray's data model.

shoyer:
One big thing that `from_dict` needs is validation. If the input dict does not match the expected format (e.g., there is a missing field), the user should get a sensible error message so they understand what went wrong.

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,47 @@ def from_dataframe(cls, dataframe):
...
+                                       'dims': list(self[k].dims),
+                                       'attrs': dict(self[k].attrs)}})
+        return d
+
+    @classmethod
+    def from_dict(cls, d):
+        """
+        Convert a dictionary into an xarray.Dataset.
+        """
+        obj = cls()
```
shoyer: Because of the way that Dataset works internally, it can be much slower (for a large number of variables) to add them incrementally to a Dataset rather than to call the Dataset constructor once. Also, the behavior could differ, because the variables are aligned incrementally rather than all at once.

So I would consider building up ordered dictionaries for `data_vars`, `coords` and `attrs` and calling `cls(data_vars, coords, attrs)` at the end.

xarray/test/test_dataset.py
```diff
@@ -1849,6 +1849,75 @@ def test_convert_dataframe_with_many_types_and_multiindex(self):
             expected = expected.drop(['h'], axis=1)
         assert roundtripped.equals(expected)
 
+    def test_to_and_from_dict(self):
+        """
+        <xarray.Dataset>
```
shoyer: We normally don't write docstrings for test methods. Put this in a multi-line comment instead.

shoyer:
It would also be good to test roundtripping arrays with some `NaN` and `NaT` values as well

xarray/test/test_dataset.py
jsignell: Ok, I think I get it. So we would want a plain `dict` for `ds['dims']`, but what would we want for `ds['data_vars']['air']['dims']`, an `OrderedDict`?

jsignell:
Ok, I made the changes that you suggested. I still need to work on the `NaN` and `NaT` roundtripping and the `from_dict` validation to prompt users to give more specific dicts.

jsignell:
I think I have done all the things that you mentioned and I added `DataArray.from_dict()`.

xarray/core/dataarray.py
```diff
@@ -1085,6 +1085,60 @@ def to_masked_array(self, copy=True):
...
+        -------
+        obj : xarray.DataArray
+
+        """
+        coords = None
+        if 'coords' in d.keys():
+            try:
+                coords = OrderedDict([(var[0], (var[1]['dims'], 
+                                                var[1]['data'], 
+                                                var[1].get('attrs'))) for var in d['coords'].items()])
```
shoyer: Tuple unpacking is more readable than using indexing: `for k, v in d['coords'].items()`

xarray/core/dataarray.py
```diff
@@ -1085,6 +1085,60 @@ def to_masked_array(self, copy=True):
...
+            except KeyError as e:
+                raise KeyError(
+                    'cannot convert dict when coords are missing {dims_data}'.format(
+                        dims_data=str(e.args[0])))
+        try:
+            obj = cls(d['data'], coords, d.get('dims'), d.get('name'), d.get('attrs'))
+        except KeyError:
+            raise KeyError('cannot convert dict with missing data')
+        return(obj)
```
shoyer: `return` is a statement, not function, so parentheses are extraneous

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
             obj[name] = (dims, data)
         return obj
 
+    def to_dict(self):
+        """
+        Convert this dataset to a dictionary following xarray naming conventions.
+        """
+        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': dict(self.dims), 
+             'data_vars': {}}
+        def time_check(val):
```
shoyer: Add a brief note about why we need this.

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
...
+        Returns
+        -------
+        obj : xarray.Dataset
+
+        """
+        import itertools
+
+        try:
+            vars = itertools.chain(d['coords'].items(), d['data_vars'].items())
+        except:
```
shoyer: Avoid blanket `except` clauses -- they can catch too many things going wrong. Instead, I would try using the `.get` argument, e.g., `d.get('coords', {})`

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
...
+            vars = d.items()
+        try:
+            OD = OrderedDict([(var[0], (var[1]['dims'],
```
shoyer: Follow PEP8 for naming. That means `snake_case` for variable names. This one might be appropriately called `variables`.

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
...
+                    dims_data=str(e.args[0])))
+        obj = cls(OD)
+
+        # what if coords aren't dims?
+        if set(['coords', 'dims']).issubset(set(d.keys())):
+            coords = (set(d['coords'].keys()) - set(d['dims']))
+            obj = obj.set_coords(coords)
+
+        if 'attrs' in d.keys():
+            obj.attrs.update(d.get('attrs'))
```
shoyer: Use `d.get('attrs', {})` instead of the conditional

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
...
+        return d
+
+    @classmethod
+    def from_dict(cls, d):
+        """
+        Convert a dictionary into an xarray.Dataset
+
+        Parameters
+        ----------
+        d : dict, with a minimum structure of {'var_0': {'dims': [..], 'data': [..]}, ..}
```
shoyer: It would be good to explain how each field is interpreted.

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
...
+
+        Returns
+        -------
+        obj : xarray.Dataset
+
```
shoyer: Add a "See also" section for `Dataset.to_dict` (and the same in that method)

xarray/test/test_dataset.py
```diff
@@ -1849,6 +1849,141 @@ def test_convert_dataframe_with_many_types_and_multiindex(self):
...
+                          'b':{
+                               'dims': 't',
+                               'data': y
+                               }   
+                          }
+           }
+        self.assertDatasetIdentical(ds, Dataset.from_dict(d))
+
+        # this is kind of a flattened version with no coords, or data_vars
+        d = {'a':{
```
shoyer: You have a few PEP8 style violations -- try running an automated tool like pep8 to catch these: https://pypi.python.org/pypi/pep8

xarray/core/dataset.py
```diff
@@ -1933,6 +1933,69 @@ def from_dataframe(cls, dataframe):
...
+                                        var[1]['data'], 
+                                        var[1].get('attrs'))) for var in vars])
+        except KeyError as e:
+            raise KeyError(
+                'cannot convert dict with missing {dims_data}'.format(
+                    dims_data=str(e.args[0])))
+        obj = cls(OD)
+
+        # what if coords aren't dims?
+        if set(['coords', 'dims']).issubset(set(d.keys())):
```
shoyer: What is `coords` is provided but not `dims`?

xarray/core/dataarray.py
```diff
@@ -1085,6 +1085,60 @@ def to_masked_array(self, copy=True):
...
+                           'data': [..], 
+                           'coords': {'dims': [..], 'data': [..]}}
+
+        Returns
+        -------
+        obj : xarray.DataArray
+
+        """
+        coords = None
+        if 'coords' in d.keys():
```
shoyer: You don't need `.keys()` here -- you can check dictionaries directly. Generally that's considered better style, especially because it can be very slow to check `.keys()` on Python 2 (because it's converted into a list).

xarray/test/test_dataset.py
```diff
@@ -1849,6 +1849,141 @@ def test_convert_dataframe_with_many_types_and_multiindex(self):
...
+        expected = ds.set_coords('b')
+        actual = Dataset.from_dict(expected.to_dict())
+
+        self.assertDatasetIdentical(expected, actual)
+
+        # test some incomplete dicts:
+        # this one has no attrs field, the dims are strings, and x, y are np.arrays
+
+        d = {
+            'coords': {
```
shoyer: Might be nice to consolidate these on fewer lines

shoyer:
This is looking pretty good! I think the main thing this needs now is a short section in the docs on IO methods :). Also, a note in "What's New".

jsignell:
Ok, I made the little changes. I will start working on the docs.

jsignell:
Ok, so I merged with the current master and added documentation and a what's-new note. I hope I did that right. @shoyer I am really new to contributing so thanks for all your help. Let me know if anything needs changing.

doc/io.rst
```diff
@@ -51,14 +51,40 @@ and lets you use xarray objects with Python modules like
    refined, we make no guarantees (at this point) that objects pickled with
    this version of xarray will work in future versions.
```
shoyer: add a more specific target here, e.g.,

    .. _dictionary io:

doc/whats-new.rst
```diff
@@ -20,9 +20,12 @@ v0.8.2 (unreleased)
 
 Enhancements
 ~~~~~~~~~~~~
-
 - New documentation on :ref:`panel transition`. By
   `Maximilian Roos <https://github.com/MaximilianR>`_.
+- New ``Dataset`` and ``DataArray`` methods :py:meth:`to_dict` and
+  :py:meth:`from_dict` to allow easy conversion between dictionaries and
+  xarray objects (:issue:`432`). See :ref:`io` for more details. By
```
shoyer: Use the more specific link, e.g.,

    :ref:`dictionary IO <dictionary io>`

xarray/core/dataarray.py
```diff
@@ -1093,6 +1093,84 @@ def to_masked_array(self, copy=True):
...
+        """
+        coords = None
+        if 'coords' in d:
+            try:
+                coords = OrderedDict([(k, (v['dims'],
+                                           v['data'],
+                                           v.get('attrs')))
+                                      for k, v in d['coords'].items()])
+            except KeyError as e:
+                raise KeyError(
```
shoyer: The user is calling a function here to convert input rather than directly looking things up in a mapping. So we should raise a `ValueError` instead.

xarray/core/dataarray.py
```diff
@@ -1093,6 +1093,84 @@ def to_masked_array(self, copy=True):
...
+                    'cannot convert dict when coords are missing '
+                    '{dims_data}'.format(dims_data=str(e.args[0])))
+        try:
+            obj = cls(d['data'], coords, d.get('dims'), d.get('name'),
+                      d.get('attrs'))
+        except KeyError:
+            raise KeyError('cannot convert dict with missing data')
```
shoyer: It's better to catch exceptions as narrowly as possible. Also, better to be clear we mean the field `'data'` in the dict, not generic "missing data". Try:

    try:
        data = d['data']
    except KeyError:
        raise ValueError("cannot convert dict without the key 'data''")
    else:
        obj = cls(data, coords, ....)

xarray/core/dataset.py
```diff
@@ -1910,6 +1910,94 @@ def from_dataframe(cls, dataframe):
...
+        -------
+        obj : xarray.Dataset
+
+        See also
+        --------
+        xarray.DataArray.from_dict
+        """
+        import itertools
+
+        variables = itertools.chain(d.get('coords', {}).items(),
```
shoyer: This should be in an `else` clause of the `if` statement below, because it isn't used otherwise.

xarray/core/dataset.py
```diff
@@ -1910,6 +1910,94 @@ def from_dataframe(cls, dataframe):
...
+                                    d.get('data_vars', {}).items())
+        if not set(['coords', 'data_vars']).issubset(set(d)):
+            variables = d.items()
+        try:
+            OD = OrderedDict([(k, (v['dims'],
```
shoyer: use a `snake_case` name instead of `OD` here

xarray/core/dataset.py
```diff
@@ -1910,6 +1910,94 @@ def from_dataframe(cls, dataframe):
...
+                                   v['data'],
+                                   v.get('attrs'))) for k, v in variables])
+        except KeyError as e:
+            raise KeyError(
```
shoyer: also should be `ValueError`

xarray/core/dataset.py
```diff
@@ -1910,6 +1910,94 @@ def from_dataframe(cls, dataframe):
...
+        """
+        Convert this dataset to a dictionary following xarray naming
+        conventions.
+
+        Useful for coverting to json.
+        """
+        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': dict(self.dims),
+             'data_vars': {}}
+
+        def time_check(val):
```
shoyer: Let's leave a link to the NumPy bug I mentioned earlier in a comment here.

shoyer:
I have a few minor suggestions, but otherwise this looks very nice!

To ensure that the documentation subpages are built for these methods (which makes the sphinx link work), you need to add them to `doc/api.rst`. These should go in the `IO / Conversion` section.

If you haven't tested the docs locally with `make html`, take a look at the development version of the docs (http://xarray.pydata.org/en/latest/) about 5-10 minutes after I merge this to make sure everything looks right. RST can be tricky to get right.

xarray/core/dataarray.py
```diff
@@ -1128,11 +1129,11 @@ def from_dict(cls, d):
 
         d = {'dims': ('t'), 'data': x}
 
-        d = {'coords': {'t': {'dims': 't', 'data': t,
-                              'attrs': {'units':'s'}}},
-             'attrs': {'title': 'air temperature'},
-             'dims': 't',
-             'data': x,
+        d = {'coords': {'t': {'dims': 't', 'data': t, \
```
jsignell: Is this bad form? It renders more cleanly in the html docs.

doc/io.rst
```diff
@@ -25,7 +25,7 @@ module:
 
 .. ipython:: python
 
-    import cPickle as pickle
+    import pickle as pickle
```
jsignell: python 3 I don't know if you want to change this

jsignell:
I made the docs locally and they look how I expect.

doc/io.rst
shoyer: Good call, let's just do `import pickle`. We're not building the docs on readthedocs with Python 3 yet.

xarray/core/dataarray.py
shoyer: I would use `::` and indentation, which will format this as code in the HTML docs: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#other-points-to-keep-in-mind

shoyer: The `\n` and the like are a little bad in the docstring because they aren't valid Python code.

jsignell: That makes sense

jsignell:
Done!

shoyer:
Thank you @jsignell -- really nice work here!

kwilcox:
@jsignell was the intention to have only python scalars/lists in the dictionary (no numpy generics or arrays)?  Right now attribute values are not being converted and results in a mixed `dict` of the data as a list but attributes still in numpy form.  Thoughts?

jsignell:
Hmm. I hadn't thought much about this. I guess the intention was to have only python scalars/lists, but I don't know if we want to get in the business of converting attribute values since there are so many options for what they could be. What do you think makes the most sense?

kwilcox:
Before I looked at the code I assumed it was going to convert... but that's just me!

As it stands a custom encoder is needed to get a JSON dump from the output, which will be a fairly common use case for this function. See https://gist.github.com/kwilcox/c41834297b1a3b732cae3ee16621f6d0.

In the least maybe a little note in the documentation on how to dump the `to_dict` output to JSON (you also need to include `decode_times=False` because datetime objects are not JSON serializable either).

shoyer:
I think we should try to convert numpy scalars/arrays, because otherwise the data won't convert directly to JSON.  Possibly could take a duck typing approach of looking for `.item()` or `.tolist()` methods?

jsignell:
Ok that makes sense. So times should be strings?

shoyer:
We should recommend `decode_times=False` in the docs but I would still make datetime/timedelta objects -- this feels like something that should be left up to users.

jsignell:
@shoyer is this something I should be working on? I am happy to, just don't know how this normally goes.

shoyer:
Up to you, but yes it would be greatly appreciated if you could work on
this.

PR #918, opened 2016-07-25
Fix dim argument for isel_points/sel_points when it's a pandas.Index

shoyer:


jhamman:
this looks good to me.

PR #930, opened 2016-08-02
WIP: align inside concat

### LINKED ISSUE #927: align() and broadcast() before concat()
I have two arrays with misaligned dimensions x and y, and I want to concatenate them on dimension y.
I can't seem to find any way to do it, because:

1. If I do not invoke align(), it will fail complaining that dimension x is not aligned
2. if I invoke align(), it will create unwanted elements on dimension y

See example:
https://gist.github.com/crusaderky/a96db5b59396d94fe1e22694bc091d55

Am I missing something obvious?
Possibly align() should accept an optional parameter e.g. `exclude==['y'] `?

Thanks in advance

shoyer:
Fixes #927

Still needs tests.

### DIFF
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -3,6 +3,7 @@ import warnings
 import pandas as pd
 
 from . import utils
+from .alignment import partial_align
 from .merge import merge
 from .pycompat import iteritems, OrderedDict, basestring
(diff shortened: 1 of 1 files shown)


shoyer:
This was completed in #963

PR #935, opened 2016-08-02
Fix silently ignored AttributeError in custom accessors

### LINKED ISSUE #933: rc 0.8: register_dataset_accessor silently ignores AttributeError
If an AttributeError is raised at construction time by the accessor, it will be ignored silently:

```python
import xarray as xr

@xr.register_dataset_accessor('stupid_accessor')
class GeoAccessor(object):
    def __init__(self, xarray_obj):
        self._obj = xarray_obj        
        raise AttributeError('Ups')

    def plot(self):
        """Plot data on a map."""
        return 'plotting!'

ds = xr.Dataset({'longitude': np.linspace(0, 10), 'latitude': np.linspace(0, 20)})
ds.stupid_accessor.plot()
```

Will raise:

```
<ipython-input-1-b47ffc33cff2> in <module>()
     12 
     13 ds = xr.Dataset({'longitude': np.linspace(0, 10), 'latitude': np.linspace(0, 20)})
---> 14 ds.stupid_accessor.plot()

/home/mowglie/.pyvirtualenvs/py3/lib/python3.4/site-packages/xarray/core/common.py in __getattr__(self, name)
    192                     return source[name]
    193         raise AttributeError("%r object has no attribute %r" %
--> 194                              (type(self).__name__, name))
    195 
    196     def __setattr__(self, name, value):

AttributeError: 'Dataset' object has no attribute 'stupid_accessor'
```

The problem with this error message is that it doesn't tell you that the problem is in your code.

Interestingly, raising other exceptions like `RuntimeError` will result in the more informative traceback:

```
<ipython-input-1-b575d0257332> in <module>()
     12 
     13 ds = xr.Dataset({'longitude': np.linspace(0, 10), 'latitude': np.linspace(0, 20)})
---> 14 ds.stupid_accessor.plot()

/home/mowglie/.pyvirtualenvs/py3/lib/python3.4/site-packages/xarray/core/extensions.py in __get__(self, obj, cls)
     17             # we're accessing the attribute of the class, i.e., Dataset.geo
     18             return self._accessor
---> 19         accessor_obj = self._accessor(obj)
     20         # Replace the property with the accessor object. Inspired by:
     21         # http://www.pydanny.com/cached-property.html

<ipython-input-1-b575d0257332> in __init__(self, xarray_obj)
      5     def __init__(self, xarray_obj):
      6         self._obj = xarray_obj
----> 7         raise RuntimeError('Ups')
      8 
      9     def plot(self):

RuntimeError: Ups
```

shoyer:
Fixes #933

### DIFF
--- a/xarray/test/test_extensions.py
+++ b/xarray/test/test_extensions.py
@@ -74,3 +74,14 @@ class TestAccessor(TestCase):
         assert array.example_accessor is array.example_accessor
         array_restored = pickle.loads(pickle.dumps(array))
         assert array.identical(array_restored)
+
+    def test_broken_accessor(self):
+        # regression test for GH933
+
+        @xr.register_dataset_accessor('stupid_accessor')
+        class BrokenAccessor(object):
+            def __init__(self, xarray_obj):
+                raise AttributeError('broken')
+
+        with self.assertRaisesRegexp(RuntimeError, 'error initializing'):
+            xr.Dataset().stupid_accessor
--- a/xarray/core/extensions.py
+++ b/xarray/core/extensions.py
@@ -1,5 +1,8 @@
+import traceback
+
 from .dataarray import DataArray
 from .dataset import Dataset
+from .pycompat import PY2
 
 
 class AccessorRegistrationError(Exception):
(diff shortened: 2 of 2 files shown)


xarray/test/test_extensions.py
```diff
@@ -74,3 +74,14 @@ def test_pickle_dataarray(self):
         assert array.example_accessor is array.example_accessor
         array_restored = pickle.loads(pickle.dumps(array))
         assert array.identical(array_restored)
+
+    def test_broken_accessor(self):
+        # regression test for GH933
+
+        @xr.register_dataset_accessor('stupid_accessor')
```
fmaussion: Maybe you want to name it otherwise ;)

Issue #939, opened 2016-08-04
Consider how to deal with the proliferation of decoder options on open_dataset


shoyer:
There are already lots of keyword arguments, and users want even more! (#843)

Maybe we should use some sort of object to encapsulate desired options?

mcgibbon:
We already have the dictionary. Users can make a decode_options dictionary, and then call what they want to with **decode_options.

shoyer:
One advantage to creating classes to encapsulate these options is that it's easier to do error checking on field names and values than a dictionary. Though a dictionary is nice and lightweight.

The simplest thing to do would be to move these options out of `open_dataset` into a single other argument, e.g., `open_dataset(filename, **kwargs)` -> `open_dataset(filename, decode_options=kwargs)`.

mcgibbon:
I would disagree with the form `open_dataset(filename, decode_options=kwargs)` over `open_dataset(filename, **kwargs)`, because the former breaks normal Python style. It would make the documentation for the arguments somewhat awkward ("decode_options is a dictionary which can have any of the following keys [...]"). It also forces the user to use a dictionary instead of having the option to use a dictionary or the regular style of entering kwargs.

What do you mean when you say it's easier to do error checking on field names and values? The xarray implementation can still use fields instead of a dictionary, with the user saying `open_dataset(filename, **kwargs)` if they feel like it. I think I'm not understanding something here.

shoyer:
My concern is that `open_dataset` has [1 required and 12 optional arguments](http://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html#xarray.open_dataset). This is too many to easily understand, and is generally considered [poor software design](http://stackoverflow.com/questions/2244860/when-a-method-has-too-many-parameters). The standard solution is to group related options into immutable objects, e.g., `DecoderOptions` or just `Decoder` (if we want to bundle related methods on the object).

[`pandas.read_csv`](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html) is a more extreme example of the same issue.

mcgibbon:
It is considered poor software design to have 13 arguments in Java and other languages which do not have optional arguments. The same isn't necessarily true of Python, but I haven't seen much discussion or writing on this.

[...]

PR #940, opened 2016-08-04
Don't convert time data to timedelta by default

### LINKED ISSUE #843: Don't convert data with time units to timedeltas by default
Don't convert data with time units to `timedeltas` by default. Most of the time this behavior is not desirable (e.g: wave period data).

@shoyer suggest:

> possibly we should add an explicit toggle for decoding `timedelta`s vs `datetime`s.

xref: https://github.com/pydata/xarray/pull/842#issuecomment-216933269

ocefpaf:
I don't really like this PR... Too much change for a simple thing. I may try again soon.

Closes #843

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -80,8 +80,9 @@ def _validate_dataset_names(dataset):
 
 
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
-                 mask_and_scale=True, decode_times=True,
-                 concat_characters=True, decode_coords=True, engine=None,
+                 mask_and_scale=True, decode_datetimes=True,
+                 decode_timedeltas=False, concat_characters=True,
+                 decode_coords=True, engine=None,
                  chunks=None, lock=None, drop_variables=None):
     """Load and decode a dataset from a file or file-like object.
 
@@ -107,9 +108,12 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         `missing_value` attribute contains multiple values a warning will be
         issued and all array values matching one of the multiple values will
         be replaced by NA.
-    decode_times : bool, optional
+    decode_datetimes : bool, optional
         If True, decode times encoded in the standard NetCDF datetime format
         into datetime objects. Otherwise, leave them encoded as numbers.
+    decode_timedeltas : bool, optional
+        If True, decode time data encoded in the standard NetCDF datetime format
+        into timedelta objects. Otherwise, leave them encoded as numbers.
     concat_characters : bool, optional
         If True, concatenate along the last dimension of character arrays to
         form string arrays. Dimensions will only be concatenated over (and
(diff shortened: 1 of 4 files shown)


ocefpaf:
This is ready for review.

[Here](http://nbviewer.jupyter.org/gist/ocefpaf/cabdcb27a5ef7da1b6110327a5f03e17) is an example of this PR in action showing how plotting and working with periods gets easier with `floats` instead of `timedeltas`.

BTW, in light of https://github.com/pydata/xarray/issues/939#issue-169274464, I wonder if `decode_timedeltas` and `decode_datetimes` (or even the original `decode_times`) are needed at all. (Just defending myself as I don't really want a new keyword argument but a better default for time data in general :grimacing:)

Maybe I am being thick and I don't know enough use cases data but I cannot see why someone might want to convert time data (which most of the time represent periods) to `timedeltas`. That breaks from the original data units and forces the user to compute it back and convert the `dtype` too before for plotting, etc.

Regarding time coordinate itself I understand that `xarray`, due to the pandas nature of the index, needs to fail to convert time when the calendar is not supported by pandas. Maybe, instead of throwing an error and asking the users to use the option `decode_(date)times=False` in case of failure, `xarray` could issue an warning and return only the "numbers" as if `decode_(date)times` were set to `False`.

I understand, and agree most of the time, that raising erros is better than issuing warnings, and creating an ambiguity in the returns. So maybe this one is harder to change than the former.

shoyer:
Decoding time units into `timedelta64` is useful if you want to be able to use them for arithmetic with datetimes. I agree that this is not always the most useful currently because, e.g., plotting doesn't work very well, and time deltas in nanoseconds are difficult to interpret. Still I'm somewhat reluctant to change the default here.

Either way, we will definitely need to support the original `decode_times` keyword argument for now (with a deprecation warning) so there is a smooth transition.

ocefpaf:
> Decoding time units into `timedelta64` is useful if you want to be able to use them for arithmetic with datetimes.

I get that but my question is how often do users perform do such operations? Again I am biased b/c with my data I never want to do that as it does not make sense. And, when it does make sense, I believe that the price of post conversion is worth the advantages of converting by `default`.

> Still I'm somewhat reluctant to change the default here.

If you want `decode_timedeltas=True` as the `default` I am not sure it is worth the extra keyword then. My goal with this PR is to avoid extra steps. If users need to set it to `True` it is not too different from doing `data.values / 1e9`. If all we get is to change one step for another I prefer to not make the `open_dataset` more complex and leave it as is.

Feel free to close this. I don't have strong feelings about what `xarray` should do by default. I just want to make it more convenient for my uses cases.

shoyer:
It might be worth querying the mailing list for more opinions here.

If the main issue is plotting, you could try fixing that upstream, too! https://github.com/pydata/pandas/issues/8711

ocefpaf:
> If the main issue is plotting, you could try fixing that upstream, too! pydata/pandas#8711

Indeed! That makes sense to matter what is decided here. Thanks for pointing that out. (Not sure if I am up to the challenge though.)

> It might be worth querying the mailing list for more opinions here.

Done!

ocefpaf:
@shoyer the more I think about this the more I don't like the addition of extra keywords. Even though I would like this behavior to be the default one I really do not like the complexity I added here.

Closing this...

PR #947, opened 2016-08-05
Multi-index levels as coordinates

benbovy:
Implements 2, 4 and 5 in #719.

Demo:

```
In [1]: import numpy as np

In [2]: import pandas as pd

In [3]: import xarray as xr

In [4]: index = pd.MultiIndex.from_product((list('ab'), range(2)),
   ...:                                    names= ('level_1', 'level_2'))

In [5]: da = xr.DataArray(np.random.rand(4, 4), coords={'x': index},
   ...:                   dims=('x', 'y'), name='test')

In [6]: da
Out[6]: 
<xarray.DataArray 'test' (x: 4, y: 4)>
array([[ 0.15036153,  0.68974802,  0.40082234,  0.94451318],
       [ 0.26732938,  0.49598123,  0.8679231 ,  0.6149102 ],
       [ 0.3313594 ,  0.93857424,  0.73023367,  0.44069622],
       [ 0.81304837,  0.81244159,  0.37274953,  0.86405196]])
Coordinates:
  * level_1  (x) object 'a' 'a' 'b' 'b'
  * level_2  (x) int64 0 1 0 1
  * y        (y) int64 0 1 2 3

In [7]: da['level_1']
Out[7]: 
<xarray.DataArray 'level_1' (x: 4)>
array(['a', 'a', 'b', 'b'], dtype=object)
Coordinates:
  * level_1  (x) object 'a' 'a' 'b' 'b'
  * level_2  (x) int64 0 1 0 1

In [8]: da.sel(x='a', level_2=1)
Out[8]: 
<xarray.DataArray 'test' (y: 4)>
array([ 0.26732938,  0.49598123,  0.8679231 ,  0.6149102 ])
Coordinates:
    x        object ('a', 1)
  * y        (y) int64 0 1 2 3

In [9]: da.sel(level_2=1)
Out[9]: 
<xarray.DataArray 'test' (level_1: 2, y: 4)>
array([[ 0.26732938,  0.49598123,  0.8679231 ,  0.6149102 ],
       [ 0.81304837,  0.81244159,  0.37274953,  0.86405196]])
Coordinates:
  * level_1  (level_1) object 'a' 'b'
  * y        (y) int64 0 1 2 3
```

Some notes about the implementation:

- I slightly modified `Coordinate` so that it allows setting different values for the names of the coordinate and its dimension. There is no breaking change.
- I also added a `Coordinate.get_level_coords` method to get independent, single-index coordinates objects from a MultiIndex coordinate.

Remaining issues:

- `Coordinate.get_level_coords` calls `pandas.MultiIndex.get_level_values` for each level and is itself called each time when indexing and for repr. This can be very costly!! It would be nice to return some kind of lazy index object instead of computing the actual level values.
- repr replace a MultiIndex coordinate by its level coordinates. That can be confusing in some cases (see below). Maybe we can set a different marker than `*`  for level coordinates.

```
In [6]: [name for name in da.coords]
Out[6]: ['x', 'y']

In [7]: da.coords.keys()
Out[7]: 
KeysView(Coordinates:
  * level_1  (x) object 'a' 'a' 'b' 'b'
  * level_2  (x) int64 0 1 0 1
  * y        (y) int64 0 1 2 3)
```

- `DataArray.level_1` doesn't return another `DataArray` object:

```
In [10]: da.level_1
Out[10]: 
<xarray.Coordinate 'level_1' (x: 4)>
array(['a', 'a', 'b', 'b'], dtype=object)
```

- Maybe we need to test the uniqueness of level names at `DataArray` or `Dataset` creation.
 
Of course still needs proper tests and docs...

### DIFF
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -166,6 +166,13 @@ class DatasetCoordinates(AbstractCoordinates):
         self._data._coord_names.update(updated_coord_names)
         self._data._dims = dict(dims)
 
+    def __setitem__(self, key, value):
+        if key in self._data._level_coords:
+            raise ValueError("cannot replace MultiIndex level %r, replace %r "
+                             "coordinate instead"
+                             % (key, self._data._level_coords[key]))
+        return super(DatasetCoordinates, self).__setitem__(key, value)
+
     def __delitem__(self, key):
         if key in self:
             del self._data[key]
@@ -208,6 +215,13 @@ class DataArrayCoordinates(AbstractCoordinates):
     def to_dataset(self):
         return self._to_dataset()
 
+    def __setitem__(self, key, value):
+        if key in self._data._level_coords:
+            raise ValueError("cannot replace MultiIndex level %r, replace %r "
+                             "coordinate instead"
+                             % (key, self._data._level_coords[key]))
+        return super(DataArrayCoordinates, self).__setitem__(key, value)
+
     def __delitem__(self, key):
         if key in self.dims:
             raise ValueError('cannot delete a coordinate corresponding to a '
@@ -216,8 +230,7 @@ class DataArrayCoordinates(AbstractCoordinates):
 
 
 class DataArrayLevelCoordinates(AbstractCoordinates):
-    """Dictionary like container for DataArray multi-index
-    level coordinates.
+    """Dictionary like container for DataArray MultiIndex level coordinates.
     """
     def __init__(self, dataarray):
         self._data = dataarray
@@ -228,7 +241,10 @@ class DataArrayLevelCoordinates(AbstractCoordinates):
 
     @property
     def variables(self):
-        return Frozen(self._data._level_coords)
+        level_coords = OrderedDict(
+            (k, self._data[v].variable.get_level_coord(k))
+            for k, v in self._data._level_coords.items())
+        return Frozen(level_coords)
 
 
 class Indexes(Mapping, formatting.ReprMixin):
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -151,12 +151,6 @@ class DatasetCoordinates(AbstractCoordinates):
     def _update_coords(self, coords):
         from .dataset import calculate_dimensions
 
-        for key in coords:
-            if key in self._data._level_coords:
-                raise ValueError("%r is already a MultiIndex level of "
-                                 "coordinate %r"
-                                 % (key, self._data._level_coords[key]))
-
         variables = self._data._variables.copy()
         variables.update(coords)
 
@@ -201,12 +195,6 @@ class DataArrayCoordinates(AbstractCoordinates):
 
     def _update_coords(self, coords):
         from .dataset import calculate_dimensions
-
-        for key in coords:
-            if key in self._data._level_coords:
-                raise ValueError("%r is already a MultiIndex level of "
-                                 "coordinate %r"
-                                 % (key, self._data._level_coords[key]))
 
         dims = calculate_dimensions(coords)
         if set(dims) != set(self.dims):
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -166,13 +166,6 @@ class DatasetCoordinates(AbstractCoordinates):
         self._data._coord_names.update(updated_coord_names)
         self._data._dims = dict(dims)
 
-    def __setitem__(self, key, value):
-        if key in self._data._level_coords:
-            raise ValueError("cannot replace MultiIndex level %r, replace %r "
-                             "coordinate instead"
-                             % (key, self._data._level_coords[key]))
-        return super(DatasetCoordinates, self).__setitem__(key, value)
-
     def __delitem__(self, key):
         if key in self:
             del self._data[key]
@@ -214,13 +207,6 @@ class DataArrayCoordinates(AbstractCoordinates):
 
     def to_dataset(self):
         return self._to_dataset()
-
-    def __setitem__(self, key, value):
-        if key in self._data._level_coords:
-            raise ValueError("cannot replace MultiIndex level %r, replace %r "
-                             "coordinate instead"
-                             % (key, self._data._level_coords[key]))
-        return super(DataArrayCoordinates, self).__setitem__(key, value)
 
     def __delitem__(self, key):
         if key in self.dims:
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -245,8 +245,8 @@ class DataArrayLevelCoordinates(AbstractCoordinates):
     @property
     def variables(self):
         level_coords = OrderedDict(
-            (k, self._data[v].variable.get_level_coord(k))
-            for k, v in self._data._level_coords.items())
+            (k, self._data[v].variable.get_level_variable(k))
+             for k, v in self._data._level_coords.items())
         return Frozen(level_coords)
 
 
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -151,6 +151,12 @@ class DatasetCoordinates(AbstractCoordinates):
     def _update_coords(self, coords):
         from .dataset import calculate_dimensions
 
+        for key in coords:
+            if key in self._data._level_coords:
+                raise ValueError("%r is already a MultiIndex level of "
+                                 "coordinate %r"
+                                 % (key, self._data._level_coords[key]))
+
         variables = self._data._variables.copy()
         variables.update(coords)
 
@@ -195,6 +201,12 @@ class DataArrayCoordinates(AbstractCoordinates):
 
     def _update_coords(self, coords):
         from .dataset import calculate_dimensions
+
+        for key in coords:
+            if key in self._data._level_coords:
+                raise ValueError("%r is already a MultiIndex level of "
+                                 "coordinate %r"
+                                 % (key, self._data._level_coords[key]))
 
         dims = calculate_dimensions(coords)
         if set(dims) != set(self.dims):
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -231,6 +231,9 @@ class DataArrayCoordinates(AbstractCoordinates):
 
 class DataArrayLevelCoordinates(AbstractCoordinates):
     """Dictionary like container for DataArray MultiIndex level coordinates.
+
+    Used for attribute style lookup. Not returned directly by any
+    public methods.
     """
     def __init__(self, dataarray):
         self._data = dataarray
--- a/xarray/core/coordinates.py
+++ b/xarray/core/coordinates.py
@@ -215,6 +215,22 @@ class DataArrayCoordinates(AbstractCoordinates):
         del self._data._coords[key]
 
 
+class DataArrayLevelCoordinates(AbstractCoordinates):
+    """Dictionary like container for DataArray multi-index
+    level coordinates.
+    """
+    def __init__(self, dataarray):
+        self._data = dataarray
+
+    @property
+    def _names(self):
+        return set(self._data._level_coords)
+
+    @property
+    def variables(self):
+        return Frozen(self._data._level_coords)
+
+
 class Indexes(Mapping, formatting.ReprMixin):
     """Ordered Mapping[str, pandas.Index] for xarray objects.
     """
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -429,7 +429,7 @@ class DataArray(AbstractArray, BaseDataObject):
         level_coords = OrderedDict()
         for cname, var in self._coords.items():
             if var.ndim == 1:
-                level_names = var.to_coord().level_names
+                level_names = var.to_index_variable().level_names
                 if level_names is not None:
                     dim = var.dims[0]
                     level_coords.update({lname: dim for lname in level_names})
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -420,10 +420,21 @@ class DataArray(AbstractArray, BaseDataObject):
 
     @property
     def _level_coords(self):
+        """Return a mapping of all MultiIndex levels and their corresponding
+        coordinate name. Raise a `ValueError`` if two or more levels have
+        the same name.
+        """
         level_coords = OrderedDict()
-        for name, var in self._coords.items():
+        for cname, var in self._coords.items():
             if var.ndim == 1:
-                level_coords.update(var.to_coord().get_level_coords())
+                level_names = var.to_coord().level_names
+                if level_names is not None:
+                    duplicate_names = set(level_names) & set(level_coords)
+                    if duplicate_names:
+                        raise ValueError("duplicate MultiIndex level names %r"
+                                         % duplicate_names)
+                    dim = var.dims[0]
+                    level_coords.update({lname: dim for lname in level_names})
         return level_coords
 
     def __getitem__(self, key):
@@ -431,11 +442,10 @@ class DataArray(AbstractArray, BaseDataObject):
             from .dataset import _get_virtual_variable
 
             try:
-                var = self._coords.get(key)
-                if var is None:
-                    var = self._level_coords[key]
+                var = self._coords[key]
             except KeyError:
-                _, key, var = _get_virtual_variable(self._coords, key)
+                _, key, var = _get_virtual_variable(
+                    self._coords, key, self._level_coords)
 
             return self._replace_maybe_drop_dims(var, name=key)
         else:
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -14,7 +14,8 @@ from . import ops
 from . import utils
 from .alignment import align
 from .common import AbstractArray, BaseDataObject, squeeze
-from .coordinates import DataArrayCoordinates, Indexes
+from .coordinates import (DataArrayCoordinates, DataArrayLevelCoordinates,
+                          Indexes)
 from .dataset import Dataset
 from .pycompat import iteritems, basestring, OrderedDict, zip
 from .variable import (as_variable, Variable, as_compatible_data, IndexVariable,
@@ -421,9 +422,8 @@ class DataArray(AbstractArray, BaseDataObject):
     def _level_coords(self):
         level_coords = OrderedDict()
         for name, var in self._coords.items():
-            if name not in self.dims:
-                continue
-            level_coords.update(var.to_coord().get_level_coords())
+            if var.ndim == 1:
+                level_coords.update(var.to_coord().get_level_coords())
         return level_coords
 
     def __getitem__(self, key):
@@ -455,7 +455,7 @@ class DataArray(AbstractArray, BaseDataObject):
     @property
     def _attr_sources(self):
         """List of places to look-up items for attribute-style access"""
-        return [self.coords, self._level_coords, self.attrs]
+        return [self.coords, DataArrayLevelCoordinates(self), self.attrs]
 
     def __contains__(self, key):
         return key in self._coords
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -431,18 +431,13 @@ class DataArray(AbstractArray, BaseDataObject):
     @property
     def _level_coords(self):
         """Return a mapping of all MultiIndex levels and their corresponding
-        coordinate name. Raise a `ValueError`` if two or more levels have
-        the same name.
+        coordinate name.
         """
         level_coords = OrderedDict()
         for cname, var in self._coords.items():
             if var.ndim == 1:
                 level_names = var.to_coord().level_names
                 if level_names is not None:
-                    duplicate_names = set(level_names) & set(level_coords)
-                    if duplicate_names:
-                        raise ValueError("duplicate MultiIndex level names %r"
-                                         % duplicate_names)
                     dim = var.dims[0]
                     level_coords.update({lname: dim for lname in level_names})
         return level_coords
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -71,6 +71,7 @@ def _infer_coords_and_dims(shape, coords, dims):
             new_coords[dim] = default_index_coordinate(dim, size)
 
     sizes = dict(zip(dims, shape))
+    level_names = {}
     for k, v in new_coords.items():
         if any(d not in dims for d in v.dims):
             raise ValueError('coordinate %s has dimensions %s, but these '
@@ -82,6 +83,15 @@ def _infer_coords_and_dims(shape, coords, dims):
                 raise ValueError('conflicting sizes for dimension %r: '
                                  'length %s on the data but length %s on '
                                  'coordinate %r' % (d, sizes[d], s, k))
+
+        if v.ndim == 1:
+            idx_level_names = v.to_coord().level_names or []
+            for n in idx_level_names:
+                if n in level_names:
+                    raise ValueError('found duplicate MultiIndex level '
+                                     'name %r for coordinates %r and %r'
+                                     % (n, k, level_names[n]))
+                level_names[n] = k
 
     return new_coords, dims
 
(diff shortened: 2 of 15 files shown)


xarray/core/dataset.py
```diff
@@ -438,6 +438,16 @@ def _subset_with_all_valid_coords(self, variables, coord_names, attrs):
...
 
+    @property
+    def _level_coords(self):
+        level_coords = OrderedDict()
+        for name in self._coord_names:
+            var = self.variables[name]
+            if name != var.dims[0]:
+                continue
+            level_coords.update(var.to_coord().get_level_coords())
+        return level_coords
```
jhamman: Am I missing something here? Wouldn't this also work without the `continue` statement?

```Python
for name in self._coord_names:
    var = self.variables[name]
    if name == var.dims[0]:
        level_coords.update(var.to_coord().get_level_coords())
    return level_coords
```

benbovy:
In the example above, `DataArray.level_1` now returns a `DataArray` object, although I haven't found another way than creating a new `coordinates.DataArrayLevelCoordinates` class for this.

I wasn't happy with the initial implementation of using levels in `.sel`, it really added too much complexity. I re-implemented it and now I find it much cleaner. However, it is not possible anymore to mix level indexers and non-dict dimension indexers, e.g.,  `da.sel(x='a', level_2=1)` doesn't work but `da.sel(x={'level_1': 'a'}, level_2=1)` does, which is already quite flexible though.

shoyer:
This is very exciting to see!

A few thoughts on implementation:

Instead of always creating a dictionary of level coordinates, I would add an attribute `level_names` to Coordinate, which would default to `None` and be set to a tuple of `MultiIndex.names` if a Coordinate is created from a MultiIndex. This would make make checking for multi-index levels very cheap, even if we do need to iterate through every coordinate to find them.

It's much cheaper to call `.get_level_values()` after slicing a MultiIndex than before, e.g.,
```
In [12]: idx = pd.MultiIndex.from_product([np.linspace(0, 1, num=500), np.arange(1000)])

In [13]: %timeit idx.get_level_values(0)[:10]
1000 loops, best of 3: 1.28 ms per loop

In [14]: %timeit idx[:10].get_level_values(0)
10000 loops, best of 3: 101 µs per loop
```
It's even more extreme for larger indexes. If possible, we should use something closer to this approach when formatting coordinates. 

> However, it is not possible anymore to mix level indexers and non-dict dimension indexers, e.g., da.sel(x='a', level_2=1) doesn't work but da.sel(x={'level_1': 'a'}, level_2=1) does, which is already quite flexible though.

I would actually be happy to disallow both, which might be even easy. It seems like a fine rule to say that you cannot call `.sel` on both a level and the dimension name at the same time. Actually, if we check uniqueness of level names at Dataset/DataArray creation (which is a good idea!), there is not much need for level indexing with a dictionary at all.

shoyer:
I would suggest putting the logic to create new variables for levels in the private `_get_virtual_variable` in `dataset.py`. We already call this function for creating variables on demand in operations like `ds['time.month']`, so it's already called in all the right places (even in `ds.coords`, and so on). We could simply extend it to also check for MultiIndex levels and build those variables on demand, too, but only when necessary.

This could get us most of the way there, but there are still a few things to watch out for:

1. What happens when you write `ds.coords['level_1'] = ...`? With the current implementation, I think this would create a new variable `level_1`. In an ideal world, maybe this would replace the MultiIndex level? For now, it is probably better to raise an error and note that the MultiIndex should be modified instead.
2. Should levels appear in `ds.keys()` or `ds.coords.keys()`? If we're printing them in the repr as peers to dimension coordinates, then maybe the answer should be yes? It could be confusing to have both the redundant levels and the multi-index coordinate in there, though. So maybe it's simpler to avoid changes here.

shoyer:
I'm conflicted about how to handle the repr. On the one hand, I like how `*` indicates indexable variables. On the other hand, it should indeed be clear that these are MultiIndex levels, not dimensions in their own right (especially if they don't appear in `ds.coords.keys()` and the like). So maybe something closer to what we had before would be better.

Let me try to sketch out some concrete proposals to encourage the peanut gallery to speak up:

Option 1: no special indicator for the MultiIndex:
```
Coordinates:
  * level_1  (x) object 'a' 'a' 'b' 'b'
  * level_2  (x) int64 0 1 0 1
  * y        (y) int64 0 1 2 3
```

Option 2: both MultiIndex and levels in repr:
```
Coordinates:
  * x        (x) MultiIndex
  * level_1  (x) object 'a' 'a' 'b' 'b'
  * level_2  (x) int64 0 1 0 1
  * y        (y) int64 0 1 2 3
```

Option 3: both MultiIndex and levels in repr, different symbol for levels:
```
Coordinates:
  * x        (x) MultiIndex
  - level_1  (x) object 'a' 'a' 'b' 'b'
  - level_2  (x) int64 0 1 0 1
  * y        (y) int64 0 1 2 3
```

Option 4: both MultiIndex and levels in repr, different symbol for levels, with indentation:
```
Coordinates:
  * x          (x) MultiIndex
    - level_1  (x) object 'a' 'a' 'b' 'b'
    - level_2  (x) int64 0 1 0 1
  * y          (y) int64 0 1 2 3
```

A separate question (if we pick one of options 2-4) is how to represent the `MultiIndex` dtype and values (everything after the dimension name):

Option A: `MultiIndex` (as shown above)
Option B: `MultiIndex[level_0, level_1]`
Option C: `object MultiIndex`
Option D: `object MultiIndex[level_0, level_1]`
Option E: `MultiIndex ('a', 0) ('a', 1) ('b', 0) ('b', 1)`
Option F: `object ('a', 0) ('a', 1) ('b', 0) ('b', 1)` (current repr)

The tradeoffs here are whether or not we include the exact dtype information (`object`), and how explicitly/redundantly we display the values.

I'm currently leaning toward Option 3A, but I don't have a strong opinion.

xarray/core/variable.py
```diff
@@ -1063,20 +1063,30 @@ class Coordinate(Variable):
...
 
     Coordinates must always be 1-dimensional. In addition to Variable methods
     and properties (attributes, encoding, broadcasting), they support some
     pandas.Index methods directly (e.g., get_indexer), even though pandas does
     not (yet) support duck-typing for indexes.
     """
 
-    def __init__(self, name, data, attrs=None, encoding=None, fastpath=False):
-        super(Coordinate, self).__init__(name, data, attrs, encoding, fastpath)
+    def __init__(self, name, data, attrs=None, encoding=None,
```
shoyer: I would slightly rather change the signature from `name` -> `dims` in the first argument (to match `Variable`), and add `name=None` as an optional parameter at the end (before `fastpath`). This could possibly break use of `Coordinate(name=..., data=...)` but I expect that usage is relatively rare -- this part of the API is usually not touched by users.

Possibly it would be a good idea to rename `Coordinate` -> `IndexVariable`, and change it's intended usage to cover anytime someone wants to represent a `pandas.Index` in xarray. This could be useful for anyone who needs support for custom pandas types like Period or Categorical, even if they won't actually use the array for indexing.

benbovy:
> It seems like a fine rule to say that you cannot call .sel on both a level and the dimension name at the same time

Agreed. I have always the tendency to want to make things as flexible as possible, but this is definitely not needed here.

> I'm currently leaning toward Option 3A, but I don't have a strong opinion.

I'm also +1 for Option 3A.  Maybe one little argument in favor of Options 3E or 3F is that they still show a consistent repr when a scalar is returned from the multi-index (see below), even though I don't like how they would display duplicate information.

```
>>> da.sel(level_1='a', level_2=0)
<xarray.DataArray 'test' (y: 4)>
array([ 0.66068869,  0.20374398,  0.43967166,  0.09447679])
Coordinates:
    x        object ('a', 0)
  * y        (y) int64 0 1 2 3
```

> Possibly it would be a good idea to rename Coordinate -> IndexVariable

As a recent xarray user, I indeed remember that I initially found confusing to have Dataset or DataArray "coordinates" that can be either `Coordinate`  or `Variable` objects. I find that the name `IndexVariable` is more representative of the object.

shoyer:
> As a recent xarray user, I indeed remember that I initially found confusing to have Dataset or DataArray "coordinates" that can be either Coordinate or Variable objects. I find that the name IndexVariable is more representative of the object.

Sounds good, I will do this in a separate PR.

benbovy:
I just made some updates.

> I would add an attribute level_names to Coordinate

Done.

> It's much cheaper to call .get_level_values() after slicing a MultiIndex than before

Done, although it currently slices an arbitrary number (30) of first elements rather than calculating the number of elements needed for display.

> It seems like a fine rule to say that you cannot call .sel on both a level and the dimension name at the same time

Done.

> check uniqueness of level names at Dataset/DataArray creation (which is a good idea!)

I tried but it broke some existing tests. It actually triggered data loading of Coordinate objects (via calls to `to_index()`). I need to further investigate this.

> there is not much need for level indexing with a dictionary at all

Right, but in the current implementation this is still used internally.

> I would suggest putting the logic to create new variables for levels in the private _get_virtual_variable in dataset.py. [...] If possible, it would be nice if ds['time.day'] works even if time is a multi-index level.

Done. It should also work with multi-index levels although not tested yet.

> I'm currently leaning toward Option 3A...

I've chosen option 3A for the repr, but I can change it depending on others' opinions.

> What happens when you write ds.coords['level_1'] = ...? [...] probably better to raise an error and note that the MultiIndex should be modified instead.

Done.

> Should levels appear in ds.keys() or ds.coords.keys()? 

They don't appear in there. If we keep Option 3A for the repr, I also think that we can avoid changes here.

xarray/core/formatting.py
```diff
@@ -186,9 +186,23 @@ def _summarize_var_or_coord(name, var, col_width, show_values=True,
...
     return front_str + values_str
 
 
+def _summarize_coord_multiindex(coord, col_width, marker):
+    first_col = pretty_print(u'  %s %s ' % (marker, coord.name), col_width)
+    return u'%s(%s) MultiIndex' % (first_col, unicode_type(coord.dims[0]))
+
+
+def _summarize_coord_levels(coord, col_width, marker):
+    # TODO: maybe slicing based on calculated number of displayed values
```
shoyer: It's almost certainly overkill, but you could write:

    max_width = OPTIONS['display_width']
    max_possibly_relevant = max(int(np.ceil(max_width / 2.0)), 1)
    relevant_coord = coord[:max_possibly_relevant]

Either way, you probably do want to slice the coordinate outside the look.

xarray/core/formatting.py
```diff
@@ -186,9 +186,23 @@ def _summarize_var_or_coord(name, var, col_width, show_values=True,
...
+
     return front_str + values_str
 
 
+def _summarize_coord_multiindex(coord, col_width, marker):
+    first_col = pretty_print(u'  %s %s ' % (marker, coord.name), col_width)
+    return u'%s(%s) MultiIndex' % (first_col, unicode_type(coord.dims[0]))
+
+
+def _summarize_coord_levels(coord, col_width, marker):
```
shoyer: Maybe set a default `marker=u'-'` here?

xarray/core/groupby.py
```diff
@@ -279,7 +279,7 @@ def _yield_binary_applied(self, func, other):
                 raise TypeError('GroupBy objects only support binary ops '
                                 'when the other argument is a Dataset or '
                                 'DataArray')
-            except KeyError:
+            except (KeyError, ValueError):
```
shoyer: how does `ValueError` get raised here?

xarray/core/indexing.py
```diff
@@ -214,6 +214,41 @@ def convert_label_indexer(index, label, index_name='', method=None,
...
+    invalid = [k for k in indexers
+               if k not in data_obj.dims and k not in data_obj._level_coords]
+    if invalid:
+        raise ValueError("dimensions or multi-index levels %r do not exist"
+                         % invalid)
+
+    level_indexers = {}
+    dim_indexers = {}
+    for key, label in iteritems(indexers):
+        dim = data_obj[key].dims[0]
```
shoyer: use tuple unpacking: `dim, = data_obj[key].dims`

xarray/core/indexing.py
```diff
@@ -214,6 +214,41 @@ def convert_label_indexer(index, label, index_name='', method=None,
...
+    into a single, dictionary indexer for that dimension (Raise a ValueError
+    if it is not possible).
+    """
+    invalid = [k for k in indexers
+               if k not in data_obj.dims and k not in data_obj._level_coords]
+    if invalid:
+        raise ValueError("dimensions or multi-index levels %r do not exist"
+                         % invalid)
+
+    level_indexers = {}
```
shoyer: could make `level_indexers` a `defaultdict(dict)`

xarray/core/coordinates.py
```diff
@@ -209,13 +216,38 @@ def _to_dataset(self, shallow_copy=True):
...
+
     def __delitem__(self, key):
         if key in self.dims:
             raise ValueError('cannot delete a coordinate corresponding to a '
                              'DataArray dimension')
         del self._data._coords[key]
 
 
+class DataArrayLevelCoordinates(AbstractCoordinates):
+    """Dictionary like container for DataArray MultiIndex level coordinates.
```
shoyer: Probably good to clarify "Used for attribute style lookup. Not returned directly by any public methods."

xarray/core/groupby.py
benbovy: `indexing.get_dim_indexers()` checks for invalid indexers and raises a `ValueError` if any (https://github.com/benbovy/xarray/blob/52b88d8660e2af7ab823954009b913e0781e9bfd/xarray/core/indexing.py#L228)

I set a `ValueError` given similar checks that I've seen in `Dataset.isel()` and `Variable.isel()`. I don't know if a `KeyError` would be more appropriate?

benbovy: I also don't know if it is the best place for that check.

shoyer: OK, makes sense. Given that we want to catch this, maybe better to make an `IndexingError` exception type that subclasses ValueError to make this more explicitly. There are probably a few existing errors that could be switched to this.

xarray/core/dataset.py
```diff
@@ -210,6 +224,7 @@ def __init__(self, data_vars=None, coords=None, attrs=None,
             coords = {}
         if data_vars is not None or coords is not None:
             self._set_init_vars_and_dims(data_vars, coords, compat)
+            self._check_multiindex_level_names()
```
shoyer: Rather than putting this check in the `DataArray` and `Dataset` constructors, let's calling this from the `Variable` constructor instead. Probably the cleanest place to put this check in `PandasIndexAdapter`, which is already used to wrap the data from all pandas Indexes:
https://github.com/pydata/xarray/blob/master/xarray/core/indexing.py#L434

benbovy: mmh not sure to understand as this checks for unique multi-index level names among all the (multi-index) coordinates of a new `DataArray` or a `Dataset` object. I guess we need to iterate over those coordinates... I can rename it if it's not clear.

shoyer: Ohhh... that makes perfect sense.

In that case, the logic should go into `merge.py`, to ensure these checks will be performed every time variables are modified, too (which doesn't necessarily use the constructor). We should be doing this check on the result of [`merge_variables()`](https://github.com/pydata/xarray/blob/master/xarray/core/merge.py#L97) in `merge_coords_without_align`, `merge_coords` and `merge_core`.

benbovy: OK I see, `Dataset._set_init_vars_and_dims` indeed depends on `merge.py` but I actually didn't go more into that module. I'll take a look.

xarray/core/dataarray.py
```diff
@@ -67,6 +69,7 @@ def _infer_coords_and_dims(shape, coords, dims):
             new_coords[dim] = default_index_coordinate(dim, size)
 
     sizes = dict(zip(dims, shape))
+    level_names = {}
```
shoyer: looks like we can drop this now

benbovy: Oups yep!

benbovy:
Not sure how to write the tests for this PR, as there is quite many small changes spread in the API (e.g., repr, data object and coordinate properties, etc.). Should I write new tests specific to multi-index or should I modify existing tests (e.g., `TestDataset.test_modify_inplace`, `TestDataset.tes_coord_properties`, `TestDataset.test_coords_modify`, etc.) to include the multi-index case?

shoyer:
> Should I write new tests specific to multi-index or should I modify existing tests (e.g., TestDataset.test_modify_inplace, TestDataset.tes_coord_properties, TestDataset.test_coords_modify, etc.) to include the multi-index case?

I would usually lean towards dedicated tests (e.g., `test_modify_multiindex_inplace`) but if it's easier to modify the original tests in a small way (e.g., for repr) feel free to take that approach

xarray/core/coordinates.py
```diff
@@ -151,6 +151,12 @@ def to_dataset(self):
     def _update_coords(self, coords):
         from .dataset import calculate_dimensions
 
+        for key in coords:
```
shoyer: Is this not already checked by all the callers, e.g., in `merge_coords_without_align`?

benbovy: No, currently it is not, though this should be indeed checked there!

xarray/core/groupby.py
benbovy: An `IndexingError` would be indeed more explicit. Maybe add this is in all the right places in another PR?

xarray/core/coordinates.py
benbovy: Added this sentence on the line below...

benbovy:
@shoyer this is ready for another round of review. I don't see any remaining issue, I added some tests and I updated the docs.

xarray/core/coordinates.py
```diff
@@ -208,13 +215,41 @@ def _to_dataset(self, shallow_copy=True):
     def to_dataset(self):
         return self._to_dataset()
 
+    def __setitem__(self, key, value):
```
shoyer: Just define this method once on `AbstractCoordinates` instead of repeating it twice

xarray/core/dataset.py
```diff
@@ -521,6 +551,12 @@ def __setitem__(self, key, value):
         if utils.is_dict_like(key):
             raise NotImplementedError('cannot yet use a dictionary as a key '
                                       'to set Dataset values')
+
+        if key in self._level_coords:
+            raise ValueError("%r is already a MultiIndex level of "
```
shoyer: I think this is already checked by `update` (which in turn calls `merge`).

But this is perhaps worth keeping anyways for the better error message? If so, please add a comment

xarray/core/groupby.py
shoyer: sounds good to me

xarray/core/variable.py
```diff
@@ -1166,16 +1175,39 @@ def to_index(self):
...
+        MultiIndex.
+        """
+        index = self.to_index()
+        if isinstance(index, pd.MultiIndex):
+            return index.names
+        else:
+            return None
+
+    @level_names.setter
+    def level_names(self, value):
```
shoyer: You don't really need this -- if you define a property without a setter, you get `AttributeError` when you try to set it, anyways

xarray/core/formatting.py
```diff
@@ -233,9 +255,24 @@ def summarize_attr(key, value, col_width=None):
...
-                       if mapping else 0)
+def _get_col_items(mapping):
+    """Get all column items to format, including both keys of `mapping`
+    and MultiIndex levels if any.
+    """
+    col_items = []
+    for k, v in mapping.items():
+        col_items.append(k)
+        var = getattr(v, 'variable', v)
+        if isinstance(var._data, PandasIndexAdapter):
```
shoyer: What about simply checking `isinstance(var, IndexVariable)` instead?

xarray/core/dataset.py
benbovy: Just checked and confirm that this is already checked by `update`.

I'd be OK to remove this. Maybe it is a good thing not to have too many different messages for similar errors...

xarray/core/formatting.py
benbovy: I tried to import `indexVariable` in `formatting` but it gives an error. Maybe because the  `variable` module imports `common`, which in turn imports `formatting`. I don't know if there is an other (and better) way to avoid that sort of cyclic import?

shoyer:
Rather than adding an independent `name` attribute to `IndexVariable`, let's deprecate/remove `IndexVariable.name` instead (I can do this in a follow-up PR). Clearly it's useful to be able to use `IndexVariable` objects even for objects that do not correspond to ticks along a dimension, but for such objects, `name` is misleading. `IndexVariable.name` was never more than a convenient shortcut, and it's outlived it's usefulness.

xarray/core/formatting.py
shoyer: Import it at the top of the method. Cyclic imports are difficult to avoid in Python projects like pandas that define lots of methods.

xarray/core/coordinates.py
shoyer: Actually, I think these should also be caught by the checks in `merge.py`

xarray/core/dataarray.py
```diff
@@ -417,14 +421,29 @@ def _item_key_to_dict(self, key):
...
+    def _level_coords(self):
+        """Return a mapping of all MultiIndex levels and their corresponding
+        coordinate name.
+        """
+        level_coords = OrderedDict()
+        for cname, var in self._coords.items():
+            if var.ndim == 1:
+                level_names = var.to_index_variable().level_names
+                if level_names is not None:
+                    dim = var.dims[0]
```
shoyer: Use tuple unpacking instead: `dim, = var.dims`

benbovy:
> Rather than adding an independent name attribute to IndexVariable, let's deprecate/remove IndexVariable.name instead (I can do this in a follow-up PR). Clearly it's useful to be able to use IndexVariable objects even for objects that do not correspond to ticks along a dimension, but for such objects, name is misleading. IndexVariable.name was never more than a convenient shortcut, and it's outlived it's usefulness.

I agree that `name` is misleading here. If we remove `IndexVariable.name`, then we can modify `IndexVariable.get_level_variable` such that it accepts one or more level names and returns an `OrderedDict` of {level_name: IndexVariable} items. That would be even better than the current implementation. Maybe we should then rename `get_level_variable` to `get_levels_as_variables`?

xarray/core/coordinates.py
benbovy: Yep! So I can remove this.

shoyer:
> If we remove IndexVariable.name, then we can modify IndexVariable.get_level_variable such that it accepts one or more level names and returns an OrderedDict of {level_name: IndexVariable} items. That would be even better than the current implementation. Maybe we should then rename get_level_variable to get_levels_as_variables?

I think what `get_level_variable` returns is an independent matter? I don't think it would be good idea to return a dict here because computing level values can be expensive and the levels aren't always necessary (they're stored internally as an Index for each level and integer codes).

BTW, I will be away over the holiday weekend (in the US), but I expect we will probably be able to merge this shortly after I get back.

benbovy:
I've made changes according to your comments.

> I don't think it would be good idea to return a dict here because computing level values can be expensive and the levels aren't always necessary (they're stored internally as an Index for each level and integer codes).

I just thought that returning a dict could be a little more convenient (i.e., using a single call) if we need to get either one particular or several or all level(s) as `IndexVariable` object(s). However, I admit that this is certainly overkill and that it is actually not related to removing the `name` attr.

> BTW, I will be away over the holiday weekend (in the US), but I expect we will probably be able to merge this shortly after I get back.

Happy holidays! (I'll be on holiday next week too).

shoyer:
I think the main (only?) thing left to do here is to remove the `name` argument you added to `IndexVariable`. For now, we can live with some `IndexVariable` objects with an inconsistent `name` (I'll remove the attribute shortly).

benbovy:
> I think the main (only?) thing left to do here is to remove the name argument you added to IndexVariable

Just removed it.

shoyer:
OK, in it goes. Big thanks to @benbovy !

PR #832, opened 2016-04-18
WIP for transitioning from Panel docs

MaximilianR:
A start for some docs on transitioning from pandas Panel to xarray. 

This is some way from the final version - but putting it out there and will iterate.

PR #955, opened 2016-08-09
Fixes to the xarray docs on pandas

shoyer:
This includes follow up to fix formatting in the new panel section from #832 and fixing where we said MultiIndex wasn't supported.

CC MaximilianR


MaximilianR:
Nice! Thank you!

PR #959, opened 2016-08-10
Make DataArray._from_temp_dataset a classmethod

shoyer:
It's a constructor, so it really should not be an instance method.


shoyer:
Actually,  this is not a good idea :)

Issue #552, opened 2015-08-26
Dataset.to_dataframe() loses datetime64 timezone localization


IamJeffG:
Not sure if feature or bug, but definitely made me look twice.

    >>> ds = xray.Dataset({'time': (['time'], [np.datetime64('2014-04-26T05:00:00-0700')])})
    >>> ds.time.values[0]
    numpy.datetime64('2014-04-26T05:00:00.000000000-0700')
    
    >>> ds.to_dataframe().index[0]
    Timestamp('2014-04-26 12:00:00')

I'd expected to see the DataFrame index maintaining its local timezone info:

    Timestamp('2014-04-26 05:00:00-0700', tz='America/Los_Angeles')

That said, the UTC that I actually get back is the same time so math should still work.  I'm opening this issue mostly as a matter for discussion -- feel free to close if you think xray should be pushing users towards UTC.

jhamman:
Is this just the pandas `repr` behavior?

shoyer:
Xray relies on numpy's datetime64 dtype, which for better or worse does not store timezone. Pandas creates its own array type (wrapping numpy) to make timezone localization work properly, but I don't want to go to those lengths in xray. So I'm happy to push users to UTC for now, and wait until upstream improvements make it easier to customize array dtypes (for things like datetime types with timezones).

PR #961, opened 2016-08-11
Update time-series.rst

NicWayand:
Thought it would be helpful to users to know that timezones are not handled here, rather than googling and finding this: https://github.com/pydata/xarray/issues/552

### DIFF
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -60,7 +60,7 @@ native representation of dates to those that fall between the years 1678 and
 2262. When a netCDF file contains dates outside of these bounds, dates will be
 returned as arrays of ``netcdftime.datetime`` objects.
 
-Timezones are not handeled.
+Timezones are not handled.
 
 Datetime indexing
 -----------------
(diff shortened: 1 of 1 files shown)


doc/time-series.rst
```diff
@@ -60,6 +60,8 @@ native representation of dates to those that fall between the years 1678 and
 2262. When a netCDF file contains dates outside of these bounds, dates will be
 returned as arrays of ``netcdftime.datetime`` objects.
 
+Timezones are not handeled.
```
MaximilianR: `handled`

jhamman: @NicWayand - I think this is a point worth adding.  Can you expand on this a bit more though?  What are the implications for this?

jhamman:
@NicWayand - when you have a minute, fix the typo that @MaximilianR mentioned and possibly expand the discussion a bit. Then we'll be happy to merge this in.

NicWayand:
Numpy's datetime64 dtype currently used by xarray does not store time zone as mentioned here #552. To prevent users from making time zone errors upon dataset creation, I think the implied assumption that UTC be used, should be made more apparent in the readthedocs. Hopefully in the future it can be added to datetime64??

shoyer:
NumPy's datetime64 is timezone naive, though indeed using UTC is generally the recommended approach. Storing timezones would need to done with another metadata layer or could be done by storing an object array of `datetime.datetime` or `pandas.Timestamp` objects.

PR #963, opened 2016-08-11
Align broadcast

crusaderky:
- Removed partial_align()
- Added exclude and indexes optional parameters to align() public API
- Added exclude optional parameter to broadcast() public API
- Added various unit tests to check that align() and broadcast() do not perform needless data copies
- broadcast() to automatically align inputs

Note: there is a failed unit test, TestDataset.test_broadcast_nocopy, which shows broadcast on dataset doing a data copy whereas it shouldn't. Could you look into it?

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,6 +24,11 @@ Enhancements
 - New documentation on :ref:`panel transition`. By
   `Maximilian Roos <https://github.com/MaximilianR>`_.
 
+- Added exclude and indexes optional parameters to align(), and exclude optional parameter
+  to broadcast(). By `Guido Imperiale <https://github.com/crusaderky>`_.
+- broadcast() and concat() will now auto-align inputs.
+  By `Guido Imperiale <https://github.com/crusaderky>`_.
+
 Bug fixes
 ~~~~~~~~~
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,10 +24,11 @@ Enhancements
 - New documentation on :ref:`panel transition`. By
   `Maximilian Roos <https://github.com/MaximilianR>`_.
 
-- Added exclude and indexes optional parameters to align(), and exclude optional parameter
-  to broadcast(). By `Guido Imperiale <https://github.com/crusaderky>`_.
-- broadcast() and concat() will now auto-align inputs.
+- Added ``exclude`` and ``indexes`` optional parameters to :py:func:`~xarray.align`,
+  and ``exclude`` optional parameter to :py:func:`~xarray.broadcast`.
   By `Guido Imperiale <https://github.com/crusaderky>`_.
+- :py:func:`~xarray.broadcast` and :py:func:`~xarray.concat` will now auto-align inputs,
+  using ``join=outer``. By `Guido Imperiale <https://github.com/crusaderky>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/xarray/core/alignment.py
+++ b/xarray/core/alignment.py
@@ -342,8 +342,11 @@ def broadcast(*args, **kwargs):
         # Add excluded dims to a copy of dims_map
         array_dims_map = dims_map.copy()
         for dim in exclude:
-            if dim in ds.dims:
-                array_dims_map[dim] = ds.shape[ds.dims.index(dim)]
+            try:
+                array_dims_map[dim] = array.shape[array.dims.index(dim)]
+            except ValueError:
+                # dim not in array.dims
+                pass                
 
         dims = tuple(array_dims_map)
         if not copy and set(dims) == set(array.dims):
@@ -352,15 +355,20 @@ def broadcast(*args, **kwargs):
         data = array.variable.expand_dims(array_dims_map)
         coords = OrderedDict(array.coords)
         coords.update(common_coords)
-        return DataArray(data, coords, dims, name=array.name,
-                         attrs=array.attrs, encoding=array.encoding)
+        res = DataArray(data, coords, dims, name=array.name,
+                        attrs=array.attrs, encoding=array.encoding)
+        if copy and res.data is array.data:
+            res = res.copy()
+        return res
 
     def _broadcast_dataset(ds):
         # Add excluded dims to a copy of dims_map    
         ds_dims_map = dims_map.copy()
         for dim in exclude:
-            if dim in ds.dims:
+            try:
                 ds_dims_map[dim] = ds.dims[dim]
+            except KeyError:
+                pass
 
         if not copy and dict(ds.dims) == dict(ds_dims_map):
             return ds
--- a/xarray/core/alignment.py
+++ b/xarray/core/alignment.py
@@ -9,7 +9,7 @@ from . import ops, utils
 from .common import _maybe_promote
 from .pycompat import iteritems, OrderedDict
 from .utils import is_full_slice, is_dict_like
-from .variable import Variable, Coordinate, broadcast_variables
+from .variable import Variable, Coordinate
 
 
 def _get_joiner(join):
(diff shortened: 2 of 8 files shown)


xarray/core/alignment.py
```diff
@@ -367,6 +391,7 @@ def _broadcast_dataset(ds):
 
 
 def broadcast_arrays(*args):
+    import warnings
```
shoyer: oops! this is probably an indicator that nobody is relying on this deprecated functionality, though ;).

xarray/core/alignment.py
```diff
@@ -322,32 +319,59 @@ def broadcast(*args):
...
+            if dim not in common_coords and dim not in exclude:
                 common_coords[dim] = arg.coords[dim].variable
                 dims_map[dim] = common_coords[dim].size
 
     def _broadcast_array(array):
-        data = array.variable.expand_dims(dims_map)
+        # Add excluded dims to a copy of dims_map
+        array_dims_map = dims_map.copy()
+        for dim in exclude:
+            try:
```
shoyer: Maybe cleaner to check rather than doing the try/except: `if dim in array.dims`.

xarray/core/alignment.py
```diff
@@ -322,32 +319,59 @@ def broadcast(*args):
...
+            var = ds.variables[k]
+            # Add excluded dims to a copy of dims_map    
+            var_dims_map = dims_map.copy()
+            for dim in exclude:
+                try:
+                    var_dims_map[dim] = var.shape[var.dims.index(dim)]
+                except ValueError:
+                    # dim not in var.dims
+                    try:
+                        del var_dims_map[dim]
```
shoyer: I don't think this ever succeeds?

xarray/core/alignment.py
```diff
@@ -322,32 +319,59 @@ def broadcast(*args):
...
+                        attrs=array.attrs, encoding=array.encoding)
+        return res
 
     def _broadcast_dataset(ds):
         data_vars = OrderedDict()
         for k in ds.data_vars:
-            data_vars[k] = ds.variables[k].expand_dims(dims_map)
+            var = ds.variables[k]
+            # Add excluded dims to a copy of dims_map    
+            var_dims_map = dims_map.copy()
```
shoyer: maybe better to write a helper function rather than writing the twice

xarray/core/alignment.py
crusaderky: That would cost double if the dim is present, and would be functionally identical... Up to you though

xarray/core/alignment.py
crusaderky: You're right I'll fix it

xarray/core/alignment.py
shoyer: this is correct, I just didn't follow the logic correctly

shoyer: Looking at how `dims_map` is initialized, we never add add variables in `exclude`  in. So now I'm not sure how `del` could fail.

xarray/test/test_dataarray.py
```diff
@@ -1636,9 +1690,45 @@ def test_broadcast_arrays(self):
...
+        # Test that input data is not copied over in case no alteration is needed
+        x = DataArray([1, 2], coords=[('a', [-1, -2])], name='x')
+        y = DataArray(3, name='y')
+        expected_x2 = DataArray([1, 2], coords=[('a', [-1, -2])], name='x')
+        expected_y2 = DataArray([3, 3], coords=[('a', [-1, -2])], name='y')
+
+        x2, y2 = broadcast(x, y)
+        self.assertDataArrayIdentical(expected_x2, x2)
+        self.assertDataArrayIdentical(expected_y2, y2)
+        assert x2.data is x.data
```
shoyer: It's a little tricky to verify that two arrays share the same data. See here for discussion:
http://stackoverflow.com/questions/11286864/is-there-a-way-to-check-if-numpy-arrays-share-the-same-data

In many cases, looking at `base` works. I wrote a little utility to help with that:
https://github.com/pydata/xarray/blob/765f17d3b31af1228c6c8def8e108f62a27d7260/xarray/test/__init__.py#L222

So maybe this would work: `source_ndarray(x2.data) is source_ndarray(x.data)`

`numpy.shares_memory` is a clean way to do this, but is only available in numpy 1.11+.

xarray/test/test_dataarray.py
```diff
@@ -1636,9 +1690,45 @@ def test_broadcast_arrays(self):
...
+        x = DataArray([[1, 2], [3, 4]], coords=[('a', [-1, -2]), ('b', [3, 4])])
+        y = DataArray([1, 2], coords=[('a', [-1, 20])])
+        expected_x2 = DataArray([[3, 4], [1, 2], [np.nan, np.nan]], coords=[('a', [-2, -1, 20]), ('b', [3, 4])])
+        expected_y2 = DataArray([[np.nan, np.nan], [1, 1], [2, 2]], coords=[('a', [-2, -1, 20]), ('b', [3, 4])])
+        x2, y2 = broadcast(x, y)
+        self.assertDataArrayIdentical(expected_x2, x2)
+        self.assertDataArrayIdentical(expected_y2, y2)
+
+    def test_broadcast_arrays_nocopy(self):
+        # Test that input data is not copied over in case no alteration is needed
```
shoyer: The guarantee is actually a little stronger than this -- the output arrays should still be views of the original data (no copies), even if broadcasting *was* required.

Alignment does require a copy if the indexes differ.

xarray/core/alignment.py
crusaderky: ah - you're right

xarray/test/test_dataarray.py
crusaderky: I changed all tests to use source_ndarray(). I'm still getting a failure in a dataset alignment that seems to be performing an unwarranted copy though. Could you take a look?

xarray/test/test_dataset.py
```diff
@@ -1056,6 +1056,35 @@ def test_align(self):
...
+
+        expected_x2 = Dataset({'foo': DataArray([[1, 2], [3, 4], [np.nan, np.nan]], dims=['x', 'y'], coords={'x': [1, 2, 3], 'y': [3, 4]})})
+        expected_y2 = Dataset({'bar': DataArray([[1, 2], [np.nan, np.nan], [3, 4]], dims=['x', 'y'], coords={'x': [1, 2, 3], 'y': [5, 6]})})
+        self.assertDatasetIdentical(expected_x2, x2)
+        self.assertDatasetIdentical(expected_y2, y2)
+
+    def test_align_nocopy(self):
+        x = Dataset({'foo': DataArray([1, 2, 3], coords={'x': [1, 2, 3]})})
+        y = Dataset({'foo': DataArray([1, 2], coords={'x': [1, 2]})})
+        x2, y2 = align(x, y, join='outer')
```
shoyer: You need to add `copy=False` here -- otherwise we do a copy by default.

shoyer:
can you add a brief note to "what's new"?

I'm going to issue another bug fix release shortly, would be nice to get this in :)

crusaderky:
Updated what's new, fixed unit test bug, and added auto-align to concat().
I plan to add auto-broadcast to it as well eventually, but I realised it's not trivial as when doing it on datasets you must perform it pairwise. I'll leave it for later.

doc/whats-new.rst
```diff
@@ -24,6 +24,11 @@ Enhancements
 - New documentation on :ref:`panel transition`. By
   `Maximilian Roos <https://github.com/MaximilianR>`_.
 
+- Added exclude and indexes optional parameters to align(), and exclude optional parameter
```
shoyer: Use back ticks to denote code:

   ``exclude``

and sphinx directives to like to function docstrings:

    :py:func:`~xarray.align`

doc/whats-new.rst
```diff
@@ -24,6 +24,11 @@ Enhancements
...
+  to broadcast(). By `Guido Imperiale <https://github.com/crusaderky>`_.
+- broadcast() and concat() will now auto-align inputs.
```
shoyer: add "using `join=outer`'

shoyer:
OK, this looks great to me! I especially appreciate all the extra test coverage.

I will merge once the CI tests are green.

shoyer:
Actually, tests were passing on the previous commit (pre doc fixes) so I decided to go ahead.

Thanks Guido!

Issue #525, opened 2015-08-11
support for units


p3trus:
One thing that bother me alot is that pandas lacks good unit support (e.g. quantities, pint, ...)

Is there a chance that xray will support it?

shoyer:
Eventually, I hope so!

Unfortunately, doing this in a feasible and maintainable way will probably require upstream fixes in NumPy. In particular, better support for duck-array types (https://github.com/numpy/numpy/issues/4164) and/or the ability to write units as a custom NumPy dtypes. Both of these are on the NumPy roadmap, though they don't have a timeframe for when that will happen.

rabernat:
Astropy has pretty good units support: 
http://astropy.readthedocs.org/en/latest/units/
Would it be possible to copy what they do?

shoyer:
Unfortunately, the astropy approach uses a numpy.ndarray subclass, which means it's mutually exclusive with dask.array. Otherwise, it does look very nice, though.

mhvk:
@shoyer -  as one who thinks unit support is probably the single best thing astropy has (and is co-maintainer of `astropy.units`), I thought I'd pipe in: why would it be a problem that astropy's `Quantity` is an `ndarray` subclass?  I must admit not having used dask arrays, but since they use `ndarray` internally for the pieces, shouldn't the fact that `Quantity` has the same interface/methods, make it relatively easy to swap `ndarray` for `Quantity` internally?  I'd be quite happy to help think about this (surely it cannot be as bad as it is for `MaskedArray` ;-).

Alternatively, maybe it is easier to tag on the outside rather than the inside. This would also not seem to be that hard, given that astropy's `Quantity` is really just a wrapper around `ndarray` that carries a `Unit` instance. I think the parts that truly wrap might be separated from those that override `ndarray` methods, and would be willing to implement that if there is a good reason (like making dask quantities possible...). It may be that in this case one would not use `Quantity` proper, but rather just the parts of `units` where the real magic happens: in the `Unit` class (which does the unit conversion) and in `quantity_helpers.py` (which tells what unit conversion is necessary for a given operation/function).

shoyer:
@mhvk It would certainly be possible to extend dask.array to handle units, in either of the ways you suggest.

Although you could allow `Quantity` objects inside dask.arrays, I don't like that approach, because static checks like units really should be done only once when arrays are constructed (akin to dtype checks) rather than at evaluation time, and for every chunk. This suggests that tagging on the outside is the better approach.

[...]

Issue #576, opened 2015-09-15
define fill value for where


mathause:
It would be nice if `where` accepts an `other` argument:

```
def where(self, cond, other=np.NaN):
    pass
```

shoyer:
Agreed. I didn't add this only because we don't yet have three-argument broadcasting operations working yet.

For now, a work around (at a slight cost in efficiency) is to use `where` followed by `fillna`.

MaximilianR:
This would be great to add - we just hit this (CC @justinkuosixty). 

>For now, a work around (at a slight cost in efficiency) is to use where followed by fillna.

This works, although if there are already `NaN`s, it requires another `where` to add those back

jhamman:
@shoyer - what do we need in order to get the 3-argument broadcasting working. I'd really like to get this into xarray.

shoyer:
It should already work with apply_ufunc -- we just need to adjust where to
use it. Take a look at the latest version of fillna for an example.

PR #964, opened 2016-08-12
New function for applying vectorized functions for unlabeled arrays to xarray objects

### LINKED ISSUE #770: Internal refactor: create a generic function for applying ufuncs-like functions to xarray objects
It would be awesome to have a generic function for making functions that act like NumPy's generalized universal functions "xarray aware".

What would `xarray.apply_ufunc(func, objs, join='inner', agg_dims=None, drop_dims=None, kwargs=None)` do?
1. If one or more of the provided objects are `Dataset` or `GroupBy` instances, dispatch to specialized loops that call the remainder of `apply_ufunc` repeatedly.
2. `align` all objects along shared labels using the indicated `join` (for some operations, e.g., `where`, a left join is appropriate rather than an inner join).
3. `broadcast` all objects against each other to expand dimensionality along all dimensions except (optionally) those listed in `agg_dims`/`drop_dims`. `drop_dims` should be moved to the end, for consistency with gufunc signatures.
4. Transform `agg_dims` (if provided) into an `axis` argument using `get_axis_num` and insert it into `kwargs`.
5. Apply `func` to the `data` argument of each array to calculate the result using the provided `kwargs`. The result is expected to have all the same dimensions in the provided arrays, except any listed in the `dims` and `drop_dims` arguments.
6. `merge` all coordinate data together (i.e., with an n-ary version of the `Coordinate.merge` method) and add these to the result array.

If any of `args` are not xarray objects (e.g., they're NumPy or dask arrays), they should be skipped in operations that don't apply to them. `xarray.Variable` don't align or have coordinates, for example.

A concrete example of similar functionality in dask.array is [`atop`](https://github.com/dask/dask/blob/0.8.0/dask/array/core.py#L1572). The most similar thing to this that we currently have in xarray are the `_unary_op` and `_binary_op` staticmethods (e.g., [on DataArray](https://github.com/pydata/xarray/blob/v0.7.1/xarray/core/dataarray.py#L1180)), but these only handle one or two arguments, don't handle aggregated dimensions and most importantly, are difficult to apply to new operations.

Here are a few concrete examples of how this could work:

``` python
def average(array, weights, dim=None):
    # still needs a bit of work to make a NaN and dask.array safe version
    # version of np.average 
    return apply_ufunc(np.average, [array, weights], agg_dims=dim)

def where(cond, first, second=None):
    if second is None:
        # need to write where2, a function that looks at first.dtype
        # to infer the appropriate NA sentinel value
        return apply_ufunc(ops.where2, [cond, first])
    else:
        return apply_ufunc(ops.where, [cond, first, second])

def dot(self, other, dim=None):
    if dim is None:
        dim = set(self.dims) ^ set(other.dims)
    return apply_ufunc(ops.tensordot, [self, other], agg_dims=dim)
```

shoyer:
This PR creates new public facing function `xarray.apply` which handles all the logic of applying [numpy generalized universal functions](http://docs.scipy.org/doc/numpy/reference/c-api.generalized-ufuncs.html) to xarray's labelled arrays, including automatic alignment, merging coordinates, broadcasting and reapplying labels to the result.

Note that although we use the gufunc interface here, this works for far more than gufuncs. Any function that handles broadcasting in the usual numpy way will do. See below for examples.

Now that this logic is all in one place, we will even be able to (in a follow-up PR) include hooks for setting output array names and attributes based on input (e.g., to allow third party libraries to add unit support #525).

Xref #770
## Examples

Calculate the vector magnitude of two arguments:

```
def magnitude(a, b):
    func = lambda x, y: np.sqrt(x ** 2 + y ** 2)
    return xr.apply_func(func, a, b)
```

Compute the mean (`.mean`)::

```
def mean(obj, dim):
    # note: apply_ufunc always moves core dimensions to the end
    sig = ([(dim,)], [()])
    kwargs = {'axis': -1}
    return xr.apply_ufunc(np.mean, obj, signature=sig, kwargs=kwargs)
```

Inner product over a specific dimension::

```
def gufunc_inner(x, y):
    result = np.matmul(x[..., np.newaxis, :], y[..., :, np.newaxis])
    return result[..., 0, 0]

def inner_product(a, b, dim):
    sig = ([(dim,), (dim,)], [()])
    return xr.apply_ufunc(gufunc_inner, a, b, signature=sig)
```

Stack objects along a new dimension (like `xr.concat`)::

```
def stack(objects, dim, new_coord):
    sig = ([()] * len(objects), [(dim,)])
    new_coords = [{dim: new_coord}]
    func = lambda *x: np.stack(x, axis=-1)
    return xr.apply_ufunc(func, *objects, signature=sig,
                          new_coords=new_coords,
                          dataset_fill_value=np.nan)
```

Singular value decomposition:

```
def dim_shape(obj, dim):
    # TODO: make this unnecessary, see #921
    try:
        return obj.dims.index(dim)
    except AttributeError:
        return obj.dims[dim]

def svd(obj, dim0, dim1, new_dim='singular_values'):
    sig = ([(dim0, dim1)], [(dim0, new_dim), (new_dim,), (new_dim, dim1)])
    K = min(dim_shape(obj, dim0), dim_shape(obj, dim1))
    new_coords = [{new_dim: np.arange(K)}] * 3
    return xr.apply_ufunc(np.linalg.svd, obj, signature=sig,
                          new_coords=new_coords,
                          kwargs={'full_matrices': False})
```
## Signature/Docstring

```
apply(func, *args, signature=None, join='inner', new_coords=None,
         exclude_dims=frozenset(), dataset_fill_value=None, kwargs=None,
         dask_array='forbidden')

Apply a vectorized function for unlabeled arrays to xarray objects.

The input arguments will be handled using xarray's standard rules for
labeled computation, including alignment, broadcasting, looping over
GroupBy/Dataset variables, and merging of coordinates.

Parameters
----------
func : callable
    Function to call like ``func(*args, **kwargs)`` on unlabeled arrays
    (``.data``). If multiple arguments with non-matching dimensions are
    supplied, this function is expected to vectorize (broadcast) over
    axes of positional arguments in the style of NumPy universal
    functions [1]_.
*args : Dataset, DataArray, GroupBy, Variable, numpy/dask arrays or scalars
    Mix of labeled and/or unlabeled arrays to which to apply the function.
signature : string or triply nested sequence, optional
    Object indicating core dimensions that should not be broadcast on
    the input and outputs arguments. If omitted, inputs will be broadcast
    to share all dimensions in common before calling ``func`` on their
    values, and the output of ``func`` will be assumed to be a single array
    with the same shape as the inputs.

    Two forms of signatures are accepted:
    (a) A signature string of the form used by NumPy's generalized
        universal functions [2]_, e.g., '(),(time)->()' indicating a
        function that accepts two arguments and returns a single argument,
        on which all dimensions should be broadcast except 'time' on the
        second argument.
    (a) A triply nested sequence providing lists of core dimensions for
        each variable, for both input and output, e.g.,
        ``([(), ('time',)], [()])``.

    Core dimensions are automatically moved to the last axes of any input
    variables, which facilitates using NumPy style generalized ufuncs (see
    the examples below).

    Unlike the NumPy gufunc signature spec, the names of all dimensions
    provided in signatures must be the names of actual dimensions on the
    xarray objects.
join : {'outer', 'inner', 'left', 'right'}, optional
    Method for joining the indexes of the passed objects along each
    dimension, and the variables of Dataset objects with mismatched
    data variables:
    - 'outer': use the union of object indexes
    - 'inner': use the intersection of object indexes
    - 'left': use indexes from the first object with each dimension
    - 'right': use indexes from the last object with each dimension
new_coords : list of dict-like, optional
    New coordinates to include on each output variable. Any core dimensions
    on outputs not found on the inputs must be provided here.
exclude_dims : set, optional
    Dimensions to exclude from alignment and broadcasting. Any inputs
    coordinates along these dimensions will be dropped. If you include
    these dimensions on any outputs, you must explicit set them in
    ``new_coords``. Each excluded dimension must be a core dimension in the
    function signature.
dataset_fill_value : optional
    Value used in place of missing variables on Dataset inputs when the
    datasets do not share the exact same ``data_vars``. Only relevant if
    ``join != 'inner'``.
kwargs: dict, optional
    Optional keyword arguments passed directly on to call ``func``.
dask_array: 'forbidden' or 'allowed', optional
    Whether or not to allow applying the ufunc to objects containing lazy
    data in the form of dask arrays. By default, this is forbidden, to
    avoid implicitly converting lazy data.

Returns
-------
Single value or tuple of Dataset, DataArray, Variable, dask.array.Array or
numpy.ndarray, the first type on that list to appear on an input.
```


xarray/core/computation.py
```diff
@@ -0,0 +1,261 @@
...
+        kwargs = {}
+
+    args = deep_align(*args, join=join, copy=False, raise_on_invalid=False)
+
+    coord_variables = [getattr(getattr(a, 'coords', {}), 'variables')
+                       for a in args]
+    coords = merge_coords_without_align(coord_variables)
+    name = result_name(args)
+
+    data_vars = [getattr(a, 'variable') for a in args]
```
MaximilianR: Here & `variables` above could be normal `a.variable` rather than `getattr...`?

xarray/core/computation.py
```diff
@@ -0,0 +1,261 @@
...
+def _calculate_unified_dim_sizes(variables):
+    dim_sizes = OrderedDict()
+
+    for var in variables:
+        try:
+            var_dims = var.dims
+        except AttributeError:
+            continue
+
+        if len(set(var_dims)) < len(var_dims):
```
MaximilianR: Will `var_dims` be undefined if the `try` statement raises?

xarray/core/computation.py
```diff
@@ -0,0 +1,261 @@
+from collections import NamedTuple
```
MaximilianR: I think it's `namedtuple`

MaximilianR:
This looks awesome! Would simplify a lot of the existing op stuff!

xarray/core/computation.py
```diff
@@ -0,0 +1,261 @@
...
+        raise ValueError('some `axis_dims` not found on any input variables: '
+                         '%r' % list(invalid))
+    axis = tuple(broadcast_dims.index(d) for d in axis_dims)
+    n_remaining_axes = len(axis_dims) - len(axis)
+    if n_remaining_axes > 0:
+        valid_core_dims_for_axis = core_dims[:remaining_axes]
+        if not set(valid_core_dims_for_axis) <= axis_dims:
+            raise ValueError('axis dimensions %r have overlap with core '
+                             'dimensions %r, but do not appear at the start'
+                             % (axis_dims, core_dims))
```
MaximilianR: Why does the order of the passed dims matter? (i.e. why not transpose them all into the order that's needed)

xarray/core/computation.py
```diff
@@ -0,0 +1,261 @@
...
+
+    if kwargs is None:
+        kwargs = {}
+
+    if combine_attrs is None:
+        combine_attrs = lambda func, attrs: None
+
+    result_attrs = combine_attrs(func, [getattr(a, 'attrs', {}) for a in args])
+
+    sig = _build_and_check_dims_signature(
```
MaximilianR: Where's this func?

MaximilianR:
When this is done & we can do `where`, I wonder whether 

```python
da[bool_array] = 5
```

...could be sugar for...
```python
da.where(bool_array, 5)
```

i.e. do we get multidimensional indexing for free?

shoyer:
@MaximilianR Two issues come to mind with remapping `da[condition] = other` -> `da = da.where(bool_array, other)`:

1. What should `da[bool_array]` return? For consistency, I think we need to support both. We could alias that to `where`, too (`da[bool_array]` -> `da.where(bool_array)`), but this would be inconsistent with the existing behavior of `da[bool_array]` if `da` and `bool_array` are one-dimensional. This suggests maybe `ds[bool_array]` -> `da.where(bool_array, drop=True)`.

2. What is `other` is not a scalar value, but is a DataArray? Currently, I don't think we align when indexing, but in this case, by necessity we would. For `__getitem__` indexing, it's pretty obvious that alignment should be a left join. For `__setitem__` indexing, I'm not sure. At the least they would be a little different from the defaults for `where` (which does an inner join like most xarray operations by default). Maybe something like: `left_join(da, inner_join(bool_array, other))`? We need to make sure that every xarray assignment like `obj[key]` or `obj[key] = value` (and also `.loc` and `.sel` and so on) works sanely with these alignment rules.

MaximilianR:
Thanks for thinking through these

> This suggests maybe `ds[bool_array] -> da.where(bool_array, drop=True).`

I think that makes sense. `drop=False` would be too confusing

> Maybe something like: `left_join(da, inner_join(bool_array, other))`? 

The way I was thinking about it: both `other` and `bool_array` need a value for every value in `da`. So they both need to be subsets. So something like:

```
assert set(other.dims) =< set(da.dims)
assert set(bool_array.dims) =< set(da.dims)

other, _ = xr.broadcast(other, da)
bool_array, _ = xr.broadcast(bool_array, da)

da.where(bool_array, other)
```

Is that consistent with the joins you were thinking of?

xarray/core/computation.py
shoyer: Nope, this should be fine -- when we catch the `AttributeError`, we `continue` to the next item in the loop.

shoyer:
This is now tested and ready for review. The API could particularly use feedback -- please take a look at the docstring and examples in the first comment. Long desired operations, like a fill value for `where` (#576) and `cumsum` (#791) should now be writable in only a few lines.

I have not yet hooked this up to the rest of xarray's code base, both because the set of changes we will be able to do with this are quite large, and because I'd like to give other contributors a chance to help/test. Note that the general version of `apply_ufunc` can include some significant overhead for doing the dispatch. For binary operations, we will probably want to use the pre-specialized versions (e.g., `apply_dataset_ufunc`).

Finally, given the generality of this operation, I'm considering renaming it from `xr.apply_ufunc` to simply `xr.apply`.

shoyer:
CC @jhamman @rabernat @crusaderky @pwolfram @spencerahill @ajdawson

xarray/core/computation.py
```diff
@@ -0,0 +1,675 @@
...
+                % (type(self).__name__,
+                   list(self.input_core_dims),
+                   list(self.output_core_dims)))
+
+
+def result_name(objects):
+    # type: List[object] -> Any
+    # use the same naming heuristics as pandas:
+    # https://github.com/blaze/blaze/issues/458#issuecomment-51936356
+    names = set(getattr(obj, 'name', None) for obj in objects)
```
MaximilianR: FWIW this could be a set comprehension

xarray/core/computation.py
```diff
@@ -0,0 +1,675 @@
...
+    if signature.n_outputs > 1:
+        return tuple(DataArray(variable, coords, name=name, fastpath=True)
+                     for variable, coords in zip(result_var, result_coords))
+    else:
+        coords, = result_coords
+        return DataArray(result_var, coords, name=name, fastpath=True)
+
+
+def ordered_set_union(all_keys):
+    # type: List[Iterable -> Iterable
```
MaximilianR: FWIW missing `]`

xarray/core/computation.py
```diff
@@ -0,0 +1,675 @@
...
+    signature : string or triply nested sequence, optional
+        Object indicating any core dimensions that should not be broadcast on
+        the input arguments, new dimensions in the output, and/or multiple
+        outputs. Two forms of signatures are accepted:
+        (a) A signature string of the form used by NumPy's generalized universal
+            functions [1], e.g., '(),(time)->()' indicating a function that
+            accepts two arguments and returns a single argument, on which all
+            dimensions should be broadcast except 'time' on the second argument.
+        (a) A triply nested sequence providing lists of core dimensions for each
+            variable, for both input and output, e.g., ([(), ('time',)], [()]).
```
MaximilianR: Currently, the _order_ of the variables is important; is that correct? I guess a future version could have a mapping that wasn't dependent on the order

shoyer: Yes, order of variables matters because that's how we match the signature up with `*args`.

To be clear, there is (currently) no special handling for matching ufunc arguments with `Dataset` variables by name.

MaximilianR:
Would it be possible to write something like [np.einsum](http://docs.scipy.org/doc/numpy/reference/generated/numpy.einsum.html) with xarray named dimensions? 

I _think_ it's possible, by supplying the dimensions to sum over, and broadcasting the others. Similar to the `inner_product` example, but taking `*dims` rather than `dims`. Is that right?

xarray/core/computation.py
```diff
@@ -0,0 +1,675 @@
...
+
+
+def safe_tuple(x):
+    # type: Iterable -> tuple
+    if isinstance(x, basestring):
+        raise ValueError('cannot safely convert %r to a tuple')
+    return tuple(x)
+
+
+class Signature(object):
```
MaximilianR: This is already a [class](https://docs.python.org/3/library/inspect.html#inspect.Signature) in Python3 FYI

xarray/core/computation.py
```diff
@@ -0,0 +1,675 @@
...
+        if dropped_dims:
+            merged = OrderedDict((k, v) for k, v in merged.items()
+                                 if dropped_dims.isdisjoint(v.dims))
+
+        output_coords.append(merged)
+
+    return output_coords
+
+
+def apply_dataarray_ufunc(func, *args, **kwargs):
```
MaximilianR: Out of interest, why all the `*args, **kwargs`, rather than the arg list?

xarray/core/computation.py
shoyer: Switched to `UFuncSignature` to minimize potential confusion.

xarray/core/computation.py
shoyer: I've gone back and forth on this one.

The basic reason is that confusing to need to keep track of which functions can be called like `f(*args)` vs `f(args)`. Most functions we will be wrapping fall in the former category, at least if we want to be able to use real ufuncs (like `np.add`) directly. I'm using lots of `functools.partial` to pass around function objects, and I wanted to avoid thinking about this.

Note that it does cost us ~5 us to parse `**kwargs` manually. This is significant compared to the cost of some operations (e.g., a unary op on `Variable` with only a few elements takes only ~10 us otherwise) but becomes negligible once you start working doing operations that involve one or more `DataArray`/`Dataset` objects (in the 400 us range). I view this as the cost of writing Python -- none of it matters once you have big enough arrays.

Note once we switch to Python 3 only the overhead should go away and everything will look pretty clean, because we can drop `**kwargs`. Even before then, writing out the signature at the top of the docstring ensures that the API docs look right (there are also signature rewriting tricks we could use to get introspection/auto-complete working properly on Python 3, but that's probably overkill).

pwolfram:
@shoyer and others, quick question-- as it turns out I need functionality from #812 for my work.  Is it best to build off that issue (with a half-baked branch) or this one?  I can form a hacky stop-gap solution in the meantime but it is clear I need to finish this off for the long term.  Thanks in advance for the advice.

pwolfram:
Also, should `py.test` work for `stable` and `master`?  I was getting various errors and this seems strange to me-- I suspect it is a configuration issue on my end.

shoyer:
@pwolfram If you don't need dask support, then I would suggest simply trying this PR. I think a slight variation of my `mean` example at the top would work.

If you do need dask support, either just use `da.nancumsum` directly with this PR (along with `dask_array='allowed'`) or we can use `core.ops._dask_or_eager_func` (e.g., as in your PR) to make a version of `nancumsum` that works on both dask and numpy arrays, and then use that version along with this PR.

shoyer:
And yes, `py.test` *should* work on master (not 100% sure for stable, which is only used for pushing doc updates). That's what we test on Travis-CI: https://travis-ci.org/pydata/xarray/builds. I'm happy to help debug, but let's open another issue for that.

pwolfram:
@shoyer, I decided to go with the more robust approach and do the PR properly: #791.

chris-b1:
A few pieces of feedback trying this out.  I'm basically learning xarray as I go (I ran into this right away), so weight appropriately.

Usecase - I have a numba gufunc I want to apply to `DataArray`, e.g. a reduction like this (`std` is just for sake of example)
```
@numba.guvectorize(['void(f8[:], f8[:])'],
                   '(n)->()')
def std_gufunc(arr, out):
    out[0] = np.std(arr)

arr = xr.DataArray(np.random.randn(100, 100, 100), dims=('x', 'y', 'z'))
```

1)  The "obvious" thing doesn't work - maybe catch and show a nicer error message here
```
xr.apply(std_gufunc, arr)
ValueError: dimensions ('x', 'y', 'z') must have the same length as the number of data dimensions, ndim=2
```

2) I personally found the non-string version of `signature` really difficult to wrap my mind around (the below took several tries to get right).  I don't have a concrete suggestion though, maybe only the string form is meant to really be the public api?
```
xr.apply(std_gufunc, arr, signature=([('x',)], [()]))
xr.apply(std_gufunc, arr, signature='(x)->()')
```


3) It would be nice to take advantage of the existing gufunc signature in some way.  Maybe this is a wrapper built on top or `xr.apply`, or expand the api to allow something like this;
```
xr.apply(std_gufunc, arr, dims=('x',))
xr.apply(std_gufunc, arr, dims={'n':'x'})
```

shoyer:
@chris-b1 thanks for giving this is a try! Using this with numba's guvectorize is exactly what I had in mind.

1) Yes, we need a better error here.

2) Agreed, it's really hard to parse a triply nested list. I don't like encouraging writing signature strings though because that artificially restricts dimensions to use strings (and it's weird to program in strings).

Maybe separate arguments for `input_core_dims` and `output_core_dims` would make more sense, e.g., `xr.apply(std_gufunc, arr, input_core_dims=[('x',)])`?

3) Yes, agreed. The main issue with `xr.apply(std_gufunc, arr, dims=('x',))` is that it's not clear what that means if `std_gufunc` is not a gufunc. For example, one might expect it to generate `axis` arguments.

rabernat:
Of course I think this is a fantastic feature which will change the way use use xarray.

I gave it a test run for a problem we come across a lot on the mailing list: estimating a linear trend along one dimension of a dataarray. A short example notebook is here:
https://gist.github.com/rabernat/a0ec6a7e947f2d928615a30f5cb91ee9

Overall it worked as I hoped, but there were a few bumps I had to overcome. My feedback is from a user perspective, regarding the api and documentation

- In the documentation, I would _not_ assume that the user is familiar with Numpy generalized universal functions. A more explicit explanation of the syntax and meaning of the signature in the docstring would be very helpful. It took me lots of trial an error to find the signature that worked.
- The function I wanted to apply, `np.polyfit`, works on the _first_ axis of the array, not the last. This required an extra swap-axis step inside a wrapper function.
- This would not work if the data ndim were > 2, because `np.polyfit` expects a 2D array. So an additional stacking step would also be required.

Perhaps I am not using this as designed, but this was the most obvious example application I could think of.

chris-b1:
@shoyer - I agree on 3) that it might too much to pack in to `xr.apply`.  As one possibility, here's a half-implemented (probably buggy!) wrapper that would allow this:

```python
@xarray_gufunc
@numba.guvectorize(['void(f8[:], f8[:])'], '(n)->()')
def std_gufunc(arr, out):
    out[0] = np.std(arr)

std_gufunc(arr, dims=('x',))
```
https://gist.github.com/chris-b1/d28c6b8e78bf65ef7eb97e1095bc87f2

shoyer:
One of the tricky things with `apply` is that there are a lot of similar but distinct use cases to disambiguate. I'll outline a few of these below.

I'd appreciate feedback on which cases are most essential and which can wait until later (this PR is already getting pretty big).

Also, I'd appreciate ideas for how to make the API more easily understood. We will have extensive docs either way, but `xarray.apply` is probably already in the realm of "too many arguments for one function". The last thing I want to do is to make a swiss army knife so flexible (like (`numpy.nditer`)[http://docs.scipy.org/doc/numpy/reference/generated/numpy.nditer.html]) that nobody uses it because they don't understand how it works.

## How `func` vectorizes

There are two main cases here:

1. Functions already written to vectorize their arguments:

  1. Scalar functions built out of NumPy primitives (e.g., `a + b + c`). These work by default.
  2. Functions that use core dimension referred to by `axis` (e.g., `np.mean`). These work if you set `axis=-1` and put the dimension in the signature, but the API is kind of awkward. You're rather that the wrapper just converts argument like `dim='time'` automatically into `axis=2`. Transposing these core dimensions to the end also feels unnecessary, though maybe not a serious concern given that transposing NumPy arrays involves no memory copies.
  3. Functions that work mostly like gufuncs, but aren't actually (e.g., `np.svd`). This is pretty common, because NumPy ufuncs have some serious limitations (e.g.., they can't handle non-vectorized arguments). These work about as well as we could hope, modulo possible improvements to the `signature` spec.
  4. True gufuncs, most likely written with `numba.guvectorize`. For these functions, we'd like a way to extract/use the signature automatically.

2. Functions for which you only have the inner loop (e.g., `np.polyfit` or `scipy.stats.pearsonr`). Running these is going to entail large Python overhead, but often that's acceptable.

  One option for these is to wrap them into something that broadcasts like a gufunc, e.g., via a new function `numpy.guvectorize` (https://github.com/numpy/numpy/pull/8054). But as a user, this is a lot of wrappers to write. You'd rather just add something like `vectorize=True` and let xarray handle all the automatic broadcasting, e.g.,
  ```python
  def poly_fit(x, y, dim='time', deg=1):
      return xr.apply(np.polyfit, x, y,
                      signature=([(dim,), (dim,)], [('poly_order',)]),
                      new_coords={'poly_order': range(deg + 1)},
                      kwargs={'deg': deg}, vectorize=True)
```

## Whether `func` applies to "data only" or "everything"

Most "computation" functions/methods in xarray (e.g., arithmetic and reduce methods) follow the rule of merging coordinates, and only applying the core function to data variables. Coordinates that are no longer valid with new dimensions are dropped. This is currently what we do in `apply`.

On the other hand, there are also function/methods that we might refer to as "organizing" (e.g., indexing methods, `concat`, `stack`/`unstack`, `transpose`), which generally apply to every variable, including coordinates. It seems like there are definitely use cases for applying these sorts of functions, too, e.g., to wrap Cartopy's `add_cyclic_point` utility (#1005). So, I think we might need another option to toggle what happens to coordinates (e.g., `variables='data'` vs `variables='all'`).

## How to handle mismatched core dimensions

Xarray methods often have fallbacks to handle data with different dimensions. For example, if you write `ds.mean(['x', 'y'])`, it matches on core dimensions to apply four different possible functions to each data variables:
- mean over `('x', 'y')`: for variables with both dimensions
- mean over `'x'`: for variables with only `'x'`
- mean over `'y'`: for variables with only `'y'`
- identity: for variables with neither `'x'` nor `'y'`

Indexing is another example -- it applies to both data and coordinates, but only to matching dimensions for each variable. If you don't have the dimensions, we ignore the variable.

Writing something like `mean` with a single call to `apply` would entail the need for something like a dispatching system to pick which function to use, e.g., instead of a singular `func`/`signature` pair you pass a `dispatcher` function that chooses `func`/`signature` based on the core dimensions of passed variable. This feels like serious over engineering.

Instead, we might support a few pre-canned options for how to deal with mismatched dimensions. For example:
- `missing_core_dims='drop'`: silently drop these variables in the output(s)
- `missing_core_dims='error'`: raise an error. This is the current default behavior, which is probably only be useful with `variables='data' -- otherwise some coordinate variables would always error.
- `missing_core_dims='keep'`: keep these variables unchanged in the output(s) (use `merge_variables` to check for conflicts)
- `missing_core_dims='broadcast'`: broadcast all inputs to have the necessary core dimensions if they don't have them already

Another option would be to consolidate this with the `variables` option to allow only two modes of operation:
- `variables='data'`: For "computation" functions. Apply only to data variables, and error if any data variables are missing a core dimension. Merge coordinates on the output(s), silently dropping conflicts for variables that don't label a dimension.
- `variables='matching'`: For "organizing" functions. Apply to every variable with matching core dimensions. Merge everything else on the outputs(s), silently dropping conflicts for variables that don't label a dimension.

**Comment by crusaderky** [comment]
Any hope to get dask support? Even with the limitation that 1 output block can be exactly produced by 1 input block, it would already be tremendously useful

**Comment by crusaderky** [comment]
I worked around the limitation. It would be nice if apply() did the below automatically!
```
from itertools import chain
from functools import wraps
import dask.array

def dask_kernel(func):
    """Invoke dask.array.map_blocks(func, *args, **kwds) if at least one
    of the arguments is a dask array; else invoke func(*args, **kwds)
    """
    @wraps(func)
    def wrapper(*args, **kwds):
        if any(isinstance(a, dask.array.Array) for a in chain(args, kwds.values())):
            return dask.array.map_blocks(func, *args, **kwds)
        else:
            return func(*args, **kwds)
    return wrapper
```

**Comment by shoyer** [comment]
I'm thinking about making a few tweaks and merging this, but not exposing it to users yet as part of public API. The public API is not quite there yet, but even as it I think it would be a useful building point for internal functionality (e.g., for #1065), and then other people could start to build on this as well.

**Comment by MaximilianR** [comment]
I'm thinking through how difficult it would be to add back-fill method to `DataArray` (that could be an argument to `fillna` or a `bfill` method - that's a separate discussion). 

Would this PR help? I'm trying to wrap my head around the options. Thanks

**Comment by shoyer** [comment]
> I'm thinking through how difficult it would be to add back-fill method to DataArray (that could be an argument to fillna or a bfill method - that's a separate discussion).
>
> Would this PR help? I'm trying to wrap my head around the options. Thanks

Yes, quite likely. In the current state, it would depend on if you want to back-fill all variables or just data variables (only the later is currently supported).

Either way, the first step is probably to write a function `backfill(values, axis)` that acts on NumPy arrays.

**Comment by MaximilianR** [comment]
> Either way, the first step is probably to write a function backfill(values, axis) that acts on NumPy arrays.

Right.
Surprisingly, I can't actually find something like this out there. The pandas code is good but highly 1-2 dimension specific.

Let me know if I'm missing (pun intended - long day) something. Is there a library of these sorts of functions over n-dims somewhere else (even R / Julia)? Or are we really the first people in the world to be doing this?

**Comment by shoyer** [comment]
> Surprisingly, I can't actually find something like this out there. The pandas code is good but highly 1-2 dimension specific.
>
> Let me know if I'm missing (pun intended - long day) something. Is there a library of these sorts of functions over n-dims somewhere else (even R / Julia)? Or are we really the first people in the world to be doing this?

Usually I check `numpy` and `bottleneck`. It actually looks like [`bottleneck.push`](http://berkeleyanalytics.com/bottleneck/reference.html#bottleneck.push) is what you're looking for. I think this is a recent addition to bottleneck, though.

**Comment by MaximilianR** [comment]
Gave this a quick spin for filling. A few questions:

- Is there an easy of way of merely translating dims into axes? Maybe that already exists?
- Is there as easy way to keep a dimension? Or should it be in the signature and a `new_dim`?

```python
da=xr.DataArray(np.random.rand(10,3), dims=('x','y'))
da = da.where(da>0.5)
In [43]: da
Out[43]:
<xarray.DataArray (x: 10, y: 3)>
array([[        nan,  0.57243305,  0.84363016],
       [        nan,  0.90788156,         nan],
       [        nan,  0.50739189,  0.93701278],
       [        nan,         nan,  0.86804167],
       [        nan,  0.50883914,         nan],
       [        nan,         nan,         nan],
       [        nan,  0.91547763,         nan],
       [ 0.72920182,         nan,  0.6982745 ],
       [ 0.73033449,  0.950719  ,  0.73077113],
       [        nan,         nan,  0.72463932]])

In [44]: xr.apply(bn.push, da) . # already better than `bn.push(da)`!
Out[44]:
<xarray.DataArray (x: 10, y: 3)>
array([[        nan,  0.57243305,  0.84363016],
       [        nan,  0.90788156,  0.90788156],
       [        nan,  0.50739189,  0.93701278],
       [        nan,         nan,  0.86804167],
       [        nan,  0.50883914,  0.50883914],
       [        nan,         nan,         nan],
       [        nan,  0.91547763,  0.91547763],
       [ 0.72920182,  0.72920182,  0.6982745 ],
       [ 0.73033449,  0.950719  ,  0.73077113],
       [        nan,         nan,  0.72463932]])

# but changing the axis is verbose and transposes the array - are there existing tools for this?

In [48]: xr.apply(bn.push, da, signature='(x)->(x)', new_coords=[dict(x=da.x)])
Out[48]:
<xarray.DataArray (y: 3, x: 10)>
array([[        nan,         nan,         nan,         nan,         nan,
                nan,         nan,  0.72920182,  0.73033449,  0.73033449],
       [ 0.57243305,  0.90788156,  0.50739189,  0.50739189,  0.50883914,
         0.50883914,  0.91547763,  0.91547763,  0.950719  ,  0.950719  ],
       [ 0.84363016,  0.84363016,  0.93701278,  0.86804167,  0.86804167,
         0.86804167,  0.86804167,  0.6982745 ,  0.73077113,  0.72463932]])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9
  o y        (y) -
```

- The triple nested signature is pretty tough to write! Two kwargs?

jhamman:
@shoyer - do we want to get this into 0.9 as a private api function and aim to complete it for the public api by 0.10 or so?

crusaderky:
@shoyer - any plans to add dask support as suggested above?

shoyer:
@crusaderky 

> any plans to add dask support as suggested above?

Yes, in fact I have a branch with some basic support for this that I was working on a few months ago. I haven't written tests yet but I can potentially push that WIP to another PR after merging this.

There are a couple of recent feature additions to `dask.array.atop` (https://github.com/dask/dask/pull/1612 and https://github.com/dask/dask/pull/1716) that should make this easier and more powerful.

@jhamman 

> do we want to get this into 0.9 as a private api function and aim to complete it for the public api by 0.10 or so?

Yes, this seems like a good goal. I'll take another look over this next week when I have the chance, to remove any work-in-progress bits that have snuck in and remove the public facing API.

shoyer:
OK, in it goes. Once again, there's no public API exposed yet.

MaximilianR:
Congrats!

MaximilianR:
FWIW the `bn.push` example still has some unanswered questions - would be interested to know if there's an easier way of doing that. Particularly if it's just a 'dim for axis' swap

shoyer:
#1245 replaces the unintuitive `signature` argument with separate `input_core_dims` and `output_core_dims`.

PR #968, opened 2016-08-14
Clarify "copy" argument for reindex

shoyer:
Also ensure that reindexing always returns a new xarray object.

xref #927

cc @crusaderky


crusaderky:
Typo in alignment.py:246
new_var = var.copy(deep=copy)
should be
new_var = var.copy(deep=True)

shoyer:
@crusaderky Nope, that's intentional. If `copy=True`, we want to do a deep copy on the variable to copy the data. If `copy=False`, we still want to return new `Variable` objects, but only shallow copies with the original data.

PR #971, opened 2016-08-17
Better error message when assigning coordinates without dimensions

### LINKED ISSUE #969: Better error message when assigning coordinates
If the user writes:
`data.coords['polygons'] = rasterized_image`

They get:
`ValueError: dimensions ('polygons',) must have the same length as the number of data dimensions, ndim=2`

This is correct, but it would be good to mention explicitly providing dimensions here.

shoyer:
Fixes #969

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -42,7 +42,7 @@ def as_variable(obj, name=None, copy=False):
         obj = obj.copy(deep=False)
     elif hasattr(obj, 'dims') and (hasattr(obj, 'data') or
                                    hasattr(obj, 'values')):
-        obj = Variable(obj.dims, getattr(obj, 'data', getattr(obj, 'values')),
+        obj = Variable(obj.dims, getattr(obj, 'data', obj.values),
                        getattr(obj, 'attrs', None),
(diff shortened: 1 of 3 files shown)


xarray/core/variable.py
```diff
@@ -42,7 +42,7 @@ def as_variable(obj, name=None, copy=False):
         obj = obj.copy(deep=False)
     elif hasattr(obj, 'dims') and (hasattr(obj, 'data') or
                                    hasattr(obj, 'values')):
-        obj = Variable(obj.dims, getattr(obj, 'data', obj.values),
+        obj = Variable(obj.dims, getattr(obj, 'data', getattr(obj, 'values')),
```
MaximilianR: Out of interest, why is this better?

shoyer: Actually I was confused when I switched this. It does literally the exact
same thing - I will revert. My concern was that this might prematurely
evaluate the data if it's a dask array.

MaximilianR: Yeah I think `getattr` is only useful if you don't know what `attr` it might be. But no harm!

PR #983, opened 2016-08-23
Update the Multimension coordinate groupby example

jhamman:
cc @rabernat @shoyer 

now using `ds = xr.tutorial.load_dataset('rasm')`


shoyer:
LGTM

PR #990, opened 2016-08-28
Added convenience method for saving DataArray to netCDF file

### LINKED ISSUE #915: Consider adding a convenience method for saving DataArrays to netCDF files
e.g.,
```python
class DataArray:
    ...
    def to_netcdf(self, *args, **kwargs):
        dataset = self.to_dataset(name='__data__')
        data.to_netcdf(*args, **kwargs)

def open_dataarray(*args, **kwargs):
    dataset = open_dataset(*args, **kwargs)
    return dataset['__data__']
```

robintw:
Added a simple function to DataArray that creates a dataset with one variable
called 'data' and then saves it to a netCDF file. All parameters are passed through
to to_netcdf().

Added an equivalent function called `open_dataarray` to be used to load from these files.

Fixes #915.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -328,8 +328,8 @@ def open_dataarray(*args, **kwargs):
 
     Notes
     -----
-    This is designed to be fully compatible with `DataArray.to_netcdf`.
-    Saving using `DataArray.to_netcdf` and then loading with this function will
+    This is designed to be fully compatible with `DataArray.to_netcdf`. Saving
+    using `DataArray.to_netcdf` and then loading with this function will
     produce an identical result.
 
     All parameters are passed directly to `xarray.open_dataset`. See that
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -267,7 +267,10 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     return maybe_decode_store(store)
 
 
-def open_dataarray(*args, **kwargs):
+def open_dataarray(filename_or_obj, group=None, decode_cf=True,
+                   mask_and_scale=True, decode_times=True,
+                   concat_characters=True, decode_coords=True, engine=None,
+                   chunks=None, lock=None, drop_variables=None):
     """
     Opens an DataArray from a netCDF file containing a single data variable.
 
@@ -339,7 +342,10 @@ def open_dataarray(*args, **kwargs):
     --------
     open_dataset
     """
-    dataset = open_dataset(*args, **kwargs)
+    dataset = open_dataset(filename_or_obj, group, decode_cf,
+                           mask_and_scale, decode_times,
+                           concat_characters, decode_coords, engine,
+                           chunks, lock, drop_variables)
 
     if len(dataset.data_vars) != 1:
         raise ValueError('Given file dataset contains more than one variable. '
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -267,6 +267,27 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     return maybe_decode_store(store)
 
 
+def open_dataarray(*args, **kwargs):
+    """
+    Opens an `xarray.DataArray` from a netCDF file, reading the variable called
+    `data`.
+
+    This is designed to read files saved with `xarray.DataArray.to_netcdf`.
+
+    Parameters
+    ----------
+    All parameters are passed directly to `xarray.open_dataset`.
+    """
+    dataset = open_dataset(*args, **kwargs)
+
+    try:
+        return dataset['data']
+    except KeyError:
+        raise ValueError('Given file dataset does not contain the variable `data`.'
+                         'If dataset was not saved using `xarray.DataArray.to_netcdf` then'
+                         'it must be loaded with `xarray.open_dataset`.')
+
+
 class _MultiFileCloser(object):
     def __init__(self, file_objs):
         self.file_objs = file_objs
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -15,6 +15,8 @@ from ..core.combine import auto_combine
 from ..core.utils import close_on_error, is_remote_uri
 from ..core.pycompat import basestring
 
+DATAARRAY_NAME = '__xarray_dataarray_name__'
+DATAARRAY_VARIABLE = '__xarray_dataarray_variable__'
 
 def _get_default_engine(path, allow_remote=False):
     if allow_remote and is_remote_uri(path):  # pragma: no cover
@@ -348,9 +350,9 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
                            chunks, lock, drop_variables)
 
     if len(dataset.data_vars) != 1:
-        raise ValueError('Given file dataset contains more than one variable. '
-                         'Please read with xarray.open_dataset and then select '
-                         'the variable you want.')
+        raise ValueError('Given file dataset contains more than one data '
+                         'variable. Please read with xarray.open_dataset and '
+                         'then select the variable you want.')
     else:
         data_array, = dataset.data_vars.values()
 
@@ -358,11 +360,11 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
 
     # Reset names if they were changed during saving
     # to ensure that we can 'roundtrip' perfectly
-    if '__xarray_dataarray_name__' in dataset.attrs:
-        data_array.name = dataset.attrs['__xarray_dataarray_name__']
-        del dataset.attrs['__xarray_dataarray_name__']
+    if DATAARRAY_NAME in dataset.attrs:
+        data_array.name = dataset.attrs[DATAARRAY_NAME]
+        del dataset.attrs[DATAARRAY_NAME]
 
-    if data_array.name == '__xarray_dataarray_variable__':
+    if data_array.name == DATAARRAY_VARIABLE:
         data_array.name = None
 
     return data_array
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -269,15 +269,75 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
 
 def open_dataarray(*args, **kwargs):
     """
-    Opens an `xarray.DataArray` from a netCDF file, and returns the single
-    variable that is within the file. If multiple variables are present then
-    a ValueError is raised.
+    Opens an DataArray from a netCDF file containing a single data variable.
 
-    This is designed to read files saved with `xarray.DataArray.to_netcdf`.
+    This is designed to read netCDF files with only one data variable. If
+    multiple variables are present then a ValueError is raised.
 
     Parameters
     ----------
-    All parameters are passed directly to `xarray.open_dataset`.
+    filename_or_obj : str, file or xarray.backends.*DataStore
+        Strings are interpreted as a path to a netCDF file or an OpenDAP URL
+        and opened with python-netCDF4, unless the filename ends with .gz, in
+        which case the file is gunzipped and opened with scipy.io.netcdf (only
+        netCDF3 supported). File-like objects are opened with scipy.io.netcdf
+        (only netCDF3 supported).
+    group : str, optional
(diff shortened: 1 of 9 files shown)


shoyer:
This is great, but it needs tests and documentation :).

shoyer:
I'm also a little concerned about the design here, because it always saves/restores the DataArray with the name `'data'`, which could override the original name. Ideally this method should save/restore a DataArray as faithfully as possible.

Instead, maybe we should only override the original name if it's `None`, or the same name as one of it's coordinates (which would otherwise result in an invalid NetCDF file).

Then we could restore by looking for a Dataset with single data variable regardless of the name, e.g.,
```
if len(dataset.data_vars) != 1:
    raise ValueError
data_array, = dataset.data_vars.values()
```

robintw:
Thanks @shoyer - I wasn't actually aware that DataArrays can have a name _before_ they're put into a Dataset (that's one of the things I love about contributing to OSS: you always learn things about the software that you never knew before!). I'll deal with the design issues ASAP.

How would you suggest testing this? I wasn't sure where to put the tests, and what exactly to test? Presumably something about the equivalence between this and saving after manually converting to a dataset, but I wasn't really sure how best to go about it.

shoyer:
I would add these tests to test_backends, probably a new test class. The
best test is probably a round-trip test where we save and load a DataArray
and verify that the result is identical to what we started with.

robintw:
I think I've dealt with the comments/issues now, but let me know if more needs doing.

xarray/backends/api.py
```diff
@@ -233,6 +233,30 @@ def maybe_decode_store(store, lock=False):
...
+    """
+    Opens an `xarray.DataArray` from a netCDF file, and returns the single
+    variable that is within the file. If multiple variables are present then
+    a ValueError is raised.
+
+    This is designed to read files saved with `xarray.DataArray.to_netcdf`.
+
+    Parameters
+    ----------
+    All parameters are passed directly to `xarray.open_dataset`.
```
shoyer: We should mention the parameters again (at least `path`). The fact that this uses `open_dataset` internally is an implementation detail.

xarray/backends/api.py
```diff
@@ -233,6 +233,30 @@ def maybe_decode_store(store, lock=False):
     return maybe_decode_store(store)
 
 
+def open_dataarray(*args, **kwargs):
+    """
+    Opens an `xarray.DataArray` from a netCDF file, and returns the single
```
shoyer: Keep the first line of docstrings to a single line.

xarray/test/test_backends.py
```diff
@@ -1063,3 +1064,29 @@ def test_extract_h5nc_encoding(self):
...
+    def test_dataarray_to_netcdf_with_name(self):
+        original_da = DataArray(np.arange(12).reshape((3, 4)),
+                                name='test')
+
+        with create_tmp_file() as tmp:
+            original_da.to_netcdf(tmp)
+
+            loaded_da = open_dataarray(tmp)
+
+        self.assertDataArrayIdentical(original_da, loaded_da)
```
shoyer: I think identical also checks name already? Worth double checking, though.

xarray/core/dataarray.py
```diff
@@ -1095,6 +1095,31 @@ def to_masked_array(self, copy=True):
...
+        Only xarray.Dataset objects can be written to netCDF files, so
+        the xarray.DataArray is converted to a xarray.Dataset object
+        containing a single variable. If the DataArray has no name, then
+        it is given the name 'data'.
+        """
+        if not self.name and self.name not in list(self.coords):
+            # If the name isn't blank/None or the same as one of the coords
+            # (as the latter is invalid for a netCDF file) then give it a name
+            # of `data`
+            dataset = self.to_dataset(name='data')
```
shoyer: Note that in principle "data" could also be the name of a coordinate!

I would be happier picking something long and arbitrary that a user is unlikely to use accidentally, e.g., `"__xarray_dataarray_variable__"`. Then, we could check for this name in `open_dataarray` and remove it from the returned DataArray if necessary.

xarray/backends/api.py
```diff
@@ -233,6 +233,30 @@ def maybe_decode_store(store, lock=False):
...
+    variable that is within the file. If multiple variables are present then
```
shoyer: single variable -> single data variable

xarray/backends/api.py
```diff
@@ -233,6 +233,30 @@ def maybe_decode_store(store, lock=False):
...
+    a ValueError is raised.
+
+    This is designed to read files saved with `xarray.DataArray.to_netcdf`.
+
+    Parameters
+    ----------
+    All parameters are passed directly to `xarray.open_dataset`.
+    """
+    dataset = open_dataset(*args, **kwargs)
```
shoyer: One downside of this approach is that it leaves a stranded file open without any possible way to close it.

We might want to add a `.close()` method to DataArray so we can clean this up. Or maybe just save that for a later PR.

xarray/core/dataarray.py
shoyer: To be extra thorough, we could save the original `name` as an attribute (e.g., `"__xarray_dataarray_name__"`) if it matches a coordinate, and then use that to set the name on the result. That could ensure that we always faithfully roundtrip the DataArray exactly.

robintw:
Thanks for the useful comments.

I've added some bits to the Sphinx docs in the places that I think they should go - but let me know if I need to add any more.

robintw:
Also, it seems like there might be an issue with not closing the Dataset - but only on Windows (see the AppVeyor build failure at https://ci.appveyor.com/project/jhamman/xarray-injyf/build/1.0.268/job/kr8ss684ijxg55be).

Any ideas how this could be sorted without having to implement `.close()` for `DataArray` (or, alternatively, how to go about implementing that, given that I don't think a DataArray contains a reference to the Dataset that it may have come from)

shoyer:
The easiest fix would be to add a decorator to skip tests on Windows, like the existing decorators that we use:
https://github.com/pydata/xarray/blob/master/xarray/test/__init__.py#L73

I'm not sure off hand how to detect if we're running on Windows but it can't be hard.

To add a close method, I would copy these lines from `dataset.py` (or better use, move them to the common base class `BaseDataObject` in `common.py`):
https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L251-L262

Then, you just need to define `_file_obj = None` in `DataArray.__init__` and set it when appropriate.

robintw:
Thanks - I've got that working with the `close()` method, and I did it 'properly' using the `BaseDataObject` class.

The only failure on here now is the same as #991 - a conda error creating one of the environments.

xarray/core/common.py
```diff
@@ -578,6 +578,19 @@ def where(self, cond, other=None, drop=False):
 
         return outobj._where(outcond)
 
+    def close(self):
+        """Close any files linked to this dataset
```
shoyer: dataset -> object

xarray/backends/api.py
```diff
@@ -233,6 +233,101 @@ def maybe_decode_store(store, lock=False):
...
+                         'Please read with xarray.open_dataset and then select '
+                         'the variable you want.')
+    else:
+        data_array, = dataset.data_vars.values()
+
+    data_array._file_obj = dataset._file_obj
+
+    # Reset names if they were changed during saving
+    # to ensure that we can 'roundtrip' perfectly
+    if '__xarray_dataarray_name__' in dataset.attrs:
```
shoyer: Set I would set these names as constants in this module and then import them, e.g.,
```
DATAARRAY_NAME = '__xarray_dataarray_name__'
DATAARRAY_VARIABLE = '__xarray_dataarray_variable__'
```

We have not been so strict about this, but it's usually considered bad form to write the same string literals multiple times in different places.

xarray/test/test_backends.py
```diff
@@ -1063,3 +1064,42 @@ def test_extract_h5nc_encoding(self):
...
+@requires_netCDF4
+class TestDataArrayToNetCDF(TestCase):
+
+    def test_dataarray_to_netcdf_no_name(self):
+        original_da = DataArray(np.arange(12).reshape((3, 4)))
+
+        with create_tmp_file() as tmp:
+            original_da.to_netcdf(tmp)
+
+            loaded_da = open_dataarray(tmp)
```
shoyer: Use the context manager to ensure things always get cleaned up (also less code), e.g.,
```
with open_dataaarray(tmp) as loaded_da:
    self.assertDataArrayIdentical(...)
```

xarray/test/test_backends.py
```diff
@@ -1063,3 +1064,42 @@ def test_extract_h5nc_encoding(self):
...
+            self.assertDataArrayIdentical(original_da, loaded_da)
+            loaded_da.close()
+
+
+    def test_dataarray_to_netcdf_coord_name_clash(self):
+        original_da = DataArray(np.arange(12).reshape((3, 4)),
+                                dims=['x', 'y'],
+                                name='r')
```
shoyer: I think you meant to pick a name that matches a coord, e.g., `name='x'`?

xarray/backends/api.py
```diff
@@ -233,6 +233,101 @@ def maybe_decode_store(store, lock=False):
     return maybe_decode_store(store)
 
 
+def open_dataarray(*args, **kwargs):
```
shoyer: as long as we're copying the docstring, let's copy the arguments for the function signature, too -- it helps auto-complete.

shoyer:
I have a few final nits, but looking very nice!

robintw:
Great, thanks. I think I've done these changes (frustratingly I can't see a way to use `*args`/`*kwargs` after using the full argument list in the function definition, so I've had to write it out. Let me know if there is a better way, but I suspect there isn't).

xarray/backends/api.py
```diff
@@ -233,6 +233,107 @@ def maybe_decode_store(store, lock=False):
...
+    --------
+    open_dataset
+    """
+    dataset = open_dataset(filename_or_obj, group, decode_cf,
+                           mask_and_scale, decode_times,
+                           concat_characters, decode_coords, engine,
+                           chunks, lock, drop_variables)
+
+    if len(dataset.data_vars) != 1:
+        raise ValueError('Given file dataset contains more than one variable. '
```
shoyer: variable -> data variable

shoyer:
I have a couple inline comments that I think you missed?

xarray/backends/api.py
shoyer: see here

xarray/core/common.py
shoyer: and here

robintw:
Oops yes, sorry!

I'm having a bit of an issue with changing the strings to constants (which, by the way, I completely agree is how it should be done). I've defined the constants in `backends/api.py` and am then trying to import them in `core/dataarray.py`.

I am doing the import as:

`from ..backends import DATAARRAY_NAME, DATAARRAY_VARIABLE`

But this gives me the following error:

```
../../anaconda3/lib/python3.5/site-packages/py/_path/local.py:650: in pyimport
    __import__(modname)
xarray/__init__.py:3: in <module>
    from .core.extensions import (register_dataarray_accessor,
xarray/core/extensions.py:3: in <module>
xarray/core/dataarray.py:9: in <module>
    from ..backends import DATAARRAY_NAME, DATAARRAY_VARIABLE
xarray/backends/__init__.py:8: in <module>
    from .netCDF4_ import NetCDF4DataStore
xarray/backends/netCDF4_.py:6: in <module>
    from .. import Variable
E   ImportError: cannot import name 'Variable'
```

It seems like my relative import is somehow mucking up another relative import inside the netCDF4 stuff. Do you have any ideas how I can solve this? Am I just doing my imports wrong?

shoyer:
We have a bit of a circular imports mess :). For now, just import them at the top of the `to_netcdf` method.

robintw:
Ah ok - that works (not entirely sure why, but it works!)

robintw:
I realised that the reason that Travis and AppVeyor weren't running properly was because there were conflicts between this PR and master. I _think_ I've managed to rebase and push properly - but let me know if I've done anything wrong.

shoyer:
Thanks!

PR #991, opened 2016-08-28
Added validation of attrs before saving to netCDF files

### LINKED ISSUE #911: KeyError on saving to NetCDF - due to objects in attrs?
I have an xarray.Dataset that I'm trying to save out to a NetCDF file. The dataset looks like this:

```python

Out[97]:
<xarray.Dataset>
Dimensions:  (x: 1240, y: 1162)
Coordinates:
  * x        (x) float64 -9.476e+05 -9.464e+05 -9.451e+05 -9.439e+05 ...
  * y        (y) float64 1.429e+06 1.428e+06 1.427e+06 1.426e+06 1.424e+06 ...
Data variables:
    data     (y, x) float32 nan nan nan nan nan nan nan nan nan nan nan nan ...
```

It has two attributes, both of which have a string key, and a value which is an object (in this case, instances of classes from the `rasterio` library).

```python
OrderedDict([('affine',
              Affine(1256.5430440955893, 0.0, -947639.6305106478,
       0.0, -1256.5430440955893, 1429277.8120091767)),
             ('crs', CRS({'init': 'epsg:27700'}))])
```

When I try to save out the NetCDF using this code:

```python
ds.to_netcdf('test.nc')
```

I get the following error:

```
<ipython-input-96-51e5ff396887> in <module>()
----> 1 ds.to_netcdf('blah3.nc')

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/core/dataset.py in to_netcdf(self, path, mode, format, group, engine, encoding)
    789         from ..backends.api import to_netcdf
    790         return to_netcdf(self, path, mode, format=format, group=group,
--> 791                          engine=engine, encoding=encoding)
    792 
    793     dump = utils.function_alias(to_netcdf, 'dump')

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/backends/api.py in to_netcdf(dataset, path, mode, format, group, engine, writer, encoding)
    354     store = store_cls(path, mode, format, group, writer)
    355     try:
--> 356         dataset.dump_to_store(store, sync=sync, encoding=encoding)
    357         if isinstance(path, BytesIO):
    358             return path.getvalue()

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/core/dataset.py in dump_to_store(self, store, encoder, sync, encoding)
    735             variables, attrs = encoder(variables, attrs)
    736 
--> 737         store.store(variables, attrs, check_encoding)
    738         if sync:
    739             store.sync()

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/backends/common.py in store(self, variables, attributes, check_encoding_set)
    226         cf_variables, cf_attrs = cf_encoder(variables, attributes)
    227         AbstractWritableDataStore.store(self, cf_variables, cf_attrs,
--> 228                                         check_encoding_set)

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/backends/common.py in store(self, variables, attributes, check_encoding_set)
    201                                 if not (k in neccesary_dims and
    202                                         is_trivial_index(v)))
--> 203         self.set_variables(variables, check_encoding_set)
    204 
    205     def set_attributes(self, attributes):

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/backends/common.py in set_variables(self, variables, check_encoding_set)
    211             name = _encode_variable_name(vn)
    212             check = vn in check_encoding_set
--> 213             target, source = self.prepare_variable(name, v, check)
    214             self.writer.add(source, target)
    215 

/Users/robin/anaconda3/lib/python3.5/site-packages/xarray/backends/netCDF4_.py in prepare_variable(self, name, variable, check_encoding)
    277             # set attributes one-by-one since netCDF4<1.0.10 can't handle
    278             # OrderedDict as the input to setncatts
--> 279             nc4_var.setncattr(k, v)
    280         return nc4_var, variable.data
    281 

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Variable.setncattr (netCDF4/_netCDF4.c:33460)()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4._set_att (netCDF4/_netCDF4.c:6171)()

/Users/robin/anaconda3/lib/python3.5/collections/__init__.py in __getitem__(self, key)
    967         if hasattr(self.__class__, "__missing__"):
    968             return self.__class__.__missing__(self, key)
--> 969         raise KeyError(key)
    970     def __setitem__(self, key, item): self.data[key] = item
    971     def __delitem__(self, key): del self.data[key]

KeyError: 0
```

The error seems slightly strange to me, but it seems to be related to saving attributes. If I change the attributes to make all of the values strings (for example, using `ds['data'].attrs = {k: repr(v) for k, v in ds['data'].attrs.items()}`) then it saves out fine.

Is there a restriction on what sort of values can be stored in `attrs` and saved out to NetCDF? If so, should this be enforced somehow? It would be ideal if any object could be stored as an attr and saved out (eg. as a pickle) - but this may be difficult (for example, for multiple python versions, if using pickle).

Any thoughts?

robintw:
This allows us to give nice errors if users try to save a Dataset with
attr values that can't be written to a netCDF file.

Fixes #911.

I've added tests to `test_backends.py` as I can't see a better place to put them. I've also made the tests fairly extensive, but also used some helper functions to stop too much repetition. Please let me know if any of this doesn't fit within the xarray style.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -6,7 +6,6 @@ from distutils.version import StrictVersion
 from glob import glob
 from io import BytesIO
 from numbers import Number
-from collections.abc import Iterable
 
 import numpy as np
 
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -94,30 +94,24 @@ def _validate_attrs(dataset):
                                  '1 or greater for serialization to netCDF '
                                  'files')
         else:
-            raise TypeError('Invalid name for attr: must be a string for '
-                            'serialization to netCDF files')
+            raise TypeError("Invalid name for attr: {} must be a string for "
+                            "serialization to netCDF files".format(name))
 
-        if not isinstance(value, basestring) and not isinstance(value, Number) and \
-           not isinstance(value, np.ndarray) and not isinstance(value, list) and \
-           not isinstance(value, tuple):
-                raise TypeError('Invalid value for attr: must be a number '
-                                'string, ndarray or a list/tuple of numbers/strings '
-                                'for serialization to netCDF files')
+        if not isinstance(value, (basestring, Number, np.ndarray, np.number,
+                                  list, tuple)):
+            raise TypeError('Invalid value for attr: {} must be a number '
+                            'string, ndarray or a list/tuple of numbers/strings '
+                            'for serialization to netCDF '
+                            'files'.format(value))
 
     # Check attrs on the dataset itself
     for k, v in dataset.attrs.items():
         check_attr(k, v)
 
     # Check attrs on each variable within the dataset
-    for var_name in dataset.data_vars:
+    for var_name in dataset.variables:
         for k, v in dataset[var_name].attrs.items():
             check_attr(k, v)
-
-    # Check attrs on each coord within the dataset
-    for coord_name in dataset.coords:
-        for k, v in dataset.coords[coord_name].attrs.items():
-            check_attr(k, v)
-
 
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=True, decode_times=True,
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -108,9 +108,10 @@ def _validate_attrs(dataset):
         check_attr(k, v)
 
     # Check attrs on each variable within the dataset
-    for var_name in dataset.variables:
-        for k, v in dataset[var_name].attrs.items():
+    for variable in dataset.variables.values():
+        for k, v in variable.attrs.items():
             check_attr(k, v)
+
 
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=True, decode_times=True,
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -5,6 +5,10 @@ import threading
 from distutils.version import StrictVersion
 from glob import glob
 from io import BytesIO
+from numbers import Number
+from collections.abc import Iterable
+
+import numpy as np
 
 from .. import backends, conventions
 from .common import ArrayWriter
@@ -77,6 +81,42 @@ def _validate_dataset_names(dataset):
 
     for k in dataset:
         check_name(k)
+
+
+def _validate_attrs(dataset):
+    """`attrs` must have a string key and a value which is either: a number
+    a string, an ndarray or a list/tuple of numbers/strings.
+    """
+    def check_attr(name, value):
+        if isinstance(name, basestring):
+            if not name:
+                raise ValueError('Invalid name for attr: string must be length '
+                                 '1 or greater for serialization to netCDF '
+                                 'files')
+        else:
+            raise TypeError('Invalid name for attr: must be a string for '
+                            'serialization to netCDF files')
+
+        if not isinstance(value, basestring) and not isinstance(value, Number) and \
+           not isinstance(value, np.ndarray) and not isinstance(value, list) and \
+           not isinstance(value, tuple):
+                raise TypeError('Invalid value for attr: must be a number '
+                                'string, ndarray or a list/tuple of numbers/strings '
+                                'for serialization to netCDF files')
+
+    # Check attrs on the dataset itself
+    for k, v in dataset.attrs.items():
+        check_attr(k, v)
+
+    # Check attrs on each variable within the dataset
+    for var_name in dataset.data_vars:
+        for k, v in dataset[var_name].attrs.items():
+            check_attr(k, v)
+
+    # Check attrs on each coord within the dataset
+    for coord_name in dataset.coords:
+        for k, v in dataset.coords[coord_name].attrs.items():
+            check_attr(k, v)
 
 
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
@@ -335,8 +375,9 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
     elif engine is None:
         engine = _get_default_engine(path)
 
-    # validate Dataset keys and DataArray names
+    # validate Dataset keys, DataArray names, and attr keys/values
     _validate_dataset_names(dataset)
+    _validate_attrs(dataset)
 
     try:
         store_cls = WRITEABLE_STORES[engine]
--- a/xarray/test/test_backends.py
+++ b/xarray/test/test_backends.py
@@ -1063,3 +1063,131 @@ class TestEncodingInvalid(TestCase):
                           {'least_sigificant_digit': 2})
         with self.assertRaisesRegexp(ValueError, 'unexpected encoding'):
             _extract_nc4_encoding(var, raise_on_invalid=True)
+
+
+class MiscObject:
+    pass
+
+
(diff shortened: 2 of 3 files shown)


xarray/backends/api.py
```diff
@@ -79,6 +83,42 @@ def check_name(name):
...
+                raise TypeError('Invalid value for attr: must be a number '
+                                'string, ndarray or a list/tuple of numbers/strings '
+                                'for serialization to netCDF files')
+
+    # Check attrs on the dataset itself
+    for k, v in dataset.attrs.items():
+        check_attr(k, v)
+
+    # Check attrs on each variable within the dataset
+    for var_name in dataset.data_vars:
```
shoyer: Use `dataset.variables`, which is a dict of [`xarray.Variable` objects](http://xarray.pydata.org/en/stable/internals.html#variable-objects)

xarray/backends/api.py
```diff
@@ -79,6 +83,42 @@ def check_name(name):
...
+        if isinstance(name, basestring):
+            if not name:
+                raise ValueError('Invalid name for attr: string must be length '
+                                 '1 or greater for serialization to netCDF '
+                                 'files')
+        else:
+            raise TypeError('Invalid name for attr: must be a string for '
+                            'serialization to netCDF files')
+
+        if not isinstance(value, basestring) and not isinstance(value, Number) and \
```
shoyer: Prefer parenthesis `()` to explicit backslashes `\`.

You can group isinstance checks, e.g., `isinstance(value, (basestring, Number, ..))`

Finally: `isinstance(numpy_value, Number)` only works on [NumPy 1.9 or later](https://github.com/numpy/numpy/issues/2951#issuecomment-38599552). Currently we still support NumPy 1.7 or newer. So you should check against `np.number` as well.

xarray/backends/api.py
```diff
@@ -79,6 +83,42 @@ def check_name(name):
...
+def _validate_attrs(dataset):
+    """`attrs` must have a string key and a value which is either: a number
+    a string, an ndarray or a list/tuple of numbers/strings.
+    """
+    def check_attr(name, value):
+        if isinstance(name, basestring):
+            if not name:
+                raise ValueError('Invalid name for attr: string must be length '
+                                 '1 or greater for serialization to netCDF '
+                                 'files')
```
shoyer: Prefer to print invalid values in error messages (it helps for debugging), e.g., 
```
raise ValueError('Invalid value: {}'.format(value))
```

xarray/test/test_backends.py
```diff
@@ -1063,3 +1063,131 @@ def test_extract_h5nc_encoding(self):
...
+
+class MiscObject:
+    pass
+
+
+class TestValidateAttrs(TestCase):
+
+    def test_invalid_attr_names(self):
+        ds = Dataset({'data': ('y', np.arange(10.0))})
+        self.assert_invalid_attr_names_for_obj(ds.attrs, ds)
```
shoyer: I would suggest moving this helper function inside this method -- that makes it easier to follow the logic of these tests.

xarray/test/test_backends.py
```diff
@@ -1063,3 +1063,131 @@ def test_extract_h5nc_encoding(self):
...
+
+        ds = Dataset({'data': ('y', np.arange(10.0))})
```
shoyer: Can we avoid repeating this line 9 times?

xarray/test/test_backends.py
```diff
@@ -1063,3 +1063,131 @@ def test_extract_h5nc_encoding(self):
...
+        self.assert_valid_attr_values_for_obj(ds.coords['y'].attrs, ds)
+
+    def assert_invalid_attr_names_for_obj(self, obj, ds):
+        # Helper function for `test_invalid_attr_names`
+        # In this function `obj` is an 'attrs' object (eg. ds.attrs or
+        # data.var.attrs, and `ds` is the dataset object
+
+        obj.clear()
+        obj[123] = 'test'
```
shoyer: I would suggest a slightly different design. Rather than using mutation/`obj.clear()`, use functions to setup brand new objects each time, e.g.,
```
def new_dataset():
    return Dataset({'data': ('y', np.arange(10.0))})

def new_dataset_and_dataset_attrs():
    ds = new_dataset()
    return ds, ds.attrs

def new_dataset_and_data_attrs():
    ds = new_dataset()
    return ds, ds.data.attrs
...

for new_dataset_and_attrs in [new_dataset_and_dataset_attrs, ....]:
    ds, attrs = new_dataset_and_attrs()
    attrs[123] = 'test'
    with self.assertRaisesRegexp(...):
        ...
```
This is a little safer, because using new objects means you can't forget to clean up old ones.

robintw:
Thanks, I'll deal with these modifications.

Also, I'm seeing errors from TravisCI saying `ValueError: cannot read or write netCDF files without netCDF4-python or scipy installed` - what do I need to do to make sure these tests work on Travis? (They pass fine on my local machine)

shoyer:
Take a look at the other netcdf tests -- they all use a decorator like
@requires_netCDF4, either on the class or the method. That ensures they are
skipped if required libraries for the functionality are not installed.

robintw:
I think I've dealt with all of these.

Refactoring the tests has made them a bit cleaner and safer due to not using mutation - however I still have quite a few repeated lines of code. I've also combined them all into one test with lots of asserts - as otherwise I couldn't keep the helper functions within the test without repeating them loads of times.

Does anything need adding to the docs for this? I'm assuming not as it's just making the code do more checking, it's not adding a feature as such?

robintw:
The Travis builds seem to have all passed except for `py27-cdat+pynio` which has failed with a conda error that seems to be unrelated to these changes (`Error: HTTPError: 403 Client Error: Forbidden for url`)

xarray/backends/api.py
```diff
@@ -79,6 +82,36 @@ def check_name(name):
...
+                            'string, ndarray or a list/tuple of numbers/strings '
+                            'for serialization to netCDF '
+                            'files'.format(value))
+
+    # Check attrs on the dataset itself
+    for k, v in dataset.attrs.items():
+        check_attr(k, v)
+
+    # Check attrs on each variable within the dataset
+    for var_name in dataset.variables:
```
shoyer: `dataset.variables` is a dictionary. So you can write:
```
for variable in dataset.variables.values():
    for k, v in variable.attrs.items():
        check_attr(k, v)
```

xarray/test/test_backends.py
```diff
@@ -1063,3 +1063,110 @@ def test_extract_h5nc_encoding(self):
...
+                ds.to_netcdf('test.nc')
+
+            # This one should work
+            ds, attrs = new_dataset_and_attrs()
+            attrs['test'] = 'test'
+            with create_tmp_file() as tmp_file:
+                ds.to_netcdf(tmp_file)
+
+
+
```
shoyer: lots of extra whitespace here

shoyer:
This is almost there now.

I agree that we don't need to update the docs, but please do make note of it in "What's New".

shoyer:
thanks!

PR #993, opened 2016-08-30
Coordinate -> IndexVariable and other deprecations

### LINKED ISSUE #727: DataArray creation prone to errors when data shares dimension shapes
It seems like there would be unexpected behavior whenever someone creates a DataArray using data with some subset of dimensions that have the same shape, and when supplying coords as a dictionary of `coord_name: coords` pairs. Since the dictionary isn't ordered, won't it be unclear which dimension it is referring to?

Maybe that's not such a good description, here's an example:
```python
data = np.random.randn(20, 5, 5)
da = xarray.DataArray(data, {'1': np.arange(20),
                             '2': np.arange(15, 20),
                             '3': np.arange(5, 10)})
print(da.coords)

da = xarray.DataArray(data, {'10': np.arange(20),
                             '11': np.arange(15, 20),
                             '12': np.arange(5, 10)})
print(da.coords)
```

For the first dimension it's no problem, since it's the only one with length 20. For the 2nd and 3rd dimensions since they are the same shape, the coordinates will be assigned depending on which order python takes the keys, which means they could sometimes be flipped depending on something arbitrary like what you named a dimension.

It seems like this could be addressed by doing a quick check (when supplying coordinates in this fashion) to see if any of the dimensions of the coordinates are the same, and throwing a warning that behavior might be unstable. Or maybe it throws an error and says "if data coordinates are same length, please supply coordinate values in a tuple or list instead"?

shoyer:
- Renamed the ``Coordinate`` class from xarray's low level API to
  ``IndexVariable``. xref https://github.com/pydata/xarray/pull/947#issuecomment-238549129
- Deprecated supplying ``coords`` as a dictionary to the ``DataArray``
  constructor without also supplying an explicit ``dims`` argument. The old
  behavior encouraged relying on the iteration order of dictionaries, which is
  a bad practice (fixes #727).
- Removed a number of methods deprecated since v0.7.0 or earlier:
  ``load_data``, ``vars``, ``drop_vars``, ``dump``, ``dumps`` and the
  ``variables`` keyword argument alias to ``Dataset``.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -21,6 +21,19 @@ v0.9.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+Deprecations
+~~~~~~~~~~~~
+
+- Renamed the ``Coordinate`` class from xarray's low level API to
+  :py:class:`~xarray.IndexVariable``.
+- Deprecated supplying ``coords`` as a dictionary to the ``DataArray``
+  constructor without also supplying an explicit ``dims`` argument. The old
+  behavior encouraged relying on the iteration order of dictionaries, which is
+  a bad practice (:issue:`727`).
+- Removed a number of methods deprecated since v0.7.0 or earlier:
+  ``load_data``, ``vars``, ``drop_vars``, ``dump``, ``dumps`` and the
+  ``variables`` keyword argument alias to ``Dataset``.
+
 Enhancements
 ~~~~~~~~~~~~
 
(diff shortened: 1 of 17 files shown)


shoyer:
> Maybe we should also rename Variable.to_coord() and IndexVariable.to_coord()?

Good catch, I meant to include this.

shoyer:
@benbovy any preference on whether I should submit this before or after your PR?

benbovy:
@shoyer feel free to merge this when you want. I can rebase my branch and resolve conflicts if needed.

shoyer:
OK, in it goes!

Issue #835, opened 2016-04-20
Merging variables with overlapping not but not conflicting values


shoyer:
It should be possible to safely merge together variables with values `[NaN, 1, 2]` and `[0, 1, NaN]` by using methods such as `combine_first` (which should also be OK with conflicting values, like the pandas method) and `merge` (which should raise if values conflict).

See this stackoverflow post for a merge example: http://stackoverflow.com/questions/36731870/how-to-merge-xarray-datasets-with-conflicting-coordinates

shoyer:
I'm currently in the process of refactoring the merge code so this should be pretty easy to add in.

shoyer:
Actually, we can just write:
```
def combine_first(left, right):
    return ops.where(ops.notnull(left), left, right)
```

shoyer:
Fixed by #1204

Issue #742, opened 2016-02-02
merge and align DataArrays/Datasets on different domains


jcmgray:
Firstly, I think `xarray` is great and for the type of physics simulations I run n-dimensional labelled arrays is _exactly_ what I need. But, and I may be missing something, is there a way to merge (or concatenate/update) DataArrays with _different_ domains on the same coordinates? 

For example consider this setup:

``` python
import xarray as xr

x1 = [100]
y1 = [1, 2, 3, 4, 5]
dat1 = [[101, 102, 103, 104, 105]]

x2 = [200]
y2 = [3, 4, 5, 6]  # different size and domain
dat2 = [[203, 204, 205, 206]]

da1 = xr.DataArray(dat1, dims=['x', 'y'], coords={'x': x1, 'y': y1})
da2 = xr.DataArray(dat2, dims=['x', 'y'], coords={'x': x2, 'y': y2})
```

I would like to aggregate such DataArrays into a new, single DataArray with `nan` padding such that:

``` python
>>> merge(da1, da2, align=True)  # made up syntax
<xarray.DataArray (x: 2, y: 6)>
array([[ 101.,  102.,  103.,  104.,  105.,   nan],
       [  nan,   nan,  203.,  204.,  205.,  206.]])
Coordinates:
  * x        (x) int64 100 200
  * y        (y) int64 1 2 3 4 5 6
```

Here is a quick function I wrote to do such but I would worried about the performance of 'expanding' the new data to the old data's size every iteration (i.e. supposing that the first argument is a large DataArray that you are adding to but doesn't necessarily contain the dimensions already).

``` python
def xrmerge(*das, accept_new=True):
    da = das[0]
    for new_da in das[1:]:
        # Expand both to have same dimensions, padding with NaN
        da, new_da = xr.align(da, new_da, join='outer')
        # Fill NaNs one way or the other re. accept_new
        da = new_da.fillna(da) if accept_new else da.fillna(new_da)
    return da
```

Might this be (or is this already!) possible in simpler form in `xarray`? I know _Datasets_ have `merge` and `update` methods but I couldn't make them work as above.
I also notice there are possible plans ( #417 ) to introduce a `merge` function for DataArrays.

shoyer:
This is actually closer to the functionality of (`concat`)[http://xarray.pydata.org/en/stable/combining.html#concatenate] than merge. Hypothetically, something like the following would do what you want:
```
# note: this is *not* valid syntax currently! the dims arguments
[...]

PR #996, opened 2016-08-31
add 'no_conflicts' as compat option for merging non-conflicting data

jcmgray:
This solves #742 and partially #835 (does not address a combine_first option yet). It essentially adds a notnull_equals method to Variable (and currently DataArray/set) that merge can use to compare objects, subsequently combining them with 'fillna'. Used as such:

```python
>>> import xarray as xr
>>> ds1 = xr.Dataset(data_vars={'x': ('a', [1, 2])})
>>> ds2 = xr.Dataset(data_vars={'x': ('a', [2, 3])}, coords={'a': [1, 2]})
>>> xr.merge([ds1, ds2], compat='notnull_equals')
<xarray.Dataset>
Dimensions:  (a: 3)
Coordinates:
  * a        (a) int64 0 1 2
Data variables:
    x        (a) float64 1.0 2.0 3.0
```

Should be very easy to add other merging options such as overwriting left to write and vice versa. 

TODO
-----
- docs
- tests for various type combinations
- some of the error excepting is mirrored from `equals` and might not be necessary (e.g. for numpy structured arrays).

ISSUES
------
- It seemed natural to add notnull_equals as a method to Dataset, but it might be unnecessary/not useful. Current version is conservative in that it still requires aligned Datasets with all the same variables.
- Due to the float nature of NaN, type is sometimes not preserved, e.g. merging two int arrays yields a float array, even when the final array has no NaN values itself.
- float rounding errors can cause unintended failures, so a notnull_allclose option might be nice.

Currently all tests passing (6 new), and no flake8 errors on the diff...

### DIFF
--- a/doc/combining.rst
+++ b/doc/combining.rst
@@ -208,5 +208,5 @@ coordinates as long as any non-missing values agree or are disjoint:
     xr.merge([ds1, ds2], compat='no_conflicts')
 
 Note that due to the underlying representation of missing values as floating
-point numbers (``NaN``) or generic objects (``None``), variable data type is not
-always conserved when merging in this manner.
+point numbers (``NaN``), variable data type is not always preserved when merging
+in this manner.
--- a/doc/combining.rst
+++ b/doc/combining.rst
@@ -208,5 +208,5 @@ coordinates as long as any non-missing values agree or are disjoint:
     xr.merge([ds1, ds2], compat='no_conflicts')
 
 Note that due to the underlying representation of missing values as floating
-point numbers (NaN) or generic objects (None), variable data type is not
+point numbers (``NaN``) or generic objects (``None``), variable data type is not
 always conserved when merging in this manner.
--- a/doc/combining.rst
+++ b/doc/combining.rst
@@ -192,3 +192,21 @@ numpy):
 
 Note that ``NaN`` does not compare equal to ``NaN`` in element-wise comparison;
 you may need to deal with missing values explicitly.
+
+Merging with 'no_conflicts'
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``compat`` argument ``'no_conflicts'`` is only available when
+combining xarray objects with ``merge``. In addition to the above comparison
+methods it allows the merging of xarray objects with locations where *either*
+have ``NaN`` values. This can be used to combine data with overlapping
+coordinates as long as any non-missing values agree or are disjoint:
+
+.. ipython:: python
+    ds1 = xr.Dataset({'a': ('x', [10, 20, 30, np.nan])}, {'x': [1, 2, 3, 4]})
+    ds2 = xr.Dataset({'a': ('x', [np.nan, 30, 40, 50])}, {'x': [2, 3, 4, 5]})
+    xr.merge([ds1, ds2], compat='no_conflicts')
+
+Note that due to the underlying representation of missing values as floating
+point numbers (NaN) or generic objects (None), variable data type is not
+always conserved when merging in this manner.
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -79,8 +79,20 @@ def _fail_on_dask_array_input(values, msg=None, func_name=None):
 
 around = _dask_or_eager_func('around')
 isclose = _dask_or_eager_func('isclose')
-isnull = _dask_or_eager_func('isnull', pd)
 notnull = _dask_or_eager_func('notnull', pd)
+_isnull = _dask_or_eager_func('isnull', pd)
+
+
+def isnull(data):
+    # GH837, GH861
+    # isnull fcn from pandas will throw TypeError when run on numpy structured
+    # array therefore for dims that are np structured arrays we assume all
+    # data is present
+    try:
+        return _isnull(data)
+    except TypeError:
+        return np.zeros(data.shape, dtype=bool)
+
 
 transpose = _dask_or_eager_func('transpose')
 where = _dask_or_eager_func('where', n_array_args=3)
@@ -125,17 +137,7 @@ def array_equiv(arr1, arr2):
         return False
 
     flag_array = (arr1 == arr2)
-
-    # GH837, GH861
-    # isnull fcn from pandas will throw TypeError when run on numpy structured array
-    # therefore for dims that are np structured arrays we skip testing for nan
-
-    try:
-
-        flag_array |= (isnull(arr1) & isnull(arr2))
-
-    except TypeError:
-        pass
+    flag_array |= (isnull(arr1) & isnull(arr2))
 
     return bool(flag_array.all())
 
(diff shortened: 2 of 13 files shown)


xarray/test/test_dataarray.py
```diff
@@ -352,6 +352,43 @@ def test_broadcast_equals(self):
         self.assertFalse(a.broadcast_equals(c))
         self.assertFalse(c.broadcast_equals(a))
 
+    def test_notnull_equals(self):
```
shoyer: Rather than writing such exhaustive test cases for each xarray class, consider writing comprehensive unit tests just for `array_notnull_equiv`.

shoyer:
This is quite clever! I'm impressed this was so easy using `fillna`.

One potential concern here is that performance is not going to be so great if you attempt to combine a bunch of variables with lazy data loaded with dask, because each comparison will trigger a separate computation. To that end, it would be nice to do the safety check in a single dask operation.

To be fair, this is already an issue if you're trying to merge lots of datasets together. But I expect this will become more of an issue when there are *useful* ways to merge a bunch of datasets, which is what this PR enables.

xarray/core/ops.py
```diff
@@ -140,6 +140,30 @@ def array_equiv(arr1, arr2):
...
+    """Like np.array_equal, but also allows values to be NaN in either or both
+    arrays
+    """
+    arr1, arr2 = as_like_arrays(arr1, arr2)
+    if arr1.shape != arr2.shape:
+        return False
+
+    flag_array = (arr1 == arr2)
+
+    # GH837, GH861
```
shoyer: I'm not sure why we missed this before, but these comments (and ignoring TypeError) should probably go in the `isnull` function instead, rather than repeating it here.

xarray/core/ops.py
```diff
@@ -140,6 +140,30 @@ def array_equiv(arr1, arr2):
...
+    # isnull fcn from pandas will throw TypeError when run on numpy structured
+    # array therefore for dims that are np structured arrays we skip testing
+    # for nan
+
+    try:
+        flag_array |= (isnull(arr1) | isnull(arr2))
```
shoyer: use two inplace operations to speed things up with numpy:
```
flag_array |= isnull(arr1)
flag_array |= isnull(arr2)
```

shoyer:
> Current version is conservative in that it still requires aligned Datasets with all the same variables.

:+1: it's much less obvious to treat missing variables as "missing data."

> Due to the float nature of NaN, type is sometimes not preserved, e.g. merging two int arrays yields a float array, even when the final array has no NaN values itself.

Yes, this is consistent with the xarray/pandas type system. (Which is not ideal, but that's another bigger issue.)

shoyer:
`compat='no_conflicts'` is possibly a better keyword argument.

shoyer:
One of the commits in this PR is not yours. Take a look here -- you want to rebase instead of merge (yes git is confusing):
http://stackoverflow.com/questions/5968964/avoid-unwanted-merge-commits-and-other-commits-when-doing-pull-request-on-github

I think you can still safely rebase now.

xarray/core/ops.py
jcmgray: Yes it would be good not to repeat. However, if put in `isnull`, then `TypeError` might be suppressed when useful. Excepting the error I think can only work in situations (such as these two functions) where an array is being *modified* by the result and the whole operation can be ignored if fails. But maybe there is another way to do it?

The only quite ugly way I can think of is something like (leaving `isnull` as-is):
```python
def isnull_with_no_type_fail(a):
    # comments about why needed etc.
    try:
        b = isnull(a)
    except TypeError:
        b = None
    return b
```
and then in `array_equiv`/`array_notnull_equiv`

```python
...
arr1_isnull = isnull_with_no_type_fail(arr1)
if arr1_isnull is not None:
    arr2_isnull = isnull_with_no_type_fail(arr2)
    if arr2_isnull is not None:
        flag_array |= (arr1_isnull | arr2_isnull)
...
```

jcmgray:
Ah sorry - yes rebased locally then mistakenly merged the remote fork... 

> `compat='no_conflicts'` is possibly a better keyword argument.

Yes I thought that might be better also, the advantages of `'notnull_equals'`, (slightly more precise operational description and slightly better 'grammar' when using as a method: `if ds1.notnull_equals(ds2): ...` vs `if ds1.no_conflicts(ds2): ...`) are probably negligible since `'no_conflicts'` is more intuitive for `merge` - and this is likely the main usage.

jcmgray:
> One potential concern here is that performance is not going to be so great if you attempt to combine a bunch of variables with lazy data loaded with dask, because each comparison will trigger a separate computation. To that end, it would be nice to do the safety check in a single dask operation.

I will have a look into how to do this, but am not that familiar with dask.

xarray/test/test_dataarray.py
jcmgray: Yes I'll move some the redundant tests to do with different types of overlap etc. into `test_ops`.

xarray/core/ops.py
shoyer: How about this version?
```
def isnull(a):
    try:
        b = isnull(a)
    except TypeError:
        b = np.zeros(a.shape, bool)
    return b
```
The `TypeError` should be pretty rare, so it's OK to do a little extra work in that case.

shoyer:
I think this is nearly ready. We just need to decide on the final keyword name (e.g., `notnull_equals` or `no_conflicts`) and add mention of this in the documentation / what's new.

I'm not sure we really need this new method for Dataset/DataArray (for the current implementation, adding it on Variable would be enough). Given that the user can catch `MergeError`, it seems straightforward enough just call `merge` to see if merging is safe.

jcmgray:
OK, I have stripped the Dataset/Array methods which I agree were largely redundant. Since this sets this type of comparison/merge slightly apart, `'no_conflicts'` seems the more intuitive wording when used only as a `compat` option so I've changed it to that.

And I've done a first pass at updating the docs.

doc/combining.rst
```diff
@@ -192,3 +192,21 @@ numpy):
...
+have ``NaN`` values. This can be used to combine data with overlapping
+coordinates as long as any non-missing values agree or are disjoint:
+
+.. ipython:: python
+    ds1 = xr.Dataset({'a': ('x', [10, 20, 30, np.nan])}, {'x': [1, 2, 3, 4]})
+    ds2 = xr.Dataset({'a': ('x', [np.nan, 30, 40, 50])}, {'x': [2, 3, 4, 5]})
+    xr.merge([ds1, ds2], compat='no_conflicts')
+
+Note that due to the underlying representation of missing values as floating
+point numbers (``NaN``) or generic objects (``None``), variable data type is not
```
shoyer: I deleted

    or generic objects (``None``)

We only use `NaN` (like pandas)

PR #1007, opened 2016-09-19
Fixes for compat='no_conflicts' and open_mfdataset

shoyer:
- `xarray.merge` now uses `compat='no_conflicts'` by default.
- `compat='no_conflicts'` preserves attributes.
- `Variable.no_conflicts` broadcasts arguments, making it a strictly more
  relaxed check than `broadcast_equals`.
- Allow `concat_dim=None` in `open_mfdataset` and `auto_combine` to disable
  trying to infer a dimension for concatenating.
- Add `compat` as an argument to `open_mfdataset` to allow more control over
  merging.

CC @jcmgray

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -21,9 +21,9 @@ v0.9.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
-- The default behavior of ``merge`` is now ``compat='no_conflict'``, so some
+- The default behavior of ``merge`` is now ``compat='no_conflicts'``, so some
   merges will now succeed in cases that previously raised
-  ``xarray.MergeError``. Set ``compat='broadcast_arrays'`` to restore the
+  ``xarray.MergeError``. Set ``compat='broadcast_equals'`` to restore the
   previous default.
 
 Deprecations
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -21,6 +21,11 @@ v0.9.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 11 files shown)


doc/whats-new.rst
```diff
@@ -21,6 +21,11 @@ v0.9.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- The default behavior of ``merge`` is now ``compat='no_conflict'``, so some
+  merges will now succeed in cases that previously raised
+  ``xarray.MergeError``. Set ``compat='broadcast_arrays'`` to restore the
+  previous default.
+
```
jcmgray: Should be 'no_conflicts' and 'broadcast_equals' here I think.

jcmgray:
This all looks great to me. Might the docstring for `auto_combine` need to be updated? Currently it seems to suggest that sorting, alignment and the addition of new data variables are not supported, whereas I think all these can occur with `compat='no_conflicts'`.

doc/whats-new.rst
shoyer: oops, thanks

shoyer:
> Might the docstring for auto_combine need to be updated? Currently it seems to suggest that sorting, alignment and the addition of new data variables are not supported, whereas I think all these can occur with compat='no_conflicts'.

Yep, with `compat='no_conflicts'` and recent align updates to concat, we are most of the way there on alignment.

It still can fail if there are different variables on datasets across the concatenated dimension, but that will be fixed when we get concat to handle different variables (#508). At that point, we might want to switch the logic from the current "concatenate by variable group, then merge" to "merge by unique coordinates, then concatenate".

Issue #1009, opened 2016-09-19
Shouldn't .where() pass the attributes of DataArrays and DataSets? 


fmaussion:
Everything is in the title! I think it should, if possible.

shoyer:
Yes, I agree. `.fillna()` should also preserve attributes.

fmaussion:
With a `keep_attrs` keyword, like `.mean()`, or always?

Since we are at it: what was the rationale behind not keeping the attributes per default everywhere?

shoyer:
I think we can keep attributes always for this operations.

The line I tried to draw was:
- For operations that rearrange/manipulate existing data, keep attributes
- For operations that do computation to produce new data, drop attributes

fmaussion:
OK, that makes sense. It gives me some trouble with the geoloc accessor I am working on, but that's another story. 

I'll try to put a PR together soon.

PR #1011, opened 2016-09-20
.where() and .fillna() should preserve attributes

fmaussion:
Fixes https://github.com/pydata/xarray/issues/1009

There is probably a more elegant way to do it, but the internals of xarray still confuse me. I'm happy to learn!

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -555,9 +555,9 @@ class BaseDataObject(AttrAccessMixin):
         if other is not None:
             raise NotImplementedError("The optional argument 'other' has not yet been implemented")
 
+        from .dataset import Dataset
         if drop:
             from .dataarray import DataArray
-            from .dataset import Dataset
             # get cond with the minimal size needed for the Dataset
             if isinstance(cond, Dataset):
                 clipcond = cond.to_array().any('variable')
@@ -576,7 +576,13 @@ class BaseDataObject(AttrAccessMixin):
             outobj = self
             outcond = cond
 
-        return outobj._where(outcond)
+        # preserve attributes
+        out = outobj._where(outcond)
+        out.attrs = self.attrs
+        if isinstance(out, Dataset):
+            for v in out:
+                out[v].attrs = self[v].attrs
+        return out
 
     def close(self):
         """Close any files linked to this object
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -555,8 +555,8 @@ class BaseDataObject(AttrAccessMixin):
         if other is not None:
             raise NotImplementedError("The optional argument 'other' has not yet been implemented")
 
-        from .dataset import Dataset
         if drop:
+            from .dataset import Dataset
             from .dataarray import DataArray
             # get cond with the minimal size needed for the Dataset
             if isinstance(cond, Dataset):
@@ -578,10 +578,7 @@ class BaseDataObject(AttrAccessMixin):
 
         # preserve attributes
         out = outobj._where(outcond)
-        out.attrs = self.attrs
-        if isinstance(out, Dataset):
-            for v in out:
-                out[v].attrs = self[v].attrs
+        out._copy_attrs_from(self)
         return out
 
     def close(self):
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -556,8 +556,8 @@ class BaseDataObject(AttrAccessMixin):
             raise NotImplementedError("The optional argument 'other' has not yet been implemented")
 
         if drop:
-            from .dataset import Dataset
             from .dataarray import DataArray
+            from .dataset import Dataset
             # get cond with the minimal size needed for the Dataset
             if isinstance(cond, Dataset):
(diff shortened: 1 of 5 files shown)


xarray/core/common.py
```diff
@@ -576,7 +576,13 @@ def where(self, cond, other=None, drop=False):
             outobj = self
             outcond = cond
 
-        return outobj._where(outcond)
+        # preserve attributes
+        out = outobj._where(outcond)
+        out.attrs = self.attrs
+        if isinstance(out, Dataset):
```
shoyer: Can you implement a private method defined on both `Dataset`/`DataArray`, something like `_copy_attrs_from(self, other)`? Then we can call `out._copy_attrs_from(self)` for a generic solution.

xarray/core/common.py
```diff
@@ -576,7 +576,13 @@ def where(self, cond, other=None, drop=False):
...
+            for v in out:
+                out[v].attrs = self[v].attrs
```
shoyer: Use `out.variables[v].attrs = self.variables[v].attrs` instead. Creating `DataArray` objects is pretty expensive compared to just pulling out existing `Variable` objects.

xarray/core/dataarray.py
```diff
@@ -1001,7 +1001,9 @@ def fillna(self, value):
         if utils.is_dict_like(value):
             raise TypeError('cannot provide fill value as a dictionary with '
                             'fillna on a DataArray')
-        return self._fillna(value)
+        out = self._fillna(value)
```
shoyer: probably can put this in `_fillna` instead.

xarray/test/test_dataset.py
```diff
@@ -2263,6 +2263,30 @@ def test_fillna(self):
         actual = ds.groupby('b').fillna(Dataset({'a': ('b', [0, 2])}))
         self.assertDatasetIdentical(expected, actual)
 
+        # attrs with groupby
+        ds.attrs['attr'] = 'ds'
+        ds.a.attrs['attr'] = 'da'
+        actual = ds.groupby('b').fillna(Dataset({'a': ('b', [0, 2])}))
+        self.assertTrue('attr' in actual.attrs)
```
shoyer: use something like `self.assertEquals(actual.attrs, {'attr': 'ds'})` instead

xarray/test/test_dataset.py
```diff
@@ -2263,6 +2263,30 @@ def test_fillna(self):
...
+        self.assertTrue(actual.attrs['attr'] == 'ds')
+        self.assertTrue(actual.a.name == 'a')
```
shoyer: use `self.assertEquals` again

xarray/core/common.py
fmaussion: Yes, good idea. Should the Dataset version check for the compatibility of ``other``, or should we just assume that ``other`` will always have the same variables?

shoyer: I think it's fine to assume compatibility, given that this is only being called by internal methods. If `other` is not a `Dataset` then something has gone seriously wrong.

On the other hand, I think it might work to call `DataArray.where` with a Dataset as the argument -- worth double checking.

fmaussion: I just tried ``actual = ds.a.where(ds > 1)``and _where() returns a ``NotImplemented`` string

fmaussion:
I implemented your comments with the exception of:
> probably can put this in _fillna instead.

Do you mean this should/could be done in ``_binary_op`` instead?

xarray/core/common.py
shoyer: OK, I think we're say then, though we probably shouldn't be returning `NotImplemented` singletons to users :).

shoyer:
> > probably can put this in _fillna instead.
> Do you mean this should/could be done in _binary_op instead?

I forgot that we hack this into `_binary_op` right now. So no, please ignore this guidance then. This will cleaner after #964.

fmaussion:
OK. Sorry for the mess with Travis, I had a lot of typos :( It's getting late here and the new review system is a bit confusing

xarray/test/test_dataset.py
```diff
@@ -2263,6 +2263,25 @@ def test_fillna(self):
         actual = ds.groupby('b').fillna(Dataset({'a': ('b', [0, 2])}))
         self.assertDatasetIdentical(expected, actual)
 
+        # attrs with groupby
+        ds.attrs['attr'] = 'ds'
+        ds.a.attrs['attr'] = 'da'
+        actual = ds.groupby('b').fillna(Dataset({'a': ('b', [0, 2])}))
+        self.assertEqual(actual.attrs, ds.attrs)
+        self.assertTrue(actual.a.name == 'a')
```
shoyer: still would be a good idea to use `assertEqual` here :)

fmaussion: this is what happens when one does too many things at the same time without thinking :-(

Should be ok now, hopefully...

shoyer:
Thanks!

PR #1018, opened 2016-09-25
Disabled auto-caching on dask; new .compute() method

crusaderky:
Disabled auto-caching dask arrays when pickling and when invoking the .values property.
Added new method .compute() to Variable, DataArray and Dataset.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -26,8 +26,8 @@ Breaking changes
   ``xarray.MergeError``. Set ``compat='broadcast_equals'`` to restore the
   previous default.
 - Pickling an xarray object based on the dask backend, or reading its
-  :py:meth:`values` property, won't automatically convert the backend to numpy
-  in the original object anymore.
+  :py:meth:`values` property, won't automatically convert the array from dask
+  to numpy in the original object anymore.
   By `Guido Imperiale <https://github.com/crusaderky>`_.
 
 Deprecations
@@ -72,8 +72,9 @@ Enhancements
   :py:func:`~xarray.open_mfdataset` to disable inferring a dimension along
   which to concatenate.
   By `Stephan Hoyer <https://github.com/shoyer>`_.
-- Added :py:meth:`compute` method to :py:class:`DataArray`, :py:class:`Dataset`,
-  and :py:class:`Variable` as a non-destructive alternative to :py:meth:`load`.
+- Added methods :py:meth:`DataArray.compute`, :py:meth:`Dataset.compute`, and
+  :py:meth:`Variable.compute` as a non-mutating alternative to
+  :py:meth:`~DataArray.load`.
   By `Guido Imperiale <https://github.com/crusaderky>`_.
 
 Bug fixes
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -25,6 +25,10 @@ Breaking changes
   merges will now succeed in cases that previously raised
   ``xarray.MergeError``. Set ``compat='broadcast_equals'`` to restore the
   previous default.
+- Pickling an xarray object based on the dask backend, or reading its
+  :py:meth:`values` property, won't automatically convert the backend to numpy
+  in the original object anymore.
+  By `Guido Imperiale <https://github.com/crusaderky>`_.
 
 Deprecations
 ~~~~~~~~~~~~
@@ -45,33 +49,32 @@ Deprecations
 Enhancements
 ~~~~~~~~~~~~
 - Add checking of ``attr`` names and values when saving to netCDF, raising useful
-error messages if they are invalid. (:issue:`911`).
-By `Robin Wilson <https://github.com/robintw>`_.
-
+  error messages if they are invalid. (:issue:`911`).
+  By `Robin Wilson <https://github.com/robintw>`_.
 - Added ability to save ``DataArray`` objects directly to netCDF files using
   :py:meth:`~xarray.DataArray.to_netcdf`, and to load directly from netCDF files
   using :py:func:`~xarray.open_dataarray` (:issue:`915`). These remove the need
   to convert a ``DataArray`` to a ``Dataset`` before saving as a netCDF file,
   and deals with names to ensure a perfect 'roundtrip' capability.
   By `Robin Wilson <https://github.com/robintw`_.
-
 - Multi-index levels are now accessible as "virtual" coordinate variables,
   e.g., ``ds['time']`` can pull out the ``'time'`` level of a multi-index
   (see :ref:`coordinates`). ``sel`` also accepts providing multi-index levels
   as keyword arguments, e.g., ``ds.sel(time='2000-01')``
   (see :ref:`multi-level indexing`).
   By `Benoit Bovy <https://github.com/benbovy>`_.
-
 - Added the ``compat`` option ``'no_conflicts'`` to ``merge``, allowing the
   combination of xarray objects with disjoint (:issue:`742`) or
   overlapping (:issue:`835`) coordinates as long as all present data agrees.
   By `Johnnie Gray <https://github.com/jcmgray>`_. See
   :ref:`combining.no_conflicts` for more details.
-
 - It is now possible to set ``concat_dim=None`` explicitly in
   :py:func:`~xarray.open_mfdataset` to disable inferring a dimension along
   which to concatenate.
   By `Stephan Hoyer <https://github.com/shoyer>`_.
+- Added :py:meth:`compute` method to :py:class:`DataArray`, :py:class:`Dataset`,
+  and :py:class:`Variable` as a non-destructive alternative to :py:meth:`load`.
+  By `Guido Imperiale <https://github.com/crusaderky>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -309,7 +309,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         """
         # access .data to coerce everything to numpy or dask arrays
         all_data = dict((k, v.data) for k, v in self.variables.items())
-        lazy_data = OrderedDict((k, v) for k, v in all_data.items()
+        lazy_data = dict((k, v) for k, v in all_data.items()
                          if isinstance(v, dask_array_type))
         if lazy_data:
             import dask.array as da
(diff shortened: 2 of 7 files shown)


xarray/core/dataset.py
```diff
@@ -319,6 +322,40 @@ def load(self):
...
+        left unaltered.
+
+        Normally, it should not be necessary to call this method in user code,
+        because all xarray functions should either work on deferred data or
+        load data automatically. However, this method can be necessary when
+        working with many file objects on disk.
+        """
+        # Can't just do the below, because new.variables[k] is the
+        # same object as self.variables[k]!
+        # new = self.copy(deep=False)
```
shoyer: Let's fix `Dataset.copy()` at the source instead so this works, e.g.,
```
    def copy(self, deep=False):
        """Returns a copy of this dataset.

        If `deep=True`, a deep copy is made of each of the componentarray.
        Otherwise, a shallow copy is made, so each array in the new dataset
        is also a array in the original dataset.
        """
        variables = OrderedDict((k, v.copy(deep=deep))
                                for k, v in iteritems(self._variables))
        # skip __init__ to avoid costly validation
        return self._construct_direct(variables, self._coord_names.copy(),
                                      self._dims.copy(), self._attrs_copy())
```

I no longer have any idea why I thought it made sense to not copy variables in `Dataset.copy()`. Copying variables is cheap, and exposing details about the `Variable` object in the public API makes the `Dataset` abstraction leaky.

xarray/core/dataset.py
```diff
@@ -306,7 +309,7 @@ def load(self):
         """
         # access .data to coerce everything to numpy or dask arrays
         all_data = dict((k, v.data) for k, v in self.variables.items())
-        lazy_data = dict((k, v) for k, v in all_data.items()
+        lazy_data = OrderedDict((k, v) for k, v in all_data.items()
```
shoyer: I don't think this matters. Variables are already assigned in the proper order in `self.variables`, and we never return this object to the user.

xarray/core/variable.py
```diff
@@ -272,9 +272,19 @@ def data(self, data):
...
-            self._data = np.asarray(self._data)
-        return self._data
+        """Load data into memory and return it.
+        Do not cache dask arrays automatically; that should
+        require an explicit load() call.
+        """
+        if isinstance(self._data, (np.ndarray, PandasIndexAdapter)):
+            new_data = self._data
+        else:
+            new_data = np.asarray(self._data)
```
shoyer: Can we make a separate method for the logic above? e.g.,
```
def _data_cast(self):
    if isinstance(self._data, (np.ndarray, PandasIndexAdapter)):
        return self._data
    else:
        return np.asarray(self._data)
```

Then both `Variable` and `IndexVariable` can share the caching logic in `_data_cached` and only reimplement `_data_cast`.

xarray/test/test_backends.py
```diff
@@ -992,6 +1020,38 @@ def test_deterministic_names(self):
                 self.assertIn(tmp, dask_name)
             self.assertEqual(original_names, repeat_names)
 
+    def test_dataarray_compute(self):
```
shoyer: These methods would belong better in `test_dask.py`

xarray/test/test_backends.py
```diff
@@ -149,6 +149,27 @@ def assert_loads(vars=None):
...
+
+        with self.roundtrip(expected) as actual:
+            # Test Dataset.compute()
+            for v in actual.variables.values():
+                self.assertFalse(v._in_memory)
+
+            computed = actual.compute()
+
+            # indexes are going to be loaded in memory anyway, even
+            # with compute(). This is by design.
```
shoyer: I think I'm missing something here -- why does this happen? I don't see any `data_vars` specific logic above.

doc/whats-new.rst
```diff
@@ -45,33 +49,32 @@ Deprecations
...
   overlapping (:issue:`835`) coordinates as long as all present data agrees.
   By `Johnnie Gray <https://github.com/jcmgray>`_. See
   :ref:`combining.no_conflicts` for more details.
-
 - It is now possible to set ``concat_dim=None`` explicitly in
   :py:func:`~xarray.open_mfdataset` to disable inferring a dimension along
   which to concatenate.
   By `Stephan Hoyer <https://github.com/shoyer>`_.
+- Added :py:meth:`compute` method to :py:class:`DataArray`, :py:class:`Dataset`,
+  and :py:class:`Variable` as a non-destructive alternative to :py:meth:`load`.
```
shoyer: I would call it "non-mutating" rather than "non-destructive"

shoyer: Also: I don't think this works:

    :py:meth:`compute`

I think it needs the `Dataset`/`DataArray` qualification to be a correct link, e.g,.

    :py:meth:`~DataArray.compute`

`~` hides the preface `DataArray` in the generated HTML.

doc/whats-new.rst
```diff
@@ -25,6 +25,10 @@ Breaking changes
   merges will now succeed in cases that previously raised
   ``xarray.MergeError``. Set ``compat='broadcast_equals'`` to restore the
   previous default.
+- Pickling an xarray object based on the dask backend, or reading its
+  :py:meth:`values` property, won't automatically convert the backend to numpy
```
shoyer: Instead of "convert the backend to numpy" let's say "convert the array from dask to numpy"

Somewhat confusingly, we use the name `backends` in xarray to refer to IO methods (e.g., netCDF libraries), not dask vs numpy.

xarray/core/dataset.py
crusaderky: on line 318, you're iterating on the values of lazy_data to generate the evaluated_data list, and then you're recombining the two by iterating on the keys of lazy_data on line 320.
_Typically_, if you iterate on a dict twice in a row you _should_ get the same order, but there's nothing guaranteeing that. So unless you use a container with order you're effectively relying on an implementation detail of CPython.

xarray/core/dataset.py
crusaderky: That was actually my first try, but hell broke loose in the regression tests when I did. I'll try again later...

xarray/core/variable.py
crusaderky: fixing it

doc/whats-new.rst
crusaderky: fixing it

doc/whats-new.rst
crusaderky: fixing it

xarray/test/test_backends.py
crusaderky: sorry, that was a comment from when I was trying to have IndexVariable do the caching and Variable do not. But then I realised that in some cases the data of a dask-backed array is actually an IndexVariable. Removing the comment.

xarray/test/test_backends.py
crusaderky: I don't think we should move test_dataarray_compute() to a different module as there's test_dataset_compute() that applies to all backends.
Moving the pickle and values tests.

xarray/core/dataset.py
crusaderky: It's done. However please see the caveat I wrote in the test script.

Further explanation to what I mean:
```
import numpy as np
from xarray import Variable, IndexVariable
from xarray.test import source_ndarray

def test_copy(data, cls):
    data = np.array(data)
    a = cls('x', data)
    print source_ndarray(data) is source_ndarray(a.data)
    b = a.copy(deep=False)
    print source_ndarray(a.data) is source_ndarray(b.data)
    
test_copy([1,2,3], Variable)
test_copy(['a', 'b', 'c'], Variable)
test_copy([1,2,3], IndexVariable)
test_copy(['a', 'b', 'c'], IndexVariable)
```
Output:
True
True
True
True
True
True
False
False

crusaderky:
All comments answered, I think I'm done now

xarray/core/dataset.py
shoyer: Iterating over *different* dicts is not guaranteed to happen in the same order, but it is guaranteed for the same dict without any modifications. See:

"If items(), keys(), values(), iteritems(), iterkeys(), and itervalues() are called with no intervening modifications to the dictionary, the lists will directly correspond."
https://docs.python.org/2/library/stdtypes.html#dict.items

"Keys and values are iterated over in an arbitrary order which is non-random, varies across Python implementations, and depends on the dictionary’s history of insertions and deletions. If keys, values and items views are iterated over with no intervening modifications to the dictionary, the order of items will directly correspond."
https://docs.python.org/3.6/library/stdtypes.html#dict-views

xarray/core/dataset.py
```diff
@@ -404,11 +420,8 @@ def copy(self, deep=False):
         Otherwise, a shallow copy is made, so each variable in the new dataset
```
shoyer: can you update this to say it makes a shallow copy of each of the component variables?

xarray/core/variable.py
```diff
@@ -288,11 +299,26 @@ def load(self):
         because all xarray functions should either work on deferred data or
         load data automatically.
         """
-        self._data_cached()
+        new_data = self._data_cached()
```
shoyer: I would just assign unambiguously here, e.g.,
```
self._data = self._data_cast()
```
It does the same thing just a little simpler (no need to account for dask separately).

crusaderky:
All done

xarray/test/test_dask.py
```diff
@@ -13,44 +13,22 @@
...
-    else:
-        assert False
-
-
 class DaskTestCase(TestCase):
     def assertLazyAnd(self, expected, actual, test):
-        expected_copy = _copy_at_variable_level(expected)
-        actual_copy = _copy_at_variable_level(actual)
+        expected.name = None
+        actual.name = None
```
shoyer: Was there a failing tests that required setting this? I would rather disable this on a case by case basis than across the board (not all objects passed to this method even have a `name` attribute).

crusaderky: Yes, on several cases the test was failing because on one side the object automatically got its name from the underlying dask.name, while on the other side it was blank.

shoyer: can you simply make a `assertLazyAndEqual` for those cases or set `name='foo'`? should be pretty quick

crusaderky: fixed and resolved conflict

crusaderky:
Sorry, still fighting against git #1024

PR #1022, opened 2016-09-30
Dask broadcast

crusaderky:
Applying broadcast() to an xarray object based on the dask backend won't accidentally convert the array from dask to numpy anymore


shoyer:
The fix looks good to me, but you accidentally pulled in your other commits from #1018

PR #1023, opened 2016-09-30
Applying broadcast() to an xarray object based on the dask backend wo…

crusaderky:
…n't accidentally convert the array from dask to numpy anymore


shoyer:
Thanks!

PR #1024, opened 2016-10-01
Disable automatic cache with dask

### LINKED ISSUE #902: Pickle and .value vs. dask backend
Pickling a xarray.DataArray with dask backend will cause it to resolve the .data to a numpy array.
This is not desirable, as there's legitimate use cases where you may want to e.g. save a computation for later, or send it somewhere across the network.

Analogously, auto-converting a dask xarray to a numpy xarray as soon as you invoke the .value property is probably nice when you are working on a jupyter terminal, but not in a general purpose situation, particularly when xarray is used at the foundation of a very complex framework. Most of my headaches so far have been caused trying to figure out when, where and why the dask backend was replaced with numpy.

IMHO a module-wide switch to disable implicit dask->numpy conversion would be a nice solution.
A new method, compute(), could explicitly convert in place from dask to numpy.

crusaderky:
Disabled auto-caching on dask; new .compute() method

Fixes #902


crusaderky:
Well, crud. This introduces a regression where DataArray.chunk() converts the data *and the coords* to dask. This becomes enormously problematic later on as pretty much nothing expects a dask-based coord.

xarray/core/dataset.py
```diff
@@ -792,13 +806,19 @@ def chunks(self):
...
         chunks = {}
-        for v in self.variables.values():
+        for v in self.data_vars.values():
             if v.chunks is not None:
                 new_chunks = list(zip(v.dims, v.chunks))
                 if any(chunk != chunks[d] for d, chunk in new_chunks
                        if d in chunks):
                     raise ValueError('inconsistent chunks')
                 chunks.update(new_chunks)
+        if chunks:
```
shoyer: Why should this need `chunks` to not be empty already? That seems strange (maybe backwards) to me.

I might simply make this:
```
for dim, size in self.dims.items():
    if dim not in chunks:
        chunks[dim] = (size,)
```

crusaderky:
What happened before this PR was that all coords were blindly converted to dask on chunk(). Then, the first time anything invoked the values property, e.g. Something as simple as DataArray.__str__., they were silently converted back to numpy. It wasn't *easy* to accidentally get them in dask format; in fact no unit test noticed before my last commit. 

If you deliberately use a dask array as a coord, it won't be converted to numpy. However I can't think of any reason why anybody would want to do it in practice. 
I'll add it to the breaking changes as if somebody did do the above, the performance of his program will degrade with this release as his coord will risk being evaluated multiple times.

xarray/core/dataset.py
crusaderky: if none of the data_vars use the dask backend, then you want chunks to return None.

crusaderky:
I added the disclaimer in the release notes.
Is there any other outstanding issue?
Thanks

crusaderky:
I can't reproduce the above failure test.test_conventions.TestEncodeCFVariable.testMethod=test_missing_fillvalue.
I suspect it might be a random failure, because
- it used to succeed until my latest commit, which eclusively changed the readme
- it suceeds on python 3

xarray/core/dataset.py
```diff
@@ -792,13 +806,19 @@ def chunks(self):
         array.
         """
         chunks = {}
-        for v in self.variables.values():
+        for v in self.data_vars.values():
```
shoyer: I am concerned about skipping non-`data_vars` here. Coordinates could still be chunked, e.g., if they were loaded from a file, or created directly from dask arrays.

xarray/core/dataset.py
shoyer: I guess this method is inconsistent with `Variable.chunks`, but it currently always returns a dict.

I would either skip this change or use something like my version.

xarray/core/dataset.py
```diff
@@ -851,6 +871,9 @@ def selkeys(dict_, keys):
             return dict((d, dict_[d]) for d in keys if d in dict_)
 
         def maybe_chunk(name, var, chunks):
+            if name not in self.data_vars:
```
shoyer: I see your point about performance, but I think that mostly holds true for indexes. So I would be inclined to adjust this to only skip variables in `self.dims` (aka indexes used for alignment).

I am still concerned about skipping coords if they are already dask arrays. If they are already dask arrays, then `.chunk()` should probably adjust their chunks anyways.

crusaderky:
I've been thinking about this... Maybe the simple, clean solution is to
simply invoke compute() on all coords as soon as they are assigned to the
DataArray / Dataset?

crusaderky:
ping - how do you prefer me to proceed?

shoyer:
> I've been thinking about this... Maybe the simple, clean solution is to
simply invoke compute() on all coords as soon as they are assigned to the
DataArray / Dataset?

I'm nervous about eager loading, especially for non-index coordinates. They can have more than one dimension, and thus can contain a lot of data. So potentially eagerly loading non-index coordinates could break existing use cases.

On the other hand, non-index coordinates indeed checked for equality in most xarray operations (e.g., for the coordinate merge in align). So it is indeed useful not to have to recompute them all the time.

Even eagerly loading indexes is potentially problematic, if loading the index values is expensive.

So I'm conflicted:
- I like the current caching behavior for `coords` and `indexes`
- But I also want to avoid implicit conversions from dask to numpy, which is problematic for all the reasons you pointed out earlier

I'm going to start throwing out ideas for how to deal with this:

### Option A

Add two new (public?) methods, something like `.load_coords()` and `.load_indexes()`. We would then insert calls to these methods at the start of each function that uses coordinates:

- `.load_indexes()`: `reindex`, `reindex_like`, `align` and `sel`
- `.load_coords()`: `merge` and anything that calls the functions in `core/merge.py` (this indirectly includes `Dataset.__init__` and `Dataset.__setitem__`)

Hypothetically, we could even have options for turning this caching systematically on/off (e.g., `with xarray.set_options(cache_coords=False, cache_indexes=True): ...`).

Your proposal is basically an extreme version of this, where we call `.load_coords()` immediately after constructing every new object.

Advantages:

- It's fairly predictable when caching happens (especially if we opt for calling `.load_cords()` immediately, as you propose).
- Computing variables is all done at once, which is much more performant than what we currently do, e.g., loading variables as needed for `.equals()` checks in `merge_variables` one at a time.
 
Downsides:

- Caching is more aggressive than necessary -- we cache indexes even if that coord isn't actually indexed.

### Option B

Like Option A, but someone infer the full set of variables that need to be cached (e.g., in a `.merge()` operation) before it's actually done. This seems hard, but maybe is possible using a variation of `merge_variables`.

This solves the downside of A, but diminishes the predictability. We're basically back to how things work now.

### Option C

Cache dask.array in `IndexVariable` but not `Variable`. This preserves performance for repeated indexing, because the hash table behind the `pandas.Index` doesn't get thrown away.

Advantages:

- Much simpler and easier to implement than the alternatives.
- Implicit conversions are greatly diminished.

Downsides:

- Non-index coordinates get thrown away after being evaluated once. If you're doing lots of operations of the form `[ds + other for ds in datasets]` where `ds` and `other` has conflicting coordinates this would probably make you unhappy.

### Option D

Load the contents of an `IndexVariable` immediately and eagerly. They no longer cache data or use lazy loading.

This has the most predictable performance, but might cause trouble for some edge use cases?

-------------------

I need to think about this a little more, but right now I am leaning towards Option C or D.

crusaderky:
Hi Stephen,
Thank you for your thinking.
IMHO option D is the cleanest and safest. Could you come up with any
example where it may be problematic?

benbovy:
Option D seems indeed the cleanest and safest option, but

> Even eagerly loading indexes is potentially problematic, if loading the index values is expensive.

I can see use cases where this might happen. For example, It is common for 1, 2 or higher-dimension unstructured meshes that the coordinates x, y, z are arranged as 1-d arrays of length that equals the number of nodes (which can be very high!). See for example [ugrid conventions](http://ugrid-conventions.github.io/ugrid-conventions/).

I admit that currently xarray is perhaps not very suited for handling unstructured meshes, but IMO it has great potential (especially considering multi-index support) and I'd love to use it here.

shoyer:
> I admit that currently xarray is perhaps not very suited for handling unstructured meshes, but IMO it has great potential (especially considering multi-index support) and I'd love to use it here.

Right now, xarray is not going to be great fit for such cases, because we already cache an index in memory for any labeled indexing operations. So at best, you could do something like `ds.isel(mesh_edge=slice(int(1e6)))`. Maybe people already do this?

I doubt very many people are relying on this, though indeed, this would include some users of an [array database we wrote at my former employer](https://github.com/TheClimateCorporation/mandoline), which did not support different chunking schemes for different variables (which could make coordinate lookup very slow). CC @ToddSmall in case he has opinions here.

For out-of-core operations with labels on big unstructured meshes, you really need a generalization of the `pandas.Index` that doesn't need to live in memory (or maybe that lives in memory on some remote server).

benbovy:
> we already cache an index in memory for any labeled indexing operations

Oh yes, true!

> So at best, you could do something like ds.isel(mesh_edge=slice(int(1e6)))

Indeed, that doesn't look very nice.

> For out-of-core operations with labels on big unstructured meshes, you really need a generalization of the pandas.Index that doesn't need to live in memory

From what I intend to do next with xarray, I'd say that extending its support for out-of-core operations to big indexes would be a great feature! I haven't seen yet how `dask.Dataframe` works internally (including `dask.Dataframe.index`and `dask.Dataframe.loc`), but I guess maybe this could be transposed in some way to the indexing logic in xarray? Though I'm certainly missing a lot of potential issues here... Anyway, I can open a new issue to discuss more about this if you think it's worth it.

shoyer:
> Anyway, I can open a new issue to discuss more about this if you think it's worth it.

Yes, please do!

@crusaderky I think we are OK going ahead here with Option D. If we do eventually extend xarray with out of core indexes, that will be done with a separate layer (not in `IndexVariable`).

crusaderky:
roger that, getting to work :)

shoyer:
Awesome, thanks for your help!

crusaderky:
Finished and waiting for code review

xarray/core/variable.py
```diff
@@ -1076,10 +1101,16 @@ def __init__(self, dims, data, attrs=None, encoding=None, fastpath=False):
                              type(self).__name__)
 
     def _data_cached(self):
-        if not isinstance(self._data, PandasIndexAdapter):
-            self._data = PandasIndexAdapter(self._data)
+        # Unlike in Variable._data_cached, always eagerly resolve dask arrays
```
shoyer: I thought we wanted to always eagerly load this into memory without caching at all?

That would suggest we should put something like `self._data = PandasIndexAdapter(self._data)` in the constructor, and make `_data_cached` and `_data_cast` on the subclass dummy methods.

xarray/core/dataset.py
```diff
@@ -874,6 +887,9 @@ def selkeys(dict_, keys):
             return dict((d, dict_[d]) for d in keys if d in dict_)
 
         def maybe_chunk(name, var, chunks):
+            if name in self.dims:
```
shoyer: Actually, maybe put this logic in `IndexVariable` instead? We could define a `chunk` method that looks like:
```
def chunk(self, ...):
    return self.copy(deep=False)
```

crusaderky:
Changed to cache IndexVariable.\_data on \_\_init\_\_. Please review...

xarray/test/test_backends.py
```diff
@@ -152,6 +156,31 @@ def assert_loads(vars=None):
...
 
+    def test_dataset_compute(self):
+        expected = create_test_data()
+
+        with self.roundtrip(expected) as actual:
+            # Test Dataset.compute()
+            for k, v in actual.variables.items():
+                # IndexVariables are eagerly cached
+                if k in actual.dims:
+                    self.assertTrue(v._in_memory)
```
shoyer: This would be slightly simpler just as `self.assertEqual(v._in_memory, k in actual.dims)`

shoyer:
Thanks for your patience! This is a nice improvement.

I have an idea for a variation that might make for a cleaner (less dask specific) way to handle the remaining caching logic -- I'll add you a reviewer on that PR.

crusaderky:
Happy to contribute!

kynan:
@crusaderky @shoyer There are still cases where dask arrays are converted to ndarrays where I think they shouldn't be: if you create a `Variable` with a dask array (e.g. in a custom data store), it gets wrapped into a `LazilyIndexedArray` at the [end of `decode_cf_variable](https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L833). This will subsequently be cast to an ndarray in `_data_cast`.

shoyer:
@kynan I think this is fixed in #1128, which has a slightly more robust solution.

kynan:
@shoyer Great, thanks, I'll give that a try.

PR #1027, opened 2016-10-02
Groupby bins empty groups

### LINKED ISSUE #1019: groupby_bins: exclude bin or assign bin with nan when bin has no values
When using groupby_bins there are cases where no values are found for some of the bins specified. Currently, it appears that in these cases, the bin is skipped, with no value neither a bin entry added to the output dataarray.

Is there a way to identify which bins have been skipped. Or preferably, is it possible to have an option to include those bins, but with nan values. This would make comparing two dataarrays easier in cases where despite the same bin intervals as inputs, the outputs result in dataarrays with different variable and coordinates lengths.

```
import xarray as xr
var = xr.open_dataset('c:\\users\\saveMWE.nc')
pop = xr.open_dataset('c:\\users\\savePOP.nc')
# binns includes very small bin to test this
binns = [-100, -50, 0, 50, 50.00001, 100]
binned = pop.p2010T.groupby_bins(var.EnsembleMean, binns).sum()
print binned
print binned.EnsembleMean_bins
```

In this case, no data falls in the 4th bin between 50 and 50.00001. 

```
<xarray.DataArray 'p2010T' (EnsembleMean_bins: 4)>
array([  2.64352214e+09,   3.46869168e+09,   3.08998110e+08,
         1.48247440e+07])
Coordinates:
  * EnsembleMean_bins  (EnsembleMean_bins) object '(0, 50]' '(-50, 0]' ...
<xarray.DataArray 'EnsembleMean_bins' (EnsembleMean_bins: 4)>
array(['(0, 50]', '(-50, 0]', '(51, 100]', '(-100, -50]'], dtype=object)
```
Obviously one can count the lengths but this doesn't indicate which bin was skipped. An option to include the empty bin with a nan value would be useful! Thanks

[bins_example.zip](https://github.com/pydata/xarray/files/499952/bins_example.zip)

rabernat:
This PR fixes a bug in `groupby_bins` in which empty bins were dropped from the grouped results. Now `groupby_bins` restores any empty bins automatically. To recover the old behavior, one could apply `dropna` after a groupby operation.

Fixes #1019

### DIFF
--- a/xarray/test/test_dataarray.py
+++ b/xarray/test/test_dataarray.py
@@ -1428,12 +1428,6 @@ class TestDataArray(TestCase):
         # (was a problem in earlier versions)
         self.assertEqual(len(array.dim_0), 4)
 
-        # now do the same
-        actual = array.groupby_bins('dim_0', bins, drop_empty_bins=True).sum()
-        expected = DataArray([6], dims='dim_0_bins',
-                        coords={'dim_0_bins': ['(0, 4]']})
-        self.assertDataArrayIdentical(expected, actual)
-
     def test_groupby_bins_multidim(self):
         array = self.make_groupby_multidim_example_array()
         bins = [0,15,20]
--- a/xarray/test/test_dataarray.py
+++ b/xarray/test/test_dataarray.py
@@ -1416,6 +1416,24 @@ class TestDataArray(TestCase):
         # (would fail with shortcut=True above)
         self.assertEqual(len(array.dim_0), 4)
 
+    def test_groupby_bins_empty(self):
+        array = DataArray(np.arange(4), dims='dim_0')
+        # one of these bins will be empty
+        bins = [0,4,5]
+        actual = array.groupby_bins('dim_0', bins).sum()
+        expected = DataArray([6], dims='dim_0_bins',
+                        coords={'dim_0_bins': ['(0, 4]']})
+        self.assertDataArrayIdentical(expected, actual)
+        # make sure original array is unchanged
+        # (was a problem in earlier versions)
+        self.assertEqual(len(array.dim_0), 4)
+
+        # now do the same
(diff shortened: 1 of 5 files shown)


xarray/test/test_dataarray.py
```diff
@@ -1416,6 +1416,25 @@ def test_groupby_bins(self):
         # (would fail with shortcut=True above)
         self.assertEqual(len(array.dim_0), 4)
 
+    def test_groupby_bins_empty(self):
+        array = DataArray(np.arange(4), dims='dim_0')
+        # one of these bins will be empty
+        bins = [0,4,5]
+        actual = array.groupby_bins('dim_0', bins, drop_empty_bins=True).sum()
+        print(actual)
```
MaximilianR: Prob don't want to keep this

shoyer:
Rather than adding a new keyword argument, I'm inclined to unilaterally set `drop_empty_bins=True` and treat the former behavior as a bug. I can't see many use cases for the previous behavior.

xarray/test/test_dataarray.py
rabernat: oops! yes forgot to take that out

rabernat:
@shoyer So do you want the keyword argument dropped altogether? Or just the default changed (as in my last commit)?

shoyer:
Yeah my inclination would be not to add the option at all.

rabernat:
I guess it's pretty easy to just do `dropna`

fmaussion:
Just for future readers like me: the point of this PR is to make ``drop_empty_bins = False`` per default, right? (not True as Stephan wrote). Maybe a new edit of @rabernat 's original post would help.

rabernat:
@fmaussion it's a little more involved than that...see edited description.

shoyer:
Thanks!

PR #1028, opened 2016-10-03
Add `set_index`, `reset_index` and `reorder_levels` methods

benbovy:
Another item in #719.

I added tests and updated the docs, so this is ready for review.


xarray/core/dataarray.py
```diff
@@ -816,6 +816,118 @@ def swap_dims(self, dims_dict):
...
+        Returns
+        -------
+        reindexed : DataArray
+            Another dataarray, with this dataarray's data but replaced coordinates.
+
+        See Also
+        --------
+        DataArray.reset_index
+        """
+        indexers = utils.combine_pos_and_kw_args(indexers, kw_indexers,
```
shoyer: I know I wrote this utility (and it is currently used by `Dataset.reindex`), but I regret it now! I feel like it's usually better to only have a single call signature, accepting either a dictionary or `**kwargs`, but not both.

This goes for the other new methods, too.

xarray/core/dataarray.py
```diff
@@ -816,6 +816,118 @@ def swap_dims(self, dims_dict):
...
+        ----------
+        indexers : dict, optional
+            Dictionary with keys given by dimension names and values given by
+            (lists of) the names of existing coordinates.
+            Any list of multiple names given for a dimension will result as
+            a MultiIndex for that dimension.
+        append : bool, optional
+            If True, append the supplied indexers to the existing indexes.
+            Otherwise replace the existing indexes (default).
+        inplace : bool, optional
```
shoyer: I'm not sure we need an `inplace=True` option. I guess it doesn't hurt. (Just more to test)

xarray/core/dataarray.py
```diff
@@ -816,6 +816,118 @@ def swap_dims(self, dims_dict):
         ds = self._to_temp_dataset().swap_dims(dims_dict)
         return self._from_temp_dataset(ds)
 
+    def set_index(self, indexers=None, append=False, inplace=False,
+                  **kw_indexers):
+        """Set DataArray (multi-)indexes using one or more existing coordinates.
+
+        Parameters
+        ----------
+        indexers : dict, optional
```
shoyer: Instead of `array.set_index({'record': ['level_0', 'level_1']})` or `array.set_index(record=['level_0', 'level_1'])`, it might suffice to simply use `array.set_index(['level_0', 'level_1'])`, without specifying which dimension these get added to. We can infer the dimension names from the variables.

The upside is that accepting a list instead of a dict is a little more succinct. The downside is that it's less explicit/self-documenting.

xarray/core/dataarray.py
```diff
@@ -816,6 +816,118 @@ def swap_dims(self, dims_dict):
...
+        """
+        indexers = utils.combine_pos_and_kw_args(indexers, kw_indexers,
+                                                 'set_index')
+        coords, _ = merge_indexes(indexers, self._coords, set(), append=append)
+        if inplace:
+            self._coords = coords
+        else:
+            return self._replace(coords=coords)
+
+    def reset_index(self, dim_levels=None, drop=False, inplace=False,
```
shoyer: Maybe switch the signature to use separate arguments like `reset_index(dim, levels=None, ...)` instead of using a dict/`**kwargs`. This would make the usual use a clearer, e.g., `array.reset_index('record')` instead of `array.reset_index(record=None)`.

Also, after #1017 (optional indexes), the ability to write `array.reset_index('x', drop=True)` for clearing an index could be nice to have.

xarray/core/dataset.py
```diff
@@ -101,6 +102,95 @@ def calculate_dimensions(variables):
...
+        names = []
+        arrays = []
+        current_index_variable = variables[dim]
+
+        if append:
+            current_index = current_index_variable.to_index()
+            if isinstance(current_index, pd.MultiIndex):
+                names.extend(current_index.names)
+                for i in range(current_index.nlevels):
+                    arrays.append(current_index.get_level_values(i))
```
shoyer: It might be premature optimization to worry about this, but something to watch out for is that this will refactorize the existing MultiIndex. It might be better to simply reuse existing `levels` and `labels`, if possible, and directly pass those into the `MultiIndex` constructor.

Note that to create `levels` and `labels` directly form an array (if necessary) you can use `Categorical` as used in `MultiIndex.from_arrays`, e.g.,
```python
cat = pd.Categorical(array, ordered=True)
levels = cat.categories
labels = cat.codes
```

xarray/core/dataarray.py
```diff
@@ -816,6 +816,118 @@ def swap_dims(self, dims_dict):
         ds = self._to_temp_dataset().swap_dims(dims_dict)
         return self._from_temp_dataset(ds)
 
+    def set_index(self, indexers=None, append=False, inplace=False,
```
shoyer: A note on terminology: in xarray/pandas, `indexer` usually refers the argument/key used for indexing, whereas `index` refers to the set of existing labels, e.g., `df.index` vs `df.loc[indexer]`. So in this case I think the argument name `indexes` would make more sense.

xarray/core/dataarray.py
benbovy: Another downside of using a single argument is that it is less consistent with other parts of the xarray's API which use keyword arguments (e.g., `reindex`, `sel`, etc.). IMHO the little additional verbosity is worth the gain in explicit/self-documenting here.

That said, a single `dim_levels` or `levels` argument may be more consistent with the signature you suggest below for reset_index: `reset_index(dim, levels=None,...)`. So finally I don't really know what is better.

xarray/core/dataarray.py
benbovy: OK, I've naively copied the signature of `reindex` but now I get it.

xarray/core/dataarray.py
benbovy: I would be for using `**kwargs` unless we choose the alternative signatures suggested in the comments above and below.

benbovy:
Some API design questions (mostly from @shoyer's review) we need to fix:

- We need to choose whether to use `dim=indexes` kwargs or fixed arg/kwarg relative to a given dimension for the signatures of `.set_index()`, `.reset_index()` and `.reorder_levels()`.
- Do we also allow `.set_index()` to rename the dimension(s) if needed, instead of doing `.set_index(...).rename(...)` ? Is this a common use case that is worth it?
- After discussion in #1017, it seems that we need an easy way to (re)set indexes *either* to no index *or* to range(n).

For point 1, my preference goes to `dim=indexes` kwargs, especially if we need 2 and 3. It is less succinct, but its more close to the signatures of other xarray methods like `.reindex()` or `.sel()`, and it allows (re)setting the indexes of multiple dimensions in a single call. Given 2, I find `set_index(new_dim_name=['level_1', 'level_3'])` a bit more elegant than `set_index(['level_1', 'level_2'], name='new_dim_name')`. Given 3, `array.reset_index('x')` seem ambiguous compared to `array.reset_index(x=None)` (no index) and, e.g., `array.reset_index(x='range')` (range(n) index).

shoyer:
> We need to choose whether to use dim=indexes kwargs or fixed arg/kwarg relative to a given dimension for the signatures of .set_index(), .reset_index() and .reorder_levels().

For `set_index` and `reorder_levels`, I like the kwargs or a dictionary. It's nice and explicit. But for `reset_index`, I think we probably want a list.

It's not at all obvious to me what `array.reset_index(x=None)` does. It could just as easily mean "reset nothing from `x`" as "reset `x` to have a null index". In fact, the former seems more consistent with how we handle levels. In contrast, `array.reset_index(['x'])` pretty clearly means that the `'x'` index should be reset.

> Do we also allow .set_index() to rename the dimension(s) if needed, instead of doing .set_index(...).rename(...) ? Is this a common use case that is worth it?

My inclination is yes -- this feels like a common thing to do. But we could also safely add this later.

> After discussion in #1017, it seems that we need an easy way to (re)set indexes either to no index or to range(n).

We definitely need a way to reset indexes to the default, but after #1017, I'm not sure we will need a way to set them to `range(n)`.

Unfortunately, if `x` is a normal (non-multi) Index, `array.reset_index('x')` is not well defined. We need a name for the variable that was formerly named `x` (or could drop it, e.g., with `array.reset_index('x', drop=True)` or `array.drop('x')`), otherwise it will still be the index for the x-axis. Or, I suppose we could rename the `x` dimension to something else.

One option is to add some sort of prefix or suffix to the index name when it becomes a new variable, e.g., `array.reset_index('x')` renames the coordinate `x` to `x_`. This seems like a probably safe choice, though I hate to add more automatic names to the API.

benbovy:
Sorry for the delay @shoyer. I've read your comments above and they all seem relevant. I'll find some time next week to get back on this.

benbovy:
Just committed review changes.

`.reset_index()` doesn't accept kwargs anymore, though I don't know what to choose between the options below (currently option A is implemented):

- option A: `reset_index(dim, levels=None)` where `dim` may accept multiple dimension names (in 
   that case `levels` must be a list of lists with the same length than `dim`, or simply `None` that
   would then be applied to all given dimensions).
- option B: same than option A, `reset_index(dim, levels=None)`, except that `dim` only accepts
   one dimension (thus a bit simpler but less flexible).
- option C: `reset_index(dim_or_levels)` where one can provide a list of dimension(s) and/or
   level(s). This is the most flexible and concise, though maybe less readable. Allow providing both
   dimensions and levels may be ambiguous too.

> if x is a normal (non-multi) Index, array.reset_index('x') is not well defined

Currently `.reset_index()` doesn't allow resetting normal indexes, but we can wait for #1017 before merging this.

xarray/core/dataset.py
```diff
@@ -102,6 +103,105 @@ def calculate_dimensions(variables):
     return dims
 
 
+def merge_indexes(indexes, variables, coord_names, append=False):
```
shoyer: Can you try adding type annotations here, like the ones I started adding in `core/merge.py`? I'm not even running mypy yet but I think these could significantly improve readability, and are lighter weight than adding a full docstring.

xarray/core/dataarray.py
```diff
@@ -821,6 +821,108 @@ def swap_dims(self, dims_dict):
...
+            If True, set new index(es) in-place. Otherwise, return a new DataArray
+            object.
+        **indexes : {dim: index, ...}
+            Keyword arguments with names matching dimensions and values given
+            by (lists of) the names of existing coordinates or variables to set
+            as new (multi-)index.
+
+        Returns
+        -------
+        reindexed : DataArray
```
shoyer: wrong name -- should not be `reindexed`

shoyer:
I kind of like Option C, given that we have guaranteed level names and variables to have no conflicts.

Did you go for allowing `set_index()` to rename variables?

benbovy:
This is ready for another round of review.

I've changed the signature of `reset_index` to option C. It is also almost ready for #1017 (just added two small TODOs). 

> Did you go for allowing set_index() to rename variables?

Not yet, but as you said we could safely add this later.

shoyer:
Can we update this for optional indexes? (now on master)

benbovy:
This should now behave correctly with optional indexes.

doc/indexing.rst
```diff
@@ -478,6 +478,49 @@ Both ``reindex_like`` and ``align`` work interchangeably between
     # this is a no-op, because there are no shared dimension names
     ds.reindex_like(other)
 
+.. _multi-index handling:
+
+Multi-index handling
```
shoyer: These docs are great, but I wouldn't call them "indexing methods" exactly. Maybe move this section to [Reshaping and reorganizing data](http://xarray.pydata.org/en/stable/reshaping.html)?

benbovy: Makes sense!

PR #1035, opened 2016-10-04
fixed broken links to data & IPython notebook

gdmcbain:
And fixed the changed name of the data-file in the example source-code


shoyer:
Thanks

I think @jhamman actually partially updated the notebook already, but it wasn't exported to the docs?

gdmcbain:
Oh.  I just followed the 'Edit on GitHub' link from http://xarray.pydata.org/en/stable/examples/monthly-means.html so figured this rst was the source rather than a generated file.  Sorry.

jhamman:
Yes, my edits are sitting on master and will be pushed to stable on our next release. Thanks for looking into this though.

Issue #1042, opened 2016-10-09
Dataset.groupby() doesn't preserve variables order


fmaussion:
Is it intentional? I think it is rather undesirable, but maybe there is some reason for this.

shoyer:
This is probably a bug. Usually, we're pretty careful to always use `OrderedDict` internally for exactly this reason. Can you give a reproducible example?

fmaussion:
Thanks @shoyer , here's a mwe:

```python
import xarray as xr
import numpy as np 
ds = xr.Dataset()
for vn in ['a', 'b', 'c']:
    ds[vn] = xr.DataArray(np.arange(10), dims=['t'])
ds.groupby('t').mean()

<xarray.Dataset>
Dimensions:  (t: 10)
Coordinates:
  * t        (t) int64 0 1 2 3 4 5 6 7 8 9
Data variables:
    a        (t) float64 0.0 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0
    c        (t) float64 0.0 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0
    b        (t) float64 0.0 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0
```

fmaussion:
@shoyer I'd be happy to provide a fix if you want. Could you give a short pointer as to where the logic is implemented?

shoyer:
So the tricky part here is that it's not obvious what is breaking here. One clue is that reducing doesn't seem to be necessary -- I can reproduce this just with applying an identity transform:
```
In [6]: identity = lambda x: x

In [7]: ds.groupby('t').apply(identity)
Out[7]:
<xarray.Dataset>
Dimensions:  (t: 10)
Coordinates:
  * t        (t) int64 0 1 2 3 4 5 6 7 8 9
Data variables:
    a        (t) int64 0 1 2 3 4 5 6 7 8 9
    c        (t) int64 0 1 2 3 4 5 6 7 8 9
    b        (t) int64 0 1 2 3 4 5 6 7 8 9
```

Actually, it looks like it's probably a `concat` bug:
```
In [17]: gb = ds.groupby('t')

In [18]: grouped = [v for _, v in gb]

In [20]: [list(g.data_vars) for g in grouped]
Out[20]:
[['a', 'b', 'c'],
 ['a', 'b', 'c'],
 ['a', 'b', 'c'],
 ['a', 'b', 'c'],
[...]

PR #1048, opened 2016-10-15
WIP: preserve variables order in Dataset.concat() 

fmaussion:
addresses https://github.com/pydata/xarray/issues/1042

The problem is that sets aren't ordered. I tried something based on an OrderedSet recipe, but I'm not sure that this is the most elegant way to go.

A remaining issue is that the coord variable is still not in place (`['t', 'a', 'b', 'c']` becomes `['a', 'b', 'c', 't']`), but I'm not sure it's possible to do it better (it happens [here](https://github.com/pydata/xarray/blob/master/xarray/core/combine.py#L273-L274)).

Any idea welcome

### DIFF
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -145,6 +145,8 @@ def _calc_concat_over(datasets, dim, data_vars, coords):
     Determine which dataset variables need to be concatenated in the result,
     and which can simply be taken from the first dataset.
     """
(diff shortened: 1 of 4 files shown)


fmaussion:
I think the alternative implementation here is better: https://github.com/pydata/xarray/pull/1049

PR #1049, opened 2016-10-15
Preserve variables order in Dataset.concat()

fmaussion:
Alternative (and much easier) implementation to https://github.com/pydata/xarray/pull/1048

The problem with the coordinate variables remains the same

### DIFF
--- a/xarray/test/test_dataset.py
+++ b/xarray/test/test_dataset.py
@@ -1858,6 +1858,21 @@ class TestDataset(TestCase):
         expected = Dataset({'foo': ('bar', [1.5, 3]), 'bar': [1, 2]})
         self.assertDatasetIdentical(actual, expected)
 
+    def test_groupby_order(self):
+        # groupby should preserve variables order
+
+        ds = Dataset()
+        for vn in ['a', 'b', 'c']:
+            ds[vn] = DataArray(np.arange(10), dims=['t'])
+        all_vars_ref = list(ds.variables.keys())
+        data_vars_ref = list(ds.data_vars.keys())
+        ds = ds.groupby('t').mean()
+        all_vars = list(ds.variables.keys())
+        data_vars = list(ds.data_vars.keys())
+        self.assertEqual(data_vars, data_vars_ref)
+        # coords are now at the end of the list, so the test below fails
+        # self.assertEqual(all_vars, all_vars_ref)
+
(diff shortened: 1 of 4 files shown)


xarray/test/test_dataset.py
```diff
@@ -1858,6 +1858,21 @@ def test_groupby_nan(self):
...
+        for vn in ['a', 'b', 'c']:
+            ds[vn] = DataArray(np.arange(10), dims=['t'])
+        all_vars_ref = list(ds.variables.keys())
+        data_vars_ref = list(ds.data_vars.keys())
+        ds = ds.groupby('t').mean()
+        all_vars = list(ds.variables.keys())
+        data_vars = list(ds.data_vars.keys())
+        self.assertEqual(data_vars, data_vars_ref)
+        # coords are now at the end of the list, so the test below fails
+        # self.assertEqual(all_vars, all_vars_ref)
```
shoyer: Actually, I think this test passes in this version?

shoyer:
Agreed, this is the better approach. Thanks for diving into this!

shoyer:
Alternatively, it looks like the issue is on [line 266](https://github.com/pydata/xarray/blob/62a7f2c00caacb5b917ff320e28aca9cb4396649/xarray/core/combine.py#L268) where we iterate `for k in concat_over`. Instead, we could just do:
```
for k in variables:
    if k in concat_over:
        ...
```
This has the advantage of preserving the set, which is important if there is a very large number of variables.

fmaussion:
Ok, done. The second problem about the coordinate variable seems trickier to solve though...

xarray/test/test_dataset.py
shoyer: Indeed, concat does not currently preserve the order of all variables (yet). But at least we can make it non-stochastic.

shoyer: You could probably do this by consolidating the two loops over all variables. But up to you if you want to go through the trouble.

fmaussion:
> You could probably do this by consolidating the two loops over all variables. But up to you if you want to go through the trouble.

@shoyer could you elaborate on that? If there's a way to make it right I am happy to try, but I'm afraid I'm not sure what is the way to go.

shoyer:
OK, I'm just going to merge this. We can iterate on this more later.

PR #1052, opened 2016-10-18
catch numpy arrays in attrs before converting to dict

jsignell:
Makes it easier to dump to json (after conversation on #917)

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1194,13 +1194,22 @@ class DataArray(AbstractArray, BaseDataObject):
         Convert this xarray.DataArray into a dictionary following xarray
         naming conventions.
 
-        Useful for coverting to json.
+        Useful for coverting to json.To avoid datetime incompatibility
+        use decode_times=False kwarg in xarrray.open_dataset.
 
         See also
         --------
         xarray.DataArray.from_dict
         """
-        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': self.dims}
+        def np_check(attrs):
+            # catch numpy arrays in attrs and convert to nested lists
+            attrs = dict(attrs)
+            for k, v in attrs.items():
+                if hasattr(v, 'tolist'):
+                    attrs.update({k: v.tolist()})
+            return attrs
+
+        d = {'coords': {}, 'attrs': np_check(self.attrs), 'dims': self.dims}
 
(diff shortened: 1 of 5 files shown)


xarray/core/dataarray.py
```diff
@@ -1194,13 +1194,22 @@ def to_dict(self):
         Convert this xarray.DataArray into a dictionary following xarray
         naming conventions.
 
-        Useful for coverting to json.
+        Useful for coverting to json.To avoid datetime incompatibility
```
shoyer: Missing a space after `json`.

Maybe add "Converts all variables and attributes to native Python objects."

xarray/core/dataarray.py
```diff
@@ -1194,13 +1194,22 @@ def to_dict(self):
...
         --------
         xarray.DataArray.from_dict
         """
-        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': self.dims}
+        def np_check(attrs):
+            # catch numpy arrays in attrs and convert to nested lists
+            attrs = dict(attrs)
+            for k, v in attrs.items():
+                if hasattr(v, 'tolist'):
+                    attrs.update({k: v.tolist()})
```
shoyer: just use `attrs[k] = v`

xarray/core/dataarray.py
```diff
@@ -1194,13 +1194,22 @@ def to_dict(self):
...
         See also
         --------
         xarray.DataArray.from_dict
         """
-        d = {'coords': {}, 'attrs': dict(self.attrs), 'dims': self.dims}
+        def np_check(attrs):
+            # catch numpy arrays in attrs and convert to nested lists
+            attrs = dict(attrs)
+            for k, v in attrs.items():
+                if hasattr(v, 'tolist'):
```
shoyer: Let's also fix NumPy scalars values, e.g., `np.float32(0)`. To convert these, use `.item()`.

jsignell:
Are the names `time_check` and `np_check` specific enough for utils.py?

shoyer:
Something more descriptive would probably be a good idea. Maybe `ensure_us_time_resolution` and `decode_numpy_dict_values`?

shoyer:
Let's add a test case including a numpy scalar in an attribute? Otherwise, looks good to me.

shoyer:
Thanks!

PR #1055, opened 2016-10-21
DOC: indexes (#1054)

### LINKED ISSUE #1054: API: are indexes public API?
Usecase -  I have a `DataArray` that will be passed into a numba function and need to pass in some slices.  Very simplified example:

```python
from numba import njit
@njit
def f(arr, slc1, slc2):
    return arr[slc1].max() - arr[slc2].min()

da = xr.DataArray([1., 2., 3.], coords={'key': ['a', 'b', 'c']})
```

Right now I'm accessing the underlying indexes to get slices, like so:

```python
f(da.values, 
  da.indexes['key'].slice_indexer('a', 'b'), 
  da.indexes['key'].slice_indexer('c', 'c'))

Out[52]: -1.0
```
First question, am I missing some obviously better way to do this?  Of course in this example I could just pass in sliced values, but in my actual usecase the data has higher dimensions and I use the slice multiple times across multiple arrays.

More broadly, should the underlying indexes be thought of as an implementation detail (e.g. could be swapped out with something else) or is it more-or-less an API guarantee that I'll get a `pandas` Index here?

chris-b1:
closes #1054

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -42,6 +42,7 @@ Attributes
    Dataset.data_vars
    Dataset.coords
    Dataset.attrs
+   Dataset.indexes
 
 Dictionary interface
 --------------------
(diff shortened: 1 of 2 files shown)


shoyer:
Thanks!

PR #1060, opened 2016-10-25
Remove obsolete NetCDF4 Error catch

fmaussion:
[edit: this original PR was replaced with another change: "remove obsolete NetCDF4 Error catch"]

This allows to give a customized NetCDF4 object to the DataStore.

I needed this for Salem's diagnostic variables: https://github.com/fmaussion/salem/blob/master/salem/sio.py#L737-L751

My workaround (a very [shallow](https://github.com/fmaussion/salem/blob/master/salem/sio.py#L653) subclass of NetCDF4DataStore) is fine for me, so I won't be offended if you decide not to merge.

On a related issue, the IndexError catch which happens [here](https://github.com/pydata/xarray/blob/master/xarray/backends/netCDF4_.py#L52) is also giving me trouble, since it hides bugs in my code. 

I know I'm kind of misusing the backend, but I find it very useful the way it is. I'm open to any suggestion to make it more elegant. Thanks!

### DIFF
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -179,13 +179,16 @@ class NetCDF4DataStore(WritableCFDataStore):
     This store supports NetCDF3, NetCDF4 and OpenDAP datasets.
     """
     def __init__(self, filename, mode='r', format='NETCDF4', group=None,
-                 writer=None, clobber=True, diskless=False, persist=False):
+                 writer=None, clobber=True, diskless=False, persist=False,
+                 ds=None):
         import netCDF4 as nc4
         if format is None:
             format = 'NETCDF4'
-        ds = nc4.Dataset(filename, mode=mode, clobber=clobber,
-                         diskless=diskless, persist=persist,
-                         format=format)
+        if ds is None:
+            # sometimes users want to give a custom NetCDF object
+            ds = nc4.Dataset(filename, mode=mode, clobber=clobber,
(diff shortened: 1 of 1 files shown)


shoyer:
I'm a little nervous about this. The entire datastore API needs some serious rethinking for extensibility.

We may be able to delete that IndexError catch -- that exists to work around a netCDF4 bug that lasted between versions 1.1.2 (Dec 16, 2014) and 1.2.1 (Oct 15, 2015).

fmaussion:
Sure, no pressure!

So what about leaving ``__init__`` the way it is and remove that IndexError catch?

fmaussion:
Reverted ``__init__`` back the way it was, but removed the error catch.

@shoyer feel free to close if you are reluctant to merge this, I don't mind.

I still think that it would be great to have a simple way to give custom "netcdf4 like" objects to xarray, though. If you have any suggestion on a good way to go I'd be happy to try an implementation

fmaussion:
@shoyer we should reach a decision about this small PR: I don't mind closing it as it's not a big deal...

shoyer:
I think the main question is: what is the oldest version of netCDF4 we want to support? Once we decide that, the action here should be pretty clear.

fmaussion:
I don't have a strong opinion on this. I guess that older systems are bound to a specific version of the C netCDF library, not with a specific python netCDF4 lib, right?

fmaussion:
Closing this, might reopen later if needed

Issue #1059, opened 2016-10-25
kwarg defaults (join="inner") for DataArray._binary_ops()


chunweiyuan:
Currently the default is join="inner".  However, there can be applications where the majority of binary operations require join="outer", not "inner".

Would it be advisable to place these default values in some config object the user can set at the beginning of the run script?  Or perhaps one already exists but I've failed to locate it.

shoyer:
Yes, we could make this a configurable option (currently it is not). See `xarray.set_options` for some related machinery.

chunweiyuan:
Thanks.  I will soon submit a pull request from my fork.  Which branch should it go to?  It seems that many people submit pull requests to master here.

shoyer:
We submit PRs to master. Please also be sure to update Dataset binary ops,
too.

chunweiyuan:
Yup, did that.  Thank you.  Will request to have you as a reviewer.  Much obliged.

PR #1065, opened 2016-10-27
Options to binary ops kwargs

chunweiyuan:
Currently the default is join="inner". However, there can be applications where the majority of binary operations require join="outer", not "inner".

Addresses https://github.com/pydata/xarray/issues/1059

The solution we come up with is adding a default key/value to OPTIONS, and dynamically access/update it in code (_binary_op in both dataarray.py and dataset.py)

We've also added test modules in test_dataarray.py and test_dataset.py.  All tests passed.


shoyer:
Let's find a more descriptive name for this option than merely `join`. Maybe: `compute_join`, `compute_align`, `arithmetic_join`, `math_join`, ...?

Also please add documentation notes about this new functionality to [this section of the docs](http://xarray.pydata.org/en/stable/computation.html#automatic-alignment) and the "what's new" page.

xarray/core/dataarray.py
```diff
@@ -1377,13 +1378,15 @@ def func(self, *args, **kwargs):
...
     @staticmethod
-    def _binary_op(f, reflexive=False, join='inner', **ignored_kwargs):
+    def _binary_op(f, reflexive=False, join=None, **ignored_kwargs):
         @functools.wraps(f)
         def func(self, other):
             if isinstance(other, (Dataset, groupby.GroupBy)):
                 return NotImplemented
             if hasattr(other, 'indexes'):
-                self, other = align(self, other, join=join, copy=False)
+                # if user does not specify join, default to OPTIONS['join']
```
shoyer: This comment just states what the code does, so it would be better to leave it out

xarray/core/dataarray.py
```diff
@@ -1377,13 +1378,15 @@ def func(self, *args, **kwargs):
...
+                how_to_join = join or OPTIONS['join']
```
shoyer: Use `if join is None` instead of implicitly testing `join`.

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,35 @@ def test_dot(self):
             da.dot(dm.values)
         with self.assertRaisesRegexp(ValueError, 'no shared dimensions'):
             da.dot(DataArray(1))
+
+    def test_binary_op_join_setting(self):
+        """
+        A test method to verify the ability to set binary operation join kwarg
+        ("inner", "outer", "left", "right") via xr.set_options().
+        """
+        # First we set up a data array
```
shoyer: These sort of comments (repeating the code) just add noise

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,35 @@ def test_dot(self):
...
+        # create a 3-by-3-by-3 data array
+        arr = xr.DataArray(np.arange(total_size).\
+                           reshape(len(xcoords),len(ycoords),len(zcoords)),
+                           [(xdim, xcoords), (ydim, ycoords),(zdim, zcoords)])
+        # now create a data array with the last x slice missing
+        missing_last_x = arr[0:-1,:,:].copy()
+        # create another data array with the last z slice missing
+        missing_last_z = arr[:,:,0:-1].copy()
+        # because the default in OPTIONS is join="inner", we test "outer" first
+        xr.set_options(join="outer")
```
shoyer: Use the context manager for tests instead (e.g, `with xr.set_options(join='outer'):`. It cleans itself up if an exception is raised in the middle of the test, which makes it more hygenic. Otherwise, if the next assert fails it would also trigger a bunch of other failing tests when the option doesn't get reset.

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,35 @@ def test_dot(self):
...
+        result = missing_last_x + missing_last_z
+        self.assertTrue(result.shape == arr.shape)
```
shoyer: Instead of checking specific properties, construct the full expected result (e.g., by calling `align` with `join='outer'` on the inputs) and use `assertDataArrayEqual` to verify that expect result is the result you get.

xarray/test/test_dataset.py
```diff
@@ -2916,6 +2917,44 @@ def test_filter_by_attrs(self):
...
+        self.assertTrue(result.foo[:,:,-1].isnull().all())
+        # now revert back to join="inner"
+        xr.set_options(join="inner")
+        result = missing_last_x + missing_last_z
+        self.assertTrue(result.foo.shape == \
+                        (len(xcoords)-1, len(ycoords), len(zcoords)-1))
+        self.assertTrue(result.foo.notnull().all())
+        self.assertFalse('c' in list(result.foo['x']))
+        self.assertFalse(2 in list(result.foo['z']))
+        # just for kicks, what happens when the dataarrays have different names?
```
shoyer: We already have tests for similar behavior -- see `test_dataset_math_auto_align`. This might belong there (more tests are great!), but it doesn't belong here.

It does raise the interesting question of what the behavior should be when `join='outer'` for Dataset objects with different data variables. Should the result now have the union of data variable names? If so, this will take a bit more retooling. (This would be easier after https://github.com/pydata/xarray/pull/964 is merged)

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,35 @@ def test_dot(self):
             da.dot(dm.values)
         with self.assertRaisesRegexp(ValueError, 'no shared dimensions'):
             da.dot(DataArray(1))
+
+    def test_binary_op_join_setting(self):
+        """
+        A test method to verify the ability to set binary operation join kwarg
```
shoyer: Usually we don't put docstrings on test methods

xarray/core/dataarray.py
chunweiyuan: This would trigger an UnboundLocalError:
if join is None:  join = OPTIONS['join']
because join is not defined in the local scope prior to that line.

So the choice is either
1.) if join is None:
         how_to_join = OPTIONS['join']
     else:
         how_to_join = join
or
2.) how_to_join = join or OPTIONS['join']

I opted for (2) for simplicity.  Is the style of (1) generally preferred here?

xarray/test/test_dataarray.py
chunweiyuan: Cool.  Will remove.

xarray/core/dataarray.py
shoyer: You could do the `if`/`else` inline, if you prefer:
`how_to_join = OPTIONS['join'] if join is None else join`

But yes, I do prefer explicitly checking against `None` for default values rather than relying on implicit falseness of `None`.

chunweiyuan: Funny, your suggestion is the same as my first edit to the code, but then my teammates suggested I should use the other options.

Nomenclature is not my forte.  I'll make the default
```
OPTIONS = {'display_width': 80, 'align_type': "inner"}
```
and change your suggestion to the following
```
align_type = OPTIONS['align_type'] if join is None else join
self, other = align(self, other, join=align_type, copy=False)
```
if there's no objection.

shoyer: We also align in other operations, e.g., `merge`, for which the default is to do an outer join. So the option name should be specific to computation/arithmetic.

chunweiyuan: `binary_op_align_type`?

chunweiyuan: alright, I'm using `arithmetic_join`....

shoyer: Yeah I think `arithmetic_join` or `compute_join` is probably the way to go. We will probably also use this when/if we ever add ternary operations, too.

chunweiyuan:
Out of curiosity, how do I test inner-join vs. outer-join using only 1-D arrays?  I can certainly simplify the current tests to be 2D + 1D.

xarray/test/test_dataarray.py
chunweiyuan: will do

shoyer:
Set coordinates on the 1D -- something like [0,1,2] for the first, [1,2,3]
for the second. Then the inner join should have [1,2] vs [0,1,2,3] for
outer join.

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,16 @@ def test_dot(self):
             da.dot(dm.values)
         with self.assertRaisesRegexp(ValueError, 'no shared dimensions'):
             da.dot(DataArray(1))
+
+    def test_binary_op_join_setting(self):
+        dim = 'x'
+        align_type = "outer"
+        coords_l, coords_r = [0, 1, 2], [1, 2, 3]
+        missing_0 = xr.DataArray(coords_l, [(dim, coords_l)])
```
shoyer: These names look backwards? `missing_0` currently has labels [0,1,2].

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,16 @@ def test_dot(self):
...
+        missing_3 = xr.DataArray(coords_r, [(dim, coords_r)])
+        with xr.set_options(arithmetic_join=align_type):
+            experimental = missing_0 + missing_3
+            missing_0_aligned, missing_3_aligned =\
+                xr.align(missing_0, missing_3, join=align_type)
+            control = missing_0_aligned + missing_3_aligned
```
shoyer: could also explicitly set `control = xr.DataArray([np.nan, 2, 4, np.nan], [(dim, [0, 1, 2, 3])])`, which is actually even better than using align and arithmetic here

xarray/test/test_dataarray.py
```diff
@@ -2268,3 +2268,16 @@ def test_dot(self):
...
             da.dot(DataArray(1))
+
+    def test_binary_op_join_setting(self):
+        dim = 'x'
+        align_type = "outer"
+        coords_l, coords_r = [0, 1, 2], [1, 2, 3]
+        missing_0 = xr.DataArray(coords_l, [(dim, coords_l)])
+        missing_3 = xr.DataArray(coords_r, [(dim, coords_r)])
+        with xr.set_options(arithmetic_join=align_type):
+            experimental = missing_0 + missing_3
```
shoyer: Drop out of the `set_options` block after this line.

doc/computation.rst
```diff
@@ -225,6 +225,23 @@ If the result would be empty, an error is raised instead:
     In [1]: arr[:2] + arr[2:]
     ValueError: no overlapping labels for some dimensions: ['x']
 
+However, one can explicitly change this default automatic alignment type ("inner")
+via :py:func:`~xarray.set_options()`
+
+.. ipython:: python
+
+    xr.set_options(arithmetic_join="outer")
```
shoyer: Can you only show the example using `with`?

I'm now wondering if we need add some mechanism to *require* using the context manager for `arithmetic_join`. If users don't use a context manager, very mysterious/nonlocal bugs could appear when this changes the behavior of unrelated code (e.g., in an import statement).

chunweiyuan: For our use case we would set `arithmetic_join="outer"` from the get-go and run all of our binary calculations like that throughout our simulation.  We'd use that as our default, and maybe occasionally switch to other join types.  Requiring context manager every time we deviate from "inner" seems onerous.

shoyer:
It occurs to me now that the other reason why people avoid global variables is because of concerns around thread safety and distributed computation. Even with the context manager, bad things will happen if you use `set_options` simultaneously from different threads.

I need to think about this a little more. Copious warnings in the docs might be enough, or maybe entering a `set_options` context should also enter an explicit `threading.Lock()` context.

On the other hand, the current implementation is basically identical to how functionality like NumPy's `np.errstate()` works, and people seem to be pretty happy with that.

@mrocklin does this seem like the sort of change we can do in xarray without seriously regretting it later? Xarray uses basically the same `set_options` machinery that dask has.

mrocklin:
I haven't noticed people getting caught by using `set_options` from multiple threads.  It definitely isn't threadsafe but it's usually only used when constructing computations, which is often done from a single thread.

It would be good to ensure that this affects the graph and that the actual tasks don't depend on the global state.

shoyer:
> It would be good to ensure that this affects the graph and that the actual tasks don't depend on the global state.

Indeed, we are safe here. `align` is turned into a bunch of NumPy or dask.array calls, so this option will simply effect how the graph is built.

chunweiyuan:
Just checking in.  Is there something else that needs to be added to this PR before merge?

shoyer:
My concern is how this setting currently does not change how we join variables in Dataset math. Consider the current behavior -- it always does an *inner join* of the data variables:
```
In [2]: import xarray as xr

In [3]: ds1 = xr.Dataset({'foo': 1, 'bar': 2})

In [4]: ds2 = xr.Dataset({'bar': 2, 'baz': 3})

In [5]: ds1
Out[5]:
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    *empty*
Data variables:
    bar      int64 2
    foo      int64 1

In [6]: ds2
Out[6]:
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    *empty*
Data variables:
    bar      int64 2
    baz      int64 3

In [7]: ds1 + ds2
Out[7]:
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    *empty*
Data variables:
    bar      int64 4
```

Logically, it makes sense for this option to switch this to an outer join instead.

chunweiyuan:
Alright, if that's desired, I'd just change code in `apply_over_both()` of `dataset._calculate_binary_op` to react to different kwarg values, and adding a method in `test_dataset.py`.  Should be able to push before the weekend is over.

xarray/core/options.py
```diff
@@ -26,6 +26,10 @@ class set_options(object):
     """
     def __init__(self, **kwargs):
         self.old = OPTIONS.copy()
+        for key in kwargs:
```
shoyer: please rebase or merge master -- I actually already added in something very similar
https://github.com/pydata/xarray/blob/master/xarray/core/options.py

xarray/core/dataset.py
```diff
@@ -2050,25 +2051,30 @@ def func(self, other):
...
-                    dest_vars[k] = f(lhs_vars[k], rhs_vars[k])
-                elif inplace:
-                    raise ValueError(
-                        'datasets must have the same data variables '
-                        'for in-place arithmetic operations: %s, %s'
-                        % (list(lhs_data_vars), list(rhs_data_vars)))
-                elif fillna:
-                    # this shortcuts left alignment of variables for fillna
+
+            for k in set(lhs_data_vars) & set(rhs_data_vars):
```
shoyer: It is important to maintain the order of data variables: should be `lhs` first, then `rhs` only variables.

So, probably should use a loop over `lhs_data_vars`, followed by `rhs_data_vars`.

xarray/core/dataset.py
```diff
@@ -2050,25 +2051,30 @@ def func(self, other):
...
+                dest_vars[k] = f(lhs_vars[k], rhs_vars[k])
+            if join in ["outer", "left"]:
+                for k in set(lhs_data_vars) - set(rhs_data_vars):
                     dest_vars[k] = lhs_vars[k]
```
shoyer: This was only the right default for `fillna`. For other operations, should be `NaN` instead (or, more specifically, the result of `f(lhs_vars[k], np.nan)`.

xarray/test/test_dataset.py
```diff
@@ -2918,17 +2918,38 @@ def test_filter_by_attrs(self):
...
+
+        # arithmetic join also applies to data_vars
+        ds1 = xr.Dataset({'foo': 1, 'bar': 2})
+        ds2 = xr.Dataset({'bar': 2, 'baz': 3})
+        expected = xr.Dataset({'bar': 4}) # default is inner joining
+        actual = ds1 + ds2
+        self.assertDatasetEqual(actual, expected)
+        
+        with xr.set_options(arithmetic_join='outer'):
+            expected = xr.Dataset({'foo':1, 'bar': 4, 'baz': 3})
```
shoyer: As noted above, should be `xr.Dataset({'foo': nan, 'bar': 4, 'baz': nan})` for consistent outer join semantics

xarray/core/dataset.py
```diff
@@ -2061,14 +2061,16 @@ def apply_over_both(lhs_data_vars, rhs_data_vars, lhs_vars, rhs_vars):
...
-                for k in set(lhs_data_vars) - set(rhs_data_vars):
-                    dest_vars[k] = lhs_vars[k]
-            if join in ["outer", "right"]:
-                for k in set(rhs_data_vars) - set(lhs_data_vars):
-                    dest_vars[k] = rhs_vars[k]
+            for k in lhs_data_vars:
+                if k in rhs_data_vars:
+                    dest_vars[k] = f(lhs_vars[k], rhs_vars[k])
+                elif join in ["left", "outer"]:
+                    dest_vars[k] = lhs_vars[k] if fillna else\
```
shoyer: per PEP8, use parenthesis to join across multiple lines instead of `\`

xarray/test/test_dataset.py
```diff
@@ -2935,6 +2936,40 @@ def test_filter_by_attrs(self):
...
+            self.assertDatasetEqual(actual, expected)
+
+        with xr.set_options(arithmetic_join='right'):
+            expected = xr.Dataset({'bar': 4, 'baz': np.nan})
+            actual = ds1 + ds2
+            self.assertDatasetEqual(actual, expected)
+
+
+
+
```
shoyer: too much white space here --  by convention, leave only 2 blank lines

xarray/test/test_dataarray.py
```diff
@@ -2278,3 +2278,16 @@ def test_dot(self):
...
+
+    def test_binary_op_join_setting(self):
+        dim = 'x'
+        align_type = "outer"
+        coords_l, coords_r = [0, 1, 2], [1, 2, 3]
+        missing_3 = xr.DataArray(coords_l, [(dim, coords_l)])
+        missing_0 = xr.DataArray(coords_r, [(dim, coords_r)])
+        with xr.set_options(arithmetic_join=align_type):
+            actual = missing_0 + missing_3
+        missing_0_aligned, missing_3_aligned =\
```
shoyer: line break after `xr.align(` instead of using the explicit `\`

xarray/core/options.py
```diff
@@ -1,7 +1,8 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-OPTIONS = {'display_width': 80}
+OPTIONS = {'display_width': 80,
+           'arithmetic_join': "inner"}
 
 
 class set_options(object):
```
shoyer: can you update the docstring for this class?

shoyer:
Thank you @chunweiyuan!

chunweiyuan:
woohoo!

Issue #790, opened 2016-03-11
ENH: Optional Read-Only RasterIO backend


jhamman:
RasterIO is a GDAL based library that provides _Fast and direct raster I/O for use with Numpy and SciPy_. I've just used it a bit but have been generally impressed with its support for a range of ASCII and binary raster formats.  It might be a nice addition to the suite of backends already available in `xarray`.

I'm envisioning a functionality akin to what we provide in the PyNIO backend, which is to say, read-only support for which ever file types RasterIO supports.

fmaussion:
:+1: Rasterio shines at reading georeferencing metadata out of any file, and I guess it would be no big deal to treat the various info as attributes in an xarray dataset. It is also possible to do lazy reading out of rasterio files.

(example with a geotiff file: https://github.com/fmaussion/salem/blob/master/salem/datasets.py#L263)

jhamman:
Thanks @fmaussion.  This was a helpful illustration of how this could be done.  The `salem` `GeoTiff` and `Grid` objects include all and more than I was hoping to implement in xarray. However, after a bit more looking into this, I have mixed feelings about whether this would work in `xarray`.  A brief summary of the pros/cons as I see them now.

Pros:
- Rasterio supports a wide range of raster formats (e.g. GeoTiff, ArcInfo ASCII Grid, etc.)
- Combined with `pyproj`, coordinate variables can be inferred
- Supports windowed reading (and writing), this would fit in well with the chunking approach already taken by `xarray`.
 - Supports lazy loading of array values, this would fit in well with the loading policies of the other `xarray` backends.

Cons:
- Would require `xarray` to adopt conventions for projecting arrays, naming (coordinates, dimensions, arrays), and handling of raster attributes. I can image ways this could be done but it may take us in down a path we don't want to go.
- The `xarray` backends generally return `Dataset`s, however, `rasterio` returns individual arrays that would better be applied to `DataArray`s.

fmaussion:
Hi @jhamman , I tend to agree with your doubts. I'll still comment on your cons:

To (1): I also think that xarray should avoid opening the projection can of worms. But the minimum things that xarray could do with rasterio is to read corner coordinates, dx and dy and define the two coordinates "x" and "y" out of it, without taking care of whether these are meters, degrees of arc or whatever. As long at the other rasterio file attributes are available as attribute of the `DataArray` or `DataSet` objects, user can do their own mixture

[...]

PR #1070, opened 2016-10-31
Feature/rasterio

NicWayand:
@jhamman started a backend for RasterIO that I have been working on. There are two issues I am stuck on that I could use some help:

1) Lat/long coords are not being decoded correctly (missing from output dataset). Lat/lon projection are correctly calculated and added here (https://github.com/NicWayand/xray/blob/feature/rasterio/xarray/backends/rasterio_.py#L117). But, it appears (with my limited knowledge of xarray) that the lat/long coords contained within `obj` are lost at this line (https://github.com/NicWayand/xray/blob/feature/rasterio/xarray/conventions.py#L930). 

2) Lazy-loading needs to be enabled. How can I setup/test this? Are there examples from other backends I could follow?

#790


xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,138 @@
...
+            raise ValueError('unknown dims')
+
+        self._attrs = OrderedDict()
+        for attr_name in ['crs', 'transform', 'proj']:
+            try:
+                self._attrs[attr_name] = getattr(self.ds, attr_name)
+            except AttributeError:
+                pass
+
+        self.coords = _try_to_get_latlon_coords(self.coords, self._attrs)
```
shoyer: You need to return these in the dict returned by `get_variables()`. `DataStore` objects don't have a `coords` attribute.

ci/requirements-py27-cdat+pynio.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python=2.7
-  - cdat-lite
+  # - cdat-lite
```
shoyer: This seems unrelated?

fmaussion:
This is awesome! 

For 2: I think you don't have much more to do than make ``__getitem_`` use the indexer to read the data with a window (example [here](https://github.com/fmaussion/salem/blob/master/salem/datasets.py#L341-L353))

xarray/backends/rasterio_.py
NicWayand: Thanks. This now pulls in lat/long if the projection was successful, otherwise it just returns the basic x/y indices. Should we also be returning the original projection coords? (currently they are lost).

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,139 @@
...
+    coords_out = {}
+    try:
+        import pyproj
+    except ImportError:
+        pyproj = False
+    if 'crs' in attrs and pyproj:
+        proj = pyproj.Proj(attrs['crs'])
+        x, y = np.meshgrid(coords['x'], coords['y'])
+        proj_out = pyproj.Proj("+init=EPSG:4326", preserve_units=True)
+        xc, yc = _transform_proj(proj, proj_out, x, y)
```
shoyer: You might try to do this calculation lazily, e.g., by making a `utils.NDArrayMixin` subclass in conventions.py and wrapping with `indexing.LazilyIndexedArray`.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,139 @@
...
+            raise ValueError('unknown dims')
+
+        self._attrs = OrderedDict()
+        for attr_name in ['crs', 'transform', 'proj']:
+            try:
+                self._attrs[attr_name] = getattr(self.ds, attr_name)
+            except AttributeError:
+                pass
+
+    # def get_vardata(self, var_id=1):
```
shoyer: should be removed

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,139 @@
...
+        # Get dims
+        if self.ds.count >= 1:
+            self.dims = ('band', 'y', 'x')
+            self.coords['band'] = self.ds.indexes
+        else:
+            raise ValueError('unknown dims')
+
+        self._attrs = OrderedDict()
+        for attr_name in ['crs', 'transform', 'proj']:
+            try:
```
shoyer: use `with pycompat.suppress(AttributeError):`

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            return copy.deepcopy(x), copy.deepcopy(y)
+
+    return pyproj.transform(p1, p2, x, y)
+
+
+def _try_to_get_latlon_coords(coords, attrs):
+    coords_out = {}
+    try:
+        import pyproj
```
shoyer: I'm a little surprised rasterio doesn't have projections built in.

ci/requirements-py35.yml
```diff
@@ -11,6 +11,7 @@ dependencies:
   - pandas
   - seaborn
   - scipy
+  - rasterio
```
shoyer: Unless rasterio only supports Python 3, should add it to other test suites, too (especially Python 2.7).

NicWayand:
Tested open_mfdataset() on 100+ geotiffs and lazyloading with rasterio does appear to be working.

ci/requirements-py35.yml
NicWayand: Rasterio supports 2.7 and 3.3-3.5 (https://github.com/mapbox/rasterio)

I added this requirement to all other test suites, was that correct? Version 1.0 is needed as many function names changed in rasterio at V1.0.

NicWayand:
Travis-ci fail is because it can't find rasterio, which comes through the conda-forge channel (https://github.com/conda-forge/rasterio-feedstock). I think it needs to be added as described here (http://conda.pydata.org/docs/travis.html#additional-steps). But I am new to Travis-ci, so don't want to mess up the current .travis.yml file.

shoyer:
Add `conda-forge` it to the `channels` section of conda requirements yml files:
https://github.com/pydata/xarray/blob/f40d323c76f6cab9615567e4ba7afde7860f7775/ci/requirements-py27-cdat%2Bpynio.yml#L3

NicWayand:
Any idea why Segmentation faults occur for 3.4 and 4.5?

xarray/backends/rasterio_.py
NicWayand: They do: https://mapbox.github.io/rasterio/topics/reproject.html

I will take a stab at doing it the rasterio way.

xarray/backends/rasterio_.py
NicWayand: Sorry I don't understand how to do this. Is there an example I can go off of?

shoyer:
I would start by adding `-v` to our invocation of `py.test` in `.travis.yml`: https://github.com/pydata/xarray/blob/a247d9314d6a8df34df887106e234bbffebe1e48/.travis.yml#L72

xarray/backends/rasterio_.py
shoyer: We have some examples of making NDArrayMixin subclasses in conventions.py, e.g.,
https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L311

shoyer: Another option would be to use `dask.array` to build the coordinate arrays, though the `NDArrayMixin` would avoid the need for dask.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,120 @@
...
+        self.array = ds.read()
+
+    @property
+    def dtype(self):
+        return np.dtype(self._ds.dtypes[0])
+
+    def __getitem__(self, key):
+        if key == () and self.ndim == 0:
+            return self.array.get_value()
+        return self.array[key]
```
jhamman: Based on what @fmaussion said above, I think this should be something like:

```Python
    def __getitem__(self, key):
        if key == () and self.ndim == 0:
            return self._ds.read()
        return self._ds.read(band, window=window)
```

Where `band` and `window` describe the slice of data to be read. Off the top of my head, I'm not exactly sure how to parse the key here though.

xarray/backends/rasterio_.py
shoyer: That's a reprojection, which does seem useful, but it seems separate from giving you `lat` and `lon` coordinates

xarray/backends/rasterio_.py
shoyer: By the time indexers get here, they should have already passed through [canonicalize_indexer](https://github.com/pydata/xarray/blob/a247d9314d6a8df34df887106e234bbffebe1e48/xarray/core/indexing.py#L41), which means you only need to handle a `key` that is a tuple of the appropriate number of dimensions (i.e., 3) composed of integers, slices and integer ndarrays.

Based on the [docstring for read](https://github.com/mapbox/rasterio/blob/a5376390829be52490a1fe6da52395262d0f6167/rasterio/_io.pyx#L234-L287), we want something like:

```python
def __getitem__(self, key):
    band = range(self.shape[0])[key[0]]
    window = []
    for win in key[1:]:
        if instance(win, slice):
            window.append((win.start, win.stop))
        elif isinstance(win, int):
            window.append((win, win + 1))
        else:  # integer ndarray
            window.append((win.min(), win.max()))
    raw_data = self._ds.read(band, window=window)
    # now, fix up raw_data to conform to numpy indexing conventions
    # - drop axes for integer band/windows
    # - stride if windows are slices with win.step != 1
    # - subset if window is an integer ndarray
```

Honestly, this logic should probably live in rasterio if possible. I'm a little surprised that they have never implemented a `__getitem__` method.

shoyer: See also: https://gist.github.com/lpinner/bd57b54a5c6903e4a6a2 (can't reuse this directly, though, because it doesn't have a license).

Anyways, I would definitely see if the rasterio folks are up for implementing a `__getitem__` method. There's no reason why this is xarray specific, to be honest.

jhamman: See also https://github.com/mapbox/rasterio/issues/920 for related discussion.

ci/requirements-py27-min.yml
```diff
@@ -1,9 +1,12 @@
 name: test_env
+channels:
+  - conda-forge
 dependencies:
   - python=2.7
   - pytest
   - numpy==1.9.3
   - pandas==0.15.0
+  - rasterio
```
jhamman: I think we can roll back these changes.  This test setup is just for required dependencies which `rasterio` will not be.

ci/requirements-py27-netcdf4-dev.yml
```diff
@@ -8,6 +10,7 @@ dependencies:
   - numpy
   - pandas
   - scipy
+  - rasterio
```
jhamman: I think we can also roll back these changes as well

ci/requirements-py33.yml
```diff
@@ -1,8 +1,11 @@
 name: test_env
+channels:
+  - conda-forge
 dependencies:
   - python=3.3
   - pytest
   - pandas
+  - rasterio
```
jhamman: roll back these changes

ci/requirements-py27-pydap.yml
```diff
@@ -8,6 +10,7 @@ dependencies:
   - numpy
   - pandas
   - scipy
+  - rasterio
```
jhamman: I don't think we need to test rasterio along with pydap so we can roll back these changes.

ci/requirements-py35-dask-dev.yml
```diff
@@ -7,6 +9,7 @@ dependencies:
   - pandas
   - scipy
   - toolz
+  - rasterio
```
jhamman: roll back these changes

ci/requirements-py35-pandas-dev.yml
```diff
@@ -7,6 +9,7 @@ dependencies:
   - netcdf4=1.1.9
   - scipy=0.16.0
   - toolz
+  - rasterio
```
jhamman: roll back these changes

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,120 @@
...
+
+from .common import AbstractDataStore
+
+__rio_varname__ = 'raster'
+
+
+class RasterioArrayWrapper(NDArrayMixin):
+    def __init__(self, ds):
+        self._ds = ds
+        self.array = ds.read()
```
jhamman: @shoyer, do we need to maintain the `array` attribute here? Would it make more sense to just populate the `_ds` attribute and set the array in `__getitem__`?

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,120 @@
...
+        if self.ds.count >= 1:
+            self.dims = ('band', 'y', 'x')
+            self.coords['band'] = self.ds.indexes
+        else:
+            raise ValueError('unknown dims')
+
+        self._attrs = OrderedDict()
+        with suppress(AttributeError):
+            for attr_name in ['crs', 'transform', 'proj']:
+                    self._attrs[attr_name] = getattr(self.ds, attr_name)
```
jhamman: I think there is some extra indentation here -- should be 4 spaces.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,120 @@
...
+        data = indexing.LazilyIndexedArray(
+            RasterioArrayWrapper(self.ds))
+        return Variable(self.dims, data, self._attrs)
+
+    def get_variables(self):
+        # Get lat lon coordinates
+        coords = _try_to_get_latlon_coords(self.coords, self._attrs)
+        vars = {__rio_varname__: self.open_store_variable(__rio_varname__)}
+        vars.update(coords)
+        return FrozenOrderedDict(vars)
```
jhamman: we should use another name other than `vars` since it is a python builtin.

xarray/test/test_backends.py
```diff
@@ -1063,6 +1063,35 @@ def test_weakrefs(self):
...
+    def test_weakrefs(self):
+        example = Dataset({'foo': ('x', np.arange(5.0))})
+        expected = example.rename({'foo': 'bar', 'x': 'y'})
+
+        with create_tmp_file() as tmp_file:
+            example.to_netcdf(tmp_file, engine='scipy')
+            on_disk = open_dataset(tmp_file, engine='rasterio')
+            actual = on_disk.rename({'foo': 'bar', 'x': 'y'})
+            del on_disk  # trigger garbage collection
+            self.assertDatasetIdentical(actual, expected)
```
jhamman: @NicWayand - take a look at the PyNio tests.  They are probably the closest analog for what we need to test here.

xarray/backends/rasterio_.py
jhamman: @NicWayand -

I think you can basically drop this in replacing the current `RasterioArrayWrapper`:

```Python
class RasterioArrayWrapper(NdimSizeLenMixin):
    """Mixin class RasterIO datasets for making wrappers of N-dimensional 
    arrays that conform to the ndarray interface required for the data
    argument to Variable objects.
    A subclass should set the `array` property and override one or more of
    `dtype`, `shape` and `__getitem__`.
    """
    def __init__(self, ds):
        self.ds = ds
    
    @property
    def dtype(self):
        if len(set(self.ds.dtypes[0])) != 1:
            raise ValueError(
                'Can only handle Rastio dataset with all bands having the same type')
        return np.dtype(self.ds.dtypes[0])

    @property
    def shape(self):
        return self.ds.shape

    def __array__(self, dtype=None):
        '''Not sure if this will work as is'''
        return np.asarray(self[...], dtype=dtype)

    def __getitem__(self, key):
        band = range(self.shape[0])[key[0]]
        window = []
        for win in key[1:]:
            if instance(win, slice):
                window.append((win.start, win.stop))
            elif isinstance(win, int):
                window.append((win, win + 1))
            else:  # integer ndarray
                window.append((win.min(), win.max()))
        raw_data = self.ds.read(band, window=window)
        # now, fix up raw_data to conform to numpy indexing conventions
        # - drop axes for integer band/windows
        # - stride if windows are slices with win.step != 1
        # - subset if window is an integer ndarray

    def __repr__(self):
        return '%s(array=%r)' % (type(self).__name__, self.ds)
```

I'm not sure the `__array__` will work as it is until we get the `__get_item__` fully functional.

shoyer: I would suggest getting something working here, with an eye toward keeping it self-contained. Later, we can try to port it upstream to rasterio.

fmaussion:
Hi, unless there are objections I'd like to try to finish this PR. I just forked @NicWayand 's branch and started to look at it, and I hope to get something done during next week or so...

NicWayand:
Hi @fmaussion, no objections here. I got it working just barely for my project, and won't have time in the near future to devote to wrap this up.

byersiiasa:
@gidden you might be interested in this

gidden:
@byersiiasa indeed I am, in fact I'm reviewing #1260 which overtook this one =)

fmaussion:
Closing this in favor of https://github.com/pydata/xarray/pull/1260

Issue #921, opened 2016-07-29
Inconsistency between the types of  Dataset.dims and DataArray.dims


shoyer:
`DataArray.dims` is currently a tuple, whereas `Dataset.dims` is a dict. This results in ugly code like this, taken from `xarray/core/group.py`:

``` python
        try:  # Dataset
            expected_size = obj.dims[group_dim]
        except TypeError:  # DataArray
            expected_size = obj.shape[obj.get_axis_num(group_dim)]
```

One way to resolve this inconsistency would be switch `DataArray.dims` to a (frozen) `OrderedDict`. The downside is that code like `x.dims[0]` wouldn't work anymore (unless we add some terrible fallback logic). You'd have to write `x.dims.keys()[0]`, which is pretty ugly. On the plus side, `x.dims['time']` would always return the size of the time dimension, regardless of whether `x` is a DataArray or Dataset.

Another option would be to add an attribute `dim_shape` (or some other name) that serves as an alias to `dims` on Dataset and an alias to `OrderedDict(zip(dims, shape))` on DataArray. This would be fully backwards compatible, but further pollute the namespace on Dataset and DataArray.

MaximilianR:
What are the arguments against having `dims` be a tuple for both `DataArray` & `Dataset`s?

shoyer:
> What are the arguments against having dims be a tuple for both DataArray & Datasets?

I guess that's a third option! In that case, we'd have to add `shape` to Dataset, too, to ensure there is some way to get size information.

The main downside is that we'd still be stuck without any idea way to get dimension sizes by name -- `ds.shape[ds.get_axis_num('time')]` is pretty bad. Also, I'm not sure how many use cases there are for actually pulling out the first dimension of a dataset (or it's shape) by position, given that dimension order is currently sorted independently of dimension order on any of its data arrays.

MaximilianR:
I frequently need to pull out the dimension names, and I've never needed to pull out their sizes. I realize that the internals need to do that though. And maybe my use case is specific.

Zooming out a bit: the access pattern for most of these (variables, coords) is indeed an `OrderedDict` of name to object. But the stickler for dim is why the size should be the value here? It seems very different from having dim objects as values.

I could certainly understand a convenience `dims_shape` (for `{dim: ds[dim].shape for dim in ds.dims}`.

But I"m coming from a place of less context without that much confidence...

shoyer:
> I frequently need to pull out the dimension names, and I've never needed to pull out their sizes.

[...]

PR #1076, opened 2016-11-03
New properties `Dataset.sizes` and `DataArray.sizes`

shoyer:
This allows for consistent access to dimension lengths on ``Dataset`` and
``DataArray``

xref #921 (doesn't resolve it 100%, but should help significantly)

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1,9 +1,10 @@
 import numpy as np
 import pandas as pd
 
-from .pycompat import basestring, iteritems, suppress, dask_array_type, bytes_type
+from .pycompat import (basestring, iteritems, suppress, dask_array_type,
+                       OrderedDict)
 from . import formatting
-from .utils import SortedKeysDict, not_implemented
+from .utils import SortedKeysDict, not_implemented, Frozen
 
 
 class ImplementsArrayReduce(object):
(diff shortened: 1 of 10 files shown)


xarray/core/common.py
```diff
@@ -335,6 +335,36 @@ def pipe(self, func, *args, **kwargs):
...
+        squeezed : same type as caller
+            This object, but with with all or a subset of the dimensions of
+            length 1 removed.
+
+        See Also
+        --------
+        numpy.squeeze
+        """
+        if dim is None:
+            dim = [d for d, s in self.sizes.item() if s == 1]
```
MaximilianR: I think `items`?

xarray/core/dataarray.py
```diff
@@ -110,6 +111,31 @@ def __setitem__(self, key, value):
         self.data_array[pos_indexers] = value
 
 
+class DataArraySizes(collections.Mapping, formatting.ReprMixin):
```
MaximilianR: Does this need a separate class that holds the array? 
Could it be a `dict` of dim-name to dim-size?

shoyer: Probably not -- could just generate an object on the fly, e.g., `Frozen(OrderedDict(zip(self.dims, self.shape)))`. Let me do a quick micro benchmark to see if that has appreciably different performance.

xarray/core/common.py
shoyer: Oops indeed

shoyer:
Updated to remove the custom class. It makes accessing a single dimension length from `DataArray.sizes` slightly slower, but within the range of usual Python overhead (2.5 vs. 5 us). Could add a cache at some later point if necessary.

PR #1079, opened 2016-11-03
New infer_intervals keyword for pcolormesh

fmaussion:
Addresses https://github.com/pydata/xarray/issues/781

@jhamman what do you think? I'm not sure if ``infer_interval_breaks`` is the best name for the keyword but I didn't come up with anything better than that. (maybe ``infer_coord_intervals``?)

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -316,6 +316,10 @@ def _plot2d(plotfunc):
         provided, extend is inferred from vmin, vmax and the data limits.
     levels : int or list-like object, optional
         Split the colormap (cmap) into discrete color intervals.
+    infer_interval_breaks : bool, optional
+        Only applies to pcolormesh. If True, the coordinate intervals are
+        passed to pcolormesh (the default). If False, the original coordinates
+        are used (this can be useful for certain map projections).
     subplot_kws : dict, optional
         Dictionary of keyword arguments for matplotlib subplots. Only applies
         to FacetGrid plotting.
@@ -341,8 +345,9 @@ def _plot2d(plotfunc):
                     col_wrap=None, xincrease=True, yincrease=True,
                     add_colorbar=None, add_labels=True, vmin=None, vmax=None,
                     cmap=None, center=None, robust=False, extend=None,
-                    levels=None, colors=None, subplot_kws=None,
-                    cbar_ax=None, cbar_kwargs=None, **kwargs):
+                    levels=None, infer_interval_breaks=True, colors=None,
+                    subplot_kws=None, cbar_ax=None, cbar_kwargs=None,
+                    **kwargs):
         # All 2d plots in xarray share this function signature.
         # Method signature below should be consistent.
 
@@ -456,8 +461,8 @@ def _plot2d(plotfunc):
                    col=None, col_wrap=None, xincrease=True, yincrease=True,
                    add_colorbar=None, add_labels=True, vmin=None, vmax=None,
                    cmap=None, colors=None, center=None, robust=False,
-                   extend=None, levels=None, subplot_kws=None,
-                   cbar_ax=None, cbar_kwargs=None, **kwargs):
+                   extend=None, levels=None, infer_interval_breaks=True,
+                   subplot_kws=None, cbar_ax=None, cbar_kwargs=None, **kwargs):
         """
         The method should have the same signature as the function.
 
@@ -555,14 +560,14 @@ def _infer_interval_breaks(coord):
 
 
 @_plot2d
-def pcolormesh(x, y, z, ax, **kwargs):
+def pcolormesh(x, y, z, ax, infer_interval_breaks=True, **kwargs):
     """
     Pseudocolor plot of 2d DataArray
 
     Wraps matplotlib.pyplot.pcolormesh
     """
 
-    if not hasattr(ax, 'projection'):
+    if infer_interval_breaks:
         x = _infer_interval_breaks(x)
         y = _infer_interval_breaks(y)
 
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -318,8 +318,10 @@ def _plot2d(plotfunc):
         Split the colormap (cmap) into discrete color intervals.
     infer_intervals : bool, optional
         Only applies to pcolormesh. If True, the coordinate intervals are
-        passed to pcolormesh (the default). If False, the original coordinates
-        are used (this can be useful for certain map projections).
+        passed to pcolormesh. If False, the original coordinates are used
+        (this can be useful for certain map projections). The default is to
+        always infer intervals, unless the mesh is irregular and plotted on
+        a map projection.
     subplot_kws : dict, optional
         Dictionary of keyword arguments for matplotlib subplots. Only applies
         to FacetGrid plotting.
@@ -345,7 +347,7 @@ def _plot2d(plotfunc):
                     col_wrap=None, xincrease=True, yincrease=True,
                     add_colorbar=None, add_labels=True, vmin=None, vmax=None,
                     cmap=None, center=None, robust=False, extend=None,
-                    levels=None, infer_intervals=True, colors=None,
+                    levels=None, infer_intervals=None, colors=None,
                     subplot_kws=None, cbar_ax=None, cbar_kwargs=None,
                     **kwargs):
         # All 2d plots in xarray share this function signature.
@@ -464,7 +466,7 @@ def _plot2d(plotfunc):
                    col=None, col_wrap=None, xincrease=True, yincrease=True,
                    add_colorbar=None, add_labels=True, vmin=None, vmax=None,
                    cmap=None, colors=None, center=None, robust=False,
-                   extend=None, levels=None, infer_intervals=True,
+                   extend=None, levels=None, infer_intervals=None,
                    subplot_kws=None, cbar_ax=None, cbar_kwargs=None, **kwargs):
         """
         The method should have the same signature as the function.
@@ -550,29 +552,56 @@ def contourf(x, y, z, ax, **kwargs):
     return ax, primitive
 
 
-def _infer_interval_breaks(coord):
+def _infer_interval_breaks(coord, axis=0):
     """
     >>> _infer_interval_breaks(np.arange(5))
     array([-0.5,  0.5,  1.5,  2.5,  3.5,  4.5])
+    >>> _infer_interval_breaks([[0, 1], [0, 1]], axis=1)
+    array([[-0.5,  0.5,  1.5],
+           [-0.5,  0.5,  1.5]])
     """
     coord = np.asarray(coord)
-    deltas = 0.5 * (coord[1:] - coord[:-1])
-    first = coord[0] - deltas[0]
-    last = coord[-1] + deltas[-1]
-    return np.r_[[first], coord[:-1] + deltas, [last]]
+    if axis == 0:
+        deltas = 0.5 * (coord[1:, ...] - coord[:-1, ...])
+        first = coord[0, ...] - deltas[0, ...]
+        last = coord[-1, ...] + deltas[-1, ...]
+        return np.r_[[first], coord[:-1, ...] + deltas, [last]]
+    elif axis == 1:
+        deltas = 0.5 * (coord[..., 1:] - coord[..., :-1])
+        first = coord[..., [0]] - deltas[..., [0]]
+        last = coord[..., [-1]] + deltas[..., [-1]]
+        return np.c_[first, coord[..., :-1] + deltas, last]
 
 
 @_plot2d
-def pcolormesh(x, y, z, ax, infer_intervals=True, **kwargs):
+def pcolormesh(x, y, z, ax, infer_intervals=None, **kwargs):
     """
     Pseudocolor plot of 2d DataArray
 
     Wraps matplotlib.pyplot.pcolormesh
     """
 
+    # decide on a default for infer_intervals (GH781)
+    x = np.asarray(x)
+    if infer_intervals is None:
+        if hasattr(ax, 'projection'):
+            if len(x.shape) == 1:
+                infer_intervals = True
+            else:
+                infer_intervals = False
+        else:
+            infer_intervals = True
+
     if infer_intervals:
-        x = _infer_interval_breaks(x)
-        y = _infer_interval_breaks(y)
+        if len(x.shape) == 1:
+            x = _infer_interval_breaks(x)
+            y = _infer_interval_breaks(y)
+        else:
+            # we have to infer the intervals on both axes
+            x = _infer_interval_breaks(x, axis=1)
+            x = _infer_interval_breaks(x, axis=0)
+            y = _infer_interval_breaks(y, axis=1)
+            y = _infer_interval_breaks(y, axis=0)
 
     primitive = ax.pcolormesh(x, y, z, **kwargs)
 
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -556,21 +556,17 @@ def _infer_interval_breaks(coord, axis=0):
     """
     >>> _infer_interval_breaks(np.arange(5))
     array([-0.5,  0.5,  1.5,  2.5,  3.5,  4.5])
-    >>> _infer_interval_breaks([[0, 1], [0, 1]], axis=1)
+    >>> _infer_interval_breaks([[0, 1], [3, 4]], axis=1)
     array([[-0.5,  0.5,  1.5],
-           [-0.5,  0.5,  1.5]])
+           [ 2.5,  3.5,  4.5]])
     """
     coord = np.asarray(coord)
-    if axis == 0:
-        deltas = 0.5 * (coord[1:, ...] - coord[:-1, ...])
-        first = coord[0, ...] - deltas[0, ...]
-        last = coord[-1, ...] + deltas[-1, ...]
-        return np.r_[[first], coord[:-1, ...] + deltas, [last]]
-    elif axis == 1:
-        deltas = 0.5 * (coord[..., 1:] - coord[..., :-1])
-        first = coord[..., [0]] - deltas[..., [0]]
-        last = coord[..., [-1]] + deltas[..., [-1]]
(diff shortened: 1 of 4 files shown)


shoyer:
I would call this just `infer_intervals`. `breaks` in the helper function name is referring to an implementation detail of how we store the intervals (by using breaks).

ocefpaf:
Is `infer_intervals` creating coordinate bounds for plotting? It looks like it. What about when we do have cell boundaries in the dataset? (See http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-boundaries). Will those be ignored and the inferred one used?

(Sorry if I am only making noise here and this does not make sense. In that case just ignore my comment.)

fmaussion:
> Is infer_intervals creating coordinate bounds for plotting? It looks like it.

Yes, it is a special case for pcolormesh though

> What about when we do have cell boundaries in the dataset? 

Currently the entire xarray workflow is built upon the implicit assumption that the coordinates are always at the grid-point center ( @shoyer correct me if I'm wrong). I'm not sure if the additional complexity added by cell boundaries is on the xarray devs priority list...

ocefpaf:
> Yes, it is a special case for pcolormesh though

A special case that I love. In the old Matlab day we had horrible hacks to get pcolor to work properly :smile: 

>  I'm not sure if the additional complexity added by cell boundaries is on the xarray devs priority list...

That is fine. Maybe a warning in the docs would be nice. So people can revert to a "manual" plotting to get the boundaries right.

fmaussion:
> In the old Matlab day we had horrible hacks to get pcolor to work properly

pcolormesh and imshow are indeed very consistent in xarray:

```python
import numpy as np
import cartopy.crs as ccrs
import xarray as xr

da = xr.DataArray(np.arange(20).reshape(4, 5), dims=['lat', 'lon'],
                  coords = {'lat': np.linspace(0, 30, 4),
                            'lon': np.linspace(-20, 20, 5)})

f = plt.figure(figsize=(5, 8))

ax = plt.subplot(3, 1, 1, projection=ccrs.PlateCarree())
da.plot.imshow(ax=ax, transform=ccrs.PlateCarree())
lon, lat = np.meshgrid(da.lon.values, da.lat.values)
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('imshow')

ax = plt.subplot(3, 1, 2, projection=ccrs.PlateCarree())
da.plot.pcolormesh(ax=ax, transform=ccrs.PlateCarree())
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('pcolormesh')

ax = plt.subplot(3, 1, 3, projection=ccrs.PlateCarree())
da.plot.pcolormesh(ax=ax, transform=ccrs.PlateCarree(), infer_intervals=False)
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('pcolormesh - no intervals')
```

figure_1

In most cases setting ``infer_intervals=False`` won't be a good idea, but maybe for some exotic projections (at the pole in @jhamman example) it might be easier this way (albeit a bit "wrong").

shoyer:
We definitely ignore cell boundaries -- they don't (yet) have any place in the xarray data model.

It would be nice to have a built in "interval" data type, along with an `IntervalIndex`, that we could use for storing and doing lookups along interval axes. At one point I was working on this for pandas (see https://github.com/pandas-dev/pandas/issues/7640), but then abandoned my effort when it seemed like it would be too much work to get it finished.

ocefpaf:
> We definitely ignore cell boundaries -- they don't (yet) have any place in the xarray data model.

Even though I would love to have that functionality I do not believe it is high priority. `cell boundaries` is a corner case, at least in my field of work, and 99% of the time it is OK to infer the intervals. Also, the name of the key word is quite clear: `infer_intervals`.

Maybe you could only add a note in the docs mentioning that the cell boundaries might exist?

xarray/plot/plot.py
```diff
@@ -555,14 +563,14 @@ def _infer_interval_breaks(coord):
...
-def pcolormesh(x, y, z, ax, **kwargs):
+def pcolormesh(x, y, z, ax, infer_intervals=True, **kwargs):
     """
     Pseudocolor plot of 2d DataArray
 
     Wraps matplotlib.pyplot.pcolormesh
     """
 
-    if not hasattr(ax, 'projection'):
+    if infer_intervals:
```
jhamman: I have mixed feelings about this change. I'm not sure I like `infer_intervals` to default to `True`. This is going to break a bunch of our plotting code. My preference, and I'm open to other ideas at this point, would be to default to `infer_intervals=None` with something like this logic:

```Python
if (infer_intervals is None) and (hasattr(ax, 'projection'))
    infer_intervals = False
    warnings.warn("infer_intervals=None is deprecated, please specify either True or False",
                  DeprecationWarning)
if infer_intervals:
    ...
```

fmaussion: I'm not sure to understand... ``infer_intervals=True`` was the *de-facto* default until V0.7.2, and from the example above it is clear that setting it to ``False`` is almost never what you want to do. Can you provide an example where this is a good thing?

I agree your concerns about breaking old code though. How about keeping the current behavior for one more version, but issuing a warning that in the next version (probably 0.9.1 or 0.10), it is going to be set to True per default? 

For my purposes this is going to imply that I'm going to use ``imshow()`` and not ``pcolormesh`` anymore for the time being.

jhamman: Setting `infer_intervals=False` is exactly what you want if you have irregularly spaced coordinates (see example here: https://github.com/pydata/xarray/issues/781#issuecomment-191595449).

I think something like the logic I showed above may be a good compromise.  I'm happy to transition to always specifying `infer_intervals=False` we should be more transparent about the change (hence the warning along with a None for a default value.  Would you be happier with this logic?

```Python
if (infer_intervals is None):
    if (hasattr(ax, 'projection'))
        infer_intervals = False
        warnings.warn("infer_intervals=None is deprecated, please specify either True or False",
                      DeprecationWarning)
    else:
        infer_intervals = True
if infer_intervals:
    ...
```

fmaussion: OK, with irregular coordinates none of the solutions is satisfying. ``infer_intervals=True`` crops one column, while ``infer_intervals=False`` crop one column and one line of data:

```python
import numpy as np
import cartopy.crs as ccrs
import xarray as xr

lon, lat = np.linspace(-20, 20, 5), np.linspace(0, 30, 4)
lon, lat = np.meshgrid(lon, lat)

da = xr.DataArray(np.arange(20).reshape(4, 5), dims=['y', 'x'],
                  coords = {'lat': (('y', 'x'), lat),
                            'lon': (('y', 'x'), lon)})

f = plt.figure()

ax = plt.subplot(2, 1, 1, projection=ccrs.PlateCarree())
da.plot.pcolormesh('lon', 'lat', ax=ax, transform=ccrs.PlateCarree(), infer_intervals=True)
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('pcolormesh')

ax = plt.subplot(2, 1, 2, projection=ccrs.PlateCarree())
da.plot.pcolormesh('lon', 'lat', ax=ax, transform=ccrs.PlateCarree(), infer_intervals=False)
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('pcolormesh - no intervals')

plt.show()
```
 
figure_1-1

I'll try to have a closer look later next week

fmaussion:
After giving this a few more thoughts:

1. if the coordinates are 1D (i.e. the grid is regular in the scipy sense: the coordinates don't have to be evenly spaced), ``infer_intervals`` should be set to ``True`` by default. This should make everybody happy (including @jhamman and me), and this will cover a huge majority of the geoscientific grids.

2. if the coordinates are 2D, what xarray was doing until now was wrong. ``_infer_interval_breaks`` was inferring intervals on the first axis only and ignoring the second, so regardless of whether we were plotting on a projection or not, the plot output was cropped (see example above)

So what do we do with this? @jhamman 's suggestion is to set  ``infer_intervals`` to ``False``, which is equivalent to saying: "this plot is going to be a bit wrong but this is not xarray's problem". I guess that it's fine to do so, but we'd have to be happy with the following:

```python
import numpy as np
import cartopy.crs as ccrs
import xarray as xr

lon, lat = np.linspace(-20, 20, 5), np.linspace(0, 30, 4)
lon, lat = np.meshgrid(lon, lat)
lon += lat/10
lat += lon/10
da = xr.DataArray(np.arange(20).reshape(4, 5), dims=['y', 'x'],
                  coords = {'lat': (('y', 'x'), lat),
                            'lon': (('y', 'x'), lon)})

ax = plt.subplot(1, 2, 1, projection=ccrs.PlateCarree())
da.plot.pcolormesh('lon', 'lat', ax=ax, transform=ccrs.PlateCarree(), infer_intervals=False)
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('2d irreg case - pcolor')

ax = plt.subplot(1, 2, 2, projection=ccrs.PlateCarree())
da.plot.contourf('lon', 'lat', ax=ax, transform=ccrs.PlateCarree())
ax.scatter(lon, lat, transform=ccrs.PlateCarree())
ax.coastlines()
ax.set_title('2d irreg case - contourf')
```

figure_1-3

Note that one line and one column of data disappeared (so that the colorbar range doesn't correspond to what we see) and the coordinates are not understood as being in the cell center. The advantage of this choice are twofold:
- it's kind of coherent with how ``contourf`` would interpret the coordinates (see example above)
- if we agree on "1D = infer per default" and "2D = don't infer per default", we might even get rid of the new keyword altogether. 

Let's call this option "Option A". 

I've been playing around with an "Option B", which includes making ``_infer_interval_breaks`` accept an ``axis`` kwarg and which includes a little bit of 2d interpolation (scipy). Regardless of the implementation (which could be meliorated), we could reach the following result:

figure_1-4

This plot is closer to what I think the grid actually represents, but I am biased towards my usage of xarray (gridded geoscientific models). The disadvantages of option B are: 
- it adds a bit of complexity
- if you think about it, the problem of plotting irregular grids is not really an xarray problem. If there was a perfect way to handle an irregular grids without having bounded coordinates, I guess that matplotlib would have a tool for it.

I currently tend to go for option A, since option B is a bit awkward and users who have bounded coordinated available somewhere (as suggested by @ocefpaf ) will have to implement their own plotting routine anyway.

fmaussion:
I've come up with a cleaner implementation which doesn't require scipy and produces a very accurate plot:
figure_1

I'm going to clean all this later today and update my PR, I think that we can come to something nice.

@jhamman : I wonder if the problem you are having in your [example](https://github.com/pydata/xarray/issues/781#issuecomment-191595449) is related to flaoting point issues at the pole. Would you mind sending me the file and the code that produces the plot so that I can have a look at it?

jhamman:
You can get data on the same grid (different variable, same coords) from the sample data `ds = xr.tutorial.load_dataset('rasm').isel(time=0)`.  From there, my example should work.

fmaussion:
@jhamman @shoyer this is ready for another review (the test failure seems unrelated).

@ocefpaf I added a note to the docs

@jhamman OK now I get why it clearly doesn't make sense to infer anything in your dataset (plot below). Just out of curiosity: does RASM have a map projection on which it has regular coordinates, too? 

figure_1-1

fmaussion:
Actually, you wouldn't like to infer intervals on these coordinates at all, regardless if you are on a map or not. Should I remove the ``if hasattr(ax, 'projection')`` altogether?

fmaussion:
> Should I remove the if hasattr(ax, 'projection') altogether?

Finally I think it's ok to keep it as it is now, since the data above is likely to always be plotted on a map anyways.

rabernat:
This looks like a really nice PR.

There is some intersection with my "multidimensional coordinates" tutorial:
http://xarray.pydata.org/en/stable/examples/multidimensional-coords.html

At the least, we should standardize the terminology in the docs. Are these "irregular coordinates" or "multidimensional coordinates"? I don't particularly care, but we should be consistent.

fmaussion:
@rabernat thanks for the link! I'll update the docs with a link to your example and with "multidimensional" instead of "irregular".

fmaussion:
@rabernat : couldn't your tutorial also use the ipython directive and run at build time instead of the using the parsed literals? I see that the other examples are also static, is there a reason for this?

rabernat:
> is there a reason for this?

My laziness / ignorance of the documentation build process.

fmaussion:
OK. Let's see what the gurus say about this, maybe there is a reason to keep it like this

shoyer:
Yes, could certainly run this at doc build time instead of using a notebook. The advantage of the notebook is that it's easier to edit/assemble, and also easier to download and adjust.

In an ideal world, I think we would check in a notebook with cleared output and then run it and convert to HTML at doc build time. Looks like that's exactly what [nbsphinx](https://github.com/spatialaudio/nbsphinx) does if anyone wants to take a go at setting that up. Could be a much nicer way to write the docs than RST.

fmaussion:
Note that in that case, [multidimensional-coords](https://raw.githubusercontent.com/pydata/xarray/master/doc/examples/multidimensional-coords.rst) and [monthly-means](https://github.com/pydata/xarray/edit/master/doc/examples/monthly-means.rst) are already formatted in rst.

shoyer:
Right, we used a static ipynb -> rst converter

xarray/plot/plot.py
```diff
@@ -542,29 +552,56 @@ def contourf(x, y, z, ax, **kwargs):
...
+    array([[-0.5,  0.5,  1.5],
+           [-0.5,  0.5,  1.5]])
     """
     coord = np.asarray(coord)
-    deltas = 0.5 * (coord[1:] - coord[:-1])
-    first = coord[0] - deltas[0]
-    last = coord[-1] + deltas[-1]
-    return np.r_[[first], coord[:-1] + deltas, [last]]
+    if axis == 0:
+        deltas = 0.5 * (coord[1:, ...] - coord[:-1, ...])
```
shoyer: You could write an axis agnostic version of this with `np.diff`, `np.take` and `np.concatenate`, e.g.,
```
deltas = 0.5 * np.diff(coord, axis)
first = np.take(coord, [0], axis)
last = np.take(coord, [-1], axis)
trim_last = tuple(slice(None, -1) if n == axis else slice(None)
                  for n in range(coord.ndim))
return np.concatenate([first, coord[trim_last] + deltas, last], axis)
```

Maybe a slight improvement? (Up to you on whether to use it)

xarray/plot/plot.py
```diff
@@ -542,29 +552,56 @@ def contourf(x, y, z, ax, **kwargs):
...
+    >>> _infer_interval_breaks([[0, 1], [0, 1]], axis=1)
+    array([[-0.5,  0.5,  1.5],
+           [-0.5,  0.5,  1.5]])
     """
     coord = np.asarray(coord)
-    deltas = 0.5 * (coord[1:] - coord[:-1])
-    first = coord[0] - deltas[0]
-    last = coord[-1] + deltas[-1]
-    return np.r_[[first], coord[:-1] + deltas, [last]]
+    if axis == 0:
```
shoyer: Add an `assert` or error to verify that `axis` is one of the expected values.

xarray/plot/plot.py
```diff
@@ -542,29 +552,56 @@ def contourf(x, y, z, ax, **kwargs):
...
+            infer_intervals = True
+
+    if infer_intervals:
+        if len(x.shape) == 1:
+            x = _infer_interval_breaks(x)
+            y = _infer_interval_breaks(y)
+        else:
+            # we have to infer the intervals on both axes
+            x = _infer_interval_breaks(x, axis=1)
+            x = _infer_interval_breaks(x, axis=0)
```
shoyer: clever!

fmaussion:
I had to modify your code a little bit for this to work, but this is now accepting any axis. Quite a useful function actually!

xarray/plot/plot.py
```diff
@@ -542,29 +552,52 @@ def contourf(x, y, z, ax, **kwargs):
...
-    deltas = 0.5 * (coord[1:] - coord[:-1])
-    first = coord[0] - deltas[0]
-    last = coord[-1] + deltas[-1]
-    return np.r_[[first], coord[:-1] + deltas, [last]]
+    deltas = 0.5 * np.diff(coord, axis=axis)
+    first = np.take(coord, [0], axis=axis) - np.take(deltas, [0], axis=axis)
+    last = np.take(coord, [-1], axis=axis) + np.take(deltas, [-1], axis=axis)
+    trim_last = tuple(slice(None, -1) if n == axis else slice(None)
+                      for n in range(coord.ndim))
+    return np.concatenate([first, coord[trim_last] + deltas, last], axis=axis)
```
ocefpaf: Nifty! Loved the use of `axis` here!

xarray/test/test_plot.py
```diff
@@ -112,6 +112,17 @@ def test__infer_interval_breaks(self):
         self.assertArrayEqual(pd.date_range('20000101', periods=4) - np.timedelta64(12, 'h'),
                               _infer_interval_breaks(pd.date_range('20000101', periods=3)))
 
+        # make a bounded 2D array that we will center and re-infer
+        xref, yref = np.meshgrid(np.arange(6), np.arange(5))
+        cx = (xref[1:, 1:] + xref[:-1, :-1]) / 2
```
shoyer: On Python 2, I think this will still have integer dtype. Can you divide by `2.0` instead?

fmaussion: I added a ``from __future__ import division`` instead

xarray/test/test_plot.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import division
```
shoyer: Can you add this to the top of literally every single Python file in xarray?

That will keep this easier to reason about.

fmaussion: :D sure

shoyer: Actually:
```
```

shoyer:
Yes, thanks!

Issue #798, opened 2016-03-21
Integration  with dask/distributed (xarray backend design)


pwolfram:
Dask (https://github.com/dask/dask) currently provides on-node parallelism for medium-size data problems.  However, large climate data sets will require multiple-node parallelism to analyze large climate data sets because this constitutes a big data problem.  A likely solution to this issue is integration of distributed (https://github.com/dask/distributed) with dask.  Distributed is now integrated with dask and its benefits are already starting to be realized, e.g., see http://matthewrocklin.com/blog/work/2016/02/26/dask-distributed-part-3.

Thus, this issue is designed to identify the steps needed to perform this integration, at a high-level. As stated by @shoyer, it will 

> definitely require some refactoring of the xarray backend system to make this work cleanly, but that's 
> OK -- the xarray backend system is indicated as experimental/internal API precisely because we 
> hadn't figured out all the use cases yet."
> 
> To be honest, I've never been entirely happy with the design we took there (we use inheritance rather 
> than composition for backend classes), but we did get it to work for our use cases. Some refactoring 
> with an eye towards compatibility with dask distributed seems like a very worthwhile endeavor. We 
> do have the benefit of a pretty large test suite covering existing use cases.

Thus, we have the chance to make xarray big-data capable as well as provide improvements to the backend.

To this end, I'm starting this issue to help begin the design process following the xarray mailing list discussion some of us have been having (@shoyer, @mrocklin, @rabernat).

Task To Do List:
- [x] Verify asynchronous access error for `to_netcdf` output is resolved (e.g., https://github.com/pydata/xarray/issues/793)
- [x] LRU-cached file IO supporting serialization to robustly support HDF/NetCDF reads

pwolfram:
The full mailing list discussion is at https://groups.google.com/d/msgid/xarray/CAJ8oX-E7Xx6NT4F6J8B4__Q-kBazoob9_qe_oFLi5hany9-%3DKQ%40mail.gmail.com?utm_medium=email&utm_source=footer

pwolfram:
See also https://github.com/dask/dask/issues/922

mrocklin:
Copying over [a comment](https://github.com/dask/dask/issues/922#issuecomment-199085431) from that issue: 

Yes, so the problem as I see it is that, for serialization and open-file reasons we want to use a function like the following:

```python
def get_chunk_of_array(filename, datapath, slice):
    with netCDF4.Dataset(filename) as f:
        return f.variables[datapath][slice]
```

[...]

PR #1083, opened 2016-11-06
Initial hack to get dask distributed working

shoyer:
WIP WIP WIP

CC
@mrocklin
@pwolfram 

xref #798

### DIFF
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -8,6 +8,12 @@ import pandas as pd
 from collections import defaultdict
 from pandas.tslib import OutOfBoundsDatetime
 
+try:
+    import distributed.protocol
+    HAS_DISTRIBUTED = True
+except ImportError:
+    HAS_DISTRIBUTED = False
+
 from .core import indexing, ops, utils
 from .core.formatting import format_timestamp, first_n_items, last_item
 from .core.variable import as_variable, Variable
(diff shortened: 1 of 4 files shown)


mrocklin:
It occurs to me after reading your DataStore API issue that we could also just implement getstate/setstate methods on all of the wrapping classes (like LazilyIndexedArray) that thought about how to serialize their contained arrays.  You wouldn't have to use dask's mechanism.  

You would still have to figure out how to effectively serialize and deserialize any contained object, which is the same problem we're solving at https://github.com/dask/distributed/pull/620 , but could stick to the pickle protocol if you wanted to stay with more general and established protocols.  You are reaching down a little bit in the object tree further than was intended, but that might be a more palatable long term solution for xarray?

mrocklin:
Not that I'm recommending this, just that it is another option.

xarray/conventions.py
```diff
@@ -470,6 +476,27 @@ def __getitem__(self, key):
...
+    header['__char_to_string_array__array_type'] = header.get('type')
+    return header, frames
+
+
+def deserialize_char_to_string_array(header, frames):
+    type_ = header.pop('__char_to_string_array__array_type', None)
+    if type_ is not None:
+        header['type'] = type_
+    array = distributed.protocol.deserialize(header, frames)
+    return CharToStringArray(array)
```
shoyer: That would be fine, but these classes (other than `LazilyIndexedArray`) are not going to be layered multiple times. So I wouldn't worry too much about that until we settle on a general solution.

shoyer:
> It occurs to me after reading your DataStore API issue that we could also just implement getstate/setstate methods on all of the wrapping classes (like LazilyIndexedArray) that thought about how to serialize their contained arrays. You wouldn't have to use dask's mechanism.

If I understand correctly, I could even just go to the bottom level, and make `NetCDF4ArrayWrapper` pickle-able, using a similar strategy to the one in the dask-distributed PR.

This does seem a little more palatable in the long term, especially if we build some more general machinery that DataStores can plug into (added into https://github.com/pydata/xarray/pull/1087)

shoyer:
Closing in favor of https://github.com/pydata/xarray/pull/1095

PR #1088, opened 2016-11-07
missing return value in sample function calls (I think)

vnoel:
sorry if I messed up, I'm not a github master

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -423,7 +423,7 @@ library::
         combined = xr.concat(dataset, dim)
         return combined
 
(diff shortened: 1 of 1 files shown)


jhamman:
I'm fine with this change but it isn't entirely necessary since these examples are really just hypotheticals.

vnoel:
I was briefly confused by the examples as at first they seemed to me to do nothing -- read data and forget it. This change prevents any similar confusion :-)

shoyer:
Thanks

PR #629, opened 2015-10-19
ENH: add tolerance argument to .sel and .reindex

shoyer:
`.sel` and `.reindex` now support the ``tolerance`` argument for controlling nearest-neighbor selection:

    In [5]: array = xray.DataArray([1, 2, 3], dims='x')

    In [6]: array.reindex(x=[0.9, 1.5], method='nearest', tolerance=0.2)
    Out[6]:
    <xray.DataArray (x: 2)>
    array([  2.,  nan])
    Coordinates:
      * x        (x) float64 0.9 1.5

This feature requires pandas v0.17 or newer.

PR #1095, opened 2016-11-08
Alternate approach to serializing netcdfs for dask.distributed

shoyer:
Alternative to https://github.com/pydata/xarray/pull/1083

Based on @mrocklin's suggestion of (ab)using pickle.

The good news is that this version can handle many more types out of the box (no datetime64 or timedelta64, though), and I don't need to do anything special for all of my other special array types further away from netCDF files.

cc @pwolfram -- maybe you'll have more luck with this version?

### DIFF
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -3,6 +3,12 @@ from functools import partial
 
 import numpy as np
 
+try:
+    import distributed.protocol
+    HAS_DISTRIBUTED = True
+except ImportError:
+    HAS_DISTRIBUTED = False
+
 from .. import Variable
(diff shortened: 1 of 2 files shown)


mrocklin:
Ah, so you've handled the leaf case and everything else follows naturally from there thanks to pickle.  Nice.

pwolfram:
This seemed to work, see #629 comment.

pwolfram:
Note, I did get a memory error for usage via `ds.sum()`.

shoyer:
Closing in favor of https://github.com/pydata/xarray/pull/1128, which simply makes all datastores pickle-able.

PR #1098, opened 2016-11-08
Docs tweaks

fmaussion:
The multidimensional coords example is now built live. I tested this locally and tried to keep the look of the figures as close as possible to the original ones. 

CC @rabernat @shoyer

(I also added a link to salem in the list of packages extending xarray - shameless self-promotion)


doc/internals.rst
```diff
@@ -133,6 +133,8 @@ They may be useful points of reference for your work:
 - `windspharm <http://ajdawson.github.io/windspharm/index.html>`_: Spherical
   harmonic wind analysis in Python.
 - `eofs <http://ajdawson.github.io/eofs/>`_: EOF analysis in Python.
+- `salem <http://salem.readthedocs.io>`_: extends xarray to add geolocalised
```
fmaussion: Sure. What do you have in mind for the more user facing page? And which of the packages are ready for prime-time? Salem is surely the less ready here, but there are also others listed here: https://aospy.hackpad.com/Summary-of-1st-brainstorming-round-lLJ3vXhCk7Z#:h=Potentially-related-or-relevan

shoyer: Take a look at https://github.com/pydata/xarray/pull/1098 -- I'll comment over there.

fmaussion:
Perfect! I removed the reference to salem from this PR, which is ready to merge/review

fmaussion:
As always: -> testing locally is not the way to go with docs. I'll have a look at what's going wrong in the build: http://xarray.pydata.org/en/latest/examples/multidimensional-coords.html

fmaussion:
Aha ok. Based on [this example](http://xarray.pydata.org/en/latest/plotting.html#maps) I assumed that everything was available on RTD, but it's also static.

I've been using cartopy and netcdf4 successfully on salem's docs with conda-force [here](https://github.com/fmaussion/salem/blob/master/docs/environment.yml) my environment file. @shoyer you OK with adding some packages to the doc builds?

fmaussion:
Alternatively, one should revert this change. I have no strong opinion about this. Doc builds are better to detect bugs in the code, but they are also tedious to maintain...

shoyer:
Yes, please do add cartopy and netcdf4 to our docs builds. The lack of these packages is only a legacy of times before RTD added conda support.

fmaussion:
OK, so now cartopy is installed properly (the map example looks good: http://xarray.pydata.org/en/latest/plotting.html#maps), but the example still fails with a netcdf3 error: http://xarray.pydata.org/en/latest/examples/multidimensional-coords.html :-(

I guess that netCDF4 didn't install properly. Next thing I'll try is to unpin the package version

Issue #1090, opened 2016-11-07
xarray.ufuncs triggering eager evaluation with dask v0.12.0


shoyer:
This is due to bug in the latest release of dask.
```
self = <xarray.test.test_dask.TestVariable testMethod=test_bivariate_ufunc>

    def test_bivariate_ufunc(self):
        u = self.eager_var
        v = self.lazy_var
        self.assertLazyAndAllClose(np.maximum(u, 0), xu.maximum(v, 0))
>       self.assertLazyAndAllClose(np.maximum(u, 0), xu.maximum(0, v))

xarray/test/test_dask.py:189:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
xarray/test/test_dask.py:57: in assertLazyAndAllClose
    self.assertLazyAnd(expected, actual, self.assertVariableAllClose)
xarray/test/test_dask.py:51: in assertLazyAnd
    self.assertIsInstance(var.data, da.Array)
E   AssertionError: array([[ 1.76405235,  0.40015721,  0.97873798,  2.2408932 ,  1.86755799,
E            0.        ],
E          [ 0.95008842,  0.        ,  0.        ,  0.4105985 ,  0.14404357,
E            1.45427351],
E          [ 0.76103773,  0.12167502,  0.44386323,  0.33367433,  1.49407907,
E            0.        ],
E          [ 0.3130677 ,  0.        ,  0.        ,  0.6536186 ,  0.8644362 ,
E            0.        ]]) is not an instance of <class 'dask.array.core.Array'>
```

PR #1100, opened 2016-11-09
Disable broken test for dask ufuncs

shoyer:
xref #1090

### DIFF
--- a/xarray/test/test_dask.py
+++ b/xarray/test/test_dask.py
@@ -7,6 +7,8 @@ import xarray.ufuncs as xu
 from xarray.core.pycompat import suppress
 from . import TestCase, requires_dask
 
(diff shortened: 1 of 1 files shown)


xarray/test/test_dask.py
```diff
@@ -182,6 +184,7 @@ def test_univariate_ufunc(self):
         v = self.lazy_var
         self.assertLazyAndAllClose(np.sin(u), xu.sin(v))
 
+    @unittest.skip('currently broken in dask, see GH1090')
```
MaximilianR: Obv not worth changing, but for future pytest-ty-ness: `@pytest.mark.skip(reason="...")`

shoyer: Thanks, I'll use that next time!

PR #1101, opened 2016-11-09
unpin package versions for doc build

fmaussion:


shoyer:
This is fine for debugging, but once we have something working, please pin again. In my experience that makes the doc build much much more reliable.

fmaussion:
Unpinning didn't help: http://xarray.pydata.org/en/latest/examples/multidimensional-coords.html 

I'm out of ideas right now - without looking at the RTD build logs I'm not sure how to proceed... Sorry for the work

shoyer:
You can actually view doc builds publicly on RTD:
http://readthedocs.org/projects/xray/

But.... congratulations... you now have admin rights to xarray on RTD anyways!

PR #1104, opened 2016-11-10
add optimization tips

### LINKED ISSUE #1103: add dask optimization tips to docs
We should add the optimization tips that @shoyer describes in this mailing list thread to @karenamckinnon.

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/xarray/11lDGSeza78/lR1uj9yWDAAJ

Specific things to try (we should add similar guidelines to xarray's docs):

1. Do your spatial and temporal indexing with .sel() earlier in the pipeline, specifically before you resample. Resample triggers some computation on all the blocks, which in theory should commute with indexing, but we haven't implemented this optimization in dask yet:
https://github.com/dask/dask/issues/746
2. Save the temporal mean to disk as a netCDF file (and then load it again with open_dataset) before subtracting it. Again, in theory, dask should be able to do the computation in a streaming fashion, but in practice this is a fail case for the dask scheduler, because it tries to keep every chunk of an array that it computes in memory:
https://github.com/dask/dask/issues/874
3. Specify smaller chunks across space when using open_mfdataset, e.g., chunks={'latitude': 10, 'longitude': 10}. This makes spatial subsetting easier, because there's no risk you will load chunks of data referring to different chunks (probably not necessary if you do my suggestion 1).

rabernat:
This adds some dask optimization tips from the mailing list (closes #1103).

### DIFF
--- a/doc/dask.rst
+++ b/doc/dask.rst
@@ -225,3 +225,14 @@ larger chunksizes.
 
     import os
     os.remove('example-data.nc')
+    
+Optimization Tips
+-----------------
+
+With analysis pipelines involving both spatial subsetting and temporal resampling, dask performance can become very slow in certain cases. Here are some optimization tips we have found through experience:
+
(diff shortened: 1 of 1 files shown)


shoyer:
Thank you Ryan!!

PR #1105, opened 2016-11-10
Another attempt to get netCDF4 working on RTD + py3

fmaussion:
Sorry @shoyer you might get a few of these debugging PRs today again.


shoyer:
You'll probably find a few places where we current have doc build failures on Python 3 -- but please feel free to clean them up! (Python 2 -> 3 is definitely a step forward)

PR #1108, opened 2016-11-11
py2 compat header

fmaussion:
As discussed in https://github.com/pydata/xarray/pull/1079

I just left out a couple of nearly empty ``__init__``s 

I also removed a couple of unused imports

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -1,3 +1,6 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
 """String formatting routines for __repr__.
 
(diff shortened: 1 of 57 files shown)


fmaussion:
The test failure seems unrelated

xarray/core/formatting.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import absolute_import
```
fmaussion: ups! yes, sorry

shoyer:
I tried restarting the failing test which looks like a conda issue -- let's see if that helps.

shoyer:
Thanks!

Issue #1106, opened 2016-11-10
Getting netCDF4 to work on RTD


fmaussion:
This is to ping @ocefpaf on whether you have an idea on what's going on with netCDF4 on Read The Docs. See the import error [here](http://xarray.pydata.org/en/latest/examples/multidimensional-coords.html).

This is our conda config file: https://github.com/pydata/xarray/blob/master/doc/environment.yml

I've found related discussions [here](https://groups.google.com/forum/#!topic/pyart-users/OGg1YfnO6ZQ) or [here](https://github.com/ioos/conda-recipes/issues/771). Note that I never saw this when building salem's doc, which installs *many* more packages  from conda-forge (see [conf file](https://github.com/fmaussion/salem/blob/master/docs/environment.yml)).

Thanks a lot for your help! And no hurry.

shoyer:
Another option would be to switch all the example files (https://github.com/pydata/xarray-data) to use netCDF3. That's probably a good idea for maximum compatibility in any case, because we can read netCDF3 with only scipy installed.

cc @jhamman

fmaussion:
I'm happy to do the conversion if you want. One could compress the file a little bit by changing the data to float32, too

fmaussion:
Yes, with the following code the file size goes down to 9mb:

```python
import xarray as xr
ds = xr.open_dataset('/home/tt/Downloads/rasm.nc')
ds.Tair.encoding['dtype'] = 'float32'
ds.to_netcdf('/home/tt/rasm.nc', format='NETCDF3_CLASSIC')
```

shoyer:
Yes, please open a PR in xarray-data!

fmaussion:
I wanted to be sure that IO works well with the scipy engine, but I get this error below. Does that look familiar to you?

```python
import xarray as xr
ds = xr.open_dataset('/home/mowglie/Downloads/rasm.nc')
ds.Tair.encoding['dtype'] = 'float32'
ds.to_netcdf('/home/mowglie/rasm.nc', format='NETCDF3_CLASSIC', engine='scipy')

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/home/mowglie/.bin/conda/envs/climate/lib/python3.5/site-packages/xarray/backends/api.py in to_netcdf(dataset, path, mode, format, group, engine, writer, encoding)
    353     try:
--> 354         dataset.dump_to_store(store, sync=sync, encoding=encoding)
    355         if isinstance(path, BytesIO):

[...]

PR #1110, opened 2016-11-11
Add default channel to conda on RTD

fmaussion:
As discussed here: https://github.com/pydata/xarray/issues/1106

I also added a section somewhere to print the packages versions on RTD. It will be useful to pin the versions again once it works, and is useful enough to stay on the docs I think. I just didn't really knew where to put it, other ideas welcome!

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -11,43 +11,56 @@
 #
 # All configuration values have a default; values that are commented out
 # serve to show the default.
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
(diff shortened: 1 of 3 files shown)


shoyer:
We print version info in `conf.py`, which shows up in the build logs on RTD:
https://github.com/pydata/xarray/blob/master/doc/conf.py#L18

fmaussion:
Uh, this is a good candidate for a ``from __future__ import print_function``

shoyer:
indeed!

Issue #1111, opened 2016-11-11
Unable to decode time axis on rasm file


fmaussion:
```python
import xarray as xr
import netCDF4

print(xr.__version__)  # 0.8.2-50-g57facab
print(netCDF4.__version__) # 1.2.4

ds = xr.tutorial.load_dataset('rasm')
/home/mowglie/Documents/git/xarray/xarray/conventions.py:389: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  result = decode_cf_datetime(example_value, units, calendar)
```

I'll have a closer look.

fmaussion:
This is due to the [0] here: https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L386

Since the units of time are ``days since 0001-01-01`` this raises an error. Any reason why this 0 is here?

shoyer:
Added in https://github.com/pydata/xarray/pull/984

Looks like the original intent was to check 0 only if other dates were *not* found. So this should probably be fixed.

PR #1112, opened 2016-11-11
Bug in DecodedCFDatetimeArray

fmaussion:
Fixes https://github.com/pydata/xarray/issues/1111

### DIFF
--- a/xarray/test/test_conventions.py
+++ b/xarray/test/test_conventions.py
@@ -497,8 +497,8 @@ class TestDatetime(TestCase):
         # this should not throw a warning (GH1111)
         with warnings.catch_warnings():
             warnings.filterwarnings('error')
-            _ = conventions.DecodedCFDatetimeArray(np.asarray([722624]),
-                                                   "days since 0001-01-01")
+            conventions.DecodedCFDatetimeArray(np.asarray([722624]),
+                                               "days since 0001-01-01")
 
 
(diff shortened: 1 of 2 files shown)


xarray/test/test_conventions.py
```diff
@@ -494,6 +494,12 @@ def test_dataset_repr_with_netcdf4_datetimes(self):
         ds = decode_cf(Dataset({'time': ('time', [0, 1], attrs)}))
         self.assertIn('(time) datetime64[ns]', repr(ds))
 
+        # this should not throw a warning (GH1111)
```
fmaussion: Right, that's much easier! I was assuming that the tutorial files were also used for your tests. Salem's test suite is based on real files (similar to xarray's) and I've never had troubles with it, but I agree that it's not optimal

xarray/test/test_conventions.py
```diff
@@ -495,10 +495,10 @@ def test_dataset_repr_with_netcdf4_datetimes(self):
         self.assertIn('(time) datetime64[ns]', repr(ds))
 
         # this should not throw a warning (GH1111)
-        from xarray.tutorial import load_dataset
         with warnings.catch_warnings():
             warnings.filterwarnings('error')
-            ds = load_dataset('rasm')
+            _ = conventions.DecodedCFDatetimeArray(np.asarray([722624]),
```
fmaussion: Right! Done.

Issue #1109, opened 2016-11-11
Fix "allowed failures" Travis-CI builds for pydap and dev versions of netCDF4 and pandas


shoyer:
These are useful for catching changes upstream before they turn into bugs for xarray users, but they are currently failing for spurious reasons.

If someone has time to investigate and fix these, that would be great!

jhamman:
I've just opened #2736 to close this issue. In actuality, this issue could be closed as is because:

1) The pydap and netcdf4 builds were removed in #2661 
2) The pandas dev build is passing in recent travis builds

PR #1118, opened 2016-11-14
Do not convert subclasses of `ndarray` unless required

burnpanck:
By changing a single `np.asarray` to `np.asanyarray`, it becomes possible to use `ndarray` subclasses within `DataArray` and  `Dataset`. No guarantee is made at this point that their behaviour is retained under all circumstances, but for some use-cases it will work.

Particularly, this allows to store physical quantities represented using the `Quantity` subclass from the package [python-quantities](https://github.com/python-quantities/python-quantities). In that sense, it is a partial fix for #525. Tests are included to highlight some of the things that do work, and some which don't. It does not work for coordinates when they are used as indexes, since pandas does not support ndarray subclasses. Most likely, it will also not work for any higher-level operation on the data such as those involving `np.concatenate`. Thus, any user of this feature should remain cautious.
Nonetheless, for me, this change provides enough added value, such that I'd consider it worthwhile to keep, assuming it does not harm anywhere else. It does pass the tests on my machine (no dask though), let's see what travis says.

I expect that astropy's units would behave similarly, though since I never worked with them yet, I did not include any tests.

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -185,7 +185,7 @@ def _as_array_or_item(data):
 
     TODO: remove this (replace with np.asarray) once these issues are fixed
     """
-    data = np.asanyarray(data)
+    data = np.asarray(data)
     if data.ndim == 0:
         if data.dtype.kind == 'M':
             data = np.datetime64(data, 'ns')
@@ -193,6 +193,19 @@ def _as_array_or_item(data):
             data = np.timedelta64(data, 'ns')
     return data
 
+def _as_any_array_or_item(data):
+    """Return the given values as a numpy array subclass instance, or as an
+    individual item if it's a 0d datetime64 or timedelta64 array.
+
+    The same caveats as for :py:meth:`_as_array_or_item` apply.
+    """
+    data = np.asanyarray(data)
+    if data.ndim == 0:
+        if data.dtype.kind == 'M':
+            data = np.datetime64(data, 'ns')
+        elif data.dtype.kind == 'm':
+            data = np.timedelta64(data, 'ns')
+    return data
 
 class Variable(common.AbstractArray, common.SharedMethodsMixin,
                utils.NdimSizeLenMixin):
@@ -267,7 +280,7 @@ class Variable(common.AbstractArray, common.SharedMethodsMixin,
         if isinstance(self._data, dask_array_type):
             return self._data
         else:
-            return self.values
+            return _as_any_array_or_item(self._data_cached())
 
     @data.setter
     def data(self, data):
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -185,7 +185,7 @@ def _as_array_or_item(data):
 
     TODO: remove this (replace with np.asarray) once these issues are fixed
     """
-    data = np.asarray(data)
+    data = np.asanyarray(data)
     if data.ndim == 0:
         if data.dtype.kind == 'M':
             data = np.datetime64(data, 'ns')
--- a/xarray/test/test_quantities.py
+++ b/xarray/test/test_quantities.py
@@ -12,7 +12,7 @@ except ImportError:
     has_quantities = False
 
 def requires_quantities(test):
-    return test if has_quantities else unittest.skip('requires dask')(test)
+    return test if has_quantities else unittest.skip('requires python-quantities')(test)
 
 @requires_quantities
 class TestWithQuantities(TestCase):
--- a/xarray/test/test_quantities.py
+++ b/xarray/test/test_quantities.py
@@ -53,3 +53,10 @@ class TestWithQuantities(TestCase):
         """
         da = self.da
         self.assertEqualWUnits(da.x.data,self.x)
+
+    def test_sel(self):
+        self.assertEqualWUnits(self.da.sel(y=self.y[0]).values,self.v[:,0])
+
+    def test_mean(self):
+        self.assertEqualWUnits(self.da.mean('x').values,self.v.mean(0))
+
--- a/xarray/test/test_quantities.py
+++ b/xarray/test/test_quantities.py
@@ -1,0 +1,55 @@
+import numpy as np
+import pandas as pd
+
+from xarray import (align, broadcast, Dataset, DataArray, Variable)
+
+from xarray.test import (TestCase, unittest)
+
+try:
+    import quantities as pq
+    has_quantities = True
+except ImportError:
+    has_quantities = False
+
+def requires_quantities(test):
+    return test if has_quantities else unittest.skip('requires dask')(test)
+
+@requires_quantities
+class TestWithQuantities(TestCase):
+    def setUp(self):
+        self.x = np.arange(10) * pq.A
+        self.y = np.arange(20)
+        self.xp = np.arange(10) * pq.J
+        self.v = np.arange(10*20).reshape(10,20) * pq.V
+        self.da = DataArray(self.v,dims=['x','y'],coords=dict(x=self.x,y=self.y,xp=(['x'],self.xp)))
+
+    def assertEqualWUnits(self,a,b):
+        self.assertIsNotNone(getattr(a, 'units', None))
+        self.assertIsNotNone(getattr(b, 'units', None))
+        self.assertEqual(a.units,b.units)
+        np.testing.assert_allclose(a.magnitude,b.magnitude)
+
+    def test_units_in_data_and_coords(self):
+        da = self.da
+        self.assertEqualWUnits(da.xp.data,self.xp)
+        self.assertEqualWUnits(da.data,self.v)
+
+    def test_arithmetics(self):
+        da = self.da
+        f = DataArray(np.arange(10*20).reshape(10,20)*pq.A,dims=['x','y'],coords=dict(x=self.x,y=self.y))
+        self.assertEqualWUnits((da*f).data, da.data*f.data)
+
+    def test_unit_checking(self):
+        da = self.da
+        f = DataArray(np.arange(10*20).reshape(10,20)*pq.A,dims=['x','y'],coords=dict(x=self.x,y=self.y))
+        with self.assertRaisesRegex(ValueError,'Unable to convert between units'):
+            da + f
+
+    @unittest.expectedFailure
+    def test_units_in_indexes(self):
+        """
+        Indexes are borrowed from Pandas, and Pandas does not support units.
+        Therefore, we currently don't intend to support units on idexes either.
+        """
+        da = self.da
+        self.assertEqualWUnits(da.x.data,self.x)
--- a/xarray/test/test_quantities.py
+++ b/xarray/test/test_quantities.py
@@ -1,3 +1,14 @@
+""" test_quantities: Test storing and using instances of
+:py:class:`quantities.Quantity` inside :py:class`xarray.DataArray`.
+
+It can be considered a stand-in for other :py:class:`numpy.ndarray`
+subclasses, particularly other units implementations such as
+astropy's. As preservation of subclasses is not a guaranteed feature of
+`xarray`, some operations will discard subclasses. This test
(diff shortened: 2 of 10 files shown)


shoyer:
This does seem pretty reasonable to me.

But is this actually running in CI? I think you need to add `quantities` to `requirements.yml` files to get this to run in Travis-CI:
https://github.com/pydata/xarray/tree/master/ci

burnpanck:
You are right. There seem to be quite a number of varying `requirements`. Should I add it to all of them? Also, I'm not very well versed in Travis-CI `.yml`: What repo are the requirements served from? I think `python-quantities` is in debian. Or should I just go for pip? There hasn't been a release in a long time, github master has progressed quite a bit, but the released version should be compatible too.

shoyer:
It either uses conda (by default) or pip:
https://github.com/conda/conda-env

Please do add it to everything, except perhaps the build with minimal dependencies (`py27-min`)

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+from xarray.test import (TestCase, unittest)
+
+try:
+    import quantities as pq
+    has_quantities = True
+except ImportError:
+    has_quantities = False
+
+def requires_quantities(test):
+    return test if has_quantities else unittest.skip('requires dask')(test)
```
dopplershift: Need to update the message here.

burnpanck:
Travis succeeds, though lots of failures under environments with allowed failure. They look unrelated to me, but I find it hard to tell. Appveyor doesn't seem to run the quantities tests, so I guess the requirements there are missing too. Where would I add requirements for Appveyor?

shoyer:
Indeed, the allowed failures are unrelated: https://github.com/pydata/xarray/issues/1109

See here for Appveyor installation: https://github.com/pydata/xarray/blob/master/appveyor.yml

It would be good to switch it use `conda env` for consistency but haven't gotten around to that yet (https://github.com/pydata/xarray/issues/1127)

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+    def assertEqualWUnits(self,a,b):
+        self.assertIsNotNone(getattr(a, 'units', None))
+        self.assertIsNotNone(getattr(b, 'units', None))
+        self.assertEqual(a.units,b.units)
+        np.testing.assert_allclose(a.magnitude,b.magnitude)
+
+    def test_units_in_data_and_coords(self):
+        da = self.da
+        self.assertEqualWUnits(da.xp.data,self.xp)
+        self.assertEqualWUnits(da.data,self.v)
```
shoyer: Can we adjust things so that `DataArray.values` still always returns a base `numpy.ndarray`, but `DataArray.data` can return a subclass? That would be consistent with our current behavior for handling duck-type arrays with dask.

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
```
shoyer: Can you add a docstring explaining that this module uses quantities as a concrete example, but really it's just checking for subclass compatibility.

Also, I might call it something more genericl ike `test_numpy_subclasses.py`.

xarray/core/variable.py
```diff
@@ -185,7 +185,7 @@ def _as_array_or_item(data):
 
     TODO: remove this (replace with np.asarray) once these issues are fixed
     """
-    data = np.asarray(data)
+    data = np.asanyarray(data)
```
shoyer: This method should probably be renamed, e.g., `_as_any_array_or_item`, just to make it super clear what it's doing (also update the TODO in the docstring)

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def test_units_in_data_and_coords(self):
+        da = self.da
+        self.assertEqualWUnits(da.xp.data,self.xp)
+        self.assertEqualWUnits(da.data,self.v)
+
+    def test_arithmetics(self):
+        da = self.da
+        f = DataArray(np.arange(10*20).reshape(10,20)*pq.A,dims=['x','y'],coords=dict(x=self.x,y=self.y))
+        self.assertEqualWUnits((da*f).data, da.data*f.data)
```
shoyer: What about arithmetic that requires broadcasting? e.g., between arrays with different dimensions and/or different dimension order?

That will require ensuring that `ops.transpose` and `ops.broadcast_to` work on subclasses (e.g., `xarray.Variable.transpose` and `xarray.Variable.expand_dims`). I think the former *should* work already because of NumPy, but `broadcast_to` will probably require adding `subok=True`.

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+        da = self.da
+        f = DataArray(np.arange(10*20).reshape(10,20)*pq.A,dims=['x','y'],coords=dict(x=self.x,y=self.y))
+        with self.assertRaisesRegexp(ValueError,'Unable to convert between units'):
+            da + f
+
+    @unittest.expectedFailure
+    def test_units_in_indexes(self):
+        """
+        Indexes are borrowed from Pandas, and Pandas does not support units.
+        Therefore, we currently don't intend to support units on indexes either.
```
shoyer: usually we use comments instead of docstrings on test methods

xarray/test/test_quantities.py
burnpanck: I see. I was rather thinking of this test-file as specialised towards physical units handling with quantities (i.e. as a regression-test database geared towards #525). I'd rather make a separate test-case for general subclasses, maybe with a dummy subclass.

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+    return test if has_quantities else unittest.skip('requires python-quantities')(test)
+
+@requires_quantities
+class TestWithQuantities(TestCase):
+    def setUp(self):
+        self.x = np.arange(10) * pq.A
+        self.y = np.arange(20)
+        self.xp = np.arange(10) * pq.J
+        self.v = np.arange(10*20).reshape(10,20) * pq.V
+        self.da = DataArray(self.v,dims=['x','y'],coords=dict(x=self.x,y=self.y,xp=(['x'],self.xp)))
```
burnpanck: Sure.

xarray/test/test_quantities.py
shoyer: Sure, `test_units_subclass.py` also seems reasonable.

xarray/test/test_quantities.py
burnpanck: Fine, but I am not so happy about this.

xarray/test/test_quantities.py
burnpanck: I'll have a go.

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        Indexes are borrowed from Pandas, and Pandas does not support units.
+        Therefore, we currently don't intend to support units on indexes either.
+        """
+        da = self.da
+        self.assertEqualWUnits(da.x.data,self.x)
+
+    @unittest.expectedFailure
+    def test_sel(self):
+        self.assertEqualWUnits(self.da.sel(y=self.y[0]).values,self.v[:,0])
```
burnpanck: Apparently yes. I didn't get around to actually look into this.

xarray/test/test_quantities.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @unittest.expectedFailure
+    def test_mean(self):
+        self.assertEqualWUnits(self.da.mean('x').values,self.v.mean(0))
```
burnpanck: Ok

xarray/core/variable.py
burnpanck: Ok

xarray/test/test_quantities.py
dopplershift: I agree, not important, but I'm really surprised by this.

shoyer: This is the convention for NumPy. To be honest I'm not sure why, but I recall being told so at one point by @charris.

I see now that this isn't followed for pandas, so I guess we can break it here. I don't feel too strongly either way.

dopplershift: pytest doesn't seem to with either `-v` or `-vv`

xarray/test/test_quantities.py
burnpanck: The question is, what should be the exact guarantee that `.values` wants to give above `.data`. For me, units are an inherent part of the data (see my comment in #525) My view here is that the behaviour of an instance of `Quantity` is indeed very close to that of the bare `ndarray`, and thus shouldn't be treated any different. The problem here is of course that this will not necessarily be the case for a general subclass. So, the question is, where to draw the line between `.values` and `.data`. Should it be bare `ndarray` vs. everything else or just `ndarray` subclass vs. things that behave like arrays but aren't. The best answer will depend both on what is out there in terms of subclasses and objects that just look like arrays, as well as what the code using xarray expects to be able to do with what it gets from `.value`. I don't want to judge that, because I think I don't know enough about the eco-system xarray lives in. (E.g. dask: Is a dask array an ndarray subclass or not?). In any case, if you think `.values` should be bare `ndarray` only, then I'll respect that.

shoyer: dask arrays are not ndarray subclasses, so agree, it's not entirely the same.

Either way, some users will need to wrap their calls to `.values` or `.data` with `np.asanyarray()` (to ensure the result is an ndarray subclass and not possibly a dask array) or `np.asarray()` (to ensure the result is a base ndarray). Actually, it's probably worth checking if `np.asanyarray(data_array)` does the right thing in your patch -- possibly `DataArray.__array__` needs to be updated to call `np.asanyarray`.

Just speaking personally -- and I'm happy for other to chime in with their opinions -- I find a shortcut for `np.asarray` more essential than `np.asanyarray`. There are lots of libraries not written to handle anything other than base numpy array objects, and being able to convert into something that is guaranteed to work is valuable.

xarray/test/test_quantities.py
burnpanck: Actually, I don't fully understand this yet: Bare `Quantities` do work fine with `nanmean`. Are you saying that `mean` is wrapped in a more complicated way than other reduction methods, or would your comment apply to any reduction. I quickly tried finding the relevant code and converged at `DataArrayGroupBy.reduce`->`.apply`, which seems to be the case for most reductions. Is that what is called when I do `.mean()` on a `DataArray`?

xarray/test/test_quantities.py
burnpanck: Ok, so if you say `.values` is really just meant to be a shortcut over `asarray` or `asanyarray`, then I agree the former is more important.

burnpanck:
With the new changes, this will now conflict with #1128, though easy to solve.

shoyer:
#1128 has been merged -- can you try rebasing? Thanks

lamorton:
Is there anything I can do to help move this forward?  I'd really like to have this capability.

shoyer:
@lamorton This branch has conflicts master (due to #1128). But otherwise, I think this is nearly ready to go in. Obviously it's not a complete fix, but it might be enough to be useful.

max-sixty:
Would anyone like to take this up? I was going to close as stale, but it's fairly close!

keewis:
Is this still relevant? If so, I'd be willing to take over. However, I am not really familiar with github yet: would I open a new pull request for that?

max-sixty:
Hi @keewis - yes! 

You can start a new PR by branching off this PR (https://help.github.com/en/articles/checking-out-pull-requests-locally). I think you'll find GH easy to use - feel free to ping here with any issues and we'll be happy to help

keewis:
closing for the same reason as #2956

PR #1119, opened 2016-11-14
Fix #1116 (rename of coordinates)

### LINKED ISSUE #1116: `DataArray.rename` of a coordinate array fails to rename the coordinate name of it's own coordinate
The documentation of `DataArray.rename` states that it renames the name of the DataArray and/or associated coordinates. However, if a name happens to be simultaneously the name of the array and of a coordinate, it only renames the array. This situation particularly arises when one tries to create a renamed coordinate to use on a different (new) array/dataset, as in that case the name of the only dimension of the coordinate array is the name of the array itself.

burnpanck:


burnpanck:
This fix handles the case `ds['somecoord'].rename({'somecoord':'newcoord'})`. It does not apply to `ds['somecoord'].rename('newcoord')`, since the documentation of `rename` does not mention coordinates at all if a string is given as argument. Nevertheless, one could argue the the two should be equivalent, in which case the latter would have to be fixed as well.

shoyer:
> This fix handles the case `ds['somecoord'].rename({'somecoord':'newcoord'})`. It does not apply to `ds['somecoord'].rename('newcoord')`

Indeed, I'm a little conflicted here. But I think this distinction makes sense, because it's useful to be able to rename an array without renaming its coordinate, too.

shoyer:
Could you please add a brief bug-fix note to what's new? Otherwise this looks good to me.

shoyer:
Thanks!

PR #1124, opened 2016-11-16
full_like, zeros_like, ones_like

### LINKED ISSUE #1102: full_like, zeros_like, ones_like
I'd like to add the following top-level functions to xarray:

```
def const_like(array, value=0):
    """Return a new array with the same shape of array and the given constant value.
    If array is dask-backed, return a new dask-backed array with the same chunks.

    :param array:
        a numpy or dask-backed xarray.DataArray
    :param value:
        any scalar number
    """
    if isinstance(array.data, dask.array.Array):
        if value == 0:
            data = dask.array.zeros(
                array.data.shape,
                chunks=array.data.chunks,
                dtype=array.data.dtype)
        else:
            data = dask.array.ones(
                array.data.shape,
                chunks=array.data.chunks,
                dtype=array.data.dtype)
    else:
        if value == 0:
            data = numpy.zeros_like(array.data)
        else:
            data = numpy.ones_like(array.data)
    if value not in (0, 1):
        data = data * value
                
    return xarray.DataArray(data, dims=array.dims, coords=array.coords, attrs=array.attrs)
  
def zeros_like(array):
    return const_like(array, 0)
  
def ones_like(array):
    return const_like(array, 1)
```

The above would need to be expanded to support Dataset and Variable objects. 
In Datasets, the data_vars would be constants whereas all other variables would be copied verbatim.
Thoughts?

crusaderky:
New top-level functions. Fixes #1102


shoyer:
Looks like one of your tests (`TestAsCompatibleData.test_full_like`) needs to use the `requires_dask` decorator so it skips if dask is not installed.

crusaderky:
fixed

crusaderky:
ping - is there anything outstanding before you can merge this?

crusaderky:
Done and ready for review

xarray/core/common.py
```diff
@@ -705,54 +705,69 @@ def _get_fill_value(dtype):
...
+    from .variable import Variable
+
+    if isinstance(other.data, dask_array_type):
+        import dask.array
+        if dtype is None:
+            dtype = other.dtype
+        data = dask.array.full(other.shape, fill_value, dtype=dtype,
+                               chunks=other.data.chunks)
+    else:
+        data = np.full_like(other, fill_value, dtype=dtype)
```
shoyer: This works, but it might be marginally safer to pass in `other.data` rather than relying on NumPy properly handling xarray's array-like `Variable` type.

shoyer:
This looks great -- thanks!

PR #1128, opened 2016-11-16
Remove caching logic from xarray.Variable

shoyer:
This is a follow-up to generalize the changes from #1024:

- Caching and copy-on-write behavior has been moved to separate array classes
  that are explicitly used in `open_dataset` to wrap arrays loaded from disk (if
  `cache=True`).
- Dask specific logic has been removed from the caching/loading logic on
  `xarray.Variable`.
- Pickle no longer caches automatically under any circumstances.
- DataStore objects are now pickleable. This makes it feasible to use xarray's IO with dask-distributed or multi-processing.

~~Still needs tests for the `cache` argument to `open_dataset`, but everything
else seems to be working.~~

@crusaderky @kynan please review

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -16,7 +16,6 @@ from .common import ArrayWriter
 from ..core import indexing
 from ..core.combine import auto_combine
 from ..core.utils import close_on_error, is_remote_uri
-from ..core.variable import Variable, IndexVariable
 from ..core.pycompat import basestring
 
 DATAARRAY_NAME = '__xarray_dataarray_name__'
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -13,12 +13,15 @@ import numpy as np
 
 from .. import backends, conventions
 from .common import ArrayWriter
+from ..core import indexing
 from ..core.combine import auto_combine
 from ..core.utils import close_on_error, is_remote_uri
+from ..core.variable import Variable, IndexVariable
 from ..core.pycompat import basestring
 
 DATAARRAY_NAME = '__xarray_dataarray_name__'
 DATAARRAY_VARIABLE = '__xarray_dataarray_variable__'
+
 
 def _get_default_engine(path, allow_remote=False):
     if allow_remote and is_remote_uri(path):  # pragma: no cover
@@ -117,10 +120,20 @@ def _validate_attrs(dataset):
             check_attr(k, v)
 
 
+def _protect_dataset_variables_inplace(dataset, cache):
+    for name, variable in dataset.variables.items():
+        if name not in variable.dims:
+            # no need to protect IndexVariable objects
+            data = indexing.CopyOnWriteArray(variable._data)
+            if cache:
+                data = indexing.MemoryCachedArray(data)
+            variable._data = data
+
+
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=True, decode_times=True,
                  concat_characters=True, decode_coords=True, engine=None,
-                 chunks=None, lock=None, drop_variables=None):
+                 chunks=None, lock=None, cache=None, drop_variables=None):
     """Load and decode a dataset from a file or file-like object.
 
     Parameters
@@ -162,14 +175,22 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         'netcdf4'.
     chunks : int or dict, optional
         If chunks is provided, it used to load the new dataset into dask
-        arrays. This is an experimental feature; see the documentation for more
-        details.
+        arrays. ``chunks={}`` loads the dataset with dask using a single
+        chunk for all arrays. This is an experimental feature; see the
+        documentation for more details.
     lock : False, True or threading.Lock, optional
         If chunks is provided, this argument is passed on to
         :py:func:`dask.array.from_array`. By default, a per-variable lock is
         used when reading data from netCDF files with the netcdf4 and h5netcdf
         engines to avoid issues with concurrent access when using dask's
         multithreaded backend.
+    cache : bool, optional
+        If True, cache data loaded from the underlying datastore in memory as
+        NumPy arrays when accessed to avoid reading from the underlying data-
+        store multiple times. Defaults to True unless you specify the `chunks`
+        argument to use dask, in which case it defaults to False. Does not
+        change the behavior of coordinates corresponding to dimensions, which
+        always load their data from disk into a ``pandas.Index``.
     drop_variables: string or iterable, optional
         A variable or list of variables to exclude from being parsed from the
         dataset. This may be useful to drop variables with problems or
@@ -190,11 +211,16 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         concat_characters = False
         decode_coords = False
 
+    if cache is None:
+        cache = chunks is not None
+
     def maybe_decode_store(store, lock=False):
         ds = conventions.decode_cf(
             store, mask_and_scale=mask_and_scale, decode_times=decode_times,
             concat_characters=concat_characters, decode_coords=decode_coords,
             drop_variables=drop_variables)
+
+        _protect_dataset_variables_inplace(ds, cache)
 
         if chunks is not None:
             try:
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -48,6 +48,13 @@ def _get_default_engine(path, allow_remote=False):
     return engine
 
 
+def _normalize_path(path):
+    if is_remote_uri(path):
+        return path
+    else:
+        return os.path.abspath(os.path.expanduser(path))
+
+
 _global_lock = threading.Lock()
 
 
@@ -251,6 +258,17 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     if isinstance(filename_or_obj, backends.AbstractDataStore):
         store = filename_or_obj
     elif isinstance(filename_or_obj, basestring):
+
+        if (isinstance(filename_or_obj, bytes) and
+                filename_or_obj.startswith(b'\x89HDF')):
+            raise ValueError('cannot read netCDF4/HDF5 file images')
+        elif (isinstance(filename_or_obj, bytes) and
+                filename_or_obj.startswith(b'CDF')):
+            # netCDF3 file images are handled by scipy
+            pass
+        elif isinstance(filename_or_obj, basestring):
+            filename_or_obj = _normalize_path(filename_or_obj)
+
         if filename_or_obj.endswith('.gz'):
             if engine is not None and engine != 'scipy':
                 raise ValueError('can only read gzipped netCDF files with '
@@ -526,8 +544,10 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
             raise ValueError('invalid engine for creating bytes with '
                              'to_netcdf: %r. Only the default engine '
                              "or engine='scipy' is supported" % engine)
-    elif engine is None:
-        engine = _get_default_engine(path)
+    else:
+        if engine is None:
+            engine = _get_default_engine(path)
+        path = _normalize_path(path)
 
     # validate Dataset keys, DataArray names, and attr keys/values
     _validate_dataset_names(dataset)
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -127,7 +127,7 @@ def _protect_dataset_variables_inplace(dataset, cache):
             data = indexing.CopyOnWriteArray(variable._data)
             if cache:
                 data = indexing.MemoryCachedArray(data)
-            variable._data = data
+            variable.data = data
 
 
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
@@ -212,7 +212,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         decode_coords = False
 
     if cache is None:
-        cache = chunks is not None
+        cache = chunks is None
 
     def maybe_decode_store(store, lock=False):
         ds = conventions.decode_cf(
@@ -300,7 +300,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
 def open_dataarray(filename_or_obj, group=None, decode_cf=True,
                    mask_and_scale=True, decode_times=True,
                    concat_characters=True, decode_coords=True, engine=None,
-                   chunks=None, lock=None, drop_variables=None):
+                   chunks=None, lock=None, cache=None, drop_variables=None):
     """
     Opens an DataArray from a netCDF file containing a single data variable.
 
@@ -354,6 +354,13 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
         used when reading data from netCDF files with the netcdf4 and h5netcdf
         engines to avoid issues with concurrent access when using dask's
         multithreaded backend.
+    cache : bool, optional
+        If True, cache data loaded from the underlying datastore in memory as
+        NumPy arrays when accessed to avoid reading from the underlying data-
+        store multiple times. Defaults to True unless you specify the `chunks`
+        argument to use dask, in which case it defaults to False. Does not
+        change the behavior of coordinates corresponding to dimensions, which
+        always load their data from disk into a ``pandas.Index``.
     drop_variables: string or iterable, optional
         A variable or list of variables to exclude from being parsed from the
         dataset. This may be useful to drop variables with problems or
@@ -375,7 +382,7 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
     dataset = open_dataset(filename_or_obj, group, decode_cf,
                            mask_and_scale, decode_times,
                            concat_characters, decode_coords, engine,
-                           chunks, lock, drop_variables)
+                           chunks, lock, cache, drop_variables)
 
     if len(dataset.data_vars) != 1:
         raise ValueError('Given file dataset contains more than one data '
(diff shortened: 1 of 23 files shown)


xarray/backends/api.py
```diff
@@ -117,10 +120,20 @@ def check_attr(name, value):
             check_attr(k, v)
 
 
+def _protect_dataset_variables_inplace(dataset, cache):
+    for name, variable in dataset.variables.items():
+        if name not in variable.dims:
+            # no need to protect IndexVariable objects
+            data = indexing.CopyOnWriteArray(variable._data)
```
kynan: I ran a test where I create a `DataSet` from a custom data store which initializes `Variables` using dask arrays for `data`. In this case the dask arrays is still converted to an ndarray when accessing the `Variable`'s `data` property, since [it checks](https://github.com/shoyer/xarray/blob/49135f20190cd024dbc7148262026659cc36e955/xarray/core/variable.py#L271) is for a dask array type, however here the array is wrapped into a `CopyOnWriteArray`, which means `Variable.values` is called, which loads eagerly.

xarray/core/dataset.py
```diff
@@ -260,17 +260,12 @@ def load_store(cls, store, decoder=None):
         return obj
 
     def __getstate__(self):
-        """Load data in-memory before pickling (except for Dask data)"""
-        for v in self.variables.values():
-            if not isinstance(v.data, dask_array_type):
-                v.load()
+        """Get this object's state for pickling"""
+        # we need a custom method to avoid
```
kynan: Incomplete

xarray/backends/api.py
shoyer: Indeed, you would need to use `cache=False` in such a case.

Xarray's decoding logic in `conventions.py` uses it's own array objects instead of dask arrays (for reasons I could get into), which unfortunately makes using dask.array objects to produce variables in a custom data store non-ideal. The problem is that the graphs from such dask arrays don't get linked up into xarray, which means that even if you rechunk the arrays in the xarray Dataset, they still get executed separately by dask. Duck typing for dask objects would probably help here (https://github.com/dask/dask/pull/1068) .

shoyer:
> In the long run I think it would be more robust to check for attributes (duck type style) rather than types in the various places.

Indeed, in particular I'm not very happy with the `isinstance` check for `indexing.MemoryCachedArray` in `Variable.copy()` -- it's rather poor separation of concerns.

It exists so that `variable.compute()` does not cache data in-memory on `variable` but only on the computed variable. Otherwise, there's basically no point to the separate compute method: if you use `cache=True`, you are stuck with caching on the original object. Likewise, it ensures that `.copy()` creates an array with a new cache, which is consistent with the current behavior of `.copy()`.

As for type checking for dask arrays in `.data`: yes, it would be nice to have a well defined array interface layer that other array types could plug into. That would entail a significant amount of further work, however.

xarray/core/dataset.py
```diff
@@ -260,17 +260,12 @@ def load_store(cls, store, decoder=None):
...
 
         # self.__dict__ is the default pickle object, we don't need to
         # implement our own __setstate__ method to make pickle work
-        state = self.__dict__.copy()
-        # throw away any references to datastores in the pickle
-        state['_file_obj'] = None
-        return state
+        return self.__dict__
```
crusaderky: Why don't you remove this method entirely?

xarray/core/utils.py
```diff
@@ -420,6 +420,13 @@ def __repr__(self):
         return '%s(array=%r)' % (type(self).__name__, self.array)
 
 
+class NoPickleMixin(object):
+    def __getstate__(self):
+        raise TypeError(
+            'cannot pickle objects of type %r: call .compute() or .load() '
+            'to load data into memory first.' % type(self))
+
+
```
crusaderky: I'm really not a fan of this design.
A better approach to pickling file-based objects is to save the file path on \_\_getstate\_\_ and open a new file descriptor on \_\_setstate\_\_. Relative vs. absolute paths should be preserved.
Of course there's no guarantee that the file will be there when you unpickle, or that it will be identical to the one that you pickled - but this is a caveat that's easily explained in the documentation.
The benefits of this are huge when you work interactively in Jupyter with plugins that pickle all session variables at once.

crusaderky: This actually hurts my xarray-based financial simulation project at my company.  I start with a DataArray loaded from disk through dask, and then heavily manipulate it. I end up with ~3000 dask-based xarrays, each with up to 80,000 key:value pairs. At that point, since just _generating_ the final objects took 5-10 minutes (no actual calculations yet), I pickle them, save them to disk and send them over the network.

Insofar I've had to load bespoke data from disk (numpy.fromfile manually wrapped with dask). In the future, however, I'd love to move to NetCDF for my on-disk storage... in other words I would have a huge dask tree with, at the very bottom of it, a dataset created by open_dataset(). And I would need to pickle the whole thing, with the obvious caveat that all your NetCDF files have to be in the same place, or pickle.load with fail with FileNotFoundError.

shoyer:
I removed the custom pickle override on `Dataset`/`DataArray` -- the issue I was working around was actually a indirect manifestation of bug on `IndexVariable.load()` (introduced in this PR).

xarray/core/utils.py
crusaderky: Sorry  I wasn't too clear - this change isn't a regression as much as my project is concerned - it just keeps the netcdf backend unusable. I agree that a proper pickling support for file-based stores can be added later on.

shoyer:
I added pickle support to DataStores. This *should* solve the basic serialization issue for dask.distributed (#798), but does not yet resolve the "too many open files" issue.

shoyer:
This isn't yet working with dask multiprocessing for reading a netCDF4 file with in-memory compression. I'm not quite sure why:
```
In [5]: from multiprocessing.pool import Pool

In [7]: ds = xr.open_dataset('big-random.nc', lock=False, chunks={'x': 2500})

In [8]: dask.set_options(pool=Pool(4))
Out[8]: <dask.context.set_options at 0x1087c3898>

In [9]: %time ds.sum().compute()
<ipython-input-9-4c43356c48db> in <module>()
----> 1 get_ipython().magic('time ds.sum().compute()')

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/IPython/core/interactiveshell.py in magic(self, arg_s)
   2156         magic_name, _, magic_arg_s = arg_s.partition(' ')
   2157         magic_name = magic_name.lstrip(prefilter.ESC_MAGIC)
-> 2158         return self.run_line_magic(magic_name, magic_arg_s)
   2159
   2160     #-------------------------------------------------------------------------

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line)
   2077                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2078             with self.builtin_trap:
-> 2079                 result = fn(*args,**kwargs)
   2080             return result
   2081

<decorator-gen-59> in time(self, line, cell, local_ns)

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/IPython/core/magic.py in <lambda>(f, *a, **k)
    186     # but it's overkill for just that one bit of state.
    187     def magic_deco(arg):
--> 188         call = lambda f, *a, **k: f(*a, **k)
    189
    190         if callable(arg):

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/IPython/core/magics/execution.py in time(self, line, cell, local_ns)
   1174         if mode=='eval':
   1175             st = clock2()
-> 1176             out = eval(code, glob, local_ns)
   1177             end = clock2()
   1178         else:

<timed eval> in <module>()

/Users/shoyer/dev/xarray/xarray/core/dataset.py in compute(self)
    348         """
    349         new = self.copy(deep=False)
--> 350         return new.load()
    351
    352     @classmethod

/Users/shoyer/dev/xarray/xarray/core/dataset.py in load(self)
    325
    326             # evaluate all the dask arrays simultaneously
--> 327             evaluated_data = da.compute(*lazy_data.values())
    328
    329             for k, data in zip(lazy_data, evaluated_data):

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/base.py in compute(*args, **kwargs)
    176         dsk = merge(var.dask for var in variables)
    177     keys = [var._keys() for var in variables]
--> 178     results = get(dsk, keys, **kwargs)
    179
    180     results_iter = iter(results)

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/threaded.py in get(dsk, result, cache, num_workers, **kwargs)
     67     results = get_async(pool.apply_async, len(pool._pool), dsk, result,
     68                         cache=cache, get_id=_thread_get_id,
---> 69                         **kwargs)
     70
     71     # Cleanup pools associated to dead threads

/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py in get_async(apply_async, num_workers, dsk, result, cache, get_id, raise_on_exception, rerun_exceptions_locally, callbacks, dumps, loads, **kwargs)
    500                     _execute_task(task, data)  # Re-execute locally
    501                 else:
--> 502                     raise(remote_exception(res, tb))
    503             state['cache'][key] = res
    504             finish_task(dsk, key, state, results, keyorder.get)

RuntimeError: NetCDF: HDF error

Traceback
---------
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py", line 268, in execute_task
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py", line 248, in _execute_task
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py", line 248, in <listcomp>
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py", line 245, in _execute_task
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py", line 245, in <listcomp>
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/async.py", line 249, in _execute_task
    return func(*args2)
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/dask/array/core.py", line 51, in getarray
    c = np.asarray(c)
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/numpy/core/numeric.py", line 482, in asarray
  File "/Users/shoyer/dev/xarray/xarray/core/indexing.py", line 417, in __array__
    return np.asarray(self.array, dtype=dtype)
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/numpy/core/numeric.py", line 482, in asarray
  File "/Users/shoyer/dev/xarray/xarray/core/indexing.py", line 392, in __array__
    return np.asarray(array[self.key], dtype=None)
  File "/Users/shoyer/conda/envs/xarray-dev/lib/python3.5/site-packages/numpy/core/numeric.py", line 482, in asarray
  File "/Users/shoyer/dev/xarray/xarray/core/indexing.py", line 392, in __array__
    return np.asarray(array[self.key], dtype=None)
  File "/Users/shoyer/dev/xarray/xarray/backends/netCDF4_.py", line 56, in __getitem__
    data = getitem(self.array, key)
  File "netCDF4/_netCDF4.pyx", line 3695, in netCDF4._netCDF4.Variable.__getitem__ (netCDF4/_netCDF4.c:37914)
  File "netCDF4/_netCDF4.pyx", line 4376, in netCDF4._netCDF4.Variable._get (netCDF4/_netCDF4.c:47134)
```

xarray/backends/h5netcdf_.py
```diff
@@ -37,24 +37,32 @@ def _read_attributes(h5netcdf_var):
...
     def __init__(self, filename, mode='r', format=None, group=None,
                  writer=None):
-        import h5netcdf.legacyapi
         if format not in [None, 'NETCDF4']:
             raise ValueError('invalid format for h5netcdf backend')
-        ds = h5netcdf.legacyapi.Dataset(filename, mode=mode)
-        with close_on_error(ds):
-            self.ds = _nc4_group(ds, group, mode)
+        opener = functools.partial(_open_h5netcdf_group, filename, mode=mode,
+                                   group=group)
```
mrocklin: Might be worth noting that this is only cloud-picklable, not stdlib pickleable.

xarray/test/test_distributed.py
```diff
@@ -0,0 +1,36 @@
...
+        with distributed.Client(('127.0.0.1', s['port']), loop=loop):
+            original = create_test_data()
+            with create_tmp_file() as filename:
+                original.to_netcdf(filename, engine=engine)
+                # TODO: should be able to serialize locks
+                restored = xr.open_dataset(filename, chunks=3, lock=False,
+                                           engine=engine)
+                assert isinstance(restored.var1.data, da.Array)
+                computed = restored.compute()
+                assert_dataset_allclose(original, computed)
```
mrocklin: This test looks good to me.

mrocklin:
Does your failure work with the following spawning pool in Python 3?

```python
In [1]: import multiprocessing

In [2]: ctx = multiprocessing.get_context('spawn')

In [3]: ctx.Pool(4)
Out[3]: <multiprocessing.pool.Pool at 0x7fec70afca20>
```

xarray/backends/h5netcdf_.py
shoyer: Can you clarify why this won't work with stdlib pickle? Is the issue doing the `h5netcdf` import inside the function definition? My understand was that `functools.partial` is pickleable.

shoyer:
> Does your failure work with the following spawning pool in Python 3?

Why, yes it does -- and it shows a nice speedup, as well! What was I missing here?

xarray/backends/h5netcdf_.py
mrocklin: Ah, you're right.  I'm surprised:

```python
In [1]: from operator import add

In [2]: from functools import partial

In [3]: from pickle import dumps, loads

In [4]: loads(dumps(partial(add, 1)))
Out[4]: functools.partial(<built-in function add>, 1)

In [5]: loads(dumps(partial(add, 1)))(2)
Out[5]: 3
```

mrocklin:
> Why, yes it does -- and it shows a nice speedup, as well! What was I missing here?

Spawn is only available in Python 3, so it's not a full solution.  Something isn't fork-safe, possibly something within the HDF5 library?

You might also want to try `forkserver` and look at this semi-related PR https://github.com/dask/distributed/pull/687

xarray/test/test_backends.py
```diff
@@ -163,24 +162,52 @@ def test_dataset_compute(self):
...
 
+    def test_pickle(self):
+        expected = Dataset({'foo': ('x', [42])})
+        with self.roundtrip(expected) as roundtripped:
+            raw_pickle = pickle.dumps(roundtripped)
+            # windows doesn't like opening the same file twice
+            roundtripped.close()
+            unpickled_ds = pickle.loads(raw_pickle)
+            self.assertDatasetIdentical(expected, unpickled_ds)
+            unpickled_ds.close()
```
shoyer: This test is only failing on Windows in `NetCDF4ViaDaskDataTest` now.

I'm not quite sure how we are missing clean up  there. In that case, the file/DataStore object is ending up in the dask graph of the arrays in `roundtripped`. We do close the file with `.close()`, but maybe those references to closed files hanging around are causing trouble?

shoyer:
OK, I'm ready to give up on the remaining test failures and merge this anyways (marking them as expected failures). They are specific to our test suite and for Windows only, due to the inability to delete files that are not closed.

If these manifest themselves as issues for real users, I am happy to revisit, especially if someone who uses Windows can help debug. The 5 minute feedback cycle of pushing a commit and then seeing what Appveyor says is too painful.

mrocklin:
@shoyer https://github.com/dask/dask/blob/master/dask/utils.py#L68-L84

shoyer:
@mrocklin OK, so one option is to just ignore the permission errors and not remove the files on Windows. But is it really better to make the test suite leak temp files?

mrocklin:
I agree that it's not great.  This was more a show of solidarity that we've also run into this same issue and had to resort to similar hacks.

shoyer:
I decided that between the choices of not running these tests on Windows and leaking a few temp files, I would rather leak some temp files. So that's exactly what I've done in the latest commit, for explicitly whitelisted tests.

shoyer:
@kynan @crusaderky Do you have concerns about merging this in the current state?

crusaderky:
All looks good, go on

kynan:
No objections, go ahead!

shoyer:
OK, in it goes!

mangecoeur:
I'm trying out the latest code to subset a set of netcdf4 files with dask.multiprocessing using `set_options(get=dask.multiprocessing.get)` but I'm still getting `TypeError: can't pickle _thread.lock objects`  - this expect or there something specific I need to do to make it work?

shoyer:
@mangecoeur You still need to use `lock=False` (or `lock=dask.utils.SerializableLock()` with the dev version of dask) and use a spawning process pool (https://github.com/pydata/xarray/pull/1128#issuecomment-261936849).

The former should be updated internally, and the later should be a documentation note.

mangecoeur:
@shoyer thanks, with a little testing it seems `lock=False` is fine (so don't automatically need   dask dev for `lock=dask.utils.SerializableLock()`). Using spawning pool is necessary, just doesn't work without. Also looks like using dask distributed ipython backend works fine (works similar to spawn pool in that the worker engines aren't forked but kinda live in their own little world) - this is really nice because ipython in turn has good support for HPC systems (SGE batch scheduling + MPI for process handling).

PR #1131, opened 2016-11-17
Fix #1040: diff dim argument should be optional

### LINKED ISSUE #1040: DataArray.diff dim argument should be optional as is in docstring
The dosctring of `DataArray.diff` lists the `dim` arg as optional, [but it isn't](https://github.com/pydata/xarray/blob/fbb4f0618eade20981bd5cfb9771b82fd88a8db5/xarray/core/dataarray.py#L1468). IMHO it should indeed be optional as it is quite convenient to apply `diff` to 1D signals without specifying the dimension.

smartass101:
* {Dataset,DataArray}.diff dim argument defaults to last dimension

* add test cases

* add changelog

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2111,13 +2111,14 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         for v in other.variables:
             self.variables[v].attrs = other.variables[v].attrs
 
-    def diff(self, dim, n=1, label='upper'):
+    def diff(self, dim=None, n=1, label='upper'):
         """Calculate the n-th order discrete difference along given axis.
 
         Parameters
         ----------
         dim : str, optional
-            Dimension over which to calculate the finite difference.
+            Dimension over which to calculate the finite difference. Defaults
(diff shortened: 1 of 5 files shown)


xarray/core/dataset.py
```diff
@@ -2156,6 +2157,10 @@ def diff(self, dim, n=1, label='upper'):
             raise ValueError('order `n` must be non-negative but got {0}'
                              ''.format(n))
 
+        # get default last dim if not specified
+        if dim is None:
+            dim = self.dims[-1]
```
shoyer: Sure.

jhamman:
In reviewing this PR, I'm not really a fan of special casing the 1-D array in the way described above. These sorts of decisions lead to unexpected/confusing behavior when scaling up analysis. Obviously there is a tradeoff here as we diverge from how the `numpy.diff` function works but I think that is something we'd prefer to live with. I think this falls under the _explicit is better than implicit_ idea...

max-sixty:
Closing as stale. Please feel free to reopen!

PR #1134, opened 2016-11-26
Further attempt to get netCDF4 working on RTD

fmaussion:
The idea came from here: https://github.com/Unidata/netcdf4-python/issues/574#issuecomment-235435628

I could test it on another repo, so I'm confident that this should work here too. Obviously, this is a temporary solution. 

hopefully fixes https://github.com/pydata/xarray/issues/1106

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -12,4 +12,5 @@ dependencies:
   - ipython=4.0.1
   - sphinx
   - netCDF4
+  - hdf4  # temporary install to get netCD4 working (GH1106)
   - cartopy


shoyer:
Okay, let's try it.

fmaussion:
Yep, that did the trick: http://xarray.pydata.org/en/latest/examples/multidimensional-coords.html

fmaussion:
Note that probably some other package than hdf4 might do it, as long as it ships that ``libmfhdf.so.0`` library... 

CC @ocefpaf

ocefpaf:
TL;DR you may consider making this change  permanent.

@fmaussion the problem is that latest conda-forge's hdf4 does ship with libmfhdf.so.0 but either defaults' version does not or an old version in conda-forge is broken. (I am traveling but I'll confirm that as soon as I get back.)

Adding hdf4 at the env creation  forces the solver to get the latest version/build number and prevented the downgrade in the subsequent conda install call. So you should probably leave hdf4 there even if it is over specifying deps because the issue may resurface.

I say that b/c, if the bad version is in conda-forge we can remove it. But if it is in defaults we have no control.

Issue #1135, opened 2016-11-26
DOCS: broken "What's new" 


fmaussion:
http://xarray.pydata.org/en/latest/whats-new.html  See the examples at the bottom.

These are all old examples relying on the "xray" package. We can either remove these examples (my suggestion) or update them to use array. 

As a general rule, I think that the what's new page shouldn't contain any code, at least not code that has to be run by RTD

fmaussion:
... there's a third solution. make a hidden ``import xarray as xray`` and hope that the examples still work.

But again, this would mean that we are supposed to maintain these examples too, although they were meant to illustrate a change made to early versions of xray/xarray

fmaussion:
After looking at the stable doc version (http://xarray.pydata.org/en/stable/whats-new.html) all examples seem to work, so for now we could cope with the ``import xarray as xray`` solution implemented in https://github.com/pydata/xarray/issues/1136

fmaussion:
Closed via https://github.com/pydata/xarray/pull/1136

PR #1136, opened 2016-11-26
WIP: remove some refs to xray

fmaussion:
There are some uses of xray that needed to be fixed.

WIP: merge after you decide what to do with https://github.com/pydata/xarray/issues/1135

Also, I'd be in favor of removing the two notebooks in https://github.com/pydata/xarray/tree/master/examples since they are also available in the docs (http://xarray.pydata.org/en/latest/examples.html). The example on RTD have the disadvantage that it's less easy to do copy/paste with them, but notebooks are more error prone since they need to be maintained externally.


shoyer:
Thanks! I agree we should remove the redundant notebooks (unless we switch to using notebooks in the doc build).

PR #1137, opened 2016-11-27
Pin package versions on RTD

fmaussion:
Now that we seem to have everything working on RTD, re-pin package versions

(follow-up to https://github.com/pydata/xarray/pull/1101)


shoyer:
Thanks!

PR #1139, opened 2016-11-28
Unpin python on RTD

fmaussion:
Pinning python to 3.5 introduced new [problems](https://readthedocs.org/projects/xray/builds/4723418/).

Despite the default python being 3 [here](https://github.com/pydata/xarray/blob/master/readthedocs.yml#L4) and also in our RTD preferences (dashboard->advanced), RTD seems to still use py2 per default...

Anyway, going back to py2 is the easiest for now, I might come back to this when I have more time.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python=3.5
+  - python
   - numpy=1.11.2
   - pandas=0.19.1
   - numpydoc=0.6.0


shoyer:
Indeed, I think `execfile` does not exist on Python 3.

For debugging the doc build on Python 3, I would definitely recommend building locally (i.e., with `make html` from the docs directory).

fmaussion:
> For debugging the doc build on Python 3, I would definitely recommend building locally (i.e., with make html from the docs directory).

I did, but my local builds still complete (with many warning but now errors: execfile seems to be simply ignored...).

I'll remove the execfile, but I'm still skeptical that this will work (py3 support is relatively recent in py3)

PR #1144, opened 2016-11-30
Update ipython on RTD

fmaussion:
Another attempt to solve a cryptic error: https://readthedocs.org/projects/xray/builds/4727274/

But this time we are getting closer! I've found a report on ipython (https://github.com/ipython/ipython/issues/8850), an I've been able to reproduce it locally and solve it by updating to ipython 5


fmaussion:
Yay, this is building now.

Issue #1138, opened 2016-11-27
.rolling(...).count() ?


MaximilianR:
Should `.count()` be available on `.rolling` objects?

```python
>       sample_count = array.rolling(**windows).count() - 1.0
E       AttributeError: 'DataArrayRolling' object has no attribute 'count'
```

shoyer:
Yes, I don't see why not?

MaximilianR:
OK will try and do a PR this week. Got a few others to get in first but this should be short

MaximilianR:
@shoyer should this do `.notnull().rolling().sum()`, or is there a specific rolling count function out there that would be better?

shoyer:
That looks like about the best we could do.

jhamman:
@MaximilianR - that is the best we can do.`bottleneck` doesn't have a `move_count` function so we will need to operate on the rolling object ourselves.

PR #1145, opened 2016-11-30
Rolling count

MaximilianR:
closes https://github.com/pydata/xarray/issues/1138

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,3 +3,6 @@ universal = 1
 
 [tool:pytest]
 python_files=test_*.py
+
+[flake8]
+max-line-length=79
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -314,6 +314,13 @@ def inject_bottleneck_rolling_methods(cls):
             name=func.__name__, da_or_ds='DataArray')
         setattr(cls, name, func)
 
+    # bottleneck doesn't offer rolling_count, so we construct it ourselves
+    func = rolling_count
+    func.__name__ = 'count'
+    func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(
+        name=func.__name__)
+    setattr(cls, 'count', func)
+
     # bottleneck rolling methods
     if has_bottleneck:
         # TODO: Bump the required version of bottlneck to 1.1 and remove all
@@ -364,5 +371,6 @@ def inject_datasetrolling_methods(cls):
     # bottleneck doesn't offer rolling_count, so we construct it ourselves
     func = rolling_count
     func.__name__ = 'count'
-    func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(name=func.__name__)
+    func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(
+        name=func.__name__)
     setattr(cls, 'count', func)
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -198,6 +198,15 @@ def _func_slash_method_wrapper(f, name=None):
     return func
 
 
+def rolling_count(rolling):
+
+    not_null = rolling.obj.notnull()
+    instance_attr_dict = {'center': rolling.center,
+                          'min_periods': rolling.min_periods,
+                          rolling.dim: rolling.window}
+    rolling_count = not_null.rolling(**instance_attr_dict).sum()
+    return rolling_count
+
 def inject_reduce_methods(cls):
     methods = ([(name, getattr(duck_array_ops, 'array_%s' % name), False)
                 for name in REDUCE_METHODS] +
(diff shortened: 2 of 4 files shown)


xarray/core/ops.py
```diff
@@ -277,6 +278,15 @@ def count(data, axis=None):
     """
     return sum(~isnull(data), axis=axis)
 
+def rolling_count(rolling):
+
+    not_null = rolling.obj.notnull()
+    instance_attr_dict = {'center': rolling.center,
+                          'min_periods': rolling.min_periods,
+                          rolling.dim: rolling.window}
+    rolling_count = not_null.rolling(**instance_attr_dict).sum()
```
MaximilianR: I couldn't find a more general / elegant way of doing this, with something like `._shallow_copy`

xarray/test/test_dataarray.py
```diff
@@ -1549,20 +1549,31 @@ def test_rolling_wrapped_bottleneck(self):
...
         rolling_obj = da.rolling(time=7, center=False)
-        for name in ('sum', 'mean', 'std', 'min', 'max', 'median'):
+        for name in ('sum', 'mean', 'std', 'min', 'max', 'median', 'count'):
             actual = getattr(rolling_obj, name)()['time']
             self.assertDataArrayEqual(actual, da['time'])
 
         # Using center=True
         rolling_obj = da.rolling(time=7, center=True)
-        for name in ('sum', 'mean', 'std', 'min', 'max', 'median'):
+        for name in ('sum', 'mean', 'std', 'min', 'max', 'median', 'count'):
```
MaximilianR: These loops are begging for pytest fixtures...

xarray/core/ops.py
```diff
@@ -517,3 +527,9 @@ def inject_bottleneck_rolling_methods(cls):
         func.__name__ = 'median'
         func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(name=func.__name__)
         setattr(cls, 'median', func)
+
+        # bottleneck doesn't offer rolling_count, so we construct it ourselves
+        func = rolling_count
+        func.__name__ = 'count'
+        func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(name=func.__name__)
+        setattr(cls, 'count', func)
```
jhamman: Is this indented too far.  Don't you want this method if we don't have bottleneck as well? I think this is causing your current test failure.

xarray/test/test_dataarray.py
jhamman: Fine by me.

setup.cfg
```diff
@@ -1,5 +1,8 @@
 [wheel]
 universal = 1
 
-[pytest]
+[tool:pytest]
 python_files=test_*.py
+
+[flake8]
+max-line-length=79
```
MaximilianR: Open to feedback here - either a different number or not having a limit at all

MaximilianR:
@jhamman updated

xarray/core/ops.py
shoyer: This looks good to me.

xarray/test/test_dataarray.py
```diff
@@ -1549,20 +1549,31 @@ def test_rolling_wrapped_bottleneck(self):
...
         with self.assertRaisesRegexp(ValueError, 'Rolling.median does not'):
             da.rolling(time=7, min_periods=1).median()
 
+    def test_rolling_count_correct(self):
+
+        da = self.make_rolling_example_array()
+        result = da.rolling(time=7, min_periods=1).count()
+        expected_rolling_count = np.r_[np.arange(1, 8), np.full((14,), 7)]
+        expected = xr.DataArray(
+            np.resize(expected_rolling_count, (4, 21)).T,
```
shoyer: Use `xarray.core.npcompat.broadcast_to`, which is just a back port of `np.broadcast_to`.

I'm honestly not sure how `np.resize` does broadcasting. It's a strange function -- very different from the low-level `resize` method.

jhamman: Can we not use `np.r_`? In my experience, its too confusing for its own good.

jhamman:
@MaximilianR - if you can resolve the conflicts on this branch, I'll give it a final review.

MaximilianR:
Any ideas why we aren't running standard CI?

MaximilianR:
I'm getting some errors running tests on master - is that well known? I can debug this end, but will avoid if others are owning

jhamman:
@MaximilianR - No idea what is going on here. It seems like github has lost the Travis hook for this PR. Maybe try closing and reopening it?

shoyer:
The open/close seems to have retrigged Travis. Looks like there's one failure for `test_rolling_count_correct`.

MaximilianR:
Yes will take a look, thanks

MaximilianR:
@jhamman @shoyer updated

xarray/core/ops.py
```diff
@@ -198,6 +198,24 @@ def func(self, *args, **kwargs):
     return func
 
 
+def rolling_count(rolling):
```
shoyer: This looks generally OK to me, but possibly a cleaner approach would be to make a function that works at the level numpy arrays instead, like bottleneck functions. I think this could be as simple as:
```
def move_count(array, window, min_count=0, axis=-1):
    not_nulls = duck_array_ops.notnull(array)
    return bottleneck.move_sum(not_nulls, window, min_count, axis=axis)
```

The only possibly awkward part about this is that the `min_count` parameter should be handled differently than the standard bottleneck reduce functions. So it it can't be easily consolidated into the existing wrapper for bottleneck functions, don't worry about it.

MaximilianR: The problem with applying a `move_sum` on the `not_nulls` is that there are no `null`s left in `not_nulls` (they are `False`), and so `min_count` won't work correctly. 

But It does feel like this is a bit awkward atm, very open to other suggestions

shoyer:
Fair enough. Let's just stick with this then.

PR #1146, opened 2016-12-01
Minor test fixes

MaximilianR:
A couple of tests weren't actually testing anything.

I may add things onto this PR if I see them


shoyer:
good catch -- thanks!

Issue #754, opened 2016-02-09
Expose testing methods


spencerahill:
Similar to numpy's [numpy.testing](http://docs.scipy.org/doc/numpy/reference/routines.testing.html) module, I would find it useful for xarray to expose some of its testing methods, particularly those in its base [TestCase](https://github.com/pydata/xarray/blob/master/xarray/test/__init__.py#L129).

I imagine other folks whose packages and/or modules derive heavily from xarray would find this useful too.  For now I am just copy-pasting xarray's code for these methods into my testing modules.

jhamman:
:+1:

shoyer:
Most of these are as simple as `assert d1.identical(d2)` or `assert d1.equals(d2)` (see [here](http://xarray.pydata.org/en/stable/combining.html#equals-and-identical) for more details on these methods). I'd be OK with adding these methods in a dedicated testing module, though.

spencerahill:
Indeed.  I'm happy to do this sometime (eventually...), but if somebody ends up with a pressing need for this in the meantime, go for it.  Thanks!

jhamman:
@spencerahill - have you made any progress here?  If not we (@dgergel) may be able to help out.

spencerahill:
Thanks for the ping.  Sorry, I haven't.  I'm actually submitting my PhD thesis in ~2 months and then defending it this summer, so I won't be able to help much until August or September.  So by all means, go for it.

fmaussion:
I'd be willing to prepare a PR, if this is still actual. Today I needed the *AllClose methods for my tests, I think it would be great to have these methods exposed.

I think I have an idea on how to proceed, but maybe I am underestimating the required amount of work.

Numpy for example has the `np.allclose` and `np.testing.assert_allclose` functions. Should xarray also have a `xr.allclose` function or are we happy with a `xr.testing.assert_allclose` alone? Also, the current xarray implementation has a different function for variables, datasets and coordinates: could these be merged in a generic `assert_allclose` which takes decisions on what to do based on the input type?

PR #1147, opened 2016-12-01
Pytest assert functions

MaximilianR:
Is this a reasonable function to use with py.test? In place of the inheritance `.assertVariableEqual` etc.

- Do we want separate functions for each class?
- There's a comment `or just switch to py.test and add an appropriate hook.` - is that different to what's here?
- This still needs to be completed, with additional functions for `assert _identical` and `assert_approx_equal`.
- Do we want more description for the failure, above just printing the arguments? That can also be added later


xarray/test/__init__.py
```diff
@@ -215,6 +215,24 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
...
+    import xarray as xr
+    ___tracebackhide__ = True
+    assert type(a) == type(b)
+    if isinstance(a, xr.DataArray):
+        assert_equal(a.data_vars, b.data_vars)
+        assert_equal(a.coords, b.coords)
+    elif isinstance(a, xr.Variable):
+        assert as_variable(a).equals(b), (a, b)
+    elif isinstance(a, xr.Dataset):
+        assert a.equals(b), (a, b)
```
shoyer: I think this (`assert a.equals(b), (a, b)`) would actually probably suffice for all of `Dataset`, `DataArray` and `Variable`.

shoyer:
> Do we want separate functions for each class?

That seems like overkill -- I think one is fine.

> There's a comment or just switch to py.test and add an appropriate hook. - is that different to what's here?

I was thinking of something like [how python lets you override the message for `assert a == b`](http://doc.pytest.org/en/latest/assert.html#defining-your-own-assertion-comparison). But it actually looks like there is no suitable hook for custom explanations when using methods. `__traceback__ = False`, like you do here, is about the best you can do.

> Do we want more description for the failure, above just printing the arguments? That can also be added later

It's nice to at least print the objects on separate lines, e.g., using something like `'%r\n%r' % (a, b)` for the assert message. Otherwise multi-line reprs end up with the start of the second object on the same line as the first.

MaximilianR:
Updated.

Is it worth 'testing the tests'? I couldn't find tests for the existing functions

xarray/test/__init__.py
```diff
@@ -215,6 +215,56 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
...
+    ___tracebackhide__ = True
+    assert type(a) == type(b)
+    if isinstance(a, xr.DataArray):
+        assert_xarray_close(a.variable, b.variable, rtol=rtol, atol=atol)
+        assert_xarray_equal(a.coords, b.coords)
+    elif isinstance(a, xr.Dataset):
+        assert sorted(a, key=str) == sorted(a, key=str)
+        assert_xarray_equal(a.coords, b.coords)
+        [assert_xarray_close(
+            a.variables[k], b.variables[k], rtol=1e-05, atol=1e-08)
```
shoyer: use `rtol=rtol, atol=atol` here

xarray/test/__init__.py
```diff
@@ -215,6 +215,56 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
...
 
+def assert_xarray_equal(a, b):
+    import xarray as xr
+    ___tracebackhide__ = True
+    assert type(a) == type(b)
+    if isinstance(a, xr.Dataset):
+        assert_xarray_equal(a.data_vars, b.data_vars)
+        assert_xarray_equal(a.coords, b.coords)
+    elif isinstance(a, (xr.Variable, xr.DataArray, xr.Coordinate)):
+        assert a.equals(b), '{}/n{}'.format(a, b)
```
shoyer: use `\n` not `/n` here and below

xarray/test/__init__.py
```diff
@@ -215,6 +215,56 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
...
         self.assertCoordinatesEqual(ar1, ar2)
 
+def assert_xarray_equal(a, b):
+    import xarray as xr
+    ___tracebackhide__ = True
+    assert type(a) == type(b)
+    if isinstance(a, xr.Dataset):
+        assert_xarray_equal(a.data_vars, b.data_vars)
+        assert_xarray_equal(a.coords, b.coords)
+    elif isinstance(a, (xr.Variable, xr.DataArray, xr.Coordinate)):
```
shoyer: `Coordinate` (actually `IndexVariable`, now) is a `Variable` subclass, so you don't need to call it out separately

xarray/test/__init__.py
```diff
@@ -215,6 +215,56 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
         self.assertVariableAllClose(ar1, ar2, rtol=rtol, atol=atol)
         self.assertCoordinatesEqual(ar1, ar2)
 
+def assert_xarray_equal(a, b):
+    import xarray as xr
+    ___tracebackhide__ = True
+    assert type(a) == type(b)
+    if isinstance(a, xr.Dataset):
```
shoyer: I would just use `.equals` for `Dataset`, as well -- no need to handle data_vars and coord separately

xarray/test/__init__.py
```diff
@@ -215,6 +215,56 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
...
+        assert_xarray_equal(a.data_vars, b.data_vars)
+        assert_xarray_equal(a.coords, b.coords)
+    elif isinstance(a, (xr.Variable, xr.DataArray, xr.Coordinate)):
+        assert a.equals(b), '{}/n{}'.format(a, b)
+    elif isinstance(a, xr.core.coordinates.AbstractCoordinates):
```
shoyer: I'm not sure we really need need this case

MaximilianR:
Thanks for the feedback and apologies for the silly errors. Updated.

xarray/test/__init__.py
```diff
@@ -215,6 +215,53 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
         self.assertVariableAllClose(ar1, ar2, rtol=rtol, atol=atol)
         self.assertCoordinatesEqual(ar1, ar2)
 
+def assert_xarray_equal(a, b):
+    import xarray as xr
+    ___tracebackhide__ = True
+    assert type(a) == type(b)
+    if isinstance(a, (xr.Variable, xr.DataArray, xr.Dataset)):
+        assert a.equals(b), '{}\n{}'.format(a, b)
+    elif isinstance(a, xr.core.coordinates.AbstractCoordinates):
```
MaximilianR: I think this is needed for coordinates - let me know if there's a better way

MaximilianR:
@shoyer green

shoyer:
This looks good to me. The last thing I would do is switch the existing test methods like `assertDatasetEqual` to use these functions, just to reduce the amount of redundant code (and also test your new functions more extensively).

MaximilianR:
> The last thing I would do is switch the existing test methods like assertDatasetEqual to use these functions, just to reduce the amount of redundant code (and also test your new functions more extensively).

All of tests? I think that's a fairly heavy lift for this PR. It also happens to be really tedious and unfortunately I don't think Find / Replace-able... If you feel strongly I'll plug in and go through though! 

Otherwise we can roll out over time as people write new tests

shoyer:
No no no... Just the definitions of the methods like assertDatasetEqual on
the bass xarray TestCase class, which you can now define as aliases to your
new functions.

xarray/test/__init__.py
```diff
@@ -215,6 +215,53 @@ def assertDataArrayAllClose(self, ar1, ar2, rtol=1e-05, atol=1e-08):
...
+    if isinstance(a, xr.DataArray):
+        assert a.name == b.name
+        assert_xarray_identical(a._to_temp_dataset(), b._to_temp_dataset())
+    elif isinstance(a, (xr.Dataset, xr.Variable)):
+        assert a.identical(b), '{}\n{}'.format(a, b)
+    else:
+        raise TypeError('{} not supported by assertion comparison'
+                        .format(type(a)))
+
+def assert_xarray_close(a, b, rtol=1e-05, atol=1e-08):
```
shoyer: It might be slightly more consistent to call this `allclose` rather than `close`.

shoyer:
You might also add these to the API docs page and what's new. I know there has been interest in public test functions for quite some time.

fmaussion:
>  I know there has been interest in public test functions for quite some time.

Yes, see also https://github.com/pydata/xarray/issues/754

xarray/test/__init__.py
MaximilianR: Why `allclose` but not `allequal`?

shoyer: This is the name of the functions in [`numpy.testing`](https://docs.scipy.org/doc/numpy/reference/routines.testing.html)

MaximilianR:
@shoyer do you have ideas for when the types are not the same? For example, many of the tests compare `Variable` to `IndexVariable`. Should we not check types? Add a `check_type=True` argument to the testing functions? Change the tests so they're compliant?

shoyer:
In most cases I would guess these are broken tests?

MaximilianR:
There are more than a dozen: https://travis-ci.org/pydata/xarray/jobs/184356287#L2611

Also some binary / unicode errors - let me know if you have context

xarray/test/__init__.py
```diff
@@ -154,67 +155,91 @@ def assertWarns(self, message):
...
+def assert_xarray_allclose(a, b, rtol=1e-05, atol=1e-08):
+    import xarray as xr
+    ___tracebackhide__ = True  # noqa: F841
+    assert type(a) == type(b)
+    if isinstance(a, xr.DataArray):
+        assert_xarray_allclose(a.variable, b.variable, rtol=rtol, atol=atol)
+        assert_xarray_equal(a.coords, b.coords)
+    elif isinstance(a, xr.Dataset):
+        assert sorted(a, key=str) == sorted(a, key=str)
+        assert_xarray_equal(a.coords, b.coords)
```
shoyer: I think this is why many (not all) of the new tests are failing (at least the unicode/bytes ones, because we have a hack that says unicode/bytes can still be "close" if they encode the same string). Previously we used `allclose` for all variables, now `coords` are required to be equal.

I can see why this makes sense, but maybe better to keep with what we have now or add a keyword argument to switch.

MaximilianR:
@shoyer I think I fixed all the tests. Running now

I didn't quite realize how many `.assertVariable...` tests were relying on coercion, so I ended up changing the tests in a lot of places. The test is explicit at the test site, at least.

There were a couple that I think are probably breaks - things like `.argsort` changing the type of a Variable. I marked those with comments but haven't fixed them - or we'll never get this in! Let me know if you have thoughts on any notes

MaximilianR:
Rebased on https://github.com/pydata/xarray/pull/1175, so that should be merged first if people agree

xarray/test/__init__.py
```diff
@@ -71,42 +72,43 @@
...
 
 
 def requires_pydap(test):
-    return test if has_pydap else unittest.skip('requires pydap.client')(test)
+    return test if has_pydap else pytest.mark.skip('requires pydap.client')(test)
 
 
 def requires_netCDF4(test):
-    return test if has_netCDF4 else unittest.skip('requires netCDF4')(test)
+    return test if has_netCDF4 else pytest.mark.skip('requires dask')(test)
```
fmaussion: dask -> netCDF4

fmaussion:
This is great! Before merge it would be good to add the new standalone ``assert_*`` functions to the API doc?

MaximilianR:
@fmaussion had a go - is that what you were thinking?

fmaussion:
Thanks @MaximilianR , this will be useful for all libraries downstream of xarray

spencerahill:
This is great.  At least in terms of the functionality I was looking for, I'd say this closes #754.

xarray/test/__init__.py
```diff
@@ -154,67 +154,91 @@ def assertWarns(self, message):
...
+        assert a.dims == b.dims
+        allclose = data_allclose_or_equiv(
+            a.values, b.values, rtol=rtol, atol=atol)
+        assert allclose, '{}\n{}'.format(a.values, b.values)
+    elif isinstance(a, xr.DataArray):
+        if hasattr(a, 'variables'):
+            [assert_xarray_allclose(a[k], b[k], rtol=rtol, atol=atol) for k in a.variables]
+        else:
+            # unsure if we should need this branch
+            # https://github.com/pydata/xarray/issues/1152
```
shoyer: Per my comment in #1152, DataArray does not have a `.variables` attribute. I think we should update this case to test both `.variable` and everything in `.coords.variables`.

xarray/test/__init__.py
```diff
@@ -154,67 +154,91 @@ def assertWarns(self, message):
...
+            assert_xarray_allclose(a.variable, b.variable)
+    elif isinstance(a, xr.Dataset):
+        assert sorted(a, key=str) == sorted(a, key=str)
+        [assert_xarray_allclose(a[k], b[k], rtol=rtol, atol=atol)
+         for k in list(a.variables) + list(a.coords)]
```
shoyer: Just a minor style point, but I find it a little unexpected to use a list comprehension when throwing away the result. I would write this with a normal `for` loop:
```
for k in list(a.variables) + list(a.coords):
    assert_xarray_allclose(a[k], b[k], rtol=rtol, atol=atol)
```

xarray/test/test_backends.py
```diff
@@ -912,7 +913,12 @@ def test_cross_engine_read_write_netcdf3(self):
                     for read_engine in valid_engines:
                         with open_dataset(tmp_file,
                                           engine=read_engine) as actual:
-                            self.assertDatasetAllClose(data, actual)
+                            # hack to allow test to work:
```
shoyer: This is fine for now, but we should look into this later because it looks like a legit bug to me.

It might be worth making a dedicated `VariablesDict` that is simply an OrderedDict with runtime type checking that verifies you can never put in anything other than a `Variable`. Or we should use pytype for static checks to ensure `Dataset._variables` and `DataArray._coords` are typed `typing.OrderedDict[Any, xarray.Variable]`.

xarray/test/test_variable.py
```diff
@@ -213,7 +213,10 @@ def test_pandas_period_index(self):
     def test_1d_math(self):
         x = 1.0 * np.arange(5)
         y = np.ones(5)
-        v = self.cls(['x'], x)
+
+        # should we need `.to_base_variable()`?
+        # probably a break that `+v` changes type?
+        v = self.cls(['x'], x).to_base_variable()
```
shoyer: I think this it's OK that any arithmetic converts `IndexVariable` to `Variable`. That seems pretty consistent to me.

shoyer: Instead of always converting `v` to a base variable, I would make another copy that is the base variable and using that for the expected variable. Thus these tests will still check that math with `IndexVariable` works even though it doesn't preserve the type.

xarray/test/test_variable.py
```diff
@@ -304,22 +309,23 @@ def test_equals_all_dtypes(self):
...
+        expected = self.cls('x', range(3), {'foo': 1}, {'bar': 2}).to_base_variable()
         for actual in [expected.T,
                        expected[...],
                        expected.squeeze(),
                        expected.isel(x=slice(None)),
                        expected.expand_dims({'x': 3}),
                        expected.copy(deep=True),
                        expected.copy(deep=False)]:
+
             self.assertVariableIdentical(expected, actual)
```
shoyer: Use `.to_base_variable()` in the assert here (on both sides). Otherwise this test doesn't have any coverage for `IndexVariable` objects.

xarray/test/test_variable.py
```diff
@@ -304,22 +309,23 @@ def test_equals_all_dtypes(self):
     def test_eq_all_dtypes(self):
         # ensure that we don't choke on comparisons for which numpy returns
         # scalars
-        expected = self.cls('x', 3 * [False])
+        expected = self.cls('x', 3 * [False]).to_base_variable()
```
shoyer: Note that you can just use `Variable('x', 3 * [False])` in this case and other ones like it, which is maybe a little clearer.

xarray/test/__init__.py
MaximilianR: 🤦‍♂️

xarray/test/test_backends.py
MaximilianR: Agreed it's a bug

MaximilianR:
@shoyer green!

xarray/test/test_variable.py
```diff
@@ -309,15 +310,15 @@ def test_equals_all_dtypes(self):
...
+        expected = Variable('x', 3 * [False])
         for v, _ in self.example_1d_objects():
             actual = 'z' == v
             self.assertVariableIdentical(expected, actual)
             actual = ~('z' != v)
             self.assertVariableIdentical(expected, actual)
 
     def test_encoding_preserved(self):
-        expected = self.cls('x', range(3), {'foo': 1}, {'bar': 2}).to_base_variable()
+        expected = Variable('x', range(3), {'foo': 1}, {'bar': 2})
```
shoyer: This should still be `self.cls`, not `Variable`.

MaximilianR:
@shoyer green!

Issue #1061, opened 2016-10-26
Consider using Sphinx gallery for examples


shoyer:
Results look pretty nice:
http://sphinx-gallery.readthedocs.io/en/latest/index.html

PR #1148, opened 2016-12-01
Expose options for axis sharing between subplots

hoonhout:
Typical usage for xarray.DataArray's for us is to store directional wave data (time, location, direction and frequency). In order to plot directional data stored in DataArray's it would be convenient to use a combination of FacetGrid and polar projections.

Currently, this is not possible with the xarray.plot modules as axes are hard-coded set to be shared among subplots, which conflicts with the fact that polar axes are by definition "box"-adjustable. Therefore we need the ability to turn off axes sharing.

The simple modification attached exposes the sharex/y options to the xarray.plot routines.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -104,6 +104,10 @@ Enhancements
 - New top-level functions :py:func:`~xarray.full_like`,
   :py:func:`~xarray.zeros_like`, and :py:func:`~xarray.ones_like`
   By `Guido Imperiale <https://github.com/crusaderky>`_.
+- Options for axes sharing between subplots are exposed to
+  :py:class:`FacetGrid` and :py:func:`~xarray.plot.plot`, so axes
+  sharing can be disabled for polar plots.
+  By `Bas Hoonhout <https://github.com/hoonhout>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/xarray/plot/facetgrid.py
+++ b/xarray/plot/facetgrid.py
@@ -72,7 +72,8 @@ class FacetGrid(object):
     """
 
     def __init__(self, data, col=None, row=None, col_wrap=None,
-                 aspect=1, size=3, subplot_kws=None):
+                 sharex=True, sharey=True, aspect=1, size=3,
+                 subplot_kws=None):
         """
         Parameters
         ----------
@@ -83,6 +84,10 @@ class FacetGrid(object):
             on separate facets in the grid.
         col_wrap : int, optional
             "Wrap" the column variable at this width, so that the column facets
+        sharex : bool, optional
+            If true, the facets will share x axes
+        sharey : bool, optional
+            If true, the facets will share y axes
         aspect : scalar, optional
             Aspect ratio of each facet, so that ``aspect * size`` gives the
             width of each facet in inches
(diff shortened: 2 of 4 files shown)


shoyer:
This seems pretty safe to me. I think this just needs a "what's new" entry, though tests are also always appreciated!

@fmaussion can you take a look?

fmaussion:
Yes, this looks quite straight forward! Tests would be good but I expect this to be difficult to do in the current testing framework, so no pressure on that.

What would be nice is to have an example of your usage of FacetGrid. You can either post it here or wait until I get to add https://github.com/pydata/xarray/issues/1061 to the docs!

hoonhout:
Here is an example of a random spectral plot that doesn't work without the suggested modification:

```python
import xarray
import numpy as np
from collections import OrderedDict
from datetime import datetime

# define dimensions
dims = OrderedDict([
    ('time',      [datetime(2016,12,1,1,0,0), datetime(2016,12,1,1,20,0)]),
    ('location',  np.arange(3)),
    ('frequency', np.linspace(.1, .5, 20)),
    ('direction', np.linspace(0., 2.*np.pi, 50)),
])

# create random data
E = np.random.rand(*[len(x) for x in dims.itervalues()])

# construct DataArray
darray = xarray.DataArray(E, dims=dims, coords=dims)

# plot data
darray.plot.pcolormesh(col='time', row='location',
                       subplot_kws=dict(projection='polar'),
                       sharex=False, sharey=False)
```

fmaussion:
Great example, thanks!

Actually this would make a good test since failing to set the new kwargs to False would raise an error. Maybe you could add this simple test [here](https://github.com/pydata/xarray/blob/master/xarray/test/test_plot.py#L918)?

ocefpaf:
Would it make sense to set the default to False? I believe false is used in mpl.

fmaussion:
This would likely brake existing code, so @shoyer has to decide upon this. The current defaults are probably inherited from [seaborn](http://seaborn.pydata.org/generated/seaborn.FacetGrid.html)

ocefpaf:
Ignore what I said. looking closer it makes sense to set the default to
True.

On Dec 2, 2016 12:18 PM, "Fabien Maussion" <notifications@github.com> wrote:

This would likely brake existing code, so @shoyer
<https://github.com/shoyer> has to decide upon this. The current defaults
are probably inherited from seaborn
<http://seaborn.pydata.org/generated/seaborn.FacetGrid.html>

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/pydata/xarray/pull/1148#issuecomment-264463656>, or mute
the thread

hoonhout:
Sure, I'll make my example into a test.

BTW, I would recommend to keep the default True.

Seaborn's default is indeed shared axes. This makes sense as a FacetGrid is intended to visualize single-source data.

MPL's default is indeed not to share axes. This also makes sense as MPL doesn't have any constraints on the data source.

fmaussion:
Thanks @hoonhout ! We're still missing an entry on [whats-new.rst](https://github.com/pydata/xarray/blob/master/doc/whats-new.rst) ;)

hoonhout:
I think this request is now ready to be pulled, right?
Otherwise I keep merging these what's new lines :-)

fmaussion:
Thanks @hoonhout !

PR #1153, opened 2016-12-05
Add drop=True argument to isel, sel and squeeze

### LINKED ISSUE #242: Add a "drop" option to squeeze

shoyer:
Fixes #242

This is useful for getting rid of extraneous scalar variables that arise from
indexing, and in particular will resolve an issue for optional indexes:
https://github.com/pydata/xarray/pull/1017#issuecomment-260777664


xarray/test/test_dataset.py
```diff
@@ -1750,6 +1770,16 @@ def get_args(v):
...
+    def test_squeeze_drop(self):
+        data = Dataset({'foo': ('x', [1])}, {'x': [0]})
+        expected = Dataset({'foo': 1})
+        selected = data.squeeze(drop=True)
+        self.assertDatasetIdentical(expected, selected)
+
+        expected = Dataset({'foo': 1}, {'x': 0})
+        selected = data.squeeze(drop=False)
+        self.assertDatasetIdentical(expected, selected)
+
```
jhamman: Can you add a few more tests to this section:

- squeeze empty array
- squeeze 2d array with a single element

shoyer: done

shoyer:
Going to merge this shortly unless anyone has objections

pwolfram:
@shoyer, I think this is really great and appreciate you doing this awesome work.  I obviously have a vested interest in the `drop=True` capability because we have found great use for it via `where`. My vote is go for it!

shoyer:
>  I obviously have a vested interest in the drop=True capability because we have found great use for it via where.

Well, to be clear `drop=True` works pretty differently on these methods than in `where`, so this might be confusing.

For `isel`, `sel` and `squeeze`, it drops scalar coordinates variables that would arise from indexing, whereas for `where` it drops some coordinate values. That said, the `drop` method already handles both cases.

PR #1155, opened 2016-12-07
RTD: fix savefig dir creation

fmaussion:
I've had a hard time finding out what was going on with missing RTD plots since we updated the packages until I came across this: https://github.com/ipython/ipython/issues/8733

The proposed fix creates the ``@savefig`` directory if not there. It works locally but I'm not sure if it will work on RTD (they might use a custom build directory).

The other changes are minor tweaks to remove some build warnings and/or making use of the latest sphinx Makefile. I also removed a part of the conf.py which I think is not needed anymore.

@shoyer do you remember what the ``inspect.findsource`` monkeypatch was good for?


doc/Makefile
```diff
@@ -47,139 +41,185 @@ help:
 	@echo "  pseudoxml  to make pseudoxml-XML files for display purposes"
 	@echo "  linkcheck  to check all external links for integrity"
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
+	@echo "  coverage   to run coverage check of the documentation (if enabled)"
+	@echo "  dummy      to check syntax errors of document sources"
 
+.PHONY: clean
 clean:
 	rm -rf $(BUILDDIR)/*
-	rm -rf generated/*
```
shoyer: I think this is still useful for clearing pages generated by sphinx for API docs.

doc/conf.py
```diff
@@ -67,99 +71,8 @@
...
     print("no netCDF4")
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-#sys.path.insert(0, os.path.abspath('.'))
-
-# Monkey patch inspect.findsource to work around a Python bug that manifests on
-# RTD. Copied from IPython.core.ultratb.
-# Reference: https://github.com/ipython/ipython/issues/1456
```
shoyer: see this page for the reference -- it's quite likely that this is no longer an issue now that we've updated ipython for our RTD builds.

doc/conf.py
```diff
@@ -409,8 +327,8 @@ def findsource(object):
 # (source start file, target name, title, author,
 #  dir menu entry, description, category)
 texinfo_documents = [
-  ('index', 'xarray', u'xarray Documentation',
-   u'xarray Developers', 'xarray', 'One line description of project.',
+  ('index', 'xarray', 'xarray Documentation',
+   'xarray Developers', 'xarray', 'N-D labeled arrays and datasets in Python.',
```
shoyer: Last time I tried (quite a while ago), the tex build did not work.

fmaussion: Yes, it does not appear as download option on RTD either. It's not very useful anyway...

doc/Makefile
fmaussion: Yes, will put it back again

doc/conf.py
shoyer: That's because I disabled that build on RTD :)

fmaussion:
OK, so this fix worked locally but not on RTD because RTD uses another build directory:

```
/home/docs/checkouts/readthedocs.org/user_builds/xray/checkouts/latest/doc/plotting.rst:None: WARNING: image file not readable: ../../../../../readthedocs/templates/sphinx/_static/plotting_example_existing_axes.png
```

This is really annoying. I'll try to dig into another possible fix mentioned here: https://github.com/ipython/ipython/issues/8733#issuecomment-234249437

PR #1157, opened 2016-12-08
PERF: Use len rather than size

MaximilianR:
Potential mitigation for https://github.com/pandas-dev/pandas/issues/14822


MaximilianR:
Are there any other places where we're looking at `.size` and know we're 1D? 

@shoyer in general, would you like to merge these or is LGTM approval enough?

shoyer:
@MaximilianR you are welcome to merge these after an approval

Issue #1159, opened 2016-12-08
Problem passing 'norm' when plotting a faceted figure


jbusecke:
I am not sure if this is a bug or user error by me.

I am basically trying to pass a SymLogNorm to all plots in a faceted plot, but it has no effect.
It works when I just plot a single axis. 
Below I reproduced the effect with an example dataset.

Matplotlib version: 1.5.1
xarray version: 0.8.2

```python
import xarray as xr
import matplotlib as mpl

# Here the passed 'norm=' keyword has the desired effect
da1 = xr.DataArray(np.random.exponential(size=[10,10]))

plt.figure()
da1.plot()

plt.figure()
da1.plot(norm=mpl.colors.SymLogNorm(0.1))


# In the faceted plot this has no effect
da2 = xr.DataArray(np.random.exponential(size=[10,10,4]))

plt.figure()
da.plot(x='dim_0',y='dim_1',col='dim_2',col_wrap=2)

plt.figure()
da.plot(x='dim_0',y='dim_1',col='dim_2',col_wrap=3,norm=mpl.colors.SymLogNorm(0.1))
```

fmaussion:
Yes, I think this is a bug. I have tried a fix in https://github.com/pydata/xarray/pull/1160

fmaussion:
Thanks for the report @jbusecke !

jbusecke:
Thanks for the quick fix @fmauisson ! Much appreciated.

PR #1160, opened 2016-12-08
Norm should be passed to facetgrid too

fmaussion:
fixes https://github.com/pydata/xarray/issues/1159

### DIFF
--- a/xarray/test/test_plot.py
+++ b/xarray/test/test_plot.py
@@ -295,11 +295,6 @@ class TestDetermineCmapParams(TestCase):
         self.assertIsNone(cmap_params['levels'])
         self.assertIsNone(cmap_params['norm'])
 
-    def test_norm(self):
-        cmap_params = _determine_cmap_params(self.data,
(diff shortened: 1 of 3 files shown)


xarray/test/test_plot.py
```diff
@@ -293,6 +295,11 @@ def test_center(self):
         self.assertIsNone(cmap_params['levels'])
         self.assertIsNone(cmap_params['norm'])
 
+    def test_norm(self):
+        cmap_params = _determine_cmap_params(self.data,
+                                             norm=mpl.colors.SymLogNorm(0.1))
+        self.assertIsNotNone(cmap_params['norm'])
+
```
jhamman: Ideally, we can test this without having to test a private function like `_determine_cmap_params`.  We should be able to check that the facet grid subplots are using the correct `norm`.

fmaussion: thanks! done

jhamman: I think we can get rid of this test then.

Issue #1161, opened 2016-12-12
Generated Dask graph is huge - performance issue?


mangecoeur:
I've been trying to get around some performance issues when subsetting a  set of netCDF files opend with `open_mfdataset`. I managed to print out the generated dask graph for one variable and it doesn't seem right - it's huge, 5000 elements, and seems to have a getitem entry for every requested element for that variable.

The code that generates this select looks roughly like:

```python

paths = WEATHER_MET['latlon'].glob('*_resampled.nc')
dataset = xr.open_mfdataset([str(p) for p in paths])
selection = dataset.sel(time=time_sel).sel_points(method='nearest', tolerance=0.1, lon=lon, lat=lat)
selection *= weights
```

and the graph for one variable in the select (the irradiance value) looks like this:

[mydask.pdf](https://github.com/pydata/xarray/files/646830/mydask.pdf)

shoyer:
Indeed, `sel_points` currently uses a very inefficient implementation -- it indexes each point and then concatenates them together.

Could be worth experimenting with `dask.array.Array.vindex` to make a more scalable version: https://github.com/dask/dask/pull/439

mangecoeur:
Ok I will have a look, where is this implemented (I always seem to have trouble pinpointing the dask-specific bits in the codebase :S )

shoyer:
https://github.com/pydata/xarray/blob/1615a0f65c331ef5f7c6be83eccf9a1c6796fa77/xarray/core/dataset.py#L1080

nothing about the logic is dask specific currently

mangecoeur:
Thanks, I've been looking around and I think i'm getting close, however i'm not sure the best way to turn the array slice i get from vindex into a DataArray variable. I'm thinking I might but together a draft PR for comments. This is what i have so far:

```python

def isel_points(self, dim='points', **indexers):
    """Returns a new dataset with each array indexed pointwise along the
    specified dimension(s).

    This method selects pointwise values from each array and is akin to
    the NumPy indexing behavior of `arr[[0, 1], [0, 1]]`, except this
    method does not require knowing the order of each array's dimensions.

    Parameters
    ----------
    dim : str or DataArray or pandas.Index or other list-like object, optional
        Name of the dimension to concatenate along. If dim is provided as a
        string, it must be a new dimension name, in which case it is added
        along axis=0. If dim is provided as a DataArray or Index or
        list-like object, its name, which must not be present in the
        dataset, is used as the dimension to concatenate along and the
[...]

PR #1162, opened 2016-12-13
#1161 WIP to vectorize isel_points

mangecoeur:
WIP to use dask vindex to point based selection


xarray/core/dataset.py
```diff
@@ -996,7 +996,7 @@ def sel(self, method=None, tolerance=None, **indexers):
         )
         return self.isel(**pos_indexers)._replace_indexes(new_indexes)
 
-    def isel_points(self, dim='points', **indexers):
+    def isel_points_old(self, dim='points', **indexers):
```
shoyer: Can you remove the old function instead? That will make GitHub show a more informative diff.

mangecoeur: done. Sorry it's a bit of a mess doing this under deadlines

mangecoeur:
So it seems to work fine in the Dask case, but I don't have a deep understanding of  how DataArrays are constructed from arrays and dims so it seems to fail in the non-dask case. Also not sure how you feel about making a special case for the dask backend here (since up till now it was all backend agnostic).

xarray/core/dataset.py
```diff
@@ -1077,12 +1077,41 @@ def relevant_keys(mapping):
...
+                                            name=name)
+            else:
+                # TODO need to make sure this still works in the non-dask case
+                # if data is ndarray should give expected behaviour
+                variables[name] = DataArray(var.data[tuple(slc)],
+                          coords=[np.arange(points_len)] + coords, dims=[dim] + non_indexed,
+                          name=name)
+
+
+        return merge([v for k, v in variables.items() if k not in self.dims])
```
shoyer: I think you can just use the normal `Dataset` constructor instead of passing these into `merge`.

xarray/core/dataset.py
```diff
@@ -1077,12 +1077,41 @@ def relevant_keys(mapping):
...
+        # need to generate a new dimension
+        points_len = lengths.pop()
+
+        variables = OrderedDict()
+
+        # add : slices for the non-indexed dimensions
+        slc = [indexers_dict[k] if k in indexers_dict else slice(None, None) for k in self.dims]
+        coords = [self.variables[k] for k in non_indexed]
+
+        # TODO need to figure out how to make sure we get the indexed vs non indexed dimensions in the right order
```
shoyer: If you transpose the arrays so that the dimensions being indexed are always leading dimensions (something like `reordered = self.transpose(*(indexed_dims + non_indexed_dims))`), indexing `.vindex` on a dask array and directly on a numpy array will give the same result.

You really don't want to allow indexed dimensions in the middle somewhere because NumPy has unintuitive results for reordering axes in some edge cases.

mangecoeur:
@shoyer I'm down to 1 test failing locally in `sel_points` but not sure what the desired behaviour is. I get:

```
<xarray.Dataset>
Dimensions:  (points: 3)
Coordinates:
  * points   (points) int64 0 1 2
Data variables:
    foo      (points) int64 0 4 8
```
instead of

```
AssertionError: <xarray.Dataset>
Dimensions:  (points: 3)
Coordinates:
  o points   (points) -
Data variables:
    foo      (points) int64 0 4 8
```

But here I'm not sure if my code is wrong or the test. It seems that the test requires `sel_points` NOT to generate a new coordinate values for points - however I'm pretty sure `isel_points` does require this (it passes in any case). Don't really see a way in my code to generate subsets without having a matching coordinate array (I don't know how to use the Dataset constructors without one for instance).

I've updated the test according to how I think it should be working, but please correct me if i misunderstood.

shoyer:
The development version of xarray includes a change that makes indexes optional. So if you use the Dataset/DataArray constructor it no longer adds new coordinates for each dimension by default.

mangecoeur:
OK I adjusted for the new behaviour and all tests pass locally, hopefully travis agrees...

xarray/core/dataset.py
```diff
@@ -1110,16 +1122,90 @@ def relevant_keys(mapping):
...
+                sel_coords[c] = self.coords[c]
+
+        variables = OrderedDict()
+
+        for name in data_vars:
+            var = self.variables[name]
+            # Transpose the var to ensure that the indexed dims come first
+            # These dims will be collapsed in the output.
+            # To avoid edge cases in numpy want to transpose to ensure the indexed dimensions are first
+            # However transpose is not lazy, so want to avoid using it for dask case (??)
```
shoyer: This is a potential concern if the data is *not* loaded in dask but is still pointing to a lazily loaded array of values on disk in a netCDF file -- calling `.transpose` will load the entire array into memory.

That said, I think this change is still worth doing, and we shouldn't add special cases for different data types. It is not unreasonable to encourage users to use dask if they are concerned about memory usage and lazy loading.

xarray/core/dataset.py
```diff
@@ -1110,16 +1122,90 @@ def relevant_keys(mapping):
...
+            var = var.transpose(*(list(d for d in indexer_dims if d in var.dims) +
```
shoyer: You might transpose the entire dataset once, instead of transposing each variable in turn. It should do basically the same thing.

xarray/core/dataset.py
```diff
@@ -1110,16 +1122,90 @@ def relevant_keys(mapping):
...
+                coord_dim = var.dims[0]  # should just be one?
+                selection = take(var, indexers_dict[coord_dim])
+
+                sel_coords[c] = (dim, selection)
+            else:
+                sel_coords[c] = self.coords[c]
+
+        variables = OrderedDict()
+
+        for name in data_vars:
```
shoyer: It is possible to use this loop for all elements in `self.variables`, instead of adding special cases for `coords` and `data_vars`?

mangecoeur: I will look into it, it's originally this way to preserve the behaviour expected in the tests but now that I understand that behaviour better I might be able to simplify

xarray/core/dataset.py
mangecoeur: I did that originally but then I thought that if there are variables that are not indexed at all we can skip applying transpose to them this way (and triggering a data load on them too). Does that make sense?

shoyer:
I believe Dataset.transpose already skips variables that don't need to be
transposed.

mangecoeur:
@shoyer Tidied up based on recommendations, now everything done in a single loop (still need to make distinction between variables and coordinates for output but still a lot neater)

xarray/core/dataset.py
```diff
@@ -1110,16 +1124,66 @@ def relevant_keys(mapping):
...
+        for name, var in reordered.variables.items():
+            if name in indexers_dict or any(d in indexer_dims for d in var.dims):
+                # if the var is an indexer or if it depends on an indexed dim, slice it
+                slc = [indexers_dict[k] if k in indexers_dict else slice(None) for k in var.dims]
+                var_dims = [dim] + [d for d in var.dims if d in non_indexed_dims]
+                selection = take(var, tuple(slc))
+
+                # Add the var to variables or coords accordingly
+                if name in coords:
+                    sel_coords[name] = (var_dims, selection)
```
shoyer: Rather than conditionally adding variables to `sel_coords` or `variables`, consider putting everything in `variables` and using the private constructor `self._replace_vars_and_dims` to return a new dataset with new updated variables. See elsewhere in this file for examples.

xarray/core/dataset.py
```diff
@@ -1074,16 +1077,27 @@ def isel_points(self, dim='points', **indexers):
 
         indexer_dims = set(indexers)
 
+        def take(variable, slices):
+            if hasattr(variable.data, 'vindex'):
+                # Special case for dask backed arrays to use vectorised list indexing
+                sel = variable.data.vindex[slices]
+            else:
+                # Otherwise assume backend is numpy array with 'fancy' indexing
```
shoyer: Add a note to remove this helper function after numpy supports `vindex`, with a reference to https://github.com/numpy/numpy/pull/6075

xarray/core/dataset.py
```diff
@@ -1034,6 +1036,7 @@ def sel(self, method=None, tolerance=None, drop=False, **indexers):
         return result._replace_indexes(new_indexes)
 
     def isel_points(self, dim='points', **indexers):
+        # type: () -> Dataset
```
shoyer: should be `# type: (...) -> Dataset`

xarray/core/dataset.py
```diff
@@ -1110,16 +1124,66 @@ def relevant_keys(mapping):
...
                                  'for the dim argument in sel_points')
+            # replace the array dim
+            dim_coord = dim.copy()
+            dim = dim.name
 
         if not utils.is_scalar(dim) and not isinstance(dim, DataArray):
-            dim = as_variable(dim, name='points')
+            name = 'points' if not hasattr(dim, 'name') else dim.name
+            dim_coord = as_variable(dim, name=name)
+            dim = name
```
shoyer: It's not immediately obvious to me what this is doing. That said, this is forgivable, given that I'm pretty sure you just duplicated it from `concat`.

xarray/core/dataset.py
shoyer: Also note: recreating variables with a tuple drops `attrs`. It would be better to write `(var_dims, selection, var.attrs)` or better yet `type(var)(var_dims, selection, var.attrs)` to be explicit about constructing variable objects (you would need to do this if you switch to use `_replace_vars_and_dims`).

xarray/core/dataset.py
mangecoeur: This is just to make sure we fallback to the default name 'points' in case the supplied index doesn't have its own name

xarray/core/dataset.py
mangecoeur: Ok I will try that. I had originally found that the tests weren't happy if I didn't make that distinction, I think because the test expects variables to be classified into vars or coords, but since both vars and coords can depend on other coords it didn't seem to infer correctly which to consider an element a 'var' and when to consider it a 'coord'.  Will see if the private api fixes that.

mangecoeur:
Completed changes based on recommendations and cleaned up old code and comments.

As for benchmarks, I don't have anything rigourous but I do have the following example `dataset` weather data from the CFSR dataset, 7 variables at hourly resolution, collected in one netCDF3 file per variable per month. In the particular case the difference is striking!

```python
%%time
data = dataset.isel_points(time=np.arange(0,1000), lat=np.ones(1000, dtype=int), lon=np.ones(1000, dtype=int))
data.load()
```

Results:

```
xarray 0.8.2
CPU times: user 1min 21s, sys: 41.5 s, total: 2min 2s
Wall time: 47.8 s

master
CPU times: user 385 ms, sys: 238 ms, total: 623 ms
Wall time: 288 ms
```

shoyer:
This looks great to me. I'll merge this shortly after releasing 0.9.0 -- could you please add a brief release note in a new section for 0.9.1?  (It won't be a long wait.)

mangecoeur:
Ok will wait for 0.9.0 to be released

xarray/core/dataset.py
```diff
@@ -1161,13 +1162,16 @@ def sel(self, method=None, tolerance=None, drop=False, **indexers):
...
     def isel_points(self, dim='points', **indexers):
+        # type: (...) -> Dataset
         """Returns a new dataset with each array indexed pointwise along the
         specified dimension(s).
 
         This method selects pointwise values from each array and is akin to
         the NumPy indexing behavior of `arr[[0, 1], [0, 1]]`, except this
         method does not require knowing the order of each array's dimensions.
 
+        Will use dask vectorised operation if available
```
shoyer: Let's not add this comment -- implementation details usually don't make it into docstrings

mangecoeur: righto

mangecoeur:
Note - waiting for 0.9.0 to be released before updating whats new, don't want to end up with conflicts in docs

shoyer:
Actually, if you want to write that now I think we can squeeze this in. We
are still a day or two away from the release.

On Mon, Jan 23, 2017 at 9:52 AM mangecoeur <notifications@github.com> wrote:

Note - waiting for 0.9.0 to be released before updating whats new, don't
want to end up with conflicts in docs

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/pydata/xarray/pull/1162#issuecomment-274564256>, or mute
the thread

mangecoeur:
OK added a performance improvements section to the docs

shoyer:
Looks good. Unfortunately there was a merge conflict with the quantile PR so you need to merge master again.

mangecoeur:
Crickey. Fixed merge hopefully it works (I hate merge conflicts)

shoyer:
OK, build passed so I'm merging. Thanks!

Issue #1163, opened 2016-12-14
User warning / more transparent _FillValue interface for .to_netcdf()


laliberte:
Since many tools (for example ``cdo``) do not work well with ``NaN`` values, the default missing value setting for the ``.to_netcdf()`` method is not well-suited to enable optimal data reuse.

One option would be to add a ``fill_value=NaN`` option to ``.to_netcdf()`` and output a warning message whenever ``NaN``s are written to netCDF: ``'The data being written to netCDF includes missing values (NaN) that are often not easily handled by other netCDF softwares. It is strongly encouraged to coerce missing data to a valid yet large floating point value (for example, 1e20 for float) by calling .to_netcdf(..., fill_value=1e20)'.``

shoyer:
I'm a little surprised that CDO cannot handle a NaN `_FillValue` -- are you sure this is actually the case? If so, it might be worth raising the issue with the CDO developers.

But actually, xarray currently does not even set `_FillValue` by default, though you can do this yourself on a per-variable basis [by setting `encoding`](http://xarray.pydata.org/en/stable/io.html#scaling-and-type-conversions). I think it would be reasonable to default to a `_FillValue` of NaN for floating point dtypes if that would help third party tools, given that that is how xarray's data model works (any other default choice would run the risk of over-writing user data).

laliberte:
I'm pretty sure that of over many versions of CDO NaNs are not properly handled. For example, all the ``setmiss*`` commands have a very hard time with NaN's. 

I have been using the ``encoding`` option but I find it not very intuitive and kind of cumbersome when there are many variables.

I think you're right that actually putting ``_FillValue = NaN`` in the variable attribute might be enough to solve most issues with third parties. If it still does not work, then it would be reasonable to raise this issue with each one of the third parties.

shoyer:
Fixed by #1165

PR #1165, opened 2016-12-15
Set a default _FillValue of NaN for float types

laliberte:
This PR attempts to solve #1163 by systematically including a default ``_FillValue`` of ``NaN`` whenever the user do not explicitly specify a FillValue. The idea is that third party software should work properly in these situations and if they don't then it is a bug on their side.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -567,7 +567,7 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
                     pass
             except ValueError:
                 pass
-            
+
     # validate Dataset keys, DataArray names, and attr keys/values
     _validate_dataset_names(dataset)
     _validate_attrs(dataset)
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -549,6 +549,25 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
             engine = _get_default_engine(path)
         path = _normalize_path(path)
 
+    for var in encoding:
+        # This section sets the default _FillValue
+        # to NaNs
+        if '_FillValue' not in encoding[var]:
+            if 'dtype' in encoding[var]:
+                dtype = encoding[var]['dtype']
+            else:
+                dtype = dataset[var].dtype
+            try:
+                # Test whether NaNs are accepted for
+                # the requested dtype:
+                np.asarray(np.nan, dtype=dtype)
+                try:
+                    encoding[var]['_FillValue'] = 'NaN'
+                except TypeError:
+                    pass
+            except ValueError:
+                pass
+            
     # validate Dataset keys, DataArray names, and attr keys/values
     _validate_dataset_names(dataset)
     _validate_attrs(dataset)
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -549,25 +549,6 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
             engine = _get_default_engine(path)
         path = _normalize_path(path)
 
-    for var in encoding:
-        # This section sets the default _FillValue
-        # to NaNs
-        if '_FillValue' not in encoding[var]:
-            if 'dtype' in encoding[var]:
-                dtype = encoding[var]['dtype']
-            else:
-                dtype = dataset[var].dtype
-            try:
-                # Test whether NaNs are accepted for
-                # the requested dtype:
-                np.asarray(np.nan, dtype=dtype)
-                try:
-                    encoding[var]['_FillValue'] = np.nan
-                except TypeError:
-                    pass
-            except ValueError:
-                pass
-
     # validate Dataset keys, DataArray names, and attr keys/values
     _validate_dataset_names(dataset)
     _validate_attrs(dataset)
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -562,7 +562,7 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
                 # the requested dtype:
                 np.asarray(np.nan, dtype=dtype)
                 try:
-                    encoding[var]['_FillValue'] = 'NaN'
+                    encoding[var]['_FillValue'] = np.nan
                 except TypeError:
                     pass
             except ValueError:
--- a/xarray/test/test_backends.py
+++ b/xarray/test/test_backends.py
@@ -424,27 +424,6 @@ class CFEncodedDataTest(DatasetIOTestCases):
             self.assertEqual(actual.x.encoding['dtype'], 'f4')
         self.assertEqual(ds.x.encoding, {})
 
-        # Test default encoding for float:
-        ds = Dataset({'x': ('y', np.arange(10.0))})
-        kwargs = dict(encoding={'x': {'dtype': 'f4'}})
-        with self.roundtrip(ds, save_kwargs=kwargs) as actual:
-            self.assertEqual(actual.x.encoding['_FillValue'],
-                             np.nan)
-        self.assertEqual(ds.x.encoding, {})
-
-        # Test default encoding for int:
-        ds = Dataset({'x': ('y', np.arange(10.0))})
-        kwargs = dict(encoding={'x': {'dtype': 'int16'}})
-        with self.roundtrip(ds, save_kwargs=kwargs) as actual:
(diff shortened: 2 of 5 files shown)


rabernat:
I can't see any problems with this.

laliberte:
The only problem I see is that in the current implementation this fix puts a ``_FillValue`` attribute even if there are no ``NaN`` in the variable. Usually, ``cdo`` would not do that. I guess there are little risks but I think it's important that this be said. Alternative implementations would be way too cumbersome so I think the approach used in this PR is the lesser of two evils.

fmaussion:
> The only problem I see is that in the current implementation this fix puts a _FillValue attribute even if there are no NaN in the variable.

Yes, I'm not very enthusiastic about that either...

laliberte:
I know but the alternative would be to scan the data before deciding but that would kind of break the whole ``dask`` integration.

xarray/backends/api.py
```diff
@@ -549,6 +549,25 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
             engine = _get_default_engine(path)
         path = _normalize_path(path)
 
+    for var in encoding:
+        # This section sets the default _FillValue
+        # to NaNs
```
fmaussion: Nit: no need to break line here

xarray/backends/api.py
```diff
@@ -549,6 +549,25 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
...
+                dtype = encoding[var]['dtype']
+            else:
+                dtype = dataset[var].dtype
+            try:
+                # Test whether NaNs are accepted for
+                # the requested dtype:
+                np.asarray(np.nan, dtype=dtype)
+                try:
+                    encoding[var]['_FillValue'] = np.nan
+                except TypeError:
```
fmaussion: In which cases will this line cause a ``TypeError``? This is not covered by your tests.

xarray/backends/api.py
```diff
@@ -549,6 +549,25 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
...
+        # This section sets the default _FillValue
+        # to NaNs
+        if '_FillValue' not in encoding[var]:
+            if 'dtype' in encoding[var]:
+                dtype = encoding[var]['dtype']
+            else:
+                dtype = dataset[var].dtype
+            try:
+                # Test whether NaNs are accepted for
+                # the requested dtype:
```
fmaussion: Nit: no need to break line here

laliberte: fixed

xarray/backends/api.py
```diff
@@ -549,6 +549,25 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
             engine = _get_default_engine(path)
         path = _normalize_path(path)
 
+    for var in encoding:
```
shoyer: Can you put this logic in `conventions.py` instead, [after this line](https://github.com/pydata/xarray/blob/7084df0e82c3c78765691910cd02b197b5772575/xarray/conventions.py#L726) where we converted to any encoded dtype? If `_FillValue` was set in `encoding` at the point, it has already been converted into `atttrs`.

xarray/test/test_backends.py
```diff
@@ -424,6 +424,27 @@ def test_encoding_kwarg(self):
             self.assertEqual(actual.x.encoding['dtype'], 'f4')
         self.assertEqual(ds.x.encoding, {})
 
+        # Test default encoding for float:
```
shoyer: can make this a separate test, e.g., `test_default_fill_value`?

laliberte:
Just to summarize:

1. Currently, where an array contains missing values, ``.to_netcdf()`` records ``NaN``s in the netcdf file but does not record that these ``NaN``s mean missing values. Then third party softwares read these values as ``NaN``s, not missing values. I don't think we can blame them for that. So this seems to me like an ``xarray`` problem.
2. The ideal option would be to add a ``_FillValue = NaN`` to float variables ONLY when there are floats in the array. This would conform to netCDF standards and should ensure seamless interaction with third parties.
3. Currently, it would seem that option 2 would require to monitor [ArrayWriter](https://github.com/pydata/xarray/blob/master/xarray/backends/common.py#L136-L157) and then modify the ``_FillValue`` when a ``NaN``s is encountered. It seems cumbersome and likely to interact poorly with ``dask``. This leaves the option proposed in this PR: use a ``_FillValue`` of ``NaN`` by default...

xarray/backends/api.py
laliberte: Would you want me to include an extra test?

xarray/backends/api.py
```diff
@@ -549,6 +549,25 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
...
+        # to NaNs
+        if '_FillValue' not in encoding[var]:
+            if 'dtype' in encoding[var]:
+                dtype = encoding[var]['dtype']
+            else:
+                dtype = dataset[var].dtype
+            try:
+                # Test whether NaNs are accepted for
+                # the requested dtype:
+                np.asarray(np.nan, dtype=dtype)
```
shoyer: I think it will be more reliable to explicitly check if the dtype is a float or complexfloat type, e.g., use `if np.issubdtype(x.dtype, np.floating) or np.issubdtype(x.dtype, np.complexfloating):`. Xarray uses dtype=object for arrays of strings, and we don't want to use `np.nan` as a fill value for those.

xarray/backends/api.py
fmaussion: Ideally, having full test coverage is always good but I have no strong opinion on this either ;)

xarray/backends/api.py
laliberte: That's a good point, I've fixed that.

fmaussion:
Yes, agreed, LGTM.

I'll let @shoyer have the final word though

shoyer:
I agree that we just want to set `_FillValue=NaN` unilaterally for float arrays. This will make for slightly less tidy netCDF files, but we really don't want to be checking arrays for NaNs when we write them to disk, which would be untenable with dask.

xarray/backends/api.py
laliberte: done

laliberte:
So I've tried with ``conventions.py`` and it works. However, allowing a NaN ``_FillValue`` for complex floats raises an exception in ``h5py``:  h5py/h5py#805. This means that until this is fixed in ``h5py`` we should restrict this default behaviour to real float types.

shoyer:
OK, I'm fine with only doing this for real types. Complex types aren't really even valid in standard netCDF files (h5py uses a custom data type), so I'm not worried about compatibility with other tools.

laliberte:
@shoyer Let me know if you need any more changes.

shoyer:
Thanks!

shoyer:
@laliberte we missed giving your credit for this fix in "What's new". Any interest in putting together another quick PR adding "By Frederic Laliberte"?

thenaomig:
Is there a way to override this default behavior and write a netCDF file *without* including _FillValue in the metadata? For example, if one were adhering to a requested output format that does not include this?

shoyer:
> Is there a way to override this default behavior and write a netCDF file without including _FillValue in the metadata? For example, if one were adhering to a requested output format that does not include this?

This is a good question. I don't think so currently.

In principle we could use a entry in the `encoding` dict to signal this, i.e., `{'_FillValue': None}`.

jhamman:
@thenaomig - are you only concerned about coordinate variables (I think your gitter post mentioned this)? We may also consider special casing these as well given the CF standard on this issue.

thenaomig:
At the moment, I am only concerned with coordinate variables. I suppose it would be less likely to come up otherwise, but someone might want to leave it off for any variable.

mmartini-usgs:
It is not desirable for us to have _FillValue = NaN for dimensions and coordinate variables.

In trying to use xarray, _FillValue was carefully kept from these variables and dimensions during the creation of the un-resampled file and then were found to appear during the to_netcdf operation.  This happens in spite of mask_and_scale=False is being used with xr.open_dataset

I would hope that downstream code would have trouble with coordinates that don't make logical sense (time or place being NaN, for instance).  We would prefer NOT to instantiate coordinate variable data with any fill value.  Keeping NaNs out of coordinate variables, dimensions and minima and maxima is part of our QA/QC process to avoid downstream issues.

Issue #897, opened 2016-07-12
Allow specification of figsize in plot methods


robintw:
Pandas allows a call to `plot` like:

``` python
df.plot(x='var1', y='var2', figsize=(10, 6))
```

but this doesn't seem to be possible in xarray - but it would be handy if it were.

It looks like fixing this would require modifying the `@_plot2d` decorator, specifically around https://github.com/pydata/xarray/blob/master/xarray/plot/plot.py#L376 (although I can't seem to find how to do the equivalent of `gca()` but set a `figsize` too).

Any thoughts or ideas?

shoyer:
I made some significant progress on this previously (https://github.com/pydata/xarray/pull/637), but stalled out when I wasn't happy with any of the API options. I would welcome someone else picking this up.

PR #637, opened 2015-10-23
size and aspect arguments for plotting methods even without faceting

shoyer:
I was finding myself writting `plt.figure(figsize=(x, y))` way too
often. This will be a convenient shortcut.

Still needs tests.

PR #1168, opened 2016-12-16
Add figsize, size, and aspect arguments to plotting methods

fmaussion:
Extends and finishes https://github.com/pydata/xarray/pull/637

I chose to keep seaborn's convention for two reasons:
- it doesn't break existing code
- now that ``figsize`` is also available, I expect the ``size`` and ``aspect`` kwargs to be less used in non-facetgrid plots

Closes https://github.com/pydata/xarray/issues/897

### DIFF
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -156,7 +156,10 @@ resulting image via the formula ``figsize = (aspect * size, size)``:
     # create a dummy figure so sphinx plots everything below normally
     plt.figure()
 
-This feature also works with :ref:`plotting.faceting`.
+This feature also works with :ref:`plotting.faceting`. For facet plots,
+``size`` and ``aspect`` refer to a single panel (so that ``aspect * size``
+gives the width of each facet in inches), while ``figsize`` refers to the
+entire figure (as for matplotlib's ``figsize`` argument).
 
 .. note::
 
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -134,6 +134,42 @@ axes created by ``plt.subplots``.
 
 On the right is a histogram created by :py:func:`xarray.plot.hist`.
 
+.. _plotting.figsize:
+
+Controlling the figure size
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can pass a ``figsize`` argument to all xarray's plotting methods to
+control the figure size. For convenience, xarray's plotting methods also
+support the ``aspect`` and ``size`` arguments which control the size of the
+resulting image via the formula ``figsize = (aspect * size, size)``:
+
+.. ipython:: python
+
+    air1d.plot(aspect=2, size=3)
+    @savefig plotting_example_size_and_aspect.png
+    plt.tight_layout()
+
+.. ipython:: python
+    :suppress:
+
+    # create a dummy figure so sphinx plots everything below normally
+    plt.figure()
+
+This feature also works with :ref:`plotting.faceting`.
+
+.. note::
+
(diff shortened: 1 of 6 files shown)


fmaussion:
After having another look, ``facetgrid`` should probably also accept a ``figsize`` kwarg...

fmaussion:
OK, this should be ready for review now

xarray/plot/plot.py
```diff
@@ -433,11 +470,16 @@ def newplotfunc(darray, x=None, y=None, ax=None, row=None, col=None,
         # This allows the user to pass in a custom norm coming via kwargs
         kwargs.setdefault('norm', cmap_params['norm'])
 
-        ax, primitive = plotfunc(xval, yval, zval, ax=ax,
-                                 cmap=cmap_params['cmap'],
-                                 vmin=cmap_params['vmin'],
-                                 vmax=cmap_params['vmax'],
-                                 **kwargs)
+        if 'imshow' == plotfunc.__name__ and isinstance(aspect, basestring):
```
shoyer: I think it would be better to avoid this special case, and simply not pass `aspect` on to `imshow`. Xarray `imshow` is already overloaded to display in data coordinates (not pixels) so I think this is pretty reasonable.

doc/plotting.rst
```diff
@@ -134,6 +134,42 @@ axes created by ``plt.subplots``.
...
+    @savefig plotting_example_size_and_aspect.png
+    plt.tight_layout()
+
+.. ipython:: python
+    :suppress:
+
+    # create a dummy figure so sphinx plots everything below normally
+    plt.figure()
+
+This feature also works with :ref:`plotting.faceting`.
```
shoyer: I think `figsize` refers to the entire figure when used with faceting, but `size` only refers to a single panel. We should either make that consistent (both referring to single panels?) or note the difference.

fmaussion: I think that ``figsize`` should always refer to the entire figure, as in matplotlib. I added a note to the docs

xarray/plot/plot.py
fmaussion: OK. I still have to catch this possibility though, otherwise the error message is not very informative

fmaussion:
OK, this should be ready for another round of reviews

xarray/plot/plot.py
shoyer: Can you clarify -- what happens if you try write `array.plot.imshow(aspect=2)`?

I think this should probably result in a plot with the same figure size as `array.plot.contourf(aspect=2)`.

fmaussion: Yes, sorry. Both methods will raise the same ``ValueError`` if called with ``aspect=2``:

```
ValueError: cannot provide `aspect` argument without `size`
```
The point of this line is to provide a more informative error message when users try to set aspect='equal', as in [this test](https://github.com/pydata/xarray/blob/master/xarray/test/test_plot.py#L901) for example. If you're happy with the message above I can remove this block.

xarray/plot/plot.py
```diff
@@ -433,11 +470,16 @@ def newplotfunc(darray, x=None, y=None, ax=None, row=None, col=None,
...
+            # forbid usage of mpl strings
+            raise ValueError("plt.imshow's `aspect` kwarg is not available " \
```
shoyer: Just use implicitly string concatenation here instead of writing `\` (you almost never want/need to actually use `\`)

xarray/plot/plot.py
shoyer: Oh OK, I was confused about this :). I agree, this makes complete sense.

Issue #1154, opened 2016-12-06
netCDF reading is not prominent in the docs


rabernat:
Just opening an issue to highlight what I think is a problem with the docs.

For me, the primary use of xarray is to read and process existing netCDF data files. @shoyer's  popular [blog post](https://www.continuum.io/content/xray-dask-out-core-labeled-arrays-python) illustrates this use case extremely well.

However, when I open the [docs](http://xarray.pydata.org/), I have to dig quite deep before I can see how to read a netCDF file. This could be turning away many potential users. The stuff about netCDF reading is hidden under "Serialization and IO". Many potential users will have no idea what either of these words mean.

IMO the solution to this is to reorganize the docs to make reading netCDF much more prominent and obvious.

shoyer:
The challenge here is that xarray has two big audiences (netCDF users and non-netCDF users), for which the right approach is very different. I tried to avoid being too netCDF specific in the docs to avoid turning off folks who aren't familiar with it. This is not to say that they would appreciate learning more about netCDF, but I know that an overemphasis on domain specific file formats can turn off others quickly.

So we almost need two separate introduction/tutorial pages, one for people who know netCDF and one for people who don't.

fmaussion:
See also: https://github.com/pydata/xarray/issues/768#issuecomment-187226020

shoyer:
@fmaussion yes, that's a really good point -- the intro for users who don't know netCDF should end with a call to arms about how great netCDF is :).

rabernat:
@shoyer I definitely see your point!

fmaussion:
Partly addressed via https://github.com/pydata/xarray/pull/1169

I'll leave the issue opened in case someone wants to add a more comprehensive netCDF doc later on

jhamman:
Closing based on the recent documentation updates (e.g. #2430) and older PRs like #1169.

Issue #768, opened 2016-02-17
save/load DataArray to numpy npz functions


jonathanstrong:
hey - 

Apologies if this is bad form: I wanted to pass this along but don't have time to do a proper pull request. 

I have found pickle to be really problematic for serializing data, so wrote these two functions to save to numpy's binary npz format and retrieve it. Generally, the numpy format is much less likely to bomb when attempting to load on another computer because of some unseen dependency. If there's interest, I could probably add this as a serialization method to DataArray in the next month or so. 

``` python
def to_npz(da, file_or_buffer):
    if 'dims' in da.dims: 
        raise ValueError('Can\'t use "dims" as a dim name.')
    if 'values' in da.dims: 
        raise ValueError('Can\'t use "values" as a dim name.')
    arrays = {}
    arrays['dims'] = da.dims
    for dim in da.dims: 
        arrays[dim] = da.indexes[dim]
    arrays['values'] = da.values
    np.savez(file_or_buffer, **arrays)

def from_npz(file_or_buffer): 
    data = np.load(file_or_buffer)
    assert hasattr(data, 'keys'), "np.load returned a {}, not a dict-like object".format(type(data))
    assert 'dims' in data, 'Can\'t locate "dims" key in file'
    assert 'values' in data, 'Can\'t locate "values" key in file'
    for dimname in data['dims']:
        assert dimname in data, 'Can\'t locate "{}" key in file'.format(dimname)
    return xray.DataArray(data['values'], dims=data['dims'], coords=dict(zip(data['dims'], [data[dimname] for dimname in data['dims']])))
```

it's pretty speedy, here is an example for a (3, 4, 5) shaped DataArray: 

```
In [42]:
def save_and_load_again(da):
    with open('/path/to/datarray.npz', 'w') as f: 
        to_npz(da, f)
    with open('/path/to/datarray.npz', 'r') as f: 
        a = from_npz(f)
    return a
%time (save_and_load_again(da) == da).all()
CPU times: user 12.6 ms, sys: 0 ns, total: 12.6 ms
Wall time: 26.2 ms
Out[42]:
<xray.DataArray ()>
array(True, dtype=bool)
```

shoyer:
This is a pretty reasonable way to save data, but my only concern is that it's not clear to me that we need another file format when netCDF already solves this problem, in a completely portable way. Have you tried using xarray's netCDF IO?

jonathanstrong:
[...]

PR #1169, opened 2016-12-16
DOC: small improvements to the netCDF docs

fmaussion:
Partly addresses https://github.com/pydata/xarray/issues/768#issuecomment-187226020 and https://github.com/pydata/xarray/issues/1154, mostly by making the tone less defensive. I also added  that we recommend the netCDF format for IO. (I think that the format is powerful enough for being useful in other disciplines than geosciences).

I agree with @rabernat that the current page title (Serialization and IO) could be changed to something more accessible, but I don't know how. Two possibilities, both quite long (it looks a bit ugly on RTD):
- "Reading and writing xarray data structures on disk"
- "Reading/writing xarray data structures"

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -4,7 +4,7 @@ Serialization and IO
 ====================
 
 xarray supports direct serialization and IO to several file formats, from
-simple :ref:`io.pickle` files to the much more flexible :ref:`io.netcdf`
+simple :ref:`io.pickle` files to the more flexible :ref:`io.netcdf`
 format.
 
 .. ipython:: python
@@ -20,8 +20,8 @@ format.
 Pickle
 ------
 
-The simplest way to serialize an xarray object is to use Python's built-in
-pickle module:
+The simplest way to serialize an xarray object is to use Python's built-in pickle
+module:
 
 .. ipython:: python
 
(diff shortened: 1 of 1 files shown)


doc/io.rst
```diff
@@ -393,6 +399,16 @@ We recommend installing PyNIO via conda::
 
 .. _combining multiple files:
 
+
+Formats supported by Pandas
+---------------------------
+
+For more options, consider exporting your objects to pandas
```
shoyer: Let's call CSV and other tabular formats in particular.

doc/io.rst
```diff
@@ -79,28 +77,36 @@ require external libraries and dicts can easily be pickled, or converted to
...
 netCDF
 ------
 
-Currently, the only disk based serialization format that xarray directly supports
-is `netCDF`__. netCDF is a file format for fully self-described datasets that
-is widely used in the geosciences and supported on almost all platforms. We use
-netCDF because xarray was based on the netCDF data model, so netCDF files on disk
+The recommended way to store xarray objects is `netCDF`__.
+netCDF is a file format for fully self-described datasets that
+is widely used and supported on almost all platforms. xarray is based on the
```
shoyer: I agree with the sentiment here, but I think it might still be a good idea to mention at least that netCDF "originated in the geosciences".

fmaussion:
Thanks! Updated

PR #1170, opened 2016-12-17
Dataset.encoding and unlimited dimensions for to_netcdf

### LINKED ISSUE #992: Creating unlimited dimensions with xarray.Dataset.to_netcdf
@shoyer you wrote in [a comment on another issue](https://github.com/pydata/xarray/issues/349#issuecomment-76883374)

> xray doesn't use or set unlimited dimensions. (It's pretty irrelevant for us, given that NumPy arrays can be stored in either row-major or column-major order.)

I see that xarray does not need `UNLIMITED` dimensions internally. But I need to create a netCDF file that I subsequently can append to (along the `time` dimension, in this case). Can this be done?

jhamman:
Add `Dataset.encoding` attribute and support unlimited dimensions for scipy/netcdf4 backends. 

closes #992


xarray/core/dataset.py
```diff
@@ -283,6 +287,18 @@ def attrs(self, value):
...
+        """Dictionary of global encoding attributes on this dataset
+        """
+        if self._encoding is None:
+            self._encoding = OrderedDict()
+        return self._encoding
+
+    @encoding.setter
+    def encoding(self, value):
+        self._encoding = OrderedDict(value)
+
```
MaximilianR: Why do you need this getter & setter? Why not just have `self.encoding = OrderedDict(encoding)` in the `__init__`? Because you need to coerce to Ordered?

xarray/core/dataset.py
```diff
@@ -283,6 +287,18 @@ def attrs(self, value):
         self._attrs = OrderedDict(value)
 
     @property
+    def encoding(self):
+        """Dictionary of global encoding attributes on this dataset
+        """
+        if self._encoding is None:
+            self._encoding = OrderedDict()
```
shoyer: I would just make it a normal dict, like `Variable.encoding`, unless you have a reason why you feel preserving order would be helpful.

xarray/core/dataset.py
```diff
@@ -519,7 +535,7 @@ def __deepcopy__(self, memo=None):
     @property
     def _attr_sources(self):
         """List of places to look-up items for attribute-style access"""
-        return [self, LevelCoordinatesSource(self), self.attrs]
+        return [self, LevelCoordinatesSource(self), self.attrs, self.encoding]
```
shoyer: I don't think this is a good idea. This lets you write things like `dataset.unlimited_dims` to pull out unlimited dimensions from `encoding`, but `dataset.encoding['unlimited_dims']` is more explicit and this runs the risk of confusing `encoding`/`attrs`.

xarray/core/dataset.py
shoyer: It might help to switch `if name != '__setstate__'` in `AttrAccessMixin.__getattr__` to  `if self._initialized`.

xarray/backends/netCDF4_.py
```diff
@@ -209,6 +217,8 @@ def __init__(self, filename, mode='r', format='NETCDF4', group=None,
         self._opener = opener
         self._filename = filename
         self._mode = 'a' if mode == 'w' else mode
+        self._unlimited_dimensions = set()
+        self.encoding = None
```
shoyer: I don't think you need an `encoding` property for datastores. Shouldn't `_unlimited_dimensions` be enough?

jhamman:
@shoyer - I'm making progress here, although I still need to cleanup/add tests.  Do you have an idea what might be causing this [test](https://travis-ci.org/pydata/xarray/jobs/185571068#L788) failure though? I'm sure it's something I broke but I'm having trouble connecting the dots...

shoyer:
So that is definitely a bug with scipy not handling `...` for assignment with unlimited dimensions. We could make that work by making a proxy object that implements `__setitem__` for `Ellipsis` by using a slice object or `assignValue` for scalars, but it would be nice to fix it upstream in scipy.

xarray/backends/common.py
```diff
@@ -143,7 +146,11 @@ def add(self, source, target):
             self.sources.append(source)
             self.targets.append(target)
         else:
-            target[...] = source
+            try:
+                target[...] = source
+            except TypeError:
+                # workaround for GH: scipy/scipy#6880
+                target[slice(None, None, None)] = source
```
jhamman: @shoyer  - I raised the `Ellipsis` issue with the scipy folks: https://github.com/scipy/scipy/issues/6880. This seems to work although maybe it isn't the most robust solution, I don't know.

shoyer: Yes, this seems reasonable to me. I would just write `target[:] = source`, though -- no need to manually make the `slice()` object. This should work because variables with unlimited dimensions will never be scalars.

jhamman:
this is ready for a full review.

xarray/backends/h5netcdf_.py
```diff
@@ -100,7 +102,12 @@ def prepare_variable(self, name, variable, check_encoding=False):
         if dtype is str:
             dtype = h5py.special_dtype(vlen=unicode_type)
 
-        self.set_necessary_dimensions(variable)
+        unlimited_dims = self.encoding.get('unlimited_dims', set())
+        if len(unlimited_dims) > 0:
+            warnings.warn('h5netcdf does not support unlimited dimensions',
```
shoyer: If `check_encoding` is `True`, this should raise an error, not just a warning.

xarray/backends/netCDF4_.py
```diff
@@ -251,6 +251,12 @@ def get_dimensions(self):
         return FrozenOrderedDict((k, len(v))
                                  for k, v in iteritems(self.ds.dimensions))
 
+    def get_encoding(self):
+        encoding = {}
+        encoding['unlimited_dims'] = set(
+            [k for k, v in self.ds.dimensions.items() if v.isunlimited()])
```
shoyer: you can use a set comprehension here, e.g.,

    encoding['unlimited_dims'] = {k for k, v in self.ds.dimensions.items()
                                  if v.isunlimited()}

xarray/core/dataset.py
```diff
@@ -306,7 +306,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
     groupby_cls = groupby.DatasetGroupBy
 
     def __init__(self, data_vars=None, coords=None, attrs=None,
-                 compat='broadcast_equals'):
+                 compat='broadcast_equals', encoding=None):
```
shoyer: Needs to be documented in the docstring.

Or we could even remove it entirely from the constructor and require setting the property.

xarray/core/dataset.py
```diff
@@ -405,6 +408,18 @@ def attrs(self, value):
         self._attrs = OrderedDict(value)
 
     @property
+    def encoding(self):
+        """Dictionary of global encoding attributes on this dataset
+        """
+        if self._encoding is None:
+            self._encoding = dict()
```
shoyer: could use `self._encoding = {}`

xarray/backends/h5netcdf_.py
shoyer: Actually, `check_encoding` is specific to variable encoding. Raising an error would make sense if you set `unlimited_dims` via an argument in `to_netcdf` (which is not yet possible).

xarray/core/dataset.py
jhamman: I think I'd like to keep it here. Often times, I end up batch constructing objects with something like `xr.Dataset(..., **kwargs)` where `kwargs = {..., 'encoding': {'unlimited_dims': ['time']}}`. 

Also, since we include `encoding` in the `DataArray` constructor, it seems appropriate to have it here as well.

shoyer: I think consistency makes for a good argument here, though I would consider deprecating the `encoding` argument to `DataArray` instead. It would also make sense to get rid of the `compat` argument to `Dataset`.

These extra arguments are not part of the fundamental xarray data model and thus are a little distracting, especially to new users.

jhamman: I'm going to open a separate issue to discuss this.

jhamman:
@shoyer - comments addressed and green.

xarray/core/dataset.py
jhamman: I've removed the encoding keyword arg for now.

xarray/backends/pynio_.py
```diff
@@ -57,5 +58,11 @@ def get_attrs(self):
     def get_dimensions(self):
         return Frozen(self.ds.dimensions)
 
+    def get_encoding(self):
+        encoding = {}
+        encoding['unlimited_dims'] = set(
+            [k for k in self.ds.dimensions if self.ds.unlimited(k)])
```
shoyer: I don't think dap can represent unlimited dimensions:
http://docs.opendap.org/index.php/DAP4:_Specification_Volume_1#Dimensions

xarray/backends/scipy_.py
```diff
@@ -116,9 +117,19 @@ def get_variables(self):
...
+        return set(k for k, v in iteritems(self.ds.dimensions) if v is None)
+
     def get_dimensions(self):
+        self._unlimited_dimensions = self._get_unlimited_dimensions()
         return Frozen(self.ds.dimensions)
 
+    def get_encoding(self):
+        encoding = {}
+        encoding['unlimited_dims'] = set(
+            [k for k, v in self.ds.dimensions.items() if v is None])
```
shoyer: can use the same set comprehension you switched to in `netCDF4_.py`

xarray/backends/scipy_.py
```diff
@@ -116,9 +117,19 @@ def get_variables(self):
     def get_attrs(self):
         return Frozen(_decode_attrs(self.ds._attributes))
 
+    def _get_unlimited_dimensions(self):
+        return set(k for k, v in iteritems(self.ds.dimensions) if v is None)
+
     def get_dimensions(self):
+        self._unlimited_dimensions = self._get_unlimited_dimensions()
```
shoyer: you don't use this currently

xarray/core/dataset.py
```diff
@@ -914,12 +929,18 @@ def to_netcdf(self, path=None, mode='w', format=None, group=None,
             Nested dictionary with variable names as keys and dictionaries of
             variable specific encodings as values, e.g.,
             ``{'my_variable': {'dtype': 'int16', 'scale_factor': 0.1, 'zlib': True}, ...}``
+        unlimited_dims : str or sequence of str, optional
```
shoyer: I think it needs to be a `sequence of str`, not a `str`.

xarray/backends/netCDF4_.py
```diff
@@ -251,6 +251,12 @@ def get_dimensions(self):
         return FrozenOrderedDict((k, len(v))
                                  for k, v in iteritems(self.ds.dimensions))
 
+    def get_encoding(self):
```
shoyer: I would lean slightly toward just creating a `get_unlimited_dims` method rather than `get_encoding`, unless we can think of other Dataset wide encodings we might possibly add in the future.

xarray/backends/api.py
```diff
@@ -565,8 +565,11 @@ def to_netcdf(dataset, path=None, mode='w', format=None, group=None,
     sync = writer is None
 
     store = store_cls(path, mode, format, group, writer)
+    # Copy dataset encoding to datastore
+    store.encoding = dataset.encoding
```
shoyer: Do we ever actually use this `encoding` state on the datastore? If not, let's not bother setting it. I think everything necessary ends up being passed on via `set_variables`.

Note that as bunch as possible, I've tried to make `DataStore` itself stateless, only storing state in the file-like object it points to.

xarray/backends/common.py
```diff
@@ -96,8 +98,9 @@ def load(self):
         This function will be called anytime variables or attributes
         are requested, so care should be taken to make sure its fast.
         """
-        variables = FrozenOrderedDict((_decode_variable_name(k), v)
-                                      for k, v in iteritems(self.get_variables()))
+        self.encoding = self.get_encoding()
```
shoyer: This is a little dangerous -- `.load()` needs to be called in order to guarantee a consistent `encoding`  state on a DataStore. I would rather we didn't set such state, and simply pulled this information out of the file linked to the DataStore as necessary.

xarray/backends/pynio_.py
jhamman: Agreed, but this is pynio which does: https://www.pyngl.ucar.edu/whatsnew.shtml#Version1.4.1

xarray/backends/netCDF4_.py
jhamman: The other encoding value that comes to mind is the dataset format (e.g. NETCDF4 vs. NETCDF3). Maybe there are others as well but nothing is mind.

xarray/backends/common.py
jhamman: Fair point, I've removed the `encoding` attribute on the `DataStore`.

xarray/backends/api.py
jhamman: We were using this but I've refactored to avoid it.

jhamman:
@shoyer - all comments addressed and tests are passing.

xarray/backends/pydap_.py
```diff
@@ -62,6 +62,7 @@ class PydapDataStore(AbstractDataStore):
     def __init__(self, url):
         import pydap.client
         self.ds = pydap.client.open_url(url)
+        self.encoding = {}
```
shoyer: delete

xarray/backends/pynio_.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, filename, mode='r'):
         self.ds = opener()
         self._opener = opener
         self._mode = mode
+        self.encoding = {}
```
shoyer: delete

xarray/backends/scipy_.py
```diff
@@ -102,6 +102,7 @@ def __init__(self, filename_or_obj, mode='r', format=None, group=None,
         self.ds = opener()
         self._opener = opener
         self._mode = mode
+        self.encoding = {}
```
shoyer: delete

xarray/backends/memory.py
```diff
@@ -21,6 +21,7 @@ class InMemoryDataStore(AbstractWritableDataStore):
     def __init__(self, variables=None, attributes=None, writer=None):
         self._variables = OrderedDict() if variables is None else variables
         self._attributes = OrderedDict() if attributes is None else attributes
+        self.encoding = {}
```
shoyer: delete

xarray/backends/scipy_.py
```diff
@@ -116,9 +117,18 @@ def get_variables(self):
     def get_attrs(self):
         return Frozen(_decode_attrs(self.ds._attributes))
 
+    def _get_unlimited_dimensions(self):
```
shoyer: I don't think you use this method anymore

xarray/backends/h5netcdf_.py
```diff
@@ -58,6 +59,7 @@ def __init__(self, filename, mode='r', format=None, group=None,
         self._opener = opener
         self._filename = filename
         self._mode = mode
+        self.encoding = {}
```
shoyer: This still should go away :)

xarray/test/test_backends.py
```diff
@@ -972,6 +983,14 @@ def test_read_byte_attrs_as_unicode(self):
             expected = Dataset(attrs={'foo': 'bar'})
             self.assertDatasetIdentical(expected, actual)
 
+    def test_encoding_unlimited_dims(self):
+        ds = Dataset({'x': ('y', np.arange(10.0))})
+        ds.encoding = {'unlimited_dims': ['y']}
+        with pytest.warns(UserWarning):
+            ds.to_netcdf('foo-bar.nc', engine='h5netcdf')
```
shoyer: use the `create_tmp_file()` context manager to ensure this file gets cleaned up

jhamman:
In we go. Thanks @shoyer for putting up with my short attention span on this one.

Issue #1015, opened 2016-09-22
'levels' tick labels for discrete colormaps


paulvija:
Hi there,

'levels' tick labels for discrete colormaps works fine for 10 intervals and tend to shorten by itself beyond 10. This behavior may exist to make colorbar exceeding 10 intervals look tidy.

However, for **discrete** colormaps it will be good to label the ticks for all intervals as they can be any number and not uniform.
xray.**version** = '0.8.2'

Please see the screenshots attached for a better understanding.

Thanks,
Vijay Paul
![image](https://cloud.githubusercontent.com/assets/12703555/18734627/6bba1b54-80ca-11e6-9121-d3138cc71af1.png)

![image](https://cloud.githubusercontent.com/assets/12703555/18734553/bf405c1c-80c9-11e6-8efa-6aa7db1afad3.png)
![image](https://cloud.githubusercontent.com/assets/12703555/18734557/ca46f530-80c9-11e6-8c6c-4c0c1985ab41.png)

shoyer:
Agreed, this would be nice. See also #920.

fmaussion:
Note that currently the colorbar labelling logic is handed over entirely to matplotlib, so its mpl which decides upon this. With the ``cbar_kwargs`` argument you can control this at your whish, so I'm not sure we have to change anything:

```python
from matplotlib import pyplot as plt
import xarray as xr

data = xr.DataArray(np.arange(12).reshape((3, 4))*10 + 10)
levels = np.append(np.arange(10)*10 + 10, 115)

f, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(14, 4))

data.plot(ax=ax1, levels=levels)
data.plot(ax=ax2, levels=levels, cbar_kwargs={'ticks':levels})
data.plot(ax=ax3, levels=levels, cbar_kwargs={'ticks':levels, 'spacing':'proportional'})
plt.tight_layout()
```
![cb](https://cloud.githubusercontent.com/assets/10050469/18742313/c72b2832-80b3-11e6-9ded-1c21875047c8.png)

paulvija:
**cbar_kwargs works well for my purpose. Thanks for the help.** 
Can this usage be shown in the documentation, at least mentioning that colorbar intervals can be placed at our wish or this comes under matplotlib?

shoyer:
@paulvija @fmaussion either of you interested in putting together a PR? :)

paulvija:
PR?

[...]

PR #1171, opened 2016-12-18
Fix tests for upcoming matplotlib v2

fmaussion:
Just to see what happens with the upcoming matplotlib version 2


fmaussion:
... three failing tests. Not too bad!

jhamman:
@fmaussion - do you want to take point on resolving these failing tests?

fmaussion:
@jhamman I planned to do so but I won't come to this in the next couple of days, so if its urgent and you want to do it go ahead! They should be easy fixes

jhamman:
@fmaussion  Great. No rush.

.travis.yml
```diff
@@ -30,6 +30,8 @@ matrix:
     env: CONDA_ENV=py35-dask-dev
   - python: 3.5
     env: CONDA_ENV=py35-pandas-dev
+  - python: 3.5
+    env: CONDA_ENV=py35-mpl2
```
jhamman: let's move this to the `allow_failures` set. This is typically how we handle development builds.

fmaussion:
This fixes the failing tests on MPL2. Two of the tests were failing because of how mpl chooses to do things, which makes me think that the tests are not very robust. My fixes are also rather simple...

For example the ``test_discrete_colormap_int_levels`` test, reproduced here:

```python
import xarray as xr
import numpy as np
import matplotlib

x = np.arange(start=0, stop=10, step=2)
y = np.arange(start=9, stop=-7, step=-3)
xy = np.dstack(np.meshgrid(x, y))
distance = np.linalg.norm(xy, axis=2)
darray = xr.DataArray(distance, list(zip(('y', 'x'), (y, x))))

darray.plot(levels=7, vmin=4, vmax=8, extend='both')
plt.title('mpl {}'.format(matplotlib.__version__))
```

produces:

mpl1 5 3
mpl2 0 0rc2

.travis.yml
fmaussion: Yes. I also renamed the environment, which is set to always use the release candidates when available. I don't know if it makes sense to keep this test environment once mpl v2 is released...

shoyer:
@fmaussion I think this is a bug, either with xarray plotting or matplotlib 2. If you supply `vmin` and `vmax` explicitly, shouldn't those be the minimum and maximum tick labels?

Also, shouldn't you get the requested number of levels? In matplotlib 2.0, I see 9 levels, not 7.

fmaussion:
>  I think this is a bug

Yes, I should've looked into this in more detail. So the problem is [here](https://github.com/pydata/xarray/blob/master/xarray/plot/utils.py#L191). ``MaxNLocator.tick_values(vmin, vmax)`` does not guarantee to land on ``vmin`` or ``vmax`` exactly, which I think is confusing. It should be either changed or documented in xarray. The change could look like: if ``vmin`` and ``vmax`` are set _explicitely_, than we use ``linspace`` to build the levels.

Furthermore, ``MaxNLocator`` is no longer guaranteed to return a max of N levels in MPLv2, as shown by this simple test:

```python
import matplotlib as mpl
ticker = mpl.ticker.MaxNLocator(7)
levels = ticker.tick_values(4, 8)
print('mpl: {}, nlevels: {}, levels: {}'.format(mpl.__version__, len(levels), levels))
```

Which prints, depending on mpl version:

```
mpl: 1.5.3, nlevels: 7, levels: [ 3.2  4.   4.8  5.6  6.4  7.2  8. ]
```

```
mpl: 2.0.0rc2, nlevels: 9, levels: [ 3.6  4.2  4.8  5.4  6.   6.6  7.2  7.8  8.4]
```

fmaussion:
I've found this [ticket](https://github.com/matplotlib/matplotlib/issues/7578), but it's not clear to me if the problem is solved or not. I'll have to test against mpl2 master I guess...

 @QuLogic do you maybe have a moment to have a look at my comment above?

shoyer:
The Matplotlib 2.0 release notes mention that the algorithm for the number of ticks in `MaxNLocator` [has been changed](http://matplotlib.org/2.0.0rc2/users/dflt_style_changes.html#number-of-ticks) but this seems at odds with the [documented behavior](http://matplotlib.org/2.0.0rc2/api/ticker_api.html#matplotlib.ticker.MaxNLocator) for specifying a minimum number of intervals.

QuLogic:
That does seem like a bug, but I'm not the one to ask. I'd file a bug and ping @efiring.

shoyer:
@fmaussion as you note, even if this is a matplotlib bug, it seems like this was not doing quite the desired thing in xarray even with matplotlib 1.5 -- we really should be using `np.linspace` for this, not `MaxNLocator`.

fmaussion:
@QuLogic : thanks, and sorry for the noise.

@shoyer : should we use ``linspace`` only if ``vmax`` and/or ``vmin`` are set explicitly, or always? ``MaxNLocator`` definitely makes "nicer" levels, but as stated by the name it might choose less than N levels if needed...

On a side note, I already encountered this problem with my students before: what most users will do in this case is to set ``levels=np.linspace(foo, bar)`` themselves

shoyer:
My feeling is that `vmin` and `vmax` should always be exactly at the specified locations, but the location of the ticks in between is a little more flexible. If both `vmin`/`vmax` are set, then using linspace seems obvious, and if neither are set, then perhaps we should differ to whatever matplotlib defaults to. If only one is set, the solution is less clear. Maybe one limit be chosen exactly and the other in a "nice" way?

efiring:
It's not a bug in the code.  The purpose of a Locator is to find "nice" values for ticks.  It is used in two related ways: to determine the axis limits when the pre-2.0 default of "round numbers" (meaning "landing on ticks")  is use, and to find the tick locations themselves.  Supporting the former application requires generating ticks beyond the specified limits; then later in the plotting process, the excess ticks might be used, or they might be trimmed off, depending on how the axis limits are determined.  In no case is it *intended* that ticks land on specified vmin and/or vmax values, though it can happen by chance if they are "nice" numbers.

In the case of the MaxNLocator, the "N" refers to intervals, not ticks; that is, in "round-number" axis limit mode, or if vmin and vmax otherwise land on ticks, there can be N+1 ticks. (This is in the docstring; and the argument is called "nbins", not "nticks".)

fmaussion:
@efiring thanks fir jumping in! I understand and I see that xarray is kind of misusing this tool

In the test above ``MaxNLocator`` returns N+2 ticks though.

fmaussion:
@shoyer 

> if neither are set, then perhaps we should differ to whatever matplotlib defaults to

AFAIK, there is no logic for ``nlevels=int`` in mpl. Furthermore, mpl uses ``vmin`` and ``vmax`` to clip the data (not to decide upon the levels), which is undesirable for us.

However, the default mpl behavior when no levels are set is to call the private method [_autolev](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/contour.py#L1119) which has a logic based on ``MaxNLocator`` too.

Long story short: I'll start another PR to address this problem specifically and then go back to MPL2 later on.

efiring:
With a vmax-vmin of 4, it is using an increment of 0.6, for a span of 4.2.  In other words, `vmax-vmin < 7 * 0.6`, and 0.6 is the smallest available increment for which this is true.  This is the intended behavior; so yes, the algorithm can return up to N+2 ticks, not just N+1.

efiring:
It looks like what you are trying to do here is analogous to the way contour levels are chosen in matplotlib's contourf.

jhamman:
I've never been completely satisfied with what we came up with for the vmin/vmax/levels logic. Its been the source of a few bugs and confusion (#1015). The feature set we were going for is mostly covered (discrete colorbar, flexibility in specification of levels, etc.) but we may want to revisit the approach.

xarray/test/test_plot.py
```diff
@@ -1116,7 +1121,7 @@ def test_figure_size(self):
             g = xplt.plot(self.darray, row=2, col='z', ax=plt.gca(), size=6)
 
     def test_num_ticks(self):
-        nticks = 100
+        nticks = 99
```
fmaussion: This change is also because ``MaxNLocator`` has changed in mplv2 ans is not guaranteed to always return a max of N intervals. With 99 it works, by chance I suppose...

fmaussion:
OK, so thanks to https://github.com/pydata/xarray/pull/1191 the required changes for the tests to pass in mpl v2 are minimal:
- deprecation of ``get_axis_bgcolor``
- ``test_num_ticks`` also changed due to changes in ``MaxNLocator``, but this is not a big deal

This is ready for review / merge

PR #1175, opened 2016-12-21
End support for py3.3?

MaximilianR:
Even at the beginning of this year its downloads were 90% lower than 2.6: https://hynek.me/articles/python3-2016/

Impetus is that pytest no longer supports it, and some of the tests I recently wrote don't work on 3.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ matrix:
     env: CONDA_ENV=py27-min
   - python: 2.7
     env: CONDA_ENV=py27-cdat+pynio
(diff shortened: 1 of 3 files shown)


jhamman:
Python 3.6.0 is scheduled for final release on Friday. Is conda-forge building packages for it yet?  If so, perhaps we just replace 3.3 with 3.6?

MaximilianR:
Yes we should add a 3.6 env. Not sure if it's on conda-forge yet.

shoyer:
Sounds good to me!

Issue #820, opened 2016-04-08
Dataset.__repr__ causing dask to be computed.


andrewdhicks:
Printing a Dataset object will show array values (#206), but it will also cause a dask-backed array to be computed so that the data variable values can be computed:

```
<xarray.Dataset>
Dimensions:  (time: 182, x: 420, y: 489)
Coordinates:
  * time     (time) datetime64[ns] 1990-03-02T23:11:16 1990-03-02T23:11:39 ...
  * y        (y) float64 -3.919e+06 -3.919e+06 -3.919e+06 -3.919e+06 ...
  * x        (x) float64 1.538e+06 1.538e+06 1.538e+06 1.538e+06 1.538e+06 ...
Data variables:
    band_20  (time, y, x) float64 729.0 612.0 579.0 629.0 862.0 1.027e+03 ...
    band_10  (time, y, x) float64 476.0 375.0 357.0 416.0 586.0 708.0 730.0 ...
    band_50  (time, y, x) float64 2.175e+03 634.0 245.0 546.0 1.788e+03 ...
...
```

When a DataArray is printed, it doesn't compute values when using dask:

```
<xarray.DataArray u'ls5_nbar_albers' (time: 182, y: 489, x: 420)>
dask.array<concate..., shape=(182, 489, 420), dtype=float64, chunksize=(1, 489, 420)>
Coordinates:
  * time      (time) datetime64[ns] 1990-03-02T23:11:16 1990-03-02T23:11:39 ...
  * y         (y) float64 -3.919e+06 -3.919e+06 -3.919e+06 -3.919e+06 ...
  * x         (x) float64 1.538e+06 1.538e+06 1.538e+06 1.538e+06 1.538e+06 ...
```

There is a check to make sure the data is not remote, but not based on dask status, see: https://github.com/pydata/xarray/blob/master/xarray/core/formatting.py#L173
Is there a way to indicate that computing a particular dask is an expensive operation and it should not be calculated?

shoyer:
Showing a preview of values in a dask dataset can be very convenient for interactive use, so I'm loathe to turn it off entirely. I agree that it's not always useful, though.

The current check for remote data is a complete hack that should probably be removed :).

What do you think about adding an user configurable option to disable printing lazily computed values in datasets? You could then write something like `xarray.set_options(preview_lazy_data=False)`:
http://xarray.pydata.org/en/stable/generated/xarray.set_options.html

andrewdhicks:
[...]

PR #1176, opened 2016-12-21
add info method to dataset

### LINKED ISSUE #244: we need some way to see the attributes on all arrays in a dataset
Options:
1. Display array attributes in the dataset repr, e.g., perhaps something like this:
   
   ```
   <xray.Dataset>
   Dimensions:         (time: 3, x: 2, y: 2)
   Coordinates:
       lat             (x, y) float64 42.25 42.21 42.63 42.59
       reference_time  datetime64[ns] 2014-09-05
       lon             (x, y) float64 -99.83 -99.32 -99.79 -99.23
     * time            (time) datetime64[ns] 2014-09-06 2014-09-07 2014-09-08
     * x               (x) int64 0 1
     * y               (y) int64 0 1
   Variables:
       precipitation   (x, y, time) float64 5.904 2.453 3.404 9.847 9.195 0.3777 8.615 7.536 ...
       temperature     (x, y, time) float64 11.04 23.57 20.77 9.346 6.683 17.17 11.6 19.54 ...
   Attributes:
       lat:
           units: degrees_east
       lon:
           units: degrees_north
       precipitation:
           units: mm
       temperature:
           units: degrees_C
   ```
2. Add a method or attribute like `attrs_map` that collects everything together into a single mapping (sort of like DAS in OpenDAP: http://www.opendap.org/api/javaDocs/opendap/dap/DAS.html).

I am leaning toward the later option right now.

### LINKED ISSUE #1150: "ncdump -h" like repr?
Sometimes it could be useful to have a view of all variables attributes at a glance. For example, this is the repr for ERA-Interim energy fluxes data:

```
(...)
Data variables:
    slhf     (month, latitude, longitude) float64 0.02852 0.02852 0.02852 ...
    tsr      (month, latitude, longitude) float64 -0.0001912 -0.0001912 ...
    strd     (month, latitude, longitude) float64 166.6 166.6 166.6 166.6 ...
    strc     (month, latitude, longitude) float64 -66.23 -66.23 -66.23 ...
    tisr     (month, latitude, longitude) float64 0.0 0.0 0.0 0.0 0.0 0.0 ...
    ssrd     (month, latitude, longitude) float64 -0.0003951 -0.0003951 ...
    ssrc     (month, latitude, longitude) float64 0.0 0.0 0.0 0.0 0.0 0.0 ...
    str      (month, latitude, longitude) float64 -40.65 -40.65 -40.65 ...
    ttr      (month, latitude, longitude) float64 -171.5 -171.5 -171.5 ...
    tsrc     (month, latitude, longitude) float64 0.0 0.0 0.0 0.0 0.0 0.0 ...
    sshf     (month, latitude, longitude) float64 10.46 10.46 10.46 10.46 ...
    ssr      (month, latitude, longitude) float64 0.0 0.0 0.0 0.0 0.0 0.0 ...
    ttrc     (month, latitude, longitude) float64 -174.9 -174.9 -174.9 ...
(...)
```
This is what my students will see when they explore the dataset for the first time. It could be nice to have a utility function (e.g. ``dumph`` or something) which would have a style closer to ``ncdump -h``:

```
(...)
	double ttr(month, latitude, longitude) ;
		ttr:least_significant_digit = 2L ;
		ttr:units = "J m**-2" ;
		ttr:long_name = "Top net thermal radiation" ;
		ttr:standard_name = "toa_outgoing_longwave_flux" ;
	double tsrc(month, latitude, longitude) ;
		tsrc:least_significant_digit = 2L ;
		tsrc:units = "J m**-2" ;
		tsrc:long_name = "Top net solar radiation, clear sky" ;
	double sshf(month, latitude, longitude) ;
		sshf:least_significant_digit = 2L ;
		sshf:units = "J m**-2" ;
		sshf:long_name = "Surface sensible heat flux" ;
		sshf:standard_name = "surface_upward_sensible_heat_flux" ;
	double ssr(month, latitude, longitude) ;
		ssr:least_significant_digit = 2L ;
		ssr:units = "J m**-2" ;
		ssr:long_name = "Surface net solar radiation" ;
		ssr:standard_name = "surface_net_downward_shortwave_flux" ;
	double ttrc(month, latitude, longitude) ;
		ttrc:least_significant_digit = 2L ;
		ttrc:units = "J m**-2" ;
		ttrc:long_name = "Top net thermal radiation, clear sky" ;
(...)
```

Or is there something like this already?

jhamman:
I don't know if this is exactly what we want but here's an idea that emulates `ncdump -h`. I'm sure people will have thoughts on the implementation and output so I'll just throw this first cut up and let people discuss.

closes #1150 
fixes #244

xref: #1044, #820

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -804,7 +804,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
     def __unicode__(self):
         return formatting.dataset_repr(self)
 
-    def attr_info(self, buf=None):
+    def info(self, buf=None):
         """
         Concise summary of a Dataset variables and attributes.
         Parameters
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -839,8 +839,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
             lines.append(u'\t:{k} = {v} ;'.format(k=k, v=v))
         lines.append(u'}')
 
-        lines = [ensure_valid_repr(line) for line in lines]
-        buf.write('\n'.join(lines))
+        buf.write(u'\n'.join(lines))
 
     @property
     def chunks(self):
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -5,6 +5,8 @@ import functools
 from collections import Mapping
 from numbers import Number
 
+import sys
+
 import numpy as np
 import pandas as pd
 
@@ -801,6 +803,42 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
 
     def __unicode__(self):
         return formatting.dataset_repr(self)
+
+    def attr_info(self, buf=None):
+        """
+        Concise summary of a Dataset variables and attributes.
+        Parameters
+        ----------
+        buf : writable buffer, defaults to sys.stdout
+
+        See Also
+        --------
+        pandas.DataFrame.assign
+        netCDF's ncdump
+        """
+
+        if buf is None:  # pragma: no cover
+            buf = sys.stdout
+
+        lines = []
+        lines.append('xarray.Dataset {')
+        lines.append('dimensions:')
+        for name, size in self.dims.items():
+            lines.append('\t{name} = {size} ;'.format(name=name, size=size))
+        lines.append('\nvariables:')
+        for name, da in self.variables.items():
+            dims = ', '.join(da.dims)
+            lines.append('\t{type} {name}({dims}) ;'.format(
+                type=da.dtype, name=name, dims=dims))
+            for k, v in da.attrs.items():
+                lines.append('\t\t{name}:{k} = {v} ;'.format(name=name, k=k,
+                                                             v=v))
+        lines.append('\n// global attributes:')
+        for k, v in self.attrs.items():
+            lines.append('\t:{k} = {v} ;'.format(k=k, v=v))
+        lines.append('}')
+
+        formatting._put_lines(buf, lines)
 
     @property
     def chunks(self):
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -27,9 +27,9 @@ from .utils import (Frozen, SortedKeysDict, maybe_wrap_array, hashable,
 from .variable import (Variable, as_variable, IndexVariable, broadcast_variables)
 from .pycompat import (iteritems, basestring, OrderedDict,
                        dask_array_type, range)
+from .formatting import ensure_valid_repr
 from .combine import concat
 from .options import OPTIONS
-
 
 # list of attributes of pd.DatetimeIndex that are ndarrays of time info
 _DATETIMEINDEX_COMPONENTS = ['year', 'month', 'day', 'hour', 'minute',
@@ -807,6 +807,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
     def info(self, buf=None):
         """
         Concise summary of a Dataset variables and attributes.
+
         Parameters
         ----------
         buf : writable buffer, defaults to sys.stdout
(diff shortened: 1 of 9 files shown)


MaximilianR:
Nice!

Why pass a buffer into that function though? Why not return a string and the user do what they want with it?

jhamman:
Good question. I tried to merge the output format from `ncdump -h` with the api of [`Pandas.DataFrame.info`](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.info.html). I'm certainly not tied to the return format I choose.

shoyer:
I never noticed that DataFrame.info() doesn't actually have a return value. It does seems strange to write to a buffer taken as an argument but I can see one reason why it sort of makes sense -- otherwise you get quotation marks printed around the returned string.

MaximilianR:
Yes interesting, I didn't know that either. 
IIRC there are also methods that are ipython & jupyter specific, that can return repl & notebook outputs

fmaussion:
Thanks @jhamman ! 

About the name: what about just ``.info()``, or ``.var_info()`` ? the additional info (which isn't in the repr already) is more about variables, right?

jhamman:
I'm also thinking `.info()` seems like a better name. It would be fairly trivial to add additional argument options to fill in the pandas `.info()` functionality.

jhamman:
Tests are passing on Python 2 now.  

@shoyer - how do you feel about `info()` for the method name?

shoyer:
I'm pretty happy with `.info()`, as long as that indicates that we aren't wedded to a display matching `ncdump -h`. In particular displaying the size of each variable in memory in appropriate units could be a nice addition (for later, doesn't need to be in this PR).

xarray/core/dataset.py
```diff
@@ -802,6 +804,42 @@ def to_netcdf(self, path=None, mode='w', format=None, group=None,
...
+                type=da.dtype, name=name, dims=dims))
+            for k, v in da.attrs.items():
+                lines.append('\t\t{name}:{k} = {v} ;'.format(name=name, k=k,
+                                                             v=v))
+        lines.append('\n// global attributes:')
+        for k, v in self.attrs.items():
+            lines.append('\t:{k} = {v} ;'.format(k=k, v=v))
+        lines.append('}')
+
+        formatting._put_lines(buf, lines)
```
shoyer: I would probably just use `buf.write(u'\n'.join(lines))` here, which will work as long as you ensure all elements in `lines` are the same (unicode/str) type.

xarray/core/dataset.py
```diff
@@ -802,6 +804,42 @@ def to_netcdf(self, path=None, mode='w', format=None, group=None,
...
+        """
+
+        if buf is None:  # pragma: no cover
+            buf = sys.stdout
+
+        lines = []
+        lines.append('xarray.Dataset {')
+        lines.append('dimensions:')
+        for name, size in self.dims.items():
+            lines.append('\t{name} = {size} ;'.format(name=name, size=size))
```
shoyer: I think these probably should all be unicode (using `u` literals), otherwise this will break for non-ASCII characters on Python 2. Take a look at what things look like in `formatting.py`.

xarray/test/test_dataset.py
```diff
@@ -12,6 +12,10 @@
     import dask.array as da
 except ImportError:
     pass
+try:
+    from io import StringIO
```
shoyer: we might put this in in `pycompat` instead

xarray/core/dataset.py
shoyer: It would also be good to add a test for attributes with non-ASCII values.

xarray/core/dataset.py
```diff
@@ -802,6 +804,44 @@ def to_netcdf(self, path=None, mode='w', format=None, group=None,
...
+                type=da.dtype, name=name, dims=dims))
+            for k, v in da.attrs.items():
+                lines.append(u'\t\t{name}:{k} = {v} ;'.format(name=name, k=k,
+                                                              v=v))
+        lines.append(u'\n// global attributes:')
+        for k, v in self.attrs.items():
+            lines.append(u'\t:{k} = {v} ;'.format(k=k, v=v))
+        lines.append(u'}')
+
+        lines = [ensure_valid_repr(line) for line in lines]
```
shoyer: I don't think we need this. I think you can write unicode to string buffers, even on Python 2 (at least it works for `sys.stdout` and `StringIO`).

jhamman: Without it, we get: 

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xae' in position 363: ordinal not in range(128)```

I've always had a tough time getting the string/bytes/unicode stuff straight so I'm open to other ideas here.

shoyer: Ah, the issue is that you're using `StringIO` from  `cStringIO`, which only handles bytes, not unicode. Instead, use `StringIO` from `io`, on both Python 2 and 3 (no need for the separate compatibility module even).

jhamman:
@shoyer --- green.

PR #1178, opened 2016-12-21
Fix inconsistent behaviour for xarray.DataArray.rename

### LINKED ISSUE #1177: Inconsistent behaviour for xarray.DataArray.rename
Hi,

thanks for creating `xarray`, I like it a lot. However I want to ask, why the behaviour of `xarray.DataArray.rename` is so incosistent:

```python
import xarray as xr
x = xr.DataArray([1, 2, 3], dims=('a',), coords={'a': [1, 2, 3]}, name='a')
x.rename({'a': 'b'})
```

renames the name of `x` not the coordinate! Could you explain the behaviour to me? I do not understand it, or consider it as inconsistent. 

Thanks, Markus

magonser:
Attempt to fix #1177

Please let me know, if PR requires further refinement.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -788,37 +788,47 @@ class DataArray(AbstractArray, BaseDataObject):
             method=method, tolerance=tolerance, copy=copy, **indexers)
         return self._from_temp_dataset(ds)
 
-    def rename(self, new_name_or_name_dict):
+    def rename(self, new_name_or_dims_dict, new_dims_dict=None):
         """Returns a new DataArray with renamed coordinates and/or a new name.
 
 
         Parameters
         ----------
-        new_name_or_name_dict : str or dict-like
+        new_name_or_dims_dict : str or dict-like
             If the argument is dict-like, it it used as a mapping from old
-            names to new names for coordinates (and/or this array itself).
-            Otherwise, use the argument as the new name for this array.
+            names to new names for coordinates. Otherwise, use the argument
+            as the new name for this array.
+
+        new_dims_dict : dict-like
+            Used as a mapping from old names to new names for coordinates.
+
 
         Returns
         -------
         renamed : DataArray
-            Renamed array or array with renamed coordinates.
+            Array with renamed name and/or coordinates.
 
         See Also
         --------
         Dataset.rename
         DataArray.swap_dims
         """
-        if utils.is_dict_like(new_name_or_name_dict):
-            name_dict = new_name_or_name_dict.copy()
-            if self.name in self.dims:
-                name = name_dict.get(self.name, self.name)
-            else:
-                name = name_dict.pop(self.name, self.name)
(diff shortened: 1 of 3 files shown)


xarray/core/dataarray.py
```diff
@@ -788,37 +788,47 @@ def reindex(self, method=None, tolerance=None, copy=True, **indexers):
...
         ----------
-        new_name_or_name_dict : str or dict-like
+        new_name_or_dims_dict : str or dict-like
             If the argument is dict-like, it it used as a mapping from old
-            names to new names for coordinates (and/or this array itself).
-            Otherwise, use the argument as the new name for this array.
+            names to new names for coordinates. Otherwise, use the argument
+            as the new name for this array.
+
+        new_dims_dict : dict-like
```
shoyer: Instead of adding a second argument here, I would simply require calling `.rename()` twice if you want to rename both the DataArray and it's coordinates at the same time.

magonser: I've been thinking about that initially, too. Fixed in f855063.

xarray/core/dataarray.py
```diff
@@ -788,16 +788,17 @@ def reindex(self, method=None, tolerance=None, copy=True, **indexers):
             method=method, tolerance=tolerance, copy=copy, **indexers)
         return self._from_temp_dataset(ds)
 
-    def rename(self, new_name_or_name_dict):
-        """Returns a new DataArray with renamed coordinates and/or a new name.
+    def rename(self, new_name_or_dims_dict):
```
shoyer: Let's keep this parameter name the same as before.

xarray/core/dataarray.py
```diff
@@ -811,12 +812,8 @@ def rename(self, new_name_or_name_dict):
         """
         if utils.is_dict_like(new_name_or_name_dict):
             name_dict = new_name_or_name_dict.copy()
```
shoyer: It occurs to me that the copying here is no longer necessary, since we no longer modify the dictionary.

magonser: I have pushed another commit fa6b02c to remove it. Keeping the PR open for another while is fine by me.

fmaussion:
Slightly OT, but it occurred to me that we took the habit not to add an attribution (e.g. "By bla") to the "Breaking changes" list. This is probably nicer for the people who are quickly reading through the changes, but not really consistent with the bugfixes or enhancements.

In that case, I think you could list your change in the "bugfix" list rather than "breaking changes"

shoyer:
@fmaussion that was my mistake here -- I didn't give myself credit on a few of the "Breaking Changes" but we should fix that.

@magonser want to insert a line giving yourself credit for this change?

magonser:
Alright, I've added attribution. Thanks by the way for being patient with me in this PR, as I am just getting started on open source contributions.

shoyer:
Thanks @magonser, we really appreciate PRs, especially from new contributors!

Issue #1173, opened 2016-12-19
Some queries


JoyMonteiro:
Hello @shoyer @pwolfram @mrocklin @rabernat ,

I was trying to write a design/requirements doc with ref. to the Columbia meetup,
and I had a few queries, on which I wanted your inputs (basically to ask whether
they make sense or not!)

1. If you serialize a labeled n-d data array using netCDF or HFD5, it gets written into
a single file, which is not really a good option if you want to eventually do distributed
processing of the data. Things like HDFS/lustreFS can split files, but that is not really
what we want. How do you think this issue could be solved within the xarray+dask
framework? 
   * is it a matter of adding some code to the dataset.to_netcdf() method or
     adding a new method that would split the DataArray (based on some user guidelines) into multiple files?
   * Or does it make more sense to add a new serialization format like Zarr?
2. Continuing along similar lines, how does xarray+dask currently decide on how to distribute the workload between dask workers? are there any heuristics to handle data locality? or does experience say that network I/O is fast enough that this is not an issue? I'm asking this question because of this article by Matt: http://blaze.pydata.org/blog/2015/10/28/distributed-hdfs/
   * If this is desirable, how would one go about implementing it?

shoyer:
For (1), take a look at [save_mfdataset](http://xarray.pydata.org/en/stable/generated/xarray.save_mfdataset.html) for saving to multiple files.

JoyMonteiro:
did not know about that, thanks!!

JoyMonteiro:
@shoyer: does this also work with dask.distributed? The doc seems to only mention a thread pool.

JoyMonteiro:
Thanks. how big of an endeavour is this? I see some free time from 2-3rd week of Jan, and
I could maybe contribute towards making this happen.

mrocklin:
Some related issues:

https://github.com/pydata/xarray/issues/798
https://github.com/dask/distributed/issues/629

Will write more shortly.

mrocklin:
There have been some efforts and progress in using many NetCDF files on a distributed POSIX filesystem (NFS, gluster, not HDFS) but there is still some pain here.  We should probably circle back up and figure out what still needs to be done (do you have a firm understanding of this @shoyer ?)

[...]

PR #1179, opened 2016-12-22
Switch to shared Lock (SerializableLock if possible) for reading/writing

### LINKED ISSUE #1172: Segmentation fault when writing to netcdf with dask-enabled xarray dataset
I have a 4 GB netcdf file and am running on a machine with 32 GB of memory.  The following works just fine without error on this large memory machine:

    ds = xarray.open_dataset('input.nc')
    ds.to_netcdf('output.nc')

This dask + pynio approach also works correctly:

    ds = xarray.open_dataset('input.nc', chunks={'a': 25, 'b': 25}, engine='pynio')
    ds.to_netcdf('output.nc')

But the following dask + default engine (netcdf4 probably?) approach slowly sucks up all the system memory, writes out a file twice as large as it should be with variable values that are extremely large, and then fails with seg fault, bus error or other low-level system errors we'd rather not be seeing in python!

    ds = xarray.open_dataset('input.nc', chunks={'a': 25, 'b': 25})
    ds.to_netcdf('output.nc')

Adding lock=True to the open_dataset call does not help.

I have two workable solutions to my problem (run without dask because I have a lot of memory available, or use engine='pynio') but this error was hard to track down so thought you would want to know.  Would be glad to hear also if I missed something in the docs and the all-too-common user error is to blame =)

shoyer:
Fixes #1172

The serializable lock will be useful for dask.distributed or multi-processing
(xref #798, #1173, among others).

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -56,7 +56,7 @@ Breaking changes
   By `Guido Imperiale <https://github.com/crusaderky>`_ and
   `Stephan Hoyer <https://github.com/shoyer>`_.
 - Pickling a ``Dataset`` or ``DataArray`` linked to a file on disk no longer
-  caches its values into memory before pickling :issue:`1128`. Instead, pickle
+  caches its values into memory before pickling (:issue:`1128`). Instead, pickle
   stores file paths and restores objects by reopening file references. This
   enables preliminary, experimental use of xarray for opening files with
   `dask.distributed <https://distributed.readthedocs.io>`_.
@@ -205,6 +205,10 @@ Bug fixes
 
 - Fixed a bug with facetgrid (the ``norm`` keyword was ignored, :issue:`1159`).
   By `Fabien Maussion <https://github.com/fmaussion>`_.
+
+- Resolved a concurrency bug that could cause Python to crash when
+  simultaneously reading and writing netCDF4 files with dask (:issue:`1172`).
+  By `Stephan Hoyer <https://github.com/shoyer>`_.
 
 .. _whats-new.0.8.2:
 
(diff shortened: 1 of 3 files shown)


mrocklin:
Is there a clear fail case we can use as a test to demonstrate the value here?

shoyer:
@mrocklin We could update our dask-distributed integration tests to avoid `lock=False`, but that will need to wait until the next dask release for tests to pass on CI.

rabernat:
> Is there a clear fail case we can use as a test to demonstrate the value here?

I have found a fail case related to distributed: attempting to use `to_netcdf()` with a dask.distributed client fails because the `threading.Lock()` can't be serialized. A `SerializableLock` would overcome this problem.

Consider this example:

```python
import dask.array as da
from distributed import Client
import xarray as xr

def create_and_store_dataset():
    shape = (10000, 1000)
    chunks = (1000, 1000)
    data = da.zeros(shape, chunks=chunks)
    ds = xr.DataArray(data).to_dataset()
    ds.to_netcdf('test_dataset.nc')
    print("Success!")

create_and_store_dataset()
client = Client()
create_and_store_dataset()
```

The first call succeeds, while the second fails with `TypeError: can't pickle thread.lock objects`.

When using the distributed client, I can successfully call `.store` on the underlying dask array if I pass `lock=SerializableLock()`.

mrocklin:
Dask 0.13.0 has been released

shoyer:
I will update the xarray/dask-distributed integration and submit this later today. @rabernat it should solve your issues with `to_netcdf`.

shoyer:
Hmm. It looks like we need dask 0.13 in conda to make the distributed integration pass.

mrocklin:
It's up now on conda-forge if you're interested in switching over.

shoyer:
In it goes. We're using `conda-forge` for all Travis-CI builds now.

PR #1180, opened 2016-12-23
Actually copy dask arrays in Variable.copy()

### LINKED ISSUE #1166: Variable.copy() should copy dask.arrays
After https://github.com/dask/dask/pull/1840, dask arrays will no longer be immutable.

shoyer:
Fixes #1166

xref https://github.com/dask/dask/pull/1840

I don't think there's a good way to test this prior to the release of the next version of dask.

For what it's worth, this will make the minimum version of dask 0.9.0, unless we want to add a hack to handle a non-existent `copy()` method.

### DIFF
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -6,7 +6,7 @@ Installation
 Required dependencies
 ---------------------
 
-- Python 2.7, 3.3, 3.4 or 3.5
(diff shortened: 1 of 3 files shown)


jhamman:
LGTM.

PR #1181, opened 2016-12-23
Add mention of in breaking change about default fill value

laliberte:
Quick PR to attribute credit to Frederic Laliberte for the default fill value in the ``.to_netcdf()`` method.


shoyer:
Thanks!

Issue #1140, opened 2016-11-28
sel with method 'nearest' fails with AssertionError


j08lue:
The following fails

```python
def test_sel_nearest():
    """Test `sel` with method='nearest'"""
    # create test data
    times = pd.date_range('2000-01-01', periods=4)
    foo = xr.DataArray([1,2,3,4], coords=dict(time=times), dims='time')
    
    # works
    expected = foo.sel(time='2000-01-01')
    
    # fails
    assert foo.sel(time='2000-01-01', method='nearest') == expected
```

with an `AssertionError` in [`xarray/core/variable.py`](https://github.com/pydata/xarray/blob/v0.8.2/xarray/core/variable.py#L379):

```
C:\Users\uuu\AppData\Local\Continuum\Miniconda2\lib\site-packages\xarray\core\dataarray.pyc in sel(self, method, tolerance, **indexers)
    623             self, indexers, method=method, tolerance=tolerance
    624         )
--> 625         return self.isel(**pos_indexers)._replace_indexes(new_indexes)
    626 
    627     def isel_points(self, dim='points', **indexers):

C:\Users\uuu\AppData\Local\Continuum\Miniconda2\lib\site-packages\xarray\core\dataarray.pyc in isel(self, **indexers)
    608         DataArray.sel
    609         """
--> 610         ds = self._to_temp_dataset().isel(**indexers)
    611         return self._from_temp_dataset(ds)
    612 

C:\Users\uuu\AppData\Local\Continuum\Miniconda2\lib\site-packages\xarray\core\dataset.pyc in isel(self, **indexers)
    910         for name, var in iteritems(self._variables):
    911             var_indexers = dict((k, v) for k, v in indexers if k in var.dims)
--> 912             variables[name] = var.isel(**var_indexers)
    913         return self._replace_vars_and_dims(variables)
    914 

C:\Users\uuu\AppData\Local\Continuum\Miniconda2\lib\site-packages\xarray\core\variable.pyc in isel(self, **indexers)
    539             if dim in indexers:
    540                 key[i] = indexers[dim]
--> 541         return self[tuple(key)]
    542 
    543     def _shift_one_dim(self, dim, count):

C:\Users\uuu\AppData\Local\Continuum\Miniconda2\lib\site-packages\xarray\core\variable.pyc in __getitem__(self, key)
    377         # orthogonal indexing should ensure the dimensionality is consistent
    378         if hasattr(values, 'ndim'):
[...]

PR #1184, opened 2016-12-25
Add test for issue 1140

j08lue:
#1140


shoyer:
Hmm... looks like things are passing on Appveyor. Might be worth trying merging master (after #1185) if we want to be completely sure.

shoyer:
Fixed by #1339

PR #1187, opened 2016-12-28
add quantile method to DataArray

### LINKED ISSUE #303: Add a percentile/quantile method
This could use numpy's `np.percentile` and `np.nanpercentile`, but would require a slightly different function signature than the other aggregation functions because it creates a new dimension. Pandas calls the same method `quantile`.

### LINKED ISSUE #561: calculate percentiles
How would I calculate percentiles over a certain dimension?

```
ds.groupby('time.month').apply(np.percentile, q=5)
```

Does not work.

jhamman:
This PR adds the `quantile` method to the DataArray.  There may be a way to better fit this in with `.apply` or `reduce` it wasn't immediately clear to me how to do that. It uses `np.nanpercentile` under the hood so this shouldn't be expected to work well with dask. The main advantage to this method, over using `apply(np.nanpercentile)` is the handling of the `quantile` coordinate.

example usage:

```Python
In [12]: x = np.random.random(size=(2, 3, 800))

In [13]: da = xr.DataArray(x, dims=('x', 'y', 'time'))

In [14]: da.quantile([1, 5, 10, 25, 50, 75, 90, 95, 99], dim='time', interpolation='lower')
Out[14]: 
<xarray.DataArray (quantile: 9, x: 2, y: 3)>
array([[[ 0.00835474,  0.01126747,  0.00778847],
        [ 0.00803924,  0.00919259,  0.01150164]],

       [[ 0.04902814,  0.05346976,  0.04493341],
        [ 0.04236611,  0.05273082,  0.05858802]],

       [[ 0.09370776,  0.09976448,  0.09256707],
        [ 0.08943787,  0.09331907,  0.08832309]],

       [[ 0.25416402,  0.22577298,  0.24407393],
        [ 0.25386087,  0.23052584,  0.24621966]],

       [[ 0.534169  ,  0.46017551,  0.49817391],
        [ 0.50968059,  0.49427688,  0.51573855]],

       [[ 0.76633412,  0.73405412,  0.77210754],
        [ 0.76759837,  0.74243892,  0.76703357]],

       [[ 0.90832116,  0.89495854,  0.91818434],
        [ 0.91492771,  0.88063773,  0.91416636]],

       [[ 0.95260527,  0.95132871,  0.95979701],
        [ 0.95988286,  0.93137055,  0.95941658]],

       [[ 0.98597133,  0.98883232,  0.99013424],
        [ 0.98951238,  0.97550784,  0.99224201]]])
Coordinates:
  * quantile  (quantile) float64 1.0 5.0 10.0 25.0 50.0 75.0 90.0 95.0 99.0
  o x         (x) -
  o y         (y) -
```

closes #303 
fixes #561


xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+        Returns the qth quantiles(s) of the array elements.
+
+        Parameters
+        ----------
+        q : float in range of [0,100] (or sequence of floats)
+            Quantile to compute, which must be between 0 and 100
+            inclusive.
+        dim : str or sequence of str, optional
+            Dimension(s) over which to apply quantile.
+        axis : int or sequence of int, optional
```
shoyer: I would not bother with the `axis` argument. It makes the logic below quite convoluted, and in my experience with xarray it's rarely useful.

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
 
         return type(self)(new_data, new_coords, new_dims)
 
+    def quantile(self, q, dim=None, axis=None, interpolation='linear'):
+        """
```
shoyer: If possible, start with a one line description (without the extra line break)

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+
+        Compute the qth quantile of the data along the specified axis,
+        while ignoring nan values.
+        Returns the qth quantiles(s) of the array elements.
+
+        Parameters
+        ----------
+        q : float in range of [0,100] (or sequence of floats)
```
shoyer: Pandas uses quantiles between 0 and 1. Since we're using the name `quantile` (rather than `percentile`), I would stick with that convention (just divide by 100 to use `nanpercentile` under the hood).

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+        -------
+        quantiles : DataArray
+            If `q` is a single quantile and `axis=None`, then the result
+            is a scalar. If multiple percentiles are given, first axis of
+            the result corresponds to the quantile and a quantile dimension
+            is added to the return array. The other axes are the axes that
+            remain after the reduction of the array. If the input
+            contains integers or floats smaller than ``float64``, the output
+            data-type is ``float64``. Otherwise, the output data-type is the
+            same as that of the input.
```
shoyer: When is "same as the input" not float64? (I guess this is an issue with numpy docstring, but I would probably just leave out the extra detail here rather than copying it)

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+        # only add the quantile dimension if q is array like
+        if not isscalar:
+            new_dims = ['quantile'] + new_dims
+
+        ps = np.nanpercentile(self.data, q, axis=axis,
+                              interpolation=interpolation)
+
+        # Construct the return DataArray
+        ps = DataArray(ps, dims=new_dims, name=self.name)
+        if not isscalar:
```
shoyer: Maybe add `quantile` as a coordinate even if it's a scalar?

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+                    new_dims.remove(self.dims[i])
+            else:
+                new_dims.remove(self.dims[axis])
+        else:
+            new_dims = []
+        # only add the quantile dimension if q is array like
+        if not isscalar:
+            new_dims = ['quantile'] + new_dims
+
+        ps = np.nanpercentile(self.data, q, axis=axis,
```
shoyer: Let's make sure this fails loudly on dask arrays rather than implicitly converting to numpy.

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
 
         return type(self)(new_data, new_coords, new_dims)
 
+    def quantile(self, q, dim=None, axis=None, interpolation='linear'):
```
shoyer: It would be nice to also have this method for `Dataset` (or maybe even for `Variable`, too). Take a look at `Dataset.reduce` for an example.

To avoid duplicate implementation, many `DataArray` methods have the core of their logic implemented for Dataset and then use `_to_temp_dataset()` and `_to_from_dataset()` to convert back and forth.

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+        """
+
+        if dim is not None and axis is not None:
+            raise ValueError("cannot supply both 'axis' and 'dim' arguments")
+
+        isscalar = np.isscalar(q)
+        if isscalar:
+            q = float(q)
+        else:
+            q = np.asarray(q, dtype=np.float64)
```
shoyer: It might be slightly cleaner to unilaterally coerce `q` to a float64 array, and then use `q.ndim != 0` to check for nonscalar q.

xarray/core/dataarray.py
```diff
@@ -1736,5 +1736,93 @@ def dot(self, other):
...
+        if not isscalar:
+            new_dims = ['quantile'] + new_dims
+
+        ps = np.nanpercentile(self.data, q, axis=axis,
+                              interpolation=interpolation)
+
+        # Construct the return DataArray
+        ps = DataArray(ps, dims=new_dims, name=self.name)
+        if not isscalar:
+            ps['quantile'] = DataArray(q, dims=('quantile', ),
```
shoyer: Maybe slightly cleaner: `ps.coords['quantile'] = ('quantile', q)` or `ps.coords['quantile'] = Variable('quantile', q)`. There's no need to construct a full DataArray here and the `name='quantile' bit in particular is redundant with the name on the left-hand-side of the assignment operation.

xarray/test/test_dataarray.py
```diff
@@ -1328,6 +1328,22 @@ def test_reduce(self):
         expected = DataArray(5, {'c': -999})
         self.assertDataArrayIdentical(expected, actual)
 
+    def test_quantile(self):
+        for method in ['linear', 'lower', 'higher', 'nearest', 'midpoint']:
```
shoyer: No need to test all these methods here, since we have no logic specific to them. It's enough to verify (once) that the `method` argument gets passed on to NumPy.

xarray/core/dataarray.py
shoyer: Probably the most consistent with the current approach is to implement the guts (calling `np.nanpercentile`) on `Variable`, and then make a wrapper for handling coordinates on `Dataset` (and call that from the `DataArray` method).

jhamman:
@shoyer - I moved the majority of the logic over to `Variable` and wrote wrapper methods for the `Dataset` and `DataArray` objects. I am getting an error on the `py27-min` build on travis that may or may not be an since fixed bug in numpy. Unpinning the version numbers in the build fixes the problem.  Compare the builds for ddd5211 and aa81a17. I'm assuming we want to keep those versions pinned as they were so thoughts on a possible fix would be great...

xarray/core/dataset.py
```diff
@@ -2538,6 +2537,57 @@ def roll(self, **shifts):
...
+        See Also
+        --------
+        np.nanpercentile, pd.Series.quantile, xr.DataArray.quantile
+        """
+
+        q = np.asarray(q, dtype=np.float64)
+
+        variables = OrderedDict()
+        for name, var in iteritems(self.variables):
+            variables[name] = var.quantile(q, dim=dim,
```
shoyer: I think we need a condition like the one in `Dataset.reduce` to skip variables that don't use `dim`.

xarray/core/variable.py
```diff
@@ -1029,6 +1029,74 @@ def no_conflicts(self, other):
...
+
+        See Also
+        --------
+        np.nanpercentile, pd.Series.quantile,
+        xr.Dataset.quantile, xr.DataArray.quantile
+        """
+
+        if isinstance(self.data, dask_array_type):
+            TypeError("quantile does not work for arrays stored as dask "
+                      "arrays. Load the data via .load() prior to calling "
```
shoyer: `.load()` should be `.compute() or .load()`

xarray/core/variable.py
```diff
@@ -1029,6 +1029,74 @@ def no_conflicts(self, other):
...
+
+        q = np.asarray(q, dtype=np.float64)
+
+        new_dims = list(self.dims)
+        if dim is not None:
+            if utils.is_scalar(dim):
+                axis = self.get_axis_num(dim)
+                new_dims.remove(dim)
+            else:
+                axis = [self.get_axis_num(d) for d in dim]
```
shoyer: minor note: `get_axis_num` actually works on either individual or lists of dimensions

doc/whats-new.rst
```diff
@@ -165,6 +165,9 @@ Enhancements
   and attributes. The method prints to a buffer (e.g. ``stdout``) with output
   similar to what the command line utility ``ncdump -h`` produces (:issue:`1150`).
   By `Joe Hamman <https://github.com/jhamman>`_.
+- New :py:meth:`~DataArray.quantile` method to calculate quantiles from
+  DataArray objects (:issue:`xxxx`).
```
shoyer: update issue number to this PR

ci/requirements-py27-min.yml
```diff
@@ -2,8 +2,8 @@ name: test_env
 dependencies:
   - python=2.7
   - pytest
-  - numpy==1.9.3
```
shoyer: This does look a possible NumPy bug. Instead of updating the required version of numpy for everything, I would just skip it explicitly for the failing test.

.travis.yml
```diff
@@ -61,9 +61,9 @@ install:
   - source activate test_env
   # scipy should not have been installed, but it's included in older versions of
   # the conda pandas package
-  - if [[ "$CONDA_ENV" == "py27-min" ]]; then
```
shoyer: better not to remove this unless necessary

xarray/core/dataset.py
shoyer: Also, coordinate variable should be skipped. See `Dataset.reduce` for the right logic.

jhamman:
@shoyer - all comments addressed and tests are passing.

xarray/core/dataset.py
```diff
@@ -2547,6 +2546,87 @@ def roll(self, **shifts):
 
         return self._replace_vars_and_dims(variables)
 
+    def quantile(self, q, dim=None, numeric_only=False, keep_attrs=False,
+                 interpolation='linear'):
```
shoyer: `numeric_only` and `keep_attrs` are missing from the docstring

doc/api.rst
```diff
@@ -270,6 +270,7 @@ Computation
    DataArray.get_axis_num
    DataArray.diff
    DataArray.dot
+   DataArray.quantile
```
shoyer: add this for `Dataset`, too

jhamman: done.

PR #1190, opened 2016-12-29
only use bottleneck version 1.0 and later

### LINKED ISSUE #808: bottleneck version
Not sure if this is possible to specify in the `setup.py` as bottleneck is not required. However, if bottleneck is used it should probably be in version 1.0. For version 0.8.0 I get the following error:

`>>> move_mean() got an unexpected keyword argument 'min_count'`

Feel free to close.

jhamman:
closes  #808

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -6,6 +6,7 @@ import contextlib
 import inspect
 import operator
 import warnings
+from distutils.version import StrictVersion
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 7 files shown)


xarray/core/ops.py
```diff
@@ -17,6 +18,10 @@
 
 try:
     import bottleneck as bn
+    if StrictVersion(bn.__version__) < StrictVersion('1.0'):
+        warnings.warn('xarray requires bottleneck version of 1.0 or greater.'
```
shoyer: Users find it really annoying to see messages just when importing a package, when they may not even be using the related functionality. Let's only issue the warning when doing rolling aggregations.

jhamman: fair enough. The logic gets spread out a bit more but we now only raise the warning when `rolling.__init__()` is called.

PR #1191, opened 2016-12-29
Integer levels and vmin/vmax

fmaussion:
Follow-up to https://github.com/pydata/xarray/pull/1171#issuecomment-269556898

From the new docstring:

```
    levels : int or list-like object, optional
        Split the colormap (cmap) into discrete color intervals. If an integer
        is provided, "nice" levels are chosen based on the data range: this can
        imply that the final number of levels is not exactly the expected one.
        Setting ``vmin`` and/or ``vmax`` with ``levels=N`` is equivalent to
        setting ``levels=np.linspace(vmin, vmax, N)``.
```

The logic overhead is quite simple, which is an argument in favor of this simple solution. It is consistent with mpl, as long as neither vmin or vmax are set.

It _might_ change the outcome of some existing plots, though. Should I mention this in the "Breaking changes" section too?

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -70,6 +70,10 @@ Breaking changes
 - Coordinates used to index a dimension are now loaded eagerly into
   :py:class:`pandas.Index` objects, instead of loading the values lazily.
   By `Guido Imperiale <https://github.com/crusaderky>`_.
+- Automatic levels for 2d plots are now guaranteed to land on ``vmin`` and
+  ``vmax`` when these kwargs are explicitly provided (:issue:`1191`). The
+  automated level selection logic also slightly changed.
+  By `Fabien Maussion <https://github.com/fmaussion>`_.
 - xarray no longer supports python 3.3 or versions of dask prior to v0.9.0.
 
 Deprecations
@@ -225,10 +229,6 @@ Bug fixes
 
 - Fix to make ``.copy()`` actually copy dask arrays, which will be relevant for
   future releases of dask in which dask arrays will be mutable (:issue:`1180`).
-
-- Automatic levels for 2d plots are now guaranteed to land on ``vmin`` and
-  ``vmax`` when these kwargs are explicitly provided (:issue:`1191`).
-  By `Fabien Maussion <https://github.com/fmaussion>`_.
 
 .. _whats-new.0.8.2:
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -226,6 +226,10 @@ Bug fixes
 - Fix to make ``.copy()`` actually copy dask arrays, which will be relevant for
   future releases of dask in which dask arrays will be mutable (:issue:`1180`).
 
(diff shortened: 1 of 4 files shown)


xarray/plot/utils.py
```diff
@@ -187,9 +191,13 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
 
     # Handle discrete levels
     if levels is not None:
-        if isinstance(levels, int):
-            ticker = mpl.ticker.MaxNLocator(levels)
-            levels = ticker.tick_values(vmin, vmax)
+        if is_scalar(levels):
+            if user_minmax or levels==1:
```
jhamman: nit/pep: spaces around ` == `

doc/whats-new.rst
```diff
@@ -226,6 +226,10 @@ Bug fixes
 - Fix to make ``.copy()`` actually copy dask arrays, which will be relevant for
   future releases of dask in which dask arrays will be mutable (:issue:`1180`).
 
+- Automatic levels for 2d plots are now guaranteed to land on ``vmin`` and
+  ``vmax`` when these kwargs are explicitly provided (:issue:`1191`).
+  By `Fabien Maussion <https://github.com/fmaussion>`_.
+
```
jhamman: I think this belongs in the "breaking changes" category.

xarray/test/test_plot.py
```diff
@@ -2,6 +2,16 @@
...
 
+try:
+    import matplotlib as mpl
+    # Using a different backend makes Travis CI work
+    # This has to happen before other mpl imports
+    mpl.use('Agg')
+    # Order of imports is important here.
+    import matplotlib.pyplot as plt
+except ImportError:
+    pass
```
clarkfitzg: If we don't have matplotlib then can we just skip all the plotting tests?

clarkfitzg:
> It is consistent with mpl, as long as neither vmin or vmax are set.

The reason for having consistency with MPL is that users can take essentially the same code and produce the same plots with or without xarray. And it's less work on this end. We're departing from that here to get something that perhaps is more natural for the user. But I think it's ok since the plots generally won't look all that different. If users want more control they can just set all the `levels` explicitly.

xarray/test/test_plot.py
fmaussion: Yes, all tests have an ``@requires_matplotlib`` decorator

fmaussion:
I've updated the PR. Let me know if there's something else we should add here

xarray/test/test_plot.py
fmaussion: Would you like me to add something here? (note that I just changed this blocks location further up in the module)

clarkfitzg: @fmaussion Maybe just update the comments to clarify what the order should be. For example:

```
# Import matplotlib before xarray plotting
try:
     import matplotlib as mpl
     # Using a different backend makes Travis CI work
     mpl.use('Agg')
 except ImportError:
     pass
```

Issue #463, opened 2015-07-10
open_mfdataset too many files


rabernat:
I am very excited to try xray.

On my first attempt, I tried to use open_mfdataset on a set of ~8000 netcdf files. I hit a "RuntimeError: Too many open files". The ulimit on my system is 1024, so clearly that is the source of the error.

I am curious whether this is the desired behavior for open_mfdataset. Does xray have to keep all the files open? If so, I will work with my sysadmin to increase the ulimit.

It seems like the whole point of this function is to work with large collections of files, so this could be a significant limitation.

rabernat:
Just a little follow up...I tried to work around the file limit by serializing the processing of the files and creating xray datasets with with fewer files in them. However, I still eventually hit this error, suggesting that the files are never being closed. For example

I would like to do
```python
ds = xray.open_mfdataset(ddir + '*.nc' % yr, engine='scipy')
EKE = (ds.variables['u']**2 + ds.variables['v']**2).mean(dim='time').load()
```
This tries to open 8031 files and produces the `error: [Errno 24] Too many open files`

So then I try to create a new dataset for each year

 ```python
EKE = []
for yr in xrange(1993,2015):
    print yr
    # this opens about 365 files
    ds = xray.open_mfdataset(ddir + '/dt_global_allsat_msla_uv_%04d*.nc' % yr, engine='scipy')
    EKE.append((ds.variables['u']**2 + ds.variables['v']**2).mean(dim='time').load())
```

This works okay for the first two years. However, by the third year, I _still_ get the `error: [Errno 24] Too many open files`. This is when the ulimit of 1024 files is exceeded.

Using xray version 0.5.1 via conda module.

**Comment by shoyer** [comment]
Yes, this is a known issue, and I agree that it is annoying. We could work around this by opening up (and closing) netCDF files inside the `__getitem__` call. If you're interested in possibly working on this, take a look at the netCDF4 backend for xray: https://github.com/xray/xray/blob/master/xray/backends/netCDF4_.py

**Comment by rabernat** [comment]
I am using the scipy backend because the netcdf4 backend doesn't work for me at all. It core dumps with the error
```
python: posixio.c:366: px_rel: Assertion `pxp->bf_offset <= offset && offset < pxp->bf_offset + (off_t) pxp->bf_extent' failed.
Aborted (core dumped)
```

Are you suggesting I work on the scipy backend?

shoyer:
[...]

PR #1198, opened 2017-01-10
Fixes OS error arising from too many files open

pwolfram:
Previously, DataStore  did not judiciously close files, resulting in opening a large number of files that
could result in an OSError related to too many files being open.  This merge provides a solution for the netCDF, scipy, and h5netcdf backends.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-import gzip
 import os.path
 from distutils.version import LooseVersion
 from glob import glob
@@ -133,7 +132,7 @@ def _protect_dataset_variables_inplace(dataset, cache):
 
 
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
-                 mask_and_scale=True, decode_times=True,
+                 mask_and_scale=True, decode_times=True, autoclose=False,
                  concat_characters=True, decode_coords=True, engine=None,
                  chunks=None, lock=None, cache=None, drop_variables=None):
     """Load and decode a dataset from a file or file-like object.
@@ -163,6 +162,10 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     decode_times : bool, optional
         If True, decode times encoded in the standard NetCDF datetime format
         into datetime objects. Otherwise, leave them encoded as numbers.
+    autoclose : bool, optional
+        If True, automatically close files to avoid OS Error of too many files
+        being open.  However, this option doesn't work with streams, e.g.,
+        BytesIO.
     concat_characters : bool, optional
         If True, concatenate along the last dimension of character arrays to
         form string arrays. Dimensions will only be concatenated over (and
@@ -251,6 +254,12 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         else:
             ds2 = ds
 
+        # protect so that dataset store isn't necessarily closed, e.g.,
+        # streams like BytesIO  can't be reopened
+        # datastore backend is responsible for determining this capability
+        if store._autoclose:
+            store.close()
+
         return ds2
 
     if isinstance(filename_or_obj, backends.AbstractDataStore):
@@ -271,33 +280,30 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
             if engine is not None and engine != 'scipy':
                 raise ValueError('can only read gzipped netCDF files with '
                                  "default engine or engine='scipy'")
-            # if the string ends with .gz, then gunzip and open as netcdf file
-            try:
-                store = backends.ScipyDataStore(gzip.open(filename_or_obj))
-            except TypeError as e:
-                # TODO: gzipped loading only works with NetCDF3 files.
-                if 'is not a valid NetCDF 3 file' in e.message:
-                    raise ValueError('gzipped file loading only supports '
-                                     'NetCDF 3 files.')
-                else:
-                    raise
-        else:
-            if engine is None:
-                engine = _get_default_engine(filename_or_obj,
-                                             allow_remote=True)
-            if engine == 'netcdf4':
-                store = backends.NetCDF4DataStore(filename_or_obj, group=group)
-            elif engine == 'scipy':
-                store = backends.ScipyDataStore(filename_or_obj)
-            elif engine == 'pydap':
-                store = backends.PydapDataStore(filename_or_obj)
-            elif engine == 'h5netcdf':
-                store = backends.H5NetCDFStore(filename_or_obj, group=group)
-            elif engine == 'pynio':
-                store = backends.NioDataStore(filename_or_obj)
             else:
-                raise ValueError('unrecognized engine for open_dataset: %r'
-                                 % engine)
+                engine = 'scipy'
+
+        if engine is None:
+            engine = _get_default_engine(filename_or_obj,
+                                         allow_remote=True)
+        if engine == 'netcdf4':
+            store = backends.NetCDF4DataStore(filename_or_obj, group=group,
+                                              autoclose=autoclose)
+        elif engine == 'scipy':
+            store = backends.ScipyDataStore(filename_or_obj,
+                                            autoclose=autoclose)
+        elif engine == 'pydap':
+            store = backends.PydapDataStore(filename_or_obj)
+        elif engine == 'h5netcdf':
+            store = backends.H5NetCDFStore(filename_or_obj, group=group,
+                                           autoclose=autoclose)
+        elif engine == 'pynio':
+            store = backends.NioDataStore(filename_or_obj,
+                                          autoclose=autoclose)
+        else:
+            raise ValueError('unrecognized engine for open_dataset: %r'
+                             % engine)
+
         if lock is None:
             lock = _default_lock(filename_or_obj, engine)
         with close_on_error(store):
@@ -479,6 +485,10 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
         Engine to use when reading files. If not provided, the default engine
         is chosen based on available dependencies, with a preference for
         'netcdf4'.
+    autoclose : bool, optional
+        If True, automatically close files to avoid OS Error of too many files
+        being open.  However, this option doesn't work with streams, e.g.,
+        BytesIO.
     lock : False, True or threading.Lock, optional
         This argument is passed on to :py:func:`dask.array.from_array`. By
         default, a per-variable lock is used when reading data from netCDF
--- a/xarray/backends/h5netcdf_.py
+++ b/xarray/backends/h5netcdf_.py
@@ -6,12 +6,18 @@ import warnings
 
 from .. import Variable
 from ..core import indexing
-from ..core.utils import FrozenOrderedDict, close_on_error, Frozen
+from ..core.utils import FrozenOrderedDict, close_on_error
 from ..core.pycompat import iteritems, bytes_type, unicode_type, OrderedDict
 
-from .common import WritableCFDataStore, DataStorePickleMixin
+from .common import WritableCFDataStore, DataStorePickleMixin, find_root
 from .netCDF4_ import (_nc4_group, _nc4_values_and_dtype,
                        _extract_nc4_variable_encoding, BaseNetCDF4Array)
+
+
+class H5NetCDFArrayWrapper(BaseNetCDF4Array):
+    def __getitem__(self, key):
+        with self.datastore.ensure_open(autoclose=True):
+            return self.get_array()[key]
 
 
 def maybe_decode_bytes(txt):
@@ -49,12 +55,18 @@ class H5NetCDFStore(WritableCFDataStore, DataStorePickleMixin):
     """Store for reading and writing data via h5netcdf
     """
     def __init__(self, filename, mode='r', format=None, group=None,
-                 writer=None):
+                 writer=None, autoclose=False):
         if format not in [None, 'NETCDF4']:
             raise ValueError('invalid format for h5netcdf backend')
         opener = functools.partial(_open_h5netcdf_group, filename, mode=mode,
                                    group=group)
         self.ds = opener()
+        if autoclose:
+            raise NotImplemented('autoclose=True is not implemented '
+                                 'for the h5netcdf backend pending further '
+                                 'exploration, e.g., bug fixes (in h5netcdf?)')
+        self._autoclose = False
+        self._isopen = True
         self.format = format
         self._opener = opener
         self._filename = filename
@@ -62,36 +74,44 @@ class H5NetCDFStore(WritableCFDataStore, DataStorePickleMixin):
         super(H5NetCDFStore, self).__init__(writer)
 
     def open_store_variable(self, name, var):
-        dimensions = var.dimensions
-        data = indexing.LazilyIndexedArray(BaseNetCDF4Array(name, self))
-        attrs = _read_attributes(var)
+        with self.ensure_open(autoclose=False):
+            dimensions = var.dimensions
+            data = indexing.LazilyIndexedArray(
+                H5NetCDFArrayWrapper(name, self))
+            attrs = _read_attributes(var)
 
-        # netCDF4 specific encoding
-        encoding = dict(var.filters())
-        chunking = var.chunking()
-        encoding['chunksizes'] = chunking if chunking != 'contiguous' else None
+            # netCDF4 specific encoding
+            encoding = dict(var.filters())
+            chunking = var.chunking()
+            encoding['chunksizes'] = chunking \
+                if chunking != 'contiguous' else None
 
-        # save source so __repr__ can detect if it's local or not
-        encoding['source'] = self._filename
-        encoding['original_shape'] = var.shape
+            # save source so __repr__ can detect if it's local or not
+            encoding['source'] = self._filename
+            encoding['original_shape'] = var.shape
 
         return Variable(dimensions, data, attrs, encoding)
 
     def get_variables(self):
-        return FrozenOrderedDict((k, self.open_store_variable(k, v))
-                                 for k, v in iteritems(self.ds.variables))
+        with self.ensure_open(autoclose=False):
+            return FrozenOrderedDict((k, self.open_store_variable(k, v))
+                                     for k, v in iteritems(self.ds.variables))
 
     def get_attrs(self):
-        return Frozen(_read_attributes(self.ds))
+        with self.ensure_open(autoclose=True):
+            return FrozenOrderedDict(_read_attributes(self.ds))
 
     def get_dimensions(self):
-        return self.ds.dimensions
+        with self.ensure_open(autoclose=True):
+            return self.ds.dimensions
 
     def set_dimension(self, name, length):
-        self.ds.createDimension(name, size=length)
+        with self.ensure_open(autoclose=False):
+            self.ds.createDimension(name, size=length)
 
     def set_attribute(self, key, value):
-        self.ds.setncattr(key, value)
+        with self.ensure_open(autoclose=False):
+            self.ds.setncattr(key, value)
 
     def prepare_variable(self, name, variable, check_encoding=False,
                          unlimited_dims=None):
@@ -129,12 +149,14 @@ class H5NetCDFStore(WritableCFDataStore, DataStorePickleMixin):
         return nc4_var, variable.data
 
     def sync(self):
-        super(H5NetCDFStore, self).sync()
-        self.ds.sync()
+        with self.ensure_open(autoclose=True):
+            super(H5NetCDFStore, self).sync()
+            self.ds.sync()
 
     def close(self):
-        ds = self.ds
-        # netCDF4 only allows closing the root group
-        while ds.parent is not None:
-            ds = ds.parent
-        ds.close()
+        if self._isopen:
+            # netCDF4 only allows closing the root group
+            ds = find_root(self.ds)
+            if not ds._closed:
+                ds.close()
+            self._isopen = False
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -9,11 +9,13 @@ import numpy as np
 from .. import Variable
 from ..conventions import pop_to
 from ..core import indexing
-from ..core.utils import (FrozenOrderedDict, NDArrayMixin,
-                          close_on_error, is_remote_uri)
+from ..core.utils import (FrozenOrderedDict, NdimSizeLenMixin,
+                          DunderArrayMixin, close_on_error,
+                          is_remote_uri)
 from ..core.pycompat import iteritems, basestring, OrderedDict, PY3
 
-from .common import WritableCFDataStore, robust_getitem, DataStorePickleMixin
+from .common import (WritableCFDataStore, robust_getitem,
+                     DataStorePickleMixin, find_root)
 from .netcdf3 import (encode_nc3_attr_value, encode_nc3_variable,
                       maybe_convert_to_char_array)
 
@@ -25,24 +27,25 @@ _endian_lookup = {'=': 'native',
                   '|': 'native'}
 
 
-class BaseNetCDF4Array(NDArrayMixin):
+class BaseNetCDF4Array(NdimSizeLenMixin, DunderArrayMixin):
     def __init__(self, variable_name, datastore):
         self.datastore = datastore
         self.variable_name = variable_name
 
-    @property
-    def array(self):
-        return self.datastore.ds.variables[self.variable_name]
+        array = self.get_array()
+        self.shape = array.shape
 
-    @property
-    def dtype(self):
-        dtype = self.array.dtype
+        dtype = array.dtype
         if dtype is str:
-            # return object dtype because that's the only way in numpy to
+            # use object dtype because that's the only way in numpy to
             # represent variable length strings; it also prevents automatic
             # string concatenation via conventions.decode_cf_variable
             dtype = np.dtype('O')
-        return dtype
+        self.dtype = dtype
+
+    def get_array(self):
+        self.datastore.assert_open()
+        return self.datastore.ds.variables[self.variable_name]
 
 
 class NetCDF4ArrayWrapper(BaseNetCDF4Array):
@@ -52,19 +55,20 @@ class NetCDF4ArrayWrapper(BaseNetCDF4Array):
         else:
             getitem = operator.getitem
 
-        try:
-            data = getitem(self.array, key)
-        except IndexError:
-            # Catch IndexError in netCDF4 and return a more informative error
-            # message.  This is most often called when an unsorted indexer is
-            # used before the data is loaded from disk.
-            msg = ('The indexing operation you are attempting to perform is '
-                   'not valid on netCDF4.Variable object. Try loading your '
-                   'data into memory first by calling .load().')
-            if not PY3:
-                import traceback
-                msg += '\n\nOriginal traceback:\n' + traceback.format_exc()
-            raise IndexError(msg)
+        with self.datastore.ensure_open(autoclose=True):
+            try:
+                data = getitem(self.get_array(), key)
+            except IndexError:
+                # Catch IndexError in netCDF4 and return a more informative
+                # error message.  This is most often called when an unsorted
+                # indexer is used before the data is loaded from disk.
+                msg = ('The indexing operation you are attempting to perform '
+                       'is not valid on netCDF4.Variable object. Try loading '
+                       'your data into memory first by calling .load().')
+                if not PY3:
+                    import traceback
+                    msg += '\n\nOriginal traceback:\n' + traceback.format_exc()
+                raise IndexError(msg)
 
         if self.ndim == 0:
             # work around for netCDF4-python's broken handling of 0-d
@@ -195,7 +199,8 @@ class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
     This store supports NetCDF3, NetCDF4 and OpenDAP datasets.
     """
     def __init__(self, filename, mode='r', format='NETCDF4', group=None,
-                 writer=None, clobber=True, diskless=False, persist=False):
+                 writer=None, clobber=True, diskless=False, persist=False,
+                 autoclose=False):
         if format is None:
             format = 'NETCDF4'
         opener = functools.partial(_open_netcdf4_group, filename, mode=mode,
@@ -203,66 +208,84 @@ class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
                                    diskless=diskless, persist=persist,
                                    format=format)
         self.ds = opener()
+        self._autoclose = autoclose
+        self._isopen = True
         self.format = format
         self.is_remote = is_remote_uri(filename)
-        self._opener = opener
         self._filename = filename
         self._mode = 'a' if mode == 'w' else mode
+        self._opener = functools.partial(opener, mode=self._mode)
         super(NetCDF4DataStore, self).__init__(writer)
 
     def open_store_variable(self, name, var):
-        dimensions = var.dimensions
-        data = indexing.LazilyIndexedArray(NetCDF4ArrayWrapper(name, self))
-        attributes = OrderedDict((k, var.getncattr(k))
-                                 for k in var.ncattrs())
-        _ensure_fill_value_valid(data, attributes)
-        # netCDF4 specific encoding; save _FillValue for later
-        encoding = {}
-        filters = var.filters()
-        if filters is not None:
-            encoding.update(filters)
-        chunking = var.chunking()
-        if chunking is not None:
-            if chunking == 'contiguous':
-                encoding['contiguous'] = True
-                encoding['chunksizes'] = None
-            else:
-                encoding['contiguous'] = False
-                encoding['chunksizes'] = tuple(chunking)
-        # TODO: figure out how to round-trip "endian-ness" without raising
-        # warnings from netCDF4
-        # encoding['endian'] = var.endian()
-        pop_to(attributes, encoding, 'least_significant_digit')
-        # save source so __repr__ can detect if it's local or not
-        encoding['source'] = self._filename
-        encoding['original_shape'] = var.shape
+        with self.ensure_open(autoclose=False):
+            dimensions = var.dimensions
+            data = indexing.LazilyIndexedArray(NetCDF4ArrayWrapper(name, self))
+            attributes = OrderedDict((k, var.getncattr(k))
+                                     for k in var.ncattrs())
+            _ensure_fill_value_valid(data, attributes)
+            # netCDF4 specific encoding; save _FillValue for later
+            encoding = {}
+            filters = var.filters()
+            if filters is not None:
+                encoding.update(filters)
+            chunking = var.chunking()
+            if chunking is not None:
+                if chunking == 'contiguous':
+                    encoding['contiguous'] = True
+                    encoding['chunksizes'] = None
+                else:
+                    encoding['contiguous'] = False
+                    encoding['chunksizes'] = tuple(chunking)
+            # TODO: figure out how to round-trip "endian-ness" without raising
+            # warnings from netCDF4
+            # encoding['endian'] = var.endian()
+            pop_to(attributes, encoding, 'least_significant_digit')
+            # save source so __repr__ can detect if it's local or not
+            encoding['source'] = self._filename
+            encoding['original_shape'] = var.shape
+
         return Variable(dimensions, data, attributes, encoding)
 
     def get_variables(self):
-        return FrozenOrderedDict((k, self.open_store_variable(k, v))
-                                 for k, v in iteritems(self.ds.variables))
+        with self.ensure_open(autoclose=False):
+            dsvars = FrozenOrderedDict((k, self.open_store_variable(k, v))
+                                       for k, v in
+                                       iteritems(self.ds.variables))
+        return dsvars
 
     def get_attrs(self):
-        return FrozenOrderedDict((k, self.ds.getncattr(k))
-                                 for k in self.ds.ncattrs())
+        with self.ensure_open(autoclose=True):
+            attrs = FrozenOrderedDict((k, self.ds.getncattr(k))
+                                      for k in self.ds.ncattrs())
+        return attrs
 
     def get_dimensions(self):
-        return FrozenOrderedDict((k, len(v))
-                                 for k, v in iteritems(self.ds.dimensions))
+        with self.ensure_open(autoclose=True):
+            dims = FrozenOrderedDict((k, len(v))
+                                     for k, v in iteritems(self.ds.dimensions))
+        return dims
 
     def get_encoding(self):
-        encoding = {}
-        encoding['unlimited_dims'] = {
-            k for k, v in self.ds.dimensions.items() if v.isunlimited()}
+        with self.ensure_open(autoclose=True):
+            encoding = {}
+            encoding['unlimited_dims'] = {
+                k for k, v in self.ds.dimensions.items() if v.isunlimited()}
         return encoding
 
     def set_dimension(self, name, length):
-        self.ds.createDimension(name, size=length)
+        with self.ensure_open(autoclose=False):
+            self.ds.createDimension(name, size=length)
 
     def set_attribute(self, key, value):
-        if self.format != 'NETCDF4':
-            value = encode_nc3_attr_value(value)
-        self.ds.setncattr(key, value)
+        with self.ensure_open(autoclose=False):
+            if self.format != 'NETCDF4':
+                value = encode_nc3_attr_value(value)
+            self.ds.setncattr(key, value)
+
+    def set_variables(self, *args, **kwargs):
+        with self.ensure_open(autoclose=False):
+            super(NetCDF4DataStore, self).set_variables(*args, **kwargs)
 
     def prepare_variable(self, name, variable, check_encoding=False,
                          unlimited_dims=None):
@@ -305,15 +328,18 @@ class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
             # set attributes one-by-one since netCDF4<1.0.10 can't handle
             # OrderedDict as the input to setncatts
             nc4_var.setncattr(k, v)
+
         return nc4_var, variable.data
 
     def sync(self):
-        super(NetCDF4DataStore, self).sync()
-        self.ds.sync()
+        with self.ensure_open(autoclose=True):
+            super(NetCDF4DataStore, self).sync()
+            self.ds.sync()
 
     def close(self):
-        ds = self.ds
-        # netCDF4 only allows closing the root group
-        while ds.parent is not None:
-            ds = ds.parent
-        ds.close()
+        if self._isopen:
+            # netCDF4 only allows closing the root group
+            ds = find_root(self.ds)
+            if ds._isopen:
+                ds.close()
+            self._isopen = False
--- a/xarray/backends/pynio_.py
+++ b/xarray/backends/pynio_.py
@@ -7,39 +7,43 @@ import functools
 import numpy as np
 
 from .. import Variable
-from ..core.utils import FrozenOrderedDict, Frozen, NDArrayMixin
+from ..core.utils import (FrozenOrderedDict, Frozen,
+                          NdimSizeLenMixin, DunderArrayMixin)
 from ..core import indexing
 
 from .common import AbstractDataStore, DataStorePickleMixin
 
 
-class NioArrayWrapper(NDArrayMixin):
+class NioArrayWrapper(NdimSizeLenMixin, DunderArrayMixin):
 
     def __init__(self, variable_name, datastore):
         self.datastore = datastore
         self.variable_name = variable_name
+        array = self.get_array()
+        self.shape = array.shape
+        self.dtype = np.dtype(array.typecode())
 
-    @property
-    def array(self):
+    def get_array(self):
+        self.datastore.assert_open()
         return self.datastore.ds.variables[self.variable_name]
 
-    @property
-    def dtype(self):
-        return np.dtype(self.array.typecode())
-
     def __getitem__(self, key):
-        if key == () and self.ndim == 0:
-            return self.array.get_value()
-        return self.array[key]
+        with self.datastore.ensure_open(autoclose=True):
+            array = self.get_array()
+            if key == () and self.ndim == 0:
+                return array.get_value()
+            return array[key]
 
 
 class NioDataStore(AbstractDataStore, DataStorePickleMixin):
     """Store for accessing datasets via PyNIO
     """
-    def __init__(self, filename, mode='r'):
+    def __init__(self, filename, mode='r', autoclose=False):
         import Nio
         opener = functools.partial(Nio.open_file, filename, mode=mode)
         self.ds = opener()
+        self._autoclose = autoclose
+        self._isopen = True
         self._opener = opener
         self._mode = mode
 
@@ -48,14 +52,17 @@ class NioDataStore(AbstractDataStore, DataStorePickleMixin):
         return Variable(var.dimensions, data, var.attributes)
 
     def get_variables(self):
-        return FrozenOrderedDict((k, self.open_store_variable(k, v))
-                                 for k, v in self.ds.variables.iteritems())
+        with self.ensure_open(autoclose=False):
+            return FrozenOrderedDict((k, self.open_store_variable(k, v))
+                                     for k, v in self.ds.variables.iteritems())
 
     def get_attrs(self):
-        return Frozen(self.ds.attributes)
+        with self.ensure_open(autoclose=True):
+            return Frozen(self.ds.attributes)
 
     def get_dimensions(self):
-        return Frozen(self.ds.dimensions)
+        with self.ensure_open(autoclose=True):
+            return Frozen(self.ds.dimensions)
 
     def get_encoding(self):
         encoding = {}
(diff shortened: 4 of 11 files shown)


pwolfram:
The intent of this PR is to address (or at least partially address) the following issues:

 * https://github.com/pydata/xarray/issues/463
 * https://github.com/pydata/xarray/issues/798
 * https://github.com/CCI-Tools/cate-core/issues/102
 * https://github.com/MPAS-Dev/MPAS-Analysis/issues/49

xarray/test/test_backends.py
```diff
@@ -492,6 +492,21 @@ def create_tmp_file(suffix='.nc', allow_cleanup_failure=False):
             if not allow_cleanup_failure:
                 raise
 
+@contextlib.contextmanager
+def create_tmp_files(nfiles, suffix='.nc', allow_cleanup_failure=False):
```
shoyer: Any reason why you can't reuse `create_tmp_file` internally here?

This would be most straightforwardly done with `contextlib.ExitStack`, though we would need a backport to Python 2.7:
https://docs.python.org/3.6/library/contextlib.html#contextlib.ExitStack

pwolfram:
@shoyer, all the checks "pass" but there are still errors in the "allowed" list.  If you get a change could you please provide me some perspective on whether these are errors on my end or not?  I'm not exactly sure how to interpret them.

Once I know I have correctness in this code I plan to fix the inlines you graciously highlighted above.  I think we are getting close here, assuming that I have enough testing to demonstrate we have accurately fixed the too many open file issue.  Any additional ideas you have for tests would be really helpful too.

shoyer:
@pwolfram the allowed failures are pre-existing, not related to this change.

pwolfram:
Thanks @shoyer, Does that mean if the checks pass the code is at least minimally correct in terms of not breaking previous design choices?  E.g., does this imply that we are ok except for cleanup / implementation details on this PR?

shoyer:
> Does that mean if the checks pass the code is at least minimally correct in terms of not breaking previous design choices? E.g., does this imply that we are ok except for cleanup / implementation details on this PR?

If the checks pass, it means that it doesn't directly break anything that we have tests for. Which should cover *most* functionality. However, we'll still need to be careful not to introduce performance regressions -- we don't have any automated performance tests yet.

pwolfram:
@shoyer, I just realized this might conflict with #1087.  Do you foresee this causing problems and what order do you plan to merge this PR and #1087 (which obviously predates this one...)?  We are running into the snag with #463 in our analysis and my personal preference would be to get some type of solution into place sooner than later.  Thanks for considering this request.

Also, I'm not sure exactly the best way to test performance either.  Could we potentially use something like the "toy" test cases for this purpose?  Ideally we would have a test case with O(100)  files to gain a clearer picture of the performance cost of this PR.

Please let me know what you want me to do with this PR-- should I clean it up in anticipation of a merge or just wait for now to see if there are extra things that need fixed via additional testing?  Note I have the full scipy, h5netcdf and pynio implementations that can also be reviewed because they weren't available when you did your review yesterday.

shoyer:
Don't worry about #1087 -- I can rebase it.

shoyer:
This should be totally fine without performance or compatibility concerns as long as we set `autoclose=False` by the default.

In the long term, it would be nice to handle autoclosing automatically (invoking it when the number of open files exceeds some limit), but we should probably be a little more clever for that.

pwolfram:
Thanks @shoyer.  This makes sense.  I think the path forward on the next round of edits should include making sure existing tests using `open_mfdataset` use both options to `autoclose`.  If we do this we could future-proof ourselves against loss due to accidental breaking of this new functionality and avoid potentially contaminating existing workflows via performance concerns.  

Documentation is also obviously required.

FYI as a heads up, I probably won't be able to get to this mid-week at the earliest but it appears we are close to a viable solution.

PeterDSteinberg:
I appreciate your work on this too-many-files-open error - I think your fixes will add a lot of value to the NetCDF multi-file functionality.  In [this notebook using K-Means clustering on multi-file NetCDF data sets](https://github.com/ContinuumIO/elm-examples/blob/master/notebooks/temperature-PDFs-clustering.ipynb) I have repeatedly experienced the too-many-open files error, even with attempts to adjust via `ulimit`.  I can test out the notebook again as this PR is finalized.

pwolfram:
@PeterDSteinberg, did this PR fix the issue for you?  I obviously need to update it but just wanted to confirm that the current branch resolved the too-many-open files error issue.  Also, do you have any idea of the performance impact of these changes I'm proposing?

xarray/test/test_backends.py
pwolfram: Thanks @shoyer, this makes the code cleaner.  However, it may be operating slower because there could be more contexts in play.  It does remove redundant code, which is always a plus.

xarray/backends/netCDF4_.py
```diff
@@ -175,6 +177,13 @@ def _extract_nc4_encoding(variable, raise_on_invalid=False, lsd_okay=True,
 
     return encoding
 
+def _close_ds(ds):
+    # netCDF4 only allows closing the root group
+    while ds.parent is not None:
+        ds = ds.parent
+    if ds.isopen():
```
pwolfram: Sure, sounds good.

xarray/backends/netCDF4_.py
```diff
@@ -206,10 +215,24 @@ def __init__(self, filename, mode='r', format='NETCDF4', group=None,
         self.ds = opener()
         self.format = format
         self.is_remote = is_remote_uri(filename)
-        self._opener = opener
         self._filename = filename
         self._mode = 'a' if mode == 'w' else mode
+        self._opener = functools.partial(_open_netcdf4_group, filename,
```
pwolfram: Changed to `self._opener = functools.partial(opener, mode=self._mode)`.  Is this the best way to do this from previous usage of 
```python
  self._opener = opener(mode=self._mode)
                                         mode=self._mode,
                                         group=group, clobber=clobber,
                                         diskless=diskless, persist=persist,
                                         format=format)
```
?

pwolfram:
@shoyer and @PeterDSteinberg I've updated this PR to reflect requested changes.

pwolfram:
There are still a few more issues that need ironed out.  I'll let you know when I've resolved them.

vnoel:
I'm just chiming in to signify my interest in seeing this issue solved. I have just hit "OSError: Too many open files". The data itself is not even huge, but it's scattered across many files and it's a PITA to revert to manual concatenation -- I've grown used to dask doing the work for me ;-)

shoyer:
@pwolfram this looks pretty close to me now -- let me know when it's ready for review.

pwolfram:
@shoyer, the pushed code represents my progress.  The initial PR had a bug-- essentially a calculation couldn't be performed following the load.  This fixes that bug and provides a test to ensure that this doesn't happen.  However, I'm having trouble with h5netcdf, which I'm not very familiar with compared to netcdf.  This represents my current progress, I just need some more time (or even inspiration from you) to sort out this last key issue...  

I'm getting the following error:

```bash
================================================================================================================== FAILURES ==================================================================================================================
___________________________________________________________________________________________ OpenMFDatasetTest.test_4_open_large_num_files_h5netcdf ___________________________________________________________________________________________

self = <xarray.tests.test_backends.OpenMFDatasetTest testMethod=test_4_open_large_num_files_h5netcdf>

    @requires_dask
    @requires_h5netcdf
    def test_4_open_large_num_files_h5netcdf(self):
>       self.validate_open_mfdataset_large_num_files(engine=['h5netcdf'])

xarray/tests/test_backends.py:1040: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
xarray/tests/test_backends.py:1018: in validate_open_mfdataset_large_num_files
    self.assertClose(ds.foo.sum().values, np.sum(randdata))
xarray/core/dataarray.py:400: in values
    return self.variable.values
xarray/core/variable.py:306: in values
    return _as_array_or_item(self._data)
xarray/core/variable.py:182: in _as_array_or_item
    data = np.asarray(data)
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/numpy/core/numeric.py:482: in asarray
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/array/core.py:1025: in __array__
    x = self.compute()
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/base.py:79: in compute
    return compute(self, **kwargs)[0]
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/base.py:179: in compute
    results = get(dsk, keys, **kwargs)
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:537: in get_sync
    raise_on_exception=True, **kwargs)
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:500: in get_async
    fire_task()
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:476: in fire_task
    callback=queue.put)
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:525: in apply_sync
    res = func(*args, **kwds)
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:268: in execute_task
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:248: in _execute_task
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:248: in <listcomp>
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:245: in _execute_task
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:245: in <listcomp>
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/async.py:249: in _execute_task
    return func(*args2)
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/dask/array/core.py:52: in getarray
    c = a[b]
xarray/core/indexing.py:401: in __getitem__
    return type(self)(self.array[key])
xarray/core/indexing.py:376: in __getitem__
    return type(self)(self.array, self._updated_key(key))
xarray/core/indexing.py:354: in _updated_key
    for size, k in zip(self.array.shape, self.key):
xarray/core/indexing.py:364: in shape
    for size, k in zip(self.array.shape, self.key):
xarray/core/utils.py:414: in shape
    return self.array.shape
xarray/backends/netCDF4_.py:37: in __getattr__
    return getattr(self.datastore.ds.variables[self.var], attr)
../../anaconda/envs/test_env_xarray35/lib/python3.5/contextlib.py:66: in __exit__
    next(self.gen)
xarray/backends/h5netcdf_.py:105: in ensure_open
    self.close()
xarray/backends/h5netcdf_.py:190: in close
    _close_ds(self.ds)
xarray/backends/h5netcdf_.py:70: in _close_ds
    find_root(ds).close()
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/h5netcdf/core.py:458: in close
    self._h5file.close()
../../anaconda/envs/test_env_xarray35/lib/python3.5/site-packages/h5py/_hl/files.py:302: in close
    self.id.close()
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper (/Users/travis/miniconda3/conda-bld/work/h5py-2.6.0/h5py/_objects.c:2840)
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper (/Users/travis/miniconda3/conda-bld/work/h5py-2.6.0/h5py/_objects.c:2798)
    ???
h5py/h5f.pyx:282: in h5py.h5f.FileID.close (/Users/travis/miniconda3/conda-bld/work/h5py-2.6.0/h5py/h5f.c:3905)
    ???
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper (/Users/travis/miniconda3/conda-bld/work/h5py-2.6.0/h5py/_objects.c:2840)
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper (/Users/travis/miniconda3/conda-bld/work/h5py-2.6.0/h5py/_objects.c:2798)
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: dictionary changed size during iteration

h5py/_objects.pyx:119: RuntimeError
============================================================================================ 1 failed, 1415 passed, 95 skipped in 116.54 seconds =============================================================================================
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x10f16e598>
  File "/Users/pwolfram/anaconda/envs/test_env_xarray35/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
```

shoyer:
I'll take a look tomorrow. Getting all these backends to behave correctly
and consistently is a constant battle.

shoyer:
I can reproduce issues with h5netcdf on my machine. Of course, I get slightly different error messages -- either an error about closing a closed file from h5py or a segfault.

So far, I've verified that it isn't a multi-threading issue -- I get the same errors when dask is run in single-threaded mode (`dask.set_options(get=dask.async.get_sync`). That's not encouraging, suggesting that this may be a legitimate h5netcdf or (perhaps more likely) h5py bug.

pwolfram:
@shoyer, it is quite possible the error is on my end because the error message implies that modification of the data prior to closing it may be the issue.  However, I'm not getting that type of an error following a similar philosophy for the other backends, which is confusing.  

Do you have advice on hunting this bug further?  You obviously have more experience with `h5py` and `h5netcdf` than I do.

shoyer:
I'm still working on it. A possible culprit is some sort of dangling reference that stops the file from being actually closed by h5py. If we can't figure it out in a few days, we can just disable autoclose for h5netcdf.

pwolfram:
@shoyer, I'll leave the ball in your court for now but am happy to take another look before we go the disabling autoclose route for h5netcdf.  Please let me know when/if you'd like me to take another look.

pwolfram:
@shoyer, I'd like to see this code integrated if possible.  Should we just disable autoclose for h5netcdf and merge?

xarray/backends/h5netcdf_.py
```diff
@@ -55,43 +86,63 @@ def __init__(self, filename, mode='r', format=None, group=None,
...
         self.ds = opener()
+        self._isopen = True
         self.format = format
         self._opener = opener
         self._filename = filename
         self._mode = mode
         super(H5NetCDFStore, self).__init__(writer)
 
+    @contextlib.contextmanager
+    def ensure_open(self, autoclose=True):
```
shoyer: This is currently defined redundantly in three or four different places. It would be better to use some sort of subclassing for this.

shoyer: Also, why does `autoclose` default to True? This will start auto-opening/closing Datasets even if they were explicitly closed by calling `.close()`. That seems undesirable.

Possibly this should be saved as another DataStore attribute.

xarray/backends/netCDF4_.py
```diff
@@ -25,14 +26,29 @@
...
+
 class BaseNetCDF4Array(NDArrayMixin):
     def __init__(self, variable_name, datastore):
         self.datastore = datastore
         self.variable_name = variable_name
 
     @property
     def array(self):
-        return self.datastore.ds.variables[self.variable_name]
+        return VariableWrapper(self.datastore, self.variable_name)
```
shoyer: This will result in a lot of churn opening/closing files just to get things like the dtype for variables.

Instead, I would save `dtype` and `shape` when this object is created, and only use `ensure_open` when doing indexing. In other words, inherit from `NdimSizeLenMixin` instead of `NDArrayMixin` -- there's no particular reason why we need to keep the `array` property.

xarray/tests/test_backends.py
```diff
@@ -992,6 +999,46 @@ def test_encoding_unlimited_dims(self):
             with pytest.warns(UserWarning):
                 ds.to_netcdf(tmp_file, engine='h5netcdf', unlimited_dims=['y'])
 
+class OpenMFDatasetTest(TestCase):
+    def validate_open_mfdataset_large_num_files(self, engine, nfiles=2000):
```
shoyer: This makes for a very expensive unit test. Maybe change the default to something faster like `nfiles=10`?

xarray/backends/h5netcdf_.py
pwolfram: @shoyer, the problem is that in order to avoid the open file limit we need to periodically open and close files.  Hence, we should open a file, read it, and then close it.  Otherwise during `open_mfdataset`, for instance, we will have too many open file and error.  However, you'll note that if we need to set properties we don't autoclose the files.  I've made this more explicit to make the code clearer by getting rid of the default `autoclose` argument.

xarray/tests/test_backends.py
pwolfram: We need to have a larger number of files opened than the limitation from the file system (> 1000) in order to demonstrate the problem.  I don't think we would have any problem for `nfiles=10` because this is much less than the open file limitation of the file system.

shoyer: I agree that we need `nfiles` to be large to make this test meaningful, though even a test with `nfiles=10` is somewhat informative. I added a follow-up in #1309.

xarray/backends/h5netcdf_.py
pwolfram: Agreed

xarray/tests/test_backends.py
pwolfram: I'm happy to do both-- that is probably better anyway.  I'll split this into two test then...

xarray/backends/h5netcdf_.py
pwolfram: I think I see what you are saying-- essentially we need to make sure that this action only happens if we specify, on loading a dataset via `open_mfdataset` that `autoclose=True`.  Otherwise, we never close a file, correct?

xarray/backends/netCDF4_.py
pwolfram: @shoyer, this is confusing to me because there are times when the `array` property is needed for a variable.  If the dataset is closed, this causes an error (hence why this code is here).  Are you suggesting we should basically cache `array.shape` and `array.dtype`?  I think I'm missing something more important here, something like `shape` and `dtype` will be read from the base array `BaseNetCDF4Array` without need for `array` if they exist.

I fully agree that there is a lot of churn when handling a large number of files, but it isn't obvious to me how to avoid this issue.

xarray/tests/test_backends.py
rabernat: One solution could be to use the [resource](https://docs.python.org/2/library/resource.html) package within the test to temporarily lower the ulimit on open files.

xarray/backends/h5netcdf_.py
shoyer: That's exactly right.

xarray/backends/netCDF4_.py
shoyer: > Are you suggesting we should basically cache array.shape and array.dtype?

Yes, exactly. We should read these in when creating these objects, when the dataset is loaded. Then we only need to go back to the underlying data on disk for pulling out array values with  `__getitem__`.

Something like:
```python
class BaseNetCDF4Array(NdimSizeLenMixin):
    def __init__(self, variable_name, datastore):
        self.datastore = datastore
        self.variable_name = variable_name
        # don't autoclose yet, because the file is already open for reading
        # variable metadata
        array = self.get_array(autoclose=False)
        self.dtype = array.dtype
        self.shape = array.shape

    def get_array(self, autoclose):
        with self.datastore.ensure_open(autoclose=autoclose):
            return self.datastore.ds.variables[self.variable_name]

    def __getitem__(self, key):
        array = self.get_array(autoclose=True)
        return array[key]

    def __array__(self, dtype=None):
        return np.asarray(self[...], dtype=dtype)
```

xarray/backends/h5netcdf_.py
shoyer: Maybe we should just ignore `autoclose=True` if the datastore wasn't opened with auto-closing.

xarray/tests/test_backends.py
pwolfram: Thanks @rabernat!  This context manager seems to do the trick for your future reference:

```python
@contextlib.contextmanager
def limit_open_num_files(nfiles=9):
    currentlim = resource.getrlimit(resource.RLIMIT_NOFILE)
    try:
        resource.setrlimit(resource.RLIMIT_NOFILE, 
                           (nfiles, resource.RLIM_INFINTY))
        yield
    finally:
        resource.setrlimit(resource.RLIMIT_NOFILE, currentlim)
        yield
```

usage 

```python
with limit_open_num_files():
     ...
```

shoyer: Oh, awesome -- that will really help.

pwolfram: Strange though, it is causing false positives on too many open file errors relative to using `nfiles=2000` from before- I'm not sure what is happening.  

Another downside here is needing to support linux/windows since `resource` is just *nix, not windows.  I think I'll revert to using fast / slow test cases for now because of these two issues.

pwolfram: I should note there is a third issue: I have a stack of context managers wrapped inside a context manager if I go this approach, which is confusing too so we will get clearer code by making slow/fast tests.  But, this is obviously at least partially unsettling because of the runtime cost.

xarray/backends/h5netcdf_.py
pwolfram: This seems like the cleanest solution to me too-- I've implemented this now.

xarray/backends/netCDF4_.py
pwolfram: Interesting- it looks like `__array__` above and `self.dtype` weren't necessarily needed.  Tests pass at least.  Should I be concerned about this issue?  For instance, perhaps they will be needed for `h5netcdf`.  I don't fully know the xarray internal structure well enough to understand the significance here, but am curious to find out if it is easy to explain.

xarray/tests/test_backends.py
pwolfram: But, this works for `nfiles=30`, not `nfiles=10` so we may be ok.  I'm going to include a commit with this and you both can rip it apart because it isn't 100% clear to me what to do here...

pwolfram:
@shoyer, this should essentially be ready to go now with the exception that I didn't come to a clean resolution on testing the too many open files issue via `resource`.  Thoughts on this issue?

pwolfram:
I obviously need to squash prior to the merge, however...

pwolfram:
Note, if we include `3961621` for resource-limited testing we will need to do something special for windows on appveyor.  We could skip this in the short-term to make this merge and deal with these types of issues in #1309, which is what probably makes the most sense to help keep the scope of this PR more limited.

pwolfram:
@shoyer, I've collapsed all the commits to a single commit.  At this point I would say this meets the scope of this PR and it should be ready to merge once CI finishes its check and it passes.  Please feel free to take another pass on the code for changes that you would like to see made.

pwolfram:
@PeterDSteinberg and @vnoel, do you mind stress-testing this PR to make sure all works as expected?  You can install easily via pip with
`pip -v install git+ssh://git@github.com/pwolfram/xarray@fix_too_many_open_files`

pwolfram:
@rabernat, I suspect you have some datasets that could be used to stress test this PR too... thanks for the help and advice yesterday.

vnoel:
Hey @pwolfram, I installed your branch and tried to `open_mfdataset` on 1474 hdf5 files.

I got the following:
```
  File "/users/noel/.conda/envs/python3/lib/python3.6/site-packages/xarray/backends/api.py", line 524, in open_mfdataset
  File "/users/noel/.conda/envs/python3/lib/python3.6/site-packages/xarray/backends/api.py", line 524, in <listcomp>
  File "/users/noel/.conda/envs/python3/lib/python3.6/site-packages/xarray/backends/api.py", line 299, in open_dataset
  File "/users/noel/.conda/envs/python3/lib/python3.6/site-packages/xarray/backends/netCDF4_.py", line 203, in __init__
  File "/users/noel/.conda/envs/python3/lib/python3.6/site-packages/xarray/backends/netCDF4_.py", line 178, in _open_netcdf4_group
  File "netCDF4/_netCDF4.pyx", line 1848, in netCDF4._netCDF4.Dataset.__init__ (netCDF4/_netCDF4.c:13992)
OSError: Too many open files
```

pwolfram:
@vnoel, did you use `autoclose=True`?  Can you please share the code snippet you used to instantiate this?

pwolfram:
Note, the force push was just to get travis-ci to "reboot".  There was no real change to the code.

pwolfram:
Is something wrong with travis-ci right now?  It doesn't look like it is running the tests...

vnoel:
with that keyword the open_mfdataset goes through !
my code crashes later because the data is too big to fit in memory, but that's another problem ;-)

pwolfram:
Awesome, thanks @vnoel for testing this for us all on your dataset!

xarray/backends/scipy_.py
```diff
@@ -37,30 +37,39 @@ def __init__(self, variable_name, datastore):
...
     def array(self):
-        return self.datastore.ds.variables[self.variable_name].data
+        with self.datastore.ensure_open(autoclose=True):
+            return self.datastore.ds.variables[self.variable_name].data
 
     @property
     def dtype(self):
-        # always use native endianness
-        return np.dtype(self.array.dtype.kind + str(self.array.dtype.itemsize))
+        with self.datastore.ensure_open(autoclose=True):
```
shoyer: We want to use the same approach for scipy as for netcdf4 to avoid repeatedly opening/closing files (i.e., we should cache dtype). Each time we open, scipy does a bunch of logic in python to read in the metadata for the entire netCDF file.

xarray/backends/scipy_.py
```diff
@@ -98,26 +107,33 @@ def __init__(self, filename_or_obj, mode='r', format=None, group=None,
...
+        self._isopen = True
         self._opener = opener
         self._mode = mode
 
         super(ScipyDataStore, self).__init__(writer)
 
     def open_store_variable(self, name, var):
-        return Variable(var.dimensions, ScipyArrayWrapper(name, self),
-                        _decode_attrs(var._attributes))
+        with self.ensure_open(autoclose=True):
```
shoyer: as above, should not autoclose here.

xarray/tests/__init__.py
```diff
@@ -146,6 +147,9 @@ def assertArrayEqual(self, a1, a2):
     def assertEqual(self, a1, a2):
         assert a1 == a2 or (a1 != a1 and a2 != a2)
 
+    def assertClose(self, a1, a2, rtol=1e-05, atol=1e-8):
```
shoyer: for consistency with the function name, I would call this `assertAllClose`

xarray/tests/test_backends.py
```diff
@@ -992,6 +999,73 @@ def test_encoding_unlimited_dims(self):
...
+    def validate_open_mfdataset_autoclose(self, engine, nfiles=10):
+        randdata = np.random.randn(nfiles)
+        original = Dataset({'foo': ('x', randdata)})
+        # test standard open_mfdataset approach with too many files
+        with create_tmp_files(nfiles) as tmpfiles:
+            for readengine in engine:
+                writeengine = readengine if readengine != 'pynio' else 'netcdf4'
+                # split into multiple sets of temp files
+                for ii in original.x.values:
+                    original.isel(x=slice(ii, ii+1))\
```
shoyer: Use parentheses instead of explicit continuation with `\`

xarray/backends/netCDF4_.py
```diff
@@ -25,35 +25,33 @@
...
 
-class BaseNetCDF4Array(NDArrayMixin):
+class BaseNetCDF4Array(NdimSizeLenMixin):
     def __init__(self, variable_name, datastore):
         self.datastore = datastore
         self.variable_name = variable_name
+        # don't autoclose yet, because the file is already open for reading
+        # variable metadata
+        array = self.get_array(autoclose=False)
+        self.shape = array.shape
```
shoyer: This made me a little nervous about our test coverage for the `dtype` property on unloaded arrays, see #1311.

pwolfram:
@shoyer, do you mind restarting the Travis-CI?

shoyer:
I restarted all the timed out tests on Travis, hopefully they pass now!

pwolfram:
They did- thank you!  I'll take a quick look at your comments now so we can hopefully get this merged soon.

xarray/backends/h5netcdf_.py
```diff
@@ -49,49 +82,60 @@ class H5NetCDFStore(WritableCFDataStore, DataStorePickleMixin):
...
     def __init__(self, filename, mode='r', format=None, group=None,
-                 writer=None):
+                 writer=None, autoclose=False):
         if format not in [None, 'NETCDF4']:
             raise ValueError('invalid format for h5netcdf backend')
         opener = functools.partial(_open_h5netcdf_group, filename, mode=mode,
                                    group=group)
         self.ds = opener()
+        if autoclose:
+            warnings.warn('autoclose=True is disabled for the h5netcdf backend '
```
pwolfram: @shoyer, do you want to explicitly let this go on without an error?  It may catastrophically error out if we just attempt to do `autoclose=True`.  My preference would be to use a `NotImplemented` error.

xarray/backends/h5netcdf_.py
```diff
@@ -2,17 +2,45 @@
...
 
-from .common import WritableCFDataStore, DataStorePickleMixin
+from .common import WritableCFDataStore, DataStorePickleMixin, find_root
 from .netCDF4_ import (_nc4_group, _nc4_values_and_dtype,
-                       _extract_nc4_variable_encoding, BaseNetCDF4Array)
-
+                       _extract_nc4_variable_encoding, NetCDF4ArrayWrapper)
+
+
+class H5NetcdFArrayWrapper(NetCDF4ArrayWrapper):
```
pwolfram: @shoyer, agreed.  Thanks for catching this goof.

xarray/tests/test_backends.py
pwolfram: That is clearer... thanks.

xarray/backends/h5netcdf_.py
shoyer: When I tested the previous version of this PR, it worked if the number of files was small -- just not if it exceeds the file limit.

But I'm happy with either option here.

xarray/tests/__init__.py
pwolfram: Done

xarray/backends/scipy_.py
pwolfram: The `NumpyIndexingAdapter` class doesn't allow setting the `dtype` property in the constructor.  Should I inherit something else?  I'm missing some key info on design intent here.

xarray/backends/pynio_.py
```diff
@@ -7,55 +7,41 @@
...
 
 from .. import Variable
+from .netCDF4_ import BaseNetCDF4Array
 from ..core.utils import FrozenOrderedDict, Frozen, NDArrayMixin
 from ..core import indexing
 
 from .common import AbstractDataStore, DataStorePickleMixin
 
 
-class NioArrayWrapper(NDArrayMixin):
```
pwolfram: I inherited from `BaseNetCDF4Array` instead to resuse the code for simplicity.  This seemed like redundant code.

xarray/backends/netCDF4_.py
```diff
@@ -195,74 +192,90 @@ class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
...
-                encoding['contiguous'] = False
-                encoding['chunksizes'] = tuple(chunking)
-        # TODO: figure out how to round-trip "endian-ness" without raising
-        # warnings from netCDF4
-        # encoding['endian'] = var.endian()
-        pop_to(attributes, encoding, 'least_significant_digit')
-        # save source so __repr__ can detect if it's local or not
-        encoding['source'] = self._filename
-        encoding['original_shape'] = var.shape
+        with self.ensure_open(autoclose=True):
```
pwolfram: Agreed

xarray/backends/netCDF4_.py
```diff
@@ -25,35 +25,33 @@
...
-    def dtype(self):
-        dtype = self.array.dtype
-        if dtype is str:
-            # return object dtype because that's the only way in numpy to
-            # represent variable length strings; it also prevents automatic
-            # string concatenation via conventions.decode_cf_variable
-            dtype = np.dtype('O')
-        return dtype
+    def __getitem__(self, key):
+        with self.datastore.ensure_open(autoclose=True):
```
pwolfram: Done

xarray/backends/netCDF4_.py
pwolfram: Done

xarray/backends/scipy_.py
shoyer: `NumpyIndexingAdapter` was just for overloading `__getitem__`. So we could simply move that into the `__getitem__` call instead, e.g., `NumpyIndexingAdapter(self.get_array())[key]`

pwolfram:
Note: tests pass before submitting changes.

xarray/backends/scipy_.py
pwolfram: Why not just use `BaseNetCDF4Array` then instead?

xarray/backends/pynio_.py
shoyer: Most likely, I think we need different work-arounds for pynio's quirks than netcdf4's quirks.

xarray/backends/scipy_.py
shoyer: We need to use `NumpyIndexAdapater` for scipy (which uses numpy arrays directly in it's netcdf reader), but not for netCDF4 (which uses it's own array class with different "orthogonal" indexing behavior)

xarray/backends/h5netcdf_.py
```diff
@@ -49,49 +82,60 @@ class H5NetCDFStore(WritableCFDataStore, DataStorePickleMixin):
...
         self._opener = opener
         self._filename = filename
         self._mode = mode
         super(H5NetCDFStore, self).__init__(writer)
 
     def open_store_variable(self, name, var):
-        dimensions = var.dimensions
-        data = indexing.LazilyIndexedArray(BaseNetCDF4Array(name, self))
-        attrs = _read_attributes(var)
+        with self.ensure_open(autoclose=True):
```
pwolfram: Done

pwolfram:
@shoyer- can you please have another look following edits as you requested?

pwolfram:
@shoyer, all tests are passing now following your recommended edits.

xarray/backends/netCDF4_.py
```diff
@@ -115,13 +118,6 @@ def _nc4_group(ds, group, mode):
         return ds
 
 
-def _ensure_fill_value_valid(data, attributes):
```
shoyer: Why did you drop this? I think we probably still need it

xarray/backends/netCDF4_.py
```diff
@@ -25,35 +25,34 @@
...
-        else:
-            getitem = operator.getitem
-
         try:
-            data = getitem(self.array, key)
+            if self.datastore.is_remote:  # pragma: no cover
+                getitem = functools.partial(robust_getitem, catch=RuntimeError)
+                data = getitem(self.array, key)
+            else:
+                getitem = super(NetCDF4ArrayWrapper, self).__getitem__
```
pwolfram: Sounds good.

xarray/backends/netCDF4_.py
```diff
@@ -25,35 +25,34 @@
...
-        if self.datastore.is_remote:  # pragma: no cover
-            getitem = functools.partial(robust_getitem, catch=RuntimeError)
-        else:
-            getitem = operator.getitem
-
         try:
-            data = getitem(self.array, key)
+            if self.datastore.is_remote:  # pragma: no cover
+                getitem = functools.partial(robust_getitem, catch=RuntimeError)
+                data = getitem(self.array, key)
```
pwolfram: Thanks for the fix.

xarray/backends/netCDF4_.py
```diff
@@ -25,35 +25,34 @@
...
-    def array(self):
-        return self.datastore.ds.variables[self.variable_name]
+    def get_array(self, autoclose):
+        with self.datastore.ensure_open(autoclose=autoclose):
+            return self.datastore.ds.variables[self.variable_name]
 
-    @property
-    def dtype(self):
-        dtype = self.array.dtype
-        if dtype is str:
```
pwolfram: Thanks, this wasn't obvious to me.

xarray/backends/netCDF4_.py
pwolfram: I pushed it back but initially removed it because it wasn't called but have put it back.

pwolfram: To clarify, it looks like I removed that code because it didn't appear to be necessary, which in hindsight was unwise-- apologies about that poor choice because I should have left the code where it was initially.  Assuming I didn't introduce an error we should be able to squash and merge, correct?

pwolfram:
The tests passed so I will squash the commits...

pwolfram:
@shoyer, we can roll back the squash if you want because for whatever reason my interactive rebase removed your name from tagged to the commit too, which is not what I expected.  This is obviously suboptimal and the choice of having two commits versus one is up to you.  I think the commit prior to the squash was `9c274eb`.

shoyer:
@pwolfram no worries, I am not concerned at all about credit for my commit :). It's possible to do gymnastics with git rebase to preserve history in cases like this but it's questionable if it's worth the trouble.

pwolfram:
Thanks @shoyer.  Ok, so we pass checks following merge of #1311.  Is there anything else that we need to do on this PR prior to merging?

mrocklin:
Just seeing this now.  I'm very glad to see progress here.  Thanks @pwolfram

xarray/backends/api.py
```diff
@@ -273,7 +281,8 @@ def maybe_decode_store(store, lock=False):
                                  "default engine or engine='scipy'")
             # if the string ends with .gz, then gunzip and open as netcdf file
             try:
-                store = backends.ScipyDataStore(gzip.open(filename_or_obj))
+                store = backends.ScipyDataStore(filename_or_obj, gzipfile=True,
+                                                autoclose=autoclose)
             except TypeError as e:
```
shoyer: I like moving `gzip.open` into ScipyDataStore. Can you move the try/except clause in there, too?

xarray/backends/h5netcdf_.py
```diff
@@ -37,6 +46,11 @@ def _read_attributes(h5netcdf_var):
 _extract_h5nc_encoding = functools.partial(_extract_nc4_variable_encoding,
                                            lsd_okay=False, backend='h5netcdf')
 
+def _close_ds(ds):
```
shoyer: This function is used exactly once. Just inline it instead.

xarray/backends/netCDF4_.py
```diff
@@ -174,6 +177,12 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
 
     return encoding
 
+def _close_ds(ds):
```
shoyer: This is defined redundantly with the function in `h5netcdf_`. But it's probably simpler just to inline it here, too -- no need to get too hung up over repeating a few lines of code.

xarray/tests/test_backends.py
```diff
@@ -1030,17 +1105,19 @@ def test_open_mfdataset(self):
...
+                        self.assertEqual(actual.foo.variable.data.chunks,
+                                         ((5, 5),))
+                        self.assertDatasetAllClose(original, actual)
+                    with open_mfdataset([tmp1, tmp2], chunks={'x': 3}, autoclose=close) as actual:
+                        self.assertEqual(actual.foo.variable.data.chunks,
+                                         ((3, 2, 3, 2),))
 
         with self.assertRaisesRegexp(IOError, 'no files to open'):
-            open_mfdataset('foo-bar-baz-*.nc')
+            for close in [True, False]:
```
shoyer: The loop needs to be outside the `with self.assertRaisesRegexp` clause.

But more generally, let's think a little carefully about whether we really need to test all of these cases for both autoclose=True and autoclose=False. My guess is that this is more than is necessary.

I would also prefer to use `pytest` style test fixtures for the `autoclose` argument. This ensures that each option ends up in a different test, which makes things much easier to debug when we get a failure.

pwolfram: Is it possible to have pytest run with and without a particular fixure automatically, e.g., via different tests with and without the fixture?  If so, this would be really helpful for this case. 

 I think it is valuable to do the full testing over both `autoclose` options because of the potential for failure based on the development challenges, e.g., braking functionality.

pwolfram: I think I understand how to use fixtures for both of these arguments now...

shoyer:
Subclasses also work in place of fixtures in many cases (we use them in
much of the existing code).

xarray/tests/test_backends.py
pwolfram: It turns out that they don't work correctly fixtures so I will just use subclassing so that each case has a unique test.

pwolfram:
@shoyer, if we generally cover test_backends for `autoclose=True`, then we should get the pickle testing for free:
```
xarray/tests/test_backends.py:181:    def test_pickle(self):
xarray/tests/test_backends.py:191:    def test_pickle_dataarray(self):
xarray/tests/test_backends.py:792:    def test_bytesio_pickle(self):
```
or was there some other test that is needed?

shoyer:
> @shoyer, if we generally cover test_backends for autoclose=True, then we should get the pickle testing for free

Agreed, that should do it.

pwolfram:
@shoyer, that subclass-based approach you outlined worked (fixture parameters really don't work with classes as far as I could tell).  We now have more comprehensive, named testing.  Note, there was one minor point that required more explicitly specification that arose from the more rigorous testing:

The scipy backend can handle objects like `BytesIO` that really aren't file handles and there doesn't appear to be a clean way to close these types of objects.  So, at present I'm explicitly setting `_autoclose=False` if they are encountered in the datastore.  If this needs to be changed, particularly since it doesn't affect existing behavior, I'd prefer this be resolved in a separate issue / PR if possible.

pwolfram:
@shoyer, I had a minor bug that is now removed.  The last caveat no longer applicable:

> The scipy backend can handle objects like BytesIO that really aren't file handles and there doesn't appear to be a clean way to close these types of objects. So, at present I'm explicitly setting _autoclose=False if they are encountered in the datastore. If this needs to be changed, particularly since it doesn't affect existing behavior, I'd prefer this be resolved in a separate issue / PR if possible.

I'll let you know when tests pass and this is ready for your final review.

pwolfram:
@shoyer, this is ready for the final review now.  Coveralls appears to have hung but other tests pass.

pwolfram:
@shoyer, all tests (including coveralls) passed.  Please let me know if you have additional concerns and if we could merge fairly soon, e.g., because of https://github.com/MPAS-Dev/MPAS-Analysis/issues/151 I would really appreciate it.

xarray/backends/h5netcdf_.py
```diff
@@ -2,18 +2,27 @@
...
 from ..core.utils import FrozenOrderedDict, close_on_error, Frozen
 from ..core.pycompat import iteritems, bytes_type, unicode_type, OrderedDict
 
-from .common import WritableCFDataStore, DataStorePickleMixin
+from .common import WritableCFDataStore, DataStorePickleMixin, find_root
 from .netCDF4_ import (_nc4_group, _nc4_values_and_dtype,
                        _extract_nc4_variable_encoding, BaseNetCDF4Array)
 
 
+class H5NetCDFFArrayWrapper(BaseNetCDF4Array):
```
shoyer: spelling: extra `F`

xarray/backends/scipy_.py
```diff
@@ -96,28 +110,38 @@ def __init__(self, filename_or_obj, mode='r', format=None, group=None,
             raise ValueError('invalid format for scipy.io.netcdf backend: %r'
                              % format)
 
+        # if the string ends with .gz, then gunzip and open as netcdf file
+        if type(filename_or_obj) is str and filename_or_obj.endswith('.gz'):
```
shoyer: Use `isinstance(filename_or_obj, basestring)` instead of `type(filename_or_obj) is str`.

Also, move this logic into `_open_scipy_netcdf` -- otherwise it won't work to reopen a gzipped file.

xarray/backends/scipy_.py
```diff
@@ -61,8 +63,18 @@ def _open_scipy_netcdf(filename, mode, mmap, version):
...
         filename = BytesIO(filename)
 
-    return scipy.io.netcdf_file(filename, mode=mode, mmap=mmap,
-                                version=version)
+    try:
+        ds = scipy.io.netcdf_file(filename, mode=mode, mmap=mmap,
+                                  version=version)
+    except TypeError as e:
+        # TODO: gzipped loading only works with NetCDF3 files.
+        if 'is not a valid NetCDF 3 file' in e.message:
```
shoyer: This should only be triggered when reading a gzipped file. Right now, it can be triggered whenever an invalid netCDF3 file is read, gzipped or not.

This should be easier to fix when you move `gzip.open` into this helper function (see my comment below).

xarray/tests/test_backends.py
```diff
@@ -992,6 +1032,80 @@ def test_encoding_unlimited_dims(self):
             with pytest.warns(UserWarning):
                 ds.to_netcdf(tmp_file, engine='h5netcdf', unlimited_dims=['y'])
 
+# tests pending h5netcdf fix
+#class H5NetCDFDataTestAutocloseTrue(H5NetCDFDataTest):
+#    autoclose = True
+
+class OpenMFDatasetTest(TestCase):
+    autoclose = True
```
shoyer: I don't think you need this class variable, since this test class does not use inheritance.

xarray/tests/test_backends.py
```diff
@@ -992,6 +1032,80 @@ def test_encoding_unlimited_dims(self):
             with pytest.warns(UserWarning):
                 ds.to_netcdf(tmp_file, engine='h5netcdf', unlimited_dims=['y'])
 
+# tests pending h5netcdf fix
+#class H5NetCDFDataTestAutocloseTrue(H5NetCDFDataTest):
+#    autoclose = True
+
+class OpenMFDatasetTest(TestCase):
```
shoyer: rename to something like `OpenMFDatasetManyFilesTest` (we have other tests of `open_mfdataset`)

xarray/tests/test_backends.py
```diff
@@ -1139,6 +1257,8 @@ def test_dataarray_compute(self):
         self.assertTrue(computed._in_memory)
         self.assertDataArrayAllClose(actual, computed)
 
+class DaskTestAutocloseTrue(DaskTest):
+    autoclose=True
```
shoyer: Please run some sort of PEP8 check, e.g., `git diff upstream/master | flake8 --diff`. There should be extra spaces around the `=` sign here.

xarray/tests/test_backends.py
```diff
@@ -992,6 +1032,80 @@ def test_encoding_unlimited_dims(self):
...
+                writeengine = readengine if readengine != 'pynio' else 'netcdf4'
+                # split into multiple sets of temp files
+                for ii in original.x.values:
+                    (
+                     original.isel(x=slice(ii, ii+1))
+                    ).to_netcdf(tmpfiles[ii], engine=writeengine)
+
+                # check that calculation on opened datasets works properly
+                ds = open_mfdataset(tmpfiles, engine=readengine,
+                                    autoclose=self.autoclose)
```
shoyer: just set `autoclose=True` here.

xarray/tests/test_backends.py
```diff
@@ -992,6 +1032,80 @@ def test_encoding_unlimited_dims(self):
...
+        randdata = np.random.randn(nfiles)
+        original = Dataset({'foo': ('x', randdata)})
+        # test standard open_mfdataset approach with too many files
+        with create_tmp_files(nfiles) as tmpfiles:
+            for readengine in engine:
+                writeengine = readengine if readengine != 'pynio' else 'netcdf4'
+                # split into multiple sets of temp files
+                for ii in original.x.values:
+                    (
+                     original.isel(x=slice(ii, ii+1))
```
shoyer: I think this would be slightly more readable with an intermediate variable, or removing the line break on the line above.

xarray/backends/scipy_.py
pwolfram: Thanks for pointing this out-- sorry for this sloppiness.

xarray/tests/test_backends.py
pwolfram: I'm going to ignore PEP8 to `xarray/core/pycompat.py` because code in there is essentially a copy / paste.

shoyer: sounds good

shoyer:
OK, in it goes!

pwolfram:
Thanks a bunch @shoyer!

PR #1200, opened 2017-01-11
DOC: fix small typo/mistake (NaN value not dtype)

gerritholl:
Fix small mistake in documentation.  It said NaN is not a valid dtype for integer dtypes, this surely should be NaN is not a valid value for integer dtypes.


shoyer:
Thanks!

PR #1204, opened 2017-01-12
combine_first by using apply_ufunc in ops.fillna

chunweiyuan:
Implementing `ops.fillna` using `apply_ufunc`, with a `join` kwarg.  The new `ops.fillna` is then used to implement `combine_first` in both DataArray and Dataset objects.


xarray/core/ops.py
```diff
@@ -279,11 +279,18 @@ def count(data, axis=None):
...
     """Fill missing values in this object with data from the other object.
     Follows normal broadcasting and alignment rules.
     """
-    return where(isnull(data), other, data)
+    from .computation import apply_ufunc
+
+    def _fillna(data, other):
+        left, right = np.broadcast_arrays(data, other)
+        result = left.copy()  # view must be copied before being written
+        result[isnull(result)] = right[isnull(result)]
```
shoyer: An advantage of using `where(isnull(data), other, data)` instead of all the logic in this function is that it works on dask arrays, too.

xarray/core/groupby.py
```diff
@@ -382,7 +382,15 @@ def fillna(self, value):
         Dataset.fillna
         DataArray.fillna
         """
-        return self._fillna(value)
+        def _yield_applied(this, other):
```
shoyer: I *think* `apply_ufunc` should even work on groupby objects, so you shouldn't need this helper function. I would be interested to know I'm wrong about that, though.

xarray/core/ops.py
```diff
@@ -279,11 +279,18 @@ def count(data, axis=None):
...
+        return result
+    return apply_ufunc(_fillna, data, other, join=join)
```
shoyer: should have `dask_array='allowed'`

xarray/core/dataset.py
```diff
@@ -1955,7 +1955,35 @@ def fillna(self, value):
...
+        return out
+
+    def combine_first(self, other):
+        """Combine two Datasets, default to data_vars of self.
+
+        The new coordinates follow the normal broadcasting and alignment rules
+        of ``join='outer'``.  Vacant cells in the expanded coordinates are
+        filled with np.nan.
+
+        Renders the same result as xr.merge([self, other]).
```
shoyer: This is not quite right -- `combine_first`/`fillna` are OK with conflicting values in `self` and `other`, but `merge` will raise an error.

xarray/core/dataarray.py
```diff
@@ -1097,10 +1097,28 @@ def fillna(self, value):
         if utils.is_dict_like(value):
             raise TypeError('cannot provide fill value as a dictionary with '
                             'fillna on a DataArray')
-        out = self._fillna(value)
+        out = ops.fillna(self, value, join=join)
         out.attrs = self.attrs
         return out
 
+    def combine_first(self, other):
```
shoyer: I'm not sure it's worth making a separate method here to save a few characters.

xarray/core/dataarray.py
```diff
@@ -1075,7 +1075,7 @@ def dropna(self, dim, how='any', thresh=None):
         ds = self._to_temp_dataset().dropna(dim, how=how, thresh=thresh)
         return self._from_temp_dataset(ds)
 
-    def fillna(self, value):
+    def fillna(self, value, join="left"):
```
shoyer: add `join` to the docstring

xarray/core/dataset.py
```diff
@@ -1955,7 +1955,35 @@ def fillna(self, value):
         -------
         Dataset
         """
-        out = self._fillna(value)
+        if utils.is_dict_like(value):
+            value_keys = value.data_vars.keys() if isinstance(value, Dataset)\
+                else value.keys()
+            if not set(value_keys) <= set(self.data_vars.keys()):
```
shoyer: I think we will need to separate the `join` argument to `apply_ufunc` into two parts:

1. The `join` used for [aligning indexes](https://github.com/pydata/xarray/blob/4d36fe2e8a1f5504d776c2d69895bc4d5a9cd31a/xarray/core/computation.py#L340).
2. The `join` used for [aligning data variables between datasets](https://github.com/pydata/xarray/blob/4d36fe2e8a1f5504d776c2d69895bc4d5a9cd31a/xarray/core/computation.py#L347).

The later should probably be renamed to something like `data_vars_join`

xarray/core/dataset.py
```diff
@@ -1955,7 +1955,35 @@ def fillna(self, value):
...
+                raise ValueError('all variables in the argument to `fillna` '
+                                 'must be contained in the original dataset')
+        out = ops.fillna(self, value, join="left")
+        out._copy_attrs_from(self)
+        return out
```
shoyer: This would be a good time to remove all the `fillna` specific logic from `Dataset._calculate_binary_op` that is no longer used

xarray/core/dataset.py
```diff
@@ -1955,7 +1955,35 @@ def fillna(self, value):
         -------
         Dataset
         """
-        out = self._fillna(value)
+        if utils.is_dict_like(value):
+            value_keys = value.data_vars.keys() if isinstance(value, Dataset)\
```
shoyer: maybe slightly cleaner to use duck typing, e.g., `value_keys = getattr(value, 'data_vars', value).keys()`

xarray/core/dataset.py
```diff
@@ -1955,7 +1955,35 @@ def fillna(self, value):
...
+                else value.keys()
+            if not set(value_keys) <= set(self.data_vars.keys()):
+                raise ValueError('all variables in the argument to `fillna` '
+                                 'must be contained in the original dataset')
+        out = ops.fillna(self, value, join="left")
```
shoyer: `join` should be a parameter for `Dataset.fillna`, too.

xarray/core/ops.py
shoyer: It might be helpful to separate the `join` argument to `apply_ufunc` into two parts:

1. The `join` used for [aligning indexes](https://github.com/pydata/xarray/blob/4d36fe2e8a1f5504d776c2d69895bc4d5a9cd31a/xarray/core/computation.py#L340).
2. The `join` used for [aligning data variables between datasets](https://github.com/pydata/xarray/blob/4d36fe2e8a1f5504d776c2d69895bc4d5a9cd31a/xarray/core/computation.py#L347).

For `fillna`, I think we would always want `data_vars_join='left'`, though I guess it's equivalent to `data_vars_join='left'` when we constrain `value` not to contain new keys.

doc/whats-new.rst
```diff
@@ -106,6 +106,13 @@ Deprecations
 
 Enhancements
 ~~~~~~~~~~~~
+
+- Added the xarray equivalent of `pandas.Dataframe.combine_first` as an instance
```
shoyer: just a note -- this will need to go in a new section for 0.9.1, since 0.9.0 will probably be released first.

doc/whats-new.rst
```diff
@@ -106,6 +106,13 @@ Deprecations
...
+  method to DataArray/Dataset objects, facilitated by the new `ops.fillna` that
+  uses `apply_ufunc` for different `join` options.
```
shoyer: I would drop implementation details like the fact that this uses `apply_ufunc` from the user facing release notes.

xarray/core/dataarray.py
chunweiyuan: I think either we have `combine_first` for both Dataset and DataArray, or we lump everything into `fillna`.  The use of `combine_first` is just that there's this concrete direct analogy from `pandas`.  What's the decision here?

xarray/core/groupby.py
chunweiyuan: So, if I replace the whole method with just `return apply_ufunc(ops.fillna, self, value)`, I'd fail line 2507 of `test_dataset.py`, where `self.assertEqual(actual.attrs, ds.attrs)` fails, because the `attrs` aren't copied over.

chunweiyuan:
So I addressed some of the comments, but not all.  Please let me know what you think.

xarray/core/groupby.py
chunweiyuan: One solution is perhaps to add something into `apply_groupby_ufunc` to make sure the `attrs` are copied over, similar to what's in `_copy_attrs_from` in dataset.py/dataarray.py.

xarray/core/ops.py
```diff
@@ -279,11 +279,34 @@ def count(data, axis=None):
...
+        - 'inner': take only overlapped variables
+        - 'left': take only variables from the first object
+        - 'right': take only variables from the last object
     """
-    return where(isnull(data), other, data)
+    from .computation import apply_ufunc
+
+    def _fillna(data, other):
+        left, right = np.broadcast_arrays(data, other)
+        return np.where(isnull(left), right, left)
```
shoyer: This should literally be:
```
def _fillna(data, other):
     return where(isnull(data), other, data)
```
The `where` and `isnull` functions defined in this module already do broadcasting, and work for both dask and numpy arrays.

xarray/core/computation.py
```diff
@@ -343,8 +344,10 @@ def apply_dataset_ufunc(func, *args, **kwargs):
     list_of_coords = build_output_coords(args, signature, exclude_dims)
 
     args = [getattr(arg, 'data_vars', arg) for arg in args]
+
     result_vars = apply_dict_of_variables_ufunc(
-        func, *args, signature=signature, join=join, fill_value=fill_value)
+        func, *args, signature=signature, join=data_vars_join,
```
shoyer: It could be nice to have a unit test that explicitly covers `data_vars_join` in `test_computation.py`.

xarray/core/computation.py
```diff
@@ -702,7 +713,8 @@ def stack(objects, dim, new_coord):
     elif any(is_dict_like(a) for a in args):
         return apply_dataset_ufunc(variables_ufunc, *args, signature=signature,
                                    join=join, exclude_dims=exclude_dims,
-                                   fill_value=dataset_fill_value)
+                                   fill_value=dataset_fill_value,
+                                   data_vars_join=data_vars_join)
```
shoyer: Add this argument to the `functools.partial` for `GroupBy` objects in the clause above this one.

xarray/core/groupby.py
shoyer: > One solution is perhaps to add something into apply_groupby_ufunc to make sure the attrs are copied over, similar to what's in _copy_attrs_from in dataset.py/dataarray.py.

I like this -- it would be useful for other functions in `apply_ufunc`, too. For consistency, I would call the new argument `keep_attrs`.

Or if you don't want to bother with that, could just use `_copy_attrs_from` in this method:
```
out = ops.fillna(self, other)
out._copy_attrs_from(self._obj)
return out
```

xarray/core/dataarray.py
shoyer: The fact that `data_vars_join` could sensibly differ for `combine_first` (as you have implemented it) suggests that a second method is reasonable. Consistency with pandas is also a nice factor. So I guess I've come around!

xarray/test/test_dataset.py
```diff
@@ -3186,6 +3186,23 @@ def test_full_like(self):
         self.assertEqual(expect['d2'].dtype, bool)
         self.assertDatasetIdentical(expect, actual)
 
+    def test_combine_first(self):  # works just like xr.merge([self, other])
```
shoyer: I would move this comment below, to just above the code block where you compare to the result of using `merge`

xarray/test/test_dataset.py
```diff
@@ -3186,6 +3186,23 @@ def test_full_like(self):
...
+        dsx1 = DataArray([1, 1], [('x', ['b', 'c'])]).to_dataset(name='dsx1')
+
+        actual = dsx0.combine_first(dsx1)
+        expected = Dataset({'dsx0': ('x', [0, 0, np.nan]),
+                            'dsx1': ('x', [np.nan, 1, 1])},
+                           coords={'x': ['a', 'b', 'c']})
+        self.assertDatasetEqual(actual, expected)
+        self.assertDatasetEqual(actual, xr.merge([dsx0, dsx1]))
+
+        dsy2 = DataArray([2, 2, 2], [('x', ['b', 'c', 'd'])]).\
```
shoyer: Per PEP8, please avoid writing `\`. Either split this on two lines or use parentheses for the implicit line continuation.

xarray/test/test_dataarray.py
```diff
@@ -2370,6 +2370,27 @@ def test_binary_op_join_setting(self):
...
+        actual = ar1.combine_first(ar0)
+        expected = DataArray([[0, 0, np.nan], [0, 1, 1], [np.nan, 1, 1]],
+                             [('x', ['a', 'b', 'c']), ('y', [-1, 0, 1])])
+        self.assertDataArrayEqual(actual, expected)
+
+        actual = ar0.combine_first(ar2)
+        expected = DataArray([[0, 0], [0, 0], [2, 2]],
+                             [('x', ['a', 'b', 'd']), ('y', [-1, 0])])
+        self.assertDataArrayEqual(actual, expected)
+
```
shoyer: nit: two line breaks should separate code blocks, not three

doc/combining.rst
```diff
@@ -116,6 +117,38 @@ used in the :py:class:`~xarray.Dataset` constructor:
...
+.. ipython:: python
+
+    ar0 = DataArray([[0, 0], [0, 0]], [('x', ['a', 'b']), ('y', [-1, 0])])
+    ar1 = DataArray([[1, 1], [1, 1]], [('x', ['b', 'c']), ('y', [0, 1])])
+    ar2 = DataArray([2], [('x', ['d'])])
+    ar0.combine_first(ar1)
+    ar1.combine_first(ar0)
+    ar0.combine_first(ar2)
+
+For datasets, ``ds0.combine_first(ds1)`` works just like ``xr.merge([ds0, ds1])``
```
shoyer: Again, this isn't quite true -- it succeeds in cases where `merge` will raise an error. It would be nice to highlight the difference here.

xarray/core/dataset.py
```diff
@@ -2324,11 +2356,6 @@ def _calculate_binary_op(self, f, other, join='inner',
                              inplace=False, fillna=False):
 
         def apply_over_both(lhs_data_vars, rhs_data_vars, lhs_vars, rhs_vars):
-            if fillna and join != 'left':
```
shoyer: you can remove the `fillna` argument from the function signature entirely (and also from `Dataset._binary_op`)

doc/combining.rst
```diff
@@ -13,6 +13,7 @@ Combining data
 
 * For combining datasets or data arrays along a dimension, see concatenate_.
 * For combining datasets with different variables, see merge_.
+# For combining datasets or data arrays with outer-join alignment, see combine_.
```
shoyer: Maybe: `For combining datasets with different coordinates or missing values`

xarray/core/groupby.py
chunweiyuan: I copped out and took your simple approach.  It's mainly because I'm not sure how to deal with more than 2 objects in `apply_ufunc` when `keep_attrs=True`.  Do I only take the first object's attributes?  Maybe best left for another PR.

xarray/core/computation.py
chunweiyuan: What is the expected behavior of the following?
```
ds0 = xr.Dataset({'a': ('x', [1, 2]), 'x': [0, 1]})
ds1 = xr.Dataset({'b': ('x', [99, 3]), 'x': [1, 2]})

def add(a, b, join, data_vars_join, **kwargs):
        return apply_ufunc(operator.add, a, b, join=join,
                           data_vars_join=data_vars_join, **kwargs)

out = add(ds0, ds1, 'outer', 'outer')
```
Currently, it raises a TypeError at `result_vars[name] = func(*variable_args)` of `apply_dict_of_variables_ufunc`.  

On the other hand, if you run
```
out = add(ds0, ds2, 'outer', 'outer', dataset_fill_value=np.nan)
```
you get
```
>> out
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 0 1 2
Data variables:
    a        (x) float64 nan nan nan
    b        (x) float64 nan nan nan
```
While I can understand why the code renders such result, as a user, I was hoping for something more like
```
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 0 1 2
Data variables:
    a        (x) float64 1.0 2.0 nan
    b        (x) float64 nan 99.0 3.0
```
which would be a little more useful.

would we want to make changes for this?

shoyer:
I would simply copy attributes from the first argument for keep_attrs. We
don't do any complex merging of attributes on any xarray methods.

xarray/core/computation.py
shoyer: You can make that work with `dataset_fill_value=0`:
```
In [11]: add(ds0, ds1, 'outer', 'outer', dataset_fill_value=0)
Out[11]:
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 0 1 2
Data variables:
    a        (x) float64 1.0 2.0 nan
    b        (x) float64 nan 99.0 3.0
```

Are you suggesting changing the default value to `0`? It's not obvious to me what is the best choice for the default `dataset_fill_value` is. Maybe we should remove the default value of `None` and raise a more descriptive error if this happens.

chunweiyuan: A couple of things:

1.) What do you think about removing the `join` kwarg in the `fillna` signature in both `dataarray.py` and `dataset.py`.  The writing of `ds0.fillna(ds1)` already implies a directionality, so adding a `join` option can lead to confusion.

2.) I've decided not to touch the default value of `dataset_fill_value=None`.  It leads to errors in many other places once you change that.   I'll look into raising some descriptive errors now that we have a `data_vars_join` option.  But if we want to make
```
out = add(ds0, ds2, 'outer', 'outer', dataset_fill_value=np.nan)
```
render
```
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 0 1 2
Data variables:
    a        (x) float64 1.0 2.0 nan
    b        (x) float64 nan 99.0 3.0
```
I think we'll have to do another PR.

shoyer: > 1.) What do you think about removing the join kwarg in the fillna signature in both dataarray.py and dataset.py. The writing of ds0.fillna(ds1) already implies a directionality, so adding a join option can lead to confusion.

Agreed, there isn't really a need for the `join` argument once we have `combine_first`.

For your example `add(ds0, ds2, 'outer', 'outer', dataset_fill_value=np.nan)`, can you clarify what the values of `ds0` and `ds2` are?

chunweiyuan: Sorry, my mistake.  That `ds2` is supposed to be `ds1`, as the one above:
```
ds0 = xr.Dataset({'a': ('x', [1, 2]), 'x': [0, 1]})
ds1 = xr.Dataset({'b': ('x', [99, 3]), 'x': [1, 2]})
```

chunweiyuan:
I'm not sure why continuous integration check is taking forever.....

shoyer:
Appveyor is acting up. I tried cancelling and restarting the build -- we'll see if that works. But the Travis-CI tests pass, so I wouldn't worry about.

shoyer:
Looks like the build is starting now.

xarray/core/computation.py
```diff
@@ -664,8 +676,10 @@ def stack(objects, dim, new_coord):
 
     signature = kwargs.pop('signature', None)
     join = kwargs.pop('join', 'inner')
+    data_vars_join = kwargs.pop('data_vars_join', 'inner')
+    keep_attrs = kwargs.pop('keep_attrs', True)
```
shoyer: This should default to `False`, as it does for arithmetic.

xarray/core/computation.py
```diff
@@ -697,16 +711,26 @@ def stack(objects, dim, new_coord):
...
     elif any(is_dict_like(a) for a in args):
-        return apply_dataset_ufunc(variables_ufunc, *args, signature=signature,
-                                   join=join, exclude_dims=exclude_dims,
-                                   fill_value=dataset_fill_value)
+        first_obj = args[0]
+        out = apply_dataset_ufunc(variables_ufunc, *args, signature=signature,
+                                  join=join, exclude_dims=exclude_dims,
+                                  fill_value=dataset_fill_value,
+                                  data_vars_join=data_vars_join)
+        if keep_attrs and (type(first_obj) is type(out)):
```
shoyer: I would move `keep_attrs` one level up into `apply_dataset_ufunc` and `apply_dataarray_ufunc`.

xarray/core/computation.py
```diff
@@ -664,8 +676,10 @@ def stack(objects, dim, new_coord):
 
     signature = kwargs.pop('signature', None)
     join = kwargs.pop('join', 'inner')
+    data_vars_join = kwargs.pop('data_vars_join', 'inner')
+    keep_attrs = kwargs.pop('keep_attrs', True)
     exclude_dims = kwargs.pop('exclude_dims', frozenset())
-    dataset_fill_value = kwargs.pop('dataset_fill_value', None)
+    dataset_fill_value = kwargs.pop('dataset_fill_value', np.nan)
```
shoyer: I would actually rather force `dataset_fill_value` to be set explicitly.

One option is to create a top level constant for the default:
```
_DEFAULT_FILL_VALUE = object()

...
dataset_fill_value = kwargs.pop('dataset_fill_value', _DEFAULT_FILL_VALUE)
```

Then, in `collect_dict_values` raise an error if you need to use the fill value and `fill_value is _DEFAULT_FILL_VALUE`.

chunweiyuan:
time to merge this guy?

chunweiyuan:
checking in on this PR...

doc/combining.rst
```diff
@@ -13,6 +13,8 @@ Combining data
 
 * For combining datasets or data arrays along a dimension, see concatenate_.
 * For combining datasets with different variables, see merge_.
+# For combining datasets with different coordinates or missing values,
```
shoyer: use `*` at the start for consistency

Maybe say "datasets or data arrays"?

Also, maybe "different indexes" instead of "different coordinates"?

doc/combining.rst
```diff
@@ -13,6 +13,8 @@ Combining data
...
+  or data arrays with outer-join alignment, see combine_.
```
shoyer: I would drop "or data arrays with outer-join alignment" -- it's a little confusing.

doc/combining.rst
```diff
@@ -13,8 +13,7 @@ Combining data
 
 * For combining datasets or data arrays along a dimension, see concatenate_.
 * For combining datasets with different variables, see merge_.
-# For combining datasets with different coordinates or missing values,
-  or data arrays with outer-join alignment, see combine_.
+* For combining datasets or data arrays with different indexes or missing values.
```
shoyer: you lost "see combine" here

xarray/core/computation.py
```diff
@@ -346,11 +342,17 @@ def apply_dataset_ufunc(func, *args, **kwargs):
...
-    data_vars_join = kwargs.pop('data_vars_join', 'inner')
+    dataset_join = kwargs.pop('dataset_join', 'inner')
     fill_value = kwargs.pop('fill_value', None)
     exclude_dims = kwargs.pop('exclude_dims', _DEFAULT_FROZEN_SET)
     keep_attrs = kwargs.pop('keep_attrs', False)
     first_obj = args[0]  # we'll copy attrs from this in case keep_attrs=True
+
+    if dataset_join != 'inner' and fill_value is _DEFAULT_FILL_VALUE:
+        raise TypeError('To apply an operation to datasets with different ',
+                        'data variables, you must supply the ',
```
shoyer: add `if dataset_join != 'inner'` to the docstring

xarray/core/computation.py
```diff
@@ -607,23 +609,23 @@ def apply_ufunc(func, *args, **kwargs):
...
         Method for joining variables of Dataset objects with mismatched
         data variables.
         - 'outer': take variables from both Dataset objects
         - 'inner': take only overlapped variables
         - 'left': take only variables from the first object
         - 'right': take only variables from the last object
+    dataset_fill_value : optional
+        Value used in place of missing variables on Dataset inputs when the
+        datasets do not share the exact same ``data_vars``. Only relevant if
+        ``dataset_join != 'inner'``.
```
shoyer: Maybe switch the last sentence to: 

   Required if dataset_join != 'inner', otherwise ignored.

xarray/core/computation.py
```diff
@@ -607,23 +609,23 @@ def apply_ufunc(func, *args, **kwargs):
         - 'inner': use the intersection of object indexes
         - 'left': use indexes from the first object with each dimension
         - 'right': use indexes from the last object with each dimension
-    data_vars_join : {'outer', 'inner', 'left', 'right'}, optional
+    dataset_join : {'outer', 'inner', 'left', 'right'}, optional
```
shoyer: Can you fix the function signature at the top of the docstring, also to add in the new arguments?

doc/whats-new.rst
```diff
@@ -13,6 +13,19 @@ What's New
...
     np.random.seed(123456)
 
+.. _whats-new.0.9.1:
+
+v0.9.1 (unreleased)
+
+Enhancements
+~~~~~~~~~~~~
+
+- Added the xarray equivalent of `pandas.Dataframe.combine_first` as an instance
```
shoyer: Actually let's move this into v0.9.0. I think this change is pretty safe, and it would be nice to get it in promptly.

shoyer:
Thanks for persevering on this, @chunweiyuan !

chunweiyuan:
tears in my eyes.......

Issue #1201, opened 2017-01-12
pass projection argument to plt.subplot when faceting with cartopy transform


vnoel:
I have a `data` 3D DataArray with `Time`, `Latitude` and `Longitude` coordinates.

I want to plot maps of this dataset, faceted by Time. The following code

```
import cartopy.crs as ccrs
proj = ccrs.PlateCarree()
data.plot(transform=proj, col='Time', col_wrap=3, robust=True)
```

fails with

``` 
ValueError: Axes should be an instance of GeoAxes, got <class 'matplotlib.axes._subplots.AxesSubplot'> 
```

this is because to plot with a transform, the axes must be a GeoAxes, which is done with something like `plt.subplot(111, projection=proj)`. The implicit subplotting done when faceting does not do that. To make the faceting works, I had to do

```
import cartopy.crs as ccrs
proj = ccrs.PlateCarree()
data.plot(transform=proj, col='Time', col_wrap=3, robust=True, subplot_kws={'projection':proj})
```

I propose that, when plot faceting is requested with a `transform` kw, the content of that keyword should be passed to the subplot function as a `projection` argument automatically by default. If a projection is provided explicitely like in the call above, use that one.

shoyer:
This seems totally reasonable to me, and would be nice because it would improve the consistency between facetted and non-facetted plotting.

Going even further, we might even default to `projection=ccrs.PlateCarree()` if a cartopy transform is provided (see https://github.com/SciTools/cartopy/issues/645).

@fmaussion @jhamman any opinions?

fmaussion:
> This seems totally reasonable to me, and would be nice because it would improve the consistency between facetted and non-facetted plotting.

I don't really understand what you mean here. Neither facetted nor non-facetted plots have any logic to greate geoaxes automatically.

Otherwise, yes, I think that any logic making working with cartopy a little easier would be a good addition. The logic creating the axes is located [here](https://github.com/pydata/xarray/blob/master/xarray/plot/utils.py#L231)

shoyer:
> I don't really understand what you mean here. Neither facetted nor non-facetted plots have any logic to greate geoaxes automatically.

Oops, I was indeed mixed up.

If we add a `projection` argument for faceted plots, we should certainly add it for non-faceted plots, too.

vnoel:
I am confused. 

I planned to identify the plotting function that accepted the `transform` keyword (used in the code sample above), and add its value in `subplot_kws` if it wasn't already.

[...]

PR #1205, opened 2017-01-13
transfer projection to implied subplots when faceting

vnoel:
this catches `transform` kw passed to `plot()` when faceting, and pass the associated projection to the subplots. This does what was suggested in issue #1201.

(sorry for the irrelevant change to reshaping.rst, it seems I've stuck myself in a git hole and can't get out)


fmaussion:
Thanks! I think we need to document this somewhere, and also some tests would be nice.

Do you feel like trying to add the same functionality to normal 2d plots also? (this will represent a little more work so if you don't have time no big deal).

Also, you should credit yourself in the "what's new" section of ``whats-new.rst``

xarray/plot/facetgrid.py
```diff
@@ -73,7 +73,7 @@ class FacetGrid(object):
 
     def __init__(self, data, col=None, row=None, col_wrap=None,
                  sharex=True, sharey=True, figsize=None, aspect=1, size=3,
-                 subplot_kws=None):
+                 transform=None, subplot_kws=None):
         """
```
fmaussion: This new kwarg needs to be documented (see examples just below)

xarray/plot/facetgrid.py
```diff
@@ -142,6 +142,8 @@ def __init__(self, data, col=None, row=None, col_wrap=None,
 
         # Set the subplot kwargs
         subplot_kws = {} if subplot_kws is None else subplot_kws
+        if transform is not None and 'projection' not in subplot_kws:
+            subplot_kws['projection'] = transform
```
fmaussion: add: ``# with a transform, we assume the user wants to plot on a projection``

(shorten if this doesn't fit in the pep8 character limit)

vnoel:
One concern I have is that it is conceivable that one might pass a `transform` keyword to `plot()` intended to interact with another library that hijacks matplotlib (i.e. not cartopy). In that case, matplotlib will get an unintended `projection` argument. That is a corner case though.

fmaussion:
> In that case, matplotlib will get an unintended projection argument. That is a corner case though.

Yes, I was about to ask @jhamman and @shoyer about their opinion about this. I'm actually not sure if this really is a corner case...

vnoel:
As it is, the xarray code becomes tied up with cartopy. If cartopy somehow changes its syntax (e.g. by renaming the `transform` kw) this code won't work. 

Maybe there's another, more abstract way to do this?

fmaussion:
> As it is, the xarray code becomes tied up with cartopy. If cartopy somehow changes its syntax (e.g. by renaming the transform kw) this code won't work.

cartopy uses ``transform`` because this is an interface provided by matlotlib: http://matplotlib.org/users/transforms_tutorial.html . This is not likely to change anytime soon.

fmaussion:
I know at least one usage where the ``projection`` keyword is needed for something else than cartopy (for polar plots, see example [here](https://github.com/pydata/xarray/pull/1148#issuecomment-264455218) by @hoonhout), but I don't know how often do people need  a ``transform`` though.

In the end, I'm still in favor of this change:
- I suspect the large majority of the use cases are for cartopy
- it's still possible to override the this default behavior by setting the ``projection`` kwarg yourself

xarray/plot/facetgrid.py
jhamman: I think we should only make this assumption if the projection is a `cartopy.crs` object. That would be easy to test here.

jhamman:
@vnoel and @fmaussion - my apologies for being quite tardy in looking at this PR. I think this is a valuable change but I do have one comment on the implementation.

PR #1203, opened 2017-01-12
add info in doc on how to facet with cartopy

vnoel:
This change explains
- how to pass projection arguments to faceted subplots
- how to access the Axes of the created subplots (not relevant only to cartopy), relevant to issue #1202

PR #1206, opened 2017-01-13
RTD: fix facetted maps example

fmaussion:
follow-up to https://github.com/pydata/xarray/pull/1203


fmaussion:
This PR also re-adds the "fix-docs" branch which is useful when debugging RDT.

I removed this branch from the travis builds to reduce noise.

PR #1213, opened 2017-01-17
Add licenses/ and docs/ to release tar-ball

### LINKED ISSUE #1209: Consider adding the docs to the distributed tarball
Linux distributions such as Debian usually source upstream Python code from PyPI. In order to produce the accompanying documentation package, I would need the source of the documentation to be distributed in future uploaded tarballs. To achieve this, please consider adding a `recursive-include docs *` entry to the `MANIFEST.in` file of the project.

Thanks,
Ghis

shoyer:
Fixes #1209

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,6 @@ nosetests.xml
 # xarray specific
 doc/_build
 doc/generated
-doc/_static/*.png
 xarray/version.py
 
 .ipynb_checkpoints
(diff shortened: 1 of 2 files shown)


shoyer:
Adding docs to the source distribution increases size from ~300KB to 1.1MB, which seems totally reasonable and worthwhile.

PR #1219, opened 2017-01-19
ENH import ufuncs module in toplevel namespace

### LINKED ISSUE #1216: Why ufuncs module not included in top-level namespace
I.e. there is no `from . import ufuncs` in `__init__.py`.  Is this deliberate?

If it was included, one could access them after `import xarray as xr` via e.g. `xr.ufuncs.sin` rather than having to do separately e.g. `from xarray.ufuncs import sin`.

Sorry if I'm missing something obvious here.

spencerahill:
- [x] closes #1216 
 - [ ] tests added / passed
 - [ ] whatsnew entry

Let me know if you want tests and/or what's new; wasn't sure since this is so small.

### DIFF
--- a/xarray/__init__.py
+++ b/xarray/__init__.py
@@ -25,3 +25,4 @@ except ImportError:  # pragma: no cover
                       'then install it in-place by running: pip install -e .')
 
 from . import tutorial
+from . import ufuncs


shoyer:
This is great, thanks!

PR #1220, opened 2017-01-19
Fix docs

fmaussion:
This PR fixes a problem we had with ipython on RTD: https://github.com/ipython/ipython/issues/8733 

We this simple workaround we can now use the latest ipython version. Thanks to @takluyver for his help!

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -211,6 +211,12 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+# attempt to fix https://github.com/ipython/ipython/issues/8733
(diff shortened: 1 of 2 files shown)


takluyver:
I'm still a bit puzzled why this is necessary. I couldn't reproduce it when building the docs locally.

takluyver:
I just made ipython/ipython#10186 - if you've got time, could you try with that branch and see if it works without the config?

fmaussion:
@takluyver thanks, just tried it: seems to work fine! (logs: https://readthedocs.org/projects/xray/builds/4918881/)

fmaussion:
Merged this with the workaround for now, we can get rid of it when the next ipython version is out

shoyer:
Thanks @fmaussion!

PR #1221, opened 2017-01-20
List "Unindexed dimensions" separately in __repr__

### LINKED ISSUE #1199: Document the new __repr__
Sorry I missed that one when it was decided upon in https://github.com/pydata/xarray/pull/1017, but I think the changes in ``repr`` should be documented somewhere (at the minimum in the "Breaking Changes" section of what's new).

I just updated Salem for it to work well with xarray 0.9.0. The changes I had to make where quite small (that's a good thing), but it took me a bit of time to understand what was going on. 

What I found confusing is following:

```python
In [1]: import xarray as xr
In [2]: ds = xr.DataArray([1, 2, 3]).to_dataset(name='var')
In [3]: ds
Out[3]: 
<xarray.Dataset>
Dimensions:  (dim_0: 3)
Coordinates:
  o dim_0    (dim_0) -
Data variables:
    var      (dim_0) int64 1 2 3

In [4]: 'dim_0' in ds.coords
Out[4]: False
```
``dim_0``is listed as coordinate, but ``'dim_0' in ds.coords`` is ``False``. I think it should remain like this, but maybe we should document somewhere what the "o" and "*" mean? 

(possibly [here](http://xarray.pydata.org/en/latest/data-structures.html#creating-a-dataset))

shoyer:
- [x] fixes #1199
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master | flake8 --diff`` (except for lines which need to be long for displaying a repr)
 - [x] whatsnew entry

Previously (on master):

    <xarray.Dataset>
    Dimensions:  (x: 1, y: 2)
    Coordinates:
      o x        (x) -
      o y        (y) -
    Data variables:
        foo      (x, y) int64 1 2

Now:

    <xarray.Dataset>
    Dimensions:  (x: 1, y: 2)
    Coordinates:
        *empty*
    Unindexed dimensions:
        x, y
    Data variables:
        foo      (x, y) int64 1 2

This version of the `__repr__` should be much more self-explanatory.


xarray/test/test_dataset.py
```diff
@@ -685,6 +686,23 @@ def test_coords_merge_mismatched_shape(self):
         actual = orig_coords.merge(other_coords)
         self.assertDatasetIdentical(expected, actual)
 
+    def test_data_vars_properties(self):
```
shoyer: It looks like we were missing tests for `data_vars` previously, which I noticed when I renamed the formatting function and an error appeared in the docs build. Oops!

Issue #1218, opened 2017-01-19
Testing functions should be moved to a separate `xarray.testing` module


shoyer:
These functions are in the `test` module, which isn't imported by default because it requires `pytest`, which we only require for testing. Hence the API docs are currently broken:
http://xarray.pydata.org/en/latest/api.html#testing

If we move these to a separate `testing` module (i.e., a `testing.py` file), we can make `xarray.testing.assert_xarray_equal` just work after `import xarray` (maybe rename it to just `xarray.testing.assert_equal` to avoid the redundant `xarray`).

To avoid confusion, we might rename the `test` directory to `tests`.

cc @MaximilianR @fmaussion in case either of you want to take a look at this -- it should probably be a blocker for 0.9.0.

fmaussion:
yes, correct. The easiest fix would be to install pytest on RTD (assuming that all users using the xarray testing functions  also have pytest installed) but this is not ideal.

 @MaximilianR let me know if you have time to do this, otherwise I'll have a look later this week

MaximilianR:
+1 on `tests` over `test`.

However much I love pytest, minimizing dependencies is a good goal.

I think we could put the testing functions in `testing.py` but the `skipif` functions in `xr.tests.__init__.py`, and then not require pytest?

MaximilianR:
@fmaussion I'm unfortunately v busy for a while, if you could take this one that would be awesome...

shoyer:
> I think we could put the testing functions in testing.py but the skipif functions in xr.tests.__init__.py, and then not require pytest?

Yes, agreed!

PR #1222, opened 2017-01-20
New testing module and tests refactor

fmaussion:
- [x] closes https://github.com/pydata/xarray/issues/1218
 - [x] tests added / passed
 - [x] whatsnew entry


xarray/testing.py
```diff
@@ -0,0 +1,95 @@
...
+        arr1 = decode_string_data(arr1)
+        arr2 = decode_string_data(arr2)
+    exact_dtypes = ['M', 'm', 'O', 'U']
+    if any(arr.dtype.kind in exact_dtypes for arr in [arr1, arr2]):
+        return ops.array_equiv(arr1, arr2)
+    else:
+        return ops.allclose_or_equiv(arr1, arr2, rtol=rtol, atol=atol)
+
+
+def assert_dataset_allclose(d1, d2, rtol=1e-05, atol=1e-08):
```
shoyer: Can we delete this method instead? I think `assert_allclose` covers its functionality.

xarray/testing.py
```diff
@@ -0,0 +1,95 @@
...
+    assert type(a) == type(b)
+    if isinstance(a, (xr.Variable, xr.DataArray, xr.Dataset)):
+        assert a.equals(b), '{}\n{}'.format(a, b)
+    else:
+        raise TypeError('{} not supported by assertion comparison'
+                        .format(type(a)))
+
+
+def assert_identical(a, b):
+    """Like assert_equal, but also checks all attributes."""
```
shoyer: Let's say "names and attributes" instead of "all attributes"

xarray/testing.py
```diff
@@ -0,0 +1,95 @@
...
+        assert_allclose(a.variable, b.variable)
+        for v in a.coords.variables:
+            # can't recurse with this function as coord is sometimes a
+            # DataArray, so call into data_allclose_or_equiv directly
+            allclose = data_allclose_or_equiv(
+                a.coords[v].values, b.coords[v].values, rtol=rtol, atol=atol)
+            assert allclose, '{}\n{}'.format(a.coords[v].values,
+                                             b.coords[v].values)
+    elif isinstance(a, xr.Dataset):
+        assert sorted(a, key=str) == sorted(b, key=str)
```
shoyer: Maybe change this to `assert set(a) == set(b)`, to avoid casting keys to strings? This probably exists for the sake of some unit test that should be fixed, maybe involving serialization to netCDF.

Also let's check `assert set(a.coords) == set(b.coords)` to verify the coordinate names match? It looks like we missed that check from `assert_dataset_allclose`.

xarray/testing.py
```diff
@@ -0,0 +1,95 @@
...
+    ___tracebackhide__ = True  # noqa: F841
+    assert type(a) == type(b)
+    if isinstance(a, xr.Variable):
+        assert a.dims == b.dims
+        allclose = data_allclose_or_equiv(
+            a.values, b.values, rtol=rtol, atol=atol)
+        assert allclose, '{}\n{}'.format(a.values, b.values)
+    elif isinstance(a, xr.DataArray):
+        assert_allclose(a.variable, b.variable)
+        for v in a.coords.variables:
```
shoyer: I think we also need to verify `assert set(a.coords) == set(b.coords)`.

xarray/testing.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def decode_string_data(data):
+    if data.dtype.kind == 'S':
+        return np.core.defchararray.decode(data, 'utf-8', 'replace')
+    return data
+
+
+def data_allclose_or_equiv(arr1, arr2, rtol=1e-05, atol=1e-08):
+    if any(arr.dtype.kind == 'S' for arr in [arr1, arr2]):
```
shoyer: If this is going to be public API, maybe we should hide this string decoding behind an option, e.g., `decode_bytes`? We could even still default to `decode_bytes=True`, that would at least signal that something is going on.

fmaussion:
I've implemented the changes you suggested, this is ready for another review.

I 've had problems locally with bottleneck (three tests didn't pass with bottleneck installed), but I haven't been able to reproduce this on travis so I guess it's just me.

this is ready for review/merge

xarray/testing.py
```diff
@@ -0,0 +1,84 @@
...
+        assert a.name == b.name
+        assert_identical(a._to_temp_dataset(), b._to_temp_dataset())
+    elif isinstance(a, (xr.Dataset, xr.Variable)):
+        assert a.identical(b), '{}\n{}'.format(a, b)
+    else:
+        raise TypeError('{} not supported by assertion comparison'
+                        .format(type(a)))
+
+
+def assert_allclose(a, b, rtol=1e-05, atol=1e-08):
```
shoyer: add `decode_bytes` here, too?

xarray/testing.py
```diff
@@ -0,0 +1,84 @@
...
+        assert a.dims == b.dims
+        allclose = data_allclose_or_equiv(
+            a.values, b.values, rtol=rtol, atol=atol)
+        assert allclose, '{}\n{}'.format(a.values, b.values)
+    elif isinstance(a, xr.DataArray):
+        assert_allclose(a.variable, b.variable)
+        for v in a.coords.variables:
+            # can't recurse with this function as coord is sometimes a
+            # DataArray, so call into data_allclose_or_equiv directly
+            assert set(a.coords) == set(b.coords)
```
shoyer: This should be outside the `for` loop.

fmaussion:
done

MaximilianR:
What do people think about the names of the functions? Numpy includes array as in `assert_array_equal`

xarray/testing.py
```diff
@@ -53,24 +53,27 @@ def assert_identical(a, b):
...
-            assert set(a.coords) == set(b.coords)
-            allclose = data_allclose_or_equiv(
-                a.coords[v].values, b.coords[v].values, rtol=rtol, atol=atol)
+            allclose = data_allclose_or_equiv(a.coords[v].values,
+                                              b.coords[v].values,
+                                              rtol=rtol, atol=atol,
+                                              decode_bytes=decode_bytes)
             assert allclose, '{}\n{}'.format(a.coords[v].values,
                                              b.coords[v].values)
     elif isinstance(a, xr.Dataset):
```
shoyer: pass `decode_bytes` on in this case, too.

shoyer:
> What do people think about the names of the functions? Numpy includes array as in `assert_array_equal`

I would lean towards `assert_equal`. Maybe `assert_data_equal` or `assert_object_equal`? I guess even `assert_xarray_equal` could be OK, though the name feels redundant to me with the namespace. I would prefer `assert_data_equal`, except that could be confused to only apply to the `.data` attribute.

shoyer:
Also, it would be a good idea to import `testing` in `xarray/__init__.py` so `xr.testing.assert_equal` works without additional imports (e.g., see #1216)

fmaussion:
I've implemented your comment and made an effort for better docstring. Let me know what you think.

As for the names, we currently have three new API-level functions. What about following names?
- assert_data_equal (this makes it clear that we look at the data and not the attributes)
- assert_identical (this makes it clear that the objects must be identical)
- assert_allclose 

I'm fine with ``assert_object_identical`` or  ``assert_xarray_identical`` if you wish.

shoyer:
> As for the names, we currently have three new API-level functions. What about following names?

My strongest preference is that all names have the same form, e.g., `assert_X_equal`, `assert_X_identical` and `assert_X_allclose` where `X` is the same token.

The problem with `assert_data_equal` is that it looks at both data and coords/dims, just not `.data`.

fmaussion:
> My strongest preference is that all names have the same form, e.g., assert_X_equal, assert_X_identical and assert_X_allclose where X is the same token.

Agreed. Currently X is nothing, but I'm happy with either option you suggest. Also, my docstrings might need a short review as well to make sure I got them right.

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+    else:
+        return ops.allclose_or_equiv(arr1, arr2, rtol=rtol, atol=atol)
+
+
+def assert_equal(a, b):
+    """Like :py:func:`numpy.testing.assert_array_equal`, but for xarray
+    objects.
+
+    Raises an AssertionError if two objects are not equal. This will match
+    data values and coordinates, but not names or attributes (except for
```
shoyer: add dimensions: "data values, dimensions and coordinates"

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+    a : xarray.Dataset or xarray.DataArray
+        The first object to compare
+    b : xarray.Dataset or xarray.DataArray
+        The second object to compare
+    rtol : float, optional
+        Relative tolerance
+    atol : float, optional
+        Absolute tolerance
+    decode_bytes : bool, optional
+        Whether byte types should be decoded to a string or not
```
shoyer: Maybe:

    Whether byte dtypes should be decoded to strings as UTF-8 or not.
    This is useful for testing serialization methods on Python 3 that return
    saved strings as bytes.

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+
+def assert_identical(a, b):
+    """Like :py:func:`xarray.testing.assert_equal`, but also matches the
+    objects' names and attributes.
+
+    Raises an AssertionError if two objects are not identical.
+
+    Parameters
+    ----------
+    a : xarray.Dataset or xarray.DataArray
```
shoyer: For all these docstrings: `xarray.Dataset, xarray.DataArray or xarray.Variable`

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+        The first object to compare
```
shoyer: End each parameter description with a period.

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+        return ops.allclose_or_equiv(arr1, arr2, rtol=rtol, atol=atol)
+
+
+def assert_equal(a, b):
+    """Like :py:func:`numpy.testing.assert_array_equal`, but for xarray
+    objects.
+
+    Raises an AssertionError if two objects are not equal. This will match
+    data values and coordinates, but not names or attributes (except for
+    Dataset objects for which the variable names must match).
```
shoyer: I would add: Arrays with NaN in the same location are considered equal.

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+    Parameters
+    ----------
+    a : xarray.Dataset or xarray.DataArray
+        The first object to compare
+    b : xarray.Dataset or xarray.DataArray
+        The second object to compare
+
+    See also
+    --------
+    assert_identical, assert_allclose, xarray.Dataset.equals,
```
shoyer: I think "See also" needs items separated by new lines, not commas. But I haven't checked what sphinx actually does.

Also: I would add `numpy.testing.assert_array_equal` here.

xarray/testing.py
```diff
@@ -0,0 +1,143 @@
...
+    rtol : float, optional
+        Relative tolerance
+    atol : float, optional
+        Absolute tolerance
+    decode_bytes : bool, optional
+        Whether byte types should be decoded to a string or not
+
+    See also
+    --------
+    assert_identical, assert_equal
```
shoyer: add `numpy.testing.assert_allclose` here.

PR #1226, opened 2017-01-24
Mention that HoloViews supports xarray objects

shoyer:


philippjfr:
Thanks @shoyer! I'll be adding more documentation about xarray support for our next release.

PR #1236, opened 2017-01-30
Better repr for coordinates without dimensions

### LINKED ISSUE #1199: Document the new __repr__

shoyer:
- [x] closes #1199
 - [x] tests added / passed
 - [x] whatsnew entry

Old:

     <xarray.Dataset>
     Dimensions:  (x: 1, y: 2)
     Coordinates:
        *empty*
     Unindexed dimensions:
        x, y
     Data variables:
         foo      (x, y) int64 1 2

New:

     <xarray.Dataset>
     Dimensions:  (x: 1, y: 2)
     Data variables:
         foo      (x, y) int64 1 2

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -13,6 +13,16 @@ What's New
     import xarray as xr
     np.random.seed(123456)
 
+
+.. _whats-new.0.9.0:
+
+v0.9.1 (29 January 2017)
+------------------------
+
+Renamed the "Unindexed dimensions" section in the ``Dataset`` and
+``DataArray`` repr (added in v0.9.0) to "Dimensions without coordinates"
+(:issue:`1199`).
+
 .. _whats-new.0.9.0:
 
 v0.9.0 (25 January 2017)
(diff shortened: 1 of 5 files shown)


doc/whats-new.rst
```diff
@@ -13,6 +13,16 @@ What's New
     import xarray as xr
     np.random.seed(123456)
 
+
+.. _whats-new.0.9.0:
+
```
fmaussion: 0.9.1

shoyer: good catch, thanks

PR #1242, opened 2017-01-31
Cleanup

jhamman:
This is just a quick PR to clean up a few style violations that have crept in. I also removed a number of unused imports.

 - [ ] closes #xxxx
 - [x] tests passed
 - [ ] whatsnew entry


MaximilianR:
Thoughts on adding flake to Travis, at least on the diffs?

shoyer:
> Thoughts on adding flake to Travis, at least on the diffs?

Yes, this would be a nice idea! I think pandas has this setup now so we could probably just copy that.

jhamman:
I'd be open to running flake8 and building the docs on travis.

PR #1245, opened 2017-02-02
Replace 'signature' argument in apply_ufunc with *_core_dims

shoyer:
This results in a much easier to use function signature than the previous
`signature` argument which required triply nested lists.

e.g., new:

    def inner_product(a, b, dim):
        return apply_ufunc(_inner, a, b, input_core_dims=[[dim], [dim]])

vs. old:

    def inner_product(a, b, dim):
        return apply_ufunc(_inner, a, b, signature=[[[dim], [dim]], []])

I also was able to remove the code related to parsing gufunc signatures from strings.

 - [x] tests added / passed


rabernat:
I like the new syntax much better. I spent 45 minutes trying to figure out the correct ``signature``.

PR #1256, opened 2017-02-08
PERF: Override equals in IndexVariable

### LINKED ISSUE #1255: PeriodIndex causes severe slow down
I need some guidance on how to handle this.

### Background

`PeriodIndex` has a 'non-numpy' dtype now: 
```python
In [2]: i = pd.PeriodIndex(start=2000, freq='A', periods=10)

In [3]: i
Out[3]:
PeriodIndex(['2000', '2001', '2002', '2003', '2004', '2005', '2006', '2007',
             '2008', '2009'],
            dtype='period[A-DEC]', freq='A-DEC')

In [6]: i.dtype
Out[6]: period[A-DEC]
```

When `.values` or `.__array__()` are called, the Periods are boxed, which is really slow. The underlying ints are stored in `._values`:

```python
In [25]: i.values
Out[25]:
array([Period('2000', 'A-DEC'), Period('2001', 'A-DEC'),
       Period('2002', 'A-DEC'), Period('2003', 'A-DEC'),
       Period('2004', 'A-DEC'), Period('2005', 'A-DEC'),
       Period('2006', 'A-DEC'), Period('2007', 'A-DEC'),
       Period('2008', 'A-DEC'), Period('2009', 'A-DEC')], dtype=object)

In [27]: all(i.__array__()==i.values)
Out[27]: True

# underlying:
In [28]: i._values
Out[28]: array([30, 31, 32, 33, 34, 35, 36, 37, 38, 39])
```

### Problem

In pandas, we limit directly calling `.values` from outside Index, instead accessing Index functions through a smaller API.

But in xarray, I think there are a fair few functions that call `.values` or implicitly call `.__array__()` by passing the index into numpy. 

As a result, there is a severe slow down when using `PeriodIndex`. As an example:

```python
In [51]: indexes = [pd.PeriodIndex(start=str((1776 + i)), freq='A', periods=300) for i in range(50)]
In [53]: das = [xr.DataArray(range(300), coords=[index]) for index in indexes]

In [54]: %timeit xr.concat(das)

# 1 loop, best of 3: 1.38 s per loop
```

vs DTI:

```python
In [55]: indexes_dt = [pd.DatetimeIndex(start=str((1776 + i)), freq='A', periods=300) for i in range(50)]
In [56]: das_dt = [xr.DataArray(range(300), coords=[index]) for index in indexes_dt]
In [57]: %timeit xr.concat(das_dt)
# 10 loops, best of 3: 69.2 ms per loop
```

...a 20x slowdown, on fairly short indexes

@shoyer do you have any ideas of how to resolve this? Is it feasible to not pass Indexes directly into numpy? I haven't gone through in enough depth to have a view there, given I was hoping you could cut through the options. Thank you.

ref https://github.com/pandas-dev/pandas/issues/14822
CC @sinhkrs @jreback

MaximilianR:
- [x] closes #1255
 - [ ] tests added / passed
 - [ ] whatsnew entry
 
I tried adding some tests but couldn't find a good way of doing that

There is a 20x speed-up though (see issues for prior):

```python
In [4]: In [51]: indexes = [pd.PeriodIndex(start=str((1776 + i)), freq='A', periods=300) for i in range(50)]
   ...: In [53]: das = [xr.DataArray(range(300), coords=[index]) for index in indexes]
   ...:
   ...: In [54]: %timeit xr.concat(das)
   ...:


10 loops, best of 3: 70.2 ms per loop
```

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -1249,7 +1249,7 @@ class IndexVariable(Variable):
 
     def equals(self, other, equiv=None):
         # if equiv is specified, super up
-        if equiv:
+        if equiv is not None:
             return super(IndexVariable, self).equals(other, equiv)
 
         # otherwise use the native index equals, rather than looking at _data
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -1247,6 +1247,19 @@ class IndexVariable(Variable):
         return type(self)(self.dims, self._data, self._attrs,
                           self._encoding, fastpath=True)
 
+    def equals(self, other, equiv=None):
+        # if equiv is specified, super up
+        if equiv:
+            return super(IndexVariable, self).equals(other, equiv)
+
+        # otherwise use the native index equals, rather than looking at _data
+        other = getattr(other, 'variable', other)
+        try:
+            return (self.dims == other.dims and
+                    self._data_equals(other))
+        except (TypeError, AttributeError):
+            return False
+
     def _data_equals(self, other):
         return self.to_index().equals(other.to_index())
 


MaximilianR:
I don't think the AppVeyor is related to this

xarray/core/variable.py
```diff
@@ -1247,6 +1247,19 @@ def copy(self, deep=True):
         return type(self)(self.dims, self._data, self._attrs,
                           self._encoding, fastpath=True)
 
+    def equals(self, other, equiv=None):
+        # if equiv is specified, super up
+        if equiv:
```
shoyer: I'd prefer `equiv is not None`, which is slightly more idiomatic for checking default values (I'd rather not rely on the truthiness of arbitrary functions/callables)

PR #1260, opened 2017-02-10
Add RasterIO backend

fmaussion:
Follow-up to https://github.com/pydata/xarray/pull/1070

This is my first backend so this is going to be a bit more work than I expected, but with your help we should be able to get through this.

A long todo list:
 - [x] closes https://github.com/pydata/xarray/issues/790
 - [x] add tests 
 - [x] whatsnew entry
 - [x] wrap __getitem__ for lazy data read
 - [x] ~~make coordinates variables lazy too~~


fmaussion:
Before I'll get more into details with what needs to be done with rasterio itself, I'd like to get some xarray internals ready first. No need to do a full review yet, but I'd appreciate help with the following points:

1. The order of the variables and coordinates is random. The current ```__repr__``` can look like this:

```
<xarray.Dataset>
Dimensions:  (band: 1, x: 4, y: 3)
Coordinates:
  * x        (x) float64 1.0 1.5 2.0 2.5
  * band     (band) int64 1
  * y        (y) float64 2.0 1.0 0.0
Data variables:
    lat      (y, x) float64 2.0 2.0 2.0 2.0 1.0 1.0 1.0 1.0 0.0 0.0 0.0 0.0
    raster   (band, y, x) float32 0.0 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 ...
    lon      (y, x) float64 1.0 1.5 2.0 2.5 1.0 1.5 2.0 2.5 1.0 1.5 2.0 2.5
Attributes:
    crs: CRS({'wktext': True, 'proj': 'longlat', 'no_defs': True, 'ellps': 'WGS84'})
    transform: [1.0, 0.5, 0.0, 2.0, 0.0, -1.0]
```
How is this possible?

2. Is it possible to make the ``x`` and ``y`` coords lazy? Could you point me on a place in the code where this has been done already?

3. I'd like to have the ``lon`` and ``lat`` variables listed as optional coordinates, and also make them lazy. Any hint?

Thanks a lot for your help, I'm afraid this is going to need a few iterations ;-)

shoyer:
> The order of the variables and coordinates is random.

Most likely you're using a dict instead of an OrderedDict somewhere.

> Is it possible to make the x and y coords lazy?

See here for an example of lazily computing arrays:
https://github.com/pydata/xarray/blob/62333208fc2a80c05848a12de67a10f00a6610a1/xarray/conventions.py#L314

There are other examples in that file, for decoding CF conventions.

> I'd like to have the lon and lat variables listed as optional coordinates

Can you clarify what you mean by an optional coordinate?

fmaussion:
> Can you clarify what you mean by an optional coordinate?

Yes sorry, I meant the have them listed as coordinates without ``*`` instead of ``Data variables``.

fmaussion:
thanks @shoyer , I think I can work on this a bit further now and I'll get back to you if I have more questions.

shoyer:
Getting optional coordinates is hard is because the current "backends" system in xarray was really only designed for handling netCDF-like libraries.

For example, every backend gets passed through `xarray.conventions.decode_cf` to use CF conventions to decode its data. You would actually need to add a "global attribute" specifying `coords='lat lon'` to set coordinates variables with this system.

One alternative would be to simply write a function for reading files with rasterio straight into xarray Dataset objects.

There are probably hacks that can get this working with data stores but this really calls out for the bigger refactor like https://github.com/pydata/xarray/pull/1087.

fmaussion:
I made some progress with the lazy indexing, I'd be glad to have a first rough feedback on whether this is going in the right direction or not.

We have a decision to make regarding the API: I think that creating the optional lon and lat coords automatically is not a good idea: 
- in some cases, the x and y coordinates are already lons and lats and the 2D coords are obsolete
- for big data files this is going to take ages and take  a lot of memory
- my initial idea to make them lazily evaluated might work, but in an ugly way: computing _both_ lons and lats needs to be done in one single operation, and I'm not sure how this can be done in an elegant way
- additionally, there is no way to make them show up as coordinates (as per @shoyer 's comment: https://github.com/pydata/xarray/pull/1260#issuecomment-279101252)

The current implementation delegates this task to a utility function (``get_latlon_coords_from_crs``). It is currently very rasterio specific, but could be made more general -- API to be defined.

Thoughts?

shoyer:
> my initial idea to make them lazily evaluated might work, but in an ugly way: computing both lons and lats needs to be done in one single operation, and I'm not sure how this can be done in an elegant way

This could be done sanely with dask.array ([example of multiple outputs from an element-wise function](https://github.com/dask/dask/blob/1e2bf8973ff3706e68a5a55d6392a5798bd072bd/dask/array/ufunc.py#L116-L133)).

> additionally, there is no way to make them show up as coordinates (as per @shoyer 's comment: #1260 (comment))

We could make this work, it just looks pretty hacky.

fmaussion:
@shoyer thanks for the tips, I think that getting the lons and lats from dask is probably the most elegant method.

After trying various things I am still struggling with dask, and in particular on how to apply ``elemwise`` to functions like ``np.meshgrid`` (I get shape broadcasting errors). To get me started, I'd be grateful for an example on how to use dask to replace the code snippet below:

```python
import xarray as xr
import numpy as np

ds = xr.DataArray(np.zeros((2, 3)),
                  coords={'x': np.arange(3), 'y': np.arange(2)},
                  dims=['y', 'x']).to_dataset(name='data')

# non-dask version
lon, lat = np.meshgrid(ds.x, ds.y)
ds['lon'] = (('y', 'x'), lon)
ds['lat'] = (('y', 'x'), lat)
ds.set_coords(['lon', 'lat'], inplace=True)
print(ds)
```

Thanks a lot!

fmaussion:
Folks, I finally managed to find a couple of hours to wrap this up: this is now ready for review. 

Everything seems to work the way I'd like it to work, and only one thing is missing: the lazy computation of lons and lats with dask (I don't know how to do this quickly and I have not enough time to spend on this right now, unfortunately). This has been waiting for too long now, so I suggest to merge this when ready and this feature later on. 

The solution retained for the API is to add a new ``open_rasterio`` top-level function: this makes sense since many keywords of ``open_dataset`` aren't relevant for rasterio datasets. It also underlines that rasterio datasets are quit different from xarray's data model.

another example I could add to the soon to come xarray gallery could be:

```python
import xarray as xr
import cartopy.crs as ccrs
ds = xr.open_rasterio('/home/mowglie/RGB.byte.tif', add_latlon=True)
ax = plt.subplot(projection=ccrs.PlateCarree())
ds.raster.sel(band=1).plot(ax=ax, x='lon', y='lat', transform=ccrs.PlateCarree());
ax.coastlines('10m');
````

rasterio_map


cc @gidden @jhamman @shoyer

**Comment by shoyer** [comment]
I haven't looked at the implementation yet, but I agree that a separate function `open_rasterio` is the way to go. Structurally, this is pretty different from opening a netCDF file, which is also why I was questioning reusing the existing xarray backends system.

**Comment by benbovy** [comment]
Excellent! I'm looking forward to see this merged!

Just a small suggestion: to me `open_raster` seems a slightly better name than `open_rasterio` as the dataset is a 'raster', not a 'rasterio'.

**Comment by fmaussion** [comment]
> which is also why I was questioning reusing the existing xarray backends system.

I am starting to understand what you mean, but in the absence of template I guess this was the easiest way to go (I overtook the design of the original PR). If you agree I'd suggest you to have a more detailed look at the current PR when you have time, and we can decide what to do from here. Since the public facing API shouldn't be affected we could also keep the current design for now and go back to it later when https://github.com/pydata/xarray/pull/1087 is ready.

> Just a small suggestion: to me open_raster seems a slightly better name than open_rasterio as the dataset is a 'raster', not a 'rasterio'.

Yes I thought about it too, but a vast majority of the datasets xarray is reading are raster datasets (although in NetCDF format), hence "open_raster" could be confusing. "open_rasterio" underlines the fact that this opens "all datasets rasterio can open". I have no strong opinion about this though

**Comment by benbovy** [comment]
> a vast majority of the datasets xarray is reading are raster datasets (although in NetCDF format), hence "open_raster" could be confusing. "open_rasterio" underlines the fact that this opens "all datasets rasterio can open".

Yep makes sense!

File: xarray/backends/api.py [hunk 1]
```diff
@@ -318,6 +318,33 @@ def maybe_decode_store(store, lock=False):
...
+
+    Returns
+    -------
+    dataset : Dataset
+        The newly created dataset.
+    """
+    ds = conventions.decode_cf(backends.RasterioDataStore(filename))
+    if add_latlon:
+        from ..core.utils import add_latlon_coords_from_crs
+        ds = add_latlon_coords_from_crs(ds)
```
gidden: Should we detect an empty CRS here? Or are we happy with the current behavior (shown below)

```
In [49]: ds = xr.open_rasterio('ssp1_g.tiff', add_latlon=True)
CRSError                                  Traceback (most recent call last)
<ipython-input-49-6f687cb4abd7> in <module>()
----> 1 ds = xr.open_rasterio('ssp1_g.tiff', add_latlon=True)

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in open_rasterio(filename, add_latlon)
    342     if add_latlon:
    343         from ..core.utils import add_latlon_coords_from_crs
--> 344         ds = add_latlon_coords_from_crs(ds)
    345     return ds
    346 

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/core/utils.pyc in add_latlon_coords_from_crs(ds, crs)
    520     # Rasterio works with 1D arrays
    521     xc, yc = transform(crs, {'init': 'EPSG:4326'},
--> 522                        x.flatten(), y.flatten())
    523     xc = np.asarray(xc).reshape((ny, nx))
    524     yc = np.asarray(yc).reshape((ny, nx))

/home/gidden/.local/lib/python2.7/site-packages/rasterio/warp.pyc in transform(src_crs, dst_crs, xs, ys, zs)
     50     coordinate reference system.
     51     """
---> 52     return _transform(src_crs, dst_crs, xs, ys, zs)
     53 
     54 

rasterio/_base.pyx in rasterio._base._transform (rasterio/_base.c:19011)()

CRSError: No translation for an empty SRS to PROJ.4 format is known.
```

gidden: I'm not sure if the CRSError is as clear as we might like. This could be caught and rethrown with a slightly more informative message.

xarray/backends/api.py
```diff
@@ -318,6 +318,33 @@ def maybe_decode_store(store, lock=False):
     return maybe_decode_store(store)
 
 
+def open_rasterio(filename, add_latlon=False):
```
gidden: Now that I've played around a little bit, I think keeping the default to `False` is the right approach

gidden: Should the crs argument be passed along to `add_latlon_coords_from_crs`? I.e., 

```
def open_rasterio(filename, add_latlon=False, crs=None)
```

or


```
def open_rasterio(filename, add_latlon=False, **kwargs)
```

xarray/core/utils.py
```diff
@@ -489,3 +489,46 @@ def ensure_us_time_resolution(val):
...
         val = val.astype('timedelta64[us]')
     return val
+
+
+def add_latlon_coords_from_crs(ds, crs=None):
+    """Computes the longitudes and latitudes out of the x and y coordinates 
+    of a dataset and a coordinate reference system (crs). If crs isn't provided
+    it will look for "crs" in the dataset's attributes. 
+    
+    Needs rasterIO to be installe.
```
gidden: installed

xarray/core/utils.py
```diff
@@ -489,3 +489,46 @@ def ensure_us_time_resolution(val):
...
+        from rasterio.warp import transform
+    except ImportError:
+        raise ImportError('add_latlon_coords_from_crs needs RasterIO.')
+
+    if crs is None:
+        if 'crs' in ds.attrs:
+            crs = ds.attrs['crs']
+        else:
+            raise ValueError('crs not found')
+
```
gidden: could check for an empty CRS here (see prior comment)

xarray/core/utils.py
```diff
@@ -489,3 +489,46 @@ def ensure_us_time_resolution(val):
...
+    ny, nx = len(ds['y']), len(ds['x'])
+    x, y = np.meshgrid(ds['x'], ds['y'])
+    # Rasterio works with 1D arrays
+    xc, yc = transform(crs, {'init': 'EPSG:4326'},
```
gidden: silly question, but is the dst transform always `EPSG:4326`?

gidden: Or would we want to pass in the `dst_crs` as an argument?

fmaussion:
Thanks @gidden for the comments! Will look into it. 

Your questions about the ``add_latlon`` makes me think a kwarg at read time isn't the right approach. Here are the scenarios where you **don't** want to have the lat/lon computed per defaut:
- when your data's crs is *already* WGS84, in that case ``x`` and ``y`` are lons and lats already
- when your file isn't georeferenced properly
- when your use case doesn't need them ([salem](http://salem.readthedocs.io/en/latest/) for example will make better use of ``crs`` than xarray)
- when your file is large: computing lons and lats on a huge 2D grid is going to be prohibitively expensive.

This latter use case is important, because it might be useful for users to first *subset* their data and then compute the lat lons. For this use case we could go for two options in place of the kwarg:
1. add a top level utility function ``get_latlon_from_crs`` which would work on any dataset with a proper crs (and could be extended)
2. compute lons and lats lazily (only when asked for)
3. do nothing and let the users do their own cuisine (consistent with xarray's general purpose)

I don't know how to do **2** because it implies using dask to compute two related variables at the same time. Furthermore, **2** requires dask while **1** could be extended towars other datasets which have a ``crs``.

Right now I tend towards **3** (because I use salem), although I now that many users will benefit from **1**...

@gidden @shoyer : thoughts?

gidden:
Hey @fmaussion, I personally think some lat/lon support should be included here. I would lean toward *1* personally, but would be interested in what others think.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __getitem__(self, key):
+
+        # make our job a bit easier
+        key = indexing.canonicalize_indexer(key, self._ndims)
+
+        # bands cannot be windowed but they can be listed
+        bands, n = key[0], self.shape[0]
+        if isinstance(bands, slice):
+            start = bands.start if bands.start is not None else 0
```
shoyer: Use `start, stop, step = bands.indices(n)` to handle the normalization

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+    def shape(self):
+        return self._shape
+
+    def __getitem__(self, key):
+
+        # make our job a bit easier
+        key = indexing.canonicalize_indexer(key, self._ndims)
+
+        # bands cannot be windowed but they can be listed
+        bands, n = key[0], self.shape[0]
```
shoyer: Put this on two lines.

Use more descriptive variable names, e.g., `bands_key` and `num_bands`

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+            bands = np.arange(start, stop)
+        # be sure we give out a list
+        bands = (np.asarray(bands) + 1).tolist()
+
+        # but other dims can only be windowed
+        window = []
+        squeeze_axis = []
+        for i, (k, n) in enumerate(zip(key[1:], self.shape[1:])):
+            if isinstance(k, slice):
+                start = k.start if k.start is not None else 0
```
shoyer: again use `.indices()`

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+                    raise IndexError(_error_mess)
+            else:
+                if is_scalar(k):
+                    # windowed operations will always return an array
+                    # we will have to squeeze it later
+                    squeeze_axis.append(i+1)
+                k = np.asarray(k).flatten()
+                start = k[0]
+                stop = k[-1] + 1
+                if (stop - start) != len(k):
```
shoyer: It is not guaranteed that `k` is in any particular order. Instead I would use `(k == np.arange(start, stop)).all()`.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+                start = k.start if k.start is not None else 0
+                stop = k.stop if k.stop is not None else n
+                if k.step is not None and k.step != 1:
+                    raise IndexError(_error_mess)
+            else:
+                if is_scalar(k):
+                    # windowed operations will always return an array
+                    # we will have to squeeze it later
+                    squeeze_axis.append(i+1)
+                k = np.asarray(k).flatten()
```
shoyer: For clarity, I would make the scalar case separate and not use `.flatten()` (which could have other undesirable consequences)

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+    rasterio = False
+
+from .. import Variable
+from ..core.utils import FrozenOrderedDict, Frozen, NDArrayMixin, is_scalar
+from ..core import indexing
+from ..core.pycompat import OrderedDict, suppress
+
+from .common import AbstractDataStore
+
+_rio_varname = 'raster'
```
shoyer: Use all capitals for module level constants, e.g., `_RIO_VARNAME`

xarray/backends/api.py
```diff
@@ -318,6 +318,33 @@ def maybe_decode_store(store, lock=False):
...
+    ----------
+    filename : str
+        path to the file to open
+    add_latlon : bool, optional
+        if the file ships with valid geoinformation, longitudes and latitudes
+        can be computed and added to the dataset as non-dimension coordinates
+
+    Returns
+    -------
+    dataset : Dataset
```
shoyer: Why return a `Dataset` rather than a `DataArray`? The later feels a little more natural to me, given that a rasterio describes a single array.

xarray/backends/api.py
```diff
@@ -318,6 +318,33 @@ def maybe_decode_store(store, lock=False):
...
+    """Open a file with RasterIO (experimental).
+
+    This should work with any file that rasterio can open (typically: 
+    geoTIFF). The x and y coordinates are generated automatically out of the 
+    file's geoinformation.
+
+    Parameters
+    ----------
+    filename : str
+        path to the file to open
```
shoyer: Here and below: please capitalize first letter and include period at end.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,131 @@
...
+            self.dims = ('band', 'y', 'x')
+            self._vars['band'] = Variable(('band',),
+                                          np.atleast_1d(self.ds.indexes))
+        else:
+            raise ValueError('Unknown dims')
+
+        self._attrs = OrderedDict()
+        with suppress(AttributeError):
+            for attr_name in ['crs']:
+                self._attrs[attr_name] = getattr(self.ds, attr_name)
```
shoyer: Can we document the type of the "crs" attribute?

xarray/tests/test_backends.py
```diff
@@ -1429,6 +1429,201 @@ class TestPyNioAutocloseTrue(TestPyNio):
...
+
+            # ref
+            expected = Dataset()
+            expected['x'] = ('x', np.arange(nx)*0.5 + 1)
+            expected['y'] = ('y', -np.arange(ny)*2 + 2)
+            expected['band'] = ('band', [1, 2, 3])
+            expected['raster'] = (('band', 'y', 'x'), data)
+
+            # tests
+            _ex = expected.isel(band=1)
```
shoyer: no need to use private variable naming for names in a test case.

fmaussion:
Thanks for looking into this!

> I'm a slightly reluctant to add a dedicated method for convert rasterio CRS objects into lat/lon arrays. It feels a little overly specialized.

OK, so this corresponds to my solution **3** above (do nothing). I will however add the relevant lines of code to the documentation so that users wanting to add lons and lats to their data can do so. This will leave room for solution **2** is someone has the time to do it later. (side note: "rasterio CRS objects" are in fact strings corresponding to a PROJ4 string that will always be understood by ``pyproj``, ``rasterio``, ``gdal``, etc.  Examples: ``'+proj=aea +lat_1=-18 +lat_2=-32 +lat_0=0 +lon_0=24 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs '`` or ``'EPSG:4326'``)

> Why return a Dataset rather than a DataArray? The later feels a little more natural to me, given that a rasterio describes a single array.

Agreed. Again this is built out of discussions on https://github.com/pydata/xarray/issues/790, but now obsolete. This speaks even stronger against the use of a ``DataStore``, right? Is there any class I should inherit from to create a DataArray from a rasterio file? I basically need an ``init()`` and a ``__getitem()``...

shoyer:
You should be able to simply wrap RasterioArrayWrapper in an `indexing.LazilyIndexedArray` and then pass it into the xarray.DataArray constructor directly.

Zac-HD:
I would also favour doing nothing (ie 3), because most users will already have some solution.  It's also easier to change later if we don't do anything now - no need to think at all about backwards compatibility, and the design can be guided by how people are using the existing parts.

shoyer:
> side note: "rasterio CRS objects" are in fact strings corresponding to a PROJ4 string that will always be understood by pyproj, rasterio, gdal, etc. Examples: '+proj=aea +lat_1=-18 +lat_2=-32 +lat_0=0 +lon_0=24 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs ' or 'EPSG:4326')

Ah, that seems already pretty clean then. I was confused by the repr for a `CRS()` object in your comment above: https://github.com/pydata/xarray/pull/1260#issuecomment-279064452

fmaussion:
Uh, right, this is obviously not a string!

gidden:
Hey @fmaussion, should we also do any attribute checking, or should we let the user fails as follows? I recently was testing this out and made a `*.tiff` with a shapefile as read from `fiona`. I got the following error:

```
In [12]: ds.to_netcdf('test.nc')
<ipython-input-12-d7687fe1d488> in <module>()
----> 1 ds.to_netcdf('test.nc')

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/core/dataset.pyc in to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims)
    975         return to_netcdf(self, path, mode, format=format, group=group,
    976                          engine=engine, encoding=encoding,
--> 977                          unlimited_dims=unlimited_dims)
    978 
    979     def __unicode__(self):

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in to_netcdf(dataset, path_or_file, mode, format, group, engine, writer, encoding, unlimited_dims)
    588     # validate Dataset keys, DataArray names, and attr keys/values
    589     _validate_dataset_names(dataset)
--> 590     _validate_attrs(dataset)
    591 
    592     try:

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in _validate_attrs(dataset)
    114     # Check attrs on the dataset itself
    115     for k, v in dataset.attrs.items():
--> 116         check_attr(k, v)
    117 
    118     # Check attrs on each variable within the dataset

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in check_attr(name, value)
    110                             'string, ndarray or a list/tuple of '
    111                             'numbers/strings for serialization to netCDF '
--> 112                             'files'.format(value))
    113 
    114     # Check attrs on the dataset itself

TypeError: Invalid value for attr: CRS({'init': u'epsg:4326'}) must be a number string, ndarray or a list/tuple of numbers/strings for serialization to netCDF files
```

fmaussion:
@shoyer @gidden tjhanks for testing, this is very useful.

Indeed rasterio uses a dict-like mapping for the PROJ4 strings (source: [rasterio docs](https://mapbox.github.io/rasterio/_modules/rasterio/crs.html)). This isn't a big deal since it provides the ``to_string()`` and ``from_string()`` methods which allow the do the round-trip.

I personally never noted the difference because pyproj (the python interface to the PROJ.4 library) can handle both representations:

```
In [1]: import xarray as xr

In [2]: ds = xr.open_rasterio('RGB.byte.tif')

In [3]: ds.crs
Out[3]: CRS({'init': 'epsg:32618'})

In [4]: import pyproj

In [5]: pyproj.Proj(ds.crs)
Out[5]: <pyproj.Proj at 0x7f12317c0468>

In [6]: pyproj.Proj(ds.crs.to_string())
Out[6]: <pyproj.Proj at 0x7f12317c0408>
```

My suggestion here (to avoid the serialisation problems you mention @gidden ) is to convert the CRS object to a string at read time

gidden:
Hey @fmaussion, I guess my question here is as follows: if a raster was generated by rasterio (thus using a dictionary representation of the CRS object even if a string is provided), should xarrary then fail to write to netcdf with that dataset?

If so, then that means that all users will have to do 

```
ds = open_rasterio('file.tiff')
ds.attrs['crs'] = ds.attrs['crs'].to_string()
ds.to_netcdf('file.nc')
```

Would it be better to use the string representation of the CRS internally after reading in? I think this would solve @shoyer's concern about onboarding non-primitive datatypes.

As an aside, even my above suggestion did not work.. I'm not sure why.

```
In [33]: ds = xr.open_rasterio('./isimip_centroid_0_5.tiff')

In [34]: ds
Out[34]: 
<xarray.Dataset>
Dimensions:  (band: 1, x: 720, y: 360)
Coordinates:
  * y        (y) float64 90.0 89.5 89.0 88.5 88.0 87.5 87.0 86.5 86.0 85.5 ...
  * x        (x) float64 -180.0 -179.5 -179.0 -178.5 -178.0 -177.5 -177.0 ...
  * band     (band) int64 1
Data variables:
    raster   (band, y, x) int32 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 ...
Attributes:
    crs:      CRS({'init': u'epsg:4326'})

In [35]: ds.attrs['crs'] = ds.attrs['crs'].to_string()

In [36]: ds
Out[36]: 
<xarray.Dataset>
Dimensions:  (band: 1, x: 720, y: 360)
Coordinates:
  * y        (y) float64 90.0 89.5 89.0 88.5 88.0 87.5 87.0 86.5 86.0 85.5 ...
  * x        (x) float64 -180.0 -179.5 -179.0 -178.5 -178.0 -177.5 -177.0 ...
  * band     (band) int64 1
Data variables:
    raster   (band, y, x) int32 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 ...
Attributes:
    crs:      +init=epsg:4326

In [37]: ds.to_netcdf('test.nc')
<ipython-input-37-d7687fe1d488> in <module>()
----> 1 ds.to_netcdf('test.nc')

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/core/dataset.pyc in to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims)
    975         return to_netcdf(self, path, mode, format=format, group=group,
    976                          engine=engine, encoding=encoding,
--> 977                          unlimited_dims=unlimited_dims)
    978 
    979     def __unicode__(self):

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in to_netcdf(dataset, path_or_file, mode, format, group, engine, writer, encoding, unlimited_dims)
    588     # validate Dataset keys, DataArray names, and attr keys/values
    589     _validate_dataset_names(dataset)
--> 590     _validate_attrs(dataset)
    591 
    592     try:

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in _validate_attrs(dataset)
    119     for variable in dataset.variables.values():
    120         for k, v in variable.attrs.items():
--> 121             check_attr(k, v)
    122 
    123 

/home/gidden/.local/lib/python2.7/site-packages/xarray-0.9.5_34_g48c7268-py2.7.egg/xarray/backends/api.pyc in check_attr(name, value)
    110                             'string, ndarray or a list/tuple of '
    111                             'numbers/strings for serialization to netCDF '
--> 112                             'files'.format(value))
    113 
    114     # Check attrs on the dataset itself

TypeError: Invalid value for attr: CRS({'init': u'epsg:4326'}) must be a number string, ndarray or a list/tuple of numbers/strings for serialization to netCDF files
```

gidden:
Note that the above `tiff` was made with something like the following code:

```
res = 0.5
nlat = 180
nlon = 360
left_lon = -180
upper_lat = 90
crs = 'epsg:4326'

profile = {
    'affine': rio.Affine(res, 0.0, left_lon,
                         0.0, -res, upper_lat),
    'dtype': 'int32',
    'height': int(nlat / res),
    'width': int(nlon / res),
    'nodata': '-1',
    'crs': crs,
}
```

fmaussion:
> Would it be better to use the string representation of the CRS internally after reading in? 

Yes this was my intention.

BTW, your serialization above doesn't work because the variable "raster" also has a CRS attr. This problem will be solved by the next iteration of my code (when data arrays will be returned instead of datasets)

gidden:
Sounds good!

fmaussion:
This is ready for another round of reviews! I think this has come out quite nicely. Everything is much simpler now.

I have:
- included all your comments
- removed the GIS part 
- added an example on how to parse lons and lats in the new "recipes" section

doc/gallery/plot_rasterio.py
```diff
@@ -0,0 +1,55 @@
...
+# Rasterio works with 1D arrays
+from rasterio.warp import transform
+lon, lat = transform(rioda.crs, {'init': 'EPSG:4326'},
+                     x.flatten(), y.flatten())
+lon = np.asarray(lon).reshape((ny, nx))
+lat = np.asarray(lat).reshape((ny, nx))
+
+# Convert the DataArray to a dataset and set them as non-dimension coordinates
+riods = rioda.to_dataset(name='img')
+riods['lon'] = (('y', 'x'), lon)
```
shoyer: I would just set `riods.coords['lon'] = ...` directly.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,116 @@
...
+    attrs = {}
+    if hasattr(riods, 'crs'):
+        # CRS is a dict-like object specific to rasterio
+        # We convert it back to a PROJ4 string using rasterio itself
+        attrs['crs'] = riods.crs.to_string()
+    # Maybe we'd like to parse other attributes here (for later)
+
+    data = indexing.LazilyIndexedArray(RasterioArrayWrapper(riods))
+    return DataArray(data=data, dims=('band', 'y', 'x'),
+                     coords=coords, attrs=attrs)
```
shoyer: Make this DataArray close-able by setting `_file_obj`, e.g.,
```python
result = DataArray(data=data, dims=('band', 'y', 'x'), ...)
result._file_obj = riods
return result
```
(arguably this should be public API, but it's currently private)

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @property
+    def dtype(self):
+        return np.dtype(self.riods.dtypes[0])
+
+    @property
+    def shape(self):
+        return self._shape
+
+    def __exit__(self, exception_type, exception_value, traceback):
```
shoyer: I don't think this method does anything -- we never use this array as a context manager.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,116 @@
...
+from ..core.utils import NDArrayMixin, is_scalar
+from ..core import indexing
+
+
+_ERROR_MSG = ('The kind of indexing operation you are trying to do is not '
+              'valid on rasterio files. Try to load your data with ds.load()'
+              'first.')
+
+
+class RasterioArrayWrapper(NDArrayMixin):
```
shoyer: You need to set an `array` property if you use `NDArrayMixin`.

In this case, I would just use the base classes `NdimSizeLenMixin` and `DunderArrayMixin`.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,116 @@
...
+    """A wrapper around rasterio dataset objects"""
+    def __init__(self, riods):
+        self.riods = riods
```
shoyer: I would prefer a slightly more explicit name, e.g., `rasterio_ds`

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,116 @@
...
+                        raise IndexError(_ERROR_MSG)
+            window.append((start, stop))
+
+        out = self.riods.read(band_key, window=window)
+        if squeeze_axis:
+            out = np.squeeze(out, axis=squeeze_axis)
+        return out
+
+
+def rasterio_to_dataarray(filename):
```
shoyer: Why not just make this function `open_rasterio` rather than using indirection?

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,116 @@
+from collections import OrderedDict
+import numpy as np
+
+try:
+    import rasterio
```
shoyer: Put this import inside the function where it's used below. That way you get a sensible error message if/when rasterio is not installed.

xarray/tests/test_backends.py
```diff
@@ -1429,6 +1429,172 @@ class TestPyNioAutocloseTrue(TestPyNio):
...
+
+            # but on x and y only windowed operations are allowed, more
+            # exotic slicing should raise an error
+            with self.assertRaisesRegexp(IndexError, 'not valid on rasterio'):
+                _ = actual.isel(x=[2, 4], y=[1, 3]).values
+            with self.assertRaisesRegexp(IndexError, 'not valid on rasterio'):
+                _ = actual.isel(x=[4, 2]).values
+            with self.assertRaisesRegexp(IndexError, 'not valid on rasterio'):
+                _ = actual.isel(x=slice(5, 2, -1)).values
+
```
shoyer: Can you add a check that `actual.isel(x=[0, 1], y=[0, 1])` works like `actual.isel(x=slice(2), y=slice(2))`?

xarray/backends/rasterio_.py
```diff
@@ -5,127 +5,109 @@
...
+    """A wrapper around rasterio dataset objects"""
+    def __init__(self, riods):
+        self.riods = riods
+        self._shape = self.riods.count, self.riods.height, self.riods.width
         self._ndims = len(self.shape)
 
     @property
     def dtype(self):
-        return np.dtype(self.ds.dtypes[0])
+        return np.dtype(self.riods.dtypes[0])
```
fmaussion: This is a good question. I just assume it *has* to be so (since rasterio's ``get_data`` returns a ndarray of uniform dtype) but I can't be sure. I added a check for this now.

If the bands have different dtypes they should probably be handled as different variables, which would be bad for the current dataarray design.

@perrygeo @sgillies if you have time could you comment here?

xarray/backends/rasterio_.py
```diff
@@ -5,127 +5,109 @@
...
+
+    # Attributes
+    attrs = {}
+    if hasattr(riods, 'crs'):
+        # CRS is a dict-like object specific to rasterio
+        # We convert it back to a PROJ4 string using rasterio itself
+        attrs['crs'] = riods.crs.to_string()
+    # Maybe we'd like to parse other attributes here (for later)
+
+    data = indexing.LazilyIndexedArray(RasterioArrayWrapper(riods))
```
fmaussion: I don't understand what you mean here, sorry :(

fmaussion:
Thanks @shoyer , I have addressed all your comments but one, which I didn't understand. Maybe we should wait for an answer of the rasterio devs about the dtype stuff before going on too.

xarray/backends/rasterio_.py
shoyer: GitHub seems to lose the specific comment link -- if you click edit on my comment here you'll find it.

To be more concrete, after this line I would suggest adding:
```
data = indexing.CopyOnWriteArray(data)
if cache:
    data = indexing.MemoryCachedArray(data)
```

`cache` should be an optional boolean argument, like it is on `open_dataset`. This also argues for adding a `chunks` argument, which is set uses dask to chunk the data and disables the cache.

fmaussion: Implementing ``cache`` went well (appart from this [bug](https://github.com/pydata/xarray/issues/1429).

For chunking I have a question: ``DataArray.chunk()`` [doesn't have](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.chunk.html#xarray.DataArray.chunk) the ``name_prefix``, ``token`` and ``lock`` keywords. Any reason for this?

shoyer:
I think DataArray.chunk should probably gain those parameters. There just
wasn't any need for it before.

xarray/backends/rasterio_.py
```diff
@@ -91,11 +93,32 @@ def open_rasterio(filename):
...
+    chunks : int, tuple or dict, optional
+        Chunk sizes along each dimension, e.g., ``5``, ``(5, 5)`` or
+        ``{'x': 5, 'y': 5}``. If chunks is provided, it used to load the new
+        DataArray into a dask array. This is an experimental feature; see the
+        documentation for more details.
+    cache : bool, optional
+        If True, cache data loaded from the underlying datastore in memory as
+        NumPy arrays when accessed to avoid reading from the underlying data-
+        store multiple times. Defaults to True unless you specify the `chunks`
+        argument to use dask, in which case it defaults to False. Does not
```
shoyer: This last sentence can be safely removed.

xarray/backends/rasterio_.py
```diff
@@ -91,11 +93,32 @@ def open_rasterio(filename):
     -------
     data : DataArray
         The newly created DataArray.
+    chunks : int, tuple or dict, optional
+        Chunk sizes along each dimension, e.g., ``5``, ``(5, 5)`` or
+        ``{'x': 5, 'y': 5}``. If chunks is provided, it used to load the new
+        DataArray into a dask array. This is an experimental feature; see the
```
shoyer: We can probably remove the "experimental feature" bit. Dask is not going away.

xarray/backends/rasterio_.py
```diff
@@ -91,11 +93,32 @@ def open_rasterio(filename):
...
+        If True, cache data loaded from the underlying datastore in memory as
+        NumPy arrays when accessed to avoid reading from the underlying data-
+        store multiple times. Defaults to True unless you specify the `chunks`
+        argument to use dask, in which case it defaults to False. Does not
+        change the behavior of coordinates corresponding to dimensions, which
+        always load their data from disk into a ``pandas.Index``.
+    lock : False, True or threading.Lock, optional
+        If chunks is provided, this argument is passed on to
+        :py:func:`dask.array.from_array`. By default, a per-variable lock is
+        used when reading data from netCDF files with the netcdf4 and h5netcdf
```
shoyer: Does rasterio/GDAL need a thread lock? Let's remove netcdf4/h5netcdf specific references here and add sane defaults.

xarray/backends/rasterio_.py
```diff
@@ -120,7 +143,36 @@ def open_rasterio(filename):
...
+    if chunks is not None:
+        # Logic borrowed from open_dataset
+        try:
+            from dask.base import tokenize
+        except ImportError:
+            # raise the usual error if dask is entirely missing
+            import dask
+            if LooseVersion(dask.__version__) < LooseVersion('0.6'):
+                raise ImportError(
+                    'xarray requires dask version 0.6 or newer')
```
shoyer: dask v0.6 is pretty old now, maybe better to replace this whole conditional logic with:
```
import dask
```

xarray/backends/rasterio_.py
sgillies: You cannot safely share a Rasterio or GDAL dataset object among threads, so a lock may be needed. See https://trac.osgeo.org/gdal/wiki/FAQMiscellaneous#IstheGDALlibrarythread-safe.

fmaussion: OK, I am on it.

@shoyer , I have trouble to understand what's going on in the ``open_dataset`` workflow. The documentation says: "By default, a per-variable lock is used when reading data from netCDF files with the netcdf4 and h5netcdf engines", but then the lock is decided once [here](https://github.com/pydata/xarray/blob/master/xarray/backends/api.py#L310). This will default to ``GLOBAL_LOCK`` and then passed to ``maybe_decode_store``. Where is the "per variable" logic happening? Or is the ``GLOBAL_LOCK`` already doing this? (it is defined [here](https://github.com/pydata/xarray/blob/371d034372bc7522098a142a0debf93916c49102/xarray/backends/common.py#L30)) 

(excuse my poor knowledge about these issues)

shoyer: The documentation is out of sync here. We use a global lock for HDF5 (and should probably use one for Rasterio, too).

fmaussion: Now I understand much more what's going on, and also why salem was having troubles when adding diagnostic variables to netcdf objects. Can I update the docs all at once here?

shoyer:
Yes, please!

fmaussion:
OK, all green.

Currently the rasterio tests are running on py36 only. Should I add rasterio to the other test suites as well?

shoyer:
> Currently the rasterio tests are running on py36 only. Should I add rasterio to the other test suites as well?

Yes, except for the one that tests bare-minimal dependencies.

gidden:
Hey @fmaussion, is this ready for me to try out again? I wanted to let you and @shoyer iterate first.

fmaussion:
@gidden yes absolutely please give it a try, I think it's ready

fmaussion:
@gidden I just updated the documentation recipe to use an accessor to compute the lons and lats, let me know what you think

doc/gallery/plot_rasterio.py
```diff
@@ -0,0 +1,75 @@
...
+import numpy as np
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+from rasterio.warp import transform
+
+
+# Define the accessor
+@xr.register_dataarray_accessor('rasterio')
+class RasterioAccessor(object):
```
shoyer: I would lean towards keeping this as just a function -- it keeps the example simpler and an accessor is probably overkill for this problem.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,166 @@
...
+                    raise IndexError(_ERROR_MSG)
+            window.append((start, stop))
+
+        out = self.rasterio_ds.read(band_key, window=window)
+        if squeeze_axis:
+            out = np.squeeze(out, axis=squeeze_axis)
+        return out
+
+
+def open_rasterio(filename, chunks=None, cache=None, lock=False):
```
shoyer: The default should be `lock=None` to use the global lock.

xarray/backends/rasterio_.py
```diff
@@ -0,0 +1,166 @@
...
+                       coords=coords, attrs=attrs)
+
+    if chunks is not None:
+        from dask.base import tokenize
+        # augment the token with the file modification time
+        mtime = os.path.getmtime(filename)
+        token = tokenize(filename, mtime, chunks)
+        name_prefix = 'open_rasterio-%s' % token
+        if lock is None:
+            lock = GLOBAL_LOCK
```
shoyer: Actually, can you make a *separate* global lock for rasterio? We can safely use GDAL and HDF5 at the same time.

xarray/tests/test_backends.py
```diff
@@ -1438,6 +1438,265 @@ class TestPyNioAutocloseTrue(TestPyNio):
     autoclose = True
 
 
+@requires_rasterio
+class TestRasterio(CFEncodedDataTest, Only32BitTypes, TestCase):
```
shoyer: I don't think there's any advantage to inheriting from the other test methods, given that this isn't an actual datastore. I think this can just inherit from `TestCase` (then you can delete the first two methods).

xarray/tests/test_backends.py
```diff
@@ -1438,6 +1438,265 @@ class TestPyNioAutocloseTrue(TestPyNio):
...
+                assert 'crs' in rioda.attrs
+                assert isinstance(rioda.attrs['crs'], basestring)
+
+                # Write it to a netcdf and read again (roundtrip)
+                with create_tmp_file(suffix='.nc') as tmp_nc_file:
+                    rioda.to_netcdf(tmp_nc_file)
+                    with xr.open_dataarray(tmp_nc_file) as ncds:
+                        assert_identical(rioda, ncds)
+
+        # Create a geotiff file in latlong proj
```
shoyer: If it's self-contained, make this a separate test method?

fmaussion:
OK, let's get this one in and see what people will report about it.

Thanks @shoyer for your patience, @NicWayand and @jhamman for the original PR, @gidden for the testing/reviews and @sgillies for rasterio!

jhamman:
Nice work @fmaussion. Thanks for bringing this home.

PeterDSteinberg:
+1 @fmaussion - Thanks!

PR #1261, opened 2017-02-11
Allow for plotting dummy netCDF4.datetime objects.

willirath:
Currently, xarray/plot.py raises a `TypeError` if the data to be plotted are not among `[np.floating, np.integer, np.timedelta64, np.datetime64]`.

This PR adds `netCDF4.datetime` objects to the list of allowed data types. These occur, because xarray/conventions.py passes undecoded `netCDF4.datetime` objects if decoding to `numpy.datetime64` fails.

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -32,7 +32,9 @@ def _ensure_plottable(*args):
     Raise exception if there is anything in args that can't be plotted on
     an axis.
     """
-    plottypes = [np.floating, np.integer, np.timedelta64, np.datetime64]
+    from netCDF4 import datetime as nc4datetime
+    plottypes = [np.floating, np.integer, np.timedelta64, np.datetime64, 
+                 nc4datetime]
 
     # Lists need to be converted to np.arrays here.
     if not any(_right_dtype(np.array(x), plottypes) for x in args):
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -13,34 +13,49 @@ import warnings
 
 import numpy as np
 import pandas as pd
+from datetime import datetime
 
 from .utils import _determine_cmap_params, _infer_xy_labels, get_axis
 from .facetgrid import FacetGrid
 from xarray.core.pycompat import basestring
 
 
-# Maybe more appropriate to keep this in .utils
-def _right_dtype(arr, types):
+def _valid_numpy_subdtype(x, numpy_types):
     """
-    Is the numpy array a sub dtype of anything in types?
+    Is any dtype from numpy_types superior to the dtype of x?
     """
-    return any(np.issubdtype(arr.dtype, t) for t in types)
+    # If any of the types given in numpy_types is understood as numpy.generic,
+    # all possible x will be considered valid.  This is probably unwanted.
+    # Warn then.
+    for t in numpy_types:
+        if np.issubdtype(np.generic, t):
+            warnings.warn('{} is treated as numpy.generic.'.format(t),
+                          RuntimeWarning, stacklevel=3)
+
+    return any(np.issubdtype(x.dtype, t) for t in numpy_types)
+
+
+def _valid_other_type(x, types):
+    """
+    Do all elements of x have a type from types?
+    """
+    return all(any(isinstance(el, t) for t in types)
+            for el in np.ravel(x))
 
 
 def _ensure_plottable(*args):
     """
-    Raise exception if there is anything in args that can't be plotted on
-    an axis.
+    Raise exception if there is anything in args that can't be plotted on an
+    axis.
     """
-    from netCDF4 import datetime as nc4datetime
-    plottypes = [np.floating, np.integer, np.timedelta64, np.datetime64, 
-                 nc4datetime]
+    numpy_types = [np.floating, np.integer, np.timedelta64, np.datetime64]
+    other_types = [datetime]
 
-    # Lists need to be converted to np.arrays here.
-    if not any(_right_dtype(np.array(x), plottypes) for x in args):
-        raise TypeError('Plotting requires coordinates to be numeric '
-                        'or dates.')
-
+    for x in args:
+        if not (_valid_numpy_subdtype(np.array(x), numpy_types)
+                or _valid_other_type(np.array(x), other_types)):
+            raise TypeError('Plotting requires coordinates to be numeric '
+                            'or dates.')
 
 def _easy_facetgrid(darray, plotfunc, x, y, row=None, col=None,
                     col_wrap=None, sharex=True, sharey=True, aspect=None,
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -37,7 +37,7 @@ def _valid_other_type(x, types):
     Do all elements of x have a type from types?
     """
     return all(any(isinstance(el, t) for t in types)
-            for el in np.ravel(x))
+               for el in np.ravel(x))
 
 
 def _ensure_plottable(*args):
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -36,8 +36,10 @@ def _valid_other_type(x, types):
     """
     Do all elements of x have a type from types?
     """
-    return all(any(isinstance(el, t) for t in types)
-               for el in np.ravel(x))
+    if not np.issubdtype(np.generic, x.dtype):
+        return False
+    else:
+        return all(any(isinstance(el, t) for t in types) for el in np.ravel(x))
 
 
 def _ensure_plottable(*args):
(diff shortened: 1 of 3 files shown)


shoyer:
Does matplotlib actually plot netcdftime datetimes correctly?

It would be nice to have a unit test to verify this works. See xarray/tests/test_plot.py for examples.

willirath:
Failing tests:
```
xarray/tests/test_plot.py::TestPlot1D::test_nonnumeric_index_raises_typeerror FAILED
xarray/tests/test_plot.py::TestContourf::test_nonnumeric_index_raises_typeerror FAILED
xarray/tests/test_plot.py::TestContour::test_nonnumeric_index_raises_typeerror FAILED
xarray/tests/test_plot.py::TestPcolormesh::test_nonnumeric_index_raises_typeerror FAILED
xarray/tests/test_plot.py::TestImshow::test_nonnumeric_index_raises_typeerror FAIL
```

All of these check strings as input to the plot functions and expect a `TypeError`.

`xarray.plot._right_dtype`, however, uses `numpy.issubdtype` to check, if the types of the given data are subtypes of the now included `netCDF4.datetime`, which is interpreted as `numpy.generic`. And as `numpy.generic` is the highest type in Numpy's type hierarchy, `xarray.plot._right_dtype` now always returns `True`.

A possible way out would be to check for numeric types and datetime-like types separately.

xarray/plot/plot.py
```diff
@@ -32,7 +32,9 @@ def _ensure_plottable(*args):
     Raise exception if there is anything in args that can't be plotted on
     an axis.
     """
-    plottypes = [np.floating, np.integer, np.timedelta64, np.datetime64]
+    from netCDF4 import datetime as nc4datetime
```
shoyer: This needs to be guarded behind a try/except clause, because netCDF4 is not a required dependency of xarray.  Use `from netcdftime import datetime` instead of netCDF4 (@jhamman is moving netcdftime to its own package). Also, put import statements at the top of the file.

willirath:
> Does matplotlib actually plot netcdftime datetimes correctly?

Yes it does. They are just treated like `datetime.datetime` objects.

xarray/plot/plot.py
willirath: What should I do to prevent `netcdftime.datetime` to be understood as `numpy.generic`?  Rewrite `xarray.plot._ensure_plottable` to handle Numpy types and `netcdftime.datetime` separately?

willirath:
- [ ] Add tests showing that netcdftime.datetime objects can be plotted (line and 2d plots).
- [ ] Rewrite `xarray.plot._ensure_plottable` to only let numeric and date-like types pass.

xarray/plot/plot.py
shoyer: Yes, looks like we need a separate check for object dtype arrays (it's okay to loop over element).

willirath:
> > Does matplotlib actually plot netcdftime datetimes correctly?
> 
> Yes it does. They are just treated like `datetime.datetime` objects.

I was wrong here: [`netCDF4.datetime` _is_ `datetime.datetime`](https://github.com/Unidata/netcdf4-python/blob/master/netCDF4/_netCDF4.pyx#L5084) while ([`netcdftime.datetime` is `DatetimeProlepticGregorian`](https://github.com/Unidata/netcdftime/blob/master/netcdftime/__init__.py#L3).

And matplotlib fails on the latter: https://gist.github.com/willirath/48c04db0481e986fac0e44549e164ab0

willirath:
a85b5c6 and e31aaab add a test plotting against a coordinate with `netCDF4.datetime` objects.

6f5e30f add `datetime.datetime` objects to the list of plottable dtypes and splits the check into numpy dtypes (using numpy's type hierarchy) and all other types.

willirath:
`netcdftime.datetime` vs. `datetime.datetime` is discussed in Unidata/netcdftime#8. For now, I only added `datetime.datetime` to the list of plottable other types (see https://github.com/willirath/xarray/blob/6f5e30f3a2f8ca907f0fd91bcd571e7d98027896/xarray/plot/plot.py#L52).

xarray/plot/plot.py
```diff
@@ -13,32 +13,49 @@
...
-    Is the numpy array a sub dtype of anything in types?
+    Is any dtype from numpy_types superior to the dtype of x?
     """
-    return any(np.issubdtype(arr.dtype, t) for t in types)
+    # If any of the types given in numpy_types is understood as numpy.generic,
+    # all possible x will be considered valid.  This is probably unwanted.
+    # Warn then.
+    for t in numpy_types:
+        if np.issubdtype(np.generic, t):
+            warnings.warn('{} is treated as numpy.generic.'.format(t),
```
shoyer: This is an internal, private function. Rather than warning let's issue a `ValueError` to ensure it doesn't happen by accident. It would also be safe to just use an `assert` here, since again it should never happen in user code.

xarray/tests/test_plot.py
```diff
@@ -1208,3 +1213,25 @@ def test_default_labels(self):
         # Top row should be labeled
         for label, ax in zip(self.darray.coords['col'].values, g.axes[0, :]):
             self.assertTrue(substring_in_axes(label, ax))
+
+
+@requires_netCDF4
```
shoyer: This test shouldn't be netCDF4 specific anymore.

xarray/tests/test_plot.py
```diff
@@ -1208,3 +1213,25 @@ def test_default_labels(self):
...
+        '''
+        Create a DataArray with a time-axis that contains netCDF4.datetime
+        objects.
+        '''
+        month = np.arange(1, 13, 1)
+        data = np.sin(2 * np.pi * month / 12.0)
+
+        darray = DataArray(data, dims=['time'])
+        darray.coords['time'] = \
+                np.array([nc4datetime(2017, m, 1) for m in month])
```
shoyer: Again, just use a normal `datetime` instance now.

xarray/tests/test_plot.py
```diff
@@ -1208,3 +1213,25 @@ def test_default_labels(self):
...
+    def setUp(self):
+        '''
+        Create a DataArray with a time-axis that contains netCDF4.datetime
+        objects.
+        '''
+        month = np.arange(1, 13, 1)
+        data = np.sin(2 * np.pi * month / 12.0)
+
+        darray = DataArray(data, dims=['time'])
+        darray.coords['time'] = \
```
shoyer: style: avoid `\` in favor of implicit line continuation with parentheses: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

xarray/plot/plot.py
willirath: Thanks for the review! I've changed it using `assert`.

xarray/tests/test_plot.py
willirath: That's right.  I've changed it to use `datetime` and to not require `netCDF4`.

xarray/tests/test_plot.py
willirath: Done.

xarray/tests/test_plot.py
willirath: Done.

willirath:
Yes, I know that xarray usually doesn't use `datetime`. For climate data past 2262 with a standard calendar, however, the dtypes of the time axis are falling back to `datetime` / `netCDF4.datetime`.

My last commit reflects what I'd consider ready to be pulled. Let me know what you think.

xarray/plot/plot.py
```diff
@@ -13,32 +13,46 @@
...
+
+def _valid_other_type(x, types):
     """
-    Raise exception if there is anything in args that can't be plotted on
-    an axis.
+    Do all elements of x have a type from types?
     """
-    plottypes = [np.floating, np.integer, np.timedelta64, np.datetime64]
+    return all(any(isinstance(el, t) for t in types)
+            for el in np.ravel(x))
```
willirath: Done.  (Line was 79 chars long anyway.)

xarray/plot/plot.py
```diff
@@ -13,32 +13,46 @@
...
+    """
+    Raise exception if there is anything in args that can't be plotted on an
+    axis.
+    """
+    numpy_types = [np.floating, np.integer, np.timedelta64, np.datetime64]
+    other_types = [datetime]
+
+    for x in args:
+        if not (_valid_numpy_subdtype(np.array(x), numpy_types)
+                or _valid_other_type(np.array(x), other_types)):
```
willirath: I've added a check to `_valid_other_type` which ensures `np.issubdtype(np.generic, x.dtype)` before looping over `x`.

shoyer:
Thanks!

PR #1272, opened 2017-02-16
Groupby-like API for resampling

### LINKED ISSUE #1269: GroupBy like API for resample
Since we wrote `resample` in xarray, pandas updated resample to have a groupyby-like API (e.g., `df.resample('24H').mean()` vs. the old `df.resample('24H')` that uses the mean by default).

It would be nice to redo the xarray resample API to match, e.g., `ds.resample(time='24H').mean()` vs `ds.resample('time', '24H')`. This would solve a few use cases, including grouped-resample arithmetic, iterating over groups and (mostly) take care of the need for `pd.TimeGrouper` support (https://github.com/pydata/xarray/issues/364). If we use `**kwargs` for matching dimension names, this could be done with a minimally painful deprecation cycle.

darothen:
This is a work-in-progress to resolve #1269. 

- [x] Basic functionality
- [x] Cleanly deprecate old API 
- [x] New test cases
- [x] Documentation / examples
- [x] "What's new"

Openly welcome feedback/critiques on how I approached this. Subclassing `Data{Array/set}GroupBy` may not be the best way, but it would be easy enough to re-write the necessary helper functions (just `apply()`, I think) so that we do not need to inherit form them directly. Additional issues I'm working to resolve:

- [x] I tried make sure that calls using the old API won't break by refactoring the old logic to `_resample_immediately()`. This may not be the best approach!
- [x] Similarly, I copied all the original test cases and added the suffix `..._old_api`; these could trivially be placed into their related test cases for the new API.
- [x] BUG: **keep_attrs** is ignored when you call it on methods chained to `Dataset.resample()`. Oddly enough, if I hard-code **keep_attrs=True** inside `reduce_array()` in `DatasetResample::reduce` it works just fine. I haven't figured out where the kwarg is getting lost.
- [x] BUG: Some of the test cases (for instance, `test_resample_old_vs_new_api`) fail because the resampling by calling `self.groupby_cls` ends up not working - it crashes because the group sizes that get computed are not what it expects. Occurs with both new and old API


xarray/core/common.py
```diff
@@ -568,9 +572,47 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+        warnings.warn("\n.resample() has been modified to defer "
+                      "calculations. Instead of passing 'dim' and "
+                      "'how=\"{how}\", instead consider using "
+                      ".resample({dim}=\"{freq}\").{how}() ".format(
+                            dim=dim, freq=freq, how=how
+                      ), DeprecationWarning, stacklevel=3)
+
         if isinstance(dim, basestring):
             dim = self[dim]
         group = DataArray(dim, [(RESAMPLE_DIM, dim)], name=RESAMPLE_DIM)
```
shoyer: Note: I think this should probably be `group = DataArray(dim, [(dim.dims, dim)], name=RESAMPLE_DIM)` but for some reason this works.

xarray/core/common.py
```diff
@@ -568,9 +572,47 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+                                      grouper=time_grouper)
+
+        return resampler
+
+    def _resample_immediately(self, freq, dim, how, skipna,
+                              closed, label, base, keep_attrs):
+        """Implement the original version of .resample() which immediately
+        executes the desired resampling operation. """
+        from .dataarray import DataArray
         RESAMPLE_DIM = '__resample_dim__'
```
shoyer: There might be better way to handle this, but currently using `RESAMPLE_DIM` is actually pretty critical here, and removing it is probably the source of some of the bugs you are seeing with your new implementation. The problem is we need to keep track of the resampled time variable in a dataset along with the original time variable. If we give the resampled group variable the same name, then xarray gets them mixed up.

darothen:
Smoothed out most of the problems from earlier and missing details. Still not sure if it's wise to refactor most of the resampling logic into a new **resample.py**, like what was done with **rolling**, but it still makes some sense to keep things in **groupby.py** because we're just subclassing existing machinery from there. 

The only issue now is the signature for **__init__()** in `Data{set,Array}Resample`, where we have to add in two keyword arguments. Python 2.x doesn't like named arguments after *\*args*. There are a few options here, mostly just playing with *\*\*kwargs* as in [this StackOverflow thread](http://stackoverflow.com/questions/15301999/python-2-x-default-arguments-with-args-and-kwargs).

shoyer:
> The only issue now is the signature for init() in Data{set,Array}Resample, where we have to add in two keyword arguments. Python 2.x doesn't like named arguments after *args. There are a few options here, mostly just playing with **kwargs as in this StackOverflow thread.

Yes, use `pop`, e.g., `dim = kwargs.pop('dim', None)`. `pop` removes the arguments from kwargs, so you can pass on the remaining ones unchanged to the super class method.

doc/time-series.rst
```diff
@@ -133,17 +133,45 @@ For example, we can downsample our dataset from hourly to 6-hourly:
...
+   ds.resample(time='6H').mean()
+
+You can also supply an arbitrary reduction function to aggregate over each
+resampling group:
+
+.. ipython: python
+
+   ds.resample(time='6H').reduce(np.mean)
+
+Resampling does not yet work for upsampling.
```
shoyer: I think we need to fix this before merging this PR, since it suggests the existing functionality would only exist in deprecated form. Pandas does this with a method called `.asfreq`, though this is basically pure sugar since in practice I think it works exactly the same as `.first` (or `.mean` if only doing pure upsampling).

darothen:
Thanks for the feedback, @shoyer! Will circle back around to continue
working on this in a few days when I have some free time.

- Daniel

darothen:
Should `.apply()` really work on non-aggregation functions? Based on the [pandas documentation](http://pandas.pydata.org/pandas-docs/stable/timeseries.html#resampling) it seems like "resample" is truly just a synonym for a transformation of the time dimension. I can't really find many examples of people using this as a substitute for time group-bys... it seems that's what the `pd.TimeGrouper` is for, in conjunction with a normal `.groupby()`.

As written, non-aggregation ("transformation"?) doesn't work because the call in `_combine()` to `_maybe_reorder()` messes things up (it drops all of the data along the resampled dimension). It shouldn't be too hard to fix this, although I'm leaning more and more to making stand-alone `Data{Array,set}Resample` classes in a separate file which only loosely inherit from their `Data{Array,set}GroupBy` cousins, since they need to re-write some really critical parts of the underlying machinery.

shoyer:
> I can't really find many examples of people using this as a substitute for time group-bys... it seems that's what the pd.TimeGrouper is for, in conjunction with a normal .groupby().

I think this is mostly because TimeGrouper has been around far longer than non-aggregating resample.

jhamman:
@darothen - can you give a summary of what's left to do here?

darothen:
I think a pull against the new releases is critical to see what breaks. Beyond that, just code clean up and testing. I can try to bump this higher on my priority list.

darothen:
I did my best to re-base everything to master... plan on spending an hour or so figuring out what's broken and at least restoring the status quo.

darothen:
### TODO

- [ ] ensure that `count()` works on `Data{Array,set}Resample` objects
- [ ] refactor `Data{Array,set}Resample` objects into a stand-alone file **core/resample.py** alongside **core/groupby.py**
- [ ] wrap `pytest.warns` around tests targeting old API 
- [ ] move old API tests into stand-alone
- [ ] Crude up-sampling. Copy/pasting Stephan's earlier comment from Feb 20:

> I think we need to fix this before merging this PR, since it suggests the existing functionality would only exist in deprecated form. Pandas does this with a method called .asfreq, though this is basically pure sugar since in practice I think it works exactly the same as .first (or .mean if only doing pure upsampling).

---

Alright @jhamman, here's the complete list of work left here. I'll tackle some of it during my commutes this week.

xarray/tests/test_dataset.py
```diff
@@ -2184,13 +2184,38 @@ def test_resample_and_first(self):
...
         self.assertDatasetIdentical(expected, actual)
 
         # upsampling
         expected_time = pd.date_range('2000-01-01', freq='3H', periods=19)
         expected = ds.reindex(time=expected_time)
+        actual = ds.resample(time='3H')
+        for how in ['mean', 'sum', 'first', 'last', ]:
+            method = getattr(actual, how)
+            result = method()
+            self.assertDatasetEqual(expected, result)
```
MaximilianR: No need to change, but in future you can use pytest parameters for these sorts of loops, and `assert_equal` rather than the `self...` methods throughout

darothen:
Finished off everything except upsampling. In pandas, all [upsampling](https://github.com/pandas-dev/pandas/blob/d02ef6f04466e4a74f67ad584cf38cdc6df56e42/pandas/core/resample.py#L890) works by constructing a new time index (which we already do) and then filling in the NaNs that result in the dataset with one of a few different rules. Not sure how involved this will be, but I anticipate this can all be implemented in core/resample.py

darothen:
Okay, it was a bit of effort but I implemented upsampling. For the padding methods I just re-index the Dataset or DataArray using the re-sampled time frequencies. I also added interpolation, but that was a bit tricky; we have to sort of break the split-apply-combine idiom to do that, so I created a `Resampler` mix-in which could contain the logic for the up-sampling. The `DatasetResampler` and `DataArrayResampler` each then implement similar logic for doing the interpolation. The up-sampling is designed to work with n-dimensional data.

The padding methods work 100% with dask arrays - since we're just calling xarray methods which themselves work with dask arrays! There are some eager computations (just the calculation of the up-sampled time frequencies) but I don't think that's a major issue; the actually re-indexing/padding is deferred. Interpolation works with dask arrays too, but eagerly does the computations. 

Could use a review from @shoyer  or @jhamman.

New **TODO** list:

- [ ] Add example chart to the timeseries doc page comparing the different upsampling options
- [ ] Additional up-sampling test cases for both `DataArray`s and `Dataset`s
- [ ] Code clean-up
- [ ] What's new

doc/time-series.rst
```diff
@@ -150,17 +150,49 @@ For example, we can downsample our dataset from hourly to 6-hourly:
...
+
+.. sample code block which shows some sample data up-sampled with the four
+different methods compared n a single plot. Use timeseries data.
+
+All of these resampling and groupby operation work on both Dataset
 and DataArray objects with any number of additional dimensions.
 
+.. note::
+
+   The ``resample`` api was updated in version ?? to reflect similar updates
```
jhamman: 0.10

xarray/core/common.py
```diff
@@ -509,23 +510,26 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
             a function that can be called like ``how(values, axis)`` to reduce
             ndarray values along the given axis. Valid choices that can be
             provided as a string include all the usual Dataset/DataArray
-            aggregations (``all``, ``any``, ``argmax``, ``argmin``, ``max``,
-            ``mean``, ``median``, ``min``, ``prod``, ``sum``, ``std`` and
-            ``var``), as well as ``first`` and ``last``.
+            aggregations (``all``, ``any``, ``argmax``, ``argmin``,
+            ``count,`` ``max``, ``mean``, ``median``, ``min``, ``prod``,
+            ``sum``, ``std`` and ``var``), as well as ``first`` and ``last``.
```
jhamman: Either indicate that `how` has been deprecated or remove it from the doc string (like pandas did).

xarray/core/common.py
```diff
@@ -539,24 +543,69 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
         """
         from .dataarray import DataArray
 
+        if dim is not None:
+            return self._resample_immediately(freq, dim, how, skipna, closed,
+                                              label, base, keep_attrs)
+
```
jhamman: Should we explicitly disallow the specification of `how` when a `indexers` is provided? I don't think we need to support that functionality since we know it will be deprecated.

doc/time-series.rst
```diff
@@ -150,17 +150,49 @@ For example, we can downsample our dataset from hourly to 6-hourly:
...
+   in pandas` ``resample`` api to be more groupby-like. Older style calls to
```
shoyer: note the stray backtick after pandas.

doc/time-series.rst
```diff
@@ -150,17 +150,49 @@ For example, we can downsample our dataset from hourly to 6-hourly:
 
 .. ipython:: python
 
-    ds.resample('6H', dim='time', how='mean')
+    ds.resample(time='6H')
 
-Resample also works for upsampling, in which case intervals without any
-values are marked by ``NaN``:
+This will create a specialized ``GroupBy`` object which saves information
```
shoyer: Probably better to call it a `Resample` object

doc/time-series.rst
```diff
@@ -150,17 +150,49 @@ For example, we can downsample our dataset from hourly to 6-hourly:
...
+
+   The ``resample`` api was updated in version ?? to reflect similar updates
+   in pandas` ``resample`` api to be more groupby-like. Older style calls to
+   ``resample`` will still be supported for a short period:
+
+   .. ipython:: python
+
+    ds.resample('6H', dim='time', how='mean')
+
+   Using the depcreated api, resample will also work for upsampling, in which
```
shoyer: spelling: decreated -> deprecated

xarray/core/common.py
shoyer: +1 that should be a TypeError

xarray/core/common.py
```diff
@@ -539,24 +543,69 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+        # More than one indexer is ambiguous, but we do in fact need one if
+        # "dim" was not provided, until the old API is fully deprecated
+        if len(indexer) != 1:
+            raise NotImplementedError(
```
shoyer: This should be `ValueError`

xarray/core/common.py
```diff
@@ -539,24 +543,69 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+                "Resampling only supported along single dimensions."
+            )
+        dim, freq = indexer.popitem()
+
+        if isinstance(dim, basestring):
+            dim_name = dim
+            dim = self[dim]
+            resample_dim = "resampled_" + dim_name
+        else:
+            raise ValueError("Dimension name should be a string; "
```
shoyer: `TypeError`

xarray/core/dataarray.py
```diff
@@ -1543,7 +1550,8 @@ def func(self, other):
             if isinstance(other, (Dataset, groupby.GroupBy)):
                 return NotImplemented
             if hasattr(other, 'indexes'):
-                align_type = OPTIONS['arithmetic_join'] if join is None else join
+                align_type = OPTIONS[
```
shoyer: Can you insert parentheses here instead, e.g.,
```python
                align_type = (OPTIONS['arithmetic_join']
                              if join is None else join)
```

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+        applied : DataArray or DataArray
+            The result of splitting, applying and combining this array.
+        """
+        if shortcut:
+            grouped = self._iter_grouped_shortcut()
+        else:
+            grouped = self._iter_grouped()
+        applied = (maybe_wrap_array(arr, func(arr, **kwargs))
+                   for arr in grouped)
+        combined = self._combine(applied, shortcut=shortcut)
```
shoyer: Can we call the `super()` method for these lines above instead?

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+        # dimension we used.
+        if self._dim in combined:
+            combined = combined.drop(self._dim)
+
+        if self._resample_dim in combined.dims:
+            combined = combined.rename({self._resample_dim: self._dim})
+
+        return combined
+
+    def reduce(self, func, dim=None, axis=None, shortcut=True,
```
shoyer: I don't think we actually need to override this method?

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+        applied : Dataset or DataArray
+            The result of splitting, applying and combining this dataset.
+        """
+        kwargs.pop('shortcut', None)  # ignore shortcut if set (for now)
+        applied = (func(ds, **kwargs) for ds in self._iter_grouped())
+        combined = self._combine(applied)
+
+        return combined.rename({self._resample_dim: self._dim})
+
+    def reduce(self, func, dim=None, keep_attrs=False, **kwargs):
```
shoyer: likewise I don't think we need this one

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+
+        f = interp1d(x, y, kind=kind, axis=axis, bounds_error=True,
+                     assume_sorted=True)
+        new_x = self._full_index.values.astype('float')
+
+        # construct new up-sampled DataArray
+        dummy = self._obj.copy()
+        dims = dummy.dims
+        coords = dummy.coords
+        coords[self._dim] = self._full_index
```
shoyer: You need to drop existing non-dimension coordinates along the resmapled dimension.

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+        def reduce_array(ar):
+            return ar.reduce(func, self._dim, axis=None, keep_attrs=keep_attrs,
+                             **kwargs)
+        return self.apply(reduce_array, shortcut=shortcut)
+
+    def _interpolate(self, kind='linear'):
+        from .dataarray import DataArray
+
+        x = self._obj[self._dim].astype('float')
+        y = self._obj.values
```
shoyer: This should use `self._obj.data` and raise `NotImplementedError` if it's a dask array.

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+
+    def _interpolate(self, kind='linear'):
+        from .dataarray import DataArray
+
+        new_x = self._full_index.values.astype('float')
+
+        arrs = []
+        # Apply the interpolation to each DataArray in our original Dataset
+        for v in self._obj.data_vars:
+            _da = self._obj[v].copy()
```
shoyer: No need to use an underscore prefix here for `_da`

xarray/tests/test_dataarray.py
```diff
@@ -1805,100 +1805,168 @@ def test_resample(self):
...
-        array = DataArray(np.arange(5), [('time', times)])
+        pass
 
-        expected_time = pd.date_range('2000-01-01', freq='12H', periods=9)
-        expected = array.reindex(time=expected_time)
-        for how in ['mean', 'median', 'sum', 'first', 'last', np.mean]:
-            actual = array.resample('12H', 'time', how=how)
-            self.assertDataArrayIdentical(expected, actual)
+    def test_upsample_interpolate(self):
+        # TODO: Implement me
```
shoyer: Yes, we definitely need this before add the interpolation code :)

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        return self._upsample('pad')
+    ffill = pad
+
+    def backfill(self):
+        """Backward fill new values at up-sampled frequency.
+        """
+        return self._upsample('backfill')
+    bfill = backfill
+
```
shoyer: It looks like we need a `nearest()` method to expose that functionality? Or can we just rely on the `interpolate` method?

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+    def interpolate(self, kind='linear'):
+        """Interpolate up-sampled data using the original data
+        as knots.
+
+        Parameters
+        ----------
+        kind : str {'linear', 'nearest', 'zero', 'slinear',
+               'quadratic', 'cubic'}
+            Interpolation method to use.
```
shoyer: Let's refer to the scipy documentation for `interp1d` for full details.

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+
+from . import ops
+from .combine import merge
+from .groupby import DataArrayGroupBy, DatasetGroupBy, GroupBy, _dummy_copy
+from .utils import maybe_wrap_array
+
+RESAMPLE_DIM = '__resample_dim__'
+
+
+class Resample(GroupBy):
```
shoyer: Let's not inherit from `GroupBy` here. We already inherit from `DataArrayGroupBy` and `DatasetGroupBy` in the resample subclasses, so this just makes the subclass hierarchy more complicated.

xarray/core/resample.py
```diff
@@ -0,0 +1,344 @@
...
+        # return result.rename({self._resample_dim: self._dim})
+
+    def _interpolate(self, kind='linear'):
+        from .dataarray import DataArray
+
+        new_x = self._full_index.values.astype('float')
+
+        arrs = []
+        # Apply the interpolation to each DataArray in our original Dataset
+        for v in self._obj.data_vars:
```
shoyer: This repetition is a little unfortunate but I don't really have a good alternative.

xarray/tests/test_dataarray.py
```diff
@@ -1805,100 +1805,168 @@ def test_resample(self):
...
-    def test_resample_first_keep_attrs(self):
+    def test_resample_bad_resample_dim(self):
         times = pd.date_range('2000-01-01', freq='6H', periods=10)
-        array = DataArray(np.arange(10), [('time', times)])
-        array.attrs['meta'] = 'data'
+        array = DataArray(np.arange(10), [('__resample_dim__', times)])
+        actual = array.resample('1D', dim='__resample_dim__', how='first')
+        self.assertRaisesRegexp(ValueError, 'Proxy resampling dimension')
+
+    # def test_resample_with_timezone(self):
```
shoyer: Fix this?

shoyer:
I like your approach here! Just some various comments on the implementation.

xarray/core/common.py
darothen: raising a `TypeError` instead of calling out to `self._resample_immediately()` will break backwards compatibility - there will be no way to use the old API. Do we want to make this break immediately?

xarray/core/resample.py
darothen: I added a method that farms out to `interpolate(kind='nearest')`. There's a slight inconsistency between `upsample('nearest')` and `interpolate(kind='nearest')` if an up-sampled point is exactly in between two of the original time coordinates; for whatever reason, `_upsample` (which relies on re-indexing) rounds that point towards the earlier knot, and the `interpolate` will round it up, which should be expected behavior. So I've removed 'nearest' as a valid option from `_upsample`.

xarray/core/resample.py
darothen: Is there a utility to do this automatically or does this need to be done manually? Skimmed **groupby.py** and a few others but couldn't really find anything.

xarray/core/resample.py
darothen: We do have to inject the **dim** argument to be `self._dim`, but I just call the super-method with that as an argument now.

xarray/tests/test_dataarray.py
darothen: Let's defer to a patch to fix #1490; it's going to take some other modifications outside of just re-sampling for timezones to work nicely. I'll remove this for now.

darothen:
All set except for my one question to @shoyer above. I've opted not to include a chart outlining the various upsampling options... couldn't really think of a nice and clean way to do so, because adding it to the time series doc page ends up being really ugly and there isn't quite enough substance for its own worked example page.

xarray/core/resample.py
shoyer: I think this needs to be done manually. In groupby operations we always apply things to sliced Dataset/DataArray objects, not arrays.

There is some similar logic in core/computation.py (see `build_output_coords()`).

shoyer:
> I've opted not to include a chart outlining the various upsampling options...

This is OK by me

darothen:
Resolved to drop auxiliary coordinates which are defined along the dimension to be re-sampled. This makes sense; if someone wants them to be interpolated or manipulated in some way, then they should promote them from coordinates to variables before doing the resampling.

In response to #1328, `count()` works just fine if you call it from a `Resample` object. Works for both resampling and up-sampling, but it will preserve the shape of the non-resampled dimensions. I think that's fine, because `count()` treats NaN as missing by default, so you can immediately know in which grid cells you have missing data :)

Final review, @shoyer, before merging in anticipation of 0.10.0?

doc/time-series.rst
```diff
@@ -15,6 +15,11 @@ core functionality.
     import numpy as np
     import pandas as pd
     import xarray as xr
+
+    # Use defaults so we don't get gridlines in generated docs
+    import matplotlib as mpl
+    mpl.rcdefaults()
```
shoyer: a simpler fix would be to update seaborn to v0.8, which no longer adjusts the default style

doc/time-series.rst
```diff
@@ -150,17 +155,41 @@ For example, we can downsample our dataset from hourly to 6-hourly:
...
+   The ``resample`` api was updated in version 0.10.0 to reflect similar
+   updates in pandas ``resample`` api to be more groupby-like. Older style
+   calls to ``resample`` will still be supported for a short period:
+
+   .. ipython:: python
+
+    ds.resample('6H', dim='time', how='mean')
+
+   Using the deprecated api, resample will also work for upsampling, in which
+   case intervals without any values are marked by ``NaN``:
```
shoyer: This makes it sound like you can't do upsampling with the new API.

I'm not quite sure what the point of this sentence is -- maybe drop it?

doc/whats-new.rst
```diff
@@ -222,6 +229,10 @@ Enhancements
 
 - New aggregation on rolling objects :py:meth:`DataArray.rolling(...).count()`
   which providing a rolling count of valid values (:issue:`1138`).
+- :py:meth:`~DataArray.resample` and :py:meth:`~Dataset.resample` have been
```
shoyer: this didn't make it into v0.9.2 :)

doc/whats-new.rst
```diff
@@ -25,6 +25,13 @@ Enhancements
   raster files are opened with :py:func:`~xarray.open_rasterio`.
   By `Greg Brener <https://github.com/gbrener>`_
 
+- A new resampling interface to match pandas' group-by-like API was added to
```
shoyer: After you merge in master, please move this into the "Breaking changes" section.

doc/whats-new.rst
```diff
@@ -25,6 +25,13 @@ Enhancements
...
+  :py:meth:`~xarray.Dataset.resample` and :py:meth:`~xarray.DataArray.resample`.
```
shoyer: Can you add a code snippet to show the difference? e.g., something like:

    Old syntax:

    In [1]: ds.resample('24H', dim='time', how='max')`
    [..]

    New syntax:

    In [2]: ds.resample(time='24H').max()
    [...]

The code doesn't even need to run.

Also please note that the old syntax is still supported for now, but with a warning.

doc/whats-new.rst
```diff
@@ -25,6 +25,13 @@ Enhancements
...
+  Timeseries resampling is fully supported for data with arbitrary dimensions
+  as is both downsampling and upsampling (including linear, quadratic, cubic,
+  and spline interpolation).
```
shoyer: This would be a nice place to link to section on resampling in the time series docs.

xarray/core/resample.py
shoyer: Hmm. I think it might be a little surprising that it deviates from the behavior of `reindex`/`sel`, but given the inconsistency between pandas and scipy it will be surprising to someone no matter which choice you make. I would lean towards the other way (to mimic pandas and make use of indexing, since it works on dask arrays), but I'm OK with other choice as long as you document it. Can you add a brief noting on the rounding behavior?

xarray/core/resample.py
```diff
@@ -0,0 +1,339 @@
...
+        """Interpolate up-sampled data using the original data
+        as knots.
+
+        Parameters
+        ----------
+        kind : str {'linear', 'nearest', 'zero', 'slinear',
+               'quadratic', 'cubic'}
+            Interpolation method to use; refer to to
+            `scipy.interpolate.interp1d` for full details.
+
```
shoyer: Use a `See also` section to reference `scipy.itnerpolate.interp1d` -- these get linked as part of the docs build.

xarray/core/resample.py
```diff
@@ -0,0 +1,339 @@
...
+            x = da[self._dim].astype('float')
+            y = da.values
+            axis = da.get_axis_num(self._dim)
+
+            f = interp1d(x, y, kind=kind, axis=axis, bounds_error=True,
+                         assume_sorted=True)
+
+            # drop any existing non-dimension coordinates along the resampling
+            # dimension
+            coords = OrderedDict()
```
shoyer: This works, but given that `coords` are shared between all data variables on a Dataset, it would be a little cleaner to only do this once.

Possibly we can leave this one to a follow-up if someone is annoyed by slow performance here. Can you add a comment noting that the lower-level Variable API could be used to speed things up here?

xarray/tests/test_dataarray.py
```diff
@@ -1805,100 +1805,240 @@ def test_resample(self):
...
+        # As frequency
+        actual = array.resample(time='3H').asfreq()
+        expected = DataArray(array.to_series().resample('3H').asfreq())
+        self.assertDataArrayIdentical(expected, actual)
+
+        # Pad
+        actual = array.resample(time='3H').pad()
+        expected = DataArray(array.to_series().resample('3H').pad())
+        self.assertDataArrayIdentical(expected, actual)
+
```
shoyer: Add a unit test for `nearest()`

xarray/core/resample.py
darothen: I definitely see the reasoning towards leaning towards pandas, so I rolled back the change to use `reindex` instead of `interpolate`. I'll also add a note in the time series documentation on this.

xarray/core/resample.py
darothen: Re-factored to do once prior to looping over all our variables to interpolate them. Looks like it works just fine, although if there's a cleaner way to get a "test" DataArray from a Dataset, that might make things a little bit prettier

xarray/core/resample.py
```diff
@@ -304,31 +307,36 @@ def _interpolate(self, kind='linear'):
...
 
-            f = interp1d(x, y, kind=kind, axis=axis, bounds_error=True,
-                         assume_sorted=True)
+        # Prepare coordinates by dropping non-dimension coordinates along the
+        # resampling dimension.
+        # note: the lower-level Variable API could be used to speed this up
+        coords = OrderedDict()
+        if self._obj.data_vars:
+            var = list(self._obj.data_vars.keys())[0]
+            da = self._obj[var].copy()
```
shoyer: I'm not certain this version is correct anymore (so maybe restore what you had before).

If you want to go down the path of selecting coordinates once, you should iterate over `Dataset.coords` (not `DataArray.coords`) to get an `OrderedDict()`, and then merge it with the interpolated data variables all at once by calling the `Dataset()` constructor instead of `merge`.

darothen: Ah, thanks for catching that. Pushing a few more commits that finish some of the documentation, and take this into consideration. I do want to do the coordinate selection all at once... but maybe we should go ahead and raise an issue to "enhance" this using Variable API, as you mention, if people run into issues.

shoyer: I think something like this would be work:
```python
old_times = self._obj[self._dim].astype(float)
new_times = self._full_index.values.astype(float)

data_vars = OrderedDict()
coords = OrderedDict()
for name, variable in self._obj.variables.items():
    if name in self._obj.coords:
        if name == self._dim:
            coords[self._dim] = self._full_index
        elif self._dim not in v.dims:
            coords[name] = variable
    else:
         if isinstance(variable.data, dask_array_type):
             raise TypeError('dask arrays not supported yet in resample().interpolate()')
         f = interp1d(old_times, variable.data, kind=kind, ....)
         interpolated = Variable(variable.dims, f(new_times), variable.attrs)
         data_vars[name] = interpolated
return Dataset(data_vars, coords)

File: xarray/core/resample.py [hunk 36]
```diff
@@ -304,31 +307,36 @@ def _interpolate(self, kind='linear'):
...
+        # Apply the interpolation to each DataArray in our original Dataset
+        for var in self._obj.data_vars:
+            da = self._obj[var].copy()
+
+            x = da[self._dim].astype('float')
+            y = da.values
+            axis = da.get_axis_num(self._dim)
+
+            f = interp1d(x, y, kind=kind, axis=axis, bounds_error=True,
+                         assume_sorted=True)
```
shoyer: Do we check somewhere earlier that `assume_sorted=True` is valid? (If not, we'll need to check `Index.is_monotonic_increasing`)

darothen: Yup! That gets checked by the logic we already use to build `GroupBy` objects, which powers resampling under the hood. See [here](https://github.com/pydata/xarray/blob/master/xarray/core/groupby.py#L234). If you try to build a non-monotonic timeseries, this error will raise long before we do any calculations.

xarray/core/resample.py
darothen: Oh that's much more succinct and nicer than my code. Hopefully it didn't take you much time to write! I made a few tweaks to it and dropped it into my code.

xarray/core/resample.py
shoyer: Sounds good, maybe add a quick comment noting that we already checked this in `__init__`? Since I already forgot :)

xarray/core/resample.py
shoyer: did you push those commits yet? I'm not seeing it yet

darothen:
Try refreshing? Latest commit is 7a767d8 and has all these changes plus
some more tweaks.

*Daniel Rothenberg *
Postdoctoral Research Associate
Center for Global Change Science
Massachusetts Institute of Technology
*A: *77 Massachusetts Ave | E18-402A
     Cambridge, MA 02139
*T: *(502) 648-7513; *T: *(617) 258-0407
*E: *darothen@mit.edu
*H: *www.danielrothenberg.com
<http://github.com/darothen> [image:
http://www.linkedin.com/in/rothenbergdaniel/]
<http://www.linkedin.com/in/rothenbergdaniel/>
<http://www.twitter.com/danrothenberg>

xarray/core/resample.py
shoyer: It looks like someone is indeed working on this for the next pandas release: https://github.com/pandas-dev/pandas/issues/17496

xarray/core/resample.py
```diff
@@ -0,0 +1,342 @@
...
+
+        # Apply the interpolation to each DataArray in our original Dataset
+        for name, variable in self._obj.variables.items():
+            if name in self._obj.coords:
+                if name == self._dim:
+                    coords[self._dim] = self._full_index
+                elif self._dim not in variable.dims:
+                    coords[name] = variable
+                else:
+                    if isinstance(variable.data, dask_array_type):
```
shoyer: Do we really want to interpolate coordinate variables? At the least, we shouldn't do this for non-numeric coordinates.

xarray/core/resample.py
```diff
@@ -0,0 +1,342 @@
...
+                        raise TypeError('dask arrays not supported yet in '
+                                        'resample.interpolate()')
```
shoyer: Let's make this error message more descriptive here: https://github.com/pydata/xarray/issues/1566

darothen:
Hmmm. Something is really screwy with my feature branch and making the task of cleaning up the merge difficult. I'll work on fixing this.

darothen:
@shoyer fixed.

xarray/core/resample.py
```diff
@@ -0,0 +1,352 @@
...
+                elif self._dim not in variable.dims:
+                    coords[name] = variable
+            else:
+                if isinstance(variable.data, dask_array_type):
+                    raise TypeError(
+                        "Up-sampling via interpolation was attempted on the the "
+                        "variable '{}', but it is a dask array; dask arrays not "
+                        "yet supprted in resample.interpolate(). Load into "
+                        "memory with Dataset.load() before resampling."
+                        .format(name)
```
shoyer: love it -- thanks!

xarray/core/resample.py
```diff
@@ -0,0 +1,352 @@
...
+
+    def interpolate(self, kind='linear'):
+        """Interpolate up-sampled data using the original data
+        as knots.
+
+        Parameters
+        ----------
+        kind : str {'linear', 'nearest', 'zero', 'slinear',
+               'quadratic', 'cubic'}
+            Interpolation scheme to use
```
shoyer: add a period

xarray/core/common.py
```diff
@@ -541,24 +520,75 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+            resample_dim = "resampled_" + dim_name
+        else:
+            raise TypeError("Dimension name should be a string; "
+                            "was passed %r" % dim)
+        group = DataArray(dim, [(dim.dims, dim)], name=resample_dim)
+        time_grouper = pd.TimeGrouper(freq=freq, closed=closed,
+                                      label=label, base=base)
+        resampler = self._resample_cls(self, group=group, dim=dim_name,
+                                      grouper=time_grouper,
+                                      resample_dim=resample_dim)
```
jhamman: fix indentation

xarray/core/common.py
```diff
@@ -541,24 +520,75 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+
+        warnings.warn("\n.resample() has been modified to defer "
+                      "calculations. Instead of passing 'dim' and "
+                      "'how=\"{how}\", instead consider using "
+                      ".resample({dim}=\"{freq}\").{how}() ".format(
+                            dim=dim, freq=freq, how=how
+                      ), DeprecationWarning, stacklevel=3)
+
         if isinstance(dim, basestring):
+            dim_name = dim
```
jhamman: `dim_name` is never used

xarray/core/common.py
```diff
@@ -541,24 +520,75 @@ def resample(self, freq, dim, how='mean', skipna=None, closed=None,
...
+            elif how == 'count':
+                result = f(dim=dim.name, keep_attrs=keep_attrs)
             else:
                 result = f(dim=dim.name, skipna=skipna, keep_attrs=keep_attrs)
         else:
             result = gb.reduce(how, dim=dim.name, keep_attrs=keep_attrs)
         result = result.rename({RESAMPLE_DIM: dim.name})
         return result
 
+
```
jhamman: PEP8: remove extra blank line

xarray/core/dataarray.py
```diff
@@ -34,7 +35,7 @@ def _infer_coords_and_dims(shape, coords, dims):
     """All the logic for creating a new DataArray"""
 
     if (coords is not None and not utils.is_dict_like(coords) and
-            len(coords) != len(shape)):
+                len(coords) != len(shape)):
```
jhamman: fix indentation

xarray/core/resample.py
```diff
@@ -0,0 +1,352 @@
...
+        from .dataarray import DataArray
+
+        if isinstance(self._obj.data, dask_array_type):
+            raise TypeError(
+                "Up-sampling via interpolation was attempted on the the "
+                "variable '{}', but it is a dask array; dask arrays not "
+                "yet supprted in resample.interpolate(). Load into "
+                "memory with Dataset.load() before resampling."
+                .format(name)
+            )
```
jhamman: Can we add a unit test that covers this `TypeError`?  Right now, I think you'll actually get a `NameError` because `name` is not defined.

xarray/core/resample.py
```diff
@@ -0,0 +1,352 @@
...
+            # is the resampling dimension
+            if k == self._dim:
+                coords[self._dim] = self._full_index
+            # else, check if resampling dim is in coordinate dimensions
+            elif self._dim not in v.dims:
+                coords[k] = v
+        return DataArray(f(new_x), coords, dims, name=dummy.name,
+                         attrs=dummy.attrs)
+
+ops.inject_reduce_methods(DataArrayResample)
```
jhamman: PEP8: add a second empty line after class definition

xarray/tests/test_dataarray.py
```diff
@@ -1836,100 +1836,249 @@ def test_resample(self):
...
+        self.assertDataArrayIdentical(new_mean, expected)
+
+        # Try other common resampling methods
+        resampler = array.resample(time='1D')
+        for method in ['mean', 'median', 'sum', 'first', 'last', 'count']:
+            # Discard attributes on the call using the new api to match
+            # convention from old api
+            new_api = getattr(resampler, method)(keep_attrs=False)
+            with pytest.warns(DeprecationWarning):
+             old_api = array.resample('1D', dim='time', how=method)
```
jhamman: indentation

xarray/tests/test_dataarray.py
```diff
@@ -1836,100 +1836,249 @@ def test_resample(self):
...
-        self.assertEqual(expected, actual)
-
-        resampled_array = array.resample('1D', dim='time', how='first',
-                                         keep_attrs=False)
-        assert resampled_array.attrs == {}
-
-    def test_resample_mean_keep_attrs(self):
+        array = DataArray(np.arange(10), [('__resample_dim__', times)])
+        actual = array.resample('1D', dim='__resample_dim__', how='first')
+        self.assertRaisesRegexp(ValueError, 'Proxy resampling dimension')
```
jhamman: I think this needs to be:

```Python
with self.assertRaisesRegexp(ValueError, 'Proxy resampling dimension'):
    actual = array.resample('1D', dim='__resample_dim__', how='first')
```

I'm not sure why this is passing actually.

xarray/tests/test_dataarray.py
```diff
@@ -1836,100 +1836,249 @@ def test_resample(self):
...
+        xx, yy = np.meshgrid(xs*5, ys*2.5)
+        tt = np.arange(len(times), dtype=int)
+        array = DataArray(data,
+                          {'time': times, 'x': xs, 'y': ys},
+                          ('x', 'y', 'time'))
+        xcoord = DataArray(xx.T, {'x': xs, 'y': ys}, ('x', 'y'))
+        ycoord = DataArray(yy.T, {'x': xs, 'y': ys}, ('x', 'y'))
+        tcoord = DataArray(tt, {'time': times}, ('time', ))
+        ds = Dataset({'data': array, 'xc': xcoord,
+                         'yc': ycoord, 'tc': tcoord})
```
jhamman: indentation

xarray/tests/test_dataset.py
```diff
@@ -2238,11 +2243,57 @@ def test_resample_by_last_discarding_attrs(self):
...
+        xx, yy = np.meshgrid(xs*5, ys*2.5)
+        tt = np.arange(len(times), dtype=int)
+        array = DataArray(data,
+                          {'time': times, 'x': xs, 'y': ys},
+                          ('x', 'y', 'time'))
+        xcoord = DataArray(xx.T, {'x': xs, 'y': ys}, ('x', 'y'))
+        ycoord = DataArray(yy.T, {'x': xs, 'y': ys}, ('x', 'y'))
+        tcoord = DataArray(tt, {'time': times}, ('time', ))
+        ds = Dataset({'data': array, 'xc': xcoord,
+                         'yc': ycoord, 'tc': tcoord})
```
jhamman: indentation

darothen:
@jhamman Gotcha, I'll clean everything up by the end of the week. If that's going to block 0.10.0, let me know and I'll shuffle some things around to prioritize this.

jhamman:
@darothen - we have a few other PRs to wrap up for 0.10 so end of the week is okay.

xarray/tests/test_dataarray.py
darothen: This was butchered in more ways than one; I fixed the test which revealed some underlying flaws, too, which are also fixed.

darothen:
@jhamman Think we're good. I deferred 4 small pep8 issues because they're in parts of the codebase which I don't think I ever touched, and i'm worried they're going to screw up the merge.

jhamman:
Thanks @darothen - can you resolve the merge conflicts?

darothen:
@jhamman done - caught me right while I was compiling GEOS-Chem, and the merge conflicts were very simple.

xarray/tests/test_dataarray.py
```diff
@@ -1836,100 +1836,267 @@ def test_resample(self):
...
+        expected_data = np.repeat(data, 2, axis=-1)
+        expected_data[..., 1::2] = expected_data[..., ::2]
+        expected_data = expected_data[..., :-1]
+        expected_times = times.to_series().resample('3H').asfreq().index
+        expected = DataArray(expected_data,
+                             {'time': expected_times, 'x': xs, 'y': ys},
+                             ('x', 'y', 'time'))
+        self.assertDataArrayIdentical(expected, actual)
+
+    def test_upsample_interpolate(self):
```
jhamman: this needs a `@requires_scipy` decorator or put

```Python
interp1d = pytest.importorskip('scipy.interpolate.interp1d')
```
in the first line of the test (instead of the import)

xarray/core/resample.py
```diff
@@ -0,0 +1,353 @@
...
+
+        if self._resample_dim in combined.dims:
+            combined = combined.rename({self._resample_dim: self._dim})
+
+        return combined
+
+    def _interpolate(self, kind='linear'):
+        """Apply scipy.interpolate.interp1d along resampling dimension."""
+        from .dataarray import DataArray
+
```
jhamman: let's import `interp1d` here, instead of at the module level. `scipy` is still a optional dependency.

darothen: Gotcha - it needed to be imported under `DatasetResample`, too.

jhamman:
@darothen - almost there. Two or three more dependency conflicts in the tests.

darothen:
@jhamman Ohhh i totally misunderstood the last readout from travis-ci. Dealing with the scipy dependency is easy enough. However, another test fails because it uses [`np.flip()`](https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.flip.html) which wasn't added to numpy until v1.12.0. Do we want to bump the numpy version in the dependencies? Or is there another aproach to take here?

shoyer:
We just bumped numpy to 1.11, but 1.12 would be too new.

Let's just add a `np.flip` backport to `core/npcompat.py`. The [whole function](https://github.com/numpy/numpy/blob/master/numpy/lib/function_base.py#L140-L209) is only a couple of lines.

jhamman:
In it goes. Thanks @darothen!

PR #1278, opened 2017-02-20
Feature/support bn 1 1

### LINKED ISSUE #1276: DataArray doesnt use bottleneck implementation on rolling variance
It appears that the DataArray is not using the bottleneck move_var when calling var on a rolling object but it does for std. I have attached some simple benchmarks to demonstrate. 

python version 3.6
pandas version 0.19.2
xarray version 0.9.1
bottleneck version 1.2.0

```python
import numpy as np
import xarray as xr
import bottleneck
import pandas as pd

data = np.random.rand(10000)

# xarray setup
ar = xr.DataArray(data, dims=['time'])
xrol = ar.rolling(time=30)

# pandas setup
df = ar.to_pandas()
prol = df.rolling(30)
```

```python
# xray std profile
%time _ = xrol.std()
```

    CPU times: user 0 ns, sys: 0 ns, total: 0 ns
    Wall time: 209 µs

```python
# xray var profile
%time _ = xrol.var()
```

    CPU times: user 8.69 s, sys: 132 ms, total: 8.82 s
    Wall time: 8.52 s

```python
# pandas std profile
%time _ = prol.std()
```

    CPU times: user 0 ns, sys: 0 ns, total: 0 ns
    Wall time: 1.13 ms

```python
# pandas var profile
%time _ = prol.var()
```

    CPU times: user 0 ns, sys: 0 ns, total: 0 ns
    Wall time: 617 µs

```python
# bottleneck std profile
%time _ = bottleneck.move_std(data, 30)
```

    CPU times: user 0 ns, sys: 0 ns, total: 0 ns
    Wall time: 138 µs

```python
# bottleneck var profile
%time _ = bottleneck.move_var(data, 30)
```

    CPU times: user 0 ns, sys: 0 ns, total: 0 ns
    Wall time: 143 µs

jhamman:
Bottleneck added support for rolling variance calculations (`bn.move_var`) as well as a few other functions in version 1.1.  This PR conditionally uses those new functions, when they are available.

 - [x] closes #1276
 - [x] tests added / passed
 - [x] whatsnew entry

cc @shoyer and @d-chambers

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -52,7 +52,9 @@ NAN_REDUCE_METHODS = ['argmax', 'argmin', 'max', 'min', 'mean', 'prod', 'sum',
 NAN_CUM_METHODS = ['cumsum', 'cumprod']
 BOTTLENECK_ROLLING_METHODS = {'move_sum': 'sum', 'move_mean': 'mean',
                               'move_std': 'std', 'move_min': 'min',
-                              'move_max': 'max'}
+                              'move_max': 'max', 'move_var': 'var',
+                              'move_argmin': 'argmin', 'move_argmax': 'argmax',
+                              'move_rank': 'rank'}
 # TODO: wrap take, dot, sort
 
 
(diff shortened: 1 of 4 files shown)


xarray/core/ops.py
```diff
@@ -52,7 +52,9 @@
 NAN_CUM_METHODS = ['cumsum', 'cumprod']
 BOTTLENECK_ROLLING_METHODS = {'move_sum': 'sum', 'move_mean': 'mean',
                               'move_std': 'std', 'move_min': 'min',
-                              'move_max': 'max'}
+                              'move_max': 'max', 'move_var': 'var',
+                              'move_argmin': 'argmin', 'move_argmax': 'argmax',
+                              'move_rank': 'rank'}
```
shoyer: I don't think we have a `rank` method `DataArray`/`Dataset`, so we should probably remove this for now for consistency (though it would be nice to add it later!). Also I think it's untested.

xarray/core/ops.py
```diff
@@ -520,24 +522,40 @@ def inject_bottleneck_rolling_methods(cls):
...
         func.__name__ = name
-        func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(name=func.__name__)
+        func.__doc__ = _ROLLING_REDUCE_DOCSTRING_TEMPLATE.format(
+            name=func.__name__)
         setattr(cls, name, func)
 
     # bottleneck rolling methods
     if has_bottleneck:
-        if LooseVersion(bn.__version__) < LooseVersion('1.0'):
+        bn_version = LooseVersion(bn.__version__)
```
shoyer: It would be nice to drop old versions of bottleneck when feasible -- all this `LooseVersion` stuff is really ugly and error prone (hard to test). But this is fine for now.

jhamman: Should we move the minimum bottleneck version to 1.1 and just get rid of all this then?

shoyer: Let's add a TODO to do that before releasing v0.10. But for now, we should keep it rather than updating the required version in a patch release (supposing we issue a v0.9 release).

jhamman: okay, done.

PR #1286, opened 2017-02-23
add assign_attrs method

### LINKED ISSUE #1281: update_attrs method
Updating attrs is an operation that consistently breaks up my method chaining. This should be easy to solve with something like

````
def update_attrs(self, *args, **kwargs):
    self.attrs.update(*args, **kwargs)
    return self
````

Thoughts?

edit: I guess it should return a modified object rather than acting as an in-place operation...is copying appropriate here? How is this usually handled?

### PR DESCRIPTION
**Pull request by hsharrison**
- [x] closes #1281 
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master | flake8 --diff`` (except for whatsnew... wasn't sure there)
 - [x] whatsnew entry

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -346,6 +346,28 @@ class BaseDataObject(AttrAccessMixin):
         data.coords.update(results)
         return data
 
+    def assign_attrs(self, *args, **kwargs):
+        """Assign new attrs to this object, returning a new object equivalent
+        to self.attrs.update(*args, **kwargs).
+
+        Parameters
+        ----------
+        args : positional arguments passed into ``attrs.update``.
+        kwargs : keyword arguments passed into ``attrs.update``.
+
+        Returns
+        -------
+        assigned : same type as caller
+            A new object with the new attrs in addition to the existing data.
+
+        See also
+        --------
+        Dataset.assign
+        """
+        out = self.copy(deep=False)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: xarray/core/common.py [hunk 1]
```diff
@@ -346,6 +346,28 @@ def assign_coords(self, **kwargs):
         data.coords.update(results)
         return data
 
+    def assign_attrs(self, *args, **kwargs):
+        """Assign new attrs to this object, returning a new object equivalent
```
shoyer: Can you move the second half of this to a new line? We try to keep the title to one line.

hsharrison: I'm sorry, I don't follow. I tried to match the conventions of the rest of the file. Do you mean this:

```
def assign_attrs(self, *args, **kwargs):
    """
    Assign new attrs to this object, returning a new object equivalent
    to...
```

or

```
def assign_attrs(self, *args, **kwargs):
    """Assign new attrs to this object.

    Returns a new object equivalent to...
```

shoyer: The second version is preferred. But you're right that `assign_coords` also has a docstring that violates this.

shoyer: If you like, it would be awesome to clean that up, too.

shoyer:
Awesome, thank you!

PR #1277, opened 2017-02-19
Restored dim order in DataArray.rolling().reduce() 

fujiisoup:
- [x] closes #1125 
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master | flake8 --diff``
 - [x] whatsnew 

Added 1 line to fix #1125.

I hope this is enough.
If another care is necessary, please let me know.

PR #1289, opened 2017-03-02
 Added a support for Dataset.rolling.

### LINKED ISSUE #859: BUG: Rolling on Dataset
This looks like it's available with `dir` / tab complete, but actually isn't:

``` python
In [13]: xr.DataArray(np.random.rand(10,3)).to_dataset('dim_1').rolling
<ipython-input-13-438d3638a0d0> in <module>()
----> 1 xr.DataArray(np.random.rand(10,3)).to_dataset('dim_1').rolling

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xarray/core/common.py in __getattr__(self, name)
    135                     return source[name]
    136         raise AttributeError("%r object has no attribute %r" %
--> 137                              (type(self).__name__, name))
    138 
    139     def __setattr__(self, name, value):

AttributeError: 'Dataset' object has no attribute 'rolling'
```

I _think_ this could be easy to implement as an `.apply` operation? (indeed, that could be a reasonable path for a whole host of operations - i.e. try and apply them to each array in the ds?)

Also, as a very narrow point, I'm not sure why `.rolling_cls` is public? Probably should be private?

Finally, the Rolling implementation is pretty sweet. I've been getting my hands dirty in the pandas one recently, and that we can have something as well featured as that with so few lines of code 👍

fujiisoup:
- [x] closes #859
 - [x] tests added / passed
 - [x] whatsnew entry

There seems to be two approaches to realize Dataset.rolling,
1. Apply rolling in each DataArrays and then combine them.
2. Apply Dataset directoly with some DataArrays that do not depend on `dim` kept aside, then merge them later.

I chose the latter approach to reuse existing `Rolling` object as much as possible,
but it results in some duplicates in `ImplementsRollingDatasetReduce`.
Any feedbacks and comments are very welcome.


xarray/core/dataset.py
```diff
@@ -307,6 +308,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
     coordinates used for label based indexing.
     """
     groupby_cls = groupby.DatasetGroupBy
+    rolling_cls = rolling.DatasetRolling
```
MaximilianR: I think this should be private (I know `groupby_cls` is public, but I think that is a mistake)

xarray/core/rolling.py
```diff
@@ -190,4 +194,65 @@ class DataArrayRolling(Rolling, ImplementsRollingArrayReduce):
...
+        """
+        from .dataset import Dataset
+        dim, window = next(iter(windows.items()))
+        # DataArrays not depending on the rolling-axis are kept aside from obj
+        self.fixed_ds = Dataset({key: da for key, da in obj.data_vars.items()
+                                 if dim not in da.dims},
+                                obj.coords)
+        # drop them from obj, then call parent's initializer.
+        obj = obj.drop(self.fixed_ds.data_vars.keys())
+        Rolling.__init__(self, obj, min_periods, center, **windows)
```
MaximilianR: Very narrow point, but this should probably be `super`

fujiisoup:
@MaximilianR 

Thanks for the review.
I made `rolling_cls` and `groupby_cls` private, and added some docstrings.

Also, I removed `_axis_num` property in Rolling class to make it compatible with Dataset.
It can be replaced by da.get_axis_num(self.dim).

xarray/core/rolling.py
```diff
@@ -192,4 +191,76 @@ class DataArrayRolling(Rolling, ImplementsRollingArrayReduce):
...
+    def reduce(self, func, **kwargs):
+        if self.center:
+            fixed = self._center_result(self.fixed_ds)
+        else:
+            fixed = self.fixed_ds
+        # call parent's reduce
+        reduced = super(DatasetRolling, self).reduce(func, **kwargs)
+        # Need to restore dimension order
+        for key, var in reduced.data_vars.items():
+            reduced = reduced.assign(**{key:
```
shoyer: I think this is unnecessary now, after #1277.

xarray/core/rolling.py
```diff
@@ -192,4 +191,76 @@ class DataArrayRolling(Rolling, ImplementsRollingArrayReduce):
...
+                Size of the moving window.
+
+        Returns
+        -------
+        rolling : type of input argument
+        """
+        from .dataset import Dataset
+        dim, window = next(iter(windows.items()))
+        # DataArrays not depending on the rolling-axis are kept aside from obj
+        self.fixed_ds = Dataset({key: da for key, da in obj.data_vars.items()
```
shoyer: Use an `OrderedDict` here instead of a dict comprehension to preserve the order of `data_vars`.

xarray/core/common.py
```diff
@@ -126,6 +169,76 @@ def wrapped_func(self, **kwargs):
...
+
+            if self.min_periods is not None:
+                raise ValueError('Rolling.median does not accept min_periods')
+
+            if any(isinstance(da, dask_array_type) for _, da
+                   in self.obj.data_vars.items()):
+                raise NotImplementedError(
+                    'Rolling window operation does not work with dask arrays')
+
+            result = Dataset({key: (da.dims,
```
shoyer: Also needs to use `OrderedDict`.

xarray/core/common.py
```diff
@@ -70,14 +70,51 @@ def wrapped_func(self, dim=None, keep_attrs=False, **kwargs):
 
 
 class ImplementsRollingArrayReduce(object):
```
shoyer: I recognize that you didn't write this class, but I wonder if we couldn't simply consolidate it into `DataArrayRolling`? It feels like one more level of complexity than is necessary. The `ImplementsArrayReduce` type classes were implemented because we have multiple array types for which we wanted similar reduce functionality.

Either way, it would be nice to move this logic into the `rolling.py` module, just to keep it in fewer places.

Perhaps @jhamman recalls why we have this separate class...

xarray/core/common.py
```diff
@@ -126,6 +169,76 @@ def wrapped_func(self, **kwargs):
...
+                    'Rolling window operation does not work with dask arrays')
+
+            # bottleneck doesn't allow min_count to be 0, although it should
+            # work the same as if min_count = 1
+            if self.min_periods is not None and self.min_periods == 0:
+                min_count = self.min_periods + 1
+            else:
+                min_count = self.min_periods
+
+            result = Dataset({key: (da.dims,
```
shoyer: Needs to use `OrderedDict` here to preserve data variable order.

doc/whats-new.rst
```diff
@@ -21,6 +21,8 @@ v0.9.2 (unreleased)
 
 Enhancements
 ~~~~~~~~~~~~
+- ``rolling`` on Dataset is now supported (:issue:`859`).
+  By `Keisuke Fujii <https://github.com/fujiisoup>`_.
```
shoyer: Please also add `Dataset.rolling` to `doc/api.rst`.

fujiisoup:
@shoyer Thanks for the review.

I didn't notice Dataset keeps the order of `data_vars`.
I think I should change the direction, because the data_var's order is destroyed after the final `merge` call.

Maybe `DatasetRolling` should not be a child of `Rolling`, but an independent class that has an `OrderedDict` of `Rolling` for each `data_var` to preserve the variable order.

xarray/core/rolling.py
fujiisoup: I will change the `DatasetRolling` structure to keep `data_var`s order.
My code destroys this order in `reduce` method by calling `ds.merge`.

xarray/core/common.py
fujiisoup: Agreed. 
I moved this class into `DataArrayRolling` in the next commit.

fujiisoup:
+ Now DatasetRolling.reduce() keeps its original `data_vars` order. (test: )
+ Dataset.rolling is added to api.rst.

I updated `DatasetRolling` so that its `reduce` method relies mostly on `DataArrayRolling.reduce` rather than `window.reduce`.
This change makes not only the reduced-Dataset keep its original order of `data_var`, but also makes the code much cleaner.
I moved many methods from `Rolling` to `DataArrayRolling` because `DatasetRolling` does not use them anymore.

However, I guess `Rolling` class is originally designed to be used also for `Dataset`, as discussed in #859. 
If someone has any requests, I will reconsider the class structure.

xarray/tests/test_dataset.py
```diff
@@ -3261,3 +3264,132 @@ def test_dir_unicode(data_set):
...
+    rolling_obj = ds.rolling(time=window, center=center,
+                             min_periods=min_periods)
+
+    # add nan prefix to numpy methods to get similar behavior as bottleneck
+    actual = rolling_obj.reduce(getattr(np, 'nan%s' % name))
+    expected = getattr(rolling_obj, name)()
+    assert_allclose(actual, expected)
+    assert ds.dims == actual.dims
+    # make sure the order of data_var are not changed.
+    assert list(ds.data_vars.keys()) == list(actual.data_vars.keys())
```
fujiisoup: Here is the test to make sure that the order of `data_vars` is not changed by `rolling().reduce()`.

shoyer:
> I didn't notice Dataset keeps the order of data_vars.

It's not a strict requirement, but we try to do it when possible because it leads to more predictable results.

fujiisoup:
>  we try to do it when possible because it leads to more predictable results.

OK. I totally agree with this.
(Today I experienced a bug caused by unexpectedly aligned data_vars in other project. I realized predictable behavior is really important :) )

I think this PR is ready for review. If anyone has comments about the following behavior, I'm happy to update the code.
+ Only data_vars depending on the given rolling-dim are reduced. The others remain unchanged but are still stored in the reduced dataset (not dropped).
+ If rolling-dim is not included in `Dataset.dims`, this class raises `KeyErrror`.
+ The order of data_vars remain unchanged even after `reduce()`.
+ All the attrs are dropped.

jhamman:
@fujiisoup - This is looking pretty good. I'll let @MaximilianR give the final review since he's spent more time here than I have. As for your four points above, I think I'm fine with all of those design decisions.

One minor point though (no change needed)

> All the attrs are dropped

I don't think this is true for data_vars that don't have rolling-dim in them.

xarray/core/rolling.py
```diff
@@ -236,31 +330,62 @@ def __init__(self, obj, min_periods=None, center=False, **windows):
...
+                reduced[key] = self.obj[key]
+        return Dataset(reduced, coords=self.obj.coords)
+
+    @classmethod
+    def _reduce_method(cls, func):
+        """
+        Return a wrapped function for injecting numpy and bottoleneck methods.
+        see ops.inject_datasetrolling_methods
+        """
+        def wrapped_func(self, **kwargs):
```
shoyer: Can you make a single helper function that takes `func` and `dataset_rolling` as arguments and applies this logic? e.g.,
```
def _reduce_rollings(rolling, func):
    reduced = OrderedDict()
    for key, da in self.obj.data_vars.items():
        if self.dim in da.dims:
            reduced[key] = func(self.rollings[key])
        else:
            reduced[key] = self.obj[key]
    return Dataset(reduced, coords=self.obj.coords)
```
Then both this method and `reduce` can reuse the same logic, just with a different callable (e.g., `lambda x: x.reduce(func, **kwargs)` for reduce).

xarray/core/ops.py
```diff
@@ -560,3 +560,14 @@ def inject_bottleneck_rolling_methods(cls):
...
         setattr(cls, 'median', func)
+
+
+def inject_datasetrolling_methods(cls):
+    # standard numpy reduce methods
+    methods = [(name, globals()[name]) for name in NAN_REDUCE_METHODS]
+    for name, f in methods:
+        func = cls._reduce_method(f)
+        func.__name__ = name
+        func.__doc__ =\
```
shoyer: prefer line breaking after parentheses, if possible, like in `inject_bottleneck_rolling_methods` above.

xarray/core/rolling.py
fujiisoup: @shoyer 
Thank you for the review.
Could you give me more detail about the helper function?
I am wondering how I could distinguish the two types of reducing, 
`rolling.reduce(np.mean)` and `rolling.mean()`, the latter uses bottleneck if it is available.

shoyer: Honestly, maybe this isn't worth the trouble here -- the redundant lines are right next to each other already.

MaximilianR:
> I'll let @MaximilianR give the final review 

Apologies for missing this. I actually think @jhamman has done much more than I to contribute to this area!

But this change looks good, thanks @fujiisoup !

fujiisoup:
@shoyer 
Thank you for the reply.
I did not yet think of smarter solution.
I hope the current implementation is acceptable. 
If I come up with cleaner script, I will send another PR.

xarray/tests/test_dataarray.py
```diff
@@ -1352,20 +1352,6 @@ def test_reduce_keep_attrs(self):
         self.assertEqual(len(vm.attrs), len(self.attrs))
         self.assertEqual(vm.attrs, self.attrs)
 
-    def test_assign_attrs(self):
```
shoyer: Did you intend to delete this test?

xarray/tests/test_dataarray.py
```diff
@@ -2544,4 +2530,3 @@ def test_rolling_reduce(da, center, min_periods, window, name):
     actual = rolling_obj.reduce(getattr(np, 'nan%s' % name))
     expected = getattr(rolling_obj, name)()
     assert_allclose(actual, expected)
-    assert da.dims == expected.dims
```
shoyer: I don't think there's any reason to delete this line?

xarray/core/rolling.py
```diff
@@ -97,16 +89,55 @@ def __repr__(self):
         return "{klass} [{attrs}]".format(klass=self.__class__.__name__,
                                           attrs=','.join(attrs))
 
+    def __len__(self):
+        return len(self.obj[self.dim])
```
shoyer: Nit: prefer `self.obj.sizes[self.dim]` instead -- it avoids creating a DataArray object (which is somewhat more expensive).

xarray/tests/test_dataarray.py
fujiisoup: It seems my mistake...
Thank you for finding it.

I'll send another PR.

xarray/tests/test_dataarray.py
fujiisoup: I thought `assert_allclose(actual, expected)` does the same thing, so I deleted it.
I experientially found that this function raises an error if two `dims`'s orders are differ.
Wasn't `assert_allclose` designed to do it?

xarray/core/rolling.py
fujiisoup: Thanks.
I'll fix it.

xarray/tests/test_dataarray.py
shoyer: Good point, I missed that this was the xarray function and not the numpy one. So it is indeed redundant (but harmless).

shoyer:
This is great -- thanks @fujiisoup !

PR #1291, opened 2017-03-02
Guess the complementary dimension when only one is passed to pcolormesh

### LINKED ISSUE #1290: guess the second coordinate when only one is passed to pcolormesh()
Say I have a DataArray `z `with dimensions `('x', 'y')`, in that order. If I call `z.plot()`, it will create a pcolormesh with `x` the vertical dimension and `y` the horizontal one. If I want to invert the axes, I need to call `z.plot(x='x', y='y')`. If I supply only one of the dimensions, i.e. `z.plot(x='x')`, I get the error message

```python
ValueError: cannot supply only one of x and y
```

I think that when calling the plot function on a 2d DataArray and passing only one coordinate, as in `z.plot(x='x')`, xarray could guess we want the unpassed coordinate to be the other one. I don't see why this would be unsafe.

vnoel:
- [x] closes #1290

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -23,7 +23,8 @@ Enhancements
 ~~~~~~~~~~~~
 
 - When `plot()` is called on a 2D DataArray and only one dimension is
-specified with `x=` or `=y`, the other dimension is now guessed.
+  specified with `x=` or `=y`, the other dimension is now guessed. By
+  `Vincent Noel <https://github.com/vnoel>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -23,7 +23,7 @@ Enhancements
 ~~~~~~~~~~~~
 
 - When `plot()` is called on a 2D DataArray and only one dimension is
-  specified with `x=` or `=y`, the other dimension is now guessed. By
+  specified with `x=` or `y=`, the other dimension is now guessed. By
   `Vincent Noel <https://github.com/vnoel>`_.
 
 Bug fixes
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -22,6 +22,9 @@ v0.9.2 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- When `plot()` is called on a 2D DataArray and only one dimension is
+specified with `x=` or `=y`, the other dimension is now guessed.
+
 Bug fixes
 ~~~~~~~~~
 - ``rolling`` now keeps its original dimension order (:issue:`1125`).
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -221,8 +221,10 @@ def _infer_xy_labels(darray, x, y):
         if darray.ndim != 2:
             raise ValueError('DataArray must be 2d')
         y, x = darray.dims
-    elif x is None or y is None:
-        raise ValueError('cannot supply only one of x and y')
+    elif x is None:
+        x = darray.dims[0] if y == darray.dims[1] else darray.dims[1]
+    elif y is None:
+        y = darray.dims[0] if x == darray.dims[1] else darray.dims[1]
     elif any(k not in darray.coords and k not in darray.dims for k in (x, y)):
         raise ValueError('x and y must be coordinate variables')
     return x, y
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -227,7 +227,7 @@ def _infer_xy_labels(darray, x, y):
         x = darray.dims[0] if y == darray.dims[1] else darray.dims[1]
     elif y is None:
         if x not in darray.dims:
-            raise ValueError('x must be a coordinate variables')
+            raise ValueError('x must be a coordinate variable')
         y = darray.dims[0] if x == darray.dims[1] else darray.dims[1]
     elif any(k not in darray.coords and k not in darray.dims for k in (x, y)):
         raise ValueError('x and y must be coordinate variables')
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -222,8 +222,12 @@ def _infer_xy_labels(darray, x, y):
             raise ValueError('DataArray must be 2d')
         y, x = darray.dims
     elif x is None:
+        if y not in darray.dims:
+            raise ValueError('y must be a coordinate variable')
         x = darray.dims[0] if y == darray.dims[1] else darray.dims[1]
     elif y is None:
+        if x not in darray.dims:
+            raise ValueError('x must be a coordinate variables')
         y = darray.dims[0] if x == darray.dims[1] else darray.dims[1]
     elif any(k not in darray.coords and k not in darray.dims for k in (x, y)):
         raise ValueError('x and y must be coordinate variables')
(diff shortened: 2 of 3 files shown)


fmaussion:
Seems pretty safe to me, thanks! Could you add some tests, too? (at least for proper code coverage)

A good place to put a test would be here: 
https://github.com/pydata/xarray/blob/master/xarray/tests/test_plot.py#L638

(I suspect your PR brakes these tests, btw)

fmaussion:
Does anyone know why Travis doesn't start?

vnoel:
I have no experience with how tests are implemented in xarray. Are there pointers somewhere? How do I know the tests pass?

vnoel:
I have added a couple of exceptions to the `_infer_xy_labels` function and updated the tests as best I could

vnoel:
thanks ! Do we need to worry about the "appveyor" build fails? The failure seems to happen in somewhere unrelated

vnoel:
Can anyone knowledgeable tell me how to fix the appveyor build? It's complaining of a failing `test_dask_distributed_integration_test`...

shoyer:
I *think* I managed to restart the Travis-CI test. Maybe try pushing again or rebasing to see if that trigger them again.

See https://github.com/pydata/xarray/issues/1292 for the failing dask distibuted tests. Those can be safely ignored here.

xarray/plot/utils.py
```diff
@@ -221,8 +221,14 @@ def _infer_xy_labels(darray, x, y):
         if darray.ndim != 2:
             raise ValueError('DataArray must be 2d')
         y, x = darray.dims
-    elif x is None or y is None:
-        raise ValueError('cannot supply only one of x and y')
+    elif x is None:
+        if y not in darray.dims:
+            raise ValueError('y must be a coordinate variable')
```
shoyer: This is a little misleading -- I would say "y must be a coordinate variable if x is not supplied"

shoyer:
Tests should pass after a rebase or merge thanks to #1293.

xarray/plot/utils.py
vnoel: I'm not sure I understand the logic. The user requests a y variable that is not in the coordinates. It would not work even if she supplied a valid x coordinate. Your messages suggests it would.

Maybe "y is not a coordinate variable" is more to the point (and the same for x)?

fmaussion: Mh, that's actually a good point. ``x`` and ``y`` can also be optional (e.g. 2D) coordinates. What about:

```y must be a dimension coordinate if x is not supplied```

?

Comment (vnoel) [inline xarray/plot/utils.py]: @fmaussion : You are proposing the same message as @shoyer above. My comment seems still valid to me. 

Why focus on 'if x is not supplied'? Why no equivalent 'if y is not supplied'? I feel like I'm missing something.

Comment (fmaussion) [inline xarray/plot/utils.py]: See the following example:

```
import xarray as xr
import numpy as np
lon, lat = np.meshgrid(np.linspace(-20, 20, 5), np.linspace(0, 30, 4))
da = xr.DataArray(np.arange(20).reshape(4, 5), dims=['y', 'x'],
                  coords = {'lat': (('y', 'x'), lat),
                            'lon': (('y', 'x'), lon)})
da.plot(x='lat')
```

Note that ``da.plot(x='lat', y='lon')`` is a valid call and that ``'lat' in da.coords`` is ``True`` (but ``'lat' in da.dims`` is ``False``).

With your current PR the code  snippet above would lead to the following error:

```x must be a coordinate variable```

Which is misleading, since ``'lat'`` *is* a coordinate variable (it is not a dimension coordinate, however). 

I just opened an issue to discuss these terminology problems: https://github.com/pydata/xarray/issues/1295

Comment (fmaussion) [inline xarray/plot/utils.py]: Ah, and I just understood your other question: yes, the message should be the same for both cases.

Comment (vnoel) [inline xarray/plot/utils.py]: Should we check for x and y in darray.coords then, instead of darray.dims? Would that work? I checked in darray.dims because the previous code was doing that.

File: doc/whats-new.rst [hunk 2]
```diff
@@ -22,6 +22,10 @@ v0.9.2 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- When `plot()` is called on a 2D DataArray and only one dimension is
+  specified with `x=` or `=y`, the other dimension is now guessed. By
```
fmaussion: typo: ``=y`` -> ``y=``

xarray/plot/utils.py
fmaussion: > Should we check for x and y in darray.coords then, instead of darray.dims?

No, I don't think this would work. 

Since we will probably land on the name "coordinate variable" for the "dimension" coords, I guess we can go back to @shoyer 's initial suggestion, or change the error message to:

```y must be a dimension name if x is not supplied```

which has the advantage to cover all cases (also when there are unindexed dimensions).

Comment (vnoel) [inline xarray/plot/utils.py]: So in your code sample above, for instance, we don't want to guess that y should be "lon"? Why not?

Comment (vnoel) [inline xarray/plot/utils.py]: Ok, I understood :-)

Comment (fmaussion) [inline xarray/plot/utils.py]: Because in 2D Plots we are guaranteed to have only 2 dims, but we can have an arbitrary number of auxiliary coordinates, in an arbitrary order. In my example above checking for coords would work, but this is not always the case. See:

```python
import xarray as xr
import numpy as np
lon, lat = np.meshgrid(np.linspace(-20, 20, 5), np.linspace(0, 30, 4))
da = xr.DataArray(np.arange(20).reshape(4, 5), dims=['y', 'x'],
                  coords = {'lat': (('y', 'x'), lat),
                            'lon': (('y', 'x'), lon),
                            'x': (('x', ), np.arange(5)),
                            'y': (('y', ), np.arange(4)),
                            })
print(da.coords)

Coordinates:
  * y        (y) int64 0 1 2 3
    lon      (y, x) float64 -20.0 -10.0 0.0 10.0 20.0 -20.0 -10.0 0.0 10.0 ...
  * x        (x) int64 0 1 2 3 4
    lat      (y, x) float64 0.0 0.0 0.0 0.0 0.0 10.0 10.0 10.0 10.0 10.0 ...
```

vnoel: Got it, thanks :-)

vnoel:
I hope everything's ok now. Thanks for the hand-holding

xarray/tests/test_plot.py
```diff
@@ -636,17 +636,24 @@ def test_xy_strings(self):
...
+        self.plotmethod(x='x')
+        ax = plt.gca()
+        self.assertEqual('x', ax.get_xlabel())
+        self.assertEqual('y', ax.get_ylabel())
 
     def test_bad_x_string_exception(self):
-        with self.assertRaisesRegexp(ValueError, 'x and y must be coordinate'):
+        with self.assertRaisesRegexp(ValueError, 'x and y must be coordinate variables'):
             self.plotmethod('not_a_real_dim', 'y')
+        with self.assertRaisesRegexp(ValueError, 'x must be a dimension name if y is not supplied'):
```
shoyer: I don't have a character count handy, but is this at 80 or more characters? If so, please split onto another line. I think flake8 or pep8 should catch this.

vnoel: 79 characters. flake8 reports lots of errors but not in this code.

fmaussion: Can you double check please? ``git diff upstream/master | flake8 --diff`` checks only your addition.



In particular,

vnoel: git had switched branch under me, sorry

fmaussion:
Good, thanks!

Issue #535, opened 2015-08-17
test_cross_engine_read_write_netcdf4 is broken on Python 3 due to some sort of upstream change


shoyer:
This has started causing test failures on master.

I had trouble tracking this down, but it seems to be related to a build of hdf5, netCDF4 or the underlying libraries which just became available on conda.

In case anyone else has time for it:
This build worked: https://travis-ci.org/xray/xray/jobs/75633366
This build failed: https://travis-ci.org/xray/xray/jobs/75777788

For now, let's just skip the test and come back to this later.

Possibly related: https://github.com/Unidata/netcdf4-python/issues/448

jhamman:
Here's a diff of the versions that went into each build environment: 

```
$ diff -y worked.txt broken.txt
    curl:            7.38.0-0         			      |	    curl:            7.43.0-1         

    hdf5:            1.8.14-0         			      |	    hdf5:            1.8.15.1-1       

    libnetcdf:       4.3.2-1          			      |	    krb5:            1.13.2-0         

    netcdf4:         1.1.8-np19py34_0 			      |	    libnetcdf:       4.3.3.1-1        
							      >
							      >	    netcdf4:         1.1.9-np19py34_0 
```

given that the failure is in an I/O test, it seems likely that this is coming from `netcdf4`.

QuLogic:
See #1298, where I changed Python 2.7 CI to install `h5netcdf` as well (it seemed to be missing.) This test doesn't work on Python 2 either right now.

PR #1298, opened 2017-03-07
netcdf4/h5netcdf cross engine test broken on Python 2

QuLogic:
This is not really a bugfix or enhancement. In fact, it's just a bug report right now. I'm having trouble running tests on Python 2 with both netcdf4 and h5netcdf installed due to the cross engine test (
`H5NetCDFDataTest.test_cross_engine_read_write_netcdf4`).

I noticed that the requirements list `netcdf4` twice; I assume this was a typo and the second one should be `h5netcdf`. By switching the duplicate out, I'm hoping to trigger the problem on CI as a way to report this bug.

 - [ ] closes #xxxx
 - [ ] tests added / passed
 - [ ] whatsnew entry

### DIFF
--- a/ci/requirements-py27-cdat+pynio.yml
+++ b/ci/requirements-py27-cdat+pynio.yml
@@ -15,7 +15,7 @@ dependencies:
   - pytest-cov
   - cyordereddict
   - h5py
-  - netcdf4
+  - h5netcdf
   - matplotlib
   - seaborn
   - pip:


QuLogic:
Fortunately, this does [fail in the same way](https://travis-ci.org/pydata/xarray/jobs/208536170#L2314-L2377) it does for me locally. Since Python 3 works, I think this might be a bytes/str thing, but the error is a bit opaque to me.

fmaussion:
The test that fails is called ``test_cross_engine_read_write_netcdf4``, so it looks like it's more of a testing bug that an issue in xarray (the tests shouldn't run when netcdf4 isn't installed).

QuLogic:
Both `netcdf4` and `h5netcdf` are installed now; that's the point.

QuLogic:
shoyer/h5netcdf#16 appears to be the upstream issue, but oddly, it seems to be reported against Python 3 when here it fails with Python 2.

QuLogic:
Oh, just realized there's a skip on Python 3, which links to #535. Apparently, this test is now broken on both Python versions.

shoyer:
Thanks for figuring this out -- I fixed our test suite in https://github.com/pydata/xarray/pull/1336. I'm looking forward to the next netCDF release, at which point h5netcdf will stop having this issue!

PR #1299, opened 2017-03-07
BUG/TST: Retain encoding upon concatenation

### LINKED ISSUE #1297: Encoding lost upon concatenation
When using `xarray.concat`, attributes are retained, but encoding information is not.  I believe encoding information should be retained, at least optionally.

```
In [64]: da = xarray.DataArray([1, 2, 3, 2, 1])

In [65]: da.attrs.update(foo="bar")

In [66]: da.encoding.update(complevel=5)

In [67]: da2 = xarray.concat((da, da), dim="new")

In [68]: print(da2.attrs)
OrderedDict([('foo', 'bar')])

In [69]: print(da2.encoding)
{}
```

gerritholl:
Retain encoding upon concatenation of DataArray or Dataset.

 - [x] closes #1297 
 - [x] tests added / passed
 - [ ] whatsnew entry

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -975,13 +975,14 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
             data = ops.stack(arrays, axis=axis)
 
         attrs = OrderedDict(first_var.attrs)
+        encoding = OrderedDict(first_var.encoding)
         if not shortcut:
             for var in variables:
                 if var.dims != first_var.dims:
                     raise ValueError('inconsistent dimensions')
                 utils.remove_incompatible_items(attrs, var.attrs)
 
-        return cls(dims, data, attrs)
+        return cls(dims, data, attrs, encoding)
 
     def equals(self, other, equiv=ops.array_equiv):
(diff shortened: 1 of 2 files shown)


jhamman:
This looks pretty good. We also should make sure that the `Dataset.encoding` attribute is passed on.

gerritholl:
Concatenation of `DataArray` and `Dataset` use the same underlying code, do you mean specifically for the test?

jhamman:
No, I mean that there is an encoding attribute on the `Dataset` in addition to the `DataArray` that should also be propagated. I think you'll need to basically repeat what you did in `variable.py` for the `Dataset`: https://github.com/pydata/xarray/blob/10fc0846b1ae86d729ffd855641f4510947085a9/xarray/core/combine.py#L223

jhamman:
@gerritholl - do have questions about my last comment? This should be pretty easy to get wrapped up and merged.  We also need a whatsnew entry.

gerritholl:
Sorry, I've been really busy, but I'll get around to it eventually!

jhamman:
I have an updated version of this PR up at #1545.

PR #1305, opened 2017-03-10
Clarify licenses for bundled code

### LINKED ISSUE #1254: Provide more detailed copyright information
So far, copyright information is summarized in the README as:
```
xarray includes portions of pandas, NumPy, Seaborn and Python itself. These licenses are included in the licenses directory.
```
Please consider providing more detailed information about what is covered by what in the code base. The Apache license states this information must be recorded in an additional `NOTICE` file alongside the existing `LICENSE` file.

That will also help me with the Debian packaging, since I need an accurate record of all copyrighted materials present in the code.

Thanks,
Ghis

shoyer:
They are all now called out explicitly in the README as well.

 - [x] closes #1254

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,7 +110,7 @@ __ http://climate.com/
 License
 -------
 
-Copyright 2014-2016, xarray Developers
+Copyright 2014-2017, xarray Developers
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


fmaussion:
Great! 

cc @ghisvail

shoyer:
To add a little more context on my considerations here, for Apache projects that bundle permissively licensed dependencies (like what we've done here), they actually add a note to LICENSE directly instead of NOTICE: http://www.apache.org/dev/licensing-howto.html#permissive-deps

We could take this approach, but it has the downside of breaking tools that automatically parse LICENSE such as GItHub itself at the top of each repository. One could argue that that's a feature, not a bug, but I think it's better to show the main license at first glance (almost every large project has some bundled dependencies). Apache Spark is a good example: https://github.com/apache/spark

ghisvail:
I really fail to understand how much of a bug deal it is to introduce a NOTICE file with the right content and keep the LICENSE file and the rest of the code base unchanged. It cannot get any simpler than this. 

Anyway, your commit provided the missing copyright information I was seeking. For the rest, I gave you a piece of advice for easy compliance (the `NOTICE` file) and you are free to follow or disregard it. Moving on.

PR #1310, opened 2017-03-15
Docs: don't mention inferring `dims` from `coords`

MareinK:
Trying this results in a deprecation warning


shoyer:
This actually works, but only if coords is provided as a list, not a dict. Can you add that clarification instead?

PR #1311, opened 2017-03-16
More explicit check for dtype roundtripping in backends

shoyer:
- [x] tests added / passed


pwolfram:
@shoyer, should this PR be merged before #1198, just in case to make sure that there aren't any "gotchas" that I've missed in #1198?

pwolfram:
Thanks!

PR #1312, opened 2017-03-16
Use LooseVersion rather than StrictVersion

shoyer:
LooseVersion plays nicer when development versions of other libraries (e.g.,
dask) are installed.


jhamman:
@shoyer - looks like travis didn't run and Appveyor failed. I'm going to close/reopen to see if that corrects things.

jhamman:
That worked. Looks good to me.

PR #1322, opened 2017-03-24
Shorter repr for attributes

### LINKED ISSUE #1319: Truncate long lines in repr of Dataset.attrs
When loading from NetCDF, `Dataset.attrs` often has a few long strings, which may even have embedded newlines (eg a multi-paragraph `summary` or `references` section).  It's lovely that these are available, but they tend to make the repr very long and poorly formatted - to the point that many Jupyter notebooks begin by discarding the `attrs`, which makes it rather pointless to store or display metadata at all!

Given that these values are already truncated at 500 characters (including the indicative `...`, but not the start point), I propose that they should instead be truncated to 80 characters including the indentation and key (as `values` are).  For the sake of pretty-printing, this should also replace newlines or tabs with spaces and truncate early if an empty line is encountered.

Another solution would be add appropriate indentation following newlines or wrapping, so that the structure remains clear.  However, I think that it is better to print a fairly minimal representation of the metadata by default.

```
>>> xr.open_dataset('http://dapds00.nci.org.au/thredds/dodsC/uc0/rs0_dev/20170215-stacked_sample/LS7_ETM_NBART_3577_15_-40.ncml')

<xarray.Dataset>
Dimensions:  (time: 246, x: 4000, y: 4000)
Coordinates:
  * y        (y) float64 -3.9e+06 -3.9e+06 -3.9e+06 -3.9e+06 -3.9e+06 ...
  * x        (x) float64 1.5e+06 1.5e+06 1.5e+06 1.5e+06 1.5e+06 1.5e+06 ...
  * time     (time) datetime64[ns] 1999-07-16T23:49:39 1999-07-25T23:43:07 ...
Data variables:
    crs      int32 ...
    blue     (time, y, x) float64 ...
    green    (time, y, x) float64 ...
    red      (time, y, x) float64 ...
    nir      (time, y, x) float64 ...
    swir1    (time, y, x) float64 ...
    swir2    (time, y, x) float64 ...
Attributes:
    date_created: 2017-03-07T11:57:26.511217
    Conventions: CF-1.6, ACDD-1.3
    history: 2017-03-07T11:57:26.511307+11:00 adh547 datacube-ncml (1.2.2+23.gd1f3512.dirty) ls7_nbart_albers.yaml, 1.0.6a, /short/v10/datacube/002/LS7_ETM_NBART/LS7_ETM_NBART_3577_15_-40.ncml, (15, -40)  # Created NCML file to aggregate multiple NetCDF files along the time dimension
    geospatial_bounds: POLYGON ((148.49626113888138 -34.828378308133452,148.638689676063308 -35.720318326735864,149.734176111491877 -35.599556747691196,149.582601578289143 -34.708911907843387,148.49626113888138 -34.828378308133452))
    geospatial_bounds_crs: EPSG:4326
    geospatial_lat_min: -35.7203183267
    geospatial_lat_max: -34.7089119078
    geospatial_lat_units: degrees_north
    geospatial_lon_min: 148.496261139
    geospatial_lon_max: 149.734176111
    geospatial_lon_units: degrees_east
    comment: -	Ground Control Points (GCP): new GCP chips released by USGS in Dec 2015 are used for re-processing
-	Geometric QA: each product undergoes geometric assessment and the assessment result will be recorded within v2 AGDC for filtering/masking purposes.
-	Processing parameter settings: the minimum number of GCPs for Ortho-rectified product generation has been reduced from 30 to 10.
-	DEM: 1 second SRTM DSM is used for Ortho-rectification.
-	Updated Calibration Parameter File (CPF): the latest/cu...
    product_suite: Surface Reflectance NBAR+T 25m
    publisher_email: earth.observation@ga.gov.au
    keywords_vocabulary: GCMD
    product_version: 2
    cdm_data_type: Grid
    references: -	Berk, A., Anderson, G.P., Acharya, P.K., Hoke, M.L., Chetwynd, J.H., Bernstein, L.S., Shettle, E.P., Matthew, M.W., and Adler-Golden, S.M. (2003) Modtran 4 Version 3 Revision 1 User s manual. Airforce Research Laboratory, Hanscom, MA, USA.
-	Chander, G., Markham, B.L., and Helder, D.L. (2009) Summary of current radiometric calibration coefficients for Landsat MSS, TM, ETM+, and EO-1 ALI sensors. Remote Sensing of Environment 113, 893-903.
-	Edberg, R., and Oliver, S. (2013) Projection-Indep...
    platform: LANDSAT-7
    keywords: AU/GA,NASA/GSFC/SED/ESD/LANDSAT,REFLECTANCE,ETM+,TM,OLI,EARTH SCIENCE
    publisher_name: Section Leader, Operations Section, NEMO, Geoscience Australia
    institution: Commonwealth of Australia (Geoscience Australia)
    acknowledgment: Landsat data is provided by the United States Geological Survey (USGS) through direct reception of the data at Geoscience Australias satellite reception facility or download.
    license: CC BY Attribution 4.0 International License
    title: Surface Reflectance NBAR+T 25 v2
    summary: Surface Reflectance (SR) is a suite of Earth Observation (EO) products from GA. The SR product suite provides standardised optical surface reflectance datasets using robust 
physical models to correct for variations in image radiance values due to atmospheric properties, and sun and sensor geometry. The resulting stack of surface reflectance
grids are consistent over space and time which is instrumental in identifying and quantifying environmental change. SR is based on radiance data from the...
    instrument: ETM
    source: LANDSAT 7 ETM+ surface observation
    publisher_url: http://www.ga.gov.au
```

Zac-HD:
NetCDF files often have tens of attributes, including multi-paragraph summaries or the full modification history of the file.  It's great to have this available in the .attrs, but we can truncate it substantially
in the repr!  Hopefully this will stop people writing `data.attrs = {}` and discarding metadata in interactive workflows for the sake of cleaner output.

 - [x] closes #1319
 - [x] test data adjusted
 - [x] whatsnew entry

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -253,8 +253,15 @@ def summarize_coord(name, var, col_width):
 
 
 def summarize_attr(key, value, col_width=None):
-    # ignore col_width for now to more clearly distinguish attributes
-    return u'    %s: %s' % (key, maybe_truncate(value))
+    """Summary for __repr__ - use ``X.attrs[key]`` for full value."""
+    # Indent key and add ':', then right-pad if col_width is not None
+    k_str = u'    %s:' % key
+    if col_width is not None:
+        k_str = pretty_print(k_str, col_width)
+    # Replace tabs and newlines, so we print on one line in known width
+    v_str = unicode_type(value).replace(u'\t', u'\\t').replace(u'\n', u'\\n')
+    # Finally, truncate to the desired display width
+    return maybe_truncate(u'%s %s' % (k_str, v_str), OPTIONS['display_width'])
 
 
 EMPTY_REPR = u'    *empty*'
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -41,6 +41,13 @@ By `Henry S. Harrison <https://hsharrison.github.io>`_.
   Note, the default is ``autoclose=False``, which is consistent with previous
   xarray behavior.  By `Phillip J. Wolfram <https://github.com/pwolfram>`_.
 
+- The ``repr()`` of ``Dataset`` and ``DataArray`` attributes uses a similar
+  format to coordinates and variables, with vertically aligned entries
+  truncated to fit on a single line.  Hopefully this will stop people writing
+  ``data.attrs = {}`` and discarding metadata in notebooks for the sake of
+  cleaner output.  The full metadata is still available as ``data.attrs``.
+  By `Zac Hatfield-Dodds <https://github.com/Zac-HD>`_.
+
 Bug fixes
 ~~~~~~~~~
 - ``rolling`` now keeps its original dimension order (:issue:`1125`).
--- a/xarray/tests/test_dataarray.py
+++ b/xarray/tests/test_dataarray.py
@@ -49,7 +49,7 @@ class TestDataArray(TestCase):
             other    int64 0
         Dimensions without coordinates: time
         Attributes:
-            foo: bar""")
+            foo:      bar""")
         self.assertEqual(expected, repr(data_array))
 
     def test_repr_multiindex(self):
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -97,7 +97,7 @@ class TestDataset(TestCase):
             var2     (dim1, dim2) float64 1.162 -1.097 -2.123 1.04 -0.4034 -0.126 ...
             var3     (dim3, dim1) float64 0.5565 -0.2121 0.4563 1.545 -0.2397 0.1433 ...
         Attributes:
-            foo: bar""") % data['dim3'].dtype
+            foo:      bar""") % data['dim3'].dtype
         actual = '\n'.join(x.rstrip() for x in repr(data).split('\n'))
         print(actual)
         self.assertEqual(expected, actual)
@@ -183,7 +183,7 @@ class TestDataset(TestCase):
         Data variables:
             *empty*
         Attributes:
-            å: ∑""" % u'ba®')
+            å:        ∑""" % u'ba®')
         actual = unicode_type(data)
         self.assertEqual(expected, actual)
 
--- a/xarray/tests/test_formatting.py
+++ b/xarray/tests/test_formatting.py
@@ -131,6 +131,17 @@ class TestFormatting(TestCase):
         result = formatting.format_timestamp(date)
         self.assertEqual(result, expected)
 
+    def test_attribute_repr(self):
+        short = formatting.summarize_attr(u'key', u'Short string')
+        long = formatting.summarize_attr(u'key', 100 * u'Very long string ')
+        newlines = formatting.summarize_attr(u'key', u'\n\n\n')
+        tabs = formatting.summarize_attr(u'key', u'\t\t\t')
+        self.assertEqual(short, '    key: Short string')
+        self.assertLessEqual(len(long), 80)
+        self.assertTrue(long.endswith(u'...'))
+        self.assertNotIn(u'\n', newlines)
+        self.assertNotIn(u'\t', tabs)
+
 
 def test_set_numpy_options():
     original_options = np.get_printoptions()
--- a/xarray/tests/test_variable.py
+++ b/xarray/tests/test_variable.py
@@ -644,7 +644,7 @@ class TestVariable(TestCase, VariableSubclassTestCases):
         array([[1, 2, 3],
                [4, 5, 6]])
         Attributes:
-            foo: bar
+            foo:      bar
         """).strip()
         self.assertEqual(expected, repr(v))
 


xarray/core/formatting.py
```diff
@@ -253,8 +253,15 @@ def summarize_coord(name, var, col_width):
...
-    return u'    %s: %s' % (key, maybe_truncate(value))
+    """Summary for __repr__ - use ``X.attrs[key]`` for full value."""
+    key = u'    %s:' % unicode_type(key)
+    if col_width is not None:
+        key = pretty_print(key, col_width)
+    output = u'%s %s' % (key, unicode_type(value).replace(u'\t', u'    '))
+    empty_line = output.find(u'\n\n')
+    if empty_line != -1:
+        output = output[:empty_line] + u'...'
+    return maybe_truncate(output.replace(u'\n', u' '), 79)
```
shoyer: Use `OPTIONS['display_width']` instead of 79 here.

xarray/core/formatting.py
```diff
@@ -253,8 +253,15 @@ def summarize_coord(name, var, col_width):
 
 
 def summarize_attr(key, value, col_width=None):
-    # ignore col_width for now to more clearly distinguish attributes
-    return u'    %s: %s' % (key, maybe_truncate(value))
+    """Summary for __repr__ - use ``X.attrs[key]`` for full value."""
+    key = u'    %s:' % unicode_type(key)
+    if col_width is not None:
+        key = pretty_print(key, col_width)
+    output = u'%s %s' % (key, unicode_type(value).replace(u'\t', u'    '))
```
shoyer: Maybe we should only do all this text formatting logic if the attribute is already a string? I can imagine this doing strange things if the attribute is a number or array.

Zac-HD: I disagree - it's just adjusting the spacing and replacing tabs and newlines before printing them.  While I don't think numbers or array representations would have spaces (and arrays with newlines are rare), surely we want to process them for display in the same way?

Zac-HD:
Ping @shoyer

shoyer:
> If the logic is "clever", time to ditch it. I've rebased to a clearer version which simply handles indentation and replaces tabs and newlines with their backslash representations.

"Clever" was possibly too strong of a word. I think both your original and new proposal are fine -- use your own best taste on which version is most readable.

It's more that we need unit tests for *any* functionality. Otherwise, things tend to brake inadvertently when someone else does code cleanup months or years later.

> What properties of this would you want to test?

This could be super simple here, e.g., just a few assert statements verifying the text substitutions and truncation. Ideally we exercise every line of logic in your code.

xarray/core/formatting.py
shoyer: If somebody is putting non-strings in attributes (which they almost certainly are, as ill-advised as that may be), there is some value in having them appear in a noticeably distinct way. I would rather have non-strings look a little funky and span multiple lines than be confused with the `repr` of the objects.

e.g., consider how the repr of an `xarray.Dataset` looks:
```
In [16]: repr(xarray.Dataset({'foo': ('x', [1, 2, 3])}))[:70] + '...'
Out[16]: '<xarray.Dataset>\nDimensions:  (x: 3)\nDimensions without coordinates: x...'
```

Though, it might possibly make sense to "indent" their reprs on any newlines to align them with the indentation of first line.

@fmaussion @pwolfram any opinions here?

pwolfram: I think that indentation would help clarify. :+1:

Zac-HD: Indentation would certainly be an improvement, but I still strongly prefer to limit each attribute to a single line.

My motivation for this pull request is seeing a common pattern in notebooks dealing with NetCDF data of simply discarding the `attrs` metadata, because (a) the formatting is a mess and (b) it's a lot of output.  Taking the example in #1319, this is a temporary file with "only" 29 attributes, of which eight are long enough to truncate and three include newlines.  *This is a short example.*  With larger workflows, this pattern destroys provenance and reproducibility - along with directly useful information!

If users need the unmodified attributes or are unsure about types, they can access them directly - so I would really prefer to keep the repr to a single line for each, no matter what.

shoyer: In the examples you've seen, how often do these datasets with long lists of attributes include non-string attributes?

shoyer: Rather, non-string attributes that have a repr that spans multiple lines.

Zac-HD: It's not uncommon to find numbers or timestamps in the attributes, but it would surprise me to find something with a multiline repr.  I'd still want to truncate it to one line in that case, but that's more of a personal preference I guess.

Zac-HD:
@shoyer - my preference is to process and truncate all strings, including object reprs.  If you want to overrule that feel free; otherwise I think it's ready to merge.

shoyer:
@Zac-HD My guess is that non-string multi-line attributes are quite rare, so in truth it doesn't really matter what we pick (even though my preference leaned the other way from yours). If someone cares enough to complain we can revisit this, but until then I'm OK with your choice.

shoyer:
Thanks for the contribution!

PR #1336, opened 2017-03-28
Marks slow, flaky, and failing tests

### LINKED ISSUE #1309: Mark slow tests and don't run them by default
We'll have a few more of these after #1198 goes in, but plotting and some of the rolling aggregation tests already fall in this category. 

For development purposes, it would be nice if these don't run automatically so the whole test suite can finish in handful of seconds. They should still always run on Travis-CI and Appveyor, which take a few minutes to setup.

See 
http://doc.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option for guidance from pytest.

pwolfram:
Closes #1309

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,46 +11,54 @@ matrix:
   fast_finish: true
   include:
   - python: 2.7
-    env: CONDA_ENV=py27-min EXTRA_FLAGS=--skip-optional-ci
+    env: CONDA_ENV=py27-min
   - python: 2.7
-    env: CONDA_ENV=py27-cdat+pynio EXTRA_FLAGS=--skip-optional-ci
+    env: CONDA_ENV=py27-cdat+pynio
   - python: 3.4
-    env: CONDA_ENV=py34 EXTRA_FLAGS=--skip-optional-ci
+    env: CONDA_ENV=py34
   - python: 3.5
-    env: CONDA_ENV=py35 EXTRA_FLAGS=--skip-optional-ci
+    env: CONDA_ENV=py35
   - python: 3.6
-    env: CONDA_ENV=py36 EXTRA_FLAGS=--skip-optional-ci
+    env: CONDA_ENV=py36
   - python: 3.6
-    env: CONDA_ENV=py36-pydap EXTRA_FLAGS=--run-flaky
+    env:
+    - CONDA_ENV=py36
+    - EXTRA_FLAGS="--run-flaky"
   - python: 3.6
-    env: CONDA_ENV=py36-netcdf4-dev EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-pydap
+  - python: 3.6
+    env: CONDA_ENV=py36-netcdf4-dev
     addons:
       apt_packages:
         - libhdf5-serial-dev
         - netcdf-bin
         - libnetcdf-dev
   - python: 3.6
-    env: CONDA_ENV=py36-dask-dev EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-dask-dev
   - python: 3.6
-    env: CONDA_ENV=py36-pandas-dev EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-pandas-dev
   - python: 3.6
-    env: CONDA_ENV=py36-condaforge-rc EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-condaforge-rc
   allow_failures:
   - python: 3.6
-    env: CONDA_ENV=py36-pydap EXTRA_FLAGS=--run-flaky
+    env:
+    - CONDA_ENV=py36
+    - EXTRA_FLAGS="--run-flaky"
   - python: 3.6
-    env: CONDA_ENV=py36-netcdf4-dev EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-pydap
+  - python: 3.6
+    env: CONDA_ENV=py36-netcdf4-dev
     addons:
       apt_packages:
         - libhdf5-serial-dev
         - netcdf-bin
         - libnetcdf-dev
   - python: 3.6
-    env: CONDA_ENV=py36-dask-dev EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-dask-dev
   - python: 3.6
-    env: CONDA_ENV=py36-pandas-dev EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-pandas-dev
   - python: 3.6
-    env: CONDA_ENV=py36-condaforge-rc EXTRA_FLAGS=--run-flaky
+    env: CONDA_ENV=py36-condaforge-rc
 
 before_install:
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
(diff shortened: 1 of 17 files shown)


pwolfram:
@shoyer, this PR should also fix the issue with travis CI failing due to testing of too many open files.  Note, when a user runs `py.test` locally, the slow tests will run.  They can be skipped via running `py.test --skip-slow`.

pwolfram:
cc  @MaximilianR

pwolfram:
Note @shoyer et al, please feel free to mark additional tests as `slow` in this PR.  I only marked the ones from #1198 that were too slow for this round.

pwolfram:
@shoyer, all tests pass and I've spot checked that this works as expected, e.g.,
```python
xarray/tests/test_backends.py::OpenMFDatasetManyFilesTest::test_1_autoclose_netcdf4 PASSED

xarray/tests/test_backends.py::OpenMFDatasetManyFilesTest::test_1_open_large_num_files_netcdf4 SKIPPED

xarray/tests/test_backends.py::OpenMFDatasetManyFilesTest::test_2_autoclose_scipy PASSED

xarray/tests/test_backends.py::OpenMFDatasetManyFilesTest::test_2_open_large_num_files_scipy SKIPPED

xarray/tests/test_backends.py::OpenMFDatasetManyFilesTest::test_3_autoclose_pynio PASSED

xarray/tests/test_backends.py::OpenMFDatasetManyFilesTest::test_3_open_large_num_files_pynio SKIPPED
```

pwolfram:
@shoyer, is there anything else you would like done on this before merging?

shoyer:
@pwolfram Thanks for putting this together.

"Slow" tests on Travis aren't a huge problem -- it takes at least a minute for Travis to complete even in the best case scenario, so having to wait a minute more is not so bad. It's really an issue for interactive, local development, where it's really valuable for tests to complete in less than 10 seconds.

Flaky tests, which fail sometimes, are the problem. So think we want a different category for these many file tests, maybe "Flaky". Ideally we would have an another Travis-CI job setup to run these that isn't required to always pass, like our existing jobs for testing development versions of other libraries.

pwolfram:
@shoyer, as we discussed, here is a robustness-ing of the testing as needed for 0.9.2

pwolfram:
This also fixes the issue noted in https://github.com/pydata/xarray/pull/1038 where flakey tests cause travis CI failure.

conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+def pytest_addoption(parser):
+    parser.addoption("--skip-flakey", action="store_true",
+            help="skips flakey tests")
```
shoyer: nit: please intent even with `(` on the previous line.

.travis.yml
```diff
@@ -76,7 +76,7 @@ install:
   - python setup.py install
 
 script:
-  - py.test xarray --cov=xarray --cov-report term-missing --verbose
+  - py.test xarray --cov=xarray --cov-report term-missing --verbose --skip-flakey
```
shoyer: Can you make `--skip-flakey` an environment variable so we can make one of our "allowed failure" matrix builds not skip flakey tests?

xarray/tests/__init__.py
```diff
@@ -114,6 +114,18 @@ def requires_bottleneck(test):
...
 
+flakey = pytest.mark.skipif(
+    pytest.config.getoption("--skip-flakey"),
+    reason="set --skip-flakey option to skip flakey tests"
+)
+
+
+slow = pytest.mark.skipif(
+    not pytest.config.getoption("--run-slow"),
+    reason="set --run-slow option to run slow tests"
```
shoyer: Just to be clear, it looks like the current default behavior is:
- Don't skip flakey tests
- Skip slow tests

I think we actually want to flip these. Consider someone running our test suite for the first time. They probably want to understand if their xarray installation worked. It's not a big deal if this is a little slow, because they probably aren't doing this many times in a row, but not testing part of the test suite would be unfortunate. Once they notice the tests are slow, they will probably quickly look for a `--skip-slow` option.

pwolfram: I've set the three categories:
 * slow (runs on default): `slow`
 * CI specific (runs on default, specified to be optional for CI): `optionalci`
 * flakey (not run on default): `flakey`

I don't currently have flakey tests.  Plotting routines are marked as slow and long tests from the too many open files issue are marked `optionalci`.  We have some test (h5netcdf) that are flakey because of that bug I've found.  They are no longer commented but are flagged via `@flakey`.

.travis.yml
```diff
@@ -76,7 +158,7 @@ install:
   - python setup.py install
 
 script:
-  - py.test xarray --cov=xarray --cov-report term-missing --verbose
+  - py.test xarray --cov=xarray --cov-report term-missing --verbose $SKIP_CI $RUN_FLAKEY
```
shoyer: Let's use a single $EXTRA_FLAGS env variable

.travis.yml
```diff
@@ -13,28 +13,63 @@ matrix:
   - python: 2.7
     env: CONDA_ENV=py27-min
   - python: 2.7
+    env: CONDA_ENV=py27-min SKIP_CI=--skip-optional-ci
```
shoyer: We have a limited number of concurrent builds on Travis-CI: only five. So extra optional builds clog up other CI infrastructure and make it slower to run new tests. So I'd strongly prefer to only have one extra build optional if possible. The 15 or so you add here is too much :)

pwolfram: @shoyer-- This is the intent of my new changes: We really only need to test the h5netcdf tests that are now marked as flakey.  I added this extra tests and make the rest of the tests to skip optional CI if they aren't marked as allowed failures.

pwolfram: Actually, we can just run flakey tests for all tests allowed to fail and skip optional CI for all tests that must pass.  This avoids having new tests and should be more economical.

pwolfram:
@shoyer, changes have been made as requested.  Thanks!

pwolfram:
@shoyer, all checks pass and this is ready for a review / merge when you have time.

xarray/tests/__init__.py
shoyer: The h5netcdf many file tests are not flaky: they *always fail*. So let's mark as a "expected failure" with `pytest.mark.xfail` instead.

I would argue that the "CI specific" tests should actually be considered flaky: if they fail on our CI, they are likely to also fail for users and redistributors as well (e.g., consider someone packaging xarray for conda-forge or ubuntu). Since these failures do not indicate a failed installation, we shouldn't run them by default.

I'll push a commit shortly with these changes...

shoyer:
Thanks @pwolfram!

QuLogic:
If slow tests are being run by default, I'm not sure they really need their own special option. You can mark them (with `@pytest.mark.slow`) and use pytest's builtin selectors to not run them: `pytest -m "not slow"`.

Are flaky tests actually _flaky_ or do they just _not work_? If flaky, and a re-run will help, then maybe try the `pytest-rerunfailures` plugin.

shoyer:
> If slow tests are being run by default, I'm not sure they really need their own special option. You can mark them (with @pytest.mark.slow) and use pytest's builtin selectors to not run them: pytest -m "not slow".

Yes, this is a good idea.

> Are flaky tests actually flaky or do they just not work? If flaky, and a re-run will help, then maybe try the pytest-rerunfailures plugin.

Thanks for pointing this out! I'm not entirely sure pytest-rerunfailures is a fit here but it's a good option to know about.

The tests are flaky on Travis-CI, but they are reliable when run locally, and we still want to keep them around as integration tests. These are slow tests that open and close quite a lot of files (2000 each).

PR #1342, opened 2017-03-30
Ensures drop=True case works with empty mask

### LINKED ISSUE #1341: where(..., drop=True) failure for empty mask on python 2.7
The following fails for 2.7 but not 3.5 (reproducible script at https://gist.github.com/89bd5bd62a475510b2611cbff8d5c67a):
```python
In [1]: import xarray as xr

In [2]: import numpy as np

In [3]: da = xr.DataArray(np.random.rand(100,10), dims=['nCells','nVertLevels'])

In [4]: mask = xr.DataArray(np.zeros((100,), dtype='bool'), dims='nCells')

In [5]: da.where(mask, drop=True)
<ipython-input-5-ca5cd9c083a9> in <module>()
----> 1 da.where(mask, drop=True)

/Users/pwolfram/src/xarray/xarray/core/common.pyc in where(self, cond, other, drop)
    681             outcond = cond.isel(**clip)
    682             indexers = {dim: outcond.get_index(dim) for dim in outcond.dims}
--> 683             outobj = self.sel(**indexers)
    684         else:
    685             outobj = self

/Users/pwolfram/src/xarray/xarray/core/dataarray.pyc in sel(self, method, tolerance, drop, **indexers)
    670             self, indexers, method=method, tolerance=tolerance
    671         )
--> 672         result = self.isel(drop=drop, **pos_indexers)
    673         return result._replace_indexes(new_indexes)
    674 

/Users/pwolfram/src/xarray/xarray/core/dataarray.pyc in isel(self, drop, **indexers)
    655         DataArray.sel
    656         """
--> 657         ds = self._to_temp_dataset().isel(drop=drop, **indexers)
    658         return self._from_temp_dataset(ds)
    659 

/Users/pwolfram/src/xarray/xarray/core/dataset.pyc in isel(self, drop, **indexers)
   1115         for name, var in iteritems(self._variables):
   1116             var_indexers = dict((k, v) for k, v in indexers if k in var.dims)
-> 1117             new_var = var.isel(**var_indexers)
   1118             if not (drop and name in var_indexers):
   1119                 variables[name] = new_var

/Users/pwolfram/src/xarray/xarray/core/variable.pyc in isel(self, **indexers)
    545             if dim in indexers:
    546                 key[i] = indexers[dim]
--> 547         return self[tuple(key)]
    548 
    549     def squeeze(self, dim=None):

/Users/pwolfram/src/xarray/xarray/core/variable.pyc in __getitem__(self, key)
    375         dims = tuple(dim for k, dim in zip(key, self.dims)
    376                      if not isinstance(k, (int, np.integer)))
--> 377         values = self._indexable_data[key]
    378         # orthogonal indexing should ensure the dimensionality is consistent
    379         if hasattr(values, 'ndim'):

/Users/pwolfram/src/xarray/xarray/core/indexing.pyc in __getitem__(self, key)
    465 
    466     def __getitem__(self, key):
--> 467         key = self._convert_key(key)
    468         return self._ensure_ndarray(self.array[key])
    469 

/Users/pwolfram/src/xarray/xarray/core/indexing.pyc in _convert_key(self, key)
    452         if any(not isinstance(k, (int, np.integer, slice)) for k in key):
    453             # key would trigger fancy indexing
--> 454             key = orthogonal_indexer(key, self.shape)
    455         return key
    456 

/Users/pwolfram/src/xarray/xarray/core/indexing.pyc in orthogonal_indexer(key, shape)
     77     """
     78     # replace Ellipsis objects with slices
---> 79     key = list(canonicalize_indexer(key, len(shape)))
     80     # replace 1d arrays and slices with broadcast compatible arrays
     81     # note: we treat integers separately (instead of turning them into 1d

/Users/pwolfram/src/xarray/xarray/core/indexing.pyc in canonicalize_indexer(key, ndim)
     65         return indexer
     66 
---> 67     return tuple(canonicalize(k) for k in expanded_indexer(key, ndim))
     68 
     69 

/Users/pwolfram/src/xarray/xarray/core/indexing.pyc in <genexpr>((k,))
     65         return indexer
     66 
---> 67     return tuple(canonicalize(k) for k in expanded_indexer(key, ndim))
     68 
     69 

/Users/pwolfram/src/xarray/xarray/core/indexing.pyc in canonicalize(indexer)
     62                                      'array indexing; all subkeys must be '
     63                                      'slices, integers or sequences of '
---> 64                                      'integers or Booleans' % indexer)
     65         return indexer
     66 

ValueError: invalid subkey array([], dtype=object) for integer based array indexing; all subkeys must be slices, integers or sequences of integers or Booleans
```

pwolfram:
Resolves error occurring for python 2.7 for the case of `where(mask, drop=True)` where the mask is empty.

 - [x] closes #1341
 - [X] tests added
 - [x]  tests passed
 - [x] whatsnew entry

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -300,7 +300,8 @@ class BaseDataObject(AttrAccessMixin):
         try:
             return self.indexes[key]
         except KeyError:
-            return pd.Index(range(self.sizes[key]), name=key)
+            # need to ensure dtype=int64 in case range is empty on Python 2
+            return pd.Index(range(self.sizes[key]), name=key, dtype=np.int64)
 
     def _calc_assign_results(self, kwargs):
         results = SortedKeysDict()
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -49,6 +49,9 @@ Bug fixes
 - Fix ``sel`` with ``method='nearest'`` on Python 2.7 and 64-bit Windows
   (:issue:`1140`).
   `Stephan Hoyer <https://github.com/shoyer>`_.
+- Fixes ``where`` with ``drop='True'`` for empty masks (:issue:`1341`).
+  By `Stephan Hoyer <https://github.com/shoyer>`_ and
+  `Phillip J. Wolfram <https://github.com/pwolfram>`_.
 
 .. _whats-new.0.9.1:
 
--- a/xarray/tests/test_dataarray.py
+++ b/xarray/tests/test_dataarray.py
@@ -106,6 +106,13 @@ class TestDataArray(TestCase):
         with self.assertRaises(KeyError):
             array.get_index('z')
 
+    def test_get_index_size_zero(self):
+        array = DataArray(np.zeros((0,)), dims=['x'])
+        actual = array.get_index('x')
+        expected = pd.Index([], dtype=np.int64)
+        assert actual.equals(expected)
+        assert actual.dtype == expected.dtype
+
     def test_struct_array_dims(self):
         """
         This test checks subraction of two DataArrays for the case
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -2602,14 +2602,6 @@ class TestDataset(TestCase):
     def test_where_drop(self):
         # if drop=True
 
-        # empty case
-        array = DataArray(np.random.rand(100, 10),
-                          dims=['nCells', 'nVertLevels'])
-        mask = DataArray(np.zeros((100,), dtype='bool'), dims='nCells')
-        actual = array.where(mask, drop=True)
-        expected = DataArray(np.zeros((0, 10)), dims=['nCells', 'nVertLevels'])
-        self.assertDatasetIdentical(expected, actual)
-
         # 1d
         # data array case
         array = DataArray(range(5), coords=[range(5)], dims=['x'])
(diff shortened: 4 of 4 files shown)


pwolfram:
@shoyer, note that we probably need this fix to be included in the 0.9.2 release.

xarray/core/indexing.py
```diff
@@ -56,7 +56,8 @@ def canonicalize(indexer):
                 if indexer.ndim != 1:
                     raise ValueError('orthogonal array indexing only supports '
                                      '1d arrays')
-                if indexer.dtype.kind == 'b':
+                if (indexer.dtype.kind == 'b' or
+                        (indexer.dtype.kind == 'O' and len(indexer) == 0)):
```
shoyer: Can we fix this on `where` instead? I'm not sure it makes sense to allow `dtype=object` for all positional indexers, even for 0-d arrays.

pwolfram: @shoyer, you are right- it is better to fix it in `where`.

pwolfram:
@shoyer the change have been made and this should be ready to merge now.

shoyer:
Something more subtle is going on here.

I can only reproduce this on Python 2, not Python 3. The difference seems to be in what `.get_index()` returns for an dimension of length 0 without labels:
```
# python 2.7
In [11]: mask[:0].get_index('nCells')
Out[11]: Index([], dtype='object', name=u'nCells')

# python 3.5
In [5]: mask[:0].get_index('nCells')
Out[5]: RangeIndex(start=0, stop=0, step=1, name='nCells')
```

I'll investigate a little further and propose a fix.

shoyer:
@pwolfram I just pushed a commit, please take a look.

shoyer:
can you add a brief bug fix note for what's new? that seems to have gotten lost...

xarray/core/common.py
```diff
@@ -300,7 +300,8 @@ def get_index(self, key):
         try:
             return self.indexes[key]
         except KeyError:
-            return pd.Index(range(self.sizes[key]), name=key)
+            # need to ensure dtype=int64 in case range is empty on Python 2
+            return pd.Index(range(self.sizes[key]), name=key, dtype=np.int64)
```
pwolfram: @shoyer, this is obvious when you see it!

pwolfram:
@shoyer, the bug fix note has now been added and commits merged!  Thank you!

xarray/core/common.py
```diff
@@ -680,6 +681,10 @@ def where(self, cond, other=None, drop=False):
                                             for adim in np.nonzero(clipcond.values)]))
             outcond = cond.isel(**clip)
             indexers = {dim: outcond.get_index(dim) for dim in outcond.dims}
+            # ensures empty indexes are non-object indexers
```
shoyer: we can drop this change now

pwolfram: Agreed @shoyer

pwolfram:
Thanks @shoyer!

PR #1345, opened 2017-03-31
new dask prefix

### LINKED ISSUE #1343: Some XArray key names don't group nicely
Some XArray loading functions provide keys that don't adhere to dask conventions used for naming.  

We can solve this in XArray by using names like `'load-' + dask.base.tokenize(stuff)` or in dask by trying to identify and avoid names like these.  It might be wise to attempt both.  I expect that this will be easier to solve in XArray (though that's also in my own self interest :))

image

rabernat:
- [x] closes #1343
 - [ ] tests added / passed
 - [ ] whatsnew entry

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -247,7 +247,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
             token = tokenize(file_arg, group, decode_cf, mask_and_scale,
                              decode_times, concat_characters, decode_coords,
                              engine, chunks, drop_variables)
-            name_prefix = '%s:%s/' % (filename_or_obj, group or '')
+            name_prefix = 'open_dataset-%s:%s/' % (filename_or_obj, group or '')
             ds2 = ds.chunk(chunks, name_prefix=name_prefix, token=token,
                            lock=lock)
             ds2._file_obj = ds._file_obj


xarray/backends/api.py
```diff
@@ -247,7 +247,7 @@ def maybe_decode_store(store, lock=False):
             token = tokenize(file_arg, group, decode_cf, mask_and_scale,
                              decode_times, concat_characters, decode_coords,
                              engine, chunks, drop_variables)
-            name_prefix = '%s:%s/' % (filename_or_obj, group or '')
+            name_prefix = 'open_dataset-%s:%s/' % (filename_or_obj, group or '')
```
shoyer: does dask automatically drop the later parts (after the dash) from task names? if not, maybe better to drop those, too, and only use `open_dataset-{token}`.

rabernat: I think only @mrocklin knows the answer to that

mrocklin: Dask splits by hyphen and then keeps all parts that are strictly letters

```python
In [1]: from dask.utils import key_split

In [2]: key_split('foo-bar-123')
Out[2]: 'foo-bar'

In [3]: key_split('foo-bar-/abc123')
Out[3]: 'foo-bar'
```

At a certain point we stopped trying to provide keynames that were indicative of the arguments like `load-/path/to/file` and started just doing `'load-' + tokenize(...)` everywhere.

shoyer: So it seems like we should just do `open_dataset-` or `open-dataset-`, dropping the filename.

mrocklin: That would follow current conventions, yes.

    name = 'open-dataset-' + tokenize(filename)

shoyer:
`Dataset.chunk` also has some name related logic that needs to be updated.

rabernat:
I am superseding this with #1411.

PR #1347, opened 2017-04-02
Support for DataArray.expand_dims()

### LINKED ISSUE #1326: Expand dimensions in xarray
Based on the post
http://stackoverflow.com/questions/34987972/expand-dimensions-xray
and issue #1323.

I think it would be great if xarray has a method to expand dimension, such as `expand_dims` in numpy.
(`xr.DataArray.squeeze` is already implemented.)

fujiisoup:
- [x] closes #1326
 - [x] tests added / passed
 - [x] whatsnew entry

I added a DataArray's method `expand_dims` based on the discussion in #1326 .
The proposed API is similar to `numpy.expand_dims` and slightly different from `Variables.expand_dims`, which requires whole sequences of `dims` of the result array.

My concern is that I do not yet fully understand the lazy data manipulation in xarray.
Does Variable.expand_dims do it?

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -864,51 +864,8 @@ class DataArray(AbstractArray, BaseDataObject):
         expanded : same type as caller
             This object, but with an additional dimension(s).
         """
-        # Make sure user does not do `expand_dims(0)`
-        if isinstance(dim, int):
-            raise ValueError('dim should be str or sequence of strs or dict')
-
-        # Make axis and dims a pair of lists
-        if isinstance(dim, basestring):
-            dim = [dim]
-        if axis is not None and not isinstance(axis, (list, tuple)):
-            axis = [axis]
-
-        # infer None arguments.
-        if axis is None:
-            axis = list(range(len(dim)))
-
-        # Error checking
-        # Make sure the length of axis and dims are identical
-        if len(dim) != len(axis):
-            raise ValueError('lengths of dim and axis should be identical.')
-        # make sure any of dim does not already exist
-        for d in dim:
-            if d in self.dims:
-                raise ValueError('{dim} already exists.'.format(dim=d))
-
-        # make sure there is no duplicates in axis or dims
-        if len(axis) != len(set(axis)):
-            raise ValueError('axis should not contain same values.')
-        if len(dim) != len(set(dim)):
-            raise ValueError('dims should not contain same values.')
-
-        # sort dim and axis, so that axis is in order.
-        axis, dim = zip(*sorted(zip(axis, dim)))
-        # dims of the result array. This will be passed to Variable.expand_dims
-        all_dims = list(self.dims)
-        for a, d in zip(axis, dim):
-            all_dims.insert(a, d)
-
-        # If dims includes a label of a scalar variables,
-        # it will be promoted to a 1D coordinate consisting of a single value.
-        coords = OrderedDict(self._coords)
-        for d in dim:
-            if d in self._coords:
-                coords[d] = coords[d].expand_dims(d)
-
-        return DataArray(self._variable.expand_dims(all_dims),
-                         dims=all_dims, coords=coords, attrs=self.attrs)
+        ds = self._to_temp_dataset().expand_dims(dim, axis)
+        return self._from_temp_dataset(ds)
 
     def set_index(self, append=False, inplace=False, **indexes):
         """Set DataArray (multi-)indexes using one or more existing coordinates.
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -870,36 +870,62 @@ class DataArray(AbstractArray, BaseDataObject):
                If the axis is a list containing identical integers
                If axis is invalid (larger than the original dimension+1)
         """
-        # Some error checking
+        def create_dim_name(src_dims):
+            """ A simple function to automatically generate dimension name. """
+            for i in range(len(src_dims)+1):
+                dim = 'dim'+str(i)
+                if dim not in src_dims:
+                    return dim
+
         if axis is None and dim is None:
-            raise ValueError('At least one of axis or dim should be specified\
-            in expand_dims')
+            axis = [0]
+            dim = [create_dim_name(self.dims)]
+        if isinstance(axis, int):
+            axis = [axis]
+        if isinstance(dim, str):
+            dim = [dim]
+        # Some error checking
         if dim is not None:
-            if isinstance(dim, str):
-                dim = [dim]
             # Make sure user does not do `expand_dims(0)`
             if not isinstance(dim, (list, tuple)):
                 raise TypeError('dim should be str or list of str or tuple of \
                                  str')
             # Make sure the length of axis and dims are identical
-            if isinstance(axis, int) and len(dim) == 1 or len(dim) != len(axis):
+            if hasattr(axis, '__len__') and len(dim) != len(axis):
                 raise TypeError('lengths of dim and axis should be identical.')
             # make sure any of dim does not already exist
             for d in dim:
                 if d in self.dims:
                     raise ValueError('{dim} already exists.'.format(dim=d))
+            # If axis is None, the new dimensions are inserted at the start.
+            if axis is None:
+                axis = list(range(len(dim)))
 
-        if axis is None:
-            axis = list(range(len(dim)))
+            # sort dim and axis, so that axis is in order.
+            axis, dim = zip(*sorted(zip(axis, dim)))
+        else:
+            # sort dim and axis, so that axis is in order.
+            axis = sorted(axis)
+            dims = self.dims
+            dims = [create_dim_name(dims) for _ in axis]
+
         # make sure there is no duplicate in axis
         if len(axis) != len(set(axis)):
             raise ValueError('axis should not contain same values.')
-        # sort dim and axis, so that axis is in order.
-        axis, dim = zip(*sorted(zip(axis, dim)))
-        new_shape = self.shape
-        for a in axis:
-            new_shape.insert(a, 1)
 
+        def expand_1dim(da, d, a):
+            """ return a new dataarray with 1 dimension inserted
+            at position `a` named `d` """
+            new_dims = list(da.dims)
+            new_dims.insert(a, d)
+            new_array = np.expand_dims(da.values, a)
+            return DataArray(new_array, dims=new_dims,
+                             coords=da.coords, attrs=da.attrs, name=da.name)
+
+        expanded = self
+        for d, a in zip(dim, axis):
+            expanded = expand_1dim(expanded, d, a)
+        return expanded
 
     def set_index(self, append=False, inplace=False, **indexes):
         """Set DataArray (multi-)indexes using one or more existing coordinates.
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -856,7 +856,7 @@ class DataArray(AbstractArray, BaseDataObject):
             Axis position(s) where new axis is to be inserted (position(s) on
             the result array). If a list (or tuple) of integers is passed,
             multiple axes are inserted. In this case, dim arguments should be
-            same length list or None. If None is passed, all the axes will be
+            same length list. If axis=None is passed, all the axes will be
             inserted to the start of the result array.
 
         Returns
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -840,6 +840,67 @@ class DataArray(AbstractArray, BaseDataObject):
         ds = self._to_temp_dataset().swap_dims(dims_dict)
         return self._from_temp_dataset(ds)
 
+    def expand_dims(self, dim=None, axis=None):
+        """Return a new object with an additional axis (or axes) inserted at the
+        corresponding position in the array shape.
+
+        Parameters
+        ----------
+        dim : str, (list, tuple) of strs, or None
+            Name(s) of new dimension.
+            If a list (or tuple) of strings is passed, multiple axes are
+            inserted. In this case, axis argument should be None or same length
+            of integers indicating new axes positions.
+        axis : integer, list (or tuple) of integers or None
+            Axis position(s) where new axis is to be inserted (position(s) on
+            the result array). If a list (or tuple) of integers is passed,
+            multiple axes are inserted. In this case, dim arguments should be
+            None or same length list. If None is passed, all the axis will be
+            inserted to the start of the result array.
+
+        Returns
+        -------
+        expanded : same type as caller
+            This object, but with an additional dimension.
+
+        Raises
+        -------
+        ValueError:
+               If the length of axis and dim are different.
+               If the axis is a list containing identical integers
+               If axis is invalid (larger than the original dimension+1)
+        """
+        # Some error checking
+        if axis is None and dim is None:
+            raise ValueError('At least one of axis or dim should be specified\
+            in expand_dims')
+        if dim is not None:
+            if isinstance(dim, str):
+                dim = [dim]
+            # Make sure user does not do `expand_dims(0)`
+            if not isinstance(dim, (list, tuple)):
+                raise TypeError('dim should be str or list of str or tuple of \
+                                 str')
+            # Make sure the length of axis and dims are identical
+            if isinstance(axis, int) and len(dim) == 1 or len(dim) != len(axis):
+                raise TypeError('lengths of dim and axis should be identical.')
+            # make sure any of dim does not already exist
+            for d in dim:
+                if d in self.dims:
+                    raise ValueError('{dim} already exists.'.format(dim=d))
+
+        if axis is None:
+            axis = list(range(len(dim)))
+        # make sure there is no duplicate in axis
+        if len(axis) != len(set(axis)):
+            raise ValueError('axis should not contain same values.')
+        # sort dim and axis, so that axis is in order.
+        axis, dim = zip(*sorted(zip(axis, dim)))
+        new_shape = self.shape
+        for a in axis:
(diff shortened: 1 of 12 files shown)


xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
...
+
+        If dim is already a scalar coordinate, it will be promoted to a 1D
+        coordinate consisting of a single value.
+
+        Parameters
+        ----------
+        dim : str or sequence of str, or None.
+            Dimensions to include on the new variable.
+            dimensions are inserted with length 1.
+            If None is passed, dimension labels are automatically generated,
```
shoyer: I would rather not support automatic dimension naming anywhere other than the DataArray constructor. It's error prone if `array.expand_dims()` usually makes an array with a first dimension `dim_0` and but sometimes calls it something else.

It's OK for DataArray constructors because (a) it mostly reminds users that they need to supply names and (b) usually every name is given a default value, so we don't need to worry about conflicts.

xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
...
+        if len(axis) != len(set(axis)):
+            raise ValueError('axis should not contain same values.')
+        if len(dim) != len(set(dim)):
+            raise ValueError('dims should not contain same values.')
+
+        # sort dim and axis, so that axis is in order.
+        axis, dim = zip(*sorted(zip(axis, dim)))
+        # dims of the result array. This will be passed to Variable.expand_dims
+        all_dims = list(self.dims)
+        [all_dims.insert(a, d) for a, d in zip(axis, dim)]
```
shoyer: Use a loop in cases where you throw away the result of a comprehension, e.g.,
```
for a, d in zip(axis, dim):
    all_dims.insert(a, d)
```

xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
...
+
+        # If dims includes a label of a scalar variables,
+        # it will be promoted to a 1D coordinate consisting of a single value.
+        coords = OrderedDict(self.coords)
```
shoyer: Use `self._coords`, which is an OrderedDict of `xarray.Variable` objects, instead of `self.coords`

xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
...
+        for d in dim:
+            if d in coords.keys():
```
shoyer: Just check `if d in coords:`

xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
...
+                coords[d] = np.atleast_1d(coords[d])
```
shoyer: Instead of calling `np.atleast_1d`, use something like `Variable.expand_dims` to ensure this plays nicely with with non-numpy arrays.

xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
         ds = self._to_temp_dataset().swap_dims(dims_dict)
         return self._from_temp_dataset(ds)
 
+    def expand_dims(self, dim=None, axis=None):
```
shoyer: Consider implementing this on `xarray.Dataset` instead (looping over all data variables), using `_to_temp_dataset()`/`_from_temp_dataset()` to write the `DataArray` method: see `swap_dims` above for an example.

xarray/core/dataarray.py
```diff
@@ -840,6 +840,99 @@ def swap_dims(self, dims_dict):
...
+            inserted to the start of the result array.
+
+        Returns
+        -------
+        expanded : same type as caller
+            This object, but with an additional dimension(s).
+
+        Raises
+        -------
+        ValueError:
```
shoyer: It's good form to document when functions can raise, but all of these conditions are already implied by the documentation for the `axis` argument, so I don't think you need to include this section. Essentially every Python function can raise `ValueError` when passed invalid arguments.

shoyer:
> The proposed API is similar to numpy.expand_dims and slightly different from Variables.expand_dims, which requires whole sequences of dims of the result array.

This looks great, but to avoid confusion, can we rename the existing `Variable.expand_dims` method? Maybe `with_dims` would be a clearer name.

> My concern is that I do not yet fully understand the lazy data manipulation in xarray.
Does Variable.expand_dims do it?

Yes, `Variable.expand_dims` is already lazy on dask arrays.

xarray/core/dataarray.py
fujiisoup: What is the reason to implement in `xarray.Dataset`?  Does it mean that we would also support `Dataset.expand_dims`? 

I think I need to make another design decision.

`axis` argument for `xarray.Dataset.expand_dims(dim, axis)` is somewhat problematic because `data_var`s are not necessarily the same dimension size.

It would be reasonable to raise an ValueError if any of `data_var` has smaller dimension that is required by `axis`?

xarray/core/dataarray.py
fujiisoup: OK. I will remove this default argument.

xarray/core/dataarray.py
shoyer: > What is the reason to implement in xarray.Dataset? Does it mean that we would also support Dataset.expand_dims?

Yes, exactly.

> It would be reasonable to raise an ValueError if any of data_var has smaller dimension that is required by axis?

Yes, of course. In practice, I suspect most people will either keep the default (to expand dims at the start) or use negative integers (to expand dims at the end).

fujiisoup:
> Maybe with_dims would be a clearer name.
How about `reshape`? 
I think verb-name gives more intuitition about the action of the method.
The current implementation of `Variable.expand_dims` actually does this when `Ordered_dict` is passed.

fujiisoup:
I moved `expand_dims` to `Dataset`.

How can I add API docs? 
Only to add a single line to api.rst?

> How about `reshape`?

I changed my mind. `reshape` sounds too general. How about `append_dims` or `attach_dims` ?

shoyer:
> How can I add API docs?
> Only to add a single line to api.rst?

Yes, that's correct.

> I changed my mind. reshape sounds too general. How about append_dims or attach_dims ?

Indeed, I would rather not do reshape, because this method does not support arbitrary reshaping (e.g., length 100 vector-> 10x10 matrix). Instead, it assigns dimensions using array broadcasting and transpose, which unlike reshape does not ever copy data. Maybe `set_dims` would be better? I also would be OK with `attach_dims`. `append_dims` makes it sound like dimensions are always appended (they aren't).

Actually, `reshape` could indeed be a nice method to add to `Dataset`/`DataArray`, as an alternative to `stack`/`unstack`.

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+            axis = list(range(len(dim)))
+
+        # Error checking
+        # Make sure the length of axis and dims are identical
+        if len(dim) != len(axis):
+            raise ValueError('lengths of dim and axis should be identical.')
+        # make sure any of dim does not already exist
+        for d in dim:
+            if d in self.dims:
+                raise ValueError('{dim} already exists.'.format(dim=d))
```
shoyer: Include the word "dimension" so it's clear what the problem is: `'dimension {dim} already exists.'`

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+
+        # make sure there is no duplicates in axis or dims
+        if len(axis) != len(set(axis)):
+            raise ValueError('axis should not contain same values.')
```
shoyer: Grammar: same values -> duplicate values

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if len(dim) != len(set(dim)):
+            raise ValueError('dims should not contain same values.')
```
shoyer: same -> duplicate, as above

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+
+        # sort dim and axis, so that axis is in order.
+        axis, dim = zip(*sorted(zip(axis, dim)))
+
+        variables = OrderedDict()
+        for k, v in iteritems(self._variables):
+            if k in self._coord_names:
+                if k not in dim: # Do not change coordinates
```
shoyer: Switch the order of these conditions to avoid checking for a negative: it results in clearer code.

Also, [per PEP8](https://www.python.org/dev/peps/pep-0008/#inline-comments) include two spaces before any inline comments.

And per xarray terminology, call these "non-dimension coordinates": http://xarray.pydata.org/en/stable/data-structures.html#coordinates

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+            multiple axes are inserted. In this case, dim arguments should be
+            the same length list or None. If None is passed, all the axes will
+            be inserted to the start of the result array.
+
+        Returns
+        -------
+        expanded : same type as caller
+            This object, but with an additional dimension(s).
+        """
+        # Make sure user does not do `expand_dims(0)`
```
shoyer: This comment (and most of the others below) is simply describing what the code does. It would be better to omit it.

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if isinstance(dim, int):
+            raise ValueError('dim should be str or sequence of strs or dict')
+
+        # Make axis and dims a pair of lists
```
shoyer: delete

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if isinstance(dim, basestring):
+            dim = [dim]
+        if axis is not None and not isinstance(axis, (list, tuple)):
+            axis = [axis]
+
+        # infer None arguments.
```
shoyer: delete

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if axis is None:
+            axis = list(range(len(dim)))
+
+        # Error checking
+        # Make sure the length of axis and dims are identical
```
shoyer: delete

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if len(dim) != len(axis):
+            raise ValueError('lengths of dim and axis should be identical.')
+        # make sure any of dim does not already exist
```
shoyer: delete

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        for d in dim:
+            if d in self.dims:
+                raise ValueError('{dim} already exists.'.format(dim=d))
+
+        # make sure there is no duplicates in axis or dims
```
shoyer: delete

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,82 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if len(axis) != len(set(axis)):
+            raise ValueError('axis should not contain same values.')
+        if len(dim) != len(set(dim)):
+            raise ValueError('dims should not contain same values.')
+
+        # sort dim and axis, so that axis is in order.
```
shoyer: Note that this is for `.insert` below.

fujiisoup:
Thanks for the review.

I completely forgot the negative axis. Thanks.
It is supported by my last commit.

I also changed `Variable.expand_dims` to `Variable.set_dims`.

xarray/core/dataset.py
```diff
@@ -1599,55 +1599,50 @@ def expand_dims(self, dim, axis=None):
...
+                        raise ValueError('axis should not contain duplicate'
+                                         ' values.')
+                    # We need to sort them to make sure `axis` equals to the
+                    # axis positions of the result array.
+                    axis_pos, dim_pos = zip(*sorted(zip(axis_pos, dim)))
+
+                    all_dims = list(v.dims)
+                    for a, d in zip(axis_pos, dim_pos):
+                        all_dims.insert(a, d)
+                    print(all_dims)
```
shoyer: Remove

xarray/core/dataset.py
```diff
@@ -1599,55 +1599,50 @@ def expand_dims(self, dim, axis=None):
...
                 else:
-                    # If dims includes a label of a scalar variables,
-                    # it will be promoted to a 1D coordinate consisting
-                    # of a single value.
-                    variables[k] = v.expand_dims(k)
+                    axis_pos = [a if a >= 0 else len(v.dims) + len(axis) + a
+                                for a in axis]
+                    if len(axis_pos) != len(set(axis_pos)):
+                        raise ValueError('axis should not contain duplicate'
+                                         ' values.')
```
shoyer: You also need to check for out of bound axes, e.g., if `result_ndim = len(v.dims) + len(axis)`, then all axes `a` must satisfy `-result_ndim <= a < result_ndim`.

This is probably easier before making axis positive so you can give an array message with the original, incorrect axis number.

xarray/core/dataset.py
```diff
@@ -1599,55 +1599,50 @@ def expand_dims(self, dim, axis=None):
...
+                    # We need to sort them to make sure `axis` equals to the
+                    # axis positions of the result array.
+                    axis_pos, dim_pos = zip(*sorted(zip(axis_pos, dim)))
```
shoyer: Your code currently zips and unzips (here) and rezips (in the loop). You can remove two of these for a small speedup.

xarray/tests/test_dataarray.py
```diff
@@ -913,6 +913,88 @@ def test_swap_dims(self):
...
+            array.expand_dims(dim=['x'])
+        # raise if duplicate
+        with self.assertRaises(ValueError):
+            array.expand_dims(dim=['y', 'y'])
+        with self.assertRaises(ValueError):
+            array.expand_dims(dim=['y', 'z'], axis=[1, 1])
+        with self.assertRaises(ValueError):
+            array.expand_dims(dim=['y', 'z'], axis=[2, -2])
+
+        # Working test
```
shoyer: Consider breaking these into smaller test methods, e.g., `test_expand_dims_errors`. It's  easier to debug when tests each one thing.

fujiisoup:
Thanks for the careful review.

+ Now it raises `IndexError` if any axis is out of bounds.
   `ValueError` is OK for the other part?
+ Removed unnecessary `print` line.
+ Removed the `unzip` -> `rezip` sequences.
+ Divide `test_dataarray.test_expand_dims` into small 2 tests.

xarray/core/variable.py
```diff
@@ -706,8 +706,9 @@ def transpose(self, *dims):
         data = ops.transpose(self.data, axes)
         return type(self)(dims, data, self._attrs, self._encoding, fastpath=True)
 
-    def expand_dims(self, dims, shape=None):
-        """Return a new variable with expanded dimensions.
+    def set_dims(self, dims, shape=None):
```
shoyer: Can you keep around `Variable.expand_dims` as an alias for `set_dims` with a deprecation warning?

e.g., see the deprecation warning for `broadcast_arrays`: https://github.com/pydata/xarray/blob/93d6963315026f87841c7cf39cc39bb78f555345/xarray/core/alignment.py#L531

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,90 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        if axis is None:
+            axis = list(range(len(dim)))
+
+        if len(dim) != len(axis):
+            raise ValueError('lengths of dim and axis should be identical.')
+        for d in dim:
+            if d in self.dims:
+                raise ValueError(
+                            'Dimension {dim} already exists.'.format(dim=d))
+            if d in self._variables.keys() and \
```
shoyer: Per PEP8, use parentheses when possible instead of explicit continuations with `\`, e.g.,
```
if (d in a_very_long_expression and
        not something_else_very_long):
```

Also, check `d in self._variables` directly instead of using `.keys()` (which is a list Python 2, making for very slow lookups).

xarray/core/dataset.py
```diff
@@ -1575,6 +1575,90 @@ def swap_dims(self, dims_dict, inplace=False):
...
+        variables = OrderedDict()
+        for k, v in iteritems(self._variables):
+            if k not in dim:
+                if k in self._coord_names:  # Do not change coordinates
+                    variables[k] = v
+                else:
+                    result_ndim = len(v.dims) + len(axis)
+                    for a in axis:
+                        if a < -result_ndim or result_ndim - 1 < a:
+                            raise IndexError(
```
shoyer: Either `IndexError` or `ValueError` would be appropriate here. NumPy has a new error type `np.AxisError` but it's only in the latest, unreleased version of NumPy.

fujiisoup:
I added a section to `reshaping.rst`, which shows an example (default) usage of `expand_dims` and `squeeze`.

shoyer:
This looks great. Thank you @fujisoup!

fujiisoup:
Oops. Found a typo in reshaping.rst. I'll send another PR.

Issue #1344, opened 2017-03-30
Dask Persist


mrocklin:
It would be convenient to load constituent dask.arrays into memory as dask.arrays rather than as numpy arrays.  This would help with distributed computations where we want to load a large amount of data into distributed memory once and then iterate on the full xarray dataset repeatedly without reloading from disk every time.

We can probably solve this from either side:

1.  XArray could make a `.persist` method that replaced all of its dask.arrays with a persisted version of that array

```python
import dask

dset.x, dset.y, dset.z = dask.persist(dset.x, dset.y, dset.z)
```

2.  We could look into the Dask duck type solution again https://github.com/dask/dask/pull/1068

cc @shoyer @jcrist @rabernat @pwolfram

shoyer:
I'm happy with either or both of these solutions.

rabernat:
since we already have `.load()`, I think adding `.persist()` directly to xarray is the simplest way to go. Ideally xarray users should be able to avoid interacting directly with the dask api. (It's a beautiful api, don't get me wrong! but new users are easily overwhelmed by multiple interacting packages.)

mrocklin:
We do eventually want to support some sort of duck typing.  This way people can do things like the following while still benefiting from shared intermediates:

    x, y, z = dask.persist(my_df, my_arrray, my_xarray)

But this may not happen as quickly as a `.persist` method.  Also we've added `.persist` to the Base dask collections class, so it's part of the standard dask API at this point.

rabernat:
In this example, are x, y and z dask objects or xarray objects?

shoyer:
> In this example, are x, y and z dask objects or xarray objects?

I think the idea is that they are mixed types, e.g., a dask-dataframe, a dask-array and an xarray Dataset or DataARray.

PR #1349, opened 2017-04-03
Add persist method to DataSet

mrocklin:
Fixes https://github.com/pydata/xarray/issues/1344

 - [x] closes #xxxx
 - [x] tests added / passed
 - [x] whatsnew entry  (not sure what to do here, is there a new section?  It looks like the last release was yesterday)

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -422,6 +422,7 @@ Dataset methods
    Dataset.from_dict
    Dataset.close
    Dataset.compute
+   Dataset.persist
    Dataset.load
    Dataset.chunk
    Dataset.filter_by_attrs
@@ -447,6 +448,7 @@ DataArray methods
    DataArray.from_cdms2
    DataArray.from_dict
    DataArray.compute
+   DataArray.persist
    DataArray.load
    DataArray.chunk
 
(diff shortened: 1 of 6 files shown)


mrocklin:
I suspect that there is more to do here, but I wanted to get the ball rolling on this.

cc @shoyer @rabernat @pwolfram

rabernat:
Would it make sense to have `persist` for DataArrays as well? My opinion is: yes.

shoyer:
Thanks @mrocklin! This is a great start.

Yes, it would be great to have this on `DataArray` as well. The implementation would be a two liner:
```
    def persist(self):
        ds = self._to_temp_dataset().persist()
```

I just pushed a commit to master adding a "what's new" template for v0.9.3.

mrocklin:
Added DataArray.persist and a whats new entry

rabernat:
Would also probably be good to document this a bit in the section on dask:
https://github.com/pydata/xarray/blob/master/doc/dask.rst

mrocklin:
Added docs.  Also removed section on dask.array not providing automatic chunk alignment (this is now supported)

rabernat:
This looks good to go from my point of view!

shoyer:
Thanks @mrocklin!

I'm glad to see your first commit on the xarray side :).

mrocklin:
Happy to contribute :)

rabernat:
> I'm glad to see your first commit on the xarray side :).

Hopefully the first of many!

Issue #367, opened 2015-03-10
ds['time.time'] is broken


shoyer:
As noted in https://github.com/xray/xray/issues/364:
```
In [32]: t = pd.date_range('2000-01-01', periods=10, freq='H')

In [33]: time = xray.DataArray(t, name='time', dims='time')

In [34]: time['time.time']
Out[34]:
<xray.DataArray 'time' (time: 10)>
array(['1999-12-31T16:00:00.000000000-0800',
       '1999-12-31T17:00:00.000000000-0800',
       '1999-12-31T18:00:00.000000000-0800',
       '1999-12-31T19:00:00.000000000-0800',
       '1999-12-31T20:00:00.000000000-0800',
       '1999-12-31T21:00:00.000000000-0800',
       '1999-12-31T22:00:00.000000000-0800',
       '1999-12-31T23:00:00.000000000-0800',
       '2000-01-01T00:00:00.000000000-0800',
       '2000-01-01T01:00:00.000000000-0800'], dtype='datetime64[ns]')
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2000-01-01T01:00:00 ...
```

It should return an array of `datetime.time` objects.

shoyer:
We could hack around this, but ultimately this is an implication of a design choice I made for the internals of `xray.DataArray`, which stores its data in an internal `xray.Dataset` object, using `self.name` as the key for the variable with the array's data. This means it is currently impossible to have a DataArray with the same name as on of its coordinates but different values.

The real solution here is to do some internal refactoring to make this possible. My current thinking is to add a constant `THIS_ARRAY = object()` in `xray.core.dataarray`, and then use `THIS_ARRAY` instead of `self.name` in all internal methods that mess with the private dataset that stores the data on DataArray objects.

shoyer:
To clarify: the issue here is that `time.time` involves the name "time" twice. If the data array is not called "time", then this works properly.

jhamman:
@shoyer, just out of curiosity, was this broken before the changes to the virtual variable syntax in #357?

shoyer:
@jhamman it was not broken before -- `ds['time.time']` returned a variable named `'time.time'`, which does not conflict with `'time'`.

PR #1356, opened 2017-04-06
Add DatetimeAccessor for accessing datetime fields via `.dt` attribute

### LINKED ISSUE #358: add .dt and .str accessors to DataArray (like pandas.Series)
Something like this should work: `observed.time.dt.to_pydatetime()`.

This could simply be wrappers of the pandas methods.

darothen:
- [x] Partially closes #358 
 - [x] tests added / passed
 - [x] whatsnew entry

This uses the `register_dataarray_accessor` to add attributes similar to those in pandas` timeseries which let the users quickly access datetime fields from an underlying array of datetime-like values. The referenced issue (#358) also asks about adding similar accessors for `.str`, but this is a more complex topic - I think a compelling use-case would help in figuring out what the critical functionality is

## Virtual time fields

Presumably this could be used to augment `Dataset._get_virtual_variable()`. A **season** field would need to be added as a special field to the accessor.

### DIFF
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -88,7 +88,22 @@ For more details, read the pandas documentation.
 Datetime components
 -------------------
 
-xarray supports a notion of "virtual" or "derived" coordinates for
+Similar `to pandas`_, the components of datetime objects contained in a
+given ``DataArray`` can be quickly computed using a special ``.dt`` accessor.
+
+.. _to pandas: http://pandas.pydata.org/pandas-docs/stable/basics.html#basics-dt-accessors
+
+.. ipython:: python
+
+    time = time = pd.date_range('2000-01-01', freq='6H', periods=365 * 4)
+    ds = xr.Dataset({'foo': ('time', np.arange(365 * 24)), 'time': time})
+    ds.time.dt.hour
+    ds.time.dt.dayofweek
+
+The ``.dt`` accessor works on both coordinate dimensions as well as
+multi-dimensional data.
+
+xarray also supports a notion of "virtual" or "derived" coordinates for
 `datetime components`__ implemented by pandas, including "year", "month",
 "day", "hour", "minute", "second", "dayofyear", "week", "dayofweek", "weekday"
 and "quarter":
@@ -100,7 +115,8 @@ __ http://pandas.pydata.org/pandas-docs/stable/api.html#time-date-components
     ds['time.month']
     ds['time.dayofyear']
 
-xarray adds ``'season'`` to the list of datetime components supported by pandas:
+For use as a derived coordinate, xarray adds ``'season'`` to the list of
+datetime components supported by pandas:
 
 .. ipython:: python
 
@@ -124,7 +140,7 @@ calculate the mean by time of day:
 
 For upsampling or downsampling temporal resolutions, xarray offers a
 :py:meth:`~xarray.Dataset.resample` method building on the core functionality
-offered by the pandas method of the same name. Resample uses essentialy the
+offered by the pandas method of the same name. Resample uses essentially the
 same api as ``resample`` `in pandas`_.
 
 .. _in pandas: http://pandas.pydata.org/pandas-docs/stable/timeseries.html#up-and-downsampling
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -121,6 +121,7 @@ datetime components supported by pandas:
 .. ipython:: python
 
     ds['time.season']
+    ds['time'].dt.season
 
 The set of valid seasons consists of 'DJF', 'MAM', 'JJA' and 'SON', labeled by
 the first letters of the corresponding months.
--- a/xarray/core/accessors.py
+++ b/xarray/core/accessors.py
@@ -9,31 +9,6 @@ from .pycompat import dask_array_type
 import numpy as np
 import pandas as pd
 
-
-def _get_date_field_from_dask(values, name):
-    """Specialized date field accessor for data contained in dask arrays
-    """
-    from dask import delayed
-    from dask.array import from_delayed, map_blocks
-    from dask.dataframe import from_array
-
-    @delayed
-    def _getattr_from_dask(accessor):
-        attr_values = getattr(accessor, name).values
-        return attr_values.compute()
-
-    def _ravel_and_access(darr):
-        raveled = darr.ravel()
-        raveled_as_series = from_array(raveled[..., np.newaxis],
-                                       columns=['_raveled_data', ])
-        field_values = _getattr_from_dask(raveled_as_series['_raveled_data'].dt)
-        field_values = from_delayed(field_values, shape=raveled.shape,
-                                    dtype=raveled.dtype)
-        return field_values.reshape(darr.shape).compute()
-
-    return map_blocks(_ravel_and_access, values, dtype=values.dtype)
-
-
 def _get_date_field(values, name):
     """Indirectly access pandas' libts.get_date_field by wrapping data
     as a Series and calling through `.dt` attribute.
@@ -46,12 +21,18 @@ def _get_date_field(values, name):
         Name of datetime field to access
 
     """
-    if isinstance(values, dask_array_type):
-        return _get_date_field_from_dask(values, name)
-    else:
+    def _access_through_series(values, name):
         values_as_series = pd.Series(values.ravel())
-    field_values = getattr(values_as_series.dt, name).values
-    return field_values.reshape(values.shape)
+        field_values = getattr(values_as_series.dt, name).values
+        return field_values.reshape(values.shape)
+
+    if isinstance(values, dask_array_type):
+        from dask.array import map_blocks
+        out_dtype = object if name == "weekday_name" else np.int64
+        return map_blocks(_access_through_series,
+                          values, name, dtype=out_dtype)
+    else:
+        return _access_through_series(values, name)
 
 
 @register_dataarray_accessor('dt')
@@ -83,12 +64,6 @@ class DatetimeAccessor(object):
                             "DataArray with datetime64 or timedelta64 dtype")
         self._obj = xarray_obj
         self._dt = self._obj.data
-
-    _field_ops = ['year', 'month', 'day', 'hour', 'minute', 'second',
-                  'weekofyear', 'week', 'weekday', 'dayofweek',
-                  'dayofyear', 'quarter', 'days_in_month',
-                  'daysinmonth', 'microsecond',
-                  'nanosecond']
 
     def _tslib_field_accessor(name, docstring=None):
         def f(self):
--- a/xarray/core/accessors.py
+++ b/xarray/core/accessors.py
@@ -30,6 +30,7 @@ def _access_through_series(values, name):
         field_values = getattr(values_as_series.dt, name).values
     return field_values.reshape(values.shape)
 
+
 def _get_date_field(values, name):
     """Indirectly access pandas' libts.get_date_field by wrapping data
     as a Series and calling through `.dt` attribute.
@@ -132,3 +133,7 @@ class DatetimeAccessor(object):
     daysinmonth = days_in_month
 
     season = _tslib_field_accessor("season", "Season of the year (ex: DJF)")
+
+    time = _tslib_field_accessor(
+        "time", "Timestamps corresponding to datetimes"
+    )
--- a/xarray/core/accessors.py
+++ b/xarray/core/accessors.py
@@ -60,7 +60,7 @@ class DatetimeAccessor(object):
                   'daysinmonth', 'microsecond',
                   'nanosecond']
 
-    def _tslib_field_accessor(name, field, docstring=None):
+    def _tslib_field_accessor(name, docstring=None):
         def f(self):
             from .dataarray import DataArray
             result = _get_date_field(self._dt, name)
@@ -71,40 +71,36 @@ class DatetimeAccessor(object):
         f.__doc__ = docstring
         return property(f)
 
-
-    year = _tslib_field_accessor('year', 'Y', "The year of the datetime")
+    year = _tslib_field_accessor('year', "The year of the datetime")
     month = _tslib_field_accessor(
-        'month', 'M', "The month as January=1, December=12"
+        'month', "The month as January=1, December=12"
     )
-    day = _tslib_field_accessor('day', 'D', "The days of the datetime")
-    hour = _tslib_field_accessor('hour', 'h', "The hours of the datetime")
-    minute = _tslib_field_accessor('minute', 'm', "The minutes of the datetime")
-    second = _tslib_field_accessor('second', 's', "The seconds of the datetime")
+    day = _tslib_field_accessor('day', "The days of the datetime")
+    hour = _tslib_field_accessor('hour', "The hours of the datetime")
+    minute = _tslib_field_accessor('minute', "The minutes of the datetime")
+    second = _tslib_field_accessor('second', "The seconds of the datetime")
     microsecond = _tslib_field_accessor(
-        'microsecond', 'us', "The microseconds of the datetime"
+        'microsecond', "The microseconds of the datetime"
     )
     nanosecond = _tslib_field_accessor(
-        'nanosecond', 'ns', "The nanoseconds of the datetime"
+        'nanosecond', "The nanoseconds of the datetime"
     )
     weekofyear = _tslib_field_accessor(
-        'weekofyear', 'woy', "The week ordinal of the year"
+        'weekofyear', "The week ordinal of the year"
     )
     week = weekofyear
     dayofweek = _tslib_field_accessor(
-        'dayofweek', 'dow', "The day of the week with Monday=0, Sunday=6"
+        'dayofweek', "The day of the week with Monday=0, Sunday=6"
     )
     weekday = dayofweek
 
     weekday_name = _tslib_field_accessor(
-        'weekday_name', 'weekday_name',
-        "The name of day in a week (ex: Friday)"
+        'weekday_name', "The name of day in a week (ex: Friday)"
     )
 
-    dayofyear = _tslib_field_accessor(
-        'dayofyear', 'doy', "The ordinal day of the year"
-    )
-    quarter = _tslib_field_accessor('quarter', 'q', "The quarter of the date")
+    dayofyear = _tslib_field_accessor('dayofyear', "The ordinal day of the year")
+    quarter = _tslib_field_accessor('quarter', "The quarter of the date")
     days_in_month = _tslib_field_accessor(
-        'days_in_month', 'dim', "The number of days in the month"
+        'days_in_month', "The number of days in the month"
     )
     daysinmonth = days_in_month
--- a/xarray/core/accessors.py
+++ b/xarray/core/accessors.py
@@ -23,6 +23,32 @@ def _get_date_field(array, name):
     field_values = getattr(series.dt, name).values
     return field_values.reshape(array.shape)
 
+_date_field_ops_and_descrs = [
+    ('year', "The year of the datetime"),
+    ('month', "The month as January=1, December=12"),
+    ('day', "The days of the datetime"),
+    ('hour', "The hours of the datetime"),
+    ('minute', "The minutes of the datetime"),
+    ('second', "The seconds of the datetime"),
+    ('microsecond', "The microseconds of the datetime"),
+    ('nanosecond', "The nanoseconds of the datetime"),
+    ('weekofyear',  "The week ordinal of the year"),
+    ('week', "The week ordinal of the year"),
+    ('weekday', "The day of the week with Monday=0, Sunday=6"),
+    ('weekday_name',
+     "The name of day in a week (ex: Friday)"),
+    ('dayofweek', "The day of the week with Monday=0, Sunday=6"),
+    ('dayofyear',  "The ordinal day of the year"),
+    ('quarter', "The quarter of the date"),
+    ('days_in_month', "The number of days in the month"),
+    ('daysinmonth', "The number of days in the month"),
+]
+
+def inject_date_field_accessors(cls):
+    for op, descr in _date_field_ops_and_descrs:
+        prop = cls._date_field_accessor(op, docstring=descr)
+        setattr(cls, op, prop)
+
 
 @register_dataarray_accessor('dt')
 class DatetimeAccessor(object):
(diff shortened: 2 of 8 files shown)


xarray/core/accessors.py
```diff
@@ -0,0 +1,104 @@
...
+            datetimes_asi8 = self._obj.values.view('i8')
+            self._dt = datetimes_asi8
+        return self._dt
+
+    # Modified from https://github.com/pandas-dev/pandas/pandas/tseries/index.py#L59
+    def _tslib_field_accessor(name, field, docstring=None):
+        def f(self):
+            from .dataarray import DataArray
+            values = self.dt
+            result = libts.get_date_field(values, field)
```
shoyer: This is using private pandas API, so it's liable to break in the future. Instead, let's wrap things in a `pandas.Series`. Something like this should do the trick: 
```python
def get_dt_field(array: np.ndarray, name: str):
    series = pd.Series(array.ravel())
    field_values = getattr(series.dt, name).values
    return field_values.reshape(array.shape)
```

xarray/core/accessors.py
```diff
@@ -0,0 +1,104 @@
...
+    """Access datetime fields for DataArrays with datetime-like dtypes.
+
+     Similar to pandas, fields can be accessed through the `.dt` attribute
+     for applicable DataArrays:
+
+        >>> ds = xarray.Dataset({'time': pd.date_range(start='2000/01/01',
+        ...                                            freq='D', periods=100)})
+        >>> ds.time.dt
+        <xarray.core.accessors.DatetimeAccessor at 0x10c369f60>
+        >>> ds.time.dt.dayofyear[5]
```
shoyer: I think this should be `[:5]`, not `[5]`?

xarray/core/accessors.py
```diff
@@ -0,0 +1,104 @@
...
+        <xarray.DataArray 'dayofyear' (time: 5)>
+        array([1, 2, 3, 4, 5], dtype=int32)
+        Coordinates:
+          * time     (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 ...
+
+     All of the pandas fields are accessible here. Note that these fields are not
+     calendar-aware; if your datetimes are encoded with a non-Gregorian calendar
+     (e.g. a 360-day calendar), then some fields like `dayofyear` may not be
```
shoyer: add "using netcdftime" just to be super explicit about where special calendars come from.

xarray/core/accessors.py
```diff
@@ -0,0 +1,104 @@
...
+     accurate.
+
+     """
+    def __init__(self, xarray_obj):
+        if not is_datetime_like(xarray_obj.dtype):
+            raise TypeError("'dt' accessor only available for "
+                            "DataArray with datetime64 dtype")
```
shoyer: add "or timedelta64 dtype"

xarray/tests/test_accessors.py
```diff
@@ -0,0 +1,43 @@
...
+        self.data = xr.DataArray(data, coords=[lons, lats, self.times],
+                                 dims=['lon', 'lat', 'time'], name='data')
+
+    def test_field_access(self):
+        years = self.times.year
+        months = self.times.month
+        days = self.times.day
+        hours = self.times.hour
+
+        self.assertArrayEqual(years, self.data.time.dt.year)
```
shoyer: Use `self.assertDataArrayEqual` to check that fields provide data arrays with the appropriate metadata.

xarray/tests/test_accessors.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertArrayEqual(months, self.data.time.dt.month)
+        self.assertArrayEqual(days, self.data.time.dt.day)
+        self.assertArrayEqual(hours, self.data.time.dt.hour)
+
+    def test_not_datetime_type(self):
+        nontime_data = self.data.copy()
+        nontime_data['time'].values = \
```
shoyer: please avoid explicit backlashes for continuation if possible, per PEP8.

xarray/tests/test_accessors.py
```diff
@@ -0,0 +1,43 @@
...
+            np.arange(len(self.data.time)).astype('int8')
+        with self.assertRaisesRegexp(TypeError, 'dt'):
+            nontime_data.time.dt.year
```
shoyer: missing newline on last line

darothen:
@shoyer I corrected things based on your comments. The last commit is an attempt to refactor things to match the way that methods like rolling/groupby functions are injected into the class; this might be totally superfluous here, but I thought it was worth trying.

xarray/core/accessors.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, xarray_obj):
+        if not is_datetime_like(xarray_obj.dtype):
+            raise TypeError("'dt' accessor only available for "
+                            "DataArray with datetime64 or timedelta64 dtype")
+        self._obj = xarray_obj
+        self._dt = None
+
+    @property
+    def dt(self):
+        """Attribute to cache a view of the underlying datetime-like
```
shoyer: I don't think we want to expose this as public facing API. So I would get ride of this property, probably.

xarray/core/accessors.py
```diff
@@ -0,0 +1,103 @@
...
+    ('weekday_name',
+     "The name of day in a week (ex: Friday)"),
+    ('dayofweek', "The day of the week with Monday=0, Sunday=6"),
+    ('dayofyear',  "The ordinal day of the year"),
+    ('quarter', "The quarter of the date"),
+    ('days_in_month', "The number of days in the month"),
+    ('daysinmonth', "The number of days in the month"),
+]
+
+def inject_date_field_accessors(cls):
```
shoyer: This is fine, but I actually did like what you did before -- and we'll probably be switching xarray to use that style in the future, instead of injecting things with a loop. Using a function to define the method directly on the class is much more explicit.

xarray/core/accessors.py
```diff
@@ -0,0 +1,103 @@
...
+        self._obj = xarray_obj
+        self._dt = None
+
+    @property
+    def dt(self):
+        """Attribute to cache a view of the underlying datetime-like
+        array for passing to pandas.tslib for date_field operations
+        """
+        if self._dt is None:
+            self._dt = self._obj.values
```
shoyer: Something possible worth thinking about, at least a little: What happens if `.dt` is called on an object that stores its values in dask array? Ideally, we would handle this transparently, by either calling `_get_date_field` directly (on numpy arrays) or using [`dask.array.map_blocks`](http://dask.pydata.org/en/latest/array-api.html#dask.array.core.map_blocks) to call `_get_date_field` (on dask arrays).

xarray/core/accessors.py
darothen: Cool, I'll revert back to the explicit style.

xarray/core/accessors.py
darothen: Yeah, I see the potential for an issue here, but from some limited testing I'm not sure where it will arise. For instance, consider this contrived example:

``` python
import numpy as np
import pandas as pd
import xarray as xr
from dask import array as da

nt = 100000
data = np.random.rand(10, 10, nt)
lons = np.linspace(0, 11, 10)
lats = np.linspace(0, 20, 10)
times = pd.date_range(start="2000/01/01", freq='D', periods=nt)

datest = xr.DataArray(data, coords=[lons, lats, times], 
                  dims=['lon', 'lat', 'time'], name='data')

tc = da.from_array(da2.time.values, chunks=5000)
datest['time2'] = (('time', ), tc)
datest.time2.dt.month
```

This works just fine. I tried creating an example where I transform the data underlying `datest['time']` into a dask array, but I can't create one that doesn't eagerly load the data into memory, no matter how I try to chunk it. Do dimensions automatically load dask arrays?

Chunking on `datest` itself was not a problem, either again, because accessing `datest.time` always gave an in-memory array.

shoyer: You can only put a dask array in xarray if it isn't an dimension coordinate, so in practice these are pretty rare for datetime64 variables but they do come up occasionally (e.g., consider a 2D variable `time` for a collection of weather forecasts with dimensions for reference time and time since the start of the forecast).

darothen: Gotcha. A few more lines of code to add to the example before:

``` python
times_arr = np.random.choice(times, size=(10, 10, nt))

dstest = datest.to_dataset()
dstest['all_times'] = (('lon', 'lat', 'time'), times_arr)

dstest = dstest.chunk({'lon': 5, 'lat': 5, 'time': 5000})
```

So now we're definitely holding a dask array:

``` python
>>> dstest.all_times
<xarray.DataArray 'all_times' (lon: 10, lat: 10, time: 100000)>
dask.array<xarray-..., shape=(10, 10, 100000), dtype=datetime64[ns], chunksize=(5, 5, 5000)>
Coordinates:
  * lon      (lon) float64 0.0 1.222 2.444 3.667 4.889 6.111 7.333 8.556 ...
  * lat      (lat) float64 0.0 2.222 4.444 6.667 8.889 11.11 13.33 15.56 ...
  * time     (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 ...
    time2    (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 ...
```

but our datetime accessors still work just fine:

``` python
>>> dstest.all_times.dt.month
<xarray.DataArray 'month' (lon: 10, lat: 10, time: 100000)>
array([[[ 3,  5, ..., 11,  5],
        [ 2, 12, ...,  8,  4],
        ..., 
        [ 9, 11, ...,  6, 11],
        [ 6,  1, ..., 12,  8]],
...
```

[Here](https://github.com/darothen/xarray/blob/8c38bf9c77f97c3e55ff175b27e4e9f659a2ba3f/xarray/core/accessors.py#L55) I cache the `values` of the DataArray passed in, and that's what the accessor checks to figure out the requested field. Are those always eagerly evaluated? If so, then it would make sense why everything works out okay - we've stored an array in memory, and we're converting that to a Pandas Series. 

Things also work fine if I create a chunked dask array from `times_arr` above and store it directly in the Dataset.

shoyer: Yes, the current version of the accessor works, but only by loading the entire dask array into memory as a NumPy array first. That's a no-no: operations on dask arrays should be lazy or raise an error if that isn't possible..

darothen: Ok I understand now. I'll see what I can do about safely handling dask arrays here; I know there's plenty of examples to emulate across the xarray code base.

shoyer: At the very least, call `fail_on_dask_array_input`: https://github.com/pydata/xarray/blob/master/xarray/core/duck_array_ops.py#L55

darothen:
Updated with support for multi-dimensional time data stored as dask array.

xarray/core/accessors.py
```diff
@@ -0,0 +1,136 @@
...
+    from dask import delayed
+    from dask.array import from_delayed, map_blocks
+    from dask.dataframe import from_array
+
+    @delayed
+    def _getattr_from_dask(accessor):
+        attr_values = getattr(accessor, name).values
+        return attr_values.compute()
+
+    def _ravel_and_access(darr):
```
shoyer: This is more complicated than it needs to, as evidenced by the nested calls to compute. The great thing about `map_blocks` is that you can make everything eager, e.g.,
```python
def _ravel_and_access(darr):
    values_as_series = pd.Series(values.ravel())
    field_values = getattr(values_as_series.dt, name).values
    return field_values.reshape(values.shape)
```
That's it!

xarray/core/accessors.py
```diff
@@ -0,0 +1,136 @@
...
+        raveled = darr.ravel()
+        raveled_as_series = from_array(raveled[..., np.newaxis],
+                                       columns=['_raveled_data', ])
+        field_values = _getattr_from_dask(raveled_as_series['_raveled_data'].dt)
+        field_values = from_delayed(field_values, shape=raveled.shape,
+                                    dtype=raveled.dtype)
+        return field_values.reshape(darr.shape).compute()
+
+    return map_blocks(_ravel_and_access, values, dtype=values.dtype)
```
shoyer: `dtype` needs to be the dtype of the return value, not the input. For most of these accessors, this is just `np.int64` (for strings, use `dtype=object`).

xarray/tests/test_accessors.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertDataArrayEqual(months, self.data.time.dt.month)
+        self.assertDataArrayEqual(days, self.data.time.dt.day)
+        self.assertDataArrayEqual(hours, self.data.time.dt.hour)
+
+    def test_not_datetime_type(self):
+        nontime_data = self.data.copy()
+        int_data = np.arange(len(self.data.time)).astype('int8')
+        nontime_data['time'].values = int_data
+        with self.assertRaisesRegexp(TypeError, 'dt'):
+            nontime_data.time.dt.year
```
shoyer: I would use simply `nontime_data.time.dt` to make the test a little more stringent.

xarray/tests/test_accessors.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @requires_dask
+    def test_dask_field_access(self):
+        import dask.array as da
+
+        # Safely pre-compute comparison fields by passing through Pandas
+        # machinery
+        def _getattr_and_reshape(arr, attr):
```
shoyer: It's OK if this unit test relies on the correctness of computing date fields on NumPy arrays -- you already check that in other tests. So I would get rid of this and simply rely on these fields already working, e.g., just use `self.times_arr.dt.year`.

xarray/core/accessors.py
```diff
@@ -0,0 +1,136 @@
...
+
+     """
+    def __init__(self, xarray_obj):
+        if not is_datetime_like(xarray_obj.dtype):
+            raise TypeError("'dt' accessor only available for "
+                            "DataArray with datetime64 or timedelta64 dtype")
+        self._obj = xarray_obj
+        self._dt = self._obj.data
+
+    _field_ops = ['year', 'month', 'day', 'hour', 'minute', 'second',
```
shoyer: you don't need this variable anymore.

xarray/core/accessors.py
darothen: Wow! The time spent on tackling this project is well worth it just for this tidbit alone. I'll circle back to clean up this stuff within the next day or so.

darothen:
Finished clean-up, added some documentation, etc. I mangled resolving a merge conflict with my update to `whats-new.rst` (5ae4e08) in terms of the commit text, but other than that I think we're getting closer to finishing this.

wrt to the virtual variables, I think some more thinking is necessary so we can come up with a plan of approach. Do we want to deprecate this feature entirely? Do we just want to wrap the datetime component virtual variables to the `.dt` accessor if they're datetime-like? We could very easily do the latter for 0.9.3, but maybe we should target a future major release to deprecate the virtual variables and instead encourage adding a few specialized (but commonly-used) accessors to xarray?

shoyer:
> wrt to the virtual variables, I think some more thinking is necessary so we can come up with a plan of approach. Do we want to deprecate this feature entirely?

This is a good question, possibly worth raising in a separate issue for discussion. I think they are a convenient shortcut, but they shouldn't be the primary interface.

> Do we just want to wrap the datetime component virtual variables to the .dt accessor if they're datetime-like?

Yes, certainly. We definitely don't want to have two separate implementations of datetime components in the code base.

xarray/core/accessors.py
```diff
@@ -0,0 +1,114 @@
...
+            raise TypeError("'dt' accessor only available for "
+                            "DataArray with datetime64 or timedelta64 dtype")
+        self._obj = xarray_obj
+        self._dt = self._obj.data
+
+    def _tslib_field_accessor(name, docstring=None):
+        def f(self):
+            from .dataarray import DataArray
+            result = _get_date_field(self._dt, name)
+            return DataArray(result, name=name,
```
shoyer: can also just do `type(self._obj)` here instead of `DataArray`, and the result will probably work better on subclasses

xarray/core/accessors.py
```diff
@@ -0,0 +1,114 @@
...
+        if not is_datetime_like(xarray_obj.dtype):
+            raise TypeError("'dt' accessor only available for "
+                            "DataArray with datetime64 or timedelta64 dtype")
+        self._obj = xarray_obj
+        self._dt = self._obj.data
+
+    def _tslib_field_accessor(name, docstring=None):
+        def f(self):
+            from .dataarray import DataArray
+            result = _get_date_field(self._dt, name)
```
shoyer: I'm not really sure why you need `self._dt`. Why not just use `self._obj.data` here?

xarray/core/accessors.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Parameters
+    ----------
+    values : np.ndarray or dask.array-like
+        Array-like container of datetime-like values
+    name : str
+        Name of datetime field to access
+
+    """
+    def _access_through_series(values, name):
```
shoyer: For serialization in dask, it's better to pass top-level functions to `map_blocks`. So I'd just move this to the top level.

xarray/core/accessors.py
```diff
@@ -0,0 +1,114 @@
...
+    """Indirectly access pandas' libts.get_date_field by wrapping data
+    as a Series and calling through `.dt` attribute.
+
+    Parameters
+    ----------
+    values : np.ndarray or dask.array-like
+        Array-like container of datetime-like values
+    name : str
+        Name of datetime field to access
+
```
shoyer: Add `Returns`

doc/time-series.rst
```diff
@@ -100,7 +115,8 @@ __ http://pandas.pydata.org/pandas-docs/stable/api.html#time-date-components
     ds['time.month']
     ds['time.dayofyear']
 
-xarray adds ``'season'`` to the list of datetime components supported by pandas:
+For use as a derived coordinate, xarray adds ``'season'`` to the list of
```
shoyer: It would be nice to move the logic for `season` and the other datetime virtual variables to use `.dt`.

xarray/core/accessors.py
```diff
@@ -0,0 +1,114 @@
...
+    if isinstance(values, dask_array_type):
+        from dask.array import map_blocks
+        out_dtype = object if name == "weekday_name" else np.int64
+        return map_blocks(_access_through_series,
+                          values, name, dtype=out_dtype)
+    else:
+        return _access_through_series(values, name)
+
+
+@register_dataarray_accessor('dt')
```
shoyer: Another thought: since this isn't being defined outside of xarray, it might actually make sense to import this class from `datarray.py` and simply add it to the DataArray class directly:
```
class DataArray:
    ...
    dt = property(DatatimeAccessor)
```

That has the advantage of being more transparent about whether it comes from.

darothen:
There's a test-case relating to #367 (**test_virtual_variable_same_name**) which is causing me a bit of grief as I re-factor the virtual variable logic. Should we really be able to access variables like `ds['time.time']`? This seems to break the logic of what a virtual variable *does*, and was implemented to help out with time GroupBys and resampling (something I'll eventually get around to finishing up a refactor for - #1272). 

Two options for fixing:

1. add a "time" field to `DateTimeAccessor`.
2. add some additional if-then-else logic to `_get_virtual_variable`.

shoyer:
> add a "time" field to DateTimeAccessor.

This would be my preferred fix.

I agree that `ds['time.time']` is a little confusing but think it's unambiguous: it says get the "time" component (an array with `datetime.time` elements) from the "time" variable (an array of dtype=datetime64), i.e., equivalent to `df['time'].dt.time` in pandas.

darothen:
Turns out it was easy enough to add an accessor for `ds['time.time']`; that's already provided via pandas.

xarray/core/dataset.py
```diff
@@ -74,20 +76,11 @@ def _get_virtual_variable(variables, key, level_vars=None, dim_sizes=None):
...
-            # TODO: move 'season' into pandas itself
-            seasons = np.array(['DJF', 'MAM', 'JJA', 'SON'])
-            month = date.month
-            data = seasons[(month // 3) % 4]
+        if is_datetime_like(ref_var.dtype):
+            ref_var = xr.DataArray(ref_var)
+            data = getattr(ref_var.dt, var_name)
         else:
-            data = getattr(date, var_name)
+            data = getattr(ref_var)
```
shoyer: This is broken -- I think this clause should be removed instead.

xarray/core/dataset.py
```diff
@@ -74,20 +76,11 @@ def _get_virtual_variable(variables, key, level_vars=None, dim_sizes=None):
...
-            raise KeyError(key)
-
-        if var_name == 'season':
-            # TODO: move 'season' into pandas itself
-            seasons = np.array(['DJF', 'MAM', 'JJA', 'SON'])
-            month = date.month
-            data = seasons[(month // 3) % 4]
+        if is_datetime_like(ref_var.dtype):
+            ref_var = xr.DataArray(ref_var)
+            data = getattr(ref_var.dt, var_name)
```
shoyer: probably should use `getattr(ref_var.dt, var_name).data` to avoid passing a DataArray instance into `xarray.Variable`.

xarray/core/accessors.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Returns
+    -------
+    datetime_fields : same type as values
+        Array-like of datetime fields accessed for each element in values
+
+    """
+    if isinstance(values, dask_array_type):
+        from dask.array import map_blocks
+        out_dtype = object if name == "weekday_name" else np.int64
```
shoyer: Should also be `object` for season and time.

Actually, rather than put the logic here, can you add a `dtype` argument to `_tslib_field_accessor` and pass it on?

xarray/core/dataset.py
darothen: I think it's a typo; I intended the whole if-then-else block to read:

``` python
        if is_datetime_like(ref_var.dtype):
            ref_var = xr.DataArray(ref_var)
            data = getattr(ref_var.dt, var_name).data
        else:
            data = getattr(ref_var, var_name).data
``` 

We only access datetime virtual fields for now, but this leaves the possibility of other types of virtual field access at some point in the future.

xarray/core/accessors.py
darothen: Yes - went ahead and did that. `dtype` now passes through the whole chain of functions, but is only used if we end up calling `map_blocks` on a dask array.

darothen:
Alrighty, patched and ready for a final look-over! I appreciate the help and patience, @shoyer!

shoyer:
Thanks @darothen! This is really nice

PR #1358, opened 2017-04-07
Include missing conftest.py

ghisvail:
The tests cannot be run from the release tarball without this file.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include LICENSE
+include LICENSE conftest.py
 recursive-include licenses *
 recursive-include doc *
 prune doc/_build


ghisvail:
Unfortunately, even if I patch the code base to add the missing file, the tests cannot be run due to the following bug: https://github.com/pytest-dev/pytest/issues/1889

Too bad you guys decided to change the design of the test suite for a point release :-(

shoyer:
This is embarrassing -- thanks for the catch.

`conftest.py` seems to be much more troublesome than the pytest documentation suggests. Another alternative is to make the test-suite robust to its absence, and not bother to packaging it, given that it is mostly useful for development purposes.

ghisvail:
> Another alternative is to make the test-suite robust to its absence, and not bother to packaging it, given that it is mostly useful for development purposes.

If by *packaging* you mean *distributing* here, then I am very disappointed by your statement. From a package maintainer's perspective, I would consider the lack of a test suite a severe regression, should the latter be removed from source distribution in the future.

I would also remind you that it is that very same test suite that lead to the triage of the regressions with `bottleneck` whilst I was bootstrapping the Debian packaging effort.

Please consider fixing the current test suite or reverting to the old design. The fact that `conftest.py` does not work reliably (see the bug linked above) should be a sign that the decision to jump to this design happened too soon.

shoyer:
> If by packaging you mean distributing here, then I am very disappointed by your statement. From a package maintainer's perspective, I would consider the lack of a test suite a severe regression, should the latter be removed from source distribution in the future.

I agree. I meant not distributing `conftests.py` in the pypi package (and making the test suite run without it), not removing the test suite.

ghisvail:
There is a special note for [pytest_addoption](https://docs.pytest.org/en/latest/_modules/_pytest/hookspec.html#pytest_addoption) which might explain the problem.

ghisvail:
I can get the tests to work at build time, but the CI tests using --pyargs (to test against the installed library under site-packages) no longer do, i.e. I get a bunch of `ValueError: no option named '--run-flaky'`

ghisvail:
> I can get the tests to work at build time

By adding conftest.py manually if that was not clear.

ghisvail:
And it looks like most projects in the Debian [archive](https://codesearch.debian.net/search?q=pytest_addoption&perpkg=1) using `pytest_addoption` have it declared either as a plugin or inside the `tests` folder, as the note in the pytest documentation says.

PR #1361, opened 2017-04-08
Fix .where(drop=True) when arguments do not have indexes

### LINKED ISSUE #1350: where(..., drop=True) error
These results appear to be incorrect unless I'm missing something:
```python
In [1]: import xarray as xr

In [2]: import numpy as np

In [3]: array = xr.DataArray(np.zeros((1,2,3)), dims=['time','x','y'], coords={'x':np.arange(2)})

In [4]: array[0,1,1] = 1

In [5]: array.where(array !=0, drop=True)
Out[5]: 
<xarray.DataArray (time: 1, x: 1, y: 1)>
array([[[ 0.]]])
Coordinates:
  * x        (x) int64 1
Dimensions without coordinates: time, y

In [5]: array.where(array !=0, drop=True).values
Out[5]: array([[[ 0.]]])

In [7]: array.values[array.values !=0]
Out[7]: array([ 1.])
```

shoyer:
There might be cleaner ways to fix this than aligning twice, but this will do for now.

 - [x] closes #1350
 - [x] tests added / passed
 - [x] whatsnew entry

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -610,21 +610,25 @@ class BaseDataObject(AttrAccessMixin):
         if drop:
             from .dataarray import DataArray
             from .dataset import Dataset
+            from .alignment import align
+
+            if not isinstance(cond, (Dataset, DataArray)):
+                raise TypeError("Cond argument is %r but must be a %r or %r" %
+                                (cond, Dataset, DataArray))
+            # align so we can use integer indexing
+            self, cond = align(self, cond)
+
             # get cond with the minimal size needed for the Dataset
             if isinstance(cond, Dataset):
                 clipcond = cond.to_array().any('variable')
-            elif isinstance(cond, DataArray):
-                clipcond = cond
             else:
-                raise TypeError("Cond argument is %r but must be a %r or %r" %
-                                (cond, Dataset, DataArray))
+                clipcond = cond
 
             # clip the data corresponding to coordinate dims that are not used
-            clip = dict(zip(clipcond.dims, [np.unique(adim)
-                                            for adim in np.nonzero(clipcond.values)]))
-            outcond = cond.isel(**clip)
-            indexers = {dim: outcond.get_index(dim) for dim in outcond.dims}
-            outobj = self.sel(**indexers)
+            nonzeros = zip(clipcond.dims, np.nonzero(clipcond.values))
+            indexers = {k: np.unique(v) for k, v in nonzeros}
+            outobj = self.isel(**indexers)
+            outcond = cond.isel(**indexers)
         else:
             outobj = self
             outcond = cond
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -2668,6 +2668,12 @@ class TestDataset(TestCase):
         expected = DataArray(np.zeros((0, 10)), dims=['nCells', 'nVertLevels'])
         self.assertDatasetIdentical(expected, actual)
 
+    def test_where_drop_no_indexes(self):
+        ds = Dataset({'foo': ('x', [0.0, 1.0])})
+        expected = Dataset({'foo': ('x', [1.0])})
+        actual = ds.where(ds == 1, drop=True)
+        self.assertDatasetIdentical(expected, actual)
+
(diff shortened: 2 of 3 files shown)


shoyer:
@pwolfram could you take a look here?

xarray/core/common.py
```diff
@@ -610,21 +610,25 @@ def where(self, cond, other=None, drop=False):
         if drop:
             from .dataarray import DataArray
             from .dataset import Dataset
+            from .alignment import align
+
+            if not isinstance(cond, (Dataset, DataArray)):
+                raise TypeError("Cond argument is %r but must be a %r or %r" %
+                                (cond, Dataset, DataArray))
```
pwolfram: This *is* cleaner and nicer now.

xarray/tests/test_dataset.py
```diff
@@ -2719,6 +2719,12 @@ def test_where_drop_empty(self):
         expected = DataArray(np.zeros((0, 10)), dims=['nCells', 'nVertLevels'])
         self.assertDatasetIdentical(expected, actual)
 
+    def test_where_drop_no_indexes(self):
```
pwolfram: We didn't have this test before because we required indices previously, correct @shoyer?

xarray/core/common.py
```diff
@@ -610,21 +610,25 @@ def where(self, cond, other=None, drop=False):
...
             # clip the data corresponding to coordinate dims that are not used
-            clip = dict(zip(clipcond.dims, [np.unique(adim)
-                                            for adim in np.nonzero(clipcond.values)]))
-            outcond = cond.isel(**clip)
-            indexers = {dim: outcond.get_index(dim) for dim in outcond.dims}
-            outobj = self.sel(**indexers)
+            nonzeros = zip(clipcond.dims, np.nonzero(clipcond.values))
+            indexers = {k: np.unique(v) for k, v in nonzeros}
+            outobj = self.isel(**indexers)
+            outcond = cond.isel(**indexers)
```
pwolfram: This is a lot cleaner now.

xarray/core/common.py
```diff
@@ -610,21 +610,25 @@ def where(self, cond, other=None, drop=False):
...
             else:
-                raise TypeError("Cond argument is %r but must be a %r or %r" %
-                                (cond, Dataset, DataArray))
+                clipcond = cond
 
             # clip the data corresponding to coordinate dims that are not used
-            clip = dict(zip(clipcond.dims, [np.unique(adim)
-                                            for adim in np.nonzero(clipcond.values)]))
-            outcond = cond.isel(**clip)
-            indexers = {dim: outcond.get_index(dim) for dim in outcond.dims}
```
pwolfram: This was the offending line, correct?  Effectively if there wasn't an index this would fail, correct?

pwolfram:
@shoyer, thanks for the fix here-- sorry this fell between the cracks.

xarray/tests/test_dataset.py
```diff
@@ -2719,6 +2719,12 @@ def test_where_drop_empty(self):
...
+        ds = Dataset({'foo': ('x', [0.0, 1.0])})
+        expected = Dataset({'foo': ('x', [1.0])})
+        actual = ds.where(ds == 1, drop=True)
+        self.assertDatasetIdentical(expected, actual)
```
MaximilianR: No need to change now, but let's gradually move to using the `assert_identical` test functions

xarray/core/common.py
shoyer: It worked, but it filled in a default index based on the length of the dimension -- which isn't doing the right thing if we clipped values.

xarray/tests/test_dataset.py
shoyer: Yes. I added some additional test coverage when we made indices optional but clearly not enough here.

xarray/tests/test_dataset.py
shoyer: +1, might be worth a large search/replace PR in the full test suite. Right now I'm pretty sure `assertDatasetIdentical` is just an alias to `assert_identical`.

Issue #1353, opened 2017-04-05
Cannot import xarray.tests due to use of pytest.config


jjhelmus:
xarray.tests cannot be imported due to the use of pytest.config:

```
$ python -c "import xarray; print(xarray.__version__); import xarray.tests"
0.9.2-3-g9479038
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/jhelmus/devel/xarray/xarray/tests/__init__.py", line 118, in <module>
    not pytest.config.getoption("--run-flaky"),
AttributeError: module 'pytest' has no attribute 'config'
```

The use of pytest.config in the `xarray/tests/__init__.py` seem to have been introduced in #1336.  From the discussion in pytest-dev/pytest#1688 it seems as if the use of `pytest.config` is discouraged.

jjhelmus:
pytest-dev/pytest#472 from [BitBucket #472](https://bitbucket.org/pytest-dev/pytest/issues/472) seems to touch on this topic and references the note found at the end of the [skip and xfail documentation](https://docs.pytest.org/en/latest/skipping.html):

> You cannot use pytest.config.getvalue() in code imported before pytest’s argument parsing takes place. 
> For example, conftest.py files are imported before command line parsing and thus config.getvalue() will not execute correctly.

shoyer:
Hmm. This seems like an easy way to add a few options to running our test suite, but maybe there is a better way to do this? I guess we could hack in a fallback with `hasattr()` but that's pretty ugly.

Just out of curiosity, what reason did you have for importing `xarray.tests`?

jjhelmus:
> Just out of curiosity, what reason did you have for importing xarray.tests?

`xarray.tests` is listed as a import test in the [conda-forge xarray-feedstock](https://github.com/conda-forge/xarray-feedstock) and was failing when I was doing a build.  I do not think that import line needs to be present, `tests` does not seem to be in the public API of xarray.  I'd be in favor of removing that line and keeping the test suite setup as-is.

shoyer:
@jjhelmus Agreed! For testing public API, being able to import `xarray` and `xarray.backends` should be enough.

shoyer:
@jjhelmus see conda-forge/xarray-feedstock#20

jjhelmus:
@shoyer Agreed, there is no need for `xarray.tests` to be importable.  Closing this issue, thanks for the quick clarification.

PR #1365, opened 2017-04-10
Fix missing config errors when running test suite

shoyer:
Fixes #1358, #1353

The approach I took here is to *not* package `conftest.py` with the xarray distribution, but rather make the test suite robust to it's absence.

Also, as suggested by @QuLogic in #1336, switched from using a custom decorator for `--run-slow` to just using `@pytest.mark.slow`.

Finally, I speed-up the test_dataset.test_rolling_reduce tests (from ~55 to ~17 seconds when bottleneck is not installed) and marked them as slow. It would be nice to get them running in a few seconds or less, but we can save that for later. On my laptop, the whole test suite now runs with `-k 'not slow'` in 36 seconds, which is not great, but acceptable.

CC @pwolfram @jhamman @ghisvail

 - [x] tests added / passed
 - [x] whatsnew entry


xarray/tests/__init__.py
```diff
@@ -78,49 +78,36 @@
...
-def requires_bottleneck(test):
-    return test if has_bottleneck else pytest.mark.skip('requires bottleneck')(test)
-
+try:
+    _SKIP_FLAKY = not pytest.config.getoption("--run-flaky")
+except ValueError:
+    # Can't get config from pytest, e.g., because xarray is installed instead
+    # of being run from a development version (and hence conftests.py is not
+    # available). Don't run flaky tests.
+    _SKIP_FLAKY = True
```
pwolfram: Nice solution!

PR #1367, opened 2017-04-10
Fix open_dataarray does not pass properly its parameters to open_dataset

### LINKED ISSUE #1359: Bug: open_dataarray does not pass properly its parameters to open_dataset
Hi,

There's now a `autoclose` parameter to the `open_dataset` function.
But `open_dataarray` uses positional arguments only, so now the arguments are no longer at their place, and `engine`, for instance, can't be specified.

Thanks !

shoyer:
- [x] closes #1359
 - [x] tests added / passed
 - [x] whatsnew entry

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -29,6 +29,10 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


pwolfram:
@shoyer, thanks for fixing this bug that I missed.

Issue #1386, opened 2017-04-27
groupby_bins tests failing with pandas 0.20


shoyer:
These are due to the introduction of `Interval` and `IntervalIndex` to hold the result of `pd.cut`:
http://pandas-docs.github.io/pandas-docs-travis/whatsnew.html#whatsnew-0200-enhancements-intervalindex

From the latest test of pandas 0.20 (currently in release candidate) on master:
https://travis-ci.org/pydata/xarray/jobs/224867239
```
=================================== FAILURES ===================================
_______________________ TestDataArray.test_groupby_bins ________________________
self = <xarray.tests.test_dataarray.TestDataArray testMethod=test_groupby_bins>
    def test_groupby_bins(self):
        array = DataArray(np.arange(4), dims='dim_0')
        # the first value should not be part of any group ("right" binning)
        array[0] = 99
        # bins follow conventions for pandas.cut
        # http://pandas.pydata.org/pandas-docs/stable/generated/pandas.cut.html
        bins = [0, 1.5, 5]
        bin_coords = ['(0, 1.5]', '(1.5, 5]']
        expected = DataArray([1, 5], dims='dim_0_bins',
                             coords={'dim_0_bins': bin_coords})
        # the problem with this is that it overwrites the dimensions of array!
        # actual = array.groupby('dim_0', bins=bins).sum()
        actual = array.groupby_bins('dim_0', bins).apply(lambda x: x.sum())
>       self.assertDataArrayIdentical(expected, actual)
xarray/tests/test_dataarray.py:1752: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
xarray/tests/__init__.py:165: in assertDataArrayIdentical
    assert_identical(ar1, ar2)
xarray/testing.py:83: in assert_identical
    assert_identical(a._to_temp_dataset(), b._to_temp_dataset())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
a = <xarray.Dataset>
Dimensions:       (dim_0_bins: 2)
Coordinates:
  * dim_0_bins    (dim_0_bins) <U8 '(0, 1.5]' '(1.5, 5]'
Data variables:
    <this-array>  (dim_0_bins) int64 1 5
[...]

Issue #974, opened 2016-08-18
Indexing with alignment and broadcasting


shoyer:
I think we can bring all of NumPy's advanced indexing to xarray in a very consistent way, with only very minor breaks in backwards compatibility.

For _boolean indexing_:
- `da[key]` where `key` is a boolean labelled array (with _any_ number of dimensions) is made equivalent to `da.where(key.reindex_like(ds), drop=True)`. This matches the existing behavior if `key` is a 1D boolean array. For multi-dimensional arrays, even though the result is now multi-dimensional, this coupled with automatic skipping of NaNs means that `da[key].mean()` gives the same result as in NumPy.
- `da[key] = value` where `key` is a boolean labelled array can be made equivalent to `da = da.where(*align(key.reindex_like(da), value.reindex_like(da)))` (that is, the three argument form of `where`).
- `da[key_0, ..., key_n]` where all of `key_i` are boolean arrays gets handled in the usual way. It is an `IndexingError` to supply multiple labelled keys if any of them are not already aligned with as the corresponding index coordinates (and share the same dimension name). If they want alignment, we suggest users simply write `da[key_0 & ... & key_n]`.

For _vectorized indexing_ (by integer or index value):
- `da[key_0, ..., key_n]` where all of `key_i` are integer labelled arrays with any number of dimensions gets handled like NumPy, except instead of broadcasting numpy-style we do broadcasting xarray-style:
  - If any of `key_i` are unlabelled, 1D arrays (e.g., numpy arrays), we convert them into an `xarray.Variable` along the respective dimension. 0D arrays remain scalars. This ensures that the result of broadcasting them (in the next step) will be consistent with our current "outer indexing" behavior. Unlabelled higher dimensional arrays triggers an `IndexingError`.
  - We ensure all keys have the same dimensions/coordinates by mapping it to `da[*broadcast(key_0, ..., key_n)]` (note that broadcast now includes automatic alignment). 
  - The result's dimensions and coordinates are copied from the broadcast keys.
  - The result's values are taken by mapping each set of integer locations specified by the broadcast version of `key_i` to the integer position on the corresponding `i`th axis on `da`.
- Labeled indexing like `ds.loc[key_0, ...., key_n]` works exactly as above, except instead of doing integer lookup, we lookup label values in the corresponding index instead.
- Indexing with `.isel` and `.sel`/`.reindex` works like the two previous cases, except we lookup axes by dimension name instead of axis position.
[...]

PR #1389, opened 2017-04-29
Sortby

### LINKED ISSUE #967: sortby() or sort_index() method for Dataset and DataArray
They should function like the pandas methods of the same name.

Under the covers, I believe it would suffice to simply remap `ds.sort_index('time')` -> `ds.isel(time=ds.indexes['time'].argsort())`.

chunweiyuan:
- [x] closes #967
 - [x] tests added / passed
 - [x] whatsnew entry

### DIFF
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -523,3 +523,24 @@ labels:
 
     array
     array.get_index('x')
+
+Sort
+------------------
+
+One may sort a dataarray/dataset via :py:meth:`~xarray.DataArray.sortby` and
+:py:meth:`~xarray.DataArray.sortby`.  The input could either be existing
+dimensions, or 1-D dataarrays that share dimensions (and have correct dimension
+lengths) as the calling object.
+
+.. ipython:: python
+
+    ds = Dataset({'A': DataArray([[1, 2], [3, 4]],
+                                 [('x', ['b', 'a']),
+                                  ('y', [1, 0])]),
+                  'B': DataArray([[5, 6], [7, 8]], dims=['x', 'y'])})
+    ds.sortby('x')
+    ds.sortby(['y', 'x'])
+    ds.sortby(['y', 'x'], ascending=False)
+    dax = DataArray([100, 99], [('x', [100, 99])])
+    day = DataArray([90, 80], [('y', [90, 80])])
+    actual = ds.sortby([day, dax])
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -541,6 +541,6 @@ lengths) as the calling object.
     ds.sortby('x')
     ds.sortby(['y', 'x'])
     ds.sortby(['y', 'x'], ascending=False)
-    dax = DataArray([100, 99], [('x', [100, 99])])
-    day = DataArray([90, 80], [('y', [90, 80])])
+    dax = DataArray([100, 99], [('x', [0, 1])])
+    day = DataArray([90, 80], [('y', [0, 1])])
     actual = ds.sortby([day, dax])
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -523,24 +523,3 @@ labels:
 
     array
     array.get_index('x')
-
-Sort
-------------------
-
-One may sort a dataarray/dataset via :py:meth:`~xarray.DataArray.sortby` and
-:py:meth:`~xarray.DataArray.sortby`.  The input could either be existing
-dimensions, or 1-D dataarrays that share dimensions (and have correct dimension
-lengths) as the calling object.
-
-.. ipython:: python
-
-    ds = Dataset({'A': DataArray([[1, 2], [3, 4]],
-                                 [('x', ['b', 'a']),
-                                  ('y', [1, 0])]),
-                  'B': DataArray([[5, 6], [7, 8]], dims=['x', 'y'])})
-    ds.sortby('x')
-    ds.sortby(['y', 'x'])
-    ds.sortby(['y', 'x'], ascending=False)
-    dax = DataArray([100, 99], [('x', [0, 1])])
-    day = DataArray([90, 80], [('y', [0, 1])])
-    actual = ds.sortby([day, dax])
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2743,6 +2743,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
 
     def sortby(self, variables, ascending=True):
         """
+        Sort object by labels or values (along an axis).
+
         Sorts the dataset, either along specified dimensions,
         or according to values of 1-D dataarrays that share dimension
         with calling object.  If multiple sorts along the same dimension is
@@ -2750,7 +2752,6 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         https://docs.scipy.org/doc/numpy/reference/generated/numpy.lexsort.html
         and the FIRST key in the sequence is used as the primary sort key,
         followed by the 2nd key, etc.
-
 
         Parameters
         ----------
@@ -2769,25 +2770,26 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         from .dataarray import DataArray
 
         if not isinstance(variables, list):
-            vs = [variables]
+            variables = [variables]
         else:
-            vs = variables
-        vs = [v if isinstance(v, DataArray) else self[v] for v in vs]
+            variables = variables
+        variables = [v if isinstance(v, DataArray)
+                     else self[v] for v in variables]
 
         vars_by_dim = defaultdict(list)
-        for d in vs:
-            if len(d.dims) > 1:
+        for data_array in variables:
+            if len(data_array.dims) > 1:
                 raise ValueError("Input DataArray has more than 1 dimension.")
             else:
-                key = d.dims[0]
-                if len(d) != self.dims[key]:
+                key = data_array.dims[0]
+                if len(data_array) != self.dims[key]:
                     raise ValueError("Input DataArray must have same "
                                      "length as dimension it is sorting.")
-            vars_by_dim[key].append(d)
+            vars_by_dim[key].append(data_array)
 
         indices = {}
-        for key, ds in vars_by_dim.items():
-            order = np.lexsort(tuple(reversed(ds)))
+        for key, arrays in vars_by_dim.items():
+            order = np.lexsort(tuple(reversed(arrays)))
             indices[key] = order if ascending else order[::-1]
         return self.isel(**indices)
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2748,6 +2748,39 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
 
         return self._replace_vars_and_dims(variables)
 
+    def sort_index(self, dims, ascending=True, inplace=False):
+        """
+        The xarray-equivalent of pandas.DataFrame.sort_index.
+        Given dimension(s), sorts dataset based on dimensional labels.
+
+        Parameters
+        ----------
+        dims: str or iterable of str.
+            Dimension(s) over which to sort by dimensional labels.
+        ascending: boolean, optional.
+            whether to sort by ascending order.
+        inplace: boolean, optional.
+            Whether to return a new object.
+
+        Returns
+        -------
+        sorted: Dataset
+            A new dataset where all the specified dims are sorted by dim
+            labels.
+        """
+        if isinstance(dims, (str or unicode)):
+            dimensions = [dims]
+        else:
+            dimensions = dims
+        if inplace:
+            self = self.isel({d: self.indexes[d].argsort()
+                              if ascending else self.indexes[d].argsort()[::-1]
+                              for d in dimensions})
+        else:
+            return self.isel({d: self.indexes[d].argsort()
+                              if ascending else self.indexes[d].argsort()[::-1]
+                              for d in dimensions})
+
     def quantile(self, q, dim=None, interpolation='linear',
                  numeric_only=False, keep_attrs=False):
         """Compute the qth quantile of the data along the specified dimension.
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2751,6 +2751,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
 
         If the input variables are dataarrays, then the dataarrays are aligned
         (via left-join) to the calling object prior to sorting by cell values.
+        NaNs are sorted to the end, following Numpy convention.
 
         If multiple sorts along the same dimension is
         given, numpy's lexsort is performed along that dimension:
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2781,20 +2781,22 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
             vs = variables
         vs = [v if isinstance(v, DataArray) else self[v] for v in vs]
 
-        dims = {}
+        vars_by_dim = defaultdict(list)
         for d in vs:
             if len(d.dims) > 1:
                 raise ValueError("Input DataArray has more than 1 dimension.")
             else:
                 key = d.dims[0]
-                val = d.argsort() if ascending else d.argsort()[::-1]
-                if len(val) != len(self[key]):
+                if len(d) != self.dims[key]:
                     raise ValueError("Input DataArray must have same "
                                      "length as dimension it is sorting.")
-            if key in dims:
-                dims[key] = np.lexsort((val, dims[key]))
-            dims.update({key: val})
-        return self.isel(**dims)
+            vars_by_dim[key].append(d)
+
+        indices = {}
+        for key, ds in vars_by_dim.items():
+            order = np.lexsort(tuple(reversed(ds)))
+            indices[key] = order if ascending else order[::-1]
+        return self.isel(**indices)
 
     def quantile(self, q, dim=None, interpolation='linear',
                  numeric_only=False, keep_attrs=False):
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import functools
-from collections import Mapping, defaultdict
+from collections import Mapping, defaultdict, Iterable
 from numbers import Number
 
 import sys
@@ -2748,19 +2748,19 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
 
         return self._replace_vars_and_dims(variables)
 
-    def sort_index(self, dims, ascending=True, inplace=False):
+    def sortby(self, variables, ascending=True):
         """
-        The xarray-equivalent of pandas.DataFrame.sort_index.
-        Given dimension(s), sorts dataset based on dimensional labels.
+        Sorts the dataset, either along specified dimensions,
+        or according to values of 1-D dataarrays that share dimension
+        with calling object.
 
         Parameters
         ----------
-        dims: str or iterable of str.
-            Dimension(s) over which to sort by dimensional labels.
+        variables: (str, DataArray, or iterable of either)
+            Name of a 1D variable in coords/data_vars whose values are used to
+            to sort the dataset.
         ascending: boolean, optional.
             whether to sort by ascending order.
-        inplace: boolean, optional.
-            Whether to return a new object.
 
         Returns
         -------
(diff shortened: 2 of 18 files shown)


xarray/core/dataset.py
```diff
@@ -2748,6 +2748,64 @@ def roll(self, **shifts):
...
+            whether to sort by ascending order.
+
+        Returns
+        -------
+        sorted: Dataset
+            A new dataset where all the specified dims are sorted by dim
+            labels.
+        """
+        from .dataarray import DataArray
+        if isinstance(variables, (str, unicode, DataArray)):
```
shoyer: To reduce the number of separate code paths, I would suggest a normalization based flow. For example:
```python
if not isinstance(variables, list):
    variables = [variables]
# variables is now a list of str/DataArray
variables = [v if isinstance(v, DataArray) else self[v]
             for v in variables]
# variables is now a list of DataArray
```
Now you can avoid the `if`/`else` branch inside the loop.

xarray/core/dataset.py
```diff
@@ -2748,6 +2748,64 @@ def roll(self, **shifts):
...
+    def sortby(self, variables, ascending=True):
+        """
+        Sorts the dataset, either along specified dimensions,
+        or according to values of 1-D dataarrays that share dimension
+        with calling object.
+
+        Parameters
+        ----------
+        variables: (str, DataArray, or iterable of either)
+            Name of a 1D variable in coords/data_vars whose values are used to
```
shoyer: Should be `Name(s)`

xarray/core/dataset.py
```diff
@@ -2748,6 +2748,64 @@ def roll(self, **shifts):
...
 
+    def sortby(self, variables, ascending=True):
+        """
+        Sorts the dataset, either along specified dimensions,
+        or according to values of 1-D dataarrays that share dimension
+        with calling object.
+
+        Parameters
+        ----------
+        variables: (str, DataArray, or iterable of either)
```
shoyer: It might make sense to limit this to only lists, rather than arbitrary iterables. Arbitrary iterables makes the dispatch rules more complicated because both strings and DataArray objects are iterable.

xarray/core/dataset.py
```diff
@@ -2748,6 +2748,64 @@ def roll(self, **shifts):
...
+                    key = d.dims[0]
+                    if key not in self:
+                        raise KeyError("Dimension {} does not exist "
+                                       "in object.".format(key))
+                    val = d.argsort() if ascending else d.argsort()[::-1]
+                    if len(val) != len(self[key]):
+                        raise ValueError("Input DataArray must have same "
+                                         "length as dimension it is sorting.")
+            dims.append((key, val))
+        return self.isel(**dict(dims))
```
shoyer: What happens if there are multiple keys provided along the same dimension? This is equivalent to sorting by multiple columns in a spreadsheet.

It's OK not to support it, but we should be sure to raise an informative error, because I'm sure somebody will try it. Note that when you put multiple items in `dict` with the same name Python *ignores* the first argument:
```
In [1]: dict([('x', 1), ('x', 2)])
Out[1]: {'x': 2}
```

If you *do* want to support it, probably the easiest way is with [`np.lexsort`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.lexsort.html)

xarray/tests/test_dataarray.py
```diff
@@ -2518,6 +2518,40 @@ def test_combine_first(self):
...
         self.assertDataArrayEqual(actual, expected)
 
+    def test_sortby(self):
+        da = DataArray([[1, 2], [3, 4]],
+                       [('x', ['b', 'a']), ('y', [1, 0])])
+
+        expected = DataArray([[4, 3], [2, 1]],
+                             [('x', ['a', 'b']), ('y', [0, 1])])
+
+        actual = da.sortby(['x', 'y'])
```
shoyer: It would be good to include a test for a single sorted, too, just `da.sortby('x')`. Likewise `da.sortby(dax)`.

doc/indexing.rst
```diff
@@ -523,3 +523,24 @@ labels:
 
     array
     array.get_index('x')
+
+Sort
+------------------
```
shoyer: I think the length here needs to line up with the previous line.

doc/indexing.rst
```diff
@@ -523,3 +523,24 @@ labels:
 
     array
     array.get_index('x')
+
+Sort
```
shoyer: `indexing.rst` is already our longest doc page, so I would put this section in "Reshaping and reorganizing data" instead (`reshaping.rst`).

doc/indexing.rst
```diff
@@ -523,3 +523,24 @@ labels:
...
+------------------
+
+One may sort a dataarray/dataset via :py:meth:`~xarray.DataArray.sortby` and
```
MaximilianR: `One may` -> `You can`

MaximilianR:
We want this as `sortby`, rather than `sort_by`? Generally the latter would be more idiomatic, but if there is prior `sortby` out there, I would vote for consistency

chunweiyuan:
Weird.  Travis seems to fail in places unrelated to my changes.

I'm rather agnostic to `sortby` vs. `sort_by`.

shoyer:
Indeed, please ignore the first CI build on Travis. For some inexplicable reason it is installing the pandas 0.20rc, which is failing due to https://github.com/pydata/xarray/issues/1386.

I am slightly in favor of `sortby` but also could go either way. The PEP8 guidance on function/method naming is "lowercase with words separated by underscores as necessary to improve readability." I don't think an underscore is necessary for readability here.

xarray/core/dataset.py
```diff
@@ -2748,6 +2748,54 @@ def roll(self, **shifts):
...
+        vs = [v if isinstance(v, DataArray) else self[v] for v in vs]
+
+        dims = {}
+        for d in vs:
+            if len(d.dims) > 1:
+                raise ValueError("Input DataArray has more than 1 dimension.")
+            else:
+                key = d.dims[0]
+                val = d.argsort() if ascending else d.argsort()[::-1]
+                if len(val) != len(self[key]):
```
shoyer: Another minor point: it's better to fail earlier is possible. Can you move this check up one line, before the sort?

xarray/core/dataset.py
```diff
@@ -2748,6 +2748,54 @@ def roll(self, **shifts):
...
+        """
+        from .dataarray import DataArray
+
+        if not isinstance(variables, list):
+            vs = [variables]
+        else:
+            vs = variables
+        vs = [v if isinstance(v, DataArray) else self[v] for v in vs]
+
+        dims = {}
```
shoyer: Maybe this variable would be better called `indices`?

xarray/core/dataset.py
```diff
@@ -2748,6 +2748,54 @@ def roll(self, **shifts):
...
+            if len(d.dims) > 1:
+                raise ValueError("Input DataArray has more than 1 dimension.")
+            else:
+                key = d.dims[0]
+                val = d.argsort() if ascending else d.argsort()[::-1]
+                if len(val) != len(self[key]):
+                    raise ValueError("Input DataArray must have same "
+                                     "length as dimension it is sorting.")
+            if key in dims:
+                dims[key] = np.lexsort((val, dims[key]))
```
shoyer: I'm not sure this is calculating the right thing. I think you need to lexsort `d`, not `val` (the result of `argsort`). Otherwise, lexsort does some sort of shuffle of the sort indices:
```
>>> np.lexsort(([2, 1, 0],))
array([2, 1, 0])
```

Maybe [`inverse_permutation`](https://github.com/pydata/xarray/blob/8f6a68e3f821689203bce2bce52b412e9fe70b5c/xarray/core/nputils.py#L38) would fix this.

Alternatively, this might be clearer -- and maybe slightly faster -- if this was done with a single call to `lexsort` per dimension. Something like:
```
vars_by_dim = collections.defaultdict(list)
for d in vs:
    ...
    vars_by_dim[key].append(d)

indices = {}
for key, vars in vars_by_dim.items():
    order = np.lexsort(tuple(reversed(vars))
    indices[key] = order if ascending else order[::-1]
```

chunweiyuan: Good catch.  Not only should I have lexsorted `d`, there was another logical error later in that code block as well (`.update` should've been after an `else`).  Fixed those and pushed.

chunweiyuan:
Almost time to merge this baby in? :)

xarray/tests/test_dataarray.py
```diff
@@ -2518,6 +2518,57 @@ def test_combine_first(self):
...
+
+        dax = DataArray([100, 90, 80], [('x', [0, 1, 2])])
+        actual = da.sortby(dax)
+        self.assertDataArrayEqual(actual, expected)
+
+        # test 1-D lexsort
+        # dax0 is sorted first to give indices of [1, 2, 0]
+        # and then dax1 would be used to move index 2 ahead of 1
+        dax0 = DataArray([100, 95, 95], [('x', [0, 1, 2])])
+        dax1 = DataArray([0, 1, 0], [('x', [7, 8, 9])])
```
shoyer: This example actually raises a good question, because `dax1` has different coordinate labels for `x` than `dax0` and `da`.

Should we align sortby arguments before using them, or ensure that they are already aligned? I think this is probably a good idea.

xarray/core/dataset.py
```diff
@@ -2741,6 +2741,56 @@ def roll(self, **shifts):
 
         return self._replace_vars_and_dims(variables)
 
+    def sortby(self, variables, ascending=True):
+        """
+        Sorts the dataset, either along specified dimensions,
```
shoyer: Can you please add a short one-line description at the top? That is the standard numpy docstring format.

xarray/core/dataset.py
```diff
@@ -2741,6 +2741,56 @@ def roll(self, **shifts):
...
+        or according to values of 1-D dataarrays that share dimension
+        with calling object.  If multiple sorts along the same dimension is
+        given, numpy's lexsort is performed along that dimension:
+        https://docs.scipy.org/doc/numpy/reference/generated/numpy.lexsort.html
+        and the FIRST key in the sequence is used as the primary sort key,
+        followed by the 2nd key, etc.
+
+
```
shoyer: should have only one space.

xarray/core/dataset.py
```diff
@@ -2741,6 +2741,56 @@ def roll(self, **shifts):
...
+                raise ValueError("Input DataArray has more than 1 dimension.")
+            else:
+                key = d.dims[0]
+                if len(d) != self.dims[key]:
+                    raise ValueError("Input DataArray must have same "
+                                     "length as dimension it is sorting.")
+            vars_by_dim[key].append(d)
+
+        indices = {}
+        for key, ds in vars_by_dim.items():
```
shoyer: This is a minor point -- but can you use slightly longer variable names instead of the abbreviations for variables that exist outside of one line? e.g.,
`vs` -> `variables`
`d` -> `data_array`
`ds` -> arrays (this one is especially confusing, because `ds` is normally used for `xarray.Dataset`)

This would help for readability (also follows PEP8 guidelines)

xarray/core/dataset.py
```diff
@@ -2741,6 +2741,56 @@ def roll(self, **shifts):
...
+            order = np.lexsort(tuple(reversed(ds)))
```
shoyer: It would be nice to add a test case verifying that this sorts a `pandas.MultiIndex` properly.

xarray/tests/test_dataarray.py
chunweiyuan: hmm, I have a dumb question: why do they need to be aligned?  In this case, we're sorting `da` along `x` using the values of two dataarrays: `dax0` and `dax1`.  So as long as both of those dataarrays have `x` and the right length, we should be able to perform our sort.  The coordinate labels feel irrelevant.

shoyer: It would mostly be for consistency with other xarray options, almost all of which require aligned objects. Indexing is the one exception, but I want to clean that up (#974).

I think I would prefer to align all of objects in `variables` with the object being sorted, if they aren't already aligned (in most cases they will be, if they are pulled out by name). You could do this with `xarray.align(self, *variables, join='left')`. Just note that missing sort labels will get sorted to the end, consistent with how NumPy's sort handles NaN.

chunweiyuan: Sounds good.  Will do.

chunweiyuan: Quick question:  If we're aligning the input args, does that mean we allow the input dataarrays to be N-D as well, since the left-join will force matching of the dims.  Then of course it opens up the possibility of using 1 input dataarray to sort _multiple_ dimensions.  Should we support that?

Align also opens up a small can of worms, because the left-join could introduce `nan` into the array values, complicating the sort.

shoyer: `align` doesn't change the dimensions on any of its arguments -- it only changes coordinate labels and dimension sizes.

> Then of course it opens up the possibility of using 1 input dataarray to sort multiple dimensions. 

What would this look like?

> Align also opens up a small can of worms, because the left-join could introduce nan into the array values, complicating the sort.

Indeed, this is why we would need to document how we sort NaN. Fortunately, NumPy already has a well defined sort order for NaN (it gets moved to the end).

chunweiyuan: Upon more thinking I think the N-D sort using 1 dataarray wouldn't make much sense, and even if it does would be an edge case.  Never mind that.

chunweiyuan:
I've added a tiny bit of extra docstring just to ease my discomfort about the NaN sort.  If you feel it's redundant I can remove it.

chunweiyuan:
Perhaps it's finally ready for prime time?

shoyer:
Looks like one the Travis-CI tests failing on the sortby tests: https://travis-ci.org/pydata/xarray/jobs/229457729#L342

It's not immediately clear to me what's going on there, but maybe it's an issue with the old version of NumPy? The error is `TypeError: merge sort not available for item 0`.

xarray/core/dataset.py
```diff
@@ -2741,6 +2741,63 @@ def roll(self, **shifts):
...
+            A new dataset where all the specified dims are sorted by dim
+            labels.
+        """
+        from .dataarray import DataArray
+
+        if not isinstance(variables, list):
+            variables = [variables]
+        else:
+            variables = variables
+        variables = [v if isinstance(v, DataArray)
```
shoyer: nit: I prefer writing this like:
```
        variables = [v if isinstance(v, DataArray) else self[v]
                     for v in variables]
```
because the `else` clause is part of the `if` conditional statement.

xarray/tests/test_dataarray.py
```diff
@@ -2519,6 +2519,69 @@ def test_combine_first(self):
                              [('x', ['a', 'b', 'd']), ('y', [-1, 0])])
         self.assertDataArrayEqual(actual, expected)
 
+    def test_sortby(self):
```
shoyer: Can you delete the test cases here that are redundant with those in `test_dataset.py`? Given the implementation (only a small wrapped for `DataArray.sortby()`), I am OK having a less comprehensive set of unit tests for the DataArray implementation -- and fewer unit tests make things easier to maintain.

chunweiyuan: To be honest, I would say the entire `test_dataarray::test_sortby` is redundant, because it's essentially a carbon copy of `test_dataset::test_sortby`.  I've gotten rid of a few lines in the latest push, but could trim more if you like.

chunweiyuan:
If I downgrade to numpy 1.10, I get that error.  Once upgraded to 1.12, it goes away......
screen shot 2017-05-09 at 11 39 20 am
screen shot 2017-05-09 at 11 39 31 am

shoyer:
> If I downgrade to numpy 1.10, I get that error. Once upgraded to 1.12, it goes away......

Does it work on numpy 1.11?

We could potentially (partially) drop compatibility for older numpy releases. Or if we can identify the issue, we can raise an informative error or use a work-around.

chunweiyuan:
1.10.4 failed with the merge sort error.  1.11.0 failed with some error that had nothing to do with merge sort (` Reason: image not found.  Make sure your test modules/packages have valid Python names.`)  Starting from 1.11.1 it works fine.

What's the best course of action now?

chunweiyuan:
From numpy 1.11.0 release notes:
```
np.lexsort now supports arrays with object data-type

The function now internally calls the generic npy_amergesort when the type does not implement a merge-sort kind of argsort method.
```

shoyer:
Ah, good to know.

In that case, let's add a check for `dtype == object` based on `numpy.__version__` (search for `LooseVersion` for examples in the xarray codebase) and raise an informative `NotImplementedError` when necessary.

chunweiyuan:
Think I just copied something over from `variable.py` and changed the message a little bit...

xarray/core/dataset.py
```diff
@@ -2742,6 +2743,68 @@ def roll(self, **shifts):
...
+        ascending: boolean, optional.
+            whether to sort by ascending order.
+
+        Returns
+        -------
+        sorted: Dataset
+            A new dataset where all the specified dims are sorted by dim
+            labels.
+        """
+        if LooseVersion(np.__version__) < LooseVersion('1.11.0'):
```
shoyer: Can you put this check inside the loop, if `data_array.dtype == object`? That way, users of old versions of numpy can at least sort other arrays (e.g., numbers).

xarray/tests/test_dataarray.py
shoyer: I would just add a single very basic test to verify that it works, and add a comment noting that more advanced functionality is tested in `test_dataset.py`.

xarray/core/dataset.py
```diff
@@ -2742,6 +2743,69 @@ def roll(self, **shifts):
...
+        variables = [v if isinstance(v, DataArray) else self[v]
+                     for v in variables]
+        aligned_vars = align(self, *variables, join='left')
+        aligned_self = aligned_vars[0]
+        aligned_other_vars = aligned_vars[1:]
+        vars_by_dim = defaultdict(list)
+        for data_array in aligned_other_vars:
+            if len(data_array.dims) > 1:
+                raise ValueError("Input DataArray has more than 1 dimension.")
+            elif data_array.dtype == object and\
```
shoyer: Prefer parentheses to `\` for the line continuation:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

xarray/core/dataset.py
```diff
@@ -2742,6 +2743,69 @@ def roll(self, **shifts):
...
+        else:
+            variables = variables
+        variables = [v if isinstance(v, DataArray) else self[v]
+                     for v in variables]
+        aligned_vars = align(self, *variables, join='left')
+        aligned_self = aligned_vars[0]
+        aligned_other_vars = aligned_vars[1:]
+        vars_by_dim = defaultdict(list)
+        for data_array in aligned_other_vars:
+            if len(data_array.dims) > 1:
```
shoyer: Should be `data_array.ndim != 1`. Scalar arrays would also be problematic.

xarray/core/dataset.py
```diff
@@ -2742,6 +2743,69 @@ def roll(self, **shifts):
...
+                raise ValueError("Input DataArray has more than 1 dimension.")
+            elif data_array.dtype == object and\
+            LooseVersion(np.__version__) < LooseVersion('1.11.0'):
+                    raise NotImplementedError(
+                        'sortby uses np.lexsort under the hood, which '
+                        'requires numpy 1.11.0 or later to support '
+                        'object data-type.')
+            else:
+                key = data_array.dims[0]
```
shoyer: Consider switching this to `(key,) = data_array.dims`. This removes the otherwise mysterious `0` literal and serves as an implicit assert statement.

xarray/core/dataset.py
shoyer: Also, can you make this just `if` instead of `elif`? This isn't a natural alternative to the dimensionality check, so it makes more sense in a separate `if`.

xarray/core/dataset.py
```diff
@@ -2742,6 +2743,69 @@ def roll(self, **shifts):
...
+        for data_array in aligned_other_vars:
+            if len(data_array.dims) > 1:
+                raise ValueError("Input DataArray has more than 1 dimension.")
+            elif data_array.dtype == object and\
+            LooseVersion(np.__version__) < LooseVersion('1.11.0'):
+                    raise NotImplementedError(
+                        'sortby uses np.lexsort under the hood, which '
+                        'requires numpy 1.11.0 or later to support '
+                        'object data-type.')
+            else:
```
shoyer: Another nit: Can you remove the `else` block here? It doesn't make sense to conditionally enter this part of the code, and it's especially weird considering the line below, which would have `key` as an undefined variable if it ever gets run.

chunweiyuan:
Looks ready now.

shoyer:
@chunweiyuan have you run flake8 on this?

chunweiyuan:
Today is the first time I've heard of flake8.  Is this how you guys standardize quality checks?  If so, do you run once for python 3.5 and once for 2.7?

shoyer:
> Today is the first time I've heard of flake8.

It's in the checklist for every new PR :).

> Is this how you guys standardize quality checks?

Yes. It would be nice to run this as a continuous integration test but we haven't set that up that.

Running it once with either Python 2 or 3 is fine -- it should give the same output either way.

chunweiyuan:
I just made some futile attempts to find that checklist, to no avail.  Mind helping me with a link? :)

shoyer:
Scroll to the top of this PR and look at your first post :)

chunweiyuan:
Haha, brilliant.

chunweiyuan:
` git diff upstream/master | flake8 --diff` only complains about a bunch of lines in `whats-new.rst` and `reshaping.rst`.  But these complaints don't make any sense to me.  Some examples:
```
doc/reshaping.rst:195:1: E112 expected an indented block
doc/reshaping.rst:203:1: E101 indentation contains mixed spaces and tabs
doc/reshaping.rst:203:1: W191 indentation contains tabs
doc/reshaping.rst:203:2: E113 unexpected indentation
```
I've played around with it a bit but not seen any changes to the complaints.  What is going on?  Should I even worry about the `.rst` files?

shoyer:
It's very strange that fake8 complains about non-Python files. I think you
can safely ignore those.

chunweiyuan:
If that's the case then I have nothing more to add.  I've run flake8 on the individual files I touched I as well, such as `flake8 xarray/core/dataset.py`, and none of my changes are flagged.

shoyer:
One last thing -- this needs an entry in `docs/api.rst` in order for the doc page to get built.

chunweiyuan:
`api.rst` should've been a checklist item :)

shoyer:
> api.rst should've been a checklist item :)

Indeed, I will add it right now!

Thanks for your contribution!!

chunweiyuan:
Thank you very much for your patience!

BTW, I think I messed up my co-worker's name in `whats-new.rst`.  The spelling is right (Kyle Heuton), but the extra indentation shouldn't be there.  As a result his name appears a bit out of place.  Would be very much obliged if you could fix that in your next push to master.  Gracias.

shoyer:
> BTW, I think I messed up my co-worker's name in whats-new.rst. The spelling is right (Kyle Heuton), but the extra indentation shouldn't be there. As a result his name appears a bit out of place. Would be very much obliged if you could fix that in your next push to master. Gracias.

No worries, `whats-new.rst` usually requires a fix-up before each release

PR #1393, opened 2017-05-02
dl tutorial files to tmp directory, then move them once successful

### LINKED ISSUE #1392: tutorial datasets can be corrupted on dl with bad connection
Corrupted datasets then do not perform as expected during tutorial exercises. See the list discussion for more details: https://groups.google.com/forum/#!topic/xarray/F5UKkZZsMec

gidden:
closes #1392

 - [x] tests added / passed
 - [x] whatsnew entry

I'm not sure how to best add tests for this. Also not sure how best to import `shutil`. I followed the current pattern for `os`, but am not sure why that pattern exists in this file. I can clean up further if this is deemed useful.

### DIFF
--- a/xarray/tests/test_tutorial.py
+++ b/xarray/tests/test_tutorial.py
@@ -1,23 +1,26 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+
 import os
+import pytest
 
 from xarray import tutorial, DataArray
 from xarray.core.pycompat import suppress
 
-from . import TestCase, unittest
+from . import TestCase, network
 
 
-@unittest.skip('TODO: make this conditional on network availability')
-class Test_load_dataset(TestCase):
+@network
+class TestLoadDataset(TestCase):
 
     def setUp(self):
         self.testfile = 'tiny'
         self.testfilepath = os.path.expanduser(os.sep.join(
             ('~', '.xarray_tutorial_data', self.testfile)))
         with suppress(OSError):
-            os.remove(self.testfilepath)
+            os.remove('{}.nc'.format(self.testfilepath))
+            os.remove('{}.md5'.format(self.testfilepath))
 
     def test_download_from_github(self):
         ds = tutorial.load_dataset(self.testfile)
--- a/xarray/tests/test_tutorial.py
+++ b/xarray/tests/test_tutorial.py
@@ -20,6 +20,7 @@ class TestLoadDataset(TestCase):
             ('~', '.xarray_tutorial_data', self.testfile)))
         with suppress(OSError):
             os.remove('{}.nc'.format(self.testfilepath))
+        with suppress(OSError):
             os.remove('{}.md5'.format(self.testfilepath))
 
     def test_download_from_github(self):
--- a/xarray/tutorial.py
+++ b/xarray/tutorial.py
@@ -10,6 +10,7 @@ from __future__ import division
 from __future__ import print_function
 
 import os as _os
+import shutil as _shutil
 
 from .backends.api import open_dataset as _open_dataset
 from .core.pycompat import urlretrieve as _urlretrieve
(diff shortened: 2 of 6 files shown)


xarray/tutorial.py
```diff
@@ -42,18 +43,27 @@ def load_dataset(name, cache=True, cache_dir=_default_cache_dir,
...
+            _os.mkdir(tmpdir)
 
         url = '/'.join((github_url, 'raw', 'master', fullname))
-        _urlretrieve(url, localfile)
+        _urlretrieve(url, tmpfile)
+
+        if not _os.path.exists(tmpfile):
+            raise ValueError('File could not be downloaded, please try again')
+
+        _shutil.move(tmpfile, localfile)
```
fmaussion: I thought that the issue wasn't that the file isn't downloaded, but rather that it is incompletely downloaded? Will this new temporary step than solve this? 

Is there a way to make ``urlretrieve`` more robust, with checksums or something?

gidden:
It seems like urlretrieve is coming directly from urllib. It seems in both
Python 2 and 3 ContentTooShort error should be thrown and isnt. Perhaps
checksums are the right way to go.

gidden:
Successful locally with pydata/xarray-data#9:

```
In [3]: xr.tutorial.load_dataset('air_temperature', github_url='https://github.com/gidden/xarray-data', branch='md5')
Out[3]: 
<xarray.Dataset>
Coordinates:
  * time     (time) datetime64[ns] 2013-01-01 2013-01-01T06:00:00 ...
Data variables:
    air      (time, lat, lon) float64 241.2 242.5 243.5 244.0 244.1 243.9 ...
Attributes:
    Conventions:  COARDS
    platform:     Model
```

shoyer:
Thanks for taking this on @gidden.

> Also not sure how best to import shutil. I followed the current pattern for os, but am not sure why that pattern exists in this file.

I think this exists to work around the fact that IPython shows everything that isn't private in auto-complete (https://github.com/ipython/ipykernel/issues/129). But we should probably switch back to normal imports in this module -- this is really more of an IPython issue.

gidden:
Ok, sounds good. `shutil` was no longer needed, but the md5 checksum was
implemented as private. So I think all is well.

xarray/tutorial.py
```diff
@@ -18,9 +20,17 @@
 _default_cache_dir = _os.sep.join(('~', '.xarray_tutorial_data'))
 
 
+def _md5(fname):
```
shoyer: Call this something more descriptive like `check_file_md5_hash`?

gidden: should it remain private (for ipython completion) or no?

shoyer: I wouldn't stress about that. Given that we use it in the script in xarray-data, it's fine either way for me.

gidden:
ok, this should be good to go

shoyer:
Thanks @gidden, this look great to me. Could you kindly add a brief note to the "Bug fixes" section of `whats-new.rst`?

shoyer:
I'd be slightly more comfortable if we had this being unit tested on Travis-CI, but that's somewhat tricky to setup because we don't wan to require network access for the test suite by default. You could do this by adding a flag like `--run-network-tests` in `conftest.py`, and then hooking it through in `xarray/tests/__init__.py` like `--run-flaky`. I would be awesome if you could take a look at that and set the Travis/Appveyor tests to test the tutorial, but isn't strictly required to merge this.

gidden:
Hey @shoyer, if I have time later I will try to address the unit test issue. If I'm too late, feel free to pull this in the meantime.

gidden:
hey @shoyer, @fmaussion. test decorator added with tutorial dataset unit test. i think this is good to go.

xarray/tests/test_tutorial.py
```diff
@@ -1,23 +1,26 @@
...
+class TestLoadDataset(TestCase):
 
     def setUp(self):
         self.testfile = 'tiny'
         self.testfilepath = os.path.expanduser(os.sep.join(
             ('~', '.xarray_tutorial_data', self.testfile)))
         with suppress(OSError):
-            os.remove(self.testfilepath)
+            os.remove('{}.nc'.format(self.testfilepath))
+            os.remove('{}.md5'.format(self.testfilepath))
```
shoyer: This should go in its own `suppress` block

gidden: done!

shoyer:
thanks!

Issue #1401, opened 2017-05-09
Verify xarray works with bottleneck 1.2


shoyer:
This is somewhat time sensitive:
https://github.com/kwgoodman/bottleneck/issues/168

fmaussion:
Looks good to me: https://travis-ci.org/pydata/xarray/jobs/230642161

(see the test-bn-master branch)

shoyer:
@fmaussion can you make a PR adding this as an optional build to our test matrix?

fmaussion:
Sure. Have to go now but will do that later today

PR #1402, opened 2017-05-10
Adds bottleneck to the test suite

fmaussion:
- [x] closes https://github.com/pydata/xarray/issues/1401
 - [ ] tests added / passed
 - [ ] whatsnew entry

I don't know if we need to install all packages to cover all the tests which rely on bottleneck, but this was the easiest.


.travis.yml
```diff
@@ -39,6 +39,8 @@ matrix:
     env: CONDA_ENV=py36-pandas-dev
   - python: 3.6
     env: CONDA_ENV=py36-condaforge-rc
+  - python: 3.6
+    env: CONDA_ENV=py36-bn
```
shoyer: For consistency with the other builds, call this py36-bottleneck-dev?

fmaussion: done

shoyer:
Thanks @fmaussion

PR #1411, opened 2017-05-16
fixed dask prefix naming

### LINKED ISSUE #1343: Some XArray key names don't group nicely

rabernat:
- [x] Closes #1343
 - [x] Tests added / passed

I am starting a new PR for this since the original one (#1345) was not branched of my own fork.

As the discussion there stood, @shoyer suggested that `dataset.chunk` should also be updated to match the latest conventions in dask naming. The relevant code is here

```python
        def maybe_chunk(name, var, chunks):
            chunks = selkeys(chunks, var.dims)
            if not chunks:
                chunks = None
            if var.ndim > 0:
                token2 = tokenize(name, token if token else var._data)
                name2 = '%s%s-%s' % (name_prefix, name, token2)
                return var.chunk(chunks, name=name2, lock=lock)
            else:
                return var

        variables = OrderedDict([(k, maybe_chunk(k, v, chunks))
                                 for k, v in self.variables.items()])
```

Currently, `chunk` has an optional keyword argument `name_prefix='xarray-'`. Do we want to keep this optional? 

IMO, the current naming logic in `chunk` is not a problem for dask and will not cause problems for the distributed bokeh dashboard (as `open_dataset` did).


rabernat:
This seems to be interacting badly with `autoclose`. 

@pwolfram: can you give any insight into what might be going on? 
This is an interesting trace from travis:
https://travis-ci.org/pydata/xarray/jobs/232969599#L3368

```
xarray/tests/__init__.py:153: in assertDatasetIdentical
    assert_identical(d1, d2)
a = <[AssertionError("internal failure: file must be open if `autoclose=True` is used.") raised in repr()] SafeRepr object at 0x7fd8d83c7e60>
b = <xarray.Dataset>
...
```

Does autoclose assume that the dask object names actually match the filenames?

shoyer:
Note that the test failures are also showing different numbers from `assert_allclose` -- the "AssertionError: internal failure" bit is a broken error message generated by the test suite (somehow due to how we close files after we use them).

I think this is actually a problem with providing unique names for dask. Previously, the filename ended up as part of the prefix explicitly, but now it needs to be explicitly included as input to `token`.

rabernat:
> I think this is actually a problem with providing unique names for dask. Previously, the filename ended up as part of the prefix explicitly, but now it needs to be explicitly included as input to token.

I thought that was already happening. Here is the token
```python
             token = tokenize(file_arg, group, decode_cf, mask_and_scale,
                               decode_times, concat_characters, decode_coords,
                               engine, chunks, drop_variables)
```

`file_arg` has the filename in it, no?

shoyer:
> file_arg has the filename in it, no?

Actually, apparently not necessarily -- look a few lines further up to where it is defined.

rabernat:
Ok, I think I fixed it.

shoyer:
Thanks @rabernat !

Issue #475, opened 2015-07-15
API design for pointwise indexing


jhamman:
There have been a number of threads discussing possible improvements/extensions to `xray` indexing. The current indexing behavior for `isel` is orthogonal indexing - in other words, each coordinate is treated independently (see #214 and #411 for more discussion).

So the question: what is the best way to incorporate diagonal or pointwise indexing in `xray`? I see two main goals / applications:
1. support simple form of [_`numpy`_ style integer array indexing](http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#purely-integer-array-indexing)
2. support pointwise array indexing along coordinates via computation of nearest-neighbor indexes - I think this can also be thought of as a form of resampling.

Input from @WeatherGod, @wholmgren, and @shoyer would be great.

shoyer:
So, the good news is that once we figure out the API for pointwise indexing, I think the nearest-neighbor part could be as simple as supplying `method='nearest'`.

The challenge is that we want to go from an DataArray that looks like this:
```
In [4]: arr = xray.DataArray([[1, 2], [3, 4]], dims=['x', 'y'])

In [5]: arr
Out[5]:
<xray.DataArray (x: 2, y: 2)>
array([[1, 2],
       [3, 4]])
Coordinates:
  * x        (x) int64 0 1
  * y        (y) int64 0 1
```

To one that looks like that:
```
In [6]: xray.DataArray([1, 4], {'x': ('c', [0, 1]), 'y': ('c', [0, 1])}, dims='c')
Out[6]:
<xray.DataArray (c: 2)>
array([1, 4])
Coordinates:
    y        (c) int64 0 1
    x        (c) int64 0 1
  * c        (c) int64 0 1
```

Somehow, we need to figure out the name for the new dimension (`c` in this example).

My thought would be to have methods `sel_points` and `isel_points` that work similarly to `sel` and `isel`. This is straightforward if you already have xray 1D objects with a labeled dimension: `arr.sel_points(x=x, y=y)`, where `x` and `y` are along the `c` dimension.

If you don't already have 1D xray objects, I suppose we could also allow `arr.sel_points(x=('c', [0, 1]), y=('c', [0, 1]))` or `arr.sel_points('c', x=[0, 1], y=[0, 1])`.

wholmgren:
Seems like if your method is going to be named ``sel_points`` then ``points`` is a reasonable dimension name. Maybe support a ``name`` kwarg?

[...]

PR #1412, opened 2017-05-17
Multiindex scalar coords, fixes #1408

### LINKED ISSUE #1408: .sel does not keep selected coordinate value in case with MultiIndex
`.sel` method usually keeps selected coordinate value as a scalar coordinate

```python
In[4] ds1 = xr.Dataset({'foo': (('x',), [1, 2, 3])}, {'x': [1, 2, 3], 'y': 'a'})

Out[4]: 
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    y        <U1 'a'
    x        int64 1
Data variables:
    foo      int64 1
```

But in MultiIndex case, does not.

```python
In[5] ds2 = xr.Dataset({'foo': (('x',), [4, 5, 6])}, {'x': [1, 2, 3], 'y': 'b'})
        ds = xr.concat([ds1, ds2], dim='y').stack(yx=['y', 'x'])
        ds.sel(x=2)
Out[5] 
<xarray.Dataset>
Dimensions:  (y: 2)
Coordinates:
  * y        (y) object 'a' 'b'
Data variables:
    foo      (y) int64 2 5
```
x is gone.
Is it a desired behavior?

fujiisoup:
- [x] Closes #1408 
 - [x] Tests added / passed

To fix #1408, 
This modification works, but actually I do not fully satisfied yet.
There are `if` statements in many places.

The major changes I made are
1. `variable.__getitem__` now returns an OrderedDict if a single element is selected from MultiIndex.
2. `indexing.remap_level_indexers` also returns `selected_dims` which is a map from the original dimension to the selected dims which will be a scalar coordinate.

Change 1 keeps level-coordinates even after `ds.isel(yx=0)`.
Change 2 enables to track which levels are selected, then the selected levels are changed to a scalar coordinate.

I guess much smarter solution should exist.
I would be happy if anyone gives me a comment.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -103,11 +103,11 @@ class _LocIndexer(object):
         return indexing.remap_label_indexers(self.data_array, key)
 
     def __getitem__(self, key):
-        pos_indexers, new_indexes = self._remap_key(key)
+        pos_indexers, new_indexes, _ = self._remap_key(key)
         return self.data_array[pos_indexers]._replace_indexes(new_indexes)
 
     def __setitem__(self, key, value):
-        pos_indexers, _ = self._remap_key(key)
+        pos_indexers, _, _ = self._remap_key(key)
         self.data_array[pos_indexers] = value
 
 
@@ -679,9 +679,9 @@ class DataArray(AbstractArray, BaseDataObject):
         Dataset.sel
         DataArray.isel
         """
-        pos_indexers, new_indexes = indexing.remap_label_indexers(
-            self, indexers, method=method, tolerance=tolerance
-        )
+        pos_indexers, new_indexes, selected_dims = \
+            indexing.remap_label_indexers(
+                self, indexers, method=method, tolerance=tolerance)
         result = self.isel(drop=drop, **pos_indexers)
         return result._replace_indexes(new_indexes)
 
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -103,8 +103,10 @@ class _LocIndexer(object):
         return indexing.remap_label_indexers(self.data_array, key)
 
     def __getitem__(self, key):
-        pos_indexers, new_indexes, _ = self._remap_key(key)
-        return self.data_array[pos_indexers]._replace_indexes(new_indexes)
+        pos_indexers, new_indexes, selected_dims = self._remap_key(key)
+        ds = self.data_array._to_temp_dataset().isel(**pos_indexers)
+        return self.data_array._from_temp_dataset(
+            ds._replace_indexes(new_indexes, selected_dims))
 
     def __setitem__(self, key, value):
         pos_indexers, _, _ = self._remap_key(key)
@@ -255,23 +257,6 @@ class DataArray(AbstractArray, BaseDataObject):
             coords = OrderedDict((k, v) for k, v in self._coords.items()
                                  if set(v.dims) <= allowed_dims)
         return self._replace(variable, coords, name)
-
-    def _replace_indexes(self, indexes):
-        if not len(indexes):
-            return self
-        coords = self._coords.copy()
-        for name, idx in indexes.items():
-            coords[name] = IndexVariable(name, idx)
-        obj = self._replace(coords=coords)
-
-        # switch from dimension to level names, if necessary
-        dim_names = {}
-        for dim, idx in indexes.items():
-            if not isinstance(idx, pd.MultiIndex) and idx.name != dim:
-                dim_names[dim] = idx.name
-        if dim_names:
-            obj = obj.rename(dim_names)
-        return obj
 
     __this_array = _ThisArray()
 
@@ -682,8 +667,9 @@ class DataArray(AbstractArray, BaseDataObject):
         pos_indexers, new_indexes, selected_dims = \
             indexing.remap_label_indexers(
                 self, indexers, method=method, tolerance=tolerance)
-        result = self.isel(drop=drop, **pos_indexers)
-        return result._replace_indexes(new_indexes)
+        ds = self._to_temp_dataset().isel(drop=drop, **pos_indexers)
+        return self._from_temp_dataset(
+            ds._replace_indexes(new_indexes, selected_dims))
 
     def isel_points(self, dim='points', **indexers):
         """Return a new DataArray whose dataset is given by pointwise integer
--- a/xarray/core/groupby.py
+++ b/xarray/core/groupby.py
@@ -367,6 +367,8 @@ class GroupBy(object):
         """ This constructs MultiIndex if applied does not have dim.
         It may happen if a single item is selected from MultiIndex-ed array.
         """
+        if not hasattr(self._group, 'to_index'):
+            return applied
         index = self._group.to_index()
         if not isinstance(index, pd.MultiIndex):
             return applied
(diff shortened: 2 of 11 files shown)


xarray/core/dataarray.py
```diff
@@ -256,23 +258,6 @@ def _replace_maybe_drop_dims(self, variable, name=__default):
                                  if set(v.dims) <= allowed_dims)
         return self._replace(variable, coords, name)
 
-    def _replace_indexes(self, indexes):
```
fujiisoup: I removed this method and use `Dataset._replace_indexes` instead, to reduce duplicates.

xarray/core/groupby.py
```diff
@@ -362,6 +362,22 @@ def _maybe_unstack(self, obj):
                     del obj.coords[dim]
         return obj
 
+    def _maybe_stack(self, applied):
```
fujiisoup: This method becomes necessary, because now we cannot do `xr.concat([ds.isel(yx=i)] for i in range(*)], dim='yx')` because `ds.isel(yx=i)` does not have `yx` anymore.

shoyer:
> variable.__getitem__ now returns an OrderedDict if a single element is selected from MultiIndex.

I don't like this change. It breaks an important invariant, which is that indexing a Variable returns another Variable.

I do agree with indexing along a MultiIndex dimension should unpacking the tuple for coordinates, but only for coordinates. So this needs to be somewhere in the `Dataset.isel` logic, not `Variable.isel`.

Consider indexing `ds['yx']` from your example in the linked issue. With the current version of xarray:
```
In [7]: ds['yx']
Out[7]:
<xarray.DataArray 'yx' (yx: 6)>
array([('a', 1), ('a', 2), ('a', 3), ('b', 1), ('b', 2), ('b', 3)], dtype=object)
Coordinates:
  * yx       (yx) MultiIndex
  - y        (yx) object 'a' 'a' 'a' 'b' 'b' 'b'
  - x        (yx) int64 1 2 3 1 2 3

In [8]: ds['yx'][0]
Out[8]:
<xarray.DataArray 'yx' ()>
array(('a', 1), dtype=object)
Coordinates:
    yx       object ('a', 1)
```
We want to change the indexing behavior to this:
```
In [8]: ds['yx'][0]
Out[8]:
<xarray.DataArray 'yx' ()>
array(('a', 1), dtype=object)
Coordinates:
    y        object 'a'
    x        int64 1
```
But we don't want to change what happens to the DataArray itself -- it should still be a scalar object array.

I tested this example on your PR branch, and it actually crashes with `KeyError`.

fujiisoup:
@shoyer 
Thanks for the comment.

> It breaks an important invariant, which is that indexing a Variable returns another Variable.

I totally agree with you.

In the last commit, I moved the unpacking functionality into `Dataset`, and restored the modification in `Variable` class I made.
I think the current is cleaner than my previous one, but I'm not yet comfortable with it.
There are a lot of functions or `if`-statements related to `MultiIndex` in different places.
I guess they should be bundled in one place.

Adding functions is easy but simplifying them are difficult...

If anyone show a direction, I will try the improvement.

xarray/core/dataarray.py
benbovy: If I remember well, we used duplicates to avoid using `_to_temp_dataset` and `_from_temp_dataset` in `__getitem__`. But now that `_replace_indexes` has more logic implemented, maybe it is a good idea to reduce duplicates?

fujiisoup: Thanks.
Yes, in `DataArray.__getitem__` and also in `DataAarray.isel`, `_to_temp_dataset` and `_from_temp_dataset` are now being used.

benbovy:
A possible direction to reduce the `if` statements in many different places would be to just return `pos_indexers` in `indexing.remap_level_indexers` - as it was the case before adding multi-index support - and instead put in  `Dataset.isel` all the logic for checking `MultiIndex` and maybe convert it to `Index` and/or scalar coordinates and maybe rename dimension.

This would simplify many things, although I haven't thought about about all other possible issues it would create (perfomance, etc.). Also, `DataArray.loc` doesn't seem to use `Dataset.isel`.

Here is another case related to this PR. From the example in the linked issue, the current behavior is

```python
In [9]: ds.isel(yx=[0, 1])
Out[9]: 
<xarray.Dataset>
Dimensions:  (yx: 2)
Coordinates:
  * yx       (yx) MultiIndex
  - y        (yx) object 'a' 'a'
  - x        (yx) int64 1 2
Data variables:
    foo      (yx) int64 1 2
```

Do we want to also change the behavior to this?

```python
In [10]: ds.isel(yx=[0, 1])
Out[10]: 
<xarray.Dataset>
Dimensions:  (x: 2)
Coordinates:
  * x        (x) int64 1 2
    y        object 'a'
Data variables:
    foo      (yx) int64 1 2
```

To me it looks like it is a bit too magical, but just wondering what you think...

shoyer:
> To me it looks like it is a bit too magical, but just wondering what you think...

Agreed, this also seems too magical to me.

fujiisoup:
I also agree. It seems too magical.

But I slightly changed my mind.
I notice what I really want to have is not particular scalar coordinate in MultiIndex,
but 'unified' interface between normal `Vraiable` and `MultiIndex`.

The current structure is illustrated as follows,

current_coord


The `MultiIndex` has different characteristics from normal `Variable`.
For example, if we do `ds.sel(x=2)`, it makes a scalar coordinate and normal Variable.
The backward process might be `.expand_dims().stack()`.
This is different from normal `Variable` behavior.
And because of it, MultiIndex should be treated in special way in every place.
(Deprecating the automatic-renaming does not change things so much.)

I am wondering if we could have the following class structure things become simpler

my_proposal

In this picture, MultiIndex can have `scalar` as its level and `.isel()` produces it.
This process can be traced backward by `.expand_dims()` or `.concat()` as in normal `Variable`.

I understand it is different from `pandas.MultiIndex` structure, and we need to expand our wrapper extensively if we decide to realize it (as written in red).
But I feel this symmetric structure could make it easy to expand `MultiIndex` functionalities in future.

Any thoughts are welcome.
(Should move discussion to another issue?)

shoyer:
@fujiisoup Yes, the solution of writing a MultiIndex wrapper for xarray looks much cleaner to me. I like the look of this proposal! (Those diagrams are also very helpful)

I guess this could be implemented as a `pandas.MultiIndex` along with a list of scalar coordinates?

benbovy:
I also agree that a MultiIndex wrapper would be to be the way to go. I think I understand the diagrams, but what remains a bit unclear to me is how this could be implemented.

In particular, how would this wrapper work with `IndexVariable`?

Currently, `IndexVariable` warps either a `pandas.Index` or a `pandas.MultiIndex` and for the latter case `IndexVariable.get_level_variable` can generate new `IndexVariable` objects so that MultiIndex levels are accessible as "virtual coordinates".

Would `IndexVariable` warp a MultiIndex wrapper instead (levels + scalars), and also be able to generate new scalar `Variable` objects that will be accessible as virtual coordinates?

This is maybe slightly off topic, but more generally I'm also wondering how this would co-exist with potentially other kinds of multi-level indexes (see this [comment](https://github.com/pydata/xarray/issues/475#issuecomment-241821366)).

fujiisoup:
@benbovy 
Thanks for the valuable comments.
Actually I can not fully imagine how the actual implementation looks like currently,
but I also think the virtual variable access needs some tricks. 
This is an essential functionality of the MultiIndex-coordinate,
I will try to investigate it.

Thanks.

fujiisoup:
Replaced by a new PR #1426 .

PR #1413, opened 2017-05-17
concat prealigned objects

### LINKED ISSUE #1385: slow performance with open_mfdataset
We have a dataset stored across multiple netCDF files. We are getting very slow performance with `open_mfdataset`, and I would like to improve this.

Each individual netCDF file looks like this:
```python
%time ds_single = xr.open_dataset('float_trajectories.0000000000.nc')
ds_single
```
```
CPU times: user 14.9 ms, sys: 48.4 ms, total: 63.4 ms
Wall time: 60.8 ms

<xarray.Dataset>
Dimensions:  (npart: 8192000, time: 1)
Coordinates:
  * time     (time) datetime64[ns] 1993-01-01
  * npart    (npart) int32 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 ...
Data variables:
    z        (time, npart) float32 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 ...
    vort     (time, npart) float32 -9.71733e-10 -9.72858e-10 -9.73001e-10 ...
    u        (time, npart) float32 0.000545563 0.000544884 0.000544204 ...
    v        (time, npart) float32 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 ...
    x        (time, npart) float32 180.016 180.047 180.078 180.109 180.141 ...
    y        (time, npart) float32 -79.9844 -79.9844 -79.9844 -79.9844 ...
```

As shown above, a single data file opens in ~60 ms.

When I call `open_mdsdataset` on 49 files (each with a different `time` dimension but the same `npart`), here is what happens:

```python
%time ds = xr.open_mfdataset('*.nc', )
ds
```
```
CPU times: user 1min 31s, sys: 25.4 s, total: 1min 57s
Wall time: 2min 4s

<xarray.Dataset>
Dimensions:  (npart: 8192000, time: 49)
Coordinates:
  * npart    (npart) int64 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 ...
  * time     (time) datetime64[ns] 1993-01-01 1993-01-02 1993-01-03 ...
Data variables:
    z        (time, npart) float64 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 -0.5 ...
    vort     (time, npart) float64 -9.717e-10 -9.729e-10 -9.73e-10 -9.73e-10 ...
    u        (time, npart) float64 0.0005456 0.0005449 0.0005442 0.0005437 ...
    v        (time, npart) float64 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 ...
    x        (time, npart) float64 180.0 180.0 180.1 180.1 180.1 180.2 180.2 ...
    y        (time, npart) float64 -79.98 -79.98 -79.98 -79.98 -79.98 -79.98 ...
```

It takes over 2 minutes to open the dataset. Specifying `concat_dim='time'` does not improve performance.

Here is `%prun` of the `open_mfdataset` command.

```
         748994 function calls (724222 primitive calls) in 142.160 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       49   62.455    1.275   62.458    1.275 {method 'get_indexer' of 'pandas.index.IndexEngine' objects}
       49   47.207    0.963   47.209    0.963 base.py:1067(is_unique)
      196    7.198    0.037    7.267    0.037 {operator.getitem}
       49    4.632    0.095    4.687    0.096 netCDF4_.py:182(_open_netcdf4_group)
      240    3.189    0.013    3.426    0.014 numeric.py:2476(array_equal)
       98    1.937    0.020    1.937    0.020 {numpy.core.multiarray.arange}
4175/3146    1.867    0.000    9.296    0.003 {numpy.core.multiarray.array}
       49    1.525    0.031  119.144    2.432 alignment.py:251(reindex_variables)
       24    1.065    0.044    1.065    0.044 {method 'cumsum' of 'numpy.ndarray' objects}
       12    1.010    0.084    1.010    0.084 {method 'sort' of 'numpy.ndarray' objects}
5227/4035    0.660    0.000    1.688    0.000 collections.py:50(__init__)
       12    0.600    0.050    3.238    0.270 core.py:2761(insert)
12691/7497    0.473    0.000    0.875    0.000 indexing.py:363(shape)
   110728    0.425    0.000    0.663    0.000 {isinstance}
       12    0.413    0.034    0.413    0.034 {method 'flatten' of 'numpy.ndarray' objects}
       12    0.341    0.028    0.341    0.028 {numpy.core.multiarray.where}
        2    0.333    0.166    0.333    0.166 {pandas._join.outer_join_indexer_int64}
        1    0.331    0.331  142.164  142.164 <string>:1(<module>)
```

It looks like most of the time is being spent on `reindex_variables`. I understand why this happens...xarray needs to make sure the dimensions are the same in order to concatenate them together.

Is there any obvious way I could improve the load time? For example, can I give a hint to xarray that this `reindex_variables` step is not necessary, since I know that all the `npart` dimensions are the same in each file?

Possibly related to #1301 and #1340.

rabernat:
- [x] Closes #1385
 - [ ] Tests added / passed

This is an initial PR to bypass index alignment and coordinate checking when concatenating datasets.

### DIFF
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -13,8 +13,8 @@ from .variable import Variable, as_variable, IndexVariable, concat as concat_var
 
 
 def concat(objs, dim=None, data_vars='all', coords='different',
-           compat='equals', positions=None, indexers=None, mode=None,
-           concat_over=None):
+           compat='equals', positions=None, prealigned=False,
+           indexers=None,mode=None, concat_over=None):
     """Concatenate xarray objects along a new or existing dimension.
 
     Parameters
@@ -66,6 +66,10 @@ def concat(objs, dim=None, data_vars='all', coords='different',
         List of integer arrays which specifies the integer positions to which
         to assign each dataset along the concatenated dimension. If not
         supplied, objects are concatenated in the provided order.
+    prealigned : bool, optional
+        If True, the objects will be assumed to be already aligned. Coordinates
+        will be taken from the first object and ignored from the subsequent
+        objects.
     indexers, mode, concat_over : deprecated
 
     Returns
@@ -117,7 +121,7 @@ def concat(objs, dim=None, data_vars='all', coords='different',
     else:
         raise TypeError('can only concatenate xarray Dataset and DataArray '
                         'objects, got %s' % type(first_obj))
-    return f(objs, dim, data_vars, coords, compat, positions)
+    return f(objs, dim, data_vars, coords, compat, positions, prealigned)
 
 
 def _calc_concat_dim_coord(dim):
@@ -195,7 +199,8 @@ def _calc_concat_over(datasets, dim, data_vars, coords):
     return concat_over
 
 
-def _dataset_concat(datasets, dim, data_vars, coords, compat, positions):
+def _dataset_concat(datasets, dim, data_vars, coords, compat, positions,
+                    prealigned):
     """
     Concatenate a sequence of datasets along a new or existing dimension
     """
@@ -207,7 +212,10 @@ def _dataset_concat(datasets, dim, data_vars, coords, compat, positions):
 
     dim, coord = _calc_concat_dim_coord(dim)
     datasets = [as_dataset(ds) for ds in datasets]
-    datasets = align(*datasets, join='outer', copy=False, exclude=[dim])
+    if not prealigned:
+        datasets = align(*datasets, join='outer', copy=False, exclude=[dim])
+    else:
+        coords = 'minimal'
 
     concat_over = _calc_concat_over(datasets, dim, data_vars, coords)
 
@@ -228,21 +236,22 @@ def _dataset_concat(datasets, dim, data_vars, coords, compat, positions):
 
     # check that global attributes and non-concatenated variables are fixed
     # across all datasets
-    for ds in datasets[1:]:
-        if (compat == 'identical' and
-                not utils.dict_equiv(ds.attrs, result_attrs)):
-            raise ValueError('dataset global attributes not equal')
-        for k, v in iteritems(ds.variables):
-            if k not in result_vars and k not in concat_over:
-                raise ValueError('encountered unexpected variable %r' % k)
-            elif (k in result_coord_names) != (k in ds.coords):
-                raise ValueError('%r is a coordinate in some datasets but not '
-                                 'others' % k)
-            elif (k in result_vars and k != dim and
-                  not getattr(v, compat)(result_vars[k])):
-                verb = 'equal' if compat == 'equals' else compat
-                raise ValueError(
-                    'variable %r not %s across datasets' % (k, verb))
+    if not prealigned:
+        for ds in datasets[1:]:
+            if (compat == 'identical' and
+                    not utils.dict_equiv(ds.attrs, result_attrs)):
+                raise ValueError('dataset global attributes not equal')
+            for k, v in iteritems(ds.variables):
+                if k not in result_vars and k not in concat_over:
+                    raise ValueError('encountered unexpected variable %r' % k)
+                elif (k in result_coord_names) != (k in ds.coords):
+                    raise ValueError('%r is a coordinate in some datasets but not '
+                                     'others' % k)
+                elif (k in result_vars and k != dim and
+                      not getattr(v, compat)(result_vars[k])):
+                    verb = 'equal' if compat == 'equals' else compat
+                    raise ValueError(
+                        'variable %r not %s across datasets' % (k, verb))
 
     # we've already verified everything is consistent; now, calculate
     # shared dimension sizes so we can expand the necessary variables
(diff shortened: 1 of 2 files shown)


rabernat:
Let me expand on what this does.

Many netCDF datasets consist of multiple files with identical coordinates, except for one (e.g. time). With xarray we can open these datasets with `open_mfdataset`, which calls `concat` on the list of individual dataset objects. `concat` calls `align`, which loads all of the dimension indices (and, optionally, non-dimension coordinates) from each file and checks them for consistency / alignment. 

This `align` step is potentially quite expensive for big collections of files with large indices. For example, an unstructured grid or particle-based dataset would just have a single dimension coordinate, with the same length as the data variables. If the user knows that the datasets are already aligned, this PR enables the alignment step to be skipped by passing the argument `prealigned=True` to `concat`.

This PR is a draft in progress. I still need to propagate the `prealigned` argument up to `auto_combine` and `open_mfdataset`.

An alternative API would be to add another option to the `coords` keywork, i.e. `coords='prealigned'`.

Feedback welcome.

xarray/core/combine.py
```diff
@@ -207,7 +212,10 @@ def _dataset_concat(datasets, dim, data_vars, coords, compat, positions):
 
     dim, coord = _calc_concat_dim_coord(dim)
     datasets = [as_dataset(ds) for ds in datasets]
-    datasets = align(*datasets, join='outer', copy=False, exclude=[dim])
+    if not prealigned:
+        datasets = align(*datasets, join='outer', copy=False, exclude=[dim])
+    else:
+        coords = 'minimal'
```
shoyer: It's bad form to unilaterally override an argument with another value -- it's better to raise an error (or maybe a warning).

The only value of `coords` that really breaks here is `'different'`, and even that value could conceivably make sense.

shoyer:
This enhancement makes a lot of sense to me.

Two things worth considering: 

1. Given a collection of datasets, how do I know if setting `prealigned=True` will work? This is where my PR adding `xr.align(..., join='exact')` could help (I can finish that up). Maybe it's worth adding `xr.is_aligned` or something similar.
2. What happens if things go wrong? It's okay if the behavior is undefined (or could give wrong results) but we should document that. Ideally we should raise sensible errors at some later time, e.g., when the dask arrays are computed. This might or might not be possible to do efficiently with dask, if the result of all the equality checks are consolidated and added into the dask graphs of the results.

xarray/core/combine.py
rabernat: What about just adding the option `coords='prealigned'`?

My initial thought was that, for prealigned data, all coords should just be drawn from the first object. But on second thought, what if there are other coords in the later dataset that do need to be concatenated, e.g. concat over `time` with an auxiliary coordinates `iteration_number` with dimension `time`.

It definitely doesn't work with `coords='different'`. I have not tried all the other options. I have a hard time conceptualizing what the different `coords` options do. Some guidance would be very welcome. I don't really understand what the function `_calc_concat_over` does.

rabernat:
> Given a collection of datasets, how do I know if setting prealigned=True will work?
I guess we would want to check that (a) the necessary variables and dimensions exist in all datasets and (b) the dimensions have the same _length_. We would want to bypass the actual reading of the indices. I agree it would be nicer to subsume this logic into `align`.

What is `xr.align(..., join='exact')` supposed to do? 

> What happens if things go wrong?
I can add more careful checks once we sort out the align question.

shoyer:
> What is `xr.align(..., join='exact')` supposed to do?

It verifies that all dimensions have the same length, and coordinates along all dimensions (used for indexing) also match. Unlike the normal version of `align`, it doesn't do any indexing -- the outputs are always the same as the inputs.

It *does not* check that the necessary dimensions and variables exist in all datasets. But we should do that as part of the logic in `concat` anyways, since the xarray data model always requires knowing variables and their dimensions.

rabernat:
As I think about this further, I realize it might be futile to avoid reading the dimensions from all the files. This is a basic part of how `open_dataset` works.

shoyer:
Well, we could potentially write a fast path constructor for loading
multiple netcdf files that avoids open_dataset. We just need another way to
specify the schema, e.g., using NCML.

rabernat:
Since the expensive part (for me) is actually reading all the coordinates, I'm not sure that this PR makes sense any more.

The same thing I am going for here could probably be accomplished by allowing the user to pass `join='exact'` via `open_mfdataset`. A related optimization would be to allow the user to pass `coords='minimal'` (or other `concat` coords options) via `open_mfdataset`. 

For really big datasets, I think we will want to go the NCML approach, generating the xarray metadata as a pre-processing step. Then we could add a function like `open_ncml_dataset` to xarray which would parse this metadata and construct the dataset in a more efficient way (i.e. not reading redundant coordinates).

shoyer:
Sounds good to me!

jhamman:
@rabernat - I'm just catching up on this issue. Is you last comment indicating that we should close this PR?

rabernat:
Yes, I think it should be closed. There are better ways to accomplish the desired goals.

jhamman:
Okay thanks, closing now. We can always reopen this if necessary.

PR #1414, opened 2017-05-18
Speed up `decode_cf_datetime`

### LINKED ISSUE #1399: `decode_cf_datetime()` slow because `pd.to_timedelta()` is slow if floats are passed
Hi,
 `decode_cf_datetime` is slowed down because it [always passes floats](https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L129) to [`pd.to_timedelta`](https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L154), while `pd.to_timedelta` is much faster when working on integers.

[Here](https://gist.github.com/cchwala/157b87d4e413b560f8ad8555a330b937#file-timing_for_timedelta64_and_pandas_timedelta-ipynb) is a notebook that shows the differences. Working with integers is approx. one order of magnitude faster.

Hence, it would be great to automatically do the conversion from raw time value floats to integers in nanoseconds where possible (likely limited to resolutions bellow days or hours to avoid coping with different durations numbers of nanoseconds within e.g. different months).

As alternative, maybe avoid forcing the cast to floats and indicate in the docstring that the raw values should be integers to speed up the conversion.

This could possibly also be resolved in `pd.to_timedelta` but I assume it will be more complicated to deal with all the edge cases there.

cchwala:
- [x] Closes #1399
 - [x] Tests added / passed

Instead of casting the input numeric dates to float, they are now
casted to nanoseconds as int64 which makes `pd.to_timedelta()`
work much faster (x100 speedup on my machine).

On my machine all existing tests for `conventions.py` pass. Overflows should be handled by [these two already existing lines](https://github.com/cchwala/xarray/commit/d7d7c01f3e2f14c38c44e62f648b30474469b078#diff-d94eba38daa73be812c57c756f01f0daR158) since everything in the valid range of `pd.to_datetime` should be save.


xarray/conventions.py
```diff
@@ -155,10 +158,24 @@ def decode_cf_datetime(num_dates, units, calendar=None):
...
 
-        dates = (pd.to_timedelta(flat_num_dates, delta) + ref_date).values
+        # Cast input dates to integers of nanoseconds because `pd.to_datetime`
+        # works much faster when dealing with integers
+        ns_per_time_delta = {'us': 1e3,
+                             'ms': 1e6,
+                             's': 1e9,
+                             'm': 1e9 * 60,
+                             'h': 1e9 * 60 * 60,
+                             'D': 1e9 * 60 * 60 * 24}
```
jhamman: should we put this in a module level scope? Not sure we need to create this dictionary each time the `decode_cf_datetime` function is called.

cchwala: I left it there because the performance impact should be negligible, but it could of course be placed somewhere outside to create it only once.

shoyer: Yes, I would suggest saving this as a module level constant, `_NS_PER_TIME_DELTA`.

cchwala:
Thanks @shoyer and @jhamman for the feedback. I will change things accordingly. 

Concerning tests, I will think again about additional checking for correct handling of overflow. I must admit, that I am not 100% sure that every case is handled correctly by the current code and checked by the current tests. Will have to think about it a little when I find time within the next days...

cchwala:
Just a short notice. Sorry, for the delay. I am still working on this PR, but I am too busy right now to finish the overflow testing. I think I found some edge cases which have to be handled. I will provide more details soon.

jhamman:
@cchwala - Any update on the testing / overflow issue you mentioned?

cchwala:
@jhamman - Sorry. I was away from office (and everything related to work) for more than a month and had to catchup with a lot of things. I will sum up my stuff and post here, hopefully after todays lunch break.

cchwala:
@jhamman - I found some differences between the old code in master an my code when decoding values close to the np.datetime64 overflow. My code produces `NaT` where the old code returned some date. 

First wanted fix that. However, I may have found that the old implementation did not behave correctly when crossing the "overflow" line just slightly.

I have summed that up in a notebook [here](https://gist.github.com/cchwala/25efac7857c9b53f6f81d6fa44135a45).

My conclusion would be, that the code in this PR here is not only faster, but also more correct than the old one. However, since it is quite late in the evening and my head needs some rest, I would like to get a second (or third) opinion...

cchwala:
...but wait. The `NaT`s that my code produces beyond the int64 overflow should be valid dates, produced using `_decode_datetime_with_netcdf4`, right? 

Hence, I should still add a check for `NaT` results and fall back to the netcdf version then.

jhamman:
@cchwala - thanks for keeping this moving. Once you've taking another pass at the code and added a Whatsnew note, I'll give it a final review.

cchwala:
hmm... it's still complicated. To avoid the `NaT`s in my code, I tried to extend [the current overflow check](https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L155) so that it switches to `_decode_datetime_with_netcdf4()` earlier. This was my attempt:

```python
(pd.to_timedelta(flat_num_dates.min(), delta) -
 pd.to_timedelta(1, 'd') +
 ref_date)
(pd.to_timedelta(flat_num_dates.max(), delta) +
 pd.to_timedelta(1, 'd') +
 ref_date)
```
But unfortunately, as shown in my notebook above, `pandas.to_timedelta()` has a bug and does not detect the overflow in those esoteric cases that I have identified... I have filed this Issue pandas-dev/pandas/issues/17037 because it should be solved there.

Since I do not think this will be fixed soon (I would gladly look at it, but have no time and probably not enough knowledge about the `pandas` core stuff), I am not sure what to do. 

Do you want to merge this PR, knowing that there still is the overflow issue that was in the code before? Or should I continue to try to fix the current overflow bug in this PR?

shoyer:
> Do you want to merge this PR, knowing that there still is the overflow issue that was in the code before?

This still sounds like an improvement to me!

jhamman:
Agreed. Let's leave the overflow fix for later. I'll give this one more review and we'll try to get this merged.

cchwala:
@jhamman @shoyer 
This should be ready to merge.

Should I open an xarray issue concerning the bug with `pandas.to_timedelta()` or is it enough to have the issue I submitted for pandas? I think the bug should be resolved in xarray when it is resolved in pandas because then the [overflow check here](https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L155) should catch the cases I discovered.

jhamman:
Thanks @cchwala!

PR #1417, opened 2017-05-21
Update weather data example

fmaussion:
- [ ] Closes #xxxx
 - [ ] Tests added / passed

For some reason a plot from the examples was rendered statically instead of live. This is now fixed, together with an update of pandas which prevents reshaping operations on indexes. 

Tested [here](http://xarray.pydata.org/en/fix-docs/examples/weather-data.html)


shoyer:
Thanks @fmaussion! I think this dates to the old days of RTD (before conda support) when getting things to build correctly was difficult.

PR #1419, opened 2017-05-21
Add sphinx-gallery to the docs

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/1061
 - [ ] Tests added / passed

This uses [sphinx-gallery](http://sphinx-gallery.readthedocs.io) to illustrate some examples of the xarray workflow. See it rendered [here](http://xarray.pydata.org/en/fix-docs/auto_gallery/index.html)

It's nicer for the gallery if there is a plot to draw in the end but it doesn't *have* to be a plot. Once the gallery gets more examples it's possible to sort them by topic (see the sphinx gallery documentation for an overview of the possibilities).

I could think of several other things we could add (including moving or copying @jhamman and @rabernat 's examples to the gallery), but I thought it's better to merge quickly in order to encourage the community to contribute with more examples before the next release.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -48,6 +48,7 @@ help:
 clean:
 	rm -rf $(BUILDDIR)/*
 	rm -rf generated/*
+	rm -rf auto_gallery/
 
 .PHONY: html
 html:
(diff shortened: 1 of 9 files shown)


shoyer:
This looks great -- but maybe move the "Gallery" section under "Examples"?

fmaussion:
>  maybe move the "Gallery" section under "Examples"?

good idea: done.

rabernat:
Agreed, great idea to use sphinx-gallery. I agree that it's best to get the basic functionality working first. Then we can migrate the other examples to sphinx-gallery once we have a working template.

shoyer:
Thinking about this a little bit more -- maybe we would rather use something like nbsphinx that lets us directly embed ipython notebooks in the docs and/or write notes in the form of notebooks?

sphinx-gallery seems very well suited to an examples that produce a single plot but not much else. Maybe we should put this gallery on the "Plotting" page instead?

Anyways, I'm happy to trust your judgment here. I'm happy to merge now and iterate later.

(Also, I'm not a huge fan of the prominent "Download all" links, they seem mostly like clutter to me -- but that's a side note.)

fmaussion:
I agree with you @shoyer , but it can still provides a little bit more than mere plotting examples. 

How about naming the gallery "Recipes" and keep it in the examples section for now?

shoyer:
> How about naming the gallery "Recipes" and keep it in the examples section for now?

Sounds good to me.

PR #1421, opened 2017-05-23
First pass adding arbitrary object serialization

richlewis42:
This adds support for object serialization using the netCDF4-python backend..

Minimum working (at least appears to..) example, no tests yet.

I added `allow_object` kwarg (rather than `allow_pickle`, no reason to firmly attach pickle to the api, could use something else for other backends).

This is now for:

- `to_netcdf`
- `AbstractDataStore` (a `True` value raises `NotImplementedError` for everything but `NetCDF4DataStore`)
- `cf_encoder` which when `True` alters its behaviour to allow `dtype('O')` through.

`NetCDF4DataStore` handles this independently from the cf_encoder/decoder.  The dtype support made it hard to decouple, plus I think object serialization is a backend dependent issue.

There's a lot of potential for refactoring, just pushed this to get opinions about whether this
was a reasonable approach - I'm relatively new to open source, so would appreciate any constructive feedback/criticisms! 

 - [ ] Closes #xxxx
 - [ ] Tests added / passed

^ these will come later!

### DIFF
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -663,7 +663,7 @@ def maybe_encode_bools(var):
     return var
 
 
-def _infer_dtype(array):
+def _infer_dtype(array, allow_object=False):
     """Given an object array with no missing values, infer its dtype from its
     first element
     """
@@ -676,12 +676,15 @@ def _infer_dtype(array):
             # the length of their first element
             dtype = np.dtype(dtype.kind)
         elif dtype.kind == 'O':
-            raise ValueError('unable to infer dtype; xarray cannot '
-                             'serialize arbitrary Python objects')
+            if allow_object:
+                dtype = np.dtype(object)
+            else:
+                raise ValueError('unable to infer dtype; xarray cannot '
+                                 'serialize arbitrary Python objects')
     return dtype
 
 
-def ensure_dtype_not_object(var):
+def maybe_infer_dtype(var, allow_object=False):
     # TODO: move this from conventions to backends? (it's not CF related)
     if var.dtype.kind == 'O':
         dims, data, attrs, encoding = _var_as_tuple(var)
@@ -689,7 +692,8 @@ def ensure_dtype_not_object(var):
         if missing.any():
             # nb. this will fail for dask.array data
             non_missing_values = data[~missing]
-            inferred_dtype = _infer_dtype(non_missing_values)
+            inferred_dtype = _infer_dtype(non_missing_values,
+                                          allow_object=allow_object)
 
             if inferred_dtype.kind in ['S', 'U']:
                 # There is no safe bit-pattern for NA in typical binary string
@@ -706,12 +710,13 @@ def ensure_dtype_not_object(var):
             data = np.array(data, dtype=inferred_dtype, copy=True)
             data[missing] = fill_value
         else:
-            data = data.astype(dtype=_infer_dtype(data))
+            data = data.astype(dtype=_infer_dtype(data,
+                                                  allow_object=allow_object))
         var = Variable(dims, data, attrs, encoding)
     return var
 
 
-def encode_cf_variable(var, needs_copy=True, name=None):
+def encode_cf_variable(var, needs_copy=True, name=None, allow_object=False):
     """
     Converts an Variable into an Variable which follows some
     of the CF conventions:
@@ -725,6 +730,9 @@ def encode_cf_variable(var, needs_copy=True, name=None):
     ----------
     var : xarray.Variable
         A variable holding un-encoded data.
+    allow_object : bool
+        Whether to allow objects to pass the encoder or to throw an error if
+        this is attempted.
 
     Returns
     -------
@@ -738,7 +746,7 @@ def encode_cf_variable(var, needs_copy=True, name=None):
     var = maybe_encode_dtype(var, name)
     var = maybe_default_fill_value(var)
     var = maybe_encode_bools(var)
-    var = ensure_dtype_not_object(var)
+    var = maybe_infer_dtype(var, allow_object=allow_object)
     return var
 
 
@@ -1058,7 +1066,7 @@ def encode_dataset_coordinates(dataset):
                                non_dim_coord_names=non_dim_coord_names)
 
 
-def cf_encoder(variables, attributes):
+def cf_encoder(variables, attributes, allow_object=False):
     """
     A function which takes a dicts of variables and attributes
     and encodes them to conform to CF conventions as much
@@ -1075,6 +1083,9 @@ def cf_encoder(variables, attributes):
         A dictionary mapping from variable name to xarray.Variable
     attributes : dict
         A dictionary mapping from attribute name to value
+    allow_object : bool
+        Whether to allow objects to pass the encoder or to throw an error if
+        this is attempted.
 
     Returns
     -------
@@ -1085,6 +1096,7 @@ def cf_encoder(variables, attributes):
 
     See also: encode_cf_variable
     """
-    new_vars = OrderedDict((k, encode_cf_variable(v, name=k))
+    new_vars = OrderedDict((k, encode_cf_variable(v, name=k,
+                                                  allow_object=allow_object))
                            for k, v in iteritems(variables))
     return new_vars, attributes
--- a/xarray/core/utils.py
+++ b/xarray/core/utils.py
@@ -9,6 +9,7 @@ import itertools
 import re
 import warnings
 from collections import Mapping, MutableMapping, Iterable
+from six.moves import cPickle as pickle
 
 import numpy as np
 import pandas as pd
@@ -489,3 +490,13 @@ def ensure_us_time_resolution(val):
     elif np.issubdtype(val.dtype, np.timedelta64):
         val = val.astype('timedelta64[us]')
     return val
+
+
+@functools.partial(np.vectorize, otypes='O')
+def encode_pickle(obj):
+    return np.frombuffer(pickle.dumps(obj), dtype=np.uint8)
+
+
+@functools.partial(np.vectorize, otypes='O')
+def decode_pickle(obj):
+    return pickle.loads(obj.tostring())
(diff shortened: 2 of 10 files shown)


xarray/conventions.py
```diff
@@ -676,20 +676,24 @@ def _infer_dtype(array):
...
+def maybe_infer_dtype(var, allow_object=False):
     # TODO: move this from conventions to backends? (it's not CF related)
     if var.dtype.kind == 'O':
         dims, data, attrs, encoding = _var_as_tuple(var)
         missing = pd.isnull(data)
         if missing.any():
             # nb. this will fail for dask.array data
             non_missing_values = data[~missing]
-            inferred_dtype = _infer_dtype(non_missing_values)
+            inferred_dtype = _infer_dtype(non_missing_values,
```
shoyer: If `_infer_dtype` fails, we want to break out of this function and return the original `var`, not copy the data and put it on a new Variable object (which happens below).

shoyer:
Thanks for giving this a shot!

> I added allow_object kwarg (rather than allow_pickle, no reason to firmly attach pickle to the api, could use something else for other backends).

I'm having a hard time imagining any other serialization formats for serializing arbitrary Python objects. `pickle` is pretty standard, though we might switch the argument for `to_netcdf` to `pickle_protocol` to allow indicating the pickle version (which would default to `None`, for don't pickle).

> NetCDF4DataStore handles this independently from the cf_encoder/decoder. The dtype support made it hard to decouple, plus I think object serialization is a backend dependent issue.

Yes, this is a little tricky. The current design is not great here. Ideally, though, we would still keep all of the encoding/decoding logic separate from the datastores. I need to think about this a little more.

-----------------------

One other concern is how to represent this data on disk in netCDF/HDF5 variables. Ideally, we would have a format that could work -- at least in principle -- with `h5netcdf`/`h5py` as well as `netCDF4-python`.

Annoyingly, these libraries currently have incompatible dtype support:

- netCDF4-python supports variable length types with custom `name`. It does not support HDF5's opaque type, though the netCDF-C libraries do support opaque types, so adding them to netCDF4-Python would be relatively straightforward.
- h5py supports variable length types, but not their `name` field. It maps `np.void` to HDF5's opaque type, which would be a pretty sensible storage type if netCDF4-Python supported it.

So if we want something that works with both, we'll need to add some additional metadata field in the form of an attribute to indicate how do decoding. Maybe something like `_PickleProtocol`, which would store the version of the pickle protocol used to write the data?

------------------------

I have some inline comments I'll add below.

xarray/core/utils.py
```diff
@@ -9,6 +9,7 @@
 import re
 import warnings
 from collections import Mapping, MutableMapping, Iterable
+from six.moves import cPickle as pickle
```
richlewis42: Wow sorry, I actually forgot it wasn't a builtin, and I lazily didn't set up a proper dev environment.  Definitely a simple fix.

xarray/core/utils.py
```diff
@@ -489,3 +490,13 @@ def ensure_us_time_resolution(val):
     elif np.issubdtype(val.dtype, np.timedelta64):
         val = val.astype('timedelta64[us]')
     return val
+
+
+@functools.partial(np.vectorize, otypes='O')
+def encode_pickle(obj):
+    return np.frombuffer(pickle.dumps(obj), dtype=np.uint8)
```
richlewis42: I think six handles the protocol issue, which is why I didn't do anything here, but no six so we can handle that manually.  I don't know much about pickle 2 and 3 compatability (i.e. dump in 2, load in 3), perhaps that would be the nicest configuration to default to?

shoyer: The way that pickle works, any version of Python can load older pickles, but old versions of Python can't load newer pickles. So `protocol=2` is a maximally backwards compatible option, but misses out on any later pickle improvements.

richlewis42:
> Thanks for giving this a shot!

No problem, I really want this feature so I can use xarray for a cheminformatics library I'm working on! Hopefully we can work out the best way to do this whilst keeping everything as nice and organised as it is was before I touched the code...
 
> I'm having a hard time imagining any other serialization formats for serializing arbitrary Python objects. pickle is pretty standard, though we might switch the argument for to_netcdf to pickle_protocol to allow indicating the pickle version (which would default to None, for don't pickle).

I couldn't think of others either - it made sense as a keyword for `cf_encoder` where no pickling currently happens which is why I changed it originally. `allow_pickle` definitely makes sense, I've used `np.save` loads but didn't know that keyword existed.  Perhaps two kws, `allow_pickle` and `pickle_protocol`, so its more explicit, and so we can set a default protocol?

> Yes, this is a little tricky. The current design is not great here. Ideally, though, we would still keep all of the encoding/decoding logic separate from the datastores. I need to think about this a little more.

Yeah, it definitely isn't great, I wanted a working example and that's what I managed to do before sleeping!  I'll keep looking through the code to familiarize myself a bit more with it - I would be interested to see what you suggest!

----

> One other concern is how to represent this data on disk in netCDF/HDF5 variables. Ideally, we would have a format that could work -- at least in principle -- with h5netcdf/h5py as well as netCDF4-python.

Yeah, that would definitely be good.

> Annoyingly, these libraries currently have incompatible dtype support:
>
> - netCDF4-python supports variable length types with custom name. It does not support HDF5's opaque type, though the netCDF-C libraries do support opaque types, so adding them to netCDF4-Python would be relatively straightforward.
> - h5py supports variable length types, but not their name field. It maps np.void to HDF5's opaque type, which would be a pretty sensible storage type if netCDF4-Python supported it.
So if we want something that works with both, we'll need to add some additional metadata field in the form of an attribute to indicate how do decoding. Maybe something like _PickleProtocol, which would store the version of the pickle protocol used to write the data?

The np.void type was pretty much designed for this sort of thing from what I can see, I was pretty surprised that netCDF4-python didn't have something similar, hence the strange `np.uint8` stuff.  We could do the same for h5py, using `h5py.special_dtype`, although as you say there is no name so it can't work the same for h5py as it does now.  Pickle works out the protocol automatically (no `protocol` keyword for `load` or `loads`), so we wouldn't really need to save the protocol as an attribute, although it would be a way to work out which variables to unpickle once saved, if we went this route.  It seems a shame not to use np.void though, so perhaps it makes sense to add the opaque types to netCDF4-python and forget the `np.uint8` trick.

richlewis42:
The pickle protocol is the second byte of any pickle btw, just had a look.

shoyer:
> Pickle works out the protocol automatically (no protocol keyword for load or loads), so we wouldn't really need to save the protocol as an attribute, although it would be a way to work out which variables to unpickle once saved, if we went this route.

I think we do want some sort of marker attribute, but I agree that it doesn't need to include the pickle version.

Maybe the attribute `_FileFormat = 'python-pickle'` would make sense? This would have the advantage of being obvious to anyone inspecting the netCDF file with standard tools (not xarray).

> It seems a shame not to use np.void though, so perhaps it makes sense to add the opaque types to netCDF4-python and forget the np.uint8 trick.

I think netCDF actually maps `np.int8` -> `NC_BYTE`, so that's at least some justification for this choice: http://www.unidata.ucar.edu/software/netcdf/docs/data_type.html

Certainly handling opaque types in netCDF4-python would be nice, though I don't think it should be a blocker for this. I suspect the reason this isn't done is that NumPy maps `bytes` -> `np.string_` even on Python 3. Thus `np.void` is used far less often than it should. Also the `repr` for `np.void` has been pretty poor, though that's being worked on currently in https://github.com/numpy/numpy/pull/8981.

shoyer:
How about something like the following:

In `encode_cf_variable`, create a new variable with pickle encoded data (if appropriate). This looks something like:
```python
def encode_cf_variable(var, allow_pickle=False):
    ...
    if var.dtype == object:
        if allow_pickle:
            var = maybe_encode_pickle(var)
        else:
            raise TypeError
    return var

def maybe_encode_pickle(var):
    if var.dtype == object:
        attrs = var.attrs.copy()
        safe_setitem('_FileFormat', 'python-pickle')
        protocol = var.encoding.pop('pickle_protocol', 2)
        data = utils.encode_pickle(var.values, protocol=protocol)
        var = Variable(var.dims, data, attrs, var.encoding)
    return var
```
This reuses the `encoding` parameter for setting the pickle protocol version, which is already what we use for similar variable specific encoding details.

In the netCDF backends, add a check for variable with `dtype == object` with a `_FileFormat` attribute. If this is the case, call a `create_vlen_int8_dtype` method to create an appropriate dtype using backend specific methods (The behavior on the base class should raise an error), and proceed with writing the data in the usual way.

For decoding, reverse the process. Convert custom vlen dtypes to `dtype=object` in the appropriate backend specific array wrapper type, but don't decode data. If `allow_pickle=True` and `var.attrs['_FileFormat'] == 'python-pickle'`, then `decode_cf_variable` should do the unpickling, moving `_FileFormat` from `attrs` to `encoding`.

For bonus points, generalize handling of vlen types with `encoding`. Something like `encoding={'dtype': {'vlen': np.int8}}` could indicate that a special vlen with `np.int8` data should be created to encode this variable's data. (This is inspired by h5py's [API for special types](http://docs.h5py.org/en/latest/special.html).)

richlewis42:
Sounds great, thanks for the feedback!  I'll probably not get to this until the weekend, but I'll have another crack at it then.

richlewis42:
Sorry for the holdup. I made progress 3 weekends ago, but didn't get a fully working example, and I've been swamped the whole of this month. This weekend I'll be free to finish this off.

xarray/core/utils.py
jhamman: this try/except statement should probably go in xarray's pycompat module.

xarray/core/utils.py
jhamman: let's set `HIGHEST_PROTOCOL` in pycompat as well.

richlewis42:
Hi all, sorry for the lack of communication on this. I'm writing up my PhD thesis at the moment, and my deadline is increasingly looming.  Once I've handed in I'll finish this off, and am more than happy to help with several other things I've encountered (I'll create the issues now). Once again, sorry for the delay!

PR #1422, opened 2017-05-23
xarray.core.variable.as_variable part of the public API

### LINKED ISSUE #1303: `xarray.core.variable.as_variable()` part of the public API?
Is it safe to use `xarray.core.variable.as_variable()` externally? I guess that currently it is not.

I have a specific use case where this would be very useful.

I'm working on a package that heavily uses and extends xarray for landscape evolution modeling, and inside a custom class for model parameters I want to be able to create `xarray.Variable` objects on the fly from any provided object, e.g., a scalar value, an array-like, a `(dims, data[, attrs])` tuple, another `xarray.Variable`, a `xarray.DataArray`... exactly what `xarray.core.variable.as_variable()` does.

Although I know that `Variable` objects are not needed in most use cases, in this specific case a clean solution would be the following

```python
import xarray as xr

class Parameter(object):

    def to_variable(self, obj):
        return xr.as_variable(obj)
        # ... some validation logic on, e.g., data type, value bounds, dimensions...
        # ... add default attributes to the created variable (e.g., units, description...)
```

I don't think it is a viable option to copy `as_variable()` and all its dependent code in my package as it seems to have quite a lot of logic implemented.

A workaround using only public API would be something like:

```python
class Parameter(object):

    def to_variable(self, obj):
        return xr.Dataset(data_vars={'v': obj}).variables['v']
```

but it feels a bit hacky.

benbovy:
- [x] Closes #1303 
 - [x] Tests added / passed
 - [x] Passes ``git diff upstream/master | flake8 --diff`` (if we ignore messages for .rst files and "imported but not used" messages for `xarray.__init__.py`)

Make `xarray.core.variable.as_variable` part of the public API and accessible as a top-level function: `xarray.as_variable`.

I changed the docstrings to follow the numpydoc format more closely.

I also removed the `copy=False` keyword arguments as apparently it was unused.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -22,6 +22,7 @@ Top-level functions
    full_like
    zeros_like
    ones_like
+   as_variable
 
 Dataset
 =======
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -22,7 +22,6 @@ Top-level functions
    full_like
    zeros_like
    ones_like
-   as_variable
 
 Dataset
 =======
@@ -490,6 +489,7 @@ Advanced API
 
    Variable
    IndexVariable
+   as_variable
    register_dataset_accessor
    register_dataarray_accessor
 
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -65,7 +65,10 @@ def as_variable(obj, name=None):
         obj = obj.copy(deep=False)
     elif hasattr(obj, 'dims') and (hasattr(obj, 'data') or
                                    hasattr(obj, 'values')):
-        obj = Variable(obj.dims, getattr(obj, 'data', obj.values),
+        obj_data = getattr(obj, 'data', None)
+        if obj_data is None:
+            obj_data = getattr(obj, 'values')
+        obj = Variable(obj.dims, obj_data,
                        getattr(obj, 'attrs', None),
                        getattr(obj, 'encoding', None))
     elif isinstance(obj, tuple):
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -35,13 +35,19 @@ def as_variable(obj, name=None):
     ----------
     obj : object
         Object to convert into a Variable.
-        If the object is already a Variable, return a shallow copy.
-        Otherwise, if the object has 'dims' and 'data' attributes, convert
-        it into a new Variable.
-        If all else fails, attempt to convert the object into a Variable by
-        unpacking it into the arguments for creating a new Variable.
+
+        - If the object is already a Variable, return a shallow copy.
+        - Otherwise, if the object has 'dims' and 'data' attributes, convert
+          it into a new Variable.
+        - If all else fails, attempt to convert the object into a Variable by
+          unpacking it into the arguments for creating a new Variable.
     name : str, optional
-        Dimension name (only if data in `obj` is 1-dimensional).
+        If provided
+
+        - `obj` can be a 1D array, which is assumed to label coordinate values
+          along a dimension of this given name.
+        - Variables with name matching one of their dimensions are converted
+          into `IndexVariable` objects.
 
     Returns
     -------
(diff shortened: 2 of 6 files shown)


fmaussion:
Many tests fail because they expect to have a ``copy=`` kwarg somehow...

benbovy:
Oh yes I haven't run the tests locally before submitting this, I didn't expected that.

Mmm I don't see any current use of the `copy` keyword argument inside the `as_variable` function. [Here](https://github.com/pydata/xarray/blob/a943419b86c1c952d07cef6acf0b10ea5784a4cc/xarray/core/groupby.py#L542) is the only place where I found `copy` to be explicitly provided.  Can I safely remove it also there? If I remove it the tests pass again (although I got 3 remaining failing tests that don't seem to be related to this issue as they fail either with or without the `copy` argument).

xarray/core/variable.py
```diff
@@ -28,14 +28,26 @@
...
-    - If all else fails, attempt to convert the object into an `Variable` by
-      unpacking it into the arguments for `Variable.__init__`.
+def as_variable(obj, name=None):
+    """Convert an object into a Variable.
+
+    Parameters
+    ----------
+    obj : object
+        Object to convert into a Variable.
+        If the object is already a Variable, return a shallow copy.
```
shoyer: Can we keep these as bullet points for sphinx? I think that works for rendered docstrings if you're careful (e.g., see the [docstring for `xarray.Dataset.__init__`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.html#xarray.Dataset)).

xarray/core/variable.py
```diff
@@ -28,14 +28,26 @@
...
+        Otherwise, if the object has 'dims' and 'data' attributes, convert
+        it into a new Variable.
+        If all else fails, attempt to convert the object into a Variable by
+        unpacking it into the arguments for creating a new Variable.
+    name : str, optional
+        Dimension name (only if data in `obj` is 1-dimensional).
```
shoyer: I would say:

> If provided:
> - `obj` can be a 1D array, which is assumed to label coordinate values along a dimension of this given name.
> - Variables with name matching one of their dimensions are converted into `IndexVariable` objects.

shoyer:
> Mmm I don't see any current use of the copy keyword argument inside the as_variable function. Here is the only place where I found copy to be explicitly provided. Can I safely remove it also there?

Yes, we can safely remove the `copy` argument here. At point point, `as_variable` did not always return a new `Variable` object, but as you can see it always does that now.

doc/api.rst
```diff
@@ -22,6 +22,7 @@ Top-level functions
    full_like
    zeros_like
    ones_like
+   as_variable
```
shoyer: I would move this to the "Advanced API" section at the bottom, next to `Variable` and `IndexVariable`.

xarray/core/variable.py
```diff
@@ -28,14 +28,32 @@
...
+    obj : object
+        Object to convert into a Variable.
+
+        - If the object is already a Variable, return a shallow copy.
+        - Otherwise, if the object has 'dims' and 'data' attributes, convert
+          it into a new Variable.
+        - If all else fails, attempt to convert the object into a Variable by
+          unpacking it into the arguments for creating a new Variable.
+    name : str, optional
+        If provided
```
shoyer: add a final colon `:`

shoyer:
We have some tests for `as_variable`, but it looks a little thin for public API: https://github.com/pydata/xarray/blob/d80248476ebe4a3845211c5d58e0af1effc73ea3/xarray/tests/test_variable.py#L609

The remaining test failures should be fixed by https://github.com/pydata/xarray/pull/1423

benbovy:
I've tried adding tests for lines in `as_variable` that were not yet covered.

shoyer:
@benbovy thanks!

PR #1423, opened 2017-05-24
Fix errors in the test suite due to pytest warning changes

shoyer:
This is causing CI failures for some builds.

This is really a pytest bug (https://github.com/pytest-dev/pytest/issues/2430), but working around this on our end is easy enough (and cleans things up a little).

### DIFF
--- a/xarray/tests/__init__.py
+++ b/xarray/tests/__init__.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-import warnings
 from contextlib import contextmanager
 from distutils.version import LooseVersion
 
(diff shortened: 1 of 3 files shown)


xarray/tests/__init__.py
```diff
@@ -119,11 +118,10 @@ def assertItemsEqual(self, first, second, msg=None):
 
     @contextmanager
     def assertWarns(self, message):
-        with warnings.catch_warnings(record=True) as w:
-            warnings.filterwarnings('always', message)
+        with pytest.warns(Warning) as w:
             yield
-            assert len(w) > 0
-            assert any(message in str(wi.message) for wi in w)
+        assert len(w) > 0
```
gidden: is the change in indentation intentional? I would expect `w` to leave scope.

shoyer: Yes, this was intentional. Python `with` blocks don't define scopes for variables, so `w` is still defined outside the block. In this case, the `with` block only indicates where warnings should be caught.

gidden:
TIL, thanks!

PR #1427, opened 2017-05-26
Fix broken holoviews link

mivade:
This is a trivial fix to point to the current location of the HoloViews documentation.


shoyer:
Thanks!

PR #1451, opened 2017-06-13
inconsistent time.units fmt in encode_cf_datetime

### LINKED ISSUE #1449: time.units truncated when saving to_netcdf
When I manually specify the `units` attribute for time, and then save the Dataset `to_netcdf` the string is truncated. See exaple

    import pandas as pd
    import xarray as xr

    time = pd.date_range('2000-01-01', '2000-01-31', freq='6h')
    ds = xr.Dataset(coords=dict(time=time))

    units = 'days since 1975-01-01 00:00:00'
    calendar = 'gregorian'
    encoding=dict(time=dict(units=units, calendar=calendar))

    ds.to_netcdf('test.nc', format='NETCDF4_CLASSIC', encoding=encoding)

    ! ncdump -h test.nc
    # time:units = "days since 1975-01-01" ;

Some programs seem to require the hours to be present to interpret the time properly (e.g. panoply). When specifying the hour, a 'T' is added.

    units = 'days since 1975-01-01 01:00:00'

    ! ncdump -h test.nc
    # time:units = "days since 1975-01-01T01:00:00" ;

When xarray defines the `time.units` it works fine.

    ds = xr.Dataset(coords=dict(time=time))
    ds.to_netcdf('test.nc', format='NETCDF4_CLASSIC',)

    ! ncdump -h test.nc
    # time:units = "hours since 2000-01-01 00:00:00" ;

xarray version 0.9.6

mathause:
- do not change user-specified units
- always format infered units as 'YYYY-mmmm-ddTHH:MM:SS'

This is my naïve approach.

 - [ ] Closes #1449
 - [ ] Tests added / passed

### DIFF
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -205,7 +205,9 @@ def infer_datetime_units(dates):
     unique_timedeltas = np.unique(np.diff(dates))
     units = _infer_time_units_from_diff(unique_timedeltas)
     reference_date = dates[0] if len(dates) > 0 else '1970-01-01'
-    return '%s since %s' % (units, pd.Timestamp(reference_date))
+    # always format as 'YYYY-mmmm-ddTHH:MM:SS'
+    reference_date = format_timestamp(reference_date, shorten=False)
+    return '%s since %s' % (units, reference_date)
 
 
 def infer_timedelta_units(deltas):
@@ -228,16 +230,6 @@ def nctime_to_nptime(times):
         dt = datetime(t.year, t.month, t.day, t.hour, t.minute, t.second)
         new[i] = np.datetime64(dt)
     return new
-
-
-def _cleanup_netcdf_time_units(units):
-    delta, ref_date = _unpack_netcdf_time_units(units)
-    try:
-        units = '%s since %s' % (delta, format_timestamp(ref_date))
-    except OutOfBoundsDatetime:
-        # don't worry about reifying the units if they're out of bounds
-        pass
-    return units
 
 
 def _encode_datetime_with_netcdf4(dates, units, calendar):
@@ -279,8 +271,6 @@ def encode_cf_datetime(dates, units=None, calendar=None):
 
     if units is None:
         units = infer_datetime_units(dates)
-    else:
-        units = _cleanup_netcdf_time_units(units)
 
     if calendar is None:
         calendar = 'proleptic_gregorian'
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -102,7 +102,7 @@ def last_item(x):
     return np.ravel(x[indexer]).tolist()
 
 
-def format_timestamp(t):
+def format_timestamp(t, shorten=True):
     """Cast given object to a Timestamp and return a nicely formatted string"""
     # Timestamp is only valid for 1678 to 2262
     try:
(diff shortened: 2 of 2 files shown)


xarray/conventions.py
```diff
@@ -205,7 +205,9 @@ def infer_datetime_units(dates):
     unique_timedeltas = np.unique(np.diff(dates))
     units = _infer_time_units_from_diff(unique_timedeltas)
     reference_date = dates[0] if len(dates) > 0 else '1970-01-01'
-    return '%s since %s' % (units, pd.Timestamp(reference_date))
+    # always format as 'YYYY-mmmm-ddTHH:MM:SS'
```
mathause: the CF conventions do not specify the use of the `T`, so not sure which convention should take preference (CF vs ISO)

xarray/conventions.py
```diff
@@ -279,8 +271,6 @@ def encode_cf_datetime(dates, units=None, calendar=None):
 
     if units is None:
         units = infer_datetime_units(dates)
-    else:
```
mathause: my approach would be not to change the user-specified units string, as this can be surprising. But there might have been a reason for this...

xarray/conventions.py
shoyer: This is for writing netCDF files, so CF should be prefered.

It looks like CF conventions avoid the `T`, so we should probably stick with that -- even though it's not strict ISO-8601.

xarray/conventions.py
shoyer: I agree with you that we shouldn't change units that are user specified, and I'm no longer sure what I was thinking here. Unfortunately the original commit (https://github.com/pydata/xarray/commit/14466e68f1ccdbaf428833aec04d8067379365a4) doesn't really help either.

If our test suite passes, I think removing should be fine.

mathause:
I tried to dig a bit into this...

The reason behind `_cleanup_netcdf_time_units` is to clean up things as `u'Hour␣␣since  1680-01-01 00:00:00'` (double space). However, it does not correct for the case (e.g. daYs remains daYs). Also, the clean up is not [tested](https://github.com/pydata/xarray/blob/master/xarray/tests/test_conventions.py#L130). This seems to be a reasonable thing to do, but, as mentioned the ref_date is changed, too. Thus we have to decide if

* we want to keep the cleaning of the "Hour since" part, and if the case should be corrected as well (daYs). If yes potentially add tests.
* the user-defined ref_date string (e.g. 2015-01-01) should be changed (I would say no). And if no, if it needs a sanity check or if pd.Timestamp(ref_date) does the sanity check for us.

In the [CF conventions](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#time-coordinate) they refer to the udunits library. And, as mentioned, provide examples without "T". But this seems to be the standard in udunits, version 1. While in the [udunits, version 2](http://www.unidata.ucar.edu/software/udunits/udunits-2.2.25/doc/udunits/udunits2lib.html#UDUNITS-Library), the timestamp format includes "T".

TIMSTAMP:
        `<year> (<month> <day>?)? "T" <hour> (<minute> <second>?)?`

But udunits2 also mentions that it should be able to parse all formats of udunits1. So leaving out the "T" seems to be the safer option.

dcherian:
@mathause can you pick this up again? It looks like it's pretty close.

mathause:
I have a feeling that there where quite some changes to this part of the code (but I'm not sure) - so I'd have to look into it first.

mathause:
Ok, I just had a look & things still look pretty much the same. Not sure if/ when I find time to do this,

keewis:
@mathause, this has been unintentionally closed while renaming the `main` branch. Feel free to reopen (you'd probably have to restore the branch on your repo first) if you think this is still valuable.

PR #1453, opened 2017-06-14
Automate interpretation of _Unsigned attribute

### LINKED ISSUE #1444: Best practice when the _Unsigned attribute is present in NetCDF files
Some (large) data providers are writing NetCDF-4-extended files but using an `_Unsigned` attribute to indicate that a signed data type should be interpreted as unsigned bytes. 

Background: https://github.com/Unidata/netcdf4-python/issues/656

From the background discussion above, it is my understanding that xarray does not honor the attribute because it’s not a part of the CF spec, is only mentioned as a proposed attribute in the [NetCDF Best Practices](http://www.unidata.ucar.edu/software/netcdf/docs/BestPractices.html
), and because "xarray wants the `Variable` dtype to be the same as the dtype of the data returned."

Taking the above as a given, it is necessary for xarray users encountering such variables to do the following after reading the data:

```dtype = data.encoding['dtype'].str.replace('i', 'u')
scale_factor = data.encoding['scale_factor']
add_offset = data.encoding['add_offset']
unscale = ((data - add_offset)/scale_factor).data.astype(dtype).astype('float64')
fixed = unscale * scale_factor + add_offset
```

The un-scaling step can be saved by turning off auto mask and scale.

In order to automate the above process while still being able to use the functionality of `Dataset`, one approach might be to automatically perform the above steps on some known list of variables, and then reassign those variables to the `Dataset`. The downside is the need to read all variables up front, which could be expensive when processing large datasets where not all variables are needed. 

Is there another approach that would preserve lazy data loading, for instance by providing pre/post hooks for transformation functions at the  `__getitem__` stage? Is there something I could do to help document that as a best practice?

deeplycloudy:
- [x] Closes #1444 
 - [x] Tests added / passed

### DIFF
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -852,13 +852,14 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
                           RuntimeWarning, stacklevel=3)
         scale_factor = pop_to(attributes, encoding, 'scale_factor')
         add_offset = pop_to(attributes, encoding, 'add_offset')
-        if ((fill_value is not None and not np.any(pd.isnull(fill_value))) or
-                scale_factor is not None or add_offset is not None):
+        has_fill = (fill_value is not None and 
+                    not np.any(pd.isnull(fill_value)))
+        if (has_fill or scale_factor is not None or add_offset is not None):
             if fill_value.dtype.kind in ['U', 'S']:
                 dtype = object
             else:
                 dtype = float
-            if is_unsigned is not None:
+            if (is_unsigned is not None) and has_fill:
                 # Need to convert the fill_value to unsigned, too
                 # According to the CF spec, the fill value is of the same
                 # type as its variable, i.e. its storage format on disk
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -777,7 +777,7 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
     mask_and_scale: bool
         Lazily scale (using scale_factor and add_offset) and mask
         (using _FillValue). If the _Unsigned attribute is present
-        treat integer arrays as unsigned. 
+        treat integer arrays as unsigned.
     decode_times : bool
         Decode cf times ('hours since 2000-01-01') to np.datetime64.
     decode_endianness : bool
@@ -833,7 +833,7 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
             else:
                 dtype = float
             data = MaskedAndScaledArray(data, fill_value, scale_factor,
-                                        add_offset, dtype, 
+                                        add_offset, dtype,
                                         is_unsigned=is_unsigned)
 
     if decode_times and 'units' in attributes:
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -49,6 +49,8 @@ def mask_and_scale(array, fill_value=None, scale_factor=None, add_offset=None,
     add_offset : number, optional
         After applying scale_factor, add this number to entries in the
         original array.
+    is_unsigned : prior to applying fill, scale, or offset, treat
+        integer array as unsigned
 
     Returns
     -------
@@ -356,6 +358,8 @@ class MaskedAndScaledArray(utils.NDArrayMixin):
         add_offset : number, optional
             After applying scale_factor, add this number to entries in the
             original array.
+         is_unsigned : prior to applying fill, scale, or offset, treat
+             integer array as unsigned
         """
         self.array = array
         self.fill_value = fill_value
@@ -770,7 +774,8 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
         example: ['h', 'e', 'l', 'l', 'o'] -> 'hello'
     mask_and_scale: bool
         Lazily scale (using scale_factor and add_offset) and mask
-        (using _FillValue).
+        (using _FillValue). If the _Unsigned attribute is present
+        treat integer arrays as unsigned. 
     decode_times : bool
         Decode cf times ('hours since 2000-01-01') to np.datetime64.
     decode_endianness : bool
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -26,7 +26,7 @@ _STANDARD_CALENDARS = set(['standard', 'gregorian', 'proleptic_gregorian'])
 
 
 def mask_and_scale(array, fill_value=None, scale_factor=None, add_offset=None,
-                   dtype=float, is_unsigned=False):
+                   dtype=float, is_unsigned=None):
     """Scale and mask array values according to CF conventions for packed and
     missing values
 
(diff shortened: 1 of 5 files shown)


deeplycloudy:
In addition to the included (basic) test I've also tested this with the real-world data that motivated the PR and #1444. While it's a working draft, I'd welcome comments on the basic approach and appropriateness of the test coverage.

shoyer:
Instead of putting this alongside the `mask_and_scale` logic, can you make a separate class to do the dtype fixing in `decode_cf_variable`? Take a look at `BoolTypeArray` for an example.

deeplycloudy:
I've created a new `UnsignedIntTypeArray` and have separated the logic from `mask_and_scale`. Lint has been removed and docs updated.

deeplycloudy:
The CI fail is for 2.7/cdat/pynio in a couple of my new tests. In one, the fill value is not being applied, while in the other the unsigned conversion isn't happening. Are there any known differences in that cdat/pynio stack that would cause these to fail while others pass?

deeplycloudy:
Tests now pass after I realized I wasn't converting the _FillValue to unsigned. 

I also turned off PyNIO's internal support for masking, in keeping with the philosophy that `xarray` should only use the backends to retrieve the bytes as represented on disk.

Note that some of the CI builds are skipping most of their tests (e.g, py=3.4; you can tell by the run time). This is a problem in other PRs as well.

xarray/conventions.py
```diff
@@ -637,6 +665,13 @@ def maybe_encode_dtype(var, name=None):
                                   'any _FillValue to use for NaNs' % name,
                                   RuntimeWarning, stacklevel=3)
                 data = duck_array_ops.around(data)[...]
+                if '_Unsigned' in encoding:
```
jhamman: is `_Unsigned = False` a valid attribute?

If so, we need this to be `if encoding.get('_Unsigned', False)`.

xarray/tests/test_backends.py
```diff
@@ -360,24 +377,53 @@ def test_roundtrip_strings_with_fill_value(self):
...
+            self.assertDatasetAllClose(encoded, actual)
+        with self.roundtrip(encoded,
+                            open_kwargs=dict(decode_cf=False)) as actual:
+            self.assertDatasetAllClose(encoded, actual)
+        # make sure roundtrip encoding didn't change the
+        # original dataset.
+        self.assertDatasetIdentical(encoded,
+                                    create_encoded_unsigned_masked_scaled_data(
+                                    )
+                                    )
```
jhamman: let's pep8 this:

```Python
        self.assertDatasetIdentical(
            encoded, create_encoded_unsigned_masked_scaled_data())

```

xarray/tests/test_backends.py
```diff
@@ -360,24 +377,53 @@ def test_roundtrip_strings_with_fill_value(self):
...
-        with self.roundtrip(encoded, open_kwargs=dict(decode_cf=False)) as actual:
+        with self.roundtrip(encoded,
+                            open_kwargs=dict(decode_cf=False)) as actual:
             self.assertDatasetAllClose(encoded, actual)
         # make sure roundtrip encoding didn't change the
         # original dataset.
         self.assertDatasetIdentical(encoded,
-                                    create_encoded_masked_and_scaled_data())
+                                    create_encoded_masked_and_scaled_data()
+                                    )
```
jhamman: pep8 / move last `)`. This whole call should actually fit on one line.

xarray/conventions.py
deeplycloudy: From the documentation I assembled as part of [the same issue in NetCDF4-python](https://github.com/Unidata/netcdf4-python/issues/656), it looks like `_Unsigned = "true" is present if the data are `_Unsigned`, with no mention of `_Unsigned = "false"`. I haven't seen an example of `False` in the wild, but, your suggestion is consistent with the [patch to NetCDF4-python](https://github.com/Unidata/netcdf4-python/pull/658/files) for this issue.

I made the change you recommended on the encode side.

On decode there is a corresponding check of the attribute, so I made a change there, too.

xarray/tests/test_backends.py
deeplycloudy: Done. That's much better.

xarray/tests/test_backends.py
deeplycloudy: Done.

shoyer:
Sounds good, I'll take another look shortly.

xarray/conventions.py
```diff
@@ -637,6 +665,13 @@ def maybe_encode_dtype(var, name=None):
                                   'any _FillValue to use for NaNs' % name,
                                   RuntimeWarning, stacklevel=3)
                 data = duck_array_ops.around(data)[...]
+                if encoding.get('_Unsigned', False):
+                    unsigned_dtype = 'i%s' % dtype.itemsize
+                    old_fill = np.asarray(attrs['_FillValue'])
```
shoyer: This block should be guarded in a check to verify that `_FillValue` is actually defined as an attribute.

xarray/conventions.py
```diff
@@ -786,6 +822,16 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
             dimensions = dimensions[:-1]
             data = CharToStringArray(data)
 
+    pop_to(attributes, encoding, '_Unsigned')
+    is_unsigned = encoding.get('_Unsigned', False)
+    if (is_unsigned) and (mask_and_scale is True):
```
shoyer: You don't need the extra parentheses here, and use implicit boolean checks instead of `is True`. So this should be: `if is_unsigned and mask_and_scale`

xarray/conventions.py
```diff
@@ -800,20 +846,25 @@ def decode_cf_variable(var, concat_characters=True, mask_and_scale=True,
...
+        if (has_fill or scale_factor is not None or add_offset is not None):
             if fill_value.dtype.kind in ['U', 'S']:
                 dtype = object
             else:
                 dtype = float
+            if is_unsigned and has_fill:
+                # Need to convert the fill_value to unsigned, too
+                # According to the CF spec, the fill value is of the same
+                # type as its variable, i.e. its storage format on disk
+                fill_value = np.asarray(fill_value, dtype=data.unsigned_dtype)
```
shoyer: Can we simply cast `fill_value` to the dtype of `data` unilaterally here? e.g., `fill_value = np.asarray(fill_value, dtype=data.dtype)` without the `if is_unsigned and has_fill` check?

That seems a little more robust to me.

xarray/tests/test_backends.py
```diff
@@ -360,24 +377,50 @@ def test_roundtrip_strings_with_fill_value(self):
...
             self.assertDatasetIdentical(expected, actual)
 
+    def test_unsigned_roundtrip_mask_and_scale(self):
+        decoded = create_unsigned_masked_scaled_data()
+        encoded = create_encoded_unsigned_masked_scaled_data()
+        with self.roundtrip(decoded) as actual:
+            self.assertDatasetAllClose(decoded, actual)
+        with self.roundtrip(decoded,
+                            open_kwargs=dict(decode_cf=False)) as actual:
+            # TODO: this assumes that all roundtrips will first
```
shoyer: please remove this redundant TODO note

xarray/tests/test_backends.py
```diff
@@ -360,24 +377,50 @@ def test_roundtrip_strings_with_fill_value(self):
...
+            # encode.  Is that something we want to test for?
+            self.assertDatasetAllClose(encoded, actual)
```
shoyer: I'm not sure `assertDatasetAllClose` checks dtypes. It would be good to check explicitly here.

xarray/conventions.py
deeplycloudy: Done.

xarray/conventions.py
deeplycloudy: Done. I had to leave the `has_fill` check because `fill_value` will be `None` in cases where there  is no `fill_value` attribute.

xarray/tests/test_backends.py
deeplycloudy: Done.

xarray/tests/test_backends.py
deeplycloudy: Done.

shoyer:
Thanks @deeplycloudy !

PR #1454, opened 2017-06-14
change NotImplemented to NotImplementedError for h5netcdf autoclose=True

MareinK:
Solves this error:

```
TypeError: 'NotImplementedType' object is not callable
```


shoyer:
Thanks!

PR #1476, opened 2017-07-12
Fix text in error message, A leftover from #993

mzuehlke:
I've spotted this wrong error message. 
In my opinion this is too small to justify an entry in `whats-new.rst`.


jhamman:
Thanks for the PR. It looks fine to me. The failing tests do not seem to be related but I'll let @shoyer check on that before we merge.

shoyer:
Yes, we're good here. The test failures are due to upstream pandas bug: https://github.com/pandas-dev/pandas/issues/16896

shoyer:
Thanks!

PR #1478, opened 2017-07-13
Fixes dataset rename bug (GH1477)

### LINKED ISSUE #1477: Dataset.rename shouldn't complain when the name-dict re-uses column names.
Consider what happens when somebody does ds.rename({'a': 'b', 'b': 'a'}). The current rename code will throw an error even though this operation is completely well defined and the current rename implementation actually handles this case correctly (except for the bad sanity check).

What this function should really check for is the there are no name collisions in the resulting dataset.

Currently we have the following code (in: [dataset.py:1527](https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L1527)): `if v in self and k != v`

I propose that we replace it with `if name in variables` just before line 1541 with an appropriate throw.

There is one problem that this can cause: right now, it is possible to overwrite variables silently. For example ds.rename({'a': 'd', 'b': 'd'}) will work and silently drop one of the columns (assuming 'd' doesn't exist in ds). My proposed changes will make the rename function throw in this case. I think the new behavior is "more correct" especially since the column dropped is pretty much random if the name_dict is a normal dictionary and the person can always use a del or something. However it is not backwards compatible which might be a concern.

Thoughts?

Thanks.
-- PG

PS. xarray is great. Thanks for writing it.

newt0311:
- [x] Closes #1477 
 - [x] Tests added / passed
 - [x] added whats-new.rst entry.

### DIFF
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -1485,7 +1485,7 @@ class TestDataset(TestCase):
             renamed['renamed_var1'].values
 
     def test_rename_old_name(self):
-        #regtest for GH1477
+        # regtest for GH1477
         data = create_test_data()
 
         with self.assertRaisesRegexp(ValueError, "'samecol' conflicts"):
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -1474,7 +1474,7 @@ class TestDataset(TestCase):
         with self.assertRaisesRegexp(ValueError, "cannot rename 'not_a_var'"):
             data.rename({'not_a_var': 'nada'})
 
-        with self.assertRaisesRegexp(ValueError, "'var1' already exists"):
+        with self.assertRaisesRegexp(ValueError, "'var1' conflicts"):
             data.rename({'var2': 'var1'})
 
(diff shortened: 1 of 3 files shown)


xarray/tests/test_dataset.py
```diff
@@ -1484,6 +1484,16 @@ def test_rename(self):
...
 
+    def test_rename_old_name(self):
+        # regtest for GH1477
+        data = create_test_data()
+
+        with self.assertRaisesRegexp(ValueError, "'samecol' conflicts"):
+            data.rename({'var1': 'samecol', 'var2': 'samecol'})
+
+        # This shouldn't cause any problems.
+        data.rename({'var1': 'var2', 'var2': 'var1'})
```
jhamman: would this raise if there is already a `var2` in the dataset? We should also test for that.

newt0311: Doesn't the original test check for that at line 1477?

-- PG.

jhamman: ah, yes it seems to.

shoyer:
This looks great, thank you!

PR #1479, opened 2017-07-14
Fix test suite failure in TestDataset.test_sel

shoyer:
This is a temporary work around for https://github.com/pandas-dev/pandas/issues/16896, which was introduced by pandas 0.20.3

We can safely revert it after the next pandas release.


jhamman:
@shoyer - lgtm. Do we want to add a note to revert this in the code itself? I'm not sure we actually need to revert this though since both are valid syntactically.

shoyer:
Yeah, this probably isn't even worth reverting.

PR #1485, opened 2017-07-21
add ISSUE_TEMPLATE for github and xr.show_versions()

jhamman:
- [x] xref #986
 - [x] Tests added / passed

This PR adds a new module level function `xr.show_versions()` and a new Github Issue Template to help with debugging user issues. Hopefully, we stop having to ask "what version of dask are you using?" Functionality is copied from [pandas](https://github.com/pandas-dev/pandas/blob/e99f56e43ab65710a34440067efe372d5f74280b/pandas/util/_print_versions.py).


fmaussion:
This is a good idea! 

But it is going to be a mess until a new version of xarray is released, since users won't have access to the utility function...

jhamman:
> But it is going to be a mess until a new version of xarray is released, since users won't have access to the utility function...

Good point. I guess we could hold off on merging for a bit then.

xarray/util/print_versions.py
```diff
@@ -0,0 +1,151 @@
...
+
+def show_versions(as_json=False):
+    sys_info = get_sys_info()
+
+    deps = [
+        # (MODULE_NAME, f(mod) -> mod version)
+        ("xarray", lambda mod: mod.__version__),
+        ("pandas", lambda mod: mod.__version__),
+        ("numpy", lambda mod: mod.version.version),
+        ("scipy", lambda mod: mod.version.version),
```
shoyer: Can we just use `mod.__version__` for numpy and scipy? That's the more standard way to access this information.

jhamman: yes, that works. Done.

jhamman:
Let's just wait until we have our next release.

Issue #1462, opened 2017-06-22
Dataset.to_dataframe loads dask arrays into memory


Zac-HD:
`to_dataframe` should return a Dask Dataframe, instead of eagerly loading data.  This is probably pretty easy to implement ([thanks to dask](http://dask.pydata.org/en/latest/dataframe-api.html#dask.dataframe.from_dask_array)), but will require some care to ensure that no intermediate results (or indices!) are loaded.  We should also check the `to_series` method.

jmunroe:
Today, I find myself in need exact functionality.  Assuming no one else is working on it, I'll give a shot at trying to fix this.

shoyer:
Closing this in favor of https://github.com/pydata/xarray/issues/1093

PR #1489, opened 2017-07-26
lazily load dask arrays to dask data frames by calling to_dask_dataframe 

### LINKED ISSUE #1093: Convert xarray dataset to dask dataframe or delayed objects
It would be great to have a function like dask's to_delayed in order to take xarray datasets and convert them to pandas dataframes chunkily.

http://stackoverflow.com/questions/40475884/how-to-convert-an-xarray-dataset-to-pandas-dataframes-inside-a-dask-dataframe

jmunroe:
- [x] Closes #1093
 - [x] Tests added / passed

Working towards on a solution for #1462 

Just some stub code for the moment.

Dask dataframes don't appear to support MultiIndex so not sure what to do about that.


shoyer:
Given that dask dataframes don't support MultiIndexes (among many other features), I have a hard time seeing them as a drop-in replacement for `pandas.DataFrame`. So maybe it would make sense to make this a separate method, e.g., `to_dask_dataframe()`?

We could also use a new method as an opportunity to slightly change the API, by not setting an index automatically. This lets us handle N-dimensional data while side-stepping the issue of MultiIndex support -- I don't think this would be very useful when limited to 1D arrays, and dask MultiIndex support seems to be a ways away (https://github.com/dask/dask/issues/1493). Also, `set_index()` in dask shuffles data, so it can be somewhat expensive.

jmunroe:
After working on this for a little while, I agree that this really should be a to_dask_dataframe() method. I'll make that change.

xarray/core/dataset.py
```diff
@@ -2384,6 +2384,34 @@ def from_dataframe(cls, dataframe):
...
+        this dataset's coordinate. Since dask DataFrames to not support
+        multi-indexes, this only works if there is one coordinate dimension.
+        """
+
+        import dask.dataframe as dd
+        import dask.array as da
+
+        columns = [k for k in self if k not in self.dims]
+
+        data = [self._variables[k].data.reshape(-1) for k in columns]
```
shoyer: You need to use `Variable.set_dims()` like in the `to_dataframe()` example. Otherwise this breaks when different variables have different shapes.

xarray/core/dataset.py
```diff
@@ -2384,6 +2384,34 @@ def from_dataframe(cls, dataframe):
...
+        df = dd.from_dask_array(da.stack(data, axis=1), columns=columns)
```
shoyer: Stacking together variables into a single dask array loses dtype information, which is not ideal. Can you look into constructing a dask dataframe from a list of dask arrays?

xarray/core/dataset.py
```diff
@@ -2384,6 +2384,34 @@ def from_dataframe(cls, dataframe):
             obj[name] = (dims, data)
         return obj
 
+    def to_dask_dataframe(self, set_index=True):
```
shoyer: I don't think we want to default to `set_index=True`, since xarray focuses on handling N-dimensional data, so data known ahead of time to be 1D is somewhat rare.

xarray/core/dataset.py
```diff
@@ -2384,6 +2384,34 @@ def from_dataframe(cls, dataframe):
...
+        columns = [k for k in self if k not in self.dims]
+
+        data = [self._variables[k].data.reshape(-1) for k in columns]
+        df = dd.from_dask_array(da.stack(data, axis=1), columns=columns)
+
+        if set_index:
+            index = self.coords.to_index(self.dims)
+
+            index = dd.from_array(index.values).repartition(divisions=df.divisions)
+            df = df.set_index(index, sort=False)
```
shoyer: The index variable should already have been added as a column, so rather than converting it into a dask array again, I would simply handle it by name.

xarray/tests/test_dask.py
```diff
@@ -393,6 +394,28 @@ def test_from_dask_variable(self):
...
         self.assertLazyAndIdentical(self.lazy_array, a)
 
+    def test_to_dask_dataframe(self):
+        # adapted from TestDataset::test_to_and_from_dataframe()
+        # but with dask DataFrames instead of pandas DataFrames
+
+        x = da.from_array(np.random.randn(10), chunks=4)
+        y = np.random.randn(10)
+        t = list('abcdefghij')
+        ds = Dataset(OrderedDict([('a', ('t', x)),
```
shoyer: We need some test cases with non-1D data.

xarray/core/dataset.py
jmunroe: Of course, that makes sense.  set_index=False is the default (index is simply an integer as would be from calling df.reset_index() )

xarray/core/dataset.py
jmunroe: I'd added that ... but I am not entirely clear if what I did was sufficient.

xarray/core/dataset.py
jmunroe: I couldn't find a simple way of doing this by wrapping something already in the dask api.  So, instead, I am constructing the dask graph directly following the pattern of the other dask dataframe constructors (like dd.from_dask_array() ).

xarray/tests/test_dask.py
jmunroe: Added a 2-d test case.

xarray/core/dataset.py
jmunroe: For the 1-d case, if the index/coordinate variable is a normal array, it needs to be partitioned into divisions consistent with the chunking pattern for the underlying dask array.  

I still need to handle the case where the coordinate variable itself is already a dask array.

jmunroe: coordinate variable can now be a dask array

xarray/core/dataset.py
shoyer: Great, thanks!

So far, we have tried to avoid adding code that actually constructs dask graphs to xarray proper as it's easier to maintain in dask. Can you check with the dask developers to see if they would be interested in adding this functionality (i.e., DataFrame from list of dask arrays) upstream in some form? I suspect they would be!

shoyer:
@jmunroe This is great functionality -- thanks for your work on this!

One concern: if possible, I would like to avoid adding explicit dask graph building code in xarray. It looks like the canonical way to transform from a list of dask/numpy arrays to a dask dataframe is to make use of `dask.dataframe.from_array` along with `dask.dataframe.concat`:
```
In [34]: import numpy as np

In [35]: import dask.dataframe as dd

In [36]: import dask.array as da

In [37]: x = da.from_array(np.arange(5), 2)

In [38]: y = da.from_array(np.linspace(-np.pi, np.pi, 5), 2)

# notice that dtype is preserved properly
In [39]: dd.concat([dd.from_array(x), dd.from_array(y)], axis=1)
Out[39]:
Dask DataFrame Structure:
                   0        1
npartitions=2
0              int64  float64
2                ...      ...
4                ...      ...
Dask Name: concat-indexed, 26 tasks
```

Can you look into refactoring your code to make use of these?

jmunroe:
I agree that using dask.dataframe.from_array and dask.dataframe.concat should work. Sorry I haven't had a chance to get back to this recently. I'll try to make the change early next week.

jmunroe:
Sorry for the delay. I think this task is now complete.

xarray/core/dataset.py
```diff
@@ -2384,6 +2384,57 @@ def from_dataframe(cls, dataframe):
...
+
+        If set_index=True, the dask DataFrame is indexed by
+        this dataset's coordinate. Since dask DataFrames to not support
+        multi-indexes, this only works if there is one coordinate dimension.
+        """
+
+        import dask.dataframe as dd
+        import dask.array as da
+
+        columns = [k for k in self if k not in self.dims]
```
shoyer: Why skip variables that are in `self.dims`? I think we want those to be part of the resulting dask dataframe, too -- even though we don't want to set them as an actual index.

jhamman:
@jmunroe - 

I added the PR checklist back to the top of this issue. The most pressing to-do item is getting some documentation written for this.

- The method will need to be added to `api.rst`
- We need a note briefly describing this feature in `whats-new.rst`
- We'll want to show an example of how this method can be used (either in the working with pandas or the dask doc sections)

xarray/core/dataset.py
jmunroe: For multiple dimensions, there wouldn't be a one-to-one correspondence between the coordinate and the data (different sizes).

shoyer: If we call `set_dims()` then all the coordinates also become the same size as the data.

There isn't unique mapping for each coordinate independently, but when you combine them together they unique identify each row (at least assuming the coordinates have unique values to begin with).

More concretely, consider something like the following example:
```
>>> ds = xr.Dataset({'z': (('x', 'y'), [[1, 2], [3, 4]])}, {'x': [1, 2], 'y': ['a', 'b']})

>>> ds
<xarray.Dataset>
Dimensions:  (x: 2, y: 2)
Coordinates:
  * x        (x) int64 1 2
  * y        (y) |S1 'a' 'b'
Data variables:
    z        (x, y) int64 1 2 3 4

>>> ds.to_dataframe()
     z
x y   
1 a  1
  b  2
2 a  3
  b  4
```
Here, `x` and `y` are part of the index, but we would probably want them to be columns.

One interesting case it how to handle dimensions without coordinates. In this case, we probably want to fill in integer columns, again similar to how `to_dataframe()` already works:
```
>>> ds2 = xr.Dataset({'z': (('x', 'y'), [[1, 2], [3, 4]])}, {'y': ['a', 'b']})

>>> ds2
<xarray.Dataset>
Dimensions:  (x: 2, y: 2)
Coordinates:
  * y        (y) |S1 'a' 'b'
Data variables:
    z        (x, y) int64 1 2 3 4

>>> ds2.to_dataframe()
     z
x y   
0 a  1
  b  2
1 a  3
  b  4
```

jhamman:
@jmunroe - can we help move this forward? I'd like to see this get into v0.10 if possible.

jmunroe:
Hi @jhamman.  Thanks for the nudge.  I'll look at this again today and either a) just get it done or b) ask for help where needed.

xarray/core/dataset.py
jmunroe: Done. Coordinate variables are now retained as dask dataframe columns.

jmunroe:
Hi @shoyer and @jhamman . Thanks for your patience. Please let me know if there is still anything needed to be done on this PR.

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,60 @@ def from_dataframe(cls, dataframe):
...
+        # convert all coordinates into data variables
+        ds = self.reset_index(coord_columns).reset_coords()
+
+        # ensure we are dealing with dask arrays with consistent chunks
+        chunked = ds.chunk(chunks=self.chunks)
+
+        # order columns so that coordinates appear before data
+        columns = list(chunked)
+        num_coords = len(coord_columns)
+        columns = columns[-num_coords:] + columns[:-num_coords]
```
shoyer: This logic is a little fragile -- it relies on the iteration order of xarray datasets, which although unlikely to change is not really guaranteed to have coordinates last.

It would be better to build this up explicitly from the dataset, e.g.,
```
columns = list(ds.dims)
columns.extend(k for k in ds.coords if k not in ds.dims)
columns.extend(ds.data_vars)
```

xarray/tests/test_dask.py
```diff
@@ -546,6 +547,92 @@ def test_from_dask_variable(self):
...
+                                  ('t', ('t', t))]))
+
+        expected_pd = pd.DataFrame({'a': x,
+                                    'b': y},
+                                   index=pd.Index(t, name='t'))
+
+        actual = ds.to_dask_dataframe(set_index=True)
+        self.assertIsInstance(actual, dd.DataFrame)
+        assert expected_pd.equals(actual.compute())
+
```
shoyer: One more test I would like to see is an xarray Dataset without coordinate labels along a dimension, e.g., `xarray.Dataset({'foo': ('x', [1, 2, 3])})`. `x` should appear as a column in the dask DataFrame as it does when converting into pandas, with default integer values `[0, ..., n-1]`.

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,60 @@ def from_dataframe(cls, dataframe):
...
+        multi-indexes, set_index only works if there is one coordinate dimension.
+        """
+
+        import dask.dataframe as dd
+
+        ordered_dims = self.dims
+        coord_columns = [c for c in self if c in ordered_dims]
+
+        # convert all coordinates into data variables
+        ds = self.reset_index(coord_columns).reset_coords()
```
shoyer: Is this actually necessary? Everything is already in `._variables`.

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,60 @@ def from_dataframe(cls, dataframe):
...
+
```
shoyer: You might need to assign default coordinates here for dimensions without coordinates. This might be easiest to do by pulling out default variables from this dataset, e.g.,
```python
ds = self.assign_coords(**{k: ds[k] for k in ds.dims})
```

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,60 @@ def from_dataframe(cls, dataframe):
...
+        # order columns so that coordinates appear before data
+        columns = list(chunked)
+        num_coords = len(coord_columns)
+        columns = columns[-num_coords:] + columns[:-num_coords]
+
+        data = [chunked._variables[k].set_dims(ordered_dims).data.reshape(-1)
+                for k in columns]
+
+        # restore names of columns that were also index names (e.g. x_ -> x)
+        columns = [c[:-1] if c[:-1] in ds.dims else c for c in columns]
```
shoyer: This feels a little fragile -- it would be nice to avoid if possible. (See my note above on avoiding `reset_index()`/`reset_coords()`)

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,60 @@ def from_dataframe(cls, dataframe):
...
+                        'set_index=True only is valid for '
+                        'for one-dimensional datasets')
+
+            # extract out first (and only) coordinate variable
+            coord_dim = list(ordered_dims)[0]
+
+            df = df.set_index(coord_dim)
+
+        if isinstance(df, dd.Series):
+            df = df.to_frame()
```
shoyer: Does `dask.dataframe.concat` really create a Series if asked to concat along `axis=1` with only one input? If so, it would be appreciated if you could file this as an dask bug (which it almost certainly is) and reference it here.

xarray/tests/test_dask.py
```diff
@@ -546,6 +547,92 @@ def test_from_dask_variable(self):
                       coords={'x': range(4)}, name='foo')
         self.assertLazyAndIdentical(self.lazy_array, a)
 
+    def test_to_dask_dataframe(self):
```
shoyer: It would be appreciated if you could break this into a few more sub-methods. We don't always follow this well currently, but smaller tests that only test one things are easier to work with.

There's no strict line limit, but aim for less than 10-20 lines if possible. Another good time to break a test into parts is when you have different input data.

xarray/core/dataset.py
jmunroe: My intent was to be able to call Dataset.chunk() so that all of the variables (both coordinates and data) were first converted to dask arrays. But IndexVariables only have a dummy .chunk() method.  

I'll change it so coordinate variables are converted with .to_base_variable()

xarray/core/dataset.py
jmunroe: I can't see the motivation for including the default coordinates in the dask dataframe for dimensions without coordinates.  But I have ensured that dimensions without coordinates do not cause a problem.

xarray/core/dataset.py
jmunroe: Ah! that is much better.  I'll do what I think is the simplest thing:

```
columns = list(self.coords) + list(self.data_vars)
```

xarray/core/dataset.py
jmunroe: I've restructured the code so that this is no longer needed.

xarray/core/dataset.py
jmunroe: Yes, that appears to be the current behaviour of dask.dataframe.concat.  And I agree that this not the same as what pd.concat does (if s is a pandas Series, then pd.concat([s], axis=1) is indeed a DataFrame).

xarray/tests/test_dask.py
jmunroe: added.

xarray/tests/test_dask.py
jmunroe: No problem.  Done.

xarray/core/dataset.py
jmunroe: Issue filed as  dask/dask#2798

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,66 @@ def from_dataframe(cls, dataframe):
...
+        chunks = self.chunks
+
+        # order columns so that coordinates appear before data
+        columns = list(self.coords) + list(self.data_vars)
+
+        data = []
+        for k in columns:
+            v = self._variables[k]
+
+            # consider coordinate variables as well as data varibles
```
shoyer: This is a good place to mention in a comment your discovery that we need to convert to base variables in order for chunk() to work properly.

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,66 @@ def from_dataframe(cls, dataframe):
...
+                v = v.to_base_variable()
+
+            # ensure all variables span the same dimensions
+            v = v.set_dims(ordered_dims)
+
+            # ensure all variables have the same chunking structure
+            if v.chunks != chunks:
+                v = v.chunk(chunks)
+
+            # reshape variable contents as a 1d array
```
shoyer: Nit: some of these comments are probably slightly overboard -- if they simply restate what's in the code it's better to omit them.

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,66 @@ def from_dataframe(cls, dataframe):
...
+            data.append(s)
+
+        df = dd.concat(data, axis=1)
+
+        if set_index:
+
+            if len(ordered_dims) != 1:
+                raise ValueError(
+                        'set_index=True only is valid for '
+                        'for one-dimensional datasets')
```
shoyer: Can you include the list of multiple dimensions in the error message?

xarray/core/dataset.py
shoyer: My reasoning here would be that the default coordinates might not be necessary in array form, but in flattened DataFrame form they're probably necessary to work with the data. For example, consider an image with dimensions `('x', 'y', 'channel')` without any labels along x and y. If you don't add the default dimensions, when you flatten the image you can no longer distinguish between the two original orderings (x, y) and (y, x).

jmunroe:
I don't understand how only test (TestDataArrayAndDataset::test_to_dask_dataframe_2D) can pass on TravisCI yet fail on Appveyor.

shoyer:
@jcrist @mrocklin @jhamman do any of you have opinions on my latest design [question above](https://github.com/pydata/xarray/pull/1489#pullrequestreview-70750344) about the order of elements in dask dataframes? Is it as important as I suspect to keep chunking/divisions consistent when converting from arrays to dataframes?

mrocklin:
I think that you would want to rechunk the dask.array so that its chunks align with the outputs divisions of the dask.dataframe.  For example if you have a 2d array and are partitioning along the x-axis then you will want to align the array so that there is no chunking along the y axis.  In this case `set_index` will also be free because your data is already aligned and you already know (I think) the division values.

shoyer:
@mrocklin are you saying that it's easier to properly rechunk data on the xarray side (as arrays) before converting to dask dataframes? That does make sense -- we have some nice structure (as multi-dimensional arrays) that is lost once the data gets put in a DataFrame.

In this case, I suppose we really should add a keyword argument like `dims_order` to `to_dask_dataframe()` that lets the user choose how they want to order dimensions on the result.

Initially, I was concerned about the resulting dask graphs when flattening out arrays in the wrong order. Although that *would* have bad performance implications if you need to stream the data from disk, I see now the total number of chunks no longer blows up, thanks to @pitrou's impressive rewrite of `dask.array.reshape()`.

xarray/core/dataset.py
```diff
@@ -2416,6 +2416,66 @@ def from_dataframe(cls, dataframe):
...
+        the columns of the DataFrame.
+
+        If set_index=True, the dask DataFrame is indexed by this dataset's
+        coordinate.  Since dask DataFrames to not support multi-indexes,
+        set_index only works if there is one coordinate dimension.
+        """
+
+        import dask.dataframe as dd
+
+        ordered_dims = self.dims
```
shoyer: We should probably add that `dims_order` keyword argument. Then this becomes something like:

```python
if dims_order is None:
    dims_order = self.dims
ordered_dims = OrderedDict((k, self.dims[k]) for k in dims_order)
```

xarray/tests/test_dask.py
```diff
@@ -546,6 +547,101 @@ def test_from_dask_variable(self):
...
+            names=['x', 'y'])
+        expected = pd.DataFrame({'w': w.reshape(-1)},
+                                index=exp_index)
+        # so for now, reset the index
+        expected = expected.reset_index(drop=False)
+
+        actual = ds.to_dask_dataframe(set_index=False)
+
+        self.assertIsInstance(actual, dd.DataFrame)
+        assert expected.equals(actual.compute()), (expected, actual.compute())
```
shoyer: Try using `pandas.testing.assert_frame_equal` to get a better error message (for easier debugging on Appveyor):

```python
# at the top
try:
    from pandas.testing import assert_frame_equal
except ImportError:
    # old location, for pandas < 0.20
    from pandas.util.testing import assert_frame_equal

# here
computed = actual.compute()
assert_frame_equal(actual, expected)

shoyer:
Just pushed a couple of commits, which should resolve the failures on Windows. It was typical int32 vs int64 NumPy on Windows nonsense.

shoyer:
@jmunroe Thanks for your help here! I'm going to merge this now and take care of my remaining clean-up requests in a follow-on PR.

jmunroe:
@shoyer Sound good. Thanks.

Issue #799, opened 2016-03-22
Support for pathlib.Path


mangecoeur:
`pathlib.Path` IMHO is one of the best additions to Python. Would be nice if it were possible to open files from `Path` without having to cast to `str`

shoyer:
:+1: we're definitely open to pull requests for this one

willirath:
+1 for this one.

PR #1513, opened 2017-08-21
WIP: Add pathlib support to `open_(mf)dataset`

willirath:
This has #799 in mind.


willirath:
Sorry, this was based on an old upstream.  I'll close and start from an up-to-date master.

/close

PR #1514, opened 2017-08-21
Add `pathlib.Path` support to `open_(mf)dataset`

### LINKED ISSUE #799: Support for pathlib.Path

willirath:
- [x] Closes #799
 - [x] Tests added / passed

This is meant to eventually make `xarray.open_dataset` and `xarray.open_mfdataset` work with `pathlib.Path` objects.  I think this can be achieved as follows:

1. In `xarray.open_dataset`, cast any `pathlib.Path` object to string

2. In `xarray.open_mfdataset`, make sure to handle generators.  This is necessary, because `pathlib.Path('some-path').glob()` returns generators.

Curently, tests with Python 2 are failing, because there is no explicit `pathlib` dependency yet.

With Python 3, everything seems to work.  I am not happy with the tests I've added so far, though.


MaximilianR:
Are you sure `pathlib` exists in py2? I had thought you needed to install `pathlib2` (I may be wrong on the specifics)

MaximilianR:
> Curently, tests with Python 2 are failing, because there is no explicit pathlib dependency yet.

I missed this, apologies.

Can we add `pathlib2` as an optional dependency and handle the case where it's not installed?

xarray/tests/test_backends.py
```diff
@@ -570,6 +585,20 @@ def create_tmp_file(suffix='.nc', allow_cleanup_failure=False):
 
 
 @contextlib.contextmanager
+def create_tmp_file_pathlib(suffix='.nc', allow_cleanup_failure=False):
```
shoyer: Rather than repeating everything from `create_tmp_file`, can you simply wrap the output from `create_tmp_file` in `Path`?

xarray/tests/test_backends.py
```diff
@@ -1275,20 +1304,22 @@ def test_dataset_caching(self):
...
-                                    autoclose=self.autoclose) as actual:
-                    self.assertIsInstance(actual.foo.variable.data, da.Array)
-                    self.assertEqual(actual.foo.variable.data.chunks,
-                                     ((5, 5),))
-                    self.assertDatasetAllClose(original, actual)
-                with open_mfdataset([tmp1, tmp2], chunks={'x': 3},
-                                    autoclose=self.autoclose) as actual:
-                    self.assertEqual(actual.foo.variable.data.chunks,
-                                     ((3, 2, 3, 2),))
+        for _create_tmp_file in [create_tmp_file, create_tmp_file_pathlib]:
```
shoyer: Can you make a separate `test_open_mfdataset_path` test rather than squeezing this into `test_open_mfdataset`?

Some copy/paste is OK, but it doesn't need to all the check we have here around chunks and dask arrays. A simple `self.assertDatasetAllClose(original, actual)` would suffice.

willirath:
Thanks for the review @shoyer !

On my machine, I've already stared adapting the other backends (`to_netcdf` et al) to support `pathlib` as well.  I'll include it here.

setup.py
```diff
@@ -37,6 +37,9 @@
 
 INSTALL_REQUIRES = ['numpy >= 1.7', 'pandas >= 0.15.0']
 TESTS_REQUIRE = ['pytest >= 2.7.1']
+if sys.version_info < (3, 0):
```
willirath: I've added a check for Python < 3 to `setup.py`.  This seems to work.  I'm not sure if this really is an accepted way of creating conditional requirements.

xarray/backends/api.py
```diff
@@ -6,6 +6,11 @@
 from glob import glob
 from io import BytesIO
 from numbers import Number
+try:
```
willirath: This prefers `pathlib2` and only falls back to `pathlib` if necessary. I'm following <https://github.com/mcmtroffaes/pathlib2/issues/8#issuecomment-162890916> here although preferring the stdlib module would feel better.

What do you think?

willirath:
> - [x] Tests added / passed

Tests are already done on my personal Travis Account (skipped intermediate commits there).

> - [x] Passes git diff upstream/master | flake8 --diff

Actually true only for `git diff upstream/master xarray/ | flake8 --diff`.  But I guess that's what should pass the linter?

To me, aae32a8 looks like pathlib support is now present whereever it makes sense.

willirath:
One more thing:

> Can we add pathlib2 as an optional dependency and handle the case where it's not installed?

Currently, I've set `pathlib2` as a mandatory requirement for Python 2. How'd we achieve `pathlib` being optional? Test for `pathlib` upon import and wrap all the `if isinstace(path, Path): path = str(path)` in a function just passing unmodified `path` if `pathlib`'s not present?

shoyer:
@willirath take a look at what we do for handling the optional dask dependency:
https://github.com/pydata/xarray/blob/bcd608101133c0cb84c74d341d22edef71ef4818/xarray/core/pycompat.py#L55-L60

willirath:
> take a look at what we do for handling the optional dask dependency

Ah, that's nice!

And then I remove the `pathlib2` dependency I've introduced in [setup.py](https://github.com/willirath/xarray/blob/799-add-pathlib-support-2/setup.py#L40) again?

shoyer:
> And then I remove the pathlib2 dependency I've introduced in setup.py again?

Yes, `pathlib2` should not be a required dependency.

xarray/tests/test_backends.py
```diff
@@ -40,6 +41,12 @@
 except ImportError:
     pass
 
+with suppress(ImportError):
+    try:
+        from pathlib import Path
+    except ImportError:
+        from pathlib2 import Path
+
```
jhamman: let's move this to `tests/__init__.py`. We'll also want a `requires_pathlib` defined there.

xarray/tests/test_backends.py
```diff
@@ -1343,6 +1362,19 @@ def test_save_mfdataset_invalid(self):
         with self.assertRaisesRegexp(ValueError, 'same length'):
             save_mfdataset([ds, ds], ['only one path'])
 
+    def test_save_mfdataset_pathlib_roundtrip(self):
```
jhamman: add `@requires_pathlib` decorator

xarray/tests/test_backends.py
```diff
@@ -1934,3 +1966,13 @@ def test_open_dataarray_options(self):
             expected = data.drop('y')
             with open_dataarray(tmp, drop_variables=['y']) as loaded:
                 self.assertDataArrayIdentical(expected, loaded)
+
+    def test_dataarray_to_netcdf_no_name_pathlib(self):
```
jhamman: add `@requires_pathlib` decorator

xarray/backends/api.py
```diff
@@ -6,6 +6,15 @@
...
+    try:
+        from pathlib import Path
+    except ImportError as e:
+        from pathlib2 import Path
+    path_type = (Path, )
+except ImportError as e:
+    path_type = ()
```
jhamman: this looks right to me but we should put it in `pycompat.py`

asv_bench/asv.conf.json
```diff
@@ -36,7 +36,7 @@
     "install_timeout": 600,
 
     // the base URL to show a commit for the project.
-    // "show_commit_url": "http://github.com/owner/project/commit/",
```
jhamman: I think this is already on master, can we roll this back?

xarray/tests/test_backends.py
willirath: I added the `requires_pathlib` decorator but couldn't remove this nested import block from `test_backends.py`, because `Path()` is explicitly used to set up the pathlib tests.

xarray/tests/test_backends.py
willirath: Done.

xarray/tests/test_backends.py
willirath: Done.

xarray/backends/api.py
willirath: Done.

asv_bench/asv.conf.json
willirath: I've rebased onto master some point.  Reverted it now.

xarray/backends/api.py
```diff
@@ -139,12 +148,12 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
...
     ----------
-    filename_or_obj : str, file or xarray.backends.*DataStore
-        Strings are interpreted as a path to a netCDF file or an OpenDAP URL
-        and opened with python-netCDF4, unless the filename ends with .gz, in
-        which case the file is gunzipped and opened with scipy.io.netcdf (only
-        netCDF3 supported). File-like objects are opened with scipy.io.netcdf
-        (only netCDF3 supported).
+    filename_or_obj : str, Path, file or xarray.backends.*DataStore
+        Strings and Path objects are interpreted as a path to a netCDF file
+        oran OpenDAP URL and opened with python-netCDF4, unless the filename
```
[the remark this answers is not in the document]
willirath: Done.

willirath:
`flake8` now fails because of the unused `pathlib` in `xarray/tests/__init__.py`.

willirath:
AppVeyor test failed with HTTP time-outs when trying to get `repodata.json`.  Can somebody trigger the build again?

doc/whats-new.rst
```diff
@@ -21,9 +21,38 @@ v0.9.7 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- Support for `pathlib.Path` objects added to
```
jhamman: nit: we typically cite the issue number (e.g. `:issue: 799:`). Would be nice to include here.

xarray/backends/api.py
```diff
@@ -496,6 +501,9 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     """
     if isinstance(paths, basestring):
         paths = sorted(glob(paths))
+    else:
+        paths = [str(p) if isinstance(p, path_type) else p for p in paths]
```
jhamman: You may have already discussed this with @shoyer but can you remind me why we're not sorting in the same way we do for the glob path above? I guess we're assuming all the paths are expanded already?

shoyer: We sort after `glob()` since the iteration order in arbitrary. But we don't sort in general, since the order of the provided filenames might be intentional.

Unfortunately, there isn't any way to detect a generator created by `pathlib`'s `glob()` method, since it's just a Python generator.

doc/whats-new.rst
shoyer: I just pushed a commit to add this

doc/whats-new.rst
```diff
@@ -21,9 +21,38 @@ v0.9.7 (unreleased)
...
+    In [3]: data_dir = Path("data/")
+
+    In [4]: one_file = data_dir / "dta_for_month_01.nc"
+
+    In [5]: print(xr.open_dataset(one_file))
+    Out[5]:
+    <xarray.Dataset>
+    [...]
+
+    In [6]: all_files = data_dir.glob("dta_for_month_*.nc")
```
shoyer: I removed this example from What's New for two reason:
1. The section was getting a little longer than essential
2. It's actually a bit of an anti-pattern, since the order of paths matters to xarray but is arbtirary from glob. The right way to write this is `sorted(data_dir.glob(...))`.

willirath:
Thanks @shoyer!

I like the shorter example. It shows the essence of why pathlib is such a
nice thing.

Is there anything more to do in this PR?


Am 31. August 2017 17:48:24 schrieb Stephan Hoyer <notifications@github.com>:

shoyer:
CI failures are all the issue with dask distributed (https://github.com/pydata/xarray/issues/1540), so I'm going ahead and merging.

Thanks @willirath !

PR #1516, opened 2017-08-23
Small documentation fixes

leezu:
- Clarify in documentation that inferring DataArray dimensions from dictionary like ``coords`` has been deprecated.
- DataArray docstring

### DIFF
--- a/doc/data-structures.rst
+++ b/doc/data-structures.rst
@@ -46,9 +46,11 @@ The :py:class:`~xarray.DataArray` constructor takes:
 
 - ``data``: a multi-dimensional array of values (e.g., a numpy ndarray,
   :py:class:`~pandas.Series`, :py:class:`~pandas.DataFrame` or :py:class:`~pandas.Panel`)
-- ``coords``: a list or dictionary of coordinates
-- ``dims``: a list of dimension names. If omitted, dimension names are
-  taken from ``coords`` if possible (deprecated).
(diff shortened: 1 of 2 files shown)


doc/data-structures.rst
```diff
@@ -48,7 +48,7 @@ The :py:class:`~xarray.DataArray` constructor takes:
   :py:class:`~pandas.Series`, :py:class:`~pandas.DataFrame` or :py:class:`~pandas.Panel`)
 - ``coords``: a list or dictionary of coordinates
 - ``dims``: a list of dimension names. If omitted, dimension names are
-  taken from ``coords`` if possible.
+  taken from ``coords`` if possible (deprecated).
```
MaximilianR: I would vote to remove the deprecated behavior from the docs

shoyer: Rather than "if possible" I would say "if coords was provided as a list"

(I think that's the only case where it's supported now.)

leezu: @shoyer when speaking of list, do you mean a list of tuples `(name, [coordinates..])`?

shoyer: Yes, exactly

leezu: Great. I have updated the PR accordingly. It seems the format of the list was previously not documented.

shoyer:
Thanks

PR #1526, opened 2017-08-25
Fixed issue #1520, by adding a if-else that checks for None

### LINKED ISSUE #1520: Rasterio CRS-attribute can be None
[open_rasterio](https://github.com/pydata/xarray/blob/b201ff784a9c74dbed3b0b178b01bbea75ae9c85/xarray/backends/rasterio_.py)  does not work when rasterio crs is None. It is also clearly visible from the code that this is the case: Here's the code from line 144 onward
```python
attrs = {}`
    if hasattr(riods, 'crs'):
        # CRS is a dict-like object specific to rasterio
        # We convert it back to a PROJ4 string using rasterio itself
        attrs['crs'] = riods.crs.to_string()
    if hasattr(riods, 'res'):
        # (width, height) tuple of pixels in units of CRS
        attrs['res'] = riods.res
    if hasattr(riods, 'is_tiled'):
        # Is the TIF tiled? (bool)
        # We cast it to an int for netCDF compatibility
        attrs['is_tiled'] = np.uint8(riods.is_tiled)
    if hasattr(riods, 'transform'):
        # Affine transformation matrix (tuple of floats)
        # Describes coefficients mapping pixel coordinates to CRS
        attrs['transform'] = tuple(riods.transform)`
```

I have version 1.0a9 for rasterio and 0.9.6 for xarray. For what I looked, the issue [#1466](https://github.com/pydata/xarray/issues/1466) is related, but not same issue. 

CRS is None when I read ENVI-dataset that has no georeferencing with rasterio.

leevei:
Made this fix for the issue #1520. I simpy added a test for None in the code and tested it with my ENVI-file that caused the problem that I reported. 

If I understood the test report correctly, the tests also passed. 

- [x] Closes #1520
- [x] Tests  passed
- [x] Passes git diff upstream/master | flake8 --diff

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -142,15 +142,11 @@ def open_rasterio(filename, chunks=None, cache=None, lock=None):
 
     # Attributes
     attrs = {}
-    if hasattr(riods, 'crs'):
+    if hasattr(riods, 'crs') and riods.crs:
         # CRS is a dict-like object specific to rasterio
-        if riods.crs is None:
-            # If crs is None we put empy string to attrs list
-            attrs['crs'] = ''
-        else:
-            # If CRS is not None, we convert it back to a PROJ4 string using
-            # rasterio itself
-            attrs['crs'] = riods.crs.to_string()
+        # If CRS is not None, we convert it back to a PROJ4 string using
+        # rasterio itself
+        attrs['crs'] = riods.crs.to_string()
     if hasattr(riods, 'res'):
         # (width, height) tuple of pixels in units of CRS
         attrs['res'] = riods.res
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -144,8 +144,13 @@ def open_rasterio(filename, chunks=None, cache=None, lock=None):
     attrs = {}
     if hasattr(riods, 'crs'):
         # CRS is a dict-like object specific to rasterio
-        # We convert it back to a PROJ4 string using rasterio itself
-        attrs['crs'] = riods.crs.to_string()
+        if riods.crs is None:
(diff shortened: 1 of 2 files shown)


xarray/backends/rasterio_.py
```diff
@@ -144,8 +144,13 @@ def open_rasterio(filename, chunks=None, cache=None, lock=None):
     attrs = {}
     if hasattr(riods, 'crs'):
         # CRS is a dict-like object specific to rasterio
-        # We convert it back to a PROJ4 string using rasterio itself
-        attrs['crs'] = riods.crs.to_string()
+        if riods.crs is None:
+            # If crs is None we put empy string to attrs list
+            attrs['crs'] = ''
```
shoyer: Is this more useful than not simply dropping the attribute?

leevei: I'm not really sure, since I don't know what it does in the first place. I'll look into it on Monday when I'm back at work.

jhamman: My vote would be to keep this simple. 
```Python
if hasattr(riods, 'crs') and riods.crs:
    attrs['crs'] = riods.crs
```

leevei: After looking into documentation of CRS-parateter, I have to agree that it should not be empty string. I will commit changes in 10 minutes.

jhamman:
This looks good to me. It would be nice to have a test for this but I'm not sure its worth the effort for what is probably a rarely seen corner case. 

We do need a whats-new note and to check the flake8 box.

leevei:
I'm getting random errors on flake8 test from whats-new.rst, but otherwise it passes.

jhamman:
This looks good now. Failing tests are unrelated so I'm going to merge. Thanks!

PR #1532, opened 2017-08-28
Avoid computing dask variables on __repr__ and __getattr__

### LINKED ISSUE #1522: Dataset.__repr__ computes dask variables
DataArray.\_\_repr\_\_ and Variable.\_\_repr\_\_ print a placeholder if the data uses the dask backend.
Not so Dataset.\_\_repr\_\_, which tries computing the data before printing a tiny preview of it. 
This issue is extremely annoying when working in Jupyter, and particularly acute if the chunks are very big or are at the end of a very long chain of computation.

For data variables, the expected behaviour is to print a placeholder just like DataArray does.
For coords, we could either
- print a placeholders (same treatment as data variables)
- automatically invoke load() when the coord is added to the dataset - see #1521 for discussion.

crusaderky:
- [x] Fixes #1522
 - [x] Tests added / passed

Stop dataset data vars and non-index dataset/dataarray coords from being loaded by repr() and getattr(). The latter is particularly acute when working in Jupyter, which does a dozen or so getattr() when printing an object.


crusaderky:
Fixes #1522

rabernat:
Just wanted to say that I'm so glad someone took this on!

Will this also fix the very slow tab-autocomplete on dask-backed xarray objects? This is something I hit frequently in interactive work with big datasets. I assume it's related.

xarray/core/formatting.py
```diff
@@ -222,38 +222,28 @@ def _summarize_coord_multiindex(coord, col_width, marker):
...
-    if source and source.startswith('http') and not var._in_memory:
-        return False
-    return True
-
-
-def summarize_var(name, var, col_width):
-    show_values = _not_remote(var)
-    return _summarize_var_or_coord(name, var, col_width, show_values)
+def summarize_datavar(name, var, col_width):
+    show_values = var._in_memory
```
shoyer: Our current heuristic uses the `_not_remote()` helper function, so it doesn't display arrays loaded over a network (via opendap), which can often be quite slow. But it does display a summary of values from netCDF files on disk, which I do think is generally helpful and for which I haven't noticed any performance issues.

Based on the current definition of `_in_memory`, we wouldn't display any of these arrays:
https://github.com/pydata/xarray/blob/4a15cfa6f3e680af6969f3e2c6a375e2cc8c3288/xarray/core/variable.py#L285-L289

So instead of using `_in_memory`, I would suggest something like `_not_remote(var) and not isinstance(var._data, dask_array_type)` as the condition for showing values.

crusaderky: @shoyer loading a NetCDF variable from disk every time you do __repr__ is a terrible idea if that variable has been compressed withouth chunking. If the variable is a single block of 100MB of zlib-compressed data, you will have to read it and decompress it every time.

crusaderky:
@rabernat could you do an example? I never noticed. If the delay was caused by a miss on __getattr__, then yes, it will be solved.

xarray/core/formatting.py
crusaderky: @shoyer also, your netcdf array might be sitting on a network file system on the opposite side of a narrowband VPN.

xarray/tests/test_dask.py
```diff
@@ -341,47 +362,98 @@ def test_dot(self):
...
+        nonindex_coord = build_dask_array('coord')
+        ds = Dataset(data_vars={'a': ('x', data)}, coords={'y': ('x', nonindex_coord)})
+        expected = dedent("""\
+        <xarray.Dataset>
+        Dimensions:  (x: 1)
+        Coordinates:
+            y        (x) int64 ...
+        Dimensions without coordinates: x
+        Data variables:
+            a        (x) int64 ...""")
```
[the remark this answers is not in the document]
crusaderky: @shoyer fixed. Now it's the same as in Variable and in the DataArray data var.

crusaderky:
@rabernat I reproduced your problem and I'm happy to confirm that this fixes it!

xarray/core/formatting.py
shoyer: That's certainly possible, but in my experience very few people writes 100MB chunks -- those are very large.

Let's summarize our options:
1. Always show a preview of data from netCDF files with `Dataset.__repr__`
2. Never show a preview for data if it isn't already in memory
3. Show a preview depending on a global option (with default choice TBD).

Reasons to show data from disk in `__repr__`:
- It's what we've always done.
- "Most" of the time it's fast and convenient.
- It provides a good experience for new users, who don't need to hunt for a separate `preview()` or `load()` command to see what's in a Dataset. You can simply print it at a console.

Reasons not to show data from disk in `__repr__`:
- IO can be slow/expensive, especially if compression or networks are involved.
- Heuristics to detect expensive IO are unreliable and somewhat distasteful.

Maybe we should solicit a few more opinions here before we change the default behavior?

Another possibility is to try loading data in a separate thread and timeout if it takes too long (say more than 0.5 seconds), but that might open up it's own set of performance issues (it's not easy to kill a thread, short of terminating a process).

jhamman: I think my vote would be to only print a preview of data that is in memory. For my uses, I typically have fill values in the first 10-20 data points so the previous `__repr__` didn't give me any information.

@pydata/xarray - anyone else have thoughts on this?

fmaussion: I am also for only showing only data in memory and never read from disk for a ``__repr__``, for the same reasons. Looking forward to see this PR merged!

jhamman: @shoyer - do we have results from your google poll on this issue yet?

shoyer: Sounds like I was wrong -- the consensus is pretty clear that we should go ahead with this

screen shot 2017-09-20 at 12 22 32 pm
screen shot 2017-09-20 at 12 22 38 pm
screen shot 2017-09-20 at 12 22 41 pm
screen shot 2017-09-20 at 12 22 48 pm

jhamman: I'm not sure this sample size is going to give us statistically significant results but I'm glad to see @delgadom and I are in agreement.

@crusaderky - are you up for implementing this?

shoyer: I think the current implementation (in this PR) is actually already correct.

xarray/core/formatting.py
```diff
@@ -208,6 +208,9 @@ def _summarize_var_or_coord(name, var, col_width, show_values=True,
     front_str = u'%s%s%s ' % (first_col, dims_str, var.dtype)
     if show_values:
         values_str = format_array_flat(var, max_width - len(front_str))
+    elif isinstance(var.data, dask_array_type):
+        chunksize = tuple(c[0] for c in var.chunks)
+        values_str = 'dask.array<shape=%s, chunksize=%s>' % (var.shape, chunksize)
```
shoyer: can you make a little helper function for this, e.g., `dask_short_repr()`?

doc/whats-new.rst
```diff
@@ -131,7 +131,11 @@ Bug fixes
   ``rtol`` arguments when called on ``DataArray`` objects.
   By `Stephan Hoyer <https://github.com/shoyer>`_.
 
-- Xarray ``quantile`` methods now properly raise a ``TypeError`` when applied to
+- Stop ``repr`` and the Jupyter Notebook from automatically computing dask
```
shoyer: This should probably go under "breaking changes" instead, given that it changes existing behavior.

crusaderky:
Will finalise over the weekend

crusaderky:
All done

xarray/core/formatting.py
crusaderky: Yep - data is eagerly loaded from disk only for index coords on \_\_init\_\_ now.

shoyer:
Thanks @crusaderky !

PR #1538, opened 2017-08-30
Fix/1120

### LINKED ISSUE #1120: Creating a `Dataset` with a coordinate given by a `DataArray` may create an invalid dataset
Consider this:
```
import xarray as xr
import numpy as np
ds = xr.Dataset(
    data_vars=dict(
        v = ('y',np.r_[0:10]),
    ),
    coords=dict(
        y=xr.DataArray(np.r_[:10]/10,dims='x'),
    ),
)
```
The dataset `ds` will end up having dimensions `('x', 'y')` because of the coordinate `y` having dimensions `('x',)`.
This seems to violate some of the internal assumptions (`ds.indexes['y'].name` turns out to be `'x'`),
consequently `ds.sel(y=0)` will throw an exception (`ValueError: cannot use a dict-like object for selection on a dimension that does not have a MultiIndex`).

I came across this situation when trying to generate a dataset with a coordinate that is a copy of another pre-existing coordinate, but under a different name. My expectation was that the coordinate would be renamed (doing so manually also doesn't work out of the box due to #1116). Of course that is not the only possible interpretation of the construct above, arguably it should raise an exception instead (which is what `DataArray` currently does in the corresponding case).

jhamman:
- [x] Closes #1120
 - [x] Tests added / passed


jhamman:
failures here appear to be related to dask distributed.

xarray/core/merge.py
```diff
@@ -365,6 +365,21 @@ def merge_data_and_coords(data, coords, compat='broadcast_equals',
     return merge_core(objs, compat, join, explicit_coords=explicit_coords)
 
 
+def assert_valid_explicit_coords(variables, explicit_coords):
+    '''raise a MergeError if an explicit coord shares a name with a dimension
+    but is comprised of arbitrary dimensions'''
```
MaximilianR: If you care: pep8 is `"""` & on their own lines

xarray/core/merge.py
```diff
@@ -365,6 +365,24 @@ def merge_data_and_coords(data, coords, compat='broadcast_equals',
...
+    """Validate explicit coordinate names/dims.
+
+    Raise a MergeError if an explicit coord shares a name with a dimension
+    but is comprised of arbitrary dimensions.
+    """
+    var_dims = []
+    for name, var in variables.items():
+        if name not in explicit_coords:
+            var_dims.extend(var.dims)
+    for coord_name in explicit_coords:
```
shoyer: I think you're missing this as a function argument

xarray/core/merge.py
```diff
@@ -365,6 +365,24 @@ def merge_data_and_coords(data, coords, compat='broadcast_equals',
...
+        if coord_name in var_dims and not all(
+                [d in var_dims for d in variables[coord_name].dims]):
```
shoyer: Rather than `not all(...)`, I think this condition should be just `variables[coord_name].dims != (coord_name,)`

xarray/core/merge.py
```diff
@@ -365,6 +365,24 @@ def merge_data_and_coords(data, coords, compat='broadcast_equals',
     return merge_core(objs, compat, join, explicit_coords=explicit_coords)
 
 
+def assert_valid_explicit_coords(variables, explicit_coords):
+    """Validate explicit coordinate names/dims.
+
+    Raise a MergeError if an explicit coord shares a name with a dimension
+    but is comprised of arbitrary dimensions.
+    """
+    var_dims = []
```
shoyer: Use a set here.

xarray/core/merge.py
jhamman: `explicit_coords` is a function argument.

xarray/core/merge.py
jhamman: yes, this works.

xarray/core/merge.py
jhamman: I don't follow. Are you just suggesting we cast `var_dims` to a set after populating it or are you looking for some different logic for determining the dimensions in the non-explicit coord variables?

xarray/core/merge.py
shoyer: I was misreading something else, nevermind

xarray/core/merge.py
shoyer: Just that a `set()` is a better data structure for contains lookups than a list.

But actually, you should use the result of `calculate_dimensions()` here, instead of calculating dimensions twice.

jhamman:
@shoyer - thanks for the review. I iterated on this a few times and landed on something that was more complex than necessary. Your suggestions have been incorporated.

jhamman:
This is ready for a final review. Tests are passing now.

xarray/core/merge.py
```diff
@@ -365,6 +365,19 @@ def merge_data_and_coords(data, coords, compat='broadcast_equals',
...
+    """Validate explicit coordinate names/dims.
+
+    Raise a MergeError if an explicit coord shares a name with a dimension
+    but is comprised of arbitrary dimensions.
+    """
+    for coord_name in explicit_coords:
+        if coord_name in dims and variables[coord_name].dims != (coord_name,):
+            raise MergeError(
+                'coordinate %s shares a name with a dimension but '
+                'includes at least one arbitrary dimensions' % coord_name)
```
shoyer: "includes at least one arbitrary dimensions" is a little confusing to me.

Let's try to be totally explicit here, even if the error message needs to be longer. Maybe:

> coordinate X shares a name with a dataset dimension, but is not a 1D variable along that dimension. This is disallowed by the xarray data model.

jhamman: done.

PR #1539, opened 2017-08-30
remove warning and raise error when dataset constructor is called wit…

### LINKED ISSUE #1197: Problem with creating coords by dict
IMO there's a problem with using dicts as coords definition for DataArrays.
E.g. an array defined by
`import xarray as xr`
`arr = xr.DataArray(np.zeros([1, 2, 3]), {'x': [1], 'y': [1, 2], 'z': [1, 2, 3]})`
will lead to an exception when tried to be indexed like
`arr[0, 0, 0]`

The error message is
`ValueError: conflicting sizes for dimension 'z': length 2 on <this-array> and length 3 on 'z'`

As far as I can imagine this is because dicts don't preserve order in python.
At least printing the coords shows a different order than used in the definition of `arr`:
`arr.coords`
`Coordinates:`
`  * x        (x) int32 1`
`  * z        (z) int32 1 2 3`
`  * y        (y) int32 1 2`

Using an ordered dict turns out to work:
`import collections`
`od = collections.OrderedDict([('x', [1]), ('y', [1, 2]), ('z', [1, 2, 3])])`
`arr = xr.DataArray(np.zeros([1, 2, 3]), od)`
And so does using a list of tuples:
`arr = xr.DataArray(np.zeros([1, 2, 3]), [('x', [1]), ('y', [1, 2]), ('z', [1, 2, 3])])`

Both lead to
`arr[0, 0, 0]`
`<xarray.DataArray ()>`
`array(0.0)`
`Coordinates:`
`    x        int32 1`
`    y        int32 1`
`    z        int32 1`
as expected.

Detected in WinPython-64bit-3.4.4.1Qt5 with xarray version 0.7.0

jhamman:
…h coords dict and not dims

 - [x] Closes #1197
 - [x] Tests added / passed


xarray/core/dataarray.py
```diff
@@ -47,14 +47,15 @@ def _infer_coords_and_dims(shape, coords, dims):
...
-                              'explicit list of ``dims`` instead.',
-                              FutureWarning, stacklevel=3)
-                dims = list(coords.keys())
+                # deprecated in GH993, removed in GH1539
+                raise ValueError('inferring DataArray dimensions from '
+                                 'dictionary like ``coords`` has been '
+                                 'deprecated. Use an explicit list of '
+                                 '``dims`` instead.')
             else:
                 for n, (dim, coord) in enumerate(zip(dims, coords)):
```
shoyer: Can remove the `else` and dedent this clause now? (This is now the guard clause pattern.)

doc/whats-new.rst
```diff
@@ -15,9 +15,16 @@ What's New
...
 
-v0.9.7 (unreleased)
+v0.10.0 (unreleased)
 -------------------
 
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+- Supplying ``coords`` as a dictionary to the ``DataArray`` constructor without
+  also supplying an explicit ``dims`` argument is no longer supported. (:issue:`727`).
```
shoyer: Say something like: was deprecated in version X but is now an error.

PR #1543, opened 2017-08-31
pass dask compute/persist args through from load/compute/perist

### LINKED ISSUE #1523: Pass arguments to dask.compute()
I work with a very large dask-based algorithm in xarray, and I do my optimization by hand before hitting compute(). In other cases, I need using multiple dask schedulers at once (e.g. a multithreaded one for numpy-based work and a multiprocessing one for pure python work).

This change proposal (which I'm happy to do) is about accepting \*args, \*\*kwds parameters in all .compute(), .load(), and .persist() xarray methods and pass them verbatim to the underlying dask compute() and persist() functions.

jhamman:
- [x] Closes #1523
 - [x] Tests added / passed

cc @crusaderky - putting this up in case its useful to you. 

I'm frankly not sure if my tests are doing what they need to but everything seems to be playing nicely.

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -325,7 +325,7 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
         dask.array.compute
         """
         if isinstance(self._data, dask_array_type):
-            self._data = np.asarray(self._data.compute(**kwargs))
+            self._data = as_compatible_data(self._data.compute(**kwargs))
         elif not isinstance(self._data, np.ndarray):
             self._data = np.asarray(self._data)
         return self
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -1036,13 +1036,14 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
             data = duck_array_ops.stack(arrays, axis=axis)
 
         attrs = OrderedDict(first_var.attrs)
+        encoding = OrderedDict(first_var.encoding)
         if not shortcut:
             for var in variables:
                 if var.dims != first_var.dims:
                     raise ValueError('inconsistent dimensions')
                 utils.remove_incompatible_items(attrs, var.attrs)
 
-        return cls(dims, data, attrs)
+        return cls(dims, data, attrs, encoding)
 
     def equals(self, other, equiv=duck_array_ops.array_equiv):
         """True if two Variables have the same dimensions and values;
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -307,19 +307,30 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
     def _indexable_data(self):
         return orthogonally_indexable(self._data)
 
-    def load(self):
+    def load(self, **kwargs):
         """Manually trigger loading of this variable's data from disk or a
         remote source into memory and return this variable.
 
         Normally, it should not be necessary to call this method in user code,
         because all xarray functions should either work on deferred data or
         load data automatically.
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Additional keyword arguments passed on to ``dask.array.compute``.
+
+        See Also
+        --------
+        dask.array.compute
         """
-        if not isinstance(self._data, np.ndarray):
+        if isinstance(self._data, dask_array_type):
+            self._data = np.asarray(self._data.compute(**kwargs))
+        elif not isinstance(self._data, np.ndarray):
             self._data = np.asarray(self._data)
         return self
 
-    def compute(self):
+    def compute(self, **kwargs):
         """Manually trigger loading of this variable's data from disk or a
         remote source into memory and return a new variable. The original is
         left unaltered.
@@ -327,9 +338,18 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
         Normally, it should not be necessary to call this method in user code,
         because all xarray functions should either work on deferred data or
         load data automatically.
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Additional keyword arguments passed on to ``dask.array.compute``.
+
+        See Also
+        --------
+        dask.array.compute
         """
         new = self.copy(deep=False)
-        return new.load()
+        return new.load(**kwargs)
 
     @property
     def values(self):
--- a/xarray/tests/test_dask.py
+++ b/xarray/tests/test_dask.py
@@ -4,6 +4,8 @@ from __future__ import print_function
 import pickle
 import numpy as np
 import pandas as pd
+import pytest
+import mock
 
 import xarray as xr
 from xarray import Variable, DataArray, Dataset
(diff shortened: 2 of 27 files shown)


xarray/core/variable.py
```diff
@@ -307,29 +307,49 @@ def data(self, data):
...
+            Additional keyword arguments passed on to ``dask.array.compute``.
+
+        See Also
+        --------
+        dask.array.compute
         """
-        if not isinstance(self._data, np.ndarray):
+        if isinstance(self._data, dask_array_type):
+            self._data = np.asarray(self._data.compute(**kwargs))
+        elif not isinstance(self._data, np.ndarray):
```
crusaderky: This clause should be removed as it causes inconsistent behaviour with numpy scalar types. I cannot think of any other use case where data is neither a dask ARRAY nor a numpy ndarray?

xarray/core/variable.py
```diff
@@ -307,29 +307,49 @@ def data(self, data):
...
+        **kwargs : dict
+            Additional keyword arguments passed on to ``dask.array.compute``.
+
+        See Also
+        --------
+        dask.array.compute
         """
-        if not isinstance(self._data, np.ndarray):
+        if isinstance(self._data, dask_array_type):
+            self._data = np.asarray(self._data.compute(**kwargs))
```
crusaderky: You don't want to invoke asarray if dask returns a scalar numpy type

xarray/tests/test_dask.py
```diff
@@ -182,6 +183,26 @@ def test_bivariate_ufunc(self):
...
+        assert_equal(b2, expected)
+        # load
+        b3 = b.load(get=dask.multiprocessing.get, num_workers=4)
+        assert b3._in_memory
+        assert_equal(b3, expected)
+        # persist
+        b4 = b.persist(get=dask.multiprocessing.get, num_workers=4)
+        assert b4._in_memory
+        assert_equal(b4, expected)
+
```
crusaderky: Redundant with the test below?

xarray/tests/test_dask.py
```diff
@@ -403,6 +450,7 @@ def kernel():
     kernel_call_count += 1
     return np.ones(1)
 
+
 def build_dask_array():
     global kernel_call_count
     kernel_call_count = 0
```
crusaderky: Missing tests for Variable

crusaderky:
@jhamman happy to go with your version

xarray/core/variable.py
jhamman: Thanks, I don't think I meant to make this change.

xarray/tests/test_dask.py
jhamman: fixed, these should have been `Variable`s.

xarray/core/variable.py
jhamman: I think this allows for on-disk type arrays. @shoyer, any thoughts on calling `np.asarray` here and in the line above?

xarray/core/variable.py
shoyer: To be honest, we don't define what can go in `_data` as carefully as we ought to. I guess there are two ways to define it:
- anything "array like" that defines at least `shape`, `dtype` and `__getitem__`
- what comes out of `xarray.core.variable.as_compatible_data`

Numpy scalars do actually pass through here (since they define all those attributes!)... but then would get converted into an array when calling `.values` anyways:
https://github.com/pydata/xarray/blob/78ca20a6ea1a42eb637ae2ef09189f481cfda9a2/xarray/core/variable.py#L334-L337

So I guess I agree, but on the other hand I'm also a little nervous that a dask routine might return a non-numpy scalar, which would definitely break if we don't wrap it in `asarray`. The safe thing to do is to leave this as is or call `as_compatible_data` on it.

xarray/tests/test_dask.py
```diff
@@ -182,6 +183,22 @@ def test_bivariate_ufunc(self):
         self.assertLazyAndAllClose(np.maximum(u, 0), xu.maximum(v, 0))
         self.assertLazyAndAllClose(np.maximum(u, 0), xu.maximum(0, v))
 
+    def test_compute_args(self):
+        a = Variable('x', [1, 2]).chunk()
+        expected = Variable('x', [1, 4])
+        b = a * a
+        # compute
+        b1 = b.compute(get=dask.multiprocessing.get)
```
shoyer: The clean way to test this is probably with [mock](https://docs.python.org/3/library/unittest.mock.html), e.g.,
```
In [63]: import numpy as np

In [64]: import dask.array as da

In [65]: from unittest import mock

In [66]: x = da.from_array(np.arange(3), chunks=(2,))

In [67]: with mock.patch.object(da.Array, 'compute', return_value=np.arange(3)) as mock_compute:
    ...:     x.compute(foo='bar')
    ...:

In [68]: mock_compute.assert_called_with(foo='bar')

In [69]: mock_compute.assert_called_with(bar='foo')
<ipython-input-69-22937cf26ca1> in <module>()
----> 1 mock_compute.assert_called_with(bar='foo')

~/conda/envs/xarray-dev/lib/python3.5/unittest/mock.py in assert_called_with(_mock_self, *args, **kwargs)
    792         if expected != actual:
    793             cause = expected if isinstance(expected, Exception) else None
--> 794             raise AssertionError(_error_message()) from cause
    795
    796

AssertionError: Expected call: compute(bar='foo')
Actual call: compute(foo='bar')
```

`unittest.mock` is part of Python 3's standard library, but there's also a widely used Python 2 backport on pypi. I think it would be perfectly fine to add it as a dependency for our test suite.

jhamman: Thanks @shoyer, I'll add mock as part of this PR.

xarray/tests/test_dask.py
```diff
@@ -410,31 +396,58 @@ def test_from_dask_variable(self):
...
-        ds.load(get=dask.multiprocessing.get, num_workers=4)
-        ds.persist(get=dask.multiprocessing.get, num_workers=4)
+
+@pytest.mark.parametrize("method", ['load', 'compute'])
+def test_dask_kwargs_variable(method):
+    x = Variable('y', da.from_array(np.arange(3), chunks=(2,)))
+    with mock.patch.object(Variable, method,
+                           return_value=np.arange(3)) as mock_method:
+        getattr(x, method)(foo='bar')
+    mock_method.assert_called_with(foo='bar')
```
shoyer: I think you can skip this first test. This verifies that `Variable.compute()` was called if you directly call `Variable.compute()`.

xarray/tests/test_dask.py
```diff
@@ -4,6 +4,8 @@
 import pickle
 import numpy as np
 import pandas as pd
+import pytest
+import mock
```
shoyer: On Python 3, use `from unittest import mock` to avoid adding the new dependency.

shoyer:
Can you also update `tests_require` in `setup.py`?
https://github.com/pydata/xarray/blob/9a8e2c58ac4f953b3763d55e840fea595d626dfa/setup.py#L39

xarray/tests/test_dask.py
```diff
@@ -4,18 +4,20 @@
...
 from xarray.core.pycompat import suppress
 from . import TestCase, requires_dask
 
-from xarray.tests import unittest
+from xarray.tests import unittest, assert_equal, mock
 
 with suppress(ImportError):
     import dask
     import dask.array as da
+    import dask.multiprocessing
```
shoyer: No longer used

PR #1546, opened 2017-09-01
return bytes from DataArray.to_netcdf when path is not provided

### LINKED ISSUE #1410: to_netcdf without path argument returns None rather than bytes
I am using:
    Ubuntu 16.04
    python 3.6.0 
    xarray 0.9.1
    scipy 0.18.1

The [documentation states](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.to_netcdf.html) that if no path argument is passed to the to_netcdf method of the DataArray class it returns a bytes object. However, it returns None. 

```python
import xarray as xr
dar = xr.DataArray([1, 2, 3])
output = dar.to_netcdf()
assert isinstance(output, bytes)  # fails because output is None
```

jhamman:
- [x] Closes #1410
 - [x] Tests added / passed

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -124,6 +124,10 @@ Bug fixes
   it is none.
   By `Leevi Annala <https://github.com/leevei>`_.
 
+- Fix :py:func:`xarray.DataArray.to_netcdf` to return bytes when if no path is
+  provided (:issue:`1410`).
+  By `Joe Hamman <https://github.com/jhamman>`_.
+
 .. _whats-new.0.9.6:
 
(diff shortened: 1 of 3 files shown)


doc/whats-new.rst
```diff
@@ -124,6 +124,10 @@ Bug fixes
   it is none.
   By `Leevi Annala <https://github.com/leevei>`_.
 
+- Fix :py:func:`xarray.DataArray.to_netcdf` to return bytes when if no path is
```
spencerahill: typo 'when if'

Issue #1542, opened 2017-08-31
Testing: Failing tests on py36-pandas-dev


jhamman:
We currently have 7 failing tests when run against the pandas development code ([travis](https://travis-ci.org/pydata/xarray/jobs/270511674)).

Question for @shoyer - can you take a look at these and see if we should try to get a fix in place prior to v.0.10.0? It looks like Pandas.0.21 is slated for release on Sept. 30.

shoyer:
It looks like two issues here:

-[ ] Something related to `resample`
-[ ] Something related to `isnull`/`isna`

I'll start digging when we I have time. Let's mark this as a 0.10 blocker for now, until we figure out what's going on.

MaximilianR:
potentially (but supposed to be warnings only) https://github.com/pandas-dev/pandas/pull/16972

shoyer:
Fixes for both of these are up for review in #1548 and #1549.

jhamman:
closed via #1548 and #1549.

PR #1549, opened 2017-09-02
Fix errors with isnull() that appear with pandas 0.21

shoyer:
xref #1542

 - [x] Tests added / passed


shoyer:
The culprit is the truly monstrous `_func_slash_method_wrapper` function where we read names from methods if need be:
https://github.com/pydata/xarray/blob/master/xarray/core/ops.py#L211-L216

We should have a more sane way to add methods on xarray objects, but this fix will have to do for now.

PR #1551, opened 2017-09-02
Load nonindex coords ahead of concat()

### LINKED ISSUE #1521: open_mfdataset reads coords from disk multiple times
I have 200x of the below dataset, split on the 'scenario' axis:

```
<xarray.Dataset>
Dimensions:      (fx_id: 39, instr_id: 16095, scenario: 2501)
Coordinates:
    currency     (instr_id) object 'GBP' 'USD' 'GBP' 'GBP' 'GBP' 'EUR' 'CHF' ...
  * fx_id        (fx_id) object 'USD' 'EUR' 'JPY' 'ARS' 'AUD' 'BRL' 'CAD' ...
  * instr_id     (instr_id) object 'property_standard_gbp' ...
  * scenario     (scenario) object 'Base Scenario' 'SSMC_1' 'SSMC_2' ...
    type         (instr_id) object 'Common Stock' 'Fixed Amortizing Bond' ...
Data variables:
    fx_rates     (fx_id, scenario) float64 1.236 1.191 1.481 1.12 1.264 ...
    instruments  (instr_id, scenario) float64 1.0 1.143 0.9443 1.013 1.176 ...
Attributes:
    base_currency:  GBP
```

I individually dump them to disk with Dataset.to_netcdf(fname, engine='h5netcdf').
Then I try loading them back up with open_mfdataset, but it's mortally slow:

```
%%time
xarray.open_mfdataset('*.nc', engine='h5netcdf')

Wall time: 30.3 s
```

The problem is caused by the coords being read from disk multiple times.
Workaround:

```
%%time
def load_coords(ds):
    for coord in ds.coords.values():
        coord.load()
    return ds
xarray.open_mfdataset('*.nc', engine='h5netcdf', preprocess=load_coords)
Wall time: 12.3 s
```

Proposed solutions:
1. Implement the above workaround directly inside open_mfdataset()
2. change open_dataset() to always eagerly load the coords to memory, regardless of the chunks parameter. Is there any valid use case where lazy coords are actually desirable?

An additional, more radical observation is that, very frequently, a user knows in advance that all coords are aligned. In this use case, the user could explicitly request xarray to blindly trust this assumption, and thus skip loading the coords not based on concat_dim in all datasets beyond the first.

crusaderky:
- [x] Closes #1521
 - [x] Tests added / passed


xarray/core/combine.py
```diff
@@ -208,6 +219,8 @@ def _dataset_concat(datasets, dim, data_vars, coords, compat, positions):
     dim, coord = _calc_concat_dim_coord(dim)
     datasets = [as_dataset(ds) for ds in datasets]
     datasets = align(*datasets, join='outer', copy=False, exclude=[dim])
+    # TODO: compute dask coords with a single invocation of dask.compute()
+    datasets = [_load_coords(ds) for ds in datasets]
```
shoyer: We only need this if the argument `coords='different'` is set. Otherwise we don't need to load coords to determine if they are equal or not.

xarray/core/combine.py
```diff
@@ -208,6 +219,8 @@ def _dataset_concat(datasets, dim, data_vars, coords, compat, positions):
     dim, coord = _calc_concat_dim_coord(dim)
     datasets = [as_dataset(ds) for ds in datasets]
     datasets = align(*datasets, join='outer', copy=False, exclude=[dim])
+    # TODO: compute dask coords with a single invocation of dask.compute()
```
shoyer: Yes, this would be even better :)

crusaderky: I didn't bother as it would be a considerable complication and it would not do any good in the most common case where the "dask" variable is just a lazy load from disk

shoyer:
Yeah, that's OK, I just liked your TODO :)

rabernat:
I think this should help with #1385.

crusaderky:
New, much better version.
There's one test failure but that's actually because now it triggers #1586.
combine.py should be final; need to add more unit tests.

crusaderky:
Finished. One more test failure, as it now triggers #1588 too.

crusaderky:
Should be ready to be merged immediately after #1589 and #1590

crusaderky:
@shoyer ready to merge if there's no outstanding critiques

xarray/tests/test_combine.py
```diff
@@ -268,6 +269,7 @@ def test_concat(self):
         with self.assertRaisesRegexp(ValueError, 'not a valid argument'):
             concat([foo, bar], dim='w', data_vars='minimal')
 
+    @pytest.mark.xfail(reason='https://github.com/pydata/xarray/issues/1586')
```
shoyer: I just merged your other PR -- can you remove this now?

crusaderky:
@shoyer ready for merge

PR #1556, opened 2017-09-07
raise error if objects other than datasets are passed to save_mfdataset

### LINKED ISSUE #1555: TypeError: DataArray.name or Dataset key must be either a string or None for serialization to netCDF files
Consider the following [gist](https://gist.github.com/kuchaale/ffb9534e60f3646d81e501e8202efba2) where I receive the following error: `TypeError: DataArray.name or Dataset key must be either a string or None for serialization to netCDF files`. 

I think that I run into a bug when `DataArray.name` is checked by `check_name` function. I think that `k` in the corresponding for loop actually represents `DataArray` not `DataArray.name` itself.  Thus, I suggest something like this to fix it:
`for k in dataset:
      check_name(k.name)
`

jhamman:
- [x] Closes #1555
 - [x] Tests added / passed

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -2,10 +2,11 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
-PAPER         =
-BUILDDIR      = _build
+SPHINXOPTS      =
+SPHINXBUILD     = sphinx-build
+SPHINXATUOBUILD = sphinx-autobuild
+PAPER           =
+BUILDDIR        = _build
 
 # Internal variables.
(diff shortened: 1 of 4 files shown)


doc/Makefile
```diff
@@ -18,6 +19,7 @@ I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
 	@echo "  html       to make standalone HTML files"
+	@echo "  livehtml   Make standalone HTML files and rebuild the documentation when a change is detected. Also includes a livereload enabled web server"
```
shoyer: Nice. I didn't realize this was a thing!

PR #1568, opened 2017-09-12
support dask arrays in rolling computations using bottleneck functions

### LINKED ISSUE #1279: Rolling window operation does not work with dask arrays
As the title says :-)

This would be very useful to downsample long time series read from multiple consecutive netcdf files.

Note that I was able to apply the rolling window by converting my variable to a pandas series with `to_series()`. I then could use panda's own rolling window methods. I guess that when converting to a pandas series the dask array is read in memory?

jhamman:
- [x] Closes #1279
 - [x] Tests added / passed

cc @shoyer, @darothen, @arbennett, @vnoel

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -10,7 +10,6 @@ from __future__ import division
 from __future__ import print_function
 
 import operator
-import warnings
 from distutils.version import LooseVersion
 
 import numpy as np
@@ -24,10 +23,7 @@ from .nputils import array_eq, array_ne
 try:
     import bottleneck as bn
     if LooseVersion(bn.__version__) < LooseVersion('1.1'):
-        warnings.warn(
-            "The installed version of bottleneck {ver} is not supported "
-            "in xarray and will be not be used\nThe minimum supported "
-            "version is 1.1\n".format(ver=bn.__version__), UserWarning)
+        # fall back to numpy
         raise ImportError(
             'bottleneck version {} is not supported'.format(bn.__version__))
     has_bottleneck = True
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -10,6 +10,7 @@ from __future__ import division
 from __future__ import print_function
 
 import operator
+import warnings
 from distutils.version import LooseVersion
 
 import numpy as np
@@ -23,7 +24,12 @@ from .nputils import array_eq, array_ne
 try:
     import bottleneck as bn
     if LooseVersion(bn.__version__) < LooseVersion('1.1'):
-        raise ImportError
+        warnings.warn(
+            "The installed version of bottleneck {ver} is not supported "
+            "in xarray and will be not be used\nThe minimum supported "
+            "version is 1.1\n".format(ver=bn.__version__), UserWarning)
+        raise ImportError(
+            'bottleneck version {} is not supported'.format(bn.__version__))
     has_bottleneck = True
 except ImportError:
     # use numpy methods instead
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -28,6 +28,11 @@ except ImportError:
     bn = np
     has_bottleneck = False
 
+try:
+    import dask.array as da
+except ImportError:
+    pass
+
 
 UNARY_OPS = ['neg', 'pos', 'abs', 'invert']
 CMP_BINARY_OPS = ['lt', 'le', 'ge', 'gt']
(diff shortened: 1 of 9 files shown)


xarray/core/ops.py
```diff
@@ -243,6 +248,42 @@ def rolling_count(rolling):
...
+    ag = da.ghost.ghost(a, depth=depth, boundary=boundary)
+    # apply rolling func
+    out = ag.map_blocks(moving_func, window, min_count=min_count,
+                        axis=axis, dtype=a.dtype)
+    # trim array
+    result = da.ghost.trim_internal(out, depth)
+    return result
+
+
+def dask_rolling_wrapper_without_min_count(moving_func, a, window, axis=-1):
```
shoyer: Can't we use the default parameter value rather than writing this twice?

xarray/core/ops.py
```diff
@@ -28,6 +28,11 @@
     bn = np
     has_bottleneck = False
 
+try:
+    import dask.array as da
```
shoyer: I moved he dask array related code to `duck_array_ops`. That's probably a more natural place for this.

Actually: maybe we should have another module  for dask array operations, and then put the duck type functions in `dask_array_ops`.

xarray/core/ops.py
jhamman: If we update the minimum support bottleneck version to 1.1, we can do away with this special case (also present for the non-dask case) all together.

shoyer:
That works for me. But note that you can also use **kwargs to insert
conditional keyword arguments.

jhamman:
I went ahead and changed the bottleneck version. It really cleans up the all the rolling methods and the testing so I think this is an overall nice change.

xarray/core/ops.py
```diff
@@ -22,6 +22,8 @@
 
 try:
     import bottleneck as bn
+    if LooseVersion(bn.__version__) < LooseVersion('1.1'):
+        raise ImportError
```
shoyer: can we add a descriptive error message here?

xarray/tests/test_dataarray.py
```diff
@@ -2698,14 +2696,23 @@ def test_rolling_wrapped_bottleneck(da, name, center, min_periods):
...
+    pytest.importorskip('dask.array')
+    pytest.importorskip('bottleneck', minversion="1.1")
+    # dask version
+    rolling_obj = da_dask.rolling(time=7, min_periods=min_periods)
+    actual = getattr(rolling_obj, name)().load()
+    # numpy version
+    rolling_obj = da_dask.load().rolling(time=7, min_periods=min_periods)
+    expected = getattr(rolling_obj, name)()
+
+    # using all-close becuase rolling over ghost cells introduces some
```
shoyer: because

xarray/tests/test_dataarray.py
```diff
@@ -2698,14 +2696,23 @@ def test_rolling_wrapped_bottleneck(da, name, center, min_periods):
...
-    with pytest.raises(ValueError) as exception:
-        da.rolling(time=7, min_periods=1).median()
-    assert 'Rolling.median does not' in str(exception)
+@pytest.mark.parametrize('name', ('sum', 'mean', 'std', 'min', 'max',
+                                  'median'))
+@pytest.mark.parametrize('center', (True, False, None))
+@pytest.mark.parametrize('min_periods', (1, None))
+def test_rolling_wrapped_bottleneck_dask(da_dask, name, center, min_periods):
+    pytest.importorskip('dask.array')
+    pytest.importorskip('bottleneck', minversion="1.1")
```
shoyer: shouldn't these tests now work regardless of whether bottleneck is installed? (obviously performance will be better with bottleneck)

xarray/core/ops.py
```diff
@@ -22,6 +23,13 @@
 
 try:
     import bottleneck as bn
+    if LooseVersion(bn.__version__) < LooseVersion('1.1'):
+        warnings.warn(
```
shoyer: Can we trigger this warning lazily (e.g., when calling actually resample)? If we do this at import time we are going to get a lot of complaints from users using bottleneck < 1.1 about irrelevant warnings.

jhamman:
> Can we trigger this warning lazily (e.g., when calling actually resample)? If we do this at import time we are going to get a lot of complaints from users using bottleneck < 1.1 about irrelevant warnings.

@shoyer - I'm not exactly sure how to do this consistently throughout the package. In d532a1f
 I removed the warning and we fall back to numpy silently. Here's some ideas of what we could do:

1. Raise an error/warning when an unsupported version of bottleneck is imported
2. Not check the version of bottleneck at all and live with some harder to predict behavior on `rolling.meadian()`
3. Silently fall back to numpy when an unsupported version of bottleneck is imported...and for rolling methods, raise a warning at either import (easy) or runtime (harder).

I'm actually leaning toward 2 right now. We don't enforce strict minimum versions on other dependencies so it seems like what we're trying to do may be a bit overkill.

shoyer:
> Not check the version of bottleneck at all and live with some harder to predict behavior on rolling.meadian()

Sounds good to me.

jhamman:
any final comments here?

PR #1587, opened 2017-09-23
Fix URL to dask example

cbarrick:
It seems that the continuum folks moved the bolg. They have a redirect at the old URL, but unfortunately it's broken. This appears to be the new url: https://www.anaconda.com/blog/developer-blog/xray-dask-out-core-labeled-arrays-python/


shoyer:
thanks!

PR #1590, opened 2017-09-24
Dataset.copy() to preserve encoding

### LINKED ISSUE #1586: Dataset.copy() drops encoding
```
ds = Dataset()
ds.encoding = {"unlimited_dims": 'x'}
ds.copy().encoding
{}
```
By looking at dataset.py, there's a lot of calls to ``Dataset._construct_direct`` that omit the encoding. Is it correct to add it in all cases?

crusaderky:
- [x] Closes #1586
 - [x] Tests added / passed


crusaderky:
Ready for merge if there's no objections

jhamman:
Pending the outcome of our conversation on #1586, this looks good to me.

xarray/tests/test_dataset.py
```diff
@@ -1434,6 +1435,7 @@ def test_copy(self):
 
         for copied in [data.copy(deep=False), copy(data)]:
             self.assertDatasetIdentical(data, copied)
+            self.assertEqual(data.encoding, copied.encoding)
```
jhamman: note on behalf of @MaximilianR -

this can just be `assert data.encoding == copied.encoding`

crusaderky: Doesn't assertEqual produce better output in case of failure?

MaximilianR: Pytest satisfies all your desires!

jhamman: @crusaderky - pytest has nice introspection for this sort of thing. https://docs.pytest.org/en/latest/assert.html#making-use-of-context-sensitive-comparisons

PR #1597, opened 2017-09-27
Add methods for combining variables of differing dimensionality

### LINKED ISSUE #1317: API for reshaping DataArrays as 2D "data matrices" for use in machine learning
Machine learning and linear algebra problems are often expressed in terms of operations on matrices rather than arrays of arbitrary dimension, and there is currently no convenient way to turn DataArrays  (or combinations of DataArrays) into a single "data matrix".

As an example, I have needed to use scikit-learn lately with data from DataArray objects. Scikit-learn requires the data to be expressed in terms of simple 2-dimensional matrices. The rows are called samples, and the columns are known as features. It is annoying and error to transpose and reshape a data array by hand to fit into this format. For instance, this [gituhub repo for xarray aware sklearn-like objects](https://github.com/wy2136/xlearn) devotes many lines of code to massaging data arrays into data matrices. I think that this reshaping workflow might be common enough to warrant some kind of treatment in xarray.

I have written some code in this [gist](https://gist.github.com/nbren12/46767d237e3b1e59f7e2e1165c1e72c5), that have found pretty convenient for doing this. This gist has an `XRReshaper` class which can be used for reshaping data to and from a matrix format. The basic usage for an EOF analysis of a dataset `A(lat, lon, time)` can be done like this
```python
feature_dims = ['lat', 'lon']

rs = XRReshaper(A)
data_matrix, _ = rs.to(feature_dims)

# Some linear algebra or machine learning
_,_, eofs = svd(data_matrix)

eofs_datarray = rs.get(eofs[0], ['mode'] + feature_dims)
```

I am not sure this is the best API, but it seems to work pretty well and I have used it [here](https://github.com/nbren12/gnl/blob/master/gnl/xlearn.py) to implement some xarray-aware sklearn-like objects for PCA, which can be used like
```
feature_dims = ['lat', 'lon']
pca = XPCA(feature_dims, n_components=10, weight=cos(A.lat))
pca.fit(A)
pca.transform(A)
eofs = pca.components_
```

Another syntax which might be helpful is some kind of context manager approach like
```python
with XRReshaper(A) as rs, data_matrix:
     # do some stuff with data_matrix
# use rs to restore output to a data array.
```

nbren12:
- [x] Closes #1317 
 - [x] Tests added / passed

While working on #1317, I settled upon combining `stack` and `to_array` to create two dimensional numpy arrays given an xarray Dataset. Unfortunately, `to_array` automatically broadcasts the variables of dataset, which is not always a desirable behavior.  For instance, I was trying to combine precipitation (a horizontal field) and temperature (a 3D field) into one array. 

This PR enables this by adding two new methods to xarray:
- `Dataset.stack_cat`, and
- `DataArray.unstack_cat`.

`stack_cat`  uses  `stack`, `expand_dims`, and `concat` to reshape a Dataset into a Dataarray with a helpful MultiIndex, and `unstack_cat` reverses the process.

I implemented this functionality as a new method since `to_array` is such a clean method already. I really appreciate your thoughts on this. Thanks!

cc @jhamman @shoyer


jhamman:
I want to let @shoyer and @rabernat weigh in on the API here.

xarray/core/dataarray.py
```diff
@@ -1086,6 +1086,35 @@ def unstack(self, dim):
...
+        ----------
+        dim : str
+            Name of existing dimension to unstack
+        level : int
+            Index of level to expand to dataset along
+
+        Returns
+        -------
+        unstacked: Dataset
+
```
shoyer: An example here would also be helpful, as well as a "See also" reference to `stack_cat`.

nbren12:
Thanks! Yah...I'm not very good at naming things. I think `Dataset.to_stacked_array` makes sense, but I would typically expect something like `from_stacked_array` to be a static method of `Dataset` (e.g. `pd.MultiIndex.from_tuples`). We could always move unstack_cat to `Dataset`, but then we miss out on the ability to sequential method calls. 

IMO the behavior of unstack_cat is kind of what I would expect `DataArray.to_dataset` to do in the case where `dim` is a stacked coordinate, so maybe we could put it in there.

nbren12:
Or maybe  `DataArray.to_dataset_unstacked` is good.

shoyer:
If we have a separate method for `to_stacked_array()`, we should have a separate method for unstacking, too.

I like `DataArray.to_unstacked_dataset()` as the complement to `Dataset.to_stacked_array()`. I agree that it should be a DataArray method for chain-ability.

nbren12:
That naming sounds good to me.

Also, I was having an issue with `unstack_cat` returning an index with dtype `object` that I would like to sort out.

nbren12:
Okay. I just changed the names of the methods and wrote a test case for the problem with the dtype of the stacked dimensions not being preserved by `to_stacked_array`.

At the moment, I am filling in the missing dimensions with None, so the resulting index has `dtype object'`. This is then concatenated with the bona fide indices of the variables which are not missing this index, so the index as a whole is cat to the lowest common denominator of `object`. Unfortunately we can't just put `Nan` in because, `NaN` is a floating point number thing, so I don't think all numpy dtypes have the equivalent of NaN. @shoyer Do you have any thoughts about how to resolve this?

xarray/core/dataset.py
```diff
@@ -1865,6 +1865,100 @@ def stack(self, **dimensions):
...
+
+            expand_dims = set(dims).difference(set(val.dims))
+            expand_dims.add('variable')
+            return val.assign_coords(**assign_coords) \
+                .expand_dims(expand_dims) \
+                .stack(**{new_dim: (variable_dim,) + dims})
+
+        # concatenate the arrays
+        Xs = [f(self[key]) for key in self.data_vars]
+        dataset = xr.concat(Xs, dim=new_dim)
```
jhamman: you should be able to pass the MultiIndex (`new_idx`) in here on the right side of the `dim` keyword.

xarray/core/dataarray.py
```diff
@@ -1086,6 +1086,35 @@ def unstack(self, dim):
...
+        """
+
+        idx = self.indexes[dim]
+        if not isinstance(idx, pd.MultiIndex):
+            raise ValueError(f"{dim} is not a stacked coordinate")
```
jhamman: this string formatting syntax will not work for Python versions < 3.6.

xarray/core/dataarray.py
```diff
@@ -1086,6 +1086,35 @@ def unstack(self, dim):
...
+        variables = idx.levels[level]
+
+        # pull variables out of datarray
+        data_dict = {}
```
jhamman: use an `OrderedDict` here.

xarray/core/dataset.py
```diff
@@ -1865,6 +1865,100 @@ def stack(self, **dimensions):
...
+        def f(val):
+            # ensure square output
+
+            assign_coords = {variable_dim: val.name}
+            for dim in dims:
+                if (dim not in val):
+                    assign_coords[dim] = None
+
+            expand_dims = set(dims).difference(set(val.dims))
+            expand_dims.add('variable')
```
jhamman: should this be: 

```Python
expand_dims.add(variable_dim)
```
?

xarray/core/dataarray.py
```diff
@@ -1086,6 +1086,35 @@ def unstack(self, dim):
...
+        for k in variables:
+            data_dict[k] = self.sel(variable=k).squeeze(drop=True)
```
jhamman: should we add a keyword argument `variable_dim` to match the `to_stacked_array` argument?
```Python
data_dict[k] = self.sel(**{variable_dim: k).squeeze(drop=True)
```

nbren12:
@rabernat Your point is well taken. I will add some docs/motivation to the reshaping page.

xarray/core/dataarray.py
nbren12: yes, definitely. good catch.

xarray/core/dataset.py
nbren12: indeed. fixed.

xarray/core/dataset.py
nbren12: the way the code is written right now, I am using the index generated by `xr.concat` to create `new_idx`. I guess it should be possible to use pandas calls to construct the MultiIndex, and then pass it to `xr.concat`.

jhamman:
@nbren12 - can you add these methods to `api.rst` and add a note to whats-new?

nbren12:
I just added some docs, but need to add to whats-new still,

nbren12:
Okay. I think I'm done with the updates to the documentation.

shoyer:
@nbren12 in the linked issue (https://github.com/pydata/xarray/issues/1317#issuecomment-337796691) you wrote:

> After using my own version of this code for the past month or so, it has occurred to me that this API probably will not support stacking arrays of with different sizes along shared arrays. For instance, I need to "stack" humidity below an altitude of 10km with temperature between 0 and 16 km. IMO, the easiest way to do this would be to change these methods into top-level functions which can take any dict or iterable of datarrays. We could leave that for a later PR of course.

Is this something that would make sense to resolve before we merge this?

nbren12:
@shoyer  If you are okay with it, I think we might want to leave that to a later date if ever. I am not exactly sure what a useful API for that would be ATM.  On the other hand, I have been using the original `stack_cat` and unstack_cat` functions for a couple of months, and they have handled my basic uses pretty well.

For more complicated uses (e.g. taking different subsets of each variable and concatenating the output), I have started working on [a project](https://github.com/nbren12/xarray_sklearn) which is similar to [sklearn-pandas](https://github.com/scikit-learn-contrib/sklearn-pandas). Since there are a million ways several xarray variables could be processed/subsetted, stacked and then concatenated, I think this functionality should probably remain in a third party package for now.

shoyer:
@nbren12 OK, sounds good to me. Yes, it's a good idea to put more speculative stuff in other packages. I will take a look at your actual implementation.

nbren12:
Sorry for random activity. I accidentally hard reset the master branch nbren12/xarray to pydata/xarray.

jhamman:
@nbren12 - can we revive this? There or some conflicts but I think this was pretty close to done.

xarray/tests/test_dataset.py
```diff
@@ -2104,6 +2104,65 @@ def test_stack_unstack(self):
...
+        ds_flat = ds.to_stacked_array('features', ['y'])
+        ds_comp = ds_flat.to_unstacked_dataset('features')
+        self.assertDatasetIdentical(ds, ds_comp)
+
+        # test for scalar variables
+        a = xr.DataArray(np.r_[:6], dims=('x'), coords={'x': np.r_[:6]})
+        ds = xr.Dataset({'a': a, 'b': 1.0})
+        ds_flat = ds.to_stacked_array('features', ['x'])
+        ds_comp = ds_flat.to_unstacked_dataset('features')
+        self.assertDatasetIdentical(ds, ds_comp)
```
jhamman: Can you break these into a bunch of smaller tests?

doc/reshaping.rst
```diff
@@ -134,6 +134,36 @@ pandas, it does not automatically drop missing values. Compare:
 We departed from pandas's behavior here because predictable shapes for new
 array dimensions is necessary for :ref:`dask`.
 
+Stacking different physical variables together
```
jhamman: nit: let's remove the word "physical" here.

xarray/tests/test_dataset.py
nbren12: Sure. I tried to split these tests into more coherent groups

doc/reshaping.rst
nbren12: done

shoyer:
> @jhamman I did a little more work on this today. How do you recommend I update this to master? rebase?

We should add a section to our contributor guide on this, but the recommended workflow is just to use `git merge`.

nbren12:
@shoyer Sorry. I rebased...

shoyer:
No worries, rebase is fine too :)

nbren12:
Okay. It looks like it is passing CI now, so I think it's ready for another look. How does it look? Are you still interested in including this functionality?

xarray/core/dataarray.py
jhamman: @nbren12 - we could still use an example here.

nbren12:
Ok. I added an example which is very similar to the `DataArray.unstack` example.

nbren12:
Hey. I just wanted to bump this PR. How does it look @rabernat @jhamman @shoyer?

doc/reshaping.rst
```diff
@@ -133,6 +133,36 @@ pandas, it does not automatically drop missing values. Compare:
...
+:py:meth:`xarray.DataArray.to_unstacked_dataset` reverses this operation. These
+methods are used like this:
+
+.. ipython:: python
+
+        arr = xr.DataArray(np.arange(6).reshape(2, 3),
+                        coords=[('x', ['a', 'b']), ('y', [0, 1, 2])])
+        data = xr.Dataset({'a': arr, 'b': arr.isel(y=0)})
+        stacked = data.to_stacked_array("z", ['y'])
+        stacked
```
rabernat: Do you want to call `print(stacked)` here? Otherwise I fear that the output won't be rendered in the docs.

xarray/core/dataset.py
```diff
@@ -2627,6 +2627,107 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+        Parameters
+        ----------
+        new_dim : str
+            Name of the new stacked coordinate
+        dims : Sequence[str]
+            Dimensions to be stacked. Not all variables in the dataset need to
+            have these dimensions.
+        variable_dim : str, optional
+            Name of the level in the MultiIndex object which corresponds to
+            the variables.
```
rabernat: I find this docstring confusing. What is "the Multindex object"? It is not obviously one of the arguments of the function.

doc/reshaping.rst
```diff
@@ -133,6 +133,36 @@ pandas, it does not automatically drop missing values. Compare:
...
+and ``unstack``, but combining multiple variables in a
+:py:class:`xarray.Dataset` is more complicated. If the variables in the dataset
+have matching numbers of dimensions, we can call
+:py:meth:`~xarray.Dataset.to_array` and then stack along the the new coordinate.
+But :py:meth:`~xarray.Dataset.to_array` will broadcast the dataarrays together,
+which will effectively tile the lower dimensional variable along the missing
+dimensions. The method :py:meth:`xarray.Dataset.to_stacked_array` allows
+combining variables of differing dimensions without this wasteful copying while
+:py:meth:`xarray.DataArray.to_unstacked_dataset` reverses this operation. These
+methods are used like this:
```
rabernat: Here I would mention explicitly that it does this by using a MultiIndex in the output. (Is that a correct interpretation?)

jhamman:
@rabernat - you had asked for some changes in the early days of this PR. Are you satisfied with the current state here?

xarray/core/dataarray.py
rabernat: Resolved below

xarray/core/dataset.py
```diff
@@ -2627,6 +2627,107 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+        idx = dataset.indexes[new_dim]
+        levels = [idx.levels[0]]\
+            + [level.astype(self[level.name].dtype)
+               for level in idx.levels[1:]]
+        new_idx = idx.set_levels(levels)
+        # patch in the new index object
+        # dataset[new_dim].variable._data.array = new_idx
+        # This commented line below is much cleaner than the junk above, but I
+        # wanted to modify the IndexVariable inplace to make sure the attrs
+        # and encodings are the same
```
rabernat: Is this comment still accurate?

doc/reshaping.rst
nbren12: I believe that it will be printed just as if I had typed `stacked<Enter>` in ipython. Most examples in the `xarray` docs seem to use this syntax.

nbren12: Indeed. I just compiled the docs and `stacked` is printed.

xarray/core/dataset.py
nbren12: I don't think so. I think it is cleaner to declare a new `IndexVariable` rather than modify the existing one in-place. I deleted the comment.

doc/reshaping.rst
nbren12: Done.

xarray/core/dataset.py
nbren12: Ah yes. This is a bit tricky, I doubt that most users would every touch this argument. I added some more description to the "Returns" section which hopefully describes it better though.

nbren12:
Hey @rabernat. Did I resolve your comments?

jhamman:
@pydata/xarray - would love to get some more eyes on this PR. Its been lingering but @nbren12 seems to have addressed all prior comments.

dcherian:
docs & tests LGTM. I don't understand stacking/unstacking/multiindexes too well...

doc/reshaping.rst
```diff
@@ -133,6 +133,38 @@ pandas, it does not automatically drop missing values. Compare:
...
+:py:meth:`xarray.DataArray.to_unstacked_dataset` reverses this operation.
+Just as with :py:meth:`xarray.Dataset.stack` the stacked coordinate is
+represented by a :py:class:`pandas.MultiIndex` object. These methods are used
+like this:
+
+.. ipython:: python
+
+        arr = xr.DataArray(np.arange(6).reshape(2, 3),
+                        coords=[('x', ['a', 'b']), ('y', [0, 1, 2])])
+        data = xr.Dataset({'a': arr, 'b': arr.isel(y=0)})
```
benbovy: I was a bit confused by the labels 'a' et 'b' used for both the names of the data variables and the 'x' coordinate labels in this example (and also in the docstrings). As the stack operation is rather complex here, maybe a slightly simpler example would help users better understand it? Something like:

```python
>>> data = xr.Dataset(
...     data_vars={'a': (('x', 'y'), [[0, 1, 2], [3, 4, 5]]),
...                'b': ('x', [6, 7])},
...     coords={'y': ['u', 'v', 'w']}
... )
<xarray.Dataset>
Dimensions:  (x: 2, y: 3)
Coordinates:
  * y        (y) <U1 'u' 'v' 'w'
Data variables:
    a        (x, y) int64 0 1 2 3 4 5
    b        (x) int64 6 7

>>> data.to_stacked_array('z', ['y'])
<xarray.DataArray (x: 2, z: 4)>
array([[0, 1, 2, 6],
       [3, 4, 5, 7]])
Coordinates:
  * z         (z) MultiIndex
  - variable  (z) object 'a' 'a' 'a' 'b'
  - y         (z) object 'u' 'v' 'w' nan
```

xarray/core/dataarray.py
```diff
@@ -1399,6 +1399,66 @@ def unstack(self, dim=None):
...
+        <xarray.Dataset>
+        Dimensions:  (x: 2, y: 3)
+        Coordinates:
+          * x        (x) <U1 'a' 'b'
+          * y        (y) int64 0 1 2
+        Data variables:
+            a        (x, y) int64 0 1 2 3 4 5
+            b        (x) int64 0 3
+        >>> stacked = data.to_stacked_array("z", ['y'])
+        >>> stacked.indexes['z']
```
benbovy: nit: I think it would be a little bit clearer to show the repr of `stacked` instead of its 'z' index here.

xarray/core/dataset.py
```diff
@@ -2643,6 +2643,107 @@ def stack(self, dimensions=None, **dimensions_kwargs):
             result = result._stack_once(dims, new_dim)
         return result
 
+    def to_stacked_array(self, new_dim, dims, variable_dim='variable'):
```
benbovy: Shouldn't we add a ``name`` keyword argument to be consistent with ``to_array``?

xarray/core/dataarray.py
```diff
@@ -1399,6 +1399,66 @@ def unstack(self, dim=None):
...
+        """Unstack DataArray expanding to Dataset along a given level of a
+        stacked coordinate.
+
+        This is the inverse operation of Dataset.to_stacked_array.
+
+        Parameters
+        ----------
+        dim : str
+            Name of existing dimension to unstack
+        level : int
```
benbovy: As we all love labels, it would be nice if `level` could also accept level names (`str`) :-). Replacing``idx.levels[level]`` with ``idx.get_level_values(level)`` below would be enough, I think.

xarray/core/dataarray.py
```diff
@@ -1399,6 +1399,66 @@ def unstack(self, dim=None):
...
+            Index of level to expand to dataset along
+
```
benbovy: ``variable_dim`` is missing in the parameters section.

nbren12:
Thanks for your review @benbovy. I'll try to take a look soon.

doc/reshaping.rst
0x0L: +1
A nice sketch in the doc would be illuminating

xarray/core/dataset.py
0x0L: Maybe we can also use the syntax `stacked = data.to_stacked_array(z=['y'])`

nbren12:
@benbovy 

> Is it expected behavior? I guess that not every combination is possible because we're not broadcasting here unlike to_array().

I believe so. In your third case, `to_stacked_array` basically wants to combine `a` and `b` into a DataArray with a dimension `y`. But that is not possible without broadcasting because `b` does not have a dimension `y`. I will try to add a more informative error message.

doc/reshaping.rst
nbren12: I am fine making this change. Not sure how to draw the sketch though.

nbren12: Okay. I used your example in `reshaping.rst` and in `to_stacked_array`.

xarray/core/dataarray.py
nbren12: Done

xarray/core/dataset.py
nbren12: Sure. added this in 23d9246e.

xarray/core/dataarray.py
nbren12: I agree it would be nice to do this. I would prefer not to overload the `level` argument, so maybe we a `label` argument could take precedence if given.

nbren12: It seems like `idx.get_level_values` has a potential problem. What happens if the labels are also integers?

nbren12: Fine...if it's good enough for pandas, it's good enough for us I guess. Updated the docs and use `idx.get_level_values(level)` as you sugget.

xarray/core/dataarray.py
nbren12: `variable_dim` is actually unnecessary since it can be inferred from `level`. I have removed it.

xarray/core/dataset.py
nbren12: While your suggestion is certainly more analogous to `stack`, I think it should be clear that this function only performs a single stacking operation. That is why I chose a more verbose signature.

nbren12:
@benbovy 
Paradoxically I think making `to_stacked_array` accept the list of dimensions which won't be stacked is clearer. These *unstacked* dims needed to shared across all variables, which is a very simple requirement. I realized this after writing a bunch of error-prone validation code. See my last commit for more of my reasoning (5ca9a1d868a). I also updated the docs and tests accordingly.

xarray/core/dataset.py
```diff
@@ -2648,6 +2648,119 @@ def stack(self, dimensions=None, **dimensions_kwargs):
             result = result._stack_once(dims, new_dim)
         return result
 
+    def to_stacked_array(self, new_dim, sample_dims, variable_dim='variable',
+                         name=None):
+        """Combine variables of differing dimensionality into a DataArray
+        without broadcasting.
+
+        This function is basically version of Dataset.to_array which does not
+        broadcast the variables.
```
mathause: I think there is a typo: 'a version'

maybe reformulate to
 This function is similar to Dataset.to_array but does not broadcast the variables.

nbren12: Okay. Changed to "This _method_ is similar to Dataset.to_array but does not broadcast variables."

nbren12:
Does anybody have an idea why the `py36-dask-env` tests are failing? None of the errors seems related to this PR.

nbren12:
It looks like 43834ac8186a851b7 might have fixed this. Let's see if the tests pass.

nbren12:
It looks like the tests passed. @benbovy How does it look now? Did I fix the issues you mentioned?

nbren12:
>  is there actual use cases where it is useful to provide multiple dimensions for sample_dims? 

Yes! My main use-case does. For example, if you have a weather dataset where every lat/lon pair should be considered a separate "sample", you would use `sample_dims=['lat', 'lon']`.

xarray/core/dataset.py
```diff
@@ -2749,6 +2757,10 @@ def f(val):
                for level in idx.levels[1:]]
         new_idx = idx.set_levels(levels)
         dataset[new_dim] = IndexVariable(new_dim, new_idx)
+
+        if name is not None:
+            dataset.name = name
+
         return dataset
```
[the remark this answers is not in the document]
nbren12: I changed this

xarray/core/dataset.py
```diff
@@ -2685,56 +2689,60 @@ def to_stacked_array(self, new_dim, dims, variable_dim='variable'):
...
-        dims = tuple(dims)
+        stacking_dims = tuple(dim for dim in self.dims
+                              if dim not in sample_dims)
+
+        for variable in self:
+            dims = self[variable].dims
+            dims_include_sample_dims = set(sample_dims) <= set(dims)
+            if not dims_include_sample_dims:
+                raise ValueError(
+                    "All DataArrays must share the dims: {}. ".format(dims)
```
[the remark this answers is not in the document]
nbren12: I can see how that is confusing, but technically it is true. I think it would be a little unwieldy to have differerent error messages for different numbers of sample_dimensions. I change the message to "All variables in the dataset must contain the dimensions {}." Hopefully, that is better.

nbren12:
Okay. I responded to @benbovy's last comments and merged the upstream changes to master. How is this looking? 

cc @jhamman

xarray/core/dataarray.py
```diff
@@ -1540,6 +1540,72 @@ def unstack(self, dim: Union[Hashable, Sequence[Hashable], None] = None
...
+        True
+
+        See Also
+        --------
+        Dataset.to_stacked_array
+        """
+
+        idx = self.indexes[dim]
+        if not isinstance(idx, pd.MultiIndex):
+            raise ValueError(dim, "is not a stacked coordinate")
```
shoyer: please add test coverage for this error

xarray/core/dataset.py
```diff
@@ -2698,6 +2698,120 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+        for variable in self:
+            dims = self[variable].dims
+            dims_include_sample_dims = set(sample_dims) <= set(dims)
+            if not dims_include_sample_dims:
+                raise ValueError(
+                    "All variables in the dataset must contain the "
+                    "dimensions {}.".format(dims)
+                )
+
+        def f(val):
```
shoyer: please give this as sensible name rather than `f`, e.g., `ensure_stacked`

xarray/core/dataset.py
```diff
@@ -2698,6 +2698,120 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+            for dim in stacking_dims:
+                if (dim not in val.dims):
+                    assign_coords[dim] = None
+
+            expand_dims = set(stacking_dims).difference(set(val.dims))
+            expand_dims.add(variable_dim)
+            # must be list for .expand_dims
+            expand_dims = list(expand_dims)
+
+            return val.assign_coords(**assign_coords) \
```
shoyer: nit: here and below, per PEP8, prefer using parentheses rather than `\` for multi-line expressions, e.g.,
```python
            return (val.assign_coords(**assign_coords)
                    .expand_dims(expand_dims)
                    .stack(**{new_dim: (variable_dim,) + stacking_dims}))

File: doc/whats-new.rst [hunk 24]
```diff
@@ -189,6 +189,8 @@ Enhancements
 - Allow ``expand_dims`` method to support inserting/broadcasting dimensions
   with size > 1. (:issue:`2710`)
   By `Martin Pletcher <https://github.com/pletchm>`_.
+- New methods for reshaping Datasets of variables with different dimensions
```
shoyer: this needs to move up to 0.12.3 now -- sorry for the churn here!

xarray/core/dataset.py
```diff
@@ -2698,6 +2698,120 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+        <xarray.Dataset>
+        Dimensions:  (x: 2, y: 3)
+        Coordinates:
+        * y        (y) <U1 'u' 'v' 'w'
+        Dimensions without coordinates: x
+        Data variables:
+            a        (x, y) int64 0 1 2 3 4 5
+            b        (x) int64 6 7
+
+        >>> data.to_stacked_array("z", ['x'])
```
shoyer: I think this example would be clearer with keyword arguments:
```suggestion
        >>> data.to_stacked_array(new_dim='z', sample_dims=['x'])
```

xarray/core/dataset.py
```diff
@@ -2698,6 +2698,120 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+                    "All variables in the dataset must contain the "
+                    "dimensions {}.".format(dims)
+                )
+
+        def f(val):
+            # ensure square output
+
+            assign_coords = {variable_dim: val.name}
+            for dim in stacking_dims:
+                if (dim not in val.dims):
```
shoyer:
```suggestion
                if dim not in val.dims:
```

xarray/core/dataarray.py
nbren12: sure. done

xarray/core/dataset.py
nbren12: Sure. I changed it to `ensure_stackable`. since the arrays aren't stacked yet.

xarray/core/dataset.py
nbren12: Sure. done.

doc/whats-new.rst
nbren12: Sure. I moved this up to that release and added a new section header
```
v0.12.3 (unreleased)
--------------------

New functions/methods
~~~~~~~~~~~~~~~~~~~~~

- New methods for reshaping Datasets of variables with different dimensions
  (:issue:`1317`). By `Noah Brenowitz <https://github.com/nbren12>`_.

```

xarray/core/dataset.py
nbren12: Added.

xarray/core/dataset.py
nbren12: Removed the parenthesis

shoyer:
you have a lint error: https://github.com/pydata/xarray/pull/1597#issuecomment-478700651

nbren12:
drat. Should be fixed now.

shoyer:
OK, I plan to merge this shortly unless there are any objections :)

xarray/core/dataarray.py
```diff
@@ -1540,6 +1540,72 @@ def unstack(self, dim: Union[Hashable, Sequence[Hashable], None] = None
...
+        This is the inverse operation of Dataset.to_stacked_array.
+
+        Parameters
+        ----------
+        dim : str
+            Name of existing dimension to unstack
+        level : int or str
+            The MultiIndex level to expand to a dataset along. Can either be
+            the integer index of the level or its name.
+        label : int, optional
```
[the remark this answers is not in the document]
nbren12: Okay. I changed this to your suggestion

xarray/core/dataarray.py
```diff
@@ -1540,6 +1540,72 @@ def unstack(self, dim: Union[Hashable, Sequence[Hashable], None] = None
...
+            raise ValueError("'{}' is not a stacked coordinate".format(dim))
+
+        level_number = idx._get_level_number(level)
+        variables = idx.levels[level_number]
+        variable_dim = idx.names[level_number]
+
+        # pull variables out of datarray
+        data_dict = OrderedDict()
+        for k in variables:
+            data_dict[k] = self.sel(**{variable_dim: k}).squeeze(drop=True)
```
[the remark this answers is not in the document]
nbren12: I have made this change.

xarray/core/dataset.py
```diff
@@ -2698,6 +2698,119 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+                    assign_coords[dim] = None
+
+            expand_dims = set(stacking_dims).difference(set(val.dims))
+            expand_dims.add(variable_dim)
+            # must be list for .expand_dims
+            expand_dims = list(expand_dims)
+
+            return (val.assign_coords(**assign_coords)
+                    .expand_dims(expand_dims)
+                    .stack(**{new_dim: (variable_dim,) + stacking_dims}))
```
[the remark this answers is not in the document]
nbren12: I made this change too. I think re-working the test suite would be a fair amount of work, and potentially make the tests harder to read. There are many places in xarray tests that use hardcoded names for dimensions and variables.

nbren12:
It looks like the CI errors above aren’t related to this PR. There seems to be an issue with `to_pandas`.

shoyer:
Thanks @nbren12!

nbren12:
phew! Thanks for all the reviews and discussion everyone!

xarray/core/dataset.py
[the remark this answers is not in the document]
max-sixty: Right, tbc the ideation was for broader than this PR! Wasn't expecting any big change now

jhamman:
Kudos to @nbren12 for sticking with this one! Looking forward to seeing this one in the wild.

nbren12:
Thanks Joe!

PR #1609, opened 2017-10-04
fix to_netcdf append bug (GH1215)

### LINKED ISSUE #1215: to_netcdf() fails to append to an existing file
The following code used to work well in v0.8.2:

```python
import os
import xarray as xr

path = 'test.nc'
if os.path.exists(path):
    os.remove(path)
    
ds = xr.Dataset()
ds['dim'] = ('dim', [0, 1, 2])
ds['var1'] = ('dim', [10, 11, 12])
ds.to_netcdf(path)

ds = xr.Dataset()
ds['dim'] = ('dim', [0, 1, 2])
ds['var2'] = ('dim', [10, 11, 12])
ds.to_netcdf(path, 'a')
```

On master, it fails with:

```
<ipython-input-1-fce5f5e876aa> in <module>()
     14 ds['dim'] = ('dim', [0, 1, 2])
     15 ds['var2'] = ('dim', [10, 11, 12])
---> 16 ds.to_netcdf(path, 'a')

/home/mowglie/Documents/git/xarray/xarray/core/dataset.py in to_netcdf(self, path, mode, format, group, engine, encoding)
    927         from ..backends.api import to_netcdf
    928         return to_netcdf(self, path, mode, format=format, group=group,
--> 929                          engine=engine, encoding=encoding)
    930 
    931     def __unicode__(self):

/home/mowglie/Documents/git/xarray/xarray/backends/api.py in to_netcdf(dataset, path, mode, format, group, engine, writer, encoding)
    563     store = store_cls(path, mode, format, group, writer)
    564     try:
--> 565         dataset.dump_to_store(store, sync=sync, encoding=encoding)
    566         if isinstance(path, BytesIO):
    567             return path.getvalue()

/home/mowglie/Documents/git/xarray/xarray/core/dataset.py in dump_to_store(self, store, encoder, sync, encoding)
    873             variables, attrs = encoder(variables, attrs)
    874 
--> 875         store.store(variables, attrs, check_encoding)
    876         if sync:
    877             store.sync()

/home/mowglie/Documents/git/xarray/xarray/backends/common.py in store(self, variables, attributes, check_encoding_set)
    219         cf_variables, cf_attrs = cf_encoder(variables, attributes)
    220         AbstractWritableDataStore.store(self, cf_variables, cf_attrs,
--> 221                                         check_encoding_set)
    222 
    223 

/home/mowglie/Documents/git/xarray/xarray/backends/common.py in store(self, variables, attributes, check_encoding_set)
    194     def store(self, variables, attributes, check_encoding_set=frozenset()):
    195         self.set_attributes(attributes)
--> 196         self.set_variables(variables, check_encoding_set)
    197 
    198     def set_attributes(self, attributes):

/home/mowglie/Documents/git/xarray/xarray/backends/common.py in set_variables(self, variables, check_encoding_set)
    204             name = _encode_variable_name(vn)
    205             check = vn in check_encoding_set
--> 206             target, source = self.prepare_variable(name, v, check)
    207             self.writer.add(source, target)
    208 

/home/mowglie/Documents/git/xarray/xarray/backends/netCDF4_.py in prepare_variable(self, name, variable, check_encoding)
    293             endian='native',
    294             least_significant_digit=encoding.get('least_significant_digit'),
--> 295             fill_value=fill_value)
    296         nc4_var.set_auto_maskandscale(False)
    297 

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Dataset.createVariable (netCDF4/_netCDF4.c:18740)()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Variable.__init__ (netCDF4/_netCDF4.c:30713)()

RuntimeError: NetCDF: String match to name in use
```

jhamman:
- [x] Closes #1215
 - [x] Tests added / passed

~TODO: 
Additional tests needed to verify this works for all the write backends and that the correct errors are raised when dims are different between writes.~

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -176,6 +176,10 @@ for dealing with datasets too big to fit into memory. Instead, xarray integrates
 with dask.array (see :ref:`dask`), which provides a fully featured engine for
 streaming computation.
 
+It is possible to append or overwrite netCDF variables using the ``mode='a'``
+argument. When using this option, all variables in the dataset will be written
+to the netCDF file, regardless if they exist in the original dataset.
+
 .. _io.encoding:
 
 Reading encoded data
@@ -390,7 +394,7 @@ over the network until we look at particular values:
 
 Some servers require authentication before we can access the data. For this
 purpose we can explicitly create a :py:class:`~xarray.backends.PydapDataStore`
-and pass in a `Requests`__ session object. For example for 
+and pass in a `Requests`__ session object. For example for
 HTTP Basic authentication::
 
     import xarray as xr
@@ -403,7 +407,7 @@ HTTP Basic authentication::
                                             session=session)
     ds = xr.open_dataset(store)
 
-`Pydap's cas module`__ has functions that generate custom sessions for 
+`Pydap's cas module`__ has functions that generate custom sessions for
 servers that use CAS single sign-on. For example, to connect to servers
 that require NASA's URS authentication::
 
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -178,7 +178,7 @@ streaming computation.
 
 It is possible to append or overwrite netCDF variables using the ``mode='a'``
 argument. When using this option, all variables in the dataset will be written
-to the netCDF file, regardless if they exist in the original dataset.
+to the original netCDF file, regardless if they exist in the original dataset.
 
 .. _io.encoding:
 
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -223,8 +223,7 @@ class AbstractWritableDataStore(AbstractDataStore):
         for vn, v in iteritems(variables):
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
-            if (vn not in self.variables or
-                    (getattr(self, '_mode', False) != 'a')):
+            if vn not in self.variables:
                 target, source = self.prepare_variable(
                     name, v, check, unlimited_dims=unlimited_dims)
             else:
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -221,6 +221,8 @@ class AbstractWritableDataStore(AbstractDataStore):
     def set_variables(self, variables, check_encoding_set,
                       unlimited_dims=None):
         for vn, v in iteritems(variables):
+            if self._mode == 'a' and vn in self.variables:
+                continue
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
             target, source = self.prepare_variable(
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -221,7 +221,7 @@ class AbstractWritableDataStore(AbstractDataStore):
     def set_variables(self, variables, check_encoding_set,
                       unlimited_dims=None):
         for vn, v in iteritems(variables):
-            if self._mode == 'a' and vn in self.variables:
+            if getattr(self, '_mode', False) == 'a' and vn in self.variables:
                 continue
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -221,12 +221,14 @@ class AbstractWritableDataStore(AbstractDataStore):
     def set_variables(self, variables, check_encoding_set,
                       unlimited_dims=None):
         for vn, v in iteritems(variables):
-            if getattr(self, '_mode', False) == 'a' and vn in self.variables:
-                continue
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
-            target, source = self.prepare_variable(
-                name, v, check, unlimited_dims=unlimited_dims)
+            if (vn not in self.variables or (getattr(self, '_mode', False) != 'a')):
+                target, source = self.prepare_variable(
+                    name, v, check, unlimited_dims=unlimited_dims)
+            else:
+                target, source = self.ds[name], v.data
+
             self.writer.add(source, target)
 
     def set_necessary_dimensions(self, variable, unlimited_dims=None):
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -223,11 +223,12 @@ class AbstractWritableDataStore(AbstractDataStore):
         for vn, v in iteritems(variables):
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
-            if (vn not in self.variables or (getattr(self, '_mode', False) != 'a')):
+            if (vn not in self.variables or
+                    (getattr(self, '_mode', False) != 'a')):
                 target, source = self.prepare_variable(
                     name, v, check, unlimited_dims=unlimited_dims)
             else:
-                target, source = self.ds[name], v.data
+                target, source = self.ds.variables[name], v.data
 
             self.writer.add(source, target)
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -964,7 +964,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         mode : {'w', 'a'}, optional
             Write ('w') or append ('a') mode. If mode='w', any existing file at
             this location will be overwritten. If mode='a', exisitng variables
-            will be over written.
+            will be overwritten.
         format : {'NETCDF4', 'NETCDF4_CLASSIC', 'NETCDF3_64BIT', 'NETCDF3_CLASSIC'}, optional
             File format for the resulting netCDF file:
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -974,7 +974,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
             default format becomes NETCDF3_64BIT).
         mode : {'w', 'a'}, optional
             Write ('w') or append ('a') mode. If mode='w', any existing file at
-            this location will be overwritten. If mode='a', exisitng variables
+            this location will be overwritten. If mode='a', existing variables
             will be overwritten.
         format : {'NETCDF4', 'NETCDF4_CLASSIC', 'NETCDF3_64BIT', 'NETCDF3_CLASSIC'}, optional
             File format for the resulting netCDF file:
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -963,7 +963,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
             default format becomes NETCDF3_64BIT).
         mode : {'w', 'a'}, optional
             Write ('w') or append ('a') mode. If mode='w', any existing file at
-            this location will be overwritten.
+            this location will be overwritten. If mode='a', exisitng variables
+            will be over written.
         format : {'NETCDF4', 'NETCDF4_CLASSIC', 'NETCDF3_64BIT', 'NETCDF3_CLASSIC'}, optional
             File format for the resulting netCDF file:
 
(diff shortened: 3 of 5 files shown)


fmaussion:
Thanks @jhamman !

With this fix existing variables will silently be ignored and won't be written, right? Maybe the expected behaviour (or specs) of the "append" option should be written somewhere in the docs and tested to avoid future regressions like we had in https://github.com/pydata/xarray/issues/1215

jhamman:
@fmaussion - you bring up a good point. There are two scenarios here.

1) appending to a file with existing data variables
2) appending to a file with existing coordinate variables

I'm wondering if we should disallow 1, in favor of being more explicit. So:

```Python
list_of_vars_to_append = ['var1', 'var2']
ds[list_of_vars_to_append].to_netcdf(filename, mode='a')
# if either var1 or var2 are in filename, raise an error?
```
as apposed to the current behavior which would silently skip all vars already in `filename` and not in  `list_of_vars_to_append`.

fmaussion:
Netcdf4 would overwrite in this situation, and I am also in favor of overwriting as this could be quite a useful usecase:

```python
from netCDF4 import Dataset
with Dataset('test.nc', 'w', format='NETCDF4') as nc:
    nc.createDimension('lat', 1)
    nc.createVariable('lat', 'f4', ('lat',))
    nc['lat'][:] = 1
with Dataset("test.nc", "a") as nc:
    nc['lat'][:] = 2
```

jhamman:
@fmaussion - I've updated the append logic slightly. I'm wondering what you think? This version more aggressively overwrites existing variables (data_vars and coords).

fmaussion:
Thanks! I like it: simple and in accordance with netcdf4's silent overwriting. It would be cool to describe this behavior this in the documentation somewhere, and add a test that the data is correctly overwritten maybe?

jhamman:
Question for those who are familiar with the `scipy` backend. I have a few failing tests here on the scipy backend and I'm not really sure what's going on. It seems like it could be related to the `mmap` feature in scipy.

shoyer:
@jhamman I think this is something to do with string/bytes. The data gets written as Unicode strings but then read back in as bytes, which is obviously not ideal.

If you want to ignore this, like the current tests, you can use `assert_allclose()` which has a flag for ignoring string/bytes issues (yeah, I know it's a nasty hack). In the long run we need to figure out how to solve issues like https://github.com/pydata/xarray/issues/1638. It seems like https://github.com/Unidata/netcdf-c/issues/402 is likely relevant.

shoyer:
Reading over https://github.com/Unidata/netcdf-c/issues/402, it seems like we should probably copy the handling of `_Encoding` from netcdf4-python (https://github.com/Unidata/netcdf4-python/pull/665) to our scipy interface. That would solve our problem of faithfully round-tripping data.

doc/io.rst
```diff
@@ -176,6 +176,10 @@ for dealing with datasets too big to fit into memory. Instead, xarray integrates
 with dask.array (see :ref:`dask`), which provides a fully featured engine for
 streaming computation.
 
+It is possible to append or overwrite netCDF variables using the ``mode='a'``
+argument. When using this option, all variables in the dataset will be written
+to the netCDF file, regardless if they exist in the original dataset.
```
fmaussion: for clarity: ``in the original file``

xarray/core/dataset.py
```diff
@@ -963,7 +963,8 @@ def to_netcdf(self, path=None, mode='w', format=None, group=None,
             default format becomes NETCDF3_64BIT).
         mode : {'w', 'a'}, optional
             Write ('w') or append ('a') mode. If mode='w', any existing file at
-            this location will be overwritten.
+            this location will be overwritten. If mode='a', exisitng variables
+            will be over written.
```
fmaussion: overwritten in one word?

xarray/tests/test_backends.py
```diff
@@ -957,6 +988,37 @@ def roundtrip(self, data, save_kwargs={}, open_kwargs={},
...
+
+    def test_append_overwrite_values(self):
+        data = create_test_data()
+        with create_tmp_file(allow_cleanup_failure=False) as tmp_file:
+            data.to_netcdf(tmp_file, mode='w', engine='scipy')
+            data['var2'][:] = -999
+            data['var9'] = data['var2'] * 3
+            data[['var2', 'var9']].to_netcdf(tmp_file, mode='a',
+                                             engine='scipy')
+            actual = open_dataset(tmp_file, engine='scipy')
```
shoyer: This is triggering a test failure on Windows, where you apparently can't open the same file twice.

xarray/tests/test_backends.py
```diff
@@ -823,6 +823,37 @@ def roundtrip(self, data, save_kwargs={}, open_kwargs={},
...
+
+    def test_append_overwrite_values(self):
+        data = create_test_data()
+        with create_tmp_file(allow_cleanup_failure=False) as tmp_file:
+            data.to_netcdf(tmp_file, mode='w', engine='netcdf4')
+            data['var2'][:] = -999
+            data['var9'] = data['var2'] * 3
+            data[['var2', 'var9']].to_netcdf(tmp_file, mode='a',
+                                             engine='netcdf4')
+            actual = open_dataset(tmp_file, autoclose=self.autoclose,
```
shoyer: Rather than allowing a clean-up failure, please close this file if possible (use a context manager). File descriptors are still a limited resource for our test suite.

xarray/tests/test_backends.py
```diff
@@ -823,6 +823,37 @@ def roundtrip(self, data, save_kwargs={}, open_kwargs={},
                               autoclose=self.autoclose, **open_kwargs) as ds:
                 yield ds
 
+    @contextlib.contextmanager
+    def roundtrip_append(self, data, save_kwargs={}, open_kwargs={},
```
shoyer: Can we put mode of these tests in one of the base classes, so we don't separately repeat them for scipy, netcdf4 and h5netcdf?

xarray/backends/common.py
```diff
@@ -223,8 +223,13 @@ def set_variables(self, variables, check_encoding_set,
         for vn, v in iteritems(variables):
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
-            target, source = self.prepare_variable(
-                name, v, check, unlimited_dims=unlimited_dims)
+            if (vn not in self.variables or
+                    (getattr(self, '_mode', False) != 'a')):
```
shoyer: I don't think we need the explicit check here for the mode here. With `mode='w'`, the file should already be starting from scratch.

xarray/tests/test_backends.py
```diff
@@ -563,6 +563,35 @@ def test_encoding_same_dtype(self):
...
+                         allow_cleanup_failure=False):
+        with create_tmp_file(
+                allow_cleanup_failure=allow_cleanup_failure) as tmp_file:
+            for i, key in enumerate(data.variables):
+                mode = 'a' if i > 0 else 'w'
+                data[[key]].to_netcdf(tmp_file, mode=mode, **save_kwargs)
+            with open_dataset(tmp_file,
+                              autoclose=self.autoclose, **open_kwargs) as ds:
```
jhamman: @shoyer - I've put these tests in a base class now. I've never quite understood how these tests work though. Are the inherited classes passing open/save_kwargs that specify the engine? How does this work?

xarray/tests/test_backends.py
```diff
@@ -563,6 +563,35 @@ def test_encoding_same_dtype(self):
...
+    def test_append_overwrite_values(self):
+        # regression for GH1215
+        data = create_test_data()
+        with create_tmp_file(allow_cleanup_failure=False) as tmp_file:
+            data.to_netcdf(tmp_file, mode='w')
+            data['var2'][:] = -999
+            data['var9'] = data['var2'] * 3
+            data[['var2', 'var9']].to_netcdf(tmp_file, mode='a')
+            with open_dataset(tmp_file, autoclose=self.autoclose) as actual:
+                assert_identical(data, actual)
```
jhamman: Regarding the issue with windows. What's the best way to do this then? Is it possible to explicitly close a netCDF file written with `to_netcdf`?

xarray/tests/test_backends.py
shoyer: Yes, that's exactly right.

One day it would be nice to split apart the gigantic `test_backends.py` into separate files for each backend...

xarray/tests/test_backends.py
shoyer: `to_netcdf()` *should* write and close the file. You only need to use the context manager to ensure things get closed with `open_dataset()`.

jhamman:
@shoyer - ready for final review.

xarray/core/dataset.py
```diff
@@ -974,7 +974,8 @@ def to_netcdf(self, path=None, mode='w', format=None, group=None,
             default format becomes NETCDF3_64BIT).
         mode : {'w', 'a'}, optional
             Write ('w') or append ('a') mode. If mode='w', any existing file at
-            this location will be overwritten.
+            this location will be overwritten. If mode='a', exisitng variables
```
shoyer: exisitng -> existing

xarray/tests/test_backends.py
```diff
@@ -563,6 +563,35 @@ def test_encoding_same_dtype(self):
...
         self.assertEqual(ds.x.encoding, {})
 
+    @contextlib.contextmanager
+    def roundtrip_append(self, data, save_kwargs={}, open_kwargs={},
+                         allow_cleanup_failure=False):
+        with create_tmp_file(
+                allow_cleanup_failure=allow_cleanup_failure) as tmp_file:
+            for i, key in enumerate(data.variables):
+                mode = 'a' if i > 0 else 'w'
+                data[[key]].to_netcdf(tmp_file, mode=mode, **save_kwargs)
```
shoyer: This is currently always using the default backend. You need to explicitly set `engine` here somewhere, though much of the logic can potentially live on the base class or helper function of some sort. There's no magic that sets this automatically for subclasses.

Take a look at how `roundtrip()` looks. I think I led you astray by suggesting that you move everything to the baseclass.

xarray/tests/test_backends.py
```diff
@@ -592,6 +627,8 @@ def create_tmp_files(nfiles, suffix='.nc', allow_cleanup_failure=False):
 
 @requires_netCDF4
 class BaseNetCDF4Test(CFEncodedDataTest):
+    engine = 'netcdf4'
+
```
jhamman: @shoyer, what do you think about this solution?

xarray/tests/test_backends.py
```diff
@@ -1024,6 +1030,8 @@ class ScipyFilePathTestAutocloseTrue(ScipyFilePathTest):
 
 @requires_netCDF4
 class NetCDF3ViaNetCDF4DataTest(CFEncodedDataTest, Only32BitTypes, TestCase):
+    engine = 'netcdf4'
```
shoyer: This isn't quite enough -- you also need to set the `format` in this case. Take a look at `roundtrip()` below.

Probably the cleanest fix would be refactor `roundtrip()` into three methods:

```python
     # on the base class
     @contextlib.contextmanager
     def roundtrip(self, data, save_kwargs={}, open_kwargs={},
                   allow_cleanup_failure=False):
         with create_tmp_file(
                 allow_cleanup_failure=allow_cleanup_failure) as path:
             self.save(data, path, **save_kwargs)
             with self.open(path, **open_kwargs) as ds:
                 yield ds

     # on subclasses, e.g., for NetCDF3ViaNetCDF4DataTest
     def save(self, dataset, path, **kwargs):
         dataset.to_netcdf(tmp_file, format='NETCDF3_CLASSIC',
                           engine='netcdf4', **kwargs)

     @contextlib.contextmanager
     def open(self, path, **kwargs):
          with open_dataset(tmp_file, engine='netcdf4',
                            autoclose=self.autoclose, **open_kwargs) as ds:
               yield ds
```

Then you could write `roundtrip_append()` in terms of `save` and `open`.

jhamman:
@shoyer - take another look. I have basically merged our two ideas and refactored the roundtrip tests. Tests are still failing but not for py2.7, on appveyor, or py3.6 locally.

shoyer:
@jhamman This looks great, thank you!

PR #1610, opened 2017-10-04
Fix the test suite on dask 0.15.3

shoyer:
Our test suite is currently failing due to https://github.com/dask/dask/issues/2718. This marks the failure as expected, which will make our CI a reliable indicator again.

### DIFF
--- a/xarray/tests/test_dask.py
+++ b/xarray/tests/test_dask.py
@@ -479,8 +479,8 @@ class TestDataArrayAndDataset(DaskTestCase):
         self.assertLazyAndIdentical(self.lazy_array, a)
 
 
-@requires_dask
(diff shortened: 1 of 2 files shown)


xarray/tests/test_dask.py
```diff
@@ -473,7 +476,6 @@ def test_from_dask_variable(self):
         self.assertLazyAndIdentical(self.lazy_array, a)
 
 
-@requires_dask
 @pytest.mark.parametrize("method", ['load', 'compute'])
```
shoyer: @jhamman For future reference, it looks like `pytest.mark.parametrize` doesn't play well with the unittest based `skiptest` decorator. So I switched the import check to use something different for this module, but it's something we'll have to watch out for in the future, too.

PR #1616, opened 2017-10-10
Default to join='exact' for apply_ufunc

shoyer:
This is safer default behavior than an inner join.

 - [x] Tests added / passed

### DIFF
--- a/xarray/tests/test_computation.py
+++ b/xarray/tests/test_computation.py
@@ -522,12 +522,12 @@ def test_dataset_join():
     with raises_regex(ValueError, 'data variable names'):
         apply_ufunc(operator.add, ds0, xr.Dataset({'b': 1}))
 
-    # we use np.nan as a fill_value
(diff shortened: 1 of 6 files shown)


xarray/tests/test_computation.py
```diff
@@ -517,7 +519,10 @@ def add(a, b, join, dataset_join):
...
     assert_identical(actual, expected)
 
-    # if variables don't match, join will perform add with np.nan
+    with raises_regex(ValueError, 'data variable names'):
+        apply_ufunc(operator.add, ds0, xr.Dataset({'b': 1}))
+
+    # we use np.nan as a fill_value
     ds2 = xr.Dataset({'b': ('x', [99, 3]), 'x': [1, 2]})
     actual = add(ds0, ds2, 'outer', 'inner')
     expected = xr.Dataset({'x': [0, 1, 2]})
```
jhamman: can you update this comment?  Its not immediately clear why/where `np.nan` is used as a fill value.

Issue #1618, opened 2017-10-10
apply_raw() for a simpler version of apply_ufunc()


shoyer:
`apply_raw()` would work like `apply_ufunc()`, but without the hard to understand broadcasting behavior and core dimensions.

The rule for `apply_raw()` would be that it directly unwraps its arguments and passes them on to the wrapped function, without any broadcasting. We would also include a `dim` argument that is automatically converted into the appropriate `axis` argument when calling the wrapped function.

Output dimensions would be determined from a simple rule of some sort:
- Default output dimensions would either be copied from the first argument, or would take on the ordered union on all input dimensions.
- Custom dimensions could either be set by adding a `drop_dims` argument (like `dask.array.map_blocks`), or require an explicit override `output_dims`.

jhamman:
Would something like the following use case be covered by `apply_raw`:

```Python

def f(a, b, c):
    '''
    inputs: 3 1D Numpy arrays of differing lengths
    outputs: 1 1D Numpy array like `a`
    '''
    new = a + b.sum() + c.min()  # just an example using all three
    return new
```

The current `apply_ufunc`, when applied on NDarrays (eg time,x,y) cannot handle this since it tries to make sure all the inputs are broadcastable with each other.

shoyer:
@jhamman you could do that core dimensions, e.g., `input_core_dims=[('time',), ('x',), ('y',)]`, `output_core_dims=[('time',)]`. But `apply_raw` would have the advantage that you don't need to think about core dimensions :).

shoyer:
More related discussion: https://github.com/pydata/xarray/issues/1074#issuecomment-258327585

PR #1619, opened 2017-10-10
Expose apply_ufunc as public API and add documentation

### LINKED ISSUE #770: Internal refactor: create a generic function for applying ufuncs-like functions to xarray objects

shoyer:
`apply_ufunc()` does not meet all our needs for wrapping unlabeled array routines with xarray (see #1618 for a proposal for `apply_raw()`), but it should be useful for many advanced users and isn't doing much in its current state as non-public API. So I'd like to make it public for the xarray 0.10 release.

 - [x] Closes #770
 - [x] Turn the example from #1517 into a separate docs page of some sort, and link to it from `computation.rst` and `dask.rst`.

@MaximilianR @rabernat @jhamman Review from any of you would be appreciated here!

### DIFF
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -346,12 +346,14 @@ Wrapping custom computation
 
 It doesn't always make sense to do computation directly with xarray objects:
 
-  - When working with small arrays (less than ~1e7 elements), applying an
-    operation with xarray can be significantly slower. Keeping track of labels
-    and ensuring their consistency adds overhead, and xarray's high level label-
-    based APIs remove low-level control over the implementation of operations.
-    Also, xarray's core itself is not especially fast, because it's written in
-    Python rather than a compiled language like C.
+  - In the inner loop of performance limited code, using xarray can add
+    considerable overhead compared to using NumPy or native Python types.
+    This is particularly true when working with scalars or small arrays (less
+    than ~1e7 elements). Keeping track of labels and ensuring their consistency
+    adds overhead, and array's core itself is not especially fast, because it's
+    written in Python rather than a compiled language like C. Also, xarray's
+    high level label-based APIs remove low-level control over how operations are
+    implemented.
   - Even if speed doesn't matter, it can be important to wrap existing code, or
     to support alternative interfaces that don't use xarray objects.
 
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -303,7 +303,7 @@ Datasets support most of the same methods found on data arrays:
     ds.mean(dim='x')
     abs(ds)
 
-Unfortunately, we current do not support NUmPy ufuncs for datasets [1]_.
+Unfortunately, we currently do not support NumPy ufuncs for datasets [1]_.
 :py:meth:`~xarray.Dataset.apply` works around this
 limitation, by applying the given function to each variable in the dataset:
 
@@ -339,7 +339,7 @@ matching data variables.
        we should be able to support this by leveraging ``__array_ufunc__``
        (:issue:`1617`).
 
-.. computation.wrapping-custom:
+.. _comput.wrapping-custom:
 
 Wrapping custom computation
 ===========================
@@ -376,7 +376,7 @@ any additional arguments:
 
     squared_error = lambda x, y: (x - y) ** 2
     arr1 = xr.DataArray([0, 1, 2, 3], dims='x')
-    xr.apply_func(squared_error, arr1, 1)
+    xr.apply_ufunc(squared_error, arr1, 1)
 
 For using more complex operations that consider some array values collectively,
 it's important to understand the idea of "core dimensions" from NumPy's
@@ -417,10 +417,8 @@ nicely with tools for building vectorized functions, like
 needs, consider using Numba's `vectorize and guvectorize <http://numba.pydata.org/numba-doc/dev/user/vectorize.html>`_.
 
 In addition to wrapping functions, ``apply_ufunc`` can automatically parallelize
-many functions when using dask by setting ``dask='parallelized'``. This is
-illustrated in a separate example.
-
-.. TODO: add link!
+many functions when using dask by setting ``dask='parallelized'``. See
+:ref:`dask.automatic-parallelization` for details.
 
 :py:func:`~xarray.apply_ufunc` also supports some advanced options for
 controlling alignment of variables and the form of the result. See the
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -349,7 +349,7 @@ It doesn't always make sense to do computation directly with xarray objects:
   - In the inner loop of performance limited code, using xarray can add
     considerable overhead compared to using NumPy or native Python types.
     This is particularly true when working with scalars or small arrays (less
-    than ~1e7 elements). Keeping track of labels and ensuring their consistency
+    than ~1e6 elements). Keeping track of labels and ensuring their consistency
     adds overhead, and array's core itself is not especially fast, because it's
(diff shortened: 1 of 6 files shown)


doc/whats-new.rst
```diff
@@ -75,6 +75,11 @@ Backward Incompatible Changes
 Enhancements
 ~~~~~~~~~~~~
 
+- New helper function :py:func:`~xarray.apply_ufunc` for wrapping functions
+  written to work on NumPy arrays to support labels on xarray objects.
+  ``apply_ufunc`` also support automatic parallelization for many functions
+  with dask. See :ref:`computation.wrapping-custom` for details.
+
```
jhamman: Don't forget: 
```
(:issue:`XXXX`). By `Stephan Hoyer <https://github.com/shoyer>`_. 
```

doc/computation.rst
```diff
@@ -329,5 +335,93 @@ Arithmetic between two datasets matches data variables of the same name:
...
+Because ``apply_ufunc`` follows a standard convention for ufuncs, it plays
+nicely with tools for building vectorized functions, like
+:func:`numpy.broadcast_arrays` and :func:`numpy.vectorize`. For high performance
+needs, consider using Numba's `vectorize and guvectorize <http://numba.pydata.org/numba-doc/dev/user/vectorize.html>`_.
+
+In addition to wrapping functions, ``apply_ufunc`` can automatically parallelize
+many functions when using dask by setting ``dask='parallelized'``. This is
+illustrated in a separate example.
+
+.. TODO: add link!
```
jhamman: I'm thinking the recipes would be a good place for this: http://xarray.pydata.org/en/stable/auto_gallery/index.html

shoyer: Yes, if I can come up with a figure!

doc/computation.rst
```diff
@@ -303,15 +303,21 @@ Datasets support most of the same methods found on data arrays:
     ds.mean(dim='x')
     abs(ds)
 
-Unfortunately, a limitation of the current version of numpy means that we
-cannot override ufuncs for datasets, because datasets cannot be written as
-a single array [1]_. :py:meth:`~xarray.Dataset.apply` works around this
+Unfortunately, we current do not support NUmPy ufuncs for datasets [1]_.
```
benbovy: currently

doc/computation.rst
```diff
@@ -329,5 +335,93 @@ Arithmetic between two datasets matches data variables of the same name:
...
+functions/methods are written using ``apply_ufunc``.
+
+Simple functions that act independently on each value should work without
+any additional arguments:
+
+.. ipython:: python
+
+    squared_error = lambda x, y: (x - y) ** 2
+    arr1 = xr.DataArray([0, 1, 2, 3], dims='x')
+    xr.apply_func(squared_error, arr1, 1)
```
benbovy: `xr.apply_ufunc`

shoyer:
@benbovy @jhamman thanks for giving this a read. The docs actually build now, and I added a section on "Automatic parallelization" to the dask docs.

doc/dask.rst
```diff
@@ -213,6 +213,93 @@ loaded into dask or not:
...
+-------------------------
+
+Almost all of xarray's built-in operations work on dask arrays. If you want to
+use a function that isn't wrapped by xarray, one option is to extract dask
+arrays from xarray objects (``.data``) and use dask directly.
+
+Another option is to use xarray's :py:func:`~xarray.apply_ufunc`, which can
+automatically parallelize functions written for processing NumPy arrays when
+applied to dask arrays. It works similarly to :py:func:`dask.array.map_blocks`
+and :py:func:`dask.array.atop`, but without requiring an immediate layer of
```
jhamman: did you mean intermediate?

shoyer: yes, indeed

doc/computation.rst
```diff
@@ -329,5 +335,91 @@ Arithmetic between two datasets matches data variables of the same name:
...
+       (:issue:`1617`).
+
+.. _comput.wrapping-custom:
+
+Wrapping custom computation
+===========================
+
+It doesn't always make sense to do computation directly with xarray objects:
+
+  - When working with small arrays (less than ~1e7 elements), applying an
```
MaximilianR: Is the point on speed is a distraction? If an array is that small, the _absolute_ difference in speed is still very small, so it really only makes a difference if you're doing those operations in a loop.

xarray/core/computation.py
```diff
@@ -724,7 +727,7 @@ def apply_ufunc(func, *args, **kwargs):
     ``apply_ufunc`` to write functions to (very nearly) replicate existing
     xarray functionality:
 
-    Calculate the vector magnitude of two arguments:
+    Calculate the vector magnitude of two arguments::
```
MaximilianR: ?

doc/computation.rst
shoyer: Agreed. I had "in the inner loop" in mind when I wrote this, but I see now that that never made it into the text. Let me know if this latest update seems more reasonable to you, or if I'm still pushing too hard on performance considerations.

shoyer:
Any additional comments? If not, I'll merge this in about 24 hours.

PR #1622, opened 2017-10-10
Improved netcdf4 error message when opening non existent file

jhamman:
- [x] Tests added / passed

This is particularly helpful when using `open_mfdataset` with a list of files. For example:

Current behavior (unclear which file is causing open_dataset to choke):
```Python-traceback
In [2]: xr.open_mfdataset(['foo.nc', 'bar.nc'], engine='netcdf4')
OSError                                   Traceback (most recent call last)
<ipython-input-2-12b24f38f025> in <module>()
----> 1 xr.open_mfdataset(['foo.nc', 'bar.nc'], engine='netcdf4')

...

/Users/jhamman/Dropbox/src/xarray/xarray/backends/netCDF4_.py in _open_netcdf4_group(filename, mode, group, **kwargs)
    183     import netCDF4 as nc4
    184
--> 185     ds = nc4.Dataset(filename, mode=mode, **kwargs)
    186
    187     with close_on_error(ds):

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Dataset.__init__()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4._ensure_nc_success()

OSError: No such file or directory
```

And the new behavior
```Python-traceback
In [2]: xr.open_mfdataset(['foo.nc', 'bar.nc'], engine='netcdf4')

...

/Users/jhamman/Dropbox/src/xarray/xarray/backends/netCDF4_.py in _open_netcdf4_group(filename, mode, group, **kwargs)
    186         ds = nc4.Dataset(filename, mode=mode, **kwargs)
    187     except OSError as e:
--> 188         raise OSError("Error opening %r" % filename, e)
    189
    190     with close_on_error(ds):

OSError: [Errno Error opening '/Users/jhamman/Dropbox/src/xarray/foo.nc'] No such file or directory

### DIFF
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -187,9 +187,10 @@ def _open_netcdf4_group(filename, mode, group=None, **kwargs):
         ds = nc4.Dataset(filename, mode=mode, **kwargs)
     except IOError as e:
         message = e.args[0]
+        err_num = e.errno
         if 'No such file or directory' in message:
-            raise FileNotFoundError(errno.ENOENT, message, filename)
-        raise IOError(e.errno, message, filename)
+            err_num = errno.ENOENT
+        raise IOError(err_num, message, filename)
 
     with close_on_error(ds):
         ds = _nc4_group(ds, group, mode)
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -3,6 +3,7 @@ from __future__ import division
 from __future__ import print_function
 import functools
 import operator
+import errno
 
 import numpy as np
 
@@ -184,8 +185,11 @@ def _open_netcdf4_group(filename, mode, group=None, **kwargs):
 
     try:
         ds = nc4.Dataset(filename, mode=mode, **kwargs)
-    except (OSError, IOError) as e:
-        raise OSError("Error opening %r" % filename, e)
+    except IOError as e:
+        message = e.args[0]
+        if 'No such file or directory' in message:
+            raise FileNotFoundError(errno.ENOENT, message, filename)
+        raise IOError(e.errno, message, filename)
 
     with close_on_error(ds):
         ds = _nc4_group(ds, group, mode)
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -182,7 +182,10 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
 def _open_netcdf4_group(filename, mode, group=None, **kwargs):
     import netCDF4 as nc4
 
-    ds = nc4.Dataset(filename, mode=mode, **kwargs)
+    try:
+        ds = nc4.Dataset(filename, mode=mode, **kwargs)
+    except OSError as e:
+        raise OSError("Error opening %r" % filename, e)
 
     with close_on_error(ds):
         ds = _nc4_group(ds, group, mode)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: xarray/backends/netCDF4_.py [hunk 1]
```diff
@@ -182,7 +182,10 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
 def _open_netcdf4_group(filename, mode, group=None, **kwargs):
     import netCDF4 as nc4
 
-    ds = nc4.Dataset(filename, mode=mode, **kwargs)
+    try:
+        ds = nc4.Dataset(filename, mode=mode, **kwargs)
+    except OSError as e:
+        raise OSError("Error opening %r" % filename, e)
```
shoyer: I don't think this is using the `OSError` constructor properly -- the first argument is supposed to be an error number:
https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

jhamman: You're right. I wasn't aware of this. I'm wondering if we should just raise an IOError here?

shoyer: That does feel like a better error type for most cases, but I'm not sure about every case. Presumably there's some reason why netCDF4 raises OSError...

jhamman: Thanks, I'll play around with this a bit more.

jhamman:
@shoyer - I think this is about as good as we can do:

Here's what's happening:
- netcdf4 is raising an `IOError` when the file is not found: https://github.com/Unidata/netcdf4-python/blob/3da921e93f158f6e443bbb9c8bfdfe681fac4351/netCDF4/_netCDF4.pyx#L2011
- but the error number is not being passed correctly to `IOError` so depending if we're on python 2 or 3, it seems this gets cast to either a `IOError` (py2) or a `OSError` (py3): 
https://github.com/Unidata/netcdf4-python/blob/3da921e93f158f6e443bbb9c8bfdfe681fac4351/netCDF4/_netCDF4.pyx#L1631-L1634

In my last commit, I think we can just catch the `IOError` and reraise a `IOError` (looks like `FileNotFoundError` on python3.

This got more complicated than it need to but I'm pretty confident this works.

xarray/backends/netCDF4_.py
```diff
@@ -182,7 +183,14 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
 def _open_netcdf4_group(filename, mode, group=None, **kwargs):
     import netCDF4 as nc4
 
-    ds = nc4.Dataset(filename, mode=mode, **kwargs)
+    try:
+        ds = nc4.Dataset(filename, mode=mode, **kwargs)
+    except IOError as e:
+        message = e.args[0]
```
shoyer: I'm a little nervious about this fix -- it seems like it's correctness relies on netCDF4 raising an error with only this single argument.

Would it make sense to fix this upstream in netCDF4 instead? This looks like where the error is raised:
https://github.com/Unidata/netcdf4-python/blob/master/netCDF4/_netCDF4.pyx#L2011

jhamman: I think this is a fair point. Let's leave this open until we see if I can get this to work in netCDF4-python.

jhamman:
see https://github.com/Unidata/netcdf4-python/pull/727

jhamman:
closed via Unidata/netcdf4-python#727

Issue #884, opened 2016-06-15
Q: Should Dataset.keys() return only variable keys?


MaximilianR:
This has been a small-but-persistent issue for me for a while. I suspect that my perspective might be dependent on my current outlook, but socializing it here to test if it's secular...

Currently `Dataset.keys()` returns both variables and coordinates (but not its `attrs` keys):
```python
In [5]: ds=xr.Dataset({'a': (('x', 'y'), np.random.rand(10,2))})
In [12]: list(ds.keys())
Out[12]: ['a', 'x', 'y']
```

Is this conceptually correct? I would posit that a Dataset is a mapping of *keys to variables*, and the coordinates contain values that _label_ that data. 

**So should `Dataset.keys()` instead return just the keys of the Variables?**

We're often passing around a dataset as a `Mapping` of keys to values - but then when we run a function across each of the keys, we get something run on both the Variables' keys, _and_ the Coordinate / label's keys.

In Pandas, `DataFrame.keys()` returns just the columns, so that conforms to what we need. While I think the xarray design is in general much better in these areas, this is one area that pandas seems to get correct - and because of the inconsistency between pandas & xarray, we're having to coerce our objects to pandas `DataFrame`s before passing them off to functions that pull out their keys (this is also why we can't just look at `ds.data_vars.keys()` - because it breaks that duck-typing).

Does that make sense?

shoyer:
An early version of xarray actually worked exactly like this, but we switched it to the current functionality. Let me see if I can dig up the relevant issues....

shoyer:
Here is one place where I discussed this with myself!
https://github.com/pydata/xarray/issues/211

Then I merged the functionality you are describing, but at some point switched it back....

fmaussion:
What would happen to the actual variables? Would `ds['longitude']` also return a `KeyError`? This would be very far from the NetCDF model and would brake many many things....

MaximilianR:
> Then I merged the functionality you are describing, but at some point switched it back....

Ha!

What are your thoughts now?

shoyer:
> What would happen to the actual variables? Would ds['longitude'] also return a KeyError? This would be very far from the NetCDF model and would brake many many things....

There is no way we would change this -- `ds['longitude']` will always return the longitude array.

> What are your thoughts now?

[...]

PR #1632, opened 2017-10-13
Support autocompletion dictionary access in ipython.

### LINKED ISSUE #1628: Add _ipython_key_completions_ to xarray.Dataset
We could use to enable autocompletion for dictionary access like `ds['temp` -> `ds['temperature']`
http://ipython.readthedocs.io/en/stable/config/integrating.html#tab-completion

fujiisoup:
- [x] Closes #1628
 - [x] Tests added / passed

Support #1628.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -74,6 +74,11 @@ Backward Incompatible Changes
 
 Enhancements
 ~~~~~~~~~~~~
+- Support `_ipython_key_completions_` to enable autocompletion for
+  dictionary access.
+  (:issue:`1628`).
+  By `Keisuke Fujii <https://github.com/fujiisoup>`_.
+
 - Support for ``data_vars`` and ``coords`` keywords added to
   :py:func:`~xarray.open_mfdataset`
   (:issue:`438`):
@@ -260,7 +265,7 @@ Bug fixes
   The previous behavior unintentionally causing additional tests to be skipped
   (:issue:`1531`). By `Joe Hamman <https://github.com/jhamman>`_.
 
-- Fix pynio backend for upcoming release of pynio with python3 support 
+- Fix pynio backend for upcoming release of pynio with python3 support
   (:issue:`1611`). By `Ben Hillman <https://github/brhillman>`_.
 
 .. _whats-new.0.9.6:
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -75,7 +75,8 @@ Backward Incompatible Changes
 Enhancements
 ~~~~~~~~~~~~
 - Support `_ipython_key_completions_` to enable autocompletion for
-  dictionary access.
+  dictionary access with IPython.
+  e.g.  ``ds['tem`` + tab -> ``ds['temperature']``
   (:issue:`1628`).
   By `Keisuke Fujii <https://github.com/fujiisoup>`_.
 
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -200,6 +200,10 @@ class AttrAccessMixin(object):
         return sorted(set(dir(type(self)) + extra_attrs))
 
     def _ipython_key_completions_(self):
+        """Provide method for the key-autocompletions in IPython.
+        See http://ipython.readthedocs.io/en/stable/config/integrating.html#tab-completion
+        For the details.
+        """
         item_lists = [item
                       for sublist in self._item_sources
                       for item in sublist
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -193,6 +193,11 @@ class AttrAccessMixin(object):
             if isinstance(item, basestring)]
         return sorted(set(dir(type(self)) + extra_attrs))
 
+    def _ipython_key_completions_(self):
+        item_lists = [item for sublist in self._item_sources for item
+                      in sublist if isinstance(item, basestring)]
+        return list(set(item_lists))
+
 
 def get_squeeze_dims(xarray_obj, dim):
     """Get a list of dimensions to squeeze out.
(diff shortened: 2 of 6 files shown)


fujiisoup:
It would be really good if we could also have an autocompletion for the keyword argument, such as 
`sel(tem` ->  `sel(temperature=`, but I have no idea though...

fujiisoup:
I think we can also have 
`ds.coords['tem` -> `ds.coords['temperature']` and `ds.data_vars['tem` -> `ds.data_vars['temperature']`.
I will update thsi PR later (maybe few days later).

xarray/core/common.py
```diff
@@ -193,6 +193,11 @@ def __dir__(self):
             if isinstance(item, basestring)]
         return sorted(set(dir(type(self)) + extra_attrs))
 
+    def _ipython_key_completions_(self):
+        item_lists = [item for sublist in self._item_sources for item
+                      in sublist if isinstance(item, basestring)]
```
shoyer: nit: can you reformat this to not line break between "for" and "in", e.g.,
```
        item_lists = [item for sublist in self._item_sources
                      for item in sublist if isinstance(item, basestring)]
```

Actually, additional line breaks might be even better:
```
        item_lists = [item
                      for sublist in self._item_sources
                      for item in sublist
                      if isinstance(item, basestring)]
```

xarray/core/dataarray.py
```diff
@@ -486,6 +486,11 @@ def _attr_sources(self):
         """List of places to look-up items for attribute-style access"""
         return [self.coords, LevelCoordinatesSource(self), self.attrs]
```
shoyer: Should `self.dims` also be in this list? I think so?

xarray/tests/test_dataset.py
```diff
@@ -3505,6 +3505,44 @@ def test_sortby(self):
         actual = ds.sortby(['x', 'y'], ascending=False)
         self.assertDatasetEqual(actual, ds)
 
+    def test_ipython_key_completion(self):
+        ds = create_test_data(seed=1)
+        actual = ds._ipython_key_completions_()
+        expected = ['var1', 'var2', 'var3', 'time', 'dim1',
+                    'dim2', 'dim3', 'numbers']
+        for item in expected:
+            ds[item]
```
shoyer: add a comment, e.g.,  `# should not raise`

xarray/tests/test_dataset.py
```diff
@@ -3505,6 +3505,44 @@ def test_sortby(self):
...
+        self.assertTrue(len(actual) == len(expected))
+        self.assertTrue(set(actual) == set(expected))
```
shoyer: `assert sorted(actual) == sorted(expected)` would be slightly cleaner than these tests

xarray/tests/test_dataset.py
```diff
@@ -3505,6 +3505,44 @@ def test_sortby(self):
...
+            ds['var1'][item]
+        self.assertTrue(len(actual) == len(expected))
+        self.assertTrue(set(actual) == set(expected))
+
+        actual = ds['var3']._ipython_key_completions_()
+        expected = ['dim3', 'dim1', 'numbers']
+        self.assertTrue(len(actual) == len(expected))
+        self.assertTrue(set(actual) == set(expected))
+
+        # drop variables
```
shoyer: I'm not sure this makes sense to test separately.

doc/whats-new.rst
```diff
@@ -74,6 +74,11 @@ Backward Incompatible Changes
 
 Enhancements
 ~~~~~~~~~~~~
+- Support `_ipython_key_completions_` to enable autocompletion for
+  dictionary access.
```
shoyer: Add an example here?

e.g., `ds['temp` -> `ds['temperature']`

xarray/core/dataarray.py
fujiisoup: You are right.
I fixed this and added tests for this.

fujiisoup:
I added a tab-completion support also for `ds.coords` and `ds.data_vars`.

I am getting to feel slight unnaturalness against the following behavior.

```python
In [1]: import xarray as xr
   ...: da = xr.Dataset({'var1': ('x', [0, 1, 2])})

In [2]: 'x' in da
Out[2]: False

In [3]: 'x' in da.keys()
Out[3]: False

In [4]: da['x']
Out[4]: 
<xarray.DataArray 'x' (x: 3)>
array([0, 1, 2])
```

I understand that it is a designed behavior, 
but it looks a little bit unintuitive.

Maybe `'x' in da` should be `True`?

xarray/tests/test_dataset.py
```diff
@@ -3505,6 +3505,64 @@ def test_sortby(self):
         actual = ds.sortby(['x', 'y'], ascending=False)
         self.assertDatasetEqual(actual, ds)
 
+    def test_attribute_access(self):
+        ds = create_test_data(seed=1)
+        for key in ['var1', 'var2', 'var3', 'time', 'dim1',
+                    'dim2', 'dim3', 'numbers']:
+            self.assertDataArrayEqual(ds[key], getattr(ds, key))
+            assert key in ds.__dir__()
```
shoyer: Nit: use `dir(ds)` in preference to `ds.__dir__()`

xarray/tests/test_dataset.py
```diff
@@ -3505,6 +3505,64 @@ def test_sortby(self):
...
+        assert 'foo' in ds['var3'].__dir__()
+
+    def test_ipython_key_completion(self):
+        ds = create_test_data(seed=1)
+        actual = ds._ipython_key_completions_()
+        expected = ['var1', 'var2', 'var3', 'time', 'dim1',
+                    'dim2', 'dim3', 'numbers']
+        for item in actual:
+            ds[item]  # should not raise
+        self.assertTrue(sorted(actual) == sorted(expected))
```
shoyer: Just use `assert sorted(actual) == sorted(expected)` (idiomatic when using pytest) or `self.assertEqual(sorted(actual), sorted(expected))`.

shoyer:
@fujiisoup I responded to your remarks on `in` over in https://github.com/pydata/xarray/issues/884#issuecomment-338446509. But yes, I agree.

fujiisoup:
@shoyer 
Thanks. Updated.

shoyer:
Thanks!

PR #1634, opened 2017-10-17
catch/supress seaborn import warning

### LINKED ISSUE #1633: seaborn.apionly module is deprecated
Xarray is using the apionly module from seaborn which is now raising this warning:

```Python
...python3.6/site-packages/seaborn/apionly.py:6: UserWarning: As seaborn no longer sets a default style on import, the seaborn.apionly module is deprecated. It will be removed in a future version.
  warnings.warn(msg, UserWarning)
```

I think the only places we use seaborn are here:

https://github.com/pydata/xarray/blob/2949558b75a65404a500a237ec54834fd6946d07/xarray/plot/utils.py#L76-L87

This shouldn't a difficult fix if/when we decide to change it.

xref: https://github.com/mwaskom/seaborn/pull/1216

jhamman:
- [x] Closes #1633
 - [x] Tests added / passed

### DIFF
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import pkg_resources
+import warnings
 
 import numpy as np
 import pandas as pd
@@ -73,13 +74,19 @@ def _color_palette(cmap, n_colors):
         # we have a list of colors
         try:
             # first try to turn it into a palette with seaborn
-            from seaborn.apionly import color_palette
-            pal = color_palette(cmap, n_colors=n_colors)
+            with warnings.catch_warnings():
+                warnings.filterwarnings('error')
+                try:
+                    from seaborn.apionly import color_palette
+                except UserWarning:
+                    from seaborn import color_palette
         except ImportError:
             # if that fails, use matplotlib
             # in this case, is there any difference between mpl and seaborn?
             cmap = ListedColormap(cmap, N=n_colors)
             pal = cmap(colors_i)
+        else:
+            pal = color_palette(cmap, n_colors=n_colors)
     elif isinstance(cmap, basestring):
         # we have some sort of named palette
         try:
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -78,7 +78,7 @@ def _color_palette(cmap, n_colors):
                 warnings.filterwarnings('error')
                 try:
                     from seaborn.apionly import color_palette
-                except UserWarning:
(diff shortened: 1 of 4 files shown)


xarray/plot/utils.py
```diff
@@ -73,13 +74,19 @@ def _color_palette(cmap, n_colors):
         # we have a list of colors
         try:
             # first try to turn it into a palette with seaborn
-            from seaborn.apionly import color_palette
-            pal = color_palette(cmap, n_colors=n_colors)
+            with warnings.catch_warnings():
+                warnings.filterwarnings('error')
+                try:
+                    from seaborn.apionly import color_palette
+                except UserWarning:
```
shoyer: Let's make this `except (UserWarning, ImportError)`, to guard against when `seaborn.apionly` is removed in some future version.

xarray/tests/__init__.py
```diff
@@ -54,6 +54,7 @@ def _importorskip(modname, minversion=None):
 has_bottleneck, requires_bottleneck = _importorskip('bottleneck')
 has_rasterio, requires_rasterio = _importorskip('rasterio')
 has_pathlib, requires_pathlib = _importorskip('pathlib')
+has_seaborn, requires_seaborn = _importorskip('matplotlib')
```
MaximilianR: Scanning this quickly so forgive me if this isn't right - but should that be `_importorskip('seaborn')`?

jhamman: you are correct. (and my tests failed so I guess I'll revisit tomorrow)

jhamman:
Looks like we've bumped into a travis build error: 

```
E   ImportError: cannot import name 'ellipkm1'
```

This seems like a scipy thing but I can't say I understand what is causing it.

shoyer:
It definitely looks like a botched scipy install on Travis but I can't see where it's coming from.

xarray/plot/utils.py
```diff
@@ -23,6 +24,21 @@ def _load_default_cmap(fname='default_colormap.csv'):
...
+            import seaborn.apionly as sns
+            if w and issubclass(w[-1].category, UserWarning):
+                raise ImportError
+        except ImportError:
+            import seaborn as sns
+        finally:
+            warnings.resetwarnings()
+    return sns
+
+
```
jhamman: @shoyer - it turns out the build failures we were seeing were because we were catching all the warnings/errors that scipy suppresses during its import. This solution seems to work but also relies on last (of many) warnings being the one from seaborn.

shoyer: You would think warnings would follow normal scoping rules like exceptions, but apparently not.

Issue #1436, opened 2017-05-31
Wrong dimension referenced in boolean indexing with .loc


d-chambers:
I am using Ubuntu 16, python 3.6, and xarary 0.9.5.

```python
import numpy as np
import xarray as xr

# setup for a simple grid
DX = 50
X = np.arange(0, 2010, DX)
Y = np.arange(0, 2010, DX)
Z = np.arange(0, 2010, DX)
grid_shape = (len(X), len(Y), len(Z))

# Create data array
dims = 'X Y Z'.split()
coords = {'X': X, 'Y': Y, 'Z': Z}
dar = xr.DataArray(np.ones(grid_shape), dims=dims, coords=coords)

# slice the data array so that all Z values are greater than 1000
dar2 = dar.loc[dar.Z > 1000]
assert np.all(dar2.Z > 1000)  # fails becase dar is sliced along X, not Z
```
Since the object returned from dar.Z > 1000 is a data array with "Z" as the only dim I would expect this to slice the "Z" dim rather than X.

d-chambers:
I should point out It does work when I specify the dimension with a dict:
```python
dar2 = dar.loc[{'Z': dar.Z > 1000}]
assert np.all(dar2.Z > 1000)  # doesn't raise
```

shoyer:
Xarray's indexing currently ignores dimension/index labels, which I agree is not ideal (and can lead to surprising behavior like this): https://github.com/pydata/xarray/issues/974

Another work-around is to use `where`, e.g., `dar.where(dar.Z > 1000)`.

shoyer:
Fixed by #1639

PR #1639, opened 2017-10-19
indexing with broadcasting

### LINKED ISSUE #1444: Best practice when the _Unsigned attribute is present in NetCDF files

fujiisoup:
- [x] Closes #1444, #1436
 - [x] Tests added / passed

This is a duplicate of #1473 originally opened by @shoyer 

Thanks, @shoyer, for giving me github's credit.

I enjoyed this PR.
I really appreciate your help to finish up this PR.


jhamman:
Nice work. @shoyer - I'll let you do the merge.

shoyer:
In it goes!

PR #1640, opened 2017-10-20
WIP: Feature/interpolate

### LINKED ISSUE #1631: Resample / upsample behavior diverges from pandas
I've found a few issues where xarray's new resample / upsample functionality is diverging from Pandas. I think they are mostly surrounding how NaNs are treated. Thoughts from @shoyer, @darothen and others.

Gist with all the juicy details: https://gist.github.com/jhamman/354f0e5ff32a39550ffd25800e7214fc#file-xarray_resample-ipynb

xref: #1608, #1272

jhamman:
- [x] Closes #1631
 - [x] Tests added / passed

Rough draft of interpolate method for filling of arbitrary nans. 

cc @darothen


xarray/core/interpolate.py
```diff
@@ -0,0 +1,95 @@
...
+        assert np.issubdtype(index.dtype, np.datetime64)
+    else:
+        index = da.get_index(dim)
+
+    # check index sorting now so we can skip it later
+    if not (np.diff(index) > 0).all():
+        raise ValueError("Index must be monotonicly increasing")
+
+    # TODO: consider using the np_interp_wrapper for linear method
+    out = np.apply_along_axis(scipy_interp_wrapper, axis, arr, index,
```
fujiisoup: I think `scipy.interpolate.interp1d` can interpolate a multiple dimensional array along one axis, though scipy is an optional requirement.
(But I did not check it is certainly faster.)

fujiisoup: ~I think `scipy.interpolate.interp1d` can interpolate a multiple dimensional array along one axis, though scipy is an optional requirement.
(But I did not check it is certainly faster.)~

EDIT:
I noticed that I did not understand correctly.
The multi-dimensional interpolation does not work for filling the missing values which are distributed randomly.

So ignore my comment.

jhamman: yes, exactly. For 1 and 2d arrays, this should behave identically to the pandas interpolate method.

shoyer: Maybe apply_ufunc with vectorize=True would work here? I guess it would reorder dimensions in a non-desired way.

xarray/core/interpolate.py
```diff
@@ -0,0 +1,95 @@
+"""Define core operations for xarray objects.
+
+TODO(shoyer): rewrite this module, making use of xarray.core.computation,
```
shoyer: This TODO note doesn't belong here :)

xarray/core/dataarray.py
```diff
@@ -1175,6 +1175,35 @@ def fillna(self, value):
         out = ops.fillna(self, value)
         return out
 
+    def interpolate(self, dim=None, method='linear', inplace=False):
```
shoyer: I like calling the `interpolate_na` rather than `interpolate` to make it clear what this method does. See here for the related pandas discussion: https://github.com/pandas-dev/pandas/issues/9340

xarray/core/interpolate.py
jhamman: That is on my list of things to try today. Getting things to behave like pandas required a bit of dancing and I've saved the obvious optimizations for later. Specifically, I want to get some more tests written first.

xarray/core/missing.py
```diff
@@ -0,0 +1,213 @@
...
+    except:
+        raise ValueError('%r is not a valid interpolator' % method)
+
+    return apply_ufunc(wrap_interpolator,
+                       interpolator, index, arr, kwargs=kwargs,
+                       input_core_dims=[[], [dim], [dim]],
+                       output_core_dims=[[dim]],
+                       dask='forbidden',
+                       vectorize=True,
+                       keep_attrs=True).transpose(*arr.dims)
```
jhamman: @shoyer - can you take a look at this? About half of my tests are passing but I suspect I'm still missing something in the `apply_ufunc` call. 

I'm getting a few errors that look like this: 

```
ValueError: inconsistent size for core dimension 'dim0': 1 vs 6
```

dcherian:
@jhamman Would it be possible to have a `limit` option like pandas?

Quoting http://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.interpolate.html
```
limit : int, default None.

    Maximum number of consecutive NaNs to fill. Must be greater than 0.
```

I suspect that a lot of the time, users want to fill in "small" gaps but not "large" ones. It is what I usually do.

jhamman:
@dcherian - Yes, we can do that. Just haven't gotten there yet.

xarray/core/missing.py
```diff
@@ -0,0 +1,258 @@
...
+        if self.cons_kwargs:
+            raise ValueError(
+                'recieved invalid kwargs: %r' % self.cons_kwargs.keys())
+
+        if fill_value is None:
+            self._left = np.nan
+            self._right = yi[-1]
+        elif isinstance(fill_value, Iterable):
+            self._left = fill_value[0]
+            self._right = fill_value[1]
```
shoyer: If this only works for a pair of values, let's restrict this to `isintance(fill_value, tuple)` and explicitly check that it has length 2.

xarray/core/missing.py
```diff
@@ -0,0 +1,258 @@
...
+        self._xi = xi
+        self._yi = yi
+
+        if self.cons_kwargs:
+            raise ValueError(
+                'recieved invalid kwargs: %r' % self.cons_kwargs.keys())
+
+        if fill_value is None:
+            self._left = np.nan
+            self._right = yi[-1]
```
shoyer: This is a little surprising to me -- I would expect both to default to NaN.

xarray/core/missing.py
```diff
@@ -0,0 +1,258 @@
...
+        if fill_value is not None:
+            raise ValueError('SplineInterpolator does not support fill_value')
+
+        self.f = UnivariateSpline(xi, yi, k=order, **self.cons_kwargs)
+
+
+def get_clean_interp_args(arr, method, dim, **kwargs):
+    '''normalize interolation arguments and choose which index to use
+
+    provides compatability with pandas interolation methods
```
shoyer: interolation -> interpolation

xarray/core/missing.py
```diff
@@ -0,0 +1,258 @@
...
+    axis = arr.get_axis_num(dim)
+    if method == 'linear':
+        index = np.arange(arr.shape[axis])
+    elif method in ['values', 'index']:
+        index = arr.get_index(dim)
+        method = 'linear'
+    elif method == 'time':
+        index = arr.get_index(dim)
+        method = 'linear'
+        assert isinstance(index, pd.DatetimeIndex)
```
shoyer: Should probably raise ValueError instead

xarray/core/dataarray.py
```diff
@@ -1185,6 +1185,37 @@ def fillna(self, value):
...
+            'time': interpolation works on daily and higher resolution data to
+                    interpolate given length of interval
+            'index', 'values': use the actual numerical values of the index
+            'nearest', 'zero', 'slinear', 'quadratic', 'cubic', 'barycentric',
+            'polynomial' is passed to scipy.interpolate.interp1d with the
+            order given both 'polynomial' and 'spline' require that you also
+            specify and order (int) e.g. da.interpolate_na(method='polynomial', order=4)
+
+        # TODO: Pandas also support: scipy wrappers of 'krogh',
+        #       'piecewise_polynomial', 'spline', and 'pchip'
```
shoyer: Personally, I find this interface from pandas a little confusing. It combines two separate choices in one argument:
- Choosing the points at which interpolation is done.
- How to interpolate between these points.

I would slightly rather that we had two separate arguments. Maybe:
- `points='coordinate'` vs `points='equal_spacing'` or a boolean `use_coordinate=True`/`use_coordinate=False`
- `method='nearest'`, `method='linear'`, `method='quadratic'`, etc, for indicating how to interpolate.

The downside is that this would be inconsistent with pandas. But in the existing interface, it is not at all obvious to me whether a method like `'polynomial'` uses coordinate points or equally spaced points.

xarray/core/missing.py
```diff
@@ -0,0 +1,258 @@
...
+            elif method == 'barycentric':
+                return interpolate.BarycentricInterpolator
+            elif method == 'krog':
+                return interpolate.KroghInterpolator
+            elif method == 'pchip':
+                return interpolate.PchipInterpolator
+            elif method == 'spline':
+                return SplineInterpolator
+        except ImportError:
+            pass
```
shoyer: It would probably be more useful to raise an error about how scipy is needed for the given interpolator rather than a generic error message.

xarray/core/missing.py
shoyer: I think the problem here is that you are treating your `interpolator` as an argument to vectorize over. This might technically work (by interpreting `interpolator` as a scalar), but was certainly was not intended.

Possibly we should have a more informative error message here.

There are various ways you can fix this (using either a class or function returned by a function), but either way probably want the way you use `apply_ufunc` to be something more like `apply_ufunc(wrap_interpolator(interpolator), index, arr, ...)`.

xarray/core/missing.py
```diff
@@ -0,0 +1,258 @@
...
+    kwargs.update(kind=method)
+
+    interpolator = _get_interpolator(method, **kwargs)
+
+    return apply_ufunc(wrap_interpolator,
+                       interpolator, index, arr, kwargs=kwargs,
+                       input_core_dims=[[], [dim], [dim]],
+                       output_core_dims=[[dim]],
+                       dask='forbidden',
+                       vectorize=True,
```
shoyer: Note that the `scipy.interpolate.interp1d` is already vectorized in the right way:
https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.interpolate.interp1d.html

xarray/core/dataarray.py
jhamman: I totally agree. As I've been working on this I've been thinking we can improve on the Pandas interface and this would be the obvious way to do that.

xarray/core/missing.py
jhamman: Agreed. I'm not sure if this is a Pandas bug or an intention feature.

xarray/core/missing.py
jhamman: I think we can only use interp1d on 1D slices. The key point here is that we should be able to interpolate at different index positions for different slices. I don't think `interp1d` will let us do that.

shoyer: Ah, good point. That would be useful for the `interpolate_at` variant.

xarray/core/missing.py
```diff
@@ -0,0 +1,261 @@
...
+    method, index = get_clean_interp_args(arr, method, dim, **kwargs)
+    kwargs.update(kind=method)
+
+    interpolator = _get_interpolator(method, **kwargs)
+
+    return apply_ufunc(wrap_interpolator,
+                       interpolator, index, arr, kwargs=kwargs,
+                       input_core_dims=[[], [dim], [dim]],
+                       output_core_dims=[[dim]],
+                       dask='forbidden',
```
shoyer: I think `dask='parallelized'` would work here -- I tested it with `vectorize=True`.

xarray/core/missing.py
```diff
@@ -0,0 +1,338 @@
...
+    interpolator = _get_interpolator(method, **kwargs)
+
+    arr = apply_ufunc(interpolator, index, arr,
+                      input_core_dims=[[dim], [dim]],
+                      output_core_dims=[[dim]],
+                      output_dtypes=[arr.dtype],
+                      output_sizes=dict(zip(arr.dims, arr.shape)),
+                      dask='parallelized',
+                      vectorize=True,
+                      keep_attrs=True).transpose(*arr.dims)
```
jhamman: @shoyer - I finally got the `apply_ufunc` to work. Turns out I just had a bug in `wrap_interpolator` Nonetheless, I am now using a partial function to simplify this call.

xarray/core/missing.py
```diff
@@ -0,0 +1,338 @@
...
+    '''Interpolate values according to different methods.'''
+
+    # this may not be possible with apply_ufunc
+    arr = self if inplace else self.copy()
+
+    if dim is None:
+        raise NotImplementedError('dim is a required argument')
+
+    if limit is not None:
+        valids = _get_valid_fill_mask(arr, dim, limit)
```
jhamman: @dcherian - the limit option is now supported.

dcherian: Awesome. Thanks!

xarray/core/dataarray.py
```diff
@@ -1214,6 +1214,56 @@ def fillna(self, value):
...
+                      spaced. default
+            'time': interpolation works on daily and higher resolution data to
+                    interpolate given length of interval
+            'index', 'values': use the actual numerical values of the index
+            'nearest', 'zero', 'slinear', 'quadratic', 'cubic', 'barycentric',
+            'polynomial' is passed to scipy.interpolate.interp1d with the
+            order given both 'polynomial' and 'spline' require that you also
+            specify and order (int) e.g. da.interpolate_na(method='polynomial',
+            order=4)
+        use_coordinate : boolean, default True
```
shoyer: needs a description.

xarray/core/dataarray.py
```diff
@@ -1214,6 +1214,56 @@ def fillna(self, value):
...
+                       **kwargs):
+        """Interpolate values according to different methods.
+
+        Parameters
+        ----------
+        dim : str
+            Specifies the dimension along which to interpolate.
+        method : {'linear', 'time', 'index', 'values', 'nearest'}
+            'linear': ignore the index and treat the values as equally
+                      spaced. default
```
shoyer: This is a minor point, but can we format this a little more consistently?

The `compat` arg on the Dataset constructor gets formatted nicely into a list by Sphinx and is still pretty readable as plain text:
https://github.com/pydata/xarray/blob/0818a3670521e7dc06b75b0a5241d2cf3f522cb2/xarray/core/dataset.py#L342-L350

xarray/core/dataarray.py
```diff
@@ -1214,6 +1214,56 @@ def fillna(self, value):
...
+            'time': interpolation works on daily and higher resolution data to
+                    interpolate given length of interval
+            'index', 'values': use the actual numerical values of the index
+            'nearest', 'zero', 'slinear', 'quadratic', 'cubic', 'barycentric',
+            'polynomial' is passed to scipy.interpolate.interp1d with the
+            order given both 'polynomial' and 'spline' require that you also
+            specify and order (int) e.g. da.interpolate_na(method='polynomial',
+            order=4)
+        use_coordinate : boolean, default True
+        limit : limit : int, default None
```
shoyer: `limit :` is repeated twice

xarray/core/dataarray.py
```diff
@@ -1214,6 +1214,56 @@ def fillna(self, value):
...
+            Maximum number of consecutive NaNs to fill. Must be greater than 0.
```
shoyer: add "or None for no limit"

xarray/core/dataarray.py
```diff
@@ -1214,6 +1214,56 @@ def fillna(self, value):
...
+        """
+        from .missing import interp_na
+        return interp_na(self, dim=dim, method=method, limit=limit,
+                         use_coordinate=use_coordinate, **kwargs)
+
+    def interpolate_at(self, dim, locs, method='linear', inplace=False,
+                       limit=None, **kwargs):
+        # this is just here so I remember the signature we discussed
+        # dim: the dimension along which to interpolate
+        # locs: a broadcastable boolean mask describing where interpolation
```
shoyer: boolean mask? I'm not quite sure how that would make sense. My immediate thought would be that `locs` should be a 1d array of coordinate values.

xarray/core/missing.py
```diff
@@ -0,0 +1,323 @@
...
+                      input_core_dims=[[dim], [dim]],
+                      output_core_dims=[[dim]],
+                      output_dtypes=[self.dtype],
+                      output_sizes=dict(zip(self.dims, self.shape)),
+                      dask='parallelized',
+                      vectorize=True,
+                      keep_attrs=True).transpose(*self.dims)
+
+    if limit is not None:
+        print(valids)
```
shoyer: remove :)

xarray/tests/test_missing.py
```diff
@@ -0,0 +1,261 @@
...
+    da = DataArray(
+        [0, np.nan, np.nan, np.nan, np.nan, 3, 4, 5, np.nan, 6, 7],
+        dims='time')
+    result = da.ffill('time')
+    expected = DataArray([0, 0, 0, 0, 0, 3, 4, 5, 5, 6, 7], dims='time')
+    assert_array_equal(result, expected)
+
+    result = da.ffill('time', limit=1)
+    expected = DataArray(
+        [0, 0, np.nan, np.nan, np.nan, 3, 4, 5, 5, 6, 7], dims='time')
```
shoyer: It would be nice to add a couple of tests using dask.

xarray/core/missing.py
```diff
@@ -0,0 +1,293 @@
...
+    f = interpolator(x[nonans], y[nonans], **kwargs)
+    out[nans] = f(x[nans])
+    return out
+
+
+def _bfill(arr, n=None, axis=-1):
+    '''inverse of ffill'''
+    import bottleneck as bn
+
+    arr = np.flip(arr, axis=axis)
```
jhamman: this requires numpy 0.12 or later so I'll need to add a compat function.

jhamman:
This is ready for some more reviews. I'm seeing some failing tests on the py27-min build that I'm struggling to detangle at the moment but I'm pretty sure they're just pytest related.

cc @darothen - I'd particularly appreciate your thoughts on the interpolators.

xarray/core/missing.py
jhamman: this was done in 19d21b8

xarray/core/dataarray.py
```diff
@@ -1220,6 +1220,91 @@ def fillna(self, value):
...
+    def interpolate_na(self, dim=None, method='linear', limit=None,
+                       use_coordinate=True,
+                       **kwargs):
+        """Interpolate values according to different methods.
+
+        Parameters
+        ----------
+        dim : str
+            Specifies the dimension along which to interpolate.
+        method : {'linear', 'time', 'index', 'values', 'nearest'}, optional
```
fujiisoup: Should probably some explanation about `method='time'` option is necessary here.

xarray/core/missing.py
```diff
@@ -0,0 +1,293 @@
...
+
+    See Also
+    --------
+    scipy.interpolate.UnivariateSpline
+    '''
+    def __init__(self, xi, yi, kind=None, fill_value=None, order=3, **kwargs):
+        from scipy.interpolate import UnivariateSpline
+
+        if kind is None:
+            raise ValueError('kind is a required argument')
```
fujiisoup: what should be passed as `kind`?

xarray/core/missing.py
```diff
@@ -0,0 +1,293 @@
...
+
+    if use_coordinate:
+        if use_coordinate is True:
+            index = arr.get_index(dim)
+        else:
+            index = arr.coords[use_coordinate]
+        if isinstance(index, pd.DatetimeIndex):
+            index = index.values.astype(np.float64)
+        # check index sorting now so we can skip it later
+        if not (np.diff(index) > 0).all():
```
fujiisoup: If we require a monotonically-increasing index, can we pass `assume_sorted=True` to `ScipyInterpolator`?

xarray/core/missing.py
```diff
@@ -0,0 +1,293 @@
...
+
+    If use_coordinate is False, the x values are set as an equally spaced
+    sequence.
+    '''
+
+    if use_coordinate:
+        if use_coordinate is True:
+            index = arr.get_index(dim)
+        else:
+            index = arr.coords[use_coordinate]
```
fujiisoup: Do we need to check `arr.coords[use_coordinate]` is certainly 1-dimensional?

xarray/core/dataarray.py
jhamman: good catch. we actually dropped the time/index/values methods in favor of a more explicit interface that uses `use_coordinate`.

xarray/core/missing.py
```diff
@@ -0,0 +1,306 @@
...
+
+    See Also
+    --------
+    scipy.interpolate.UnivariateSpline
+    '''
+    def __init__(self, xi, yi, method='spline', fill_value=None, order=3,
+                 **kwargs):
+        from scipy.interpolate import UnivariateSpline
+
+        if method is not 'spline':
```
fujiisoup: `if method is not 'spline':` -> `if method != 'spline':`

xarray/core/dataarray.py
```diff
@@ -1220,6 +1220,93 @@ def fillna(self, value):
...
+            other words, if there is a gap with more than this number of
+            consecutive NaNs, it will only be partially filled. Must be greater
+            than 0 or None for no limit.
+
+        Returns
+        -------
+        DataArray
+        '''
+        from .missing import bfill
+        return bfill(self, dim, limit=limit)
```
fujiisoup: Do we need bottleneck installed to use `bfill` or `ffill`? Maybe it should be noted in docstrings.

xarray/tests/test_missing.py
```diff
@@ -0,0 +1,378 @@
...
+        actual = da.interpolate_na(method=method, dim=dim)
+        expected = df.interpolate(method=method, axis=da.get_axis_num(dim))
+        np.testing.assert_allclose(actual.values, expected.values)
+
+
+@pytest.mark.parametrize('shape', [(8, 8), (1, 20), (20, 1)])
+@pytest.mark.parametrize('frac_nan', [0, 0.5, 1])
+@pytest.mark.parametrize('method', ['time', 'index', 'values', 'linear',
+                                    'nearest', 'zero', 'slinear',
+                                    'quadratic', 'cubic'])
```
fujiisoup: Do we need to test also for 'barycentric', 'krog', 'pchip', 'spline' methods?

xarray/core/dataarray.py
```diff
@@ -1220,6 +1220,93 @@ def fillna(self, value):
...
+                       **kwargs):
+        """Interpolate values according to different methods.
+
+        Parameters
+        ----------
+        dim : str
+            Specifies the dimension along which to interpolate.
+        method : {'linear', 'nearest', 'zero', 'slinear', 'quadratic', 'cubic',
+                  'polynomial', 'barycentric', 'krog', 'pchip',
+                  'spline'}, optional
```
fujiisoup: Pandas seems to support `Akima`.
Maybe we can also add this?
Scipy provides the [similar interface](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.interpolate.Akima1DInterpolator.html).

xarray/core/dataarray.py
```diff
@@ -1220,6 +1220,93 @@ def fillna(self, value):
...
+            String indicating which method to use for interpolation:
+
+            - 'linear': linear interpolation (Default). Additional keyword
```
fujiisoup: I noticed `pandas.interpolate(method='linear')` behaves differently.
Though I think our definition is more natural,  we can note this here?

xarray/core/missing.py
fujiisoup: Maybe we can also raise an appropriate error if coordinate is `MultiIndex`

jhamman: @fujiisoup - do we have a handy way to tell if an coordinate is a xarray DataArray/Multindex? Its not a subset of `pd.Multindex` but I need something like:

```Python
isinstance(index, (pd.MultiIndex, xr.MultiIndexCoordinate))
```

shoyer: Probably the cleanest and most comprehensive thing to do is to try to always cast the interpolation index to `float64`, while catching any error messages (to replace them with something more informative). When a `MultiIndex` is cast to numpy array it ends up with object dtype.

xarray/core/dataarray.py
jhamman: Are you referring to which values Pandas uses for the index?

fujiisoup: Yes.

pandas' [docstring](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.interpolate.html) says 
> ‘linear’: ignore the index and treat the values as equally spaced.
> ‘index’: use the actual numerical values of the index.   

the difference of which may correspond to our `use_coordinate` option [False/True].

Our `linear` option means a linear interpolation. 
I've just thought the same name with different meanings is sometimes confusing.
It is a very trivial point (and maybe a kind of overconcern).

jhamman:
@fujiisoup - care to take another pass here? I think this is done but would appreciate a few more reviews.

doc/computation.rst
```diff
@@ -70,10 +71,25 @@ for working with missing data from pandas:
...
 Like pandas, xarray uses the float value ``np.nan`` (not-a-number) to represent
 missing values.
 
+xarray objects also have an :py:meth:`~xarray.DataArray.interpolate_na` method
+for filling missing values via 1D interpolation.
+
+.. ipython:: python
+
+    x = xr.DataArray([0, 1, np.nan, np.nan, 2], dims=['x'],
+                     coords={'xx': [0, 1, 1.1, 1.9, 3]})
```
fujiisoup: I think here is a typo. Maybe
```python
x = xr.DataArray([0, 1, np.nan, np.nan, 2], dims=['x'],
                    coords={'xx': ('x', [0, 1, 1.1, 1.9, 3])})
```
or
```python
x = xr.DataArray([0, 1, np.nan, np.nan, 2], dims=['x'],
          coords={'x': [0, 1, 1.1, 1.9, 3]})
```
?

xarray/core/missing.py
fujiisoup: Maybe `npcompat.flip`?
And then `numpy=1.11` is also supported?

doc/computation.rst
jhamman: This was intentional.

fujiisoup: Really?
I got
`ValueError: coordinate xx has dimensions ('xx',), but these are not a subset of the DataArray dimensions ['x']`
from this script.

jhamman: You were right. I was missing the Variable constructor.

fujiisoup:
This looks good to me.
Thanks for the great work :)

I will merge it in this weekend if there are no further comments.

xarray/core/missing.py
jgerardsimcock: Would it make sense to have them both default to NaN in Xarray?

xarray/core/dataarray.py
```diff
@@ -1228,6 +1228,97 @@ def fillna(self, value):
...
+        --------
+        numpy.interp
+        scipy.interpolate
+        """
+        from .missing import interp_na
+        return interp_na(self, dim=dim, method=method, limit=limit,
+                         use_coordinate=use_coordinate, **kwargs)
+
+    def ffill(self, dim, limit=None):
+        '''Fill NaN values by propogating values forward
```
maxim-lian: no need to change now, but FYI PEP8 is `"""` on its own line IIRC

xarray/core/dataarray.py
```diff
@@ -1228,6 +1228,97 @@ def fillna(self, value):
         out = ops.fillna(self, value)
         return out
 
+    def interpolate_na(self, dim=None, method='linear', limit=None,
```
maxim-lian: Probably too late to be helpful - but are we sure about the name here? We don't generally add `_na` onto methods (`bfill_na`?), and pandas is `interpolate` only

xarray/core/missing.py
```diff
@@ -0,0 +1,330 @@
...
+        self.call_kwargs['nu'] = kwargs.pop('nu', 0)
+        self.call_kwargs['ext'] = kwargs.pop('ext', None)
+
+        if fill_value is not None:
+            raise ValueError('SplineInterpolator does not support fill_value')
+
+        self.f = UnivariateSpline(xi, yi, k=order, **self.cons_kwargs)
+
+
+def _apply_over_vars_with_dim(func, self, dim=None, **kwargs):
```
maxim-lian: This looks good! Could we use this elsewhere in the future? Or even put it on `Dataset`?

xarray/core/missing.py
```diff
@@ -0,0 +1,330 @@
...
+    '''wrapper for datasets'''
+    from .dataset import Dataset
+
+    ds = Dataset(coords=self.coords, attrs=self.attrs)
```
maxim-lian: If people are subclassing `Dataset` (we do internally), this will void their subclass. I think generally we should use `type(self)` or `self.__class__`

xarray/core/missing.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def ffill(arr, dim=None, limit=None):
+    '''forward fill missing values'''
+    import bottleneck as bn
+
+    axis = arr.get_axis_num(dim)
+
+    # work around for bottleneck 178
+    _limit = limit if limit is not None else arr.shape[axis]
```
maxim-lian: 👍

xarray/tests/test_missing.py
```diff
@@ -0,0 +1,441 @@
...
+
+    for (shape, frac_nan, method) in itertools.product(shapes, frac_nans,
+                                                       methods):
+
+        da, df = make_interpolate_example_data(shape, frac_nan)
+
+        for dim in ['time', 'x']:
+            actual = da.interpolate_na(method=method, dim=dim)
+            expected = df.interpolate(method=method, axis=da.get_axis_num(dim))
+            np.testing.assert_allclose(actual.values, expected.values)
```
maxim-lian: Perennial note for future tests: `xarray.tests.assert_close`

xarray/tests/test_missing.py
```diff
@@ -0,0 +1,441 @@
...
+        actual = da.interpolate_na(method=method, dim='time', **kwargs)
+        assert (da.count('time') <= actual.count('time')).all()
+
+
+@requires_np112
+@requires_scipy
+def test_interpolate_pd_compat_non_uniform_index():
+    shapes = [(8, 8), (1, 20), (20, 1), (100, 100)]
+    frac_nans = [0, 0.5, 1]
+    methods = ['time', 'index', 'values']
```
maxim-lian: obv don't change but for future tests: pytest handles these parameterizations v well

maxim-lian:
This is great! While I left some comments, I don't think any warrant changes now, as long as ppl are good with the `interpolate_na` name

xarray/tests/test_missing.py
jhamman: @maxim-lian - I was using the pytest parameterizations here but they weren't playing nice with the `@requires_foo` decorators.

xarray/core/missing.py
jhamman: Yes, certainly. I'm not 100% convinced we don't already have something like this but I didn't seem to find it. @shoyer, do you have thoughts on this?

xarray/tests/test_missing.py
maxim-lian: I can have a look. Do you remember what the setup was that didn't work? No stress if not

jhamman: I think skipped tests are returned as functions without arguments. The parameterizations then were passed to functions that didn't take arguments and that obviously didn't work.

So: 
```Python
@parameterize('foo', [1, 2])
@skip
def bar(foo):
    pass
```

becomes 
```Python
@parameterize('foo', [1, 2])
def bar():
    pass
```
when skipped.

xarray/core/dataarray.py
jhamman: See comment from @shoyer above: https://github.com/pydata/xarray/pull/1640#discussion_r145888076

xarray/tests/test_missing.py
jhamman: expected is a Pandas.DataFrame so I don't think this will work with `xarray.tests.assert_close`

xarray/tests/test_missing.py
maxim-lian: I think we should use `pytest.importorskip` - while @jreback is right that it [https://github.com/pandas-dev/pandas/issues/18190#issuecomment-353460062](doesn't allow as much control) as writing a function ourselves, and @WillAyd gives some [more benefits](https://github.com/pandas-dev/pandas/issues/18190#issuecomment-353460062) , we're not using anything that the standard pytest function can't do, in any of our functions.

If people agree, I could add a note to this function: https://github.com/pydata/xarray/pull/1640/files/48505a51e69d6b3663783db414962f3011e94aa1#diff-e84d0339b6da4d18c02f6b2d54fbd5bdR47

xarray/core/missing.py
shoyer: +1 let's use NaN here.

xarray/core/dataarray.py
```diff
@@ -1228,6 +1228,97 @@ def fillna(self, value):
...
+        -------
+        DataArray
+        '''
+        from .missing import ffill
+        return ffill(self, dim, limit=limit)
+
+    def bfill(self, dim, limit=None):
+        '''Fill NaN values by propogating values backward
+
+        *Requires bottleneck and Numpy v1.12.0 or later.*
```
shoyer: I don't think NumPy 1.12 is still required with our `flip` backport

xarray/core/dataset.py
```diff
@@ -2410,6 +2410,105 @@ def fillna(self, value):
...
+        '''
+        from .missing import ffill, _apply_over_vars_with_dim
+
+        new = _apply_over_vars_with_dim(ffill, self, dim=dim, limit=limit)
+        return new
+
+    def bfill(self, dim, limit=None):
+        '''Fill NaN values by propogating values backward
+
+        *Requires bottleneck and Numpy v1.12.0 or later.*
```
shoyer: as above: numpy 1.12 is not required with our backport.

xarray/tests/test_missing.py
shoyer: I recall we had some difficulty with decorators based on pytest and subclass methods, but that doesn't apply here.

fujiisoup:
ready to go?

fujiisoup:
Thanks for this great work!

PR #1642, opened 2017-10-20
Fix rasterio backend for rasterio=1.0a10

### LINKED ISSUE #1641: Rasterio backend broken with rasterio=1.0a10
The latest alpha of rasterio (`1.0a10`) breaks the xarray backend (tested against `xarray=0.9.6`) because it expects either a `tuple` or `rasterio.windows.Window` passed as the `window` keyword argument to `read`. Tuples get converted to a `Window` in the latest alpha, but xarray currently passes the window as a list. Simply converting the window argument to a tuple fixes the backend ([see here for the fix I'm proposing](https://github.com/ceholden/xarray/commit/8dc7e748e1af273ed13f0d3dacc48a80e4741c21)).

Somewhat related, the CI services are currently testing against `rasterio=0.36` while this backend was originally tested against one of the 1.0 alphas. The cause is the `rasterio-feedstock` repo switched the development alpha releases to be distributed on the `conda-forge/label/dev` channel, instead of the regular `conda-forge` channel. Switching to the development channel for `1.0a10` should cause this issue to manifest itself on the CI services.

Happy to address these (and add to whats-new, etc) in the same or separate issues/PRs, depending on what you think makes sense. Thanks for all your work!

ceholden:
- [x] Closes #1641
 - [x] Tests added / passed

First commit should fix rasterio backend for the current alpha release `rasterio=1.0a10` and for future stable releases. Behind the scenes rasterio converts to a `Window` using `Window.from_slices`. I thought about doing the conversion within xarray and passing a `Window` directly, but that would require importing rasterio within the `__getitem__`, which seems not very great.

Second commit updates the CI service configurations to use the conda-forge development channel that the rasterio alphas are published on. I put the dev channel behind existing channels, and pinned the version of rasterio to anything `1.0` and beyond. Not so sure what the best course of action here is and I'm happy to take suggestions.

Still have to add to the `whats-new`. I'm assuming I can ignore flake8 warnings from the conda environment files (e.g., `git diff upstream/master -- '*.py' | flake8 --diff`)?

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -74,7 +74,7 @@ class RasterioArrayWrapper(NdimSizeLenMixin, DunderArrayMixin):
                     raise IndexError(_ERROR_MSG)
             window.append((start, stop))
 
-        out = self.rasterio_ds.read(band_key, window=window)
+        out = self.rasterio_ds.read(band_key, window=tuple(window))
         if squeeze_axis:
             out = np.squeeze(out, axis=squeeze_axis)
         return out
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,8 @@ matrix:
     env: CONDA_ENV=py36-condaforge-rc
   - python: 3.6
     env: CONDA_ENV=py36-pynio-dev
+  - python: 3.6
+    env: CONDA_ENV=py36-rasterio1.0alpha
   allow_failures:
   - python: 3.6
     env:
@@ -63,6 +65,8 @@ matrix:
     env: CONDA_ENV=py36-condaforge-rc
   - python: 3.6
     env: CONDA_ENV=py36-pynio-dev
+  - python: 3.6
+    env: CONDA_ENV=py36-rasterio1.0alpha
 
 before_install:
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
--- a/ci/requirements-py36-rasterio1.0alpha.yml
+++ b/ci/requirements-py36-rasterio1.0alpha.yml
@@ -1,0 +1,24 @@
+name: test_env
+channels:
+  - conda-forge
+  - conda-forge/label/dev
+dependencies:
+  - python=3.6
+  - dask
+  - distributed
+  - h5py
+  - h5netcdf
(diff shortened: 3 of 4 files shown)


xarray/backends/rasterio_.py
```diff
@@ -74,7 +74,7 @@ def __getitem__(self, key):
                     raise IndexError(_ERROR_MSG)
             window.append((start, stop))
 
-        out = self.rasterio_ds.read(band_key, window=window)
```
jhamman: will this change work with rasterio < v1.0?

ceholden: The [docstring](https://github.com/mapbox/rasterio/blob/96e63f02b5929b43db56a79e06efd748b8d89fa9/rasterio/_io.pyx#L674) for 0.36 says it takes a tuple of pairs of ints, so it should work (as it did on my machine). I've added another environment file specific to rasterio<=1.0 to test these conditions.

fmaussion: Actually the default should be to test against 0.36 (which is the current stable version) and to add an "allowed failure" test for 1.Xa in the matrix. But maybe I'm being pedantic here...

jhamman: I'm +1 on @fmaussion's idea here. I think we want to test against 0.36 until rasterio has an official 1.0 release. We can add an "allowed failure" test to integrate 1.0 into our test suite. 

The good news is that this change seems to work for both versions so there's not anything huge to worry about here.

ceholden: Makes sense and certainly reduced what I needed to change! I added a new entry to the test matrix for py36 and rasterio>=1.* from the conda-forge development channel.

Unfortunately I don't think Travis sees the new matrix entry. Not sure if it's related, but I backed out the commits that had changed many of the environment files, hoping to tidy up all the modifications. Sorry if this complicates the PR.

fmaussion: > Unfortunately I don't think Travis sees the new matrix entry

You need to add the entry to both the `matrix` and `allow_failures` lists

ceholden: Thanks for catching this! Should be fixed now.

Issue #687, opened 2015-12-28
Dataset.T property creates conflict with data named 'T'


brian-rose:
I frequently find myself working with datasets that include a variable named `'T'` (for temperature!)

The problem with this is that, unlike any other variable, `mydataset.T` does not return the variable, because there is a built-in property `.T` which is just a map to the `mydataset.transpose()` method.

This is mostly just a minor annoyance, because of course I can access the data as `mydataset['T']`, but it necessitates an odd asymmetry in any code that accesses `T` as well as other data. And, maybe more importantly, `mydataset.T` does NOT immediately raise an error in code that is attempting to access the temperature data.

I have not been able to simply remove the property interactively from the `dataset` objects, e.g.
`del mydataset.T`
raises an error
`AttributeError: can't delete attribute`

Since the transpose functionality is already easily accessible through `mydataset.transpose()` (which is more readable, in my opinion), perhaps the `T` property should just be removed?

MaximilianR:
Removing `.T` would break so much existing code, I don't think that's going to get much support. It's also consistent with numpy.

If you use IPython, tab-complete will work, so if your column is called `temperature`, you can access it almost as quickly...

jhamman:
Agreed, I don't think we want to get rid of the transpose method `T`.  A better solution for you would probably be to rename your temperature variable using the `rename` method.  I often do this when comparing multiple datasets that use different variable naming conventions.

brian-rose:
Brilliant, somehow I missed the `rename` method in the documentation. Also, I missed the fact that `.T` is consistent with numpy. I will just make a habit of renaming all my variables called `T` from now on. Thanks for these comments.

PR #1643, opened 2017-10-21
Deprecate Dataset.T as an alias for Dataset.transpose()

### LINKED ISSUE #1232: Remove the `Dataset.T` property?
The fact that `ds.T` transposes a dataset rather than returning `ds['T']` is a repeated point of confusion:
http://stackoverflow.com/questions/41879269/xarray-dataarray-opens-as-a-dataset/41885639#41885639
http://stackoverflow.com/questions/40270929/how-to-read-a-variable-called-t-with-xarray

shoyer:
- [x] Closes #1232
 - [x] Tests added / passed

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -158,7 +158,6 @@ Computation
 :py:attr:`~Dataset.imag`
 :py:attr:`~Dataset.round`
 :py:attr:`~Dataset.real`
-:py:attr:`~Dataset.T`
 :py:attr:`~Dataset.cumsum`
 :py:attr:`~Dataset.cumprod`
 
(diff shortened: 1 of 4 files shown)


rabernat:
This is great. I hit this annoyance all the time with any dataset from the IRI data library:
http://iridl.ldeo.columbia.edu/

The time dimension is always called `T`.

brian-rose:
I'm happy to see this too!
I raised this way back in https://github.com/pydata/xarray/issues/687, but at the time the consensus was that `.T` notation was preferred for transposing because of consistency with numpy. 
I've been using `.rename` on all variables called `T` ever since -- usually temperature in my use cases.

MaximilianR:
Good to go back and read your own comments: 

> Removing .T would break so much existing code, I don't think that's going to get much support. It's also consistent with numpy.

Issue #1267, opened 2017-02-14
"in" operator does not work as expected on DataArray dimensions


hottwaj:
As an example I have a DataArray called "my_dataarray" that looks something like this:

```
<xarray.DataArray 'values' (Type: 3)>
array([1, 2, 3])
Coordinates:
  * Type        (Type) object 'Type 1' 'Type 2' 'Type 3'
```

'Type' is a dimension on my DataArray.  Note that 'Type' is also a DataArray that looks like this:

```
OrderedDict([('Type', <xarray.IndexVariable 'Type' (Type: 3)>
array(['Type 1', 'Type 2', 'Type 3'], 
      dtype='object'))])
```

Let's say I run:

```
'Type 1' in my_dataarray.Type
```

The result is False, even though 'Type 1' is in the "Type" dimension.

To get the result I was expecting I need to run:

```
'Type 1' in my_dataarray.Type.values
```

Stepping through the code, the problematic line is here:
https://github.com/pydata/xarray/blob/20ec32430fac63a8976699d9528b5fdc1cd4125d/xarray/core/dataarray.py#L487

The test used for `__contains__(self, key)` on the Type dimension is whether the key is in the `_coords` of Type.  

This is probably the right thing to do when the DataArray is used for storing data, but probably not what we want if the DataArray is being used as a dimension -  it should instead check if 'Type 1' is in the *values* of Type?

shoyer:
I agree, this is strange. `my_dataarray['Type']` is really just a convenient shortcut for `my_dataarray.coords['Type']`, but `DataArray` is meant to be primarily array-like so `in` should look at the values (we have `Dataset` for the dict-like container).

I would be in favor of changing this to look at values like `in` on NumPy arrays in the next major release (0.10.0). That's probably a little ways off though, so let's hold off on the pull request for now :).

shoyer:
https://github.com/pydata/xarray/pull/1645 adds a FutureWarning. We'll actually change this in v0.11.

PR #1645, opened 2017-10-22
Deprecate checking coordinates in DataArray.__contains__

shoyer:
In the future, this should check array values. For now, issue a FutureWarning.

 - [x] xref #1267
 - [x] Tests added / passed

### DIFF
--- a/xarray/core/rolling.py
+++ b/xarray/core/rolling.py
@@ -205,7 +205,10 @@ class DataArrayRolling(Rolling):
                    for _, window in self]
 
         # Find valid windows based on count
-        concat_dim = self.window_labels if self.dim in self.obj else self.dim
+        if self.dim in self.obj.coords:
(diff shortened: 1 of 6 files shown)


xarray/core/rolling.py
```diff
@@ -205,7 +205,10 @@ def reduce(self, func, **kwargs):
                    for _, window in self]
 
         # Find valid windows based on count
-        concat_dim = self.window_labels if self.dim in self.obj else self.dim
+        if self.dim in self.obj.coords:
+            concat_dim = self.window_labels
+        else:
+            concat_dim = self.dim
```
jhamman: Was this the only place we were looking for coords in this way?

shoyer: There were a couple more. Verified by converting the warning into an error and running the test suite.

PR #1654, opened 2017-10-24
[DOCS] PyNIO is now available on conda-forge

jsignell:
Just a docs change. Updated instructions for installing PyNIO to use conda-forge.


shoyer:
Thanks!

PR #1657, opened 2017-10-25
fixes for warnings related to unit tests and nan comparisons

### LINKED ISSUE #1652: Resolve warnings issued in the xarray test suite
82 warnings are currently issued in the process of running our test suite:
https://gist.github.com/shoyer/db0b2c82efd76b254453216e957c4345

Some of can probably be safely ignored, but others are likely noticed by users, e.g.,
https://stackoverflow.com/questions/41130138/why-is-invalid-value-encountered-in-greater-warning-thrown-in-python-xarray-fo/41147570#41147570

It would be nice to clean up all of these, either by catching the appropriate upstream warning (if irrelevant) or changing our usage to avoid the warning. There may very well be a lurking FutureWarning in there somewhere that could cause issues when another library updates.

Probably the easiest way to get started here is to get the test suite running locally, and use `py.test -W error` to turn all warnings into errors.

jhamman:
- [x] Closes #1652
 - [x] Tests added / passed

This first commit just includes the fixes for the warnings issued in our unit tests. I've temporarily added `-W error` to the travis builds so we can see the warnings/errors for all environments. Next step is to address the warnings in the xarray code itself.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1590,10 +1590,9 @@ class DataArray(AbstractArray, BaseDataObject):
             other_variable = getattr(other, 'variable', other)
             other_coords = getattr(other, 'coords', None)
 
-            with np.errstate(all='ignore'):
-                variable = (f(self.variable, other_variable)
-                            if not reflexive
-                            else f(other_variable, self.variable))
+            variable = (f(self.variable, other_variable)
+                        if not reflexive
+                        else f(other_variable, self.variable))
             coords = self.coords._merge_raw(other_coords)
             name = self._result_name(other)
 
@@ -1615,8 +1614,7 @@ class DataArray(AbstractArray, BaseDataObject):
             other_coords = getattr(other, 'coords', None)
             other_variable = getattr(other, 'variable', other)
             with self.coords._merge_inplace(other_coords):
-                with np.errstate(all='ignore'):
-                    f(self.variable, other_variable)
+                f(self.variable, other_variable)
             return self
 
         return func
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1571,7 +1571,9 @@ class DataArray(AbstractArray, BaseDataObject):
     def _unary_op(f):
         @functools.wraps(f)
         def func(self, *args, **kwargs):
-            return self.__array_wrap__(f(self.variable.data, *args, **kwargs))
+            with np.errstate(all='ignore'):
+                return self.__array_wrap__(f(self.variable.data, *args,
+                                             **kwargs))
 
         return func
 
@@ -1588,9 +1590,10 @@ class DataArray(AbstractArray, BaseDataObject):
             other_variable = getattr(other, 'variable', other)
             other_coords = getattr(other, 'coords', None)
 
-            variable = (f(self.variable, other_variable)
-                        if not reflexive
-                        else f(other_variable, self.variable))
+            with np.errstate(all='ignore'):
+                variable = (f(self.variable, other_variable)
+                            if not reflexive
+                            else f(other_variable, self.variable))
             coords = self.coords._merge_raw(other_coords)
             name = self._result_name(other)
 
@@ -1612,7 +1615,8 @@ class DataArray(AbstractArray, BaseDataObject):
             other_coords = getattr(other, 'coords', None)
             other_variable = getattr(other, 'variable', other)
             with self.coords._merge_inplace(other_coords):
-                f(self.variable, other_variable)
+                with np.errstate(all='ignore'):
+                    f(self.variable, other_variable)
             return self
 
         return func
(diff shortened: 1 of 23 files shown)


jhamman:
Down to 96 (was 372) warnings in my py36 test environment.

Two changes that I could uses some input on: 

##### 1 - Numpy element wise comparison

```Python
    def assertEqual(self, a1, a2):
>       assert a1 == a2 or (a1 != a1 and a2 != a2)
E       DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
```

@shoyer - you and I put together the assertEqual method a while back. Do we need to check the size of arrays before comparing the values?

##### 2 - Numpy invalid value comparison

e.g.: 
```Python
RuntimeWarning: invalid value encountered in greater_equal 
```

I'm hoping there is a succinct way of wrapping many of our nan comparing functions in `numpy.errstat`. Thoughts here? I've run out of time for tonight but will pick this up again tomorrow.

jhamman:
@shoyer - I'm wondering if we should attack #1652 in stages. The remaining warnings are going to take a bit more effort and this first block included a lot of line changes. I'm a bit concerned that if we let this sit, we will end up with a million merge conflicts.

shoyer:
> I'm wondering if we should attack #1652 in stages

Yes, absolutely. Let's do it in stages.

jhamman:
Okay, so this knocks off the lowest hanging fruit. I'll pull the pytest error out of the travis build and we'll move forward with this.

xarray/core/dataarray.py
```diff
@@ -1588,9 +1590,10 @@ def func(self, other):
             other_variable = getattr(other, 'variable', other)
             other_coords = getattr(other, 'coords', None)
 
-            variable = (f(self.variable, other_variable)
-                        if not reflexive
-                        else f(other_variable, self.variable))
+            with np.errstate(all='ignore'):
```
shoyer: I don't think we need this here -- this should be caught at the level of the `Variable` wrapper.

xarray/core/dataarray.py
```diff
@@ -1612,7 +1615,8 @@ def func(self, other):
             other_coords = getattr(other, 'coords', None)
             other_variable = getattr(other, 'variable', other)
             with self.coords._merge_inplace(other_coords):
-                f(self.variable, other_variable)
+                with np.errstate(all='ignore'):
```
shoyer: Also probably unneeded

shoyer:
@jhamman sorry, looks like you'll need to merge in master, too -- I just put in some conflicting changes with my unicode to netCDF3 PR.

jhamman:
@shoyer - updated. If all the tests pass, I'll merge tonight.

shoyer:
This looks good to me. Can you add a note to "what's new" (under bug fixes) about removing the warning when comparing with NaN?

xarray/tests/test_backends.py
```diff
@@ -1040,7 +1043,7 @@ def test_nc4_scipy(self):
             with nc4.Dataset(tmp_file, 'w', format='NETCDF4') as rootgrp:
                 rootgrp.createGroup('foo')
 
-            with self.assertRaisesRegexp(TypeError, 'pip install netcdf4'):
+            with raises_regex(TypeError, 'pip install netcdf4'):
                 open_dataset(tmp_file, engine='scipy')
```
jhamman: @shoyer - any ideas why this test is failing on Appveyor Py3.6: https://ci.appveyor.com/project/shoyer/xray/build/1.0.2844/job/uarr5gsny3wpx9qm#L3326

xarray/tests/test_backends.py
```diff
@@ -1019,14 +1018,18 @@ def create_store(self):
...
-        with self.assertRaisesRegexp(ValueError, 'must be 1-dimensional'):
+        with raises_regex(ValueError, 'must be 1-dimensional'):
             with self.roundtrip(ds) as roundtripped:
                 pass
 
     def test_roundtrip_example_1_netcdf_gz(self):
-        with open_example_dataset('example_1.nc.gz') as expected:
-            with open_example_dataset('example_1.nc') as actual:
-                self.assertDatasetIdentical(expected, actual)
+        if sys.version_info[:2] < (2, 7):
```
shoyer: please revert to my changes here

xarray/tests/test_backends.py
shoyer: Well, we aren't closing this file properly, so I'm not too surprised that Windows is complaining. I don't know why it's only showing up now, though.

For now, just add `allow_cleanup_failure=True` to the `create_tmp_file()` call. We'll tackle this later in #1668.

jhamman:
@shoyer - I think this is all good now. We seem to have acquired another unrelated build failures though...

shoyer:
Everything passed after the pandas 0.21 fixes in #1669

PR #1665, opened 2017-10-27
DOC: added apply example to core/dataarray.py

raybellwaves:
- [NA] Closes #xxxx
 - [NA] Tests added / passed
 - [NA] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

I added an example to apply as I just learnt about it from my Q on stackoverflow (https://stackoverflow.com/questions/46980823/python-xarray-compute-the-absolute-values-of-a-dataset/46981027#46981027).

Also my first time using git and making edits on the command line rather than the web browser.


xarray/core/dataset.py
```diff
@@ -2431,6 +2431,25 @@ def apply(self, func, keep_attrs=False, args=(), **kwargs):
         attrs = self.attrs if keep_attrs else None
         return type(self)(variables, attrs=attrs)
 
+        Examples
```
shoyer: This needs to be in the docstring, not the python code!

xarray/core/dataset.py
```diff
@@ -2431,6 +2431,25 @@ def apply(self, func, keep_attrs=False, args=(), **kwargs):
...
+        --------
```
shoyer: add an extra line break here

shoyer:
Thanks @raybellwaves! This will be helpful.

xarray/core/dataset.py
raybellwaves: I shouldn't do PR's on a Friday afternoon!

shoyer:
Thanks!

PR #1667, opened 2017-10-28
Tweak to to_dask_dataframe()

shoyer:
Follow on to #1489.

- Add a `dim_order` argument
- Always write columns for each dimension
- Docstring to NumPy format

 - [x] Tests added / passed

CC @jmunroe @jhamman

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2606,6 +2606,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         dask.dataframe.DataFrame
         """
 
+        import dask.array as da
         import dask.dataframe as dd
 
         if dim_order is None:
@@ -2621,14 +2622,15 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
         columns.extend(k for k in self.coords if k not in self.dims)
         columns.extend(self.data_vars)
 
-        data = []
+        series_list = []
         for name in columns:
             try:
                 var = self.variables[name]
             except KeyError:
                 # dimension without a matching coordinate
-                values = np.arange(self.dims[name], dtype=np.int64)
-                var = Variable((name,), values)
+                size = self.dims[name]
+                data = da.arange(size, chunks=size, dtype=np.int64)
+                var = Variable((name,), data)
 
             # IndexVariable objects have a dummy .chunk() method
             if isinstance(var, IndexVariable):
@@ -2636,9 +2638,9 @@ class Dataset(Mapping, ImplementsDatasetReduce, BaseDataObject,
 
             dask_array = var.set_dims(ordered_dims).chunk(self.chunks).data
             series = dd.from_array(dask_array.reshape(-1), columns=[name])
-            data.append(series)
+            series_list.append(series)
 
-        df = dd.concat(data, axis=1)
+        df = dd.concat(series_list, axis=1)
 
         if set_index:
             if len(dim_order) == 1:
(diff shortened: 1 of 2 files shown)


xarray/core/dataset.py
```diff
@@ -2577,63 +2577,76 @@ def from_dataframe(cls, dataframe):
...
+                'dim_order {} does not match the set of dimensions on this '
+                'Dataset: {}'.format(dim_order, list(self.dims)))
 
-            # ensure all variables span the same dimensions
-            v = v.set_dims(ordered_dims)
+        ordered_dims = OrderedDict((k, self.dims[k]) for k in dim_order)
 
-            # ensure all variables have the same chunking structure
-            if v.chunks != chunks:
-                v = v.chunk(chunks)
```
shoyer: @jmunroe was there was reason why you didn't just chunk everything here?

jmunroe: On the assumption (probably mistaken) that there was a cost to calling .chunk(chunks) on a variable that already had that chunking structure.  If that assumption was not correct, then, yes, everything could just be chunked.

shoyer: Rechunk is nearly free if chunks are unchanged -- it actually returns the same dask array object.

shoyer:
@jhamman can you please take a look here when you have the chance?

xarray/tests/test_dask.py
```diff
@@ -595,12 +598,24 @@ def test_to_dask_dataframe_2D(self):
...
+    @pytest.mark.xfail(raises=NotImplementedError)
+    def test_to_dask_dataframe_2D_set_index(self):
+        # This will fail until dask implements MultiIndex support
+        w = da.from_array(np.random.randn(2, 3), chunks=(1, 2))
+        ds = Dataset({'w': (('x', 'y'), w)})
+        ds['x'] = ('x', np.array([0, 1], np.int64))
+        ds['y'] = ('y', list('abc'))
+
+        expected = ds.compute().to_dataframe()
+        actual = ds.to_dask_dataframe(set_index=True)
```
jhamman: rather than using `xfail` above, use `raises_regex` to make sure we raise an error in the correct line.

shoyer: My reasoning on using xfail was that that makes this test more robust. If/when dask implementing MultiIndex, we'll just get an unexpected xfail rather than a failing test. `NotImplementedError` seems specific enough (unlike, e.g., ValueError) that I'm not concerned about grepping for  the exact error message.

xarray/core/dataset.py
```diff
@@ -2577,63 +2577,77 @@ def from_dataframe(cls, dataframe):
...
-
-        data = []
-        for k in columns:
-            v = self._variables[k]
-
-            # consider coordinate variables as well as data varibles
-            if isinstance(v, xr.IndexVariable):
-                v = v.to_base_variable()
+        if dim_order is None:
+            dim_order = list(self.dims)
```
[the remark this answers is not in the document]
shoyer: For Dataset, it's a SortedKeysDict (i.e., the dimensions in alphabetical order).

xarray/core/dataset.py
```diff
@@ -2577,63 +2577,77 @@ def from_dataframe(cls, dataframe):
...
 
-            # reshape variable contents as a 1d array
-            d = v.data.reshape(-1)
+        data = []
+        for name in columns:
+            try:
+                var = self.variables[name]
+            except KeyError:
+                # dimension without a matching coordinate
+                values = np.arange(self.dims[name], dtype=np.int64)
```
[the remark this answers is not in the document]
shoyer: Yes, good idea, will do

PR #1676, opened 2017-10-31
Support orthogonal indexing in MemoryCachedArray (Fix for #1429)

### LINKED ISSUE #1429: Orthogonal indexing and MemoryCachedArray
While working on https://github.com/pydata/xarray/pull/1260 I came upon this which looks like a bug in caching:

```python
import numpy as np
import xarray as xr
from xarray.core import indexing

nx, ny = 8, 10
data = np.arange(nx*ny).reshape(ny, nx)
cached = indexing.MemoryCachedArray(data)

data = xr.DataArray(data=data, dims=('y', 'x'))
cached = xr.DataArray(data=cached, dims=('y', 'x'))

a = data.isel(x=[2, 4], y=[3, 5])
b = cached.isel(x=[2, 4], y=[3, 5])
```
The last line raises:

```
<ipython-input-13-45cd1493cf6b> in <module>()
     11 
     12 a = data.isel(x=[2, 4], y=[3, 5])
---> 13 b = cached.isel(x=[2, 4], y=[3, 5])

/home/mowglie/Documents/git/xarray/xarray/core/dataarray.py in isel(self, drop, **indexers)
    668         DataArray.sel
    669         """
--> 670         ds = self._to_temp_dataset().isel(drop=drop, **indexers)
    671         return self._from_temp_dataset(ds)
    672 

/home/mowglie/Documents/git/xarray/xarray/core/dataset.py in isel(self, drop, **indexers)
   1141         for name, var in iteritems(self._variables):
   1142             var_indexers = dict((k, v) for k, v in indexers if k in var.dims)
-> 1143             new_var = var.isel(**var_indexers)
   1144             if not (drop and name in var_indexers):
   1145                 variables[name] = new_var

/home/mowglie/Documents/git/xarray/xarray/core/variable.py in isel(self, **indexers)
    547             if dim in indexers:
    548                 key[i] = indexers[dim]
--> 549         return self[tuple(key)]
    550 
    551     def squeeze(self, dim=None):

/home/mowglie/Documents/git/xarray/xarray/core/variable.py in __getitem__(self, key)
    380         # orthogonal indexing should ensure the dimensionality is consistent
    381         if hasattr(values, 'ndim'):
--> 382             assert values.ndim == len(dims), (values.ndim, len(dims))
    383         else:
    384             assert len(dims) == 0, len(dims)

AssertionError: (1, 2)
```

fujiisoup:
- [x] Closes #1429
 - [x] Tests added / passed

This bug originates from the complicated structure around the array wrappers and their indexing,
i.e. different array wrappers support different indexing types, and moreover, some can store another array wrapper in it.

I made some cleanups.
+ Now every array wrapper is a subclass of `NDArrayIndexable`
+ Every array wrapper should implement its own `__getitem__` or just store another `NDArrayIndexable`.

I think I added enough test for it, but I am not yet fully accustomed with xarray's backend.
There might be many combinations of their hierarchical relation.

I would appreciate any comments.

### DIFF
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -57,8 +57,10 @@ class BaseNetCDF4Array(indexing.NDArrayIndexable):
 class NetCDF4ArrayWrapper(BaseNetCDF4Array):
     def __getitem__(self, key):
         if isinstance(key, indexing.VectorizedIndexer):
-            raise NotImplementedError('{} does not support vectorized '
-                                      'indexing'.format(self.__class__))
+            raise NotImplementedError(
+             'Vectorized indexing for {} is not implemented. Load your '
+             'data first with .load() or .compute().'.format(type(self)))
+
         key = indexing.to_tuple(key)
 
         if self.datastore.is_remote:  # pragma: no cover
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -10,9 +10,7 @@ from .. import conventions
 from .. import Variable
 from ..conventions import pop_to
 from ..core import indexing
-from ..core.utils import (FrozenOrderedDict, NdimSizeLenMixin,
-                          DunderArrayMixin, close_on_error,
-                          is_remote_uri)
+from ..core.utils import FrozenOrderedDict, close_on_error, is_remote_uri
 from ..core.pycompat import iteritems, basestring, OrderedDict, PY3, suppress
 
 from .common import (WritableCFDataStore, robust_getitem,
@@ -27,13 +25,13 @@ _endian_lookup = {'=': 'native',
                   '|': 'native'}
 
 
-class BaseNetCDF4Array(NdimSizeLenMixin, DunderArrayMixin):
+class BaseNetCDF4Array(indexing.NDArrayIndexable):
     def __init__(self, variable_name, datastore):
         self.datastore = datastore
         self.variable_name = variable_name
 
         array = self.get_array()
-        self.shape = array.shape
+        self._shape = array.shape
 
         dtype = array.dtype
         if dtype is str:
@@ -41,7 +39,15 @@ class BaseNetCDF4Array(NdimSizeLenMixin, DunderArrayMixin):
             # represent variable length strings; it also prevents automatic
             # string concatenation via conventions.decode_cf_variable
             dtype = np.dtype('O')
-        self.dtype = dtype
+        self._dtype = dtype
+
+    @property
+    def shape(self):
+        return self._shape
+
+    @property
+    def dtype(self):
+        return self._dtype
 
     def get_array(self):
         self.datastore.assert_open()
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -56,6 +56,9 @@ class BaseNetCDF4Array(indexing.NDArrayIndexable):
 
 class NetCDF4ArrayWrapper(BaseNetCDF4Array):
     def __getitem__(self, key):
+        if isinstance(key, indexing.VectorizedIndexer):
+            raise NotImplementedError('{} does not support vectorized '
+                                      'indexing'.format(self.__class__))
         key = indexing.to_tuple(key)
 
         if self.datastore.is_remote:  # pragma: no cover
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -10,7 +10,9 @@ from .. import conventions
 from .. import Variable
 from ..conventions import pop_to
 from ..core import indexing
-from ..core.utils import FrozenOrderedDict, close_on_error, is_remote_uri
+from ..core.utils import (FrozenOrderedDict, NdimSizeLenMixin,
+                          DunderArrayMixin, close_on_error,
+                          is_remote_uri)
 from ..core.pycompat import iteritems, basestring, OrderedDict, PY3, suppress
 
 from .common import (WritableCFDataStore, robust_getitem,
@@ -25,13 +27,14 @@ _endian_lookup = {'=': 'native',
                   '|': 'native'}
 
 
-class BaseNetCDF4Array(indexing.NDArrayIndexable):
+class BaseNetCDF4Array(NdimSizeLenMixin, DunderArrayMixin,
+                       indexing.NDArrayIndexable):
     def __init__(self, variable_name, datastore):
         self.datastore = datastore
         self.variable_name = variable_name
 
         array = self.get_array()
-        self._shape = array.shape
+        self.shape = array.shape
 
         dtype = array.dtype
         if dtype is str:
@@ -39,15 +42,7 @@ class BaseNetCDF4Array(indexing.NDArrayIndexable):
             # represent variable length strings; it also prevents automatic
             # string concatenation via conventions.decode_cf_variable
             dtype = np.dtype('O')
-        self._dtype = dtype
-
-    @property
-    def shape(self):
-        return self._shape
-
-    @property
-    def dtype(self):
-        return self._dtype
+        self.dtype = dtype
 
     def get_array(self):
         self.datastore.assert_open()
--- a/xarray/core/indexing.py
+++ b/xarray/core/indexing.py
@@ -303,7 +303,11 @@ class VectorizedIndexer(IndexerTuple):
     """ Tuple for vectorized indexing """
 
 
-class LazilyIndexedArray(utils.NDArrayMixin):
+class NDArrayIndexable(utils.NDArrayMixin):
+    """ A base array wrapper to support vectorized indexing """
+
+
+class LazilyIndexedArray(NDArrayIndexable):
     """Wrap an array that handles orthogonal indexing to make indexing lazy
     """
     def __init__(self, array, key=None):
@@ -356,7 +360,7 @@ class LazilyIndexedArray(utils.NDArrayMixin):
         return tuple(shape)
 
     def __array__(self, dtype=None):
-        array = xarray_indexable(self.array)
+        array = as_indexable(self.array)
         return np.asarray(array[self.key], dtype=None)
 
     def __getitem__(self, key):
@@ -379,9 +383,9 @@ def _wrap_numpy_scalars(array):
         return array
 
 
-class CopyOnWriteArray(utils.NDArrayMixin):
+class CopyOnWriteArray(NDArrayIndexable):
     def __init__(self, array):
-        self.array = array
+        self.array = as_indexable(array)
         self._copied = False
 
     def _ensure_copied(self):
@@ -400,9 +404,9 @@ class CopyOnWriteArray(utils.NDArrayMixin):
         self.array[key] = value
 
 
-class MemoryCachedArray(utils.NDArrayMixin):
+class MemoryCachedArray(NDArrayIndexable):
     def __init__(self, array):
-        self.array = _wrap_numpy_scalars(array)
+        self.array = _wrap_numpy_scalars(as_indexable(array))
 
     def _ensure_cached(self):
         if not isinstance(self.array, np.ndarray):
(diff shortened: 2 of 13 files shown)


xarray/backends/netCDF4_.py
```diff
@@ -27,21 +25,29 @@
...
             # use object dtype because that's the only way in numpy to
             # represent variable length strings; it also prevents automatic
             # string concatenation via conventions.decode_cf_variable
             dtype = np.dtype('O')
-        self.dtype = dtype
+        self._dtype = dtype
+
+    @property
+    def shape(self):
+        return self._shape
```
MaximilianR: Why are all the private methods / public properties needed?

xarray/tests/test_backends.py
```diff
@@ -399,6 +399,24 @@ def test_orthogonal_indexing(self):
...
+        # Make sure vectorized_indexing works or at leaset raises
+        # NotImplementedError
+        in_memory = create_test_data()
+        with self.roundtrip(in_memory) as on_disk:
+            indexers = {'dim1': DataArray([0, 2, 0], dims='a'),
+                        'dim2': DataArray([0, 2, 3], dims='a')}
+            expected = in_memory.isel(**indexers)
+            try:
+                actual = on_disk.isel(**indexers)
+                self.assertDatasetIdentical(expected, actual)
```
MaximilianR: FYI for the future `xarray.tests.assert_identical` is a better interface

xarray/tests/test_variable.py
```diff
@@ -1681,3 +1683,75 @@ def test_raise_no_warning_for_nan_in_binary_ops():
...
+    def check_orthogonal_indexing(self, v):
+        assert np.allclose(v.isel(x=[8, 3], y=[2, 1]),
+                           self.d[[8, 3]][:, [2, 1]])
+
+    def check_vectorized_indexing(self, v):
+        ind_x = Variable('z', [0, 2])
+        ind_y = Variable('z', [2, 1])
+        assert np.allclose(v.isel(x=ind_x, y=ind_y), self.d[ind_x, ind_y])
+
+    def test_NumpyIndexingAdapter(self):
```
MaximilianR: FYI for the future, I think these combinations could be handled through pytest parameterization well (but within the existing framework, this looks good)

xarray/backends/netCDF4_.py
fujiisoup: This is because `NDArrayMixin` (the parent of `NDArrayIndexable`) requires this property.
We can avoid this by using the double inheritance... but I think the above is simpler.

xarray/tests/test_backends.py
fujiisoup: I adopted `assert_identical` only where I added. 
After the approval, I will change the other part of `test_backends`

fujiisoup:
@fmaussion 
How did you notice issue #1429?
I found test_backends.py already checked the orthogonal indexing.

I guess there are some untested use cases.

xarray/tests/test_variable.py
fujiisoup: Thanks.
Do you mean something similar to [this](https://github.com/pydata/xarray/blob/master/xarray/tests/test_dataset.py#L4063)?
If more tests are needed, I will adopt this feature.

xarray/tests/test_backends.py
```diff
@@ -393,11 +393,29 @@ def test_orthogonal_indexing(self):
...
+            expected = in_memory.isel(**indexers)
+            try:
+                actual = on_disk.isel(**indexers)
+                assert_identical(expected, actual)
+                # do it twice, to make sure we're switched from
+                # orthogonal -> numpy when we cached the values
+                actual = on_disk.isel(**indexers)
+                assert_identical(expected, actual)
+            except NotImplementedError:
+                pass
```
jhamman: I think I understand what this test is going for but its not all that clear when its going to pass because the indexing is supported and when its going to pass because it raised a `NotImplementedError`. Can we tease this into two tests somehow?

xarray/tests/test_variable.py
MaximilianR: Yes exactly! Thanks

xarray/backends/netCDF4_.py
MaximilianR: Ah yes, I've had this problem in python before. Agree this is a better compromise than double inheritance

xarray/tests/test_backends.py
fujiisoup: Thanks. I separated the test. Currently, only the dask-related wrappers support vindex.

xarray/core/indexing.py
```diff
@@ -419,14 +423,21 @@ def __setitem__(self, key, value):
...
+    if isinstance(array, NDArrayIndexable):
+        return array
     if isinstance(array, np.ndarray):
         return NumpyIndexingAdapter(array)
     if isinstance(array, pd.Index):
         return PandasIndexAdapter(array)
     if isinstance(array, dask_array_type):
         return DaskIndexingAdapter(array)
-    return array
+    raise TypeError('Invalid array type: {}'.format(type(array)))
```
shoyer: The default assumption previously was that any array inside an `xarray.Variable` supported indexing. With this change, we only support indexing when objects are instances of `NDArrayIndexable`. It seems like it might make more sense to rather never allow creating `Variable` objects with this invalid data in the first place, e.g., put this check in `xarray.core.variable.as_compatible_data()` instead.

xarray/core/indexing.py
```diff
@@ -303,7 +303,11 @@ class VectorizedIndexer(IndexerTuple):
     """ Tuple for vectorized indexing """
 
 
-class LazilyIndexedArray(utils.NDArrayMixin):
+class NDArrayIndexable(utils.NDArrayMixin):
```
shoyer: Rather than inheriting an implementation (which requires all the various changes to implementations above), why not simply make this a dummy mixin class instead?

We could still inherit from `NDArrayIndexable` to mark that a class supports indexing, but it would not add any methods, e.g.,
```python
class NDArrayIndexable(object):
    """Mixin to mark support for IndexerTuple subclasses in indexing."""
```

shoyer:
Also, I should say: thank you again for diving into the internals here. This sort of clean-up work is extremely important, even though it isn't as visible as new features.

fmaussion:
>  @fmaussion How did you notice issue #1429?
> I found test_backends.py already checked the orthogonal indexing.
> I guess there are some untested use cases.

I don't remember exactly, but it happened when I wrote tests for rasterio's indexing: https://github.com/pydata/xarray/blob/master/xarray/tests/test_backends.py#L1855-L1884 (I just tested all possible combinations on index operations I could think of). The caching mechanism has some non-intuitive consequences, for example that some indexing operations not allowed on a rasterio file stored on disk become available once the data is cached:  https://github.com/pydata/xarray/blob/master/xarray/tests/test_backends.py#L1916-L1927

xarray/core/variable.py
```diff
@@ -197,6 +190,10 @@ def as_compatible_data(data, fastpath=False):
         else:
             data = np.asarray(data)
 
+    # validate whether the data is valid data types
+    if not isinstance(data, SUPPORT_ARRAY_TYPES):
+        data = np.asarray(data)
```
fujiisoup: I think we should check `data` is an array-like and raise an Error if not.
(`np.asarray` converts any object to a `np.ndarray`.)

Maybe we need to follow [this](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects) definition of `array-like`?
Do we already have any utility function for this?

xarray/core/variable.py
```diff
@@ -31,6 +31,7 @@
     pass
 
 
+SUPPORT_ARRAY_TYPES = (indexing.NDArrayIndexable, np.ndarray, pd.Series)
```
shoyer: For our sanity, I think this list should match the list of types accepted as valid by `as_indexable()`. So I think that would mean we should drop `pd.Series` but add `dask_array_type` (note that it's a tuple) and `pd.Index`.

xarray/tests/test_backends.py
```diff
@@ -393,11 +393,34 @@ def test_orthogonal_indexing(self):
...
-            self.assertDatasetIdentical(expected, actual)
+            assert_identical(expected, actual)
             # do it twice, to make sure we're switched from orthogonal -> numpy
             # when we cached the values
             actual = on_disk.isel(**indexers)
-            self.assertDatasetIdentical(expected, actual)
+            assert_identical(expected, actual)
+
+    def _test_vectorized_indexing(self, vindex_support=True):
+        # Make sure vectorized_indexing works or at leaset raises
```
shoyer: leaset -> least

xarray/tests/test_variable.py
```diff
@@ -1676,8 +1679,97 @@ def test_ones_like(self):
...
+        v = Variable(dims=('x', 'y'), data=NumpyIndexingAdapter(self.d))
+        self.check_orthogonal_indexing(v)
+        self.check_vectorized_indexing(v)
+        # could not doublly wrapping
```
shoyer: doublly -> doubly

xarray/tests/test_backends.py
```diff
@@ -393,11 +393,34 @@ def test_orthogonal_indexing(self):
...
+                # orthogonal -> numpy when we cached the values
+                actual = on_disk.isel(**indexers)
+                assert_identical(expected, actual)
+            else:
+                with raises_regex(NotImplementedError, 'Vectorized indexing '):
+                    actual = on_disk.isel(**indexers)
+
+    def test_vectorized_indexing(self):
+        # This test should be overwritten if vindex is supported
+        self._test_vectorized_indexing(vindex_support=False)
```
shoyer: I think scipy should also support vectorized indexing? But maybe not because of caching?

fujiisoup: Yes. `LazilyIndexingArray` does not yet support vectorized indexing.
Maybe it should be improved.

shoyer: Would it be confusing to make vectorized indexing trigger loading data into memory, i.e., it's not eager, not lazy? That would seem pretty reasonable to me.

xarray/core/variable.py
```diff
@@ -159,12 +160,7 @@ def as_compatible_data(data, fastpath=False):
     if isinstance(data, Variable):
         return data.data
 
-    # add a custom fast-path for dask.array to avoid expensive checks for the
-    # dtype attribute
-    if isinstance(data, dask_array_type):
-        return data
-
-    if isinstance(data, pd.Index):
+    if isinstance(data, SUPPORT_ARRAY_TYPES):
```
shoyer: I don't think this is quite right -- we still cast certain types of ndarray arguments at the end of this function (e.g., with `_possibly_convert_objects`). I'm somewhat surprised tests aren't failing.

xarray/tests/test_backends.py
shoyer: We could also save this for later.

xarray/core/variable.py
fujiisoup: `SUPPORT_ARRAY_TYPES` does not include `np.ndarray` (I changed from the previous commit), and some type-castings for `np.ndarray` in below are still working here.

(But maybe my naming `SUPPORT_ARRAY_TYPES` is not good as `np.ndarray` is also supported.)

fujiisoup:
@fmaussion 
Thanks for the information. Is there any need to add indexing tests for `rasterio` backend, after this issue #1429 is resolved?

fmaussion:
> Is there any need to add indexing tests for rasterio backend, after this issue #1429 is resolved?

No, I don't think so. Thanks!

shoyer:
I renamed `SUPPORT_ARRAY_TYPES` to `NON_NUMPY_SUPPORTED_ARRAY_TYPES` which is longer but more obvious in its meaning. I think this is good to go in now.

shoyer:
Thanks again @fujiisoup !

PR #1679, opened 2017-11-01
Use numpy's normalize_axis_index, if available

eric-wieser:
Curious if you think this is a good idea


shoyer:
Yes, this looks like a good idea. Though I'd prefer if the logic were in `core/npcompat.py` instead, which is where the rest of our NumPy backports are.

max-sixty:
@eric-wieser any interest in moving this to `npcompat.py`?

PR #1682, opened 2017-11-02
Add option “engine”

ghost:
Implements a new xarray option `engine` for setting the default backend data read/write engine. Inspired by [this](https://stackoverflow.com/a/47041952) Stack Overflow answer.

This PR is not ready for merge yet but I wanted to verify if the code changes are on the right track.

The default `engine` option value is `None`. If this option is set the `_get_default_engine()` function will return its value without going through the import statements chain.

 - [ ] Closes #xxxx
 - [ ] Tests added / passed


xarray/core/options.py
```diff
@@ -6,6 +6,7 @@
 OPTIONS = {
     'display_width': 80,
     'arithmetic_join': 'inner',
+    'engine': None,
```
MaximilianR: Should this have the current default rather than `None`?

ajelenak-thg: I did not want to alter the current behavior. My understanding is that unless a user specifies an engine it is effectively `None` which then invokes a `_get_default_engine()` call to figure out the most appropriate one based on the file path.

MaximilianR:
This looks great so far!

But have you tested this? I have a suspicion that putting `OPTIONS['engine']` as a default arg won't work: whenever the relevant function is imported, the default will be set to the value at that time, and won't change if the options dictionary is later modified. 

You could have a default arg of `None` and then look up the option within the function, if my suspicions are right

shoyer:
I like the direction this is going in, but as @MaximilianR notes it won't work to set this in default function arguments.

Also, I would probably pick a more descriptive name for the option, e.g., `netcdf_engine`.

ajelenak-thg:
How about `io_engine` for the option's name? The data can come from an OPeNDAP server as well.

I have reverted to the original `engine=None` in the functions/methods and use
```python
engine = engine or OPTIONS['io_engine']
```
to assign the correct `engine` value.

shoyer:
We should probably have separate options for opendap/netcdf. It would not be so useful to try to use pydap to read a netcdf file.

ajelenak-thg:
Yes, there could be more I/O engine options. How about `file_engine` and `web_engine`? Keeping the naming more generic as there may be other file-based formats or data web services in the future.

On the other hand, setting this global option should indicate a willingness to accept the consequences. If automatic selection of the optional I/O engine is preferred, this global option should not be set.

shoyer:
Maybe it would make more sense to set a list of engines to try in order? e.g., `io_engines=['pydap', 'h5netcdf', 'scipy']` could use pydap for web requests, h5netcdf for netCDF4 and scipy for netCDF3. The default would be something like `['netcdf4', 'pydap', 'scipy', 'h5netcdf']`

ajelenak-thg:
Let's see if we can get this PR over the line... 😄 

A list of engines would need some way of declaring their I/O capabilities: only file-based, only HTTP-based, or both. Something like:

```python
io_engines = [
    {'engine': 'netcdf4', 
     'capabilities': ['file', 'http']},

    {'engine': 'pydap', 
     'capabilities': ['http']},

    {'engine': 'scipy', 
     'capabilities': ['file']},
     
    {'engine': 'h5netcdf',
     'capabilities': ['file']},
]
```

On xarray import or any time this option would change, the list of engines would be checked to remove unavailable engines.

shoyer:
One more minor complexity is that there are two version of netCDF, not all of which are supported by all tools:

- netcdf4 support netcdf3 and netcdf4
- scipy only supports netcdf3
- h5netcdf only supports netcdf4.
- pynio supports both netcdf3 and netcdf4.

So will need to distinguish between URLs (opendap) and these two types of netCDF files. To do this, we'll need to look at the first few bytes of each file:
- netCDF4/HDF5 starts with the 8-byte signature `b'\x89HDF\r\n\x1a\n'`
- netCDF4 starts with `b'CDF\x00'` or `b'CDF\x01'`

ajelenak-thg:
I've refactored setting the I/O engine option as per our discussion. Hopefully, it captures now all the requested functionality.

xarray/backends/api.py
```diff
@@ -15,34 +15,40 @@
...
 DATAARRAY_NAME = '__xarray_dataarray_name__'
 DATAARRAY_VARIABLE = '__xarray_dataarray_variable__'
 
 
+def _file_startswith(path, *args):
+    """Check that the file statswith with the sepcific byte sequence."""
+    with open(path, 'rb') as f:
+        for b in args:
+            sig = f.read(len(b))
+            print(sig, b)
```
shoyer: probably should remove :)

xarray/backends/api.py
```diff
@@ -15,34 +15,40 @@
...
-                engine = 'scipy'
-            except ImportError:
-                raise ValueError('cannot read or write netCDF files without '
-                                 'netCDF4-python or scipy installed')
+        for engine in OPTIONS['io_engines']:
+            if engine in ('netcdf4', 'h5netcdf') \
+                    and _file_startswith(path, b'\x89HDF\r\n\x1a\n'):
+                return engine
+            elif engine == 'scipy' \
+                    and _file_startswith(path, b'CDF\x00', b'CDF\x01'):
```
shoyer: The netcdf4 package can also read netcdf3 files. Yes, it's a confusingly named package!

xarray/backends/api.py
```diff
@@ -15,34 +15,40 @@
...
-        except ImportError:  # pragma: no cover
-            try:
-                import scipy.io.netcdf
-                engine = 'scipy'
-            except ImportError:
-                raise ValueError('cannot read or write netCDF files without '
-                                 'netCDF4-python or scipy installed')
+        for engine in OPTIONS['io_engines']:
+            if engine in ('netcdf4', 'h5netcdf') \
+                    and _file_startswith(path, b'\x89HDF\r\n\x1a\n'):
```
shoyer: Rather than reopening files each time an engine is checked, can we do this once?

Maybe utility function like `_get_file_format(path)`  which returns either 'netcdf3', 'netcdf4' or 'unknown'?

xarray/core/options.py
```diff
@@ -1,11 +1,37 @@
...
+    _AVAILABLE_IO_ENGINES.append('scipy')
+except ImportError:
+    pass
+try:
+    import h5netcdf.legacyapi  # noqa
+    _AVAILABLE_IO_ENGINES.append('h5netcdf')
+except ImportError:
+    pass
+if len(_AVAILABLE_IO_ENGINES) == 0:
+    raise RuntimeError('No suitable I/O engines found installed')
```
shoyer: None of these backends are required for using xarray, so this error will need to get raised when opening/saving a file instead, not merely when importing `xarray.core.options`.

xarray/core/options.py
```diff
@@ -18,6 +44,9 @@ class set_options(object):
       Default: ``80``.
     - ``arithmetic_join``: DataArray/Dataset alignment in binary operations.
       Default: ``'inner'``.
+    - ``io_engines``: List of backend data I/O engines. Default: All the
```
shoyer: should add something like "to try in order when saving/loading data without an explicit engine"

xarray/backends/api.py
```diff
@@ -15,34 +15,40 @@
...
-            engine = 'netcdf4'
-        except ImportError:
-            try:
-                import pydap
-                engine = 'pydap'
-            except ImportError:
-                raise ValueError('netCDF4 or pydap is required for accessing '
-                                 'remote datasets via OPeNDAP')
+        for engine in OPTIONS['io_engines']:
+            if engine in ('netcdf4', 'pydap', 'h5netcdf'):
```
shoyer: h5netcdf doesn't work for opendap

ajelenak-thg: OPeNDAP is not the only data web service protocol. There could be others in the future, like OGC Web Coverage Service, etc., all with URLs beginning with `http`.

shoyer: Certainly... but as yet h5netcdf/h5py doesn't support any of them :)

xarray/core/options.py
ajelenak-thg: This exception will be raised only when none of the known I/O engines can be imported, netCDF4 included. I assume netCDF4 is the only one required so I think it is safe to keep it as is.

shoyer: Not even netCDF4 is a required dependency. We only require numpy and pandas: http://xarray.pydata.org/en/stable/installing.html

shoyer: Obviously, you need an I/O engine to read/write files. But not to import xarray.

ajelenak-thg:
Now that the tests are passing again, is there anything else left to change?

xarray/backends/api.py
```diff
@@ -15,35 +15,53 @@
...
-    return engine
+        filefmt = _guess_file_format(path)
+        for engine in OPTIONS['io_engines']:
+            if filefmt == 'hdf5' and engine in ('netcdf4', 'h5netcdf'):
+                return engine
+            elif filefmt == 'netcdf3' and engine in ('netcdf4', 'scipy'):
+                return engine
+        else:
+            raise ValueError('cannot read or write netCDF files without '
+                             'netCDF4-python, scipy, or h5netcdf installed')
```
shoyer: We need a different error message here for the case that `filefmt == 'unknown'`. Currently, if you try to open a non-netCDF file (e.g., a text file) with `open_dataset()` you will get this error message telling you that you don't have a netCDF library installed.

xarray/backends/api.py
```diff
@@ -15,35 +15,53 @@
...
+def _guess_file_format(path):
+    """Make an educated guess about the file format based on the first few
+    bytes.
+
+    If the file does not exist, assume netcdf3.
+    """
+    try:
+        f = open(path, 'rb')
+    except IOError:
+        return 'netcdf3'
```
shoyer: Is this really better than raising an error for a missing file? I think it's probably fine to surface the default error message from `open()`, which looks something like: `FileNotFoundError: [Errno 2] No such file or directory: 'asdolfkjalsdkfj'`

xarray/core/options.py
```diff
@@ -41,6 +69,21 @@ def __init__(self, **kwargs):
             raise ValueError('argument names %r are not in the set of valid '
                              'options %r' % (invalid_options, set(OPTIONS)))
         self.old = OPTIONS.copy()
+
+        if 'io_engines' in kwargs:
```
shoyer: This clause needs test coverage -- somewhere in https://github.com/pydata/xarray/blob/master/xarray/tests/test_options.py

xarray/core/options.py
```diff
@@ -41,6 +69,21 @@ def __init__(self, **kwargs):
...
+            if isinstance(kwargs['io_engines'], basestring):
+                kwargs['io_engines'] = [kwargs['io_engines']]
+            new_engines = list()
+            for e in kwargs['io_engines']:
+                if e in _AVAILABLE_IO_ENGINES:
+                    new_engines.append(e)
+                else:
+                    raise ValueError('I/O engine %s not installed' % e)
```
shoyer: since you only raise an error for invalid engines, I think this logic could be simplified:
```
new_engines  = list(kwargs['io_engines'])
for engine in new_engines:
    if engine not in _AVAILABLE_IO_ENGINES:
        raise ValueError(...)
```

xarray/core/options.py
```diff
@@ -41,6 +69,21 @@ def __init__(self, **kwargs):
...
+            if not new_engines:
+                raise ValueError('No I/O engines')
```
shoyer: I can see use cases for disabling the default IO engine altogether (e.g., if you want to always require an explicit choice). So let's not disallow this.

xarray/backends/api.py
```diff
@@ -15,35 +15,53 @@
...
-            try:
-                import scipy.io.netcdf
-                engine = 'scipy'
-            except ImportError:
-                raise ValueError('cannot read or write netCDF files without '
-                                 'netCDF4-python or scipy installed')
-    return engine
+        filefmt = _guess_file_format(path)
+        for engine in OPTIONS['io_engines']:
+            if filefmt == 'hdf5' and engine in ('netcdf4', 'h5netcdf'):
```
shoyer: nit: usual Python style is to avoid abbreviation in variable names. Prefer `file_format` instead.

xarray/backends/api.py
shoyer: Also, please include the list of available default IO engines in the error message.

xarray/core/options.py
```diff
@@ -18,6 +42,10 @@ class set_options(object):
       Default: ``80``.
     - ``arithmetic_join``: DataArray/Dataset alignment in binary operations.
       Default: ``'inner'``.
+    - ``io_engines``: List of backend data I/O engines to try in order when
+      saving/loading data without an explicit engine. Default: All the
+      installed engines on start up from this list: ['netcdf4', 'pydap',
+      'scipy', 'h5netcdf'].
```
shoyer: Thinking about the interface here again, would it make sense to use a mapping {file_format: engine} instead? That would be more flexible than a single list with priority order. For example: `{'netcdf3': 'scipy', 'netcdf4': 'h5netcdf', 'remote': 'pydap'}`.

Or perhaps these should all be separate options, e.g., `default_netcdf3_engine`, `default_netcdf4_engine` and `default_remote_engine`.

xarray/backends/api.py
```diff
@@ -15,35 +15,53 @@
 from ..core.combine import auto_combine
 from ..core.utils import close_on_error, is_remote_uri
 from ..core.pycompat import basestring, path_type
+from ..core.options import OPTIONS
 
 DATAARRAY_NAME = '__xarray_dataarray_name__'
 DATAARRAY_VARIABLE = '__xarray_dataarray_variable__'
 
 
+def _guess_file_format(path):
```
shoyer: A unit test for this function would be quite valuable (and would be pretty easy to write).

xarray/backends/api.py
```diff
@@ -15,35 +15,53 @@
...
+
+    if sig[:8] == b'\x89HDF\r\n\x1a\n':
+        return 'hdf5'
+    elif sig[:4] == b'CDF\x02' or sig[:4] == b'CDF\x01':
+        return 'netcdf3'
+    else:
+        return 'unknown'
+
+
 def _get_default_engine(path, allow_remote=False):
```
shoyer: It would also be nice to have a test here.

xarray/backends/api.py
ajelenak-thg: There were tests that needed to save data to a netCDF file and they were failing here because these files did not exist. Recommending `netcdf3` here for such cases made sense to me because it enabled selection of an appropriate I/O engine (either `netcdf4` or `scipy`).

xarray/core/options.py
ajelenak-thg: I think how `_get_default_engine()` does it is sufficient for now.

xarray/core/options.py
ajelenak-thg: This is allowed: `xarray.set_options(io_engines='pydap')`; this should not be allowed: `xarray.set_options(io_engines='foo')` unless `foo` engine is added to xarray.

xarray/core/options.py
shoyer: @jhamman @fmaussion any thoughts on API design here?

jhamman: I think I'm generally okay here with just providing a list. Since this option only affects the netcdf-like engines, it would be good to be clear about that in the doc string. Pynio, rasterio, and zarr are also io_engines but this option does not control their behavior. Perhaps the name of the option should be changed. Alternatively, an approach like @shoyer recommends would allow for an explicit mapping for any number of file types (not just netcdf).

ajelenak-thg: The current design allows adding other Python packages and associating them with the file format they support. Let me know what other packages you would like to see added.

For remote data access engines, I think nothing super-smart can be done because all of their URLs will start with `http`. Setting a default engine option (this PR) or using the `engine` keyword is inevitable.

shoyer:
@ajelenak-thg thanks for your patience here. I need to think a little bit harder about what the right API for controlling the default engine is, but this work is definitely appreciated (and will probably make it in for the next release).

PR #1696, opened 2017-11-07
ffill & bfill methods

### LINKED ISSUE #1651: ENH: Forward & back fill methods
I think with `np.flip` and `bn.push`, this should be simple. They're both fairly new and so would require version checks / upgrading the minimums.

One small issue, I wonder if anyone has come across this: `bottleneck` returns the numpy array rather than the DataArray - is that because it's not operating with the correct numpy interface?

Forward fill:
```
array.values = bn.push(array.values, axis=array.get_axis_num(axis_name))
```

Backfill:
```
axis = array.get_axis_num(axis_name)

# reverse for bfill
array = np.flip(array, axis=axis)

# fill 
array.values = bn.push(array.values, axis=axis)

# reverse back to original
result = np.flip(scaling, axis=date_axis)
```

MaximilianR:
- [x] Closes #1651 
 - [x] Tests added / passed

No docs / docstrings. No backfill.

But otherwise is this a reasonable layout? 
Do we prefer `fillna` with options (back / forward / replacement), or `ffill`, `bfill`, `fillna`?

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -153,6 +153,22 @@ def fillna(data, other, join="left", dataset_join="left"):
                        dataset_fill_value=np.nan,
                        keep_attrs=True)
 
+def ffill(data, dim, limit=None):
+    axis_num = data.get_axis_num(dim)
+
+    if not has_bottleneck:
+        raise ImportError('ffill requires bottleneck to be installed')
+
+    data = data.copy()
+
+    if limit is None:
+        # bottleneck raises an error if you pass `None`
+        data.values = bn.push(data.values, axis=axis_num)
+    else:
+        data.values = bn.push(data.values, axis=axis_num, n=limit)
+
+    return data
+
 
 def where_method(self, cond, other=dtypes.NA):
     """Return elements from `self` or `other` depending on `cond`.
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1214,6 +1214,9 @@ class DataArray(AbstractArray, BaseDataObject):
         out = ops.fillna(self, value)
         return out
 
+    def ffill(self, dim, limit=None):
+        return ops.ffill(self, dim, limit)
+
     def combine_first(self, other):
         """Combine two DataArray objects, with union of coordinates.
 
--- a/xarray/tests/test_dataarray.py
+++ b/xarray/tests/test_dataarray.py
@@ -3146,3 +3146,21 @@ def test_raise_no_warning_for_nan_in_binary_ops():
     with pytest.warns(None) as record:
         xr.DataArray([1, 2, np.NaN]) > 0
     assert len(record) == 0
+
+
+@pytest.mark.parametrize('da', (1, 2), indirect=True)
+def test_ffill_functions(da):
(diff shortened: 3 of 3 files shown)


shoyer:
> Do we prefer fillna with options (back / forward / replacement), or ffill, bfill, fillna?

I like separate methods. It keeps things logically well separated.

xarray/core/ops.py
```diff
@@ -153,6 +153,22 @@ def fillna(data, other, join="left", dataset_join="left"):
...
+    axis_num = data.get_axis_num(dim)
+
+    if not has_bottleneck:
+        raise ImportError('ffill requires bottleneck to be installed')
+
+    data = data.copy()
+
+    if limit is None:
+        # bottleneck raises an error if you pass `None`
+        data.values = bn.push(data.values, axis=axis_num)
```
shoyer: Try `apply_ufunc` + `transpose()` afterwards instead? That would work on arbitrary xarray objects. It's awkward to need to restore the original dimension order, but it adds almost no overhead given NumPy's memory model. @jhamman is doing something like that in #1640.

I think you could even use `dask='parallelized'` here, to allow for doing multiple forward fills at the same time. 

Parallelizing across the pushed axis with dask would be harder (and certainly isn't necessary for a first pass), because you would need something hierarchical that can push data between chunks. But I think you could still make this work with multiple passes to get forward fill-values on the boundaries.

jhamman:
@MaximilianR - I have rough implementation of both of these on my interpolate branch: 

https://github.com/jhamman/xarray/blob/56f37144f2dc827b61aa61f94c69277e6a79c262/xarray/core/missing.py#L206-L234

some slight modifications gives this:

```Python
def _bfill(arr, n=None, axis=-1):
    '''inverse of ffill'''
    import bottleneck as bn

    arr = np.flip(arr, axis=axis)

    # fill
    arr = bn.push(arr, axis=axis, n=n)

    # reverse back to original
    return np.flip(arr, axis=axis)

def ffill(self, dim=None, limit=None):
    ''' '''
    import bottleneck as bn

    return apply_ufunc(bn.push, self,
                       dask='forbidden',
                       keep_attrs=True,
                       kwargs=dict(n=limit, axis=self.get_axis_num(dim))).transpose(*arr.dims)

def bfill(self, dim=None, limit=None):
    ''' '''
    return apply_ufunc(_bfill, self,
                       dask='forbidden',
                       keep_attrs=True,
                       kwargs=dict(n=limit, axis=self.get_axis_num(dim))).transpose(*arr.dims)
```

As @shoyer says, it may be possible to change to `dask='parallelized'` but I have tried it yet.

Feel free to take what you want from this snippet, I am moving pretty slowly on my interpolation branch but it would be great if you can get this in as a stand alone feature.

jhamman:
@MaximilianR - quick update. My implementation of ffill/bfill in #1640 is a bit cleaner now and seems to be working. It probably makes sense for me to just finish off this feature in that PR. Maybe you could give it a review when you get a chance?

MaximilianR:
@jhamman Perfect! I had a read through yours and it looks great, excited to get this merged.
I'll close this

PR #1702, opened 2017-11-08
fix empty dataset from_dict

chrisbarber:
- [ ] Closes #xxxx
 - [x] Tests added / passed

Not sure if you want an issue or a whats-new entry for a small fix like this.

Also not sure how xarray tends to handle `np.array([])`. One option that I did not take was to provide ndims to `as_compatible_data` in some fashion and make sure the result comes back as having shape `(0,)*ndim` instead of just `(0,)`.

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -427,7 +427,7 @@ class Variable(common.AbstractArray, utils.NdimSizeLenMixin):
         if isinstance(dims, basestring):
             dims = (dims,)
         dims = tuple(dims)
-        if len(dims) != self.ndim:
+        if self.shape != (0,) and len(dims) != self.ndim:
             raise ValueError('dimensions %s must have the same length as the '
                              'number of data dimensions, ndim=%s'
                              % (dims, self.ndim))
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -2871,6 +2871,26 @@ class TestDataset(TestCase):
                                      "without the key 'dims'"):
             Dataset.from_dict(d)
 
+    def test_to_and_from_dict_with_zero_dim(self):
+        ds = Dataset(OrderedDict([('a', ('t', np.array([], dtype='float64'))),
+                                  ('b', ('t', np.array([], dtype='float64'))),
+                                  ('t', ('t', np.array([], dtype='U1')))]))
+        expected = {'coords': {'t': {'dims': ('t',),
+                                     'data': [],
+                                     'attrs': {}}},
+                    'attrs': {},
+                    'dims': {'t': 0},
+                    'data_vars': {'a': {'dims': ('t',),
+                                        'data': [],
+                                        'attrs': {}},
+                                  'b': {'dims': ('t',),
+                                        'data': [],
+                                        'attrs': {}}}}
+
+        actual = ds.to_dict()
+        self.assertEqual(expected, actual)
+        self.assertDatasetIdentical(ds, Dataset.from_dict(actual))
+
     def test_to_and_from_dict_with_time_dim(self):
         x = np.random.randn(10, 3)
         y = np.random.randn(10, 3)


xarray/core/variable.py
```diff
@@ -427,7 +427,7 @@ def _parse_dimensions(self, dims):
         if isinstance(dims, basestring):
             dims = (dims,)
         dims = tuple(dims)
-        if len(dims) != self.ndim:
+        if self.shape != (0,) and len(dims) != self.ndim:
```
shoyer: This would allow you to create an empty Variable with the wrong number of dimensions, so I don't think this is the right way to fix this.

Something like `xarray.Variable((), [])` should still be an error, because we require labeling all dimensions on a Variable, regardless of their size.

xarray/tests/test_dataset.py
```diff
@@ -2871,6 +2871,26 @@ def test_to_and_from_dict(self):
...
+                    'data_vars': {'a': {'dims': ('t',),
+                                        'data': [],
+                                        'attrs': {}},
+                                  'b': {'dims': ('t',),
+                                        'data': [],
+                                        'attrs': {}}}}
+
+        actual = ds.to_dict()
+        self.assertEqual(expected, actual)
+        self.assertDatasetIdentical(ds, Dataset.from_dict(actual))
```
jhamman: We're trying to move away from using the UnitTest assert methods. Instead, try this:

```Python
assert actual == expected
assert_identical(ds, Dataset.from_dict(actual))
```

xarray/core/variable.py
chrisbarber: Huh, I was thinking that `[]` would be valid with any number of dims. So for example, `xarray.Variable((), [])` would be equivalent to `xarray.Variable((), np.empty(()))`, or `xarray.Variable(('a', 'b'), [])` would be like `xarray.Variable(('a', 'b'), np.empty((0,0)))`.

This interpretation of `[]` would be consistent with the way `to_dict` currently deals with empty variables. For example, `xr.Dataset({'foo': (['a', 'b'], np.empty((0,0)))}).to_dict()['data_vars']['foo']['data']` is `[]`.

But `Variable` seems to interpret `[]` not as having ambiguous dimensionality, but rather `ndim=1`. `to_dict` could instead write `[[]]` for data of shape `(0,0)`, but then there is nothing sensible to write for data of shape `()` that I can think of. Also, `Variable` gives things like `[[]]` or `[[[]]]` leading sizes of 1, like `(1,0)` or `(1,1,0)`, which brings up another issue with trying to infer shape information from empty arrays.

Without changing `Variable`'s current interpretation of the shape of `[]`, `[[]]`, `[[[]]]`, ..., I guess this will have to be explicitly handled in `from_dict`, either locally or by providing a `shape` override on `Variable`.

shoyer: Variable is just casting its inputs with `np.array()` here. Note that `np.array([]).shape == (0,)`.

The culprit seems to be that NumPy's `.tolist()` method is not invertible for size zero arrays. All trailing dimensions seem to be dropped:
```
>>> np.empty((0,)).tolist()
[]
>>> np.empty((0, 2, 3)).tolist()
[]
>>> np.empty((2, 3, 0)).tolist()
[[[], [], []], [[], [], []]]
```

So it appears that we will indeed need to special our logic for restoring these arrays, to take note of the variable's dimensions and their sizes.

shoyer:
> Not sure if you want an issue or a whats-new entry for a small fix like this.

Yes, this would be still be appreciated. Something like "Fix Dataset.from_dict() for size zero inputs" in the bug-fixes section would suffice.

xarray/core/variable.py
chrisbarber: Here's an especially weird one:
```
>>> np.empty(()).tolist()
1.309863837007e-312
```
As expected, it also shows up here:
```
>>> xr.Dataset({'a': np.empty(())}).to_dict()
{'coords': {}, 'attrs': {}, 'dims': {}, 'data_vars': {'a': {'data': 4.4e-323, 'dims': (), 'attrs': {}}}}
```
From a cursory look it appears that xarray does try to handle 0d arrays, so I guess this is another thing that technically ought to be handled.

chrisbarber:
Just doing some garbage-collection; it looks like this was somehow fixed. This works in 0.10.0 and 0.10.3:
```
>>> ds = xr.Dataset({'a': ('b', [])})
>>> xr.Dataset.equals(ds, xr.Dataset.from_dict(ds.to_dict()))
True
```

Issue #1617, opened 2017-10-10
__array_ufunc__ for xarray


shoyer:
Now that `__array_ufunc__` has appeared in NumPy 1.13, we should be able to resolve a long-standing annoyance with xarray: the inability to use NumPy ufuncs like `np.sin()` directly on all xarray objects, such as `xarray.Dataset`.

I believe this could be straightforwardly accomplished by wrapping `xarray.core.computation.apply_ufunc`.

Some care should be taken to ensure that properly defer to other array types, as suggested [in the NEP](https://docs.scipy.org/doc/numpy-1.13.0/neps/ufunc-overrides.html). This also could be a good time to resolve some of the xarray's broader consistency issues when interacting with and wrapping other array-like types.

shoyer:
A related task would be finishing up this PR:

- #1118: Do not convert subclasses of `ndarray` unless required

Ultimately, it would be nice to have some sort of registration system, especially for objects that aren't ndarray subclasses, that indicates that a class can safely be wrapped by xarray.

I'm thinking something like `xarray.register_data_type(ArrayWithUnits)` which indicates that `.data` for xarray objects can now be an `ArrayWithUnits` object. This would also possibly require passing a reference to a set of function to use on data containing these objects (e.g., for stacking or concatenating).

shoyer:
I'm starting to work on this.

PR #1705, opened 2017-11-09
Make Indexer classes not inherit from tuple.

shoyer:
I'm not entirely sure this is a good idea. The advantage is that it ensures that
all our indexing code is entirely explicit: everything that reaches a backend
*must* be an ExplicitIndexer. The downside is that it removes a bit of
internal flexibility: we can't just use tuples in place of basic indexers
anymore. On the whole, I think this is probably worth it but I would appreciate
feedback.

@fujiisoup can you take a look?

 - [x] Tests added / passed

### DIFF
--- a/xarray/core/indexing.py
+++ b/xarray/core/indexing.py
@@ -3,6 +3,7 @@ from __future__ import division
 from __future__ import print_function
 from datetime import timedelta
 from collections import defaultdict, Hashable
+import operator
 import numpy as np
 import pandas as pd
 
@@ -300,6 +301,17 @@ class ExplicitIndexer(object):
         return '{}({})'.format(type(self).__name__, self.tuple)
 
 
+def as_integer_or_none(value):
+    return None if value is None else operator.index(value)
+
+
+def as_integer_slice(value):
+    start = as_integer_or_none(value.start)
+    stop = as_integer_or_none(value.stop)
+    step = as_integer_or_none(value.step)
+    return slice(start, stop, step)
+
+
 class BasicIndexer(ExplicitIndexer):
     """Tuple for basic indexing.
 
@@ -307,6 +319,22 @@ class BasicIndexer(ExplicitIndexer):
     rules for basic indexing: each axis is independently sliced and axes
     indexed with an integer are dropped from the result.
     """
+    def __init__(self, key):
+        if not isinstance(key, tuple):
+            raise TypeError('key must be a tuple: {!r}'.format(key))
+
+        new_key = []
+        for k in key:
+            if isinstance(k, integer_types):
+                k = int(k)
+            elif isinstance(k, slice):
+                k = as_integer_slice(k)
+            else:
+                raise TypeError('unexpected indexer type for {}: {!r}'
+                                .format(type(self).__name__, k))
+            new_key.append(k)
+
+        super(BasicIndexer, self).__init__(new_key)
 
 
 class OuterIndexer(ExplicitIndexer):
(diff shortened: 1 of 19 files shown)


fujiisoup:
This looks pretty clean and less error-prone.

For more cleanliness, I'm wondering if we could more clearly distinguish between raw array-wrappers (such as `NumpyIndexingAdapter`) and wrappers of array-wrapper (such as `MemoryCachedArray`).
But as a whole, I like this idea.

Regarding the more array-type support in the future (as suggested in [comment](https://github.com/pydata/xarray/issues/1617#issuecomment-340290288)),
is there something to prepare in this PR?
I guess there are some typical indexing types, such as Fortran-type and Numpy-type.
Can we have some abstract classes (maybe too early)?

shoyer:
There's been discussion about abstract classes on the NumPy mailing list.
It would be nice to use something standard if possible:
https://mail.python.org/pipermail/numpy-discussion/2017-November/thread.html#77312

xarray/core/variable.py
```diff
@@ -766,6 +769,7 @@ def chunk(self, chunks=None, name=None, lock=False):
             if utils.is_dict_like(chunks):
                 chunks = tuple(chunks.get(n, s)
                                for n, s in enumerate(self.shape))
+            data = indexing.WrapImplicitIndexing(data, indexing.OuterIndexer)
```
fujiisoup: Why is this line necessary here? (and only OuterIndexer?)
I think it would be better if we could have a simple comment above.

xarray/core/indexing.py
```diff
@@ -277,84 +278,224 @@ def _index_indexer_1d(old_indexer, applied_indexer, size):
...
+            VectorizedIndexer: 'Vectorized'
+        }[type(key)]
+        raise NotImplementedError(
+            '{} indexing for {} is not implemented. Load your data first with '
+            '.load(), .compute() or .persist(), or disable caching by setting '
+            'cache=False in open_dataset.'.format(key_type_name, type(target)))
+    return key.tuple
+
+
+class WrapImplicitIndexing(utils.NDArrayMixin):
```
fujiisoup: This name sounds like a function which behaves opposite to `unwrap_explicit_indexer`. Can we change this something like `ImplicitIndexedArrayWrapper`?
(But as the below comment, I did not yet understand how this wrapper will be used.)

xarray/core/indexing.py
```diff
@@ -277,84 +278,224 @@ def _index_indexer_1d(old_indexer, applied_indexer, size):
...
+        super(VectorizedIndexer, self).__init__(new_key)
+
+
+class ExplicitlyIndexed(object):
+    """Mixin to mark support for Indexer subclasses in indexing."""
 
 
-class VectorizedIndexer(IndexerTuple):
-    """ Tuple for vectorized indexing """
+class ExplicitNDArrayMixin(utils.NDArrayMixin, ExplicitlyIndexed):
```
fujiisoup: I like a more explicit name such as `ExplicitlyIndexedNDArrayMixin`. Is it too long?

xarray/core/indexing.py
```diff
@@ -277,84 +278,224 @@ def _index_indexer_1d(old_indexer, applied_indexer, size):
...
+    indexed with an integer are dropped from the result.
     """
+    def __init__(self, key):
+        if not isinstance(key, tuple):
+            raise TypeError('key must be a tuple: {!r}'.format(key))
+
+        new_key = []
+        for k in key:
+            if isinstance(k, integer_types):
+                k = int(k)
```
fujiisoup: This looks a good idea over checking an integer type in array wrappers :+1:

xarray/core/variable.py
shoyer: Added a comment to explain, with a link to the upstream dask issue (https://github.com/dask/dask/issues/2883).

shoyer:
I plan to merge this shortly and issue another release candidate unless any one has objections.

It does occur to me that this could break custom backends, since they will not longer be getting tuples as indexers. @rabernat any ideas for how to minimize that pain?

xarray/core/variable.py
```diff
@@ -1261,7 +1270,7 @@ def equals(self, other, equiv=duck_array_ops.array_equiv):
             return (self.dims == other.dims and
                     (self._data is other._data or
                      equiv(self.data, other.data)))
-        except (TypeError, AttributeError):
+        except (TypeError, AttributeError) as e:
```
jhamman: it doesn't look like `e` is ever used.

shoyer: Good catch, this was introduced in debugging

shoyer:
Actually, looking over @rabernat's custom backend code in xmitgcm, I think that should be OK. This would only be a problem if you pass custom duck arrays into `xarray.Variable` objects, rather than numpy or dask arrays.

shoyer:
OK, merging. Hopefully we'll catch any issues in the next release candidate!

PR #1707, opened 2017-11-11
Fix "Chunksize cannot exceed dimension size"

### LINKED ISSUE #1225: “ValueError: chunksize cannot exceed dimension size” when trying to write xarray to netcdf
Reported on StackOverflow: http://stackoverflow.com/questions/39900011/valueerror-chunksize-cannot-exceed-dimension-size-when-trying-to-write-xarray

Unfortunately, the given example is not self-contained:
```
import xarray as xr  
ds=xr.open_dataset("somefile.nc",chunks={'lat':72,'lon':144}  
myds=ds.copy()
#ds is 335 (time) on 720 on 1440 and has variable var  
def some_function(x):
  return x*2
myds['newvar']=xr.DataArray(np.apply_along_axis(some_function,0,ds['var']))  
myds.drop('var')  
myds.to_netcdf("somenewfile.nc")
```

Apparently this works if `engine='scipy'` in `to_netcdf`!

Something strange is definitely going on, I suspect a bug.

shoyer:
- [x] Closes #1225
 - [x] Tests added / passed

### DIFF
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -156,7 +156,7 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
     if lsd_okay:
         valid_encodings.add('least_significant_digit')
 
-    if not raise_on_invalid and 'chunksizes' in encoding:
+    if not raise_on_invalid and encoding.get('chunksizes') is not None:
         chunks_too_big = any(
             c > d for c, d in zip(encoding['chunksizes'], variable.shape))
         changed_shape = encoding.get('original_shape') != variable.shape
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -156,10 +156,12 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
     if lsd_okay:
         valid_encodings.add('least_significant_digit')
 
-    if (encoding.get('chunksizes') is not None and
-            (encoding.get('original_shape', variable.shape) !=
-                variable.shape) and not raise_on_invalid):
-        del encoding['chunksizes']
+    if not raise_on_invalid and 'chunksizes' in encoding:
+        chunks_too_big = any(
+            c > d for c, d in zip(encoding['chunksizes'], variable.shape))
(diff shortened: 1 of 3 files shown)


xarray/backends/netCDF4_.py
```diff
@@ -156,10 +156,12 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
...
-    if (encoding.get('chunksizes') is not None and
-            (encoding.get('original_shape', variable.shape) !=
-                variable.shape) and not raise_on_invalid):
-        del encoding['chunksizes']
+    if not raise_on_invalid and 'chunksizes' in encoding:
+        chunks_too_big = any(
+            c > d for c, d in zip(encoding['chunksizes'], variable.shape))
+        changed_shape = encoding.get('original_shape') != variable.shape
+        if chunks_too_big or changed_shape:
+            del encoding['chunksizes']
```
jhamman: this looks fine. Can you add a comment here that explains that we are dropping the encoding chunksizes so that netCDF4-python can write this dataset

xarray/backends/netCDF4_.py
```diff
@@ -156,10 +156,12 @@ def _extract_nc4_variable_encoding(variable, raise_on_invalid=False,
     if lsd_okay:
         valid_encodings.add('least_significant_digit')
 
-    if (encoding.get('chunksizes') is not None and
-            (encoding.get('original_shape', variable.shape) !=
-                variable.shape) and not raise_on_invalid):
-        del encoding['chunksizes']
+    if not raise_on_invalid and 'chunksizes' in encoding:
+        chunks_too_big = any(
+            c > d for c, d in zip(encoding['chunksizes'], variable.shape))
```
jhamman: It looks like you'll need to make sure `encoding['chunksizes']` is an iterable of length `variable.ndim`

shoyer:
I updated the logic to only drop `chunksizes` if the dimension is no longer unlimited, which is the case where an error comes up. This allows us to faithfully keep chunksizes for unlimited dimensions (e.g., if you directly read/write a complete netCDF file).

jklymak:
I had this error in 0.9.5.  0.10.0 definitely fixes it.  Thanks!

Issue #1686, opened 2017-11-03
Incorrect y-coordinates for non-georeferenced data from open_rasterio


maaleske:
For a non-georeferenced dataset, `open_rasterio()` results in y-coordinates from the raster height to two times the height, instead of the expected 0 .. height range:

```python
import numpy as np
import rasterio
from xarray import open_rasterio

tmp_file = 'no_transform.tif'
nx, ny, nz = 4, 3, 3 
data = np.arange(nx*ny*nz,
                 dtype=rasterio.float32).reshape(nz, ny, nx)
with rasterio.open(
              tmp_file, 'w', 
              driver='GTiff', height=ny, width=nx, count=nz,
              dtype=rasterio.float32) as s:
        s.write(data)

open_rasterio(tmp_file)
```

>     <xarray.DataArray (band: 3, y: 3, x: 4)>
>     array( ... )
>     Coordinates:
>       * band     (band) int64 1 2 3
>       * y        (y) float64 3.5 4.5 5.5
>       * x        (x) float64 0.5 1.5 2.5 3.5
>     Attributes:
>          res:        (1.0, -1.0)
>          is_tiled:   0
>          transform:  (0.0, 1.0, 0.0, 0.0, 0.0, 1.0)

Passing `transform=from_origin(0,3,1,1)` (from `rasterio.transforms`) to `rasterio.open()` in the above code seems to give the expected result with y running down from 2.5:

>     <xarray.DataArray (band: 3, y: 3, x: 4)>
>     array( ... )
>     Coordinates:
>       * band     (band) int64 1 2 3
>       * y        (y) float64 2.5 1.5 0.5
>       * x        (x) float64 0.5 1.5 2.5 3.5
>     Attributes:
>         res:        (1.0, 1.0)
>         is_tiled:   0
>         transform:  (0.0, 1.0, 0.0, 3.0, 0.0, -1.0)

I'm not sure whether there's something amiss in the xarray coordinate calculations or the rasterio default transform logic. Looking at the code, I feel like there's some mismatch between the rasterio `res` property logic and the xarray coordinate generation (namely, the sign of `res[1]`):

https://github.com/mapbox/rasterio/blob/fcd361c49cca9c4ad32a2ac1f5d66b967f4b6cd4/rasterio/_base.pyx#L611-L619

https://github.com/pydata/xarray/blob/f83361c76b6aa8cdba8923080bb6b98560cf3a96/xarray/backends/rasterio_.py#L131-L139

[...]

PR #1712, opened 2017-11-12
Use rasterio's transform instead of homemade coordinates 

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/1686
 - [x] Tests added / passed


fmaussion:
OK, I think this is ready for review (@shoyer or @jhamman)

xarray/backends/rasterio_.py
```diff
@@ -131,15 +133,18 @@ def open_rasterio(filename, chunks=None, cache=None, lock=None):
         raise ValueError('Unknown dims')
     coords['band'] = np.asarray(riods.indexes)
 
-    # Get geo coords
+    # Get coordinates
+    if LooseVersion(rasterio.__version__) < '1.0':
+        transform = riods.affine
+    else:
+        transform = riods.transform
```
shoyer: Can we use attribute lookup fallback instead, e.g.,
```
try:
    transform = riods.transform
except AttributeError:
    transform = riods.affine
```

That's usually slightly more robust than explicitly checking versions.

xarray/backends/rasterio_.py
```diff
@@ -131,15 +133,18 @@ def open_rasterio(filename, chunks=None, cache=None, lock=None):
...
     nx, ny = riods.width, riods.height
-    dx, dy = riods.res[0], -riods.res[1]
-    x0 = riods.bounds.right if dx < 0 else riods.bounds.left
-    y0 = riods.bounds.top if dy < 0 else riods.bounds.bottom
-    coords['y'] = np.linspace(start=y0 + dy/2, num=ny,
-                              stop=(y0 + (ny - 1) * dy) + dy/2)
-    coords['x'] = np.linspace(start=x0 + dx/2, num=nx,
-                              stop=(x0 + (nx - 1) * dx) + dx/2)
+    x, _ = (np.arange(nx), np.zeros(nx)) * transform
+    _, y = (np.zeros(ny), np.arange(ny)) * transform
```
shoyer: I'm hardly an expert here... but is there a reason why this "obvious" version doesn't work?
```
x, y = (np.arange(nx), np.arange(ny)) * transform
```

(it might be worth explaining in a comment!)

shoyer:
@maaleske would you mind taking a look at this to see if it looks right to you?

xarray/backends/rasterio_.py
maaleske: It would only work if the image was square (width == height), in which case the transform would be applied to the coordinates on the diagonal. The `np.zeros` version transforms the coordinate points of the edges (that could also probably be `np.zeros_like`?)

There's a slightly deeper problem here actually, since given a rotated /sheared dataset, we can in any case only generate correct coordinates for the edge pixels:
```python
import numpy as np
from affine import Affine

transform = Affine.rotation(30.0)
x = np.arange(5)

(x, np.zeros_like(x)) * transform
# (array([ 0.        ,  0.8660254 ,  1.73205081,  2.59807621,  3.46410162]),
#  array([ 0. ,  0.5,  1. ,  1.5,  2. ]))

(x, np.ones_like(x)) * transform
# (array([-0.5       ,  0.3660254 ,  1.23205081,  2.09807621,  2.96410162]),
#  array([ 0.8660254,  1.3660254,  1.8660254,  2.3660254,  2.8660254]))
```

Unless there's some way to implement spatial indexing using the Affine transform, I think there should be a warning thrown when the transform has non-zero cross-coordinate elements.

xarray/backends/rasterio_.py
maaleske: The `transform` attribute exists on the older version as well, but there it is a tuple specifying the transform in GDAL format. Maybe it'd be better to check whether `transform` is an instance of `Affine` instead?

xarray/backends/rasterio_.py
fmaussion: >> The most general way would probably be this:
>> xy = np.meshgrid(np.arange(nx), np.arange(ny)) * transform
>> and make it a multidimensional coordinate.

Yes, but I'd like to avoid this when the coordinates are regular (again, I haven't met a single geotiff file which isn't regular in it's reference coordinate system: when defined in UTM for example, the x, y coordinates are 1D, the lon lats are not, as in the documentation example: http://xarray.pydata.org/en/latest/auto_gallery/plot_rasterio.html#sphx-glr-auto-gallery-plot-rasterio-py).

Do you know a way to know for sure if we need 2d coordinates or not?

(another compelling reason is that many geotiff files are huge, and having non-lazy 2d coordinates isn't an option)

maaleske: You could check for `transform.is_rectilinear` to determine whether 1d coordinates can be used.

I still think it would be best to always distinguish the geocoordinates from the pixel coordinates regardless of the dataset, i.e. x, y should always be from 0.5 to width/height, with possible other coordinates added if crs/transform is found. Is there a way to build a lazy 2d coordinate array that would work as a DataArray?

fmaussion: > You could check for transform.is_rectilinear to determine whether 1d coordinates can be used.

Thanks! Will have a look

> I still think it would be best to always distinguish the geocoordinates from the pixel coordinates regardless of the dataset, i.e. x, y should always be from 0.5 to width/height

I don't understand why this would be useful. For pixel based selection `xarray` has `.isel`, and the coordinates are the correct representation of the data in space. Do you have an example?

maaleske: > I don't understand why this would be useful. For pixel based selection xarray has .isel, and the coordinates are the correct representation of the data in space. Do you have an example?

You're right that `isel` is enough for pixel selection (though then you don't have center-of-pixel pixel coordinates). I guess I'm used to having x and y always denote pixel coordinates, with any other coordinates being explicitly labeled, and I'm a bit worried about the meaning of the coordinate changing if I change the dataset transform without it being reflected in the coordinate name.

fmaussion: > if I change the dataset transform without it being reflected in the coordinate name

If geotiffs had a general way to name the coordinates I would be glad to include it here. But the way I see it the `[x, y]` pair we are producing here can be understood as Cartesian coordinates in a given reference system. 

Most of the cases I'm dealing with, the reference system is a proper map projection (in this case `[x, y]` are "eastings" and "northings", in m) or an equirectangular projection (in which case `[x, y]` are "longitudes" and "latitudes"). 

We agreed before ([this comment](https://github.com/pydata/xarray/pull/1260#issuecomment-302949290)) that xarray should not care about coordinate transforms because they're too specialized, but I think that parsing the cartesian coordinates is OK (provided that we are able to compute them in a robust way and know if the coords are 1D or 2D). 

Would you be in favor of removing the coordinates altogether? (this is possible in the xarray data model).

maaleske: I am leaning slightly towards removal; I mostly deal with non-referenced images where each person will place `(0,0)` in a different corner of the image by default, so having less coordinates to worry about is usually for the better (especially since the rasterio convention on dealing with transformless data is not really clearly defined). 

> We agreed before (this comment) that xarray should not care about coordinate transforms because they're too specialized, but I think that parsing the cartesian coordinates is OK (provided that we are able to compute them in a robust way and know if the coords are 1D or 2D).

I agree on not caring about specialized stuff in xarray, but it does make it necessary to check that all the required data is accessible for doing it later (which is what I'm struggling a bit with #1583). If you do use the transform, I think you should commit to it; If I were doing stuff involving both rectilinear and non-rectilinear stuff, not having the code work on one of the two would raise my suspicions on it's correctness for the other case (and I'm not too sure if we aren't still missing something important w.r.t. the transform).

Also: There is probably a desync happening currently with the pixel center coordinates and the transform attribute; Since the coordinates are shifted by a half, shouldn't the reverse be applied to the relevant elements in the transform? Otherwise when trying to use the included transform for generating the lon/lat/etc. coordinates the results will be off.

xarray/backends/rasterio_.py
fmaussion: > Maybe it'd be better to check whether transform is an instance of Affine instead?

I don't know... Reading the transform attribute with rio < 1.0 raises a warning, reading the affine attribute with rio >= 1.0 raises a warning, so we'll have to catch the warnings in order to test this. Would that be more elegant?

xarray/backends/rasterio_.py
fmaussion: > I am leaning slightly towards removal;

I think that for most geoscientific purposes, the coordinates are quite useful. Would a `parse_coordinates=True/False` kwarg be a possible compromise?

> Also: There is probably a desync happening currently with the pixel center coordinates and the transform attribute; Since the coordinates are shifted by a half, shouldn't the reverse be applied to the relevant elements in the transform? Otherwise when trying to use the included transform for generating the lon/lat/etc. coordinates the results will be off.

I needed some time to understand what you mean here, but now I get it. I guess you are right, but at the moment users start to play with the transform attribute, they probably don't need our coordinates either (a possible case for the `parse_coordinates` kwarg). 

Btw I replaced the code with the simpler:

```python
        x, _ = (np.arange(nx)+0.5, np.zeros(nx)+0.5) * transform
        _, y = (np.zeros(ny)+0.5, np.arange(ny)+0.5) * transform
```
(i.e. no Affine shift), which does the same thing.

See also: https://github.com/mapbox/rasterio/issues/1199

maaleske: @fmaussion Sorry, I had missed the shift you had in the transform previously (`transform.translate(0.5, 0.5)`); That was more correct, while the new one matches the old code with the desync issue. You are right that if the transform has already been used to generate the coordinates it should not really be needed anymore, and vice versa, but I would still expect it to match the coordinates I'm seeing regardless of the case, but doing the shift should be documented clearly.

>I think that for most geoscientific purposes, the coordinates are quite useful. Would a parse_coordinates=True/False kwarg be a possible compromise?

I'm really fine with any version that doesn't generate incorrect coordinates, regardless of parameters. If you do add a parameter it might be better to name it `geo_coords` to be clear on what you are actually generating.

fmaussion:
I added a `parse_coordinates=True` kwarg and added a check for 2D coordinates.

I proposed an implementation for the 2D case, but it feels a bit clunky. A second option would be to *not* parse 2D coords and raise a Warning in case of ``parse_coordinates=True`` and 2D coordinates. 
(I don't think that 2D coords happen very often).

Thoughts?

snowman2:
Would it be possible to put the coordinate generation into a separate function? It would be nice to pass in the affine, width, height, and is_2d as args and get back the spatial coordinates.

fmaussion:
@snowman2 I added an example in the code on how to do this - can be useful for downstream libraries

@maaleske @jhamman this is ready for review / opinions. This PR still closes https://github.com/pydata/xarray/issues/1686 in a backwards compatible manner and adds a special case for irregular coordinates. I don't think that 2D coordinates happen very often (never encountered one in my work), but xarray should handle them now. The PR also makes some minor other modifications.

maaleske:
@fmaussion Seems okay to me. The rasterio version check is a bit iffy but I can't think of a way to do it better.

I do agree with @snowman2 though, having the coordinate generation separately exposed would be nice (and easier to write unit tests for). Something like
```python
def xy_pixel_coordinates(nx, ny, transform=None)
    transform = Affine.identity if not transform else transform

    # xarray coordinates are pixel centered
    x, _ = (np.arange(nx)+0.5, np.zeros(nx)+0.5) * transform
    _, y = (np.zeros(ny)+0.5, np.arange(ny)+0.5) * transform
    return x,y
```
would probably suffice, though I don't know which module this would fall under. I'd imagine it might be useful for backends besides rasterio, though.

fmaussion:
Thanks @maaleske for having a look.

> having the coordinate generation separately exposed would be nice

I agree, but this could be done in a separate PR as this needs another kind of discussion: do we want these kind of domain-specific tools in the codebase (and where?), and is `Affine` sufficiently used to justify this? Here we use `Affine` because it is a `rasterio` dependency, but I don't know if it is used outside of it.

fmaussion:
I wonder if someone from @pydata/xarray wants to have a look -- if possible I'd like to get this in before the next point release. Thanks!

xarray/backends/rasterio_.py
```diff
@@ -160,18 +178,38 @@ def open_rasterio(filename, chunks=None, cache=None, lock=None):
...
-    coords['x'] = np.linspace(start=x0 + dx/2, num=nx,
-                              stop=(x0 + (nx - 1) * dx) + dx/2)
+    # Get coordinates
+    if LooseVersion(rasterio.__version__) < '1.0':
+        transform = riods.affine
+    else:
+        transform = riods.transform
+    if transform.is_rectilinear:
+        # 1d coordinates
+        parse = True if (parse_coordinates is None) else parse_coordinates
```
shoyer: nit: you could remove parentheses if you like here

xarray/tests/test_backends.py
```diff
@@ -2189,15 +2224,44 @@ def test_platecarree(self):
...
 
+    def test_notransform(self):
+        # regression test for https://github.com/pydata/xarray/issues/1686
+        import rasterio
+        import warnings
+
+        # Create a geotiff file
+        with warnings.catch_warnings():
+            # This warning is expected
+            warnings.filterwarnings('ignore', category=UserWarning)
```
shoyer: which specific warning is expected here?

`pytest.assert_warns()` might be more appropriate, if we want to test the warning as part of the API

xarray/backends/rasterio_.py
fmaussion: Done

xarray/tests/test_backends.py
fmaussion: It's rasterio which sends a warning here. I explain it in more detail know, and catch only this specific warning.

fmaussion:
Thanks for the review @shoyer ! I will wait for https://github.com/pydata/xarray/pull/1817 to be merged before going on here.

fmaussion:
All green! Will try to get to the tests in the upcoming days

Issue #1285, opened 2017-02-23
FAQ page could use some updating


shoyer:
Along the same lines as https://github.com/pydata/xarray/issues/1282, we haven't done much updating for frequently asked questions -- it's mostly still the original handful of FAQ entries I wrote in the first version of the docs.

Topics worth addressing:

- [ ] How xarray handles missing values
- [ ] File formats -- how can I read format *X* in xarray? (Maybe we should make a table with links to other packages?)

(please add suggestions for this list!)

StackOverflow may be a helpful reference here: http://stackoverflow.com/questions/tagged/python-xarray?sort=votes&pageSize=50

PR #1718, opened 2017-11-15
1D line plot with data on both axes + errorbar plotting added

### LINKED ISSUE #575: 1D line plot with data on the x axis
Consider the following Dataset, representing a function f = cos(z)

``` python
z = np.arange(10)
ds = xray.Dataset( {'f': ('z', np.cos(z))}, coords={'z': z})
```

If I call

``` python
ds.f.plot()
```

xray naturally puts "z" on the x-axis.

However, since z represents the vertical dimension, it would be more natural do put it on the y-axis, i.e. 

``` python
plt.plot(ds.f, ds.z)
```

This is conventional in atmospheric science and oceanography for buoy data or balloon data.

Is there an easy way to do this with xray's plotting functions? I scanned the code and didn't see an obvious solution, but maybe I missed it.

kuchaale:
- [ ] Closes #575 
 - [ ] Tests added / passed

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -20,7 +20,6 @@ from .utils import (_determine_cmap_params, _infer_xy_labels, get_axis,
 from .facetgrid import FacetGrid
 from xarray.core.pycompat import basestring
 
-
 def _valid_numpy_subdtype(x, numpy_types):
     """
     Is any dtype from numpy_types superior to the dtype of x?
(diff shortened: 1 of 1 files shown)


shoyer:
I think you're on the right track here, but rather than adding these to the exist `_PlotMethods` object, I think we would need a new dedicated PlotMethods class for use on Dataset instead of DataArray objects. This could probably involve duplicating some of the existing code to setup the `.plot` accessor.

kuchaale:
[#1285](https://github.com/mwaskom/seaborn/pull/1285)

kuchaale:
https://github.com/mwaskom/seaborn/pull/#1285

kuchaale:
mwaskom/seaborn/#1285

kuchaale:
mwaskom/seaborn#1285

PR #1723, opened 2017-11-16
Fix unexpected behavior of .set_index() since pandas 0.21.0

### LINKED ISSUE #1722: Change in behavior of .set_index() from pandas 0.20.3 to 0.21.0
I use xarray 0.9.6 for both examples below.

With pandas 0.20.3, `Dataset.set_index` gives me what I expect (i.e., the `grid__x` data variable becomes a coordinate `x`):

```python
In [1]: import xarray as xr

In [2]: import pandas as pd

In [3]: pd.__version__
Out[3]: '0.20.3'

In [4]: ds = xr.Dataset({'grid__x': ('x', [1, 2, 3])})

In [5]: ds.set_index(x='grid__x')
Out[5]: 
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 1 2 3
Data variables:
    *empty*
```

With pandas 0.21.0, it creates a `MultiIndex`, which is not what I expect here when setting an index with only one data variable: 

```python
In [1]: import xarray as xr

In [2]: import pandas as pd

In [3]: pd.__version__
Out[3]: '0.21.0'

In [4]: ds = xr.Dataset({'grid__x': ('x', [1, 2, 3])})

In [5]: ds.set_index(x='grid__x')
Out[5]: 
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) MultiIndex
  - grid__x  (x) int64 1 2 3
Data variables:
    *empty*
```

benbovy:
- [x] Closes #1722 
 - [x] Tests added / passed

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -136,6 +136,14 @@ def merge_indexes(
         names, labels, levels = [], [], []
         current_index_variable = variables.get(dim)
 
+        for n in var_names:
+            var = variables[n]
+            if ((current_index_variable is not None) and
+                    (var.dims != current_index_variable.dims)):
+                raise ValueError(
+                    "dimension mismatch between %r %s and %r %s"
+                    % (dim, current_index_variable.dims, n, var.dims))
+
         if current_index_variable is not None and append:
             current_index = current_index_variable.to_index()
             if isinstance(current_index, pd.MultiIndex):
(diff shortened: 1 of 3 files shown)


xarray/core/dataset.py
```diff
@@ -136,6 +136,14 @@ def merge_indexes(
         names, labels, levels = [], [], []
         current_index_variable = variables.get(dim)
 
+        for n in var_names:
+            var = variables[n]
+            if ((current_index_variable is not None) and
```
shoyer: nice: `is not` and `!=` have higher precedence than `and`, so you could remove these nested parantheses

PR #1724, opened 2017-11-17
Fix unexpected loading after ``print``

### LINKED ISSUE #1720: Possible regression with PyNIO data not being lazily loaded
@weathergod reports on the mailing list:

> I just tried [0.10.0 rc2] out in combination with the pynio engine (v1.5.0 from conda-forge), and doing a print on a dataset object causes all of the data to get loaded into memory.

fujiisoup:
- [x] Closes #1720
 - [x] Tests added / passed

Only a single missing underscore causes this issue :)
Added tests.

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -492,6 +492,14 @@ class DatasetIOTestCases(object):
             actual = on_disk.dropna(dim='x')
             assert_identical(expected, actual)
 
+    def test_ondisk_after_print(self):
+        """ Make sure print does not load file into memory """
+        in_memory = create_test_data()
+        with self.roundtrip(in_memory) as on_disk:
+            if not on_disk['var1']._in_memory:
+                print(on_disk)
+                assert not on_disk['var1']._in_memory
+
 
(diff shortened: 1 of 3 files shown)


xarray/tests/test_backends.py
```diff
@@ -492,6 +492,14 @@ def test_dropna(self):
             actual = on_disk.dropna(dim='x')
             assert_identical(expected, actual)
 
+    def test_ondisk_after_print(self):
+        """ Make sure print does not load file into memory """
+        in_memory = create_test_data()
+        with self.roundtrip(in_memory) as on_disk:
+            if not on_disk['var1']._in_memory:
```
shoyer: Can we write this without the conditional check? It's usually best to avoid control flow in tests -- it makes it more obvious what is going on.

xarray/tests/test_backends.py
```diff
@@ -492,6 +492,14 @@ def test_dropna(self):
...
+                print(on_disk)
```
shoyer: Just call `repr(on_disk)` to avoid actually printing to stdout.

PR #1733, opened 2017-11-21
Rank Methods

### LINKED ISSUE #1731: Rank function
Hi,

I think xarray is missing a rank function.
Is there any reason not to expose a wrapper to `bottleneck.nanrankdata` ?

See also https://github.com/pydata/xarray/issues/1635

[edit]
Although moving rank is mentioned in the whats-new for v0.9.2 I wasn't able to find that functionality nor a trace of it in the code :)

0x0L:
- [x] Closes #1731
- [x] Tests added / passed

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1972,7 +1972,7 @@ class DataArray(AbstractArray, BaseDataObject):
                                               interpolation=interpolation)
         return self._from_temp_dataset(ds)
 
-    def rank(self, dim, pct=False):
+    def rank(self, dim, pct=False, keep_attrs=False):
         """Ranks the data.
 
         Equal values are assigned a rank that is the average of the ranks that
@@ -1987,6 +1987,7 @@ class DataArray(AbstractArray, BaseDataObject):
         ----------
         dim : str
         pct : bool, optional
+        keep_attrs : bool, optional
 
         Returns
         -------
@@ -2002,17 +2003,8 @@ class DataArray(AbstractArray, BaseDataObject):
         array([ 1.,   2.,   3.])
         Dimensions without coordinates: x
         """
-        import bottleneck as bn
-        axis = self.get_axis_num(dim)
-        func = bn.nanrankdata if self.dtype.kind is 'f' else bn.rankdata
-        ranks = apply_ufunc(func, self,
-                            dask='parallelized',
-                            keep_attrs=True,
-                            output_dtypes=[np.float64],
-                            kwargs=dict(axis=axis)).transpose(*self.dims)
-        if not pct:
-            return ranks
-        return (ranks - 1) / (self.count(dim) - 1)
+        ds = self._to_temp_dataset().rank(dim, pct=pct, keep_attrs=keep_attrs)
+        return self._from_temp_dataset(ds)
 
 
 # priority most be higher than Variable to properly work with binary ufuncs
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1981,6 +1981,8 @@ class DataArray(AbstractArray, BaseDataObject):
 
         NaNs in the input array are returned as NaNs.
 
+        The `bottleneck` library is required.
+
         Parameters
         ----------
         dim : str
@@ -2006,7 +2008,7 @@ class DataArray(AbstractArray, BaseDataObject):
         ranks = apply_ufunc(func, self,
                             dask='parallelized',
                             keep_attrs=True,
-                            output_dtypes=[np.float_],
+                            output_dtypes=[np.float64],
                             kwargs=dict(axis=axis)).transpose(*self.dims)
         if not pct:
             return ranks
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1972,18 +1972,19 @@ class DataArray(AbstractArray, BaseDataObject):
                                               interpolation=interpolation)
         return self._from_temp_dataset(ds)
 
-    def rank(self, dim):
+    def rank(self, dim, pct=False):
         """Ranks the data.
 
         Equal values are assigned a rank that is the average of the ranks that
         would have been otherwise assigned to all of the values within that set.
-        Ranks begin at 1, not 0.
+        Ranks begin at 1, not 0. If pct is True, computes percentage ranks.
 
         NaNs in the input array are returned as NaNs.
 
         Parameters
         ----------
-        dim : str, optional
+        dim : str
+        pct : bool, optional
 
         Returns
         -------
@@ -2002,11 +2003,14 @@ class DataArray(AbstractArray, BaseDataObject):
         import bottleneck as bn
         axis = self.get_axis_num(dim)
         func = bn.nanrankdata if self.dtype.kind is 'f' else bn.rankdata
-        return apply_ufunc(func, self,
-                           dask='parallelized',
-                           keep_attrs=True,
-                           output_dtypes=[np.float_],
-                           kwargs=dict(axis=axis)).transpose(*self.dims)
+        ranks = apply_ufunc(func, self,
+                            dask='parallelized',
+                            keep_attrs=True,
+                            output_dtypes=[np.float_],
+                            kwargs=dict(axis=axis)).transpose(*self.dims)
+        if not pct:
+            return ranks
+        return (ranks - 1) / (self.count(dim) - 1)
 
 
 # priority most be higher than Variable to properly work with binary ufuncs
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1986,8 +1986,13 @@ class DataArray(AbstractArray, BaseDataObject):
         Parameters
         ----------
         dim : str
+            Dimension over which to compute rank.
         pct : bool, optional
+            If True, compute percentage ranks, otherwise compute integer ranks.
         keep_attrs : bool, optional
+            If True, the dataset's attributes (`attrs`) will be copied from
+            the original object to the new one.  If False (default), the new
+            object will be returned without attributes.
 
         Returns
         -------
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -19,6 +19,7 @@ from . import utils
 from .accessors import DatetimeAccessor
 from .alignment import align, reindex_like_indexers
 from .common import AbstractArray, BaseDataObject
+from .computation import apply_ufunc
 from .coordinates import (DataArrayCoordinates, LevelCoordinatesSource,
                           Indexes, assert_coordinate_consistent,
                           remap_label_indexers)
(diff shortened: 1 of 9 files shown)


0x0L:
This fails for datasets. I thought the call would already be dispatched to individual arrays.

shoyer:
Thanks for putting this together!

Rather than using the injection stuff in `ops.py` (which we would like to eliminate), I would prefer to simply implement this as a method on `Dataset` and `DataArray`.

Also take a look at the use of `apply_ufunc` in https://github.com/pydata/xarray/pull/1640 for a straightforward example of wrapping external functions. I think that would make things easier here.

xarray/core/duck_array_ops.py
```diff
@@ -23,6 +23,8 @@
 try:
     import bottleneck as bn
     has_bottleneck = True
+    # monkeypatch numpy with rankdata
+    np.rankdata = bn.rankdata
```
shoyer: Please don't monkeypatch -- this would affect NumPy for all xarray users.

0x0L:
@shoyer Thanks for the comment and the tips!  I'll make the changes asap, hopefully this week-end

0x0L:
Dropped support for Dataset. I don't think there's much use for it anyway.

xarray/core/dataarray.py
```diff
@@ -1964,6 +1965,42 @@ def quantile(self, q, dim=None, interpolation='linear', keep_attrs=False):
...
+        Examples
+        --------
+
+        >>> arr = xr.DataArray([5, 6, 7], dims='x')
+        >>> arr.rank('x')
+        <xarray.DataArray (x: 3)>
+        array([ 1.,   2.,   3.])
+        Dimensions without coordinates: x
+        """
+        import bottleneck as bn
```
0x0L: IMO the exception is explicit enough

xarray/core/dataarray.py
```diff
@@ -1964,6 +1965,46 @@ def quantile(self, q, dim=None, interpolation='linear', keep_attrs=False):
...
+        axis = self.get_axis_num(dim)
+        func = bn.nanrankdata if self.dtype.kind is 'f' else bn.rankdata
+        ranks = apply_ufunc(func, self,
+                            dask='parallelized',
+                            keep_attrs=True,
+                            output_dtypes=[np.float_],
```
jhamman: based on the docs, it looks like the return type is always `np.float64`.

xarray/core/dataarray.py
```diff
@@ -1964,6 +1965,46 @@ def quantile(self, q, dim=None, interpolation='linear', keep_attrs=False):
...
 
+    def rank(self, dim, pct=False):
+        """Ranks the data.
+
+        Equal values are assigned a rank that is the average of the ranks that
+        would have been otherwise assigned to all of the values within that set.
+        Ranks begin at 1, not 0. If pct is True, computes percentage ranks.
+
+        NaNs in the input array are returned as NaNs.
+
```
jhamman: Add a note here that this requires bottleneck.

xarray/tests/test_dataarray.py
```diff
@@ -2979,6 +2979,21 @@ def test_sortby(self):
...
+        expect_1 = DataArray([[2, 3, np.nan, 1]])
+        self.assertDataArrayEqual(ar.rank('dim_0'), expect_0)
+        self.assertDataArrayEqual(ar.rank('dim_1'), expect_1)
+        # int
+        x = DataArray([3,2,1])
+        self.assertDataArrayEqual(x.rank('dim_0'), x)
+        # str
+        y =  DataArray(['c', 'b', 'a'])
+        self.assertDataArrayEqual(y.rank('dim_0'), x)
+
```
jhamman: need to add test coverage for the `pct=True` option.

0x0L:
@shoyer when the ranking dimension is missing from an array in the dataset, should we do nothing, or put 1s everywhere ? None of these looks appealing or natural.

shoyer:
> when the ranking dimension is missing from an array in the dataset, should we do nothing, or put 1s everywhere ?

Good point, maybe it needs to be slightly more complicated than apply.

Two other options are raise an error (safest) or to drop all such variables from the result. The later is similar to what we do in aggregations like `sum()` if an argument is non-numeric. I think this is probably the best choice: it strikes a good balance between not returning an invalid result and not raising errors so often as to be annoying.

0x0L:
Should i move the implementation on Variable and use temp datasets like `quantile` does ?

shoyer:
> Should i move the implementation on Variable and use temp datasets like quantile does ?

I would be happy either way here. In some ways, yes, that would be the cleanest solution.

xarray/core/dataarray.py
```diff
@@ -1995,11 +1996,14 @@ def rank(self, dim):
...
-                           output_dtypes=[np.float_],
-                           kwargs=dict(axis=axis)).transpose(*self.dims)
+        ranks = apply_ufunc(func, self,
+                            dask='parallelized',
+                            keep_attrs=True,
+                            output_dtypes=[np.float_],
+                            kwargs=dict(axis=axis)).transpose(*self.dims)
+        if not pct:
+            return ranks
+        return (ranks - 1) / (self.count(dim) - 1)
```
0x0L: @shoyer with `pct=True` this is not the same implementation as pandas which returns `rank / count`. I think this one makes more sense but users might be suprised

shoyer: Good catch, thanks for calling this out. The pandas definition makes a little more sense to me, as "fraction of the data with this rank or higher". I don't know how I would describe your version in a simple sentence.

0x0L: Normalized ? It lies in [0, 1] not [1/Nsamples, 1]

My main issue is that I would expect the pct rank to average to 0.5 but with pandas we have
```python
In [2]: pd.Series([1,2,3]).rank(pct=True)
Out[2]: 
0    0.333333
1    0.666667
2    1.000000
```

shoyer: > My main issue is that I would expect the pct rank to average to 0.5

I agree that this would be a nice property, but I'm not sure it's worth the loss in interpretability.

0x0L: How about a str arg for specifying the normalization scheme ?
'pct', 'norm' and 'polar'

shoyer: What would `polar` mean?

All things being equal, we do try to lean towards doing what pandas does in xarray because consistency makes it easier to use both packages together.

0x0L: polar would be [-1, 1]
I get your point of view, i'll make the change to mimick pandas

xarray/core/dataarray.py
```diff
@@ -1964,6 +1965,40 @@ def quantile(self, q, dim=None, interpolation='linear', keep_attrs=False):
...
+
+        NaNs in the input array are returned as NaNs.
+
+        The `bottleneck` library is required.
+
+        Parameters
+        ----------
+        dim : str
+        pct : bool, optional
+        keep_attrs : bool, optional
```
jhamman: One last comment, it would be great if you could expand the parameters section of the docstrings to include a 1 line description of each. For example:

```
         Parameters
         ----------
         dim : str
             Dimension(s) over which to compute rank.
         pct : bool, optional
             If True, compute percentage ranks, otherwise compute integer ranks.
         keep_attrs : bool, optional
             If True, the dataset's attributes (`attrs`) will be copied from
             the original object to the new one.  If False (default), the new
             object will be returned without attributes.

**Comment by maxim-lian** [comment]
This looks great @0x0L thanks & congrats on your first contribution!

File: xarray/tests/test_variable.py [hunk 8]
```diff
@@ -1353,6 +1355,28 @@ def test_quantile_dask_raises(self):
...
+        # floats
+        v = Variable(['x', 'y'], [[3, 4, np.nan, 1]])
+        expect_0 = bn.nanrankdata(v.data, axis=0)
+        expect_1 = bn.nanrankdata(v.data, axis=1)
+        np.testing.assert_allclose(v.rank('x').values, expect_0)
+        np.testing.assert_allclose(v.rank('y').values, expect_1)
+        # int
+        v = Variable(['x'], [3,2,1])
+        expect = bn.rankdata(v.data, axis=0)
+        np.testing.assert_allclose(v.rank('x').values, expect)
```
maxim-lian: FYI for the future, `xarray.test` has these natively, and `assert_equal` rather than the older `self.` methods

0x0L: A shameless copy/paste from  from the test above :)

xarray/core/dataset.py
```diff
@@ -3246,6 +3246,48 @@ def quantile(self, q, dim=None, interpolation='linear',
...
+            Variables that do not depend on `dim` are dropped.
+        """
+        if dim not in self.dims:
+            raise ValueError('Dataset does not contain the dimension: %s' % dim)
+
+        variables = OrderedDict()
+        for name, var in iteritems(self.variables):
+            if name in self.data_vars and dim in var.dims:
+                variables[name] = var.rank(dim, pct=pct)
+                variables.update({
```
shoyer: This looks a little complicated to me, and also will drop non-dimension coordinates. As a simpler rule, how about simply keeping all existing coordinates?

This could look like:
```python
if name in self.data_vars:
    if dim in var.dims:
        variables[name] = var.rank(dim, pct=pct)
else:
    variables[name] = var
```

shoyer: It would also be good at add test cases verifying:
1. that coordinates stick around
2. that invalid data variables are dropped

xarray/core/variable.py
```diff
@@ -1374,6 +1373,47 @@ def quantile(self, q, dim=None, interpolation='linear'):
...
+        ranked : Variable
+
+        See Also
+        --------
+        Dataset.rank, DataArray.rank
+        """
+        import bottleneck as bn
+
+        if isinstance(self.data, dask_array_type):
+            raise TypeError("rank does not work for arrays stored as dask "
```
shoyer: please add a test that this error is raised, e.g., using `pytest.raises` or `raises_regex`

xarray/core/dataset.py
```diff
@@ -3246,6 +3246,48 @@ def quantile(self, q, dim=None, interpolation='linear',
...
+            the original object to the new one.  If False (default), the new
+            object will be returned without attributes.
+
+        Returns
+        -------
+        ranked : Dataset
+            Variables that do not depend on `dim` are dropped.
+        """
+        if dim not in self.dims:
+            raise ValueError('Dataset does not contain the dimension: %s' % dim)
```
shoyer: please add test coverage for this condition

xarray/core/dataset.py
```diff
@@ -3222,6 +3222,48 @@ def quantile(self, q, dim=None, interpolation='linear',
...
+
+        variables = OrderedDict()
+        for name, var in iteritems(self.variables):
+            if name in self.data_vars:
+                if dim in var.dims:
+                    variables[name] = var.rank(dim, pct=pct)
+            else:
+                variables[name] = var
+
+        coord_names = set(k for k in self.coords if k in variables)
```
shoyer: could be simplified as just `coord_names = set(self.coords)` now

shoyer:
Thank you!

PR #1734, opened 2017-11-21
pandas casting issues

### LINKED ISSUE #1726: Behavior of dataarray with no dimensions
Consider

```python
type(np.array([1.0]).mean())
# -> numpy.float64

type(pd.Series([1.0]).mean())
# -> float

type(xr.DataArray([1.0]).mean())
# -> xarray.core.dataarray.DataArray
```

The issue is that this dimensionless data array won't be cast into float by numpy/pandas when constructing a new ndarray/dataframe. You'll have to do it explicitly. Not a big deal but it feels weird.

I'm sure there's a real technical reason (keeping metadata ?) behind this behavior but I couldn't find any discussion about it.

0x0L:
- [x] Closes #1726 

Added a comment about constructing `pandas` objects from `xarray` objects


fmaussion:
Thanks for the PR! Doc improvements are always appreciated.

Maybe this would be better placed in the computations section? (http://xarray.pydata.org/en/latest/computation.html)

Also, I don't understand what you mean with ``Constructing `pandas` objects sometimes require explicit casting.`` Maybe we should refer to the working with pandas page then? (http://xarray.pydata.org/en/latest/pandas.html)

0x0L:
@fmaussion 

In a nutshell,

```python
x = xr.DataArray([1,2,3])
pd.Series({'mean': x.mean()})
```
does not work as expected. You'll want one of

```python
pd.Series({'mean': float(x.mean())})
pd.Series({'mean': x.mean()}, dtype=float)
pd.Series({'mean': x.mean()}).astype(float)
```

shoyer:
Can you merge master? https://github.com/pydata/xarray/pull/1747 should fix the build failure

0x0L:
@shoyer all good

doc/faq.rst
```diff
@@ -58,6 +58,15 @@ fundamentally multi-dimensional. If your data is unstructured or
 one-dimensional, stick with pandas.
 
 
+Gotchas
```
shoyer: Can you frame this as a question, so it looks more similar to the other headings?

e.g., "Why don't aggregations return Python scalars?"

doc/faq.rst
```diff
@@ -58,6 +58,15 @@ fundamentally multi-dimensional. If your data is unstructured or
...
+-------
+
+xarray tries hard to be self-consistent. Operations on xarray objects return
+other xarray objects. In particular, operations like `mean` or `sum` - even
```
shoyer: also mention indexing here?

doc/faq.rst
```diff
@@ -58,6 +58,15 @@ fundamentally multi-dimensional. If your data is unstructured or
...
+when applied to all axes - will return xarray objects. Constructing `pandas`
+objects sometimes require explicit casting.
```
shoyer: I don't quite understand what this last comment means, maybe an example would help?

doc/faq.rst
```diff
@@ -58,6 +58,15 @@ fundamentally multi-dimensional. If your data is unstructured or
 one-dimensional, stick with pandas.
 
 
+Gotchas
+-------
+
+xarray tries hard to be self-consistent. Operations on xarray objects return
```
shoyer: I would note that the self-consistency is that the DataArray methods already return DataArrays (and the same for Dataset), i.e., it preserves the type. This isn't true for NumPy or pandas, which return scalars.

0x0L: I just noticed but numpy aggregation funcs return boxed objects that look like an ndarray

shoyer: Yes, NumPy and pandas return boxed scalar types (subclasses of `np.generic`).

doc/faq.rst
0x0L: I'll add something similar to my reply to @fmaussion above

doc/faq.rst
```diff
@@ -58,13 +66,29 @@ fundamentally multi-dimensional. If your data is unstructured or
...
+does not yield the pandas DataFrame we expected. We need to do the type
+conversion ourselves:
+
+.. ipython:: python
 
-xarray tries hard to be self-consistent. Operations on xarray objects return
-other xarray objects. In particular, operations like `mean` or `sum` - even
-when applied to all axes - will return xarray objects. Constructing `pandas`
-objects sometimes require explicit casting.
+    pd.Series({'x': arr[0], 'mean': arr.mean(), 'std': arr.std()}).astype(float)
```
shoyer: Can you also note `float(arr[0])` and `arr[0].item()` as two others ways to convert scalars to builtin types?

0x0L:
@shoyer Thanks for the help for the writing, most of these are your own words ^^

doc/whats-new.rst
```diff
@@ -18,6 +18,11 @@ What's New
 v0.10.1 (unreleased)
 --------------------
 
+Documentation
+~~~~~~~~~~~~~
+
+- New entry `Why don’t aggregations return Python scalars?` in the :doc:`faq`.
```
shoyer: Consider giving yourself credit here :)

jhamman:
@0x0L - I gave you credit on the whats new page and synced with master. I'll merge once the tests pass here again.

0x0L:
@jhamman Thank you

PR #1735, opened 2017-11-22
Colorbar to FacetGrid plots

### LINKED ISSUE #1717: colorbars in facet grids
Hello,

In the 0.9.6 version, it does not appear to be possible to pass any arguments to the colorbar plotting
routine.

https://github.com/pydata/xarray/blob/8267fdb1093bba3934a172cf71128470698279cd/xarray/plot/facetgrid.py#L239

explicitly sets set_colorbar = False, which makes sense.

However, if we want horizontal colorbars, or any way of adjusting the colorbar plotted (it is huge and unwieldy), it would be good if the plotting routine checks for and passes suitable arguments to
https://github.com/pydata/xarray/blob/8267fdb1093bba3934a172cf71128470698279cd/xarray/plot/facetgrid.py#L256

I tried hacking something together, I can do something like the following now:

```python
import xarray

data = xarray.open_dataset('/data/ERSST/sst.mnmean.old.nc').sst

data = data.loc[dict(time=slice('1999-1', '1999-4'))]
data.plot.contourf(col='time', col_wrap=2, levels=12, cbar_kwargs=dict(orientation='horizontal',
                                                pad=0.1, aspect=30, shrink=0.6, ticks=[0, 10, 20 ,30]))
```

which produces:

figure_1

Is something like this available in the development version? If not, and it seems like a useful feature, I can create a PR.

Joy

JoyMonteiro:
- [x] Closes #1717
 - [ ] Tests added / passed
 - [x] Passes ``flake8 ``

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -21,6 +21,8 @@ v0.10.1 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- Facet grid plots now accept `cbar_ax` and `cbar_kwargs` arguments.
+
 Bug fixes
 ~~~~~~~~~
 
--- a/xarray/plot/facetgrid.py
+++ b/xarray/plot/facetgrid.py
@@ -236,7 +236,7 @@ class FacetGrid(object):
         # Order is important
         func_kwargs = kwargs.copy()
         func_kwargs.update(cmap_params)
-        func_kwargs.update({'add_colorbar': False, 'add_labels': False})
+        func_kwargs.update({'cbar_ax': None, 'cbar_kwargs': None, 'add_colorbar': False, 'add_labels': False})
 
         # Get x, y labels for the first subplot
         x, y = _infer_xy_labels(darray=self.data.loc[self.name_dicts.flat[0]],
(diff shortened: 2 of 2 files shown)


JoyMonteiro:
Happy to make any changes required!!

fmaussion:
Thanks for the PR! It seems that your changes broke [some tests](https://travis-ci.org/pydata/xarray/jobs/305625730#L4264). Could you have a look? 

Also, it would be good to add a test for this feature. [Here](https://github.com/pydata/xarray/blob/9d09c1659741dafb1fadeed49c81f9e90a548b07/xarray/tests/test_plot.py#L1011) are the tests for facetgrids, and [here](https://github.com/pydata/xarray/blob/9d09c1659741dafb1fadeed49c81f9e90a548b07/xarray/tests/test_plot.py#L724) is an example for the 2d plots cbar kwargs tests as a template.

Finally, if you have time it would be great to add an example in the docs ;). The easiest would be to add a `cbar_kwargs` kw to the call to facetgrid in [this example](http://xarray.pydata.org/en/latest/plotting.html#other-features). Thanks a lot!

PR #1741, opened 2017-11-23
Auto flake

mrocklin:
- [ ] Closes #xxxx
 - [x] Tests added / passed
 - [ ] Passes ``flake8 xarray``

I had a free half hour so I decided to run autoflake and autopep8 tools on the codebase.  `flake8 xarray` passes.  I copied over the exclusions that we use within `dask/distributed` and extended the line length to 120.   You may wish to review these decisions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ install:
 script:
   - python -OO -c "import xarray"
   - py.test xarray --cov=xarray --cov-config ci/.coveragerc --cov-report term-missing --verbose $EXTRA_FLAGS
-  - git diff upstream/master **/*py | flake8 --diff --exit-zero || true
+  - flake8 -j auto xarray
 
 after_success:
   - coveralls
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,4 +5,25 @@ universal = 1
 python_files=test_*.py
 
 [flake8]
-max-line-length=79
+# References:
+# https://flake8.readthedocs.io/en/latest/user/configuration.html
+# https://flake8.readthedocs.io/en/latest/user/error-codes.html
+
+# Note: there cannot be spaces after comma's here
+exclude = __init__.py
+ignore =
+    # Extra space in brackets
+    E20,
+    # Multiple spaces around ","
+    E231,E241,
+    # Comments
+    E26,
+    # Import formatting
+    E4,
+    # Comparing types instead of isinstance
+    E721,
+    # Assigning lambda expression
+    E731,
+    # Ambiguous variable names
+    E741
+max-line-length = 120
--- a/xarray/tests/test_dataarray.py
+++ b/xarray/tests/test_dataarray.py
@@ -203,11 +203,11 @@ class TestDataArray(TestCase):
     def test_encoding(self):
         expected = {'foo': 'bar'}
         self.dv.encoding['foo'] = 'bar'
-        assert expected, self.d == encoding
+        assert expected == self.dv.encoding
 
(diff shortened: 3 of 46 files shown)


xarray/tests/test_dataarray.py
```diff
@@ -203,11 +203,11 @@ def test_sizes(self):
     def test_encoding(self):
         expected = {'foo': 'bar'}
         self.dv.encoding['foo'] = 'bar'
-        assert expected, self.d == encoding
+        assert expected == self.dv.encoding
```
mrocklin: This change was odd.  I'm not sure what was expected here.

mrocklin:
I'm not sure how to fix these lines in `xarray/tests/test_computation.py::test_keep_attrs`

    a = xr.Dataset({'x': ('x', [1, 2]), 'x': [0, 1]})
    b = xr.Dataset({'x': ('x', [1, 1]), 'x': [0, 1]})

We're reusing the same key, `'x'`, twice

xarray/tests/test_dataarray.py
shoyer: Yeah, I think this change is correct. I have no idea what was going on previously -- `self.d` is not even a valid attribute.

setup.cfg
```diff
@@ -5,4 +5,25 @@ universal = 1
...
+    E26,
+    # Import formatting
+    E4,
+    # Comparing types instead of isinstance
+    E721,
+    # Assigning lambda expression
+    E731,
+    # Ambiguous variable names
+    E741
+max-line-length = 120
```
shoyer: Let's leave this as the default (79 per PEP8)

shoyer:
@mrocklin oof, you're catching all our questionable code!

Try `xr.Dataset({'x': [1, 2]})` for both of these.

xarray/tests/test_dataarray.py
dopplershift: Previously, it asserted that `expected` was true-ish; if it wasn’t, it would print the value of `self.d == encoding` as the message, but that expression is only evaluated if the `assert` fails.

dopplershift: I mean, that’s how the code was able to pass without error. Not sure what happened to cause that to be written. 😁

.travis.yml
```diff
@@ -91,7 +91,7 @@ install:
 script:
   - python -OO -c "import xarray"
   - py.test xarray --cov=xarray --cov-config ci/.coveragerc --cov-report term-missing --verbose $EXTRA_FLAGS
-  - git diff upstream/master **/*py | flake8 --diff --exit-zero || true
+  - flake8 -j auto xarray
```
jhamman: question: do we want to have travis fail when flake8 finds something awry?

.travis.yml
```diff
@@ -91,7 +91,7 @@ install:
 script:
   - python -OO -c "import xarray"
   - py.test xarray --cov=xarray --cov-config ci/.coveragerc --cov-report term-missing --verbose $EXTRA_FLAGS
-  - git diff upstream/master **/*py | flake8 --diff --exit-zero || true
```
Zac-HD: Instead of replacing this, I'd actually enforce that the diff must be flake8-clean:

    git diff master **/*py | flake8 --diff

And then add a second check that covers all files but ignores some errors in existing code:

    flake8 --ignore E20,E231,E241,E26,E4,E721,E731,E741 xarray

Which means that the config section can (needs to be) be left as-is.

PR #1750, opened 2017-11-30
xarray to and from Iris

### LINKED ISSUE #621: An iris cube converter?

duncanwp:
- [x] Closes #621 and #37 - finishes #814
 - [x] Tests added / passed

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -415,7 +415,9 @@ Dataset methods
    open_dataset
    open_mfdataset
    open_rasterio
+   open_zarr
    Dataset.to_netcdf
+   Dataset.to_zarr
    save_mfdataset
    Dataset.to_array
    Dataset.to_dataframe
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -446,6 +446,7 @@ DataArray methods
    DataArray.to_index
    DataArray.to_masked_array
    DataArray.to_cdms2
+   DataArray.to_iris
    DataArray.to_dict
    DataArray.from_series
    DataArray.from_cdms2
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -487,6 +487,7 @@ Exceptions
    :toctree: generated/
 
    MergeError
+   SerializationWarning
 
 Advanced API
 ============
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -447,6 +447,7 @@ DataArray methods
    DataArray.to_masked_array
    DataArray.to_cdms2
    DataArray.to_iris
+   DataArray.from_iris
    DataArray.to_dict
    DataArray.from_series
    DataArray.from_cdms2
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
 By default, xarray uses the 'proleptic_gregorian' calendar and units of the smallest time
 difference between values, with a reference time of the first time value.
 
+.. _io.iris:
+
+Iris
+----
+
+The Iris_ tool allows easy reading of common meteorological and climate model formats
+(including GRIB and UK MetOffice PP files) into ``Cube``s which are in many ways very
+similar to to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
+installaed xarray can convert a ``Cube`` into a ``DataArray`` using
+:py:meth:`~xarray.Dataset.from_iris`:
+
+.. ipython:: python
+
+    da_cube = xr.Dataset.from_iris(cube)
+    da_cube
+
+
+Conversly, We can create a new cube object from a ``DataArray`` using
+:py:meth:`~xarray.Dataset.from_dict`:
+
+.. ipython:: python
+
+    cube = da.to_iris()
+    cube
+
+.. _Iris: http://scitools.org.uk/iris
+
+
 OPeNDAP
 -------
 
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -344,9 +344,9 @@ Iris
 ----
 
 The Iris_ tool allows easy reading of common meteorological and climate model formats
-(including GRIB and UK MetOffice PP files) into ``Cube``s which are in many ways very
+(including GRIB and UK MetOffice PP files) into ``Cube`` objects which are in many ways very
 similar to to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
-installaed xarray can convert a ``Cube`` into a ``DataArray`` using
+installed xarray can convert a ``Cube`` into a ``DataArray`` using
 :py:meth:`~xarray.Dataset.from_iris`:
 
 .. ipython:: python
@@ -355,8 +355,8 @@ installaed xarray can convert a ``Cube`` into a ``DataArray`` using
     da_cube
 
 
-Conversly, We can create a new cube object from a ``DataArray`` using
-:py:meth:`~xarray.Dataset.from_dict`:
+Conversly, we can create a new cube object from a ``DataArray`` using
+:py:meth:`~xarray.Dataset.to_iris`:
 
 .. ipython:: python
 
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -346,24 +346,26 @@ Iris
 The Iris_ tool allows easy reading of common meteorological and climate model formats
 (including GRIB and UK MetOffice PP files) into ``Cube`` objects which are in many ways very
 similar to ``DataArray`` objects, while enforcing a CF-compliant data model. If iris is
-installed xarray can convert a ``Cube`` into a ``DataArray`` using
-:py:meth:`~xarray.Dataset.from_iris`:
+installed xarray can convert a ``DataArray`` into a ``Cube`` using
+:py:meth:`~xarray.DataArray.to_iris`:
 
 .. ipython:: python
-    :verbatim:
 
-    da_cube = xr.Dataset.from_iris(cube)
-    da_cube
-
-
-Conversly, we can create a new cube object from a ``DataArray`` using
-:py:meth:`~xarray.Dataset.to_iris`:
-
-.. ipython:: python
-    :verbatim:
+    da = xr.DataArray(np.random.rand(4, 5), dims=['x', 'y'],
+                      coords=dict(x=[10, 20, 30, 40],
+                                  y=pd.date_range('2000-01-01', periods=5)))
 
     cube = da.to_iris()
     cube
+
+Conversely, we can create a new ``DataArray`` object from a ``Cube`` using
+:py:meth:`~xarray.DataArray.from_iris`:
+
+.. ipython:: python
+
+    da_cube = xr.Dataarray.from_iris(cube)
+    da_cube
+
 
 .. _Iris: http://scitools.org.uk/iris
 
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -363,7 +363,7 @@ Conversely, we can create a new ``DataArray`` object from a ``Cube`` using
 
 .. ipython:: python
 
-    da_cube = xr.Dataarray.from_iris(cube)
+    da_cube = xr.DataArray.from_iris(cube)
     da_cube
 
 
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -345,11 +345,12 @@ Iris
 
 The Iris_ tool allows easy reading of common meteorological and climate model formats
 (including GRIB and UK MetOffice PP files) into ``Cube`` objects which are in many ways very
-similar to to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
+similar to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
 installed xarray can convert a ``Cube`` into a ``DataArray`` using
 :py:meth:`~xarray.Dataset.from_iris`:
 
 .. ipython:: python
+    :verbatim:
 
     da_cube = xr.Dataset.from_iris(cube)
     da_cube
@@ -359,6 +360,7 @@ Conversly, we can create a new cube object from a ``DataArray`` using
 :py:meth:`~xarray.Dataset.to_iris`:
 
 .. ipython:: python
+    :verbatim:
 
     cube = da.to_iris()
     cube
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -295,7 +295,7 @@ string encoding for character arrays in netCDF files was
 Technically, you can use
 `any string encoding recognized by Python <https://docs.python.org/3/library/codecs.html#standard-encodings>`_ if you feel the need to deviate from UTF-8,
 by setting the ``_Encoding`` field in ``encoding``. But
-`we don't recommend it<http://utf8everywhere.org/>`_.
+`we don't recommend it <http://utf8everywhere.org/>`_.
 
 .. warning::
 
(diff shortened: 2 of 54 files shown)


xarray/convert.py
```diff
@@ -53,3 +64,129 @@ def set_cdms2_attrs(var, attrs):
...
+    """ Convert a DataArray into a Iris Cube
+    """
+    # Iris not a hard dependency
+    import iris
+    try:
+        from iris.fileformats.netcdf import parse_cell_methods
+    except ImportError:
+        # prior to v1.10
+        from iris.fileformats._pyke_rules.compiled_krb.fc_rules_cf_fc \
+            import _parse_cell_methods as parse_cell_methods
```
jhamman: would it be wroth using `importlib` to clean this up?

doc/whats-new.rst
```diff
@@ -23,7 +23,11 @@ Enhancements
 
 - :py:func:`~plot.contourf()` learned to contour 2D variables that have both a 1D co-ordinate (e.g. time) and a 2D co-ordinate (e.g. depth as a function of time).
   By `Deepak Cherian <https://github.com/dcherian>`_.
+- Added :py:meth:`DataArray.to_iris <xray.DataArray.to_iris>` for
+  converting a data array into an Iris_ Cube with the same data and coordinates (:issue:`621` and :issue:`37`).
+  By `Neil Parley <https://github.com/nparley>`_.
```
jhamman: Also note the `from_iris` method.

doc/api.rst
```diff
@@ -446,6 +446,7 @@ DataArray methods
    DataArray.to_index
    DataArray.to_masked_array
    DataArray.to_cdms2
+   DataArray.to_iris
```
jhamman: also note the `from_iris` method

shoyer:
@pelson any thoughts here?

I think this may be a reasonable place to start, though it is certainly not leveraging Iris's full metadata decoding capabilities.

It would also be nice to pass dask arrays back and forth, but that will probably need to wait until after Iris 2.0 and https://github.com/pydata/xarray/issues/1372 is solved.

doc/api.rst
duncanwp: Done

doc/whats-new.rst
duncanwp: Done

xarray/convert.py
duncanwp: I wonder if it's worth just removing this catch and requiring iris > 1.10. It's at 1.13 now and nearly 2.0... @pelson?

duncanwp:
> I think this may be a reasonable place to start, though it is certainly not leveraging Iris's full metadata decoding capabilities.
It seams like a decent starting point, were there specific decodings you had in mind?

> It would also be nice to pass dask arrays back and forth, but that will probably need to wait until after Iris 2.0 and #1372 is solved.
Agreed.

Do you have a feel for if/where else this should be documented? I'm not sure if really fits in I/O...

fmaussion:
> Do you have a feel for if/where else this should be documented? I'm not sure if really fits in I/O...

I think it's OK in I/O (for now). This is also the place where we document serialization in pickle, and the `from_dict` methods.

shoyer:
> I think it's OK in I/O (for now). This is also the place where we document serialization in pickle, and the from_dict methods.

Agreed. In the long term, the IO docs have gotten pretty long, it could make sense to split them up into several subpages.

doc/io.rst
```diff
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
 By default, xarray uses the 'proleptic_gregorian' calendar and units of the smallest time
 difference between values, with a reference time of the first time value.
 
+.. _io.iris:
+
+Iris
+----
+
+The Iris_ tool allows easy reading of common meteorological and climate model formats
+(including GRIB and UK MetOffice PP files) into ``Cube``s which are in many ways very
```
shoyer: Write "`Cube` objects".  Sphinx won't render ``Cube``s properly.

doc/io.rst
```diff
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
+similar to to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
+installaed xarray can convert a ``Cube`` into a ``DataArray`` using
```
shoyer: installaed -> installed

doc/io.rst
```diff
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
+:py:meth:`~xarray.Dataset.from_iris`:
+
+.. ipython:: python
+
+    da_cube = xr.Dataset.from_iris(cube)
+    da_cube
+
+
+Conversly, We can create a new cube object from a ``DataArray`` using
```
shoyer: We -> we

doc/io.rst
```diff
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
+:py:meth:`~xarray.Dataset.from_dict`:
```
shoyer: to_iris?

xarray/convert.py
```diff
@@ -53,3 +64,124 @@ def set_cdms2_attrs(var, attrs):
...
+            aux_coords.append((iris_coord, axis))
+
+    args = _get_iris_args(dataarray.attrs)
+    args['var_name'] = dataarray.name
+    args['dim_coords_and_dims'] = dim_coords
+    args['aux_coords_and_dims'] = aux_coords
+    if 'cell_methods' in dataarray.attrs:
+        args['cell_methods'] = parse_cell_methods(dataarray.attrs['cell_methods'])
+
+    cube = iris.cube.Cube(dataarray.to_masked_array(), **args)
```
shoyer: Rather than loading every dataset into memory, can we raise `NotImplementedError` for now if a DataArray's data is a dask array? `isinstance(dataarray.dask, dask_array_type)` should do it, where `dask_array_type` is imported from xarray.core.pycompat.

xarray/convert.py
```diff
@@ -53,3 +64,124 @@ def set_cdms2_attrs(var, attrs):
...
+        except iris.exceptions.CoordinateNotFoundError:
+            dims.append("dim_{}".format(i))
+
+    coords = OrderedDict()
+
+    for coord in cube.coords():
+        coord_attrs = _iris_obj_to_attrs(coord)
+        coord_dims = [dims[i] for i in cube.coord_dims(coord)]
+        if not coord.var_name:
+            raise ValueError('Coordinate has no var_name')
```
shoyer: Can we print the offending coordinate in the error message? That makes things easier to debug.

xarray/convert.py
```diff
@@ -53,3 +64,124 @@ def set_cdms2_attrs(var, attrs):
...
+    """ Convert a Iris cube into an DataArray
+    """
+    import iris.exceptions
+    name = cube.var_name
+    dims = []
+    for i in range(cube.ndim):
+        try:
+            dim_coord = cube.coord(dim_coords=True, dimensions=(i,))
+            dims.append(dim_coord.var_name)
+        except iris.exceptions.CoordinateNotFoundError:
```
shoyer: This line doesn't have an test coverage:
https://coveralls.io/builds/14513466/source?filename=xarray%2Fconvert.py

xarray/convert.py
shoyer: This line also need coverage in a test.

xarray/convert.py
duncanwp: Raising ``NotImplementedError`` seems a bit drastic, especially since Iris 2.0 is just around the corner. The latest commit lets Iris decide what to do with the array - it keeps it as a dask array if your version of Iris supports it but will read it into memory otherwise. 

Otherwise we have to work out which version of Iris we're converting to/from which would be a pain... If you're happy with this approach I can add another set of tests for the various dask/numpy Cube/DataArray permutations.

duncanwp:
Thanks for the feedback @shoyer and sorry for all the typos!

xarray/convert.py
```diff
@@ -8,8 +8,10 @@
 
 from .core.dataarray import DataArray
 from .core.pycompat import OrderedDict, range
+from .core.dtypes import get_fill_value
 from .conventions import (
     maybe_encode_timedelta, maybe_encode_datetime, decode_cf)
+from dask.array import ma
```
shoyer: This needs to be guarded inside a function, as dask is still not a hard requirement for xarray.

xarray/convert.py
```diff
@@ -117,7 +119,7 @@ def to_iris(dataarray):
     if 'cell_methods' in dataarray.attrs:
         args['cell_methods'] = parse_cell_methods(dataarray.attrs['cell_methods'])
 
-    cube = iris.cube.Cube(dataarray.to_masked_array(), **args)
+    cube = iris.cube.Cube(ma.masked_invalid(dataarray), **args)
```
shoyer: This converts everything into a dask array, even numpy arrays. Instead, we probably only want to convert xarray objects that already store their data in dask arrays into masked dask arrays. You can dispatch based on the type of `dataarray.data`.

xarray/convert.py
```diff
@@ -181,7 +183,9 @@ def from_iris(cube):
     cell_methods = _iris_cell_methods_to_str(cube.cell_methods)
     if cell_methods:
         array_attrs['cell_methods'] = cell_methods
-    dataarray = DataArray(cube.data, coords=coords, name=name,
+
+    cube_data = ma.filled(cube.core_data(), get_fill_value(cube.dtype)) if hasattr(cube, 'core_data') else cube.data
```
shoyer: Likewise, this should not coerce all data into dask arrays.

xarray/convert.py
shoyer: OK, if it's easy enough to add support for dask, then that is great!

xarray/convert.py
duncanwp: OK, I hadn't appreciated that dask wasn't a hard requirement. I could leave them as whatever type they're currently stored in - how would xarray cope with a dask masked array?

shoyer:
I don't know exactly what would go wrong, but I'm pretty sure masked Dask
arrays would break xarray in some subtle ways. It would be better to
convert them to using unmasked Dask arrays using NaN.

xarray/convert.py
jhamman: would it be worth supporting dask masked arrays in `DataArray.to_masked_array()`?

shoyer: Yes that's probably a good idea.

jhamman: xref: #1769

duncanwp:
OK, here's another go. It's not pretty, but it seems to work. Once the various issues referenced above are solved it should clean it up, but I'd rather not wait for those since this has been in the works a while now.

doc/io.rst
```diff
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
 difference between values, with a reference time of the first time value.
 
+.. _io.iris:
+
+Iris
+----
+
+The Iris_ tool allows easy reading of common meteorological and climate model formats
+(including GRIB and UK MetOffice PP files) into ``Cube`` objects which are in many ways very
+similar to to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
```
fmaussion: double "to"

doc/io.rst
```diff
@@ -338,6 +338,34 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
+installed xarray can convert a ``Cube`` into a ``DataArray`` using
+:py:meth:`~xarray.Dataset.from_iris`:
+
+.. ipython:: python
+
+    da_cube = xr.Dataset.from_iris(cube)
+    da_cube
+
```
fmaussion: Are these examples working? If you'd like this code to run at build time you'll have to add iris to the doc build environment and define the ``cube`` variable first.

Alternatively, you can use the ``:verbatim:`` environment (see OPeNDAP example below) which is static and therefore easier (and more error prone on the long term, though)

duncanwp:
Great - thanks!

doc/io.rst
```diff
@@ -338,6 +338,36 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
 difference between values, with a reference time of the first time value.
 
+.. _io.iris:
+
+Iris
+----
+
+The Iris_ tool allows easy reading of common meteorological and climate model formats
+(including GRIB and UK MetOffice PP files) into ``Cube`` objects which are in many ways very
+similar to ``DataArray``s, while enforcing a CF-compliant data model. If iris is
```
shoyer: Needs to be "`DataArray` object"

(RST can't handle anything adjacent to backticks which is why the color looks weird here in GitHub's syntax highlighting. Yes, it's not very intuitive.)

xarray/convert.py
```diff
@@ -53,3 +65,153 @@ def set_cdms2_attrs(var, attrs):
...
+# TODO: Add converting bounds from xarray to Iris and back
+def to_iris(dataarray):
+    """ Convert a DataArray into a Iris Cube
+    """
+    # Iris not a hard dependency
+    import iris
+    from iris.fileformats.netcdf import parse_cell_methods
+    try:
+        from dask.array import ma as dask_ma
+        from dask.array import Array
```
shoyer: Instead of testing for Array, use:
```
from xarray.core.pycompat import dask_array_type

...
if isinstance(data, dask_array_type):
    # this is always defined, even if dask is not installed
```

xarray/convert.py
```diff
@@ -53,3 +65,153 @@ def set_cdms2_attrs(var, attrs):
...
+    args = _get_iris_args(dataarray.attrs)
+    args['var_name'] = dataarray.name
+    args['dim_coords_and_dims'] = dim_coords
+    args['aux_coords_and_dims'] = aux_coords
+    if 'cell_methods' in dataarray.attrs:
+        args['cell_methods'] = parse_cell_methods(dataarray.attrs['cell_methods'])
+
+    # Create the right type of masked array (should be easier after #1769)
+    if isinstance(dataarray.data, Array):
+        masked_data = dask_ma.masked_invalid(dataarray)
```
shoyer: just import `from dask.array import ma as dask_ma` right about this line to avoid the need for guarding the import statement in try/except.

duncanwp:
dask/dask#2977 would certianly make it cleaner, but [95b0197](https://github.com/pydata/xarray/pull/1750/commits/95b0197ebb0b5258db8deee48b2657fab7d86ca4 ) should work in the meantime, unless there's a specific case I've missed?

xarray/convert.py
```diff
@@ -53,3 +65,153 @@ def set_cdms2_attrs(var, attrs):
...
+    array_attrs = _iris_obj_to_attrs(cube)
+    cell_methods = _iris_cell_methods_to_str(cube.cell_methods)
+    if cell_methods:
+        array_attrs['cell_methods'] = cell_methods
+
+    # Deal with iris 1.* and 2.*
+    cube_data = cube.core_data() if hasattr(cube, 'core_data') else cube.data
+
+    # Deal with dask and numpy masked arrays
+    if dask_ma and isinstance(cube_data, Array):
```
shoyer: Does Iris *always* use dask masked arrays? Or does it ever use dask numpy arrays? @pelson?

(I would guess not, since we can't tell these apart.)

shoyer: Also, again it would be better to use `dask_array_type` andd import `dask.array.ma` here rather than above. That would ensure that an error is raised if somehow a cube contains a dask array but has the wrong version of dask installed, rather than letting the error pass silently.

duncanwp: I don't think it matters, if the given array isn't masked then ma.filled is just a no-op

xarray/convert.py
duncanwp: Much nicer - thanks!

doc/io.rst
duncanwp: Good catch, thanks

xarray/convert.py
shoyer: OK, good to hear!

doc/io.rst
```diff
@@ -338,6 +338,36 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
...
+----
+
+The Iris_ tool allows easy reading of common meteorological and climate model formats
+(including GRIB and UK MetOffice PP files) into ``Cube`` objects which are in many ways very
+similar to ``DataArray`` objects, while enforcing a CF-compliant data model. If iris is
+installed xarray can convert a ``Cube`` into a ``DataArray`` using
+:py:meth:`~xarray.Dataset.from_iris`:
+
+.. ipython:: python
+    :verbatim:
```
shoyer: `:verbatim:` is basically designed for copying/pasting output from the IPython command line.

I would definitely prefer adding Iris to our docs builds if that's feasible. See:
https://github.com/pydata/xarray/blob/master/doc/environment.yml

duncanwp: OK, I'll give it a shot. How does the environment work, are all the examples run in the same interpreter session? And are there pre-built datarrays I can just use as examples?

duncanwp:
How's this? (I turned the example around so that I create the cube as a DataArray then convert it...)

shoyer:
@duncanwp that example looks good to me. I assume it runs locally with this branch?

duncanwp:
Yep, it's not a very exciting cube, but it runs fine.

duncanwp:
Ah yes, sorry about that - fixed now.

duncanwp:
This looks good to go now - thanks for all your help!

pelson:
Apologies for the delay. This is a great first pass. Definitely more metadata sharing possible, but I'm glad this has been merged - let's grow the capability as required. If any issues crop up, please feel free to ping me & @bjlittle. 👍

PR #1752, opened 2017-12-01
Refactor xarray.conventions into VariableCoder

### LINKED ISSUE #1372: decode_cf() loads chunked arrays
Currently using xarray version 0.9.2 and dask version 0.14.0.

Suppose you load a NetCDF file with the **chunks** parameter:

``` python
ds = xr.open_dataset("my_data.nc", decode_cf=False, chunks={'lon': 10, 'lat': 10})
```

The data is loaded as dask arrays, as expected. But if we then manually call `xarray.decode_cf()`, it'll eagerly load the data. Is this the expected behavior, or should `decode_cf()` preserve the laziness of the data?

shoyer:
Building off of discussion in #1087, I would like to propose refactoring xarray.conventions to use an interface based on VariableCoder objects with encode() and decode() methods.

The idea is make it easier to write new backends, by making decoding variables according to CF conventions as simple as calling decode() on each coder in a list of coders, with encoding defined by calling the same list of encoders in the opposite order.

As a proof of concept, here I implement only a single Coder. In addition to making use of xarray's existing lazy indexing behavior, I have written it so that dask arrays are decoded using dask (which would solve #1372)

Eventually, we should port all the coders in xarray.conventions to this new format. This is probably best saved for future PRs -- help would be appreciated!

 - [x] Tests added / passed

### DIFF
--- a/xarray/coding/variables.py
+++ b/xarray/coding/variables.py
@@ -85,12 +85,12 @@ def lazy_elemwise_func(array, func, dtype):
 
     Returns
     -------
-    Either a dask.array.Array or _LazyFuncArray.
+    Either a dask.array.Array or _ElementwiseFunctionArray.
     """
     if isinstance(array, dask_array_type):
         return array.map_blocks(func, dtype=dtype)
     else:
-        return _LazyFuncArray(array, func, dtype)
+        return _ElementwiseFunctionArray(array, func, dtype)
 
 
 def unpack_for_encoding(var):
--- a/xarray/coding/variables.py
+++ b/xarray/coding/variables.py
@@ -1,0 +1,194 @@
+"""Coders for individual Variable objects."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from functools import partial
+import warnings
+
+import numpy as np
+import pandas as pd
+
+from ..core import dtypes
+from ..core import duck_array_ops
+from ..core import indexing
+from ..core import utils
+from ..core.pycompat import dask_array_type
+from ..core.variable import Variable
+
+
+class SerializationWarning(RuntimeWarning):
+    """Warnings about encoding/decoding issues in serialization."""
+
+
+class VariableCoder(object):
+    """Base class for encoding and decoding transformations on variables.
+
+    We use coders for transforming variables between xarray's data model and
+    a format suitable for serialization. For example, coders apply CF
+    conventions for how data should be represented in netCDF files.
+
+    Subclasses should implement encode() and decode(), which should satisfy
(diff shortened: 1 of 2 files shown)


xarray/coding/variables.py
```diff
@@ -0,0 +1,194 @@
...
+                                 "attrs and decoding explicitly using "
+                                 "xarray.decode_cf()."
+                                 .format(name, attrs['_FillValue'],
+                                         attrs['missing_value']))
+            attrs['_FillValue'] = attrs.pop('missing_value')
+
+        if '_FillValue' in attrs:
+            raw_fill_value = pop_to(attrs, encoding, '_FillValue', name=name)
+            encoded_fill_values = [
+                fv for fv in np.ravel(raw_fill_value) if not pd.isnull(fv)]
```
jhamman: Is it even possible to specify a sequence of fill values in a netcdf file?

shoyer: Apparently? This came from a pull request, though I refactored the logic here slightly.

alexamici:
@shoyer I'm writing a small helper that looks very much like a decoding filter, but I'm not sure it fits the VariableCoder model.

The intent is to use CF attributes to identify coordinate variables, for example {'units': 'degrees_north'} identify a 'latitude' coordinate, and apply a set of transformations to harmonise datasets from different providers: I rename the coordinates to a known name, for example 'lat' for latitude, and convert to a known unit, for example to 'Pa' for pressure.

To my understanding some of the transformations need to be at a DataArray / Dataset level, not Variable one, because I need to manipulate the names of the coordinate variables and keep the name of the dimensions in sync, for example I'm currently using DataArray.rename a lot, but I also have an example of using DataArray.swap_dims. A simplified example is:

```
def harmonise_latitude(dataset):
    # (xr.Dataset) -> xr.Dataset
    for coord in dataset.coords.values():
        if 'units' in coord.attrs and coord.attrs['units'] == 'degrees_north':
            # update the other CF attributes that reference coord.name, for example 'coordinates'.
            return dataset.rename({coord.name: 'lat'})
    return dataset
```

Would you be open to add support for Dataset -> Dataset decoding filters similar to the Variable -> Variable?

shoyer:
The intent with decoders is handle the invertible transformations between a raw file on disk and xarray's data model. So we will definitely need some transformations that act on a collection of variables (e.g., to decode coordinates vs data variables), but not in the form of a Dataset yet. These decoders/encoders will probably take/return three arguments: `variables` (an OrderedDict of `xarray.Variable` objects with both coordinates and data), `dims` (dict of dimension names to sizes) and (global) `attrs`.

PR #1753, opened 2017-12-01
Clarify our pull request template.

shoyer:
I have two goals here:
1. I'd like to clarify for new contributors what we actually expect here.
2. I'd like to ensure smaller, self-contained changes, e.g., for major new
   features or refactoring, and make it clear that it's OK to merge parts of
   larger projects before they are entirely documented.

Large changes (more a few hundred lines of code) are hard to review and
collaborate on. As long as the code is tested (even if some are marked
xfail), we should try to get these changes in faster so we can iterate on them.
It's OK if documentation comes later, as long as we document everything before
it's released. It's also OK to copy code when refactoring and delete old code
later: that is often the easiest way to make large changes.

@rabernat's recent zarr PR (#1528) is probably a good example of something that
would have been good to have merged in smaller pieces :).

CC @pydata/xarray

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,5 @@
  - [ ] Closes #xxxx (remove if there is no corresponding issue, which should only be the case for minor changes)
- - [ ] Tests added / passed (remove if this change only consists of documentation or performance improvements)
+ - [ ] Tests added (for all bug fixes or enhancements)
+ - [ ] Tests passed (for all non-documentation changes)
  - [ ] Passes ``git diff upstream/master **/*py | flake8 --diff`` (remove if you did not edit any Python files)
  - [ ] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,4 @@
- - [ ] Closes #xxxx
- - [ ] Tests added / passed
- - [ ] Passes ``git diff upstream/master **/*py | flake8 --diff``
- - [ ] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API
+ - [ ] Closes #xxxx (remove if there is no corresponding issue, which should only be the case for minor changes)
+ - [ ] Tests added / passed (remove if this change only consists of documentation or performance improvements)
```
jhamman: probably drop the performance improvements part. In those cases, we at least will want the tests to pass.

maxim-lian:
> As long as the code is tested (even if some are marked xfail), we should try to get these changes in faster so we can iterate on them.

I very much agree with this, both as something that I feel and for newer contributors. Is there a way of making this more explicit? It's not in the PR template. Maybe a Contributor statement on the README? 

Also we could put some guidance for how to do tests there!

shoyer:
> I very much agree with this, both as something that I feel and for newer contributors. Is there a way of making this more explicit? It's not in the PR template. Maybe a Contributor statement on the README?

We should really add a `CONTRIBUTING.md` file. GitHub surfaces those when users go to the page to open an issue or a pull request.

PR #1758, opened 2017-12-04
DOC: Added parameter list to open_dataarray

### LINKED ISSUE #1757: open_dataarray docs still contains *args, **kwargs
I noticed the [open_dataset](http://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html) docs provides the full input parameters whereas, [open_dataarray](http://xarray.pydata.org/en/stable/generated/xarray.open_dataarray.html) docs still lists `*args, **kwargs`
If you point me to where this is created I can change it, if the change is preferred

raybellwaves:
Closes #1757

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -397,7 +397,11 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
     --------
     open_dataset
     """
-    dataset = open_dataset(*args, **kwargs)
+    dataset = open_dataset(filename_or_obj, group=None, decode_cf=True,
+                           mask_and_scale=True, decode_times=True, 
+                           autoclose=False, concat_characters=True, 
+                           decode_coords=True, engine=None,chunks=None, 
+                           lock=None, cache=None, drop_variables=None)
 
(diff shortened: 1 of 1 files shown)


shoyer:
You need to pass these new arguments on to open_dataset.

xarray/backends/api.py
```diff
@@ -394,7 +397,11 @@ def open_dataarray(*args, **kwargs):
     --------
     open_dataset
     """
-    dataset = open_dataset(*args, **kwargs)
+    dataset = open_dataset(filename_or_obj, group=None, decode_cf=True,
```
spencerahill: You need to pass in the values from the `open_dataarray` call, not the defaults: `open_dataset(filename_or_obj, group=group, decode_cf=decode_cf, ...)`.  As written, none of the values other than `filename_or_obj` that the user specifies will get used: `mask_and_scale` will always be `True`, etc.

raybellwaves:
Apologies. Should have had a better looked at how `open_dataset` was laid out. Thanks for baring with me

fujiisoup:
Thanks, @raybellwaves.

PR #1762, opened 2017-12-06
ENH: Add dt.date accessor.

dcherian:
This PR lets us access `dt.date` thereby removing all higher frequency time information. 

*Use case:* Just like `dayofyear` but easier to interpret when only looking at 1 year of data. 

*Example:* Start with `da.time`
```
<xarray.DataArray 'time' (time: 8737)>
  array(['2014-01-01T00:00:00.000000000', '2014-01-01T01:00:00.000000000',
         '2014-01-01T02:00:00.000000000', ..., '2014-12-30T22:00:00.000000000',
         '2014-12-30T23:00:00.000000000', '2014-12-31T00:00:00.000000000'], dtype='datetime64[ns]')
  Coordinates:
    * time     (time) datetime64[ns] 2014-01-01 2014-01-01T01:00:00 ...
```

then `da.time.dt.date` yields
```
<xarray.DataArray 'date' (time: 8737)>
  array(['2014-01-01T00:00:00.000000000', '2014-01-01T00:00:00.000000000',
         '2014-01-01T00:00:00.000000000', ..., '2014-12-30T00:00:00.000000000',
         '2014-12-30T00:00:00.000000000', '2014-12-31T00:00:00.000000000'], dtype='datetime64[ns]')
  Coordinates:
    * time     (time) datetime64[ns] 2014-01-01 2014-01-01T01:00:00 ...
```

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -21,6 +21,9 @@ v0.10.1 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- `DatetimeAccessor` learned to return `date` i.e. `'2014-01-01T01:00:00.000000000'` is transformed to `'2014-01-01T00:00:00.000000000'`.
+  By `Deepak Cherian <https://github.com/dcherian>`_.
+
 - :py:func:`~plot.contourf()` learned to contour 2D variables that have both a 1D co-ordinate (e.g. time) and a 2D co-ordinate (e.g. depth as a function of time).
   By `Deepak Cherian <https://github.com/dcherian>`_.
 
--- a/xarray/core/accessors.py
+++ b/xarray/core/accessors.py
@@ -26,8 +26,11 @@ def _access_through_series(values, name):
     if name == "season":
         months = values_as_series.dt.month.values
         field_values = _season_from_months(months)
+    elif name == "date":
+        field_values = values_as_series.dt.floor('1D').values
     else:
         field_values = getattr(values_as_series.dt, name).values
+
     return field_values.reshape(values.shape)
 
 
@@ -145,4 +148,8 @@ class DatetimeAccessor(object):
 
     time = _tslib_field_accessor(
         "time", "Timestamps corresponding to datetimes", object
+    )
+
+    date = _tslib_field_accessor(
+        "date", "Date with hours, minutes etc. dropped", object
     )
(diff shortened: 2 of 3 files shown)


shoyer:
My main concern here is that pandas returns an array of `datetime.date` objects when you access `.date`. I think that's why we left it off in original implementation here:
```
In [1]: import pandas as pd

In [2]: t = pd.date_range('2010-01-01', periods=12, freq='3H')

In [3]: t
Out[3]:
DatetimeIndex(['2010-01-01 00:00:00', '2010-01-01 03:00:00',
               '2010-01-01 06:00:00', '2010-01-01 09:00:00',
               '2010-01-01 12:00:00', '2010-01-01 15:00:00',
               '2010-01-01 18:00:00', '2010-01-01 21:00:00',
               '2010-01-02 00:00:00', '2010-01-02 03:00:00',
               '2010-01-02 06:00:00', '2010-01-02 09:00:00'],
              dtype='datetime64[ns]', freq='3H')

In [4]: t.date
Out[4]:
array([datetime.date(2010, 1, 1), datetime.date(2010, 1, 1),
       datetime.date(2010, 1, 1), datetime.date(2010, 1, 1),
       datetime.date(2010, 1, 1), datetime.date(2010, 1, 1),
       datetime.date(2010, 1, 1), datetime.date(2010, 1, 1),
       datetime.date(2010, 1, 2), datetime.date(2010, 1, 2),
       datetime.date(2010, 1, 2), datetime.date(2010, 1, 2)], dtype=object)
```

Possibly implementing `.dt.floor()`, `.dt.ceil()` and `.dt.round()` (like pandas) would be a better way to do this? You would write something like `da.time.dt.floor('1D')` for this use-case.

dcherian:
@shoyer Didn't know `.floor()` was an option. I've used your suggestion now.

shoyer:
@dcherian I think I was unclear in my earlier comment. I don't like the current API, because it is inconsistent with `pandas.Series.dt.date` which returns an Series of `datetime.date` objects with `dtype=object`. Instead, why not implement methods `xarray.DataArray.dt.floor()`, `xarray.DataArray.dt.ceil()` and `xarray.DataArray.dt.round()`, like the pandas methods of the same name?

dcherian:
Oh, I see your point. Once implemented, would the round/ceil/floor methods work as arguments to groupby too? That's really what I wanted to do. 

Are you in favor of renaming the current functionality to`dt.yeardate` or some such for convenient usage? I expect it would be a common use case

shoyer:
> Once implemented, would the round/ceil/floor methods work as arguments to groupby too? 

Sure. You could write, e.g., `ds.groupby(ds.time.dt.floor('1D')).mean()`

> Are you in favor of renaming the current functionality todt.yeardate or some such for convenient usage? 

I would be inclined to stick to the datetime methods/properties from pandas: https://pandas.pydata.org/pandas-docs/stable/api.html#datetimelike-properties

PR #1777, opened 2017-12-12
Respect PEP 440

### LINKED ISSUE #1300: git version label yields version in violation of PEP 440
When an `xarray` installation does not match a released version ,it has a version number like `0.9.1-28-g769f120`.

This violates [PEP 440](https://www.python.org/dev/peps/pep-0440/), which leads to multiple problems:

* `pip install --update` will revert `xarray` back to `0.9.1`, because it does not recognise that `0.9.1-28-g769f120 > 0.9.1`
* packages with an `xarray` dependencies will be considered not satisfied.  Running a script through `setuptools` `load_entry_point` fails with `pkg_resources.ContextualVersionConflict: (xarray 0.9.1-28-g769f120 (/dev/shm/gerrit/venv/stable-3.5/lib/python3.5/site-packages), Requirement.parse('xarray>=0.8')`

Instead, the version number above should be written as `0.9.1+r10345` or so, which would satisfy PEP 440 and not cause problems with `pip` or `setuptools`.

gerritholl:
Change unreleased version numbers as to respect PEP 440.  Rather than
`0.10.0-9-g89a1a98` we will have `0.10.0+dev9.g89a1a98`.  This means
automated setuptools requirements can be respected.  Closes #1300.

 - [x] Closes #1300  (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,6 +110,9 @@ if not ISRELEASED:
         # Strip leading v from tags format "vx.y.z" to get th version string
         FULLVERSION = rev.lstrip('v')
 
+        # make sure we respect PEP 440
+        FULLVERSION = FULLVERSION.replace("-", "+dev", 1).replace("-", ".")
+
 else:
     FULLVERSION += QUALIFIER
 
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ if not ISRELEASED:
             # partial clone, manually construct version string
             # this is the format before we started using git-describe
             # to get an ordering on dev version strings.
-            rev = "v%s.dev-%s" % (VERSION, rev)
+            rev = "v%s+dev.%s" % (VERSION, rev)
 
         # Strip leading v from tags format "vx.y.z" to get th version string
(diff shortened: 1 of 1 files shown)


gerritholl:
Does this needs tests or a `whats-new.rst` entry?

gerritholl:
I'm aware that the longer term plan should be to use `versioneer`, but I think this fix is useful until we make such a transition.

shoyer:
This is fine by me. No need for tests or docs. (Travis-CI does install things from source, so that the fact that it installs is a good sign.)

setup.py
```diff
@@ -110,6 +110,9 @@
         # Strip leading v from tags format "vx.y.z" to get th version string
         FULLVERSION = rev.lstrip('v')
 
+        # make sure we respect PEP 440
+        FULLVERSION = FULLVERSION.replace("-", "+dev", 1).replace("-", ".")
```
shoyer: does line 108 need to be updated, too?

shoyer:
Thanks!

PR #1785, opened 2017-12-15
plot.line(): Draw multiple lines for 2D DataArrays.

dcherian:

Adds support for plotting multiple lines if `plot.line()` is provided with a 2D dataarray.

kwarg `x` lets you specify co-ordinate for x-axis.

Example:

xarray-multiple-line

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -177,7 +177,7 @@ def line(darray, *args, **kwargs):
         Axis on which to plot this figure. By default, use the current axis.
         Mutually exclusive with ``size`` and ``figsize``.
     x : string, optional
-        Coordinate for x axis (2D inputs only). If None use darray.dims[1]
+        Coordinate for x axis (2D inputs only). If None use longer dimension.
     add_legend : boolean, optional
         Add legend with y axis coordinates (2D inputs only).
     *args, **kwargs : optional
@@ -207,6 +207,9 @@ def line(darray, *args, **kwargs):
         x = darray.coords[xlabel]
 
     else:
+        if x is None:
+            x = darray.dims[np.argmax(darray.shape)]
+
         xlabel, ylabel = _infer_xy_labels(darray=darray, x=x, y=None)
         x = darray.coords[xlabel]
         darray = darray.transpose(xlabel, ylabel)
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -208,7 +208,8 @@ def line(darray, *args, **kwargs):
     if ndims == 1:
         xlabel, = darray.dims
         if x is not None and xlabel != x:
-            raise ValueError('Input does not have specified dimension ' + x)
+            raise ValueError('Input does not have specified dimension'
+                             + ' {!r}'.format(x))
 
         x = darray.coords[xlabel]
 
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -156,7 +156,7 @@ def plot(darray, row=None, col=None, col_wrap=None, ax=None, rtol=0.01,
 # matplotlib format strings
 def line(darray, *args, **kwargs):
     """
-    Line plot of 1 dimensional DataArray index against values
+    Line plot of DataArray index against values
 
     Wraps :func:`matplotlib:matplotlib.pyplot.plot`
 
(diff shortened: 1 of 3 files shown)


dcherian:
Now adds a legend too. Can be turned off with `add_legend=False`

xarray-multiple-line

fmaussion:
Looks quite good! Thanks. 

Sorry to be annoying ;), but there should be tests for the features you document:

> +    x : string, optional
> +        Coordinate for x axis (2D inputs only). If None use darray.dims[1]
> +    add_legend : boolean, optional
> +        Add legend with y axis coordinates (2D inputs only).

Currently, choosing another dim for the x axis and the legend kwargs are not tested.

dcherian:
@fmaussion I've added more tests. Let me know if you'd like more changes.

shoyer:
Is there a keyword argument name that it would make sense to use for the dimension that is repeated into multiple lines? Maybe `hue`, as is used by Seaborn?

I would like there to be a fully explicit way to make these plots.

dcherian:
@fmaussion I've added one more commit. If `x` kwarg is not provided, we automatically choose the longer dimension. So if you have a 10000x3 DataArray, this will plot 3 lines instead of 10000.

Extended an existing test

dcherian:
@shoyer `hue` sounds perfect. I'll work on adding that.

shoyer:
>  If x kwarg is not provided, we automatically choose the longer dimension. So if you have a 10000x3 DataArray, this will plot 3 lines instead of 10000.

I am very nervous about automated heuristics for choosing behavior. I would much rather we raise an error message in cases like this, rather than guessing. (The problem is that heuristics can make it very hard to predict/understand how code will work without trying it.)

dcherian:
I've added support for a `hue` kwarg in the latest commit. Let me know what you think.

> I am very nervous about automated heuristics for choosing behavior. I would much rather we raise an error message in cases like this, rather than guessing. (The problem is that heuristics can make it very hard to predict/understand how code will work without trying it.)

Well, this behaviour is analogous to automatically choosing `x,y` in `_plot2d`. xarray is already pretty opinionated in this respect.

The current behaviour for a 10000x3 array is to plot 10000 lines which is bad. I'm OK with adding an error message but strongly feel that choosing to plot 3 lines (i.e. always smallest number of lines) is a good default. 

Re:error, would that be a message stating that either `x` or `hue` be specified if the input is 2D?

shoyer:
> Well, this behaviour is analogous to automatically choosing x,y in _plot2d. xarray is already pretty opinionated in this respect.

We use the order of the dimensions on the array for choosing how to plot it. The analogous behavior would be to always plot longer dimension along the x-axis, which isn't what we do.

> The current behaviour for a 10000x3 array is to plot 10000 lines which is bad. I'm OK with adding an error message but strongly feel that choosing to plot 3 lines (i.e. always smallest number of lines) is a good default.

I agree that users probably rarely want 10,000 lines :). That's a good reason to require an explicit choice here.

The problem are edge cases like a 5x6 array. Do you want 5 lines of 6 points each or 6 lines of 5 points each? If we make the heuristic depend on the size of the array, then it will give very hard to understand what happens when the array shape changes slightly.

> Re:error, would that be a message stating that either x or hue be specified if the input is 2D?

Yes, that sounds right.

dcherian:
@shoyer , @fmaussion : Updated to require either `x` or `hue` kwarg if input is 2D.

xarray/plot/plot.py
```diff
@@ -193,11 +199,26 @@ def line(darray, *args, **kwargs):
...
+    add_legend = kwargs.pop('add_legend', True)
 
     ax = get_axis(figsize, size, aspect, ax)
 
-    xlabel, = darray.dims
-    x = darray.coords[xlabel]
+    if ndims == 1:
+        xlabel, = darray.dims
+        if x is not None and xlabel != x:
+            raise ValueError('Input does not have specified dimension ' + x)
```
shoyer: please use string formatting instead so this still works even if the dimensions is not a string, e.g.,
```
raise ValueError('Input does not have specified dimension {!r}'.format(x))
```
(Using non-strings as dimension names is rare but we still try to support it when feasible.)

dcherian:
@shoyer Used your suggested change.

xarray/plot/plot.py
```diff
@@ -208,7 +208,8 @@ def line(darray, *args, **kwargs):
     if ndims == 1:
         xlabel, = darray.dims
         if x is not None and xlabel != x:
-            raise ValueError('Input does not have specified dimension ' + x)
+            raise ValueError('Input does not have specified dimension'
+                             + ' {!r}'.format(x))
```
shoyer: just for future reference: you can skip the `+` character and rely on implicit string concatenation.

dcherian: Didn't know about that. Thanks.

shoyer:
Thanks!

shoyer:
I merged this, but it might also be nice to add an example (or at least brief mention) to the narrative docs for 1D plotting. See:
http://xarray.pydata.org/en/stable/plotting.html#one-dimension
https://github.com/pydata/xarray/blob/master/doc/plotting.rst

PR #1786, opened 2017-12-15
_plot2d: Rotate x-axis ticks if datetime subtype

dcherian:
Rotate x-axis dateticks by default, just as for plot.line()


xarray/plot/plot.py
```diff
@@ -530,6 +530,10 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
 
         _update_axes_limits(ax, xincrease, yincrease)
 
+        # Rotate dates on xlabels
+        if np.issubdtype(xval.dtype, np.datetime64):
+            plt.gcf().autofmt_xdate()
```
rabernat: Since the axis object is already known, better to use `ax.autofmt_xdate` than `plt.gcf()...`.

dcherian: `autofmt_xdate` is a `Figure` method so that won't work. I could set the rotation manually though...

rabernat: ah right, my mistake.

You should be able to use [`ax.get_figure()`](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.get_figure.html#matplotlib.axes.Axes.get_figure) to get the figure. For some reason, `gcf()` seems error prone, but I don't really have any evidence for that.

dcherian: Sounds good to me. Changed the corresponding call in `plot.line()` too

xarray/tests/test_plot.py
```diff
@@ -615,6 +615,14 @@ def test_xyincrease_true_changes_axes(self):
         diffs = xlim[0] - 0, xlim[1] - 14, ylim[0] - 0, ylim[1] - 9
         self.assertTrue(all(abs(x) < 1 for x in diffs))
 
+    def test_x_ticks_are_rotated_for_time(self):
+        time = pd.date_range('2000-01-01', '2000-01-10')
+        a = DataArray(np.random.randn(2, len(time)),
+                      [('xx', [1, 2]), ('t', time)])
+        a.plot(x='t')
+        rotation = plt.gca().get_xticklabels()[0].get_rotation()
+        self.assertFalse(rotation == 0)
```
shoyer: Nit: prefer `self.assertEqual(rotation, 0)`

dcherian: Yup. Made the change.

fmaussion:
Oh I just noticed I'm able to do it myself. Will wait for the tests to pass and merge

PR #1787, opened 2017-12-15
Include units (if set) in plot labels

### LINKED ISSUE #1630: Option to include units in .plot()
I find myself editing the label on the colorbar of xarray plots often to include the units of the plotted variable.
Would it be possible to extract the value of DataArray.attrs['units] (if present) and add it to the colorbar label?

Given an example DataArray with name='temperature' and attrs['units']='deg C', the current colorbar label says 'temperature'.

A 'plot_units' keyword could possibly be used like this for example:

```python
da.plot(plot_units=True)
```
Then the colorbar label should be 'temperature [deg C]'.

Zac-HD:
- [x] Closes #1630
 - [x] Tests passed
 - [x] Fully documented, including `whats-new.rst` for all changes - details of label not previously documented

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -38,11 +38,14 @@ Enhancements
   By `Matti Eskelinen <https://github.com/maaleske>`_.
 - :py:func:`~plot()` learned to rotate x-axis ticks if x-axis is time.
   By `Deepak Cherian <https://github.com/dcherian>`_.
+- Plots of a :obj:`DataArray` with ``units`` in the attrs dict are now
+  labelled ``name (units)`` instead of just ``name`` (:issue:`1630`).
+  By `Zac Hatfield-Dodds <https://github.com/Zac-HD>`_.
 - :py:func:`~plot.line()` learned to draw multiple lines if provided with a
   2D variable.
   By `Deepak Cherian <https://github.com/dcherian>`_.
 
-.. _Zarr: http://zarr.readthedocs.io/
+.. _Zarr: https://zarr.readthedocs.io/
 
 .. _Iris: http://scitools.org.uk/iris
 
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -228,7 +228,10 @@ def line(darray, *args, **kwargs):
     ax.set_title(darray._title_for_slice())
 
     if darray.name is not None:
-        ax.set_ylabel(darray.name)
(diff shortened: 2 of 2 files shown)


fujiisoup:
I am slightly wondering if there are some relavents to 'units',  such as ' unit', 'Units', or 'UNIT', as xarray does not define how units should be stored (Actually my data has 'Unit').

Zac-HD:
[`units` is required by the CF Conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#units) and most similar conventions, and very common in other landscape and hydrology data I work with.

Zac-HD:
Ping @fmaussion for review?  (the build error is Travis download trouble, not failing tests)

fmaussion:
In general I'm in favor for this change: adding units to the colorbar is the the first thing I teach my students in class. afaik, we don't do anything with units in xarray (except time handling), so this would be a first time thing and @shoyer maybe wants to chime in here.

Also:
- other labels could print their units too: coordinates on x and y axis labels might have units, e.g. lons/lats (the same way you added it to the y-label for 1d plots).
- are parentheses `Velocity (m/s)` the correct way to represent units in graphics? I got used to brackets `Velocity [m/s]` and thought this was the standard but internet and stackoverflow seem to have various opinions on the matter
- there should be tests for this. The fact that your change didn't break anything is that we have no data attribute called `units` in the graphics test suite: we should change this if we go for this change.

rabernat:
Perhaps we could default to “units” for the unit attribute but allow a custom attribute to be set via xarray.set_options?

Zac-HD:
- Including units in labels more generally sounds good to me, but I'll hold off until the general concept is approved 😉
- I don't think there's an objectively correct way to label units, but parentheses seem to be fairly standard and least prone to misinterpretation.
- Happy to add tests, though again I'll wait for in-principle approval.  Adding a `units` attribute to some of the test data is easy enough!
- `units` is the convention, and if this is unsatisfactory it's quite trivial to override already - I'd hold off on custom options until someone has an actual use-case for the feature.

Zac-HD:
This is now far enough down my list of priorities that I'm unlikely to get to it any time soon, and (it turns out) there are also some design issues that would need to be resolved.

I therefore suggest that anyone wishing to take this on opens an issue first to clarify the scope of any such change, and then make a PR.

PR #1791, opened 2017-12-19
Cookbook docs

### LINKED ISSUE #1790: 'Cookbook' page
Would there be interest in adding a 'cookbook' section to the docs a-la [Pandas](http://pandas.pydata.org/pandas-docs/stable/cookbook.html)? The current Recipes section might then be better renamed as Gallery?

It's useful for the kind of thing which isn't a full-fledged example, but might be useful nonetheless. I'll hapily put together a pull-request if there's interest.

duncanwp:
Closes #1790 

This is a bit empty at the moment so feel free to add any snippets you think would be useful :-)

### DIFF
--- a/doc/examples/cookbook.rst
+++ b/doc/examples/cookbook.rst
@@ -7,11 +7,8 @@ This is a repository for short and sweet examples and links for useful xarray re
 
 Adding interesting links and/or inline examples to this section is a great First Pull Request.
 
-Simplified, condensed, new-user friendly, in-line examples have been inserted where possible to augment the Stack-Overflow and GitHub links. Many of the links contain expanded information, above what the in-line examples offer.
-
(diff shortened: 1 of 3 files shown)


doc/examples/cookbook.rst
```diff
@@ -0,0 +1,36 @@
...
+
+.. ipython:: python
+
+   da = xr.DataArray(np.random.rand(4, 5), dims=['x', 'y'],
+                     coords=dict(x=[40, 30, 20, 10],
+                                 y=pd.date_range('2000-01-01', periods=5)))
+
+   da
+
+   np.flip(da, da.dims.index('x'))
```
maxim-lian: The cookbook would be great! Thanks for giving it an initial go.

Unfortunately `flip` is the one function that is about to be in the library!

Other good starting points might be:
- Constructing Datasets & DataArrays (I know there's a learning curve 'bump' vs pandas here)
- Merge & concats
- Shifting while extending a dimension (?)

PR #1793, opened 2017-12-19
fix distributed writes

### LINKED ISSUE #1464: Writing directly to a netCDF file while using distributed
I've been experimenting with `distributed` recently and have run into an issue when saving a result directly to a file using the `netcdf4` engine.  I've found if I compute things before saving to a file (thus loading the result into memory before calling `to_netcdf`) things work OK.  I attached a minimum working example below.  

Can others reproduce this?   Part of me thinks there must be something wrong with my setup, because I'm somewhat surprised something like this wouldn't have come up already (apologies in advance if that's the case).

```
In [1]: import dask

In [2]: import distributed

In [3]: import netCDF4

In [4]: import xarray as xr

In [5]: dask.__version__
Out[5]: '0.15.0'

In [6]: distributed.__version__
Out[6]: '1.17.1'

In [7]: netCDF4.__version__
Out[7]: '1.2.9'

In [8]: xr.__version__
Out[8]: '0.9.6'

In [9]: da = xr.DataArray([1., 2., 3.])

In [10]: da.to_netcdf('no-dask.nc')

In [11]: da.chunk().to_netcdf('dask.nc')  # Not using distributed yet

In [12]: c = distributed.Client()  # Launch a LocalCluster (now using distributed)

In [13]: c
Out[13]: <Client: scheduler='tcp://127.0.0.1:44576' processes=16 cores=16>

In [14]: da.chunk().to_netcdf('dask-distributed-netcdf4.nc', engine='netcdf4')
EOFError                                  Traceback (most recent call last)
<ipython-input-14-98490239a35f> in <module>()
----> 1 da.chunk().to_netcdf('dask-distributed-netcdf4.nc', engine='netcdf4')

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/core/dataarray.py in to_netcdf(self, *args, **kwargs)
   1349             dataset = self.to_dataset()
   1350
-> 1351         dataset.to_netcdf(*args, **kwargs)
   1352
   1353     def to_dict(self):

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/core/dataset.py in to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims)
    975         return to_netcdf(self, path, mode, format=format, group=group,
    976                          engine=engine, encoding=encoding,
--> 977                          unlimited_dims=unlimited_dims)
    978
    979     def __unicode__(self):

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/backends/api.py in to_netcdf(dataset, path_or_file, mode, format, group, engine, writer, encoding, unlimited_dims)
    571     try:
    572         dataset.dump_to_store(store, sync=sync, encoding=encoding,
--> 573                               unlimited_dims=unlimited_dims)
    574         if path_or_file is None:
    575             return target.getvalue()

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/core/dataset.py in dump_to_store(self, store, encoder, sync, encoding, unlimited_dims)
    916                     unlimited_dims=unlimited_dims)
    917         if sync:
--> 918             store.sync()
    919
    920     def to_netcdf(self, path=None, mode='w', format=None, group=None,

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/backends/netCDF4_.py in sync(self)
    334     def sync(self):
    335         with self.ensure_open(autoclose=True):
--> 336             super(NetCDF4DataStore, self).sync()
    337             self.ds.sync()
    338

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/backends/common.py in sync(self)
    200
    201     def sync(self):
--> 202         self.writer.sync()
    203
    204     def store_dataset(self, dataset):

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/xarray/backends/common.py in sync(self)
    177             import dask
    178             if LooseVersion(dask.__version__) > LooseVersion('0.8.1'):
--> 179                 da.store(self.sources, self.targets, lock=GLOBAL_LOCK)
    180             else:
    181                 da.store(self.sources, self.targets)

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/dask/array/core.py in store(sources, targets, lock, regions, compute, **kwargs)
    922     dsk = sharedict.merge((name, updates), *[src.dask for src in sources])
    923     if compute:
--> 924         Array._get(dsk, keys, **kwargs)
    925     else:
    926         from ..delayed import Delayed

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/dask/base.py in _get(cls, dsk, keys, get, **kwargs)
    102         get = get or _globals['get'] or cls._default_get
    103         dsk2 = optimization_function(cls)(ensure_dict(dsk), keys, **kwargs)
--> 104         return get(dsk2, keys, **kwargs)
    105
    106     @classmethod

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/client.py in get(self, dsk, keys, restrictions, loose_restrictions, resources, sync, **kwargs)
   1762         if sync:
   1763             try:
-> 1764                 results = self.gather(packed)
   1765             finally:
   1766                 for f in futures.values():

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/client.py in gather(self, futures, errors, maxsize, direct)
   1261         else:
   1262             return self.sync(self._gather, futures, errors=errors,
-> 1263                              direct=direct)
   1264
   1265     @gen.coroutine

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/client.py in sync(self, func, *args, **kwargs)
    487             return future
    488         else:
--> 489             return sync(self.loop, func, *args, **kwargs)
    490
    491     def __str__(self):

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/utils.py in sync(loop, func, *args, **kwargs)
    232         e.wait(1000000)
    233     if error[0]:
--> 234         six.reraise(*error[0])
    235     else:
    236         return result[0]

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    684         if value.__traceback__ is not tb:
    685             raise value.with_traceback(tb)
--> 686         raise value
    687
    688 else:

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/utils.py in f()
    221                 raise RuntimeError("sync() called from thread of running loop")
    222             yield gen.moment
--> 223             result[0] = yield make_coro()
    224         except Exception as exc:
    225             logger.exception(exc)

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/tornado/gen.py in run(self)
   1013
   1014                     try:
-> 1015                         value = future.result()
   1016                     except Exception:
   1017                         self.had_exception = True

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/tornado/concurrent.py in result(self, timeout)
    235             return self._result
    236         if self._exc_info is not None:
--> 237             raise_exc_info(self._exc_info)
    238         self._check_done()
    239         return self._result

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/tornado/util.py in raise_exc_info(exc_info)

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/tornado/gen.py in run(self)
   1019
   1020                     if exc_info is not None:
-> 1021                         yielded = self.gen.throw(*exc_info)
   1022                         exc_info = None
   1023                     else:

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/client.py in _gather(self, futures, errors, direct)
   1154                             six.reraise(type(exception),
   1155                                         exception,
-> 1156                                         traceback)
   1157                     if errors == 'skip':
   1158                         bad_keys.add(key)

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    683             value = tp()
    684         if value.__traceback__ is not tb:
--> 685             raise value.with_traceback(tb)
    686         raise value
    687

/nbhome/skc/miniconda3/envs/research/lib/python3.6/site-packages/distributed/protocol/pickle.py in loads()
     57 def loads(x):
     58     try:
---> 59         return pickle.loads(x)
     60     except Exception:
     61         logger.info("Failed to deserialize %s", x[:10000], exc_info=True)

EOFError: Ran out of input
```

If I load the data into memory first by invoking `compute()` things work OK:

```
In [15]: da.chunk().compute().to_netcdf('dask-distributed-netcdf4.nc', engine='netcdf4')
```

jhamman:
- [x] Closes #1464 
 - [x] Tests added
 - [x] Tests passed

Right now, I've just modified the dask distributed integration tests so we can all see the [failing tests](https://travis-ci.org/jhamman/xarray/jobs/317603224#L4400-L4571). 

I'm happy to push this further but I thought I'd see if either @shoyer or @mrocklin have an idea where to start?


shoyer:
yes, see https://github.com/pydata/xarray/issues/1464#issuecomment-341329662

mrocklin:
The zarr test seems a bit different.  I think your issue here is that you are trying to use synchronous API with the async test harness.  I've changed your test and pushed to your branch (hope you don't mind).  Relevant docs are here: http://distributed.readthedocs.io/en/latest/develop.html#writing-tests

Async testing is nicer in many ways, but does require you to be a bit familiar with the async/tornado API.  I also suspect that operations like `to_zarr` really aren't yet async friendly.

xarray/backends/netCDF4_.py
```diff
@@ -41,6 +41,11 @@ def __init__(self, variable_name, datastore):
             dtype = np.dtype('O')
         self.dtype = dtype
 
+    def __setitem__(self, key, value):
+        with self.datastore.ensure_open(autoclose=True):
+            data = self.get_array()
+            data[key] = value
+
```
jhamman: @shoyer, is this what you were describing in https://github.com/pydata/xarray/issues/1464#issuecomment-341329662

shoyer: Yes, this looks right to me.

jhamman:
@mrocklin - 

I have a [test failing here](https://travis-ci.org/pydata/xarray/jobs/327790224#L5643) with a familiar message. 

    E       TypeError: 'Future' object is not iterable

We saw this last week when debugging some pangeo things. Can you remind me what our solution was?

mrocklin:
I don't know.  I would want to look at the fail case locally.  I can try to do this near term, no promises though :/

jhamman:
I've just taken another swing at this and come up empty. I open to ideas in the following areas:

1. scipy backend is failing to roundtrip a length 1 datetime array: https://travis-ci.org/pydata/xarray/jobs/333068098#L4504
2. scipy, netcdf4, and h5netcdf backends are all failing inside dask-distributed: https://travis-ci.org/pydata/xarray/jobs/333068098#L4919

The good news here is that only 8 tests are failing after applying the array wrapper so I suspect we're quite close.  I'm hoping @shoyer may have some ideas on (1) since I think he had implemented some scipy workarounds in the past. @mrocklin, I'm hoping you can point me in the right direction.

All of these tests are reproducible locally.

xarray/backends/scipy_.py
```diff
@@ -55,6 +55,18 @@ def __getitem__(self, key):
             copy = self.datastore.ds.use_mmap
             return np.array(data, dtype=self.dtype, copy=copy)
 
+    def __setitem__(self, key, value):
+        with self.datastore.ensure_open(autoclose=True):
+            data = self.get_array()
```
shoyer: This needs to be `self.datastore.ds.variables[self.variable_name]` (a `netcdf_variable` object), not `self.get_array()`, which returns the `.data` (a numpy array). You can't enlarge a numpy array, but *can* enlarge a scipy netcdf variable.

(This manifests itself in writing a netcdf file with a time dimension of length 0. Xarray then crashes when attempting to a decode a length 0 time variable, which is an unrelated bug.)

jhamman: Thanks @shoyer - this appears to fix the scipy issue.

rabernat:
Kudos for pushing this forward. I don't have much help to offer, but I wanted to recognize your effort...this is hard stuff!.

shoyer:
Has anyone successfully used `dask.array.store()` with the distributed scheduler?

mrocklin:
I can take a look at the future not iterable issue sometime tomorrow.  

> Has anyone successfully used dask.array.store() with the distributed scheduler?

My guess is that this would be easy with a friendly storage target.  I'm not sure though.  cc @jakirkham who has been active on this topic recently.

jakirkham:
Yep, using `dask.array.store` regularly with the `distributed` scheduler both on our cluster and in a local Docker image for testing. Am using Zarr Arrays as the targets for `store` to write to. Basically rechunk the data to match the chunking selected for the Zarr Array and then write out in parallel lock-free.

Our cluster uses NFS for things like one's home directory. So these are accessible across nodes. Also there are other types of storage available that are a bit faster and still remain accessible across nodes. So these work pretty well.

jhamman:
Yes, the zarr backend here in xarray is also using `dask.array.store` and seems to work with distributed just fine.

rabernat:
I have definitely used the distributed scheduler with `dask.array.store` both via Zarr and via a custom store class I wrote: https://gist.github.com/rabernat/e54755e7de4eb5a93cc4e7f9f903e3cc

But I cannot recall if I ever got it to work with netCDF. Probably not.

jhamman:
xref: https://github.com/pydata/xarray/issues/798 and https://github.com/dask/dask/issues/2488 which are both seem to be relevant to this discussion.

I'm also remembering @pwolfram was quite involved with the original distributed integration so pinging him to see if he is interested in this.

shoyer:
Looking into this a little bit, this looks like a dask-distributed bug to me. Somehow `Client.get()` is returning a `tornado.concurrent.Future` object, even though `sync=True`.

mrocklin:
For reference, the line

    computed = restored.compute()

would have to be replaced with

    (computed,) = yield c.compute(restored)

To get the same result.  However there were a few more calls to compute hidden in various functions (like `to_netcdf`) that would be tricky to make asynchronous, so I opted to swich to synchronous style instead.

jhamman:
Thanks @mrocklin for taking a look here. I reworked the tests a bit more to put the `to_netcdf` inside the distributed cluster section.

Bad news is that the tests are failing again. The good news is we have a semi-informative error message that indicates we're missing a `lock` somewhere.

jhamman:
The tests failure indicates that the netcdf4/h5netcdf libraries cannot open the file in write/append mode, and it seems that is because the file is already open (by another process).

Two questions: 

1. `autoclose` is False `to_netcdf`. That generally makes sense to me but I'm concerned that we're not being explicit enough about closing the file after each process is done interacting with it. Do we have a way to lock until the file is closed?
2. The lock we're using is dask's `SerializableLock`. Is that the correct Lock to be using? There is also the `distributed.Lock`.

xref: https://github.com/dask/dask/issues/1892

mrocklin:
SerializableLock isn't appropriate here if you want inter process locking.
Dask's lock is probably better here if you're running with the distributed
scheduler.

shoyer:
We might always need to use `autoclose=True` with distributed. The problem is that in xarray's default mode of operation, we open a netCDF file (without using dask) to create variables, dimensions and attributes, keeping the file open. Then we write the data using dask (via `AbstractWritableDataStore.sync()`), but the original file is still open.

As for the lock, we need locking both:
- Per process: only one thread can use HDF5 for reading/writing at the same time.
- Per file: only one worker can read/write a file at the same time.

shoyer:
A simpler way to handle locking for now (but with possibly subpar performance) would be to use a single global distributed lock.

As for `autoclose`, perhaps we should make the default `autoclose=None`, which becomes `True` if using dask-distributed (or maybe using dask in general?) and otherwise is `False`.

mrocklin:
Performance-wise Dask locks will probably add 1-10ms of communication overhead (probably on the lower end of that), plus whatever contention there will be from locking.  You can make these locks as fine-grained as you want, for example by defining a lock-per-filename with `Lock(filename)` with no cost (which would presumably reduce contention).

jhamman:
@mrocklin - What is the preferred method for determining which scheduler is being used?

mrocklin:
Honestly we don't have a very clean mechanism for this.  Probably you want to look at `dask.context._globals['get']`.  This should either be `None`, which means "use the collection's default" (`dask.threaded.get` in your case) or a callable.  If you're using the distributed scheduler then this will be a method of a `Client` object.  Again, not a very clean thing to test for.  My apologies.

xarray/backends/common.py
```diff
@@ -27,8 +22,41 @@
...
+        return Lock
+    elif scheduler == 'multiprocessing':
+        import multiprocessing as mp
+        return mp.Manager().Lock
+    elif scheduler == 'threaded':
+        from dask.utils import SerializableLock
+        return SerializableLock
+    else:
+        from threading import Lock
+        return Lock
```
jhamman: @mrocklin - this is what I came up with. Similar to  `dask.utils.get_scheduler_lock` except it breaks the determination of the scheduler and the lock selection into two steps, and it includes the distributed lock. Its a little ugly but seems to be doing the write thing. I'm thinking something like this may have a place in dask (or distributed)

xarray/backends/common.py
```diff
@@ -178,24 +206,21 @@ def add(self, source, target):
...
-                target[...] = source
-            except TypeError:
-                # workaround for GH: scipy/scipy#6880
-                target[:] = source
+            target[...] = source
 
     def sync(self):
         if self.sources:
             import dask.array as da
+            print('self.lock == ', self.lock)
```
jhamman: Now we pass `dask.distributed.Lock(filename)` here. Interesting thing is that the test distributed test cases are failing when `da.store(..., lock=dask.distributed.Lock(filename))` is called with a 

```
TypeError: can't pickle _thread.lock objects
```

So I'm clearly missing another lock. Honestly, I'm creeping pretty far outside my comfort zone on whether or not I know what I'm doing so thoughts/suggestions at this point would be very welcome.

mrocklin: This is an oversight on the Dask side.  Historically people have explicitly created locks within their functions rather than passed them around.  I believe that this is easily resolved in https://github.com/dask/distributed/pull/1738

However now I get different errors like the following in the XArray tests

```python
distributed.worker - WARNING -  Compute Failed
Function:  store_chunk
args:      (array([[ 0.02069011, -1.15082717, -0.04842995],
       [ 0.54649439,  0.39982083,  1.22082541],
       [-0.8788512 , -1.49633607, -0.49022316],
       [-1.16940102, -0.03950083,  1.15521853]]), <xarray.backends.netCDF4_.NetCDF4ArrayWrapper object at 0x7f14e7e4ab38>, (slice(0, 4, None), slice(0, 3, None)), <distributed.lock.Lock object at 0x7f14e7e7e0f0>, None, False)
kwargs:    {}
Exception: KeyError('var2',)
```

jhamman: Thanks @mrocklin - This looks like a problem I can handle.

xarray/backends/common.py
```diff
@@ -27,8 +22,41 @@
...
+    except ImportError:
+        return None
+
+
+def get_scheduler_lock(scheduler):
+    if scheduler == 'distributed':
+        from dask.distributed import Lock
+        return Lock
+    elif scheduler == 'multiprocessing':
+        import multiprocessing as mp
```
shoyer: No need for the guarded imports here for multiprocessing or threading. These are in the standard library, so they can be imported at the top level.

xarray/backends/common.py
```diff
@@ -27,8 +22,41 @@
...
+def get_scheduler(get=None, collection=None):
+    try:
+        from dask.utils import effective_get
+        actual_get = effective_get(get, collection)
+        try:
+            from dask.distributed import Client
+            if isinstance(actual_get.__self__, Client):
+                return 'distributed'
+        except (ImportError, AttributeError):
+            import dask.multiprocessing
```
shoyer: I think we need to guard against this line potentially raising `ImportError` because cloudpickle is missing (which is only required for dask.multiprocessing).

xarray/backends/common.py
```diff
@@ -27,8 +22,41 @@
...
+    elif scheduler == 'threaded':
+        from dask.utils import SerializableLock
+        return SerializableLock
+    else:
+        from threading import Lock
+        return Lock
+
+
+SCHEDULER = get_scheduler()
+GLOBAL_LOCK = get_scheduler_lock(SCHEDULER)()
```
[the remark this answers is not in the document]
shoyer: Actually, for what we need here (we should call it `HDF5_LOCK` for clarity), we probably don't even need something to be scheduler specific at all. It can be a SerializableLock if dask is installed, or otherwise just a threading.Lock.

xarray/backends/api.py
```diff
@@ -617,8 +617,15 @@ def to_netcdf(dataset, path_or_file=None, mode='w', format=None, group=None,
     # if a writer is provided, store asynchronously
     sync = writer is None
 
+    # TODO Move this logic outside of this function
+    from .common import get_scheduler, get_scheduler_lock
+    scheduler = get_scheduler()
+    lock = get_scheduler_lock(scheduler)(path_or_file)
```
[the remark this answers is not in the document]
shoyer: It occurs to me that this becomes even more complicated if we allow for reading/writing netCDF3 files with SciPy, which certainly does not need an HDF5 lock. But perhaps it's not worth worrying about that yet.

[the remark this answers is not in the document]
mrocklin: > Only a single process will be creating file_lock, so we don't need to worry about including the filename as part of the lock

Do you ever save to multiple files at the same time?  Are we locking around the file or around the HDF5 library?  If we have many files and we're locking around the file then having a different lock per file might reduce contention, especially if there are very many files.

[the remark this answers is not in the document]
jhamman: Yes, there is the [`save_mfdataset`](http://xarray.pydata.org/en/latest/generated/xarray.save_mfdataset.html#xarray.save_mfdataset) function. I haven't gotten there yet but this is a use case we'll want to support.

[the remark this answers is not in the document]
shoyer: > Are we locking around the file or around the HDF5 library?

Per process, only one thread can use the HDF5 library (it is not threadsafe). Across multiple processes, there can be multiple readers per file but only one writer.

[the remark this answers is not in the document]
mrocklin: I see, so your choice of ordering locks here is quite intentional:

```python
lock = CombinedLock([HDF5_LOCK, file_lock])
```

Grab the lock for this process first, then grab the lock for the file that we're trying to read.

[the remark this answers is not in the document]
shoyer: That's a good point that the ordering of the locks may matter. I hadn't thought that through yet.

It might actually be better to lock files first, if we expect the HDF5 lock to have more contention. I suppose this could be answered experimentally.

Potentially it could be useful for the dask scheduler to understand these limitations, too, by encouraging it to schedule writes to the same file in the same process.

jhamman:
I think we're getting close. We're currently failing during the `sync` step and I'm hypothesizing that it is due to the file not being closed after the setup steps. That said, I wasn't able to pinpoint why/where we're missing a close.  I think this traceback is pretty informative: 

```
  File "/Users/jhamman/anaconda/envs/xarray36/lib/python3.6/site-packages/distributed/worker.py", line 1255, in add_task
    self.tasks[key] = _deserialize(function, args, kwargs, task)
  File "/Users/jhamman/anaconda/envs/xarray36/lib/python3.6/site-packages/distributed/worker.py", line 641, in _deserialize
    args = pickle.loads(args)
  File "/Users/jhamman/anaconda/envs/xarray36/lib/python3.6/site-packages/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
  File "/Users/jhamman/Dropbox/src/xarray/xarray/backends/common.py", line 445, in __setstate__
    self.ds = self._opener(mode=self._mode)
  File "/Users/jhamman/Dropbox/src/xarray/xarray/backends/netCDF4_.py", line 204, in _open_netcdf4_group
    ds = nc4.Dataset(filename, mode=mode, **kwargs)
  File "netCDF4/_netCDF4.pyx", line 2015, in netCDF4._netCDF4.Dataset.__init__
  File "netCDF4/_netCDF4.pyx", line 1636, in netCDF4._netCDF4._ensure_nc_success
OSError: [Errno -101] NetCDF: HDF error: b'/var/folders/v0/qnh7jvgx5gnglpxfztxdlhk00000gn/T/tmpn_mo662_/temp-0.nc'
distributed.scheduler - ERROR - error from worker tcp://127.0.0.1:63248: [Errno -101] NetCDF: HDF error: b'/var/folders/v0/qnh7jvgx5gnglpxfztxdlhk00000gn/T/tmpn_mo662_/temp-0.nc'
HDF5-DIAG: Error detected in HDF5 (1.10.1) thread 140736093991744:
  #000: H5F.c line 586 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1305 in H5F_open(): unable to lock the file
    major: File accessibilty
    minor: Unable to open file
  #002: H5FD.c line 1839 in H5FD_lock(): driver lock request failed
    major: Virtual File Layer
    minor: Can't update object
  #003: H5FDsec2.c line 940 in H5FD_sec2_lock(): unable to lock file, errno = 35, error message = 'Resource temporarily unavailable'
    major: File accessibilty
    minor: Bad file ID accessed
```

jhamman:
@shoyer - I have this working with the `netcdf4` backend for with the `NETCDF3_CLASSIC` file format. I'm still having some locking issues with the HDF library and I'm not sure why.

xarray/backends/common.py
```diff
@@ -374,14 +436,19 @@ class DataStorePickleMixin(object):
     def __getstate__(self):
         state = self.__dict__.copy()
         del state['ds']
+        state['_isopen'] = False
```
jhamman: This was a tricky one but it was messing up autoclose.

xarray/backends/common.py
```diff
@@ -374,14 +436,19 @@ class DataStorePickleMixin(object):
...
         return state
 
     def __setstate__(self, state):
         self.__dict__.update(state)
-        self.ds = self._opener(mode=self._mode)
+        if self._autoclose:
+            with self.ensure_open(True):
+                self.ds = self._opener(mode=self._mode)
+        else:
+            self.ds = self._opener(mode=self._mode)
```
jhamman: This isn't pretty but we don't really want to open the dataset when un-pickling. This actually opens then closes. Maybe this needs a special lock statement?

xarray/backends/common.py
```diff
@@ -391,7 +458,7 @@ def ensure_open(self, autoclose):
 
         Use requires `autoclose=True` argument to `open_mfdataset`.
         """
-        if self._autoclose and not self._isopen:
+        if autoclose and not self._isopen:
```
jhamman: I'm not sure, but this may have been a bug. We might consider some additional logic to handle what to do when `self._autoclose` and `autoclose` conflict.

xarray/backends/common.py
shoyer: I think we could only restore `self.ds` if not using autoclose. There's no point in restoring the dataset only to immediately close it, given that we assign `self.ds` in `ensure_open()` anyways if using autoclose.

So this could just be:
```python
if not self._autoclose:
    self.ds = self._opener(mode=self._mode)
```

jhamman: Right. I ended up doing something a bit different by writing `ds` as a property. I'm not sure I like it but it seems to work.

jhamman:
I've this down to 4 test failures: 

```
test_dask_distributed_netcdf_integration_test[NETCDF3_CLASSIC-True-scipy]
test_dask_distributed_netcdf_integration_test[NETCDF3_CLASSIC-False-scipy]
test_dask_distributed_netcdf_integration_test[NETCDF4_CLASSIC-False-netcdf4]
test_dask_distributed_netcdf_integration_test[NETCDF4-False-netcdf4]
```

I think I'm ready for an initial review. I've made some changes to autoclose and sync so I'd like to get feedback on my approach before I spend too much time sorting out the last few failures.

xarray/tests/test_distributed.py
```diff
@@ -25,28 +28,54 @@
...
     ENGINES.append('h5netcdf')
 
+NC_FORMATS = {'netcdf4': ['NETCDF3_CLASSIC', 'NETCDF3_64BIT_OFFSET',
+                          'NETCDF3_64BIT_DATA', 'NETCDF4_CLASSIC', 'NETCDF4'],
+              'scipy': ['NETCDF3_CLASSIC', 'NETCDF3_64BIT'],
+              'h5netcdf': ['NETCDF4']}
+TEST_FORMATS = ['NETCDF3_CLASSIC', 'NETCDF4_CLASSIC', 'NETCDF4']
+
+
+os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
```
jhamman: thanks to @jakirkham for uncovering this doozy (https://github.com/dask/dask/pull/3179#issuecomment-366813164).

jhamman:
The battle of inches continues. Turning off HDF5's file locking fixes all the tests for netCDF4 (🎉 ). Scipy is not working and `h5netcdf` doesn't support `autoclose` so it isn't expected to work.

@shoyer - I don't totally understand the scipy constraints on incremental writes but could that be playing a factor here?

rabernat:
Given the fact that distributed writes never actually worked, I would be happy at this point if they worked for *some* backends (e.g. netCDF4 and zarr) but not all (e.g. scipy, h5netcdf). We could document this and make sure to raise an appropriate `NotImplementedError` if the user tries a distributed write with a non-supported backend.

I make this suggestion because I know Joe has been slogging through this for a long time and it may be better to get it merged and live to fight another day. On the other hand, if Joe is optimistic about resolving the remaining backends, then by all means carry on!

shoyer:
>  I don't totally understand the scipy constraints on incremental writes but could that be playing a factor here?

I'm pretty sure SciPy supports incremental reads but not incremental writes. In general the entire netCDF file needs to get written at once. Certainly it's not possible to update only part of an array -- scipy needs it in memory as a NumPy array to copy its raw data to the netCDF file.

jbusecke:
I would echo @rabernat in the sense that if distributed writes work for some backends, I would love to see it merged as soon as possible. Thanks for working on this, I am very excited about this feature.

jhamman:
Thanks all for the comments. I will clean this up a bit and request a full review later this week. 

A few things to note:

1. I have not tested `save_mfdataset` yet. In theory, it should work now but it will require some testing. I'll save that for another PR.
2. I will raise an informative error message when either h5netcdf or scipy are used to write files along with distributed.

xarray/tests/test_distributed.py
shoyer: If this turns up in our test suite, my guess is that this will also manifest itself to users. So it is probably also worth mentioning in the user facing documentation.

It might also be worth checking if this resolves the test failure with h5netcdf and autoclose (no worries if it doesn't).

jakirkham: Glad to help. :)

Out of curiosity, where were you running into this issue (e.g. OS, file system, etc.)?

jhamman: @jakirkham - I'm also on OSX with a standard apple fusion file system.

xarray/backends/common.py
```diff
@@ -373,27 +456,41 @@ class DataStorePickleMixin(object):
...
         """
         Helper function to make sure datasets are closed and opened
         at appropriate times to avoid too many open file errors.
 
         Use requires `autoclose=True` argument to `open_mfdataset`.
         """
-        if self._autoclose and not self._isopen:
+
+        if autoclose is None:
+            autoclose = self._autoclose
```
jhamman: This could probably use some additional thinking.

jhamman:
I've added some additional tests and cleaned up the implementation a bit. I'd like to get reviews from a few folks and hopefully get this merged later this week.

xarray/backends/api.py
```diff
@@ -129,6 +130,32 @@ def _protect_dataset_variables_inplace(dataset, cache):
...
+    except TypeError:
+        # threading/multiprocessing lock
+        locks.append(SchedulerLock())
+
+    # When we have more than one lock, use the CombinedLock wrapper class
+    lock = CombinedLock(locks) if len(locks) > 1 else locks[0]
+
+    # Question: Should we be dropping one of these two locks when they are they
+    # are basically the same. For instance, when using netcdf4 and dask is not
+    # installed, locks will be [threading.Lock(), threading.Lock()]
```
shoyer: I think this is harmless, as long as these are different lock instances. 

On the other hand, something like `CombinedLock([lock, lock])` would never be satisfied because it's impossible to unlock a lock twice.

xarray/backends/api.py
```diff
@@ -129,6 +130,32 @@ def _protect_dataset_variables_inplace(dataset, cache):
...
+    locks = []
+    SchedulerLock = get_scheduler_lock(scheduler)
+    if format in ['NETCDF4', None] and engine in ['h5netcdf', 'netcdf4']:
+        locks.append(HDF5_LOCK)
+
+    try:
+        # per file lock
+        # Dask locks take a name argument (e.g. filename)
+        locks.append(SchedulerLock(path_or_file))
+    except TypeError:
```
shoyer: It would be less error prone to pass the name to `get_scheduler_lock` and have it return a lock instance.

xarray/backends/common.py
```diff
@@ -77,6 +126,39 @@ def robust_getitem(array, key, catch=Exception, max_retries=6,
...
+    def __exit__(self, *args):
+        for lock in self.locks:
+            lock.__exit__(*args)
+
+    @property
+    def locked(self):
+        return any(lock.locked for lock in self.locks)
+
+    def __repr__(self):
+        return "CombinedLock(%s)" % [repr(lock) for lock in self.locks]
```
shoyer: Nit: I think you could equivalently substitute `"CombinedLock(%r)" % list(self.locks)` here.

xarray/tests/test_distributed.py
```diff
@@ -34,29 +40,96 @@
...
 
+NC_FORMATS = {'netcdf4': ['NETCDF3_CLASSIC', 'NETCDF3_64BIT_OFFSET',
+                          'NETCDF3_64BIT_DATA', 'NETCDF4_CLASSIC', 'NETCDF4'],
+              'scipy': ['NETCDF3_CLASSIC', 'NETCDF3_64BIT'],
+              'h5netcdf': ['NETCDF4']}
+TEST_FORMATS = ['NETCDF3_CLASSIC', 'NETCDF4_CLASSIC', 'NETCDF4']
+
+
+# Does this belong elsewhere?
+os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
```
shoyer: It would be better to use a context manager or decorator on the test, something along the lines of https://stackoverflow.com/questions/2059482/python-temporarily-modify-the-current-processs-environment

jhamman: good idea. I think we can actually do this with pytest/monkeypatch.

doc/whats-new.rst
```diff
@@ -38,6 +38,13 @@ Documentation
 Enhancements
 ~~~~~~~~~~~~
 
+- Support for writing netCDF files from xarray datastores (scipy and netcdf4 only)
```
rabernat: Should this be "*to* xarray datastores"?

xarray/backends/zarr.py
```diff
@@ -356,6 +356,8 @@ def prepare_variable(self, name, variable, check_encoding=False,
 
         fill_value = _ensure_valid_fill_value(attrs.pop('_FillValue', None),
                                               dtype)
+        if variable.encoding == {'_FillValue': None} and fill_value is None:
+            variable.encoding = {}
```
rabernat: Does this fix a specific issue?

doc/whats-new.rst
rabernat: Nevermind, I think it makes sense as is.

Maybe "Support for writing xarray datasets to netCDF files..."

xarray/backends/zarr.py
jhamman: IIRC, this crept in from #1869.

doc/whats-new.rst
shoyer: It's nice to see you were able to get this to work with SciPy!

xarray/backends/zarr.py
rabernat: Could this fix #1955?

xarray/tests/test_distributed.py
```diff
@@ -1,23 +1,29 @@
...
+import pickle
+import tempfile
 
 import pytest
 
 dask = pytest.importorskip('dask')  # isort:skip
 distributed = pytest.importorskip('distributed')  # isort:skip
 
 from dask import array
+from dask.distributed import Client, Lock
```
jhamman: @mrocklin - would you mind looking at the test implementation I have here and let us know if you see anything that would be causing the default (global) dask scheduler to be permanently overridden. In #1971, I pointed to some test failures that appear coming from a unexpected scheduler.

mrocklin: Everything seems fine to me.  What is `dask._config.globals['get']`?  Is it the `get` method of a client?  To debug you might consider giving each of your clients a name

    with Client(s['address'], name='test-foo') as client

and then seeing which one isn't getting cleaned up?  You can also try `distributed.client.default_client()`.  We clean things up in the `__exit__()` call though, so as long as you're using context managers or `@gen_cluster` everything *should* work fine.

xarray/backends/common.py
```diff
@@ -77,6 +126,39 @@ def robust_getitem(array, key, catch=Exception, max_retries=6,
...
 
+class CombinedLock(object):
+    """A combination of multiple locks.
+
+    Like a locked door, a CombinedLock is locked if any of its constituent
+    locks are locked.
+    """
+
+    def __init__(self, locks):
+        self.locks = tuple(set(locks))  # remove duplicates
```
jhamman: previous test failures were having trouble in `__enter__` when iterating over a `set` of locks. casting to list/tuple seems to have resolved that.

jhamman:
Any final comments on this? If not, I'll probably merge this in the next day or two.

xarray/backends/netCDF4_.py
```diff
@@ -399,7 +404,9 @@ def prepare_variable(self, name, variable, check_encoding=False,
             # OrderedDict as the input to setncatts
             nc4_var.setncattr(k, v)
 
-        return nc4_var, variable.data
+        target = NetCDF4ArrayWrapper(name, self)
+
+        return target, variable.data
```
fujiisoup: @jhamman 
This is too late, but I think `nc4_var` is never used. Is it correct?

jhamman: @fujiisoup - it is used in line 405 (`nc4_var.setncattr(k, v)`).

PR #1794, opened 2017-12-20
DOC: added sortby example

raybellwaves:
 - [NA] Passes ``git diff upstream/master **/*py | flake8 --diff`` (remove if you did not edit any Python files)


shoyer:
Thanks!

PR #1796, opened 2017-12-20
Support RGB[A] arrays in plot.imshow()

Zac-HD:

This patch brings `xarray.plot.imshow` up to parity with `matplotlib.pyplot.imshow`:

- As well as 2D images (greyscale / luminance, using a colormap), it now supports a third dimension for RGB or RGBA channels.  For consistency with 2D arrays, missing data is plotted as transparent pixels
- Being Xarray, users need not care about the order of their dimensions - we infer the right one for color, and warn if it's ambiguous.
- ~~Using `robust=True` for easy saturation is really nice.  Having it adjust each channel and facet in the same way is essential for this to work, which it does.~~
- ~~Matplotlib wraps out-of-range colors, leading to crazy maps and serious interpretation problems if it's only a small region.  Xarray clips (ie saturates) to the valid range instead.~~

*I'm going to implement clip-to-range and color normalization upstream in matplotlib, then open a second PR here so that Xarray can use the same interface.*

And that's the commit log!  It's not really a big feature, but each of the parts can be fiddly so I've broken
 the commits up logically 😄 

Finally, a motivating example: visible-light Landsat data before, during (top-right), and after a fire at Sampson's Flat, Australia:

    arr = ds['red green blue'.split()].to_array(dim='band') / (2 ** 12)
    arr.plot.imshow(col='time', col_wrap=5, robust=True)

image

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -443,6 +443,10 @@ def _plot2d(plotfunc):
         # Decide on a default for the colorbar before facetgrids
         if add_colorbar is None:
             add_colorbar = plotfunc.__name__ != 'contour'
+        if (plotfunc.__name__ == 'imshow' and
+            darray.ndim == (3 + (row is not None) + (col is not None))):
+            # Don't add a colorbar when showing an image with explicit colors
+            add_colorbar = False
 
         # Handle facetgrids first
         if row or col:
@@ -470,7 +474,8 @@ def _plot2d(plotfunc):
                           "Use colors keyword instead.",
                           DeprecationWarning, stacklevel=3)
 
-        xlab, ylab = _infer_xy_labels(darray=darray, x=x, y=y)
+        xlab, ylab = _infer_xy_labels(darray=darray, x=x, y=y,
+                                      imshow=plotfunc.__name__ == 'imshow')
 
         # better to pass the ndarrays directly to plotting functions
         xval = darray[xlab].values
@@ -595,6 +600,10 @@ def imshow(x, y, z, ax, **kwargs):
 
     Wraps :func:`matplotlib:matplotlib.pyplot.imshow`
 
+    While other plot methods require the DataArray to be strictly
+    two-dimensional, ``imshow`` also accepts a 3D array where the third
+    dimension can be interpreted as RGB or RGBA color channels.
+
     .. note::
         This function needs uniformly spaced coordinates to
         properly label the axes. Call DataArray.plot() to check.
@@ -631,6 +640,15 @@ def imshow(x, y, z, ax, **kwargs):
 
     # Allow user to override these defaults
     defaults.update(kwargs)
+
+    if z.ndim == 3:
+        # matplotlib imshow uses black for missing data, but Xarray makes
+        # missing data transparent.  We therefore add an alpha channel if
+        # there isn't one, and set it to transparent where data is masked.
+        if z.shape[-1] == 3:
+            z = np.ma.concatenate((z, np.ma.ones(z.shape[:2] + (1,))), 2)
+        z = z.copy()
+        z[np.any(z.mask, axis=-1), -1] = 0
 
     primitive = ax.imshow(z, **defaults)
 
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -443,14 +443,17 @@ def _plot2d(plotfunc):
         # Decide on a default for the colorbar before facetgrids
         if add_colorbar is None:
             add_colorbar = plotfunc.__name__ != 'contour'
-        if (plotfunc.__name__ == 'imshow' and
-            darray.ndim == (3 + (row is not None) + (col is not None))):
+        imshow_rgb = (
+            plotfunc.__name__ == 'imshow' and
+            darray.ndim == (3 + (row is not None) + (col is not None)))
+        if imshow_rgb:
             # Don't add a colorbar when showing an image with explicit colors
             add_colorbar = False
 
         # Handle facetgrids first
         if row or col:
             allargs = locals().copy()
+            allargs.pop('imshow_rgb')
             allargs.update(allargs.pop('kwargs'))
 
             # Need the decorated plotting function
@@ -474,13 +477,19 @@ def _plot2d(plotfunc):
                           "Use colors keyword instead.",
                           DeprecationWarning, stacklevel=3)
 
-        xlab, ylab = _infer_xy_labels(darray=darray, x=x, y=y,
-                                      imshow=plotfunc.__name__ == 'imshow')
+        rgb = kwargs.pop('rgb', None)
+        xlab, ylab = _infer_xy_labels(
+            darray=darray, x=x, y=y, imshow=imshow_rgb, rgb=rgb)
+
+        if rgb is not None and plotfunc.__name__ != 'imshow':
+            raise ValueError('The "rgb" keyword is only valid for imshow()')
+        elif rgb is not None and not imshow_rgb:
+            raise ValueError('The "rgb" keyword is only valid for imshow()'
+                             'with a three-dimensional array (per facet)')
 
         # better to pass the ndarrays directly to plotting functions
         xval = darray[xlab].values
         yval = darray[ylab].values
-        zval = darray.to_masked_array(copy=False)
 
         # check if we need to broadcast one dimension
         if xval.ndim < yval.ndim:
@@ -491,8 +500,19 @@ def _plot2d(plotfunc):
 
         # May need to transpose for correct x, y labels
         # xlab may be the name of a coord, we have to check for dim names
-        if darray[xlab].dims[-1] == darray.dims[0]:
-            zval = zval.T
+        if imshow_rgb:
+            # For RGB[A] images, matplotlib requires the color dimension
+            # to be last.  In Xarray the order should be unimportant, so
+            # we transpose to (y, x, color) to make this work.
+            yx_dims = (ylab, xlab)
+            dims = yx_dims + tuple(d for d in darray.dims if d not in yx_dims)
+            if dims != darray.dims:
+                darray = darray.transpose(*dims)
+        elif darray[xlab].dims[-1] == darray.dims[0]:
+            darray = darray.transpose()
+
+        # Pass the data as a masked ndarray too
+        zval = darray.to_masked_array(copy=False)
 
         _ensure_plottable(xval, yval)
 
@@ -601,8 +621,9 @@ def imshow(x, y, z, ax, **kwargs):
     Wraps :func:`matplotlib:matplotlib.pyplot.imshow`
 
     While other plot methods require the DataArray to be strictly
-    two-dimensional, ``imshow`` also accepts a 3D array where the third
-    dimension can be interpreted as RGB or RGBA color channels.
+    two-dimensional, ``imshow`` also accepts a 3D array where some
+    dimension can be interpreted as RGB or RGBA color channels and
+    allows this dimension to be specified via the kwarg ``rgb=``.
 
     .. note::
         This function needs uniformly spaced coordinates to
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -258,17 +258,21 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
                 levels=levels, norm=norm)
 
 
-def _infer_xy_labels(darray, x, y):
+def _infer_xy_labels(darray, x, y, imshow=False):
     """
     Determine x and y labels. For use in _plot2d
 
-    darray must be a 2 dimensional data array.
+    darray must be a 2 dimensional data array, or 3d for imshow only.
     """
 
     if x is None and y is None:
         if darray.ndim != 2:
-            raise ValueError('DataArray must be 2d')
-        y, x = darray.dims
+            if not imshow:
+                raise ValueError('DataArray must be 2d')
+            elif darray.ndim != 3 or darray.shape[2] not in (3, 4):
+                raise ValueError('DataArray for imshow must be 2d, MxNx3 for '
+                                 'RGB image, or MxNx4 for RGBA image.')
+        y, x, *_ = darray.dims
     elif x is None:
         if y not in darray.dims:
             raise ValueError('y must be a dimension name if x is not supplied')
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -258,21 +258,70 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
                 levels=levels, norm=norm)
 
 
-def _infer_xy_labels(darray, x, y, imshow=False):
+def _infer_xy_labels_3d(darray, x, y, rgb):
+    """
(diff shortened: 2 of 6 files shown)


xarray/tests/test_plot.py
```diff
@@ -1024,6 +1026,34 @@ def test_2d_coord_names(self):
         with raises_regex(ValueError, 'requires 1D coordinates'):
             self.plotmethod(x='x2d', y='y2d')
 
+    def test_plot_rgb_image(self):
+        DataArray(
+            easy_array((10, 15, 3), start=0),
+            dims=['y', 'x', 'band'],
+        ).plot.imshow()
```
rabernat: Is it possible to check that the colors have actually rendered correctly by introspecting deeper into the figures / axes that are generated by `imshow`?

shoyer:
I like this. My main concern is that I'd like a fully explicit way to opt into this behavior. Perhaps `rgb='band'` and/or `rgba='band'`?

xarray/tests/test_plot.py
Zac-HD: Well, it's almost certainly *possible* - but I have absolutely no idea where to start!

Zac-HD:
> I like this. My main concern is that I'd like a fully explicit way to opt into this behavior. Perhaps rgb='band' and/or rgba='band'?

I have a pretty strong preference for RGB images 'just working' - *requiring* a new argument would mean imshow has a different signature to all other plot methods without adding any information (at worst you can supply `x` and `y` dims, so IMO it's not needed).

In many ways showing an RGB image is a special case, but this is the least-special-case I could make work - and it's consistent with the obvious meaning and upstream behavior of `imshow`.

shoyer:
I should clarify: I don't think we should *require* the explicit name for the RGB dimension, but I think we should have an argument around so that's an option if desired. This is nice because it gives better error messages when things go wrong and makes it easier to write self-documented code. This would be consistent with the rest of our visualization functions, which provide an option for being fully explicit but don't require it.

Zac-HD:
Optional argument coming up then 😄 

Anything else I need to do?

Zac-HD:
Ugh, it looks like this broke saturation handling somewhere in the rebase.  I'll take another look later.

Zac-HD:
Done - now with optional argument and nicer commit history,  Anything else?

xarray/plot/plot.py
```diff
@@ -419,10 +419,32 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+            add_colorbar = False
+            # Manually stretch colors for robust cmap
+            if robust:
+                flat = darray.values.ravel(order='K')
+                flat = flat[~np.isnan(flat)]
+                if flat.size == 0:
+                    # All data will be masked, so skip percentile calculation
+                    vmin, vmax = 0, 1
+                if vmin is None:
+                    vmin = np.percentile(flat, ROBUST_PERCENTILE)
```
shoyer: I'm not quite sure that this handling of the robust argument makes sense. Does this resemble what vmin/vmax do by default on imshow?

Parts of this that seem strange to me:
- `robust=True` does more than merely clipping the data to the robust range, unlike the behavior with colormaps. The data gets extra normalization, even if it all already fell within the central 95%.
- all channels are treated together, rather than independently, including the alpha channel!

I'm sure this behavior is useful sometimes, but maybe it's better to leave it out for now. We could make robust=True an error for now, and potentially save color normalization for another keyword argument.

xarray/plot/plot.py
```diff
@@ -446,12 +468,13 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
                           "Use colors keyword instead.",
                           DeprecationWarning, stacklevel=3)
```
shoyer: Can you add `rgb` as an actual argument inline, after x and y?

xarray/plot/utils.py
```diff
@@ -253,12 +254,67 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
...
+
+def _infer_xy_labels(darray, x, y, imshow=False, rgb=None):
     """
     Determine x and y labels. For use in _plot2d
 
-    darray must be a 2 dimensional data array.
+    darray must be a 2 dimensional data array, or 3d for imshow only.
     """
+    if imshow and darray.ndim == 3:
+        return _infer_xy_labels_3d(darray, x, y, rgb)
```
shoyer: Please make sure this raises an informative error message if RGB is provided and the input has the wrong number of dimensions or isn't being plotted with imshow.

xarray/plot/utils.py
```diff
@@ -253,12 +254,67 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
                 levels=levels, norm=norm)
 
 
-def _infer_xy_labels(darray, x, y):
+def _infer_xy_labels_3d(darray, x, y, rgb):
+    """
+    Determine x and y labels for showing RGB images.
+
+    Attempts to infer which dimension is RGB/RGBA by size and order of dims.
```
shoyer: Rather than all this complex logic with warnings in ambiguous cases, why not always treat the last and/or remaining (after explicit x/y labels) dimension as RGB? I think that solves the convenience use cases, without hard to understand/predict inference logic.

xarray/tests/test_plot.py
```diff
@@ -1024,6 +1026,35 @@ def test_2d_coord_names(self):
...
+
+    def test_warns_ambigious_dim(self):
+        arr = DataArray(easy_array((3, 3, 3)), dims=['y', 'x', 'band'])
+        with pytest.warns(UserWarning):
+            arr.plot.imshow()
+        # but doesn't warn if dimensions specified
+        arr.plot.imshow(rgb='band')
+        arr.plot.imshow(x='x', y='y')
+
```
shoyer: Please add tests for errors related to `rgb`.

xarray/tests/test_plot.py
shoyer: I'm okay relying on matplotlib to plot colors properly. One sane option would be to mock `plt.imshow` to ensure it gets called properly, but we don't bother with that for our current tests. But that shouldn't get in the way of integration tests that verify that we can actually call matplotlib without any errors being raised.

xarray/plot/plot.py
```diff
@@ -567,6 +601,16 @@ def imshow(x, y, z, ax, **kwargs):
...
     Wraps :func:`matplotlib:matplotlib.pyplot.imshow`
 
+    While other plot methods require the DataArray to be strictly
+    two-dimensional, ``imshow`` also accepts a 3D array where some
+    dimension can be interpreted as RGB or RGBA color channels and
+    allows this dimension to be specified via the kwarg ``rgb=``.
+    In this case, ``robust=True`` will saturate the image in the
+    usual way, consistenly between all bands and facets.
+
+    This method will clip oversaturated pixels to the valid range,
```
shoyer: Is this actually strictly better? If so, why didn't it get changed during matplotlib's recent style overhaup? Was it simply overlooked or a more careful decision?

xarray/plot/plot.py
Zac-HD: I *could*, but this would be a breaking change for anyone using positional arguments.  I'd therefore prefer not to, as it's just as easy to do the breaking bit later.

xarray/plot/plot.py
Zac-HD: This logic makes `robust=True` work exactly the same way on single-band and RGB greyscale images.  The handling is only distinct because the 'normalise to [0, 1]' step is implicit for single bands but must be explicit for RGB.

It does make sense to normalise RGB without considering or changing the alpha channel though; I'll play with this a little more.

As to usefulness, it's basically essential to sensibly visualise surface reflectance data (ie high-quality satellite images) - they're represented for analysis as albedo, ie fraction of light reflected, and if you display that directly it's a very dark and muddy image (example below; same scene as above).  So I'd strongly prefer to keep this in!

download

xarray/plot/plot.py
```diff
@@ -419,10 +419,32 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+                    vmin, vmax = 0, 1
+                if vmin is None:
+                    vmin = np.percentile(flat, ROBUST_PERCENTILE)
+                if vmax is None:
+                    vmax = np.percentile(flat, 100 - ROBUST_PERCENTILE)
+                darray = (darray - vmin) / (vmax - vmin)
+                robust = False
+                del flat
+            # Clip range to [0, 1] to avoid visual artefacts
+            darray.values[:] = np.clip(darray.values, 0, 1)
```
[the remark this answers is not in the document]
Zac-HD: Oh dear, I can't!  Will fix this.

xarray/plot/utils.py
Zac-HD: Because `some_dataset[['red', 'green', 'blue']].to_array().plot.imshow()` is *really tempting*.  The only difference is that if the last dimension is too large, we try the first dimension before resuming 'last dim is color' behaviour.

On reflection I think you're right though - so long as we can do the right thing when only one dim is the right size, a simpler heuristic for order preference makes good sense.  Will change.

xarray/plot/plot.py
Zac-HD: IMO it is strictly better, because it avoids the ambiguity a colour could actually represent *any value at all* modulo the range.

I believe it was overlooked because relatively few people start with crudely-scaled RGB images, though I wasn't involved closely enough to know for sure (or fix it then!).  Unfortunately such "scaling" is pretty common to go from albedo to useful dynamic range in imagery, as I mentioned about `robust` above.

Zac-HD:
Thanks for the review @shoyer!  I'll do what I can in the next few days, but that might not be much at all before I get back from a no-internet camping trip around Jan 8th.  Items:

- improve `robust` with an alpha channel (exclude from scaling but still clip to valid)
- add validation and tests for `rgb` argument
- don't mutate the input data!  (I *think* it doesn't, but obviously this needs to be checked)
- simplify order heuristic, to "use last dimension of size 3 or 4 as rgb; warn if there are multiple possible"

xarray/plot/plot.py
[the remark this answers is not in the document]
Zac-HD: Yep, I've tossed in a `.copy(deep=True)` to guarantee that there's no mutation.

Zac-HD:
OK, @shoyer & @fmaussion - I think I'm done again!  

The only thing I *haven't* done is change the handling of `robust=True`, as anything beyond "scale and clip 2nd-to-98th percentiles to [0, 1]" seems fraught with complications.  If this is not what the user intended, it's at least the obvious default behaviour and should be simple to understand.

Hopefully this covers the substantial changes; but let me know if there's anything else as I'd really like this merged before the 0.10.1 release - and ideally released in time to for demos at my summer school in January 😄

xarray/tests/test_plot.py
Zac-HD: Done.

xarray/plot/plot.py
shoyer: This still feels strange to me. With colormaps, all `robust=True` does is explicitly set vmin and vmax based on the data. But vmin and vmax are entirely ignored here.

What if we interpreted vmin/vmax with rgb plots as setting the bounds for normalization? The defaults would be (0, 1) for floats and (smallest int, largest int) for int dtypes. Then, robust=True could again simply set vmin/vmax based on the data.

Zac-HD: I actually agree that the current behavior is strange and that your suggestion is an improvement (though I'd use [0, 255] for all ints) - but it's consistent with matplotlib!  My proposed plan has some extra steps though:

1. Merge this pull.  We will need the stretching machinery regardless, unless Xarray drops support for matplotlib<2.2
2. Patch matplotlib/matplotlib#9391 upstream.  (I'm already working on a patch for this)
3. Implement the new matplotlib interface in Xarray.  I *could* do that in this pull, but then if the matplotlib implementation doesn't match we'd have to break API compatibility - I'd really prefer to have a smaller API now and get it right the first time, later.

Zac-HD:
More review fixes 😄 

I'm pretty sure the build failure is just Travis being flaky - it passes on my machine.  Could someone restart it?

fmaussion:
Just restarted them, let's see how it goes

Zac-HD:
Okay... I've now seen three Travis runs.  In every one there's been a pass, a fail, and three errors... but different jobs each time.  At this point I'm ready to give up and wait for the Travis team to fix it 😕

The code is ready to go though! 🎉

xarray/plot/plot.py
shoyer: > I could do that in this pull, but then if the matplotlib implementation doesn't match we'd have to break API compatibility - I'd really prefer to have a smaller API now and get it right the first time, later.

I agree. But I'm also concerned about the behavior for `robust=True` potentially needing to change to match upstream.

My preference would be to leave out support for `robust` with RGB plots until the matplotlib behavior lands. You could write a small helper function for RGB normalization (maybe even it include it in the docs as an example), e.g.,
```python
import xarray.ufuncs as xu

def robust_stretch(data):
  vmin, vmax = data.quantile(q=[0.02, 0.98])
  bounded = xu.minimum(xu.maximum(data, vmin), vmax)
  returned (bounded - vmin) / (vmax - vmin)
```
Then you could plot things like `arr.pipe(robust_stretch).plot.imshow(col='time', col_wrap=5)`.

It's a few more characters but not a terrible work around.

Zac-HD:
@shoyer makes some good points about things that should be fixed upstream in Matplotlib - namely normalization of RGB images, but I'm also going to move my color fixes (clipping instead of modulo) upstream instead.  Timeline:

- I edit this PR, it gets merged, Xarray 10.1 (ETA soonish?) can display RGB images just like Matplotlib
- I finish my upstream PR, Matplotlib 2.2 (ETA ~Feb sometime) supports normalization of RGB images and fixes the modulo/clip bug
- Once the matplotlib API is merged, I make a new PR against Xarray to support the same interface (including on older matplotlib versions)

shoyer:
@Zac-HD  Sounds good. At this point I don't think there are any blockers left for the 0.10.1 release (I wanted to get the pydap fix in, which we recently merged).

Zac-HD:
`test_datetime_line_plot` failed on Travis, but I'm pretty sure that's nothing to do with me 😕 

And otherwise, this pull is ready to merge 🎉   
<sup>...this time for sure!</sup>

Zac-HD:
Looks like `TestDatetimePlot.test_datetime_line_plot` is failing, but I'm pretty sure that's nothing to do with me (and it passed twice on Travis!).

shoyer:
Should be fixed by https://github.com/pydata/xarray/pull/1814 which I just merged

Zac-HD:
matplotlib/matplotlib#10220 is the upstream pull request 🎉 

I still like the plan of merging this without `robust` support, but if you can then hold off on an Xarray release for a few days there's a good chance that steps two and three will also be done this week - matplotlib 2.1.2 is scheduled for 12 or 13 Jan and issue is marked "release critical", so I'm hoping for a fast turnaround 😄

shoyer:
Thanks @Zac-HD !

PR #1799, opened 2017-12-21
move backend append logic to the prepare_variable methods

### LINKED ISSUE #1798: BUG: set_variables in backends.commons loads target dataset
#### Problem description

In #1609 we (I) implemented a fix for appending to datasets with existing variables. In doing so, it looks like I added a regression wherein the `variables` property on the `AbstractWritableDataStore` is repeatedly queried. This property calls `.load()` on the underlying dataset.

This was discovered while diagnosing some problems with the zarr backend (#1770, https://github.com/pangeo-data/pangeo/issues/48#issuecomment-353223737). 

I have a potential fix for this that I will post once the tests pass.

cc @rabernat, @mrocklin

<details>
INSTALLED VERSIONS
------------------
commit: 20f957db105a9348b0f7d2dac076c17c31cbccee
python: 3.6.0.final.0
python-bits: 64
OS: Darwin
OS-release: 17.3.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0+dev18.g4a9c1e3
pandas: 0.21.0
numpy: 1.13.3
scipy: 0.19.1
netCDF4: 1.3.0
h5netcdf: 0.5.0
Nio: None
zarr: 2.1.4
bottleneck: 1.2.1
cyordereddict: None
dask: 0.15.4
distributed: 1.19.3
matplotlib: 2.0.2
cartopy: 0.15.1
seaborn: 0.8.1
setuptools: 33.1.0.post20170122
pip: 9.0.1
conda: None
pytest: 3.2.3
IPython: 5.2.2
sphinx: 1.6.3
</details>

jhamman:
- [x] Closes #1798
 - [ ] Tests added (ideas for how to test that load is not called? Regression tests from #1609 are passing)
 - [x] Tests passed

### DIFF
--- a/xarray/backends/common.py
+++ b/xarray/backends/common.py
@@ -7,7 +7,6 @@ import time
 import traceback
 import contextlib
 from collections import Mapping
-from distutils.version import LooseVersion
 
 from ..conventions import cf_encoder
 from ..core import indexing
@@ -183,11 +182,7 @@ class ArrayWriter(object):
     def sync(self):
         if self.sources:
             import dask.array as da
-            import dask
-            if LooseVersion(dask.__version__) > LooseVersion('0.8.1'):
-                da.store(self.sources, self.targets, lock=self.lock)
-            else:
-                da.store(self.sources, self.targets)
+            da.store(self.sources, self.targets, lock=self.lock)
             self.sources = []
             self.targets = []
 
@@ -232,11 +227,8 @@ class AbstractWritableDataStore(AbstractDataStore):
         for vn, v in iteritems(variables):
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
-            if vn not in self.variables:
-                target, source = self.prepare_variable(
-                    name, v, check, unlimited_dims=unlimited_dims)
-            else:
-                target, source = self.ds.variables[name], v.data
+            target, source = self.prepare_variable(
+                name, v, check, unlimited_dims=unlimited_dims)
 
             self.writer.add(source, target)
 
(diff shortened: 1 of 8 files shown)


xarray/backends/common.py
```diff
@@ -183,11 +182,7 @@ def add(self, source, target):
     def sync(self):
         if self.sources:
             import dask.array as da
-            import dask
-            if LooseVersion(dask.__version__) > LooseVersion('0.8.1'):
-                da.store(self.sources, self.targets, lock=self.lock)
-            else:
-                da.store(self.sources, self.targets)
```
jhamman: we now have a minimum dask version of 0.9 so this check is no longer needed.

xarray/backends/common.py
```diff
@@ -232,11 +227,8 @@ def set_variables(self, variables, check_encoding_set,
         for vn, v in iteritems(variables):
             name = _encode_variable_name(vn)
             check = vn in check_encoding_set
-            if vn not in self.variables:
```
shoyer: Can we deprecate the variables property as part of this change? That would help avoid this in the future. In general it's a bad idea for adapter layers (like our backend classes) to hold state.

jhamman: @shoyer - done. I actually removed the the `attrs` and `dimensions` properties too since they weren't really used. I left a deprecation error in there for now but can remove it assuming all the tests pass.

shoyer:
I'm glad to see we weren't using this properties very much internally :). Let's keep them as FutureWarning for now since there is a decent chance some third party backend subclass is using them, and there's no reason why we need to make things more difficult than necessary.

jhamman:
@shoyer - this should be good to go following my last two commits. I'm not sure how to write a regression test for this but if you think its needed, I can try to whip something up.

PR #1800, opened 2017-12-26
WIP: Performance improvements for zarr backend

jhamman:
- [x] Closes #https://github.com/pangeo-data/pangeo/issues/48

This is building on top of #1799. Based on the suggestion from @alimanfoo in https://github.com/pangeo-data/pangeo/issues/48#issuecomment-353807691, I have reworked the handling of attributes in the zarr backend. There is more to do here, particularly in the `set_dimensions` arena but this is giving almost a 2x speedup in writing to GCP.

cc @rabernat, @mrocklin and @alimanfoo


xarray/backends/zarr.py
```diff
@@ -331,29 +331,45 @@ def get_variables(self):
...
+        except KeyError:
+            raise KeyError("Zarr object is missing the attribute `%s`, which "
+                           "is required for xarray to determine variable "
+                           "dimensions." % (_DIMENSION_KEY))
         return dimensions
 
-    def set_dimension(self, name, length, is_unlimited=False):
-        if is_unlimited:
+    # TODO: we need these checks one way or another
+    # def set_dimension(self, name, length, is_unlimited=False):
```
shoyer: It seems that we should probably rewrite our backend logic to write/check dimensions once rather than doing it each time a Variable is written.

jhamman: This seems like a good idea. I've gone ahead and done this and will push some more changes later tonight.

jhamman:
@shoyer - I've added a `set_dimensions` method to the writable datastores. This seems to work quite well except for string variables. I have a hacky fix with a comment in there right now, this works for all backends except h5netcdf. Thoughts on a better way to encode variables prior to setting dimensions would be great.

xarray/backends/common.py
```diff
@@ -234,12 +238,22 @@ def set_variables(self, variables, check_encoding_set,
...
+
+        dims = {}
+        for v in variables.values():
+            dims.update(dict(zip(v.dims, v.shape)))
+
+        for d, l in dims.items():
+
+            if d in existing_dims and l != existing_dims[d]:
+                raise ValueError("Unable to update size for existing dimension"
+                                 "%r (%d != %d)" % (d, l, existing_dims[d]))
```
rabernat: I don't think we have any tests for this sort of error.

xarray/backends/zarr.py
```diff
@@ -331,29 +331,42 @@ def get_variables(self):
                                  for k, v in self.ds.arrays())
 
     def get_attrs(self):
-        _, attributes = _get_zarr_dims_and_attrs(self.ds, _DIMENSION_KEY)
+        attributes = HiddenKeyDict(self.ds.attrs, [_DIMENSION_KEY])
```
rabernat: 👍

xarray/backends/zarr.py
```diff
@@ -331,29 +331,42 @@ def get_variables(self):
...
+
+        dims = {}
+        for v in variables.values:
+            dims.update(dict(zip(v.dims, v.shape)))
+
+        update_dims = {}
+        for d, l in dims.items():
+            if d in existing_dims and l != existing_dims[d]:
+                raise ValueError("Unable to update size for existing dimension"
+                                 "%r (%d != %d)" % (d, l, existing_dims[d]))
```
rabernat: I don't think we have any tests for this sort of error.

xarray/backends/zarr.py
```diff
@@ -363,41 +376,30 @@ def prepare_variable(self, name, variable, check_encoding=False,
...
-        # filters=None, cache_metadata=True, **kwargs)
+        encoded_attrs = OrderedDict()
+        # the magic for storing the hidden dimension data
+        encoded_attrs[_DIMENSION_KEY] = dims
+        for k, v in iteritems(attrs):
+            encoded_attrs[k] = _encode_zarr_attr_value(v)
+
         if name in self.ds:
             zarr_array = self.ds[name]
+            zarr_array.attrs.update(encoded_attrs)
```
rabernat: I'm trying to understand when we would ever hit this case. It seems like it would only happen if we were trying to modify an existing store in place. That is not currently supported.

xarray/backends/common.py
```diff
@@ -216,7 +216,11 @@ def store_dataset(self, dataset):
 
     def store(self, variables, attributes, check_encoding_set=frozenset(),
               unlimited_dims=None):
+        # This seems out of place
+        variables = OrderedDict([(k, maybe_encode_as_char_array(v))
+                                 for k, v in variables.items()])
```
jhamman: Should each backend define encode/decode methods? For string variables, we need to encode before setting the dimensions.

xarray/tests/test_backends.py
```diff
@@ -1288,6 +1313,35 @@ def test_roundtrip_string_encoded_characters(self):
...
+    def test_roundtrip_timedelta_data(self):
+        super(CFEncodedDataTest, self).test_roundtrip_timedelta_data()
+
+    @pytest.mark.xfail(reason="Zero-dimension variables are broken")
+    def test_roundtrip_datetime_data(self):
+        super(CFEncodedDataTest, self).test_roundtrip_datetime_data()
+
+    @pytest.mark.xfail(reason="Zero-dimension variables are broken")
+    def test_roundtrip_coordinates_with_space(self):
+        super(CFEncodedDataTest, self).test_roundtrip_coordinates_with_space()
```
jhamman: @rabernat - I have not been able to get zero-dim variables to work with zarr. I know you iterated on this but did you ever get it to work?

xref: https://github.com/zarr-developers/zarr/issues/150

xarray/tests/test_backends.py
```diff
@@ -1205,10 +1229,11 @@ def test_hidden_zarr_keys(self):
...
             # verify that the dataset fails to open if dimension key is missing
-            del zarr_group.attrs[self.DIMENSION_KEY]
-            with pytest.raises(KeyError):
-                with xr.decode_cf(store) as actual:
-                    pass
+            # this is not passing because the store is not read only
+            # del zarr_group.attrs[self.DIMENSION_KEY]
+            # with pytest.raises(KeyError):
+            #     with xr.decode_cf(store) as actual:
+            #         pass
```
jhamman: I still need to figure out why this wasn't raising an error.

xarray/backends/zarr.py
```diff
@@ -363,72 +363,27 @@ def prepare_variable(self, name, variable, check_encoding=False,
...
-
+        encoded_attrs = OrderedDict()
         # the magic for storing the hidden dimension data
-        zarr_array.attrs[_DIMENSION_KEY] = dims
-        _, attributes = _get_zarr_dims_and_attrs(zarr_array, _DIMENSION_KEY)
-
+        encoded_attrs[_DIMENSION_KEY] = dims
         for k, v in iteritems(attrs):
-            attributes[k] = _encode_zarr_attr_value(v)
+            encoded_attrs[k] = _encode_zarr_attr_value(v)
```
jhamman: can the attribute encoding be moved to the `encode_attributes`?

xarray/backends/zarr.py
```diff
@@ -331,29 +323,37 @@ def get_variables(self):
...
-                                 "encoded in Zarr attributes are incompatible "
-                                 "with newly specified dimension `%s`: %g" %
-                                 (self.ds.attrs[_DIMENSION_KEY], name, length))
-        self.ds.attrs[_DIMENSION_KEY][name] = length
-
-    def set_attribute(self, key, value):
-        _, attributes = _get_zarr_dims_and_attrs(self.ds, _DIMENSION_KEY)
-        attributes[key] = _encode_zarr_attr_value(value)
+
+        dims = {}
```
jhamman: make this an `OrderedDict`

xarray/tests/test_backends.py
rabernat: This issue was fixed in  zarr-developers/zarr#160. But zarr has not had a release since then. So you need to be installing zarr from github master in order to get zero-dim variables.

Zero dim variables are covered by the test suite
https://github.com/pydata/xarray/blob/master/xarray/tests/test_backends.py#L170
And you can see the test passing in the `zarr-dev` environment in my original zarr PR:
https://travis-ci.org/pydata/xarray/jobs/315571803#L1558

xarray/tests/test_backends.py
rabernat: Probably because the error checking occurs in `_get_zarr_dims_and_attrs`, but you have bypassed that function in many cases for performance optimizations.

What does the dataset look like if you open it after deleting `DIMENSION_KEY`? It should be broken in some way.

xarray/backends/zarr.py
```diff
@@ -297,9 +292,6 @@ def __init__(self, zarr_group, writer=None):
                 raise KeyError("Zarr group can't be read by xarray because "
                                "it is missing the `%s` attribute." %
                                _DIMENSION_KEY)
-            else:
-                # initialize hidden dimension attribute
-                self.ds.attrs[_DIMENSION_KEY] = {}
```
rabernat: I think the removal of these lines is relevant for the dimension key test error.

xarray/backends/zarr.py
```diff
@@ -331,29 +323,37 @@ def get_variables(self):
...
         return _decode_zarr_attrs(attributes)
 
     def get_dimensions(self):
-        dimensions, _ = _get_zarr_dims_and_attrs(self.ds, _DIMENSION_KEY)
+        try:
+            dimensions = self.ds.attrs[_DIMENSION_KEY].asdict()
+        except KeyError:
+            raise KeyError("Zarr object is missing the attribute `%s`, which "
+                           "is required for xarray to determine variable "
+                           "dimensions." % (_DIMENSION_KEY))
```
rabernat: This should be raising when you delete the dimension key.

xarray/tests/test_backends.py
jhamman: you're right. I thought I had Zarr-master in my environment. but I didn't These all work now.

xarray/tests/__init__.py
```diff
@@ -71,7 +71,7 @@ def _importorskip(modname, minversion=None):
 has_bottleneck, requires_bottleneck = _importorskip('bottleneck')
 has_rasterio, requires_rasterio = _importorskip('rasterio')
 has_pathlib, requires_pathlib = _importorskip('pathlib')
-has_zarr, requires_zarr = _importorskip('zarr', minversion='2.2.0')
+has_zarr, requires_zarr = _importorskip('zarr', minversion='2.1.0')
```
jhamman: change this back to 2.2.

xarray/backends/zarr.py
```diff
@@ -331,29 +313,34 @@ def get_variables(self):
...
-        dimensions, _ = _get_zarr_dims_and_attrs(self.ds, _DIMENSION_KEY)
+        dimensions = OrderedDict()
+        for k, v in self.ds.arrays():
+            try:
+                for d, s in zip(v.attrs[_DIMENSION_KEY], v.shape):
+                    dimensions[d] = s
+            except KeyError:
+                raise KeyError("Zarr object is missing the attribute `%s`, "
+                               "which is required for xarray to determine "
+                               "variable dimensions." % (_DIMENSION_KEY))
```
jhamman: @rabernat - I want to get your take on this change. What I've done is remove the `_DIMENSION_KEY` attribute from the dataset all together and I'm just pulling it from the individual arrays/variables. I realized we were storing this information multiple places and it was getting too tricky keeping everything straight.

rabernat: I see your point that the coord information was stored twice.

My understanding was that having `_DIMENSION_KEY` in the global dataset attributes was necessary for `set_coords` to work on non-dimensions coordinates. But maybe this is wrong--I can't find any justification for it by looking through the code.

rabernat: I suppose one option here is to just have no dimensions if there is no `_DIMENSION_KEY` on the variable. That would allow us to open any zarr store, even those not created by xarray.

jhamman: >  My understanding was that having _DIMENSION_KEY in the global dataset attributes was necessary for set_coords to work on non-dimensions coordinates. But maybe this is wrong--I can't find any justification for it by looking through the code.

 Right, I think that may have been the case but I've worked around that now.

> I suppose one option here is to just have no dimensions if there is no _DIMENSION_KEY on the variable. That would allow us to open any zarr store, even those not created by xarray.

I think this is doable but I want to leave that for another PR since we'll need to make a number of decisions about how to name dimensions that were not specified in the store.

xarray/backends/zarr.py
```diff
@@ -331,29 +313,34 @@ def get_variables(self):
...
+        attributes = OrderedDict(self.ds.attrs.asdict())
+        return attributes
 
     def get_dimensions(self):
-        dimensions, _ = _get_zarr_dims_and_attrs(self.ds, _DIMENSION_KEY)
+        dimensions = OrderedDict()
+        for k, v in self.ds.arrays():
+            try:
+                for d, s in zip(v.attrs[_DIMENSION_KEY], v.shape):
+                    dimensions[d] = s
```
jhamman: I think we need a consistency check here to ensure all the Zarr arrays have the same length on their labeled dimensions.

jhamman:
This is ready for some final reviews. Although the title refers to Zarr, I ended up doing some minor refactoring of all the backends so a critical eye there would be nice. I'm also hoping to use @shoyer's coding/encoding work in #1803 here once that is merged.

xarray/backends/common.py
```diff
@@ -234,23 +253,42 @@ def set_variables(self, variables, check_encoding_set,
 
             self.writer.add(source, target)
 
-    def set_necessary_dimensions(self, variable, unlimited_dims=None):
+    def set_dimensions(self, variables, unlimited_dims=None):
```
rabernat: Can you add a docstring or comment explaining what this method does?

This would help new developers (including myself) come onboard with backend development.

xarray/backends/zarr.py
rabernat: 👍

xarray/backends/zarr.py
rabernat: maybe leave this as a #TODO?

xarray/tests/test_backends.py
```diff
@@ -1189,30 +1213,21 @@ def test_hidden_zarr_keys(self):
...
             for var in expected.variables.keys():
-                assert (zarr_group[var].attrs[self.DIMENSION_KEY]
-                        == list(expected[var].dims))
+                dims = zarr_group[var].attrs[self.DIMENSION_KEY]
+                assert dims == list(expected[var].dims)
 
             with xr.decode_cf(store) as actual:
                 # make sure it is hidden
-                assert self.DIMENSION_KEY not in actual.attrs
+                # assert self.DIMENSION_KEY not in actual.attrs
```
rabernat: remove comment

xarray/tests/test_backends.py
```diff
@@ -1189,30 +1213,21 @@ def test_hidden_zarr_keys(self):
...
 
-            # verify that the dataset fails to open if dimension key is missing
-            del zarr_group.attrs[self.DIMENSION_KEY]
-            with pytest.raises(KeyError):
-                with xr.decode_cf(store) as actual:
-                    pass
-
             # put it back and try removing from a variable
-            zarr_group.attrs[self.DIMENSION_KEY] = {}
+            # zarr_group.attrs[self.DIMENSION_KEY] = {}
```
rabernat: remove comment

jhamman:
@rabernat - thanks for the review. I've addressed all your comments. I've also tested using this to push data to GCS. We're still getting the best performance we've been able to get when this is combined with gcsf:master and zarr:master.

rabernat:
Eventually we will want to think about testing xarray / zarr / gcsfs integration automatically somewhere.

xarray/tests/test_backends.py
```diff
@@ -1273,10 +1273,6 @@ def test_group(self):
         with self.roundtrip(original, save_kwargs={'group': group},
                             open_kwargs={'group': group}) as actual:
             self.assertDatasetIdentical(original, actual)
-        with pytest.raises(KeyError):
-            with self.roundtrip(original,
-                                save_kwargs={'group': group}) as actual:
-                self.assertDatasetIdentical(original, actual)
```
rabernat: Why did you remove this part of the test?

jhamman: This is an interesting one. I wanted to see if the tests all passed after this change before bringing this up for discussion. 

The key error that is raised here was being raised when looking for the `_DIMENSIONS_KEY` on the group. Now that each group doesn't have that key, it obviously isn't raised. So the question up for discussion is what should we do when we open a zarr group that doesn't have anything in it? Currently, an empty group is returned as an empty Dataset.

rabernat: So this is related to my earlier comment. I think we should pick a behavior we want to support. Either
- You can open any zarr group with xarray
- You can only open zarr groups that themselves were created by xarray

The current docs say the latter. However, having eliminated `_DIMENSION_KEY` from the group attributes, there is no way to check for whether the group was created by xarray if there are no variables.

I can't imagine any scenario in which a user would want to open an empty zarr group in read-only mode. That would be useless. I imagine that this means the dataset was written into a group So let's raise an error for now if the group is empty. 

In a future PR, we can figure out how to more comprehensively support the reading of zarr datasets that do not have xarray's special attributes.

jhamman: Just to play devils advocate for a moment, The current behavior is consistent with how we treat groups in netCDF4. Consider the following example:

```Python
# create a dataset with a group and nothing in the root group
In [16]: import netCDF4 as nc4

In [17]: rg = nc4.Dataset('/Users/jhamman/workdir/junk.nc', 'w')

In [18]: g1 = rg.createGroup('some/random/path')

In [19]: g1
Out[19]:
<class 'netCDF4._netCDF4.Group'>
group /some/random/path:
    dimensions(sizes):
    variables(dimensions):
    groups:

In [20]: g1.createDimension('time', None)
Out[20]: <class 'netCDF4._netCDF4.Dimension'> (unlimited): name = 'time', size = 0

In [22]: time = g1.createVariable('time', 'f8', ('time', ))

In [23]: time[:] = 30

In [25]: rg
Out[25]:
<class 'netCDF4._netCDF4.Dataset'>
root group (NETCDF4 data model, file format HDF5):
    dimensions(sizes):
    variables(dimensions):
    groups: some

In [26]: rg.close()

# opening the dataset in xarray returns an empty dataset
In [27]: xr.open_dataset('/Users/jhamman/workdir/junk.nc')
Out[27]:
<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*

# but if you specify the group, you get the expected dataset
In [28]: xr.open_dataset('/Users/jhamman/workdir/junk.nc', group='some/random/path')
Out[28]:
<xarray.Dataset>
Dimensions:  (time: 1)
Coordinates:
  * time     (time) float64 30.0
Data variables:
    *empty*
```

If we were to raise an error in the Zarr backend, would we want to check for the presence of variables or attributes?

rabernat: Ok, that's a pretty good point.

I'm fine with leaving as is.

jhamman:
Time for final comments. I'll merge tomorrow if nobody has any further changes.

xarray/backends/common.py
```diff
@@ -234,23 +318,55 @@ def set_variables(self, variables, check_encoding_set,
...
-        AbstractWritableDataStore.store(self, cf_variables, cf_attrs,
+        variables, attributes = cf_encoder(variables, attributes)
+        variables = OrderedDict([(k, self.encode_variable(v))
+                                 for k, v in variables.items()])
+        attributes = OrderedDict([(k, self.encode_attribute(v))
+                                  for k, v in attributes.items()])
+        return variables, attributes
+
+    def store(self, variables, attributes, *args, **kwargs):
+        AbstractWritableDataStore.store(self, variables, attributes,
```
shoyer: You can remove this method -- the implementation is exactly the parent class method.

xarray/backends/common.py
```diff
@@ -195,6 +198,37 @@ def __init__(self, writer=None):
             writer = ArrayWriter()
         self.writer = writer
 
+    def encode(self, variables, attributes):
```
shoyer: This is definitely cleaner than what we had before, but I am reluctant to give the idea that this this is a new supported method for third-party DataStore classes. Maybe we can call this `_encode` for now, or add a warning about implementing it?

Eventually, I would like to remove all implementations from the DataStore base classes, and leave them as purely abstract. This will make it clearer to new backend implementers what they actually should/can implement.

So instead of implementing an `encode()` method, data store classes could have a list of default `encoders` (see xarray.coding) used when reading/writing data. But xarray.coding isn't quite ready for this yet...

rabernat: Clearly some refactoring will be needed to this once the overall backend refactoring moves forward. For now, however, this seems like a reasonable compromise.

jhamman: @shoyer - I'm a bit confused here. As you'll see in the Zarr backend, the `encode_variable` method is applying a list of encoders. Where in the `WritableDataStore` were you envisioning the application of the encoders?

jhamman: @shoyer - What would you say to merging this in its current state and leaving the encoders refactor to a separate PR? I'm happy to make more changes here but a) I'm not sure how to address your last comment, and b) I've already drifted a fair ways off track with this PR.

shoyer: I'm fine with that. We may want to change it more in the future but this is a clear improvement for now.

mrocklin:
Woot.

PR #1804, opened 2017-12-31
Docs for multiple line plots.

dcherian:
As requested, I've added a section to the 1D plotting documentation.

I don't like the title. but couldn't think of a better one...

### DIFF
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -174,14 +174,42 @@ entire figure (as for matplotlib's ``figsize`` argument).
 
 .. _not equivalent to matplotlib's: https://github.com/mwaskom/seaborn/issues/746
 
+
+Multiple lines showing variation along a dimension
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to make line plots of two-dimensional data by calling ``xarray.plot.line()`` with appropriate arguments. Consider the 3D variable ``air``
+
+.. ipython:: python
+
+    air
+
+We can use line plots to check the variation of air temperature at three different latitudes along a longitude line:
+
+.. ipython:: python
+
+    @savefig plotting_example_multiple_lines_x_kwarg.png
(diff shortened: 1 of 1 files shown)


doc/plotting.rst
```diff
@@ -174,14 +174,42 @@ entire figure (as for matplotlib's ``figsize`` argument).
...
 
+
+Multiple lines showing variation along a dimension
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to make line plots of two-dimensional data by calling ``xarray.plot.line()`` with appropriate arguments. Consider the 3D variable ``air``
+
+.. ipython:: python
+
+    air
```
shoyer: We use `air` as a running example on this page. I don't think we need to print it again here.

doc/plotting.rst
```diff
@@ -174,14 +174,42 @@ entire figure (as for matplotlib's ``figsize`` argument).
...
+It is required to explicitly specify either
+
+1. ``x``: the dimension to be used for the x-axis, or
+2. ``hue``: the dimension you want to represent by multiple lines.
+
+Thus, we can make the previous plot by specifying ``hue='lat'`` instead of ``x='time'``. If required, the automatic legend can be turned off using ``add_legend=False``.
+
+.. ipython:: python
+
+    @savefig plotting_example_multiple_lines_hue_kwarg.png
```
shoyer: I would omit this example. `add_legend=False` doesn't need an example to make it clear what it does, and plots take up lots of space in the docs.

dcherian: It also illustrates hue, but we could just leave that as a text description.

Is there eat you would prefer?

shoyer: I'm OK with it in the text description.

fmaussion:
Ups, I realized that was a bit fast of a merge. I hope everyone agrees.

Thanks @dcherian !

fmaussion:
See it rendered [here](http://xarray.pydata.org/en/latest/plotting.html#multiple-lines-showing-variation-along-a-dimension)

PR #1808, opened 2018-01-03
Add gallery example for multiple lines plot

fmaussion:
Extends https://github.com/pydata/xarray/pull/1804 by adding some usage examples to the gallery.

Here is how the plot looks like:

sphx_glr_plot_lines_from_2d_001

cc @dcherian and @shoyer


dcherian:
lgtm. thanks.

PR #1811, opened 2018-01-07
WIP: Compute==False for to_zarr and to_netcdf

### LINKED ISSUE #1784: Add compute=False keywords to `to_foo` functions
When working with @jhamman profiling the `to_zarr` method on large datasets I wanted the ability to run through the `to_zarr` setup code, but avoid waiting on the dask computation to finish.  In many functions in Dask proper our `to_foo` methods have a `compute=False` keyword that returns a `dask.delayed` object on which people can call `compute` later if desired.

cc @jhamman @rabernat @jakirkham (who has been looking at similar questions within `dask.array.Array.store`)

jhamman:
review of this can wait until after #1800 is merged. 

 - [x] Closes #1784

cc @mrocklin


xarray/backends/api.py
```diff
@@ -743,19 +748,23 @@ def save_mfdataset(datasets, paths, mode='w', format=None, groups=None,
...
+    stores = [to_netcdf(ds, path, mode, format, group, engine, writer,
+                        compute=compute)
               for ds, path, group in zip(datasets, paths, groups)]
+
     try:
-        writer.sync()
+        delayed = writer.sync(compute=compute)
         for store in stores:
-            store.sync()
+            store.sync(compute=compute)
```
jhamman: I'm not sure why/if we need this sync line here. My understanding is that syncing the writer should be enough here.

xarray/backends/common.py
```diff
@@ -264,19 +264,22 @@ def add(self, source, target):
         else:
             target[...] = source
 
-    def sync(self):
+    def sync(self, compute=True):
         if self.sources:
             import dask.array as da
-            da.store(self.sources, self.targets, lock=self.lock)
+            delayed = da.store(self.sources, self.targets, lock=self.lock,
+                               compute=compute, flush=True)
```
rabernat: I guess this is where the magic happens. You are just exposing `dask.array.store(... compute=False)` as an option from xarray.

xarray/backends/h5netcdf_.py
```diff
@@ -183,10 +183,11 @@ def prepare_variable(self, name, variable, check_encoding=False,
 
         return target, variable.data
 
-    def sync(self):
+    def sync(self, compute=True):
         with self.ensure_open(autoclose=True):
-            super(H5NetCDFStore, self).sync()
+            super(H5NetCDFStore, self).sync(compute=compute)
             self.ds.sync()
+        return
```
rabernat: This sync logic feels repetitive across the different backend classes. Maybe there is a way to refactor a bit to avoid redundancy.

jhamman: I totally agree. I'd like to get this to work for h5netcdf before I embark on a refactor though.

xarray/backends/api.py
```diff
@@ -664,9 +665,12 @@ def to_netcdf(dataset, path_or_file=None, mode='w', format=None, group=None,
     if not sync:
         return store
```
shoyer: If you look above, we only set `sync = None` if a writer is provided explicitly, which is only used for `open_mfdataset`. This return value should probably be eliminated -- instead we could always return a delayed object for open_mfdataset.

xarray/backends/api.py
shoyer: Sometimes you need to sync a backend object to actually save data to disk. Synchronizing the writer only saves data to the backend's arrays, but that data is sometimes queued up before writing (e.g., with scipy, which only writes full netCDF files).

xarray/backends/h5netcdf_.py
shoyer: an empty `return` is a pointless if it's the last nice of a function -- you should remove these.

shoyer:
One potential issue here is the lack of clean-up (which may be unnecessary if `autoclose=True`). You want to construct a single dask graph with a structure like the following:
- Tasks for writing all array data (i.e., from `ArrayWriter`).
- Tasks for calling sync() and close() on each datastore object. These should depend on the appropriate writing tasks.
- A single tasks that depends on writing all datastores. This is what the delayed object returned by `save_mfdataset` should return.

shoyer:
To elaborate a little bit on my last comment (which I submitted very quickly when my bus was arriving), the way to make dependent tasks with dask.delayed is to add dummy function arguments, e.g.,
```python
def finalize_store(store, write):
    del write  # unused
    store.sync()
    store.close()

write = dask.array.store(..., compute=False)
write_and_close = dask.delayed(finalize_store)(store, write)
write_and_close.compute()  # writes and syncs
```

Potentially some of this logic could get moved into `ArrayWriter.sync()`

xarray/backends/api.py
jhamman: @shoyer - did you mean `save_mfdataset` instead of `open_mfdataset`?

shoyer: Yes, my mistake.

jhamman:
@shoyer - I did my best to implement the finalize store concept you've described above. This all seems to work nicely except for `h5netcdf`. During the store operation, `h5netcdf` is raising the following error:

```Python
self = <h5netcdf.core._LazyObjectLookup object at 0x122d8b748>, key = 'foo'

    def __getitem__(self, key):
>       if self._objects[key] is not None:
E       KeyError: 'foo'
```

xarray/backends/api.py
```diff
@@ -743,19 +763,26 @@ def save_mfdataset(datasets, paths, mode='w', format=None, groups=None,
...
               for ds, path, group in zip(datasets, paths, groups)]
+
     try:
-        writer.sync()
+        delayed = writer.sync(compute=compute)
+        if not compute:
+            import dask
+            # or should this return an iterable of delayed objects
+            # (one for each store)
+            return dask.delayed(_finalize_stores)(stores, delayed)
```
jhamman: I'm open to ideas from folks on this one. Do we want to return a list of delayed objects or a single delayed object when using this option with `save_mfdataset`?

xarray/backends/api.py
```diff
@@ -743,19 +763,26 @@ def save_mfdataset(datasets, paths, mode='w', format=None, groups=None,
...
         for store in stores:
-            store.sync()
+            store.sync(compute)
```
jhamman: I don't think this `compute` needs to be here anymore. (note to self)

rsignell-usgs:
@jhamman , I'm trying to test `compute=False` out this code:
```python
# Write National Water Model data to Zarr

import pandas as pd
import xarray as xr
import s3fs
import zarr

if __name__ == '__main__':

    client = Client()

    root = '/projects/water/nwm/data/forcing_short_range/'                      # Local Files
#   root = 'http://tds.renci.org:8080/thredds/dodsC/nwm/forcing_short_range/'   # OPenDAP


    bucket_endpoint='https://s3.us-west-1.amazonaws.com/'
#   bucket_endpoint='https://iu.jetstream-cloud.org:8080'

    f_zarr = 'rsignell/nwm/test_week'

    dates = pd.date_range(start='2018-04-01T00:00', end='2018-04-07T23:00', freq='H')
    urls = ['{}{}/nwm.t{}z.short_range.forcing.f001.conus.nc'.format(root,a.strftime('%Y%m%d'),a.strftime('%H')) for a in dates]

    ds = xr.open_mfdataset(urls, concat_dim='time', lock=True)
    ds = ds.drop(['ProjectionCoordinateSystem'])

    fs = s3fs.S3FileSystem(anon=False, client_kwargs=dict(endpoint_url=bucket_endpoint))
    d = s3fs.S3Map(f_zarr, s3=fs)

    compressor = zarr.Blosc(cname='zstd', clevel=3, shuffle=2)
    encoding = {vname: {'compressor': compressor} for vname in ds.data_vars}

    delayed_store = ds.to_zarr(store=d, mode='w', encoding=encoding, compute=False)
    persist_store = delayed_store.persist(retries=100)
```
and after 20 seconds or so, the process dies with this error:

```python-traceback
home/rsignell/my-conda-envs/zarr/lib/python3.6/site-packages/distributed/worker.py:742: UserWarning: Large object of size 1.23 MB detected in task graph:
  (<xarray.backends.zarr.ZarrStore object at 0x7fb49 ... deedecefab224')
Consider scattering large objects ahead of time
with client.scatter to reduce scheduler burden and
keep data on workers

    future = client.submit(func, big_data)    # bad

    big_future = client.scatter(big_data)     # good
    future = client.submit(func, big_future)  # good
  % (format_bytes(len(b)), s))
```
How should I modify my code?

jhamman:
@rsignell-usgs - can you repeat this example with a single threaded scheduler. It will be slow but it should work (or return a more informative error).

rsignell-usgs:
@jhamman, I tried the same code with a single-threaded scheduler:
```python
    ...
    delayed_store = ds.to_zarr(store=d, mode='w', encoding=encoding, compute=False)
    persist_store = delayed_store.persist(retries=100, get=dask.local.get_sync)
```
and it ran to completion with no errors.   What should I try next?

jhamman:
@rsignell-usgs - This is going to require some debugging on your part but here are a few suggestions:

- try smaller chunksize
- use a LocalCluster before using the KubeCluster
- turn up dask's logging level for the scheduler/workers

> and after 20 seconds or so, the process dies with this error:

looking above, I don't see any error. Your goal in the next step should be to find the error. If workers are dying, they should report that in the worker logs. If your cluster is dying, that should be reported in the notebook.

rsignell-usgs:
I was just using `client = Client()`.   Should I be using `LocalCluster` instead?

jhamman:
I see you were already using the LocalCluster (Client). Disregard my comment on switching clusters. I seem to be getting my Github issues mixed up. 

It may be good to take this offline to one of the pangeo/zarr/dask/xarray issues (there are a few).

doc/whats-new.rst
```diff
@@ -36,6 +36,11 @@ Enhancements
 
 - Support writing lists of strings as netCDF attributes (:issue:`2044`).
```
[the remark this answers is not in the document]
jhamman: No. This is just showing the two lines above where the diff takes place.

xarray/backends/api.py
```diff
@@ -764,10 +776,17 @@ def save_mfdataset(datasets, paths, mode='w', format=None, groups=None,
...
               for ds, path, group in zip(datasets, paths, groups)]
+
     try:
-        writer.sync()
+        delayed = writer.sync(compute=compute)
+        if not compute:
+            import dask
+            # or should this return an iterable of delayed objects
+            # (one for each store)
+            return dask.delayed(_finalize_stores)(delayed, *stores)
```
shoyer: Thinking about this a little more, this is a sub-optimal way to setup the task graph for dask:
- Currently, you have: all write tasks -> close all files -> done
- Rather, we should have: set of all {write one file -> close the file}  -> done

The later is far less likely to run into issues with too many open files, because it closes each file immediately when it's done writing.

To do this, I think we should setup `to_netcdf()` to return delayed objects rather than stores if `compute=False`, in which case we can simply collect them all together in this result task.

We might also make dask required for `save_mfdataset` since it really is entirely useless otherwise and that would let us simplify this logic (e.g., to get ride of the `Writer` class).

jhamman: @shoyer - I think this is a good idea. However, I'm at a loss as to how to implement it. Should we just call `_finalize_store` in `to_netcdf` or should that happen after `writer.sync(compute=compute)` is called?

jhamman:
@shoyer I'm getting a [test failure](https://travis-ci.org/pydata/xarray/jobs/378274606#L5269-L5326) in the h5netcdf backend that seems unrelated. Do you know if something has changed in the string handling of h5netcdf recently?

jhamman:
@shoyer / @mrocklin - I think this is ready for another review. Since I asked for reviews last, I have: 

- reworked the task graph to write one file / close one file rather than write all / close all. 
- moved the tests to the `TestDask` class which seemed more appropriate.

xarray/backends/api.py
```diff
@@ -144,6 +144,12 @@ def _get_lock(engine, scheduler, format, path_or_file):
     return lock
 
 
+def _finalize_store(write, store):
+    del write
```
shoyer: add a comment, e.g., `# ensure writing is done first`

doc/dask.rst
```diff
@@ -100,6 +100,20 @@ Once you've manipulated a dask array, you can still write a dataset too big to
...
+
+By setting the ``compute`` argument to ``False``, :py:meth:`~xarray.Dataset.to_netcdf`
+will return a dask delayed object that can be computed later.
+
+.. ipython:: python
+
+    from dask.diagnostics import ProgressBar
+    delayed_obj = ds.to_netcdf('manipulated-example-data.nc', compute=False)
+    with ProgressBar():
+        results = delayed_obj.compute()
```
mrocklin: Note that this example will not provide a progress bar if they are using the distributed scheduler.

xarray/backends/common.py
mrocklin: @jakirkham has been working on `da.store` a bit recently.  He may have thoughts here.

xarray/backends/common.py
```diff
@@ -264,19 +264,22 @@ def add(self, source, target):
...
             self.sources = []
             self.targets = []
+            return delayed
```
mrocklin: It may be wise to use another term here to avoid confusion with the `dask.delayed` function itself.

shoyer:
> Yes, just tried again. I'm open to ideas but would also like to move this issue along first, if possible.

Sounds good, let's go ahead and merge this!

shoyer:
(assuming tests pass)

xarray/backends/common.py
jakirkham: Not sure that `flush` does anything actually. Generic `**kwargs` get forwarded onto `compute`/`persist` depending. 

FWIW my typical usage is `compute=True, return_stored=True`. This is pretty nice as one gets back a Dask Array with the ability to load in data that has been stored to disk. Pretty nice for inspecting the result as it computes. Can also be tossed into a Dask progress bar to get details about how much has completed.

PR #1819, opened 2018-01-11
Normalisation for RGB imshow

Zac-HD:
Follow-up to #1796, where normalisation and clipping of RGB[A] values were deferred so that we could match any upstream API.  matplotlib/matplotlib#10220 implements clipping to the valid range, but a strong consensus *against* RGB normalisation in matplotlib has emerged.

This pull therefore implements normalisation, and clips values only where our normalisation has pushed them out of range.

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -15,8 +15,8 @@ import numpy as np
 import pandas as pd
 from datetime import datetime
 
-from .utils import (_determine_cmap_params, _infer_xy_labels, get_axis,
-                    import_matplotlib_pyplot)
+from .utils import (ROBUST_PERCENTILE, _determine_cmap_params,
+                    _infer_xy_labels, get_axis, import_matplotlib_pyplot)
 from .facetgrid import FacetGrid
 from xarray.core.pycompat import basestring
 
@@ -326,6 +326,39 @@ class _PlotMethods(object):
         return line(self._da, *args, **kwargs)
 
 
+def _rescale_imshow_rgb(darray, vmin, vmax, robust):
+    assert robust or vmin is not None or vmax is not None
+    # There's a cyclic dependency via DataArray, so we can't import from
+    # xarray.ufuncs in global scope.
+    from xarray.ufuncs import maximum, minimum
+    # Calculate vmin and vmax automatically for `robust=True`
+    if robust:
+        if vmax is None:
+            vmax = np.nanpercentile(darray, 100 - ROBUST_PERCENTILE)
+        if vmin is None:
+            vmin = np.nanpercentile(darray, ROBUST_PERCENTILE)
+    # If not robust and one bound is None, calculate the default other bound
+    # and check that an interval between them exists.
+    elif vmax is None:
+        vmax = 255 if np.issubdtype(darray.dtype, np.integer) else 1
+        if vmax < vmin:
+            raise ValueError(
+                'vmin=%r is less than the default vmax (%r) - you must supply '
+                'a vmax > vmin in this case.' % (vmin, vmax))
+    elif vmin is None:
+        vmin = 0
+        if vmin > vmax:
+            raise ValueError(
+                'vmax=%r is less than the default vmin (0) - you must supply '
+                'a vmin < vmax in this case.' % vmax)
+    # Scale interval [vmin .. vmax] to [0 .. 1], with darray as 64-bit float
+    # to avoid precision loss, integer over/underflow, etc with extreme inputs.
+    # After scaling, downcast to 32-bit float.  This substantially reduces
+    # memory usage after we hand `darray` off to matplotlib.
+    darray = ((darray.astype('f8') - vmin) / (vmax - vmin)).astype('f4')
+    return minimum(maximum(darray, 0), 1)
+
+
 def _plot2d(plotfunc):
     """
     Decorator for common 2d plotting logic
@@ -449,6 +482,11 @@ def _plot2d(plotfunc):
         if imshow_rgb:
             # Don't add a colorbar when showing an image with explicit colors
             add_colorbar = False
+            # Matplotlib does not support normalising RGB data, so do it here.
+            # See eg. https://github.com/matplotlib/matplotlib/pull/10220
+            if robust or vmax is not None or vmin is not None:
+                darray = _rescale_imshow_rgb(darray, vmin, vmax, robust)
+                vmin, vmax, robust = None, None, False
 
         # Handle facetgrids first
         if row or col:
@@ -624,6 +662,11 @@ def imshow(x, y, z, ax, **kwargs):
     two-dimensional, ``imshow`` also accepts a 3D array where some
     dimension can be interpreted as RGB or RGBA color channels and
     allows this dimension to be specified via the kwarg ``rgb=``.
+
+    Unlike matplotlib, Xarray can apply ``vmin`` and ``vmax`` to RGB or RGBA
+    data, by applying a single scaling factor and offset to all bands.
+    Passing  ``robust=True`` infers ``vmin`` and ``vmax``
+    :ref:`in the usual way <robust-plotting>`.
 
     .. note::
         This function needs uniformly spaced coordinates to
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -1126,6 +1126,16 @@ class TestImshow(Common2dMixin, PlotTestCase):
         with pytest.raises(ValueError):
             arr.plot.imshow(rgb='band')
 
+    def test_normalize_rgb_imshow(self):
+        for kwds in (
+            dict(vmin=-1), dict(vmax=2),
+            dict(vmin=-1, vmax=1), dict(vmin=0, vmax=0),
+            dict(vmin=0, robust=True), dict(vmax=-1, robust=True),
+        ):
+            da = DataArray(easy_array((5, 5, 3), start=-0.6, stop=1.4))
+            arr = da.plot.imshow(**kwds).get_array()
+            assert 0 <= arr.min() <= arr.max() <= 1, kwds
+
 
 class TestFacetGrid(PlotTestCase):
     def setUp(self):
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -1136,6 +1136,16 @@ class TestImshow(Common2dMixin, PlotTestCase):
             arr = da.plot.imshow(**kwds).get_array()
             assert 0 <= arr.min() <= arr.max() <= 1, kwds
 
+    def test_normalize_rgb_one_arg_error(self):
+        da = DataArray(easy_array((5, 5, 3), start=-0.6, stop=1.4))
+        # If passed one bound that implies all out of range, error:
+        for kwds in [dict(vmax=-1), dict(vmin=2)]:
+            with pytest.raises(ValueError):
+                da.plot.imshow(**kwds)
+        # If passed two that's just moving the range, *not* an error:
+        for kwds in [dict(vmax=-1, vmin=-1.2), dict(vmin=2, vmax=2.1)]:
+            da.plot.imshow(**kwds)
+
 
 class TestFacetGrid(PlotTestCase):
     def setUp(self):
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -305,6 +305,8 @@ example, consider the original data in Kelvins rather than Celsius:
 The Celsius data contain 0, so a diverging color map was used. The
 Kelvins do not have 0, so the default color map was used.
 
+.. _robust-plotting:
+
 Robust
 ~~~~~~
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -41,6 +41,7 @@ Enhancements
 - Support for using `Zarr`_ as storage layer for xarray.
   By `Ryan Abernathey <https://github.com/rabernat>`_.
 - :func:`xarray.plot.imshow` now handles RGB and RGBA images.
+  Saturation can be adjusted with ``vmin`` and ``vmax``, or with ``robust=True``.
   By `Zac Hatfield-Dodds <https://github.com/Zac-HD>`_.
 - Experimental support for parsing ENVI metadata to coordinates and attributes
   in :py:func:`xarray.open_rasterio`.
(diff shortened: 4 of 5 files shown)


Zac-HD:
CC @shoyer - hopefully I'm in time for Xarray 0.10.1 😉

xarray/plot/plot.py
```diff
@@ -449,6 +450,21 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
         if imshow_rgb:
             # Don't add a colorbar when showing an image with explicit colors
             add_colorbar = False
+            # Calculate vmin and vmax automatically for `robust=True`
+            if robust:
+                vmin_tmp, vmax_tmp = darray.quantile(
+                    q=[ROBUST_PERCENTILE, 1 - ROBUST_PERCENTILE])
+                vmin = vmin if vmin is not None else vmin_tmp
+                vmax = vmax if vmax is not None else vmax_tmp
+                del vmin_tmp, vmax__tmp
```
jklymak: typo. `vmax_tmp`

xarray/plot/plot.py
```diff
@@ -449,6 +450,21 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+                robust=False
+            # Scale interval [vmin .. vmax] to [0 .. 1] and clip to bounds
+            if vmin is not None or vmax is not None:
+                vmin = vmin if vmin is not None else darray.min()
+                vmax = vmax if vmax is not None else darray.max()
+                darray = (darray.astype(np.float32) - vmin) / (vmax - vmin)
```
jklymak: We ran into dynamic range issues w/ `uint32` being converted to `float32`, particularly as some people use an extreme high value as a bad-data flag.  https://github.com/matplotlib/matplotlib/pull/10133

xarray/plot/plot.py
```diff
@@ -625,6 +641,10 @@ def imshow(x, y, z, ax, **kwargs):
     dimension can be interpreted as RGB or RGBA color channels and
     allows this dimension to be specified via the kwarg ``rgb=``.
 
+    Unlike matplotlib, Xarray can apply ``vmin`` and ``vmax`` to RGB or RGBA
+    data, by applying a single scaling factor and offset to all bands.
+    Passing  ``robust=True`` infers ``vmin`` and ``vmax`` in the usual way.
```
jklymak: ... infers vmin and vmax from the 96-th percentile of the data distribution. (Or reference the “usual way” somehow).

xarray/plot/plot.py
```diff
@@ -449,6 +450,21 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
             # Don't add a colorbar when showing an image with explicit colors
             add_colorbar = False
+            # Calculate vmin and vmax automatically for `robust=True`
+            if robust:
+                vmin_tmp, vmax_tmp = darray.quantile(
+                    q=[ROBUST_PERCENTILE, 1 - ROBUST_PERCENTILE])
+                vmin = vmin if vmin is not None else vmin_tmp
+                vmax = vmax if vmax is not None else vmax_tmp
+                del vmin_tmp, vmax__tmp
+                robust=False
```
jhamman: 2 nits: 1) let garbage collection handle deleting your temp variables, 2) space in `robust = False`

xarray/tests/test_plot.py
```diff
@@ -1115,6 +1115,16 @@ def test_rgb_errors_bad_dim_sizes(self):
...
 
+    @pytest.mark.parametrize('kwds', [
+        dict(vmin=-1), dict(vmax=2), dict(vmin=-1, vmax=1),
+        dict(vmin=0, robust=True), dict(vmax=-1, robust=True),
+    ])
+    def test_normalise_rgb_imshow(kwds):
+        # Test for crash when normalising RGB data.  (further tests require a
+        # reference image, as the args used to be silently ignored instead)
+        da = easy_array((5, 5, 3), start=-0.6, stop=1.4)
+        da.plot.imshow(**kwds)
```
jhamman: I think you can test that the range of the values in this plot are correct.

xarray/plot/plot.py
```diff
@@ -449,6 +450,21 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
         if imshow_rgb:
             # Don't add a colorbar when showing an image with explicit colors
             add_colorbar = False
+            # Calculate vmin and vmax automatically for `robust=True`
+            if robust:
+                vmin_tmp, vmax_tmp = darray.quantile(
+                    q=[ROBUST_PERCENTILE, 1 - ROBUST_PERCENTILE])
+                vmin = vmin if vmin is not None else vmin_tmp
+                vmax = vmax if vmax is not None else vmax_tmp
```
jhamman: perhaps we should skip the quantile calculation when both vmin/vmax are set. Then again, if robust is True, the behavior is poorly defined and maybe an error would be better.

xarray/plot/plot.py
shoyer: I think we use `locals()` somewhere below, which is why this is necessary (but yes it's pretty ugly)

Zac-HD:
@jhamman - all done 😄

xarray/plot/plot.py
```diff
@@ -449,10 +449,29 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+            # Scale interval [vmin .. vmax] to [0 .. 1] and clip to bounds
+            if vmin is not None or vmax is not None:
+                vmin = vmin if vmin is not None else darray.min()
+                vmax = vmax if vmax is not None else darray.max()
+                darray = darray.astype('f4' if vmax - vmin < 2 ** 32 else 'f8')
+                darray = ((darray - vmin) / (vmax - vmin)).astype('f4')
+                vmin, vmax = None, None
+                # There's a cyclic dependency via DataArray, so we can't
+                # import xarray.ufuncs in global or outer scope.
+                import xarray.ufuncs as xu
```
jhamman: we should be able to fix this.

xarray/plot/plot.py
```diff
@@ -449,10 +449,29 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+                if vmin is None:
+                    vmin = np.nanpercentile(darray, ROBUST_PERCENTILE)
+                if vmax is None:
+                    vmax = np.nanpercentile(darray, 100 - ROBUST_PERCENTILE)
+                robust = False
+            # Scale interval [vmin .. vmax] to [0 .. 1] and clip to bounds
+            if vmin is not None or vmax is not None:
+                vmin = vmin if vmin is not None else darray.min()
+                vmax = vmax if vmax is not None else darray.max()
+                darray = darray.astype('f4' if vmax - vmin < 2 ** 32 else 'f8')
```
jhamman: I think you can use `np.finfo('f4').max` here instead of `2**32`.

jklymak: I think it was uint32 that gave us trouble on Matplotlib.  i.e. way less than f4.max (I think?).  See https://github.com/matplotlib/matplotlib/pull/10133.  I was too lazy to figure out the actual limit (I guess its 2^N, where N is how many bits a float has dynamic range over.  I think its 32-1-8 = 2**23?

jhamman: So you can use `np.iinfo` instead:

```In [12]: np.iinfo(np.uint32)
Out[12]: iinfo(min=0, max=4294967295, dtype=uint32)
```

jklymak: Yeah, but thats not the problem.  The problem is that if you convert (a-vmin)/(1e9) then you can't tell the difference between a=1 and a=2 anymore if you only use float32.  

This came up in Matplotlib because the conversion to float happened with an automatic vmin and vmax, and the user had a large value as a bad-data flag.  *Then* the user subsequently set vmin - vmax = 10 or so, but the convert to float32 had erased all the dynamic between 0 and 10.  

Maybe `xarray` doesn't have this problem because you trust the user to set `vmin` and `vmax` during the `imshow` call.  But if someone grabs the image handle and tries to set the limits after the call they could have this problem.

Zac-HD: The more deeply I analyse matplotlib/matplotlib#10133, the more suspicious I get - there are actually *two* places you could get loss of precision: subtracting `vmin` (if <<0, loss of precision for high values), and dividing by `(vmin - vmax)`  (loss if vmax is large).

I think for Xarray, the best option is actually just to use `np.float64` in all cases.  We already have a habit of upcasting all the time anyway\*, and if that's lossy you're in trouble anyway.

\* (including `uint16` to `float64` when scaling, which is silly - issue incoming for that)

jklymak: vmin and A are still ints when vmin is removed from A.

Zac-HD: In which case you are vulnerable to underflow, and hence wraparound of values - that's why I cast to float64 here *first*!

jklymak: Not sure I follow.    

The case we were running into was 
```python
A = np.array( [[1, 2, 1e9], [3, 4, 5]], dtype=np.uint32)
im = plt.imshow(A)
im.set_clim([0., 5.])
```

When the data in the image is first input, `vmin = 1` and `vmax = 1e9`, and the underflow occurred when we cast to `np.float32`, which we did to help accommodate very large images. Hence the check on `vmax - vmin` to see if it is larger than 1e8 and to use `np.float64` for the case instead.  

Casting to `float64` first is great but apparently people sometimes have huge images, hence the extra rigamarole w/ using `float32` if possible.

Zac-HD: > vmin and A are still ints when vmin is removed from A.

Going and reading the code again, it's not - a_min is subtracted from A_scaled, which has a floating dtype.  In which case you're also safe from overflow/underflow, and have just added handling for loss of precision 😄

I remain happy with this solution - for Xarray, including passing non-RGB images directly to matplotlib - and suggest we take any further discussion to the upstream pull.

Zac-HD:
@jhamman - all done 😄

xarray/plot/plot.py
```diff
@@ -449,10 +449,28 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+            # Scale interval [vmin .. vmax] to [0 .. 1] and clip to bounds
+            if vmin is not None or vmax is not None:
+                vmin = vmin if vmin is not None else darray.min()
+                vmax = vmax if vmax is not None else darray.max()
+                darray = ((darray.astype('f8') - vmin) / (vmax - vmin))
+                vmin, vmax = None, None
+                # There's a cyclic dependency via DataArray, so we can't
+                # import xarray.ufuncs in global or outer scope.
+                import xarray.ufuncs as xu
+                darray = xu.minimum(xu.maximum(darray.astype('f4'), 0), 1)
```
jhamman: do we need to type cast this back to a 4byte float?

jhamman: @shoyer - do you have a solution off had for the cyclic import problem here?

Zac-HD: We don't *need* to downcast, in that it will work even if we don't, but downcasting ~halves memory use further inside matplotlib which can be a 5-10x multiple of the input data size.  Happy to remove if you don't think that's worth it.

The best alternative solution I can see is refactoring `xarray.ufuncs` to do all the imports inside a function or class, and pay the performance cost when creating ufuncs and slightly increased overhead in dispatch.  Personally, I prefer a slight slowdown when creating scaled RGB images - much less common, and it's probably negligible given the array operations involved anyway.

Zac-HD:
Ping @shoyer, @jhamman - I think we're out of blockers and ready to merge 😄

xarray/plot/plot.py
```diff
@@ -449,10 +449,28 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+            if robust:
+                if vmin is None:
+                    vmin = np.nanpercentile(darray, ROBUST_PERCENTILE)
+                if vmax is None:
+                    vmax = np.nanpercentile(darray, 100 - ROBUST_PERCENTILE)
+                robust = False
+            # Scale interval [vmin .. vmax] to [0 .. 1] and clip to bounds
+            if vmin is not None or vmax is not None:
+                vmin = vmin if vmin is not None else darray.min()
+                vmax = vmax if vmax is not None else darray.max()
```
shoyer: Shouldn't the default here for vmin/vmax be 0/1 rather than the min/max values of the array? That would seem a little less surprising to me. Otherwise setting one of `vmin`/`vmax` would cause the other to be automatically set which seems strange.

xarray/plot/plot.py
shoyer: I'm happy with downcasting, let's just add a comment to explain why.

I don't think we can avoid the cycle import issue. This is a downside of our method heavy design. However, it would be nice if we moved the import statement to the top of the method.

xarray/plot/plot.py
```diff
@@ -449,10 +449,28 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
         if imshow_rgb:
             # Don't add a colorbar when showing an image with explicit colors
             add_colorbar = False
+            # Calculate vmin and vmax automatically for `robust=True`
```
shoyer: Can we move this logic to a separate helper function, e.g., ` _rescale_imshow_rgb(darray, vmin, vmax, robust)`? This method is already getting pretty long.

Zac-HD:
Method extracted, default for missing bounds fixed, comment added, import moved.  Also added a check that you haven't accidentally reversed the bounds while supplying only one.

xarray/plot/plot.py
```diff
@@ -326,6 +326,39 @@ def line(self, *args, **kwargs):
...
+        if vmax is None:
+            vmax = np.nanpercentile(darray, 100 - ROBUST_PERCENTILE)
+        if vmin is None:
+            vmin = np.nanpercentile(darray, ROBUST_PERCENTILE)
+    # If not robust and one bound is None, calculate the default other bound
+    # and check that an interval between them exists.
+    elif vmax is None:
+        vmax = 255 if np.issubdtype(darray.dtype, np.integer) else 1
+        if vmax < vmin:
+            raise ValueError(
```
shoyer: These error checks look great, thanks! Can you add a test that covers them?

Zac-HD: Done 😄

Zac-HD:
Ping @jhamman?

jhamman:
In it goes. Thanks @Zac-HD.

PR #1824, opened 2018-01-13
Make `flake8 xarray` pass

Zac-HD:
Closes #1741 by @mrocklin (who did most of the work I'm presenting here).  I had an evening free, so I rebased the previous pull on master, fixed the conflicts, and then made everything pass with `flake8`'s default settings (including line length).  My condolences to whoever gets to *review* this diff!

The single change any non-pedant will notice: Travis now fails if there is a flake8 warning anywhere.  My experience in other projects is that this is the only way to actually *keep* flake8 passing - it's just unrealistic to expect perfect attention to detail from every contributor, but "make the build green before we merge" is widely understood 😄

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -456,7 +456,7 @@ class DatasetIOTestCases(object):
                         assert isinstance(obj, indexing.PandasIndexAdapter)
                     else:
                         raise TypeError('{} is wrapped by {}'.format(
-                                            type(obj.array), type(obj)))
+                            type(obj.array), type(obj)))
 
         for k, v in ds.variables.items():
             find_and_validate_array(v._data)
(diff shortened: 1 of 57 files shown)


mrocklin:
Huzzah.  Thanks Zac for picking this up

xarray/tests/test_backends.py
```diff
@@ -1045,10 +1048,10 @@ def test_unsorted_index_raises(self):
         with self.roundtrip(ds) as ondisk:
             inds = np.argsort(dim1)
             ds2 = ondisk.isel(dim1=inds)
-            try:
-                print(ds2.randovar.values)  # should raise IndexError in netCDF4
-            except IndexError as err:
-                self.assertIn('first by calling .load', str(err))
+            with pytest.raises(IndexError) as err:
```
shoyer: This modified test is failing in Travis. I think we actually want to keep something similar to what we had before. The idea was that netcdf4 was raising an error that we wanted to be sure we were reraising with a better error message. NetCDF4 stopped raising the error in recent versions,so we can silently ignore this behavior if a newer version of netcdf4 is installed.

xarray/tests/test_indexing.py
```diff
@@ -27,7 +27,7 @@ def set_to_zero(self, x, i):
     def test_expanded_indexer(self):
         x = np.random.randn(10, 11, 12, 13, 14)
         y = np.arange(5)
-        I = ReturnItem()
+        I = ReturnItem()  # noqa: E741  # allow ambiguous name
```
shoyer: These meaningless noqa numbers are one reason why I like pylint over flake8 sometimes...

xarray/tests/test_backends.py
Zac-HD: Makes sense - I'll improve the comment instead 😆

Zac-HD:
@shoyer; done 🎉

shoyer:
Thanks!

PR #1825, opened 2018-01-13
Testing deprecations

maxim-lian:
Re comment here: https://github.com/pydata/xarray/pull/1640#discussion_r158384496

The `TestCase` functions I think we should deprecate 
`_importorskip` I'm less sure about - there are some reasons to keep them if we need flexibility in the future (and it doesn't matter much)

### DIFF
--- a/xarray/tests/__init__.py
+++ b/xarray/tests/__init__.py
@@ -45,6 +45,12 @@ except ImportError:
 
 
 def _importorskip(modname, minversion=None):
+    """
+    This is deprecated - instead use pytest.importorskip. For example:
+
+    matplotlib = pytest.importorskip('matplotlib', minversion='1.9.0')
+
+    """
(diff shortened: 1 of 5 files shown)


xarray/tests/__init__.py
```diff
@@ -45,6 +45,12 @@
 
 
 def _importorskip(modname, minversion=None):
+    """
+    This is deprecated - instead use pytest.importorskip. For example:
+
+    matplotlib = pytest.importorskip('matplotlib', minversion='1.9.0')
+
+    """
```
jhamman: I don't this can work until https://github.com/pytest-dev/pytest/issues/568 is fixed.

maxim-lian:
Closing in favor of https://github.com/pydata/xarray/pull/1828

PR #1827, opened 2018-01-14
Support dt.floor(), dt.ceil() and dt.round() accessors.

dcherian:

Following on from #1762 , I've tried to implement floor, ceil and round as in pandas.


xarray/core/accessors.py
```diff
@@ -146,3 +189,23 @@ def f(self, dtype=dtype):
     time = _tslib_field_accessor(
         "time", "Timestamps corresponding to datetimes", object
     )
+
+    def _tslib_round_accessor(self, name, freq, dtype=None):
```
shoyer: The `pandas.Series.dt.round()` and the other methods always return datetime64, so I don't think you need the `dtype` argument.

xarray/core/accessors.py
```diff
@@ -58,6 +58,49 @@ def _get_date_field(values, name, dtype):
         return _access_through_series(values, name)
 
 
+def _round_series(values, name, freq):
+    """Coerce an array of datetime-like values to a pandas Series and
+    apply requested rounding
+    """
+    values_as_series = pd.Series(values.ravel())
+    if name == 'floor':
+        field_values = values_as_series.dt.floor(freq=freq).values
```
shoyer: can you rewrite this to avoid the conditionals? maybe something like:
```python
method = getattr(values_as_series.dt, name)
field_values = method(freq=freq).values
```

doc/time-series.rst
```diff
@@ -129,6 +129,8 @@ the first letters of the corresponding months.
 
 You can use these shortcuts with both Datasets and DataArray coordinates.
 
+In addition, xarray supports rounding operations ``floor``, ``ceil``, and ``round``. These operations require you to supply a rounding frequency as a string argument (see list of frequency aliases at http://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases).
+
```
shoyer: Can you add a quick example here?

doc/time-series.rst
```diff
@@ -129,6 +129,8 @@ the first letters of the corresponding months.
 
 You can use these shortcuts with both Datasets and DataArray coordinates.
 
+In addition, xarray supports rounding operations ``floor``, ``ceil``, and ``round``. These operations require you to supply a rounding frequency as a string argument (see list of frequency aliases at http://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases).
```
shoyer: please make a link instead of printing the URL, e.g.,

```
These operations require you to supply a rounding `frequency as a string argument`_.

.. _frequency as a string argument: http://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases
```

xarray/core/accessors.py
```diff
@@ -58,6 +58,49 @@ def _get_date_field(values, name, dtype):
...
+    rounded timestamps : same type as values
+        Array-like of datetime fields accessed for each element in values
+
+    """
+    if isinstance(values, dask_array_type):
+        from dask.array import map_blocks
+        return map_blocks(_round_series,
+                          values, name, freq=freq, dtype=dtype)
+    else:
+        return _round_series(values, name, freq)
```
shoyer: At some point, we should probably consolidate this function with the similar looking one above. But I'm OK with this for now.

xarray/core/accessors.py
dcherian: Yes! I didn't know how to do this. The above works.

xarray/core/accessors.py
dcherian: Ya, I have no experience with dask so I just copied the previous one.

xarray/core/accessors.py
dcherian: Fixed

dcherian:
@shoyer Addressed all comments and rebased on master. Let me know if you want any more changes.

xarray/core/accessors.py
```diff
@@ -147,3 +184,21 @@ def f(self, dtype=dtype):
...
     )
+
+    def _tslib_round_accessor(self, name, freq):
+        obj_type = type(self._obj)
+        result = _round_field(self._obj.data, name, freq)
+        return obj_type(result, name=name,
+                        coords=self._obj.coords, dims=self._obj.dims)
+
+    def floor(self, freq):
+        ''' Round timestamp downward. '''
```
shoyer: It would be nice to have a docstring that explains what the freq argument should be.

xarray/core/accessors.py
```diff
@@ -58,6 +58,43 @@ def _get_date_field(values, name, dtype):
...
+    Returns
+    -------
+    rounded timestamps : same type as values
+        Array-like of datetime fields accessed for each element in values
+
+    """
+    if isinstance(values, dask_array_type):
+        from dask.array import map_blocks
+        return map_blocks(_round_series,
+                          values, name, freq=freq)
```
shoyer: Please add dtype=np.datetime64 here (it's much more efficient for dask)

xarray/tests/test_accessors.py
```diff
@@ -95,3 +95,14 @@ def test_seasons(self):
...
+    def test_rounders(self):
+        dates = pd.date_range("2014-01-01", "2014-05-01", freq='H')
+        xdates = xr.DataArray(np.arange(len(dates)),
+                              dims=['time'], coords=[dates])
+        assert_array_equal(dates.floor('D').values,
+                           xdates.time.dt.floor('D').values)
+        assert_array_equal(dates.ceil('D').values,
+                           xdates.time.dt.ceil('D').values)
+        assert_array_equal(dates.round('D').values,
+                           xdates.time.dt.round('D').values)
```
shoyer: Please add a test that exercises the dask codepath. Right now it doesn't have any test coverage at all.

dcherian:
@shoyer All done.

PR #1828, opened 2018-01-14
pytest-ification

maxim-lian:
So far, only the changes that `unittest2pytest` could fix. Need to work out how to sort through the exisiting xarray ones

EDIT: ready to merge (and keen to get this in before there are divergences from other branches)

### DIFF
--- a/xarray/core/dtypes.py
+++ b/xarray/core/dtypes.py
@@ -22,9 +22,11 @@ def maybe_promote(dtype):
     # N.B. these casting rules should match pandas
     if np.issubdtype(dtype, np.floating):
         fill_value = np.nan
-    elif np.issubdtype(dtype, np.signedinteger):
-        # convert to floating point so NaN is valid
-        dtype = float
+    elif np.issubdtype(dtype, np.integer):
+        if dtype.itemsize <= 2:
+            dtype = np.float32
+        else:
+            dtype = np.float64
         fill_value = np.nan
     elif np.issubdtype(dtype, np.complexfloating):
         fill_value = np.nan + np.nan * 1j
--- a/xarray/core/dtypes.py
+++ b/xarray/core/dtypes.py
@@ -20,13 +20,13 @@ def maybe_promote(dtype):
     fill_value : Valid missing value for the promoted dtype.
     """
     # N.B. these casting rules should match pandas
-    if np.issubdtype(dtype, float):
+    if np.issubdtype(dtype, np.floating):
         fill_value = np.nan
-    elif np.issubdtype(dtype, int):
+    elif np.issubdtype(dtype, np.signedinteger):
         # convert to floating point so NaN is valid
         dtype = float
         fill_value = np.nan
-    elif np.issubdtype(dtype, complex):
+    elif np.issubdtype(dtype, np.complexfloating):
         fill_value = np.nan + np.nan * 1j
     elif np.issubdtype(dtype, np.datetime64):
         fill_value = np.datetime64('NaT')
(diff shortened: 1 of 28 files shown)


maxim-lian:
I think this is now passing throughout except for Python3.4, which I think is because pytest is on version [3.0.5](https://travis-ci.org/pydata/xarray/jobs/328818925#L1118), which doesn't have the `match` keyword from pytest 3.3.

Does anyone know why an old version of pytest is being installed there? From what I can see on the [pytest home page](https://docs.pytest.org/en/latest/getting-started.html), 3.4 is fully supported, and the [3.4 ci requirements](https://github.com/pydata/xarray/blob/master/ci/requirements-py34.yml#L5) don't force a lower version of pytest.

I could unravel that but if anyone has insight (and is anyone using 3.4? Though this PR prob not the place to make that decision)

I think we should try and merge this asap, as there are going to be (manageable) merge conflicts with other branches

maxim-lian:
> Are you going to make our tests more pytest-ized (e.g. by adopting more @pytest.mark.parametrize)?

Not yet, I think that's a bigger exercise, and more difficult to script

xarray/core/dtypes.py
```diff
@@ -20,13 +20,13 @@ def maybe_promote(dtype):
     fill_value : Valid missing value for the promoted dtype.
     """
     # N.B. these casting rules should match pandas
-    if np.issubdtype(dtype, float):
+    if np.issubdtype(dtype, np.floating):
         fill_value = np.nan
-    elif np.issubdtype(dtype, int):
+    elif np.issubdtype(dtype, np.signedinteger):
```
Zac-HD: This should also apply to unsigned integers, and the resulting dtype should be given as a numpy dtype.  Plus, we can make this much more efficient for small integers:

    elif np.issubdtype(dtype, np.integer):
        if dtype.itemsize <= 2:
            dtype = np.float32
        else:
            dtype = np.float64
        fill_value = np.nan

maxim-lian:
Tests now pass, including flake8, apart from the pytest.warns.

Given how much code this touches, I'm keen to merge this ASAP, before divergences creep in. What's the best course on 3.4? pandas is in the process of [dropping](https://github.com/pandas-dev/pandas/issues/15251) it. 

Alternatively I could revert the `pytest.warns` changes, though there isn't one clean commit to revert.

jhamman:
@maxim-lian - I opened https://github.com/pydata/xarray/issues/1829 to discuss dropping Python 3.4.

fujiisoup:
Using pip seems to temporary solve this issue.
ci/requirements-py34.yml
```
name: test_env
dependencies:
  - python=3.4
  - bottleneck
  - flake8
  - pandas
  - pip:
    - coveralls
    - pytest>=3.3
    - pytest-cov
```

maybe not the best solution, though.

fujiisoup:
There are two comment-outed lines using `self.assertEqual` in test_dataarray and test_dataset.

We are still using some methods of `TestCase` class, such as `assertVariableNotEqual`.
Can we add equivalent function to testings.py?

maxim-lian:
> There are two comment-outed lines using self.assertEqual in test_dataarray and test_dataset.

That are comment-out as part of this PR? Or prior? If prior, then I think we should leave until we attempt to add those back in

> We are still using some methods of TestCase class, such as assertVariableNotEqual.
Can we add equivalent function to testings.py?

I think the best way of handling those is xfailing an `assert_equal` test, rather than having a extra `-not` test for each variant. I also think we should do that in a separate PR (generally am keen to get this in before there are divergences)

maxim-lian:
> Using pip seems to temporary solve this issue.

Nice, thanks. I've put this in. If we're on the way to deprecate 3.4, from which source we install pytest is flexible IMHO

PR #1835, opened 2018-01-16
Add warning for netCDF4 bug

### LINKED ISSUE #1745: open_mfdataset() memory error in v0.10
#### Code Sample

```python
import xarray

ncfiles = '/example/path/to/wrf/netcdfs/*'
dropvars = ['list', 'of', 'many', 'vars', 'to', 'drop']

dset = xarray.open_mfdataset(ncfiles, drop_variables=dropvars, concat_dim='Time',  
                                                autoclose=True, decode_cf=False)
```
#### Problem description

I am trying to load 73 model (WRF) output files using ``open_mfdataset()``. (Thus, 'Time' is a new dimension). Each netcdf has dimensions ``{'x' : 405, 'y' : 282, 'z': 37}`` and roughly 20 variables (excluding the other ~20 in ``dropvars``). 

When I run the above code with **v0.9.6**, it completes in roughly 7 seconds. But with **v0.10**, it crashes with the following error:

``*** Error in `~/anaconda3/bin/python': corrupted size vs. prev_size: 0x0000560e9b6ca7b0 ***``

which, as I understand, means I'm exceeding my memory allocation. Any thoughts on what could be the source of this issue?

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.0-3-amd64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: C
LANG: C
LOCALE: None.None

xarray: 0.10.0
pandas: 0.20.3
numpy: 1.13.1
scipy: 0.19.1
netCDF4: 1.2.4
h5netcdf: 0.5.0
Nio: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.0
matplotlib: 2.0.2
cartopy: None
seaborn: 0.8.0
setuptools: 27.2.0
pip: 9.0.1
conda: 4.3.29
pytest: 3.1.3
IPython: 6.1.0
sphinx: 1.6.2

</details>

braaannigan:
Closes #1745 by adding a warning message for a bug in netCDF4.  The bug occurs when the filepath has 88 characters and the netCDF4 engine has been used with version less than 1.3.1.  

The warning message has been added at the first point where the length of the file path is known and the netcdf4 version number is known.  

I don't think there is any corresponding test, as the bug leads to a seg fault rather than a python output/exception.

Passes ``git diff upstream/master **/*py | flake8 --diff`` (remove if you did not edit any Python files)

Updated  `whats-new.rst`

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -5,7 +5,7 @@ import os.path
 from glob import glob
 from io import BytesIO
 from numbers import Number
-import warnings
+
 
 import numpy as np
 
@@ -281,14 +281,6 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
             engine = _get_default_engine(filename_or_obj,
                                          allow_remote=True)
         if engine == 'netcdf4':
-            import netCDF4
-            if len( filename_or_obj ) == 88 and float(netCDF4.__version__[:3]) < 1.3:
-                warnings.warn('\nA segmentation fault may occur when the   \n'
-                'file path has exactly 88 characters.  The issue is known  \n'
-                'to occur with version 1.2.4 of netCDF4 and can be  \n'
-                'addressed by upgrading netCDF4 to at least version 1.3.1.  \n'
-                'More details can be found here:  \n'
-                'https://github.com/pydata/xarray/issues/1745  \n')
             store = backends.NetCDF4DataStore.open(filename_or_obj,
                                                    group=group,
                                                    autoclose=autoclose)
@@ -321,7 +313,10 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     return maybe_decode_store(store)
 
 
-def open_dataarray(*args, **kwargs):
+def open_dataarray(filename_or_obj, group=None, decode_cf=True,
+                   mask_and_scale=True, decode_times=True, autoclose=False,
+                   concat_characters=True, decode_coords=True, engine=None,
+                   chunks=None, lock=None, cache=None, drop_variables=None):
     """Open an DataArray from a netCDF file containing a single data variable.
 
     This is designed to read netCDF files with only one data variable. If
@@ -402,7 +397,13 @@ def open_dataarray(*args, **kwargs):
     --------
     open_dataset
     """
-    dataset = open_dataset(*args, **kwargs)
+    dataset = open_dataset(filename_or_obj, group=group, decode_cf=decode_cf,
+                           mask_and_scale=mask_and_scale,
+                           decode_times=decode_times, autoclose=autoclose,
+                           concat_characters=concat_characters,
+                           decode_coords=decode_coords, engine=engine,
+                           chunks=chunks, lock=lock, cache=cache,
+                           drop_variables=drop_variables)
 
     if len(dataset.data_vars) != 1:
         raise ValueError('Given file dataset contains more than one data '
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -5,7 +5,7 @@ import os.path
 from glob import glob
 from io import BytesIO
 from numbers import Number
-
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


xarray/backends/api.py
```diff
@@ -281,6 +281,14 @@ def maybe_decode_store(store, lock=False):
             engine = _get_default_engine(filename_or_obj,
                                          allow_remote=True)
         if engine == 'netcdf4':
+            import netCDF4
+            if len( filename_or_obj ) == 88 and float(netCDF4.__version__[:3]) < 1.3:
```
shoyer: Use `LooseVersion()` here: https://stackoverflow.com/questions/11887762/compare-version-strings-in-python

shoyer: Please remove spaces around `filename_or_obj`, per pep8.

xarray/backends/api.py
```diff
@@ -281,6 +281,14 @@ def maybe_decode_store(store, lock=False):
...
+                warnings.warn('\nA segmentation fault may occur when the   \n'
+                'file path has exactly 88 characters.  The issue is known  \n'
+                'to occur with version 1.2.4 of netCDF4 and can be  \n'
+                'addressed by upgrading netCDF4 to at least version 1.3.1.  \n'
+                'More details can be found here:  \n'
+                'https://github.com/pydata/xarray/issues/1745  \n')
             store = backends.NetCDF4DataStore.open(filename_or_obj,
```
shoyer: Can you put this inside `NetCDF4DataStore.open()` instead? We try to keep the backend specific logic in the backend classes.

braaannigan:
Thanks for the advice @shoyer 

I've moved the warning to NetCDF4DataStore.open() and reverted backends/api.py back to the original.

I've used LooseVersion for the version comparison and made the pep8 change.

There is a problem with the test.  I've added the test to NetCDF4DataTest.  However, when the filename has 88 characters the seg fault occurs and the test gets stopped.  The output from
```
$py.test tests/test_backends.py
```
is just:
```
============================= test session starts ==============================
platform linux -- Python 3.6.2, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /my_python_path/xarray, inifile: pytest.ini
plugins: pylama-7.4.3
collected 1042 items

tests/test_backends.py .
*** Error in `/path_to/python': double free or corruption (!prev):  ***
Aborted (core dumped)
```

If the filename doesn't have 88 characters then the test result is "Failed: DID NOT WARN.", which is fine as it shouldn't trigger the warning.

I've been trying to see how people handle this issue, but haven't been able to find examples of people trying to trigger seg faults in a test yet

braaannigan:
There are some other small issues I need to deal with, I'll leave a message when I've sorted those out

shoyer:
@braaannigan How about we turn warnings into errors, and then catch the appropriate error instead? e.g.,
```python
    warnings.simplefilter("error")
    with raises_regex(Warning, 'segmentation fault'):
        ...
```
That should abort (due to the error) before the seg fault.

xarray/backends/netCDF4_.py
```diff
@@ -244,6 +245,16 @@ def __init__(self, netcdf4_dataset, mode='r', writer=None, opener=None,
     def open(cls, filename, mode='r', format='NETCDF4', group=None,
              writer=None, clobber=True, diskless=False, persist=False,
              autoclose=False):
+        import netCDF4 as nc4
+        from distutils.version import LooseVersion
```
shoyer: You can put this import at the top level. (We only import netCDF4 inside functions/methods to avoid the hard dependency, but disutils is part of Python's standard library.)

xarray/backends/api.py
```diff
@@ -281,6 +281,14 @@ def maybe_decode_store(store, lock=False):
             engine = _get_default_engine(filename_or_obj,
                                          allow_remote=True)
         if engine == 'netcdf4':
+            import netCDF4
```
shoyer: It looks like you accidentally kept your changes in this file?

braaannigan: I thought that I had checked out the original version again to replace all of my edits, but that doesn't seem to have worked.  I'm looking for a solution.

shoyer: You can simply revert this file (e.g., copy/paste from master) and then add another commit

braaannigan: I was worried that would cause some new unforeseen problem. I've commited this now.

shoyer:
Merged in 47c3b62432c27a8d10c167ffbea2c2dc86d0a1e5.

Thanks!

Issue #1142, opened 2016-11-29
win_type for rolling() ?


peterkamatej:
Hi, is there any plan to include support for various win_types such as "gaussian" to the xarray's rolling() methods like in pandas rolling()? I started using something like DataArray.to_pandas().rolling(...).. .to_xarray(), but in this way it's tricky to adhere to the xarray's flexibility when it comes to N-dimensionality..

shoyer:
There are no immediate plans that I know of, but I think we are certainly open to adding this functionality if someone wants to look into it.

cc @jhamman @MaximilianR

MaximilianR:
Think it would be useful generally. @jhamman will have a better view of how difficult it is to implement. Our use cases are fine with simple windows, although an `ewm` object would be helpful.

One API choice is to what extent we want `.rolling`, `.expanding` & `.ewm` to be separate objects vs kwargs to `.rolling`

jhamman:
Certainly open to adding this functionality.  Bottleneck isn't going to help so the code will all live in xarray. I think it makes sense to have a bit of a design discussion here prior to getting started. Questions:

#### 1. What window types are you interested in adding?  Pandas includes:

- boxcar
- triang
- blackman
- hamming
- bartlett
- parzen
- bohman
- blackmanharris
- nuttall
- barthann
- kaiser (needs beta)
- gaussian (needs std)
- general_gaussian (needs power, width)
- slepian (needs width)

#### 2. Can we maintain compatibility with pandas and bottleneck.

We have tried to maintain compatibility with both pandas and bottleneck in our rolling implementation.  This is proving somewhat difficult (e.g. #1046) but is a design consideration that we should keep in mind. 

Also, our current implementation falls back to operating on individual slices of the DataArray when bottleneck cannot be utilized. We should think a bit about the applicability of other windows when using other window types. Presumably, each `win_type` listed above would just provide a different set of weights to be associated with the members of each window.

#### 3. What about nd windows?

Our current implementation left open the possibility of n-dimensional windows.  While we haven't implemented it yet, the utility of the rolling object for a 2+ dimensional smoother would be quite a nice feature. That said, I'd be hesitant to implement anything on the rolling object that would not allow us to make this addition in the future.

peterkamatej:
Sorry for not replying sooner. So far it works fine for me when I switch to
pandas, use their gaussian rolling window, and then switch back to xarray.
As I'm in a hurry with something else now, I will get back to this
discussion a bit later.

[...]

Issue #819, opened 2016-04-06
N-D rolling


forman:
Dear xarray Team,

We just discovered xarray and it seems to be a fantastic candidate to serve as a core library for our climate data toolbox we are about to implement. While investigating the API we recognized that the `windows` kwargs in

```
DataArray.rolling(min_periods=None, center=False, **windows)
```

is limited to a single `dim=window_size` entry. Are there any plans to make it rolling in N-D?
This could be very useful for efficient gap filling, filtering or other methodologies that use grid cell neighbourhoods in multiple dimensions. 

Actually, I also asked myself why the `groupby` and `resample` methods don't take an N-D `dim` argument. This would allow for performing not only a temporal resampling but also a spatial resampling in the lat/lon plane or even a spatio-temporal resampling (including up- and downsampling in either dim).

Anyway, thanks for xarray!

Regards
    Norman

jhamman:
@forman - 

The main reason this isn't supported yet is that we haven't implemented it yet. I recently added the `Rolling` object to `xarray` and our initial application aimed to wrap the [`bottleneck` moving window functions](http://berkeleyanalytics.com/bottleneck/reference.html#bottleneck.move_mean). Because the bottleneck moving window functions only supports one windowing axis, we have also adopted that constraint for now. 

At this point, I don't know of any plans to extend this functionality although we could discuss further.  The complexity of N-D rolling aggregations does increase a fair bit and we would ideally like to let a lower level package (e.g. bottleneck) handle most of that. Since you seem to have a tangible application for the N-D rolling feature, maybe this is something you want to contribute to?

forman:
Thanks for the prompt reply!

Once we have decided to use xarray for our project(s) and once we familiarized with its internals, we'll be happy to contribute and support you! Currently we all feel a bit dizzy about the many options we have and how to decide which way to go: Create our own library using xarray or build on UK MetOffice's Iris, Apache OCW, or Max-Planck-Institute's CDO, etc.

magonser:
Hello together, 

additional to forman's comments I would like to note that for the xarray rolling operation a stride would be useful in some use-cases. I have understood, that no work is currently planned, but wanted to leave this remark for further contributions. 

I.e. 

    DataArray.rolling(min_periods=None, center=False, **windows)

with 

    windows={dimname: (size, stride=1)}

shoyer:
[...]

PR #1837, opened 2018-01-18
Rolling window with `as_strided`

### LINKED ISSUE #1831: Slow performance of rolling.reduce

```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: 
   ...: da = xr.DataArray(np.random.randn(1000, 100), dims=['x', 'y'],
   ...:                   coords={'x': np.arange(1000)})
   ...: 

In [2]: %%timeit
   ...: da.rolling(x=10).reduce(np.sum)
   ...: 
2.04 s ± 8.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
#### Problem description

In `DataArray.rolling`, we index by `.isel` method for every window, constructing huge number of `xr.DataArray` instances. This is very inefficient.

Of course, we can use bottleneck methods if available, but this provides only a limited functions.
(This also limits possible extensions of rolling, such as ND-rolling (#819), window type (#1142), strides (#819).)

I am wondering if we could skip any sanity checks in our `DataArray.isel -> Variable.isel` path in indexing.
Or can we directly construct a single large `DataArray` instead of a lot of small `DataArray`s?

fujiisoup:
- [x] Closes #1831, #1142, #819
 - [x] Tests added
 - [x] Tests passed

I started to work for refactoring rollings.
As suggested in [#1831 comment](https://github.com/pydata/xarray/issues/1831#issuecomment-357828636), I implemented `rolling_window` methods based on `as_strided`.

I got more than 1,000 times speed up! yey!

```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: 
   ...: da = xr.DataArray(np.random.randn(10000, 3), dims=['x', 'y'])
```
with the master
```python
%timeit da.rolling(x=5).reduce(np.mean)
1 loop, best of 3: 9.68 s per loop
```
with the current implementation
```python
%timeit da.rolling(x=5).reduce(np.mean)
100 loops, best of 3: 5.29 ms per loop
```
and with the bottleneck
```python
%timeit da.rolling(x=5).mean()
100 loops, best of 3: 2.62 ms per loop
```

My current concerns are 
+ Can we expose the new `rolling_window` method of `DataArray` and `Dataset` to the public?
I think this method itself is useful for many usecases, such as short-term-FFT and convolution.
This also gives more flexible rolling operation, such as windowed moving average, strided rolling, and ND-rolling.

+ Is there any dask's equivalence to numpy's `as_strided`?
Currently, I just use a slice->concatenate path, but I don't think it is very efficient.
(Is it already efficient, as dask utilizes out-of-core computation?)

Any thoughts are welcome.


xarray/core/dataarray.py
```diff
@@ -2132,6 +2132,50 @@ def rank(self, dim, pct=False, keep_attrs=False):
         ds = self._to_temp_dataset().rank(dim, pct=pct, keep_attrs=keep_attrs)
         return self._from_temp_dataset(ds)
 
+    def rolling_window(self, dim, window, window_dim, center=True):
```
shoyer: Can we make this a `Rolling` method instead? Maybe `construct` or `build` would be a good name for the method?

The API could look something like: `da.rolling(b=3).construct('window_dim')`

That helps keep the main DataArray/Dataset namespace a little more organized.

xarray/core/variable.py
```diff
@@ -936,6 +936,43 @@ def shift(self, **shifts):
...
+        ----------
+        **pad_width: keyword arguments of the form {dim: (before, after)}
+            Number of values padded to the edges of each dimension.
+        """
+        dtype, fill_value = dtypes.maybe_promote(self.dtype)
+
+        if isinstance(self.data, dask_array_type):
+            array = self.data
+
+            for d, pad in pad_widths.items():
```
shoyer: add a comment noting https://github.com/dask/dask/issues/1926 here

xarray/core/variable.py
```diff
@@ -936,6 +936,43 @@ def shift(self, **shifts):
             result = result._shift_one_dim(dim, count)
         return result
 
+    def _pad(self, **pad_widths):
```
shoyer: Call this `_pad_with_fill_value`?

xarray/core/indexing.py
```diff
@@ -815,6 +820,17 @@ def __setitem__(self, key, value):
         array, key = self._indexing_array_and_key(key)
         array[key] = value
 
+    def rolling_window(self, axis, window):
```
shoyer: Can we put this in `duck_array_ops` instead? I'd like to keep the indexing adapters simple.

xarray/core/indexing.py
```diff
@@ -815,6 +820,17 @@ def __setitem__(self, key, value):
...
+        """
+        Make an ndarray with a rolling window of axis-th dimension.
+        The rolling dimension will be placed at the last dimension.
+        """
+        
+        axis = nputils._validate_axis(self.array, axis)
+        rolling = nputils.rolling_window(np.swapaxes(self.array, axis, -1),
```
shoyer: maybe add an `axis` argument directly to `nputils.rolling_window`?

xarray/core/indexing.py
```diff
@@ -850,6 +866,25 @@ def __setitem__(self, key, value):
...
+            raise ValueError(
+                "`window` must be at least 1. Given : {}".format(window))
+        if window > self.array.shape[axis]:
+            raise ValueError("`window` is too long. Given : {}".format(window))
+
+        axis = nputils._validate_axis(self.array, axis)
+        size = self.array.shape[axis] - window + 1
+        arrays = [self.array[(slice(None), ) * axis + (slice(w, size + w), )]
+                  for w in range(window)]
+        return da.stack(arrays, axis=-1)
```
shoyer: Probably the most efficient way to do this would be to use dask's ghost cell support, but this is fine for now:
http://dask.pydata.org/en/latest/array-ghost.html

You would want to map the new efficient rolling window computation over each block.

jhamman: see here as well: https://github.com/pydata/xarray/blob/f3deb2f2495220af819021b199a5305b0d62ef36/xarray/core/dask_array_ops.py#L11-L25

xarray/core/dataarray.py
fujiisoup: How about `to_dataarray` and `to_dataset`?
Personally, `construct` sounds something necessary to compute rolling method.

fujiisoup:
During the work, I notice a somehow unexpected behavior of rolling.

I expected that with `min_periods=1` option, we will get an array without `nan`,
It is true with `center=False` and also `pd.rolling`.

```python
In [2]: da = xr.DataArray(np.arange(10), dims='x')
In [3]: da.rolling(x=3, min_periods=1, center=False).sum()
Out[3]: 
<xarray.DataArray (x: 10)>
array([  0.,   1.,   3.,   6.,   9.,  12.,  15.,  18.,  21.,  24.])

In [5]: s = pd.Series(np.arange(10))
 s.rolling(3, min_periods=1, center=True).sum()
Out[7]: 
0     1.0
1     3.0
2     6.0
3     9.0
4    12.0
5    15.0
6    18.0
7    21.0
8    24.0
9    17.0
dtype: float64
```
But with `center=True`, we have a `nan` at the end.
```
In [4]: da.rolling(x=3, min_periods=1, center=True).sum()
Out[4]: 
<xarray.DataArray (x: 10)>
array([  1.,   3.,   6.,   9.,  12.,  15.,  18.,  21.,  24.,  nan])
```

It is because we make `shift` operation after the rolling.
https://github.com/pydata/xarray/blob/74d8318c68be884134d449afad18dfe731d48b72/xarray/core/rolling.py#L263-L269

If we `pad` the array before the rolling operation instead of `shift`, we will not get the last `nan` and the result would be the same to pandas.
(`rolling.to_dataarray('window_dim')` does this).

I think this path is more intuitive.
Any thoughts?

xarray/core/dtypes.py
```diff
@@ -34,6 +34,8 @@ def maybe_promote(dtype):
         fill_value = np.datetime64('NaT')
     elif np.issubdtype(dtype, np.timedelta64):
         fill_value = np.timedelta64('NaT')
+    elif dtype.kind == 'b':
+        fill_value = False
```
fujiisoup: This is convenient for me, but it is not very clear whether `False` is equivalent to `nan` for boolean arrays.
If anyone has objections, I will consider different approach.

xarray/tests/test_dataarray.py
```diff
@@ -3376,12 +3388,9 @@ def test_rolling_pandas_compat(da, center, window, min_periods):
...
                             min_periods=min_periods).mean()
-    # pandas does some fancy stuff in the last position,
-    # we're not going to do that yet!
-    np.testing.assert_allclose(s_rolling.values[:-1],
-                               da_rolling.values[:-1])
-    np.testing.assert_allclose(s_rolling.index,
-                               da_rolling['index'])
+
+    np.testing.assert_allclose(s_rolling.values, da_rolling.values)
+    np.testing.assert_allclose(s_rolling.index, da_rolling['index'])
```
fujiisoup: I updated the logic for the `center=True` case. Now our result is equivalent to `pandas`'s rolling, including the last position.

xarray/core/variable.py
```diff
@@ -936,6 +936,43 @@ def shift(self, **shifts):
             result = result._shift_one_dim(dim, count)
         return result
 
+    def pad_with_fill_value(self, **pad_widths):
```
fujiisoup: I want to expose this to the public, which is used in my new logic in rolling.

xarray/core/indexing.py
fujiisoup: Thank you for the help.
But I don't yet come up with the solution for dask...
(Do you mean that we want to do `as_strided`-like operation for each ghosted-chunk?)

I think this could be in another PR, as it would take for along time for me to think of the correct path.

xarray/core/dtypes.py
shoyer: Indeed, let's consider other options here. This is used for the default value when reindexing/aligning.

xarray/core/rolling.py
```diff
@@ -203,26 +214,18 @@ def reduce(self, func, **kwargs):
...
-        result = concat(windows, dim=concat_dim)
-        # restore dim order
-        result = result.transpose(*self.obj.dims)
+        windows = self.to_dataarray('_rolling_window_dim')
+        result = windows.reduce(func, dim='_rolling_window_dim', **kwargs)
 
+        # Find valid windows based on count.
+        # We do not use `reduced.count()` because it constructs a larger array
+        # (notice that `windows` is just a view)
+        counts = (~self.obj.isnull()).rolling(
```
shoyer: For formatting long chains of method calls, I like to add extra parentheses and break every operation at the start of the line, e.g.,
```python
counts = ((~self.obj.isnull())
          .rolling(center=self.center, **{self.dim: self.window})
          .to_dataarray('_rolling_window_dim')
          .sum(dim='_rolling_window_dim'))
```
I find this makes it easier to read

xarray/core/rolling.py
```diff
@@ -203,26 +214,18 @@ def reduce(self, func, **kwargs):
...
+            center=self.center, **{self.dim: self.window}).to_dataarray(
+                '_rolling_window_dim').sum(dim='_rolling_window_dim')
         result = result.where(counts >= self._min_periods)
-
-        if self.center:
-            result = self._center_result(result)
-
-        return result
+        # restore dim order
+        return result.transpose(*self.obj.dims)
```
shoyer: I don't think we need to restore dimension order any more. The result should already be calculated correctly.

xarray/tests/test_dataarray.py
shoyer: Awesome, thanks!

doc/computation.rst
```diff
@@ -158,20 +158,31 @@ Aggregation and summary methods can be applied directly to the ``Rolling`` objec
...
-Finally, we can manually iterate through ``Rolling`` objects:
+We can also manually iterate through ``Rolling`` objects:
 
 .. ipython:: python
 
    @verbatim
    for label, arr_window in r:
       # arr_window is a view of x
 
+Finally, the rolling object has ``to_dataarray`` method, which gives a
```
shoyer: Maybe add: (`to_dataset` for Rolling objects from Dataset)

doc/whats-new.rst
```diff
@@ -27,6 +27,13 @@ Documentation
...
 Enhancements
 ~~~~~~~~~~~~
+- Improve :py:func:`~xarray.DataArray.rooling` logic for speed up.
+  :py:func:`~xarray.DataArrayRolling` object now support ``to_dataarray``
+  method that returns a view of the DataArray object with the rolling-window
+  dimension added to the last position. This enables more flexible operation,
+  such as strided rolling, windowed rolling, ND-rolling, and convolution.
+  (:issue:`1831`, :issue:`1142`, :issue:`819`)
+  By `Keisuke Fujii <https://github.com/fujiisoup>`_.
 - Added nodatavals attribute to DataArray when using :py:func:`~xarray.open_rasterio`. (:issue:`1736`).
```
shoyer: Add a bug fix note for the aggregations of the last element with `center=True`?

xarray/core/rolling.py
shoyer: Maybe we should add a short-cut here that doesn't bother to compute `counts` if the array's dtype cannot hold NaN? I think that would solve the issue with changing `maybe_promote` for booleans.

You could add a utility function to determine this based on whether the result of `maybe_promote()` has the same dtype as the input.

xarray/core/variable.py
```diff
@@ -936,6 +936,48 @@ def shift(self, **shifts):
...
+        """
+        Return a new Variable with paddings.
+
+        Parameters
+        ----------
+        **pad_width: keyword arguments of the form {dim: (before, after)}
+            Number of values padded to the edges of each dimension.
+        """
+        if self.dtype.kind == 'b':
```
fujiisoup: Is there a better way to get an appropriate fill_value for non-float arrays?

xarray/tests/test_dataarray.py
```diff
@@ -3435,7 +3448,7 @@ def test_rolling_count_correct():
     result = da.rolling(time=11, min_periods=None).count()
     expected = DataArray(
         [np.nan, np.nan, np.nan, np.nan, np.nan, np.nan,
-         np.nan, np.nan, np.nan, np.nan, 8], dims='time')
+         np.nan, np.nan, np.nan, np.nan, np.nan], dims='time')
```
fujiisoup: I think the last element should be `np.nan` rather than 8, because `8 < min_periods=11`.

xarray/core/ops.py
```diff
@@ -227,20 +227,8 @@ def func(self, *args, **kwargs):
...
-    if rolling.min_periods is None:
-        return rolling_count
-
-    # otherwise we need to filter out points where there aren't enough periods
-    # but not_null is False, and so the NaNs don't flow through
-    # array with points where there are enough values given min_periods
-    enough_periods = rolling_count >= rolling.min_periods
-
+    rolling_count = rolling._counts()
+    enough_periods = rolling_count > rolling._min_periods - 0.5
```
shoyer: 0.5 is a little strange to see when the other veggies are integers.

xarray/core/variable.py
shoyer: Maybe this function should take a fill value argument, which could default to `dtypes.NA`?

xarray/core/ops.py
shoyer: Other variables, not veggies (I blame autocorrect!)

shoyer:
Would it make sense to add a `stride` argument to `to_dataarray`? You can always index afterwards, but that could be a nice shortcut. It will be much faster for dask, certainly. (We can also add this later if desired)

xarray/core/ops.py
fujiisoup: Oops. I will fix.

doc/api.rst
```diff
@@ -467,6 +467,35 @@ DataArray methods
...
    DataArray.chunk
 
+Rolling objects
+===============
+
+.. autosummary::
+   :toctree: generated/
+
+   core.rolling.DataArrayRolling
+   core.rolling.DataArrayRolling.to_dataarray
```
fujiisoup: I want to add them (also groupby) to readthedocs.

fujiisoup:
I think this is ready for the final review, including docs.
I would like to add rolling methods to airspeed benchmark, but in another PR.

xarray/core/rolling.py
```diff
@@ -97,92 +95,114 @@ def __len__(self):
...
-
-    in ops.inject_bottleneck_rolling_methods.
-
-    After the injection, the Rolling object will have `name` (such as `mean` or
-    `median`) methods,
-    e.g. it enables the following call,
-    >>> data.rolling().mean()
+    def __init__(self, obj, min_periods=None, center=False, **windows):
+        """
+        Moving window object for DataArray.
```
shoyer: Let's add that you should use the `DataArray.rolling()` method to construct this object instead of the class constructor.

xarray/core/variable.py
```diff
@@ -936,6 +936,48 @@ def shift(self, **shifts):
...
+    def pad_with_fill_value(self, fill_value=None, **pad_widths):
+        """
+        Return a new Variable with paddings.
+
+        Parameters
+        ----------
+        **pad_width: keyword arguments of the form {dim: (before, after)}
+            Number of values padded to the edges of each dimension.
+        """
+        if fill_value is None:
```
shoyer: Use `dtypes.NA` as the default value here -- that's a sentinel value specifically designed for this use case. Otherwise you can't pad with `None` as a fill value.

xarray/core/rolling.py
```diff
@@ -97,92 +95,114 @@ def __len__(self):
...
-                                        coords=self.obj[self.dim].coords)
-
         self.window_indices = [slice(start, stop)
                                for start, stop in zip(starts, stops)]
 
-    def _center_result(self, result):
-        """center result"""
-        shift = (-self.window // 2) + 1
-        return result.shift(**{self.dim: shift})
+    def to_dataarray(self, window_dim, stride=None):
```
shoyer: Let's make the default stride `1` instead. It's equivalent to `stride=None` but more self-descriptive.

xarray/core/variable.py
```diff
@@ -1542,7 +1542,7 @@ def rolling_window(self, dim, window, window_dim, center=False):
         else:
             pads = (window - 1, 0)
 
-        fill_value = False if self.dtype.kind == 'b' else None
+        fill_value = False if self.dtype.kind == 'b' else dtypes.NA
```
shoyer: Maybe we should add the `fill_value` argument to this function?

I suspect special logic for boolean dypes would be surprising to many users.

xarray/core/rolling.py
```diff
@@ -228,6 +228,13 @@ def reduce(self, func, **kwargs):
         reduced : DataArray
             Array with summarized data.
         """
+        # Reduce functions usually assumes numeric type.
+        # For non-number array such as bool, We cast them to float
+        if self.obj.dtype.kind not in 'iufcm':
+            return DataArrayRolling(
+                self.obj.astype(float), center=self.center,
+                min_periods=self.min_periods,
+                **{self.dim: self.window}).reduce(func, **kwargs)
```
fujiisoup: I think boolean array rolling should work, such as
```python
array.isnull().rolling(dim=1).sum()
```
We are actually using this [here](https://github.com/pydata/xarray/blob/e31cf43e8d183c63474b2898a0776fda72abc82c/xarray/core/missing.py#L332).

So I cast non-number array to `float` here.

xarray/core/rolling.py
```diff
@@ -282,6 +278,12 @@ def wrapped_func(self, **kwargs):
             padded = self.obj.variable
             if self.center:
                 shift = (-self.window // 2) + 1
+
+                if (LooseVersion(np.__version__) < LooseVersion('1.13') and
```
fujiisoup: I notice that with the older numpy and benchmark, reducing of `np.nan`-and-boolean-mixed arrays gives a wrong result.
Here is a workaround, where we cast boolean arrays to float.

xarray/core/rolling.py
```diff
@@ -254,45 +275,41 @@ def wrapped_func(self, **kwargs):
...
 
-            if isinstance(self.obj.data, dask_array_type):
+            padded = self.obj.variable
+            if self.center:
+                shift = (-self.window // 2) + 1
+
+                if (LooseVersion(np.__version__) < LooseVersion('1.13') and
+                        self.obj.dtype.kind == 'b'):
+                    # with numpy < 1.13 bottleneck cannot handle np.nan-Boolean
+                    # mixed array correctly. We cast boolean array to float.
```
fujiisoup: I notice that with the older numpy and benchmark, reducing `np.nan`-and-boolean-mixed arrays gives a wrong result.
Here is a workaround, where we cast boolean arrays to float.

asv_bench/benchmarks/rolling.py
```diff
@@ -0,0 +1,48 @@
...
+time_rolling.params = (['mean', 'count'], [True, False])
+
+
+def time_rolling_np(window_, min_periods):
+    ds.rolling(x=window_, center=False, min_periods=min_periods).reduce(
+        getattr(np, 'nanmean'))
+
+
+time_rolling_np.param_names = ['window_', 'min_periods']
+time_rolling_np.params = ([20, 40], [5, None])
```
fujiisoup: I expected the memory usage keeps almost the same (excluding the padded region) 
even if we changed the window size.
But with a doubled window size, the memory usage increases almost twice.

Is it the nature of numpy's reducing function? 
Or I'm doing something wrong...

shoyer: I've experienced something similar before. I think NumPy's reducing functions may need to copy data copy into a C or Fortran contiguous array to do computations.

shoyer: Probably the best way to get a definitive answer would be to ask on the NumPy issue tracker or mailing list.

fujiisoup: It looks that the rolling-reduction with a bare `np.ndarray` does not consume so much memory.
I think the problem is in my code...

Here is a script I checked with `np.ndarray`
```python
In [1]: import numpy as np
   ...: 
   ...: def rolling_window(a, window):
   ...:     shape = a.shape[:-1] + (a.shape[-1] - window + 1, window)
   ...:     strides = a.strides + (a.strides[-1],)
   ...:     return np.lib.stride_tricks.as_strided(a, shape=shape, strides=strid
   ...: es)
   ...: 
   ...: a = np.arange(20000000, dtype=float).reshape(1000, 20000)
   ...: a_rolling1 = rolling_window(a, 100)
   ...: a_rolling2 = rolling_window(a, 300)
   ...: a_rolling1.shape, a_rolling2.shape
Out[1]: ((1000, 19901, 100), (1000, 19701, 300))

In [2]: from memory_profiler import memory_usage
   ...: memory_usage((np.mean, (a_rolling1, ), {'axis': 2}))
Out[2]: -> 355 MB

In [3]: memory_usage((np.mean, (a_rolling2, ), {'axis': 2}))
Out[3]:  -> 354 MB
```

fujiisoup: OK. Found. `np.nanmean` copies the data.

fujiisoup: They seem to copy the original array with `nan` replaced by 0.
I think we can avoid this to replace `nan` *before* making the rolling window.

fujiisoup:
Although nan-reduce methods (such as `nanmean`) might be sped up by replacing nan by a certain number before constructing a rolling window, this improvement brings another complexity into the code (and bottleneck already does this).

I want to finish this PR first and left such improvements to another chance.
Please let me know if anyone has comments of this PR.

asv_bench/benchmarks/rolling.py
```diff
@@ -6,43 +6,48 @@
...
+                    'x_coords': ('x', np.linspace(1.1, 2.1, nx))})
 
+    def time_rolling(self, func, center):
+        getattr(self.ds.rolling(x=window, center=center), func)()
 
-def time_rolling_np(window_, min_periods):
-    ds.rolling(x=window_, center=False, min_periods=min_periods).reduce(
-        getattr(np, 'nanmean'))
+    time_rolling.param_names = ['func', 'center']
+    time_rolling.params = (['mean', 'count'], [True, False])
```
shoyer: Maybe we should write our own `@parameterized` decorator, as suggested in https://github.com/airspeed-velocity/asv/issues/482?

xarray/core/rolling.py
```diff
@@ -202,27 +228,22 @@ def reduce(self, func, **kwargs):
...
-        # restore dim order
-        result = result.transpose(*self.obj.dims)
-
-        result = result.where(counts >= self._min_periods)
-
-        if self.center:
-            result = self._center_result(result)
-
-        return result
+        fill_value = dtypes.reduceable_fill_value(self.obj.dtype)
```
shoyer: I'm not sure this makes sense in the long term. This feels like a work-around for how we don't support `skipna=True` for object dtypes, which leads to an inconsistency with pandas:
```
>>> import xarray
>>> da = xarray.DataArray([True, False], [('x', [0, 1])]).reindex(x=[0, 1, 2])
>>> da
<xarray.DataArray (x: 3)>
array([True, False, nan], dtype=object)
Coordinates:
  * x        (x) int64 0 1 2

>>> da.to_series()
x
0     True
1    False
2      NaN
dtype: object

>>> da.to_series().sum()
1

>>> da.sum()
<xarray.DataArray ()>
array(nan)

>>> da.sum(skipna=True)
NotImplementedError: skipna=True not yet implemented for sum with dtype object
```

fujiisoup: I think `np.nansum(da.values)` works for `DataArray` in your example, but why do we decided not to support it for `object` dtype?
https://github.com/pydata/xarray/blob/009291139fde0c859ee565141cdb3b6a3d28cba0/xarray/core/duck_array_ops.py#L188-L192 
Also, do you know how pandas supports this?

shoyer: It looks like the problematic op was `nanmean`, though there might be others, too:
```
>>> np.nanmean(da.values)
AttributeError: 'int' object has no attribute 'dtype'
```

I suppose this is why pandas has its own nanmean implementation:
https://github.com/pandas-dev/pandas/blob/d3f7d2a666aa824e2df98083aa5c1fd9bb63252e/pandas/core/nanops.py#L356

fujiisoup: Thanks. 
A tentative workaround would be to support `skipna=True` for `nansum` (and `nanmax`, etc..), but not for `nanmean`.
I guess we can import pandas nan-reduce methods in the future.

xarray/core/duck_array_ops.py
```diff
@@ -229,26 +243,19 @@ def f(values, axis=None, skipna=None, **kwargs):
...
-argmin = _create_nan_agg_method('argmin', coerce_strings=True,
-                                support_object_type=True)
+argmax = _create_nan_agg_method('argmax', coerce_strings=True)
+argmin = _create_nan_agg_method('argmin', coerce_strings=True)
 max = _create_nan_agg_method('max', coerce_strings=True,
                              support_object_type=True)
 min = _create_nan_agg_method('min', coerce_strings=True,
                              support_object_type=True)
 sum = _create_nan_agg_method('sum', numeric_only=True,
                              support_object_type=True)
```
fujiisoup: Just a workaround for `.sum` to work with object-dtype.
I tried to generalize this to other ops, but it was not very easy because dask, bottleneck, and numpy work differently.
I will raise another issue for this but I want to finish this up with the smallest workaround.

fujiisoup:
Hmm...
The support of `nansum` for object-dtype is much more difficult than I expected.

Now, I am stacked in supporting numpy<1.13.
With numpy<1.13, `nansum` for the object type gives just a `nan`.

`nansum` scripts defined in numpy==1.13 does not help because `!=` operator used [here](https://github.com/numpy/numpy/blob/6914bb41f0fb3c1ba500bae4e7d671da9536786f/numpy/lib/nanfunctions.py#L68)
behaves differently in numpy<1.13.

I think our best choice is to revert this PR to 4189d71 and left this issue to #1866.

shoyer:
I'll take a look, but I'm open to only fixing this nansum for numpy 1.13+
(or even requiring numpy 1.13 for rolling window arrays), as long as we
don't break existing behavior on earlier versions of numpy.

fujiisoup:
I sent a separate PR (#1883) for nan-reduce methods.
I will continue / finish up this PR after #1883 is merged.

xarray/testing.py
```diff
@@ -140,3 +147,57 @@ def assert_allclose(a, b, rtol=1e-05, atol=1e-08, decode_bytes=True):
     else:
         raise TypeError('{} not supported by assertion comparison'
                         .format(type(a)))
+
+
+def assert_allclose_with_nan(a, b, rtol=1e-05, atol=1e-08, decode_bytes=True):
+    """Like assert_allclose, but except for nan.
```
shoyer: Can we consolidate this into `assert_allclose()`, like `numpy.testing.assert_allclose()`, which has an `equal_nan=True` argument?

xarray/core/duck_array_ops.py
```diff
@@ -213,7 +227,8 @@ def f(values, axis=None, skipna=None, **kwargs):
         with _ignore_warnings_if(using_numpy_nan_func):
             try:
                 return func(values, axis=axis, **kwargs)
-            except AttributeError:
+            except AttributeError as e:
+                print(e)
```
shoyer: delete

xarray/core/dask_array_ops.py
```diff
@@ -24,3 +25,43 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
     # trim array
     result = da.ghost.trim_internal(out, depth)
     return result
+
+
+def rolling_window(a, window, axis=-1):
+    """ Dask's equivalence to np.utils.rolling_window """
```
shoyer: very nice!

fujiisoup:
> What's the current status here vs #1883? Do you still want to merge that one first?

Yes. 
It looks I reverted to a wrong commit.
I will overwrite most modifications in `duck_array_ops` and `test_duck_array_ops` after #1883 is merged..

xarray/testing.py
fujiisoup: This will overwrite by #1883. Sorry for letting you review outdated code.

xarray/core/dask_array_ops.py
```diff
@@ -24,3 +25,44 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
...
+    # inputs for ghost
+    if axis < 0:
+        axis = a.ndim + axis
+    depth = {d: 0 for d in range(a.ndim)}
+    if window % 2 == 0:
+        depth[axis] = int((window - 1) / 2 + 1)
+        offset = 1
+    else:
+        depth[axis] = int((window - 1) / 2)
```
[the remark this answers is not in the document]
shoyer: we should indeed add these future imports:
```
```

xarray/tests/test_duck_array_ops.py
```diff
@@ -301,3 +308,17 @@ def test_argmin_max_error():
     da = construct_dataarray(2, np.bool_, contains_nan=True, dask=False)
     with pytest.raises(ValueError):
         da.argmin(dim='y')
+
+
+@requires_dask
+@pytest.mark.parametrize('axis', [0, -1])
+@pytest.mark.parametrize('window', [3, 8, 11])
+def test_dask_rolling(axis, window):
```
fujiisoup: It looks the test fails if the environment does not have dask.
It says 
```
../../../miniconda/envs/test_env/lib/python2.7/site-packages/_pytest/python.py:798: in parametrize
    self.function, name, arg))
E   ValueError: <function test_dask_rolling at 0x7f963326d050> uses no argument 'window'
```
Can we use both `@requires_dask` and `@pytest.mark.parametrize`?

jhamman: @fujiisoup - I've found that this doesn't always work. In some cases, I've reverted to using `importorskip` inside test functions.

shoyer: I think it might work if you make `@requires_dask` the inner-most decorator. But that's just a guess...

fujiisoup: Thanks. 

I noticed that moving this one to the inner-most decorator does not solve this issue.
I decided to use `pytest.skipif(has_dask, ...)` instead.

xarray/core/missing.py
```diff
@@ -329,4 +330,8 @@ def _get_valid_fill_mask(arr, dim, limit):
     '''helper function to determine values that can be filled when limit is not
     None'''
     kw = {dim: limit + 1}
-    return arr.isnull().rolling(min_periods=1, **kw).sum() <= limit
+    # we explicitly use to_dataarray method to avoid copy.
+    new_dim = rolling._get_new_dimname(arr.dims, '_window')
+    return (arr.isnull().rolling(min_periods=1, **kw)
+            .to_dataarray(new_dim, fill_value=False)
+            .sum(new_dim, skipna=False)) <= limit
```
fujiisoup: In my environment, this is faster than the original one with bottlneck (especially for small arrays)
`arr.isnull().rolling(min_periods=1, **kw).sum() <= limit`

xarray/testing.py
shoyer: Can we still move this into `assert_allclose()`? Most of the time, I think we do want to use `equal_nan=True`.

xarray/tests/test_variable.py
```diff
@@ -734,6 +734,51 @@ def test_getitem_error(self):
...
+        data = np.full_like(data, False, dtype=bool).reshape(4, 3, 2)
+        v = self.cls(['x', 'y', 'z'], data)
+        for xr_arg, np_arg in zip(xr_args, np_args):
+            actual = v.pad_with_fill_value(fill_value=False, **xr_arg)
+            expected = np.pad(np.array(v.data), np_arg,
+                              mode='constant', constant_values=False)
+            assert_array_equal(actual, expected)
+
+    def test_rolling_window(self):
+        # Just a working test. See test_nputils fot the algorithm validation
```
shoyer: fot -> for

shoyer:
My final concern is about the name for these methods. Currently, you have two separate method names, `to_dataset` and `to_dataarray()`. These names are very descriptive, but unfortunately they are different for DataArray and Dataset, which means that users will need to write different code for different xarray types. I would slightly rather that we had a single method name that could be used for both, perhaps `construct()` or `build()`. Then you could write code like `xarray_obj.rolling(time=100).construct(rolling_dim='window')` without needing to know whether `xarray_obj` is a Dataset or DataArray.

xarray/testing.py
fujiisoup: I noticed the current `assert_allclose` assumes `equal_nan=True` by default. I dropped `assert_allclose_with_nan`.

fujiisoup:
@shoyer , thanks for the review.

> These names are very descriptive, but unfortunately they are different for DataArray and Dataset, which means that users will need to write different code for different xarray types.

Agreed. Changed `to_dataarray / to_dataset` -> `construct`.

doc/computation.rst
```diff
@@ -158,20 +158,42 @@ Aggregation and summary methods can be applied directly to the ``Rolling`` objec
...
+view of the original ``DataArray`` with the windowed dimension attached to
+the last position.
+You can use this for more advanced rolling operations, such as strided rolling,
+windowed rolling, convolution, short-time FFT, etc.
+
+.. ipython:: python
+
+    rolling_da = r.construct('window_dim')
+    rolling_da
+    # rolling mean with 2-point stride
```
shoyer: Can we show the example by calling `construct('window_dim', stride=2)` instead?

doc/whats-new.rst
```diff
@@ -43,6 +43,13 @@ Documentation
 
 Enhancements
 ~~~~~~~~~~~~
+- Improve :py:func:`~xarray.DataArray.rooling` logic.
```
shoyer: rooling -> rolling

doc/whats-new.rst
```diff
@@ -43,6 +43,13 @@ Documentation
...
+  :py:func:`~xarray.DataArrayRolling` object now supports ``construct``
```
shoyer: `construct` ->

    :py:func:`~xarray.DataArrayRolling.construct`

xarray/core/dask_array_ops.py
```diff
@@ -24,3 +26,44 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
...
+
+    if depth[axis] > min(a.chunks[axis]):
+        raise ValueError(
+            "The window size %d is larger than your\n"
+            "smallest chunk size %d + 1. Rechunk your array\n"
+            "with a larger chunk size or a chunk size that\n"
+            "more evenly divides the shape of your array." %
+            (window, min(a.chunks[axis])))
+
+    boundary = {d: np.nan for d in range(a.ndim)}
```
shoyer: Should use this `fill_value` passed on from the Variable method instead of always being `np.nan`?

fujiisoup:
@shoyer , thanks for the detailed review.

I noticed the benchmark test is still failing. 
After fixing this, I will merge this. 

Thanks :)

xarray/core/dask_array_ops.py
```diff
@@ -24,3 +26,70 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
...
+
+    # pad the original array before the operation in order to avoid copying
+    # the output array (output array is just a view).
+    if center:
+        start = int(window / 2)  # 10 -> 5,  9 -> 4
+        end = window - 1 - start
+    else:
+        start, end = window - 1, 0
+
+    drop_size = depth[axis] - offset - np.maximum(start, end)
```
shoyer: Normally I think of `size` as a positive integer, but below you use `-drop_size` to make it positive. I think this would be clearer as `drop_size = max(start, end) - offset - depth[axis]` (use `max()` vs `np.maximum` as `start` and `end` are Python integers)

xarray/core/dask_array_ops.py
```diff
@@ -24,3 +26,70 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
...
+
+    if depth[axis] > min(a.chunks[axis]):
+        raise ValueError(
+            "For window size %d, every chunk should be larger than %d, "
+            "but the smallest chunk size is %d. Rechunk your array\n"
+            "with a larger chunk size or a chunk size that\n"
+            "more evenly divides the shape of your array." %
+            (window, depth[axis], min(a.chunks[axis])))
+
+    # We temporary use `reflect` boundary here, but the edge portion is
```
shoyer: No longer correct?

xarray/core/variable.py
```diff
@@ -1458,6 +1505,57 @@ def rank(self, dim, pct=False):
...
+               [[nan, nan, 4], [nan, 4, 5], [4, 5, 6], [5, 6, 7]]])
+
+        >>> v.rolling_window(x, 'b', 3, 'window_dim', center=True)
+        <xarray.Variable (a: 2, b: 4, window_dim: 3)>
+        array([[[nan, 0, 1], [0, 1, 2], [1, 2, 3], [2, 3, nan]],
+               [[nan, 4, 5], [4, 5, 6], [5, 6, 7], [6, 7, nan]]])
+        """
+        if fill_value is dtypes.NA:  # np.nan is passed
+            dtype, fill_value = dtypes.maybe_promote(self.dtype)
+            array = self.astype(dtype).data
```
shoyer: Use `self.data.astype(dtype, copy=False)` to avoid copying for numpy arrays unless necessary.

xarray/core/dask_array_ops.py
fujiisoup: You are right. I thought it becomes sometimes negative.
Fixed.

fujiisoup:
@shyer, do you have further suggestions?
I think this is almost ready.

jhamman:
nice work @fujiisoup!

xarray/core/variable.py
```diff
@@ -934,6 +934,53 @@ def shift(self, **shifts):
...
+        if fill_value is dtypes.NA:  # np.nan is passed
+            dtype, fill_value = dtypes.maybe_promote(self.dtype)
+        else:
+            dtype = self.dtype
+
+        if isinstance(self.data, dask_array_type):
+            array = self.data
+
+            # Dask does not yet support pad. We manually implement it.
+            # https://github.com/dask/dask/issues/1926
```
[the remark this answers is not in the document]
fujiisoup: Nice :)
Thanks for letting me know this.
I will update here after your PR is merged.

PR #1847, opened 2018-01-22
Use getitem_with_mask in reindex_variables

shoyer:
This is an internal refactor of reindexing/alignment to use `Variable.getitem_with_mask`.

As noted back in https://github.com/pydata/xarray/pull/1751#issuecomment-348380756, there is a nice improvement for alignment with dask (~100x improvement) but we are slower in several cases with NumPy (2-3x).

ASV results (smaller ratio is better):
```
    before     after       ratio
  [e31cf43e] [5830f2f8]
     4.85ms     4.86ms      1.00  reindexing.Reindex.time_1d_coarse
    98.15ms    98.97ms      1.01  reindexing.Reindex.time_1d_fine_all_found
+   96.88ms   210.71ms      2.17  reindexing.Reindex.time_1d_fine_some_missing
    24.47ms    25.18ms      1.03  reindexing.Reindex.time_2d_coarse
   433.26ms   437.19ms      1.01  reindexing.Reindex.time_2d_fine_all_found
+  245.20ms   711.36ms      2.90  reindexing.Reindex.time_2d_fine_some_missing
-   23.78ms    12.79ms      0.54  reindexing.Reindex.time_reindex_coarse
-  409.89ms   230.75ms      0.56  reindexing.Reindex.time_reindex_fine_all_found
+  233.41ms   369.48ms      1.58  reindexing.Reindex.time_reindex_fine_some_missing
    14.39ms    14.20ms      0.99  reindexing.ReindexDask.time_1d_coarse
   184.07ms   182.64ms      0.99  reindexing.ReindexDask.time_1d_fine_all_found
-     1.44s   277.03ms      0.19  reindexing.ReindexDask.time_1d_fine_some_missing
    95.49ms    94.49ms      0.99  reindexing.ReindexDask.time_2d_coarse
   910.11ms   916.47ms      1.01  reindexing.ReindexDask.time_2d_fine_all_found
     failed   997.33ms       n/a  reindexing.ReindexDask.time_2d_fine_some_missing
```
Note that `reindexing.ReindexDask.time_2d_fine_some_missing` timed out previously, which I think indicates that it took longer than 60 seconds.


asv_bench/asv.conf.json
```diff
@@ -63,7 +63,7 @@
         "netcdf4": [""],
         "scipy": [""],
         "bottleneck": ["", null],
-        "dask": ["", null],
+        "dask": [""],
```
shoyer: I don't think we need to run benchmarks separately with/without dask. This just makes the whole suite take twice as long as necessary. We already need to write separate benchmark cases, given that the syntax for using dask is different.

asv_bench/benchmarks/reindexing.py
```diff
@@ -0,0 +1,45 @@
...
+    def time_2d_coarse(self):
+        self.ds.reindex(x=np.arange(0, 100, 2), y=np.arange(0, 100, 2)).load()
+
+    def time_2d_fine_all_found(self):
+        self.ds.reindex(x=np.arange(0, 100, 0.5), y=np.arange(0, 100, 0.5),
+                        method='nearest').load()
+
+    def time_2d_fine_some_missing(self):
+        self.ds.reindex(x=np.arange(0, 100, 0.5), y=np.arange(0, 100, 0.5),
+                        method='nearest', tolerance=0.1).load()
```
jhamman: do we have unit test coverage for this test w/ dask? I see it was failing in the ASV benchmark before the changes in `alignment.py` so my guess is we don't have test coverage here.

shoyer: I'll double check but I think the test failure was just ASV timing out. When I run a similar test case before (see my comment on the other PR) it was extremely slow.

shoyer:
@fujiisoup Indeed, that seems to be the case. I made a fixed version of `duck_array_ops.where()`, relying on the new `dtypes.result_type()` function that's basically a fixed version of `np.result_type()`.

I believe this is ready for review now.

xarray/core/dtypes.py
```diff
@@ -60,3 +60,26 @@ def is_datetime_like(dtype):
     """
     return (np.issubdtype(dtype, np.datetime64) or
             np.issubdtype(dtype, np.timedelta64))
+
+
+def result_type(*arrays_and_dtypes):
+    """Like np.result_type, but number + string -> object (not string).
```
fujiisoup: I think we should also take care of bool-float pair
```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: 
   ...: x = xr.DataArray([True, False])
   ...: x.where(x)
   ...: 
Out[1]: 
<xarray.DataArray (dim_0: 2)>
array([ 1., nan])
```

shoyer:
I added fixes (bool, float) -> object and (str, bytes) -> object, and also applied these to `concatenate`/`stack` as well as `where` in `duck_array_ops`.

From a user-facing perspective, this now fixes dtypes `xarray.concat()` for mixed dtypes as well.

smithsp:
Mention https://github.com/gafusion/OMFIT-source/pull/2734.

shoyer:
@smithsp that link doesn't work. Did you mean to refer to something else?

shoyer:
@smithsp OK, I'd be happy to take a look if you're able to track it down into a self-contained example.

smithsp:
I think that the behavior is consistent with expected upcasting, (so no bug on your side), but we haven't figured out where we are using unicode vs string.

Issue #1771, opened 2017-12-09
Needs performance check / improvements in value assignment of DataArray


fujiisoup:
https://github.com/pydata/xarray/blob/5e801894886b2060efa8b28798780a91561a29fd/xarray/core/dataarray.py#L482-L489

In #1746, we added a validation in `xr.DataArray.__setitem__` whether the coordinates of array, key, and arrays are consistent.
In the current implementation, we call `xr.DataArray.__getitem__` to use the existing coordinate validation logic, but it does unnecessary indexing and it may decrease the `__setitem__` performance if the arrray is multidimensional.

We may need to optimize the logic here.

Is it reasonable to constantly monitor the performance of basic operations, such as `Dataset` construction, indexing, and assignment?
(or are these operations too light to make a performance monitor?)

cc @jhamman @shoyer

jhamman:
@fujiisoup in #1457, we added a framework ([Airspeed-velocity](https://github.com/airspeed-velocity/asv)) for benchmarking xarray operations. It is certainly within the scope of that framework to include indexing performance benchmarks. I just implemented a few IO related benchmarks with the expectation that more issues, like this one, would be added later on.

PR #1851, opened 2018-01-23
Indexing benchmarking

fujiisoup:
- [x] Relates to #1771 

Just added some benchmarks for basic, outer, and vectorized indexing and assignments.


.gitignore
```diff
@@ -34,6 +34,9 @@ nosetests.xml
 .cache
 .ropeproject/
 
+# asv environments
+.asv
```
fujiisoup: Is there any problem if `.asv` is added to `.gitignore`?
A bunch of files in this directory makes my gui-git-client freeze.

asv_bench/benchmarks/__init__.py
```diff
@@ -29,3 +29,12 @@ def randn(shape, frac_nan=None, chunks=None):
         x.flat[inds] = np.nan
 
     return x
+
+
+def randint(low, high=None, size=None, frac_minus=None):
```
shoyer: Can we use a seed for all these random numbers? That should decrease the variance for these test results.

asv_bench/benchmarks/indexing.py
```diff
@@ -0,0 +1,116 @@
...
+
+basic_assignment_values = [
+    xr.DataArray(randn((3, ny), frac_nan=0.1), dims=['x', 'y']),
+    xr.DataArray(randn(int(ny / 3) + 1, frac_nan=0.1), dims=['y']),
+    xr.DataArray(randn(int((nx - 6) / 3), frac_nan=0.1), dims=['x']),
+]
+
+
+def time_indexing_basic():
+    for ind in basic_indexes:
```
shoyer: Rather than a loop, another option would be to separate these cases into separate benchmarks. That would give more granular information.

asv_bench/benchmarks/indexing.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def time_assignment_vectorized():
+    tmp = ds.copy(deep=True)
+    for ind, val in zip(vectorized_indexes, vectorized_assignment_values):
+        tmp['var1'][ind.get('x', slice(None)), ind.get('y', slice(None))] = val
+
+
+try:
+    ds_dask = ds.chunk({'x': 100, 'y': 50, 't': 50})
```
shoyer: Another option is to use a subclass, like I did in https://github.com/pydata/xarray/pull/1847

asv_bench/benchmarks/indexing.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def time_indexing_basic_dask():
+        for ind in basic_indexes:
+            ds_dask.isel(**ind)
```
shoyer: It's probably a good idea to call `.load()` on these to load them into numpy. Otherwise it's only measure the time to construct the dask graph, not to evaluate it. Usually evaluation time dominates.

asv_bench/benchmarks/__init__.py
fujiisoup: See [line 9](https://github.com/pydata/xarray/pull/1851/files#diff-85a6ac883680893ebfac241d3ce89fe1L9).

jhamman was careful enough :)

fujiisoup:
Thanks. All done.

asv_bench/benchmarks/__init__.py
fujiisoup: >  it means that results will vary depending upon whether we run the full benchmark suite or not, and depending upon the order in which benchmark tests are run.

Thanks for the details.
Understood. 
Done.

fujiisoup:
It might be another issue, but I think we can have a clear link to our benchmark result page.
(maybe in our doc, README or in our PR template?)

jhamman:
@fujiisoup - I don't see any problem including a link in the docs (probably on the testing section) and/or on the readme. Its maybe too soon to include in the PR template since we just don't have very good coverage yet.

PR #1863, opened 2018-01-28
test decoding num_dates in float types

### LINKED ISSUE #1859: Time decoding has round-off error 0.10.0. Gone now.
Note: This problem occurs with version 0.10.0, but is gone when using current `master` (0.10.0+dev44.g0a0593d).

Here is a complete example:
https://gist.github.com/j08lue/34498cf17b176d15933e778278ba2921

#### Problem description

I have this time variable from a netCDF file:
```
float32 time(time)
    units: days since 1980-1-1 0:0:0
    standard_name: time
    calendar: gregorian
```
And when I open the file with `xr.open_dataset(..., decode_times=True)` the first time stamp becomes `1982-12-31T23:59:59.560122368`, while it should be `1983-01-01`. For reference, `netCDF4.num2date` gets it right.

I tracked the problem down to `xarray.conventions.decode_cf_datetime`. But then also noticed that you in https://github.com/pydata/xarray/commit/50b0a69a7aa0fb7ac3afb28e7bd971cf08055f99 made major changes to the decoding.

So I also tested with current `master` (0.10.0+dev44.g0a0593d) and the problem is gone. 

Up to you what you make of this. 😄 Maybe you can just close the issue.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 60 Stepping 3, GenuineIntel
byteorder: little
LC_ALL: None
LANG: ZZ
LOCALE: None.None

xarray: 0.10.0
pandas: 0.21.1
numpy: 1.13.3
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: None
bottleneck: 1.2.1
cyordereddict: 1.0.0
dask: 0.16.1
matplotlib: 2.1.1
cartopy: None
seaborn: None
setuptools: 38.2.4
pip: 9.0.1
conda: None
pytest: 3.3.1
IPython: 6.2.1
sphinx: None
In [4]:
</details>

j08lue:
- [x] Closes #1859
 - [x] try to find the origin of the difference in behaviour between `v0.10.0` and current base

### DIFF
--- a/xarray/tests/test_coding_times.py
+++ b/xarray/tests/test_coding_times.py
@@ -27,6 +27,8 @@ class TestDatetime(TestCase):
         import netCDF4 as nc4
         for num_dates, units in [
             (np.arange(10), 'days since 2000-01-01'),
+            (np.arange(10).astype('float64'), 'days since 2000-01-01'),
+            (np.arange(10).astype('float32'), 'days since 2000-01-01'),
             (np.arange(10).reshape(2, 5), 'days since 2000-01-01'),
             (12300 + np.arange(5), 'hours since 1680-01-01 00:00:00'),
             # here we add a couple minor formatting errors to test


j08lue:
For comparison, I added the same test to the `0.10.0` version in https://github.com/j08lue/xarray/tree/0p10-f4-time-decode, where it fails:

```
pytest -v xarray\tests\test_conventions.py

...

>               self.assertArrayEqual(expected, actual_cmp)
E               AssertionError:
E               Arrays are not equal
E
E               (mismatch 90.0%)
E                x: array([datetime.datetime(2000, 1, 1, 0, 0),
E                      datetime.datetime(2000, 1, 2, 0, 0),
E                      datetime.datetime(2000, 1, 3, 0, 0),...
E                y: array(['2000-01-01T00:00:00.000000', '2000-01-02T00:00:00.003211',
E                      '2000-01-03T00:00:00.006422', '2000-01-04T00:00:00.001245',
E                      '2000-01-05T00:00:00.012845', '2000-01-06T00:00:00.024444',...
```

j08lue:
I believe the issue originates in these lines:

https://github.com/pydata/xarray/blob/v0.10.0/xarray/conventions.py#L174-L175

where we multiply the `num_dates` with some `float` value and then cast to `int64`. 

If `num_dates` is `float32`, numpy keeps `float32` when multiplying with e.g. 1e9 and that somehow introduces an error. Here is a stripped version of the above:

```python
flat_num_dates = np.arange(100).astype('float32')
n = 1e9
roundtripped = (flat_num_dates * n).astype(np.int64) / n
assert np.all(flat_num_dates == roundtripped)
```

By the way, the factor has to be large, like `1e9`. E.g. `1e6` (`'ms since ...'`) won't give the error.

The weird thing is that the corresponding code in the current `master` is identical:

https://github.com/pydata/xarray/blob/50b0a69a7aa0fb7ac3afb28e7bd971cf08055f99/xarray/coding/times.py#L151-L152

I will look into why the result is still different from `v0.10.0`.

Also, if this really is the origin of the error, there are two easy ways to avoid this:

1. Cast `flat_num_dates` to `float64`: `(flat_num_dates.astype(np.float64) * n).astype(np.int64)`
2. Store `_NS_PER_TIME_DELTA` values as `int`, then `numpy` will do the casting.

j08lue:
> Store _NS_PER_TIME_DELTA values as int, then numpy will do the casting.

Haha, OK, this is actually what you implemented in https://github.com/pydata/xarray/commit/50b0a69a7aa0fb7ac3afb28e7bd971cf08055f99:

https://github.com/pydata/xarray/blob/50b0a69a7aa0fb7ac3afb28e7bd971cf08055f99/xarray/coding/times.py#L32-L37

So that is why it works now.

Case closed, I guess.

shoyer:
Nice, thank you @j08lue for tracking this down and adding tests!

PR #1869, opened 2018-01-30
Add '_FillValue' to set of valid_encodings for netCDF4 backend

### LINKED ISSUE #1865: Dimension Co-ordinates incorectly saving _FillValue attribute

```python
import xarray as xr
import pandas as pd
import numpy as np

temp = 273.15 + 25 * np.random.randn(2, 2)
lon = [0.0, 5.0]
lat = [10.0, 20.0]

ds = xr.Dataset({'temperature': (['lat', 'lon'], temp)},
                coords={'lat': lat,
                        'lon': lon})

ds['lat'].attrs = {
    'standard_name': 'latitude',
    'long_name': 'latitude',
    'units': 'degrees_north',
    'axis': 'Y'}
ds['lon'].attrs = {
    'standard_name': 'longitude',
    'long_name': 'longitude',
    'units': 'degrees_east',
    'axis': 'X'}
ds['temperature'].attrs = {
    'standard_name': 'air_temperature',
    'units': 'K'}
ds.attrs = {
    ('title', 'non-conforming CF 1.6 data produced by xarray 0.10'),
    ('Conventions', 'CF-1.6')}

ds.to_netcdf('/tmp/test.nc')

```
#### Problem description

According to the last sentence of the first paragraph of 2.5.1. *Missing data, valid and actual range of data* in NetCDF Climate and Forecast (CF) Metadata Conventions 1.7:

> Missing data is not allowed in coordinate variables.

When I use the [conformance checker](http://pumatest.nerc.ac.uk/cgi-bin/cf-checker-dev.pl?cfversion=auto) it issues an INFO message to this point for the co-ordinate variables.

Output of CF-Checker follows...

```
CHECKING` NetCDF FILE: /tmp/29428.nc
=====================
Using CF Checker Version 3.0.0
Checking against CF Version CF-1.6
Using Standard Name Table Version 48 (2017-11-28T15:32:48Z)
Using Area Type Table Version 6 (22 February 2017)


------------------
Checking variable: temperature
------------------

------------------
Checking variable: lat
------------------
INFO: attribute _FillValue is being used in a non-standard way

------------------
Checking variable: lon
------------------
INFO: attribute _FillValue is being used in a non-standard way

ERRORS detected: 0
WARNINGS given: 0
INFORMATION messages: 2
```

#### Expected Output

Co-ordinate variables should not store a _FillValue attribute

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 4.14.15-1-ARCH
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
xarray: 0.10.0
pandas: 0.22.0
numpy: 1.14.0
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.1
matplotlib: 2.1.2
cartopy: None
seaborn: None
setuptools: 38.4.0
pip: 9.0.1
conda: None
pytest: None
IPython: None
sphinx: None

</details>

czr137:
- [x] Closes #1865 (remove if there is no corresponding issue, which should only be the case for minor changes)
    - Relevant tests passed on my cloned version of the code. I had some skipped tests as I did not install rasterio, Nio, or Zarr. Four unrelated tests failed (2 in netcdf3 and 2 in h5netcdf)


shoyer:
Oops! Yes, this is was definitely a bug.

Can you add a unit test for this, so we don't have a regression?

Take a look at the test `test_explicitly_omit_fill_value ` added in https://github.com/pydata/xarray/pull/1659 and `test_encoding_same_dtype` below it for examples.

czr137:
Unit tests added in 9210dfa45428842c69b8d1835013159b1230c572. The first and the third added tests use the encoding kwarg and fail without b8fd86d5614d28254406d6f9b6d4e37383efc1be applied and pass with b8fd86d5614d28254406d6f9b6d4e37383efc1be applied, so they should catch a regression.

The second added test passes both before and after b8fd86d5614d28254406d6f9b6d4e37383efc1be and was added for completeness as it pairs with the original test_explicitly_omit_fill_value, but for a coordinate.

I still think nothing needs to change in the documentation as this now follows the intent of the design.

I will leave it to you @shoyer or your delegate to address whether anything needs to be stated in the changelog / what's new.

czr137:
Wait to merge... I'm still getting failed tests using the scipy backend

czr137:
Now the tests pass (see comment beside the checkbox)

Honestly, I'm not happy with where I put the fix in order to get the tests to pass when using the scipy backend by adding an additional test, but I didn't know where else to put the test. The test to see if the variable would be checked in prepare_variables came before the encoding could be managed in encode_cf_variable in conventions.py

czr137:
@shoyer, do you have an opinion on whether or not to include the scipy fix or leave it out of this PR?
Also what is your opinion on the documentation? I feel nothing is needed as with this PR the code will operate as the documentation states.

shoyer:
We don't need any additional documentation, but we should note this bug fix in "What's new"

xarray/backends/scipy_.py
```diff
@@ -188,8 +188,9 @@ def encode_variable(self, variable):
     def prepare_variable(self, name, variable, check_encoding=False,
                          unlimited_dims=None):
         if check_encoding and variable.encoding:
-            raise ValueError('unexpected encoding for scipy backend: %r'
-                             % list(variable.encoding))
+            if variable.encoding != {'_FillValue': None}:
```
shoyer: Yeah, this is a little awkward.

I think this is probably OK (backend classes already need to know about encodings), but it would be good to think about why we don't already have validation for encodings with scipy. We should probably be raising an error if somebody attempts an encoding like `{'zlib': True}` but right now it looks like we don't do that.

czr137:
@shoyer, I think everything is in order to apply this bug fix.

jhamman:
@shoyer - this looks good to me. At some point, we should formalize how backends validate their encodings but I don't think that needs to happen here.

shoyer:
Yes, thanks @czr137 !

PR #1872, opened 2018-01-31
added contributing guide

### LINKED ISSUE #640: Use pytest to simplify unit tests
xray's unit testing system uses Python's standard `unittest` framework.  [pytest](http://pytest.org/latest/) offers a more flexible framework requiring less boilerplate code.  I recently (#638) introduced pytest into xray's CI builds.  This issue proposes incrementally migrating and simplifying xray's unit testing framework to pytest.

jhamman:
- [x] Closes #640 

This is something we've talked about for a while and I'm capitalizing on a moment of inspiration.   Full disclosure, I've taken most of this from Pandas and edited it just where it makes sense for Xarray.

If others would like specific changes to this,  please comment only on `doc/contributing.rst`, ~`CONTRIBUTING.md` is auto generated with `pandoc`~. @pydata/xarray, feel free to push directly to this branch if there are larger edits you'd like to add.


doc/contributing.rst
```diff
@@ -0,0 +1,874 @@
...
+.. _contributiong.dev_python:
+
+Creating a Python Environment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before starting any development, you'll need to create an isolated xarray
+development environment:
+
+- Install either `Anaconda <https://www.anaconda.com/download/>`_ or `miniconda
+  <https://conda.io/miniconda.html>`_
```
maxim-lian: Is it worth mentioning that `pip` is an acceptable alternative?

doc/contributing.rst
```diff
@@ -0,0 +1,874 @@
...
+See the full conda docs `here <http://conda.pydata.org/docs>`__.
+
+Creating a branch
+-----------------
+
+You want your master branch to reflect only production-ready code, so create a
+feature branch for making your changes. For example::
+
+    git branch shiny-new-feature
+    git checkout -b shiny-new-feature
```
maxim-lian: I think there's an extra `-b` here? You have the combined command below

doc/contributing.rst
```diff
@@ -0,0 +1,874 @@
...
+   git diff master -u -- "*.py" | flake8 --diff
+
+This command will catch any stylistic errors in your changes specifically, but
+be beware it may not catch all of them. For example, if you delete the only
+usage of an imported function, it is stylistically incorrect to import an
+unused function. However, style-checking the diff will not catch this because
+the actual import is not part of the diff. Thus, for completeness, you should
+run this command, though it will take longer::
+
+   git diff master --name-only -- "*.py" | grep "xarray/" | xargs -r flake8
```
maxim-lian: Running `flake8 xarray` takes 5.7s on my system, so that's may be the right slower-but-more-complete command

doc/contributing.rst
```diff
@@ -0,0 +1,874 @@
...
+Then you will need to push the branch (see below) forcefully to replace the current
+commits with the new ones::
+
+    git push origin shiny-new-feature -f
+
+.. note::
+
+   GitHub now supports squashing commits when merging pull requests. Unless
+   there is a specific reason for squashing/rebasing your feature branch, we
+   recommend letting Github do the work.
```
maxim-lian: Given this note, is the section less relevant? Would have been v useful prior to the GH change

maxim-lian:
This is superb! I haven't done a comparison with the pandas docs, but from memory this feels much nicer. Great work!

doc/contributing.rst
jhamman: I think we can add that later if/when there is interest. The requirements files in ci/ are formatted for conda so I suggest we leave this as is for now.

doc/contributing.rst
jhamman: I moved the `.. Note` to the top of this section. I think having some instructions on how/why to rebase is still useful so I've left the majority of this section in tact.

shoyer: My preference would be to not encourage rebasing. In my experience, it can be a source of confusion for many new contributor. Also, from a review perspective GitHub makes it easier to see incremental changes if you only append commits. A rebase in a large PR means I have to start review over from scratch.

doc/contributing.rst
maxim-lian: 👍 

Tangentially, are we the only ppl who don't use conda?

jhamman: Perhaps. Personally, conda has become pretty baked into my workflow but I know that is not the case for everyone. 

If you want to push a description of how you setup your development environment, I think that would be a welcome addition here.

fmaussion: Personally I lost faith in conda some time ago when the [sky fell](https://github.com/conda-forge/pyproj-feedstock/issues/10). Since wheels are available for most packages, installing with pip is (almost) as easy as with conda nowadays.

fmaussion:
I don't think we need to have both the ``.md`` and ``.rst`` versions. I've written something very similar for OGGM recently and github recognises the ``contributing.rst`` in the docs as well. [Demo](https://github.com/OGGM/oggm/issues/new)

doc/contributing.rst
maxim-lian: @jhamman interesting. TBC, I didn't mean that I had any edge here. Fewer options are better - I vote to leave conda only.

Any hardcore pippers can deal with the issues themselves

doc/contributing.rst
```diff
@@ -0,0 +1,812 @@
...
+through the `GitHub "issues" tab <https://github.com/pydata/xarray/issues>`_
+to find issues that interest you. There are a number of issues listed under
+`Documentation <https://github.com/pydata/xarray/issues?q=is%3Aissue+is%3Aopen+label%3Adocumentation>`_
+and `good first issue
+<https://github.com/pydata/xarray/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>`_
+where you could start out. Once you've found an interesting issue, you can
+return here to get your development environment setup.
+
+Feel free to ask questions on the `mailing list
+<https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!forum/xarray>`_ or on `Gitter`_.
```
fujiisoup: We previously dropped a gitter support in #1673 (or just removed a link from README), due to our limited resource.
I think we can remve this.

jhamman:
Anymore comments here, if not, I'll merge at COB on Monday.

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+.. _contributing.forking:
+
+Forking
+-------
+
+You will need your own fork to work on the code. Go to the `xarray project
+page <https://github.com/pydata/xarray>`_ and hit the ``Fork`` button. You will
+want to clone your fork to your machine::
+
+    git clone https://github.com/your-user-name/xarray.git xarray-yourname
```
shoyer: Let's simplify this by removing `xarray-yourname`

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+    cd xarray-yourname
```
shoyer: This should just be `cd xarray`

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+    git remote add upstream https://github.com/pydata/xarray.git
+
+This creates the directory `xarray-yourname` and connects your repository to
```
shoyer: likewise, `xarray-yourname` -> `xarray`

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+
+   # Build and install xarray
+   python setup.py develop
+
+At this point you should be able to import xarray from your locally built version::
+
+   $ python  # start an interpreter
+   >>> import xarray
+   >>> print(xarray.__version__)
+   0.22.0.dev0+29.g4ad6d4d74
```
shoyer: maybe we should fix this? :)

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+and report benchmarks that changed by more than 10%.
+The command uses ``conda`` by default for creating the benchmark
+environments. If you want to use virtualenv instead, write::
+
+    asv continuous -f 1.1 -E virtualenv upstream/master HEAD
+
+The ``-E virtualenv`` option should be added to all ``asv`` commands
+that run benchmarks. The default value is defined in ``asv.conf.json``.
+
+Running the full test suite can take up to one hour and use up to 3GB of RAM.
```
shoyer: "test suite" -> "benchmark suite"

Can it really take a full hour for xarray? I suspect it's currently much faster, just a few minutes in my experience

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+submitting a pull request.
+
+For more, see the `pytest <http://doc.pytest.org/en/latest/>`_ documentation.
+
+Running the performance test suite
+----------------------------------
+
+Performance matters and it is worth considering whether your code has introduced
+performance regressions.  *xarray* is in the process of migrating to
+`asv benchmarks <https://github.com/spacetelescope/asv>`__
```
shoyer: Instead of saying that we're in the process of migrating to asv, say that we're starting to write asv tests. I don't think we're migrating from anything else :)

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+#. Navigate to your repository on GitHub
+#. Click on the ``Pull Request`` button
+#. You can then click on ``Commits`` and ``Files Changed`` to make sure everything looks
+   okay one last time
+#. Write a description of your changes in the ``Preview Discussion`` tab
+#. Click ``Send Pull Request``.
+
+This request then goes to the repository maintainers, and they will review
+the code. If you need to make more changes, you can make them in
+your branch, push them to GitHub, and the pull request will be automatically
```
shoyer: Let's add "add them to a new commit"

doc/contributing.rst
```diff
@@ -0,0 +1,809 @@
...
+updated.  Pushing them to GitHub again is done by::
+
+    git push -f origin shiny-new-feature
```
shoyer: Let's remove `-f` here. I'd rather discourage force-pushing.

jhamman:
Thanks @shoyer for the review. I've addressed each of your comments in ffb44dd.

maxim-lian:
💯  @jhamman

doc/index.rst
```diff
@@ -53,6 +52,8 @@ Documentation
    plotting
    api
    internals
+   contributing
+   whats-new
```
shoyer: I noticed that this change moved `whats-new` to the end. I'm not sure this is a good idea.

It's true that the details of whats-new is unlikely to be relevant to new users. But I like putting it first because I think it's something most users will want to see:
1. New users will check it to see if the project is well-maintained.
2. Existing users want to know what's new.

shoyer: That said... Google analytics shows "What's new" ranking pretty lowly in popularity:
https://docs.google.com/spreadsheets/d/1F6IZ0aSQlX7yZDZbS2R9RZN9PAt70Dl_g60BM5agDIc/edit?usp=sharing

fmaussion: Still, I think your argument is valid. People won't return to a "what's new" page as often as they do for the other docs, but the page shows that things happen on the project

jhamman: I think my thinking was that new users are probably best served by getting to the why xarray/tutorial/examples pages before. I'm fine moving it back if we think it is worth it.

Issue #1789, opened 2017-12-18
Broken link to github on ReadTheDocs 


JiaweiZhuang:
The "Edit on GitHub" button on [ReadTheDocs](http://xarray.pydata.org/en/stable/) is linked to
https://github.com/pydata/xarray/blob/origin/stable/doc/index.rst
which doesn't exist. The correct URL would be without "origin":
https://github.com/pydata/xarray/blob/stable/doc/index.rst

shoyer:
Thanks... I don't know if these links have ever worked properly for us. If anyone can figure out how to fix them, a PR would be appreciated!

raybellwaves:
I started to have a look at this. 

I needed to install `rasterio` to build the docs but unfortunately the `conda` installation isn't working (see https://github.com/conda-forge/rasterio-feedstock/issues/53) nor is the `pip` installation (see https://github.com/mapbox/rasterio/issues/1244). 

Anyone  who uses a Mac and mostly installs via `conda` have any advice installing rasterio?

fmaussion:
I'm not sure it'll help, but have you tried creating the environment with the same pinned versions as we use for building the docs? i.e:

```
conda env create -n xarray-docs -f /path/to/xarray/doc/environment.yml
```

shoyer:
I have also had trouble getting a rasterio build working on OS X.

To be honest, I have been commenting out the bit in the docs config that sets up sphinx gallery when testing locally:
https://github.com/pydata/xarray/blob/5a28b89d8f32a16d8529d6514c04992b5ee7a349/doc/conf.py#L56 
It would be nice if a lack of rasterio did not fail the docs build.

shoyer:
It looks like `expected_failing_examples` would take care of it:
https://sphinx-gallery.github.io/advanced_configuration.html#don-t-fail-the-build-on-exit

Zac-HD:
Related to rtfd/readthedocs.org#1820, I think.  Searching reveals that there have been a variety of problems with the 'Edit on Github' link over time.

shoyer:
Interesting. Reading that link, I see that RTD now automatically creates
the "stable" tag. I've still been pushing manually to a branch! I guess the
problem manifests the same way, though.

Issue #1880, opened 2018-02-02
Should imshow() recognise 0-255 images?


fmaussion:
#### Code Sample, a copy-pastable example if possible

```python
import os
import urllib.request
import xarray as xr
import matplotlib.pyplot as plt

# Download the file from rasterio's repository
url = 'https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif'
urllib.request.urlretrieve(url, 'RGB.byte.tif')

# Read the data
da = xr.open_rasterio('RGB.byte.tif')

f, (ax1, ax2) = plt.subplots(1, 2, figsize=(9, 4))
da.plot.imshow(ax=ax1)
(da / 255).plot.imshow(ax=ax2)
plt.tight_layout()
plt.show()

# Delete the file
os.remove('RGB.byte.tif')
```

![figure_1](https://user-images.githubusercontent.com/10050469/35730809-c157c7ba-0813-11e8-8909-24b6d61e685c.png)

#### Problem description

In https://github.com/pydata/xarray/pull/1796, @Zac-HD added support for RGBA images. If an alpha channel is not found, it is added ([code](https://github.com/pydata/xarray/blob/master/xarray/plot/plot.py#L708))

The problem is that adding this alpha channel requires the images to be normalized to 0-1, while plotting an image in 0-255 range without alpha channel works fine in matplotlib. Removing https://github.com/pydata/xarray/blob/master/xarray/plot/plot.py#L708-L715 would solve the problem, but I guess it was added for a reason. 

@Zac-HD , thoughts?

Zac-HD:
Ouch, this is definitely a bug in the code I wrote - that's terrible 😭 

Probably related to matplotlib/matplotlib/pull/10220, where I'm looking at clipping instead of wrapping colours from the other side.

Fixing this issue *may* be as simple as setting alpha to 255 instead of 1 for integer dtypes, or it might be more involved.  Either way I'll try to open a pull within the next few days.

PR #1881, opened 2018-02-02
Fix rasterio example in docs

fmaussion:
https://github.com/pydata/xarray/pull/1796 introduced a bug in the doc gallery. This PR reverts the code to the previous greyscale example and adds a new case using imshow (the use case is different, as I tried to explain in the descriptions).

I also took care of https://github.com/pydata/xarray/issues/1789#issuecomment-356068358 : the docs should now build even when rasterio is not installed.

cc @Zac-HD , @shoyer

### DIFF
--- a/doc/gallery/plot_rasterio_rgb.py
+++ b/doc/gallery/plot_rasterio_rgb.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""
+.. _recipes.rasterio_rgb:
+
+============================
+imshow() and map projections
+============================
+
+Using rasterio's projection information for more accurate plots.
+
+This example extends :ref:`recipes.rasterio` and plots the image in the
+original map projection instead of relying on pcolormesh and a map
+transformation.
+"""
+
+import os
+import urllib.request
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+
+# Download the file from rasterio's repository
+url = 'https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif'
+urllib.request.urlretrieve(url, 'RGB.byte.tif')
+
+# Read the data
+da = xr.open_rasterio('RGB.byte.tif')
+
+# Normalize the image
+da = da / 255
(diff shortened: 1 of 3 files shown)


jhamman:
LGTM. Thanks @fmaussion. I had noticed this was broken but hadn't figured out why yet.

doc/gallery/plot_rasterio_rgb.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Download the file from rasterio's repository
+url = 'https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif'
+urllib.request.urlretrieve(url, 'RGB.byte.tif')
+
+# Read the data
+da = xr.open_rasterio('RGB.byte.tif')
+
+# Normalize the image
+da = da / 255
```
shoyer: I don't think this is strictly necessary with RGB support in imshow, though I suppose it doesn't hurt.

doc/gallery/plot_rasterio_rgb.py
```diff
@@ -0,0 +1,44 @@
...
+
+# The data is in UTM projection
+# Until https://github.com/SciTools/cartopy/issues/813 is implemented
+# we have to do this manually
+crs = ccrs.UTM('18N')
+
+# Plot on a map
+ax = plt.subplot(projection=crs)
+da.plot.imshow(ax=ax, transform=crs)
```
shoyer: For pedagogical purposes, it might be nice to explicitly write `rga='band'`.

doc/gallery/plot_rasterio_rgb.py
fmaussion: That's unfortunately necessary, see https://github.com/pydata/xarray/issues/1880

doc/gallery/plot_rasterio_rgb.py
fmaussion: yes, will do

shoyer:
OK, LGTM

fmaussion:
Thanks!

PR #1883, opened 2018-02-02
Support nan-ops for object-typed arrays

### LINKED ISSUE #1866: aggregation ops for object-dtype are missing
This issue arises in [#1837 comment](https://github.com/pydata/xarray/pull/1837#discussion_r163999738), where we need to make a summation of object-dtype array, such as
```python
xr.DataArray(np.array([True, True, False, np.nan], dtype=object), dims='x').sum('x', skipna=True)
```
Currently, it raises NotImplementedError.

pandas support this by having their own nan-aggregation methods.

fujiisoup:
- [x] Closes #1866
 - [x] Tests passed

I am working to add aggregation ops for object-typed arrays, which may make #1837 cleaner.
I added some tests but maybe not sufficient.
Any other cases which should be considered?
e.g. `[True, 3.0, np.nan]` etc...


xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,87 @@ def _ignore_warnings_if(condition):
...
+    data = _dask_or_eager_func(func)(value, axis=axis, **kwargs)
+    # dask seems return non-integer
+    if isinstance(value, dask_array_type):
+        data = data.astype(int)
+    if not hasattr(data, 'dtype'):  # scalar case
+        # TODO should we raise ValueError if all-nan slice encountered?
+        data = -1 if valid_count == 0 else int(data)
+        return np.array(data)  # return 0d-array
+    # convert all nan part axis to nan
+    return where_method(data, valid_count != 0, -1)
```
fujiisoup: In numpy, `nanargmin` raise ValueError if there are all-NaN slice/axis.
Do we follow this?
Considering our `getitem_with_mask` method, I think it is consistent to return -1 for such a case, 
but it could be confusing sometimes.

doc/whats-new.rst
```diff
@@ -41,7 +41,16 @@ Documentation
 
 Enhancements
 ~~~~~~~~~~~~
-- reduce methods such as :py:func:`DataArray.sum()` now accepts ``dtype``
+- Reduce methods such as :py:func:`DataArray.sum()` now handles object-type array.
+
+  .. ipython:: python
+
+    da = xray.DataArray(np.array([True, False, np.nan], dtype=object), dims='x')
```
maxim-lian: `xr` / `xarray`?

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,79 @@ def _ignore_warnings_if(condition):
...
+        data = -1 if valid_count == 0 else int(data)
+        return np.array(data)
+    return where_method(data, valid_count != 0, -1)
+
+
+def _nanmean_ddof(ddof, value, axis=None, **kwargs):
+    """ In house nanmean. ddof argument will be used in _nanvar method """
+    valid_count = count(value, axis=axis)
+    value = fillna(value, 0.0)
+    # TODO numpy's mean does not support object-type array, so we assume float
```
shoyer: Why not use `sum` in this function and simply divide by `valid_count - ddof` instead of rescaling the mean?

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,79 @@ def _ignore_warnings_if(condition):
         yield
 
 
+def _nansum(value, axis=None, **kwargs):
```
shoyer: Can we give these functions a more explicit names like `_nansum_object`?

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,79 @@ def _ignore_warnings_if(condition):
...
+    return _dask_or_eager_func('sum')(value, axis=axis, **kwargs)
+
+
+def _nan_minmax(func, fill_value, value, axis=None, **kwargs):
+    """ In house nanmin and nanmax for object array """
+    valid_count = count(value, axis=axis)
+    filled_value = fillna(value, fill_value)
+    data = _dask_or_eager_func(func)(filled_value, axis=axis, **kwargs)
+    if not hasattr(data, 'dtype'):  # scalar case
+        data = np.nan if valid_count == 0 else data
```
shoyer: Instead passing in the `fill_value`, let's figure it out from `dtypes.maybe_promote()`

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,79 @@ def _ignore_warnings_if(condition):
...
+    kwargs_mean.pop('keepdims', None)
+    value_mean = _nanmean_ddof(ddof=0, value=value, axis=axis, keepdims=True,
+                               **kwargs_mean)
+    squared = _dask_or_eager_func('square')(value.astype(value_mean.dtype) -
+                                            value_mean)
+    return _nanmean_ddof(ddof, squared, axis=axis, **kwargs)
+
+
+_nan_funcs = {'sum': _nansum,
+              'min': partial(_nan_minmax, 'min', np.inf),
```
shoyer: I am concerned that this will break on arrays of strings. On Python 2, the code probably works (but gives an incorrect result), but on Python 3 `np.inf > 'abc'` raises a TypeError.

Given that these are only used for object arrays, maybe we should use special objects for this instead, e.g.,
```python
@functools.total_ordering
class AlwaysLessThan(object):
    def __lt__(self, other):
        return True
    def __eq__(self, other):
        return isinstance(other, type(self))
```

We should probably also add some unit tests for object arrays of strings/NaN (probably do this first!). Currently these raise an error, but I think this code could fix them:
```
>>> xr.DataArray(np.array([np.nan, 'foo'], dtype=object)).min()
TypeError: '<=' not supported between instances of 'float' and 'str'
```

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,79 @@ def _ignore_warnings_if(condition):
...
+    return where_method(data, valid_count != 0)
+
+
+def _nanvar(value, axis=None, **kwargs):
+    ddof = kwargs.pop('ddof', 0)
+    kwargs_mean = kwargs.copy()
+    kwargs_mean.pop('keepdims', None)
+    value_mean = _nanmean_ddof(ddof=0, value=value, axis=axis, keepdims=True,
+                               **kwargs_mean)
+    squared = _dask_or_eager_func('square')(value.astype(value_mean.dtype) -
```
shoyer: You could potentially just use the operator `** 2` instead of the dask_or_eager_func here.

xarray/core/duck_array_ops.py
shoyer: You could potentially copy at least part of the implementation from NumPy's own `mean`:
https://github.com/numpy/numpy/blob/e06d3614182e7b97d5e0d90291642027d147744b/numpy/core/_methods.py#L53

fujiisoup: > Why not use sum in this function and simply divide by valid_count - ddof instead of rescaling the mean?

I feel pretty dumb now. Updated.

xarray/core/duck_array_ops.py
```diff
@@ -205,22 +206,17 @@ def _nan_argminmax_object(func, fill_value_typ, value, axis=None, **kwargs):
...
     """ In house nanmean. ddof argument will be used in _nanvar method """
     valid_count = count(value, axis=axis)
-    value = fillna(value, 0.0)
-    # TODO numpy's mean does not support object-type array, so we assume float
-    dtype = kwargs.get('dtype', None)
+    value = fillna(value, 0)
+    # As dtype inference is impossible for object dtype, we assume float
+    dtype = kwargs.pop('dtype', None)
     if dtype is None and value.dtype.kind == 'O':
         dtype = value.dtype if value.dtype.kind in ['cf'] else float
```
fujiisoup: Is there a good workaround to infer the output dtype of the object-typed array?
We need to pass this to dask for the next division but `dtype=object` is not allowed.

xarray/core/duck_array_ops.py
fujiisoup: Done

xarray/core/duck_array_ops.py
fujiisoup: Done

xarray/tests/test_duck_array_ops.py
```diff
@@ -109,3 +113,158 @@ def test_types(self, val1, val2, val3, null):
...
+
+    # without nan
+    da = construct_dataarray(dim_num, dtype, contains_nan=False, dask=dask)
+    actual = getattr(da, func)(skipna=skipna)
+    expected = getattr(np, 'nan{}'.format(func))(da.values)
+    assert np.allclose(actual.values, np.array(expected), rtol=rtol)
+
+
+@pytest.mark.parametrize('dim_num', [1, 2])
+@pytest.mark.parametrize('dtype', [float, int, np.float32, np.bool_, str])
```
fujiisoup: Added a test for str-type

xarray/core/dtypes.py
```diff
@@ -40,7 +64,7 @@ def maybe_promote(dtype):
     return np.dtype(dtype), fill_value
 
 
-def get_fill_value(dtype):
+def get_fill_value(dtype, fill_value_typ=None):
```
shoyer: Can we make separate functions for this, maybe `get_pos_infinity` and `get_neg_infinity`? It feels a little strange to put it all in one function, and with separate functions you can avoid the need to validate the `fill_value_typ` argument.

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,76 @@ def _ignore_warnings_if(condition):
...
+    type """
+    fill_value = dtypes.get_fill_value(value.dtype, fill_value_typ)
+    valid_count = count(value, axis=axis)
+    value = fillna(value, fill_value)
+    data = _dask_or_eager_func(func)(value, axis=axis, **kwargs)
+    # dask seems return non-integer type
+    if isinstance(value, dask_array_type):
+        data = data.astype(int)
+    if not hasattr(data, 'dtype'):  # scalar case
+        # TODO should we raise ValueError if all-nan slice encountered?
```
shoyer: For consistency with `nanargmin()`, we probably should still raise `ValueError('All-NaN slice encountered')` for now. `-1` would make sense, but it would need to documented. NaN could also make sense, but would not be so useful since floats are not valid indexers.

xarray/core/duck_array_ops.py
shoyer: Is this fixed by your dask PR https://github.com/dask/dask/pull/3137?

If so, can we maybe say this requires using the latest dask release?

xarray/core/duck_array_ops.py
```diff
@@ -171,6 +171,83 @@ def _ignore_warnings_if(condition):
...
 
+def _nansum_object(value, axis=None, **kwargs):
+    """ In house nansum for object array """
+    value = fillna(value, 0)
+    return _dask_or_eager_func('sum')(value, axis=axis, **kwargs)
+
+
+def _nan_minmax_object(func, fill_value_typ, value, axis=None, **kwargs):
+    """ In house nanmin and nanmax for object array """
+    if fill_value_typ == '+inf':
```
shoyer: Nit: Instead of passing a separate string, we might just pass the function to make the fill value directly (`dtypes.get_pos_infinity` or `dtypes.get_neg_infinity`).

That would let us drop these conditionals and error prone string matching.

xarray/core/duck_array_ops.py
```diff
@@ -197,6 +197,79 @@ def _ignore_warnings_if(condition):
...
+
+def _nanmean_ddof_object(ddof, value, axis=None, **kwargs):
+    """ In house nanmean. ddof argument will be used in _nanvar method """
+    valid_count = count(value, axis=axis)
+    value = fillna(value, 0)
+    # As dtype inference is impossible for object dtype, we assume float
+    # https://github.com/dask/dask/issues/3162
+    dtype = kwargs.pop('dtype', None)
+    if dtype is None and value.dtype.kind == 'O':
+        dtype = value.dtype if value.dtype.kind in ['cf'] else float
```
DimitriPapadopoulos, line 243: Wasn't the intent `value.dtype.kind in "cf"` instead of `value.dtype.kind in ["cf"]`?

In any case, this is a no-op because `value.dtype.kind == "O"`. So remove the test altogether?

```suggestion
    if dtype is None and value.dtype.kind == 'O':
        dtype =  float
```

I am attempting to fix this https://github.com/pydata/xarray/pull/10465#discussion_r2173717231.

PR #1885, opened 2018-02-03
Raise when pcolormesh coordinate is not sorted

### LINKED ISSUE #1852: bug: 2D pcolormesh plots are wrong when coordinate is not ascending order

```python
import numpy as np
import xarray as xr


x = np.arange(10)
y = np.arange(20)

np.random.shuffle(x)

x = xr.DataArray(x, dims=['x'], coords={'x': x})
y = xr.DataArray(y, dims=['y'], coords={'y': y})

z = x + y
z_sorted = z.isel(x=np.argsort(x.values))


# make plot
fig, axs= plt.subplots(1, 2, figsize=(6,3))
z_sorted.plot(ax=axs[0])
axs[0].set_title("X is sorted")

z.plot(ax=axs[1])
axs[1].set_title("X is not unsorted")
plt.tight_layout()
```
#### Problem description

Sometime the coordinates in an xarray dataset are not always sorted in ascending order. I recently had an issue where the time coordinate of a 2D datasets was scrambled, so calling `x.plot` gave very strange results. In my opinion, `x.plot` should probably sort the data along the coordinates, or at least provide a warning if the coordinates are unsorted.

#### Expected Output
Here is the image generated by the snippet above:

image

The left and right panels should be the same.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Darwin
OS-release: 16.0.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0+dev50.ga988dc2
pandas: 0.20.3
numpy: 1.13.1
scipy: 0.19.1
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.15.2
distributed: 1.18.3
matplotlib: 2.0.2
cartopy: None
seaborn: 0.8.0
setuptools: 36.5.0.post20170921
pip: 9.0.1
conda: 4.3.29
pytest: 3.2.1
IPython: 6.1.0
sphinx: 1.6.3
</details>

nbren12:
- [x] Closes #1852 (remove if there is no corresponding issue, which should only be the case for minor changes)

I added a simple warning to `_infer_interval_breaks` in `xarray/plot/plot.py`. The warning does not currently say the name of the coordinate, because that would requiring introducing a new function or potentially passing a name argument, which seems overly complicated for such a small dit. Hopefully, this isn't a problem because the user can easily figure out which coordinate is not sorted by process of elimination.

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -779,7 +779,9 @@ def _infer_interval_breaks(coord, axis=0):
         raise ValueError("The input coordinate is not sorted in increasing "
                          "order along axis %d. This can lead to unexpected "
                          "results. Consider calling the `sortby` method on "
-                         "the input DataArray." % axis)
+                         "the input DataArray. To plot data with categorical "
+                         "axes, consider using the `heatmap` function from "
+                         "the `seaborn` statistical plotting library." % axis)
 
     deltas = 0.5 * np.diff(coord, axis=axis)
     if deltas.size == 0:
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -758,8 +758,10 @@ def _is_monotonic(coord, axis=0):
         return True
     else:
         n = coord.shape[axis]
-        delta_pos = coord.take(np.arange(1, n), axis=axis) >= coord.take(np.arange(0, n-1), axis=axis)
-        delta_neg = coord.take(np.arange(1, n), axis=axis) <= coord.take(np.arange(0, n-1), axis=axis)
+        delta_pos = (coord.take(np.arange(1, n), axis=axis)
+                     >= coord.take(np.arange(0, n-1), axis=axis))
+        delta_neg = (coord.take(np.arange(1, n), axis=axis)
+                     <= coord.take(np.arange(0, n-1), axis=axis))
         return np.all(delta_pos) or np.all(delta_neg)
 
 
@@ -774,10 +776,10 @@ def _infer_interval_breaks(coord, axis=0):
     coord = np.asarray(coord)
 
     if not _is_monotonic(coord, axis=axis):
-        warnings.warn("The input coordinate is not sorted in increasing order "
-                      "along axis %d. This can lead to unexpected results. "
-                      "Consider calling the `sortby` method on the input "
-                      "DataArray." % axis)
+        raise ValueError("The input coordinate is not sorted in increasing "
+                         "order along axis %d. This can lead to unexpected "
+                         "results. Consider calling the `sortby` method on "
+                         "the input DataArray." % axis)
 
     deltas = 0.5 * np.diff(coord, axis=axis)
     if deltas.size == 0:
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -759,7 +759,8 @@ def _is_monotonic(coord, axis=0):
     else:
         n = coord.shape[axis]
         delta_pos = coord.take(np.arange(1, n), axis=axis) >= coord.take(np.arange(0, n-1), axis=axis)
-        return np.all(delta_pos) or np.all(~delta_pos)
+        delta_neg = coord.take(np.arange(1, n), axis=axis) <= coord.take(np.arange(0, n-1), axis=axis)
+        return np.all(delta_pos) or np.all(delta_neg)
 
 
 def _infer_interval_breaks(coord, axis=0):
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -757,9 +757,9 @@ def _is_monotonic(coord, axis=0):
     if coord.shape[axis] < 3:
         return True
     else:
-        delta = np.diff(coord, axis=axis)
-        # This ensures all the deltas have the same sign
-        return np.abs(delta).sum() == np.abs(delta.sum())
+        n = coord.shape[axis]
+        delta_pos = coord.take(np.arange(1, n), axis=axis) >= coord.take(np.arange(0, n-1), axis=axis)
+        return np.all(delta_pos) or np.all(~delta_pos)
 
 
 def _infer_interval_breaks(coord, axis=0):
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -745,6 +745,23 @@ def contourf(x, y, z, ax, **kwargs):
     return primitive
 
 
+def _is_monotonic(coord, axis=0):
+    """
+    >>> _is_monotonic(np.array([0, 1, 2]))
+    True
+    >>> _is_monotonic(np.array([2, 1, 0]))
+    True
+    >>> _is_monotonic(np.array([0, 2, 1]))
+    False
+    """
+    if coord.shape[axis] < 3:
+        return True
+    else:
+        delta = np.diff(coord, axis=axis)
+        # This ensures all the deltas have the same sign
+        return np.abs(delta).sum() == np.abs(delta.sum())
+
+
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks a lot @nbren12 ! Indeed `_infer_interval_breaks` is probably the right place to add this warning.

A few comments:
- the tests are now failing, likely because of a type issue with time coordinates
- coordinates don't have to be ascending for pcolormesh to work (see example below). Monotonically ascending and descending coordinates should work.
- we need a test for this new behavior ([asserting warnings](https://docs.pytest.org/en/3.0.2/recwarn.html))
- we follow the pep8 conventions, and I think your lines are longer than the 79 character rule

```python
import numpy as np
import xarray as xr
x = np.arange(5)[::-1]
y = np.arange(6)[::-1]
x = xr.DataArray(x, dims=['x'], coords={'x': x})
y = xr.DataArray(y, dims=['y'], coords={'y': y})
z = x + y
z.plot();
```

index

nbren12:
Thanks for your help @fmaussion. Hopefully the commit I just pushed fixes the failing tests. I will work on adding a test and fixing the formatting.

nbren12:
Ok. I think everything is ready now.

xarray/plot/plot.py
```diff
@@ -750,6 +767,13 @@ def _infer_interval_breaks(coord, axis=0):
            [ 2.5,  3.5,  4.5]])
     """
     coord = np.asarray(coord)
+
+    if not _is_monotonic(coord, axis=axis):
+        warnings.warn("The input coordinate is not sorted in increasing order "
```
shoyer: Rather an a warning, why not make this an error? I don't see any use-cases for 2d plots with non-monotonic coordinates. With the current version of xarray, these plots always end up wrong in some way, either by not plotting everything or with bad axis labels.

nbren12:
@shoyer That would work with me. Is there any chance people would want to
make heatmaps involving categorical variables though?

If we do decide to raise an error, why not go one step further and just
sort the coordinates automatically?

shoyer:
> Is there any chance people would want to make heatmaps involving categorical variables though?

Certainly, but the currently version of 2D plots in xarray doesn't work for these!

If they want to make plots involving categorical variables, I would currently recommend using `seaborn.heatmap` (we can mention it in the error message, too).

> If we do decide to raise an error, why not go one step further and just
sort the coordinates automatically?

One reason is that it's not obvious if they would like increasing or decreasing coordinates.

nbren12:
>  One reason is that it's not obvious if they would like increasing or decreasing coordinates.

For me at least, `plt.pcolormesh` automatically displays in increasing order even if one of the input arrays is sorted in descending order. This happens all the time with meteorological data available in pressure coordinates (pressure goes down with height). I usually have to manually call plt.gca().invert_yaxis() to flip the y axis. Does `xarray.plot`'s behave differently?

xarray/plot/plot.py
```diff
@@ -741,6 +741,23 @@ def contourf(x, y, z, ax, **kwargs):
...
+    True
+    >>> _is_monotonic(np.array([0, 2, 1]))
+    False
+    """
+    if coord.shape[axis] < 3:
+        return True
+    else:
+        delta = np.diff(coord, axis=axis)
+        # This ensures all the deltas have the same sign
+        return np.abs(delta).sum() == np.abs(delta.sum())
```
fmaussion: Is this robust wrt floating point accuracy?

nbren12:
That’s a good point. Probably not.

fmaussion:
Which, I guess, illustrates our point quite well: while it's easy for users to find out when their data is messy, it's much harder to find a robust and logical way to handle all cases for a general library like xarray. I guess it's the reason why matplotlib also doesn't warn or error when unordered data is given to pcolormesh.

Don't get me wrong: I think it would be valuable to be able to provide a better "service" than matplotlib. Probably we should even error earlier, at the 2d plot level (and not only at the pcolormesh level like it is now). It's just going to be a bit hard because of all the special cases (dtypes, 2d coords, and map projections moslty).

Ordering "per default" might be the most straightforward way, but:
- what about performance issues?
- I don't like the idea of silently doing this with user data

I'll let @shoyer decide on what's best here. I am slightly in favor of doing nothing and giving the responsibility to the user.

nbren12:
There. I think this code should work for all dtypes and 2D coords.

fmaussion:
Thanks! This looks good.

nbren12:
Yah. I knew my solution before was probably too cute.

fmaussion:
Sorry @nbren12 there is still a pep8 error somewhere, and you can freely make it an error as @shoyer suggests.

Note that currently the test for monotonic coords  is for pcolormesh only, while it could be for all 2d plots as well. What I like with the current implementation is that the check happens in `_infer_interval_breaks`, which is only called when certain conditions are met (1D coords, or 2D coords without transform kwarg). If this check lives in the common 2d function the same conditions should be applied.

If we start to check input for sanity we might as well raise an error when coordinates are not regularly spaced in the imshow case, although I can imagine this might be more difficult to check.

Thoughts?

nbren12:
> Sorry @nbren12 there is still a pep8 error somewhere, and you can freely make it an error as @shoyer suggests.

Hopefully, the commit I just pushed fixes this. I had some global flake8 settings that were messing with  my local linting.

> Note that currently the test for monotonic coords is for `pcolormesh` only, while it could be for all 2d plots as well.

I don't think I was having any issues with `contour` or `contourf` when I first opened the issue, but it probably does break `imshow`. 

> If we start to check input for sanity we might as well raise an error when coordinates are not regularly spaced in the imshow case.

Maybe we could leave this to a later PR.

xarray/plot/plot.py
```diff
@@ -741,6 +741,26 @@ def contourf(x, y, z, ax, **kwargs):
...
+    True
+    >>> _is_monotonic(np.array([0, 2, 1]))
+    False
+    """
+    if coord.shape[axis] < 3:
+        return True
+    else:
+        n = coord.shape[axis]
+        delta_pos = (coord.take(np.arange(1, n), axis=axis)
+                     >= coord.take(np.arange(0, n-1), axis=axis))
```
fmaussion: To make flake8 happy, the operator should be on the first line, i.e.:

```python
delta_pos = (coord.take(np.arange(1, n), axis=axis) >= 
             coord.take(np.arange(0, n-1), axis=axis))
```

xarray/plot/plot.py
```diff
@@ -741,6 +741,26 @@ def contourf(x, y, z, ax, **kwargs):
...
+        delta_neg = (coord.take(np.arange(1, n), axis=axis)
+                     <= coord.take(np.arange(0, n-1), axis=axis))
```
fmaussion: Here too

fmaussion:
Sorry, our new flake8 tool can be quite strict ;)

xarray/plot/plot.py
nbren12: ok. my last commit should fix this. sry, i can't seem to get my flake8 to do the same things as travis.

nbren12:
no problem. I have always preferred putting operators on new lines, though. Didn't realize that was against pep8.  oh well 🤷‍♂️ .

nbren12:
I just rebased onto master.

shoyer:
Thanks @nbren12 !

nbren12:
cool! Thanks

PR #1893, opened 2018-02-07
Use correct dtype for RGB image alpha channel

### LINKED ISSUE #1880: Should imshow() recognise 0-255 images?

Zac-HD:
- [x] Closes #1880
 - [x] Fully documented (bugfix for earlier change, no additional note)

The cause of the bug in #1880 was that I had forgotten to specify the dtype when creating an alpha channel, and therefore concatenating it cast the all the data to float64.  I've fixed that, corrected the alpha value for integer arrays, and avoided a pointless copy to save memory.

### DIFF
--- a/doc/gallery/plot_rasterio_rgb.py
+++ b/doc/gallery/plot_rasterio_rgb.py
@@ -26,9 +26,6 @@ urllib.request.urlretrieve(url, 'RGB.byte.tif')
 # Read the data
 da = xr.open_rasterio('RGB.byte.tif')
 
-# Normalize the image
-da = da / 255
-
 # The data is in UTM projection. We have to set it manually until
 # https://github.com/SciTools/cartopy/issues/813 is implemented
 crs = ccrs.UTM('18N')
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -710,8 +710,12 @@ def imshow(x, y, z, ax, **kwargs):
         # missing data transparent.  We therefore add an alpha channel if
         # there isn't one, and set it to transparent where data is masked.
         if z.shape[-1] == 3:
-            z = np.ma.concatenate((z, np.ma.ones(z.shape[:2] + (1,))), 2)
-        z = z.copy()
+            alpha = np.ma.ones(z.shape[:2] + (1,), dtype=z.dtype)
+            if np.issubdtype(z.dtype, np.integer):
+                alpha *= 255
+            z = np.ma.concatenate((z, alpha), 2)
+        else:
+            z = z.copy()
         z[np.any(z.mask, axis=-1), -1] = 0
 
     primitive = ax.imshow(z, **defaults)
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -713,7 +713,7 @@ def imshow(x, y, z, ax, **kwargs):
             alpha = np.ma.ones(z.shape[:2] + (1,), dtype=z.dtype)
             if np.issubdtype(z.dtype, np.integer):
                 alpha *= 255
-            z = np.ma.concatenate((z, alpha), 2)
+            z = np.ma.concatenate((z, alpha), axis=2)
         else:
             z = z.copy()
         z[np.any(z.mask, axis=-1), -1] = 0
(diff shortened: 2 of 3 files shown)


Zac-HD:
Ping @fmaussion

fmaussion:
Thanks @Zac-HD ! Again (and sorry for the hassle ;), this needs a regression test. Also, could you change the rasterio example in the docs and see if your change indeed makes the normalisation obsolete? (https://github.com/pydata/xarray/blob/master/doc/gallery/plot_rasterio_rgb.py#L27)

doc/gallery/plot_rasterio_rgb.py
```diff
@@ -26,11 +26,9 @@
 # Read the data
 da = xr.open_rasterio('RGB.byte.tif')
 
-# Normalize the image
-da = da / 255
-
 # The data is in UTM projection. We have to set it manually until
-# https://github.com/SciTools/cartopy/issues/813 is implemented
+# https://github.com/SciTools/cartopy/issues/813 is implemented, so that
+# the coastlines are drawn in the same coordinate system as the image.
```
fmaussion: This is not exactly what I meant. Right now we *can* draw the coastlines in the same coordinate system as the image. What we need https://github.com/SciTools/cartopy/issues/813 for is a tool to convert the PROJ4 string (available as attribute ``da.crs``) to a ``cartopy.Proj`` instance (see https://github.com/SciTools/cartopy/issues/813#issuecomment-355051930). To be honest I don't understand why Cartopy made the choice to create their own new projection classes with their own kwargs instead of relying on an existing system, but this is another topic...

fmaussion:
Appart from the comment in the gallery example and the flake8 error, this looks good to me! Thanks @Zac-HD !

doc/gallery/plot_rasterio_rgb.py
Zac-HD: Right, I've just reverted this comment - I don't understand it well enough to improve the explanation, and it's not really part of this PR.

xarray/plot/plot.py
```diff
@@ -710,8 +710,12 @@ def imshow(x, y, z, ax, **kwargs):
         # missing data transparent.  We therefore add an alpha channel if
         # there isn't one, and set it to transparent where data is masked.
         if z.shape[-1] == 3:
-            z = np.ma.concatenate((z, np.ma.ones(z.shape[:2] + (1,))), 2)
-        z = z.copy()
+            alpha = np.ma.ones(z.shape[:2] + (1,), dtype=z.dtype)
+            if np.issubdtype(z.dtype, np.integer):
+                alpha *= 255
+            z = np.ma.concatenate((z, alpha), 2)
```
shoyer: nit: can you use `axis=2` here, for clarity? (I loathe undocumented literals)

fmaussion:
The failing tests in appveyor should be fixed by https://github.com/pydata/xarray/pull/1904

fmaussion:
Thanks @Zac-HD !

Issue #1725, opened 2017-11-17
repr in backend arrays


fujiisoup:
Ported from #1724, [comment](https://github.com/pydata/xarray/pull/1724#pullrequestreview-77354985) by @shoyer 

> In the long term, it would be nice to get ride of these uses of _data, maybe by switching entirely from our lazy array classes to Dask.

https://github.com/pydata/xarray/blob/1a012080e0910f3295d0fc26806ae18885f56751/xarray/core/formatting.py#L200-L218

shoyer:
This comment has the full context: https://github.com/pydata/xarray/issues/1372#issuecomment-293748654. To repeat myself:

----------------------------

You might ask why this separate lazy compute machinery exists. The answer is that dask fails to optimize element-wise operations like `(scale * array)[subset]` -> `scale * array[subset]`, which is a critical optimization for lazy decoding of large datasets.

See https://github.com/dask/dask/issues/746 for discussion and links to PRs about this. jcrist had a solution that worked, but it slowed down every dask array operations by 20%, which wasn't a great win.

I wonder if this is worth revisiting with a simpler, less general optimization pass that doesn't bother with broadcasting. See the subclasses of `NDArrayMixin` in [xarray/conventions.py](https://github.com/pydata/xarray/blob/master/xarray/conventions.py) for examples of the sorts of functionality we need:
- Casting (e.g., `array.astype(bool)`).
- Chained arithmetic with scalars (e.g., `0.5 + 0.5 * array`).
- Custom element-wise operations (e.g., `map_blocks(convert_to_datetime64, array, dtype=np.datetime64)`)
- Custom aggregations that drop a dimension (e.g., `map_blocks(characters_to_string, array, drop_axis=-1)`)

If we could optimize all these operations (and ideally chain them), then we could drop all the lazy loading stuff from xarray in favor of dask, which would be a real win.

----------------------------------

The downside of this switch is that lazy loading of data from disk would now *require* dask, which would be at least slightly annoying to some users. But it's probably worth the tradeoff from a maintainability perspective, and also to fix issues like #1372.

jhamman:
> The downside of this switch is that lazy loading of data from disk would now require dask, which would be at least slightly annoying to some users. But it's probably worth the tradeoff from a maintainability perspective, and also to fix issues like #1372.

[...]

Issue #1194, opened 2017-01-06
Use masked arrays while preserving int


gerritholl:
A great beauty of numpys masked arrays is that it works with any dtype, since it does not use `nan`.  Unfortunately, when I try to put my data into an `xarray.Dataset`, it converts ints to float, as shown below:

```
In [137]: x = arange(30, dtype="i1").reshape(3, 10)

In [138]: xr.Dataset({"count": (["x", "y"], ma.masked_where(x%5>3, x))}, coords={"x": range(3), "y":
     ...: range(10)})
Out[138]:
<xarray.Dataset>
Dimensions:  (x: 3, y: 10)
Coordinates:
  * y        (y) int64 0 1 2 3 4 5 6 7 8 9
  * x        (x) int64 0 1 2
Data variables:
    count    (x, y) float64 0.0 1.0 2.0 3.0 nan 5.0 6.0 7.0 8.0 nan 10.0 ...
```

This happens in the function [`_maybe_promote`](https://github.com/pydata/xarray/blob/master/xarray/core/common.py#L693).

Such type “promotion” is unaffordable for me; the memory consumption of my multi-gigabyte arrays would explode by a factor 4.  Secondly, many of my integer-dtype fields are bit arrays, for which floating point representation is not desirable.

It would greatly benefit `xarray` if it could use masking while preserving the dtype of input data.

(See also: [Stackoverflow question](http://stackoverflow.com/q/41505699/974555))

shoyer:
I answered your question on StackOverflow.

I agree that this is unfortunate. The cleanest solution would be an integer dtype with missing value support in NumPy itself, but that isn't going to happen anytime soon.

I'm not entirely opposed to the idea of adding (limited) support for masked arrays in xarray (see also https://github.com/pydata/xarray/pull/1118), but this could be a lot of work for relatively limited return.

I definitely recommend trying dask for processing multi-gigabyte arrays. You might even find the performance boost compelling enough that you could forgive the limitation that it doesn't handle masked arrays, either.

gerritholl:
[...]

PR #1899, opened 2018-02-09
Vectorized lazy indexing

### LINKED ISSUE #1897: Vectorized indexing with cache=False

```python
import numpy as np
import xarray as xr
n_times = 4; n_lats = 10; n_lons = 15
n_points = 4

ds = xr.Dataset({'test_var': (['time', 'latitude', 'longitude'],
                              np.random.random((n_times, n_lats, n_lons)))})
ds.to_netcdf('test.nc')

rand_lons = xr.Variable('points', np.random.randint(0, high=n_lons, size=n_points))
rand_lats = xr.Variable('points', np.random.randint(0, high=n_lats, size=n_points))

ds = xr.open_dataset('test.nc', cache=False)
points = ds['test_var'][:, rand_lats, rand_lons]
```
yields:
```
NotImplementedError                       Traceback (most recent call last)
<ipython-input-7-f16e4cae9456> in <module>()
     12 
     13 ds = xr.open_dataset('test.nc', cache=False)
---> 14 points = ds['test_var'][:, rand_lats, rand_lons]

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/dataarray.py in __getitem__(self, key)
    478         else:
    479             # xarray-style array indexing
--> 480             return self.isel(**self._item_key_to_dict(key))
    481 
    482     def __setitem__(self, key, value):

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/dataarray.py in isel(self, drop, **indexers)
    759         DataArray.sel
    760         """
--> 761         ds = self._to_temp_dataset().isel(drop=drop, **indexers)
    762         return self._from_temp_dataset(ds)
    763 

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/dataset.py in isel(self, drop, **indexers)
   1390         for name, var in iteritems(self._variables):
   1391             var_indexers = {k: v for k, v in indexers_list if k in var.dims}
-> 1392             new_var = var.isel(**var_indexers)
   1393             if not (drop and name in var_indexers):
   1394                 variables[name] = new_var

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/variable.py in isel(self, **indexers)
    851             if dim in indexers:
    852                 key[i] = indexers[dim]
--> 853         return self[tuple(key)]
    854 
    855     def squeeze(self, dim=None):

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/variable.py in __getitem__(self, key)
    620         """
    621         dims, indexer, new_order = self._broadcast_indexes(key)
--> 622         data = as_indexable(self._data)[indexer]
    623         if new_order:
    624             data = np.moveaxis(data, range(len(new_order)), new_order)

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/indexing.py in __getitem__(self, key)
    554 
    555     def __getitem__(self, key):
--> 556         return type(self)(_wrap_numpy_scalars(self.array[key]))
    557 
    558     def __setitem__(self, key, value):

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/indexing.py in __getitem__(self, indexer)
    521 
    522     def __getitem__(self, indexer):
--> 523         return type(self)(self.array, self._updated_key(indexer))
    524 
    525     def __setitem__(self, key, value):

~/anaconda/envs/pangeo/lib/python3.6/site-packages/xarray/core/indexing.py in _updated_key(self, new_key)
    491                 'Vectorized indexing for {} is not implemented. Load your '
    492                 'data first with .load() or .compute(), or disable caching by '
--> 493                 'setting cache=False in open_dataset.'.format(type(self)))
    494 
    495         iter_new_key = iter(expanded_indexer(new_key.tuple, self.ndim))

NotImplementedError: Vectorized indexing for <class 'xarray.core.indexing.LazilyIndexedArray'> is not implemented. Load your data first with .load() or .compute(), or disable caching by setting cache=False in open_dataset.
```

#### Problem description

Raising a `NotImplementedError` here is fine but it instructs the user to "disable caching by setting cache=False in open_dataset" which I've already done. So my questions are 1) should we expect this to work and 2) if not

#### Expected Output

Ideally, we can get the same behavior as: 

```python
ds = xr.open_dataset('test2.nc', cache=False).load()
points = ds['test_var'][:, rand_lats, rand_lons]

<xarray.DataArray 'test_var' (time: 4, points: 4)>
array([[0.939469, 0.406885, 0.939469, 0.759075],
       [0.470116, 0.585546, 0.470116, 0.37833 ],
       [0.274321, 0.648218, 0.274321, 0.383391],
       [0.754121, 0.078878, 0.754121, 0.903788]])
Dimensions without coordinates: time, points
```

without needing to use `.load()`

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 3.10.0-693.5.2.el7.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0+dev55.g1d32399
pandas: 0.22.0
numpy: 1.14.0
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.1
distributed: 1.20.2
matplotlib: 2.1.2
cartopy: 0.15.1
seaborn: 0.8.1
setuptools: 38.4.0
pip: 9.0.1
conda: None
pytest: 3.4.0
IPython: 6.2.1
sphinx: None
</details>

fujiisoup:
- [x] Closes #1897

I tried to support lazy vectorised indexing inspired by #1897.
More tests would be necessary but I want to decide whether it is worth to continue.

My current implementation is
+ For outer/basic indexers, we combine successive indexers (as we are doing now).
+ For vectorised indexers, we just store them as is and index sequentially when the evaluation.

The implementation was simpler than I thought, but it has a clear limitation.
It requires to load array before the vectorised indexing (I mean, the evaluation time).
If we make a vectorised indexing for a large array, the performance significantly drops and it is not noticeable until the evaluation time.

I appreciate any suggestions.


fujiisoup:
I noticed the lazy vectorized indexing can be (sometimes) optimized by decomposing the vectorized indexers into successive outer and vectorized indexers, so that the size of the array to be loaded into memory is minimized.

shoyer:
I figured out how to consolidate two vectorized indexers, as long as they don't include any `slice` objects:
```python
import numpy as np

def index_vectorized_indexer(old_indexer, applied_indexer):
    return tuple(o[applied_indexer] for o in np.broadcast_arrays(*old_indexer))

for x, old, applied in [
    (np.arange(10), (np.arange(2, 7),), (np.array([3, 2, 1]),)),
    (np.arange(10), (np.arange(6).reshape(2, 3),), (np.arange(2), np.arange(1, 3))),
    (-np.arange(1, 21).reshape(4, 5), (np.arange(3)[:, None], np.arange(4)[None, :]), (np.arange(3), np.arange(3))),
]:
    new_key = index_vectorized_indexer(old, applied)
    np.testing.assert_array_equal(x[old][applied], x[new_key])
```

We could probably make this work with `VectorizedIndexer` if we converted the slice objects to arrays. I think we might even already have some code to do that conversion somewhere. So another option would be to convert `BasicIndexer` and `OuterIndexer` -> `VectorizedIndexer` if necessary and then use this path.

fujiisoup:
Thanks, @shoyer 
Do you think it is possible to consolidate `transpose` also?
We need it to keep our logic in `Variable._broadcast_indexing`.

I am wondering what computation cost we want to avoid by the lazy indexing.
1. The indexing itself is expensive so we want to minimize the number of indexing operation? 
2. The original data is too large to fit into memory, and we want to load the smallest subset of the original array by the lazy indexing?

If the reason 2 is the common case, I think it is not a good idea to consolidate all the lazy indexing as `VectorizedIndexer`, since most of the backend does not support vectorized indexing, which means we need to load all the array into memory before any indexing operation.
(But still it would be valuable to consolidate all the indexers *after* the first vectorized indexer, since we can decompose any VectorizedIndexer into successive outer- and smaller vectorized-indexers pair.)

And I am also wondering as pointed out in #1725, what I am doing now was already implemented in dask.

shoyer:
Reason 2 is the primary one. We want to load the minimum amount of data possible into memory, mostly because pulling data from disk is slow.

shoyer:
I think the design choice here really comes down to whether we want to enable VectorizedIndexing on arbitrary data on disk or not:

Is it better to:
1. Always allow vectorized indexing by means of (lazily) loading all indexed data into memory as a single chunk. This could potentially be very expensive for IO or memory in hard to predict ways.
2. Or to only allow vectorized indexing if a backend supports it directly. This ensures that when vectorized indexing works it works efficiently. Vectorized indexing is still possibly but you have to explicitly write `.compute()`/`.load()`.

I think I slightly prefer option (2) but I can see the merits in either decision.

shoyer:
> in case we want to get three diagonal elements (1, 1), (2, 2), (3, 3) from a 1000x1000 array. What we want is
array[[1, 2, 3], [1, 2, 3]].
It can be decomposed to
array[1: 4, 1:4][[0, 1, 2], [0, 1, 2]].
We only need to load 3 x 3 part of the 1000 x 1000 array, then take its diagonal elements.

OK, this is pretty clever.

There are some obvious fail cases, e.g., if they want to pull out indices `array[[1, -1], [1, -1]]`, in which case the entire array needs to be sliced. I wonder if we should try to detect these with some heuristics, e.g., if the size of the result is much (maybe 10x or 100x) smaller than the size of sliced arrays.

Also, we would want to avoid separating basic/vectorized for backends that support efficient vectorized indexing (scipy and zarr).

fujiisoup:
> There are some obvious fail cases, e.g., if they want to pull out indices array[[1, -1], [1, -1]], in which case the entire array needs to be sliced. 

If the backend supports the orthogonal indexing (not only the basic indexing), 
we can do `array[[1, -1]][:, [1, -1]]`, load the 2x2 array, then apply the vectorized indexing `[[0, 1], [0, 1]]`.

But if we want a full diagonal, we need a full slice anyway...

> Also, we would want to avoid separating basic/vectorized for backends that support efficient vectorized indexing (scipy and zarr).

OK. Agreed. We may need a flag that can be accessed from the array wrapper.

fujiisoup:
Based on the [suggestion](https://github.com/pydata/xarray/pull/1899#issuecomment-364529325), I implemented the lazy vectorized indexing with index-consolidation.

Now, every backend is compatible to all the indexer types, i.e. basic-, outer- and vectorized-indexers.

It sometimes consume large amount of memory if the indexer is unable to decompose efficiently, but it is always better than loading the full slice.
The drawback is the unpredictability of how many data will be loaded.

xarray/core/indexing.py
```diff
@@ -602,23 +656,23 @@ def _outer_to_vectorized_indexer(key, shape):
...
     shape : tuple
         Shape of the array subject to the indexing.
 
     Returns
     -------
     tuple
         Tuple suitable for use to index a NumPy array with vectorized indexing.
-        Each element is an integer or array: broadcasting them together gives
-        the shape of the result.
+        Each element is array: broadcasting them together gives the shape of
```
jhamman: Each element is ***an*** array

xarray/core/indexing.py
```diff
@@ -654,6 +708,45 @@ def _outer_to_numpy_indexer(key, shape):
         return _outer_to_vectorized_indexer(key, shape)
 
 
+def _decompose_vectorized_indexer(indexer):
+    """ Decompose vectorized indexer to outer and vectorized indexers,
+    array[indexer] == array[oindex][vindex]
+    such that array[oindex].shape becomes smallest.
```
jhamman: becomes smallest what?

Just trying to make sure I understand the intent here. Is this algorithm trying to optimize (minimize) the size of `array[oindex]`?

xarray/backends/rasterio_.py
```diff
@@ -41,8 +41,10 @@ def shape(self):
         return self._shape
 
     def __getitem__(self, key):
-        key = indexing.unwrap_explicit_indexer(
-            key, self, allow=(indexing.BasicIndexer, indexing.OuterIndexer))
+        # TODO support indexing.decompose_indexer
```
fujiisoup: I think I need someone's help to update indexing scheme for rasterio backend.

WeatherGod:
Oh, wow... this worked like a charm for the netcdf4 backend! I have a ~13GB (uncompressed) 4-D netcdf4 variable that was giving me trouble for slicing a 2D surface out of. Here is a snippet where I am grabbing data at random indices in the last dimension. First for a specific latitude, then for the entire domain.
```
>>> CD_subset = rough['CD'][0]
>>> wind_inds_decorated
<xarray.DataArray (latitude: 3501, longitude: 7001)>
array([[33, 15, 25, ..., 52, 66, 35],
       [ 6,  8, 55, ..., 59,  6, 50],
       [54,  2, 40, ..., 32, 19,  9],
       ..., 
       [53, 18, 23, ..., 19,  3, 43],
       [ 9, 11, 66, ..., 51, 39, 58],
       [21, 54, 37, ...,  3,  0, 65]])
Dimensions without coordinates: latitude, longitude
>>> foo = CD_subset.isel(latitude=0, wind_direction=wind_inds_decorated[0])
>>> foo
<xarray.DataArray 'CD' (longitude: 7001)>
array([ 0.004052,  0.005915,  0.002771, ...,  0.005604,  0.004715,  0.002756], dtype=float32)
Coordinates:
    scales          int16 60
    latitude        float64 54.99
  * longitude       (longitude) float64 -130.0 -130.0 -130.0 -130.0 -130.0 ...
    wind_direction  (longitude) int16 165 75 125 5 235 345 315 175 85 35 290 ...
>>> foo = CD_subset.isel(wind_direction=wind_inds_decorated)
>>> foo
<xarray.DataArray 'CD' (latitude: 3501, longitude: 7001)>
[24510501 values with dtype=float32]
Coordinates:
    scales          int16 60
  * latitude        (latitude) float64 54.99 54.98 54.97 54.96 54.95 54.95 ...
  * longitude       (longitude) float64 -130.0 -130.0 -130.0 -130.0 -130.0 ...
    wind_direction  (latitude, longitude) int64 165 75 125 5 235 345 315 175 ...
```
All previous attempts at this would result in having to load the entire 13GB array into memory just to get 93.5 MB out. Or, I would try to fetch each individual point, which took way too long. This worked faster than loading the entire thing into memory, and it used less memory, too (I think I maxed out at about 1.2GB of total usage, which is totally acceptable for my use case).

I will try out similar things with the pynio and rasterio backends, and get back to you. Thanks for this work!

xarray/backends/h5netcdf_.py
```diff
@@ -17,15 +17,20 @@
...
         # h5py requires using lists for fancy indexing:
         # https://github.com/h5py/h5py/issues/992
-        # OuterIndexer only holds 1D integer ndarrays, so it's safe to convert
-        # them to lists.
-        key = tuple(list(k) if isinstance(k, np.ndarray) else k for k in key)
+        key = tuple(list(k) if isinstance(k, np.ndarray) else k for k in
+                    key.tuple)
         with self.datastore.ensure_open(autoclose=True):
-            return self.get_array()[key]
+            array = np.asarray(self.get_array()[key], dtype=self.dtype)
```
WeatherGod: shouldn't this be np.asanyarray() to allow masked arrays to come out of netcdf4?

xarray/core/indexing.py
```diff
@@ -531,6 +514,55 @@ def __repr__(self):
...
+            self.key = _arrayize_vectorized_indexer(key, array.shape)
+        self.array = as_indexable(array)
+        self.order = np.arange(self.ndim)
+
+    @property
+    def shape(self):
+        return np.broadcast(*self.key.tuple).shape
+
+    def __array__(self, dtype=None):
+        return np.asarray(self.array[self.key], dtype=None)
```
WeatherGod: Shouldn't this be `np.asanyarray()`? Could this interfere with dask arrays? Pint support?

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    first indexer will be used to index backend arrays, while the second one
+    is used to index loaded on-memory np.ndarray.
+
+    Parameters
+    ----------
+    indexer: VectorizedIndexer
+    mode: str, one of ['basic' | 'outer' | 'outer_1vector' | 'vectorized']
+        basic: for backends that support onle basic indexer
+        outer: for backends that support basic / outer indexer
+        outer_1vector: for backends that support outer indexer inluding at most
```
WeatherGod: "inluding" --> "including"

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    >>> array[np_indexer]  # vectorized indexing for on-memory np.ndarray.
+    """
+    assert isinstance(indexer, VectorizedIndexer)
+
+    if mode == 'vectorized':
+        return indexer, ()
+
+    backend_indexer = []
+    np_indexer = []
+    # posify
```
WeatherGod: "posify"???

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    indexer = [np.where(k < 0, k + s, k) if isinstance(k, np.ndarray) else k
+               for k, s in zip(indexer.tuple, shape)]
+
+    for k in indexer:
+        if isinstance(k, slice):
+            backend_indexer.append(k)
+            np_indexer.append(slice(None))
```
WeatherGod: Might help for future readers to add a bit of prose to each of these. Like: "If it is a slice, then we will slice it as-is (`k`) in the backend, and then use all of it (`slice(None)`) for the in-memory portion."

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    backend_indexer, np_indexers = _decompose_outer_indexer(
+        backend_indexer, shape, mode)
+    return backend_indexer, (np_indexers[0], np_indexer)
+
+
+def _decompose_outer_indexer(indexer, shape, mode):
+    """
+    Decompose ouer indexer to the successive two indexers, where the
+    first indexer will be used to index backend arrays, while the second one
+    is ised to index the loaded on-memory np.ndarray.
```
WeatherGod: "ised" --> "used"

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    shape = tuple(s if isinstance(k, slice) else len(k) for k, s in
+                  zip(backend_indexer.tuple, shape))
+    backend_indexer, np_indexers = _decompose_outer_indexer(
+        backend_indexer, shape, mode)
+    return backend_indexer, (np_indexers[0], np_indexer)
+
+
+def _decompose_outer_indexer(indexer, shape, mode):
+    """
+    Decompose ouer indexer to the successive two indexers, where the
```
WeatherGod: "ouer" --> "outer"

Also, the phrasing is odd. I would go with "Decompose outer indexer into two indexers" instead, as it is more concise.

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    As an example, let us consider to index a few elements from a backend array
+    with a orthogonal indexer ([0, 3, 1], [2, 3, 2]).
+    Even if the backend array only supports basic indexing, it is more
+    efficient to load a subslice of the array than loading the entire array,
+
+    >>> backend_indexer = BasicIndexer(slice(0, 3), slice(2, 3))
+    >>> array = array[backend_indexer]  # load subslice of the array
+    >>> np_indexer = OuterIndexer([0, 2, 1], [0, 1, 0])
+    >>> array[np_indexer]  # outer indexing for on-memory np.ndarray.
+    """
```
WeatherGod: The modes aren't documented. Maybe a reference to a list of them?

xarray/core/indexing.py
jhamman: This should be okay, neither xarray or dask support using masked arrays under the hood. See  #1194.

WeatherGod:
Hmm, came across a bug with the pynio backend. Working on making a reproducible example, but just for your own inspection, here is some logging output:
```
<xarray.Dataset>
Dimensions:    (xgrid_0: 1799, ygrid_0: 1059)
Coordinates:
    lv_HYBL0   float32 8.0
    longitude  (ygrid_0, xgrid_0) float32 ...
    latitude   (ygrid_0, xgrid_0) float32 ...
Dimensions without coordinates: xgrid_0, ygrid_0
Data variables:
    UGRD       (ygrid_0, xgrid_0) float32 ...
    VGRD       (ygrid_0, xgrid_0) float32 ...
DEBUG:hiresWind.downscale:shape of a data: (50, 1059, 1799)
```
The first bit is the repr of my DataSet. The last line is output of `ds['UGRD'].values.shape`. It is supposed to be 3D, not 2D.

If I revert back to v0.10.0, then the shape is (1059, 1799}, just as expected.

WeatherGod:
I can also confirm that the shape comes out correctly using master, so this is definitely isolated to this PR.

WeatherGod:
Ah, interesting... so, this dataset was created by doing an isel(0) on the original:
```
>>> ds['UGRD_P0_L105_GLC0']
<xarray.DataArray 'UGRD_P0_L105_GLC0' (lv_HYBL0: 50, ygrid_0: 1059, xgrid_0: 1799)>
[95257050 values with dtype=float32]
Coordinates:
  * lv_HYBL0   (lv_HYBL0) float32 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0 ...
    gridlat_0  (ygrid_0, xgrid_0) float32 ...
    gridlon_0  (ygrid_0, xgrid_0) float32 ...
Dimensions without coordinates: ygrid_0, xgrid_0
```
So, the original data has a 50x1059x1799 grid, and the new indexer isn't properly composing the indexer so that it fetches [0, slice(None), slice(None)] when I grab it's `.values`.

xarray/backends/pynio_.py
```diff
@@ -27,14 +27,18 @@ def get_array(self):
         return self.datastore.ds.variables[self.variable_name]
 
     def __getitem__(self, key):
-        key = indexing.unwrap_explicit_indexer(
-            key, target=self, allow=indexing.BasicIndexer)
+        key, np_inds = indexing.decompose_indexer(key, self.shape,
+                                                  mode='outer')
```
WeatherGod: PyNIO actually has a very robust indexing interface. We might be over-simplifying here, and this could be leading to bugs. I am not exactly sure we even need to do all of this indirection. PyNIO allows for the creation of a "selection object" that can be passed to the pynio array objects. Perhaps we should be creating that object instead of forcing the pynio array object into a numby index adapter?

WeatherGod: https://www.pyngl.ucar.edu/NioExtendedSelection.shtml

xarray/backends/rasterio_.py
WeatherGod: rasterio is strictly a slicing approach for fetching data, or pulling out individual data points. In this case, you want to avoid converting the slices into numpy arrays of indexes. Maybe it would make sense to have a function that could re-compose numpy arrays of indexes into groups of slices? So, if we were indexing `[1, 3, 4, 5, 2, 6]`, we would get back a `slice(1, 7)`? Maybe even make it smart enough to take the chunk hints so that indexes like `[1, 4, 5]` is kept as `slice(1, 6)`, rather than `(slice(1, 2), slice(4, 6))`?

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+    Decompose vectorized indexer to the successive two indexers, where the
+    first indexer will be used to index backend arrays, while the second one
+    is used to index loaded on-memory np.ndarray.
+
+    Parameters
+    ----------
+    indexer: VectorizedIndexer
+    mode: str, one of ['basic' | 'outer' | 'outer_1vector' | 'vectorized']
+        basic: for backends that support onle basic indexer
```
WeatherGod: "onle" --> "only"?

WeatherGod:
Just did some more debugging, putting in some debug statements within `NioArrayWrapper.__getitem__()`:
```
diff --git a/xarray/backends/pynio_.py b/xarray/backends/pynio_.py
index c7e0ddf..b9f7151 100644
--- a/xarray/backends/pynio_.py
+++ b/xarray/backends/pynio_.py
@@ -27,16 +27,24 @@ class NioArrayWrapper(BackendArray):
         return self.datastore.ds.variables[self.variable_name]
 
     def __getitem__(self, key):
+        import logging
+        logger = logging.getLogger(__name__)
+        logger.addHandler(logging.NullHandler())
+        logger.debug("initial key: %s", key)
         key, np_inds = indexing.decompose_indexer(key, self.shape,
                                                   mode='outer')
+        logger.debug("Decomposed indexers:\n%s\n%s", key, np_inds)
 
         with self.datastore.ensure_open(autoclose=True):
             array = self.get_array()
+            logger.debug("initial array: %r", array)
             if key == () and self.ndim == 0:
                 return array.get_value()
 
             for ind in np_inds:
+                logger.debug("indexer: %s", ind)
                 array = indexing.NumpyIndexingAdapter(array)[ind]
+                logger.debug("intermediate array: %r", array)
 
             return array
 
```

And here is the test script (data not included):
```
import logging
import xarray as xr
logging.basicConfig(level=logging.DEBUG)
fname1 = '../hrrr.t12z.wrfnatf02.grib2'
ds = xr.open_dataset(fname1, engine='pynio')
subset_isel = ds.isel(lv_HYBL0=7)
sp = subset_isel['UGRD_P0_L105_GLC0'].values.shape
```

And here is the relevant output:
```
DEBUG:xarray.backends.pynio_:initial key: BasicIndexer((slice(None, None, None),))
DEBUG:xarray.backends.pynio_:Decomposed indexers:
BasicIndexer((slice(None, None, None),))
()
DEBUG:xarray.backends.pynio_:initial array: <Nio.NioVariable object at 0x7f0f3c339210>
DEBUG:xarray.backends.pynio_:initial key: BasicIndexer((slice(None, None, None),))
DEBUG:xarray.backends.pynio_:Decomposed indexers:
BasicIndexer((slice(None, None, None),))
()
DEBUG:xarray.backends.pynio_:initial array: <Nio.NioVariable object at 0x7f0f3c339b90>
DEBUG:xarray.backends.pynio_:initial key: BasicIndexer((slice(None, None, None),))
DEBUG:xarray.backends.pynio_:Decomposed indexers:
BasicIndexer((slice(None, None, None),))
()
DEBUG:xarray.backends.pynio_:initial array: <Nio.NioVariable object at 0x7f0f3c339d50>
DEBUG:xarray.backends.pynio_:initial key: BasicIndexer((slice(None, None, None),))
DEBUG:xarray.backends.pynio_:Decomposed indexers:
BasicIndexer((slice(None, None, None),))
()
DEBUG:xarray.backends.pynio_:initial array: <Nio.NioVariable object at 0x7f0f3c339d90>
DEBUG:xarray.backends.pynio_:initial key: BasicIndexer((7, slice(None, None, None), slice(None, None, None)))
DEBUG:xarray.backends.pynio_:Decomposed indexers:
BasicIndexer((7, slice(None, None, None), slice(None, None, None)))
()
DEBUG:xarray.backends.pynio_:initial array: <Nio.NioVariable object at 0x7f0f3c339190>
DEBUG:xarray.backends.pynio_:initial key: BasicIndexer((7, slice(None, None, None), slice(None, None, None)))
DEBUG:xarray.backends.pynio_:Decomposed indexers:
BasicIndexer((7, slice(None, None, None), slice(None, None, None)))
()
DEBUG:xarray.backends.pynio_:initial array: <Nio.NioVariable object at 0x7f0f3c339190>
(50, 1059, 1799)
```

So, the `BasicIndexer((7, slice(None, None, None), slice(None, None, None)))` isn't getting decomposed correctly, it looks like?

xarray/backends/pynio_.py
```diff
@@ -27,14 +27,18 @@ def get_array(self):
...
         with self.datastore.ensure_open(autoclose=True):
             array = self.get_array()
             if key == () and self.ndim == 0:
                 return array.get_value()
-            return array[key]
+
+            for ind in np_inds:
+                array = indexing.NumpyIndexingAdapter(array)[ind]
+
+            return array
```
WeatherGod: Ah-hah! I see the bug! `key` never gets used!

WeatherGod: So, this should probably be something like:
```
array = array[key.tuple]
for ind in np_inds:
    array = indexing.NumpyIndexingAdapter(array)[ind]

return array
```

xarray/backends/h5netcdf_.py
WeatherGod: What's the reasoning behind doing `asarray` here, but not for other backends?

WeatherGod:
It looks like the pynio backend isn't regularly tested, as several of them currently fail when I run the tests locally. Some of them are failing because they are asserting NotImplementedErrors that are now implemented.

jhamman:
@WeatherGod - you are right, all the pynio tests are being skipped on travis. I'll open a separate issue for that. Yikes!

WeatherGod:
I am working on re-activating those tests. I think PyNio is now available
for python3, too.

xarray/backends/rasterio_.py
fujiisoup: @WeatherGod Thanks for the review.
> Maybe it would make sense to have a function that could re-compose numpy arrays of indexes into groups of slices?

Agreed, it would be an improvement. But I am wondering the combination will explode if the dimension increases. Some heuristics would be also necessary how many chunks we used. 
I do not have yet a clear solution. I am open to any idea for the improvement.

xarray/backends/pynio_.py
fujiisoup: Thanks for the debugging this!!
We certainly need test coverage for this.

xarray/backends/pynio_.py
fujiisoup: Thanks for the suggestion. 
I am not familiar with pynio, but it looks their selection object is very specific to pynio...
I want to leave this improvement to another PR and focus on the common feature in this PR.

xarray/core/indexing.py
shoyer: I'll take the blame for this one! I think I used this made-up word elsewhere :).

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+def _decompose_vectorized_indexer(indexer, shape, mode):
+    """
+    Decompose vectorized indexer to the successive two indexers, where the
+    first indexer will be used to index backend arrays, while the second one
+    is used to index loaded on-memory np.ndarray.
+
+    Parameters
+    ----------
+    indexer: VectorizedIndexer
+    mode: str, one of ['basic' | 'outer' | 'outer_1vector' | 'vectorized']
```
shoyer: This is a minor point, but instead of using strings for modes (which can be error prone), it would be better to use objects.

This is a perfect use-case for [enums](https://docs.python.org/3/library/enum.html), but it's probably not worth adding a new dependency just for this. Still, we can use objects here, e.g.,
```python
# like enums, the values for these class variables are arbitrary
class IndexingSupport(object):  # could inherit from enum.Enum on Python 3
    BASIC = 'BASIC'
    OUTER = 'OUTER'
    OUTER_1VECTOR = 'OUTER_1VECTOR'
    VECTORIZED = 'VECTORIZED'
```

xarray/core/indexing.py
```diff
@@ -654,6 +686,185 @@ def _outer_to_numpy_indexer(key, shape):
...
+
+    backend_indexer = []
+    np_indexer = []
+    # posify
+    indexer = [np.where(k < 0, k + s, k) if isinstance(k, np.ndarray) else k
+               for k, s in zip(indexer.tuple, shape)]
+
+    if mode == 'outer_1vector':
+        # some backends such as h5py supports only 1 vector in indexers
+        # We choose the most efficient axis
```
shoyer: This is very nice :)

xarray/backends/pynio_.py
shoyer: I'm not sure how selection objects would really help us here. I guess they can used to do a form of outer indexing, but putting integer indices into a string seems very inefficient. I agree that we should save it for later.

Are we actually sure that PyNIO even supports "outer" indexing? I had it marked down as only supporting basic indexing before. "Outer" indexing would include indexing with 1D arrays like `array[[0, 1, 2], :]`, but I don't see that support documented anywhere. (Maybe it does work, but just isn't documented.)

xarray/backends/rasterio_.py
WeatherGod: Yeah, we certainly have time to think about it. This feature doesn't seem to change anything for indexing that worked before, so I don't think we have to worry about regressions. Improvements to indexing can be thought of holistically as a separate task.

For our immediate needs though, for rasterio, we probably should just create a failback indexer that would just load up that entire dimension and index it as given if the original indexer is anything but scalar integers or slices.

xarray/core/indexing.py
```diff
@@ -705,12 +722,7 @@ def _decompose_vectorized_indexer(indexer, shape, mode):
     Parameters
     ----------
     indexer: VectorizedIndexer
-    mode: str, one of ['basic' | 'outer' | 'outer_1vector' | 'vectorized']
-        basic: for backends that support onle basic indexer
-        outer: for backends that support basic / outer indexer
-        outer_1vector: for backends that support outer indexer inluding at most
-                       1 vector.
-        vectorized: for backends that support full vectorized indexer.
+    indexing_support: one of IndexerSupport entries
```
WeatherGod: IndexerSupport --> `IndexingSupport`

Having the right name will be important for searches. Also, it would be valuable if this was a reference. xarray uses sphinx, right?

xarray/core/indexing.py
WeatherGod: Good job with the prose above, but the other spots are going to need similar prose as well.

xarray/core/indexing.py
```diff
@@ -654,6 +692,204 @@ def _outer_to_numpy_indexer(key, shape):
...
 
+class IndexingSupport(object):  # could inherit from enum.Enum on Python 3
+    # for backends that support only basic indexer
+    BASIC = 'BASIC'
+    # for backends that support basic / outer indexer
+    OUTER = 'OUTER'
+    # for backends that support outer indexer including at most 1 vector.
+    OUTER_1VECTOR = 'OUTER_1VECTOR'
+    # for backends that support full vectorized indexer.
+    VECTORIZED = 'VECTORIZED'
```
WeatherGod: I was just thinking. Is there a page some where that neatly and clearly documents the terms "basic", "outer", and "vectorized" indexing? I honestly keep getting myself mixed up what is what, and it would be good to have a single spot that explains the difference, and have that page referenced in difference places throughout the codebase, like here.

xarray/core/indexing.py
```diff
@@ -654,6 +692,204 @@ def _outer_to_numpy_indexer(key, shape):
...
+    n_dim = arrays[0].ndim if len(arrays) > 0 else 0
+    i_dim = 0
+    new_key = []
+    for v, size in zip(indexer.tuple, shape):
+        if isinstance(v, np.ndarray):
+            new_key.append(np.reshape(v, v.shape + (1, ) * len(slices)))
+        else:  # slice
+            shape = ((1,) * (n_dim + i_dim) + (-1,) +
+                     (1,) * (len(slices) - i_dim - 1))
+            new_key.append(np.arange(*v.indices(size)).reshape(shape))
```
WeatherGod: According to the tests, `v` is sometimes an integer at this point, and thus has an attribute error for `v.indices`.
From the test `TestLazyArray.test_vectorized_lazily_indexed_array`:

```
indexer = BasicIndexer((slice(1, 3, 2), 0)), shape = (1, -1)
    def _arrayize_vectorized_indexer(indexer, shape):
        """ Return an identical vindex but slices are replaced by arrays """
        slices = [v for v in indexer.tuple if isinstance(v, slice)]
        if len(slices) == 0:
            return indexer
    
        arrays = [v for v in indexer.tuple if isinstance(v, np.ndarray)]
        n_dim = arrays[0].ndim if len(arrays) > 0 else 0
        i_dim = 0
        new_key = []
        for v, size in zip(indexer.tuple, shape):
            if isinstance(v, np.ndarray):
                new_key.append(np.reshape(v, v.shape + (1, ) * len(slices)))
            else:  # slice
                shape = ((1,) * (n_dim + i_dim) + (-1,) +
                         (1,) * (len(slices) - i_dim - 1))
>               new_key.append(np.arange(*v.indices(size)).reshape(shape))
E               AttributeError: 'int' object has no attribute 'indices'
```

xarray/core/indexing.py
shoyer: https://github.com/numpy/numpy/pull/6256 is probably the closest. When I have time, I want to finish that up and merge it....

WeatherGod:
I can confirm that with the latest changes, the pynio tests now pass locally for me. Now, as to whether or not the tests in there are actually exercising anything useful is a different question.

xarray/core/indexing.py
fujiisoup: Thanks. 
Our VectorizedIndexer only consists of np.ndarrays and slices, it is OK here.
Indeed, the error was due to another bug. 
Fixed.

doc/whats-new.rst
```diff
@@ -48,6 +48,16 @@ Enhancements
...
+- Reduce methods such as :py:func:`DataArray.sum()` now handles object-type array.
+
+  .. ipython:: python
+
+    da = xr.DataArray(np.array([True, False, np.nan], dtype=object), dims='x')
+    da.sum()
+
+  (:issue:`1866`)
+  By `Keisuke Fujii <https://github.com/fujiisoup>`_.
+- Reduce methods such as :py:func:`DataArray.sum()` now accepts ``dtype``
```
WeatherGod: This has nothing to do with this PR. And it also duplicates a line.

fujiisoup: Thanks. Fixed.

xarray/backends/rasterio_.py
fujiisoup: @fmaussion can you help me for this? 
What I want to do here is the 2-staged indexing.
1. index backend array by the 1st indexer.
2. index the loaded np.ndarray by the 2nd vectorized indexer.

These two indexers (for backend and np.ndarray) can be computed by
```python
key, np_inds = indexing.decompose_indexer(
    key, self.shape, indexing.IndexingSupport.OUTER)
```

I am not familiar with rasterio data structure (I do not yet understand what is going on below...), 
I would appreciate if you could instruct me.

xarray/core/indexing.py
```diff
@@ -645,13 +682,237 @@ def _outer_to_numpy_indexer(key, shape):
...
+
+    Parameters
+    ----------
+    indexer: VectorizedIndexer
+    indexing_support: one of IndexerSupport entries
+
+    Returns
+    -------
+    backend_indexer: OuterIndexer or BasicIndexer
+    np_indexers: a sequence of vectorizedIndexer
```
fujiisoup: I slightly changed a direction here.
Previously, this function returns a sequence of `VectorizedIndexer`s for in-memory np.ndarray, but now it just returns a single `VectorizedIndexer` by combining multiple `VectorizedIndexer`s into one. It may slightly improve the indexing efficiency.

WeatherGod:
Something changed. Now the indexing for pynio is forcing a full loading of the data.

WeatherGod:
correction... the problem isn't with pynio... it is in the netcdf4 backend

WeatherGod:
Ah, not a change in behavior, but a possible bug exposed by a tiny change on my part. So, I have a 4D data array, `CD` and a data array for indexing, `wind_inds`. The following does not trigger a full loading: `CD[0][wind_direction=wind_inds]`, which is good! But, this does: `CD[scales=0, wind_direction=wind_inds]`, which is bad.

So, somehow, the indexing system is effectively treating these two things as different.

fujiisoup:
Can you share how you tested this?
The test I added says it is still in memory after vectroized indexing.

WeatherGod:
`CD` by the way, has dimensions of `scales, latitude, longitude, wind_direction`.

WeatherGod:
also, at this point, I don't know if this is limited to the netcdf4 backend, as this type of indexing was only done on a variable I have in a netcdf file. I don't have 4-D variables in other file types.

WeatherGod:
huh... now I am not so sure about that... must be something else triggering the load.

fujiisoup:
@WeatherGod, Thanks for testing.
Can you share more detail? 
With your example, what does `wind_inds` look like? Can you share the shape and dimension names?

WeatherGod:
Ah-hah! Ok, so, the problem isn't some weird difference between the two examples I gave. The issue is that calling `np.asarray(foo)` triggered a full loading of the data!

xarray/backends/rasterio_.py
shoyer: Rasterio (as wrapped by xarray) supports a strange form of indexing:
- Outer indexing for the first indexer element only.
- Basic indexing for remaining two indexer elements (every rasterio array is 3-dimensional).

I think the simplest fix would be to extend our rasterio wrapper to support arbitrary outer indexing.  This would entail changing the final clause below, where rasterio currently raises an error:
https://github.com/pydata/xarray/blob/e0621c7d66c13b486b1890f67a126caec2990da7/xarray/backends/rasterio_.py#L72-L78

Instead, we would want to create a slice object based on the minimum/maximum elements, and then add another indexing to `np_inds` to pull out the listed elements (after subtracting the `start` from the slice).

xarray/core/indexing.py
```diff
@@ -543,6 +543,10 @@ def _updated_key(self, new_key):
         return _combine_indexers(self.key, self.shape, new_key)
 
     def __getitem__(self, indexer):
+        # load if the indexed array becomes a scalar
```
shoyer: Is this actually a good idea? I think it's a little surprising to load data immediately just because the result is a scalar.

xarray/backends/rasterio_.py
```diff
@@ -40,48 +40,72 @@ def dtype(self):
...
         window = []
         squeeze_axis = []
-        for i, (k, n) in enumerate(zip(key[1:], self.shape[1:])):
+        for i, (k, n) in enumerate(zip(key.tuple[1:], self.shape[1:])):
             if isinstance(k, slice):
                 start, stop, step = k.indices(n)
-                if step is not None and step != 1:
-                    raise IndexError(_ERROR_MSG)
+                np_inds2.append(slice(None, None, step))
+                new_shape.append(stop - start)
```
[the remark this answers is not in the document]
fujiisoup: Thanks. I completely forgot about the negative step. 
fixed.

fujiisoup:
This looks some backends do not support negative step slices.
I'm going to wrap this maybe this weekend.

WeatherGod:
I did some more investigation into the memory usage problem I was having. I had assumed that the vectorized indexed result of a lazily indexed data array would be an in-memory array. So, when I then started to use the result, it was then doing a read of all the data at once, resulting in a near-complete load of the data into memory.

I have adjusted my code to chunk out the indexing in order to keep the memory usage under control at reasonable performance penalty. I haven't looked into trying to identify the ideal chunking scheme to follow for an arbitrary dataarray and indexing. Perhaps we can make that a task for another day. At this point, I am satisfied with the features (negative step-sizes aside, of course).

jhamman:
@fujiisoup - is this ready for a final review? I see you have all the tests passing 💯 !

fujiisoup:
I think it's ready :)

xarray/core/indexing.py
```diff
@@ -643,13 +684,282 @@ def _outer_to_numpy_indexer(key, shape):
...
+        else:  # slice:  convert positive step slice for backend
+            bk_slice, np_slice = _decompose_slice(k, s)
+            backend_indexer.append(bk_slice)
+            np_indexer.append(np_slice)
+
+    return (BasicIndexer(tuple(backend_indexer)),
+            OuterIndexer(tuple(np_indexer)))
+
+
+def _arrayize_vectorized_indexer(indexer, shape):
```
shoyer: Oops -- I wrote a basically identical helper function `xarray.backends.zarr._replace_slices_with_arrays` along with some unit tests.

Can you confirm that we can remove the zarr helper function, and that all its unit tests work with this function instead? I don't have a preference on which implementation to use, assuming that all tests pass with both.

fujiisoup:
All done :)

shoyer:
OK, in it goes. Thanks @fujiisoup !

fujiisoup:
Thanks, @WeatherGod , for your feedback.
This is finally merged!

WeatherGod:
:tada:

dopplershift:
So did this remove/rename `LazilyIndexedArray` in 0.10.2? Because I'm getting an attribute in the custom xarray backend I wrote: https://github.com/Unidata/siphon/blob/master/siphon/cdmr/xarray_support.py

I don't mind updating, but I wanted to make sure this was intentional.

fujiisoup:
Yes, 
LazilyIndexedArray was renamed to `LazilyOuterIndexedArray` and `LazilyVectorizedIndexedArray` was newly added.
These two backend arrays are selected depending on what kind of indexer is used.

Issue #1896, opened 2018-02-08
_color_palette not using full range of colors if seaborn is installed


mathause:
```python
from xarray.plot.utils import import_seaborn
_color_palette('Greys', 3)
```
Returns
0.85, ...
0.58, ...
0.31, ...

if seaborn is installed, and

1.00, ...
0.58, ...
0.00, ...

otherwise

#### Problem description

the return value of `_color_palette('Greys', 3)` is different when seaborn is installed or not.

The relevant code is here:
https://github.com/pydata/xarray/blob/6aa225f5dae9cc997e232c11a63072923c8c0238/xarray/plot/utils.py#L115
https://github.com/pydata/xarray/blob/6aa225f5dae9cc997e232c11a63072923c8c0238/xarray/plot/utils.py#L143

The same logic in seaborn
https://github.com/mwaskom/seaborn/blob/0beede57152ce80ce1d4ef5d0c0f1cb61d118375/seaborn/palettes.py#L411

Intuitively I prefer the xarray solution because this uses the full range of colors which I find beneficial, however there may be a reason for this I'm not aware of.

Maybe @mwaskom will  answer:
mwaskom/seaborn#1372

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.4.final.0
python-bits: 64
OS: Linux
OS-release: 3.13.0-141-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0
pandas: 0.21.0
numpy: 1.13.3
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.0
matplotlib: 2.1.1
cartopy: 0.15.1
seaborn: 0.8.1
setuptools: 38.2.4
pip: 9.0.1
conda: None
pytest: None
IPython: 6.2.1
sphinx: None

</details>

~~

edit:

sorry I pressed the button to early

shoyer:
Using seaborn as a fall-back here was probably a mistake. It's better to give consistent results independently of whether an optional dependency is installed.

fmaussion:
See https://github.com/pydata/xarray/pull/1902 for a fix

PR #1902, opened 2018-02-10
_color_palette consistent with or without seaborn

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/1896
 - [ ] Tests added 
 - [ ] Tests passed 

Instead of relying on seaborn per default we now use mpl and fall back to seaborn if cmap is not recognized (e.g. with `cmap="husl"`).


mathause:
Looks good to me. This will change the default colors. Also, seaborn cycles qualitative (colorbrewer) colormaps when you ask for more colors than available. Not saying this needs to be supported, but something to be aware of.

PR #1905, opened 2018-02-12
Added PNC backend to xarray 

barronh:
PNC is used for GEOS-Chem, CAMx, CMAQ and other atmospheric data formats
including NOAA arl packed bit and HYSPLIT related files. The goal is to 
provide access to many file types that have their own file formats and 
meta-data conventions. It can provide a CF compliant netCDF-like interface.

 - [x] Tests were run using example datasets, but datasets were not added to repository as they are large and diverse. Only files that provide CF compliant structures will successfully open. Confirmed by @bbakernoaa


barronh:
p.s., I just noticed my edits to whats-new.rst were not pushed until after my pull request.

shoyer:
First of all -- this is very cool, thanks for putting this together!

In the long term, I think we would prefer to move more specialized backends out of xarray proper. But at the current time, it's difficult to do this easily, and the backend interface itself is not entirely stable. So it probably makes sense to add this directly into xarray for now.

To merge this into xarray, I'm afraid that an automated test suite of some sort is non-negotiable. Are there example datasets can you can create on the fly from Python?

barronh:
First, I too quickly tried to fast forward and clearly some test was failing. I have updated my code to pass all tests with py.test.

Before closing this request and opening another, I want to make sure I am clear one the extent of what I should add. I can create binary data from within python and then read it, but all those tests are in my software package. Duplicating that seems like a bad idea.

I have added a NetCDF3Only testcase to test_backends.py and it passes. That doesn't stress the multi-format capabilities of pnc, but as I've said all the numerical assertions for the other formats are in my system's test cases. Is the NetCDF3Only test sufficient in this case?

Further, below are some simple applications that download my test data for CAMx and GEOS-Chem and plot it.

Thanks for the input.

```
import xarray as xr
from urllib.request import urlretrieve


# CAMx test

urlretrieve('https://github.com/barronh/pseudonetcdf/blob/master/src/PseudoNetCDF/testcase/camxfiles/uamiv/test.uamiv?raw=true', 'test.uamiv')
xf = xr.open_dataset('test.uamiv', engine = 'pnc')
pm = xf.O3.isel(TSTEP = 0, LAY = 0).plot()
pm.axes.figure.savefig('test_camx.png')

pm.axes.figure.clf()

# GEOS-Chem Test

urlretrieve("https://github.com/barronh/pseudonetcdf/blob/master/src/PseudoNetCDF/testcase/geoschemfiles/test.bpch?raw=true", "test.bpch")
urlretrieve("https://github.com/barronh/pseudonetcdf/blob/master/src/PseudoNetCDF/testcase/geoschemfiles/tracerinfo.dat?raw=true", "tracerinfo.dat")
urlretrieve("https://github.com/barronh/pseudonetcdf/blob/master/src/PseudoNetCDF/testcase/geoschemfiles/diaginfo.dat?raw=true", "diaginfo.dat")
xf = xr.open_dataset('test.bpch', engine = 'pnc')
xa = getattr(xf, 'IJ-AVG-$_Ox')
xa2d = xa.isel(time = 0, layer3 = 0)
pm = xa2d.plot()
pm.axes.figure.savefig('test_bpch.png')
```

jhamman:
> Before closing this request and opening another, I want to make sure I am clear one the extent of what I should add.

First, don't close this PR. You can just keep adding commits to your `pnc-backend`. 

In terms of test coverage, let's start by detailing what functionality we expect the PseudoNetCDF backend to have.

- What file formats are supported by PseudoNetCDF?
- What roundtrip capability is expected when PseudoNetCDF is used (i.e. if I write a netCDF3 file with scipy, can I open it with PNC and get exactly the same dataset)?
- What are PseudoNetCDF's dependencies (Python 3?, netCDF4?), are there optional dependencies?

Finally, two things that would really help grease the skids here would be:

1. Continuous integration testing in PseudoNetCDF, at a minimum on TravicCI, but Appveyor would be great too.
2. A conda-forge package to facilitate easy installs of PseudoNetCDF

Neither of these last two are strictly required but both are satisfied by all the other Xarray supported backends.

shoyer:
>  I can create binary data from within python and then read it, but all those tests are in my software package. Duplicating that seems like a bad idea.

Right.

The goal isn't to duplicate your tests, but to provide a meaningful integration test for the xarray backend. Can we read data from PNC into xarray and do everything an xarray user would want to do with it? Testing your API with a netCDF3 file would probably be enough, assuming you have good test coverage internally.

We already have a somewhat complete test suite for netCDF data that you could probably hook into, but for reference sorts of issues that tend to come up include:
- Indexing support. Do you only support basic-indexing like `x[0, :5]` or is indexing with integer arrays also supported?
- Serialization/thread-safety. Can we simultaneously read a file with another process or thread using dask?
- API consistency for scalar arrays. Do these require some sort of special API compared to non-scalar arrays?
- Data types support. Are strings and datetimes converted properly into the format xarray expects?

barronh:
> Indexing support. Do you only support basic-indexing like x[0, :5] or is indexing with integer arrays also supported?

My variable objects present a pure numpy array, so they follow numpy indexing precisely with one exception. If the files are actually netCDF4, they have the same limitations of the netCDF4.Variable object.

> Serialization/thread-safety. Can we simultaneously read a file with another process or thread using dask?

I have not tested separate processes. In many cases, I use numpy memmap. So that will be the limitation.

> API consistency for scalar arrays. Do these require some sort of special API compared to non-scalar arrays?

Same as numpy, but also has support for the netCDF4 style.

> Data types support. Are strings and datetimes converted properly into the format xarray expects?

I use relative dates following netcdf time conventions. Within my software, there are special functions for translation, but I have seen this be treated by xarray separately.

> Continuous integration testing in PseudoNetCDF, at a minimum on TravicCI,, but Appveyor would be great too.

I added TravisCI, but haven't looked Appveyor.

> A conda-forge package to facilitate easy installs of PseudoNetCDF

I added a ci/requirements-py36-netcdf4-dev.yml as a part of my TravisCI integration. I am also working on a recipe (like https://github.com/conda-forge/xarray-feedstock).

xarray/backends/pnc_.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class PncDataStore(AbstractDataStore, DataStorePickleMixin):
+    """Store for accessing datasets via PseudoNetCDF
+    """
+
+    def __init__(self, filename, mode='r', autoclose=False):
+        from PseudoNetCDF import pncopen
+        try:
+            opener = functools.partial(pncopen, filename,
```
shoyer: For new backend classes, we're trying to move to a style where there is a dedicated classmethod `open` that handles opening a file from scratch and the backend constructor accepts an already open File object. Take a look at NetCDF4Backend for an example.

xarray/backends/pnc_.py
```diff
@@ -0,0 +1,91 @@
...
+                                       mode=mode, format='netcdf')
+            self.ds = opener()
+        except Exception:
```
shoyer: Why do you need all these fallbacks? This logic is usually best avoided -- it's better to require users to be explicit.

You should pick some default behavior for `open` (which will be used by `open_dataset`) and require explicitly opening the backend object if that does not suffice for some use-case.

xarray/backends/api.py
```diff
@@ -295,6 +295,9 @@ def maybe_decode_store(store, lock=False):
         elif engine == 'pynio':
             store = backends.NioDataStore(filename_or_obj,
                                           autoclose=autoclose)
+        elif engine == 'pnc':
```
shoyer: I would slightly rather use the full name "pseudonetcdf" than an abbreviation like "pnc". We don't use abbreviations for any of the other backend names (to be fair, yes, they are also shorter).

shoyer:
> My variable objects present a pure numpy array, so they follow numpy indexing precisely with one exception. If the files are actually netCDF4, they have the same limitations of the netCDF4.Variable object.

OK, we will need to surface this information in some way for xarray -- maybe as an attribute of some sort on the psuedonetcdf side? The good news is that we already have handling for indexing like NumPy arrays and netCDF4 variables, but we need to know which behavior to expect to make all of indexing operations work efficiently.

It's also safe to say for now that only basic indexing is supported, but that will result in sub-optimal indexing behavior ( slower and more memory intensive than necessary).

Note that we are currently in the process of refactoring how we handle indexers in backends, so you'll probably need to update things after https://github.com/pydata/xarray/pull/1899 is merged.

bbakernoaa:
What is the status of this? I would like to start using this ASAP but cannot convince my team to use it until it is an official backend of xarray

barronh:
@bbakernoaa - It may be a bit. I'm

xarray/backends/pnc_.py
barronh: This would require changes to many of my classes that accept paths. Is this a preconditions?

shoyer: Can you elaborate a little bit -- how do you normally call `pncopen`? I assume it accepts some set of required and optional arguments?

For the usual case, users would write:
```
# this would call pncopen(filename) internally
ds = xr.open_dataset(path, engine='pseudonetcdf')
```
For advanced cases, users would write something like:
```
# this would call pncopen(filename, mode='r', format='netcdf') internally
store = xr.backends.PsuedoNetCDFBackend.open(filename, mode='r', format='netcdf')
ds = xr.open_dataset(store)
```

xarray/backends/api.py
barronh: Done.

xarray/backends/pnc_.py
barronh: I can remove the first except wrap, but the second is present because not all constructors except the mode keyword. Individual constructors are wrapped by pncopen.

shoyer: Can you simply leave out the `mode` argument in that case? There's no particular reason xarray needs to include it, as long as the default behavior supports reads.

xarray/backends/pnc_.py
```diff
@@ -0,0 +1,86 @@
...
+        self.shape = array.shape
+        self.dtype = np.dtype(array.dtype)
+
+    def get_array(self):
+        self.datastore.assert_open()
+        return self.datastore.ds.variables[self.variable_name]
+
+    def __getitem__(self, key):
+        key = indexing.unwrap_explicit_indexer(
+            key, target=self, allow=indexing.BasicIndexer)
```
fujiisoup: By #1899, we improved the indexing logic for backend arrays, 
to support flexible indexing for all the backend.

Can you update here as similar to the following?
```python
key, np_inds = indexing.decompose_indexer(
    key, self.shape, indexing.IndexingSupport.OUTER_1VECTOR)

array = self.get_array()[key]  # index backend array
if len(np_inds.tuple) > 0:
    array = indexing.NumpyIndexingAdapter(array)[np_inds]  # index the loaded np.ndarray
return array
```

For example, see
https://github.com/pydata/xarray/blob/54468e1924174a03e7ead3be8545f687f084f4dd/xarray/backends/h5netcdf_.py#L18-L32

Here, this
```python
key, np_inds = indexing.decompose_indexer(
    key, self.shape, indexing.IndexingSupport.OUTER_1VECTOR)
```
decomposes the given indexers into those for backend arrays (`key`) and for loaded `np.ndarray` (`np_inds`).

The third argument (`OUTER_1VECTOR`, `OUTER`, `VECTORIZED`, `BASIC`) indicates what kind of indexing the backend support.

xarray/backends/pnc_.py
barronh: A normal call looks like this

```
myfile = pncopen(path, format=fmt, **kwds)
```

The only required argument is path. The format keyword is a named format and is highly recommended. If format is omitted, pncopen will cycle through all the formats it knows using the isMine class method to find a class that can open it. The other keywords are optional from pncopen's perspective, but may be required if one is opening file formats that are not fully self-describing.

I had not realized that you can provide the `open` interface and the resulting store to open_dataset. I was not planning to provide support for these because I assume your __init__ interfaces need to be identical. This seems like a nice alternative.

I'll make an adjustment here.

xarray/backends/pnc_.py
barronh: I can leave out mode, but that will prevent the formats that do support mode from using it. I could add a `**kwds` option to `open` much like pncopen. The default behavior would, as you suggest, leave out mode.

xarray/backends/pnc_.py
shoyer: We should probably add an optional `backend_kwargs` argument to `open_dataset` to make this a little easier.

xarray/tests/test_backends.py
```diff
@@ -2198,6 +2197,39 @@ class TestPyNioAutocloseTrue(TestPyNio):
     autoclose = True
 
 
+@requires_scipy
+@requires_pnc
+class TestPnc(NetCDF3Only, TestCase):
```
fujiisoup: `NetCDF3Only` class does not do anything (see line 127) but it is just a mixin that is used if the backend does not support int64 (see line 190).
Can you make this inherit `CFEncodedDataTest`?
 (if PNC only supports netcdf3, we need both `CFEncodedDataTest` and `NetCDF3Only` as `ScipyFileObjectTest` does)
It would cover more cases.

If pnc does not support a certain feature that `CFEncodedDataTest` tries to test, I think it would be OK now to explicitly skip it.

barronh: Can you be more clear about this?

fujiisoup: You can see here
https://github.com/pydata/xarray/blob/7c2c43ce6b1792e1193635ab9b64fd248266f632/xarray/tests/test_backends.py#L127-L128

Inheriting `NetCDF3Only` does not do anything.

We have prepared general test cases `CFEncodedDataTest` and almost all of our backends are tested using this class.
I think pnc should be tested using `CFEncodedDataTest`.

xarray/tests/test_backends.py
```diff
@@ -2198,6 +2197,72 @@ class TestPyNioAutocloseTrue(TestPyNio):
...
+        with create_tmp_file() as tmp_file:
+            example.to_netcdf(tmp_file, engine='netcdf4')
+            on_disk = open_dataset(tmp_file, engine='pseudonetcdf')
+            actual = on_disk.rename({'foo': 'bar', 'x': 'y'})
+            del on_disk  # trigger garbage collection
+            assert_identical(actual, expected)
+
+    def test_pickle(self):
+        # pnc does not support pickling
+        pass
```
fujiisoup: Thanks :)
Very minor comment.
Maybe we can use `@pytest.mark.skip` which clearly indicates that the test is skipped not passing.
Example
https://github.com/pydata/xarray/blob/7c2c43ce6b1792e1193635ab9b64fd248266f632/xarray/tests/test_backends.py#L1452-L1454

bbakernoaa:
You guys are great. Thanks @barronh I’m going to be implementing this in MONET really soon.

barronh:
I have added a recipe to conda-forge and it is passing all tests. I don't know when it will be in feedstock.

barronh:
Help me understand, there are now failures in scipy that seem unrelated to my changes. In fact, I had to switch the writer in my tests to netcdf4 to bypass the scipy problem. Is this going to hold up my branch?

fujiisoup:
This looks due to scipy 1.0.1 and not related to this change.

> I have added a recipe to conda-forge and it is passing all tests. I don't know when it will be in feedstock.

Thanks :)
Let's wait until it will be in feedstock.

shoyer:
I opened a new issue for the scipy 1.0.1 failures: https://github.com/pydata/xarray/issues/2019

(I'll try to take a look at the shortly)

shoyer:
The tests are fixed on master, please merge in master to fix this.

barronh:
@fujiisoup - I have a recipe in conda-forge that is passing all tests. Do you know how long it typically takes to get added or what I need to do to get it added?

https://github.com/conda-forge/staged-recipes/pull/5449

fujiisoup:
> Do you know how long it typically takes to get added or what I need to do to get it added?

Sorry, but I do not have experience with conda-forge.
@shoyer , could you help?

shoyer:
If you have a pseudonetcdf up on pypi, we can install it for our tests with pip instead of conda. That should work for now until the conda recipe is merged. You could simply add it to the end of the "pip" section in this file: https://github.com/pydata/xarray/blob/master/ci/requirements-py36.yml

barronh:
I've added the latest version to pip. Still waiting on the recipe in conda-forge.

barronh:
The conda recipe was approved and merged. Feedstock should be ready soon.

https://github.com/conda-forge/staged-recipes/pull/5449

xarray/tests/test_backends.py
```diff
@@ -2213,6 +2212,77 @@ class PyNioTestAutocloseTrue(PyNioTest):
...
+    def open(self, path, **kwargs):
+        with open_dataset(path, engine='pseudonetcdf',
+                          autoclose=self.autoclose,
+                          **kwargs) as ds:
+            yield ds
+
+    def save(self, dataset, path, **kwargs):
+        dataset.to_netcdf(path, engine='netcdf4', **kwargs)
+
+    def test_weakrefs(self):
```
shoyer: We don't need this test -- it was specifically for some issue with pynio.

xarray/tests/test_backends.py
```diff
@@ -2213,6 +2212,77 @@ class PyNioTestAutocloseTrue(PyNioTest):
...
+            actual = on_disk.rename({'foo': 'bar', 'x': 'y'})
+            del on_disk  # trigger garbage collection
+            assert_identical(actual, expected)
+
+    @pytest.mark.skip(reason='cannot pickle file objects')
+    def test_pickle(self):
+        pass
+
+    @pytest.mark.skip(reason='cannot pickle file objects')
+    def test_pickle_dataarray(self):
```
shoyer: I'm not sure we should be skipping these tests.

They are skipped for `ScipyFileObjectTest` because it's not possible to pickle a `BytesIO()` object.

But it should work perfectly file to pickle the `opener()` function, which just keeps track of the file to open. If you use the model of the existing backend classes this should probably just work -- do you get an error message when you don't skip these?

xarray/tests/test_backends.py
```diff
@@ -2213,6 +2212,77 @@ class PyNioTestAutocloseTrue(PyNioTest):
...
+        pass
+
+    @pytest.mark.skip(reason='does not support boolean dtype')
+    def test_roundtrip_boolean_dtype(self):
+        # pnc does not support boolean dtype
```
shoyer: This is done by xarray's conventions handling logic. The underlying data is stored as integers. Unless you explicitly strip out attributes I think this should still work?

xarray/tests/test_backends.py
```diff
@@ -2213,6 +2212,77 @@ class PyNioTestAutocloseTrue(PyNioTest):
...
+        pass
+
+    @pytest.mark.skip(reason='does not support auto masking and scaling')
+    def test_roundtrip_mask_and_scale(self):
+        # pnc does not support auto masking and scaling
```
shoyer: Again, this is done in the xarray layer, by parsing the `_FillValue`, `scale_factor` and `add_offset` attributes.

Does pnc handle masking of data in some way? Do any of the file formats it works with have some way of marking data as invalid?

xarray/tests/test_backends.py
```diff
@@ -2213,6 +2212,77 @@ class PyNioTestAutocloseTrue(PyNioTest):
...
+        pass
+
+    @pytest.mark.skip(reason='does not support object types')
+    def test_roundtrip_object_dtype(self):
+        # pnc does not support object types
```
shoyer: Also done by xarray -- no backends support storing object dtype

xarray/backends/api.py
```diff
@@ -147,7 +147,8 @@ def _get_lock(engine, scheduler, format, path_or_file):
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=True, decode_times=True, autoclose=False,
                  concat_characters=True, decode_coords=True, engine=None,
-                 chunks=None, lock=None, cache=None, drop_variables=None):
+                 chunks=None, lock=None, cache=None, drop_variables=None,
+                 backend_kwargs={}):
```
shoyer: We can add this new `backend_kwargs` argument, but (1) it should pass arguments to each backends (otherwise it would be surprising) and (2) we should add it to `open_dataarray()`, too.

shoyer: Also, it's generally considered bad form to use mutable arguments like dictionaries as default values. Better to use `backend_kwargs=None` and then in the body of the function:
```
if backend_kwargs is None:
    backend_kwargs = {}
```

doc/io.rst
```diff
@@ -661,6 +661,34 @@ exporting your objects to pandas and using its broad range of `IO tools`_.
 
 .. _IO tools: http://pandas.pydata.org/pandas-docs/stable/io.html
 
+.. _io.PseudoNetCDF:
+
+
+Formats supported by PseudoNetCDF
```
shoyer: can you move this after PyNIO and before pandas? I think that would make a little more logical sense.

doc/io.rst
```diff
@@ -661,6 +661,34 @@ exporting your objects to pandas and using its broad range of `IO tools`_.
...
+---------------------------------
+
+xarray can also read CAMx, BPCH, ARL PACKED BIT, and many other file
+formats supported by PseudoNetCDF, if PseudoNetCDF is installed.
```
shoyer: Please make the first reference to "PseudoNetCDF" a link.

xarray/tests/test_backends.py
shoyer: On the other hand, maybe all these conventions aren't relevant for pseudonetcdf, since it's not designed to open arbitrary netcdf files?

doc/io.rst
barronh: Sure.

doc/io.rst
barronh: Sure

xarray/backends/api.py
barronh: Done.

barronh: I have added it to open_dataarray. Are you asking me to implement the keyword treatment for all other backends? Or just implement the passing and leave it unused?

xarray/tests/test_backends.py
barronh: I do get an error message.

```
E       AttributeError: 'PncDataStore' object has no attribute '_mode'
```

xarray/tests/test_backends.py
barronh: Referring to test_weakrefs right? Done.

xarray/tests/test_backends.py
barronh: If I add `_mode`, then I start getting another error:

```
    def assert_open(self):
        if not self._isopen:
>           raise AssertionError('internal failure: file must be open '
                                 'if `autoclose=True` is used.')
E           AssertionError: internal failure: file must be open if `autoclose=True` is used.
```

xarray/tests/test_backends.py
barronh: Same as above.

I am not stripping out attributes, but...

```
    def test_roundtrip_boolean_dtype(self):
        original = create_boolean_data()
        self.assertEqual(original['x'].dtype, 'bool')
        with self.roundtrip(original) as actual:
>           assert_identical(original, actual)
E           AssertionError: <xarray.Dataset>
E           Dimensions:  (t: 4)
E           Dimensions without coordinates: t
E           Data variables:
E               x        (t) bool True False False True
E           <xarray.Dataset>
E           Dimensions:  (t: 4)
E           Dimensions without coordinates: t
E           Data variables:
E               x        (t) int8 ...
```

xarray/tests/test_backends.py
barronh: The format that is being used by default is NetCDF4. Oddly, it seems to fail very early on in the test. It is not obvious to me where pnc is even used yet.

```
        decoded = create_masked_and_scaled_data()
        encoded = create_encoded_masked_and_scaled_data()
            assert_allclose(decoded, actual, decode_bytes=False)
```

xarray/tests/test_backends.py
barronh: Fails on my machine

```
    def test_roundtrip_object_dtype(self):
        floats = np.array([0.0, 0.0, 1.0, 2.0, 3.0], dtype=object)
        floats_nans = np.array([np.nan, np.nan, 1.0, 2.0, 3.0], dtype=object)
        bytes_ = np.array([b'ab', b'cdef', b'g'], dtype=object)
        bytes_nans = np.array([b'ab', b'cdef', np.nan], dtype=object)
        strings = np.array([u'ab', u'cdef', u'g'], dtype=object)
        strings_nans = np.array([u'ab', u'cdef', np.nan], dtype=object)
        all_nans = np.array([np.nan, np.nan], dtype=object)
        original = Dataset({'floats': ('a', floats),
                            'floats_nans': ('a', floats_nans),
                            'bytes': ('b', bytes_),
                            'bytes_nans': ('b', bytes_nans),
                            'strings': ('b', strings),
                            'strings_nans': ('b', strings_nans),
                            'all_nans': ('c', all_nans),
                            'nan': ([], np.nan)})
        expected = original.copy(deep=True)
        with self.roundtrip(original) as actual:
            try:
            except AssertionError:
                # Most stores use '' for nans in strings, but some don't.
                # First try the ideal case (where the store returns exactly)
                # the original Dataset), then try a more realistic case.
                # This currently includes all netCDF files when encoding is not
                # explicitly set.
                # https://github.com/pydata/xarray/issues/1647
                expected['bytes_nans'][-1] = b''
                expected['strings_nans'][-1] = u''
>               assert_identical(expected, actual)
E               AssertionError: <xarray.Dataset>
E               Dimensions:       (a: 5, b: 3, c: 2)
E               Dimensions without coordinates: a, b, c
E               Data variables:
E                   floats        (a) object 0.0 0.0 1.0 2.0 3.0
E                   floats_nans   (a) object nan nan 1.0 2.0 3.0
E                   bytes         (b) object b'ab' b'cdef' b'g'
E                   bytes_nans    (b) object b'ab' b'cdef' b''
E                   strings       (b) object 'ab' 'cdef' 'g'
E                   strings_nans  (b) object 'ab' 'cdef' 
E                   all_nans      (c) object nan nan
E                   nan           float64 nan
E               <xarray.Dataset>
E               Dimensions:       (a: 5, b: 3, c: 2)
E               Dimensions without coordinates: a, b, c
E               Data variables:
E                   floats        (a) float64 0.0 0.0 1.0 2.0 3.0
E                   floats_nans   (a) float64 nan nan 1.0 2.0 3.0
E                   bytes         (b) |S4 b'ab' b'cdef' b'g'
E                   bytes_nans    (b) |S4 b'ab' b'cdef' b''
E                   strings       (b) object 'ab' 'cdef' 'g'
E                   strings_nans  (b) object 'ab' 'cdef' ''
E                   all_nans      (c) float64 nan nan
E                   nan           float64 0.0

xarray/tests/test_backends.py:325: AssertionError
```

barronh: Maybe this is not pnc?

xarray/tests/test_backends.py
barronh: Okay, I think I get it. PNC allows use of netCDF4, but netCDF4 has to disable auto mask and scale to work with xarray.

Most formats supported by pnc (uamiv, ffi1001, bpch, arl packedbit) do not use CF masking and scaling. I can add a test-case file for one of these formats, but I'd still just end up disabling the test.

Please advise.

xarray/tests/test_backends.py
barronh: I agree that it is not designed to open arbitrary netcdf files. I could upload another test file (e.g., uamiv formatted) and disable this test... or just disable this test.

xarray/tests/test_backends.py
barronh: I figured out the pickle problem. So, this is no longer an issue.

shoyer:
OK, I pushed a couple of small changes to your branch. Generally this is looking pretty good. I have a couple of other minor comments that I will post inline.

Your issues with the skipped tests (which I switched to xfail) are part of a large issue in xarray, which is that we don't have good ways to handle backend specific decoding (https://github.com/pydata/xarray/issues/2061).

I agree that we can probably put this most of them for now (especially the string encoding), but I'm somewhat concerned about how decoding could end up with reading incorrect data values, e.g., if a source using scale/offset encoding. Consider this failed test:
```
    def test_roundtrip_mask_and_scale(self):
        decoded = create_masked_and_scaled_data()
        encoded = create_encoded_masked_and_scaled_data()
>           assert_allclose(decoded, actual, decode_bytes=False)
E           AssertionError: [ nan  nan 10.  10.1 10.2]
E           [  nan   nan 11.   11.01 11.02]
```

These sort of bugs can be pretty insidious, so if there's any chance that someone would use PNC to read a netCDF file with this sort of encoding we should try to fix this before merging this in.

One simple approach would be to raise an error for now if `mask_and_scale=True` in `open_dataset()`, that is, to force the user to explicitly disable masking and scaling with `xr.open_dataset(filename, engine='pseudonetcdf', mask_and_scale=False)`.

Alternatively, I suppose we could switch the default value to `mask_and_scale=None`, and pick `True` or `False` based on the choice of backend.

xarray/tests/test_backends.py
```diff
@@ -2243,60 +2243,53 @@ def open(self, path, **kwargs):
...
+    @pytest.mark.xfail(reason='https://github.com/pydata/xarray/issues/2061')
     def test_roundtrip_string_encoded_characters(self):
-        # pnc string treatment does not meet xarray expectations
-        # working on the right approach
         pass
 
-    def test_roundtrip_test_data(self):
-        expected = create_test_data()
-        with self.roundtrip(expected) as actual:
-            # dim3 sting type is <U instead of <U1
```
shoyer: The issue with `U` vs `U1` dtypes is something else, actually unrelated to this change. I'm going to try to fix it over in https://github.com/pydata/xarray/pull/2058

xarray/tests/test_backends.py
```diff
@@ -2243,60 +2243,53 @@ def open(self, path, **kwargs):
                           **kwargs) as ds:
             yield ds
 
+    def test_uamiv_format_read(self):
```
shoyer: Something in this test is triggering a warning when I run it:
```
xarray/tests/test_backends.py::PseudoNetCDFTest::test_uamiv_format_read
  /Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/PseudoNetCDF/conventions/ioapi/_ioapi.py:134: UserWarning: IOAPI_ISPH is assumed to be 6370000.; consistent with WRF
    warn('IOAPI_ISPH is assumed to be ' + ENV_IOAPI_ISPH + '; consistent with WRF')
```

Can you please silence this warning by using a context manager: https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings

doc/io.rst
```diff
@@ -650,7 +650,31 @@ We recommend installing PyNIO via conda::
...
+To use PseudoNetCDF to read such files, supply
+``engine='pseudonetcdf'`` to :py:func:`~xarray.open_dataset`.
+
+Add ``backend_kwargs={'format': '<format name>'}`` where `<format name>`
+options are listed on the PseudoNetCDF page.
+
+PseudoNetCDF is available via pip and has a recipe awaiting approval 
+that will enable installing PseudoNetCDF via conda::
+
+    conda install -c conda-forge PseudoNetCDF
```
shoyer: Please remove these paragraphs on installation here and add a brief note into the "Installation" docs page instead.

doc/whats-new.rst
```diff
@@ -181,6 +181,9 @@ Documentation
 
 Enhancements
 ~~~~~~~~~~~~
+- added backend for many Atmospheric data formats ranging including GEOS-Chem,
```
shoyer: "backend" -> "PseuodoNetCDF backend"? :)

doc/whats-new.rst
```diff
@@ -181,6 +181,9 @@ Documentation
...
+  CAMx, NOAA arlpacked bit and many others.
+- reduce methods such as :py:func:`DataArray.sum()` now accepts ``dtype``
```
shoyer: please delete

doc/whats-new.rst
```diff
@@ -181,6 +181,9 @@ Documentation
 
 Enhancements
 ~~~~~~~~~~~~
+- added backend for many Atmospheric data formats ranging including GEOS-Chem,
+  CAMx, NOAA arlpacked bit and many others.
```
shoyer: please add your name, if you're OK being credited for this PR.

Also, this section needs to be moved up on to docs for 0.10.4.

xarray/tests/test_backends.py
barronh: Done.

doc/whats-new.rst
barronh: Not sure what you're referring to since I am also editing and moving the first two lines? The third line was not mine.

doc/whats-new.rst
barronh: Done.

xarray/tests/test_backends.py
barronh: So, skip for now?

barronh:
Anything else needed?

doc/whats-new.rst
```diff
@@ -31,59 +31,38 @@ What's New
 v0.10.4 (unreleased)
 --------------------
 
+Documentation
+~~~~~~~~~~~~~
+
 Enhancements
 ~~~~~~~~~~~~
 
-- Support writing lists of strings as netCDF attributes (:issue:`2044`).
```
shoyer: please fix -- you are inadvertently editing release notes for 0.10.3.

shoyer:
@barronh pleas see my comment above: https://github.com/pydata/xarray/pull/1905#issuecomment-381467470

I would be OK not applying masking/scaling at all. But applying masking/scaling *twice* by default seems problematic and hard to debug. If there's any chance someone will use psuedonetcdf to access a dataset with these attributes, we need to fix this behavior.

doc/whats-new.rst
barronh: Sorry. Fixed it.

barronh: Fixed.

barronh:
I tried disabling mask and scale, but many other tests fail. At its root this is because I am implicitly supporting netCDF4 and other formats.

I see two ways to solve this. Right now, it is only important to add non-netcdf support to xarray via PseudoNetCDF. I am currently allowing dynamic identification of the file format, which implicitly supports netCDF. I could disable implicit format support, and require the format keyword. In that case, PseudoNetCDF tests no longer should be CFEncodedDataTest. Instead, I can simply test some round tripping with the other formats (uamiv and possibly adding one or two other formats).

What do you think?

barronh:
> I trust that none of the other formats PNC supports use _FillValue, add_offset or scale_factor attributes?

The challenge here is that if specific formats use a similar functionality, it has already been applied and may or may not use the CF keywords. So, it should be disabled by default.

> If it is possible to detect the inferred file format from PNC, then another option (other than requiring the explicit format argument) would be to load the data and raise an error if the detected file format is netCDF.

I was worried this would be hard to implement, but it was actually easier. So that is what I did.

barronh:
Depends on the format and expected size of data in that format. Some formats support lazy; others load immediately into memory.  Sill others use memmaps, so that virtual memory is used immediately.

xarray/backends/api.py
```diff
@@ -222,6 +228,10 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     --------
     open_mfdataset
     """
+    
+    if mask_and_scale is None:
+        mask_and_scale = not engine == 'pseudonetcdf'
```
shoyer: Actually, let's do this for `decode_cf` instead of `mask_and_scale`. I don't think the other netCDF specific decoding like concatenating characters is relevant to psuedonetcdf, either.

Also, let's be sure to change this for `open_mfdataset` (and `open_dataarray`) as well.

xarray/backends/pseudonetcdf_.py
```diff
@@ -0,0 +1,121 @@
...
+        return lhs not in _genericncf
+
+
+class PseudoNetCDFDataStore(AbstractDataStore, DataStorePickleMixin):
+    """Store for accessing datasets via PseudoNetCDF
+    """
+    @classmethod
+    def open(cls, filename, format=None, writer=None,
+             autoclose=False, **format_kwds):
+        from PseudoNetCDF._getreader import getreader, getreaderdict
```
shoyer: Importing from private APIs (with the underscore prefix) makes me nervous. In general we try to avoid doing this because private APIs can change at any time, though of course that isn't true here because you're the PNC maintainer. I'm also nervous about including constants like `_genericncf = ('Dataset', 'netcdf', 'ncf', 'nc')` in xarray because that feels like logic that should live in libraries like PNC instead.

So maybe we should stick with the standard open method that you were using earlier. In particular, if we change the default to `decode_cf=False` for PNC in `open_dataset()` then I don't think it's essential to check the format.

xarray/backends/pseudonetcdf_.py
```diff
@@ -0,0 +1,121 @@
...
+        readerdict = getreaderdict()
+        _genreaders = tuple([readerdict[rn] for rn in _genericncf
+                             if rn in readerdict])
+        reader = getreader(filename, format=format, **format_kwds)
+        if isinstance(reader, _genreaders):
+            raise ValueError(('In xarray, PseudoNetCDF should not be used ' +
```
shoyer: I think we can avoid this check/error message if we default to `decode_cf=False` for PNC in `open_dataset()`.

xarray/tests/test_backends.py
```diff
@@ -2236,6 +2236,209 @@ class PyNioTestAutocloseTrue(PyNioTest):
...
+        pncf.variables = {k: pnc.PseudoNetCDFVariable(pncf, k, v.dtype.char,
+                                                      v.dims,
+                                                      values=v.data[...],
+                                                      **v.attrs)
+                          for k, v in dataset.variables.items()}
+        for pk, pv in dataset.attrs.items():
+            setattr(pncf, pk, pv)
+
+        pnc.pncwrite(pncf, path, **save_kwargs)
+
```
shoyer: Would it make sense to add a test for `open_mfdataset` with PNC if that's something we want to support?

xarray/backends/api.py
barronh: Next push:
* I switched from `mask_and_scale` to `decode_cf`.
* `open_mfdataset` uses a pass-through of keywords.
* `open_dataarray` has been updated

FYI - This leads to an odd outcome -- if I want to `decode_times`, I have to explicitly enable `decode_cf` and then disable all the others individually, *and* enable `decode_times`.

xarray/backends/pseudonetcdf_.py
barronh: I think it will be confusing for users to have an indirect use of `netCDF4` via `PseudoNetCDF` that responds differently than the typical backend. This code is designed to address that xarray-specific issue. Typical behavior in PNC is to mask_and_scale via netCDF4's default functionality, so with `decode_cf` a netCDF file would still be decoded. This is confusing to me, let alone a user. For this reason, I think it is right to exclude formats that already have specifically optimized backends.

I agree completely about private API. The standard pncopen, however, does not currently support the ability to enable a subset of readers. I will add this a feature request to PseudoNetCDF, and make a subsequent update there. As soon as I update pncopen, I would make the update here.

The `_genericncf` is not logic that should live in PNC. It is logic designed specifically to address an xarray issue. In this case, `xarray` has a feature that should pre-empt PNC typical capabilities.

The alternative is to update PNC, update the conda installation, and update this code again.

xarray/backends/pseudonetcdf_.py
barronh: see previous comment.

xarray/tests/test_backends.py
barronh: I think so. I'll add that to my next push.

xarray/backends/api.py
shoyer: Oh... I didn't realize you do want to decode times. In that case, I guess disabling only mask_and_scale is a reasonable choice.

xarray/backends/pseudonetcdf_.py
shoyer: Are you thinking of someone explicitly calling `xarray.decode_cf()` on a dataset loaded from PNC? I agree that that would give unexpected results, but I don't see that being done very often.

Another way to make all of this safe (and which wouldn't require disabling `mask_and_scale` in `open_dataset()`) would be to explicitly drop `_FillValue`, `missing_value`, `add_offset` and `scale_factor` attributes from variables loaded with PNC, moving them to `Variable.encoding` instead.

barronh: Imagine that you have a netcdf file. If you open it with PNC, it uses netCDF4 as the library to open it. In xarray, however, you might get a different processing. If you write it out, it might have different properties. This seems undesirable at best. That is why I prefer removing the netCDF functionality.

In PNC, the goal of using netCDF4 is to make sure all pnc functionality is available for any type of file. In PNC, some plotting and processing tools are available (similar to xarray). I think plotting and processing in xarray is often better than PNC. The goal of adding PNC to xarray is to bring xarray formats that PNC can read. Since xarray already has mechanisms for supporting netCDF, I think bringing this functionality is unnecessary.

barronh:
The two failures were not pnc:

```
xarray/tests/test_backends.py::TestRasterio::test_serialization FAILED   [ 26%]
...
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_open_dataarray_options FAILED [ 26%]
```

Neither seems related to PNC at all...

I'm pulling the master and remerging to see if that helps...

xarray/backends/pseudonetcdf_.py
shoyer: My general feeling is that we should make it easy to do the right thing, but not put up artificial barriers to stop expert users. I think this is broadly consistent with most Python projects, including the Python language itself. Even if it's strange, somebody might have reasons for trying PNC/netCDF4 together, and as long as it will not give wrong results there is no reason to disable it.

That said, my real objection here are all the private imports from PNC, and the four different aliases we use to check for netCDF files. If you want to put a public API version of this in PNC before merging this PR, that would be OK with me.

barronh: Okay. I removed the type checking and private imports. I updated the testing. 

I get one failed that I cannot identify as related to PNC: `test_open_mfdataset_manyfiles` files with:

```
E           OSError: [Errno 24] Too many open files: '/var/folders/g2/hwlpd21j4vl8hvldb4tyrz400000gn/T/tmp0bteny75'
```

I'm going to run it through the build to see if it is just my setup.

barronh: It didn't fail on the automated test, so all passing [x] and no private imports [x].

barronh:
@shoyer - Thanks for all the help and guidance. I learned a lot. In the development version of pnc, it is now flake8 compliant. I now use pytest to organize my unittests. I have TravisCI implemented. I have a conda-forge recipe. 

I’m grateful for the interactions.

bbakernoaa:
@barronh Truly great work. I’m very excited about this and think it will do wonders for the climate and AQ data analysis. 

@shoyer thanks for being patient and help with getting this going.

PR #1907, opened 2018-02-13
drop zarr variable name from the dask chunk name

### LINKED ISSUE #1894: Zarr keys include variable name
When using open_zarr on a dataset with many variables the keynames include the variable name, like 

    ('zarr-temperature-1234', 1, 3, 2)

In the distributed scheduler these keynames get shortened to prefixes like `zarr-temperature` and used both for scheduling heuristics (all keys with the same prefix are expected to take similar-ish amounts of time) and for diagnostics, such as in the progress bar plot below:

image

We may want to avoid including the variable name into the keyname here in order to avoid breaking these out into several groups.  Instead you might consider putting the variable name within the key as another member of the tuple like the following:

    ('zarr-1234', 'temperature', 1, 3 ,2)

jhamman:
- [x] Closes #1894 

cc @mrocklin

### DIFF
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -482,7 +482,7 @@ def open_zarr(store, group=None, synchronizer=None, auto_chunk=True,
             if (var.ndim > 0) and (chunks is not None):
                 # does this cause any data to be read?
                 token2 = tokenize(name, var._data)
-                name2 = 'zarr-%s-%s' % (name, token2)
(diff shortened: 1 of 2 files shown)


jhamman:
I thought about this when I was making the change. I also don't think it requires a test. Obviously we weren't testing this before and its really just a cosmetic change. If anyone else feels differently, I can certainly look into writing a simple test.

PR #1911, opened 2018-02-15
DOC: Removed a line in setting up toy weather data

raybellwaves:


jhamman:
@raybellwaves - what is the motivation for removing this line? Can you provide some context to the change?

raybellwaves:
Apologies. Closing. I thought it was a rouge line as it was not declared to a variable and I couldn't see that it was being using anywhere. But I was not familiar with [np.random.seed](https://stackoverflow.com/questions/21494489/what-does-numpy-random-seed0-do) but I am now,

PR #1915, opened 2018-02-15
h5netcdf new API support

### LINKED ISSUE #1536: Better compression algorithms for NetCDF
As of today, ``Dataset.to_netcdf()`` exclusively allows writing uncompressed or compressed with zlib.
zlib was absolutely revolutionary when it was released... in 1995. Time has passed, and much better compression algorithms have appeared over time.
Good news is, h5py supports LZF out of the box, and is extensible with plugins to support theoretically any other algorithm. h5netcdf exposes such interface through its new (non-legacy) API; however ``Dataset.to_netcdf(engine='h5netcdf')`` supports the legacy API exclusively. 

I already tested that, once you manage to write to disk with LZF (using h5netcdf directly), ``open_dataset(engine='h5netcdf')`` transparently opens the compressed store.

Options:
- write a new engine for ``Dataset.to_netcdf()`` to support the new h5netcdf API.
- switch the whole ``engine='h5netcdf'`` to the new API. Drop support for the old parameters in ``to_netcdf()``. This is less bad than it sounds, as people can switch to another engine in case of trouble. This is the cleanest solution, but also the most disruptive one.
- switch the whole ``engine='h5netcdf'`` to the new API; have ``to_netcdf()`` accept both new and legacy parameters, and implement a translation layer of parameters from the legacy API to the new API. The benefit here is that, as long as the user sticks to the legacy API, he can hop between engines transparently. On the other hand I have a hard time believing anybody would care.
- ?

crusaderky:
Closes #1536

Support arbitrary compression plugins through the h5netcdf new API.

Done:
- public API and docstrings (untested)
- implementation
- unit tests
- What's New


jhamman:
@crusaderky - if you sync with master, you'll stop getting the stickler-ci messages.

xarray/backends/hdf5.py
```diff
@@ -0,0 +1,190 @@
+from __future__ import absolute_import
```
crusaderky: TODO - didn't do anything useful here yet, besides copy-pasting from h5netcdf_.py and renaming variables

xarray/backends/h5netcdf_ng.py
```diff
@@ -0,0 +1,197 @@
+from __future__ import absolute_import, division, print_function
```
crusaderky: TODO
I didn't do anything in this module yet - this is still using the old API.

shoyer:
@crusaderky Yes, I like the look of that better. It's slightly unfortunate that it's taking a while for the netCDF4 library to figure this out so it's not clear the API will be yet. We might not want to keep the separate new API backend with h5netcdf in the long term, but we certainly would not remove/deprecate it until everything is supported with the legacy API.

One minor concern is the name. I don't know what `h5netcdf-ng` stands for ("next generation?"). I would prefer `h5netcdf-new`.

xarray/backends/h5netcdf_new.py
```diff
@@ -0,0 +1,197 @@
+from __future__ import absolute_import, division, print_function
```
crusaderky: No actual work done on this module yet

crusaderky:
Almost finished - only one outstanding unit test failure when specifying the "chunks" parameter

crusaderky:
@shoyer finished and ready for final code review and integration

shoyer:
Generally looks great. 

What if both `zlib=True` *and* `compression='lzf'` is set in encoding? This could potentially happen, e.g., if reading from a file with zlib encoding and attempting to write as lzf. I think the `compression` argument should take a precedence, but we should add a test case to verify that this works.

crusaderky:
@shoyer done

crusaderky:
@shoyer / @jhamman , stickler-ci has been pending for a day now - I'm afraid it should be pronounced dead

jhamman:
closing/reopening seemed to do the trick for @stickler-ci.

xarray/backends/h5netcdf_.py
```diff
@@ -175,10 +175,12 @@ def prepare_variable(self, name, variable, check_encoding=False,
                                           raise_on_invalid=check_encoding)
         kwargs = {}
 
-        if encoding.get('zlib'):
-            encoding['compression'] = 'gzip'
-            del encoding['zlib']
-            encoding['compression_opts'] = encoding.pop('complevel', None)
+        # Convert from NetCDF4-Python style compression settings to h5py style
+        if encoding.pop('zlib', False):
+            # If both styles are used together, h5py takes precedence
```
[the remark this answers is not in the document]
crusaderky: @shoyer I changed the code, but I don't know how to write a test for it?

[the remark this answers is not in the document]
crusaderky: @shoyer ping - please advise how I should write a unit test for ``check_encoding=True`` - is it something that can be requested from ``DataArray.to_netcdf``?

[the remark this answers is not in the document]
shoyer: Yes, `check_encoding=True` is triggered when `encoding` is supplied via `to_netcdf()` rather than on variables directly.

For example, this should be an error: `ds.to_netcdf(..., encoding={'foo': {'zlib': True', 'compression': 'other'}})`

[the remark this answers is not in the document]
crusaderky: done

crusaderky:
@shoyer all green and ready for merge

xarray/backends/h5netcdf_.py
```diff
@@ -166,18 +175,38 @@ def prepare_variable(self, name, variable, check_encoding=False,
                                           raise_on_invalid=check_encoding)
         kwargs = {}
 
-        for key in ['zlib', 'complevel', 'shuffle',
-                    'chunksizes', 'fletcher32']:
+        # Convert from NetCDF4-Python style compression settings to h5py style
+        # If both styles are used together, h5py takes precedence
+        # If set_encoding=True, raise ValueError in case of mismatch
+        if encoding.pop('zlib', False):
+            if check_encoding and encoding.get('compression') \
```
shoyer: nit: per pep8, prefer uses parentheses over explicit line continuation `\`.

xarray/backends/h5netcdf_.py
```diff
@@ -166,18 +175,38 @@ def prepare_variable(self, name, variable, check_encoding=False,
...
         kwargs = {}
 
-        for key in ['zlib', 'complevel', 'shuffle',
-                    'chunksizes', 'fletcher32']:
+        # Convert from NetCDF4-Python style compression settings to h5py style
+        # If both styles are used together, h5py takes precedence
+        # If set_encoding=True, raise ValueError in case of mismatch
+        if encoding.pop('zlib', False):
+            if (check_encoding and encoding.get('compression')
+                    not in (None, 'gzip')):
```
shoyer: note: per PEP8, I replaced the explicit continuation `\` with parentheses.

shoyer:
Thanks @crusaderky !

PR #1916, opened 2018-02-16
COMPAT: MultiIndex checking is fragile (#1833)

### LINKED ISSUE #1833: COMPAT: MultiIndex checking is fragile
https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L2755

checking for a MI via 
```
if hasattr(idx, 'levels'):
    # is a MI
```

should be changed to
```
if idx.nlevels > 1:
    # is a MI
```

see [here](https://github.com/pandas-dev/pandas/issues/3268#issuecomment-357925936). This is not currently implemented but good to get ahead of this hard to detect usage of the API.

floriankrb:
- [X] Closes #1833
 - [ 0] Tests added 
 - [ 0] Tests passed

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -492,6 +492,9 @@ Bug fixes
   ``apionly`` module was deprecated.
   (:issue:`1633`). By `Joe Hamman <https://github.com/jhamman>`_.
 
+- Fix COMPAT: MultiIndex checking is fragile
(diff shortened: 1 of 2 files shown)


fujiisoup:
Thanks for sending a PR. 
The second commit (ec3498f) looks a duplicate of #1918 .
Could you reset this?

shoyer:
Indeed, thanks!

PR #1918, opened 2018-02-16
`axis` keyword ignored when applying `np.squeeze` to `DataArray` (#1487)

### LINKED ISSUE #1487: `axis` keyword ignored when applying `np.squeeze` to `DataArray`
Thank you all for creating xarray, I'm a new user and I love it! 

I believe I came across a bug. It appears the axis keyword is ignored when applying `np.squeeze` to a `DataArray`. Here is an illustration:

```
import numpy as np
import xarray as xr

test = np.random.random((1,10,1))
np.squeeze(test, axis=0).shape
# (10, 1)
np.squeeze(xr.DataArray(test), axis=0).shape
# (10,)
np.__version__
# '1.13.0'
xr.__version__
# '0.9.6'
```

floriankrb:
- [X] Closes #1487 
 - [x] Tests added
 - [x] Tests passed 


xarray/core/common.py
```diff
@@ -211,14 +211,17 @@ def _ipython_key_completions_(self):
...
+def get_squeeze_dims(xarray_obj, dim, axis=None):
     """Get a list of dimensions to squeeze out.
     """
-    if dim is None:
+    if dim is None and axis is None:
         dim = [d for d, s in xarray_obj.sizes.items() if s == 1]
     else:
         if isinstance(dim, basestring):
             dim = [dim]
+        if isinstance(axis, int):
```
fujiisoup: `np.squeeze` can accept a tuple of ints, also.

xarray/core/common.py
```diff
@@ -211,14 +211,17 @@ def _ipython_key_completions_(self):
         return list(set(item_lists))
 
 
-def get_squeeze_dims(xarray_obj, dim):
+def get_squeeze_dims(xarray_obj, dim, axis=None):
     """Get a list of dimensions to squeeze out.
     """
-    if dim is None:
+    if dim is None and axis is None:
```
fujiisoup: Maybe we can raise an Error if dim and axis contradict each other?

xarray/core/common.py
```diff
@@ -211,14 +211,27 @@ def _ipython_key_completions_(self):
...
+        if isinstance(axis, int):
+            alldims = list(xarray_obj.sizes.keys())
+            dim = [alldims[axis]]
+        if isinstance(axis, tuple):
+            for a in axis:
+                if not isinstance(a, int):
+                    raise ValueError(
+                        'parameter `axis` must be int or tuple of int.')
+            alldims = list(xarray_obj.sizes.keys())
+            dim = [alldims[a] for a in axis]
```
fujiisoup: I think this can be simpler
`dim = [xarray_obj.dims[a] for a in axis]`

xarray/core/common.py
```diff
@@ -211,14 +211,27 @@ def _ipython_key_completions_(self):
...
+        raise ValueError('cannot use both parameters `axis` and `dim`')
+
+    if dim is None and axis is None:
         dim = [d for d, s in xarray_obj.sizes.items() if s == 1]
     else:
         if isinstance(dim, basestring):
             dim = [dim]
+        if isinstance(axis, int):
+            alldims = list(xarray_obj.sizes.keys())
+            dim = [alldims[axis]]
```
fujiisoup: I think `axis = (axis, )` here is cleaner.

xarray/core/common.py
shoyer: `xarray_obj` can be either a Dataset or a DataArray, so I think this is indeed necessary.

xarray/tests/test_dataarray.py
```diff
@@ -1668,6 +1668,11 @@ def test_squeeze_drop(self):
         actual = array.squeeze(drop=False)
         assert_identical(expected, actual)
 
+        array = DataArray([[[0., 1.]]], dims=['dim_0', 'dim_1', 'dim_2'])
+        expected = DataArray([[0., 1.]], dims=['dim_1', 'dim_2'])
+        actual = array.squeeze(axis=0)
+        assert_identical(expected, actual)
+
```
fujiisoup: For the better coverage, can you add tests for tuple axis case?
Also, it would be nice if we could have tests to make sure our code certainly raises an Error if both dim and axis are given.

xarray/core/common.py
fujiisoup: You are right. Let's keep this line.

floriankrb:
Here it is. Thanks for your code reviews and your helpful comments !

fujiisoup:
Thanks, @floriankrb !

PR #1252, opened 2017-02-06
WIP: NetCDFTimeIndex

spencerkclark:
- [ ] closes #1084
 - [ ] tests added / passed
 - [ ] passes ``git diff upstream/master | flake8 --diff``
 - [ ] whatsnew entry

This work in progress PR is a start on implementing a ```NetCDFTimeIndex```, a subclass of pandas.Index, which closely mimics pandas.DatetimeIndex, but uses ```netcdftime._netcdftime.datetime``` objects.  Currently implemented in the new index are:
- Partial datetime-string indexing (using strictly [ISO8601-format strings](https://en.wikipedia.org/wiki/ISO_8601), using a date parser implemented by @shoyer in https://github.com/pydata/xarray/issues/1084#issuecomment-274372547)
- Field-accessors for year, month, day, hour, minute, second, and microsecond, to enable ```groupby``` operations on attributes of date objects

This index is meant as a step towards improving the handling of non-standard calendars and dates outside the range ```Timestamp('1677-09-21 00:12:43.145225')``` to ```Timestamp('2262-04-11 23:47:16.854775807')```.

--------------

For now I have pushed only the code and some tests for the new index; I want to make sure the index is solid and well-tested before we consider integrating it into any of xarray's existing logic or writing any documentation.  

Regarding the index, there are a couple remaining outstanding issues (that at least I'm aware of):

1. Currently one can create non-sensical datetimes using ```netcdftime._netcdftime.datetime``` objects.  This means one can attempt to index with an out-of-bounds string or datetime without raising an error.  Could this possibly be addressed upstream?  For example:
```
In [1]: from netcdftime import DatetimeNoLeap

In [2]: DatetimeNoLeap(2000, 45, 45)
Out[2]: netcdftime._netcdftime.DatetimeNoLeap(2000, 45, 45, 0, 0, 0, 0, -1, 1)
```
2. I am looking to enable this index to be used in pandas.Series and pandas.DataFrame objects as well;  this requires implementing a ```get_value``` method.  I have taken @shoyer's suggested simplified approach from https://github.com/pydata/xarray/issues/1084#issuecomment-275963433, and tweaked it to also allow for slice indexing, so I think this is most of the way there.  A remaining to-do for me, however, is to implement something to allow for integer-indexing outside of ```iloc```, e.g.  if you have a pandas.Series ```series```, indexing with the syntax ```series[1]``` or ```series[1:3]```.

[...]

Issue #1084, opened 2016-11-06
Towards a (temporary?) workaround for datetime issues at the xarray-level


spencerahill:
Re: #789.  The consensus is that upstream fixes in Pandas are not coming anytime soon, and there is an acute need amongst many xarray users for a workaround in the meantime.  There are two separate issues: (1) date-range limitations due to nanosecond precision, and (2) support for non-standard calendars.

@shoyer, @jhamman , @spencerkclark, @darothen, and I briefly discussed offline a potential workaround that I am (poorly) summarizing here, with hope that others will correct/extend my snippet.

The idea is to extend either PeriodIndex or (more involved but potentially more robust) Int64Index, either through subclassing or composition, to implement all of the desired functionality: slicing, resampling, groupby, and serialization.

For reference, @spencerkclark nicely summarized the limitations of PeriodIndex and the netCDF4.datetime objects, which are often used as workarounds currently: https://github.com/spencerahill/aospy/issues/98#issuecomment-256043833

shoyer:
To be clear, the subclassing question is whether we should subclass Index or not. This would entail overriding a lot of methods, and would probably break in the future (with pandas 2.0). I definitely do not recommend subclassing PeriodIndex -- that could be very fragile.

spencerkclark:
@shoyer @jhamman this is my first time digging in to the internals of pandas / xarray, so please forgive me if this is off-track, but I've started experimenting with a subclass of ```pandas.Index``` (which I've called ```NetCDFTimeIndex```) that could be a start toward addressing this issue.      Here's a [link to a Jupyter notebook that sketches it out](https://nbviewer.jupyter.org/gist/spencerkclark/158ddbd0b0df7de1332020429e4c97c2), and includes a few examples of its use.

[...]

PR #1920, opened 2018-02-16
Add netcdftime as an optional dependency. 

jhamman:
- [x] Helps with #1084 

I've added a temporary travis build with the master branch of `netcdftime`. After a while, we can probably remove that. 

This is helping us move towards https://github.com/Unidata/netcdf4-python/issues/601 and #1252

cc @jswhit and @spencerkclark


jhamman:
@spencerkclark, mind giving this a review?

PR #1924, opened 2018-02-20
isort

maxim-lian:
Not sure if we want this?

Probably too strict to enforce on every commit, more permissible to do a one-time update, assuming it doesn't cause merge conflicts. 

You can get the same result by running `isort -y`

### DIFF
--- a/xarray/tests/test_coding_times.py
+++ b/xarray/tests/test_coding_times.py
@@ -1,16 +1,14 @@
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+from __future__ import absolute_import, division, print_function
 
 import warnings
 
 import numpy as np
 import pandas as pd
+import pytest
 
 from xarray import Variable, coding
-from . import (
-    TestCase, requires_netCDF4, assert_array_equal)
-import pytest
+
+from . import TestCase, assert_array_equal, requires_netCDF4
 
 
 @np.vectorize
--- a/xarray/tests/test_coding_times.py
+++ b/xarray/tests/test_coding_times.py
@@ -9,8 +9,7 @@ import pytest
 from xarray import Variable, coding
 from xarray.coding.times import _import_netcdftime
 
-from . import (
-    TestCase, assert_array_equal, requires_netCDF4, requires_netcdftime)
+from . import TestCase, assert_array_equal, requires_netcdftime
 
 
 @np.vectorize
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -2,7 +2,7 @@ linters:
   flake8:
     max-line-length: 79
     fixer: true
-    ignore: I002
+    ignore: 
     exclude:
       - 'doc/'
   py3k:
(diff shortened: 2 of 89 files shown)


maxim-lian:
Requires https://github.com/pydata/xarray/pull/1925 to make @stickler-ci happy (and this needs to stop the `I002` disabling when that is merged)

maxim-lian:
Updated

shoyer:
Is it possible to enforce this programmatically, e.g., with flake8?

I'm a little surprised flake8 doesn't already require sorted imports by default. We use a modified version of pylint internally at Google that does this.

xarray/tests/test_coding_times.py
```diff
@@ -1,17 +1,16 @@
...
 import pandas as pd
+import pytest
 
 from xarray import Variable, coding
 from xarray.coding.times import _import_netcdftime
+
 from . import (
-    TestCase, requires_netcdftime, assert_array_equal)
-import pytest
+    TestCase, assert_array_equal, requires_netCDF4, requires_netcdftime)
```
jhamman: if you remove `requires_netCDF4`, I think the flake8 tests will pass.

dopplershift:
I recommend [flake8-import-order](https://github.com/PyCQA/flake8-import-order). If you install that plugin, then flake8 will enforce import ordering and grouping.

maxim-lian:
We use `flake8-isort`, which implements `isort` itself in `flake8`. @dopplershift do you have a view between them?

One advantage of `isort` is that there's a link between the formatter and the checker, so running `isort` can never cause errors.

But stepping back, we should think about whether we want to mandate this on every PR. It doesn't matter _that_ much (less than flake8 imo), and it can be frustrating to have formatting errors get in the way of a PR. A tool that automatically corrects the code like @stickler-ci is great, but it doesn't support this one. 

A compromise could be to put it in allowed failures in Travis? Or if people think we should add this, let's do it.

dopplershift:
I don't have much preference between the two--I picked one and it worked for me. I just fixed up the initial import cleanups by hand and didn't find the need for a tool to do so.

In my experience, having formatting errors hold up a PR hasn't been a problem. IMO, import ordering is a more important style issue than many of the things that flake8 catches.

shoyer:
I would also be OK with just fixing this up in occasional cleanup PR.

If Stickler-CI supported this that might be a decisive consideration in terms of enforcing it. Certainly it's a better experience for contributors to get formatting errors as comments on a PR rather than a failed build.

maxim-lian:
> a better experience for contributors to get formatting errors as comments on a PR

+ the really cool feature is that it'll add commits with changes, where it can

maxim-lian:
> I would also be OK with just fixing this up in occasional cleanup PR.

Would you prefer in allowed failures, or that's a bad middle-ground?

maxim-lian:
@jhamman @shoyer green

maxim-lian:
@jhamman @shoyer Green!

No auto-check

PR #1926, opened 2018-02-20
Add x,y kwargs for plot.line().

### LINKED ISSUE #575: 1D line plot with data on the x axis

dcherian:
- [x] Partially closes #575 
 - [x] Tests added
 - [x] Tests passed

### Description
`plot.line` now supports both 1D and 2D DataArrays as input. I've changed  some variable names to make code clearer:
   1. set `xplt`, `yplt` to be values that are passed to `ax.plot()`
   2. `xlabel`, `ylabel` are axes labels
   3. `xdim`, `ydim` are dimension names

### Example
This code
```z = np.arange(10)
da = xr.DataArray(np.cos(z), dims=['z'], coords=[z], name='f')

xy = [[None, None],
     [None, 'f'],
     [None, 'z'],
     ['f', None],
      ['z', None],
      ['z', 'f'],
      ['f', 'z']]

f, ax = plt.subplots(2,4)

for aa, (x,y) in enumerate(xy):
    da.plot(x=x, y=y, ax=ax.flat[aa])
    ax.flat[aa].set_title('x='+str(x)+ ' | '+'y='+str(y))
```

yields

image

### Feedback requested

Should I refactor out the kwarg checking?

### DIFF
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -197,10 +197,10 @@ It is required to explicitly specify either
 Thus, we could have made the previous plot by specifying ``hue='lat'`` instead of ``x='time'``.
 If required, the automatic legend can be turned off using ``add_legend=False``.
 
-Coordinate along y-axis
-~~~~~~~~~~~~~~~~~~~~~~~
+Dimension along y-axis
+~~~~~~~~~~~~~~~~~~~~~~
 
-It is also possible to make line plots such that the data are on the x-axis and a coordinate is on the y-axis. This can be done by specifying the ``x`` and ``y`` keyword arguments.
+It is also possible to make line plots such that the data are on the x-axis and a dimension is on the y-axis. This can be done by specifying the appropriate ``y`` keyword argument.
 
 .. ipython:: python
 
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -197,6 +197,16 @@ It is required to explicitly specify either
 Thus, we could have made the previous plot by specifying ``hue='lat'`` instead of ``x='time'``.
 If required, the automatic legend can be turned off using ``add_legend=False``.
 
+Data along x-axis
+~~~~~~~~~~~~~~~~~
+
+It is also possible to make line plots such that the data are on the x-axis and a co-ordinate is on the y-axis. This can be done by specifying the ``x`` and ``y`` keyword arguments.
+
+.. ipython:: python
+
+    @savefig plotting_example_xy_kwarg.png
+    air.isel(time=10, lon=[10, 11]).plot.line(y='lat', hue='lon')
+
 Two Dimensions
 --------------
 
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -197,10 +197,10 @@ It is required to explicitly specify either
 Thus, we could have made the previous plot by specifying ``hue='lat'`` instead of ``x='time'``.
 If required, the automatic legend can be turned off using ``add_legend=False``.
 
-Data along x-axis
-~~~~~~~~~~~~~~~~~
+Coordinate along y-axis
+~~~~~~~~~~~~~~~~~~~~~~~
 
-It is also possible to make line plots such that the data are on the x-axis and a co-ordinate is on the y-axis. This can be done by specifying the ``x`` and ``y`` keyword arguments.
+It is also possible to make line plots such that the data are on the x-axis and a coordinate is on the y-axis. This can be done by specifying the ``x`` and ``y`` keyword arguments.
 
 .. ipython:: python
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -141,6 +141,8 @@ Enhancements
   encoding/decoding of datetimes with non-standard calendars without the
   netCDF4 dependency (:issue:`1084`).
   By `Joe Hamman <https://github.com/jhamman>`_.
+- :py:func:`~plot.line()` learned to make plots with data on x-axis if so specified.
+  By `Deepak Cherian <https://github.com/dcherian>`_.
 
 .. _Zarr: http://zarr.readthedocs.io/
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -141,7 +141,7 @@ Enhancements
   encoding/decoding of datetimes with non-standard calendars without the
   netCDF4 dependency (:issue:`1084`).
   By `Joe Hamman <https://github.com/jhamman>`_.
-- :py:func:`~plot.line()` learned to make plots with data on x-axis if so specified.
+- :py:func:`~plot.line()` learned to make plots with data on x-axis if so specified. (:issue:`575`)
   By `Deepak Cherian <https://github.com/dcherian>`_.
 
 .. _Zarr: http://zarr.readthedocs.io/
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -46,6 +46,8 @@ Enhancements
   rolling, windowed rolling, ND-rolling, short-time FFT and convolution.
   (:issue:`1831`, :issue:`1142`, :issue:`819`)
   By `Keisuke Fujii <https://github.com/fujiisoup>`_.
+- :py:func:`~plot.line()` learned to make plots with data on x-axis if so specified. (:issue:`575`)
+  By `Deepak Cherian <https://github.com/dcherian>`_.
 
 Bug fixes
 ~~~~~~~~~
@@ -141,8 +143,6 @@ Enhancements
   encoding/decoding of datetimes with non-standard calendars without the
   netCDF4 dependency (:issue:`1084`).
   By `Joe Hamman <https://github.com/jhamman>`_.
-- :py:func:`~plot.line()` learned to make plots with data on x-axis if so specified. (:issue:`575`)
-  By `Deepak Cherian <https://github.com/dcherian>`_.
 
 .. _Zarr: http://zarr.readthedocs.io/
 
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -206,7 +206,7 @@ def line(darray, *args, **kwargs):
     ax = get_axis(figsize, size, aspect, ax)
 
     error_msg = ('must be either None or one of ({0:s})'
-                 .format(', '.join(['\'' + dd + '\'' for dd in darray.dims])))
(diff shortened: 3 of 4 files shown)


doc/plotting.rst
```diff
@@ -197,6 +197,16 @@ It is required to explicitly specify either
 Thus, we could have made the previous plot by specifying ``hue='lat'`` instead of ``x='time'``.
 If required, the automatic legend can be turned off using ``add_legend=False``.
 
+Data along x-axis
```
fmaussion: "Rotated line plots" as title? Just a suggestion

doc/plotting.rst
```diff
@@ -197,6 +197,16 @@ It is required to explicitly specify either
...
+~~~~~~~~~~~~~~~~~
+
+It is also possible to make line plots such that the data are on the x-axis and a co-ordinate is on the y-axis. This can be done by specifying the ``x`` and ``y`` keyword arguments.
```
fmaussion: "co-ordinate" -> coordinate

doc/whats-new.rst
```diff
@@ -97,6 +97,8 @@ Enhancements
   encoding/decoding of datetimes with non-standard calendars without the
   netCDF4 dependency (:issue:`1084`).
   By `Joe Hamman <https://github.com/jhamman>`_.
+- :py:func:`~plot.line()` learned to make plots with data on x-axis if so specified.
```
fmaussion: add ``(:issue:`575`)``

xarray/plot/plot.py
```diff
@@ -175,9 +175,13 @@ def line(darray, *args, **kwargs):
...
     hue : string, optional
-        Coordinate for which you want multiple lines plotted (2D inputs only).
+        Coordinate for which you want multiple lines plotted
+        (2D DataArrays only).
     x : string, optional
-        Coordinate for x axis.
+        1D and 2D DataArrays: Coordinate for x axis.
+    y : string, optional
+        1D DataArray: Can be coordinate name or DataArray.name
+        2D DataArray: Coordinate for y axis.
```
fmaussion: I'm not sure this is really helping. I'd prefer to stick to the simple ``Coordinate for x/y axis.``

xarray/plot/plot.py
```diff
@@ -198,43 +202,78 @@ def line(darray, *args, **kwargs):
...
+            raise ValueError('x ' + error_msg)
+
+        if y not in [None, dim, darray.name]:
+            raise ValueError('y ' + error_msg)
 
-        x = darray.coords[xlabel]
+        if x is not None and y is not None and x == y:
+            raise ValueError('Cannot make a plot with x=%r and y=%r' % (x, y))
+
+        if (x is None and y is None) or x == dim or y is darray.name:
```
fmaussion: ``y is darray.name`` -> ``y == darray.name``

xarray/plot/plot.py
```diff
@@ -198,43 +202,78 @@ def line(darray, *args, **kwargs):
...
+            xplt = darray.coords[dim]
+            yplt = darray
+            xlabel = dim
+            ylabel = darray.name
+
+        elif y == dim or x is darray.name:
```
fmaussion: here too

dcherian:
@fmaussion when I added the `hue` argument, @shoyer suggested that we have a "fully explicit way to make these plots". 
In this case this would mean
```da.plot(x='temp', y='time', hue='lat')``` 
which seems a lot more readable than 
```da.plot(y='coordinate', hue='lat')```

It would also mean that `x,y` are different from `hue` (for `hue` we do specify dimension name).

@shoyer Thoughts?

File: doc/plotting.rst (hunk 1 above)
Comment (dcherian) [reply doc/plotting.rst]: Rotation makes me think of rotating the lines by an angle. How about "Coordinate along y-axis"?

**Comment by fmaussion** [comment]
> Do you mean da.plot(y='time', x=None, hue='lat')

Yes, I meant that there should be one and only way to specify which axis should do what: my understanding of your current implementation is that there are two ways to reach the same goal: either by (i) specifying the name of the variable to the axis you want to plot it onto, or (ii) by specifying the name of the coordinate you want to use for the axis. Since (ii) is the default and only option for 1D and 2D plots until now, I just wondered if (i) is very necessary. (or maybe I missed something)

**Comment by shoyer** [comment]
My main thought is that this API would feel much more natural on a Dataset object, alongside a `.plot.scatter()` method.

That said, I suppose this could still be useful and I don't think it's harmful to expand the API here. It does feel a little strange that if you had a DataArray with non-dimension coordinates you could make a plot without including any of the DataArray values, e.g., `xr.DataArray(..., dims=['x'], coords={'x': ..., 'y': ('x', ...)}, name='f').plot.line(x='x', y='y')`.

**Comment by dcherian** [comment]
> either by (i) specifying the name of the variable to the axis you want to plot it onto, or (ii) by specifying the name of the coordinate you want to use for the axis. Since (ii) is the default and only option for 1D and 2D plots until now, I just wondered if (i) is very necessary.

> It does feel a little strange that if you had a DataArray with non-dimension coordinates you could make a plot without including any of the DataArray values

Good points. I'll change it so that only one of `x` and `y` can be specified.

> My main thought is that this API would feel much more natural on a Dataset object, alongside a .plot.scatter() method.
It's very useful in oceanography (and meteorology) where we want to make plots of physical quantities like temperature against depth; it is much more natural to see depth on the y-axis than the x-axis. This is the original usecase @rabernat mention in #575

File: xarray/plot/plot.py [hunk 7]
```diff
@@ -198,43 +199,72 @@ def line(darray, *args, **kwargs):
...
     hue = kwargs.pop('hue', None)
     x = kwargs.pop('x', None)
+    y = kwargs.pop('y', None)
     add_legend = kwargs.pop('add_legend', True)
 
     ax = get_axis(figsize, size, aspect, ax)
 
+    error_msg = 'must be either None or one of %r' % list(darray.dims)
+
+    if x not in [None, *darray.dims]:
```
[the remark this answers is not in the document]
dcherian: I don't know what's wrong here. It runs locally; the tests pass and flake8 goes through plot.py without errors. Is it a python2 vs python3 thing?

shoyer:
Yes, this is not valid syntax on Python 2.

fmaussion:
Thanks! In view of @shoyer ' [comment](https://github.com/pydata/xarray/pull/1926#issuecomment-367383795) we could revisit the API for datasets at a later stage.

xarray/plot/plot.py
```diff
@@ -198,43 +199,73 @@ def line(darray, *args, **kwargs):
     ax = kwargs.pop('ax', None)
     hue = kwargs.pop('hue', None)
     x = kwargs.pop('x', None)
+    y = kwargs.pop('y', None)
     add_legend = kwargs.pop('add_legend', True)
 
     ax = get_axis(figsize, size, aspect, ax)
 
+    error_msg = ('must be either None or one of ({0:s})'
+                 .format(', '.join(['\'' + dd + '\'' for dd in darray.dims])))
```
shoyer: just use `repr(dd)` here instead of `'\'' + dd + '\''`

xarray/plot/plot.py
```diff
@@ -198,43 +199,73 @@ def line(darray, *args, **kwargs):
...
+
+    if x is not None and x not in darray.dims:
+        raise ValueError('x ' + error_msg)
+
+    if y is not None and y not in darray.dims:
+        raise ValueError('y ' + error_msg)
+
+    if x is not None and y is not None:
+        raise ValueError('You cannot specify both x and y kwargs.')
```
shoyer: This should probably be qualified by "for line plots"

dcherian:
Done and rebased on master.

xarray/plot/plot.py
```diff
@@ -176,9 +176,10 @@ def line(darray, *args, **kwargs):
         Axis on which to plot this figure. By default, use the current axis.
         Mutually exclusive with ``size`` and ``figsize``.
     hue : string, optional
-        Coordinate for which you want multiple lines plotted (2D inputs only).
-    x : string, optional
-        Coordinate for x axis.
+        Coordinate for which you want multiple lines plotted
+        (2D DataArrays only).
+    x, y : string, optional
+        Coordinates for x, y axis. Only one of these may be specified.
```
shoyer: This should probably say something like "The other coordinate plots values from the DataArray on which this plot method is called."

xarray/tests/test_plot.py
```diff
@@ -91,14 +91,44 @@ def setUp(self):
...
+
+        xy = [[None, None],
+              [None, 'z'],
+              ['z', None]]
+
+        f, ax = plt.subplots(2, 4)
+
+        for aa, (x, y) in enumerate(xy):
+            da.plot(x=x, y=y, ax=ax.flat[aa])
+            ax.flat[aa].set_title('x=' + str(x) + ' | ' + 'y=' + str(y))
```
shoyer: Is there any point to this line?

xarray/tests/test_plot.py
```diff
@@ -91,14 +91,44 @@ def setUp(self):
...
 
+    def test_1d_x_y_kw(self):
+        z = np.arange(10)
+        da = DataArray(np.cos(z), dims=['z'], coords=[z], name='f')
+
+        xy = [[None, None],
+              [None, 'z'],
+              ['z', None]]
+
+        f, ax = plt.subplots(2, 4)
```
shoyer: You only have two cases now, so no need for 2x4 subplots.

xarray/tests/test_plot.py
dcherian: No. I copied the test over from my debugging plots. I'll remove it (I assume you mean the `set_title` bit).

shoyer:
@fmaussion any further concerns here? This looks good to me.

fmaussion:
Thanks @dcherian ! This is a nice feature

Issue #1923, opened 2018-02-19
Local test failure in test_backends


max-sixty:
I'm happy to debug this further but before I do, is this an issue people have seen before? I'm running tests on master and hit an issue very early on. 

FWIW I don't use netCDF, and don't think I've got that installed

#### Code Sample, a copy-pastable example if possible

```python
========================================================================== FAILURES ==========================================================================
_________________________________________________________ ScipyInMemoryDataTest.test_bytesio_pickle __________________________________________________________

self = <xarray.tests.test_backends.ScipyInMemoryDataTest testMethod=test_bytesio_pickle>

    @pytest.mark.skipif(PY2, reason='cannot pickle BytesIO on Python 2')
    def test_bytesio_pickle(self):
        data = Dataset({'foo': ('x', [1, 2, 3])})
        fobj = BytesIO(data.to_netcdf())
        with open_dataset(fobj, autoclose=self.autoclose) as ds:
>           unpickled = pickle.loads(pickle.dumps(ds))
E           TypeError: can't pickle _thread.lock objects

xarray/tests/test_backends.py:1384: TypeError
```
#### Problem description

[this should explain **why** the current behavior is a problem and why the expected output is a better solution.]

#### Expected Output

Skip or pass backends tests

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: d00721a3560f57a1b9226c5dbf5bf3af0356619d
python: 3.6.4.final.0
python-bits: 64
OS: Darwin
OS-release: 17.4.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.7.0-38-g1005a9e  # not sure why this is tagged so early. I'm running on latest master
pandas: 0.22.0
numpy: 1.14.0
scipy: 1.0.0
netCDF4: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: None
distributed: None
matplotlib: 2.1.2
cartopy: None
seaborn: 0.8.1
setuptools: 38.5.1
pip: 9.0.1
conda: None
pytest: 3.4.0
IPython: 6.2.1
sphinx: None
</details>

shoyer:
I think this test would pass if you had dask installed.

SciPy comes with read/write support for netCDF3, which is why this is being picked up.

smithsp:
[...]

PR #1927, opened 2018-02-20
Some backend tests require dask

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/1923 (remove if there is no corresponding issue, which should only be the case for minor changes)

LMK if these are the right ones - I basically added the decorator to anything that was failing. 
Though not sure we need to be 100% accurate here - worst case we could skip the file - either people are writing backend code and have dask installed, or they're not...

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -1384,6 +1384,7 @@ def test_replace_slices_with_arrays():
 
 
 @requires_scipy
+@requires_dask
 class ScipyInMemoryDataTest(CFEncodedDataTest, NetCDF3Only, TestCase):
     engine = 'scipy'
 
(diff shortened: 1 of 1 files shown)


xarray/tests/test_backends.py
```diff
@@ -1384,6 +1384,7 @@ def test_replace_slices_with_arrays():
 
 
 @requires_scipy
+@requires_dask
```
shoyer: This isn't quite right -- we still want to check the reading/writing a netCDF file with SciPy work without dask. I'm going to look alternate fixes for the `can't pickle _thread.lock objects` issue.

maxim-lian: OK! I'll leave this open for the moment

mathause:
As #1923 & without really looking - I think this can be closed given that we test all but dask now: https://github.com/pydata/xarray/blob/master/ci/requirements/py38-all-but-dask.yml#L1

max-sixty:
Yes, thanks!

PR #1929, opened 2018-02-20
Use requires_netcdftime decorators in test_coding_times.py

spencerkclark:
@jhamman I'm sorry I missed this in #1920.  The time decoding tests in the temporary Travis build with the new `netcdftime` library are all skipped because they are tagged with `@requires_netCDF4` decorators rather than `@requires_netcdftime` ones.  This PR fixes that.

In a local environment (after swapping these decorators) with the new `netcdftime`, I'm actually getting a failure, so there may be a bug we need to sort out upstream.

### DIFF
--- a/xarray/tests/test_coding_times.py
+++ b/xarray/tests/test_coding_times.py
@@ -8,8 +8,9 @@ import numpy as np
 import pandas as pd
 
 from xarray import Variable, coding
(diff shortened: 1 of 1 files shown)


jhamman:
Thanks for catching this @spencerkclark!

PR #1934, opened 2018-02-22
DOC: Add contributing section in README.rst and README.rst in doc/

raybellwaves:

These were inspired by a discussion on the mailing list:
https://groups.google.com/forum/#!topic/xarray/FyCFkuWbFSk

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -292,7 +292,7 @@ Building the documentation
 
 So how do you build the docs? 
 
-Firstly, make sure you have `sphinx` and `sphinx_rtd_theme` installed::
+Firstly, make sure you have ``sphinx`` and ``sphinx_rtd_theme`` installed::
 
     conda install -c anaconda sphinx sphinx_rtd_theme
 
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -290,7 +290,13 @@ First, you need to have a development environment to be able to build xarray
 Building the documentation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


doc/contributing.rst
```diff
@@ -290,7 +290,13 @@ First, you need to have a development environment to be able to build xarray
 Building the documentation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-So how do you build the docs? Navigate to your local
+So how do you build the docs? 
+
+Firstly, make sure you have ``sphinx`` and ``sphinx_rtd_theme`` installed::
+
+    conda install -c anaconda sphinx sphinx_rtd_theme
```
shoyer: This feels a little incongruent with the paragraph beginning with "First" under "Requirements" above

shoyer: Maybe: "In your development environment, install sphinx and sphinx_rtd_theme:"

fmaussion:
Note that in the upcoming version of xarray the contributing page will be more prominent and visible in the docs: http://xarray.pydata.org/en/latest/contributing.html 

The two links in the README files link to the github page, which is less nice than the rendered readthedocs HTML

doc/contributing.rst
raybellwaves: Agreed. I also dropped the question as i felt it was superfluous.

raybellwaves:
@fmaussion Agreed. Changed the URsL to link to the latest version of the online docs.

jhamman:
LGTM. Thanks @raybellwaves!

PR #1946, opened 2018-02-27
DOC: add main sections to toc

benbovy:
Not a big change, but adds a little more clarity IMO.

I'm open to any suggestion for better section names and/or organization. Also I let "What's new" at the top, but not sure if "Getting started" is the right section.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,4 +1,3 @@
-
 .. image:: _static/dataset-diagram-logo.png
    :width: 300 px
    :align: center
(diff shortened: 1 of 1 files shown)


doc/index.rst
```diff
@@ -32,14 +31,44 @@ describing scientific data in widespread use in the Earth sciences:
 Documentation
 -------------
 
+**Getting Started**
+
+* :doc:`whats-new`
```
shoyer: Let's move this to the start of "Help & reference" instead.

benbovy: Or move it to the very end? I kind of like a linear flow from new user to contributor (i.e., getting started -> user guide -> api -> developer zone...) and keep "what's new" out of that.

Also, a "what's new" section at the bottom of the toc may sound a little bit weird. Maybe rename it to "Release Notes"?

For example pandas has both "What's new" at the top and "Release Notes" at the bottom. I like it but I don't think we need both in xarray, though.

No strong opinion on this.

shoyer: Yeah, I don't know :). We can also just leave it here.

benbovy: Yeah that was probably overthinking it :). Moved to to the start of "Help & reference".

PR #1950, opened 2018-03-02
Fix doc for missing values.

### LINKED ISSUE #1944: building doc is failing for the release 0.10.1
I found the following page fails
http://xarray.pydata.org/en/stable/examples/weather-data.html

fujiisoup:
- [x] Closes #1944

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -113,6 +113,22 @@ Indexing
    Dataset.reset_index
    Dataset.reorder_levels
 
+Missing value handling
+----------------------
+
+.. autosummary::
+   :toctree: generated/
+
+   Dataset.isnull
+   Dataset.notnull
+   Dataset.count
+   Dataset.dropna
+   Dataset.fillna
+   Dataset.ffill
+   Dataset.bfill
+   Dataset.interpolate_na
+   Dataset.where
+
 Computation
 -----------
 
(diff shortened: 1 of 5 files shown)


doc/api.rst
```diff
@@ -113,6 +113,22 @@ Indexing
...
+   Dataset.isnull
+   Dataset.notnull
+   Dataset.count
+   Dataset.dropna
+   Dataset.fillna
+   Dataset.ffill
+   Dataset.bfill
+   Dataset.interpolate_na
+   Dataset.where
+
```
fujiisoup: I think it would be nice if we could see docstrings for these methods in rtd.

shoyer: Works for me, but these should be removed from `api-hidden.rst` now.

fujiisoup:
I'm wondering this doc is some commits ahead of the stable (e.g. it has docs about `rolling.construct`).
Should I modify stable branch?

doc/api.rst
fujiisoup: Done.

shoyer:
@fujiisoup What I do for updating the stable docs is to cherry-pick commits with doc fixes to the stable branch. As long as those apply cleanly (which is usually the case) we can update the stable docs.

shoyer:
I just cherry-picked the merge commit from this PR into `stable`. So RTD should automatically update when the build finishes in ~15 minutes.

fujiisoup:
Thanks, @shoyer 

But this part 
http://xarray.pydata.org/en/stable/computation.html#missing-values
is still failing (I think this was working on my environment).
Is there any missing dependencies on rtd?

PR #1965, opened 2018-03-05
avoid integer overflow when decoding large time numbers

j08lue:
The issue: `int32` time data in seconds or so leads to an overflow in time decoding. 

This is in way the back side of #1859: By ensuring that `_NS_PER_TIME_DELTA` is integer, we got rid of round-off errors that were due to casting to float but now we are getting `int` overflow in this line:

https://github.com/pydata/xarray/blob/0e73e240107caee3ffd1a1149f0150c390d43251/xarray/coding/times.py#L169-L170

e.g. '2001-01-01' in `seconds since 1970-01-01` means `np.array([978307200]) * int(1e9)` which gives `288686080` that gets decoded to '1970-01-01T00:00:00.288686080' -- note also the trailing digits. Something is very wrong here.

### DIFF
--- a/asv_bench/benchmarks/dataset_io.py
+++ b/asv_bench/benchmarks/dataset_io.py
@@ -5,7 +5,7 @@ import pandas as pd
 
 import xarray as xr
 
-from . import randn, requires_dask
+from . import randn, randint, requires_dask
 
 try:
     import dask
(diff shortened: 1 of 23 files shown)


j08lue:
Seems like this issue only exists on Windows. See https://ci.appveyor.com/project/shoyer/xray/build/1.0.3782/job/yjodrbenev1tw5y7#L4686

The Travis failure is due to a failing build.

j08lue:
I can see `netcdftime` is casting the `num_dates` to `float64`:

https://github.com/Unidata/netcdftime/blob/e745434547de728d53cde6316cae75b62db102e2/netcdftime/_netcdftime.pyx#L241

like I suggested for https://github.com/pydata/xarray/pull/1863#issuecomment-361089977. 

How about we do that, too?

shoyer:
OK, this seems reasonable to me. Can you add a brief note on the bug fix in whats-new?

jhamman:
@j08lue - mind freshening up this PR? It seems there are some conflicts that need to be resolved before merging.

j08lue:
> mind freshening up this PR

I deleted my fork in the meantime. I opened a new PR at #2096 and fixed the merge conflict there.

PR #1967, opened 2018-03-06
Fix RGB imshow with X or Y dim of size one

### LINKED ISSUE #1966: imshow should work with third dimension of len 1

```python
import xarray as xr
import numpy as np

da = xr.DataArray(np.arange(9).reshape((1, 3, 3)))

da.plot()  # works
da.plot.imshow()  # fails
```
Error log:

<details>

```
/home/mowglie/Documents/git/xarray/xarray/plot/utils.py:295: UserWarning: Several dimensions of this array could be colors.  Xarray will use the last possible dimension ('dim_2') to match matplotlib.pyplot.imshow.  You can pass names of x, y, and/or rgb dimensions to override this guess.
  'and/or rgb dimensions to override this guess.' % rgb)

<ipython-input-7-a0592d3e8758> in <module>()
----> 1 da.plot.imshow()

~/Documents/git/xarray/xarray/plot/plot.py in plotmethod(_PlotMethods_obj, x, y, figsize, size, aspect, ax, row, col, col_wrap, xincrease, yincrease, add_colorbar, add_labels, vmin, vmax, cmap, colors, center, robust, extend, levels, infer_intervals, subplot_kws, cbar_ax, cbar_kwargs, **kwargs)
    679         for arg in ['_PlotMethods_obj', 'newplotfunc', 'kwargs']:
    680             del allargs[arg]
--> 681         return newplotfunc(**allargs)
    682 
    683     # Add to class _PlotMethods

~/Documents/git/xarray/xarray/plot/plot.py in newplotfunc(darray, x, y, figsize, size, aspect, ax, row, col, col_wrap, xincrease, yincrease, add_colorbar, add_labels, vmin, vmax, cmap, center, robust, extend, levels, infer_intervals, colors, subplot_kws, cbar_ax, cbar_kwargs, **kwargs)
    553         rgb = kwargs.pop('rgb', None)
    554         xlab, ylab = _infer_xy_labels(
--> 555             darray=darray, x=x, y=y, imshow=imshow_rgb, rgb=rgb)
    556 
    557         if rgb is not None and plotfunc.__name__ != 'imshow':

~/Documents/git/xarray/xarray/plot/utils.py in _infer_xy_labels(darray, x, y, imshow, rgb)
    308     assert x is None or x != y
    309     if imshow and darray.ndim == 3:
--> 310         return _infer_xy_labels_3d(darray, x, y, rgb)
    311 
    312     if x is None and y is None:

~/Documents/git/xarray/xarray/plot/utils.py in _infer_xy_labels_3d(darray, x, y, rgb)
    297 
    298     # Finally, we pick out the red slice and delegate to the 2D version:
--> 299     return _infer_xy_labels(darray.isel(**{rgb: 0}).squeeze(), x, y)
    300 
    301 

~/Documents/git/xarray/xarray/plot/utils.py in _infer_xy_labels(darray, x, y, imshow, rgb)
    312     if x is None and y is None:
    313         if darray.ndim != 2:
--> 314             raise ValueError('DataArray must be 2d')
    315         y, x = darray.dims
    316     elif x is None:

ValueError: DataArray must be 2d
```

</details>

Zac-HD:
- [x] Closes #1966

Not much more to say, really.  Thanks to @fmaussion for pinging me - definitely faster to track down when you know the code!

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -56,6 +56,9 @@ Bug fixes
   By `Keisuke Fujii <https://github.com/fujiisoup>`_.
 - Fix kwarg `colors` clashing with auto-inferred `cmap` (:issue:`1461`)
   By `Deepak Cherian <https://github.com/dcherian>`_.
+- Fix :py:func:`~xarray.plot.imshow` error when passed an RGB array with
+  size one in a spatial dimension.
+  By `Zac Hatfield-Dodds <https://github.com/Zac-HD>`_.
 
 .. _whats-new.0.10.1:
 
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -296,7 +296,7 @@ def _infer_xy_labels_3d(darray, x, y, rgb):
     assert rgb is not None
 
     # Finally, we pick out the red slice and delegate to the 2D version:
-    return _infer_xy_labels(darray.isel(**{rgb: 0}).squeeze(), x, y)
+    return _infer_xy_labels(darray.isel(**{rgb: 0}), x, y)
 
 
 def _infer_xy_labels(darray, x, y, imshow=False, rgb=None):
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -1197,6 +1197,11 @@ class TestImshow(Common2dMixin, PlotTestCase):
         assert out.dtype == np.uint8
         assert (out[..., :3] == da.values).all()  # Compare without added alpha
 
+    def test_regression_rgb_imshow_dim_size_one(self):
+        # Regression: https://github.com/pydata/xarray/issues/1966
+        da = DataArray(easy_array((1, 3, 3), start=0.0, stop=1.0))
+        da.plot.imshow()
+
 
 class TestFacetGrid(PlotTestCase):
     def setUp(self):


fmaussion:
Wow that was fast, thanks! 

Would it be possible to add the test to the ``Common2dMixin`` class instead? (exmaple [here](https://github.com/Zac-HD/xarray/blob/9816bd149f2b33612e19ebdf2a05ab8d8fe2ac16/xarray/tests/test_plot.py#L660))

Zac-HD:
No problem - I had an hour free and no open pulls waiting on me, so the timing was good.

As a regression test it's specific to imshow, so I'm not sure what you'd want here (or whether it would work at all on the 2d mixin). More details please?

fmaussion:
> As a regression test it's specific to imshow, so I'm not sure what you'd want here (or whether it would work at all on the 2d mixin). More details please?

Yes sorry: I think it should work same for all four 2d plot methods (we should have had a test for this)

Zac-HD:
Ah, I see what you mean but don't think we need any change or additional test.

There's a catch though - `xarray.plot()` is special, because it (and only it) [squeezes the dimensions of the array]() before plotting it.  Therefore, the following calls all produce the same plot:

    xr.DataArray(np.arange(9).reshape((3,3))).plot()
    xr.DataArray(np.arange(9).reshape((1,3,3))).plot()
    xr.DataArray(np.arange(9).reshape((1,1,3,1,1,1,1,3,1,1))).plot()

My view is that the test you linked to is sufficient for the test you're asking for - `imshow` is a special case because it *can* accept 3D input for RGB plots.

TLDR - working as intended IMO, it's just that nobody reads the docs.  Changing the API would avoid this but at cost of convenience which is the whole point of `DataArray.plot()`.

shoyer:
@Zac-HD your link on squeezing goes back to this page :).

Possibly we should deprecate and/or remove the auto-squeezing behavior? That sort of magic can be pretty frustrating to debug.

Zac-HD:
Not any more :wink: - it links to the docs that "[`xarray.plot.plot()`] calls an xarray plotting function based on the dimensions of darray.squeeze()".

I'd actually like to keep the magic in `plot()` - the whole point of this is that we can do *something* with the data, not that we do anything in particular.  For interactive use, this is valuable enough - in my view - to justify keeping an inconsistent API; particularly when there is an obvious non-magic version to use in a script.  In any case, that's a topic for another pull.

shoyer:
OK, I'm going to merge this. Thanks @Zac-HD

PR #1968, opened 2018-03-06
einsum for xarray

### LINKED ISSUE #1951: einsum for xarray

I sometimes want to make more flexible dot product of two data arrays, 
where we sum up along a part of common dimensions.

```python
# Your code here
da_vals = np.arange(6 * 5 * 4).reshape((6, 5, 4))
da = DataArray(da_vals, dims=['x', 'y', 'z'])
dm_vals = np.arange(6 * 4).reshape((6, 4))
dm = DataArray(dm_vals, dims=['x', 'z'])

# I want something like this
da.dot(dm, 'z')  # -> dimensions of the output array: ['x', 'y']
```

It's an intermediate path of `np.matmul` and `np.tensordot`.

Is this feature sufficiently universal?

EDIT: 
I just noticed dask does not have `einsum` yet (dask/dask#732).
We maybe need to wait or decide to support only numpy arrays.

fujiisoup:
- [x] Closes #1951 
 - [x] Tests added

Currently, lazy-einsum for dask is not yet working.

@shoyer 
I think `apply_ufunc` supports lazy computation, but I did not yet figure out how to do this.
Can you give me a help?


xarray/core/computation.py
```diff
@@ -926,6 +926,86 @@ def earth_mover_distance(first_samples,
...
+    ('a', 'd')
+    """
+    dims = kwargs.pop('dims', None)
+    arrays = args
+    if dims is None and isinstance(args[-1], (list, tuple, basestring)):
+        dims = args[-1]
+        arrays = args[:-1]
+
+    if len(arrays) < 2:
+        raise TypeError('More than two arrays must be provided')
```
maxim-lian: More than one :)

xarray/core/computation.py
```diff
@@ -926,6 +926,86 @@ def earth_mover_distance(first_samples,
...
+                                output_core_dims[0] + dims + broadcast_dims]
+        all_dims += [d for d in arr.dims if d not in all_dims]
+
+    einsum_axes = 'abcdefghijklmnopqrstuvwxyz'
+    dim_map = {d: einsum_axes[i] for i, d in enumerate(all_dims)}
+
+    subscripts = ''
+    for ds in input_core_dims:
+        subscripts += '...' + ''.join([dim_map[d] for d in ds]) + ','
+    subscripts = subscripts[:-1]  # remove last comma
```
shoyer: It would probably be cleaner to build up subscripts as a list and use `','.join(subscripts_list)` once at the end.

xarray/core/computation.py
```diff
@@ -926,6 +926,86 @@ def earth_mover_distance(first_samples,
...
+    subscripts += '->...' + ''.join([dim_map[d] for d in output_core_dims[0]])
+
+    result = apply_ufunc(np.einsum, subscripts, *arrays,
+                         input_core_dims=[[]] + input_core_dims,
+                         output_core_dims=output_core_dims, dask='allowed')
```
shoyer: I think `dask='parallelized'` is what you want here -- that generate the wrapper to do this with dask. This will require also determining the result data type, probably with `dtypes.result_type` or even `np.result_type` (we don't need support for non-numeric types in einsum, so I'm pretty sure NumPy's casting rules would work fine).

`dask='allowed'` would be appropriate if `np.einsum` already supported dask arrays (but it does not).

It's possible that a dask specific `einsum` could be much more efficient than the auto-generated wrapper here, but certainly this is good enough for now.

xarray/core/computation.py
```diff
@@ -926,6 +926,86 @@ def earth_mover_distance(first_samples,
...
+    dims = kwargs.pop('dims', None)
+    arrays = args
+    if dims is None and isinstance(args[-1], (list, tuple, basestring)):
+        dims = args[-1]
+        arrays = args[:-1]
+
+    if len(arrays) < 2:
+        raise TypeError('More than two arrays must be provided')
+
+    if any(not hasattr(arr, 'dims') for arr in arrays):
```
shoyer: Dataset also defines `dims`. It's probably better to explicitly use an `isinstance()` check.

xarray/core/dataarray.py
```diff
@@ -1968,23 +1972,12 @@ def dot(self, other):
...
-        axes = (self.get_axis_num(dims), other.get_axis_num(dims))
-        new_data = duck_array_ops.tensordot(self.data, other.data, axes=axes)
-
-        new_coords = self.coords.merge(other.coords)
-        new_coords = new_coords.drop([d for d in dims if d in new_coords])
-        new_dims = ([d for d in self.dims if d not in dims] +
-                    [d for d in other.dims if d not in dims])
-
-        return type(self)(new_data, new_coords.variables, new_dims)
+        # backward compat: if there is no shared dimension, we rais an Errror
```
shoyer: It might be better to eliminate this special case. Then users can understand `DataArray.dot` as a simple short-cut for `xarray.dot()`.

xarray/core/computation.py
```diff
@@ -926,6 +926,86 @@ def earth_mover_distance(first_samples,
...
+    >>> dot(da_a, da_b, dims=['a']).dims
+    ('b', 'c')
+    >>> dot(da_a, da_b, da_c, dims=['b', 'c']).dims
+    ('a', 'd')
+    """
+    dims = kwargs.pop('dims', None)
+    arrays = args
+    if dims is None and isinstance(args[-1], (list, tuple, basestring)):
+        dims = args[-1]
+        arrays = args[:-1]
```
shoyer: I think it is better to require specifying `dims` with a keyword argument.

xarray/core/computation.py
```diff
@@ -926,6 +926,86 @@ def earth_mover_distance(first_samples,
         return apply_array_ufunc(func, *args, dask=dask)
 
 
+def dot(*args, **kwargs):
+    """ dot(*arrays, dims=None)
```
shoyer: `dot(*arrays, *, dims=None)` is the way to write this with Python 3's keyword only arguments.

xarray/core/computation.py
fujiisoup: Thanks. I noticed that my current implementation is not very efficient for dask.
Maybe smaller number of `input_core_dims` is better for dask?

I think I need some improvement.

xarray/core/computation.py
fujiisoup: Our previous `dot` does not require `dim`. This assumes to sum over along all the common dimensions.
I think `dim=None` is not surprising.

xarray/core/computation.py
shoyer: `dask='parallelized'` will only parallelize over broadcast dimensions, i.e., ones that don't appear in either `input_core_dims` or `output_core_dims`. So yes, it will probably be slow in many cases.

I'm still OK adding the non-optimal einsum for now and improving it later.

xarray/core/computation.py
shoyer: I agree, the default `dims=None` should be OK. I meant that `dims` should be a keyword only argument, not a required argument.

Here you are supporting `xr.dot(a, b, 'x')`, where `'x'` denotes a dimension. I would require writing `xr.dot(a, b, dim='x')` or omitting `dim` altogether.

xarray/core/computation.py
```diff
@@ -994,15 +996,30 @@ def dot(*args, **kwargs):
...
+    out_dtype = dtypes.result_type(*arrays)
+
+    # we use tensordot if available, because it is more efficient for dask
+    if len(broadcast_dims) == 0 and len(arrays) == 2:
+        axes = [[arr.get_axis_num(d) for d in arr.dims if d in dims]
+                for arr in arrays]
+        return apply_ufunc(duck_array_ops.tensordot, *arrays, dask='allowed',
+                           input_core_dims=input_core_dims,
+                           output_core_dims=output_core_dims,
+                           kwargs={'axes': axes})
```
fujiisoup: Thanks. I added a path for tensordot, which dask can compute more efficiently.

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+    """ dot(*arrays, *, dims=None)
+
+    einsum for xarray object, but providing simpler interface based on
+    the array dimensions.
+
+    Parameters
+    ----------
+    arrays: multiple DataArrays
+        arrays to compute.
+    dims: tuple of strings, optional
```
shoyer: str or tuple of strings

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
         return apply_array_ufunc(func, *args, dask=dask)
 
 
+def dot(*arrays, **kwargs):
+    """ dot(*arrays, *, dims=None)
+
+    einsum for xarray object, but providing simpler interface based on
+    the array dimensions.
```
shoyer: We should lead with a more general description. Maybe:

    Generalized dot product for xarray objects. Like np.einsum, but 
    provides a simpler interface based on array dimensions.

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+
+    Parameters
+    ----------
+    arrays: multiple DataArrays
```
shoyer: `*arrays: DataArray objects`

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+        arrays to compute.
```
shoyer: Arrays

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+    dims: tuple of strings, optional
+        Along which dimensions to be summed over.
```
shoyer: Which dimensions to sum over.

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+        If not speciified, then all the common dimensions are summed over.
+
+    Returns
+    -------
+    dot: same type to input.
```
shoyer: Probably should just "DataArray"?

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+
+    if any(not isinstance(arr, (DataArray, Variable)) for arr in arrays):
+        raise TypeError('Only xr.DataArray and xr.Variable are supported.')
+
+    if isinstance(dims, basestring):
+        dims = [dims]
+
+    common_dims = set(arrays[0].dims)
+    for arr in arrays[1:]:
+        common_dims = common_dims.intersection(set(arr.dims))
```
shoyer: This is a slightly different choice of default dimensions than `np.einsum`:
- `np.einsum` sums over any dimensions that are defined in two over more inputs.
- This sums only over dimensions that are defined on *all* inputs.

Should we switch this behavior to match `einsum`?

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+
+    Examples
+    --------
+
+    >>> da_a = xr.DataArray(np.arange(3 * 4).reshape(3, 4), dims=['a', 'b'])
+    >>> da_b = xr.DataArray(np.arange(3 * 4 * 5).reshape(3, 4, 5),
+                            dims=['a', 'b', 'c'])
+    >>> da_c = xr.DataArray(np.arange(5 * 6).reshape(5, 6), dims=['c', 'd'])
+
+    >>> dot(da_a, da_b, dims=['a', 'b']).dims
```
shoyer: These should use the full name `xr.dot`.

xarray/core/computation.py
```diff
@@ -926,6 +926,103 @@ def earth_mover_distance(first_samples,
...
+    >>> dot(da_a, da_b, da_c, dims=['b', 'c']).dims
+    ('a', 'd')
+    """
+    from .dataarray import DataArray
+    from .variable import Variable
+
+    dims = kwargs.pop('dims', None)
+
+    if len(arrays) < 2:
+        raise TypeError('More than one arrays must be provided')
```
shoyer: Do we need this special case? If not, let's remove this. For consistency, it is nice to use the same logic even for edge cases when possible. This makes it easier to think about the function.

In this case, I think a dot product of 1 array would consistently defined by summing over dimensions listed explicitly in `dims`.

xarray/core/computation.py
fujiisoup: Maybe we would keep this as `dot(*arrays, **kwargs)` as we did not yet drop python 2 support?

xarray/core/computation.py
```diff
@@ -974,27 +977,30 @@ def dot(*arrays, **kwargs):
...
+
+    einsum_axes = 'abcdefghijklmnopqrstuvwxyz'
+    dim_map = {d: einsum_axes[i] for i, d in enumerate(all_dims)}
 
     if dims is None:
-        dims = list(common_dims)
+        # find dimensions that exist in more than two arrays
+        whole_dims = []
+        for arr in arrays:
+            whole_dims += [d for d in arr.dims]
```
shoyer: This might be a nice use for `collections.Counter()`, e.g.,
```
dim_counts = Counter():
for arr in arrays:
    dim_counts.update(arr.dims)
```

xarray/core/computation.py
```diff
@@ -974,27 +977,30 @@ def dot(*arrays, **kwargs):
         dims = [dims]
 
     common_dims = set(arrays[0].dims)
+    all_dims = []
```
shoyer: would it work to make `all_dims` a set instead of a list? I think that would be slightly more efficient.

xarray/core/computation.py
```diff
@@ -927,45 +927,48 @@ def earth_mover_distance(first_samples,
...
 
     dims = kwargs.pop('dims', None)
+    if len(kwargs) > 0:
+        raise TypeError('Invalid keyward arguments {} are given'.format(
+            list(kwargs.keys())))
 
-    if len(arrays) < 2:
-        raise TypeError('More than one arrays must be provided')
+    if len(arrays) < 2 and dims is None:
+        raise TypeError('dim must be provided for one array computation.')
```
shoyer: If there's only one array, wouldn't `dims` just be any repeated dimensions on the single array?

xarray/core/computation.py
shoyer: I was confused. `def dot(*arrays, *, dims=None)` is not valid syntax in Python 3, either. (There can only be one single `*`)

xarray/core/computation.py
```diff
@@ -926,6 +927,108 @@ def earth_mover_distance(first_samples,
...
+
+    if any(not isinstance(arr, DataArray) for arr in arrays):
+        raise TypeError('Only xr.DataArray and xr.Variable are supported.'
+                        'Given {}.'.format([type(arr) for arr in arrays]))
+
+    if len(arrays) == 0:
+        raise TypeError('At least one array should be given.')
+
+    if isinstance(dims, basestring):
+        dims = (dims, )
```
maxim-lian: FWIW you don't need the parentheses

xarray/core/computation.py
```diff
@@ -926,6 +927,108 @@ def earth_mover_distance(first_samples,
...
+    elif isinstance(dims, list):
+        dims = tuple(dims)
```
maxim-lian: FWIW `dims=tuple(dims)` doesn't create any copies if `dims` is already a tuple, so you could skip the `if isinstance` check

fujiisoup:
Thanks, @maxim-lian
added `xr.Variable` support for `xr.dot`.

xarray/core/computation.py
fujiisoup: I personally like parentheses, as I think it is more descriptive.

maxim-lian:
This is awesome. Beautiful code, immediately impactful, and the API is so simple - a testament to the benefits of named dims

Thank you @fujiisoup !

maxim-lian:
Do you know why the tests are failing? Do you want me to have a look?

The arrays _look_ the same: https://travis-ci.org/pydata/xarray/jobs/350640898#L5182. Would `assert_close` help?

fujiisoup:
I just noticed the test failings.
This was a bug caused by the undefined order of `set`.
Fixed. Thanks :)

fujiisoup:
I'm going to merge this tomorrow if there are no further comments.

PR #1969, opened 2018-03-06
Switched broken pypip.in badges to shields.io

raybellwaves:

This seems to be happening to a lot of projects as highlighted on the recent list of people who have cited this issue https://github.com/badges/shields/issues/716.

Here is a project with a recent commit to resolve this which explains the issue better that I can https://github.com/landscapeio/pylint-celery/pull/7

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,7 +10,7 @@ xarray: N-D labeled arrays and datasets
 .. image:: https://readthedocs.org/projects/xray/badge/?version=latest
    :target: http://xarray.pydata.org/
 .. image:: https://img.shields.io/pypi/v/xarray.svg
-   :target: https://pypi.python.org/pypi/xarray/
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -10,7 +10,7 @@ xarray: N-D labeled arrays and datasets
 .. image:: https://readthedocs.org/projects/xray/badge/?version=latest
    :target: http://xarray.pydata.org/
 .. image:: https://img.shields.io/pypi/v/xarray.svg
-   :target: https://pypi.python.org/pypi/xarray/
```
shoyer: This is a link to pypi (the Python package index), not pypip.in.

shoyer:
Thanks for looking into this, but I don't think this is actually relevant for us -- we already do not use pypip.in badges.

PR #1972, opened 2018-03-07
Starter property-based test suite

### LINKED ISSUE #1846: Add a suite of property-based tests with Hypothesis
[Hypothesis](https://hypothesis.readthedocs.io/en/master/) is a library for writing property-based tests in Python: you describe input data and make assertions that should be true for all examples, then Hypothesis tries to find a counterexample.  This came up in #1840, because `data == decode(encode(data))` is a classic property.

We could add a (initially small) suite of property-based tests, to complement the traditional example-based tests Xarray is already using.  Keeping them in independent files will ensure that they run in CI but the dependency on Hypothesis remains optional for local development.

I have moved jobs and don't have time to do this myself, but I'd be very happy to help anyone who does 😄

Zac-HD:
- [x] Closes #1846
 - [x] Tests added - **you bet**
 - [x] Tests passed - well, the code under test hasn't changed...

This is a small property-based test suite, to give two examples of the kinds of tests that we could write for Xarray using Hypothesis.

1. For any array, encoding and decoding it with a CF coder outputs an identical array.  As you would hope, these tests pass.
2. For any 2D array, you can call the 2D plotting methods without raising an exception.  Alas, this is *not* the case, and Hypothesis will show you the failing inputs (and matplotlib-related tracebacks) to prove it.
   (Contributing a very small feature to matplotlib was shockingly painful, so I'm not planning to take a similar suite upstream myself unless something changes)

Things that I would like to know:

- Have I build-wrangled something reasonable here?
- Will anyone else contribute property-based tests?  I'm happy to help people debug or work out how to test something, but I simply don't have the time to write another test suite for free.
- Is this something you *want?*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 *.py[cod]
 __pycache__
 
+# example caches from Hypothesis
+.hypothesis/
+
 # temp files from docs build
 doc/auto_gallery
 doc/example.nc
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,8 @@ matrix:
     env: CONDA_ENV=py36-zarr-dev
   - python: 3.5
     env: CONDA_ENV=docs
+  - python: 3.6
+    env: CONDA_ENV=py36-hypothesis
   allow_failures:
   - python: 3.6
     env:
@@ -104,6 +106,8 @@ script:
   - if [[ "$CONDA_ENV" == "docs" ]]; then
       conda install -c conda-forge sphinx sphinx_rtd_theme sphinx-gallery numpydoc;
       sphinx-build -n -j auto -b html -d _build/doctrees doc _build/html;
+    elif [[ "$CONDA_ENV" == "py36-hypothesis" ]]; then
+      pytest properties ;
     else
       py.test xarray --cov=xarray --cov-config ci/.coveragerc --cov-report term-missing --verbose $EXTRA_FLAGS;
     fi
--- a/ci/requirements-py36-hypothesis.yml
+++ b/ci/requirements-py36-hypothesis.yml
@@ -1,0 +1,27 @@
+name: test_env
+channels:
+  - conda-forge
+dependencies:
+  - python=3.6
+  - dask
+  - distributed
+  - h5py
+  - h5netcdf
+  - matplotlib
+  - netcdf4
+  - pytest
+  - flake8
+  - numpy
+  - pandas
+  - scipy
+  - seaborn
+  - toolz
+  - rasterio
+  - bottleneck
+  - zarr
+  - pip:
+    - coveralls
+    - pytest-cov
+    - pydap
+    - lxml
+    - hypothesis
--- a/properties/README.md
+++ b/properties/README.md
@@ -1,0 +1,22 @@
+# Property-based tests using Hypothesis
+
+This directory contains property-based tests using a library
+called [Hypothesis](https://github.com/HypothesisWorks/hypothesis-python).
+
+The property tests for Xarray are a work in progress - more are always welcome.
+They are stored in a separate directory because they tend to run more examples
+and thus take longer, and so that local development can run a test suite
+without needing to `pip install hypothesis`.
+
+## Hang on, "property-based" tests?
+
+Instead of making assertions about operations on a particular piece of
+data, you use Hypothesis to describe a *kind* of data, then make assertions
+that should hold for *any* example of this kind.
+
+For example: "given a 2d ndarray of dtype uint8 `arr`,
+`xr.DataArray(arr).plot.imshow()` never raises an exception".
+
+Hypothesis will then try many random examples, and report a minimised
+failing input for each error it finds.
+[See the docs for more info.](https://hypothesis.readthedocs.io/en/master/)
--- a/properties/test_encode_decode.py
+++ b/properties/test_encode_decode.py
@@ -1,0 +1,46 @@
+"""
+Property-based tests for encoding/decoding methods.
+
+These ones pass, just as you'd hope!
+
+"""
+from __future__ import absolute_import, division, print_function
+
+from hypothesis import given, settings
+import hypothesis.strategies as st
+import hypothesis.extra.numpy as npst
+
+import xarray as xr
+
+# Run for a while - arrays are a bigger search space than usual
+settings.deadline = None
+
+
+an_array = npst.arrays(
+    dtype=st.one_of(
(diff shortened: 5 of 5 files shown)


Zac-HD:
Ping @fmaussion / @shoyer - would love your opinions on this, including high-value targets to test.

(btw Appveyor had an internal error; build is otherwise green)

shoyer:
This looks like a great start to me -- thank you!

It's impressive that it's possible to break every plotting type with matplotlib :).

Zac-HD:
> This looks like a great start to me -- thank you!

You're welcome!  Same questions as above though, plus "is there anything else you need in this initial PR?".  If not, can we merge it 

> It's impressive that it's possible to break every plotting type with matplotlib :).

As much as I love matplotlib, it's a steaming pile of hacks and I want to avoid it more than I want it cleaned up 😥  (entirely because the process is dysfunctional, not the code)

shoyer:
One thing that comes to mind is organization... would it make sense to put this alongside the current xarray tests, e.g., have `xarray/tests/unit` and `xarray/tests/property`?

I guess one downside of this would be that it could change how we need to invoke py.test by default, if we don't want to trigger all the property based tests.

tacaswell:
What do the failing data sets look like?  Does it get easier or harder to find failures if you go up to 10x10?  What sort of exceptions are you getting?

You can shove a fair amount of configuration in to `pytest.ini` to make these opt-in.

Zac-HD:
@shoyer - that depends mostly on whether you want to run these tests as part of a standard run.  A test-time dependency on Hypothesis is very cheap compared to the other dev dependencies, so I'd think more about the impact on eg CI resources than contributors.

Upside is more power and coverage of edge-cases with odd data; downside is that they take a lot longer by virtue of trying hundreds of examples, and in this case also having to generate arrays takes a while (~log(elements) average via sparse filling).

-------

@tacaswell - I would be delighted to write a test suite like this for matplotlib!  The only reason I *haven't* is because I thought it would be rude to report so many bugs that I don't have time to help fix.  If we can get a group project going though I'd be very enthusiastic 😄 

For this suite I was passing in 2D arrays of unsigned ints, signed ints, or floats (in all sizes), with edge sizes between 2 and 10.

The failing outputs were reported as 2x2 arrays (~30 times), or 3x2 (once - there's a cap on number of shrinks that I think I hit there).

Values tended to be either uniform small integers - 0 or 1, dtype int8 or uint8; or for some an array of floats with one corner zero and the others *very* large.

I didn't keep the exact tracebacks, but I remember seeing many come from overflow in tick spacing calculations.  Again, happy to write a test suite and make more detailed reports upstream if people want to fix this - in which case let's open an issue there!

maxim-lian:
If we don't want to trigger by default, we can do something like [this](https://docs.pytest.org/en/latest/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs) and require passing this to run them:

```
pytest --property-tests
```

jklymak:
As pointed out on the matplotlib gitter: 

If you run 

```python
import numpy as np
import xarray as xr

for i in range(200):
    xr.DataArray(np.array([[0, 0], [0, 0]], dtype=np.uint8)).plot.pcolormesh()
```
at step 165 you will get:
```
File "/Users/jklymak/matplotlib/lib/matplotlib/figure.py", line 236, in update
    raise ValueError('left cannot be >= right')
ValueError: left cannot be >= right
```
Why?  Because you have made a plot that if it displays looks like:

fig1

Are you sure your test isn't doing something similar?   At some point there just isn't room for more colorbars!  Adding a `plt.clf()` can cure the problem.  

Its also is possible you are hitting floating point overflows with your test.  At some point Matplotlib needs to be able to manipulate the data that comes in, and if you operate near the maximum number your data type can handle, you'll have problems.  Just like you would if you just did 

```python
a = 2*xr.DataArray(np.array([[0, 0], [0, 1e308]]))
```
you will get: 
```
/Users/jklymak/anaconda3/envs/matplotlibdev/lib/python3.6/site-packages/xarray/core/variable.py:1165: RuntimeWarning: overflow encountered in multiply
```
So maybe your hypothesis tester could be constrained to stay away from floating point overflows?  

Matplotlib indeed has flaws and quirks, but if you are finding bugs it would be good to isolate them.

Zac-HD:
...that also explains why I was having trouble reproducing the error, whoops.  I'll see how it goes with those problems excluded later tonight!

Zac-HD:
@jklymak - I'm getting `RuntimeError: Invalid DISPLAY variable` in the Qt backend now that I've added `plt.clf()`.  It works on my machine now (:tada:, thanks!) - any suggestions for Travis?

(I'm also getting Zarr errors, but I assume those will go away soon as I didn't cause them)

tacaswell:
Set the backend to Agg on travis as you don't have a xserever running. You probably want to manually force a draw as well.

Zac-HD:
@shoyer & @fmaussion - I've just given up on the plotting tests as being more effort than they're worth.  Are there any:

- blockers to merging this as-is?
- other APIs you think it would be reasonably easy to write property tests for?  [Here's a nice list of properties](https://fsharpforfunandprofit.com/posts/property-based-testing-2/) to test :smile:

shoyer:
> blockers to merging this as-is?

This looks pretty good to me in its current state. I would say we should merge it now and iterate in future PRs.

> other APIs you think it would be reasonably easy to write property tests for? Here's a nice list of properties to test 😄

Almost anywhere where we currently make heavy use of `parametrize` would be a good candidate. Some other possibilities:

- Consistency with pandas for groupby/rolling aggregations.
- Roundtrip writing/reading data to netCDF. There are a couple of known exceptions (e.g., dtypes not supported by netCDF and MultiIndex) but otherwise every xarray object should be serializable to netCDF and back without data loss.
- Roundtrip to/from pandas Series/DataFrame with `to_series()`/`to_dataframe()`/`to_xarray()`.
- Indexing consistency tests for backends: all indexing operations should be supported consistently on data accessed from any backend.
- NumPy vs Dask: any operation on dask arrays should be consistent (after calling `compute()`) with the operation on numpy arrays.
- Indexing followed by `xarray.concat`: should get back the same result.
- Binary arithmetic on xarray objects with Python operators  (`+`, `-`, etc) and NumPy ufuncs (`np.add`, `np.subtract`, etc).

Zac-HD:
> I would say we should merge it now and iterate in future PRs.

Merge away then!

fmaussion:
Thanks @Zac-HD !

PR #1983, opened 2018-03-13
Parallel open_mfdataset

### LINKED ISSUE #1981: use dask to open datasets in parallel

```python
xr.open_mfdataset('path/to/many/files*.nc', method='parallel')
```
#### Problem description

We have many issues describing the less than stelar performance of open_mfdataset (e.g. #511, #893, #1385, #1788, #1823). The problem can be broken into three pieces: 1) open each file, 2) decode/preprocess each datasets, and 3) merge/combine/concat the collection of datasets. We can perform (1) and (2) in parallel (performance improvements to (3) would be a separate task). Lately, I'm finding that for large numbers of files, it can take many seconds to many minutes just to open all the files in a multi-file dataset of mine.

I'm proposing that we use something like `dask.bag` to parallelize steps (1) and (2). I've played around with this a bit and it "works" almost right out of the box, provided you are using the  "autoclose=True" option. A concrete example:

We could change the line:
```Python
datasets = [open_dataset(p, **open_kwargs) for p in paths]
```
to
```Python
import dask.bag as db
paths_bag = db.from_sequence(paths)
datasets = paths_bag.map(open_dataset, **open_kwargs).compute()
```

I'm curious what others think of this idea and what the potential downfalls may be.

jhamman:
- [x] Closes #1981
 - [x] Tests added 
 - [x] Tests passed 

I'm sharing this in the hopes of getting comments from @mrocklin and @pydata/xarray.

 What this does:

- implements a `dask.bag` map/apply on the xarray `open_dataset` and `preprocess` steps in `open_mfdataset`
- adds a new `parallel` option to `open_mfdataset`
- provides about a 40% speedup in opening a multifile dataset when using the distributed scheduler (I tested on 1000 netcdf files that took about 9 seconds to open/concatenate in the default configuration)

What it does not do (yet):

- check that `autoclose=True` when multiple processes are being use (multiprocessing/distributed scheduler)
- provide any speedup with the multiprocessing backend (I do not understand why this is)

------

### Benchmark Example

```Python
In [1]: import xarray as xr
   ...: import dask
   ...: import dask.threaded
   ...: import dask.multiprocessing
   ...: from dask.distributed import Client
   ...:

In [2]: c = Client()
   ...: c
   ...:
Out[2]: <Client: scheduler='tcp://127.0.0.1:59576' processes=4 cores=4>

In [4]: %%time
   ...: with dask.set_options(get=dask.multiprocessing.get):
   ...:     ds = xr.open_mfdataset('../test_files/test_netcdf_*nc', autoclose=True, parallel=True)
   ...:
CPU times: user 4.76 s, sys: 201 ms, total: 4.96 s
Wall time: 7.74 s

In [5]: %%time
   ...: with dask.set_options(get=c.get):
   ...:     ds = xr.open_mfdataset('../test_files/test_netcdf_*nc', autoclose=True, parallel=True)
   ...:
   ...:
CPU times: user 1.88 s, sys: 60.6 ms, total: 1.94 s
Wall time: 4.41 s

In [6]: %%time
   ...: with dask.set_options(get=dask.threaded.get):
   ...:     ds = xr.open_mfdataset('../test_files/test_netcdf_*nc')
   ...:
CPU times: user 7.77 s, sys: 247 ms, total: 8.02 s
Wall time: 8.17 s

In [7]: %%time
   ...: with dask.set_options(get=dask.threaded.get):
   ...:     ds = xr.open_mfdataset('../test_files/test_netcdf_*nc', autoclose=True)
   ...:
   ...:
CPU times: user 7.89 s, sys: 202 ms, total: 8.09 s
Wall time: 8.21 s

In [8]: ds
Out[8]:
<xarray.Dataset>
Dimensions:  (lat: 45, lon: 90, time: 1000)
Coordinates:
  * lon      (lon) float64 0.0 4.045 8.09 12.13 16.18 20.22 24.27 28.31 ...
  * lat      (lat) float64 -90.0 -85.91 -81.82 -77.73 -73.64 -69.55 -65.45 ...
  * time     (time) datetime64[ns] 1970-01-01 1970-01-02 1970-01-11 ...
Data variables:
    foo      (time, lon, lat) float64 dask.array<shape=(1000, 90, 45), chunksize=(1, 90, 45)>
    bar      (time, lon, lat) float64 dask.array<shape=(1000, 90, 45), chunksize=(1, 90, 45)>
    baz      (time, lon, lat) float32 dask.array<shape=(1000, 90, 45), chunksize=(1, 90, 45)>
Attributes:
    history:  created for xarray benchmarking

```


xarray/tests/test_backends.py
```diff
@@ -1658,88 +1658,62 @@ class H5NetCDFDataTestAutocloseTrue(H5NetCDFDataTest):
...
-    @pytest.mark.slow
-    def test_4_open_large_num_files_h5netcdf(self):
-        self.validate_open_mfdataset_large_num_files(engine=['h5netcdf'])
+def skip_if_not_engine(engine):
+    if engine == 'netcdf4':
+        pytest.importorskip('netCDF4')
+    elif engine == 'pynio':
+        pytest.importorskip('Nio')
+    else:
+        pytest.importorskip(engine)
```
jhamman: @maxim-lian - do you know a better way to skip parameterized tests based on some condition?

xarray/tests/test_backends.py
```diff
@@ -1658,88 +1658,62 @@ class H5NetCDFDataTestAutocloseTrue(H5NetCDFDataTest):
...
-    @requires_dask
-    @requires_netCDF4
-    def test_1_autoclose_netcdf4(self):
-        self.validate_open_mfdataset_autoclose(engine=['netcdf4'])
 
-    @requires_dask
-    @requires_scipy
-    def test_2_autoclose_scipy(self):
-        self.validate_open_mfdataset_autoclose(engine=['scipy'])
+@pytest.fixture(params=[1, 10, 1000])
```
jhamman: it would also be nice to mark the `1000` entry here as slow.

xarray/backends/api.py
```diff
@@ -562,12 +565,25 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
-                             **kwargs) for p in paths]
-    file_objs = [ds._file_obj for ds in datasets]
 
-    if preprocess is not None:
-        datasets = [preprocess(ds) for ds in datasets]
+    open_kwargs = dict(engine=engine, chunks=chunks or {}, lock=lock,
+                       autoclose=autoclose, **kwargs)
+
+    if parallel:
+        import dask.bag as db
```
shoyer: This is a minor point, but I find `dask.delayed()` slightly more general and easier to understand than `dask.bag`.

xarray/backends/api.py
```diff
@@ -562,12 +565,25 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
+        paths_bag = db.from_sequence(paths)
+        datasets = paths_bag.map(open_dataset,
+                                            **open_kwargs).compute()
```
shoyer: What exactly does `compute()` do when called on a dask bag of xarray datasets of dask arrays? I'm assuming that only the bags get computed, but not the array values? This would be good to clarify in some comments on the code.

xarray/backends/api.py
```diff
@@ -562,12 +565,25 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
-    datasets = [open_dataset(p, engine=engine, chunks=chunks or {}, lock=lock,
-                             **kwargs) for p in paths]
-    file_objs = [ds._file_obj for ds in datasets]
 
-    if preprocess is not None:
-        datasets = [preprocess(ds) for ds in datasets]
+    open_kwargs = dict(engine=engine, chunks=chunks or {}, lock=lock,
+                       autoclose=autoclose, **kwargs)
+
+    if parallel:
```
shoyer: Consider making the default `parallel=None` and using auto-detection logic for setting `parallel=True` if using dask-distributed.

xarray/backends/api.py
jhamman: I admit I don't fully understand the tradeoffs between these two approaches.

xarray/backends/api.py
jhamman: That is my understanding. In other words, `delayed(open_dataset)(...).compute()` is going to return whatever `open_dataset` would have. I have not tested this thoroughly and I would be interested to hear from @mrocklin on how dask is going to treat this situation.

xarray/backends/api.py
shoyer: They are basically equivalent. The only meaningful difference I think of aside from the API is that dask.delayed has fewer required dependencies (a subset of those required for dask.array):
https://github.com/dask/dask/blob/47d025476312c6a8155192196b8da3b5af0c13e4/setup.py#L10-L17

jhamman:
@shoyer - I updated this to use dask.delayed. I actually like it more because I only have to call compute once. Thanks for the suggestion.

xarray/backends/api.py
```diff
@@ -562,12 +566,26 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
+        import dask
+        datasets = [delayed(open_dataset)(p, **open_kwargs) for p in paths]
+        # important: get the file_objs before calling preprocess
+        file_objs = [ds._file_obj for ds in datasets]
+        if preprocess is not None:
+            datasets = [delayed(preprocess)(ds) for p in datasets]
+
+        # calling compute here will return compute the datasets/file_objs lists
+        # the underlying datasets will still be stored as dask arrays
+        dask.compute([datasets, file_objs])
```
shoyer: Don't you need to assign these computed results, e.g., `datasets, file_objs = dask.compute([datasets, file_objs])`?

jhamman:
If anyone understands Windows file handling with Python, I'm all ears as to why this is failing on AppVeyor. I'm tempted to just skip this test there but thought I should ask for help first...

jhamman:
> I'm tempted to just skip this test there but thought I should ask for help first...

I've skipped the offending test on appveyor for now. Objectors speak up please. I don't have a windows machine to test on and iterating via appveyor is not something a sane person does 😉.

jhamman:
> Have you tested this with both a local system and an HPC cluster?

I have. See below for a simple example using this feature on [Cheyenne](https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne).

```python
In [1]: import xarray as xr
   ...:
   ...: import glob
   ...:

In [2]: pattern = '/glade/u/home/jhamman/workdir/LOCA_daily/met_data/CESM1-BGC/16th/rcp45/r1i1p1/*/*nc'

In [3]: len(glob.glob(pattern))
Out[3]: 285

In [4]: %time ds = xr.open_mfdataset(pattern)
CPU times: user 15.5 s, sys: 2.62 s, total: 18.1 s
Wall time: 42.4 s

In [5]: ds.close()

In [6]: %time ds = xr.open_mfdataset(pattern, parallel=True)
CPU times: user 18.4 s, sys: 5.28 s, total: 23.6 s
Wall time: 30.7 s

In [7]: ds.close()

In [8]: from dask.distributed import Client

In [9]: client = Client()
clien
In [10]: client
Out[10]: <Client: scheduler='tcp://127.0.0.1:39853' processes=72 cores=72>

In [11]: %time ds = xr.open_mfdataset(pattern, parallel=True, autoclose=True)
CPU times: user 10.8 s, sys: 808 ms, total: 11.6 s
Wall time: 12.4 s
```

xarray/backends/api.py
```diff
@@ -534,7 +535,10 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
             those corresponding to other dimensions.
           * list of str: The listed coordinate variables will be concatenated,
             in addition the 'minimal' coordinates.
-
+    parallel : bool, optional
+        If True, the open and preprocess steps of this function will be performed
+        in parallel using ``dask.delayed``. Default is False unless the dask's
+        distributed scheduler is being used in which case the default is True.
```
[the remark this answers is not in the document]
rabernat: I'm curious about the logic of defaulting to parallel when using distributed.

xarray/tests/test_backends.py
```diff
@@ -1658,88 +1658,75 @@ class H5NetCDFDataTestAutocloseTrue(H5NetCDFDataTest):
...
+                       else 'netcdf4')
+        # split into multiple sets of temp files
+        for ii in original.x.values:
+            subds = original.isel(x=slice(ii, ii + 1))
+            subds.to_netcdf(tmpfiles[ii], engine=writeengine)
+
+        # check that calculation on opened datasets works properly
+        actual = open_mfdataset(tmpfiles, engine=readengine,
+                                autoclose=autoclose,
+                                chunks=chunks)
```
rabernat: Don't you want to have a test that explicitly calls `open_mfdataset(parallel=parallel)`?

jhamman: nice catch :)

jhamman:
> I'm curious about the logic of defaulting to parallel when using distributed.

I'm not tied to the behavior. It was [suggested](https://github.com/pydata/xarray/pull/1983#discussion_r173990300) by @shoyer a while back. Perhaps we try this and evaluate how it works in the wild?

xarray/tests/test_backends.py
jhamman: nice catch :)

rabernat:
Can we imagine cases where it might actually *degrade* performance?

jhamman:
I image there will be a small performance cost when the number of files is small. That cost is probably lost in the noise in most i/o operations.

jhamman:
All the tests are passing here? Any final objectors?

doc/whats-new.rst
```diff
@@ -37,6 +37,15 @@ Documentation
 Enhancements
 ~~~~~~~~~~~~
 
+- Added the ``parallel`` option to :py:func:`open_mfdataset`. This option uses
+  ``dask.delayed`` to parallelize the open and preprocessing steps withing
+  ``open_mfdataset``. This is expected to provide performance improvements when
+  opening many files, particularly when used in conjunction with
+  ``dask.distributed`` (:issue:`1981`).
+  By `Joe Hamman <https://github.com/jhamman>`_.
+ - Some speed improvement to construct :py:class:`~xarray.DataArrayRolling`
```
shoyer: This item ended up duplicated when you merged in master.

xarray/backends/api.py
```diff
@@ -562,13 +566,32 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
-    file_objs = [ds._file_obj for ds in datasets]
 
+    open_kwargs = dict(engine=engine, chunks=chunks or {}, lock=lock,
+                       autoclose=autoclose, **kwargs)
+
+    if parallel or (parallel is None and get_scheduler() == 'distributed'):
+        # wrap the open_dataset, getattr, and preprocess with delayed
+        import dask
+        parallel = True
+        _open = dask.delayed(open_dataset)
```
shoyer: nit: per PEP8, we usually post-fix with an underscore for variables named to avoid conflicts with builtins, e.g., `open_` rather than `_open`.

xarray/backends/api.py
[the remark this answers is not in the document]
shoyer: Maybe a better choice would be to default to true when using distributed or multi-processing? That would probably be a simpler choice.

Actually, this is probably fine even with using multi-threading -- it just won't give you any noticeable speedup. So I guess we could default to `parallel=True`.

xarray/backends/api.py
jhamman: Thanks, wasn't aware of this detail.

rabernat:
I recently tried this branch with my data server and got an error.

I opened a dataset this way
```python
# works fine with parallel=False
ds = xr.open_mfdataset(os.path.join(ddir, '*V1_1.204*.nc'), decode_cf=False, parallel=True)
```

and got the following error.

```
distributed.utils - ERROR - NetCDF: HDF error
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/distributed/utils.py", line 237, in f
    result[0] = yield make_coro()
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/distributed/client.py", line 1356, in _gather
    traceback)
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/six.py", line 692, in reraise
    raise value.with_traceback(tb)
  File "/home/rpa/.conda/envs/dask_distributed/lib/python3.5/site-packages/dask/compatibility.py", line 48, in apply
    return func(*args, **kwargs)
  File "/home/rpa/xarray/xarray/backends/api.py", line 318, in open_dataset
    return maybe_decode_store(store, lock)
  File "/home/rpa/xarray/xarray/backends/api.py", line 238, in maybe_decode_store
    drop_variables=drop_variables)
  File "/home/rpa/xarray/xarray/conventions.py", line 594, in decode_cf
    vars, attrs = obj.load()
  File "/home/rpa/xarray/xarray/backends/common.py", line 217, in load
    for k, v in self.get_variables().items())
  File "/home/rpa/xarray/xarray/backends/netCDF4_.py", line 319, in get_variables
    iteritems(self.ds.variables))
  File "/home/rpa/xarray/xarray/core/utils.py", line 308, in FrozenOrderedDict
    return Frozen(OrderedDict(*args, **kwargs))
  File "/home/rpa/xarray/xarray/backends/netCDF4_.py", line 318, in <genexpr>
    for k, v in
  File "/home/rpa/xarray/xarray/backends/netCDF4_.py", line 311, in open_store_variable
    encoding['original_shape'] = var.shape
  File "netCDF4/_netCDF4.pyx", line 3381, in netCDF4._netCDF4.Variable.shape.__get__ (netCDF4/_netCDF4.c:34388)
  File "netCDF4/_netCDF4.pyx", line 2759, in netCDF4._netCDF4.Dimension.__len__ (netCDF4/_netCDF4.c:27006)
RuntimeError: NetCDF: HDF error
```

Without the distributed scheduler (but with `parallel=True`), I get no error, but the command never returns, and eventually I have to restart the kernel.

Any idea what could be going on? (Sorry for the non-reproducible bug report...I figured some trials "in the field" might be useful.)

jhamman:
@rabernat - my last commit(s) seem to have broken the CI so I'll need to revisit this.

jhamman:
@rabernat - I just pushed a few more commits here. Can I ask two questions:

When using the distributed scheduler, what configuration are you using? Can you try:
  - `autoclose=True` (in open_mfdataset)
 - `processes=True` (in client)

xarray/tests/test_backends.py
```diff
@@ -1675,88 +1676,65 @@ class H5NetCDFDataTestAutocloseTrue(H5NetCDFDataTest):
...
-    def validate_open_mfdataset_large_num_files(self, engine):
-        self.validate_open_mfdataset_autoclose(engine, nfiles=2000)
+@pytest.fixture(params=[
+    pytest.mark.skipif(not has_scipy, 'scipy', reason='requires scipy'),
+    pytest.mark.skipif(not has_netCDF4, 'netcdf4', reason='requires netcdf4'),
+    pytest.mark.skipif(not has_h5netcdf, 'h5netcdf',
+                       reason='requires h5netcdf'),
+    pytest.mark.skipif(not has_pynio, 'pynio', reason='requires pynio')])
+def readengine(request):
+    return request.param
```
jhamman: Can anyone guess as to why the pynio tests are not being properly skipped using this fixture configuration? 

@mrocklin - I think you suggested this approach, which I like, but something weird is happening.

jhamman: I should note, this is working on python 2.7 and 3.4. Just not for 3.5 or 3.6.

jhamman:
@rabernat - I got the tests passing here again. If you can make the time to try your example/test again, it would be great to figure out what wasn't working before.

NicWayand:
Thanks @jhamman for working on this! I did a test on my real world data (1202 ~3mb files) on my local computer and am not getting results I expected: 
1) No speed up with parallel=True
2) _Slow down_ when using distributed (processes=16 cores=16). 

I am I missing something?

```python
nc_files = glob.glob(E.obs['NSIDC_0081']['sipn_nc']+'/*.nc')
print(len(nc_files))
1202

# Parallel False
%time ds = xr.open_mfdataset(nc_files, concat_dim='time', parallel=False, autoclose=True)
CPU times: user 57.8 s, sys: 3.2 s, total: 1min 1s
Wall time: 1min

# Parallel True with default scheduler 
%time ds = xr.open_mfdataset(nc_files, concat_dim='time', parallel=True, autoclose=True)
CPU times: user 1min 16s, sys: 9.82 s, total: 1min 26s
Wall time: 1min 16s

# Parallel True with distributed
client = Client()
print(client)
<Client: scheduler='tcp://127.0.0.1:43291' processes=16 cores=16>
%time ds = xr.open_mfdataset(nc_files, concat_dim='time', parallel=True, autoclose=True)
CPU times: user 2min 17s, sys: 12.3 s, total: 2min 29s
Wall time: 3min 48s

```

On feature/parallel_open_netcdf commit 280a46f13426a462fb3e983cfd5ac7a0565d1826

jhamman:
@NicWayand - Thanks for giving this a go. Some thoughts on your problem...

I'm have been using this feature for the past few days and have been seeing a speedup on datasets with many files along the lines of what I showed above. I am applying my tests on perhaps the perfect test architecture (parallel shared fs, fast interconnect, etc.). I think there are many reasons/cases where this won't work as well.

shoyer:
It sounds like the right resolution for now would be to leave the default as `parallel=False` and leave this as an optional feature.

jhamman:
I think that makes sense for now. We need to experiment with this a bit more but I don't see a problem merging the basic workflow we have now (with a minor change to the default behavior).

jhamman:
With my last commits here, this feature is completely optional and defaults to the current behavior. I cleaned up the tests a bit further and am now ready to merge this. Baring any objections, I'll merge this  on Friday.

PR #1994, opened 2018-03-15
Make constructing slices lazily.

### LINKED ISSUE #1993: DataArray.rolling().mean() is way slower than it should be

From @RayPalmerTech in https://github.com/kwgoodman/bottleneck/issues/186:
```python
import numpy as np
import pandas as pd
import time
import bottleneck as bn
import xarray

N = 30000200 # Number of datapoints
Fs = 30000 # sample rate
T=1/Fs # sample period
duration = N/Fs # duration in s
t = np.arange(0,duration,T) # time vector
DATA = np.random.randn(N,)+5*np.sin(2*np.pi*0.01*t) # Example noisy sine data and window size
w = 330000

def using_bottleneck_mean(data,width):
  return bn.move_mean(a=data,window=width,min_count = 1)

def using_pandas_rolling_mean(data,width):
  return np.asarray(pd.DataFrame(data).rolling(window=width,center=True,min_periods=1).mean()).ravel()

def using_xarray_mean(data,width):
  return xarray.DataArray(data,dims='x').rolling(x=width,min_periods=1, center=True).mean()

start=time.time()
A = using_bottleneck_mean(DATA,w)
print('Bottleneck: ', time.time()-start, 's')
start=time.time()
B = using_pandas_rolling_mean(DATA,w)
print('Pandas: ',time.time()-start,'s')
start=time.time()
C = using_xarray_mean(DATA,w)
print('Xarray: ',time.time()-start,'s')
```

This results in:
```
Bottleneck:  0.0867006778717041 s
Pandas:  0.563546895980835 s
Xarray:  25.133142709732056 s
```

Somehow xarray is way slower than pandas and bottleneck, even though it's using bottleneck under the hood!

#### Problem description

Profiling shows that the majority of time is spent in `xarray.core.rolling.DataArrayRolling._setup_windows`. Monkey-patching that method with a dummy rectifies the issue:
```
xarray.core.rolling.DataArrayRolling._setup_windows = lambda *args: None
```

Now we obtain:
```
Bottleneck:  0.06775331497192383 s
Pandas:  0.48262882232666016 s
Xarray:  0.1723031997680664 s
```

The solution is to make setting up windows done lazily (in `__iter__`), instead of doing it in the constructor.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.96+
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.2
pandas: 0.22.0
numpy: 1.14.2
scipy: 0.19.1
netCDF4: None
h5netcdf: None
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: None
distributed: None
matplotlib: 2.1.2
cartopy: None
seaborn: 0.7.1
setuptools: 36.2.7
pip: 9.0.1
conda: None
pytest: None
IPython: 5.5.0
sphinx: None
</details>

fujiisoup:
- [x] Closes #1993 
 - [x] Tests passed
 - [x] Fully documented, including `whats-new.rst` for all changes.

Quick fix of #1993.

With this fix, the script shown in #1993 runs
Bottleneck:  0.08317923545837402 s
Pandas:  1.3338768482208252 s
Xarray:  1.1349339485168457 s

### DIFF
--- a/asv_bench/benchmarks/rolling.py
+++ b/asv_bench/benchmarks/rolling.py
@@ -8,26 +8,39 @@ import xarray as xr
 from . import parameterized, randn, requires_dask
 
 nx = 3000
+long_nx = 30000200
 ny = 2000
 nt = 1000
 window = 20
+
+randn_xy = randn((nx, ny), frac_nan=0.1)
+randn_xt = randn((nx, nt))
+randn_t = randn((nt, ))
+randn_long = randn((long_nx, ), frac_nan=0.1)
 
 
 class Rolling(object):
     def setup(self, *args, **kwargs):
         self.ds = xr.Dataset(
-            {'var1': (('x', 'y'), randn((nx, ny), frac_nan=0.1)),
-             'var2': (('x', 't'), randn((nx, nt))),
-             'var3': (('t', ), randn(nt))},
+            {'var1': (('x', 'y'), randn_xy),
+             'var2': (('x', 't'), randn_xt),
+             'var3': (('t', ), randn_t)},
             coords={'x': np.arange(nx),
                     'y': np.linspace(0, 1, ny),
                     't': pd.date_range('1970-01-01', periods=nt, freq='D'),
                     'x_coords': ('x', np.linspace(1.1, 2.1, nx))})
+        self.da_long = xr.DataArray(randn_long, dims='x',
+                                    coords={'x': np.arange(long_nx) * 0.1})
 
     @parameterized(['func', 'center'],
                    (['mean', 'count'], [True, False]))
     def time_rolling(self, func, center):
         getattr(self.ds.rolling(x=window, center=center), func)()
+
+    @parameterized(['func', 'center'],
+                   (['mean', 'count'], [True, False]))
+    def time_rolling_long(self, func, center):
+        getattr(self.da_long.rolling(x=window, center=center), func)()
 
     @parameterized(['window_', 'min_periods'],
                    ([20, 40], [5, None]))
@@ -47,3 +60,4 @@ class RollingDask(Rolling):
         requires_dask()
         super(RollingDask, self).setup(**kwargs)
         self.ds = self.ds.chunk({'x': 100, 'y': 50, 't': 50})
+        self.da_long = self.da_long.chunk({'x': 100, 'y': 50, 't': 50})
--- a/asv_bench/benchmarks/rolling.py
+++ b/asv_bench/benchmarks/rolling.py
@@ -8,7 +8,7 @@ import xarray as xr
 from . import parameterized, randn, requires_dask
 
 nx = 3000
-long_nx = 30000200
+long_nx = 30000000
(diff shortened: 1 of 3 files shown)


xarray/core/rolling.py
```diff
@@ -151,34 +151,22 @@ def __init__(self, obj, min_periods=None, center=False, **windows):
         """
         super(DataArrayRolling, self).__init__(obj, min_periods=min_periods,
                                                center=center, **windows)
-        self.window_indices = None
-        self.window_labels = None
 
-        self._setup_windows()
+        self.window_labels = self.obj[self.dim]
+        self._stops = np.arange(1, len(self.window_labels) + 1)
```
shoyer: can we do this only when iteration is requested? that would be a bit more efficient for the common case where iteration is not done.

My guess is that iteration is slow enough that the overhead of recreating these objects will not be noticeable.

shoyer:
The benchmark does not need to be complete, but it would be good to ensure we don't have a regression here.

xarray/core/rolling.py
fujiisoup: Done.

asv_bench/benchmarks/rolling.py
```diff
@@ -37,10 +37,14 @@ def setup(self, *args, **kwargs):
...
+    @parameterized(['func', 'pandas'],
                    (['mean', 'count'], [True, False]))
-    def time_rolling_long(self, func, center):
-        getattr(self.da_long.rolling(x=window, center=center), func)()
+    def time_rolling_long(self, func, pandas):
+        if pandas:
+            se = self.da_long.to_series()
+            getattr(se.rolling(window=window), func)()
+        else:
+            getattr(self.da_long.rolling(x=window), func)()
```
fujiisoup: Also added a benchmark using pandas for long 1d-array, in order to make it easier to find the cause of the regression.
But is it too verbose as pandas might make the similar benchmark tests.

shoyer: I'm OK with this.

xarray/core/rolling.py
```diff
@@ -151,34 +151,21 @@ def __init__(self, obj, min_periods=None, center=False, **windows):
...
-        self.window_indices = None
-        self.window_labels = None
 
-        self._setup_windows()
+        self.window_labels = self.obj[self.dim]
 
     def __iter__(self):
-        for (label, indices) in zip(self.window_labels, self.window_indices):
-            window = self.obj.isel(**{self.dim: indices})
+        _stops = np.arange(1, len(self.window_labels) + 1)
```
shoyer: nit: no need to preface with an underscore for variables that are already limited in scope to this method.

PR #2003, opened 2018-03-21
Extraneous full stop in 88 char warning

edoddridge:
The warning introduced in #1835 has an extraneous full stop. This PR removes it.


shoyer:
thanks!

Issue #2010, opened 2018-03-23
rolling.mean vs rolling.construct.mean


mathause:
#### Code Sample, a copy-pastable example if possible

```python
import xarray as xr
import numpy as np
x = np.arange(4.)
ds = xr.DataArray(x, dims=('dim', ))
ds.rolling(dim=3, center=True).mean()
# RESULT: array([nan,  1.,  2., nan])
ds.rolling(dim=3, center=True).construct('window').mean('window')
# RESULT: array([0.5, 1. , 2. , 2.5])
```
#### Problem description

`ds.rolling(...).mean()` and `ds.rolling(...).construct().mean()` yields different results. Because `mean` does `skipna=True` per default.

#### Expected Output

I would expect both ways to yield the same result.

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.114-42-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8

xarray: 0.10.2+dev6.g9261601
pandas: 0.22.0
numpy: 1.14.2
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: None
h5py: None
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: 1.0.0
dask: 0.17.1
distributed: 1.21.3
matplotlib: 2.2.2
cartopy: None
seaborn: None
setuptools: 39.0.1
pip: 9.0.2
conda: None
pytest: 3.4.2
IPython: 6.2.1
sphinx: None
</details>

fujiisoup:
Hi, @mathause. Thanks for reporting.

I agree that this behavior is a little surprising, but this is something expected.
The difference is not `skipna=True` (both use this), but actually in `min_periods`.
If you specify `min_periods=1`, then the results become identical.
```python
In [7]: ds.rolling(dim=3, center=True, min_periods=1).mean()
Out[7]: 
<xarray.DataArray (dim: 4)>
array([0.5, 1. , 2. , 2.5])
Dimensions without coordinates: dim
```

I am not sure how we can make both the behaviors the same with keeping the generality of `construct` method.

fujiisoup:
I think we need to document more clearly that `min_periods` option is just neglected if using `construct` method.

mathause:
[...]

PR #2011, opened 2018-03-23
rolling: periodic

### LINKED ISSUE #2007: rolling with periodic boundary conditions

```python
import numpy as np
import xarray as xr

x = np.arange(1, 366)
y = np.random.randn(365)
ds = xr.DataArray(y, dims=dict(dayofyear=x))

ds.rolling(center=True, dayofyear=31).mean()
```
#### Problem description

`rolling` cannot directly handle periodic boundary conditions (lon, dayofyear, ...), but could be very helpful to e.g. calculate climate indices. Also I cannot really think of an easy way to append the first elements to the end of the dataset and then calculate rolling.

Is there a way to do this? Should xarray support this feature?

This might also belong to SO...

mathause:
- [x] Closes #2007

---

Ok, this was easier to do than initially thought, we can use `np.pad(a, pads, mode='wrap')` in `nputils.rolling_window`. However, I'm not sure if that is enough already<sup>*</sup>.

I added an initial test, but could use some pointers where else you want this to be tested.

Questions:
* is `fill_value='periodic'` a good api?
* should the `fill_value` keyvalue be ported to `rolling`?
* should this also be mentioned in the docs for `rolling`  (I only learned about `rolling.construct` yesterday)

---
<sup>*</sup>`rolling` is present in`core/dataset.py`, `core/dataarray.py`, `core/variable.py`, `core/rolling.py`, `core/dask_array_ops.py`, `core/nputils.py`, `core/ops.py`, `core/common.py`, `core/missing.py`, and `core/duck_array_ops.py` that can be a bit daunting...

### DIFF
--- a/xarray/core/nputils.py
+++ b/xarray/core/nputils.py
@@ -146,7 +146,12 @@ def rolling_window(a, axis, window, center, fill_value):
         pads[axis] = (start, end)
     else:
         pads[axis] = (window - 1, 0)
-    a = np.pad(a, pads, mode='constant', constant_values=fill_value)
+
+    if fill_value == 'periodic':
+        a = np.pad(a, pads, mode='wrap')
+    else:
+        a = np.pad(a, pads, mode='constant', constant_values=fill_value)
+
     return _rolling_window(a, window, axis)
 
 
--- a/xarray/tests/test_nputils.py
+++ b/xarray/tests/test_nputils.py
@@ -53,3 +53,20 @@ def test_rolling():
     actual = rolling_window(x, axis=-1, window=3, center=False, fill_value=0.0)
     expected = np.stack([expected, expected * 1.1], axis=0)
     assert_array_equal(actual, expected)
+
+    x = np.array([1, 2, 3, 4], dtype=float)
+    actual = rolling_window(x, axis=-1, window=3, center=True,
+                            fill_value='periodic')
+    expected = np.array([[4, 1, 2],
+                         [1, 2, 3],
+                         [2, 3, 4],
+                         [3, 4, 1]], dtype=float)
+    assert_array_equal(actual, expected)
+
+    actual = rolling_window(x, axis=-1, window=3, center=False,
+                            fill_value='periodic')
+    expected = np.array([[3, 4, 1],
+                         [4, 1, 2],
+                         [1, 2, 3],
+                         [2, 3, 4]], dtype=float)
+    assert_array_equal(actual, expected)


mathause:
We would probably have to test how `min_periods` and `periodic` play together..., see #2010

fujiisoup:
Thanks for sending a PR.

> is fill_value='periodic' a good api?

Maybe better API would be `boundaries='peiodic'` (see [comments](https://github.com/pydata/xarray/issues/2007#issuecomment-375828864)), but changing the public API is not an easy task.
Let's keep it now as is and decide it later.

> should the fill_value keyvalue be ported to rolling?

I don't think so. I think we can keep rolling as similar to that of pandas,
and this option should be specific to `construct` method.

> should this also be mentioned in the docs for rolling (I only learned about rolling.construct yesterday)

I think at least we need to add this to the docstring of `construct`.

fujiisoup:
Some comments.

+ We need to also support a dask array.
Can you update `dask_array_op.rolling_window`? Maybe [this page](http://dask.pydata.org/en/latest/array-ghost.html#boundaries) would help for the implementation.

We use this here 
https://github.com/pydata/xarray/blob/7c2c43ce6b1792e1193635ab9b64fd248266f632/xarray/core/dask_array_ops.py#L78-L79

I suspect this would be already working, but it would be nice if we can add some test cases.

+ Maybe we can also add `boundary='reflect'` option to make this as similar to `dask.ghost`.

mathause:
Thanks for the discussion. I could not get it to work with dask - the problem is the 'padding before ghosting'. 

https://github.com/pydata/xarray/blob/7c2c43ce6b1792e1193635ab9b64fd248266f632/xarray/core/dask_array_ops.py#L60-L62

I see no way how to get the `boundary='periodic'` working afterwards...

fujiisoup:
Ah, you are right.
The only way to realize rolling.periodic with dask would be to manually attach a chunk in the other side to the boundaries, construct rolling window array and then remove unnecessary boundaries by slicing.

rabernat:
Rather than letting this PR lie dormant, maybe we should just move forward without dask support and raise an appropriate error if the user tries it with dask arrays.

shoyer:
It might make sense to start with the higher level `pad()` with periodic boundary conditions first (https://github.com/pydata/xarray/issues/2605). Then rolling with periodic boundary conditions is a simple matter of padding, then rolling.

shoyer:
> * is `fill_value='periodic'` a good api?

`fill_value='periodic'` almost suggests padding with the string 'periodic'. Maybe `padding='periodic'` vs `padding=None` (default)?

PR #2012, opened 2018-03-23
Add weighted mean docs.

dcherian:
I like @fujiisoup's weighted mean demo in this stack overflow example:
https://stackoverflow.com/questions/48510784/xarray-rolling-mean-with-weights

I thought it'd be a useful addition to the docs on rolling.

### DIFF
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -185,7 +185,7 @@ windowed rolling, convolution, short-time FFT etc.
 
 Because the ``DataArray`` given by ``r.construct('window_dim')`` is a view
 of the original array, it is memory efficient.
-One can also use ``construct`` to compute a weighted mean:
+You can also use ``construct`` to compute a weighted mean:
 
 .. ipython:: python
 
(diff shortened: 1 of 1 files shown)


doc/computation.rst
```diff
@@ -185,6 +185,12 @@ windowed rolling, convolution, short-time FFT, etc.
 
 Because the ``DataArray`` given by ``r.construct('window_dim')`` is a view
 of the original array, it is memory efficient.
+One can also use ``construct`` to compute a weighted mean:
```
maxim-lian: `weighted rolling`
And nit: we've used `You` or `We` throughout - let's not add another voice

doc/computation.rst
```diff
@@ -185,7 +185,7 @@ windowed rolling, convolution, short-time FFT etc.
 
 Because the ``DataArray`` given by ``r.construct('window_dim')`` is a view
 of the original array, it is memory efficient.
-One can also use ``construct`` to compute a weighted mean:
+You can also use ``construct`` to compute a weighted mean:
```
maxim-lian: I do think this should explicitly say `weighted rolling mean`, may have got lost from https://github.com/pydata/xarray/pull/2012#discussion_r176850037

doc/computation.rst
dcherian: Ah, I missed that

maxim-lian:
Thanks a lot @dcherian ! 
And @fujiisoup for the feature and SO post!

Issue #2002, opened 2018-03-21
Unexpected decoded time in xarray >= 0.10.1


JanisGailis:
#### Problem description

Given the original time dimension:
```python
ds = xr.open_mfdataset("C:\\Users\\janis\\.cate\\data_stores\\local\\local.SST_should_fail\\*.nc", decode_cf=False)
```
```
<xarray.DataArray 'time' (time: 32)>
array([788961600, 789048000, 789134400, 789220800, 789307200, 789393600,
       789480000, 789566400, 789652800, 789739200, 789825600, 789912000,
       789998400, 790084800, 790171200, 790257600, 790344000, 790430400,
       790516800, 790603200, 790689600, 790776000, 790862400, 790948800,
       791035200, 791121600, 791208000, 791294400, 791380800, 791467200,
       791553600, 791640000], dtype=int64)
Coordinates:
  * time     (time) int64 788961600 789048000 789134400 789220800 789307200 ...
Attributes:
    standard_name:  time
    axis:           T
    comment:        
    bounds:         time_bnds
    long_name:      reference time of sst file
    _ChunkSizes:    1
    units:          seconds since 1981-01-01
    calendar:       gregorian
```
Produces this decoded time dimension with `xarray >= 0.10.1`:
```python
ds = xr.open_mfdataset("C:\\Users\\janis\\.cate\\data_stores\\local\\local.SST_should_fail\\*.nc", decode_cf=True)
```
```
<xarray.DataArray 'time' (time: 32)>
array(['1981-01-01T00:00:00.627867648', '1980-12-31T23:59:58.770774016',
       '1981-01-01T00:00:01.208647680', '1980-12-31T23:59:59.351554048',
       '1981-01-01T00:00:01.789427712', '1980-12-31T23:59:59.932334080',
       '1980-12-31T23:59:58.075240448', '1981-01-01T00:00:00.513114112',
       '1980-12-31T23:59:58.656020480', '1981-01-01T00:00:01.093894144',
       '1980-12-31T23:59:59.236800512', '1981-01-01T00:00:01.674674176',
       '1980-12-31T23:59:59.817580544', '1980-12-31T23:59:57.960486912',
       '1981-01-01T00:00:00.398360576', '1980-12-31T23:59:58.541266944',
       '1981-01-01T00:00:00.979140608', '1980-12-31T23:59:59.122046976',
       '1981-01-01T00:00:01.559920640', '1980-12-31T23:59:59.702827008',
[...]

PR #2015, opened 2018-03-24
Fix an overflow bug in decode_cf_datetime

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/2002
 - [x] Tests added 
 - [x] Tests passed

Not sure yet if this is the best way to do this.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -47,6 +47,9 @@ Bug fixes
 - Fixed labeled indexing with slice bounds given by xarray objects with
   datetime64 or timedelta64 dtypes (:issue:`1240`).
   By `Stephan Hoyer <https://github.com/shoyer>`_.
+- Fixed a bug in decode_cf_datetime where ``int32`` arrays weren't parsed
+  correctly (:issue:`2002`).
+  By `Fabien Maussion <https://github.com/fmaussion>`_.
 
 .. _whats-new.0.10.2:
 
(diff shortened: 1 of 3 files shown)


fmaussion:
The problem comes from a weird numpy casting behavior. Consider:

```python
In [1]: import numpy as np

In [2]: (np.int32(2) * np.int64(2)).dtype
Out[2]: dtype('int64')

In [3]: (np.array(np.int32(2)) * np.int64(2)).dtype
Out[3]: dtype('int64')

In [4]: (np.array([np.int32(2)]) * np.int64(2)).dtype
Out[4]: dtype('int32')
```

In the offending line:

```python
flat_num_dates *  _NS_PER_TIME_DELTA[delta]
```
``flat_num_dates`` is an array of dtype ``int32`` and ``_NS_PER_TIME_DELTA[delta]`` is a scalar of dtype ``int64``. Is there a way to make sure this line casts to a secure dtype better than my proposed solution to cast everything to a ``float64`` first?

fujiisoup:
Thanks @fmaussion .

> Is there a way to make sure this line casts to a secure dtype better than my proposed solution to cast everything to a float64 first?

How about `np.array([np.int32(2)]) * np.array([np.int64(2)])`?

I am not sure whether numpy explains this casting behavior, but I guess this is intentional.
This is consistent with
```python
In [11]: (np.array([1.0], dtype=np.float32) * 1.0).dtype
Out[11]: dtype('float32')
```
where we expect the array dtype takes precedence.

fujiisoup:
I am curious why this test is failing in appveyor.
The casting rule of numpy should not depend on the operating system.

fmaussion:
> I am curious why this test is failing in appveyor.
> The casting rule of numpy should not depend on the operating system.

I'm curious too, but I don't have a windows machine to test it. The tests are green with the float64 casting though

PR #2016, opened 2018-03-26
Allow _FillValue and missing_value to differ (Fixes #1749)

### LINKED ISSUE #1749: _FillValue and missing_value not allowed for the same variable
#### Problem description

Currently opening the file which has a variable with both _FillValue and missing_value defined, results in a ValueError "'Discovered conflicting _FillValue and missing_value...", which can be overridden by setting `decode_cf=False` 

Code in `xarray/conventions.py` have comments that `missing_value is deprecated`, however this seems not to be true anymore, see [ticket #67](https://cf-trac.llnl.gov/trac/ticket/67) in http://cfconventions.org/cf-conventions/cf-conventions.html

Also, see description of meanings of _FillValue and missing_value here: 
https://github.com/Unidata/netcdf4-python/issues/94
"Note that missing_value and _FillValue are very different. _FillValue is used by the C library to fill in values in variables that have not yet been written to. missing_value is meant to represent values set by the user to represent data points that do not exist."

#### Expected Output
Accept both _FillValue and missing_value and if both present, mask output only for _FillValue

0.9.6

dopplershift:
The CF standard permits both values, and them to have different values, so we should not be treating this as an error--just mask out all of them.

 - [x] Closes #1749 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/coding/variables.py
+++ b/xarray/coding/variables.py
@@ -7,7 +7,7 @@ from functools import partial
 import numpy as np
 import pandas as pd
 
-from ..core import dtypes, duck_array_ops, indexing, utils
+from ..core import dtypes, duck_array_ops, indexing
 from ..core.pycompat import dask_array_type
 from ..core.variable import Variable
 
@@ -152,26 +152,12 @@ class CFMaskCoder(VariableCoder):
     def decode(self, variable, name=None):
         dims, data, attrs, encoding = unpack_for_decoding(variable)
 
-        if 'missing_value' in attrs:
-            # missing_value is deprecated, but we still want to support it as
-            # an alias for _FillValue.
-            if ('_FillValue' in attrs and
-                not utils.equivalent(attrs['_FillValue'],
-                                     attrs['missing_value'])):
-                raise ValueError("Conflicting _FillValue and missing_value "
-                                 "attrs on a variable {!r}: {} vs. {}\n\n"
-                                 "Consider opening the offending dataset "
-                                 "using decode_cf=False, correcting the "
-                                 "attrs and decoding explicitly using "
-                                 "xarray.decode_cf()."
-                                 .format(name, attrs['_FillValue'],
-                                         attrs['missing_value']))
-            attrs['_FillValue'] = attrs.pop('missing_value')
-
-        if '_FillValue' in attrs:
-            raw_fill_value = pop_to(attrs, encoding, '_FillValue', name=name)
-            encoded_fill_values = [
-                fv for fv in np.ravel(raw_fill_value) if not pd.isnull(fv)]
+        raw_fill_values = [pop_to(attrs, encoding, attr, name=name)
+                           for attr in ('missing_value', '_FillValue')]
+        if raw_fill_values:
+            encoded_fill_values = {fv for option in raw_fill_values
+                                   for fv in np.ravel(option)
+                                   if not pd.isnull(fv)}
 
             if len(encoded_fill_values) > 1:
                 warnings.warn("variable {!r} has multiple fill values {}, "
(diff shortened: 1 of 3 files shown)


dopplershift:
Test failures look unrelated--they also happen to me locally. Installing SciPy 1.0.0 locally fixes them, so I'm guessing they're caused by SciPy 1.0.1.

shoyer:
please merge in master, which fixes the failing tests.

dopplershift:
Rebased on current master.

dopplershift:
Fixed flake8 (unnecessary import)

dopplershift:
Reworked due to corner case with PyNio on Python 2.7. Final solution looks simpler to me.

xarray/coding/variables.py
```diff
@@ -152,26 +152,12 @@ def encode(self, variable, name=None):
...
-                fv for fv in np.ravel(raw_fill_value) if not pd.isnull(fv)]
+        raw_fill_values = [pop_to(attrs, encoding, attr, name=name)
+                           for attr in ('missing_value', '_FillValue')]
+        if raw_fill_values:
+            encoded_fill_values = [fv for option in raw_fill_values
+                                   for fv in np.ravel(option)
+                                   if not pd.isnull(fv)]
 
             if len(encoded_fill_values) > 1:
                 warnings.warn("variable {!r} has multiple fill values {}, "
```
shoyer: Is this warning this useful? Or should it be removed?

xarray/tests/test_conventions.py
```diff
@@ -156,12 +156,16 @@ def test(self):
 
 
 def test_decode_cf_with_conflicting_fill_missing_value():
-    var = Variable(['t'], np.arange(10),
+    expected = Variable(['t'], [np.nan, np.nan, 2], {'units': 'foobar'})
+    expected[[0, 1]] = np.nan
```
shoyer: I think this line is unnecessary, since you already set these values to NaN when you construct the variable.

dopplershift: Oops, leftover from trying to get the thing to work. 😁

xarray/coding/variables.py
dopplershift: Personally, I don't need it. I'm not sure if others would like to be made of this case, though. I think my thoughts on leaving it were that it might be good to know since it would affect round-tripping.

One thing I notice nowis that this will now fire if `missing_value` and `_FillValue` are both present, regardless of their values. That's probably a bad thing. So I guess I'm 👍 on removing it.

shoyer: Yeah, I think it probably makes sense to keep if it's an indication that round-tripping will likely fail.

dopplershift: Ok, leaving the warning. I updated the comprehensions to make a set, so that should suppress false positives with the warning.

shoyer:
thanks!

himikof:
It seems that this PR exposed a new encoding key 'missing_value', and this makes the encoding fail to round-trip, as it is not explicitly allowed in `xarray.backends.netCDF4_._extract_nc4_variable_encoding` function (similar to #1869).
This leads to the following error when saving with an explicitly preserved encoding: 
ValueError: unexpected encoding parameters for 'netCDF4' backend:  ['missing_value']

dopplershift:
@shoyer Happy to put in a PR to address...provided you can tell me what that should actually look like. 😁

shoyer:
Well, we're not going to perfectly round-trip data that uses both `_FillValue` *and* `missing_value`. That isn't possible with xarray's data model.

But we should certainly not raise errors when round-tripping. I see a few reasonable options:
- Don't store `missing_value` in encoding. This would entail removing the use of `pop_to` with `missing_value` in `decode()`.
- Add some logic to `encode()` to check which of `_FillValue` and/or `missing_value` are defined. Use the attribute is defined and write it to attribute. If both are defined, ignore/drop `missing_value`.

As a regression test, we should verify that `code.encode(coder.decode(var)) == var` (including explicitly checking `encoding`).

PR #2020, opened 2018-03-28
One pytest config

maxim-lian:
Currently our pytest settings in `setup.cfg` are ignored, because they're overridden by the presence of a `pytest.ini`

This restores pytest looking at `setup.cfg`, gives us a single place to put configs, and retains the 'no hypothesis by default'. Hypothesis is still run in travis, which specifically calls `pytest properties`

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,0 @@
-[pytest]
-addopts = -p no:hypothesis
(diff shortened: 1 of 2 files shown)


shoyer:
thanks!

PR #2031, opened 2018-03-30
Isin

maxim-lian:

This is an initial implementation of `isin`. It works for DataArrays, but isn't yet implemented further.

I've been away from these sorts features for too long: Is there a canonical place to put these? What's the canonical approach to extending to datasets? I can see a few approaches in the code.

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -746,6 +746,22 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         self._file_obj = None
 
     def isin(self, test_elements):
+        """Tests each value in the array for whether it is in the supplied list
+        Requires NumPy >= 1.13 
+
+        Parameters
+        ----------
+        element : array_like
+            Input array.
+        test_elements : array_like
+            The values against which to test each value of `element`.
+            This argument is flattened if it is an array or array_like.
+            See notes for behavior with non-array-like parameters.
+
+        -------
+        isin : same as object, bool
+            Has the same shape as object
+        """
         if LooseVersion(np.__version__) < LooseVersion('1.13.0'):
             raise ImportError('isin requires numpy version 1.13.0 or later')
         from .computation import apply_ufunc
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -755,8 +755,8 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
             Input array.
         test_elements : array_like
             The values against which to test each value of `element`.
-            This argument is flattened if it is an array or array_like.
-            See notes for behavior with non-array-like parameters.
+            This argument is flattened if an array or array_like.
+            See numpy notes for behavior with non-array-like parameters.
 
         -------
         isin : same as object, bool
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import warnings
+from distutils.version import LooseVersion
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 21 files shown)


maxim-lian:
Fails on Numpy pre 1.13. Is that too recent to upgrade min version? 1.14.2 is current, so would be aggressive

shoyer:
Let's just skip the tests if numpy is too old.

xarray/core/dataarray.py
```diff
@@ -2113,6 +2113,15 @@ def rank(self, dim, pct=False, keep_attrs=False):
         ds = self._to_temp_dataset().rank(dim, pct=pct, keep_attrs=keep_attrs)
         return self._from_temp_dataset(ds)
 
+    def isin(self, test_elements):
```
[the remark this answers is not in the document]
maxim-lian: Great. Anyone recall the newest way to run across Variables in a Dataset (i.e. more standard than a loop)? 

I'm sure I saw it in a recent PR but can't find it, either in the code or the recent PRs

Otherwise loop is fine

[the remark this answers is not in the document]
shoyer: `apply_ufunc` can handle an xarray.Dataset as an argument.

Are you thinking of using a Dataset for test_elements as well?

xarray/tests/test_dataset.py
```diff
@@ -4032,9 +4032,33 @@ def test_ipython_key_completion(self):
...
+    ).astype('bool')
+
+    result = (
+        data_set
+        .round(0)
+        .isin([0])
+        .sel(dim2=0, dim3='a')
+        .isel(dim1=[0, 1])
+        .drop(['time', 'dim3', 'dim2', 'numbers'])
+        .squeeze()
```
maxim-lian: Open to feedback on this formatting...

maxim-lian:
Any thoughts on this approach of writing out the result on a slice of a sample dataset / dataarray? 

I've been thinking about expect tests, as described by @yminsky [here](https://blog.janestreet.com/testing-with-expectations/). 
That would be something like: 
- Have some example datasets (similar to what we do now, though with a well known seed)
- Run our functions and save to a file, as a known good output
- During tests, compare the result to the known good output
- Where different, raise and show the diff

That's a bit harder with numerical data than with small lists of words (the example in the link), but also helpful - we don't have to manually construct the result in python - just check the first time & commit the result. And would enable tests across moderately sized data, rather than only 'toy' examples.

xarray/core/common.py
```diff
@@ -744,6 +745,35 @@ def close(self):
...
+            Has the same shape as object
+        """
+        if LooseVersion(np.__version__) < LooseVersion('1.13.0'):
+            raise ImportError('isin requires numpy version 1.13.0 or later')
+        from .computation import apply_ufunc
+
+        return apply_ufunc(
+            np.isin,
+            self,
+            kwargs=dict(test_elements=test_elements),
```
shoyer: It's probably a better idea to explicitly unwrap `.data` from `test_elements` if it's an xarray object, and explicitly raise for `xarray.Dataset`. Otherwise numpy will probably give a really strange error message.

shoyer: Indeed, see #2032 for what converting a Dataset to a numpy array does.

maxim-lian: I merged your branch - is that better than an additional check? 

Why extract `.data` - won't the standard machinery take care of that? I added a test

shoyer: I suppose you're right, the standard machinery will work fine here for now. In the future when dask supports isin (https://github.com/dask/dask/pull/3363) we'll want to use `.data` so we can keep it as a dask array.

xarray/tests/test_dataarray.py
```diff
@@ -3339,6 +3347,31 @@ def da_dask(seed=123):
...
+
+    expected = DataArray(
+        np.asarray([[0, 0, 1], [1, 0, 0]]),
+        dims=list('yx'),
+        coords={'x': list('abc'),
+                'y': list('de')},
+    ).astype('bool')
+    result = da.isin([2, 3]).sel(y=list('de'), z=0)
+    assert_equal(result, expected)
+
```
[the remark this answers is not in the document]
maxim-lian: (done in `test_dataset`, lmk if better to put in both)

maxim-lian:
Green! @shoyer

maxim-lian:
I'll merge this later tonight given @shoyer 's previous approval, unless there's any feedback

dcherian:
@shoyer the cookbook might be a good place for that

shoyer:
Indeed, but combined `where`/`isin` is also basically equivalent to indexing, so I think it's appropriate on that doc page, too.

maxim-lian:
Yes good idea. I'll add that to my (metaphorical) list.

PR #2038, opened 2018-04-05
isin: better docs, support older numpy and use dask.array.isin.

shoyer:
I added a section to the indexing docs showing how to use `isin` with `where` to do indexing.

Adding support for older numpy turned out to be pretty easy, so I added a copy of np.isin in npcompat.

This PR also makes use of dask.array.isin, so that we can support a dask array properly as the second argument:
https://github.com/dask/dask/pull/3363


maxim-lian:
Looks great!! Thank you

xarray/tests/test_dataset.py
```diff
@@ -4099,6 +4093,12 @@ def test_isin_dask(test_elements):
     assert_equal(result, expected)
 
 
+def test_isin_dataset():
+    ds = Dataset({'x': [1, 2]})
+    with pytest.raises(TypeError):
+        ds.isin(ds)
```
maxim-lian: You're right, this wasn't really an `xfail`

doc/whats-new.rst
```diff
@@ -37,18 +37,20 @@ Documentation
...
 
-- `~xarray.DataArray.isin` and `~xarray.Dataset.isin` methods, which test each value 
-  in the array for whether it is contained in the supplied list, returning a bool array.
-   Similar to the ``np.isin`` function. Requires NumPy >= 1.13 
-By `Maximilian Roos <https://github.com/maxim-lian>`
+- :py:meth:`~xarray.DataArray.isin` and :py:meth:`~xarray.Dataset.isin` methods,
+  which test each value in the array for whether it is contained in the
+  supplied list, returning a bool array. See :ref:`selecting values with isin`
+  for full details. Similar to the ``np.isin`` function.
+  By `Maximilian Roos <https://github.com/maxim-lian>`_.
```
maxim-lian: Feel free to change! Thank you for adding the docs

PR #2045, opened 2018-04-09
Implement setncattr_string for attributes that are lists of strings

### LINKED ISSUE #2044: Feature request: writing xarray list-type attributes to netCDF
Migrated from [Stack Overflow](https://stackoverflow.com/questions/49581072/writing-xarray-list-type-attributes-to-netcdf).

NetCDF supports the [NC_STRING](https://www.unidata.ucar.edu/software/netcdf/netcdf-4/newdocs/netcdf-c/Arrays-of-Strings.html) type, which can stores arrays of strings in attributes. **Xarray already supports reading arrays of strings from attributes in netCDF files, and It would be great if it also supported writing the same.**

### Reading already works

```python
import xarray as xr
import netCDF4 as nc

rg = nc.Dataset('test_string.nc', 'w', format='NETCDF4')
rg.setncattr_string('testing', ['a', 'b'])
rg.close()
ds = xr.open_dataset('test_string.nc')
print(ds)
```
gives
```
<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*
Attributes:
    testing:  ['a', 'b']
```

This works because I used the `setncattr_string` method. Setting the attributes like `rg.testing = ['a', 'b']` does not work and results in a concatenated list (just like the xarray example below).

### Writing doesn't work

```python
import xarray as xr
ds = xr.Dataset()
ds.attrs['testing'] = ['a', 'b']
ds.to_netcdf('asdf.nc')
ds = xr.open_dataset('asdf.nc', autoclose=True)
print(ds)
```
gives
```
<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*
Attributes:
    testing:  ab
```

Note the list elements have been concatenated. So this is a request for xarray to implement something like netCDF4's `setncattr_string`. I would be happy to help do this if someone pointed me in the right direction; I looked through the code but got lost pretty quickly. Thanks.

dnowacki-usgs:
- [X] Closes #2044 
 - [X] Tests added (for all bug fixes or enhancements)

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -34,6 +34,9 @@ v0.10.4 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- Support writing lists of strings as netCDF attributes (:issue:`2044`).
+  By `Dan Nowacki <https://github.com/dnowacki-usgs>`_.
+
 Bug fixes
 ~~~~~~~~~
 
--- a/xarray/backends/netCDF4_.py
+++ b/xarray/backends/netCDF4_.py
@@ -229,6 +229,31 @@ def _disable_auto_decode_group(ds):
         _disable_auto_decode_variable(var)
 
 
+def _is_list_of_strings(value):
+    if (np.asarray(value).dtype.kind in ['U', 'S'] and
+            np.asarray(value).size > 1):
+        return True
+    else:
+        return False
+
+
+def _set_nc_attribute(obj, key, value):
+    if _is_list_of_strings(value):
+        # encode as NC_STRING if attr is list of strings
+        try:
+            obj.setncattr_string(key, value)
+        except AttributeError:
+            # Inform users with old netCDF that does not support
+            # NC_STRING that we can't serialize lists of strings
+            # as attrs
+            msg = ('Attributes which are lists of strings are not '
+                   'supported with this version of netCDF. Please '
+                   'upgrade to netCDF4-python 1.2.4 or greater.')
+            raise AttributeError(msg)
+    else:
+        obj.setncattr(key, value)
+
+
 class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
     """Store for reading and writing data via the Python-NetCDF4 library.
 
@@ -347,7 +372,7 @@ class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
         with self.ensure_open(autoclose=False):
             if self.format != 'NETCDF4':
                 value = encode_nc3_attr_value(value)
-            self.ds.setncattr(key, value)
+            _set_nc_attribute(self.ds, key, value)
 
     def set_variables(self, *args, **kwargs):
         with self.ensure_open(autoclose=False):
@@ -402,7 +427,7 @@ class NetCDF4DataStore(WritableCFDataStore, DataStorePickleMixin):
         for k, v in iteritems(attrs):
             # set attributes one-by-one since netCDF4<1.0.10 can't handle
             # OrderedDict as the input to setncatts
-            nc4_var.setncattr(k, v)
+            _set_nc_attribute(nc4_var, k, v)
 
         target = NetCDF4ArrayWrapper(name, self)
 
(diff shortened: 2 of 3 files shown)


dnowacki-usgs:
OK, after some false starts, I think this is working, but the CI build is failing on CONDA_ENV=py35 and py36... but it's failing on a different test (specific to netcdf3, complaining about a temp file) and I can't see how it is impacted by my changes... thoughts?

xarray/backends/netCDF4_.py
```diff
@@ -343,11 +343,22 @@ def set_dimension(self, name, length, is_unlimited=False):
...
+            return False
+
     def set_attribute(self, key, value):
         with self.ensure_open(autoclose=False):
             if self.format != 'NETCDF4':
                 value = encode_nc3_attr_value(value)
-            self.ds.setncattr(key, value)
+            if self.is_list_of_strings(value) and (self.format == 'NETCDF4'):
+                # encode as NC_STRING if attr is list of strings and NETCDF4
+                self.ds.setncattr_string(key, value)
```
shoyer: Can we adjust this to raise a sensible error message if the user has an old version of netCDF4-python installed which doesn't have this method?

xarray/backends/netCDF4_.py
```diff
@@ -343,11 +343,22 @@ def set_dimension(self, name, length, is_unlimited=False):
...
+            return True
+        else:
+            return False
+
     def set_attribute(self, key, value):
         with self.ensure_open(autoclose=False):
             if self.format != 'NETCDF4':
                 value = encode_nc3_attr_value(value)
-            self.ds.setncattr(key, value)
+            if self.is_list_of_strings(value) and (self.format == 'NETCDF4'):
```
shoyer: We probably want to use this in all cases, even if a different format from NETCDF4 is selected. It's better to error than to improperly serialize data.

dnowacki-usgs: OK, I tested with netCDF4 1.2.2 (last release before setncattr_string was introduced) and successfully catch the error with this latest commit. Is this along the lines of what you were thinking?

xarray/backends/netCDF4_.py
```diff
@@ -402,7 +423,20 @@ def prepare_variable(self, name, variable, check_encoding=False,
...
+                except AttributeError:
+                    # Inform users with old netCDF that does not support
+                    # NC_STRING that we can't serialize lists of strings
+                    # as attrs
+                    msg = ('Attributes which are lists of strings are not '
+                           'supported with this version of netCDF. Please '
+                           'upgrade to netCDF4-python 1.2.4 or greater.')
+                    raise AttributeError(msg)
+            else:
+                nc4_var.setncattr(k, v)
```
shoyer: Could you consolidate this logic into a helper function (e.g., `_set_attribute()`) rather than repeating it twice?

dnowacki-usgs:
I only have a vague grasp of proper object oriented programming practices; not sure if the helper functions should be within the class or what, but it does seem to work this way.

xarray/tests/test_backends.py
```diff
@@ -552,6 +552,19 @@ def test_ondisk_after_print(self):
             repr(on_disk)
             assert not on_disk['var1']._in_memory
 
+    def test_setncattr_string(self):
```
shoyer: You might need to add this test specifically for the class with netCDF4 tests, making use of the `self.roundtrip()` method. Otherwise this is also going to run for cases where the list of strings attribute cannot be roundtripped (e.g., writing a netCDF3 file).

xarray/tests/test_backends.py
```diff
@@ -552,6 +552,19 @@ def test_ondisk_after_print(self):
...
+        """
+        Test to ensure setncattr_string is called on lists of strings
+        with NetCDF4
+        """
```
shoyer: Nit: we generally avoid writing docstrings for test method unless they add substantial information beyond what's already obvious from the class/method name.

dnowacki-usgs:
@shoyer Thank you for the helpful feedback. I think I addressed all of your comments, and also expanded the tests.

doc/whats-new.rst
```diff
@@ -34,6 +34,10 @@ v0.10.4 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- Support reading and writing lists of strings as attributes to netCDF
```
shoyer: Minor nit: this should be "Support writing lists" only. We can already read lists of strings.

shoyer:
thanks!

PR #2047, opened 2018-04-10
Fix decode cf with dask

### LINKED ISSUE #1372: decode_cf() loads chunked arrays

rabernat:
- [x] Closes #1372
 - [x] Tests added
 - [x] Tests passed

This was a very simple fix for an issue that has vexed me for quite a while. Am I missing something obvious here?

### DIFF
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -9,7 +9,7 @@ import pandas as pd
 from .coding import times, variables
 from .coding.variables import SerializationWarning
 from .core import duck_array_ops, indexing
-from .core.pycompat import OrderedDict, basestring, iteritems
+from .core.pycompat import OrderedDict, basestring, iteritems, dask_array_type
 from .core.variable import IndexVariable, Variable, as_variable
 
 
@@ -490,8 +490,9 @@ def decode_cf_variable(name, var, concat_characters=True, mask_and_scale=True,
         del attributes['dtype']
         data = BoolTypeArray(data)
 
-    return Variable(dimensions, indexing.LazilyOuterIndexedArray(data),
-                    attributes, encoding=encoding)
+    if not isinstance(data, dask_array_type):
+        data = indexing.LazilyOuterIndexedArray(data)
+    return Variable(dimensions, data, attributes, encoding=encoding)
 
 
 def decode_cf_variables(variables, attributes, concat_characters=True,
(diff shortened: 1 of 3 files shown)


xarray/conventions.py
```diff
@@ -490,8 +490,9 @@ def decode_cf_variable(name, var, concat_characters=True, mask_and_scale=True,
         del attributes['dtype']
         data = BoolTypeArray(data)
 
-    return Variable(dimensions, indexing.LazilyOuterIndexedArray(data),
-                    attributes, encoding=encoding)
+    if not isinstance(data, dask_array_type):
+        data = indexing.LazilyOuterIndexedArray(data)
```
rabernat: This is the offending line which caused the dask arrays to stop being dask arrays. I have simply bypassed it for dask arrays. This didn't break any tests. But can we think of any unforeseen negative consequences? What does `LazilyOuterIndexedArray` do here? I have to admit that I don't understand the details of our new indexing wrappers.

xarray/tests/test_conventions.py
```diff
@@ -332,6 +332,17 @@ def test_decode_cf_datetime_transition_to_invalid(self):
...
+    @requires_dask
+    def test_decode_cf_with_dask(self):
+        import dask
+        original = Dataset({
+            't': ('t', [0, 1, 2], {'units': 'days since 2000-01-01'}),
+            'foo': ('t', [0, 0, 0], {'coordinates': 'y', 'units': 'bar'}),
+            'y': ('t', [5, 10, -999], {'_FillValue': -999})
+        }).chunk({'t': 1})
+        decoded = conventions.decode_cf(original)
+        assert dask.is_dask_collection(decoded.y.data)
```
rabernat: Is this an appropriate place to test for this? There is no other dask stuff in `test_conventions.py`, but it seemed to make sense there.

rabernat:
Travis failures are for `GenericNetCDFDataTest.test_cross_engine_read_write_netcdf3` and appear to be unrelated to this PR.

xarray/tests/test_conventions.py
shoyer: This looks like a fine place to me.

xarray/conventions.py
shoyer: `LazilyOuterIndexedArray` makes indexing on non-dask arrays lazy. It's not necessary or useful for dask arrays.

PR #2054, opened 2018-04-13
Updates for the renaming of netcdftime to cftime

spencerkclark:
Addresses https://github.com/pydata/xarray/pull/1252#issuecomment-381131366

Perhaps I should have waited until `cftime` was up on conda-forge, but once that happens I can update this PR to use that in setting up the CI environments rather than pip.  

I made updates to the installing and time series pages of the docs.  Does this need a what's new entry?    I'm not sure which heading I would classify it under.

### DIFF
--- a/xarray/coding/times.py
+++ b/xarray/coding/times.py
@@ -38,24 +38,20 @@ TIME_UNITS = frozenset(['days', 'hours', 'minutes', 'seconds',
                         'milliseconds', 'microseconds'])
 
 
-def _import_netcdftime():
+def _import_cftime():
     '''
-    helper function handle the transition to netcdftime as a stand-alone
-    package
+    helper function handle the transition to netcdftime/cftime
+    as a stand-alone package
     '''
     try:
-        # Try importing netcdftime directly
-        import netcdftime as nctime
-        if not hasattr(nctime, 'num2date'):
-            # must have gotten an old version from netcdf4-python
-            raise ImportError
+        import cftime
     except ImportError:
         # in netCDF4 the num2date/date2num function are top-level api
         try:
-            import netCDF4 as nctime
+            import netCDF4 as cftime
         except ImportError:
-            raise ImportError("Failed to import netcdftime")
-    return nctime
+            raise ImportError("Failed to import cftime")
+    return cftime
 
 
 def _netcdf_to_numpy_timeunit(units):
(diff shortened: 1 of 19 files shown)


xarray/coding/times.py
```diff
@@ -38,24 +38,20 @@
...
-        import netcdftime as nctime
-        if not hasattr(nctime, 'num2date'):
-            # must have gotten an old version from netcdf4-python
-            raise ImportError
+        import cftime
     except ImportError:
         # in netCDF4 the num2date/date2num function are top-level api
         try:
-            import netCDF4 as nctime
+            import netCDF4 as cftime
```
shoyer: should this be `import netcdftime as cftime` or will that not work properly?

spencerkclark: Unfortunately it will not work properly, because `num2date` and `date2num` are not part of the `netcdftime` library that is packaged with `netCDF4` (they are top-level API in `netCDF4`).  `num2date` and `date2num` were added to `cftime` when it was split out of `netCDF4`.  

I agree - it is a little odd to be using `cftime` as an alias for `netCDF4`.  Ultimately this logic is in place to determine which version of the `num2date` and `date2num` methods to use.  Priority is placed on using the most up-to-date versions which are part of `cftime`, but if those cannot be found, then the versions in `netCDF4` are used.  Would it make more sense to have this helper function return those two methods themselves (rather than an alias for the module)?  For example:

```python
def _import_num2date_and_date2num():
    try:
        from cftime import num2date, date2num
    except ImportError:
        try:
            from netCDF4 import num2date, date2num
        except ImportError:
            raise ImportError('Failed to import cftime.num2date and cftime.date2num')
    return num2date, date2num
```
In instances where we only use one of those functions, we could always just ignore one or the other:
```python
_, date2num = _import_num2date_and_date2num()
```

shoyer: Eventually we will probably want to distinguish cftime from mere support for num2date and date2num from netCDF4. But we can tackle that later, as necessary. I think this is fine for now.

spencerkclark:
@jhamman I see that the builds were successful in https://github.com/conda-forge/staged-recipes/pull/5647 and it was subsequently merged, but I'm still not seeing it on conda-forge.  Can it take time for it to become available?
```
$ conda install -c conda-forge cftime
Fetching package metadata .............

PackageNotFoundError: Packages missing in current channels:

  - cftime

We have searched for the packages in the following channels:

  - https://conda.anaconda.org/conda-forge/osx-64
  - https://conda.anaconda.org/conda-forge/noarch
  - https://repo.continuum.io/pkgs/main/osx-64
  - https://repo.continuum.io/pkgs/main/noarch
  - https://repo.continuum.io/pkgs/free/osx-64
  - https://repo.continuum.io/pkgs/free/noarch
  - https://repo.continuum.io/pkgs/r/osx-64
  - https://repo.continuum.io/pkgs/r/noarch
  - https://repo.continuum.io/pkgs/pro/osx-64
  - https://repo.continuum.io/pkgs/pro/noarch
```

spencerkclark:
Thanks @ocefpaf and @jhamman! `cftime` is up on conda-forge now.  I updated the CI environments accordingly.

PR #2058, opened 2018-04-15
Refactor string coding and fix string support in zarr

### LINKED ISSUE #2057: Problem reading dtype=S64 with open_zarr
@jhamman suggested I raise [this SO question](https://stackoverflow.com/questions/49756981/round-tripping-zarr-data-from-xarray/49831196#49831196) as an issue here. 

I have a dataset that looks like:
```
<xarray.Dataset>
Dimensions:                     (nv: 2, reference_time: 746, time: 746, x: 4608, y: 3840)
Coordinates:
  * reference_time              (reference_time) datetime64[ns] 2018-03-07 ...
  * x                           (x) float64 -2.304e+06 -2.303e+06 -2.302e+06 ...
  * y                           (y) float64 -1.92e+06 -1.919e+06 -1.918e+06 ...
  * time                        (time) datetime64[ns] 2018-03-07T01:00:00 ...
Dimensions without coordinates: nv
Data variables:
    time_bounds                 (time, nv) datetime64[ns] dask.array<shape=(746, 2), chunksize=(1, 2)>
    ProjectionCoordinateSystem  (time) |S64 b'' b'' b'' b'' b'' b'' b'' b'' ...
    T2D                         (time, y, x) float64 dask.array<shape=(746, 3840, 4608), chunksize=(1, 3840, 4608)>
```

When writing this dataset using `ds.to_zarr` containing the `ProjectionCoordinateSystem` variable with `dtype=S64` , there can be an issue reading it using `ds.open_zarr` with the default `auto_chunk=True`. 

This example illustrates the problem:
```python
import xarray as xr
import s3fs

f_zarr = 'rsignell/nwm/test02'   
fs = s3fs.S3FileSystem(anon=False)
d = s3fs.S3Map(f_zarr, s3=fs)
xr.open_zarr(d)
```
returning
```python-traceback
<ipython-input-3-fff1cd753e7c> in <module>()
      2 fs = s3fs.S3FileSystem(anon=False)
      3 d = s3fs.S3Map(f_zarr, s3=fs)
----> 4 xr.open_zarr(d)

/opt/conda/lib/python3.6/site-packages/xarray/backends/zarr.py in open_zarr(store, group, synchronizer, auto_chunk, decode_cf, mask_and_scale, decode_times, concat_characters, decode_coords, drop_variables)
    476 
    477         variables = OrderedDict([(k, maybe_chunk(k, v))
--> 478                                  for k, v in ds.variables.items()])
    479         return ds._replace_vars_and_dims(variables)
    480     else:

/opt/conda/lib/python3.6/site-packages/xarray/backends/zarr.py in <listcomp>(.0)
    476 
    477         variables = OrderedDict([(k, maybe_chunk(k, v))
--> 478                                  for k, v in ds.variables.items()])
    479         return ds._replace_vars_and_dims(variables)
    480     else:

/opt/conda/lib/python3.6/site-packages/xarray/backends/zarr.py in maybe_chunk(name, var)
    471                 token2 = tokenize(name, var._data)
    472                 name2 = 'zarr-%s' % token2
--> 473                 return var.chunk(chunks, name=name2, lock=None)
    474             else:
    475                 return var

/opt/conda/lib/python3.6/site-packages/xarray/core/variable.py in chunk(self, chunks, name, lock)
    820             data = indexing.ImplicitToExplicitIndexingAdapter(
    821                 data, indexing.OuterIndexer)
--> 822             data = da.from_array(data, chunks, name=name, lock=lock)
    823 
    824         return type(self)(self.dims, data, self._attrs, self._encoding,

/opt/conda/lib/python3.6/site-packages/dask/array/core.py in from_array(x, chunks, name, lock, asarray, fancy, getitem)
   1988     >>> a = da.from_array(x, chunks=(1000, 1000), lock=True)  # doctest: +SKIP
   1989     """
-> 1990     chunks = normalize_chunks(chunks, x.shape)
   1991     if name in (None, True):
   1992         token = tokenize(x, chunks)

/opt/conda/lib/python3.6/site-packages/dask/array/core.py in normalize_chunks(chunks, shape)
   1918             raise ValueError(
   1919                 "Chunks and shape must be of the same length/dimension. "
-> 1920                 "Got chunks=%s, shape=%s" % (chunks, shape))
   1921 
   1922     if shape is not None:

ValueError: Chunks and shape must be of the same length/dimension. Got chunks=(3, 64), shape=(3,)
```
The full notebook is at https://gist.github.com/rsignell-usgs/dce09aae4f7cd174a141247a56ddea2c

shoyer:
Fixes #2057

This refactor will make it easier to use different coding for different backends. It also makes string decoding/encoding work properly with dask arrays, which I think was the issue responsible for #2057.

It includes using a special dtype with metadata (like h5py) to identify variable length strings. So in principle, we could now handle fixed width and variable length unicode differently.

I also took this opportunity to fix handling of strings in zarr. We now no longer convert into character arrays, but instead store as fixed length bytes (with UTF-8 encoding, if necessary). This wasn't really necessary, but it is a little cleaner, since zarr doesn't have the restriction of only supporting character arrays.

CC @rabernat @jhamman

### DIFF
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -236,7 +236,13 @@ def encode_zarr_variable(var, needs_copy=True, name=None):
     var = conventions.maybe_default_fill_value(var)
     var = conventions.maybe_encode_bools(var)
     var = conventions.ensure_dtype_not_object(var, name=name)
-    var = conventions.maybe_encode_string_dtype(var, name=name)
+
+    # zarr allows unicode, but not variable-length strings, so it's both
+    # simpler and more compact to always encode as UTF-8 explicitly.
+    # TODO: allow toggling this explicitly via dtype in encoding.
+    coder = coding.strings.EncodedStringCoder(allows_unicode=False)
+    var = coder.encode(var, name=name)
+
     return var
 
 
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-from base64 import b64encode
 from itertools import product
 from distutils.version import LooseVersion
 
@@ -25,21 +24,9 @@ def _encode_zarr_attr_value(value):
     # this checks if it's a scalar number
     elif isinstance(value, np.generic):
         encoded = value.item()
-        # np.string_('X').item() returns a type `bytes`
-        # zarr still doesn't like that
-        if type(encoded) is bytes:  # noqa
-            encoded = b64encode(encoded)
     else:
         encoded = value
     return encoded
-
-
-def _ensure_valid_fill_value(value, dtype):
-    if dtype.type == np.string_ and type(value) == bytes:  # noqa
-        valid = b64encode(value)
-    else:
-        valid = value
-    return _encode_zarr_attr_value(valid)
 
 
 class ZarrArrayWrapper(BackendArray):
@@ -221,27 +208,14 @@ def encode_zarr_variable(var, needs_copy=True, name=None):
         A variable which has been encoded as described above.
     """
 
-    if var.dtype.kind == 'O':
-        raise NotImplementedError("Variable `%s` is an object. Zarr "
-                                  "store can't yet encode objects." % name)
-
-    for coder in [coding.times.CFDatetimeCoder(),
-                  coding.times.CFTimedeltaCoder(),
-                  coding.variables.CFScaleOffsetCoder(),
-                  coding.variables.CFMaskCoder(),
-                  coding.variables.UnsignedIntegerCoder()]:
-        var = coder.encode(var, name=name)
-
-    var = conventions.maybe_encode_nonstring_dtype(var, name=name)
-    var = conventions.maybe_default_fill_value(var)
-    var = conventions.maybe_encode_bools(var)
-    var = conventions.ensure_dtype_not_object(var, name=name)
+    var = conventions.encode_cf_variable(var, name=name)
 
     # zarr allows unicode, but not variable-length strings, so it's both
     # simpler and more compact to always encode as UTF-8 explicitly.
     # TODO: allow toggling this explicitly via dtype in encoding.
     coder = coding.strings.EncodedStringCoder(allows_unicode=False)
     var = coder.encode(var, name=name)
+    var = coding.strings.ensure_fixed_length_bytes(var)
 
     return var
 
@@ -345,8 +319,7 @@ class ZarrStore(AbstractWritableDataStore):
         dtype = variable.dtype
         shape = variable.shape
 
-        fill_value = _ensure_valid_fill_value(attrs.pop('_FillValue', None),
-                                              dtype)
+        fill_value = attrs.pop('_FillValue', None)
         if variable.encoding == {'_FillValue': None} and fill_value is None:
             variable.encoding = {}
 
--- a/xarray/coding/strings.py
+++ b/xarray/coding/strings.py
@@ -49,7 +49,7 @@ class EncodedStringCoder(VariableCoder):
             if '_FillValue' in attrs:
                 raise NotImplementedError(
                     'variable {!r} has a _FillValue specified, but '
-                    '_FillValue is yet supported on unicode strings: '
+                    '_FillValue is not yet supported on unicode strings: '
                     'https://github.com/pydata/xarray/issues/1647'
                     .format(name))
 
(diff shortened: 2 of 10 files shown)


xarray/backends/zarr.py
```diff
@@ -221,22 +208,15 @@ def encode_zarr_variable(var, needs_copy=True, name=None):
...
-                  coding.variables.CFMaskCoder(),
-                  coding.variables.UnsignedIntegerCoder()]:
-        var = coder.encode(var, name=name)
-
-    var = conventions.maybe_encode_nonstring_dtype(var, name=name)
-    var = conventions.maybe_default_fill_value(var)
-    var = conventions.maybe_encode_bools(var)
-    var = conventions.ensure_dtype_not_object(var, name=name)
-    var = conventions.maybe_encode_string_dtype(var, name=name)
+    var = conventions.encode_cf_variable(var, name=name)
```
rabernat: omg so much simpler 👍

xarray/coding/strings.py
```diff
@@ -0,0 +1,219 @@
...
+        dims, data, attrs, encoding = unpack_for_encoding(variable)
+
+        contains_unicode = is_unicode_dtype(data.dtype)
+        encode_as_char = 'dtype' in encoding and encoding['dtype'] == 'S1'
+
+        if contains_unicode and (encode_as_char or not self.allows_unicode):
+            if '_FillValue' in attrs:
+                raise NotImplementedError(
+                    'variable {!r} has a _FillValue specified, but '
+                    '_FillValue is yet supported on unicode strings: '
```
rabernat: *not* yet supported

xarray/coding/strings.py
```diff
@@ -0,0 +1,219 @@
...
+                    'https://github.com/pydata/xarray/issues/1647'
+                    .format(name))
+
+            string_encoding = encoding.pop('_Encoding', 'utf-8')
+            safe_setitem(attrs, '_Encoding', string_encoding, name=name)
+            # TODO: figure out how to handle this in a lazy way with dask
+            data = encode_string_array(data, string_encoding)
```
rabernat: so does this evaluate eagerly?

xarray/coding/strings.py
```diff
@@ -0,0 +1,219 @@
...
+    """Convert numpy/dask arrays from fixed width bytes to characters."""
+    if arr.dtype.kind != 'S':
+        raise ValueError('argument must have a fixed-width bytes dtype')
+
+    if isinstance(arr, dask_array_type):
+        import dask.array as da
+        return da.map_blocks(_numpy_bytes_to_char, arr,
+                             dtype='S1',
+                             chunks=arr.chunks + ((arr.dtype.itemsize,)),
+                             new_axis=[arr.ndim])
```
rabernat: nice use of `map_blocks`

xarray/coding/strings.py
shoyer: Indeed, that's the case currently. We could probably do *this step* lazily with dask, but when we call `ensure_fixed_length_bytes` below it's also going to evaluate this into memory.

The challenge is that to write fixed-width strings to disk we need to know how much space to allocate *before* we start writing the data.

Alternative options:
1. Try to figure out the amount of space to allocate first by checking the maximum string length in a separate pass (for dask arrays). This would involve doing the calculation twice, but is probably better than eager evaluation.
2. Use encoding of some sort to indicate the data-type ahead of time, e.g., `dtype='S10'` to indicate that the data should be converted first into a string array of size 10. This might even already work in some form, I'm not quite sure with how the encoding logic is currently layered.

Potentially we should implement both options, issuing a warning in case (1). I don't know how common dask arrays of strings are yet, but I imagine handling this well will be useful eventually.

rabernat: In my work, I personally have never encountered a string array big enough to require chunking.

I just wish there were some way to be more explicit about any potential eager evaluation. In some cases, the user would rather see an error than have a big dask array silently start computing, since the later can potentially crash your whole system.

shoyer: I agree with you about being explicit. I guess we could add a keyword argument to `to_netcdf()`, e.g., `allow_computing_object_arrays`, defaulting to `False`?

We currently have to load string arrays with `dtype=object` twice:
- Once to figure out if they hold bytes or unicode (i.e., the true dtype for each chunk).
- A second time to figure out the maximum string length, if necessary.

In each case, this could be done by mapping over the dask array converting each chunk into a scalar. This would be safe from a memory consumption perspective (vs. computing the entire arrays), because dask already stores a fixed amount of metadata per chunk.

shoyer: I want to leave this to a follow-up issue, since it will be a significant under-taking and is not a regression from current behavior: https://github.com/pydata/xarray/issues/2069

shoyer:
OK, I'm planning on merging this shortly unless anyone else has comments....

PR #2073, opened 2018-04-22
recommend xr.Dataset.from_dataframe(df) instead of xr.Dataset(df) for…

DansYU:
… Multi-dimensional data


maxim-lian:
Should these return the same result?

shoyer:
I agree that `from_dataframe` should be preferred -- which is exactly why we mention in first on this doc page.

I'm not sure this change makes sense, since the leading paragraph introduces this example with "Both the DataArray and Dataset constructors..."

Issue #2088, opened 2018-04-27
Errors on indexing doc page


shoyer:
in this section: http://xarray.pydata.org/en/v0.10.3/indexing.html#more-advanced-indexing
```
In [73]: arr.sel(space=xr.DataArray(['IA', 'IL', 'IN'], dims=['new_time']),
   ....:         time=times)
   ....: 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-73-012885a177c1> in <module>()
      1 arr.sel(space=xr.DataArray(['IA', 'IL', 'IN'], dims=['new_time']),
----> 2         time=times)
      3 

~/checkouts/readthedocs.org/user_builds/xray/conda/v0.10.3/lib/python3.6/site-packages/xarray-0.10.3-py3.6.egg/xarray/core/dataarray.py in sel(self, method, tolerance, drop, **indexers)
    765         """
    766         ds = self._to_temp_dataset().sel(drop=drop, method=method,
--> 767                                          tolerance=tolerance, **indexers)
    768         return self._from_temp_dataset(ds)
    769 

~/checkouts/readthedocs.org/user_builds/xray/conda/v0.10.3/lib/python3.6/site-packages/xarray-0.10.3-py3.6.egg/xarray/core/dataset.py in sel(self, method, tolerance, drop, **indexers)
   1470         """
   1471         pos_indexers, new_indexes = remap_label_indexers(self, method,
-> 1472                                                          tolerance, **indexers)
   1473         result = self.isel(drop=drop, **pos_indexers)
   1474         return result._replace_indexes(new_indexes)

~/checkouts/readthedocs.org/user_builds/xray/conda/v0.10.3/lib/python3.6/site-packages/xarray-0.10.3-py3.6.egg/xarray/core/coordinates.py in remap_label_indexers(obj, method, tolerance, **indexers)
    344 
    345     pos_indexers, new_indexes = indexing.remap_label_indexers(
--> 346         obj, v_indexers, method=method, tolerance=tolerance
    347     )
    348     # attach indexer's coordinate to pos_indexers

~/checkouts/readthedocs.org/user_builds/xray/conda/v0.10.3/lib/python3.6/site-packages/xarray-0.10.3-py3.6.egg/xarray/core/indexing.py in remap_label_indexers(data_obj, indexers, method, tolerance)
    235     new_indexes = {}
    236 
--> 237     dim_indexers = get_dim_indexers(data_obj, indexers)
[...]

PR #2091, opened 2018-04-30
DOC: uniformize variable names in indexing.rst

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/2088

https://github.com/pydata/xarray/commit/6402391cf206fd04c12d44773fecd9b42ea0c246 overwrote an array which was needed later on. In general the whole page was a bit messy with the same initial array being forgotten and then re-used much later.

This PR attempts to uniformize the variable names in order to make self-consistent subsection examples. I hope I didn't brake something on the way, but it looks ok on my local build.

(also renamed a gallery example plot which wasn't showing up correctly)


shoyer:
Looks great to me, thank you!

PR #2087, opened 2018-04-27
Drop conflicted coordinate when assignment.

fujiisoup:
- [x] Closes #2068
 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

After this, when assigning a dataarray to a dataset, non-dimensional and conflicted coordinates of the dataarray are dropped.

example
```
In [2]: ds = xr.Dataset({'da': ('x', [0, 1, 2])}, 
   ...:                 coords={'y': (('x',), [0.1, 0.2, 0.3])})
   ...: ds
   ...: 
Out[2]: 
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
    y        (x) float64 0.1 0.2 0.3
Dimensions without coordinates: x
Data variables:
    da       (x) int64 0 1 2

In [3]: other = ds['da']
   ...: other['y'] = 'x', [0, 1, 2]  # conflicted non-dimensional coordinate
   ...: ds['da'] = other
   ...: ds
   ...: 
Out[3]: 
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
    y        (x) float64 0.1 0.2 0.3  # 'y' is not overwritten
Dimensions without coordinates: x
Data variables:
    da       (x) int64 0 1 2
```

PR #2100, opened 2018-05-02
Fix a bug introduced in #2087

### LINKED ISSUE #2099: Dataset.update wrongly handles the coordinate

I noticed a bug introduced by #2087 (my PR) 

```python
import xarray as xr

ds = xr.Dataset({'var': ('x', [1, 2, 3])}, 
                coords={'x': [0, 1, 2], 'z1': ('x', [1, 2, 3]), 
                        'z2': ('x', [1, 2, 3])})

ds['var'] = ds['var'] * 2
```
It claims a ValueError.

#### Problem description
https://github.com/pydata/xarray/blob/0cc64a08c672e6361d05acea3fea9f34308b62ed/xarray/core/merge.py#L564

Here should be 
```python
other[k] = obj.drop(coord_names)
```
not 
```python
other[k] = obj.drop(*coord_names)
```

fujiisoup:
- [x] Closes #2099 
 - [x] Tests added
 - [x] Tests passed

A quick fix for #2099

### DIFF
--- a/xarray/core/merge.py
+++ b/xarray/core/merge.py
@@ -561,7 +561,7 @@ def dataset_update_method(dataset, other):
             coord_names = [c for c in obj.coords
                            if c not in obj.dims and c in dataset.coords]
             if coord_names:
-                other[k] = obj.drop(*coord_names)
+                other[k] = obj.drop(coord_names)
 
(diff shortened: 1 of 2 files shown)


shoyer:
good catch!

PR #2101, opened 2018-05-03
DOC: Add resample e.g. Edit rolling e.g. Add groupby e.g.

raybellwaves:

Added a `resample` example to reflect the `DeprecationWarning: 
.resample() has been modified to defer calculations. Instead of passing 'dim' and 'how="mean", instead consider using...`. There were also some missing parameters in the docs.

Made a minor edit to my `rolling` example which uses the parameter `center`. I can't remember if that parameter was there last time but I think it's useful to return the time value for the middle of window.

Added a `groupby` example. Not sure if this should go here. See text below for my motivation of putting an example in the docstring of the class/object.

I learnt `xarray` before I learnt `pandas` so these example will hopefully be useful to other beginners as it took me a minute to get the syntax right.

Whilst there are great examples of these functions in the docs, a google search of a function for example `xarray rolling` often returns the doc entry that that object. Having a little example there is helpful for me to get the syntax right. The key/value pair in a dictionary for example is very powerful and always takes me a few tries to get the entry right.

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -426,7 +426,27 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         grouped : GroupBy
             A `GroupBy` object patterned after `pandas.GroupBy` that can be
             iterated over in the form of `(unique_value, grouped_array)` pairs.
-
+            
+        Examples
+        --------
+        Calculate daily anomalies for daily data:
+        
+        >>> da = xr.DataArray(np.linspace(0, 1826, num=1827),
+        ...                   coords=[pd.date_range('1/1/2000', '31/12/2004',
+        ...                           freq='D')],
+        ...                   dims='time')
+        >>> da
+        <xarray.DataArray (time: 1827)>
+        array([0.000e+00, 1.000e+00, 2.000e+00, ..., 1.824e+03, 1.825e+03, 1.826e+03])
+        Coordinates:
+          * time     (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 ...
+        >>> da.groupby('time.dayofyear') - da.groupby('time.dayofyear').mean('time')
+        <xarray.DataArray (time: 1827)>
+        array([-730.8, -730.8, -730.8, ...,  730.2,  730.2,  730.5])
+        Coordinates:
+          * time       (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 ...
+            dayofyear  (time) int64 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 ...
+        
         See Also
         --------
         core.groupby.DataArrayGroupBy
@@ -514,7 +534,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         --------
         Create rolling seasonal average of monthly data e.g. DJF, JFM, ..., SON:
 
-        >>> da = xr.DataArray(np.linspace(0,11,num=12),
+        >>> da = xr.DataArray(np.linspace(0, 11, num=12),
         ...                   coords=[pd.date_range('15/12/1999',
         ...                           periods=12, freq=pd.DateOffset(months=1))],
         ...                   dims='time')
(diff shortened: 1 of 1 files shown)


xarray/core/common.py
```diff
@@ -572,12 +598,53 @@ def resample(self, freq=None, dim=None, how=None, skipna=None,
...
+        >>> da
+        <xarray.DataArray (time: 12)>
+        array([  0.,   1.,   2.,   3.,   4.,   5.,   6.,   7., 8.,   9.,  10.,  11.])
+        Coordinates:
+          * time     (time) datetime64[ns] 1999-12-15 2000-01-15 2000-02-15 ...
+        >>> da.resample(time="Q-DEC").mean()
+        <xarray.DataArray (time: 4)>
+        array([ 1.,  4.,  7., 10.])
+        Coordinates:
+          * time     (time) datetime64[ns] 2000-02-29 2000-05-31 2000-08-31 2000-11-30
```
shoyer: Maybe this would be a good place to show upsampling, e.g., from monthly to daily? Maybe:
```
In [10]: da.resample(time='1D').interpolate('linear')
Out[10]:
<xarray.DataArray (time: 337)>
array([ 0.      ,  0.032258,  0.064516, ..., 10.935484, 10.967742, 11.      ])
Coordinates:
  * time     (time) datetime64[ns] 1999-12-15 1999-12-16 1999-12-17 ...
```

xarray/core/common.py
```diff
@@ -556,6 +576,12 @@ def resample(self, freq=None, dim=None, how=None, skipna=None,
```
shoyer: Could you please delete the sentence starting with "Upsampling with filling is not supported"? This is no longer true :)

xarray/core/common.py
```diff
@@ -572,12 +598,53 @@ def resample(self, freq=None, dim=None, how=None, skipna=None,
...
+        ...                           np.linspace(0,1,num=2)],
+        ...                   dims=['time', 'latitude'])
+        >>> da
+        <xarray.DataArray (time: 2, latitude: 2)>
+        array([[0.229808, 0.780577],
+               [0.725998, 0.666351]])
+        Coordinates:
+          * time      (time) datetime64[ns] 2000-01-01 2000-01-02
+          * latitude  (latitude) float64 0.0 1.0
+        >>> da.resample(freq='M', dim='time', how='mean')
```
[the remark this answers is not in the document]
shoyer: I'm also not entirely sure we need a separate example for N-dimensional data.

xarray/core/common.py
raybellwaves: Done

xarray/core/common.py
raybellwaves: Done

raybellwaves:
For `rolling` I just realized there is good section in the docs on the use of `.construct()` and `.reduce()`. Perhaps in the docstring of `rolling` I can point to http://xarray.pydata.org/en/stable/computation.html#rolling-window-operations although not sure where this would go in the docstring? Examples? See Also?

I opened up an issue on `resample` to use the new syntax with an n-d array. Once That's squared up i'll drop the 2-d example with the old syntax https://github.com/pydata/xarray/issues/2102

raybellwaves:
Closes https://github.com/pydata/xarray/issues/2102

raybellwaves:
Snuck in a assign_coords example. I often use climate data which has longitude values 0-359. I then may subset the data crossing the meridian. Making the longitude values from -180 - 180 allows an easy look at the data using `.plot()`. My example is based on the example here: https://gis.stackexchange.com/questions/205871/xarray-slicing-across-the-antimeridian

xarray/core/common.py
```diff
@@ -556,6 +576,12 @@ def resample(self, freq=None, dim=None, how=None, skipna=None,
 
         Parameters
         ----------
+        freq : str, pandas Offset Alias
+            Frequency for which to resample to.
+        dim : str, 
+            Name of dimension to resample.
+        how : str, optional
+            Method of resampling e.g. mean.
```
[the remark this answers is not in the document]
shoyer: This section should still be removed

Issue #1931, opened 2018-02-21
apply_ufunc produces illegal coordinate sizes


JiaweiZhuang:
If `func` changes the size of the core dimension, `apply_ufunc(func, ..)` will only change the data variable dimension but keep the coordinate dimension and value unchanged. The resulting DataArray cannot be saved to a NetCDF file due to dimension inconsistency. Please see this [GitHub gist](https://gist.github.com/JiaweiZhuang/89a6bce7d5adc473134c7014bec065a0) for a simple illustration. 

Would it be more natural to drop the core coordinate by default? This is safer (will not produce illegal NetCDF file) and also makes more physical sense (the core coordinate is likely to change).

I'd like use `apply_ufunc` to track metadata in my xESMF package, but the output DataArray will have incorrect `lon` and `lat` coordinate dimension and values. It is easy to manually correct this afterwards but I want to bring up this issue.

#### Output of ``xr.show_versions()``
<details>

commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Darwin
OS-release: 16.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0
pandas: 0.22.0
numpy: 1.13.3
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.0
matplotlib: 2.0.2
cartopy: 0.15.1
seaborn: 0.8.0
setuptools: 38.2.5
pip: 9.0.1
conda: None
pytest: 3.2.5
IPython: 6.1.0
sphinx: 1.6.5
</details>

jhamman:
@JiaweiZhuang - thanks for the report. Based on your example, I think we should be raising an error when constructing the output object. That won't help your use case which we do want to support so I think another option to drop the coordinate may also be in order.

rabernat:
👍 I agree we should try to fix these issues upstream (in xarray) rather than compelling @JiaweiZhuang to work around them in xESGM. Thanks for the bug report.

JiaweiZhuang:
@jhamman @rabernat Thanks for the help!

Raising an error when encountering this issue and adding `keep_core_coords=False` to optionally drop the coordinate would be a good solution for me. 

[...]

Issue #1552, opened 2017-09-03
Flow chart for choosing indexing operations


shoyer:
We have a lot of indexing operations, even though `sel_points` and `isel_points` are about to be deprecated (#1473).

A flow chart / decision tree to help users pick the right indexing operation might be helpful (e.g., like [this skimage FlowChart](http://scikit-learn.org/stable/tutorial/machine_learning_map/index.html)). It would ask various questions (e.g., do you have labels or integer positions? do you want to select or impose coordinates?) and then suggest appropriate the indexer methods.

cc @fujiisoup

PR #2104, opened 2018-05-04
implement interp()

### LINKED ISSUE #2079: New feature: interp1d
I've written a series of wrappers for the 1-dimensional scipy interpolators.

Prototype code and colourful demo plots:
https://gist.github.com/crusaderky/b0aa6b8fdf6e036cb364f6f40476cc67

# Features
- Interpolate a ND array on any arbitrary dimension
- Nearest-neighbour, linear, quadratic, cubic, Akima, PCHIP, and custom interpolators are supported
- dask supported on both on the interpolated array and x_new
- Supports ND x_new arrays
- The CPU-heavy interpolator generation (splrep) is executed only once and then can be applied to multiple x_new (splev)
- Pickleable and distributed-friendly 

# Design hacks
- Depends on dask module, even when all inputs are based on plain numpy.
- Abuses attrs and the ability to invoke a.attrname to get the user experience of a new DataArray method.
- Abuses the fact that the chunks of a ``dask.array.Array`` can contain anything and you won't notice until you compute them.
# Limitations
- Can't dump to netcdf. Not solvable without hacking into the implementation details of scipy.
- Datasets are not supported. Trivial to fix after solving #1699.
- Chunks are not supported on x_new. Trivial to fix after solving #1995.
- Chunks are not supported along the interpolated dimension. This is very complicated to solve. If x and x_new were always monotonic ascending,it would be (not trivially) solvable with dask.array.ghost.ghost. If you make no assumptions about monotonicity, things become way more complicated. A solution would need to go in the dask module, and then be invoked trivially from here with dask='allowed'.

fujiisoup:
- [x] Closes #2079 (remove if there is no corresponding issue, which should only be the case for minor changes)

I started working to add `interpolate_at` to xarray, as discussed in issue #2079 (but without caching).

I think I need to take care of more edge cases, but before finishing up this PR, I want to discuss what the best API is.

I would like to this method working similar to `isel`, which may support *vectorized* interpolation.
Currently, this works as follwos

```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: 
   ...: da = xr.DataArray([0, 0.1, 0.2, 0.1], dims='x', coords={'x': [0, 1, 2, 3]})
   ...: 

In [2]: # simple linear interpolation
   ...: da.interpolate_at(x=[0.5, 1.5])
   ...: 
Out[2]: 
<xarray.DataArray (x: 2)>
array([0.05, 0.15])
Coordinates:
  * x        (x) float64 0.5 1.5

In [3]: # with cubic spline interpolation
   ...: da.interpolate_at(x=[0.5, 1.5], method='cubic')
   ...: 
Out[3]: 
<xarray.DataArray (x: 2)>
array([0.0375, 0.1625])
Coordinates:
  * x        (x) float64 0.5 1.5

In [4]: # interpolation at one single position
   ...: da.interpolate_at(x=0.5)
   ...: 
Out[4]: 
<xarray.DataArray ()>
array(0.05)
Coordinates:
    x        float64 0.5

In [5]: # interpolation with broadcasting
   ...: da.interpolate_at(x=xr.DataArray([[0.5, 1.0], [1.5, 2.0]], dims=['y', 'z']))
   ...: 
Out[5]: 
<xarray.DataArray (y: 2, z: 2)>
array([[0.05, 0.1 ],
       [0.15, 0.2 ]])
Coordinates:
    x        (y, z) float64 0.5 1.0 1.5 2.0
Dimensions without coordinates: y, z

In [6]: da = xr.DataArray([[0, 0.1, 0.2], [1.0, 1.1, 1.2]], 
   ...:                   dims=['x', 'y'], 
   ...:                   coords={'x': [0, 1], 'y': [0, 10, 20]})
   ...: 

In [7]: # multidimensional interpolation
   ...: da.interpolate_at(x=[0.5, 1.5], y=[5, 15])
   ...: 
Out[7]: 
<xarray.DataArray (x: 2, y: 2)>
array([[0.55, 0.65],
       [ nan,  nan]])
Coordinates:
  * x        (x) float64 0.5 1.5
  * y        (y) int64 5 15

In [8]: # multidimensional interpolation with broadcasting
   ...: da.interpolate_at(x=xr.DataArray([0.5, 1.5], dims='z'), 
   ...:                   y=xr.DataArray([5, 15], dims='z'))
   ...: 
Out[8]: 
<xarray.DataArray (z: 2)>
array([0.55,  nan])
Coordinates:
    x        (z) float64 0.5 1.5
    y        (z) int64 5 15
Dimensions without coordinates: z
```

#### Design question

1. How many interpolate methods should we support?
  Currently, I only implemented [scipy.interpolate.interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html) for 1dimensional interpolation and [scipy.interpolate.RegularGridInterpolator](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html) for multidimensional interpolation.
I think 90% usecases are linear, but there are more methods in scipy.

2. How do we handle `nan`?
  Currently this raises ValueError if nan is present. It may be possible to carry out the interpolation with skipping nan, but in this case the performance would be significantly drops because it cannot be vectorized.

3. Do we support interpolation along *dimension without coordinate*?
  In that case, do we attach new coordinate to the object?

4. How should we do if new coordinate has the dimensional coordinate for the dimension to be interpolated?
e.g. in the following case, 
```python
da = xr.DataArray([0, 0.1, 0.2, 0.1], dims='x', coords={'x': [0, 1, 2, 3]})
rslt = da.interpolate_at(x=xr.DataArray([0.5, 1.5], dims=['x'], coords={'x': [1, 3]})
```
what would be `rslt['x']`?

I appreciate any comments.


fmaussion:
@fujiisoup this is awesome! So glad you are looking into this. 

To point 1: ``How many interpolate methods should we support?``. 

I think that your implementation is a very good first step. From the geosciences point of view I think that [Spline interpolation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RectBivariateSpline.html#scipy.interpolate.RectBivariateSpline) would be a very good candidate too, but it only works in the 2-dimensional case which makes it a no-go for the general case...

jhamman:
@fujiisoup - This is great. I wonder if we can draw more from the interpolator adapters I built in `missing.py`: 

https://github.com/pydata/xarray/blob/b9f40cc1da9c45b3dd33a3434b69c3d8fce57138/xarray/core/missing.py#L284-L323

I just gave your implementation a quick skim and I think there is currently some duplication of efforts so maybe we look there to see what can be combined.

shoyer:
First of all, this is awesome!

One question: since I think this new interpolation method will be used quite frequently (more often than `interpolate_na` and `reindex`), perhaps we should give it a shorter name. Maybe `interp`?

We should also consider adding `interp_like` or `interpolate_like` by analogy to `reindex_like`.

> I would like to this method working similar to isel, which may support vectorized interpolation.

Yes, I agree. Interpolation should be vectorized similarly to `isel`/`sel`.

In particular, in the long term I think we should aim to make `interpolate_at(..., method='nearest')` and `reindex(..., method='nearest')` equivalent.

> How many interpolate methods should we support?

I think this is fine to start (we can always add more later).

> How do we handle nan?

Which NaN are you referring to?

1. `data` on the arrays being indexed
2. Coordinates on the arrays being indexed
3. Points at which to interpolate

Case (1) should definitely be supported. Especially if data is stored in dask arrays, we cannot necessarily check if there are NaNs.

Cases (2) and (3) are not important, because there are relatively few use-cases for NaN in coordinate arrays.

> Do we support interpolation along dimension without coordinate?
In that case, do we attach new coordinate to the object?

Currently we don't support this for `reindex()`, but I suppose we could do so unambiguous.

> How should we do if new coordinate has the dimensional coordinate for the dimension to be interpolated?

I think the new coordinates should take priority, and the dimension coordinates on the new coordinate should be dropped. This is similar to what we do for `sel()`:

```
In [10]: da = xr.DataArray([0, 0.1, 0.2, 0.1], dims='x', coords={'x': [0, 1, 2, 3]})
    ...:

In [11]: da.sel(x=xr.DataArray([1, 2], dims=['x'], coords={'x': [1, 3]}))
    ...:
Out[11]:
<xarray.DataArray (x: 2)>
array([0.1, 0.2])
Coordinates:
  * x        (x) int64 1 2
```

fujiisoup:
@jhamman,
> I just gave your implementation a quick skim and I think there is currently some duplication of efforts so maybe we look there to see what can be combined.

Agreed. I will try to generalize your interpolator-adaptor more.

@shoyer 
> perhaps we should give it a shorter name. Maybe interp?

Agreed. `interp` sounds nice.

> Which NaN are you referring to?
> 1. data on the arrays being indexed
> 2. Coordinates on the arrays being indexed
> 3. Points at which to interpolate

I meant 1.

`Scipy.interpolate.interp1d` does not support arrays with NaN.
So if we want to support this, we need to interpolate the array line by line as `interpolate_na` does, i.e. no vectorization.
For the case of multidimensional interpolation of arrays with NaN, it is equivalent with the interpolation on *non-grided data*. It is possible, but the performance will significantly drop.

xarray/core/interp.py
```diff
@@ -0,0 +1,152 @@
...
+    Only consider a subspace that is needed for the interpolation
+    """
+    for dim, [x, new_x] in index_coord.items():
+        try:
+            imin = x.to_index().get_loc(np.min(new_x), method='ffill')
+            imax = x.to_index().get_loc(np.max(new_x), method='bfill')
+
+            idx = slice(np.maximum(imin - 1, 0), imax + 1)
+            index_coord[dim] = (x[idx], new_x)
+            obj = obj.isel(**{dim: idx})
```
fujiisoup: Currently, only a small portion of arrays will be used for the interpolation with `method='linear' / 'nearest '`. So if there are no NaN in this region, we can interpolate this array.

shoyer:
I still need to look at the code here, but I don't think we need separate code paths for NaN vs no-NaNs.  For example, suppose we have a DataArray with a coordinate `x=[0, 1, 2]`, then `da.interp(x=0.5, method='linear')` should be equal to `0.5 * (da.sel(x=0) + da.sel(x=1))`, regardless of the values of the DataArray. If either `da[0]` or `da[1]` is NaN, then the result would be NaN, too.

fujiisoup:
@shoyer ,
Yes, `linear` and `nearest` can handle NaN intuitively, but other methods need to invert some matrices. For example, cubic spline can not be used with nan-including arrays,
```python
In [1]: x = np.arange(4)
   ...: y = np.array([0, 1, np.nan, 3])
   ...: 

In [2]: # 'linear' works as expected
   ...: interp1d(x, y, kind='linear')([0.5, 1.5])
   ...: 
Out[2]: array([0.5, nan])

In [3]: # 'cubic' returns an all nan array
   ...: interp1d(x, y, kind='cubic')([0.5, 1.5])
   ...: 
Out[3]: array([nan, nan])
```

fujiisoup:
> RegularGridInterpolator handles NaNs pretty much the same way @shoyer describes it.

I did not certainly realize how RegularGridInterpolator works, but it seems to work with nan including array too as you pointed out. 
But higher order interpolations (RegularGridInterpolator is `linear`), such as `RectBivariateSpline`, would not work (return an all nan array) when just 1 nan is included in the target array.
It is the same with higher order interpolations in `scipy.interpolate.interp1d` such as `cubic`.

> Why don't leave the NaN handling part to the actual function doing the interpolation? 

It is the original behavior of `scipy.interpolate` functions (except for `linear` or `nearest` methods), which do some matrix inversions in its algorithm.

I think the addition of the linear interpolation is still nice, but it would be nicer if we could also implement nan-skipping interpolations, which will enable us to use higher order interpolation with nan-including arrays.

fmaussion:
> I think the addition of the linear interpolation is still nice, but it would be nicer if we could also implement nan-skipping interpolations, which will enable us to use higher order interpolation with nan-including arrays.

Do you mean you want to use unstructured data interpolation routines? (https://docs.scipy.org/doc/scipy/reference/interpolate.html#multivariate-interpolation)

fujiisoup:
Yes, I am thinking this for multidimensional interpolation of nan-including arrays.
(For 1-dimensional interpolation, we can interpolate them row by row)
Although this should be another work (PR), I am wondering what the best API is.

Is `skipna=True` option handy? If so, which should be the default?

shoyer:
My inclination would be to limit this method to interpolation on regular grids, i.e., the methods supported by [scipy.interpolate.interpn](https://docs.scipy.org/doc/scipy-1.0.0/reference/generated/scipy.interpolate.interpn.html) or interp1d.

I think a `skipna=True` option could be useful, but I don't know how we could do that for regular interpolation with multi-dimensional data. So I think it might make sense to save `skipna=True` for interpolation on unstructured grids.

fujiisoup:
> So I think it might make sense to save skipna=True for interpolation on unstructured grids.

I agreed.
I updated some of this PR.

The proposed API is
```python
da.interp(x=[0, 1, 2], method='linear', 
          kwargs={'bounds_error': False, 'fill_value': 'extrapolate'})
```

Honestly, I am not very happy with `kwargs={}` arguments (it would be more beautiful if they can be keyword arguments), but I think it is a necessary pain.

fmaussion:
> Honestly, I am not very happy with kwargs={} arguments (it would be more beautiful if they can be keyword arguments), but I think it is a necessary pain

I don't think it's *that* bad. It makes it clear that the arguments are passed over to the underlying function, and that xarray is not doing all the magic by itself.

fmaussion:
Let us know when this is ready for review! I'm happy to have a look at it.

fujiisoup:
Thanks, @fmaussion.
I think it is almost ready.

The remaining things are 
+ I am looking for a good-looking example on rtd.
+ Just before the merge, I'm going to rename missing.py to interp.py so that the filename is more descriptive.

doc/interpolation.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. ipython:: python
+   :suppress:
+
+    import numpy as np
+    import pandas as pd
+    import xarray as xr
+    np.random.seed(123456)
+
+xarray offers flexible interpolating routines, which has a similar interface
```
fmaussion: flexible interpolati**on** routines, with a similar interface to :ref:`indexing <indexing>`.

doc/interpolation.rst
```diff
@@ -0,0 +1,183 @@
...
+
+    # label lookup
+    da.sel(time=[2, 3])
+
+    # interpolation
+    da.interp(time=[2.5, 3.5])
+
+.. note::
+
+  Currently, our interpolation only works for gridded array.
```
fmaussion: gridded (structured) arrays

doc/interpolation.rst
```diff
@@ -0,0 +1,183 @@
...
+  Therefore, as similar to :py:meth:`~xarray.DataArray.sel`,
+  only dimensional coordinates can be used for the original coordinate to be
```
fmaussion: In theory it should work with non-dimension coordinates with the same shape and dimension as the dimension coordinate, right? (http://xarray.pydata.org/en/latest/data-structures.html#coordinates)

doc/interpolation.rst
```diff
@@ -0,0 +1,183 @@
...
+    da.interp(time=[1.5, 2.5], space=[0.15, 0.25])
+
+
+Interpolation methods
+---------------------
+
+We use scipy.interpolate.interp1d for 1-dimensional interpolation and
+scipy.interpolate.interpn for multi-dimensional interpolation.
+
+The interpolation method can be specified by ``method`` optional argument.
```
fmaussion: **the** ``method``

doc/interpolation.rst
```diff
@@ -0,0 +1,183 @@
...
+you can pass two 1-dimensional :py:class:`~xarray.DataArray` s with
+a common dimension as new coordinate.
+
+.. image:: _static/advanced_selection_interpolation.svg
+    :height: 200px
+    :width: 400 px
+    :alt: advanced indexing and interpolation
+    :align: center
+
+The example snippest is shown below,
```
fmaussion: For example:

doc/interpolation.rst
```diff
@@ -0,0 +1,183 @@
...
+
+For the details of the advanced indexing,
+see :ref:`more advanced indexing <more_advanced_indexing>`.
+
+
+Interpolating arrays with NaN
+-----------------------------
+
+Our :py:meth:`~xarray.DataArray.interp` works with arrays with NaN
+as the same way that `scipy.interpolate.interp1d` and scipy.interpolate.interpn` do.
```
fmaussion: remove "as"

doc/interpolation.rst
fujiisoup: Yes, it should work in theory. But currently, it is not supported, just because of the analogy to our `.sel()`.
To support this, we may need to make an additional design decision, 
+ which coordinate should be the dimensional coordinate if the array is interpolated along non-dimensional coordinate?
+ how should we do if the argument is a DataArray?

I suppose in the future we may support unstructured data, where we interpolate data along non-dimensional coordinates.
I think it is probably cleaner that the current `.interp()` only supports dimensional coordinates, and leave the interpolation along non-dimensional coordinates to the future implementation of the unstructured array support.

fujiisoup:
Thanks, @fmaussion 

The current design question is

[ ] should we support the interpolation along the non-dimensional coordinate?
[ ] what should be the default value of 'bounds_error' and `fill_value`. 
  scipy's interpolator assumes `bounds_error=True` by default, but I think `bounds_error=False` is more xarray-like.

shoyer:
> should we support the interpolation along the non-dimensional coordinate?

As I think I've mentioned before, I think we should save interpolation of unstructured grids for later -- if only to keep the complexity of this PR to the minimum.

Interpolation on unstructured grids needs some of auxiliary data structure to be done efficiently, so it might make sense to wait until after we support flexible index types such as a KDTree index saved on xarray objects. That said, we could also add it now (immediately after this PR), and add support for caching KDTrees later.

I think the API for unstructured grid interpolation is pretty straightforward and could even reuse the same `interp`/`interpolate_at` method. Unstructured grid interpolation would be triggered whenever multiple existing coordinates that share some of the same dimensions are interpolated.

> what should be the default value of 'bounds_error' and fill_value.
scipy's interpolator assumes bounds_error=True by default, but I think bounds_error=False is more xarray-like, but it may be unintuitive for users very familiar with scipy's behavior.

I agree, by default we should switch to `bounds_error=False`, and replace extrapolated values with NaN, which is already the default fill value for `interp1d`.

fujiisoup:
> As I think I've mentioned before, I think we should save interpolation of unstructured grids for later -- if only to keep the complexity of this PR to the minimum.

I agreed that it would be better to concentrate on structured interpolation only in this PR.
I concerned about the interpolation along **1-dimensional** but non-dimensional coordinate,
which @fmaussion suggested in a review [comment](https://github.com/pydata/xarray/pull/2104#discussion_r186745479).

It is actually the structured interpolation, but currently, I do not support this just because of the analogy to `.sel()`.
If we want to support this, we may need to decide its behavior in particular the dimension or coordinate of the interpolated array.
I prefer not to support this now for keeping the interpolation rule simple.

fmaussion:
> I prefer not to support this now for keeping the interpolation rule simple.

Agreed. Users can turn their 1D non dimension coordinates into dimension coordinates if they want to.

fujiisoup:
Thanks, @fmaussion
I didn't realize that scipy.interpolate.interpn does not sort the original coordinates (interp1d does).
Thanks for pointing this out!

> The implementation for you would be straightforward, you'd just have to pass np.arange(NN) to scipy instead of the actual coordinates.

As the interpolation routine can also be used for non-uniform gridded data, I don't think passing np.arange(NN) to scipy is a good idea (it will change the result if higher order method such as 'cubic' is used).
Instead, I would like to call `sortby` in our `interp` routine, so that the array passed to scipy is always sorted in the ascending order.

The new API I would propose is
```python
da.interp(x=[0, 1, 2], method='linear', assume_sorted=True,
          kwargs={'bounds_error': False, 'fill_value': 'extrapolate'})
```
with a new keyword `assume_sorted`.

xarray/core/dataset.py
```diff
@@ -1775,6 +1783,74 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
...
+
+        indexers_list = self._validate_indexers(coords)
+
+        obj = self if assume_sorted else self.sortby([k for k in coords])
+
+        def maybe_variable(obj, k):
+            try:
+                return obj._variables[k]
+            except KeyError:
+                return as_variable((k, range(obj.dims[k])))
```
fujiisoup: Do we have a cleaner solution to get variable corresponding to a dimension without coordinate?
I found `_get_virtual_variables`, but it looks overkill.

fmaussion:
> As the interpolation routine can also be used for non-uniform gridded data, I don't think passing np.arange(NN) to scipy is a good idea (it will change the result if higher order method such as 'cubic' is used).

Right. Another feature of geospatial grids is that they are regularly spaced, which is not always the case here. Forget about my idea for now.

I tested your updated branch and we now come to the same results with my test data! I took the liberty to edit your [comment](https://github.com/pydata/xarray/pull/2104#issuecomment-387914727) because in your current implementation ``assume_sorted`` is ``False`` per default. 

Another wish (sorry! ;): it would be great to implement the ``keep_attrs`` kwarg as well, because interpolation often conserves the units of data. Thanks!!!!

fujiisoup:
Thank you for the double checking :+1: 

> it would be great to implement the keep_attrs kwarg as well, because interpolation often conserves the units of data.

Agreed.
Is there any problem if we always keep attrs in the interpolation?
`.sel` always keeps attrs and I think it would be better if users have less choice in each method.

BTW, do you have any idea about an example for the new interp feature?
It would be nice if we have a good-looking example on the doc.

fmaussion:
> BTW, do you have any idea about an example for the new interp feature?

A quick shot would be to use the existing sample dataset:

```python
import xarray as xr
import numpy as np

# Raw data
ds = xr.tutorial.load_dataset('air_temperature')
ds.air.isel(time=0).plot()
plt.title('Raw data')

# Interpolated data
new_lon = np.linspace(ds.lon[0], ds.lon[-1], ds.dims['lon'] * 4)
new_lat = np.linspace(ds.lat[0], ds.lat[-1], ds.dims['lat'] * 4)
dsi = ds.interp(lon=new_lon, lat=new_lat)
dsi.air.isel(time=0).plot()
plt.title('Interpolated data')
```
Which produces the plots:

index
index2

fmaussion:
> Is there any problem if we always keep attrs in the interpolation?

In my opinion it should be the default, yes.

fujiisoup:
@fmaussion , thanks for the example! It's a nicely simple but descriptive example. Updated docs.

After your suggestion about attrs, I found a bug in `apply_ufunc` (#2114). The tests are added but now failing because of this issue.

dcherian:
I think it would be useful to have a table in the docs showing the differences between /  when to use `interp`, `interpolate_na`, `reindex`, `sel(..., method='nearest')`, `resample`.

fujiisoup:
> I think it would be useful to have a table in the docs showing the differences between / when to use interp, interpolate_na, reindex, sel(..., method='nearest'), resample.

Agreed. Thanks for the nice suggestion :+1:
We have discussed a cheat sheet #1552 previously, but no work has been done yet.

I added a simple [illustration](https://github.com/fujiisoup/xarray/blob/b3c76d7fec45e912b5ff688b5ceb74467cab1e51/doc/_static/advanced_selection_interpolation.svg) presenting the difference between advanced-`sel` and -`interp` for this addition, and I think we can enrich this kind of figure to clarify the differences among these set of methods.

dcherian:
@fujiisoup That illustration is great. Thanks for all this great work. Yes, adding more information to that would work well.

fujiisoup:
I think this is ready for final review

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
     return (arr.isnull().rolling(min_periods=1, **kw)
             .construct(new_dim, fill_value=False)
             .sum(new_dim, skipna=False)) <= limit
+
+
+def _assert_single_chunk(obj, axes):
+    for axis in axes:
+        if len(obj.chunks[axis]) > 1 or obj.chunks[axis][0] < obj.shape[axis]:
+            raise ValueError('Chunk along the dimension to be interpolated '
```
shoyer: Chunk -> Chunking

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+def _localize(obj, indexes_coords):
+    """ Speed up for linear and nearest neighbor method.
+    Only consider a subspace that is needed for the interpolation
+    """
+    for dim, [x, new_x] in indexes_coords.items():
+        try:
+            imin = x.to_index().get_loc(np.min(new_x.values), method='nearest')
+            imax = x.to_index().get_loc(np.max(new_x.values), method='nearest')
+
+            idx = slice(np.maximum(imin - 2, 0), imax + 2)
```
shoyer: nit: prefer the builtin `max()` on scalar values

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+            indexes_coords[dim] = (x[idx], new_x)
+            obj = obj.isel(**{dim: idx})
+        except ValueError:  # if index is not sorted.
```
shoyer: I thought we already ensure coordinates are sorted with `assume_sorted`?

xarray/core/dataset.py
```diff
@@ -1782,6 +1790,74 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
...
+        See Also
+        --------
+        scipy.interpolate.interp1d
+        scipy.interpolate.interpn
+        """
+        from . import missing
+
+        indexers_list = self._validate_indexers(coords)
+
+        obj = self if assume_sorted else self.sortby([k for k in coords])
```
shoyer: At some point it might be worth profiling to see if sorting adds significant overhead (due to copying data in the underlying array). In that case, we could potentially check the `pandas.Index.is_monotonic_increasing` property to avoid needing to sort.

xarray/core/dataset.py
```diff
@@ -1782,6 +1790,74 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
...
+            coords can be an integer, array-like or DataArray.
+            If DataArrays are passed as coords, their dimensions are used
+            for the broadcasting.
+        method: {'linear', 'nearest'} for multidimensional array,
+            {'linear', 'nearest', 'zero', 'slinear', 'quadratic', 'cubic'}
+            for 1-dimensional array.
+        assume_sorted: boolean, optional
+            If False, values of x can be in any order and they are sorted
+            first. If True, x has to be an array of monotonically increasing
+            values.
```
shoyer: This should have a disclaimer, something like: "If True, interpolated coordinates are assumed to be an array of monotonically increasing values"

xarray/core/dataset.py
```diff
@@ -1782,6 +1790,74 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
...
+        **coords : {dim: new_coordinate, ...}
+            Keyword arguments with names matching dimensions and values.
+            coords can be an integer, array-like or DataArray.
+            If DataArrays are passed as coords, their dimensions are used
+            for the broadcasting.
+        method: {'linear', 'nearest'} for multidimensional array,
+            {'linear', 'nearest', 'zero', 'slinear', 'quadratic', 'cubic'}
+            for 1-dimensional array.
+        assume_sorted: boolean, optional
+            If False, values of x can be in any order and they are sorted
```
shoyer: rather than "x" this should be "coordinates that are interpolated over"

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+        'cubic'}. For multidimensional interpolation, only
+        {'linear', 'nearest'} can be used.
+    kwargs:
+        keyword arguments to be passed to scipy.interpolate
+
+    Returns
+    -------
+    Interpolated Variable
+    """
+    if len(indexes_coords) == 0:
```
shoyer: just use "if not indexes_coords`

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+
+
+def _localize(obj, indexes_coords):
+    """ Speed up for linear and nearest neighbor method.
+    Only consider a subspace that is needed for the interpolation
+    """
+    for dim, [x, new_x] in indexes_coords.items():
+        try:
+            imin = x.to_index().get_loc(np.min(new_x.values), method='nearest')
+            imax = x.to_index().get_loc(np.max(new_x.values), method='nearest')
```
shoyer: it's better to reuse a single index rather than calling `to_index()` twice.

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+
+    # default behavior
+    kwargs['bounds_error'] = kwargs.get('bounds_error', False)
+    # as coords should be always sorted, neglect this option
+    kwargs.pop('assume_sorted', None)
+
+    # target dimensions
+    dims = list(indexes_coords)
+    x = [indexes_coords[d][0] for d in dims]
+    new_x = [indexes_coords[d][1] for d in dims]
```
shoyer: `zip(*...)` is a nice way to do this unwrapping:
```
x, new_x = zip(*[indexes_coords[d] for d in dims])
```

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+    destination = broadcast_variables(*new_x)
+
+    rslt = apply_ufunc(interp_func, obj,
```
shoyer: Prefer spelling out the full variable name `result` rather than abbreviating it. Omitting vowels is generally avoided in Python except for a few standard abbreviations like `kwargs` and `cls`.

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+                       input_core_dims=[dims],
+                       output_core_dims=[destination[0].dims],
+                       output_dtypes=[obj.dtype], dask='allowed',
+                       kwargs={'x': x, 'new_x': destination, 'method': method,
+                               'kwargs': kwargs},
+                       keep_attrs=True)
+
+    if all(x1.dims == new_x1.dims for x1, new_x1 in zip(x, new_x)):
+        return rslt.transpose(*obj.dims)
```
shoyer: Rather than making this conditional on all the dimensions matching, can we either *always* or *never* reorder dimensions into the position of `dims`? I'm worried that this will make behavior hard to predict.

In particular, let's make sure that the dimension order of the result always matches that from `.isel()`.

xarray/core/missing.py
```diff
@@ -332,3 +369,158 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+
+    # target dimensions
+    dims = list(indexes_coords)
+    x = [indexes_coords[d][0] for d in dims]
+    new_x = [indexes_coords[d][1] for d in dims]
+    destination = broadcast_variables(*new_x)
+
+    rslt = apply_ufunc(interp_func, obj,
+                       input_core_dims=[dims],
+                       output_core_dims=[destination[0].dims],
```
shoyer: I'm a little worried that we shouldn't be reusing the same output dimensions as inputs with potentially different sizes in `apply_ufunc`. This certainly won't work once we fix https://github.com/pydata/xarray/issues/1931. So `output_core_dims` should probably be temporarily renamed, and then restored afterwards.

Alternatively, maybe it isn't worth using `apply_ufunc` here at all. If a function only needs to work on `xarray.Variable` objects, then it's just as easy to work with the variable directly, e.g.,
```python
new_data = interp_func(obj.data, x=x, new_x=destination, method=method, kwargs=kwargs)
new_dims = destination[0].dims
result = Variable(new_dims, new_data, obj.attrs)
```

rabernat: 👍 to operating on variables.

agoodm:
Very nice! I noticed that the interpolation is performed among dimensions rather than coordinates in this PR. However the limitation to that is interpolation to/from curvilinear grids is not supported, which I think is a common enough use case, and would be extremely nice to have. Pretty sure scipy's interpolation tools work out of the box with curvilinear grids. Is an updated interface which works on coordinate variables rather than dimensions planned?

shoyer:
> Very nice! I noticed that the interpolation is performed among dimensions rather than coordinates in this PR. However the limitation to that is interpolation to/from curvilinear grids is not supported, which I think is a common enough use case, and would be extremely nice to have. Pretty sure scipy's interpolation tools work out of the box with curvilinear grids. Is an updated interface which works on coordinate variables rather than dimensions planned?

Yes, our intent is to eventually extend this same interface to handle these cases, but not in this initial PR.

doc/interpolation.rst
```diff
@@ -0,0 +1,260 @@
...
+``(x, y) = ((0.5, 0.15), (1.5, 0.25), (2.5, 0.35))`` are obtained by the
+2-dimensional interpolation and mapped along a new dimension ``z``.
+
+If you want to add a coordinate to the new dimension ``z``, you can supply
+:py:class:`~xarray.DataArray` s with a coordinate,
+
+.. ipython:: python
+
+    x = xr.DataArray([0.5, 1.5, 2.5], dims='z', coords={'z': ['a', 'b','c']})
+    y = xr.DataArray([0.15, 0.25, 0.35], dims='z')
```
jhamman: for clarity, would it make sense to specify the z-coordinate for both x and y here?

xarray/core/missing.py
```diff
@@ -332,3 +369,162 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+    """
+    if not indexes_coords:
+        return obj.copy()
+
+    # simple speed up for the local interpolation
+    if method in ['linear', 'nearest']:
+        obj, indexes_coords = _localize(obj, indexes_coords)
+
+    if method not in ['linear', 'nearest', 'zero', 'slinear', 'quadratic',
+                      'cubic', 'polynomial']:
```
jhamman: I think you can replace this list with `interp1d_methods` (defined above).

xarray/core/dataarray.py
```diff
@@ -882,10 +882,46 @@ def reindex(self, method=None, tolerance=None, copy=True, **indexers):
...
+        ----
+        scipy is required.
+
+        See Also
+        --------
+        scipy.interpolate.interp1d
+        scipy.interpolate.interpn
+        """
+        ds = self._to_temp_dataset().interp(
+            method=method, kwargs=kwargs, **coords)
```
jhamman: should you be passing `assume_sorted` through here?

xarray/core/missing.py
```diff
@@ -332,3 +369,162 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+        raise ValueError('{} is not a valid method.'.format(method))
+
+    # default behavior
+    kwargs['bounds_error'] = kwargs.get('bounds_error', False)
+    # as coords should be always sorted, neglect this option
+    kwargs.pop('assume_sorted', None)
```
jhamman: I'm wondering if this check is actually needed. I'm not seeing where this kwarg is coming from.

xarray/core/missing.py
```diff
@@ -332,3 +369,162 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+        return obj
+
+    if isinstance(obj, dask_array_type):
+        import dask.array as da
+
+        _assert_single_chunk(obj, range(obj.ndim - len(x), obj.ndim))
+        chunks = obj.chunks[:-len(x)] + new_x[0].shape
+        drop_axis = range(obj.ndim - len(x), obj.ndim)
+        new_axis = range(obj.ndim - len(x), obj.ndim - len(x) + new_x[0].ndim)
+        # call this function recursively
```
jhamman: I think the recursive use of this function could use a better explanation/comment.

fujiisoup: I removed this recursive call.

xarray/core/missing.py
fujiisoup: Agreed. removed.

xarray/core/dataarray.py
fujiisoup: Good catch! Thanks.
fixed.

xarray/core/missing.py
fujiisoup: I removed this line and use the sanity check carried out above.

jhamman:
Is there more to do here or is this ready to go in? Seems to me its ready to go.

shoyer:
Sorry, I still need to find time to finish reviewing this — hopefully very
soon!

xarray/core/missing.py
```diff
@@ -332,3 +374,160 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+    original_dims = broadcast_dims + dims
+    new_dims = broadcast_dims + list(destination[0].dims)
+    interped = interp_func(obj.transpose(*original_dims).data,
+                           x, destination, method, kwargs)
+
+    result = Variable(new_dims, interped, attrs=obj.attrs)
+
+    # transpose the result so that the new dimensions are inserted to the
+    # original position
+    if all(x1.dims == new_x1.dims for x1, new_x1 in zip(x, new_x)):
```
shoyer: Do we need this special case? I don't think it would help much for performanc.

fujiisoup: This line is intended to make the dimension order consistent to `.isel`, not the efficiency.
I'm wondering if this implementation is the cleanest solution l, but the removal of this block makes the test failing.

xarray/core/missing.py
```diff
@@ -332,3 +374,160 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+        index = x.to_index()
+        imin = index.get_loc(np.min(new_x.values), method='nearest')
+        imax = index.get_loc(np.max(new_x.values), method='nearest')
+
+        indexes[dim] = slice(max(imin - 2, 0), imax + 2)
+        indexes_coords[dim] = (x[indexes[dim]], new_x)
+    return obj.isel(**indexes), indexes_coords
+
+
+def interp(obj, indexes_coords, method, **kwargs):
```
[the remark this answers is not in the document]
fujiisoup: done

xarray/core/missing.py
```diff
@@ -312,6 +330,8 @@ def _get_interpolator(method, **kwargs):
             interp_class = interpolate.PchipInterpolator
         elif method == 'spline':
             kwargs.update(method=method)
+            if 'fill_value' in kwargs:
+                del kwargs['fill_value']
```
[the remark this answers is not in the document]
fujiisoup: This was my mistake.
The current `interp` no longer supports 'spline' method interpolation.
Removed.

xarray/core/dataarray.py
```diff
@@ -895,10 +895,47 @@ def reindex(self, method=None, tolerance=None, copy=True, **indexers):
             method=method, tolerance=tolerance, copy=copy, **indexers)
         return self._from_temp_dataset(ds)
 
+    def interp(self, method='linear', kwargs={}, assume_sorted=False,
+               **coords):
```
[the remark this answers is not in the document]
fujiisoup: done

xarray/core/missing.py
```diff
@@ -332,3 +369,160 @@ def _get_valid_fill_mask(arr, dim, limit):
     return (arr.isnull().rolling(min_periods=1, **kw)
             .construct(new_dim, fill_value=False)
             .sum(new_dim, skipna=False)) <= limit
+
+
+def _assert_single_chunk(var, axes):
+    for axis in axes:
+        if len(var.chunks[axis]) > 1 or var.chunks[axis][0] < var.shape[axis]:
+            raise ValueError('Chunking along the dimension to be interpolated '
```
shoyer: please see my previous comment about making this NotImplementedError

xarray/core/missing.py
```diff
@@ -332,3 +369,160 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+    return var.isel(**indexes), indexes_coords
+
+
+def interp(var, indexes_coords, method, **kwargs):
+    """ Make an interpolation of Variable
+
+    Parameters
+    ----------
+    var: Variable
+    index_coord:
```
shoyer: name is inconsistent with the function argument

xarray/core/missing.py
```diff
@@ -332,3 +369,160 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+        mapping from dimension name to a pair of original and new coordinates.
+        Original coordinates should be sorted in strictly ascending order.
+    method: string
+        One of {'linear', 'nearest', 'zero', 'slinear', 'quadratic',
+        'cubic'}. For multidimensional interpolation, only
+        {'linear', 'nearest'} can be used.
+    kwargs:
```
shoyer: add `**`

xarray/core/missing.py
```diff
@@ -332,3 +369,160 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+
+
+def interp(var, indexes_coords, method, **kwargs):
+    """ Make an interpolation of Variable
+
+    Parameters
+    ----------
+    var: Variable
+    index_coord:
+        mapping from dimension name to a pair of original and new coordinates.
```
shoyer: can you note that these coordinates are Variable objects?

xarray/core/missing.py
```diff
@@ -332,3 +369,168 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+    result = Variable(new_dims, interped, attrs=var.attrs)
+
+    # transpose the result so that the new dimensions are inserted to the
+    # original position
+    if all(x1.dims == new_x1.dims for x1, new_x1 in zip(x, new_x)):
+        result_dims = var.dims
+    else:
+        start = min(var.get_axis_num(d) for d in dims)
+        result_dims = (broadcast_dims[:start] + list(destination[0].dims) +
+                       broadcast_dims[start:])
```
shoyer: If I recall correctly, `Variable.__getitem__` orders dimensions based on broadcasting:
- Unindexed dimensions are assumed to be indexed with `slice(None)`
- `slice()` is equivalent to `range(...)`
- Unlabeled arrays are equivalent to indexing along the matching dimension
- Dimensions on the broadcast result appear in order of appearance.

I'm not confident this version is correct in all edge cases, so I think it would be best to match this logic directly, e.g., by adding dimensions from all variables to an `OrderedSet` (from `xarray.utils`).

fujiisoup: OK. Done. Thanks for the suggestion.

xarray/core/dataset.py
```diff
@@ -1806,6 +1814,79 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
...
+
+        def maybe_variable(obj, k):
+            try:
+                return obj._variables[k]
+            except KeyError:
+                return as_variable((k, range(obj.dims[k])))
+
+        variables = OrderedDict()
+        for name, var in iteritems(obj._variables):
+            if name not in [k for k, v in indexers_list]:
```
shoyer: Given that you already need `indexers_list` as a dict down below, why not do so before this loop so you can use `if name not in indexers_dict` instead of the comprehension? (The comprehension inside the `if` clause looks ugly to me.)

xarray/core/dataset.py
```diff
@@ -1806,6 +1814,79 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
...
+                var_indexers = {k: (maybe_variable(obj, k), v) for k, v
+                                in indexers_list if k in var.dims}
+                variables[name] = missing.interp(
+                    var, var_indexers, method, **kwargs)
+
+        coord_names = set(variables).intersection(obj._coord_names)
+        selected = obj._replace_vars_and_dims(variables,
+                                              coord_names=coord_names)
+        # attach indexer as coordinate
+        variables.update({k: v for k, v in indexers_list})
```
shoyer: Note that `dict(indexers_list)` is equivalent to `{k: v for k, v in indexers_list}`.

If you call it inside update, you don't even need to call `dict()` on it.

shoyer:
This is very elegant and I think very close to being ready to merge.

There are two design issues that I would like to discuss first:
1. Should we interpolate both `data_vars` and `coords` or only `data_vars`?
2. How should we handle non-numeric variables?

The current behavior of this PR is:
1. Interpolate all variables, both `data_vars` and `coords`.
2. Non-numeric variables raise an error, because they can't be converted into float.

It might make sense to copy the behavior from `Dataset.reduce`:
1. Only interpolate `data_vars`.
2. Drop non-numeric variables instead of trying to interpolate them.

shoyer:
My opinions:
1. I suppose it probably does make sense to interpolate numeric coordinates, too. So this probably doesn't need to change.
2. I do think we should drop non-numeric variables. There just isn't any way to meaningfully interpolate them (especially strings). This means we will probably need to add a check in `DataArray.interp` to ensure that the variable dtype is numeric and otherwise raise an error.

shoyer:
I would wait before trying to do interpolation with datetime types. There might be use cases for this but I can't think of them offhand. `resample()` would seem to cover most of these.

On the other hand, it could indeed be interesting to extend `resample()` to handling numbers. I proposed this a while ago for pandas (https://github.com/pandas-dev/pandas/issues/12828).

fujiisoup:
Done!

But, I am slightly worrying if automatic dropping non-numeric array is not surprising.
Do we have other methods that behave similarly?

Or maybe we can raise a warning?

shoyer:
> Do we have other methods that behave similarly?

We do the same thing for aggregations like `mean()`, behavior originally copied from pandas:
http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.mean.html

We could add a `numeric_only=True` argument (like we do for reduction methods) to make this behavior more obvious and allow it to be controlled.

fujiisoup:
Thanks, @shoyer 

> We do the same thing for aggregations like mean(), behavior originally copied from pandas:
http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.mean.html

Thanks. I have not been surprised when `mean` drops the object type array. So it would be OK also for `interp`.

Maybe this PR is ready, then.

BTW, I noticed that our `mean()` does not accept the `numeric_only` keyword argument.
Our docs do not say about `numeric_only` keyword either.
Is it intended? 
(This behavior looks already accepted, and I think it is OK if we do not support this.)

shoyer:
Hmm. I guess numeric_only may currently only be an argument for the reduce method. We probably should add it for methods like mean, too.

xarray/core/dataset.py
```diff
@@ -1815,24 +1815,29 @@ def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
         return self._replace_vars_and_dims(variables, coord_names)
 
     def interp(self, coords=None, method='linear', assume_sorted=False,
-               kwargs={}, **coords_kwargs):
+               keep_attrs=False, kwargs={}, **coords_kwargs):
```
shoyer: My inclination would be to have `keep_attrs=True` by default, and maybe not even bother to include the argument at all. This would be consistent with `reindex`/`sel`/`isel`.

shoyer:
Do you want to add `interp_like` here or should we leave it for a later PR?

xarray/core/dataset.py
fujiisoup: OK. Then, I would remove this argument.

fujiisoup:
`interp_like` sounds a good idea. I like it in another PR as this is already huge.

xarray/core/missing.py
```diff
@@ -332,3 +369,170 @@ def _get_valid_fill_mask(arr, dim, limit):
...
             .construct(new_dim, fill_value=False)
             .sum(new_dim, skipna=False)) <= limit
+
+
+def _assert_single_chunk(var, axes):
+    for axis in axes:
+        if len(var.chunks[axis]) > 1 or var.chunks[axis][0] < var.shape[axis]:
+            raise NotImplementedError(
+                'Chunking along the dimension to be interpolated '
+                '({}) is not supported.'.format(axis))
```
shoyer: "not supported" -> "not yet supported"

xarray/core/missing.py
```diff
@@ -332,3 +369,170 @@ def _get_valid_fill_mask(arr, dim, limit):
...
+    Note
+    ----
+    This requiers scipy installed.
+
+    See Also
+    --------
+    scipy.interpolate.interp1d
+    """
+    if not x:
+        return var
```
shoyer: For consistency on views/copies, it's better to return `var.copy()` here

xarray/core/missing.py
```diff
@@ -320,7 +335,29 @@ def _get_interpolator(method, **kwargs):
...
+        from scipy import interpolate
+    except ImportError:
+        raise ImportError(
+            'Interpolation with method `%s` requires scipy' % method)
+
+    if method in valid_methods:
+        kwargs.update(method=method)
+        interp_class = interpolate.interpn
+    else:
+        raise ValueError('%s is not a valid interpolator' % method)
```
shoyer: add: "for interpolating over multiple dimensions"

fujiisoup:
Thanks, @shoyer.

> I think the last thing to do here is remove keep_attrs?

Done!
(I thought I reverted this change already...)

fujiisoup:
Merged!

Thanks, @shoyer, @fmaussion, @jhamman, for the detailed review.

rabernat:
Congratulations on this great new feature!

yohai:
@fujiisoup Great feature! thanks.
There's a typo ('rondomloy') in the documentation: https://github.com/pydata/xarray/blame/master/doc/interpolation.rst#L192

Issue #2085, opened 2018-04-26
units = 'days' leads to timedelta64 for data variable


mathause:
#### Code Sample

```python

import numpy as np
import xarray as xr

def test_units(units):

    x = np.arange(10)
    data = np.random.randn(10)

    ds = xr.Dataset(data_vars=dict(data=('x', data)), coords=dict(x=('x', x)))
    ds.data.attrs['units'] = units

    ds.to_netcdf('tst.nc')

    decoded = xr.open_dataset('tst.nc')

    print(units.ljust(8), decoded.data.dtype)


    ds.close()
    decoded.close()

test_units('seconds')
test_units('second')

test_units('minutes')
test_units('minute')

test_units('days')
test_units('day')

test_units('months')
test_units('years')
```
#### Problem description

Returns:
```
seconds  timedelta64[ns]
second   float64
minutes  timedelta64[ns]
minute   float64
days     timedelta64[ns]
day      float64
months   float64
years    float64
```
#### Expected Output

I would expect type float for all of them. Or is this expected behaviour?

I have a dataset that reports 'consecutive dry days' and the dataset creator correctly set the units of the data to 'days', but I don't want this to be decoded (but the time axis should)....

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.126-48-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8

xarray: 0.10.2+dev6.g9261601
pandas: 0.22.0
numpy: 1.14.2
scipy: 1.0.1
netCDF4: 1.3.1
h5netcdf: 0.5.0
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: 1.0.0
dask: 0.17.2
distributed: 1.21.3
matplotlib: 2.2.2
cartopy: None
seaborn: None
setuptools: 39.0.1
pip: 9.0.2
conda: None
pytest: 3.4.2
IPython: 6.2.1
sphinx: None
</details>

shoyer:
This is expected behavior, but you're not the first person to complain, and it seems like there is a consensus around changing it: https://github.com/pydata/xarray/issues/1621

I'm going to close this issue in favor of the other one.

ocefpaf:
[...]

PR #2105, opened 2018-05-04
Deprecate decode timedelta

### LINKED ISSUE #1621: Undesired decoding to timedelta64 (was: units of "seconds" translated to time coordinate)
When using open_dataset( ), it is translating data variables with units of "seconds" to time coordinates. For example, measurements of wave period. I don't believe xarray should treat variables as time coordinates unless their units are of "seconds since...". I have noticed that changing my units to "second" or "sec" or "s" prevents xarray from translating the variable to datetime64 and keeps it float64, as desired. More details and an OPeNDAP example posted on github here: https://stackoverflow.com/questions/46552078/xarray-wave-period-in-seconds-ingested-as-timedelta64

ocefpaf:
- [X] Closes #1621 (remove if there is no corresponding issue, which should only be the case for minor changes)
 - [X] Tests added (for all bug fixes or enhancements)
 - [X] Tests passed (for all non-documentation changes)
 - [X] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

I'll add tests, docs, and the whats-new entry later if I'm on the right path here.

xref.: #843, #940, and #2085

See http://nbviewer.jupyter.org/gist/ocefpaf/e736c07faf3d4c9361ecf546a692c2cd

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -335,6 +335,10 @@ like ``'days'`` for ``timedelta64`` data. ``calendar`` should be one of the cale
 supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'noleap',
 '365_day', '360_day', 'julian', 'all_leap', '366_day'.
 
+The automatic decoding of untis like ``'days'`` to ``timedelta64`` is deprecated and
+will be removed in xarray version 0.11. The default behavior will keep the same numeric
+type of the original data instead of converting to ``timedelta64``.
+
 By default, xarray uses the 'proleptic_gregorian' calendar and units of the smallest time
 difference between values, with a reference time of the first time value.
 
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -41,8 +41,8 @@ converted automatically when used as arguments in xarray objects:
     import datetime
     xr.Dataset({'time': datetime.datetime(2000, 1, 1)})
 
-When reading or writing netCDF files, xarray automatically decodes datetime and
-timedelta arrays using `CF conventions`_ (that is, by using a ``units``
+When reading or writing netCDF files, xarray automatically decodes datetime
+arrays using `CF conventions`_ (that is, by using a ``units``
 attribute like ``'days since 2000-01-01'``).
 
 .. _CF conventions: http://cfconventions.org
--- a/xarray/coding/times.py
+++ b/xarray/coding/times.py
@@ -180,6 +180,9 @@ def decode_cf_timedelta(num_timedeltas, units):
     """Given an array of numeric timedeltas in netCDF format, convert it into a
     numpy timedelta64[ns] array.
     """
+    warnings.warn('Decoding timedeltas been deprecated and '
+                  'will be removed in xarray v??.',
+                  FutureWarning, stacklevel=2)
     num_timedeltas = np.asarray(num_timedeltas)
     units = _netcdf_to_numpy_timeunit(units)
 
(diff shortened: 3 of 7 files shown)


xarray/coding/times.py
```diff
@@ -180,6 +180,9 @@ def decode_cf_timedelta(num_timedeltas, units):
     """Given an array of numeric timedeltas in netCDF format, convert it into a
     numpy timedelta64[ns] array.
     """
+    warnings.warn('Decoding timedeltas been deprecated and '
```
ocefpaf: I'm not sure if you want to deprecate `decode_cf_timedelta` too or if you want to keep that around in case anyone wants to manually convert data to timedeltas. I don't have a user case for that but it was the `default` behavior, so maybe someone uses this.

xarray/coding/times.py
```diff
@@ -362,10 +365,12 @@ def encode(self, variable, name=None):
             data, units = encode_cf_timedelta(
                 data, encoding.pop('units', None))
             safe_setitem(attrs, 'units', units, name=name)
-
         return Variable(dims, data, attrs, encoding)
 
     def decode(self, variable, name=None):
+        warnings.warn('Decoding timedeltas been deprecated and '
+                      'will be removed in xarray v??.',
```
ocefpaf: What version of xarray would be OK to make this default?

xarray/conventions.py
```diff
@@ -293,10 +294,15 @@ def decode_cf_variable(name, var, concat_characters=True, mask_and_scale=True,
                       variables.CFScaleOffsetCoder()]:
             var = coder.decode(var, name=name)
 
+    enable_future_time_unit_decoding = OPTIONS['enable_future_time_unit_decoding']
```
ocefpaf: An suggestions for a shorter name here?

xarray/coding/times.py
ocefpaf: Makes sense.

xarray/conventions.py
ocefpaf: I wasn't really worried about the linter but rather with the long name :smile:
But naming things **is** hard!

doc/io.rst
```diff
@@ -335,6 +335,10 @@ like ``'days'`` for ``timedelta64`` data. ``calendar`` should be one of the cale
 supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'noleap',
 '365_day', '360_day', 'julian', 'all_leap', '366_day'.
 
+The automatic decoding of untis like ``'days'`` to ``timedelta64`` is deprecated and
+will be removed in xarray version 0.11. The default behavior will keep the same numeric
+type of the original data instead of converting to ``timedelta64``.
```
ocefpaf: @shoyer b/c english is not my mother tongue I really appreciate your help wording this part.

doc/time-series.rst
```diff
@@ -41,8 +41,8 @@ converted automatically when used as arguments in xarray objects:
     import datetime
     xr.Dataset({'time': datetime.datetime(2000, 1, 1)})
 
-When reading or writing netCDF files, xarray automatically decodes datetime and
-timedelta arrays using `CF conventions`_ (that is, by using a ``units``
+When reading or writing netCDF files, xarray automatically decodes datetime
+arrays using `CF conventions`_ (that is, by using a ``units``
```
ocefpaf: I think that removing the mention of timedelta here is better b/c the netCDF original data is rarely a timedelta. The case of wave periods, or any other phenomena period, isn't a timedelta IMO.

ocefpaf:
@shoyer this is ready for a second round of reviews.

> The last thing I'd like to see is support for automatically serializing/deserializing timedelta64 by saving an attribute dtype='timedelta64[ns]'. This will preserve the ability to roundtrip this data to netCDF, which I do think is useful for some users -- otherwise users will have xarray objects that they can no longer save to netCDF.

I'm not sure I follow. Roundtrip is easier now since the original time unit dtype is preserved, no? Ping @rsignell-usgs who is a netCDF/CF specialist :wink:

ocefpaf:
> I would like to be able to round-trip something like the following dataset to netCDF.

OK, I though you meant roundtrip from the netCDF file and back. In my line of work handling high level python objects serialization like that is usually not desired as the user should be responsible of how s/he wants to save the data. (I also never had an application that required timedelta, so it was hard for me to contextualize that.)

I'll be away the next weeks, so if someone wants to pick it up from here please feel free to do so. Otherwise I'll try to get back to this once I return to the office.

PR #2089, opened 2018-04-27
xarray.dot is now based on da.einsum

crusaderky:
Closes #2074 
Requires dask >= 0.17.3

PR #2106, opened 2018-05-04
xarray.dot to pass **kwargs to einsum

crusaderky:
Late addition to #2089

### DIFF
--- a/xarray/tests/test_computation.py
+++ b/xarray/tests/test_computation.py
@@ -844,6 +844,9 @@ def test_dot(use_dask):
     assert (actual.data == np.zeros(actual.shape)).all()
 
     # Invalid cases
+    if not use_dask or LooseVersion(dask.__version__) > LooseVersion('0.17.4'):
+        with pytest.raises(TypeError):
+            xr.dot(da_a, dims='a', invalid=None)
     with pytest.raises(TypeError):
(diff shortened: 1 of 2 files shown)


crusaderky:
@shoyer sorry, late addition.
Note that currently split_every is ignored by dask (https://github.com/dask/dask/issues/3471)

xarray/tests/test_computation.py
```diff
@@ -835,13 +836,24 @@ def test_dot(use_dask):
...
-        xr.dot(da_a, dims='a', invalid=None)
+    # Invalid cases
     with pytest.raises(TypeError):
         xr.dot(da_a.to_dataset(name='da'), dims='a')
     with pytest.raises(TypeError):
         xr.dot(dims='a')
 
+    # einsum parameters
+    xr.dot(da_a, da_b, dims=['b'], order='F')
+    actual = xr.dot(da_a, da_b, dims=['b'])
```
shoyer: I'm a little confused here -- the result of the first line is thrown away rather than checked, and the second line doesn't use parameters at all?

crusaderky: @shoyer Sorry, you're right it did not make much sense. Fixed.

crusaderky:
@shoyer ready for merge if there are no further objections

shoyer:
thanks!

PR #2107, opened 2018-05-06
ENH: added FacetGrid functionality to line plots

### LINKED ISSUE #2084: New Feature: Add FacetGrid functionality to plot.line()
When data is more than 2 dimensional, `plot()` creates a `FacetGrid` where each facet gets plots two dimensional data. I would find it very useful to have similar functionality for line plots. 

Let's define three `DataArray`s, of 2,3 and 4 dimensions, named `d2`, `d3` and `d4`.

```python
d4=xr.DataArray(np.random.randn(3,4,5,6),
                dims=['x','y','z','w'],
                coords=[range(3),range(4),range(5),range(6)])
d3=d4[...,0].drop('w')
d2=d3[...,0].drop('z')

d4.plot(row='x', col='y')   #creates a 2d facet grid, each facet is a pcolormesh
d3.plot(col='x')           #creates a 1d facet grid, each facet is a pcolormesh
plt.show()
```

It would be very useful to have a similar interface for `line` plots too. That is, I would expect
```python
d2.plot.line(row='y', x='x')
```
to create a row of panels, each having one line. Similarly, I would expect
```python
d3.plot.line(col='y', x='x')
#or alternatively
d3.plot.line(col='y', hue='z')
```
to create a row of panels, each having a series of lines, with `hue` corresponding to `z` and so forth.  Finally, 
```python
d4.plot.line(row='z',col='y', x='x')
#or alternatively
d4.plot.line(row='z',col='y', hue='w')
```
should create a 2d matrix of panels, each having a series of lines, with `hue` corresponding to `w`.

This seems to me like a natural generalization of the behaviour of `plot()`, and I found myself coding manually loops to do that a few times.

yohai:
closes #2084
    a) plot.line can now accept also 'row' and 'col' keywords.
    b) If 'hue' is passed as a keyword to DataArray.plot() it generates
       a line plot FacetGrid.
    c) Line plots are automatically generated if the number of dimensions
       after faceting along row and/or col is one.

 - [x] Closes #xxxx (remove if there is no corresponding issue, which should only be the case for minor changes)


yohai:
Example code to demonstrate functionality:

```
import xarray as xr
import numpy as np
import pandas as pd
np.random.seed(0)

#set up data:
tm=pd.date_range('2000-01-01',end='2000-01-02',freq='H')
d4=xr.DataArray(np.random.randn(len(tm),4,3,3),
                dims=['time', 'y', 'z', 'w'],
                coords=[tm,['foo', 'bar', 'foobar', 'barfoo'], range(3), range(3)],
                name='foo')
d3 = d4[..., 0].drop('w')
d2 = d3[..., 0].drop('z')

#generate a faceted 4-dimensional plot:
d4.plot(row='w', col='y', hue='z')
#or equivalently:
d4.plot.line(row='w', col='y', hue='z')
#or equivalently:
d4.plot.line(row='w', col='y', x='time')

#also works for plotting along y axis:
d3.plot.line(col='y', y='time')
```

shoyer:
I'm really excited about this!

Could you kindly paste in examples of what the generated plots look like in your comment? (GitHub will upload an image when you paste it in.)

yohai:
was simpler than I thought. @shoyer I added the plots.

jhamman:
@yohai - this looks great. Can we add some basic unit tests covering this functionality?

xarray/plot/facetgrid.py
```diff
@@ -266,6 +265,46 @@ def map_dataarray(self, func, x, y, **kwargs):
...
+            darray=self.data.loc[self.name_dicts.flat[0]],
+            x=x, y=y, hue=hue)
+
+        self._finalize_grid(xlabel, ylabel)
+
+        if add_legend and huelabel:
+            self.fig.legend(handles=self._mappables[-1],
+                            labels=list(self.data.coords[huelabel].values),
+                            title=huelabel,
+                            loc="center right")
```
shoyer: It would be great to draw the legend outside the plots, similar to what Seaborn does by default:
image

I think the resulting plots look a little cleaner -- otherwise the legend can overlap with the data.

We don't need to copy the full logic from Seaborn (it's pretty complicated!), but its `add_legend()` method could be a useful reference:
https://github.com/mwaskom/seaborn/blob/102cf08f9e40ef7037daf4b3b60b94ee144dc5cb/seaborn/axisgrid.py#L40

This file is already derived from Seaborn, so feel free to copy anything that is relevant.

yohai:
@jhamman @shoyer Sorry for being a noob, but what exactly do you mean by "add some basic unit tests"? you mean writing new functions that will verify that stuff works as it should, or that the new addition does not mess up previous plotting behavior?

fmaussion:
> writing new functions that will verify that stuff works as it should

Yes. See [here](https://github.com/pydata/xarray/blob/master/xarray/tests/test_plot.py#L1223) for example tests for FacetGrid. Testing plotting functions is not trivial but we usually test only very
basic functionality. You might find some examples in here.

> or that the new addition does not mess up previous plotting behavior?

This is usually covered by continuous integration, which already ran the tests on your current PR
(the "all checks have passed" green button above). So on this side you are safe for now, unless
our previous tests weren't efficient enough ;-)

xarray/plot/facetgrid.py
```diff
@@ -276,6 +313,42 @@ def _finalize_grid(self, *axlabels):
...
+            title=huelabel,
+            loc="center right")
+
+        # self._legend = figlegend
+        # figlegend.set_title(title)
+
+        # Set the title size a roundabout way to maintain
+        # compatability with matplotlib 1.1
+        # prop = matplotlib.font_manager.FontProperties(size=title_size)
+        # figlegend._legend_title_box._text.set_font_properties(prop)
```
shoyer: It's probably best to delete this commented out code

xarray/plot/plot.py
```diff
@@ -123,26 +149,28 @@ def plot(darray, row=None, col=None, col_wrap=None, ax=None, rtol=0.01,
     plot_dims = set(darray.dims)
     plot_dims.discard(row)
     plot_dims.discard(col)
+    plot_dims.discard(hue)
 
     ndims = len(plot_dims)
 
     error_msg = ('Only 2d plots are supported for facets in xarray. '
```
shoyer: This should be updated, to something like "1d or 2 plots"

xarray/tests/test_plot.py
```diff
@@ -1456,6 +1456,50 @@ def test_default_labels(self):
...
+        bottomright = g.axes[-1, -1]
+        assert not bottomright.has_data()
+        assert not bottomright.get_visible()
+
+    def test_set_axis_labels(self):
+        g = self.darray.plot(row='row', col='col', hue='hue')
+        g.set_axis_labels('longitude', 'latitude')
+        alltxt = text_in_fig()
+        for label in ['longitude', 'latitude']:
+            assert label in alltxt
```
shoyer: nit: for unit tests, we prefer to write things out in long form when practical rather than writing loops. This makes things easier to debug, because test runners usually only report the line on which the error occurs, e.g.,
```
assert 'longitude' in alltxt
assert 'latitude' in alltxt
```

xarray/plot/plot.py
```diff
@@ -151,6 +179,55 @@ def plot(darray, row=None, col=None, col_wrap=None, ax=None, rtol=0.01,
...
+
+    if x is not None and x not in darray.dims:
+        raise ValueError('x ' + error_msg)
+
+    if y is not None and y not in darray.dims:
+        raise ValueError('y ' + error_msg)
+
+    if x is not None and y is not None:
+        raise ValueError('You cannot specify both x and y kwargs'
+                         'for line plots.')
```
shoyer: It would be good to add tests that trigger each of these error conditions, e.g.,
```
    ...  # do something wrong
```

yohai:
Thanks @shoyer. I fixed everything except for the coverall part, which is weird. The lines it marks as untested are the function `_line_facetgrid`, which the main function that performs all of the testing. I suspect that it simply doesn't run my new tests for some reason. When I run `pytest` locally these lines do get tested (I verified by adding `print` commands there).
Should I add decorators like `pytest.mark.something`for coverall to notice these tests?

xarray/plot/plot.py
```diff
@@ -204,64 +289,22 @@ def line(darray, *args, **kwargs):
...
-            xplt = darray.coords[xlabel]
-            yplt = darray.transpose(xlabel, huelabel)
-
-        else:
-            ylabel, huelabel = _infer_xy_labels(darray=darray, x=y, y=hue)
-            xlabel = darray.name
-            xplt = darray.transpose(ylabel, huelabel)
-            yplt = darray.coords[ylabel]
-
+    xplt, yplt, xlabel, ylabel, huelabel = _infer_line_data(darray, x, y, hue)
```
dcherian: I like this refactor!

xarray/plot/plot.py
```diff
@@ -151,6 +179,55 @@ def plot(darray, row=None, col=None, col_wrap=None, ax=None, rtol=0.01,
...
+        raise ValueError('x ' + error_msg)
+
+    if y is not None and y not in darray.dims:
+        raise ValueError('y ' + error_msg)
+
+    if x is not None and y is not None:
+        raise ValueError('You cannot specify both x and y kwargs'
+                         'for line plots.')
+
+    if ndims == 1:
```
dcherian: This will need to be changed to follow #2151. One of us should rebase on the other's branch,. How do you think we should proceed?

yohai: The question is whether it's likely that there will more commits in #2151 or not. If not I can rebase and continue from there. If yes, it's probably better the other way around.

(As said above, it's my first contribution to open source software stuff so I don't really know how this works and what would be simpler)

shoyer:
> Thanks @shoyer. I fixed everything except for the coverall part, which is weird. The lines it marks as untested are the function _line_facetgrid, which the main function that performs all of the testing. I suspect that it simply doesn't run my new tests for some reason. When I run pytest locally these lines do get tested (I verified by adding print commands there).
Should I add decorators like pytest.mark.somethingfor coverall to notice these tests?

The specific problematic lines are those marked red in coveralls, e.g.,
https://coveralls.io/builds/17253745/source?filename=xarray%2Fplot%2Fplot.py#L101
https://coveralls.io/builds/17253745/source?filename=xarray%2Fplot%2Fplot.py#L176
https://coveralls.io/builds/17253745/source?filename=xarray%2Fplot%2Fplot.py#L291

These are all cases where an error is raised.

xarray/plot/plot.py
shoyer: I ended up merging #2151 first, so you should you merge the latest updates into this PR, e.g.,
```
git checkout master
git pull upstream master
git checkout line_FacetGrid
git merge master
```

yohai: OK, I'm pretty sure I messed this up but I might have fixed it too. All this rebasing is new to me. Sorry. Was it normal that after I did all that I got a list of many many modified files that I had to commit?

yohai:
OK, somehow the rebasing thing broke everything and I am at a loss. @shoyer, @dcherian please guide me on how to proceed. Sorry.

shoyer:
It looks like the merge went OK, but you have a couple of tests that are failing. Click on the "Travis-CI" link above next to the red "X" to see details:
https://travis-ci.org/pydata/xarray/jobs/386973465

xarray/plot/plot.py
```diff
@@ -218,66 +304,23 @@ def line(darray, *args, **kwargs):
...
-    if ylabel is not None:
-        ax.set_ylabel(ylabel)
+        if ylabel is not None:
+            ax.set_ylabel(ylabel)
 
-    ax.set_title(darray._title_for_slice())
+        ax.set_title(darray._title_for_slice())
 
     if darray.ndim == 2 and add_legend:
         ax.legend(handles=primitive,
```
shoyer: `huecoords` on the line below this one is no longer defined -- that's why your tests are failing

dcherian: Ya, the merge is a little screwed. Here's the commit you need to fix it: https://github.com/dcherian/xarray/commit/3e0429cb3b43db8e74ece681e6b1d2f5f2b638d9. 

@yohai I can't push to your branch so you need to make those changes somehow. Here's a gist with a patch file. Download and apply it using `git apply` (https://www.devroom.io/2009/10/26/how-to-create-and-apply-a-patch-with-git/)

xarray/plot/plot.py
```diff
@@ -83,8 +83,32 @@ def _easy_facetgrid(darray, plotfunc, x, y, row=None, col=None,
...
+                    size=None, subplot_kws=None, **kwargs):
+    """
+    Convenience method to call xarray.plot.FacetGrid for line plots
+    kwargs are the arguments to pyplot.plot()
+    """
+    ax = kwargs.pop('ax', None)
+    figsize = kwargs.pop('figsize', None)
+    if ax is not None:
+        raise ValueError("Can't use axes when making faceted plots.")
+    if aspect is None:
```
dcherian: Use `aspect=1`, `size=3` as default arguments in the `def _line_facetgrid` line?

shoyer: It's intentional that you can't pass `figsize` to facetted plots, because it's ambiguous whether you are referring to the size of each panel or the size of the entire plot. The later is what `figsize` means in matplotlib.

yohai: If I change the signature of `_easy_factgrid` to have `aspect=1` and `size=3`  and remove the lines 

````python
if aspect is None:
    aspect = 1
if size is None:
    size = 3
````
then some tests fail, presumably because some functions explicitly pass `None` as arguments. I suggest that for consistency we leave the signature of `_line_facetgrid` similar to `_easy_facetgrid`, but if you think it's not important then I can change it.

dcherian: Best to leave it as it is then.

shoyer:
thanks @yohai!

omergra:
This is an amazing addition.
Thanks @yohai

PR #2110, opened 2018-05-09
DOC: Update link to documentation of Rasterio

hgrif:
The documentation of Rasterio used to be at https://mapbox.github.io/rasterio/, but now you'll get a 404 and a pointer to https://rasterio.readthedocs.io/en/latest/. I've updated the links in `io.rst` and `whats-new.rst`.

Don't think I need to check the boxes below. :)


jhamman:
Thanks @hgrif!

PR #2115, opened 2018-05-10
Fix docstring formatting for load().

dopplershift:
Need '::' to introduce a code literal block. This was causing MetPy's
doc build to warn (since we inherit AbstractDataStore).


shoyer:
thanks!

PR #2117, opened 2018-05-11
Remove 'out' from docstrings for wrapped numpy unary functions

### LINKED ISSUE #1997: can't do in-place clip() with DataArrays.
Where `foo` is a DataArray, there doesn't seem to be a nice way to use `clip()` in-place.
```python
foo.clip(0, None, out=foo)
    foo.clip(0, None, out=foo)
  File "/rd22/scratch/broot/Programs/xarray/xarray/core/dataarray.py", line 1726, in func
    **kwargs))
  File "/rd22/scratch/broot/Programs/xarray/xarray/core/ops.py", line 205, in func
    return _call_possibly_missing_method(self, name, args, kwargs)
  File "/rd22/scratch/broot/Programs/xarray/xarray/core/ops.py", line 192, in _call_possibly_missing_method
    return method(*args, **kwargs)
TypeError: output must be an array
```
You get a similar exception if you do `np.clip(foo, ..., out=foo)`.

#### Problem description
Note the docstring for `DataArray.clip()`:
```
Help on method clip in module xarray.core.ops:

clip(self, *args, **kwargs) method of xarray.core.dataarray.DataArray instance
    a.clip(min=None, max=None, out=None)
    
    Return an array whose values are limited to ``[min, max]``.
    One of max or min must be given.
    
    Refer to `numpy.clip` for full documentation.
    
    See Also
    --------
    numpy.clip : equivalent function
```
So, the docstring advertises support for `out`.

dcherian:
- [x] Closes #1997

Inserting the string "The out keyword argument is not supported" is hacky. Can we do without that and just remove `out=None` from the signature?

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -199,8 +199,22 @@ def _values_method_wrapper(name):
 def _method_wrapper(name):
     def func(self, *args, **kwargs):
         return _call_possibly_missing_method(self, name, args, kwargs)
+
+    def format_docstring(doc):
+        '''Unlike numpy we don't support the out kwarg.
+           Remove it from the docstring'''
+
+        if doc.find('out=None') != -1:
+            idx = doc.find('for full documentation.')
+            doc = ((doc[:idx] +
+                    'for full documentation.\n\n' +
+                    '    The out keyword argument is not supported.' +
+                    doc[idx + 23:]).replace(', out=None', ''))
+
+        return doc
+
     func.__name__ = name
-    func.__doc__ = getattr(np.ndarray, name).__doc__
+    func.__doc__ = format_docstring(getattr(np.ndarray, name).__doc__)
     return func
 
 


fmaussion:
Thanks!

I wonder is this is the best way to do this but I don't have a better idea either. Surely @shoyer can weight in.

Anyways, this doesn't pass through the tests because sometimes ``__doc__`` is ``None``. Also I think we need some kind of test for this.

dcherian:
Ugh, serves me right for submitting PRs late at night. Thanks, @fmaussion. I'll wait for feedback on the approach before fixing it.

shoyer:
The cleaner solution here would be to try to get rid of `_method_wrapper` by writing explicit methods/functions implemented with `apply_ufunc`:
https://github.com/pydata/xarray/blob/a52540505f606bd7619536d82d43f19f2cbe58b5/xarray/core/ops.py#L34-L40

This is the strategy suggested in this TODO note:
https://github.com/pydata/xarray/blob/a52540505f606bd7619536d82d43f19f2cbe58b5/xarray/core/ops.py#L3-L5

This is a bit more verbose but it would be much more obvious what's going on.

`clip` would probably be the place to start, since it's the only one with an invalid `out` argument. You could just copy the relevant parts of the docstring from NumPy.

dcherian:
OK It'll take me some time to get to that. Closing.

PR #2120, opened 2018-05-11
Prevent Inf from screwing colorbar scale.

dcherian:

The current version uses `pd.isnull` to remove invalid values from input data when making a colorbar. `pd.isnull([np.inf])` is False which means `_determine_cmap_params` returns Inf for colorbar limits which screws everything up. This PR changes `pd.isnull` to `np.isfinite`.

### DIFF
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -160,7 +160,7 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
     """
     import matplotlib as mpl
 
-    calc_data = np.ravel(plot_data[~pd.isnull(plot_data)])
+    calc_data = np.ravel(plot_data[np.isfinite(plot_data)])
 
     # Handle all-NaN input data gracefully
     if calc_data.size == 0:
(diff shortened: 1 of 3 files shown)


xarray/plot/utils.py
```diff
@@ -160,7 +160,7 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
     """
     import matplotlib as mpl
 
-    calc_data = np.ravel(plot_data[~pd.isnull(plot_data)])
+    calc_data = np.ravel(plot_data[np.isfinite(plot_data)])
```
jhamman: I think [`pd.notnull`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.notnull.html#pandas.notnull) should work here?

dcherian: `pd.notnull(np.inf)` is True but we want it to be False. `calc_data` needs to contain all data values that are not inf/nan and `np.isfinite` does that.

shoyer:
This looks good to me, thanks.

Data with a color-bar can't be non-numeric, so we don't need `pd.isnull()`.

PR #2124, opened 2018-05-13
Raise an Error if a coordinate with wrong size is assigned to a dataarray

### LINKED ISSUE #2112: Sanity check when assigning a coordinate to DataArray

I think we can raise an Error if the newly assigned coordinate to a DataArray has an invalid shape.

```python
In [1]: import xarray as xr
   ...: 
   ...: da = xr.DataArray([0, 1, 2], dims='x')
   ...: da['x'] = [0, 1, 2, 3]  # no error
   ...: da
   ...: 
Out[1]: 
<xarray.DataArray (x: 3)>
array([0, 1, 2])
Coordinates:
  * x        (x) int64 0 1 2 3
```

#### Problem description

It is more user-friendly if we make some sanity checks when a new coordinate is assigned to a xr.DataArray.

Dataset raises an appropriate error,
```python
In [2]: ds = xr.Dataset({'da': ('x', [0, 1, 2])})
   ...: ds['x'] = [0, 1, 2, 3]  # -> raises ValueError
```

#### Expected Output
ValueError

fujiisoup:
- [x] Closes #2112
 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API.

Now uses `dataset_merge_method` when a new coordinate is assigned to a xr.DataArray

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -474,10 +474,7 @@ class DataArray(AbstractArray, DataWithCoords):
 
     def __setitem__(self, key, value):
         if isinstance(key, basestring):
-            # convert to dataset first to use merge.dataset_update_method
-            ds = self._to_temp_dataset()
-            ds[key] = value
-            self._coords[key] = ds[key].variable
+            self.coords[key] = value
         else:
             # Coordinates in key, value and self[key] should be consistent.
             # TODO Coordinate consistency in key is checked here, but it
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -474,7 +474,10 @@ class DataArray(AbstractArray, DataWithCoords):
 
     def __setitem__(self, key, value):
         if isinstance(key, basestring):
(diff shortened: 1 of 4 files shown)


xarray/core/dataarray.py
```diff
@@ -474,7 +474,10 @@ def __getitem__(self, key):
 
     def __setitem__(self, key, value):
         if isinstance(key, basestring):
-            self.coords[key] = value
+            # convert to dataset first to use merge.dataset_update_method
```
shoyer: Can we fix this in `DataArray.coords.__setitem__` instead? Otherwise that will still be broken:
```
In [33]: da = xr.DataArray([0, 1, 2], dims='x')

In [34]: da.coords['x'] = [0, 1, 2, 3]
```

fujiisoup: You are right.
Updated. Thanks.

shoyer:
Looks good to me, merging...

PR #2126, opened 2018-05-13
Add cftime to doc/environment.yml

spencerkclark:
cftime is now needed to build the documentation: http://xarray.pydata.org/en/latest/time-series.html#non-standard-calendars-and-dates-outside-the-timestamp-valid-range

Sorry I neglected this in #1252!


fmaussion:
Thanks! Let's see if it builds correctly now

PR #2131, opened 2018-05-14
Feature/pickle rasterio

### LINKED ISSUE #2121: rasterio backend should use DataStorePickleMixin (or something similar)

```Python
In [1]: import xarray as xr

In [2]: ds = xr.open_rasterio('RGB.byte.tif')

In [3]: ds
Out[3]:
<xarray.DataArray (band: 3, y: 718, x: 791)>
[1703814 values with dtype=uint8]
Coordinates:
  * band     (band) int64 1 2 3
  * y        (y) float64 2.827e+06 2.826e+06 2.826e+06 2.826e+06 2.826e+06 ...
  * x        (x) float64 1.021e+05 1.024e+05 1.027e+05 1.03e+05 1.033e+05 ...
Attributes:
    transform:   (101985.0, 300.0379266750948, 0.0, 2826915.0, 0.0, -300.0417...
    crs:         +init=epsg:32618
    res:         (300.0379266750948, 300.041782729805)
    is_tiled:    0
    nodatavals:  (0.0, 0.0, 0.0)

In [4]: import pickle

In [5]: pickle.dumps(ds)
<ipython-input-5-a165c2473431> in <module>()
----> 1 pickle.dumps(ds)

TypeError: can't pickle rasterio._io.RasterReader objects
```

#### Problem description

Originally reported by @rsignell-usgs in https://github.com/pangeo-data/pangeo/issues/249#issuecomment-388445370, the rasterio backend is not pickle-able. This obviously causes problems when using dask-distributed. We probably need to use `DataStorePickleMixin` or something similar on rasterio datasets to allow multiple readers of the same dataset.

#### Expected Output

```python
pickle.dumps(ds)
```

returns a pickled dataset.

<details>
xr.show_versions()
/Users/jhamman/anaconda/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.3
pandas: 0.22.0
numpy: 1.14.2
scipy: 1.0.1
netCDF4: 1.3.1
h5netcdf: 0.5.1
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.2
distributed: 1.21.6
matplotlib: 2.2.2
cartopy: 0.16.0
seaborn: 0.8.1
setuptools: 39.0.1
pip: 9.0.3
conda: 4.5.1
pytest: 3.5.1
IPython: 6.3.1
sphinx: 1.7.4
</details>

jhamman:
- [x] Closes #2121
 - [x] Tests added
 - [x] Tests passed 

cc @rsignell-usgs


xarray/backends/common.py
```diff
@@ -503,3 +503,43 @@ def assert_open(self):
...
                                  'if `autoclose=True` is used.')
+
+
+class PickleByReconstructionWrapper(object):
+
+    def __init__(self, opener, *args, **kwargs):
+
+        self.opener = opener
+        self.open_args = args
+        self.open_kwargs = kwargs
```
shoyer: For simplicity, let's pool these into a single `functools.partial` object. Note that you can get out keyword arguments if desired from `partial.keywords`.

xarray/backends/common.py
```diff
@@ -503,3 +503,43 @@ def assert_open(self):
...
+
+        self._ds = None
+        self._isopen = False
+
+    @property
+    def value(self):
```
shoyer: I'm a little concerned about the complexity of making file opening lazy. State makes things harder to reason about. For example, what happens with the `mode` argument if never actually accessed the file?

For now can we stick with constructing `value` eagerly?

xarray/backends/common.py
```diff
@@ -503,3 +503,43 @@ def assert_open(self):
...
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._ds = None
+        self._isopen = False
+
+    def __getitem__(self, key):
+        return self.value[key]
+
+    def __getattr__(self, name):
```
shoyer: This magic is convenient but I think it could possibly be a source of bugs. I would stick to requiring pulling out `.value`.

xarray/backends/common.py
```diff
@@ -503,3 +503,43 @@ def assert_open(self):
         if not self._isopen:
             raise AssertionError('internal failure: file must be open '
                                  'if `autoclose=True` is used.')
+
+
+class PickleByReconstructionWrapper(object):
```
shoyer: It would be nice to add a unit test verifies that this works properly independently of any concrete datastore.

Maybe something simple with `open()`?

xarray/tests/test_backends.py
```diff
@@ -2649,6 +2649,14 @@ def test_chunks(self):
                 ex = expected.sel(band=1).mean(dim='x')
                 assert_allclose(ac, ex)
 
+    def test_pickle_rio(self):
+        # regression test for https://github.com/pydata/xarray/issues/2121
+        with create_tmp_geotiff() as (tmp_file, expected):
+            with xr.open_rasterio(tmp_file) as rioda:
+                temp = pickle.dumps(rioda)
+                actual = pickle.loads(temp)
```
shoyer: Something to think about here -- I'm pretty sure this leaks a file descriptor. I'm not quite sure what the best way to deal with this is (maybe it's not worth worrying about yet), but in the past this has caused issues with tests on various platforms (especially Windows).

shoyer: And indeed, it's causing a test failure on Windows :).

I *think* it should work to explicitly open `actual` in a context manager, e.g.,
```
with pickle.loads(temp) as actual:
    ...
```

xarray/backends/common.py
jhamman: should we just have the backends pass in a partial function then?

xarray/backends/common.py
```diff
@@ -503,3 +503,43 @@ def assert_open(self):
...
+    def value(self):
+        if self._ds is not None and self._isopen:
+            return self._ds
+        self._ds = self.opener(*self.open_args, **self.open_kwargs)
+        self._isopen = True
+        return self._ds
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        del state['_ds']
```
fmaussion: shouldn't we close the file here? ``state['_ds'].close()``

xarray/backends/common.py
```diff
@@ -503,3 +503,43 @@ def assert_open(self):
...
+        del state['_isopen']
+        if self.open_kwargs.get('mode', None) == 'w':
+            # file has already been created, don't override when restoring
+            state['open_kwargs']['mode'] = 'a'
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._ds = None
```
fmaussion: Here too?

jhamman:
@rsignell-usgs - I think this should work in its current form. I'm writing some additional tests now and have some cleanup tasks as indicated in the reviews but I do not expect the behavior to change.

xarray/tests/test_backends.py
```diff
@@ -2649,6 +2649,14 @@ def test_chunks(self):
                 ex = expected.sel(band=1).mean(dim='x')
                 assert_allclose(ac, ex)
 
+    def test_pickle_rio(self):
+        # regression test for https://github.com/pydata/xarray/issues/2121
+        with create_tmp_geotiff() as (tmp_file, expected):
+            with xr.open_rasterio(tmp_file) as rioda:
+                temp = pickle.dumps(rioda)
+                with pickle.loads(temp) as actual:
+                    assert_equal(actual, rioda)
```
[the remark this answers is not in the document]
jhamman: Done in 55a3abc.

xarray/backends/common.py
jhamman: Done in 8b286c0

xarray/backends/common.py
jhamman: The basic idea is that most open file handles can't be pickled so we need to provide a mechanism to remove the existing handle and generate a new one in the dump/load steps in the pickling. In most cases, we do want to keep the original file open.

xarray/backends/common.py
jhamman: I'm actually not a big fan of dropping the lazy opening. For one, there will be a non-negligible performance penalty. Perhaps a better argument is that we've been using this logic successfully in xarray for a while now. That said, its not clear to me what the potential problems with this approach so I'm open to convincing otherwise.

xarray/backends/common.py
shoyer: If we don't anticipate adding any argument, then I think it's just as easy to make the partial function here. I suppose there's some potential overlap in needs with auto-closing a file, so I'm OK either way.

xarray/backends/common.py
shoyer: Don't we always open files immediately to pull out metadata? (e.g., variables and dimensions)

shoyer:
Thinking about this a little more, maybe we should consolidate pickling with autoclosing into a single, composable wrapper class. Usage would look something like:
```python
wrapper = FileWrapper(netCDF4.Dataset, path, mode='r', autoclose=False, kwargs=kwargs)
with wrapper:
    # will close/open wrapped file
    with wrapper.acquire() as ds:
        # context manager is dummy, simply returning the open file object
        ... # do stuff with ds, a netCDF4.Dataset instance
# or could write wrapper.close() instead of using the context manager

wrapper = FileWrapper(netCDF4.Dataset, path, mode='r', autoclose=True, kwargs=kwargs)
with wrapper:
    # explicit opening/closing the wrapper is a no-op if autoclose=True
    with wrapper.acquire() as ds:
        # context manager opens/closes file
        # if file is already open, it's a no-op
        ... # do stuff with ds, a netCDF4.Dataset instance
```

We could then expose `xarray.backends.FileWrapper` as part of our external API for backend authors (without coupling it into the Datastore class directly).

Eventually, we might even change the default for `autoclose` to automatically use LRU caching (but that obviously would come later).

rsignell-usgs:
It's working for me!
https://gist.github.com/rsignell-usgs/ef81fb4306dac3a2406d0adb575b340f

rsignell-usgs:
@jhamman , still very much interested in this -- could the functionality as-is be merged and enhanced later?

xarray/tests/test_backends.py
```diff
@@ -2913,3 +2922,22 @@ def test_dataarray_to_netcdf_no_name_pathlib(self):
 
             with open_dataarray(tmp) as loaded_da:
                 assert_identical(original_da, loaded_da)
+
+
+def test_pickle_reconstructor():
+
+    lines = ['foo bar spam eggs']
+
+    with create_tmp_file() as tmp:
```
shoyer: This needs to add `allow_cleanup_failure=ON_WINDOWS` for the test to pass on windows

jhamman: Great, that did fix the windows failure.

jhamman:
All the tests are passing here. Its been a while so perhaps it would be good to get another review from @pydata/xarray. 

@rsignell-usgs - can you give this another whirl on your use case?

rsignell-usgs:
@jhamman , although I'm getting distributed workers to compute the mean from a bunch of images, I'm getting a "Failed to Serialize" error in cell [23] of this notebook:
https://gist.github.com/rsignell-usgs/90f15e2da918e3c6ba6ee5bb6095d594
If this is a bug, I think it was there before the recent changes.

xarray/backends/common.py
```diff
@@ -507,3 +508,30 @@ def assert_open(self):
...
+        self.mode = mode
+        self._ds = None
+
+    @property
+    def value(self):
+        self._ds = self.opener()
+        return self._ds
+
+    def __getstate__(self):
+        del self._ds
```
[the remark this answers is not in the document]
jhamman: Thanks, I had just noticed this problem and it should be fixed/tested it in 35520c0.

rsignell-usgs:
@jhamman woohoo!  Cell [20] completes nicely now: 
https://gist.github.com/rsignell-usgs/90f15e2da918e3c6ba6ee5bb6095d594
I'm getting some errors during the 2nd call, but I think those are unrelated and didn't affect the successful completions of the task, right

rsignell-usgs:
Might this PR warrant a new minor release?

shoyer:
@rsignell-usgs Sure, though `interp()` (https://github.com/pydata/xarray/pull/2104) is almost ready, too, so we might wait for that.

rsignell-usgs:
Sounds good.  Thanks @shoyer!

shoyer:
v0.10.7 is out

PR #2136, opened 2018-05-16
Fix test suite with pandas 0.23

shoyer:


jhamman:
Thanks. I think this was causing the tests to fail in #1811.

PR #2144, opened 2018-05-16
Add strftime() to datetime accessor

### LINKED ISSUE #2090: strftime and/or format support for DatetimeAccessor
There's no easy way currently to control the conversion to string of time values/series. For this purpose, Panda's own `.dt` attribute has an implementation of [strftime](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.dt.strftime.html).

Is there interest in adding similar functionality to xarray?

dopplershift:
This matches pandas and makes it possible to pass a datetime dataarray
to something expecting to be able to use strftime().

 - [x] Closes #2090 

### DIFF
--- a/xarray/core/accessors.py
+++ b/xarray/core/accessors.py
@@ -258,3 +258,24 @@ class DatetimeAccessor(object):
             Array-like of datetime fields accessed for each element in values
         '''
         return self._tslib_round_accessor("round", freq)
+
+    def strftime(self, date_format):
+        '''
+        Return an array of formatted strings specified by date_format, which
+        supports the same string format as the python standard library. Details
+        of the string format can be found in `python string format doc
+        <https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior>`__
+
+        Parameters
+        ----------
+        date_format : str
+            date format string (e.g. "%Y-%m-%d")
+
+        Returns
+        -------
+        ndarray of formatted strings
+        '''
+        values = self._obj.data
+        values_as_series = pd.Series(values.ravel())
+        strs = values_as_series.dt.strftime(date_format)
+        return strs.values.reshape(values.shape)
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -82,6 +82,10 @@ Enhancements
   currently only supported by the netCDF4 and zarr backends. (:issue:`1784`).
   By `Joe Hamman <https://github.com/jhamman>`_.
 
+- ``strftime`` method added to ``.dt`` accessor, making it simpler to hand
+  datetime ``DataArray`` to other code expecting to format dates and times.
+  (:issue:`2090`). By `Ryan May <https://github.com/dopplershift>`_.
+
 Bug fixes
 ~~~~~~~~~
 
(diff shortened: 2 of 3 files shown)


dopplershift:
Any chance this makes it into 0.10.4?

xarray/core/accessors.py
```diff
@@ -258,3 +258,24 @@ def round(self, freq):
             Array-like of datetime fields accessed for each element in values
         '''
         return self._tslib_round_accessor("round", freq)
+
+    def strftime(self, date_format):
+        values = self._obj.data
+        values_as_series = pd.Series(values.ravel())
+        strs = values_as_series.dt.strftime(date_format)
+        return strs.values.reshape(values.shape)
+    strftime.__doc__ = '''
```
jhamman: I may be missing something but what's the rational for assigning the doc string this way, as apposed to under the method definition?

dopplershift: It came from pandas, and I think the purpose was to avoid the long line caused by the URL...maybe? I'm happy to adjust this to something more straightforward.

shoyer: I would lean toward just writing the docstring in the normal way, and silencing the lint error.

xarray/core/accessors.py
```diff
@@ -258,3 +258,24 @@ def round(self, freq):
             Array-like of datetime fields accessed for each element in values
         '''
         return self._tslib_round_accessor("round", freq)
+
+    def strftime(self, date_format):
+        values = self._obj.data
+        values_as_series = pd.Series(values.ravel())
+        strs = values_as_series.dt.strftime(date_format)
+        return strs.values.reshape(values.shape)
```
shoyer: For consistency with the other time methods, it would be good to:
1. support dask arrays
2. return an `xarray.DataArray`

spencerkclark:
It might also be worth considering adding this accessor to `CFTimeIndex`, since `cftime.datetime` [implements a version of `strftime`](https://unidata.github.io/cftime/api.html) (that would enable use with dates from non-standard calendars).  No worries if you'd rather put that off until later though.

shoyer:
Sorry, it's too late for v0.10.4 (But my intent is not to wait too long for 0.10.5.)

dopplershift:
Since this can wait for 0.10.5, I can go slowly and get this right.

abrammer:
Was looking for this functionality and came across the parent [issue](https://github.com/pydata/xarray/issues/2090) and this work around,  did this just get forgotten or was there more work to be done?  It'd be great to have this functionality readily available without having to work around.

dopplershift:
Just haven't had the cycles to bring this to the finish line.

abrammer:
ok, Thanks.  Anything an outsider can help with?

ahuang11:
Would also like to see this move forward and would be happy to help if needed.

max-sixty:
Given we're ~a year since the last commit and @dopplershift has taken this so far, @dopplershift are you OK if someone takes this to the finish line?

abrammer:
I took a stab at this after being reminded about it.  https://github.com/abrammer/xarray/tree/strftime_addition
Can open a new PR if this one has gone stale.

jhamman:
@abrammer - please go ahead and open a PR. We can leave this open in the meantime. Thanks!

dopplershift:
So I finally have time to work on this...but if someone has working code to do this instead, I'm totally fine with that going in instead.

abrammer:
@dopplershift I built from your start here, over in #2989 and keeping your initial code and tests along with an edited version of your what's-new.  Kept you in the what's new, so hope that's cool otherwise lmk.

PR #2146, opened 2018-05-17
Add favicon to docs?

fmaussion:
Don't know if we want this, but it's possible to replace the default RTD icon. I'm not even sure if the cool kids are making bookmarks anymore, but I still do. This is how it looks like on my Firefox fav tab:

selection_054


fmaussion:
Tabs also display it on firefox or chromium, so it does have some visibility:

selection_056
:

benbovy:
Ha great! This will certainly help people like me who are struggling with many open tabs including a bunch of RTD pages for different libraries :)

PR #2151, opened 2018-05-17
Plot labels use CF convention information if available.

### LINKED ISSUE #2135: use CF conventions to enhance plot labels
Elsewhere in xarray we use CF conventions to help with automatic decoding of datasets. Here I propose we consider using [CF metadata conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html) to improve the automatic labelling of plots. If datasets declare `long_name`, `standard_name`, and `units` attributes, we could use these instead of the variable name to label the relevant axes / colorbars. This feature would have helped me avoid several past mistakes due to my failure to examine the `units` attribute (e.g. data given in cm when I assumed m).

Here I create some data with relevant attributes

```python
import xarray as xr
import numpy as np
ds = xr.Dataset({'foo': ('x', np.random.rand(10),
                         {'long_name': 'height',
                          'units': 'm'})},
                coords={'x': ('x', np.arange(10),
                              {'long_name': 'distance',
                               'units': 'km'})})
ds.foo.plot()
```

image

#### Problem description

We have neglected the variable attributes, which would provide better axis labels.

#### Expected Output

Consider this instead:
```python
def label_from_attrs(da):
    attrs = da.attrs
    if 'long_name' in attrs:
        name = attrs['long_name']
    elif 'standard_name' in attrs:
        name = attrs['standard_name']
    else:
        name = da.name
    if 'units' in da.attrs:
        units = ' [{}]'.format(da.attrs['units'])
    label = name + units
    return label
ds.foo.plot()
plt.xlabel(label_from_attrs(ds.x))
plt.ylabel(label_from_attrs(ds.foo))
```
image

I feel like this would be a sensible default. But it would be a breaking change. We could make it optional with a keyword like `labels_from_attrs=True`.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.111+
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.3+dev13.g98373f0
pandas: 0.22.0
numpy: 1.14.3
scipy: 1.0.1
netCDF4: 1.3.1
h5netcdf: 0.5.1
h5py: 2.7.1
Nio: None
zarr: 2.2.1.dev2
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.4
distributed: 1.21.8
matplotlib: 2.2.2
cartopy: None
seaborn: None
setuptools: 39.1.0
pip: 9.0.1
conda: 4.3.29
pytest: 3.5.1
IPython: 6.3.1
sphinx: None
</details>

dcherian:
Uses attrs long_name/standard_name, units if available.

 - [x] Closes #2135, #1630, #1787

This PR basically uses @rabernat's `label_from_attrs` in #2135. 

Examples:
image

image

image
image
image

I also changed the way histograms are labelled
**Old:**
image

**New**:
image

### DIFF
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -66,6 +66,10 @@ For these examples we'll use the North American air temperature dataset.
     # Convert to celsius
     air = airtemps.air - 273.15
 
+    # copy attributes to get nice figure labels and change Kelvin to Celsius
+    air.attrs = airtemps.air.attrs
+    air.attrs['units'] = 'deg C'
+
 
 One Dimension
 -------------
@@ -73,7 +77,7 @@ One Dimension
 Simple Example
 ~~~~~~~~~~~~~~
 
-xarray uses the coordinate name to label the x axis.
+xarray uses the coordinate name along with any associated `CF-compliant metadata <http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html>`_ (the `attrs` 'long_name', 'standard_name' and 'units') to label the axes.
 
 .. ipython:: python
 
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -89,7 +89,7 @@ The simplest way to make a plot is to call the :py:func:`xarray.DataArray.plot()
     @savefig plotting_1d_simple.png width=4in
     air1d.plot()
 
-xarray uses the coordinate name along with  metadata ``attrs.long_name``, ``attrs.standard_name``, ``DataArray.name`` and ``attrs.units`` (if available) to label the axes. The names ``long_name``, ``standard_name`` and ``units`` are defined by the `CF-conventions <http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html>`_. When choosing names, the order of precedence is ``long_name``, ``standard_name`` and finally ``DataArray.name``. The y-axis label in the above plot was constructed from the ``long_name`` and ``units`` attributes of ``air1d``.
+xarray uses the coordinate name along with  metadata ``attrs.long_name``, ``attrs.standard_name``, ``DataArray.name`` and ``attrs.units`` (if available) to label the axes. The names ``long_name``, ``standard_name`` and ``units`` are copied from the `CF-conventions spec <http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html>`_. When choosing names, the order of precedence is ``long_name``, ``standard_name`` and finally ``DataArray.name``. The y-axis label in the above plot was constructed from the ``long_name`` and ``units`` attributes of ``air1d``.
 
 .. ipython:: python
 
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -70,6 +70,9 @@ For these examples we'll use the North American air temperature dataset.
     air.attrs = airtemps.air.attrs
     air.attrs['units'] = 'deg C'
 
+.. note::
+   Until :issue:`1614` is solved, you might need to copy over the metadata in ``attrs`` to get informative figure labels (as was done above).
+
 
 One Dimension
 -------------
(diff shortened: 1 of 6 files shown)


dcherian:
The units in the images say "deg K" but really should be "deg C". I've fixed that

xarray/plot/utils.py
```diff
@@ -354,3 +355,28 @@ def get_axis(figsize, size, aspect, ax):
         ax = plt.gca()
 
     return ax
+
+
+def label_from_attrs(da):
+    ''' Makes informative labels if variable metadata (attrs) follows
+        CF conventions. '''
```
rabernat: Should this be a private method?

xarray/tests/test_plot.py
```diff
@@ -89,6 +89,20 @@ class TestPlot(PlotTestCase):
...
+        da.attrs['units'] = 'a_units'
+        da.attrs['long_name'] = 'a_long_name'
+        da.attrs['standard_name'] = 'a_standard_name'
+        assert 'a_long_name [a_units]' == label_from_attrs(da)
+        da.attrs.pop('long_name')
+        assert 'a_standard_name [a_units]' == label_from_attrs(da)
+        da.attrs.pop('standard_name')
+        assert 'a [a_units]' == label_from_attrs(da)
+        da.attrs.pop('units')
+        assert 'a' == label_from_attrs(da)
```
rabernat: perfect test

xarray/plot/utils.py
dcherian: I thought it might be useful to have it accessible.

dcherian:
> Do we want to have a way to disable this new feature and revert back to old behavior?

I can add an option if needed.

rabernat:
Anyone else have thoughts on this? @Zac-HD, @fmaussion, @fujiisoup, and @jbusecke have all contributed to those earlier discussions / PRs.

jbusecke:
This looks good to me! thanks for the implementation! This will save lots of seconds that add up ;-)

xarray/tests/test_plot.py
Zac-HD: Good test, sure, but it doesn't cover what happens if the units are present without various combinations of name metadata.

I'd also find this easier to follow using [`pytest.mark.parametrize`](https://docs.pytest.org/en/latest/reference.html#pytest-mark-parametrize) to separate each combination.

dcherian:
Rebased on master and updated based on review comments.

I haven't used `pytest.mark.parameterize` because I found it a little confusing.

doc/plotting.rst
```diff
@@ -66,14 +66,18 @@ For these examples we'll use the North American air temperature dataset.
...
+
 
 One Dimension
 -------------
 
 Simple Example
 ~~~~~~~~~~~~~~
 
-xarray uses the coordinate name to label the x axis.
+xarray uses the coordinate name along with any associated `CF-compliant metadata <http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html>`_ (the `attrs` 'long_name', 'standard_name' and 'units') to label the axes.
```
shoyer: Can you elaborate a little bit more on how these are used and what the expectations are? It would be good to entirely self-contained, so users don't need to reference CF conventions (which are long and mostly irrelevant to this)

dcherian: Yes, I've added a new commit. Let me know if you want more changes.

doc/plotting.rst
```diff
@@ -86,6 +89,12 @@ xarray uses the coordinate name along with any associated `CF-compliant metadata
     @savefig plotting_1d_simple.png width=4in
     air1d.plot()
 
+xarray uses the coordinate name along with  metadata ``attrs.long_name``, ``attrs.standard_name``, ``DataArray.name`` and ``attrs.units`` (if available) to label the axes. The names ``long_name``, ``standard_name`` and ``units`` are defined by the `CF-conventions <http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html>`_. When choosing names, the order of precedence is ``long_name``, ``standard_name`` and finally ``DataArray.name``. The y-axis label in the above plot was constructed from the ``long_name`` and ``units`` attributes of ``air1d``.
```
shoyer: Maybe same "copied from" instead of "defined by"?

dcherian: done.

shoyer:
OK, I think this is ready to merge. Anyone else have concerns?

PR #2152, opened 2018-05-17
ENH: Plotting for groupby_bins

maahn:
DataArrays created with e.g. groupy_bins have coords arrays consisting of pd._libs.interval.Interval. Therefore, they cannot be plotted. The small patch replaces the the pd._libs.interval.Interval values with the interval's center point and adds `_center` to the label name. It looks like this: https://gist.github.com/maahn/91da0a8d299ef6567827749cbe2f1913
I don't think there is any need for additional documentation except the `whats-new.rst` or tests(?), but I'm also happy to add them if you think it is required. 


rabernat:
This seems like a good idea. 

`groupby_bins` has a `labels` option, which can override the default labels generated by pandas. That's what is done in the (multidimensional groupby example](http://xarray.pydata.org/en/stable/examples/multidimensional-coords.html#multidimensional-groupby).

I wonder if it's worth just making this behavior the default, or generating an additional coordinate for the bin centers automatically.

maahn:
Thanks, for the comment, but I wouldn't use center labels when using `groupby_bins` by default, that could be misleading in case of non-uniform (e.g. exponential) bin spacing. I guess that's acceptable for a plot, but not for a DataArray, because the information about the boundaries would be lost.  And from a user perspective, I would find it a bit confusing if an additional coordinate would show up when using `groupby_bins`.

xarray/plot/plot.py
```diff
@@ -48,14 +48,27 @@ def _ensure_plottable(*args):
...
+
+
+def _interval_to_mid_points(array, label):
+    """
+    Helper function which checks whether array consists of
+    pd._libs.interval.Interval. If yes, it returns an array
+    with the Intervals' mid points.In addition, _center is
+    appended to the label
+    """
+    if _valid_other_type(array, [pd._libs.interval.Interval]):
```
shoyer: Please use `pd.Interval` directly, which is how this is exposed as public API.

xarray/plot/plot.py
```diff
@@ -267,6 +280,8 @@ def line(darray, *args, **kwargs):
 
     _ensure_plottable(xplt)
 
+    xplt.values, xlabel = _interval_to_mid_points(xplt.values, xlabel)
```
shoyer: Rather than mutating `xplt.values`, please assign a new variable. Otherwise this could mutate an inline argument to this function.

xarray/plot/plot.py
maahn: Thanks, changed.

xarray/plot/plot.py
maahn: OK, changed it here and in lines 628f. Note that this changes the type of xplt passed to ax.plot from DataArray to np.array, but I think this shouldn't matter.

shoyer:
A couple of other thoughts:

1. For 2D plots that already show regions (imshow/pcolormesh), I'm not sure it makes sense to update the label to include the word "center". This plot around shows intervals pretty clearly:
image
2. This needs tests, at least something to make sure that plotting with an Intervals on an axis does not crash in xarray/tests/test_plot.py

xarray/plot/plot.py
```diff
@@ -267,7 +276,14 @@ def line(darray, *args, **kwargs):
 
     _ensure_plottable(xplt)
 
-    primitive = ax.plot(xplt, yplt, *args, **kwargs)
+    # Remove pd.Intervals if contained in xplt.values.
+    if _valid_other_type(xplt.values, [pd.Interval]):
+        xplt_val = _interval_to_mid_points(xplt.values)
```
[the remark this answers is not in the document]
shoyer: Yeah, this is probably better default behavior. Potentially there could be a flag to choose.

xarray/tests/test_plot.py
```diff
@@ -975,13 +972,9 @@ def test_cmap_and_color_both(self):
 
     def test_2d_coord_with_interval(self):
         for dim in self.darray.dims:
```
maahn: I left the loop here, because for the 2d plots, x and y axis are treated separately.

maahn:
Ok, I added the interval_step_plot kwarg (default True) to the 1D line plot to make a step plot with the real (i.e. not interpolated) boundaries. After that I had the feeling that it's inconsistent if line uses the real boundaries but pcolormesh doesn't. So I also patched that, but I had to disable infer_intervals for these cases. See also:
https://gist.github.com/maahn/91da0a8d299ef6567827749cbe2f1913/530ed7bf77c9c3257cab46c3894c1412085a217a

Let me know what you think and I will also add some documentation.

shoyer:
Rather than using `interval_step_plot=True`, what about making an entirely new plot method, `plot.step()`? That would feel a little cleaner to me.

I'm not sure that a step plot would be preferred by most users for a plot over an axis on which groupby_bins was applied. Personally, I would probably prefer a line + scatter plot (e.g., `marker='s'`). Maybe @rabernat (groupby_bins author) has opinions here?

xarray/plot/plot.py
```diff
@@ -310,9 +309,14 @@ def line(darray, *args, **kwargs):
 
     # Remove pd.Intervals if contained in xplt.values.
     if _valid_other_type(xplt.values, [pd.Interval]):
-        if interval_step_plot:
+        if kwargs.get('linestyle', '').startswith('steps-'):
             xplt_val, yplt_val = _interval_to_double_bound_points(xplt.values,
                                                                   yplt.values)
+            # just to be sure that matplotlib is not confused
+            kwargs['linestyle'] = kwargs['linestyle'].replace(
```
maahn: I know this is quite ugly, but does it make sense to import re only for this one line?

maahn:
Good idea, it turned out the step function was quite easy to implement by using https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/axes/_axes.py#L1735 
So now, 1D data defaults to the standard line plot, but you can use plot.step() instead. plot.step() works also with normal xarrays. plot.step() 
See https://gist.github.com/maahn/91da0a8d299ef6567827749cbe2f1913

xarray/plot/plot.py
```diff
@@ -48,14 +49,53 @@ def _ensure_plottable(*args):
...
+                            'datetime.datetime or pd.Interval.')
+
+
+def _interval_to_mid_points(array):
+    """
+    Helper function which returns an array
+    with the Intervals' mid points.
+    """
+
+    return np.asarray(list(map(lambda x: x.mid, array)))
```
shoyer: Consider writing these with list comprehensions, e.g., `np.array([x.mid for x in array])`

xarray/plot/plot.py
```diff
@@ -48,14 +49,53 @@ def _ensure_plottable(*args):
...
+    """
+    Helper function to deal with a xarray consisting of pd.Intervals. Each
+    interval is replaced with both boundaries. I.e. the length of xarray
+    doubles. yarray is modified so it matches the new shape of xarray.
+    """
+
+    xarray1 = np.asarray(list(map(lambda x: x.left, xarray)))
+    xarray2 = np.asarray(list(map(lambda x: x.right, xarray)))
+
+    xarray = ([x for x in itertools.chain.from_iterable(
```
shoyer: This can just be `list(itertools.chain.from_iterable(zip(xarray1, xarray2)))`

xarray/plot/plot.py
```diff
@@ -661,6 +766,28 @@ def newplotfunc(darray, x=None, y=None, figsize=None, size=None,
...
+        if _valid_other_type(yval, [pd.Interval]):
+            if 'pcolormesh' == plotfunc.__name__:
+                yplt = _interval_to_bound_points(yval)
+                ylab_extra = ''
+            else:
+                yplt = _interval_to_mid_points(yval)
+                ylab_extra = '_center'
+        else:
+            yplt = yval
+            ylab_extra = ''
```
shoyer: Can we put this logic in a helper function? This function is already getting pretty long :)

maahn:
Thanks for the review.

doc/plotting.rst
```diff
@@ -224,6 +224,38 @@ It is also possible to make line plots such that the data are on the x-axis and
     @savefig plotting_example_xy_kwarg.png
     air.isel(time=10, lon=[10, 11]).plot(y='lat', hue='lon')
 
+
+Step plots
+~~~~~~~~~~
+
+As an alternative, also a step lot similar to matplotlib's ``plt.step`` can be
```
yohai: "step lot" should be step plot?

maahn: of course, thanks!

maahn:
Sorry, for the delay, I finally merged upstream. Looks like the failed builds are unrelated to my changes, so it should be ready for merging?

dcherian:
Really sorry for the delay, @maahn. 

I've merged master, refactored out the utility functions to `utils.py`, fixed the tests locally and added a whats-new entry. I'll merge once this round of tests pass.

maahn:
great, thanks for taking care of that!

dcherian:
Failed test is cfgrib test.

Thanks @maahn

Issue #2036, opened 2018-04-03
better error message for to_netcdf -> unlimited_dims


mathause:
#### Code Sample, a copy-pastable example if possible

```python
# Your code here
import numpy as np
import xarray as xr
x = np.arange(10)
da = xr.Dataset(data_vars=dict(data=('dim1', x)), 
               coords=dict(dim1=('dim1', x), dim2=('dim2', x)))
da.to_netcdf('tst.nc', format='NETCDF4_CLASSIC', unlimited_dims='dim1')

```
#### Problem description

This creates the error `RuntimeError: NetCDF: NC_UNLIMITED size already in use`. With `format='NETCDF4'` silently creates the dimensions `d`, `i`, `m`, and `\1`.

The correct syntax is `unlimited_dims=['dim1']`. 

With `format='NETCDF4_CLASSIC'` and  `unlimited_dims=['dim1', 'dim2']`, still raises the not-so-helpful `NC_UNLIMITED` error.

I only tested with netCDF4 as backend.

#### Expected Output

* better error message
*  work with `unlimited_dims='dim1'`

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.120-45-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8

xarray: 0.10.2
pandas: 0.22.0
numpy: 1.14.2
scipy: 1.0.1
netCDF4: 1.3.1
h5netcdf: 0.5.0
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: 1.0.0
dask: 0.17.2
distributed: 1.21.5
matplotlib: 2.2.2
cartopy: 0.16.0
seaborn: 0.8.1
setuptools: 39.0.1
pip: 9.0.3
conda: None
pytest: 3.5.0
IPython: 6.3.0
sphinx: 1.7.2
</details>

mathause:
This could be relevant here:
https://github.com/pydata/xarray/pull/1170/files

mathause:
https://github.com/shoyer/h5netcdf/issues/42

jhamman:
closed via #2154

PR #2154, opened 2018-05-17
fix unlimited dims bug

### LINKED ISSUE #2134: unlimited_dims generates 0-length dimensions named as letters of unlimited dimension
I'm not sure I understand how the `unlimited_dims` option to `to_netcdf()` is supposed to work. Consider the following:
```python
ds = xr.Dataset()
ds['time'] = xr.DataArray(pd.date_range('2000-01-01', '2000-01-10'), dims='time')
ds.to_netcdf('timedim.cdf', unlimited_dims='time')
```
This results in a file that looks like this:
```
$ ncdump timedim.cdf
netcdf timedim {
dimensions:
	t = UNLIMITED ; // (0 currently)
	i = UNLIMITED ; // (0 currently)
	m = UNLIMITED ; // (0 currently)
	e = UNLIMITED ; // (0 currently)
	time = UNLIMITED ; // (10 currently)
variables:
	int64 time(time) ;
		time:units = "days since 2000-01-01 00:00:00" ;
		time:calendar = "proleptic_gregorian" ;
data:

 time = 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 ;
}
```
Note the dimensions named `t`, `i`, `m`, `e` all with zero length. The `time` dimension (which is the only one that should exist) is properly set to `UNLIMITED` but we shouldn't have the four extra dimensions. What's going on here? The same behavior occurs when setting via `ds.encoding['unlimited_dims'] = 'time'`. Everything is as expected without the `unlimited_dims` option (but the `time` dimension is not `UNLIMITED`, of course). 

I thought it could be related to the variable and dimension having the same name, but this also happens when they are different.

#### Expected Output

There shouldn't be extra 0-length dimensions

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Darwin
OS-release: 16.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

xarray: 0.10.3
pandas: 0.22.0
numpy: 1.14.3
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
h5py: 2.7.1
Nio: None
zarr: 2.2.0
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.1
distributed: 1.20.2
matplotlib: 2.2.2
cartopy: 0.16.0
seaborn: None
setuptools: 36.5.0.post20170921
pip: 9.0.1
conda: 4.5.3
pytest: None
IPython: 6.3.1
sphinx: 1.7.1

</details>

jhamman:
- [x] Closes #2134
 - [x] Tests added
 - [ ] Tests passed

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -39,7 +39,9 @@ Enhancements
 
 Bug fixes
 ~~~~~~~~~
-
+- Fixed a bug where `to_netcdf(..., unlimited_dims='bar'` yielded NetCDF files
+  with spurious 0-length dimensions (i.e. `b`, `a`, and `r`) (:issue:`2134`).
+  By `Joe Hamman <https://github.com/jhamman>`_.
 
 .. _whats-new.0.10.4:
 
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -686,6 +686,9 @@ def to_netcdf(dataset, path_or_file=None, mode='w', format=None, group=None,
 
     if unlimited_dims is None:
         unlimited_dims = dataset.encoding.get('unlimited_dims', None)
+    if isinstance(unlimited_dims, basestring):
+        unlimited_dims = [unlimited_dims]
+
     try:
         dataset.dump_to_store(store, sync=sync, encoding=encoding,
                               unlimited_dims=unlimited_dims, compute=compute)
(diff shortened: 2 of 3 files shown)


shoyer:
This should also fix #2036

PR #2155, opened 2018-05-17
Validate output dimension sizes with apply_ufunc

### LINKED ISSUE #1931: apply_ufunc produces illegal coordinate sizes

shoyer:
Fixes #1931

Uses of apply_ufunc that change dimension size now raise an explicit error,
e.g.,

    >>> xr.apply_ufunc(lambda x: x[:5], xr.Variable('x', np.arange(10)))
    ValueError: size of dimension 'x' on inputs was unexpectedly changed by
    applied function from 10 to 5. Only dimensions specified in ``exclude_dims``
    with xarray.apply_ufunc are allowed to change size.

 - [x] Closes #1931 (remove if there is no corresponding issue, which should only be the case for minor changes)


xarray/core/computation.py
```diff
@@ -559,20 +559,29 @@ def func(*arrays):
...
-            output.append(var)
-        return tuple(output)
-    else:
-        dims, = output_dims
-        var = Variable(dims, result_data)
+    if signature.num_outputs == 1:
+        result_data = [result_data]
+
+    output = []
+    for dims, data in zip(output_dims, result_data):
```
fujiisoup: I think it would be nice if we raise an informative error message when the number of `output_dims` and `result_data` is different.
Currently this will silently drop some items if they have a different number of items.

shoyer: Good point -- done!

shoyer:
I plan to merge this shortly, unless anyone else has feedback here

PR #2156, opened 2018-05-17
Bug fixes for Dataset.reduce() and n-dimensional cumsum/cumprod

### LINKED ISSUE #1470: Dataset.mean drops coordinates
Hi,

Dataset.mean drops coordinates, unlike DataArray.mean:
```
x = xarray.DataArray([2., 4.], 
                     coords={'a':[1., 2.], 'b': 3.}, 
                     dims=['a'],
                     name='name')
print 'DataArray mean:'
print x.mean(dim='a')
print 'Dataset mean:'
print x.to_dataset().mean(dim='a')['name']
```
Output
```
DataArray mean
<xarray.DataArray 'name' ()>
array(3.0)
Coordinates:
    b        float64 3.0
Dataset mean
<xarray.DataArray 'name' ()>
array(3.0)
```

Thanks
Ignacio

shoyer:
Fixes GH1470, "Dataset.mean drops coordinates"

Fixes a bug where non-scalar data-variables that did not include the
aggregated dimension were not properly reduced:

    Previously::

        >>> ds = Dataset({'x': ('a', [2, 2]), 'y': 2, 'z': ('b', [2])})
        >>> ds.var('a')
        <xarray.Dataset>
        Dimensions:  (b: 1)
        Dimensions without coordinates: b
        Data variables:
            x        float64 0.0
            y        float64 0.0
            z        (b) int64 2

    Now::

        >>> ds.var('a')
        <xarray.Dataset>
        Dimensions:  (b: 1)
        Dimensions without coordinates: b
        Data variables:
            x        int64 0
            y        int64 0
            z        (b) int64 0

Finally, adds support for n-dimensional cumsum() and cumprod(), reducing
over all dimensions of an array. (This was necessary as part of the above fix.)

 - [x] Closes #1470 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2597,7 +2597,6 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
             if name in self.coords:
                 if not reduce_dims:
                     variables[name] = var
-                # drop other coordinates
             else:
                 if (not numeric_only or
                         np.issubdtype(var.dtype, np.number) or
@@ -2615,7 +2614,6 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
                                                  keep_attrs=keep_attrs,
                                                  allow_lazy=allow_lazy,
                                                  **kwargs)
-                # drop other data_vars
 
         coord_names = set(k for k in self.coords if k in variables)
         attrs = self.attrs if keep_attrs else None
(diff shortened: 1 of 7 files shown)


xarray/core/dataset.py
```diff
@@ -2594,26 +2594,28 @@ def reduce(self, func, dim=None, keep_attrs=False, numeric_only=False,
...
+                    elif len(reduce_dims) == var.ndim:
+                        # prefer to aggregate over axis=None rather than
+                        # axis=(0, 1) if they will be equivalent, because
+                        # the former is often more efficient
+                        reduce_dims = None
+                    variables[name] = var.reduce(func, dim=reduce_dims,
+                                                 keep_attrs=keep_attrs,
+                                                 allow_lazy=allow_lazy,
+                                                 **kwargs)
+                # drop other data_vars
```
jhamman: errant comment?

shoyer: This was intentional, but I don't think it actually adds much value.

PR #2162, opened 2018-05-18
Test suite: explicitly ignore irrelevant warnings

### LINKED ISSUE #2161: Regression: Dataset.update(Dataset)
Dataset().update(Dataset())

 

FutureWarning: iteration over an xarray.Dataset will change in xarray v0.11 to only include data variables, not coordinates. Iterate over the Dataset.variables property instead to preserve existing behavior in a forwards compatible manner.

This is a regression in xarray 0.10.4.
@shoyer this isn't serious enough to warrant an immediate release on its own, but we're already doing one so we might as well include it.

shoyer:
Includes silencing two warnings that show up in xarray's public API:

1. `Dataset.update(Dataset)` #2161
2. `.equals()` with scalar timedelta64/datetime64 arrays containing NaT:

```
In [1]: import numpy as np

In [2]: import xarray as xr

In [3]: array = xr.DataArray([np.datetime64('NaT')])

In [4]: array[0].equals(array[0])
/Users/shoyer/dev/xarray/xarray/core/duck_array_ops.py:148: FutureWarning: In the future, 'NAT == x' and 'x == NAT' will always be False.
  flag_array = (arr1 == arr2)
Out[16]: True
```

Fixes #2161


shoyer:
@fujiisoup any thoughts?

Do you agree that we should have different behavior for overriding `coords` with `__setitem__` vs. `update()`?

fujiisoup:
I didn't notice that iterating over dataset gives a FutureWarning.

> Do you agree that we should have different behavior for overriding coords with __setitem__ vs. update()?

I actually do not yet understand well the difference between `ds.update(other)` and `ds.merge(other)`.

The docs says
> If any dimensions would have inconsistent sizes in the updated dataset.

`update` raises a ValueError, but it does not.
Maybe it should work identically with `xr.merge(other)`?

shoyer:
-[ ] make a decision on coordinate conflicts with `update()`

shoyer:
OK, updated to only silence the warning in `Dataset.update(Dataset)` (by refactoring the logic)

fujiisoup:
Do we update coordinates if other is Dataset?

shoyer:
Yes, if other is a Dataset coordinates get overwritten. We probably already
had test coverage for this, but I also added some explicit tests.

shoyer:
I plan to merge this shortly (with a day or two) unless anyone else has comments

xarray/core/merge.py
```diff
@@ -547,21 +547,24 @@ def dataset_merge_method(dataset, other, overwrite_vars, compat, join):
...
-    other = other.copy()
-    for k, obj in other.items():
-        if isinstance(obj, (Dataset, DataArray)):
-            # drop duplicated coordinates
-            coord_names = [c for c in obj.coords
-                           if c not in obj.dims and c in dataset.coords]
-            if coord_names:
-                other[k] = obj.drop(coord_names)
+    if not isinstance(other, Dataset):
+        other = OrderedDict(other)
```
fujiisoup: I am slightly wondering this behavior may change after we finally support #884, i.e. we will not include coordinates in `.key()`.

shoyer:
This line only applies to non-Dataset objects (see the isintance check on
the line above), so I think we are OK here.

fujiisoup:
I just missed this line. 
Thanks. 
LGTM.

PR #2163, opened 2018-05-18
Versioneer

### LINKED ISSUE #1300: git version label yields version in violation of PEP 440

jhamman:
- [x] Closes #1300 (in a more portable way)

This eliminates the need to edit `setup.py` before / after release and is a nice step towards simplifying xarray's release process.


doc/conf.py
```diff
@@ -102,7 +102,7 @@
 # built documents.
 #
 # The short X.Y version.
-version = xarray.version.short_version
+version = xarray.__version__.split('+')[0]
```
jhamman: this is the one change that worries me. It is possible people were relying on the `xarray.version` module...

dopplershift:
Versioneer has worked great for us. Cutting a release is triggered just by making a new release on GitHub.

shoyer:
Very nice!

I tested this out locally and it seems to properly with picking up new tags 👍

doc/conf.py
shoyer: If they were... then they will have to fix up their code.

Issue #2173, opened 2018-05-22
Formatting error in conjunction with pandas.DataFrame


Hoeze:
#### Code Sample, a copy-pastable example if possible

```python
import pandas as pd
import numpy as np
import xarray as xr

sample_data = np.random.uniform(size=[2,2000,10000])
x = xr.Dataset({"sample_data": (sample_data.shape, sample_data)})
print(x)

df = pd.DataFrame({"x": [1,2,3], "y": [2,4,6]})
x["df"] = df
print(x)
```
![grafik](https://user-images.githubusercontent.com/1200058/40392190-0e333486-5e1b-11e8-9fcb-08b05eb40cb5.png)

#### Problem description
Printing a xarray.Dataset results in an error when containing a pandas.DataFrame

#### Expected Output
Should print string representation of Dataset

#### Output of ``xr.show_versions()``

<details>
/usr/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.16.9-1-ARCH
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: de_DE.UTF-8
LOCALE: de_DE.UTF-8

xarray: 0.10.4
pandas: 0.22.0
numpy: 1.14.3
scipy: 1.0.1
netCDF4: None
h5netcdf: 0.5.1
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: None
distributed: None
matplotlib: 2.2.2
cartopy: None
seaborn: None
setuptools: 39.2.0
pip: 10.0.1
conda: None
pytest: None
IPython: 6.3.1
sphinx: None

</details>

shoyer:
This is a manifestation of the same issue as https://github.com/pydata/xarray/issues/2172: you are using integers as dimension names, which we don't support very well.

In this case, you end up with mixed integers/strings as dimension names, which means our internal use of `sorted()` in `repr(x)` fails.

max-sixty:
Closed by #2174

keewis:
[...]

Issue #1231, opened 2017-01-26
Consider accepting a positional dict argument in place of keyword arguments


shoyer:
Using `**kwargs` in xarray operations is incredibly convenient, but it means there is no way to use a keyword argument that conflicts with an existing argument name. For example, there is no way to use `.sel` to select along `'method'`, `'tolerance'` or `'drop'` dimensions.

For interactive use, this is fine -- these reserved names are rarely used and it's nice to save the keystrokes it takes to write `dict()`. But it's a problem for writing reliable complex software, which may be calling xarray methods from somewhere very different. In fact, knowing how this works, it's possible to trigger bugs without even using `**kwargs`, e.g., by indexing a DataArray:
```
In [34]: array = xr.DataArray([1, 2, 3], dims='drop')

# result should be a scalar!
In [35]: array[0]
Out[35]:
<xarray.DataArray (drop: 3)>
array([1, 2, 3])
Unindexed dimensions:
    drop
```

One option to resolve this is to make the first argument to every function like this (including `sel`, `isel`, `sel_points`, `isel_points`, `stack`, ...) accept a dictionary, which is interpreted exactly like `**kwargs`. In fact, we already do this in a highly inconsistent fashion for [`Dataset.reindex`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.reindex.html) only. For all cases where axis names are set dynamically (and certainly within xarray), we could encourage using the dictionary form, e.g., `array.isel(kwargs, drop=True)` rather than `array.isel(drop=True, **kwargs)`.

max-sixty:
I missed this issue and last year created https://github.com/pydata/xarray/issues/2188. I'll leave that one open as it has tracking
The good news is that it's mostly complete, mainly due to @fujiisoup

PR #2174, opened 2018-05-23
Datasets more robust to non-string keys

### LINKED ISSUE #2172: Errors on pycharm completion
```python
# execute:
import numpy as np
import xarray as xr

sample_data = np.random.uniform(size=[2,2000,10000])
x = xr.Dataset({"sample_data": (sample_data.shape, sample_data)})

x2 = x["sample_data"]

# now type by hand:
x2.

```
grafik

#### Problem description
I'm not completely sure if it's a xarray problem, but each time I enter [some dataset]**.** (note the point) inside PyCharm's python console, I get a python exception instead of some autocompletion.

#### Expected Output
grafik

<details>
/usr/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.16.9-1-ARCH
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: de_DE.UTF-8
LOCALE: de_DE.UTF-8

xarray: 0.10.4
pandas: 0.22.0
numpy: 1.14.3
scipy: 1.0.1
netCDF4: None
h5netcdf: 0.5.1
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: None
distributed: None
matplotlib: 2.2.2
cartopy: None
seaborn: None
setuptools: 39.2.0
pip: 10.0.1
conda: None
pytest: None
IPython: 6.3.1
sphinx: None
</details>

maxim-lian:
- [x] Closes #2172 & #2173 
 - [x] Tests added 
 - [x] Tests passed 

I don't think this is the most efficient way of doing this, though it does work. Any ideas for a more efficient implementation?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,6 +8,7 @@ testpaths=xarray/tests
 [flake8]
 max-line-length=79
 ignore=
+  W503
 exclude=
   doc/
 
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -785,7 +785,7 @@ class DataArray(AbstractArray, DataWithCoords):
             indexers=indexers, drop=drop, method=method, tolerance=tolerance)
         return self._from_temp_dataset(ds)
 
-    def isel_points(self, indexers=None, dim='points', **indexers_kwargs):
+    def isel_points(self, dim='points', **indexers):
         """Return a new DataArray whose dataset is given by pointwise integer
         indexing along the specified dimension(s).
 
@@ -793,13 +793,11 @@ class DataArray(AbstractArray, DataWithCoords):
         --------
         Dataset.isel_points
         """
-        indexers = combine_pos_and_kw_args(
-            indexers, indexers_kwargs, 'isel_points')
         ds = self._to_temp_dataset().isel_points(dim=dim, **indexers)
         return self._from_temp_dataset(ds)
 
-    def sel_points(self, indexers=None, dim='points', method=None,
-                   tolerance=None, **indexers_kwargs):
+    def sel_points(self, dim='points', method=None, tolerance=None,
+                   **indexers):
         """Return a new DataArray whose dataset is given by pointwise selection
         of index labels along the specified dimension(s).
 
@@ -807,8 +805,6 @@ class DataArray(AbstractArray, DataWithCoords):
         --------
         Dataset.sel_points
         """
-        indexers = combine_pos_and_kw_args(
-            indexers, indexers_kwargs, 'sel_points')
         ds = self._to_temp_dataset().sel_points(
             dim=dim, method=method, tolerance=tolerance, **indexers)
         return self._from_temp_dataset(ds)
@@ -860,12 +856,18 @@ class DataArray(AbstractArray, DataWithCoords):
         return self.reindex(method=method, tolerance=tolerance, copy=copy,
                             **indexers)
 
-    def reindex(self, method=None, tolerance=None, copy=True, **indexers):
+    def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
+                **indexers_kwargs):
         """Conform this object onto a new set of indexes, filling in
         missing values with NaN.
 
         Parameters
         ----------
+        **indexers : dict
+            Dictionary with keys given by dimension names and values given by
+            arrays of coordinates tick labels. Any mis-matched coordinate
+            values will be filled in with NaN, and any mis-matched dimension
+            names will simply be ignored.
         copy : bool, optional
             If ``copy=True``, data in the return value is always copied. If
             ``copy=False`` and reindexing is unnecessary, or can be performed
@@ -883,11 +885,8 @@ class DataArray(AbstractArray, DataWithCoords):
             Maximum distance between original and new labels for inexact
             matches. The values of the index at the matching locations most
             satisfy the equation ``abs(index[indexer] - target) <= tolerance``.
-        **indexers : dict
-            Dictionary with keys given by dimension names and values given by
-            arrays of coordinates tick labels. Any mis-matched coordinate
-            values will be filled in with NaN, and any mis-matched dimension
-            names will simply be ignored.
+        **indexers_kwargs : {dim: indexer, ...}
+            The keyword arguments form of ``indexers``
 
         Returns
         -------
@@ -900,8 +899,10 @@ class DataArray(AbstractArray, DataWithCoords):
         DataArray.reindex_like
         align
         """
+        indexers = combine_pos_and_kw_args(
+            indexers, indexers_kwargs, 'reindex')
         ds = self._to_temp_dataset().reindex(
-            method=method, tolerance=tolerance, copy=copy, **indexers)
+            indexers=indexers, method=method, tolerance=tolerance, copy=copy)
         return self._from_temp_dataset(ds)
 
     def rename(self, new_name_or_name_dict):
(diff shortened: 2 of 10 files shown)


setup.cfg
```diff
@@ -8,6 +8,7 @@ testpaths=xarray/tests
 [flake8]
 max-line-length=79
 ignore=
+  W503
```
maxim-lian: This is line break before operator, which has been changed in PEP8

xarray/core/dataarray.py
```diff
@@ -466,7 +466,17 @@ def _getitem_coord(self, key):
         return self._replace_maybe_drop_dims(var, name=key)
 
     def __getitem__(self, key):
-        if isinstance(key, basestring):
+        try:
+            is_coord_key = any([
+                isinstance(key, basestring),
+                key in set(self.dims).union(self.coords)
```
shoyer: I would hesitate to do this fix -- it means `array[2]` could pull out a coordinate if `2` is the name of a dimension. More generally, `DataArray.__getitem__` is only meant as a convenient short-cut for getting coordinates, which we might even want to deprecate entirely. `DataArray.coords.__getitem__` is the reliable way to get a coordinate out.

Instead, I think the better option would be to avoid making use of `**` unpacking below, in the expression `self.isel(**self._item_key_to_dict(key))`. This would be a little more involved, but I think the right fix would be to make most of these functions accept a positional dictionaries instead of `**kwargs`, as suggested in https://github.com/pydata/xarray/issues/1231.

maxim-lian: Yes, completely agree, thanks

maxim-lian:
I added an initial attempt at allowing for supplying a dict; on `isel` only at first

Let me know ppl's thoughts:
- Arg name?
- Combine or raise if supplied with kwargs?
- Is there an encapsulation leak? Should we have a user API layer which is then translated to a more robust representation (i.e. kwargs to a dict), and those be separate methods? My initial inclination is that the current approach is good and managable

maxim-lian:
> `DataArray.__getitem__` is only meant as a convenient short-cut for getting coordinates, which we might even want to deprecate entirely

👍

shoyer:
> Arg name?

I'd like a convention of matching names to make it clear in docstrings that these are the same thing, e.g.,
- `something_dict`/`**something`, or
- `something`/`**something_kwargs`

Probably the last is best since users will never have cause to type the longer argument name ending with `_kwargs`.

In practice, I suspect these first arguments are almost always going to be used positionally. We might even imagine making them positional only argument if [PEP 570](https://www.python.org/dev/peps/pep-0570/) is ever accepted. So I don't think it matters too much.

> Combine or raise if supplied with kwargs?

Let's raise (like your current implementation) if both are provided. Combining could be error prone and seems unnecessarily complicated.

This is the same logic I implemented in the somewhat misleadingly named `utils.combine_pos_and_kw_args()` utility function (used by `Dataset.reindex()`).

> Is there an encapsulation leak? Should we have a user API layer which is then translated to a more robust representation (i.e. kwargs to a dict), and those be separate methods? My initial inclination is that the current approach is good and managable

I agree -- I think the current version is manageable.

For internal use, we should always use the positional argument, but there's not much harm in leaving `**kwargs` around.

maxim-lian:
AppVeyor failure unrelated

xarray/core/dataset.py
```diff
@@ -1404,12 +1406,16 @@ def isel(self, drop=False, **indexers):
         Dataset.sel
         DataArray.isel
         """
+
+        indexers = combine_pos_and_kw_args(indexers, indexers_kwargs, 'isel')
+        assert isinstance(drop, bool)
```
shoyer: This should either be dropped for now, or raise `TypeError`. `assert` isn't appropriate for external APIs.

xarray/core/dataarray.py
```diff
@@ -851,12 +856,18 @@ def reindex_like(self, other, method=None, tolerance=None, copy=True):
...
 
-    def reindex(self, method=None, tolerance=None, copy=True, **indexers):
+    def reindex(self, indexers=None, method=None, tolerance=None, copy=True,
+                **indexers_kwargs):
         """Conform this object onto a new set of indexes, filling in
         missing values with NaN.
 
         Parameters
         ----------
+        **indexers : dict
```
shoyer: shouldn't start with `**`

xarray/core/dataarray.py
```diff
@@ -18,7 +18,9 @@
 from .formatting import format_item
 from .options import OPTIONS
 from .pycompat import OrderedDict, basestring, iteritems, range, zip
-from .utils import decode_numpy_dict_values, ensure_us_time_resolution
+from .utils import (
+    combine_pos_and_kw_args, decode_numpy_dict_values,
```
shoyer: While we're at it, maybe we should rename `combine_pos_and_kw_args` to something clearer, maybe `either_dict_or_kwargs`?

xarray/core/dataarray.py
```diff
@@ -874,11 +885,8 @@ def reindex(self, method=None, tolerance=None, copy=True, **indexers):
             Maximum distance between original and new labels for inexact
             matches. The values of the index at the matching locations most
             satisfy the equation ``abs(index[indexer] - target) <= tolerance``.
-        **indexers : dict
-            Dictionary with keys given by dimension names and values given by
-            arrays of coordinates tick labels. Any mis-matched coordinate
-            values will be filled in with NaN, and any mis-matched dimension
-            names will simply be ignored.
+        **indexers_kwargs : {dim: indexer, ...}
+            The keyword arguments form of ``indexers``
```
shoyer: We should add something like: Only `indexers` or `indexer_kwargs` may be provided.

shoyer:
Looking through Dataset/DataArray methods that support `**kwargs`, I think this may be the full list:
- [ ] stack
- [ ] shift
- [ ] roll
- [ ] set_index
- [ ] reorder_levels
- [ ] rolling
- [ ] resample (not yet, we still support old behavior for the first positional arguments with a warning)

shoyer:
But I think just supporting this for `isel/`sel`/`reindex` for now would be enough to be valuable. Note that currently it's on Dataset.reindex() but not DataArray.reindex().

xarray/core/dataset.py
```diff
@@ -1444,6 +1452,15 @@ def sel(self, method=None, tolerance=None, drop=False, **indexers):
 
         Parameters
         ----------
+        ----------
```
shoyer: delete

xarray/core/utils.py
```diff
@@ -183,7 +183,7 @@ def is_full_slice(value):
     return isinstance(value, slice) and value == slice(None)
 
 
-def combine_pos_and_kw_args(pos_kwargs, kw_kwargs, func_name):
+def either_dict_or_kwargs(pos_kwargs, kw_kwargs, func_name):
```
shoyer: We don't need to test every use, but we should at least add unit test for this helper function -- I don't think we have any test coverage currently.

shoyer:
thanks @maxim-lian !

maxim-lian:
thanks for working through the feedback, as ever, @shoyer ! 
It's like free coding classes

PR #2189, opened 2018-05-27
docs: add xyzpy to projects

jcmgray:
Justs adds [xyzpy](http://xyzpy.readthedocs.io) to the list of projects utilizing ``xarray``.


shoyer:
Looks great, thanks!

(And thanks for writing xyzpy, it looks pretty handy!)

PR #2193, opened 2018-05-28
Deprecated paths in tests

maxim-lian:
Responding to a couple of warnings


shoyer:
These fixes are also in #2162 -- any chance I could tempt you to review that PR instead? :)

Issue #1603, opened 2017-10-04
Future of MultiIndex


fujiisoup:
I think we can continue the discussion we have in #1426 about `MultiIndex` here.

In [comment](https://github.com/pydata/xarray/pull/1426#issuecomment-304778433) , @shoyer recommended to remove `MultiIndex` from public API.

I agree with this, as long as my codes work with this improvement.

I think if we could have a list of possible `MultiIndex` use cases here, 
it would be easier to deeply discuss and arrive at a consensus of the future API.

fujiisoup:
I'm using `MultiIndex` a lot, 
but I noticed that it is just a workaround to index along multiple kinds of coordinate.

Consider the following example,
```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: da = xr.DataArray(np.arange(5), dims=['x'],
   ...:                   coords={'experiment': ('x', [0, 0, 0, 1, 1]),
   ...:                           'time': ('x', [0.0, 0.1, 0.2, 0.0, 0.15])})
   ...: 

In [2]: da
Out[2]: 
<xarray.DataArray (x: 5)>
array([0, 1, 2, 3, 4])
Coordinates:
    experiment  (x) int64 0 0 0 1 1 
    time        (x) float64 0.0 0.1 0.2 0.0 0.15
Dimensions without coordinates: x

```

I want to do something like this
```python
da.sel(experiment=0).sel(time=0.1)
```
but it cannot. 
MultiIndexing enables this,

```python
In [2]: da = da.set_index(exp_time=['experiment', 'time'])
   ...: da
   ...: 
Out[2]: 
<xarray.DataArray (x: 5)>
array([0, 1, 2, 3, 4])
Coordinates:
  * exp_time    (exp_time) MultiIndex
  - experiment  (exp_time) int64 0 0 0 1 1 
  - time        (exp_time) float64 0.0 0.1 0.2 0.0 0.15
Dimensions without coordinates: x
```

If we could make a selection from a non-index coordinate, 
`MultiIndex` is not necessary for this case.

I think there should be other important usecases of `MultiIndex`.
I would be happy if anyone could list them in this issue.

shoyer:
One API design challenge here is that I think we still want a explicit notation of "indexed" variables. We could *possibly* allow operations like `.sel()` on non-indexed variables, but they would be slower, because we would not want to create expensive hash-tables (i.e., `pandas.Index`) in a non-transparent fashion.

shoyer:
[...]

PR #2195, opened 2018-05-29
WIP: explicit indexes

shoyer:
Some utility functions that should be useful for https://github.com/pydata/xarray/issues/1603

Still very much a work in progress -- it would be great if someone has time to finish writing any of these in another PR!

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -114,14 +114,7 @@ def as_variable(obj, name=None):
     return obj
 
 
-def as_index_or_compatible_data(data):
-    if isinstance(data, pd.Index):
-        return data
-    else:
-        return as_compatible_data(data)
-
-
-def as_variables_with_multiindex_expansion(obj, name):
+def maybe_expand_multiindex(obj, name):
     """Expand an object into one or more Variable objects.
 
     Parameters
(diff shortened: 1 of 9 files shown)


xarray/core/coordinates.py
```diff
@@ -314,6 +315,71 @@ def __unicode__(self):
         return formatting.indexes_repr(self)
 
 
+def normalize_indexes(indexes, coords, sizes):
```
maxim-lian: (will need `=None`)

xarray/core/coordinates.py
```diff
@@ -314,6 +315,71 @@ def __unicode__(self):
...
+
+    Returns
+    -------
+    Mapping[Any, pandas.Index] mapping indexing keys (levels/dimension names)
+    to indexes used for indexing along that dimension.
+    """
+    indexes = {} if indexes is None else dict(indexes)
+
+    # default indexes
+    for key in sizes:
```
maxim-lian: Do we want to allow the sizes to be optional, and drawn from the indexes / coords? At the moment they must be supplied in `sizes`

xarray/core/variable.py
```diff
@@ -117,6 +117,88 @@ def as_variable(obj, name=None):
...
+
+        if isinstance(data, pd.MultiIndex):
+            raise NotImplementedError('TODO: expand MultiIndex objects.')
+
+        if (name,) == dims or isinstance(data, pd.Index):
+            variable = IndexVariable(dims, data, *args, fastpath=True)
+        else:
+            variable = Variable(dims, data, *args, fastpath=True)
+
+    return OrderedDict([(name, variable)])
```
maxim-lian: 👍

xarray/core/coordinates.py
shoyer: I was imagining calling this function *after* the Dataset/DataArray is filled in with fully specified `xarray.Variable` objects, so we would already have determined sizes for all dimensions.

maxim-lian: Perfect, that makes sense

maxim-lian: Eventually (and I realize from the other PR this is a sketch atm) then we likely want to check the sizes are consistent if they're also supplied in `indexes` or `coords`

xarray/core/coordinates.py
fujiisoup: I'm trying to follow this PR up, but not yet sure what this utility function does.

Can you suggest an example usage of this?
What I currently think of is 
+ Without MultiIndex
```python
    coords = OrderedDict()
    coords['x'] = Variable(('x', ), [0, 1, 2])
    coords['y'] = Variable(('y', ), ['a', 'b'])
    sizes = {'x': 3, 'y': 2}
    actual = normalize_indexes({}, coords, sizes)
    expected = {'x': coords['x'].to_index(), 'y': coords['y'].to_index()}
```
+ With MultiIndex
```python
    coords = OrderedDict()
    coords['x'] = Variable(('z', ), [0, 1, 2])
    coords['y'] = Variable(('z', ), ['a', 'b'])
    sizes = {'x': 6}
    actual = normalize_indexes({}, coords, sizes)
    expected = {'x': pd.Index([0, 0, 1, 1, 2, 2]),
                'y': pd.Index(['a', 'b', 'a', 'b', 'a', 'b'])}
```
Or do we need a `MultiIndex` for `'z'` ?

shoyer: Your "without MultiIndex" example look correct, but I think your "with MultiIndex" case should raise an error, because the dimension "z" has inconsistent sizes.

I was thinking something more like:
```
    indexes = {
        'x': pandas.Index([0, 0, 1, 1, 2, 2]),
        'y': pandas.Index(['a', 'b', 'a', 'b', 'a', 'b']),
    }
    coords = {
        'x': Variable(('z',), [0, 0, 1, 1, 2, 2]),
        'y': Variable(('z',), ['a', 'b', 'a', 'b', 'a', 'b']),
    }
    sizes = {'z': 6}
    actual = normalize_indexes(indexes, coords, sizes)
    combined_index = pandas.MultiIndex.from_arrays([indexes['x'], indexes['y']], names=['x', 'y'])
    expected = {'x': combined_index, 'y': combined_index, 'z': combined_index}
```

The resulting object would supporting indexing with any of `.sel(x=...)`, `.sel(y=...)` or `.sel(z=...)`, all based on `combined_index`.

Issue #2203, opened 2018-05-30
Update minimum version of dask


jhamman:
Xarray currently states that it supports dask version 0.9 and later. However, 1) I don't think this is true and my quick test shows that some of our tests fail using dask 0.9, and 2) we have a growing number of tests that are being skipped for older dask versions:

```
$ grep -irn "dask.__version__" xarray/tests/*py
xarray/tests/__init__.py:90:    if LooseVersion(dask.__version__) < '0.18':
xarray/tests/test_computation.py:755:        if LooseVersion(dask.__version__) < LooseVersion('0.17.3'):
xarray/tests/test_computation.py:841:    if not use_dask or LooseVersion(dask.__version__) > LooseVersion('0.17.4'):
xarray/tests/test_dask.py:211:    @pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_dask.py:223:    @pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_dask.py:284:    @pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_dask.py:296:    @pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_dask.py:387:        if LooseVersion(dask.__version__) == LooseVersion('0.15.3'):
xarray/tests/test_dask.py:784:    pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_dask.py:802:    pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_dask.py:818:@pytest.mark.skipif(LooseVersion(dask.__version__) <= '0.15.4',
xarray/tests/test_variable.py:1664:        if LooseVersion(dask.__version__) <= LooseVersion('0.15.1'):
xarray/tests/test_variable.py:1670:        if LooseVersion(dask.__version__) <= LooseVersion('0.15.1'):
```

I'd like to see xarray bump the minimum version number of dask to something around 0.15.4 (Oct. 2017) or 0.16 (Nov. 2017).

cc @mrocklin, @pydata/xarray

fujiisoup:
`DataArray.dot()` is using dask.einsum, and it only woks with dask >= 0.17.5 (#2089).

Does this requirement sound is too strict? If so, we maybe better to add another pass for older dask.

[...]

PR #2205, opened 2018-05-31
Support dot with older dask

fujiisoup:
- [x] Related with #2203 
 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented

Related with #2203, I think it is better if `xr.DataArray.dot()` is working even with older dask, at least in the simpler case (as this is a very primary operation).

The cost is a slight complication of the code.
Any comments are welcome.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -38,7 +38,7 @@ Enhancements
 ~~~~~~~~~~~~
 
 - `:py:meth:`~DataArray.dot` and :py:func:`~dot` are partly supported with older
-  dask<01.7.4. (related to :issue:`2203`)
+  dask<0.17.4. (related to :issue:`2203`)
   By `Keisuke Fujii <https://github.com/fujiisoup`_.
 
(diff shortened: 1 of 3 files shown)


xarray/tests/test_computation.py
```diff
@@ -784,6 +781,10 @@ def test_dot(use_dask):
     assert (actual.data == np.einsum('ij,ijk->k', a, b)).all()
     assert isinstance(actual.variable.data, type(da_a.variable.data))
 
+    import dask
```
shoyer: This needs to be guarded behind `use_dask`. Currently the tests without dask installed are failing here on Travis-CI.

doc/whats-new.rst
```diff
@@ -37,6 +37,10 @@ Documentation
 Enhancements
 ~~~~~~~~~~~~
 
+- `:py:meth:`~DataArray.dot` and :py:func:`~dot` are partly supported with older
+  dask<01.7.4. (related to :issue:`2203`)
```
shoyer: 01.7.4 -> 0.17.4

PR #2210, opened 2018-06-01
Remove height=12in from facetgrid example plots.

### LINKED ISSUE #2208: Poorly stretched plots in plotting docs with latest sphinx-rtd-theme
This is now showing up on ReadTheDocs:
http://xarray.pydata.org/en/latest/plotting.html#facetgrid-objects

Example:
screen shot 2018-06-01 at 8 45 06 am

dcherian:
- [x] Closes #2208  (remove if there is no corresponding issue, which should only be the case for minor changes)

This fixes it for me locally. The height was forced to be 12in, while width is 100%. 

I'm not sure why this was added in the first place.

### DIFF
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -436,7 +436,7 @@ arguments to the xarray plotting methods/functions. This returns a
 
 .. ipython:: python
 
-    @savefig plot_facet_dataarray.png height=12in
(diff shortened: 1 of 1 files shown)


shoyer:
Thanks!

I think that was just an arbitrary choice to shrink the size of some plot images, which apparently broke with the latest RTD style.

PR #2222, opened 2018-06-09
implement interp_like

### LINKED ISSUE #2218: interp_like
Just as a reminder of the remaining extension of #2104 .

We might add `interp_like` that behaves like `reindex_like` but using `interp()`.

fujiisoup:
- [x] Closes #2218
 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API.

This adds `interp_like`, that behaves like `reindex_like` but using interpolation.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -951,6 +951,52 @@ class DataArray(AbstractArray, DataWithCoords):
             **coords_kwargs)
         return self._from_temp_dataset(ds)
 
+    def interp_like(self, other, method='linear', assume_sorted=False,
+                    kwargs={}):
+        """Interpolate this object onto the coordinates of another object,
+        filling out of range values with NaN.
+
+        Parameters
+        ----------
+        other : Dataset or DataArray
+            Object with an 'indexes' attribute giving a mapping from dimension
+            names to an 1d array-like, which provides coordinates upon
+            which to index the variables in this dataset.
+        method: string, optional.
+            {'linear', 'nearest'} for multidimensional array,
+            {'linear', 'nearest', 'zero', 'slinear', 'quadratic', 'cubic'}
+            for 1-dimensional array. 'linear' is used by default.
+        assume_sorted: boolean, optional
+            If False, values of coordinates that are interpolated over can be
+            in any order and they are sorted first. If True, interpolated
+            coordinates are assumed to be an array of monotonically increasing
+            values.
+        kwargs: dictionary, optional
+            Additional keyword passed to scipy's interpolator.
+
+        Returns
+        -------
+        interpolated: xr.DataArray
+            Another dataarray by interpolating this dataarray's data along the
+            coordinates of the other object.
+
+        Note
+        ----
+        scipy is required.
+
+        See Also
+        --------
+        DataArray.interp
+        DataArray.reindex_like
+        """
+        if self.dtype.kind not in 'uifc':
+            raise TypeError('interp only works for a numeric type array. '
+                            'Given {}.'.format(self.dtype))
+
+        ds = self._to_temp_dataset().interp_like(
+            other, method=method, kwargs=kwargs, assume_sorted=assume_sorted)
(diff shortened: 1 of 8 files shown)


fujiisoup:
Slightly wondering the case when an object has non-numeric dimensions.
The current implementation raises TypeError, even when the target destination is just a subset of the current one.

shoyer:
> Slightly wondering the case when an object has non-numeric dimensions.
The current implementation raises TypeError, even when the target destination is just a subset of the current one.

Do you have an example?

It might make sense to interpolate on numeric dimensions, and directly reindex on non-numeric dimensions.

fujiisoup:
> Do you have an example?

We raise a TypeError in the following example, because the dataset has datetime-type dimension.

```python
In [3]: ds = xr.tutorial.load_dataset('air_temperature')

In [4]: ds
Out[4]: 
<xarray.Dataset>
Coordinates:
  * time     (time) datetime64[ns] 2013-01-01 2013-01-01T06:00:00 ...
Data variables:
    air      (time, lat, lon) float32 241.2 242.5 243.5 244.0 244.09999 ...
Attributes:
    Conventions:  COARDS
    platform:     Model

In [5]: ds.interp_like(ds.isel(lat=[0, 1, 2]))  # -> TypeError
```

> It might make sense to interpolate on numeric dimensions, and directly reindex on non-numeric dimensions.

I agree for the object type dimension, such as string, reindex is be appropriate and users might expect it.
But for datetime dimension, I think users expect the interpolation rather than reindexing.

xarray/core/dataset.py
```diff
@@ -1893,6 +1893,58 @@ def maybe_variable(obj, k):
...
+        numeric_coords = OrderedDict()
+        object_coords = OrderedDict()
+        for k, v in coords.items():
+            if v.dtype.kind in 'uifcMm':
+                numeric_coords[k] = v
+            else:
+                object_coords[k] = v
+        # We do not support interpolation along object coordinate.
+        # reindex instead.
+        ds = self.reindex(object_coords)
```
shoyer: Reindexing makes a copy of the dataset, so let's only do this if there are object coordinates.

xarray/core/missing.py
```diff
@@ -395,6 +395,24 @@ def _localize(var, indexes_coords):
...
+    x = list(x)
+    new_x = list(new_x)
+    for i in range(len(x)):
+        if x[i].dtype.kind in 'Mm':
+            # Scipy cast np.float_ for coordinate, but 64bit float is not
+            # sufficient for datetime64 (uses 64bit integer).
+            # We assumes x - min(x) can be well presented by 64bit float.
+            xmin = np.min(x[i])
+            x[i] = (x[i] - xmin).astype(np.float_)
+            new_x[i] = (new_x[i] - xmin).astype(float)
```
shoyer: Can we use the same `astype(np.float64)` on both these lines?

xarray/tests/test_interp.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+import pandas as pd
```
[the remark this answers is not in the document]
shoyer: looks like we can remove this?

xarray/core/dataarray.py
```diff
@@ -951,6 +951,52 @@ def interp(self, coords=None, method='linear', assume_sorted=False,
             **coords_kwargs)
         return self._from_temp_dataset(ds)
 
+    def interp_like(self, other, method='linear', assume_sorted=False,
+                    kwargs={}):
+        """Interpolate this object onto the coordinates of another object,
+        filling out of range values with NaN.
+
```
shoyer: Can we add a note that if there are coordinates with object dtype, these are reindexed instead, not interpolated?

fujiisoup:
Thanks for the review :) merged.

PR #2229, opened 2018-06-13
Bugfix for faceting line plots.

### LINKED ISSUE #2239: Error in docs/plottings
There is an error on rtd.
http://xarray.pydata.org/en/stable/plotting.html#id4

dcherian:
Closes #2239 

Fixes a broken doc image: http://xarray.pydata.org/en/stable/plotting.html#id4

The tests passed previously because there was no metadata associated with the test DataArray. I've assigned some now, that should be good enough.

### DIFF
--- a/xarray/plot/facetgrid.py
+++ b/xarray/plot/facetgrid.py
@@ -297,10 +297,12 @@ class FacetGrid(object):
             darray=self.data.loc[self.name_dicts.flat[0]],
             x=x, y=y, hue=hue)
 
+        self._hue_var = hueplt
+        self._hue_label = huelabel
         self._finalize_grid(xlabel, ylabel)
 
         if add_legend and hueplt is not None and huelabel is not None:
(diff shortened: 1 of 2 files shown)


xarray/plot/facetgrid.py
```diff
@@ -314,10 +314,10 @@ def _finalize_grid(self, *axlabels):
             if namedict is None:
                 ax.set_visible(False)
 
-    def add_line_legend(self, huelabel):
+    def _add_line_legend(self, hueplt, huelabel):
```
shoyer: Do we want this method to be private? Or maybe call it just `add_legend` for consistency with `add_colorbar` and seaborn?

dcherian: I made it private because it wasn't usable without giving it data. Fixed that now.

dcherian:
The failing test is the one documented in #2238

PR #2236, opened 2018-06-18
Refactor nanops

### LINKED ISSUE #2230: Inconsistency between Sum of NA's and Mean of NA's: resampling gives 0 or 'NA'
#### Problem description

For datamining with xarray there is always the following issue with the resampling-method.  
If i resample e.g. a daily timeseries over one month and if the data are 'NA' at each day, I get zero as a result. That is annoying considering a timeseries of precipitation. **It is definitely a difference if the monthly precipitation is zero for one month (each day zero precipitation) or the monthly precipitation was not measured due to problems with the device (each day NA)** 

#### Data example

I have a dataset with hourly values for 5 month 'fcut'.
```python
<xarray.Dataset>
Dimensions:       (bnds: 2, time: 3672)
Coordinates:
    rlon          float32 22.06
    rlat          float32 5.06
  * time          (time) datetime64[ns] 2006-05-01 2006-05-01T01:00:00 ...
Dimensions without coordinates: bnds
Data variables:
    rotated_pole  int32 1
    time_bnds     (time, bnds) float64 1.304e+07 1.305e+07 1.305e+07 ...
    TOT_PREC      (time) float64 nan nan nan nan nan nan nan nan nan nan nan ...
Attributes:
```
Doing a resample process gives only zero values for each month.

```
In [10]: fcut.resample(dim='time',freq='M',how='sum')
Out[10]: 
<xarray.Dataset>
Dimensions:       (bnds: 2, time: 5)
Coordinates:
  * time          (time) datetime64[ns] 2006-05-31 2006-06-30 2006-07-31 ...
Dimensions without coordinates: bnds
Data variables:
    rotated_pole  (time) int64 1 1 1 1 1
    time_bnds     (time, bnds) float64 1.07e+10 1.07e+10 1.225e+10 1.225e+10 ...
    TOT_PREC      (time) float64 0.0 0.0 0.0 0.0 0.0
```
**But I expect to have NA for each month, as it is the case for the operator 'mean'**

I know that there is an ongoing discussion about that topic (see for example  https://github.com/pandas-dev/pandas/issues/9422).

For earth science it would be nice to have an option telling xarray what to do in case of a sum over values being all NA. Do you see a chance to have a fast fix for that issue in the model code?

fujiisoup:
- [x] Closes #2230
 - [x] Tests added
 - [x] Tests passed

In #2230, the addition of `min_count` keywords for our reduction methods was discussed, but 
our `duck_array_ops` module is becoming messy (mainly due to nan-aggregation methods for dask, bottleneck and numpy) and it looks a little hard to maintain them.

I tried to refactor them by moving nan-aggregation methods to `nanops` module.

I think I still need to take care of more edge cases, but I appreciate any comment for the current implementation.

Note:
In my implementation, **bottleneck is not used when `skipna=False`**.
bottleneck would be advantageous when `skipna=True` as numpy needs to copy the entire array once, but I think numpy's method is still OK if `skipna=False`.


xarray/tests/test_variable.py
```diff
@@ -1508,8 +1508,8 @@ def test_reduce_funcs(self):
         assert_identical(v.all(dim='x'), Variable([], False))
 
         v = Variable('t', pd.date_range('2000-01-01', periods=3))
-        with pytest.raises(NotImplementedError):
-            v.argmax(skipna=True)
```
fujiisoup: Does anyone remember why we did not implement `nanargmax` for `pd.date_range`? It looks `np.nanargmax` takes care of this dtype.

shoyer:
Very nice!

> In my implementation, bottleneck is not used when skipna=False.
bottleneck would be advantageous when skipna=True as numpy needs to copy the entire array once, but I think numpy's method is still OK if skipna=False.

I think this is correct -- bottleneck does not speed up non-NaN skipping functions.

xarray/core/duck_array_ops.py
```diff
@@ -296,76 +223,65 @@ def f(values, axis=None, skipna=None, **kwargs):
...
-                    assert using_numpy_nan_func
-                    msg = ('%s is not available with skipna=False with the '
-                           'installed version of numpy; upgrade to numpy 1.12 '
-                           'or newer to use skipna=True or skipna=None' % name)
-                raise NotImplementedError(msg)
-    f.numeric_only = numeric_only
+            if dtype is None:
+                func = _dask_or_eager_func(name)
+            else:
+                func = getattr(np_module, name)
```
shoyer: This looks a little dangerous -- it could silently convert a dask array into a NumPy array. Can we raise an error instead?

xarray/core/nanops.py
```diff
@@ -0,0 +1,278 @@
...
+        # object arrays do not support `isnan` (gh-9009), so make a guess
+        mask = a != a
+    elif issubclass(a.dtype.type, np.inexact):
+        mask = np.isnan(a)
+    else:
+        mask = None
+
+    if mask is not None:
+        if isinstance(a, dask_array_type):
+            return dask_array.where(mask, val, a), mask
```
shoyer: This is a little surprising to me. I would rather use `duck_array_ops.where()` here than use explicit type checks.

Likewise, instead of logic above for `mask`, use `duck_array_ops.isnull()`

xarray/core/duck_array_ops.py
fujiisoup: It looks a kind of holdover.
Removed.

xarray/core/nanops.py
fujiisoup: Done.

fujiisoup:
> I think it would make sense to restructure this a little bit to have two well defined layers:

> A module of bottleneck/numpy functions that act on numpy arrays only.
> A module of functions that act on numpy or dask arrays (or these could be moved into duck_array_ops).

Could you explain more detail about this idea?

shoyer:
>> A module of bottleneck/numpy functions that act on numpy arrays only.
>>A module of functions that act on numpy or dask arrays (or these could be moved into duck_array_ops).

> Could you explain more detail about this idea?

OK, let me try:

1. On numpy arrays, we use bottleneck eqiuvalents of numpy functions when possible because bottleneck is faster than numpy
2. On dask arrays, we use dask equivalents of numpy functions.
3. We also want to add some extra features on top of what numpy/dask/bottleneck provide, e.g., handling of `min_count`

We could implement this with:
- `nputils.nansum()` is equivalent to `numpy.nansum()` but uses `bottleneck.nansum()` internally instead when possible.
- `duck_array_ops.nansum()` uses `numpy_nansum()` or `dask.array.nansum()`, based upon the type of the inputs.
- `duck_array_ops.sum()` uses `numpy.sum()` or `dask.array.sum()`, based upon the type of the inputs.
- `duck_array_ops.sum_with_mincount()` adds `mincount` and `skipna` support and is used in the `Dataset.sum()` implementation. Its is written using `duck_array_ops.nansum()`, `duck_array_ops.sum()`, `duck_array_ops.where()` and `duck_array_ops.isnull()`.

rpnaut:
Hello from me hopefully contributing some needfull things.

At first, I would like to comment that I checked out your code.

I ran the following code example:
``

rpnaut:
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!IT SEEMS TO BE A BUG!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

If I do the resample process using the old nomenclature:
` data_aggreg = data["TOT_PREC"].resample(dim="time",how="sum",freq="M") `
it works. 
DO we have a bug in xarray?

rpnaut:
Okay. Using the resample-nomenclature I tried to compare the results with your modified code and with the tagged version 0.10.7. I did not saw any nan's in the resampled fields although I expect to have over water only nan's.

fujiisoup:
@shoyer , thanks for the details. I think I understood your idea.
This sounds a cleaner solution.
I will update the code again, but it will take some more days (or a week).

fujiisoup:
Thanks, @rpnaut .
Actually, I'm changing the code also around `sum`. So it looks my change caused the bug you reported.
I think we do not have a good test coverage around the dataset.reduction.

I will add the test also. Thanks again for the testing :)

fujiisoup:
I think this is ready for another review.

rpnaut:
Thanks, @fujiisoup .

I have good news and i have bad news. Your min_count argument still seems to work only if using the new syntax for resample, i.e. `data.resample($dim=$freq).sum()`

rpnaut:
To wrap it up. Your implementation works for timeseries - data. There is something strange with time-space data, which should be fixed. If this is fixed, it is worth to test in my evaluation environment.
Do you have a feeling, why the new syntax is giving such strange behaviour? Shall we put the bug onto the issue list?

And maybe, it would be interesting to have in the future the min_count argument also available for datasets and not only dataarrays.

fujiisoup:
I noticed `min_count` is working also on `resample`d object.
[Your issue](#2356) might be for the API of `resample`.

fujiisoup:
Can anyone give further review?

xarray/core/nanops.py
```diff
@@ -0,0 +1,211 @@
...
+
+    return result
+
+
+def _nan_argminmax_object(func, fill_value, value, axis=None, **kwargs):
+    """ In house nanargmin, nanargmax for object arrays. Always return integer
+    type """
+    valid_count = count(value, axis=axis)
+    value = fillna(value, fill_value)
+    data = getattr(np, func)(value, axis=axis, **kwargs)
```
shoyer: This function is doing eager evaluation by calling the numpy function -- can we call the dask function instead on dask arrays?

Alternatively, we could make this function error on dask arrays? I just don't want to silently compute them.

xarray/core/nanops.py
```diff
@@ -0,0 +1,211 @@
...
+    # dask seems return non-integer type
```
shoyer: It would be good to file an upstream bug report with dask about this

xarray/core/nanops.py
```diff
@@ -0,0 +1,211 @@
...
+    if isinstance(value, dask_array_type):
+        data = data.astype(int)
+
+    if (valid_count == 0).any():
```
shoyer: This would also evaluate dask arrays, which can be costly. I don't know if there's much to do about this but it should be noted.

xarray/core/nanops.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def nanmax(a, axis=None, out=None):
+    if a.dtype.kind == 'O':
+        return _nan_minmax_object(
+            'max', dtypes.get_neg_infinity(a.dtype), a, axis)
+
+    if isinstance(a, dask_array_type):
+        return dask_array.nanmax(a, axis=axis)
+    return nputils.nanmax(a, axis=axis)
```
shoyer: Maybe this could be more cleanly written as:
```python
    module = nptuils if isinstance(a, dask_array_type) else dask_array
    return module.nanmin(a, axis=axis)
```

xarray/core/nanops.py
```diff
@@ -0,0 +1,211 @@
...
+        return _nan_argminmax_object('argmin', fill_value, a, axis=axis)
+    a, mask = _replace_nan(a, fill_value)
+    if isinstance(a, dask_array_type):
+        res = dask_array.argmin(a, axis=axis)
+    else:
+        res = np.argmin(a, axis=axis)
+
+    if mask is not None:
+        mask = np.all(mask, axis=axis)
+        if np.any(mask):
```
shoyer: It's more dask friendly to use the`.all()` and `.any()` methods -- that way at least most of the check can be done in parallel, and only a single value (the boolean scalar) gets evaluates.

xarray/tests/test_duck_array_ops.py
```diff
@@ -272,21 +294,29 @@ def test_reduce(dim_num, dtype, dask, func, skipna, aggdim):
...
+
+        # make sure the compatiblility with pandas' results.
         if func == 'var':
             expected = series_reduce(da, func, skipna=skipna, dim=aggdim,
                                      ddof=0)
             assert_allclose(actual, expected, rtol=rtol)
             # also check ddof!=0 case
             actual = getattr(da, func)(skipna=skipna, dim=aggdim, ddof=5)
+            if dask:
+                isinstance(da.data, dask_array_type)
```
shoyer: Needs `assert`?

xarray/tests/test_duck_array_ops.py
```diff
@@ -297,11 +327,14 @@ def test_reduce(dim_num, dtype, dask, func, skipna, aggdim):
...
         if func not in ['max', 'min']:
             actual = getattr(da, func)(skipna=skipna, dim=aggdim, dtype=float)
+            assert_dask_array(actual, dask)
             assert actual.dtype == float
 
         # without nan
         da = construct_dataarray(dim_num, dtype, contains_nan=False, dask=dask)
         actual = getattr(da, func)(skipna=skipna)
+        if dask:
+            isinstance(da.data, dask_array_type)
```
shoyer: also needs `assert`

xarray/tests/test_variable.py
shoyer: nope, no idea! :)

xarray/core/nanops.py
fujiisoup: This looks a misunderstanding or already fixed in upstream.
Removed this line.

fujiisoup:
Thanks, @shoyer.
All done.

doc/whats-new.rst
```diff
@@ -70,6 +70,15 @@ Bug fixes
 
 - Tests can be run in parallel with pytest-xdist
 
+- Follow up the renamings in dask; from dask.ghost to dask.overlap
+  By `Keisuke Fujii <https://github.com/fujiisoup>`_.
+
+  By `Tony Tung <https://github.com/ttung>`_.
```
shoyer: this should reference the line above, not your change here

fujiisoup:
Thanks for the review. Merging.

st-bender:
Hi,
just to let you know that `.std()` does not accept the `ddof` keyword anymore (it worked in 0.10.8)
Should I open a new bugreport?

fujiisoup:
Thanks, @st-bender, for the bug report.
I copied your comment to #2440.

PR #2246, opened 2018-06-22
Small error in open_zarr docs

maxim-lian:
I tried setting this up on GCS as an experiment, and found there's no `mode` kwarg to `open_zarr`. I'm not sure whether there are broader issues but thought I'd put this in regardless

(also - if anyone has any experience in deploying xarray over https://github.com/dask/dask-kubernetes and GCS, I'd be interested in your experience. We're currently working with either a) BigQuery for big data or b) in-memory xarray for small data, and I was interested whether there's a way of retaining the xarray model for bigger data without too much overhead)

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -603,7 +603,7 @@ pass to xarray::
     # write to the bucket
     ds.to_zarr(store=gcsmap)
     # read it back
-    ds_gcs = xr.open_zarr(gcsmap, mode='r')
+    ds_gcs = xr.open_zarr(gcsmap)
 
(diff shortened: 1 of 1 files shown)


shoyer:
Thanks!

I'm pretty sure pangeo does most of their dask deployments on GCS: http://pangeo-data.org/setup_guides/cloud.html

maxim-lian:
Great, thanks for the pointer!

jhamman:
@maxim-lian - we'd love to hear about your use case on the pangeo issue tracker: https://github.com/pangeo-data/pangeo/issues

PR #2255, opened 2018-06-27
Add automatic chunking to open_rasterio

### LINKED ISSUE #2093: Default chunking in GeoTIFF images
Given a tiled GeoTIFF image I'm looking for the best practice in reading it as a chunked dataset.  I did this in [this notebook](https://gist.github.com/mrocklin/3df315e93d4bdeccf76db93caca2a9bd) by first opening the file with rasterio, looking at the block sizes, and then using those to inform the argument to `chunks=` in `xarray.open_rasterio`.  This works, but is somewhat cumbersome because I also had to dive into the rasterio API.  Do we want to provide defaults here?

In dask.array every time this has come up we've always shot it down, automatic chunking is error prone and hard to do well.  However in these cases the object we're being given usually also conveys its chunking in a way that matches how dask.array thinks about it, so the extra cognitive load on the user has been somewhat low.  Rasterio's model and API feel much more foreign to me though than a project like NetCDF or H5Py.  I find myself wanting a `chunks=True` or `chunks='100MB'` option.

Thoughts on this?  Is this in-scope?  If so then what is the right API and what is the right policy for how to make xarray/dask.array chunks larger than GeoTIFF chunks?

mrocklin:
This uses the automatic chunking in dask 0.18+ to chunk rasterio
datasets in a nicely aligned way.

Currently this doesn't implement tests due to a difficulty in creating
chunked tiff images.

This also uncovered some inefficiencies in how Dask doesn't align rechunking to existing chunk schemes.

 - [x] Closes #2093 

I could use help on how the following:

-  How to create tiled TIFF files in the tests
-  The right way to merge different dtypes and block shapes in the tiff file.  Currently I'm assuming that they're uniform

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -286,28 +286,7 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
     # this lets you write arrays loaded with rasterio
     data = indexing.CopyOnWriteArray(data)
     if chunks in (True, 'auto'):
-        if not attrs.get('is_tiled', False):
-            msg = "Data store is not tiled. Automatic chunking is not sensible"
-            raise ValueError(msg)
-
-        import dask.array
-        if dask.__version__ < '0.18.0':
-            msg = ("Automatic chunking requires dask.__version__ >= 0.18.0 . "
-                   "You currently have version %s" % dask.__version__)
-            raise NotImplementedError(msg)
-
-        img = riods._ds
-        block_shapes = set(img.block_shapes)
-        block_shape = (1,) + list(block_shapes)[0]
-        previous_chunks = tuple((c,) for c in block_shape)
-        shape = (img.count, img.height, img.width)
-        dtype = img.dtypes[0]
-        chunks = dask.array.core.normalize_chunks(
-            'auto',
-            shape=shape,
-            previous_chunks=previous_chunks,
-            dtype=dtype
-        )
+        chunks = (1, 'auto', 'auto')
 
     if cache and (chunks is None):
         data = indexing.MemoryCachedArray(data)
@@ -327,6 +306,30 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         name_prefix = 'open_rasterio-%s' % token
         if lock is None:
             lock = RASTERIO_LOCK
+
+        if not attrs.get('is_tiled', False):
+            msg = "Data store is not tiled. Automatic chunking is not sensible"
+            raise ValueError(msg)
+
+        import dask.array
+        if dask.__version__ < '0.18.0':
+            msg = ("Automatic chunking requires dask.__version__ >= 0.18.0 . "
+                   "You currently have version %s" % dask.__version__)
+            raise NotImplementedError(msg)
+
+        img = riods._ds
+        block_shapes = set(img.block_shapes)
+        block_shape = (1,) + list(block_shapes)[0]
+        previous_chunks = tuple((c,) for c in block_shape)
+        shape = (img.count, img.height, img.width)
+        dtype = img.dtypes[0]
+        chunks = dask.array.core.normalize_chunks(
+            chunks,
+            shape=shape,
+            previous_chunks=previous_chunks,
+            dtype=dtype
+        )
+
         result = result.chunk(chunks, name_prefix=name_prefix, token=token,
                               lock=lock)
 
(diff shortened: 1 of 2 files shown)


mrocklin:
```python
import os
if not os.path.exists('myfile.tif'):
    import requests
    response = requests.get('https://oin-hotosm.s3.amazonaws.com/5abae68e65bd8f00110f3e42/0/5abae68e65bd8f00110f3e43.tif')
    with open('myfile.tif', 'wb') as f:
        f.write(response.content)

import dask
dask.config.set({'array.chunk-size': '1MiB'})

import xarray as xr
ds = xr.open_rasterio('myfile.tif', chunks=True)  # this only reads metadata to start

>>> ds.chunks
((1, 1, 1),
 (1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 136),
 (1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 995))
```

Also depends on https://github.com/dask/dask/pull/3679 .  Without that PR it will use values that are similar, but don't precisely align with 1024.

xarray/backends/rasterio_.py
```diff
@@ -301,6 +306,30 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
...
+                   "You currently have version %s" % dask.__version__)
+            raise NotImplementedError(msg)
+
+        img = riods._ds
+        block_shapes = set(img.block_shapes)
+        block_shape = (1,) + list(block_shapes)[0]
+        previous_chunks = tuple((c,) for c in block_shape)
+        shape = (img.count, img.height, img.width)
+        dtype = img.dtypes[0]
+        chunks = dask.array.core.normalize_chunks(
```
shoyer: Should we expose `normalize_chunks()` as a top-level API in dask.array, e.g., `dask.array.normalize_chunks`? I'm generally a little nervous about dipping into `dask.array.core`.

snowman2:
> The right way to merge different dtypes and block shapes in the tiff file. Currently I'm assuming that they're uniform

I have yet to run into a raster that varies dtypes and block shapes across bands. Most of the time, they are single band rasters. And if they are not, they have had the same dtype and block shape. So, I think your assumption is a good one for most use cases.

snowman2:
> How to create tiled TIFF files in the tests

One thing I would like to note is that the automatic chunking would be useful if the raster is tiled or not. I tested out a raster that was not tiled, but it still had chunks. This is due to the raster being written in stripes. So, I would recommend removing the restriction to only tiled rasters.

Also, to create a tiled raster:
```python
import rasterio
import numpy
from affine import Affine

with rasterio.open(
    "tiled.tif",
    "w",
    driver="GTiff",
    count=2,
    width=1024,
    height=1024,
    crs="+init=epsg:4326",
    transform=Affine(0.0083333333, 0.0, -180.00416666665, 0.0, -0.0083333333, 75.00416666665),
    dtype=rasterio.float32,
    tiled=True,
    blockxsize=512,
    blockysize=512,
) as rds:
    rds.write((numpy.random.rand(2, 1024, 1024)*10).astype(numpy.float32))
```

mrocklin:
I've abandoned this PR.  If anyone has time to pick it up, that would be welcome.  I think that it would have positive impact.

snowman2:
> I've abandoned this PR. If anyone has time to pick it up, that would be welcome.

I appreciate you staring this! Based on this PR, I added the feature into `rioxarray` here: https://github.com/corteva/rioxarray/pull/31 (released in version 0.0.9).

mrocklin:
I'm glad to hear it!  I'm curious, are there features in rioxarray that
could be pushed upstream?

snowman2:
> I'm curious, are there features in rioxarray that could be pushed upstream?

Depends on what the xarray maintainers would like to add. I would definitely like to see the `open_rasterio` function in `rioxarray` absorbed back into `xarray`. Things have just been really slow moving with the other PRs/issues.

fmaussion:
>> I'm curious, are there features in rioxarray that could be pushed upstream?

> Depends on what the xarray maintainers would like to add. I would definitely like to see the open_rasterio function in rioxarray absorbed back into xarray. Things have just been really slow moving with the other PRs/issues.

Put another way: why don't we put all the logic in rioxarray and make rioxarray an optional dependency of xarray to open rio files?

snowman2:
> Put another way: why don't we put all the logic in rioxarray and make rioxarray an optional dependency of xarray to open rio files?

That is an option. All of the logic has already been moved over.

keewis:
with #4697 and the fact that this seems to have been included in `rioxarray`: should we close this?

PR #2258, opened 2018-06-29
BUG: unnamed args in faceted line plots

### LINKED ISSUE #2248: unnamed args in faceted line plots
Unnamed arguments that are passed to line plots are directly passed to `plot`. 
```python
import xarray as xr
import numpy as np

d3 = xr.DataArray(np.random.randn(10, 6, 3),
                  dims=['x', 'y', 'z'],
                  coords=[range(10), range(6), ['foo', 'bar', 'foobar']])
d2 = d3[..., 0].squeeze(drop=True)
# This will make triangle markers and dashed lines:
d2.plot.line('^--',x='x')
```
We [recently implemented ](https://github.com/pydata/xarray/pull/2107) faceted line plots:
```python
d3.plot(col='z', hue='x') #should work from commit `bc52f8a` onwards.
```

However, current implementation does not allow for passing of unnamed arguments:
```python
d3.plot('^--', col='z', hue='x') #gives error
d3.plot.line('^--', col='z', hue='x') #gives different error
```
I'd be happy to fix this (it's probably an easy fix, I didn't check) but I'm not sure what the correct user interface should be. Should `DataArray.plot` with `hue` as a keyword behave identically to `DataArray.plot.line`?

yohai:
- [x] Closes #2248 (remove if there is no corresponding issue, which should only be the case for minor changes)

Fixed a syntax inconsistency, allowing to pass unnamed arguments to faceted line plots. 

```python
Example code:
import xarray as xr
import numpy as np

np.random.seed(0)
da = xr.DataArray(np.random.randn(3, 3, 3, 3),
                  dims=['x', 'y', 'z',  'w'],
                  coords=[range(3),  [3.5, 4.9, 6.7], ['a','b','c'], ['foo', 'bar', 'foobar']],
                  name='arr_name')
da.plot.line('o--',row='y', col='w',hue='z')
```

sdojn

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -310,6 +310,8 @@ def line(darray, *args, **kwargs):
     yincrease = kwargs.pop('yincrease', True)
     add_legend = kwargs.pop('add_legend', True)
     _labels = kwargs.pop('_labels', True)
+    if args is ():
+        args = kwargs.pop('args', ())
 
     ax = get_axis(figsize, size, aspect, ax)
     xplt, yplt, hueplt, xlabel, ylabel, huelabel = \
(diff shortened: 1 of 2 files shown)


yohai:
Sorry, I messed this up. Opening a new fresh pull request

PR #2262, opened 2018-07-02
Add curvilinear grid support to to_cdms2 and fix mask bug

stefraynaud:
 - [x] Fully documented, including ``whats-new.rst`` for all changes and api.rst for new API (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

### DIFF
--- a/xarray/convert.py
+++ b/xarray/convert.py
@@ -42,14 +42,16 @@ def from_cdms2(variable):
     dims = variable.getAxisIds()
     coords = {}
     for axis in variable.getAxisList():
-        coords[axis.id] = DataArray(np.asarray(axis), dims=[axis.id],
+        coords[axis.id] = DataArray(
+            np.asarray(axis), dims=[axis.id],
             attrs=_filter_attrs(axis.attributes, cdms2_ignored_attrs))
     grid = variable.getGrid()
     if grid is not None:
         ids = [a.id for a in grid.getAxisList()]
         for axis in grid.getLongitude(), grid.getLatitude():
             if axis.id not in variable.getAxisIds():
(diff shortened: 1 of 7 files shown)


xarray/tests/test_dataarray.py
```diff
@@ -2940,6 +2942,39 @@ def test_to_and_from_cdms2(self):
         roundtripped = DataArray.from_cdms2(actual)
         assert_identical(original, roundtripped)
```
shoyer: Could you add separate test methods for this? It makes things a little easier to debug when things go wrong :).

xarray/convert.py
```diff
@@ -38,15 +39,28 @@ def from_cdms2(variable):
     """
     values = np.asarray(variable)
     name = variable.id
-    coords = [(v.id, np.asarray(v),
-               _filter_attrs(v.attributes, cdms2_ignored_attrs))
-              for v in variable.getAxisList()]
+    dims = variable.getAxisIds()
+    coords = {}
```
shoyer: Can you switch this to `OrderedDict` instead? We usually try to use those to keep the order of coordinates more predictable (though it's increasingly less relevant with ordered dicts by default on Python 3).

stefraynaud:
Is it up to me to resolve the whats-new.rst conflict?

fmaussion:
> Is it up to me to resolve the whats-new.rst conflict?

I just took care of it: you'll have to pull from origin if you want to add further changes though

stefraynaud:
Thanks @fmaussion

shoyer:
thanks @stefraynaud !

PR #2264, opened 2018-07-02
WIP DataArray.plot() can now plot errorbars with kwargs xerr, yerr

dcherian:

Added support for errorbar plotting. This works by providing the kwargs `xerr` and/or `yerr` to `plot()` or `plot.line()`. It will only work for 1D data. 

Errorbars for plots that use`hue` argument require that we loop and plot each line individually. I'm happy to add this is you think it's a good idea.

Example from docs: 
`air.isel(time=10, lon=10).plot.line(y='lat', xerr=3, yerr=1.5, ecolor='r')`

image

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -270,6 +270,7 @@ def line(darray, *args, **kwargs):
         Coordinates for x, y axis. Only one of these may be specified.
         The other coordinate plots values from the DataArray on which this
         plot method is called.
+    xerr, yerr : Errorbar sizes (optional)
     xincrease : None, True, or False, optional
         Should the values on the x axes be increasing from left to right?
         if None, use the default for the matplotlib function.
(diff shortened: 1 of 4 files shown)


shoyer:
Very cool!

> Errorbars for plots that usehue argument require that we loop and plot each line individually. I'm happy to add this is you think it's a good idea.

I'm pretty sure that under the covers matplotlib uses a loop when `plt.plot` results in multiple lines. So this seems pretty reasonable to me.

xarray/plot/plot.py
```diff
@@ -270,6 +270,7 @@ def line(darray, *args, **kwargs):
         Coordinates for x, y axis. Only one of these may be specified.
         The other coordinate plots values from the DataArray on which this
         plot method is called.
+    xerr, yerr : Errorbar sizes (optional)
```
jhamman: Can we expand this description?

marccheneau:
Hi there. I have read this thread with a great interest as I was wondering whether it was possible to plot data from a Dataset with errorbars. Yet it seems that the changes discussed here were not merged in the end. At least I can't find the corresponding piece of code in the latest version of xarray/plot/plot.py.
What is the current status of this feature?

dcherian:
This never went in because I never finished it :)

One bit of weirdness is that the return types are different for `plt.plot` and `plt.errorbar` so the return type for `xarray.plot` will change when `xerr` or `yerr` is specified. I don't know if we should care too much.

Feel free to take over

marccheneau:
I will have a look at it, starting with your initial commit. Regarding the change of return type, I won't be able to judge whether this can cause trouble or not.

PR #2271, opened 2018-07-08
dev/test build for python 3.7

jhamman:
- [x] Tests added


jhamman:
It looks like travis isn't ready for 3.7. We can revisit in a few days. I did get a bare bones environment up and running with 3.7 and the minimal test suite passed, so that's a good sign.

jhamman:
Python 3.7 tests seem happy now. I have a minimal environment setup for 3.7. It may be a bit until all the optional deps have 3.7 support so maybe we just merge this as is and migrate our more comprehensive test environments to 3.7 as needed.

shoyer:
Nice, thanks!

To be honest, I don't know if there's any particular need to even specify a Python version with Travis, given that we use conda environments for everything.

PR #2285, opened 2018-07-13
ENH: format_array_flat() always displays first and last items.

### LINKED ISSUE #1186: Including last coordinate values when displaying coordinates
I first posted this to https://groups.google.com/forum/#!topic/xarray/wjPtXMr91sg .

I'm afraid I'm not set up to submit a PR, but I think something like the following should work once one has created a ``last_n_items()`` similar to the existing ``first_n_items()``.

Also, I wonder if it would make sense to change the minimum number of items displayed from 1 to min(2, items_ndarray.size).

```
def format_array_flat(items_ndarray, max_width):
    """Return a formatted string for as many items in the flattened version of
    items_ndarray that will fit within max_width characters
    """
    # every item will take up at least two characters, but we always want to
    # print at least one item
    max_possibly_relevant = max(int(np.ceil(max_width / 2.0)), 1)
    relevant_items = sum(([x, y] for (x, y) in
                          zip(first_n_items(items_ndarray, (max_possibly_relevant + 1) // 2),
                              reversed(last_n_items(items_ndarray, max_possibly_relevant // 2)))),
                         [])
    pprint_items = format_items(relevant_items)

    cum_len = np.cumsum([len(s) + 1 for s in pprint_items]) - 1
    if (max_possibly_relevant < items_ndarray.size or
            (cum_len > max_width).any()):
        padding = u' ... '
        count = max(np.argmax((cum_len + len(padding)) > max_width), 1)
    else:
        count = items_ndarray.size
        padding = u'' if (count <= 1) else u' '

    pprint_str = u' '.join(np.take(pprint_items, range(0, count, 2))) + padding + \
                 u' '.join(np.take(pprint_items, range(count - (count % 2) - 1, 0, -2)))
    return pprint_str
```

seth-p:
- [ ] Closes #1186 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import, division, print_function
 import contextlib
 import functools
 from datetime import datetime, timedelta
+from itertools import zip_longest
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


shoyer:
Can you share a few examples of what `xarray.Dataset` objects look like with this change?

Originally, I intentionally did not show values at both the start and end because I was concerned that `Dataset.__repr__()` could load lazy data from disk or a remote server. But we changed the Dataset repr somewhat recently (I think in v0.20), so it no longer does that.

seth-p:
For some reason (presumably due to Github's outage today), this PR isn't updating to reflect the latest commits in my branch. So I'm going to close this one and create a new one.

seth-p:
Replaced with https://github.com/pydata/xarray/pull/2293.

Issue #2244, opened 2018-06-22
Implement shift for CFTimeIndex 


aidanheerdegen:
#### Code Sample

```python
import numpy as np
import xarray as xr
import pandas as pd

from cftime import num2date, DatetimeNoLeap

times = num2date(np.arange(730), calendar='noleap', units='days since 0001-01-01')
da = xr.DataArray(np.arange(730), coords=[times], dims=['time'])
```

#### Problem description

I am trying to shift a time index as I need to align datasets to a common start point.

Directly incrementing one of the `CFTimeIndex` values works:
```python
>>> da.time.get_index('time')[0] + pd.Timedelta('365 days')
cftime.DatetimeNoLeap(2, 1, 1, 0, 0, 0, 0, -1, 1)
```
Trying to use `shift` does not:
```python
>>> da.time.get_index('time').shift(1,'Y')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/g/data3/hh5/public/apps/miniconda3/envs/analysis3-18.04/lib/python3.6/site-packages/pandas/core/indexes/base.py", line 2629, in shift
    type(self).__name__)
NotImplementedError: Not supported for type CFTimeIndex
```

If I want to shift a time index is the only way currently is to loop over all the individual elements of the index and add a time offset to each.

#### Expected Output

I would expect to have CFTimeIndex shifted by the desired time delta.

#### Output of ``xr.show_versions()``

<details>
# Paste the output here xr.show_versions() here

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 3.10.0-693.17.1.el6.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: C
LOCALE: None.None

xarray: 0.10.7
pandas: 0.23.1
numpy: 1.14.5
scipy: 1.1.0
netCDF4: 1.3.1
h5netcdf: 0.5.1
h5py: 2.8.0
Nio: None
zarr: 2.2.0
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.5
distributed: 1.21.8
matplotlib: 1.5.3
cartopy: 0.16.0
seaborn: 0.8.1
setuptools: 39.2.0
pip: 9.0.3
conda: None
pytest: 3.6.1
IPython: 6.4.0
sphinx: None
</details>

aidanheerdegen:
[...]

Issue #2191, opened 2018-05-28
Adding resample functionality to CFTimeIndex


spencerahill:
Now that CFTimeIndex has been implemented (#1252), one thing that remains to implement is resampling.  @shoyer provided a sketch of how to implement it: https://github.com/pydata/xarray/pull/1252#issuecomment-380593243.  In the interim, @spencerkclark provided a sketch of a workaround for some use-cases using groupby: https://github.com/pydata/xarray/issues/1270#issuecomment-390973986.

I thought it would be useful to have a new Issue specifically on this topic from which future conversation can continue.  @shoyer, does that sketch you provided still seem like a good starting point?

shoyer:
Yes, I think so. The main thing we need is a function to map from datetime -> datetime at start of frequency.

naomi-henderson:
I am trying to combine the monthly CMIP5 rcp85 ts datasets (go past 2064AD)  with the myriad calendars, so I love the new CFTimeIndex!  But I need resample(time='MS') in order to force them all to start on the first of each month
thanks!

spencerkclark:
@naomi-henderson thanks!  In the meantime here's a possible workaround, in case you haven't figured one out already:
```python
import numpy as np
import xarray as xr

from cftime import num2date, DatetimeNoLeap


times = num2date(np.arange(730), calendar='noleap', units='days since 0001-01-01')
da = xr.DataArray(np.arange(730), coords=[times], dims=['time'])

month_start = [DatetimeNoLeap(date.dt.year, date.dt.month, 1) for date in da.time]
da['MS'] = xr.DataArray(month_start, coords=da.time.coords)
resampled = da.groupby('MS').mean('time').rename({'MS': 'time'})
```

naomi-henderson:
@spencerkclark thanks!  I hadn't figured out that particular workaround, but it works, albeit quite slow. For now it will get me to the next step, but just changing to first-of-the-month takes longer than regridding all models to a common grid!

spencerkclark:
Indeed what I had above is quite slow!

```python
In [6]: %%timeit
   ...: month_start = [DatetimeNoLeap(date.dt.year, date.dt.month, 1) for date in da.time]
   ...:
1 loop, best of 3: 588 ms per loop
```

Iterating over the contents of `da.time` generates DataArray instances encapsulating single dates.  We can iterate over the dates themselves directly, which is much (over 1000x) faster:

```python
In [7]: %%timeit
[...]

PR #2301, opened 2018-07-19
WIP Add a CFTimeIndex-enabled xr.date_range function

### LINKED ISSUE #2142: add CFTimeIndex enabled date_range function
Pandas' [`date_range`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.date_range.html) function is a fast and flexible way to create `DateTimeIndex` objects. Now that we have a functioning `CFTimeIndex`, it would be great to add a version of the `date_range` function that supports other calendars and dates out of range for Pandas. 

#### Code Sampl and expected output

```python
In [1]: import xarray as xr

In [2]: xr.date_range('2000-02-26', '2000-03-02')
Out[2]:
DatetimeIndex(['2000-02-26', '2000-02-27', '2000-02-28', '2000-02-29',
               '2000-03-01', '2000-03-02'],
              dtype='datetime64[ns]', freq='D')

In [3]: metsim.datetime.date_range('2000-02-26', '2000-03-02', calendar='noleap')
Out[3]:
CFTimeIndex(['2000-02-26', '2000-02-27', '2000-02-28', '2000-03-01',
               '2000-03-02'],
              dtype='cftime.datetime', freq='D')
```

spencerkclark:
- [ ] Closes #2142 

I took the approach first discussed [here](https://github.com/pydata/xarray/pull/1252#issuecomment-380593243) by @shoyer and followed pandas by creating simplified offset classes for use with cftime objects to implement a `CFTimeIndex`-enabled `date_range` function.  I still may clean things up a bit and add a few more tests, but I wanted to post this in its current state to show some progress, as I think it is more or less working.  I will try to ping folks when it is ready for a more detailed review.

Here are a few examples:
```
In [1]: import xarray as xr

In [2]: xr.date_range('2000-02-01', '2002-05-05', freq='3M')
Out[2]:
DatetimeIndex(['2000-02-29', '2000-05-31', '2000-08-31', '2000-11-30',
               '2001-02-28', '2001-05-31', '2001-08-31', '2001-11-30',
               '2002-02-28'],
              dtype='datetime64[ns]', freq='3M')

In [3]: xr.date_range('2000-02-01', '2002-05-05', freq='3M', calendar='noleap')
Out[3]:
CFTimeIndex([2000-02-28 00:00:00, 2000-05-31 00:00:00, 2000-08-31 00:00:00,
             2000-11-30 00:00:00, 2001-02-28 00:00:00, 2001-05-31 00:00:00,
             2001-08-31 00:00:00, 2001-11-30 00:00:00, 2002-02-28 00:00:00],
            dtype='object')
In [4]: xr.date_range('2000-02-01', periods=4, freq='3A-JUN')
Out[4]: DatetimeIndex(['2000-06-30', '2003-06-30', '2006-06-30', '2009-06-30'], dtype='datetime64[ns]', freq='3A-JUN')

In [5]: xr.date_range('2000-02-01', periods=4, freq='3A-JUN', calendar='noleap')
Out[5]:
CFTimeIndex([2000-06-30 00:00:00, 2003-06-30 00:00:00, 2006-06-30 00:00:00,
             2009-06-30 00:00:00],
            dtype='object')

In [6]: xr.date_range(end='2000-02-01', periods=4, freq='3A-JUN')
Out[6]: DatetimeIndex(['1990-06-30', '1993-06-30', '1996-06-30', '1999-06-30'], dtype='datetime64[ns]', freq='3A-JUN')

In [7]: xr.date_range(end='2000-02-01', periods=4, freq='3A-JUN', calendar='noleap')
Out[7]:
CFTimeIndex([1990-06-30 00:00:00, 1993-06-30 00:00:00, 1996-06-30 00:00:00,
             1999-06-30 00:00:00],
            dtype='object')
```

Hopefully the offset classes defined here would also be useful for implementing things like `resample` for `CFTimeIndex` objects (#2191) and `CFTimeIndex.shift` (#2244).


shoyer:
I haven't had the chance to look in detail at this yet, but one small point I would suggest is renaming it from `xarray.date_time` to `xarray.cftime_range`. This makes it more obvious that the function is really for users of cftime.

jhamman:
@spencerkclark - sync this branch with master and we'll get this all wrapped up.

spencerkclark:
Thanks @jhamman!  This should be synced up and ready for review.  The test failure under the dask-dev build doesn't appear to be related.

spencerkclark:
Awesome, thanks @shoyer.

xarray/coding/cftime_offsets.py
```diff
@@ -0,0 +1,765 @@
...
+    def __sub__(self, other):
+        import cftime
+
+        if isinstance(other, cftime.datetime):
+            raise TypeError('Cannot subtract a cftime.datetime '
+                            'from a time offset.')
+        elif type(other) == type(self):
+            return type(self)(self.n - other.n)
+        else:
+            raise NotImplementedError
```
shoyer: Use `return NotImplemented` here (which Python will reraise as TypeError)

xarray/coding/cftime_offsets.py
```diff
@@ -0,0 +1,765 @@
...
+            'standard': cftime.DatetimeProlepticGregorian
+        }
+        return calendars[calendar]
+
+
+class BaseCFTimeOffset(object):
+    _freq = None
+
+    def __init__(self, n=1):
+        self.n = n
```
shoyer: do you want to add checks here to verify that `n` is an integer? or maybe floats are OK, too?

xarray/coding/cftime_offsets.py
```diff
@@ -0,0 +1,765 @@
...
+            raise ValueError(
+                'If converting a string to a cftime.datetime object, '
+                'a calendar type must be provided')
+        date, _ = _parse_iso8601_with_reso(get_date_type(calendar),
+                                           date_str_or_date)
+        return date
+    elif isinstance(date_str_or_date, cftime.datetime):
+        return date_str_or_date
+    else:
+        raise ValueError('date_str_or_date must be a string or a '
```
shoyer: should be `TypeError`

xarray/coding/cftime_offsets.py
```diff
@@ -0,0 +1,765 @@
...
+    - Compound sub-monthly frequencies are not supported, e.g. '1H1min', as
+      these can easily be written in terms of the finest common resolution,
+      e.g. '61min'.
+
+    Examples
+    --------
+
+    For dates from standard calendars within the ``pandas.Timestamp``-valid
+    range, this function operates as a thin wrapper around
+    ``pandas.date_range``.
```
shoyer: I don't think this use-case is important to support, and interfaces that can return different types can be difficult to understand. I would instead always return a `CFTimeIndex`.

xarray/coding/cftimeindex.py
```diff
@@ -149,10 +182,22 @@ class CFTimeIndex(pd.Index):
                                   'The microseconds of the datetime')
     date_type = property(get_date_type)
 
-    def __new__(cls, data):
```
shoyer: Let's keep the constructor the same here, rather than overriding it to match the signature of `cftime_range`. (I know this is a deviation from pandas, but frankly I think the pandas behavior is a confusing mistake.)

xarray/coding/cftimeindex.py
```diff
@@ -254,3 +299,28 @@ def __contains__(self, key):
...
+        data = self._format_data()
+        attrs = self._format_attrs()
+        space = self._format_space()
+
+        if self.date_type is None:
+            attrs.append(('calendar', None))
+        else:
+            attrs.append(('calendar', repr(infer_calendar_name(self._data))))
+
+        prepr = (pd.compat.u(",%s") %
```
shoyer: you can just use `u",%s"` rather than `pd.compat.u` (which is only needed for old versions of Python 3 that we no longer support)

xarray/coding/cftimeindex.py
```diff
@@ -254,3 +299,28 @@ def __contains__(self, key):
...
         return self.__contains__(key)
+
+    def __unicode__(self):
+        """Return a string representation for this object.
+
+        Adds a calendar attribute to denote the calendar type of the index.
+        Adapted from pandas.core.indexes.base.__unicode__
+        """
+        klass = self.__class__.__name__
+        data = self._format_data()
```
shoyer: This is a little dangerous: we're using private methods here which might go away without notice. I have been burned by this in the past (specifically in xarray with pandas).

The other reason to hold off on copying the pandas repr() is that we don't implement casting like `pd.to_datetime()` instead the CFTimeIndex constructor, e.g.,
```
In [2]: pd.DatetimeIndex(['2000', '2001'])
Out[2]: DatetimeIndex(['2000-01-01', '2001-01-01'], dtype='datetime64[ns]', freq=None)
```

(to be clear, I don't think this is a good idea either, but at least it ensures that the repr roundtrips faithfully)

xarray/tests/test_cftime_offsets.py
```diff
@@ -0,0 +1,723 @@
...
+import numpy as np
+import pandas as pd
+
+from xarray.coding.cftime_offsets import (
+    BaseCFTimeOffset, YearBegin, YearEnd, MonthBegin, MonthEnd,
+    Day, Hour, Minute, Second, _days_in_month,
+    to_offset, get_date_type, _MONTH_ABBREVIATIONS, _cftime_range,
+    to_cftime_datetime, cftime_range)
+from xarray import CFTimeIndex
+from . import has_cftime
```
shoyer: Instead of checking `has_cftime` and `skipif` decorators on every function, use `cftime = pytest.importorskip('cftime')`.

xarray/tests/test_cftime_offsets.py
```diff
@@ -0,0 +1,723 @@
...
+
+
+@pytest.mark.skipif(not has_cftime, reason='cftime not installed')
+@pytest.mark.parametrize(
+    ('start', 'end', 'periods', 'freq', 'name'),
+    [('0001', None, 5, 'A', 'foo'),
+     ('2000', None, 5, 'A', 'foo'),
+     ('2000', '1999', None, 'A', 'foo')]
+)
+def test_cftime_range(start, end, periods, freq, name, calendar):
```
shoyer: Can you add selection of calendar specific tests, e.g., verify that generating generating a year's worth of dates with noleap, all_leap and 360_day does the right thing?

xarray/coding/cftime_offsets.py
```diff
@@ -0,0 +1,765 @@
+"""Time offset classes for use with cftime.datetime objects"""
```
shoyer: Assuming that most of this file is copied/adapted from pandas, please add copy of the pandas copyright notice at the top.

xarray/coding/cftime_offsets.py
spencerkclark: Indeed `n` must be an integer.  I added a check here.  I also added a check on the `month` argument provided to the `YearOffset` classes.

xarray/coding/cftime_offsets.py
spencerkclark: I added a copy of the license for pandas to the top of the module.  Is that what you meant?  Should we do the same in `cftimeindex.py`?

shoyer: > I added a copy of the license for pandas to the top of the module. Is that what you meant? 

Yes, this is what I had in mind. A comment would also be OK rather than the docstring.

> Should we do the same in cftimeindex.py?

Sure, this is probably a good idea

xarray/coding/cftime_offsets.py
```diff
@@ -240,7 +287,7 @@ def __sub__(self, other):
         elif type(other) == type(self) and other.month == self.month:
             return type(self)(self.n - other.n, month=self.month)
         else:
-            raise NotImplementedError
+            raise NotImplemented
```
shoyer: should be `return NotImplemented`, not `raise NotImplemented`

xarray/coding/cftimeindex.py
spencerkclark: Sure thing -- I retained my addition of the optional name argument so that we can support it in `cftime_range` (and obviously the constructor here as well).

spencerkclark:
Many thanks for the initial review @shoyer.  I think I got to everything so far.  I agree restricting `cftime_range` to only return CFTimeIndexes makes things much simpler to reason about.

xarray/coding/cftime_offsets.py
```diff
@@ -0,0 +1,736 @@
...
+        return self.__add__(other)
+
+    def __rsub__(self, other):
+        if isinstance(other, BaseCFTimeOffset) and type(self) != type(other):
+            raise TypeError('Cannot subtract cftime offsets of differing '
+                            'types')
+        return -self + other
+
+    def __apply__(self):
+        raise NotImplementedError
```
shoyer: should also be `return NotImplemented` to differ implementation to subclasses

Issue #2298, opened 2018-07-18
Making xarray math lazy


shoyer:
At SciPy, I had the realization that it would be relatively straightforward to make element-wise math between xarray objects lazy. This would let us support lazy coordinate arrays, a feature that has quite a few use-cases, e.g., for both geoscience and astronomy.

The trick would be to write a lazy array class that holds an element-wise vectorized function and passes indexers on to its arguments. I haven't thought too hard about this yet for vectorized indexing, but it could be quite efficient for outer indexing. I have some prototype code but no tests yet.

The question is how to hook this into xarray operations. In particular, supposing that the inputs to a function do no hold dask arrays:
- Should we try to make *every* element-wise operation with vectorized functions (ufuncs) lazy by default? This might have negative performance implications and would be a little tricky to implement with xarray's current code, since we still implement binary operations like `+` with separate logic from `apply_ufunc`.
- Should we make every element-wise operation that explicitly uses `apply_ufunc()` lazy by default?
- Or should we only make element-wise operations lazy with `apply_ufunc()` if you use some special flag, e.g., `apply_ufunc(..., lazy=True)`?

I am leaning towards the last option for now but would welcome other opinions.

fujiisoup:
This sounds interesting.
I am curious what the practical difference from dask is.
Does it mean some maths are lazy by default (without any external library)?

fujiisoup:
Thanks, @shoyer 

> we have enough related functionality (e.g., for lazy and explicit indexing) 

Agreed.
Actually, it sounds very fun to code the lazy arithmetics.

> Ideally, this logic would live outside xarray. 

Yes, I concerned about this.
We have discussed to support more kinds of array-likes (e.g. dask, sparse, cupy) in #1938, 
and I thought the lazy array can be (ideally) one of them.

But in practice, it should take a long time to realize the any-array-like support and it might be a good idea to natively support the lazy mathematics for now.
If we are heading to any-array-like support, I think that the implementation of the lazy array should be as isolated from xarray core logic as possible so that we can move smoothly to the any-array-like support in the future.

Therefore, personally, I'd like to see this lazy math by implementing a lazy array.
The API I thought of is `.to_lazy()` which converts the backend to the lazy array,
as similar to that `.chunk()` converts the backend to dask array.

shoyer:
> Therefore, personally, I'd like to see this lazy math by implementing a lazy array.
> The API I thought of is .to_lazy() which converts the backend to the lazy array,
[...]

PR #2302, opened 2018-07-20
WIP: lazy=True in apply_ufunc()

shoyer:
- [x] Closes https://github.com/pydata/xarray/issues/2298
 - [ ] Tests added
 - [ ] Tests passed

Still needs more tests and documentation.

### DIFF
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -11,6 +11,7 @@ from collections import Counter
 import numpy as np
 
 from . import duck_array_ops
+from . import indexing
 from . import utils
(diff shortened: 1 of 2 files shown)


fujiisoup:
Super great. 
I am excited the lazy math is realized so simple and clear implementation.

A trivial comment.
Will we support non-elementwise lazy math such as `dot` or `sum` in the near future?
(I guess they are not so difficult)

In that case, `LazyFunctionArray` should need `signature` argument in the construction, 
so maybe `LazyElementwiseFunctionArray` can also have the same API?

PR #2303, opened 2018-07-20
Rename "Recipes" to "Gallery"

fmaussion:
Simply because it makes more sense after all.

The link address won't change so it's not a biggie (http://xarray.pydata.org/en/latest/auto_gallery/index.html)


shoyer:
makes sense to me 👍

PR #2309, opened 2018-07-24
DOC: add initial draft of a development roadmap for xarray

shoyer:
See here for the rendered version:
https://github.com/shoyer/xarray/blob/4dd4463d5f79c07eb3a01f67b7329f2f4a60dd31/doc/roadmap.rst

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -74,6 +74,7 @@ Documentation
 * :doc:`whats-new`
 * :doc:`api`
 * :doc:`internals`
+* :doc:`roadmap`
 * :doc:`contributing`
 
 .. toctree::
@@ -84,6 +85,7 @@ Documentation
    whats-new
    api
    internals
+   roadmap
    contributing
 
 See also
--- a/doc/roadmap.rst
+++ b/doc/roadmap.rst
@@ -191,16 +191,18 @@ non-computer scientists or users less familiar with the pydata ecosystem
 might find it difficult to learn xarray and realize how xarray can help
 them in their daily work.
 
-In order to lower this adoption barrier, we propose to: - Develop
-entry-level tutorials for users with different backgrounds. For example,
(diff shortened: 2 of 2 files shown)


fmaussion:
Any comment/suggestion from the larger xarray community?

jhamman:
I think we merge this to make it more visible. 

@fmaussion - we got a few comments asking if/how the core changes proposed in xarray would make using the library more difficult. Those worries seemed to be easily put to rest. Beyond that, things seemed mostly positive/quite.

shoyer:
Here are the responses on Twitter: https://twitter.com/shoyer/status/1021788358048010240

aseyboldt:
This roadmap sounds great!
We plan on using xarray to store all traces in pymc4.
The biggest item on my wish list for that is better support for plotting. I regularly convert traces to pandas and use the plotting functions there, or I convert it so that I can use seaborn.
Better support for hierarchical indexes sounds useful as well. They can be a bit surprising right now, and they can't be serialised to netCDF.
The documentation is quite important our use case. As it is, we are asking our users to learn Bayesian stats, diagnosing sampler issues and possibly some theano or tensorflow. When we switch to xarray, our users have to learn the basics of that as well. Most won't even have heard about netCDF. One thing in particular that I noticed is that many people seem to get confused about the difference between coordinates and dimensions at some point.

shoyer:
Thanks for the feedback!

@dcherian and @yohai have been making some nice improvements to plotting recently. It seems like each release has a few new features. Fortunately plotting goes in the category of "straightforward to improve without a major refactoring", which is probably why we didn't put it on the roadmap!

aseyboldt:
I noticed and appreciate those plotting additions :-)

dcherian:
@aseyboldt Thanks! Please file issues for any feature requests you might have.

Issue #2306, opened 2018-07-24
Timeouts errors on readthedocs


fmaussion:
We are reaching the 900s build time limit on readthedocs more often than not (https://readthedocs.org/projects/xray/builds/). I have the same problem with all of my OS projects. 

The bottleneck is the conda environment installation, which took  457s on the latest failed build. I'm going to try to spare some time in a subsequent PR, but we might have to get in touch with the RTD people to get more build time.

fmaussion:
This is actually a bigger problem since the last point releases did not build correctly: http://xarray.pydata.org/en/v0.10.8/ 
http://xarray.pydata.org/en/v0.10.7/

shoyer:
See also https://github.com/pydata/xarray/issues/2209

Recently I've only been able to get builds to complete successfully after repeatedly restarting them.

fmaussion:
closing in favor of https://github.com/pydata/xarray/issues/2209

PR #2310, opened 2018-07-24
Make RTD builds faster

fmaussion:
See https://github.com/pydata/xarray/issues/2306

This makes the builds a little bit faster. But we are still very close to the 900 s limit.

I tried to pin more packages this morning but this didn't work out. I'll merge this first (the latest build worked with this config, https://readthedocs.org/projects/xray/builds/) and try again at a later stage.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -26,7 +26,7 @@ print("python exec:", sys.executable)
 print("sys.path:", sys.path)
 for name in ('numpy scipy pandas matplotlib dask IPython seaborn '
              'cartopy netCDF4 rasterio zarr iris flake8 '
-             'sphinx-gallery cftime').split():
+             'sphinx_gallery cftime').split():
     try:
         module = importlib.import_module(name)
         if name == 'matplotlib':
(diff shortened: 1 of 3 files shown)


shoyer:
Thanks for looking into this! I've found a pretty high variance on RTD depending on network latency. With our current setup the build completes about one in four times.

fmaussion:
Yes, and now it comes that a C dependency is missing from the conda channels: http://xarray.pydata.org/en/fix-docs/auto_gallery/plot_rasterio_rgb.html#sphx-glr-auto-gallery-plot-rasterio-rgb-py

Will unpin everything and see if it works.

fmaussion:
OK so I managed to get a successful build will all packages unpinned this morning, but not after pinning them back to fixed versions (although you would think this makes things faster).

Since its always a few seconds that are missing I opened https://github.com/rtfd/readthedocs.org/issues/4432 by RTD to see if we can get an extension...

fmaussion:
See resolution here: https://github.com/rtfd/readthedocs.org/issues/4432

Issue #2284, opened 2018-07-12
interp over time coordinate


czr137:
Before I start, I'm very excited about the interp addition in 0.10.7. Great addition and thanks to @fujiisoup and @shoyer.

I see there was a bit of a discussion in the interp pull request,  #2104, about interpolating over times and that it was suggested to wait for use cases. I can think of an immediate use case in my line of work.
I frequently use regular gridded geophysical data (time, lat, lon), not unlike the sample tutorial air_temperature data,  and the data must be interpolated to line up with corresponding satellite measurements that are irregularly spaced in lat, lon and time.

Being able to interpolate in one quick step would be fantastic. For example:

```python
ds = xr.tutorial.load_dataset('air_temperature')
ds.interp(lat=60.5, lon=211, time='2013-01-01T03:14:37')
```
#### Problem description
Currently issues `TypeError: cannot perform reduce with flexible type`.

#### Desired Output
```
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    lat      float64 60.5
    lon      int64 211
    time    datetime64[ns] 2013-01-01T03:14:37
Data variables:
    air      float64 273.5
```

shoyer:
Thanks for bringing this up.

@fujiisoup actually implemented interpolation for datetime64 as part of writing `interp_like()` , which will be in the next release of xarray (0.10.8). With the current development version of xarray:
```
In [6]: import numpy as np

In [7]: ds.interp(lat=60.5, lon=211, time=np.datetime64('2013-01-01T03:14:37'))
Out[7]:
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    lat      float64 60.5
    lon      int64 211
    time     datetime64[ns] 2013-01-01T03:14:37
Data variables:
    air      float64 273.5
Attributes:
    Conventions:  COARDS
    title:        4x daily NMC reanalysis (1948)
    description:  Data is from NMC initialized reanalysis\n(4x/day).  These a...
    platform:     Model
    references:   http://www.esrl.noaa.gov/psd/data/gridded/data.ncep.reanaly...
```

However, you do currently need to wrap your indexer explicitly in a `np.datetime64` object. It would be nice to support casting strings to dates, somewhat like what we do in `.sel()`. This would be a good candidate for a pull request.

(I'll try to release 0.10.8 soon, maybe within the next week)

czr137:
Thanks, I will gladly exercise patience until then.

PR #2312, opened 2018-07-25
DOC: add interp example

raybellwaves:

I wanted to add an example as googling 'xarray interp' takes me to http://xarray.pydata.org/en/stable/generated/xarray.Dataset.interp.html.

However, there lots of great examples in http://xarray.pydata.org/en/stable/interpolation.html
Perhaps I should edit it to:
```
Examples
--------
See http://xarray.pydata.org/en/stable/interpolation.html
```

I was also thinking about adding a note about interpolating time as it came up in https://github.com/pydata/xarray/issues/2284 but not sure what would be the best way to do so.

### DIFF
--- a/doc/interpolation.rst
+++ b/doc/interpolation.rst
@@ -58,12 +58,12 @@ wrap it explicitly in a :py:func:`numpy.datetime64` object.
                            [('time', pd.date_range('1/1/2000', '1/3/2000', periods=2))])
     da_dt64.interp(time=np.datetime64('2000-01-02'))
 
-The interpolated data can be merged into the original :py:class:`~xarray.DataArray`
-using :py:meth:`~xarray.concat`.
+The interpolated data can be merged into the original :py:class:`~xarray.DataArray` 
+by specifing the time periods required.
 
 .. ipython:: python
 
-    xr.concat([da_dt64, da_dt64.interp(time=np.datetime64('2000-01-02'))], dim='time')
+    da_dt64.interp(time=pd.date_range('1/1/2000', '1/3/2000', periods=3))
 
 .. note::
 
--- a/doc/interpolation.rst
+++ b/doc/interpolation.rst
@@ -48,6 +48,23 @@ array-like, which gives the interpolated result as an array.
     # interpolation
     da.interp(time=[2.5, 3.5])
 
+
+To interpolate data with a :py:func:`numpy.datetime64` coordinate you have to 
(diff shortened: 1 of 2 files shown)


fujiisoup:
Thanks for adding this to doc.
It is very welcome if you could add an example for interp with datetime.

To do so, it would be best place to put it in here
https://github.com/pydata/xarray/blob/master/doc/interpolation.rst

raybellwaves:
@fujiisoup I would like to add another example on how to update a `DataArray` with the interpolated data. I looked in http://xarray.pydata.org/en/stable/combining.html but couldn't work out how best to do it.

I tried using `merge` but it requires the `DataArray`'s to be named.

I used `concat` in this example but it is not correct as I would like the coordinates to be in order (in-place may be the right word). 

```
import xarray as xr
import pandas as pd
import numpy as np

da_dt64 = xr.DataArray([1, 3], [('time', pd.date_range('1/1/2000', '1/3/2000', periods=2))])
a = da_dt64.interp(time=np.datetime64('2000-01-02'))
xr.concat([da_dt64, a], dim='time')
#<xarray.DataArray (time: 3)>
#array([ 1.,  3.,  2.])
#Coordinates:
#  * time     (time) datetime64[ns] 2000-01-01 2000-01-03 2000-01-02
```

fujiisoup:
Thanks.

I think you can do
```python
da_dt64.interp(time=pd.date_range('1/1/2000', '1/3/2000', periods=3))
```

doc/interpolation.rst
```diff
@@ -48,6 +48,23 @@ array-like, which gives the interpolated result as an array.
     # interpolation
     da.interp(time=[2.5, 3.5])
 
+
+To interpolate data with a :py:func:`numpy.datetime64` coordinate you have to 
+wrap it explicitly in a :py:func:`numpy.datetime64` object.
```
fujiisoup: It is currently true, but I think we could support `da_dt64.interp(time='2000-01-02')` in the future.
Do you mind to send another PR to do this?

raybellwaves: I'm trying to wrap up another PR. Unfortunately, I don't want to over promise and may not be able to get to this until a couple of weeks. If you have ideas of how to do this we can move the discussion to https://github.com/pydata/xarray/issues/2284

fujiisoup:
Thanks, @raybellwaves.

Issue #2316, opened 2018-07-26
rasterio released v1 as stable


fmaussion:
conda-forge now ships v1.0.1 per default.

After two years of betas and release candidates this is very welcome!

We have very little code specifically handling pre-v1 and post-v1, but we should keep it around for a couple more months.

Will update the CI to reflect this change.

dcherian:
Closed by #2589

PR #2317, opened 2018-07-26
Remove test on rasterio rc and test for 0.36 instead

fmaussion:
See https://github.com/pydata/xarray/issues/2316


fmaussion:
The failing tests seem unrelated

PR #2325, opened 2018-07-28
interp() now accepts date strings as desired co-ordinate locations

### LINKED ISSUE #2284: interp over time coordinate

dcherian:
- [x] Closes #2284

```python
da = xr.DataArray([1, 5],
                  dims=['time'],
                  coords={'time': [np.datetime64('2014-05-06'),
                                   np.datetime64('2014-05-10')]})

da.interp(time='2014-05-07')
```
```
<xarray.DataArray ()>
array(2.)
Coordinates:
    time     datetime64[ns] 2014-05-07
```


xarray/core/dataset.py
```diff
@@ -1328,6 +1330,12 @@ def _validate_indexers(self, indexers):
                 raise TypeError('cannot use a Dataset as an indexer')
             else:
                 v = np.asarray(v)
+
+                if (v.dtype.kind == 'U'
```
dcherian: only the first half is really required (the tests all pass), but I added the second bit to be safe.

fujiisoup:
Thanks, @dcherian .

It looks already great, but can you add a bit more test cases to keep it working with the future modification?
It would be nicer if we can check
+ If `xnew` is a single datetime unicode such as '2014-05-07`
+ If `xnew` is pd.datetime

I think they will be passing though.

fujiisoup:
Can you merge master? It will solve some failures in CI.
I also noticed tests in Python=2.7 are failing.

xarray/tests/test_interp.py
```diff
@@ -462,19 +462,20 @@ def test_interp_like():
...
 
 @requires_scipy
-def test_datetime():
-    da = xr.DataArray(np.random.randn(24), dims='time',
+@pytest.mark.parametrize('x_new, expected', [
+    (pd.date_range('2000-01-02', periods=3), [1, 2, 3]),
+    (np.array([np.datetime64('2000-01-01T12:00'),
+               np.datetime64('2000-01-02T12:00')]), [0.5, 1.5]),
+    (['2000-01-01T12:00', '2000-01-02T12:00'], [0.5, 1.5]),
+    (['2000-01-01T12:00'], 0.5)  # needs to be list, see #2327
```
fujiisoup: Can you remove the comment and add a test with a string (not a list)?

xarray/tests/test_interp.py
```diff
@@ -462,19 +462,30 @@ def test_interp_like():
 
 
 @requires_scipy
-def test_datetime():
-    da = xr.DataArray(np.random.randn(24), dims='time',
+@pytest.mark.parametrize('x_new, expected', [
```
fujiisoup: I think `requires_scipy` does not work with another decorator. (We need to improve it though)

A workaround is to `pytest.importorskip('scipy')` *inside* the function.

fujiisoup: OK. Forget my comment. I will be merging it later.

fujiisoup:
Thanks, @dcherian. merging.

xarray/tests/test_interp.py
shoyer: It looks it would still be a good idea to fix the scalar case?

xarray/tests/test_interp.py
```diff
@@ -462,19 +462,31 @@ def test_interp_like():
...
-    actual = da.interp(time=x_new)
-    assert_allclose(actual.isel(time=0).drop('time'),
-                    0.5 * (da.isel(time=0) + da.isel(time=1)))
-    assert_allclose(actual.isel(time=1).drop('time'),
-                    0.5 * (da.isel(time=1) + da.isel(time=2)))
+    assert_allclose(actual, expected_da)
+
+
+@requires_scipy
+def test_datetime_single_string():
```
dcherian: @shoyer the scalar case works and is tested here. I didn't want to make the previous test too complicated (it fails because the result of the scalar case has no dims; but the test checks both dims and coords).

shoyer: OK, cool :)

PR #2261, opened 2018-07-01
xarray.backends refactor

shoyer:
A major refactor of xarray backend classes:

- A new least-recently-used cache facilitates keeping track of open files rather than requiring them to be reopened.
- A number of internal simplifications/clean-ups that should make backends easier to write and maintain:
     - `PickleByReconstructionWrapper` and `DataStorePickleMixin` have been eliminated in favor of `CachingFIleManager`.
     - Backends no longer store any state beyond their wrapped files. This fixes a bug where dask graphs were accidentally being serialized in pickled backend classes (oops!).
     - Locks are now setup by backend classes directly, rather than separately in `to_netcdf`/`open_dataset`.

xref #2121 
fixes https://github.com/pydata/xarray/issues/1738
fixes https://github.com/pydata/xarray/issues/2376

 - [x] Tests added
 - [x] Tests passed

Benchmark numbers:
```
    before     after       ratio
  [66a8f8dd] [c8fbadcc]
+   65.27ms    89.91ms      1.38  dataset_io.IOReadSingleNetCDF3.time_load_dataset_netcdf4
+  866.65ms   960.37ms      1.11  dataset_io.IOReadMultipleNetCDF3Dask.time_load_dataset_netcdf4_with_time_chunks_multiprocessing
-  676.18ms   610.92ms      0.90  dataset_io.IOReadMultipleNetCDF3Dask.time_load_dataset_netcdf4_with_block_chunks
-   61.16ms    55.20ms      0.90  dataset_io.IOReadSingleNetCDF4.time_load_dataset_netcdf4
-  167.54ms   150.41ms      0.90  dataset_io.IOReadSingleNetCDF3Dask.time_load_dataset_netcdf4_with_time_chunks
-     1.11s   981.78ms      0.89  dataset_io.IOReadMultipleNetCDF3Dask.time_load_dataset_scipy_with_time_chunks
-  247.82ms   219.50ms      0.89  dataset_io.IOReadSingleNetCDF3Dask.time_load_dataset_netcdf4_with_block_chunks_vindexing
-   81.04ms    71.70ms      0.88  dataset_io.IOReadMultipleNetCDF4Dask.time_open_dataset_netcdf4_with_block_chunks_multiprocessing
-   79.17ms    70.01ms      0.88  dataset_io.IOReadMultipleNetCDF4Dask.time_open_dataset_netcdf4_with_time_chunks_multiprocessing
[...]

PR #2328, opened 2018-07-29
Silence some warnings.

dcherian:

Remove some warnings.

### DIFF
--- a/xarray/core/missing.py
+++ b/xarray/core/missing.py
@@ -181,9 +181,9 @@ def get_clean_interp_index(arr, dim, method, use_coordinate=True, **kwargs):
                 # The division can change the nearest-neighbour
                 # when compared to pandas (which does not divide).
                 # Let's keep that compatitibility
-                index = (index - index.mean())
+                index = (index - index.min())
                 if len(index) > 1:
-                    index /= np.max(np.abs(index))
+                    index /= index.std()
 
         except (TypeError, ValueError):
             # pandas raises a TypeError
@@ -417,13 +417,11 @@ def _floatize_x(x, new_x):
         # We assume that the most of the bits are used to represent the
         # offset (min(x)) and the variation (x - min(x)) can be
         # represented by float.
-        # Let's be defensive and always rescale x to be in [0, 1]
-        # Remove minimum instead of mean allows us to handle datetime
-        xfloat = (x[i] - np.min(x[i])).astype(np.float64)
-        newxfloat = (new_x[i] - np.min(x[i])).astype(np.float64)
-        xmax = np.max(xfloat)
-        x[i] = xfloat / xmax
-        new_x[i] = newxfloat / xmax
+        # Let's be defensive and always rescale (x)
+        xmin = np.min(x[i])
+        xstd = np.std(x[i].astype(np.float64))
+        x[i] = (x[i] - xmin).astype(np.float64) / xstd
+        new_x[i] = (new_x[i] - xmin).astype(np.float64) / xstd
     return x, new_x
 
 
--- a/xarray/core/missing.py
+++ b/xarray/core/missing.py
@@ -153,7 +153,7 @@ def _apply_over_vars_with_dim(func, self, dim=None, **kwargs):
     return ds
 
 
-def get_clean_interp_index(arr, dim, method, use_coordinate=True, **kwargs):
+def get_clean_interp_index(arr, dim, use_coordinate=True, **kwargs):
     '''get index to use for x values in interpolation.
 
     If use_coordinate is True, the coordinate that shares the name of the
@@ -176,15 +176,6 @@ def get_clean_interp_index(arr, dim, method, use_coordinate=True, **kwargs):
         # raise if index cannot be cast to a float (e.g. MultiIndex)
         try:
             index = index.values.astype(np.float64)
-            if method != 'nearest':
-                # rescale index to avoid overflow/underflow
-                # The division can change the nearest-neighbour
-                # when compared to pandas (which does not divide).
-                # Let's keep that compatitibility
-                index = (index - index.min())
-                if len(index) > 1:
-                    index /= index.std()
-
         except (TypeError, ValueError):
             # pandas raises a TypeError
             # xarray/nuppy raise a ValueError
@@ -211,8 +202,7 @@ def interp_na(self, dim=None, use_coordinate=True, method='linear', limit=None,
         valids = _get_valid_fill_mask(self, dim, limit)
 
     # method
-    index = get_clean_interp_index(self, dim, method=method,
-                                   use_coordinate=use_coordinate,
+    index = get_clean_interp_index(self, dim, use_coordinate=use_coordinate,
                                    **kwargs)
     interp_class, kwargs = _get_interpolator(method, **kwargs)
     interpolator = partial(func_interpolate_na, interp_class, **kwargs)
--- a/xarray/core/missing.py
+++ b/xarray/core/missing.py
@@ -412,15 +412,16 @@ def _floatize_x(x, new_x):
     x = list(x)
     new_x = list(new_x)
     for i in range(len(x)):
-        if x[i].dtype.kind in 'Mm':
-            # Scipy casts coordinates to np.float64, which is not accurate
-            # enough for datetime64 (uses 64bit integer).
-            # We assume that the most of the bits are used to represent the
-            # offset (min(x)) and the variation (x - min(x)) can be
-            # represented by float.
-            xmin = np.min(x[i])
-            x[i] = (x[i] - xmin).astype(np.float64)
-            new_x[i] = (new_x[i] - xmin).astype(np.float64)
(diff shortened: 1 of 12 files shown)


dcherian:
Most of the remaining warnings are: 
`
RuntimeWarning: Cannot close a netcdf_file opened with mmap=True, when netcdf_variables or arrays referring to its data still exist. All data arrays obtained from such files refer directly to data on disk, and must be copied before the file can be cleanly closed. (See netcdf_file docstring for more information on mmap.)
`

xarray/tests/test_dataarray.py
```diff
@@ -693,7 +693,7 @@ def test_isel_fancy(self):
         da.isel(time=(('points',), [1, 2]), x=(('points',), [2, 2]),
                 y=(('points',), [3, 4]))
         np.testing.assert_allclose(
-            da.isel_points(time=[1], x=[2], y=[4]).values.squeeze(),
+            da.isel(time=[1], x=[2], y=[4]).values.squeeze(),
```
fujiisoup: This does not give the same result.
I think we can just ignore the warning by `@pytest.mark.filterwarnings`.

xarray/tests/test_dataset.py
```diff
@@ -1417,7 +1420,7 @@ def test_sel_fancy(self):
         assert_identical(actual['b'].drop('y'), idx_y['b'])
 
         with pytest.raises(KeyError):
-            data.sel_points(x=[2.5], y=[2.0], method='pad', tolerance=1e-3)
+            data.sel(x=[2.5], y=[2.0], method='pad', tolerance=1e-3)
```
fujiisoup: Here also, we can just ignore the deprecation warning.

jhamman:
> Most of the remaining warnings are:
> RuntimeWarning: Cannot close a netcdf_file opened with mmap=True, when netcdf_variables or arrays referring to its data still exist. All data arrays obtained from such files refer directly to data on disk, and must be copied before the file can be cleanly closed. (See netcdf_file docstring for more information on mmap.)


@dcherian - take a look here: https://github.com/pydata/xarray/pull/2261#issuecomment-408642640. These may be fixed soon.

Also, xrefing: #1652, #1657

dcherian:
@jhamman, @fujiisoup, @shoyer Thanks!

xarray/core/missing.py
```diff
@@ -176,6 +176,15 @@ def get_clean_interp_index(arr, dim, use_coordinate=True, **kwargs):
         # raise if index cannot be cast to a float (e.g. MultiIndex)
         try:
             index = index.values.astype(np.float64)
+            if method != 'nearest':
```
dcherian: rescaling is necessary to avoid underflow/overflow RuntimeWarnings with datetime64. I've made a similar change to interp() below.

xarray/core/missing.py
```diff
@@ -176,6 +176,15 @@ def get_clean_interp_index(arr, dim, use_coordinate=True, **kwargs):
...
+                # rescale index to avoid overflow/underflow
+                # The division can change the nearest-neighbour
+                # when compared to pandas (which does not divide).
+                # Let's keep that compatitibility
+                index = (index - index.min())
+                if len(index) > 1:
+                    index /= index.std()
```
shoyer: Standard deviation is *probably* fine here, but I wonder if it would be better to *guarantee* that are all values fall within some fixed range [-MAX, MAX], e.g., by using max() instead:
```
index = index - index.mean()
index /= EPSILON * np.maximum(index.max(), 1)
```
where EPSILON is some constant, e.g., 1e-8.

xarray/core/missing.py
```diff
@@ -181,9 +181,9 @@ def get_clean_interp_index(arr, dim, method, use_coordinate=True, **kwargs):
                 # The division can change the nearest-neighbour
                 # when compared to pandas (which does not divide).
                 # Let's keep that compatitibility
-                index = (index - index.min())
+                index = (index - index.mean())
                 if len(index) > 1:
-                    index /= index.std()
+                    index /= np.max(np.abs(index))
```
shoyer: This still needs a fix to account for possibly dividing by zero.

Also, I don't think you need abs() here because subtracting by mean() already centers the data.

dcherian: The `abs()` makes sure that it's always within [-1, 1] right? e.g. if `index = [-2, 0, 0.9]` and as long as `index` is not all zeroes, it should always work? :thinking: 
Sorry, I feel like I'm missing something.

shoyer: Yes, this would work fine unless the array is all zeros/constant.

It still might be worth considering that possibility, though at this point I'm pretty sure we are just refining total edge case behavior.

Maybe:
```
min_value = index.min()
max_value = index.max()
range_ = max_value - min_value
if range_:
    midpoint = (min_value + max_value) / 2
    index = index - midpoint
    index /= range_
```

shoyer: The other option is to not really worry about this at all, besides silencing the warning. float64 is about the best we can do as far as precision, and it's a simple fact that you cannot represent every int64 value exactly in float64. In most cases this isn't a problem.

WeatherGod: The people doing lightning research probably don't think that scenario is all that unusual (don't know for sure, I just remember a conversation with a lightning researcher when he complained about matplotlib's lack of support for nanosecond tickers).

dcherian: > The other option is to not really worry about this at all, besides silencing the warning. float64 is about the best we can do as far as precision, and it's a simple fact that you cannot represent every int64 value exactly in float64. In most cases this isn't a problem.

Ya, so the choice is between using `filterwarnings` or rescaling to avoid the warning. I have a mild preference for rescaling because it "solves" the problem. Hopefully, the lightning research people will chime in / send in a PR if this is actually an issue.

shoyer: Let's just silence the warning here. I don't think rescaling actually fixes anything, and it's less efficient than silencing it.

xarray/tests/test_dataarray.py
```diff
@@ -3532,6 +3532,8 @@ def test_rolling_reduce(da, center, min_periods, window, name):
 @pytest.mark.parametrize('min_periods', (None, 1, 2, 3))
 @pytest.mark.parametrize('window', (1, 2, 3, 4))
 @pytest.mark.parametrize('name', ('sum', 'max'))
+@pytest.mark.filterwarnings('ignore:Using a non-tuple sequence')
```
dcherian: Silences a warning that seems to be thrown by code in `bottleneck`

```
xarray/tests/test_dataarray.py::test_rolling_reduce_nonnumeric[sum-1-3-False]
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/bottleneck/slow/move.py:149: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    nidx1 = n[idx1]
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/bottleneck/slow/move.py:150: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    nidx1 = nidx1 - n[idx2]
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/bottleneck/slow/move.py:152: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    idx[idx1] = nidx1 < min_count
  /home/travis/miniconda/envs/test_env/lib/python3.6/site-packages/bottleneck/slow/move.py:153: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
    idx[idx3] = n[idx3] < min_count
```

shoyer: Good case -- can you file a report in bottleneck? I expect this would be easy to fix.

dcherian: OK. Filed a report upstream https://github.com/kwgoodman/bottleneck/issues/194 and reverted this commit.

xarray/tests/test_dask.py
```diff
@@ -24,8 +24,11 @@
 
 class DaskTestCase(TestCase):
     def assertLazyAnd(self, expected, actual, test):
-        with dask.set_options(get=dask.get):
+
+        with (dask.config.set(get=dask.get) if hasattr(dask, 'config')
+              else dask.set_options(get=dask.get)):
```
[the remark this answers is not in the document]
dcherian: Ya it's basically a version check. I've made it an explicit version check now.

Issue #2056, opened 2018-04-13
Warning on supplying a Dataset to the Dataset constructor


max-sixty:
```python

In [1]: import xarray as xr

In [3]: ds=xr.Dataset({'a':xr.DataArray([1,2,3])})

In [4]: xr.Dataset(ds)
/Users/maximilian/drive/workspace/xarray/xarray/core/dataset.py:373: FutureWarning: iteration over an xarray.Dataset will change in xarray v0.11 to only include data variables, not coordinates. Iterate over the Dataset.variables property instead to preserve existing behavior in a forwards compatible manner.
  both_data_and_coords = [k for k in data_vars if k in coords]
Out[4]:
<xarray.Dataset>
Dimensions:  (dim_0: 3)
Dimensions without coordinates: dim_0
Data variables:
    a        (dim_0) int64 1 2 3
```
#### Problem description

Currently we run `both_data_and_coords = [k for k in data_vars if k in coords]`, where `data_vars` is the first arg to the constructor. So, when `data_vars` is a `Dataset`, we'll raise a warning.

More importantly: this raises the question of how we should handle `xr.Dataset(ds)`
Currently we retain the `coords`, but discard `attrs`. That seems arbitrary?
I'd have to check, but when we change `__iter__` to only reference `data_vars`, we may also discard `coords`

I don't have a strong view. If you think that `Foo(foo)` should always return `foo` (which is often the case in pandas), then we should retain the full object.
If you think that `foo` is supplied to the `data_vars` arg, then retaining only the `data_vars` seems reasonable

#### Output of ``xr.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: a9d1f3a36229636f0d519eb36a8d4a7c91f6e1cd
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.4.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.2+dev31.g0a24429
pandas: 0.22.0
numpy: 1.14.2
scipy: 1.0.0
netCDF4: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.2
distributed: None
matplotlib: 2.1.2
cartopy: None
seaborn: 0.8.1
setuptools: 39.0.1
pip: 9.0.3
conda: None
pytest: 3.5.0
[...]

PR #2330, opened 2018-07-31
Supplying a dataset to the dataset constructor

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/2056

- Do we want to raise if coords / attrs are passed in addition to a Dataset? Do we want to warn for a couple versions first?
- Is this the best way to implement this? Do we want a consistent method to "make this object a shallow copy of the object that's been passed in"?


max-sixty:
> Do we want to raise if coords / attrs are passed in addition to a Dataset? Do we want to warn for a couple versions first?

An alternative approach - more friendly but less explicit - is for any other argument (coords, attrs) supplied to overwrite the Dataset's

shoyer:
> An alternative approach - more friendly but less explicit - is for any other argument (coords, attrs) supplied to overwrite the Dataset's

I think this would be slightly preferred, and also more consistent with the model of a Dataset as only iterating over `data_vars`. This suggests that at the very least, `Dataset(ds)` and `Dataset(ds.data_vars)` should be consistent after we finish the changes described in https://github.com/pydata/xarray/issues/884. Note that `ds.data_vars` will pull in coordinates associated with the DataArray objects.

To be more precise, I would probably make:
```
ds2 = Dataset(ds1, coords, attrs)
```
equivalent to
```
ds2 = ds1.copy()
ds2.coords.update(coords)
ds2.attrs.update(attrs)
```

max-sixty:
OK good!

And, to confirm, you think it's OK that :

`xr.Dataset(ds, coords=some_coords).coords != some_coords`

max-sixty:
> And, to confirm, you think it's OK that :

Moving ahead with this, please let me know any objections from anyone

fujiisoup:
> `xr.Dataset(ds, coords=some_coords).coords != some_coords`

I thought that `xr.Dataset(ds, coords=some_coords).coords == some_coords`.
I think Dataset should be regarded as a dict of dataarrays (data_vars), and thus
`xr.Dataset(ds, coords=some_coords)` should be equivalent to `xr.Dataset(ds.data_vars, coords=some_coords)`.

shoyer:
Now I'm having second thoughts. It does make me a little nervous to add special cases that distinguish between `xr.Dataset(ds, coords=some_coords)` and `xr.Dataset(ds.data_vars, coords=some_coords)`.

Maybe my proposed solution above is too complex, and it would be better simply to default the `coords` and `attrs` argument to those arguments from a passed Dataset object. This is basically what we do for DataArray.

That said, I do think we should try to resist the notion that every class constructor must also work for coercion to that class. This tends to result in a slightly confused interface. A separate `as_dataset()` function (even in user code) could give exactly the desired result in only be a few lines of code.

fmaussion:
From https://github.com/pydata/xarray/issues/2056 and this discussion I don't really understand the use case for this constructor - maybe an example of a real use case that could be added to the doc would help?

max-sixty:
@fmaussion good question

The main reason I think this is worthwhile is the change in behavior between 0.10.x and 0.11.0, when iterating over a Dataset will iterate only over its `data_vars`, and so will change `Dataset(ds)` from `ds` to `Dataset(ds.data_vars)`. 
I don't have a strong view on which option we choose, but I do think we should have a deliberate decision rather than making a breaking-change-by-coincidence.

To answer your question though, we hit this when inheriting from a Dataset: the object is initialized with a dataset, which is super-ed up to the Dataset constructor. (I realize this is narrow!). Alternatively, there is another repo that linked to the issue [here](https://github.com/jupyter-widgets/ipyleaflet/pull/178)

max-sixty:
I think we should do _something_ here, rather than change behavior by coincidence.

@shoyer do you want to make the decision? 

I would vote +0.1 for the current version in the PR. V happy to be outweighed!

max-sixty:
Any votes?

I think the current version dominates the current master (i.e. it doesn't disable any useful functionality, it adds some clarity). So I would vote to merge unless anyone has any alternative suggestions (which I will gladly implement)

Thanks team

max-sixty:
Yes, I agree - given the reluctance on the principle around "calling a constructor on an object of the same type needs to work" - it's right not to make this change. Thanks for holding that principle.

I'll close this; let me know if there's any change around this I make in its place.

PR #2333, opened 2018-08-01
Let test_unicode_data pass on big-endian systems too

ginggs:


fujiisoup:
Thanks for sending the patch.
Looks good to me.

Can you merge the current master?
The test error in travis (actually the second one) will be fixed by this.

PR #2339, opened 2018-08-03
Use sorted fixtures to ensure that test gathering is consistent.

ttung:
pytest-xdist expects that test gathering produces the same results and in the same order.  Because sets, prior to python 3.7, do not guarantee ordering, this can cause pytest -n (from pytest-xdist) to fail.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -61,6 +61,7 @@ Bug fixes
   attribute being set.
   (:issue:`2201`)
   By `Thomas Voigt <https://github.com/tv3141>`_.
+- Tests can be run in parallel with pytest-xdist
 
 .. _whats-new.0.10.8:
 
--- a/xarray/tests/test_coding_times.py
+++ b/xarray/tests/test_coding_times.py
@@ -16,10 +16,12 @@ from . import (assert_array_equal, has_cftime_or_netCDF4,
                requires_cftime_or_netCDF4, has_cftime, has_dask)
 
 
-_NON_STANDARD_CALENDARS = {'noleap', '365_day', '360_day',
-                           'julian', 'all_leap', '366_day'}
-_ALL_CALENDARS = _NON_STANDARD_CALENDARS.union(
-    coding.times._STANDARD_CALENDARS)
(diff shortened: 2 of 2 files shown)


doc/whats-new.rst
```diff
@@ -61,6 +61,7 @@ Bug fixes
   attribute being set.
   (:issue:`2201`)
   By `Thomas Voigt <https://github.com/tv3141>`_.
+- Tests can be run in parallel with pytest-xdist
```
max-sixty: You can give yourself attribution IYL

xarray/tests/test_coding_times.py
```diff
@@ -16,10 +16,12 @@
...
 
 
-_NON_STANDARD_CALENDARS = {'noleap', '365_day', '360_day',
-                           'julian', 'all_leap', '366_day'}
-_ALL_CALENDARS = _NON_STANDARD_CALENDARS.union(
-    coding.times._STANDARD_CALENDARS)
+_NON_STANDARD_CALENDARS_SET = {'noleap', '365_day', '360_day',
+                               'julian', 'all_leap', '366_day'}
+_ALL_CALENDARS = sorted(_NON_STANDARD_CALENDARS_SET.union(
+    coding.times._STANDARD_CALENDARS))
```
max-sixty: These could be lists throughout (a single cast for `_STANDARD_CALENDARS`) with the union as a simple `+`, but this works fine

xarray/tests/test_coding_times.py
```diff
@@ -161,7 +163,7 @@ def test_decode_cf_datetime_non_iso_strings():
 @pytest.mark.skipif(not has_cftime_or_netCDF4, reason='cftime not installed')
 @pytest.mark.parametrize(
     ['calendar', 'enable_cftimeindex'],
-    product(coding.times._STANDARD_CALENDARS, [False, True]))
+    product(_STANDARD_CALENDARS, [False, True]))
```
max-sixty: For any future pytesters - no need to make the `product` - put a `@...parametrize` for each param and pytest will do the product for you

PR #2342, opened 2018-08-05
apply_ufunc now raises a ValueError when the size of input_core_dims is inconsistent with number of argument

### LINKED ISSUE #2341: apply_ufunc silently neglects arguments if `len(input_core_dims) < args`
From [SO](https://stackoverflow.com/questions/51680659/disparity-between-result-of-numpy-gradient-applied-directly-and-applied-using-xa/51690873#51690873)

In the following script, the second argument is silently neglected,
```python
da = xr.DataArray(np.random.randn(4, 3), coords={'x': [5, 7, 9, 11]}, dims=('x', 'y'))
xr.apply_ufunc(np.gradient, da, da.coords['x'].values, kwargs={'axis': -1},
               input_core_dims=[['x']], output_core_dims=[['x']],
               output_dtypes=[da.dtype])
```
This is because we need to the same number of `input_core_dims` to the number of arguments,

https://github.com/pydata/xarray/blob/56381ef444c5e699443e8b4e08611060ad5c9507/xarray/core/computation.py#L535-L538

The correct scipt might be
`input_core_dims=[['x']]` -> `input_core_dims=[['x'], []]`.

I think we can raise a more friendly error if the size of `input_core_dims` are wrong.

EDIT:
Or we can automatically insert an empty tuple or `None` for non-xarray object?
`input_core_dims` for non-xarray object sounds a little strange.

fujiisoup:
- [x] Closes #2341
 - [x] Tests added
 - [x] Tests passed

Now raises a ValueError when the size of input_core_dims is inconsistent with number of argument.

### DIFF
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -920,6 +920,11 @@ def apply_ufunc(func, *args, **kwargs):
 
     if input_core_dims is None:
         input_core_dims = ((),) * (len(args))
+    if len(input_core_dims) != len(args):
+        raise ValueError(
+            'input_core_dims must be None or a tuple with the length same to '
+            'the number of arguments. Given input_core_dims: {}, '
+            'number of args: {}.'.format(input_core_dims, len(args)))
 
     signature = _UFuncSignature(input_core_dims, output_core_dims)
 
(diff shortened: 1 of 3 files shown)


xarray/core/computation.py
```diff
@@ -920,6 +920,11 @@ def earth_mover_distance(first_samples,
 
     if input_core_dims is None:
         input_core_dims = ((),) * (len(args))
+    if len(input_core_dims) != len(args):
```
shoyer: This could use `elif` instead -- there's no need to check this if `input_core_dims` was just assigned.

PR #2344, opened 2018-08-05
FutureWarning: creation of DataArrays w/ coords Dataset

hmaarrfk:
Previously, this would raise a:

FutureWarning:
iteration over an xarray.Dataset will change in xarray v0.11 to
only include data variables, not coordinates. Iterate over the
Dataset.variables property instead to preserve existing behavior
in a forwards compatible manner.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -60,8 +60,13 @@ def _infer_coords_and_dims(shape, coords, dims):
     new_coords = OrderedDict()
 
     if utils.is_dict_like(coords):
-        for k, v in coords.items():
-            new_coords[k] = as_variable(v, name=k)
+        if isinstance(coords, Dataset):
+            for k in coords.variables:
+                v = coords.variables[k]
+                new_coords[k] = as_variable(v, name=k)
+        else:
+            for k, v in coords.items():
+                new_coords[k] = as_variable(v, name=k)
     elif coords is not None:
         for dim, coord in zip(dims, coords):
             var = as_variable(coord, name=dim)
--- a/xarray/tests/test_creation.py
+++ b/xarray/tests/test_creation.py
@@ -2,10 +2,11 @@ import xarray as xr
 import numpy as np
 import warnings
 
+
 def test_coords_are_dataset():
     # Setup an array with coordinates
     n = np.zeros(3)
-    coords={'x': np.arange(3)}
+    coords = {'x': np.arange(3)}
     c = xr.Dataset(coords=coords)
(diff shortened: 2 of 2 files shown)


hmaarrfk:
I came across this when serializing/deserializing my coordinates to a json file.

shoyer:
I'm not sure we want to add this special case.

Logically, if a iterating over a Dataset only includes `data_vars` (as of xarray v0.11), then it would seem unexpected to use `variables` instead when a Dataset is passed as coords into the DataArray constructor.

If you want to copy coordinates `coords` or `variables`, you can do that explicitly, e.g., `DataArray(array, coords=dataset.coords)`.

hmaarrfk:
It seems like this warning isn't benign though. I will take your suggestion though (`coords=dataset.coords`)

I feel like I'm not the only one who probably did this. 
Should you raise an other warning explicitly?

hmaarrfk:
Is there a better way to serialize coordinates only?

hmaarrfk:
Maybe the issue that I am facing is that I want to deal with the storage of my metadata and data seperately.

I used to have my own library that was replicating much of xarray's functionality, but your code is much nicer than anything I would be able to write in a finite time. :smile: 

Following the information here: http://xarray.pydata.org/en/stable/data-structures.html#coordinates-methods

Currently, my serialization pipeline is:
```python
import xarray as xr
import numpy as np

# Setup an array with coordinates
n = np.zeros(3)
coords={'x': np.arange(3)}
m = xr.DataArray(n, dims=['x'], coords=coords)

coords_dataset_dict = m.coords.to_dataset().to_dict()
coords_dict = coords_dataset_dict['coords']

# Read/Write dictionary to JSON file

# This works, but I'm essentially creating an emtpy dataset for it
coords_set = xr.Dataset.from_dict(coords_dataset_dict)
coords2 = coords_set.coords  # so many `coords` :D
m2 = xr.DataArray(np.zeros(shape=m.shape), dims=m.dims, coords=coords2)
# I used to just pass the dataset to "coords"
m3 = xr.DataArray(np.zeros(shape=m.shape), dims=m.dims, coords=coords_set)
```

shoyer:
If you really want to save the data / metadata separately for a DataArray, you might as well roll your own version of `to_dict`/`from_dict`. The code for these methods is remarkably simple, all using public APIs: 
https://github.com/pydata/xarray/blob/ded0a684136540962bcc409e6272b1cebb5af30a/xarray/core/dataarray.py#L1662-L1674

hmaarrfk:
Thanks!

PR #2350, opened 2018-08-07
DOC: Add xskillscore to project lists

### LINKED ISSUE #1115: Feature request: Compute cross-correlation (similar to pd.Series.corr()) of gridded data
As a earth scientist regularly dealing with 3D data (time, latitude, longitude), I believe it would be great to be able to perform cross-correlation on DataArrays by specifying the axis. It's usage could look like: a.corr(b, axis = 0). It would be even more useful if the two arrays need not have the same dimensions (e.g. 'b' could be a time series). 

Currently, the only way to compute this that I am aware of, is by looping through each grid, converting the time series to pd.Series(), and then computing the correlation. This takes a long time. 
Would also appreciate suggestions to a faster algorithm.

raybellwaves:
Potentially fixes #1115 as I have a correlation function. http://xarray.pydata.org/en/stable/dask.html?highlight=spearman#automatic-parallelization probably fixes #1115 as well


dcherian:
Thanks!

Issue #2332, opened 2018-07-31
Test failures on master with DataArray.to_cdms2


shoyer:
See https://travis-ci.org/pydata/xarray/jobs/410459646

Example failure:
```
=================================== FAILURES ===================================
_________________ TestDataArray.test_to_and_from_cdms2_classic _________________
self = <xarray.tests.test_dataarray.TestDataArray testMethod=test_to_and_from_cdms2_classic>
    def test_to_and_from_cdms2_classic(self):
        """Classic with 1D axes"""
        pytest.importorskip('cdms2')
    
        original = DataArray(
            np.arange(6).reshape(2, 3),
            [('distance', [-2, 2], {'units': 'meters'}),
             ('time', pd.date_range('2000-01-01', periods=3))],
            name='foo', attrs={'baz': 123})
        expected_coords = [IndexVariable('distance', [-2, 2]),
                           IndexVariable('time', [0, 1, 2])]
        actual = original.to_cdms2()
>       assert_array_equal(actual, original)
E       ValueError: 
E       error during assertion:
E       
E       Traceback (most recent call last):
E         File "/home/travis/miniconda/envs/test_env/lib/python2.7/site-packages/numpy/testing/_private/utils.py", line 752, in assert_array_compare
E           x, y = x[~flagged], y[~flagged]
E         File "/home/travis/miniconda/envs/test_env/lib/python2.7/site-packages/cdms2/avariable.py", line 1177, in __getitem__
E           speclist = self._process_specs([key], {})
E         File "/home/travis/miniconda/envs/test_env/lib/python2.7/site-packages/cdms2/avariable.py", line 938, in _process_specs
E           if Ellipsis in specs:
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
E       
E       
E       Arrays are not equal
E        x: TransientVariable([[0, 1, 2],
E                          [3, 4, 5]])
E        y: array([[0, 1, 2],
E              [3, 4, 5]])
```

jhamman:
I think @stefraynaud worked on the cdms2 converters last. Any thoughts on what's going on here?

shoyer:
I suppose we probably just need to convert these TransientVariable objects explicitly into numpy arrays before using assert_array_equal?

stefraynaud:
A call to ``.getValue()`` before the comparison of the content should do the trick for both axes (coordinates) and variables.
[...]

PR #2354, opened 2018-08-08
Mark some tests related to cdat-lite as xfail

fujiisoup:
I just mark some to_cdms2 tests xfail. See #2332 for the details.
It is a temporal workaround and we may need to keep #2332 open until it is solved.


max-sixty:
I vote to merge ASAP

PR #2360, opened 2018-08-10
Add option to not roll coords

### LINKED ISSUE #1875: roll doesn't handle periodic boundary conditions well
DataArray.roll() currently rolls both data variables and coordinates:
```
>>> arr = xr.DataArray(range(4), [('x', range(0, 360, 90))])
>>> arr.roll(x=2)
<xarray.DataArray (x: 4)>
array([2, 3, 0, 1])
Coordinates:
  * x        (x) int64 180 270 0 90
```

This is sort of makes sense, but the labels are now all non-monotonic, so you can't even plot the data with xarray. In my experience, you probably want coordinate labels that either look like:

1. The unrolled original coordinates: [0, 90, 180, 270]
2. Shifted coordinates: [-180, -90, 0, 90]

It should be easier to accomplish this is in xarray. I currently resort to using roll and manually fixing up coordinates after the fact.

I'm actually not sure if there are any use-cases for the current behavior. Choice (1) would have the virtue of being consistent with shift():
```
>>> arr.shift(x=2)
<xarray.DataArray (x: 4)>
array([nan, nan,  0.,  1.])
Coordinates:
  * x        (x) int64 0 90 180 270
```
We could potentially add optional another argument for shifting labels, too, or requiring fixing that up by subtraction.

Note: you might argue that this is overly geoscience specific, and it would be, if this was only for handling a longitude coordinate. But periodic boundary conditions are common in many areas of the physical sciences.

ahuang11:
- [x] Closes #1875

image

Will add the others stuff from the checklist soon.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2052,10 +2052,8 @@ class DataArray(AbstractArray, DataWithCoords):
         Coordinates:
           * x        (x) int64 2 0 1
         """
-        shifts = either_dict_or_kwargs(shifts,
-                                       shifts_kwargs,
-                                       'roll')
-        ds = self._to_temp_dataset().roll(roll_coords=roll_coords, **shifts)
+        ds = self._to_temp_dataset().roll(
+            shifts=shifts, roll_coords=roll_coords, **shifts_kwargs)
         return self._from_temp_dataset(ds)
 
     @property
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2015,14 +2015,20 @@ class DataArray(AbstractArray, DataWithCoords):
         variable = self.variable.shift(**shifts)
         return self._replace(variable)
 
-    def roll(self, **shifts):
+    def roll(self, coords, **shifts):
         """Roll this array by an offset along one or more dimensions.
 
-        Unlike shift, roll rotates all variables, including coordinates. The
-        direction of rotation is consistent with :py:func:`numpy.roll`.
+        Unlike shift, roll may rotate all variables, including coordinates
+        if specified. The direction of rotation is consistent with
+        :py:func:`numpy.roll`.
 
         Parameters
         ----------
+        coords : bool
+            Indicates whether to  roll the coordinates by the offset
+            The current default of coords (None, equivalent to True) is
+            deprecated and will change to False in a future version.
+            Explicitly pass coords to silence the warning and sort.
         **shifts : keyword arguments of the form {dim: offset}
             Integer offset to rotate each of the given dimensions. Positive
             offsets roll to the right; negative offsets roll to the left.
@@ -2046,7 +2052,7 @@ class DataArray(AbstractArray, DataWithCoords):
         Coordinates:
           * x        (x) int64 2 0 1
         """
-        ds = self._to_temp_dataset().roll(**shifts)
+        ds = self._to_temp_dataset().roll(coords=coords, **shifts)
         return self._from_temp_dataset(ds)
 
     @property
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2015,7 +2015,7 @@ class DataArray(AbstractArray, DataWithCoords):
         variable = self.variable.shift(**shifts)
         return self._replace(variable)
 
-    def roll(self, coords, **shifts):
+    def roll(self, roll_coords=None, **shifts):
         """Roll this array by an offset along one or more dimensions.
 
         Unlike shift, roll may rotate all variables, including coordinates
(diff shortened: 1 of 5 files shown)


xarray/core/dataarray.py
```diff
@@ -2015,14 +2015,20 @@ def shift(self, **shifts):
         variable = self.variable.shift(**shifts)
         return self._replace(variable)
 
-    def roll(self, **shifts):
+    def roll(self, coords, **shifts):
```
fujiisoup: This argument should have a default.

xarray/core/dataset.py
```diff
@@ -3395,9 +3401,16 @@ def roll(self, **shifts):
 
         variables = OrderedDict()
         for name, var in iteritems(self.variables):
-            var_shifts = dict((k, v) for k, v in shifts.items()
-                              if k in var.dims)
-            variables[name] = var.roll(**var_shifts)
+            if name in self.data_vars or roll_coords:
```
fujiisoup: I think this logic should be updated.
All the variables in `data_vars` should be rolled, independent from `roll_coords` value.
Other variables (variables in `self.coord_names`) may depend on whether `roll_coords` is True or not.

xarray/core/dataset.py
```diff
@@ -3400,12 +3400,20 @@ def roll(self, roll_coords=None, **shifts):
...
+                                  if k in var.dims)
+                variables[name] = var.roll(**var_shifts)
+            elif name in coord_names and roll_coords:
+                var_shifts = dict((k, v) for k, v in shifts.items()
+                                  if k in var.dims)
+                variables[name] = var.roll(**var_shifts)
+            elif name in coord_names and roll_coords is None:
+                warnings.warn("roll_coords will be set to False in the future."
+                              " Explicitly set roll_coords to silence warning.",
+                               DeprecationWarning, stacklevel=3)
```
fujiisoup: My previous comment seemed confusing...
I think your previous commit is cleaner.

Or how about something like this?
```python
if roll_coords is None:
    warnings.warn("roll_coords will be set to False in the future."
                  " Explicitly set roll_coords to silence warning.",
                  DeprecationWarning, stacklevel=3)
    roll_coords = True

unrolled_vars = () if roll_coords else self.coord_names
variables = OrderedDict()
for k, v in self.variables:
    if k not in unrolled_vars:
        variables[k] = v.roll(**coords)
    else:
        variables[k] = v
```

ahuang11: Yes, that definitely looks cleaner.

xarray/core/dataset.py
```diff
@@ -3393,11 +3399,20 @@ def roll(self, **shifts):
         if invalid:
             raise ValueError("dimensions %r do not exist" % invalid)
 
+        if roll_coords is None:
+            warnings.warn("roll_coords will be set to False in the future."
+                          " Explicitly set roll_coords to silence warning.",
+                          DeprecationWarning, stacklevel=3)
```
shoyer: Use `FutureWarning` here, which doesn't get silenced by default and is intended for warning about future changes in behavior.

shoyer: Also -- consider using `stacklevel=2` here. `stacklevel=3` here would be confusing if `Dataset.roll()` is called directly.

I think it's *probably* better to incorrect indicate that the error is being raised internally inside xarray rather than to incorrectly indicate that it's being raised by some arbitrary user code.

xarray/tests/test_dataarray.py
```diff
@@ -3096,9 +3096,21 @@ def test_shift(self):
...
+
+    def test_roll_no_coords(self):
+        arr = DataArray([1, 2, 3], coords={'x': range(3)}, dims='x')
+        actual = arr.roll(x=1, roll_coords=False)
+        expected = DataArray([3, 1, 2], coords=[('x', [0, 1, 2])])
+        assert_identical(expected, actual)
+
+    def test_roll_coords_none(self):
+        arr = DataArray([1, 2, 3], coords={'x': range(3)}, dims='x')
+        actual = arr.roll(x=1, roll_coords=None)
```
shoyer: Please verify that the warning is raised, e.g.,
```
with pytest.raises(FutureWarning):
    ...
```

xarray/core/dataarray.py
```diff
@@ -2015,14 +2015,20 @@ def shift(self, **shifts):
...
+        if specified. The direction of rotation is consistent with
+        :py:func:`numpy.roll`.
 
         Parameters
         ----------
+        roll_coords : bool
+            Indicates whether to  roll the coordinates by the offset
+            The current default of roll_coords (None, equivalent to True) is
+            deprecated and will change to False in a future version.
+            Explicitly pass roll_coords to silence the warning and sort.
```
shoyer: What does "and sort" refer to?

xarray/core/dataarray.py
```diff
@@ -2015,14 +2015,20 @@ def shift(self, **shifts):
         variable = self.variable.shift(**shifts)
         return self._replace(variable)
 
-    def roll(self, **shifts):
+    def roll(self, roll_coords=None, **shifts):
```
shoyer: It might be nice to use `utils.either_dict_or_kwargs` here, to handle potential (unlikely) conflicts between the name "roll_coords" and dimension names.

Example usage:
https://github.com/pydata/xarray/blob/ded0a684136540962bcc409e6272b1cebb5af30a/xarray/core/dataarray.py#L755

ahuang11: Not too sure I understand this util; let me know if I implemented it inaccurately.

fujiisoup:
Thanks!

PR #2366, opened 2018-08-14
Future warning for default reduction dimension of groupby

### LINKED ISSUE #2363: Reduction APIs for groupby, groupby_bins, resample, rolling
From #2356 

APIs for `groupby`, `groupby_bins`, `resample`, `rolling` are different, especially for multi-dimensional array.

```python
import numpy as np
import xarray as xr
import pandas as pd

time = pd.date_range('2000-01-01', freq='6H', periods=365 * 4)
ds = xr.Dataset({'foo': (('time', 'x'), np.random.randn(365 * 4, 5)), 'time': time, 
                 'x': [0, 1, 2, 1, 0]})

ds.rolling(time=2).mean()  # result dims : ('time', 'x')
ds.resample(time='M').mean()  # result dims : ('time', 'x')
ds['foo'].resample(time='M').mean()  # result dims : ('time', )  maybe a bug #2362
ds.groupby('time.month').mean()  # result dims : ('month', )
ds.groupby_bins('time', 3).mean()  # result dims : ('time_bins', )
```

+ In `rolling` and `resample`(for Dataset), reduction without argument is carried out along grouped dimension
+ In `rolling`, reduction along other dimesnion is not possible
+ In `groupby` and `groupby_bins`, reduction is applied to the *grouped* objects and if without argument, it reduces alongall the dimensions of each grouped object.

I think `rolling`s API is most clean, but I am not sure it is worth to change these APIs.

The possible options would be 
1. Change APIs of `groupby` and `groupby_bins` so that they share similar API with `rolling`.
2. Document clearly how to perform `resample` or `groupby` with multidimensional arrays.

fujiisoup:
- [ ] Closes #xxxx
 - [x] Tests added
 - [x] Tests passed

Started to fix #2363.
Now warns a futurewarning in groupby if default reduction dimension is not specified.
As a side effect, I added `xarray.ALL_DIMS`. With `dim=ALL_DIMS` always reduces along all the dimensions.


fujiisoup:
I think it is ready for a review.

I am wondering when we can change the default behavior (it only affects groupby/resample for multi-dimensional arrays). 
v.0.11? or do we need more time?

xarray/core/common.py
```diff
@@ -9,7 +9,11 @@
 from . import duck_array_ops, dtypes, formatting, ops
 from .arithmetic import SupportsArithmetic
 from .pycompat import OrderedDict, basestring, dask_array_type, suppress
-from .utils import Frozen, SortedKeysDict
+from .utils import Frozen, SortedKeysDict, ReprObject
+
+
+# Use as a sentinel value to indicate a all dimensions, which
```
shoyer: This sentence looks incomplete

xarray/core/dataset.py
```diff
@@ -2750,6 +2751,8 @@ def reduce(self, func, dim=None, keep_attrs=False, numeric_only=False,
             Dataset with this object's DataArrays replaced with new DataArrays
             of summarized data and the indicated dimension(s) removed.
         """
+        if dim == ALL_DIMS:
```
shoyer: I usually prefer using `is` for comparison to sentinel objects -- it looks a little more idiomatic, and is more similar to what we do for comparison to `None`.

xarray/core/groupby.py
```diff
@@ -567,10 +568,41 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=False,
...
+            # TODO change this to dim = self._group_dim after
+            # the deprecation process
+            if self._obj.ndim > 1:
+                warnings.warn(
+                    "Default reduction dimension will be changed to the "
+                    "grouped dimension after xarray 0.11. To silence this "
+                    "warning, pass dim=xarray.ALL_DIMS explicitly.",
+                    FutureWarning, stacklevel=2)
+        elif dim is None:
+            dim = self._group_dim
```
shoyer: I don't think we should add this yet -- this means that code that uses `groupby.reduce(... dim=None)` would change its behavior. There probably isn't much code like this but I don't see a good reason to change it yet.

PR #2372, opened 2018-08-17
[MAINT] Avoid using duck typing

### LINKED ISSUE #2179: Stop using duck typing internally for xarray objects
We currently use duck typing internally in xarray in various somewhat ad-hoc ways to check for xarray objects, e.g.,
https://github.com/pydata/xarray/blob/48d55eea052fec204b843babdc81c258f3ed5ce1/xarray/core/combine.py#L130-L131

or in the `DataArray` constructor.

We should try to avoid doing this going forward (and change things in existing code), since these sorts of checks often result in hard to debug/understand behavior. Attribute names like `.values`, `.data` and `.name` are common enough that it's a bad idea to use them to recognize xarray objects.

fujiisoup:
- [x] Closes #2179 
 - [x] Tests passed

### DIFF
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -181,8 +181,6 @@ def _call_possibly_missing_method(arg, name, args, kwargs):
         method = getattr(arg, name)
     except AttributeError:
         duck_array_ops.fail_on_dask_array_input(arg, func_name=name)
-        if hasattr(arg, 'data'):
-            duck_array_ops.fail_on_dask_array_input(arg.data, func_name=name)
         raise
     else:
         return method(*args, **kwargs)
--- a/xarray/core/ops.py
+++ b/xarray/core/ops.py
@@ -187,6 +187,8 @@ def _call_possibly_missing_method(arg, name, args, kwargs):
         method = getattr(arg, name)
     except AttributeError:
         duck_array_ops.fail_on_dask_array_input(arg, func_name=name)
+        if hasattr(arg, 'data'):
+            duck_array_ops.fail_on_dask_array_input(arg.data, func_name=name)
         raise
     else:
         return method(*args, **kwargs)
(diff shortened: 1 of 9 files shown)


xarray/core/variable.py
```diff
@@ -64,22 +64,15 @@ def as_variable(obj, name=None):
...
-    # pandas objects.
-    if hasattr(obj, 'variable'):
+    if isinstance(obj, DataArray):
         # extract the primary Variable from DataArrays
         obj = obj.variable
 
     if isinstance(obj, Variable):
         obj = obj.copy(deep=False)
-    elif hasattr(obj, 'dims') and (hasattr(obj, 'data') or
-                                   hasattr(obj, 'values')):
```
fujiisoup: Do we need to consider Variable-like object here?

xarray/core/ops.py
```diff
@@ -187,8 +187,6 @@ def _call_possibly_missing_method(arg, name, args, kwargs):
         method = getattr(arg, name)
     except AttributeError:
         duck_array_ops.fail_on_dask_array_input(arg, func_name=name)
-        if hasattr(arg, 'data'):
-            duck_array_ops.fail_on_dask_array_input(arg.data, func_name=name)
```
fujiisoup: What is this block doing? Tests are passing without this.

shoyer: I'm sure we would get a bug report if it matters!

But for now it's probably best to leave it just in case, and save removing it for a refactor where we eliminate the `_call_possibly_missing_method` function entirely.

xarray/core/variable.py
shoyer: I don't think there's a good reason for keeping this clause. It was entirely undocumented.

fujiisoup:
Thanks @shoyer. Merged

PR #2373, opened 2018-08-17
More support of non-string dimension names

fujiisoup:

Following to #2174

In some methods, consistency of the dictionary arguments and keyword arguments are checked twice in `Dataset` and `Variable`. 
Can we change the API of Variable so that it does not take kwargs-type argument for dimension names?

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -290,7 +290,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
                 results[k] = v
         return results
 
-    def assign_coords(self, coords=None, **coords_kwargs):
+    def assign_coords(self, **kwargs):
         """Assign new coordinates to this object.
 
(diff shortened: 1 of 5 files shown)


xarray/core/common.py
```diff
@@ -526,24 +530,24 @@ def groupby_bins(self, group, bins, right=True, labels=None, precision=3,
...
-    def rolling(self, min_periods=None, center=False, **windows):
+    def rolling(self, dim=None, min_periods=None, center=False, **dim_kwargs):
         """
         Rolling window object.
 
         Parameters
         ----------
+        dim: dict, optional
+            Mapping from the dimension name to create the rolling iterator
+            along (e.g. `time`) to its moging window size.
```
max-sixty: small typo!

fujiisoup:
Thanks @max-sixty and @shoyer for your kind review.

PR #2375, opened 2018-08-21
Make `dim` optional on unstack

jsignell:

Not sure if this is a desirable change but I thought it could be easily discussed as a PR. Just for context I was looking at flattening spatial data for machine learning pipelines and then reshaping after the output had been acquired. 

I have a NxM array called `b` and am flattening it like:

```python
flat_input = b.stack(z=('y', 'x'))
```

Then I use that flat_input in my ML pipeline and get back an `np.array` (called `output`) which I want to unstack using all the metadata that went into stacking my original NxM array b.

```python
xr.full_like(flat_input, output).unstack(dim='z')
```

This PR just makes the `dim` argument optional in unstack so that we can use 

```python
xr.full_like(flat_input, output).unstack()
```

As a follow on PR I was thinking of making a function called `xr.unstack_like(other: xr.DataArray|xr.Dataset, data: array_like)` to encompass this functionality. A couple questions:

1) Would something like `xr.unstack_like` be desirable?
2) Should we be using `xr.full_like` in this way? The documentation in xarray and numpy only mentions scalars, but arrays work fine in both. If this is a supported behavior I could add docs, and if not perhaps a copy and overwrite is a better approach?

Here is a gist of the workflow using a tweaked datashader example and datashader example data https://gist.github.com/jsignell/79a6cf2da5c1458211d9dcf34d4417df

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1247,7 +1247,7 @@ class DataArray(AbstractArray, DataWithCoords):
         ds = self._to_temp_dataset().stack(dimensions, **dimensions_kwargs)
         return self._from_temp_dataset(ds)
 
-    def unstack(self, *dims):
+    def unstack(self, dim=None):
         """
         Unstack existing dimensions corresponding to MultiIndexes into
         multiple new dimensions.
@@ -1256,9 +1256,9 @@ class DataArray(AbstractArray, DataWithCoords):
 
         Parameters
         ----------
-        *dims : str, optional
-            Names of the existing dimensions to unstack. By default (if
-            no dims provided), unstacks all MultiIndexes.
+        dim : str or sequence of str, optional
+            Dimension(s) over which to unstack. By default unstacks all
+            MultiIndexes.
 
         Returns
         -------
@@ -1285,7 +1285,7 @@ class DataArray(AbstractArray, DataWithCoords):
         --------
         DataArray.stack
         """
-        ds = self._to_temp_dataset().unstack(*dims)
+        ds = self._to_temp_dataset().unstack(dim)
         return self._from_temp_dataset(ds)
 
     def transpose(self, *dims):
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1247,18 +1247,18 @@ class DataArray(AbstractArray, DataWithCoords):
         ds = self._to_temp_dataset().stack(dimensions, **dimensions_kwargs)
         return self._from_temp_dataset(ds)
 
-    def unstack(self, dim=None):
+    def unstack(self, *dims):
         """
-        Unstack an existing dimension corresponding to a MultiIndex into
+        Unstack existing dimensions corresponding to MultiIndexes into
         multiple new dimensions.
 
         New dimensions will be added at the end.
 
         Parameters
         ----------
-        dim : str
-            Name of the existing dimension to unstack.By default (if
-            ``dim is None``), unstacks first MultiIndex in dims.
+        *dims : str, optional
+            Names of the existing dimensions to unstack. By default (if
+            no dims provided), unstacks all MultiIndexes.
 
         Returns
         -------
@@ -1285,7 +1285,7 @@ class DataArray(AbstractArray, DataWithCoords):
         --------
         DataArray.stack
         """
-        ds = self._to_temp_dataset().unstack(dim)
+        ds = self._to_temp_dataset().unstack(*dims)
         return self._from_temp_dataset(ds)
 
     def transpose(self, *dims):
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1247,7 +1247,7 @@ class DataArray(AbstractArray, DataWithCoords):
         ds = self._to_temp_dataset().stack(dimensions, **dimensions_kwargs)
         return self._from_temp_dataset(ds)
 
-    def unstack(self, dim):
+    def unstack(self, dim=None):
         """
         Unstack an existing dimension corresponding to a MultiIndex into
         multiple new dimensions.
@@ -1257,7 +1257,8 @@ class DataArray(AbstractArray, DataWithCoords):
         Parameters
         ----------
         dim : str
-            Name of the existing dimension to unstack.
+            Name of the existing dimension to unstack.By default (if
+            ``dim is None``), unstacks first dim.
 
         Returns
         -------
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1258,12 +1258,28 @@ class DataArray(AbstractArray, DataWithCoords):
         ----------
         dim : str
             Name of the existing dimension to unstack.By default (if
-            ``dim is None``), unstacks first dim.
+            ``dim is None``), unstacks first MultiIndex in dims.
 
         Returns
         -------
         unstacked : DataArray
             Array with unstacked data.
+
+        Examples
+        --------
+
+        >>> arr = DataArray(np.arange(6).reshape(2, 3),
+        ...                 coords=[('x', ['a', 'b']), ('y', [0, 1, 2])])
+        >>> arr
+        <xarray.DataArray (x: 2, y: 3)>
+        array([[0, 1, 2],
+               [3, 4, 5]])
+        Coordinates:
+          * x        (x) |S1 'a' 'b'
+          * y        (y) int64 0 1 2
+        >>> roundtripped = arr.stack(z=('x', 'y')).unstack()
+        >>> arr.identical(roundtripped)
+        True
 
         See also
         --------
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1277,7 +1277,12 @@ class DataArray(AbstractArray, DataWithCoords):
         Coordinates:
           * x        (x) |S1 'a' 'b'
           * y        (y) int64 0 1 2
-        >>> roundtripped = arr.stack(z=('x', 'y')).unstack()
+        >>> stacked = arr.stack(z=('x', 'y'))
+        >>> stacked.indexes['z']
+        MultiIndex(levels=[[u'a', u'b'], [0, 1, 2]],
+                   labels=[[0, 0, 0, 1, 1, 1], [0, 1, 2, 0, 1, 2]],
+                   names=[u'x', u'y'])
+        >>> roundtripped = stacked.unstack()
         >>> arr.identical(roundtripped)
         True
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2361,33 +2361,23 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         --------
         Dataset.stack
         """
-        dim_from_kwarg = dim is not None
-
         if isinstance(dim, basestring):
-            dims = set([dim])
+            dims = [dim]
         elif dim is None:
-            dims = set(self.dims)
+            dims = self.dims
         else:
-            dims = set(dim)
+            dims = dim
 
-        missing_dims = [dim for dim in dims if dim not in self.dims]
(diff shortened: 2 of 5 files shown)


fujiisoup:
Thanks, @jsignell.

I like this idea (`unstack` without explicit dimension names), but I think we may need to decide what API would be the best.
My particular concern is 
+ what should be done if DataArray or Dataset has multiple MultiIndexes.
  Maybe do we unstack all the MultiIndexes?
+ we have similar method `reset_index`. Do we also want to make `dim` optional?

For `unstack_like`, I'm not sure it is worth adding as a top level function as `xr.full_like(other, data).unstack()` is simple enoguh...

jsignell:
> what should be done if DataArray or Dataset has multiple MultiIndexes.
Maybe do we unstack all the MultiIndexes?

I like the idea of unstacking all the MultiIndexes, but in that case should we allow passing in multiple dims? It seems weird to do a recursive unstack in the case of no argument passed without allowing the user to specifically choose multiple dims along which to unstack. 

I think it is probably better to just choose a default dim to unstack like this PR does. But maybe it is better to choose the last dim and check that it's a MultiIndex rather than the first dim. That way if you do a stack().unstack() you will roundtrip your data since the stacked index gets added to the end of dims. And if you just pass an object with one MultiIndex (probably the most common scenario) unstack will do the right thing.

jsignell:
>we have similar method reset_index. Do we also want to make dim optional?

I don't have an opinion on that except to say that `reset_index` takes an iter of dims so it is at least slightly different. So I guess I think that it seems find to only make dim optional on unstack.

shoyer:
I think `unstack()` unstacking all dimensions by default would make sense.

> Should we be using xr.full_like in this way?

I'm not really opposed to `full_like` working this way, but it does look a little strange to my eye. The "full" part of the name doesn't really make sense to me. I would usually suggest using the DataArray constructor here, e.g., `xr.DataArray(output_values, flat_input.coords, flat_input.dims, flat_inputs.attrs)`.

Maybe we can figure a better way to spell "label these arrays like this template xarray object" that doesn't require referencing `flat_input` multiple times. Maybe `xarray.label_like(array, source)` or `source.with_data(array)`?

> Would something like xr.unstack_like be desirable?

I'm not sure that a dedicated function `unstack_like` would make sense for xarray. This is the sort of helper function that you can write yourself in a couple of lines.

jsignell:
Ok so in this PR I will make unstack accept multiple dims like `xr.DataFrame.unstack(*dims)`. The order of the dims will only be roundtripped if all dims are stacked into one, but I think that is reasonable.

In a follow on PR I will make `xarray.label_like(array, other)`. I think that notation speaks more to what we are trying to convey, but I do think the position of the arguments isn't intuitive.

jsignell:
I chose to use *dims rather than a list of dims so that this change will have a very small impact on people. Most people probably do something like `unstack('z')` right now, and that will still work.

shoyer:
> I chose to use *dims rather than a list of dims so that this change will have a very small impact on people. Most people probably do something like unstack('z') right now, and that will still work.

Usually we prefer to stick to a single argument, but use isinstance checks to support both single dimensions and lists of dimensions, e.g., see how `dim` is parsed in `Dataset.reduce`:
https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L2774-L2779

jsignell:
I can change it. I guess I was looking at `Dataset.transpose`: https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L2498

shoyer:
Dataset.transpose accepts *args based on the design of
numpy.ndarray.transpose, but that API is probably a mistake (both in NumPy
and xarray). Everything else uses an axis/dim argument that can take a
scalar or sequence value.

jsignell:
Thanks for the context!

xarray/core/dataset.py
```diff
@@ -2366,6 +2339,58 @@ def unstack(self, dim):
...
+        non_multi_dims = [dim for dim in dims
+                          if not isinstance(self.get_index(dim), pd.MultiIndex)]
+        if non_multi_dims and dim_from_kwarg:
+            raise ValueError('cannot unstack dimensions that do not '
+                             'have a MultiIndex: %s' % non_multi_dims)
+
+        dims = dims - set(non_multi_dims)
+        if len(dims) == 0:
+            raise ValueError('cannot unstack an object that does not have '
+                             'MultiIndex dimensions')
```
fujiisoup: I think that we can allow to `unstack` an object without `MultiIndex`, which just returns as is.
It would be useful if users want to remove any MultiIndexes from an object.

xarray/core/dataset.py
```diff
@@ -2366,6 +2339,58 @@ def unstack(self, dim):
...
+        Dataset.stack
+        """
+        dim_from_kwarg = dim is not None
+
+        if isinstance(dim, basestring):
+            dims = set([dim])
+        elif dim is None:
+            dims = set(self.dims)
+        else:
+            dims = set(dim)
```
fujiisoup: Maybe we can use `OrderedSet` instead of `set` so that the resultant dimension order is fixed.

shoyer: I agree -- actually a list might be more appropriate here.

xarray/core/dataset.py
```diff
@@ -2366,6 +2339,58 @@ def unstack(self, dim):
...
+        See also
+        --------
+        Dataset.stack
+        """
+        dim_from_kwarg = dim is not None
+
+        if isinstance(dim, basestring):
+            dims = set([dim])
+        elif dim is None:
+            dims = set(self.dims)
```
shoyer: As a slightly tweaked version of this, rather than using `dim_from_kwarg`, why not set the default dimensions to the list of all dimensions containing a MultiIndex?

xarray/core/dataset.py
```diff
@@ -2366,6 +2339,58 @@ def unstack(self, dim):
...
+        else:
+            dims = set(dim)
+
+        missing_dims = [dim for dim in dims if dim not in self.dims]
```
shoyer: Minor point: use a new variable name other than `dim` here (even `d` would be fine) to avoid masking the original `dim` argument.

xarray/tests/test_dataarray.py
```diff
@@ -1663,6 +1663,16 @@ def test_stack_unstack(self):
...
+    def test_stack_unstack_with_no_dim_kwarg(self):
+        data = np.random.rand(1, 2, 3, 2, 1)
+        dims = ['a', 'b', 'c', 'd', 'e']
+        orig = xr.DataArray(data, dims=dims)
+        stacked = orig.stack(ab=['a', 'b'], cd=['c', 'd'])
+
+        unstacked = stacked.unstack()
+        roundtripped = unstacked.drop(['a', 'b', 'c', 'd']).transpose(*dims)
+        assert_identical(orig, roundtripped)
+
```
shoyer: Can we also add a test that explicitly calls `stacked.unstack(['ab', 'cd'])`?

xarray/core/dataarray.py
```diff
@@ -1247,23 +1247,40 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+        >>> arr = DataArray(np.arange(6).reshape(2, 3),
+        ...                 coords=[('x', ['a', 'b']), ('y', [0, 1, 2])])
+        >>> arr
+        <xarray.DataArray (x: 2, y: 3)>
+        array([[0, 1, 2],
+               [3, 4, 5]])
+        Coordinates:
+          * x        (x) |S1 'a' 'b'
+          * y        (y) int64 0 1 2
+        >>> roundtripped = arr.stack(z=('x', 'y')).unstack()
```
shoyer: I love this example! To make it a little clearer, could you use what `arr.stack(z=('x', 'y'))` looks like? That would make it more obvious what `unstack()` specifically does.

xarray/tests/test_dataarray.py
```diff
@@ -1660,9 +1660,22 @@ def test_dataset_math(self):
...
         actual = orig.stack(z=['x', 'y']).unstack('z').drop(['x', 'y'])
         assert_identical(orig, actual)
 
+        dims = ['a', 'b', 'c', 'd', 'e']
+        orig = xr.DataArray(np.random.rand(1, 2, 3, 2, 1), dims=dims)
+        stacked = orig.stack(ab=['a', 'b'], cd=['c', 'd'])
+
+        for dim in (None, ['ab', 'cd'], ['ab', 'cd', 'e']):
+            unstacked = stacked.unstack(dim=dim)
+            roundtripped = unstacked\
```
shoyer: another minor point: please use an extra set of parentheses rather than writing `\` for line continuations. This is preferred by PEP8 (and looks a little cleaner, in my opinion).

xarray/core/dataset.py
```diff
@@ -2366,6 +2339,48 @@ def unstack(self, dim):
...
+
+        missing_dims = [d for d in dims if d not in self.dims]
+        if missing_dims:
+            raise ValueError('Dataset does not contain the dimensions: %s'
+                             % missing_dims)
+
+        multi_dims = [d for d in dims if isinstance(self.get_index(d),
+                                                    pd.MultiIndex)]
+
+        result = self
```
shoyer: This is a minor point, but please make a shallow copy rather than ever returning the original object, e.g., `self.copy(deep=False)`. We try to do this consistently in xarray operations so users aren't surprised when they mutate objects.

Example of why this can be problematic:
```python
def unstack_with_history(xarray_obj):
    # this looks like a pure function, but it can actually modify the
    # input object if there's nothing to unstack
    unstacked = xarray_obj.unstack()
    unstacked.attrs['history'] = 'This was unstacked!'
    return unstacked
```

jsignell: Thanks for the explanation. That makes a lot of sense.

xarray/tests/test_dataarray.py
jsignell: I agree that is cleaner.

xarray/core/dataset.py
```diff
@@ -2366,6 +2339,48 @@ def unstack(self, dim):
...
+            dims = [dim]
+        elif dim is None:
+            dims = self.dims
+        else:
+            dims = dim
+
+        missing_dims = [d for d in dims if d not in self.dims]
+        if missing_dims:
+            raise ValueError('Dataset does not contain the dimensions: %s'
+                             % missing_dims)
```
shoyer: Looking at the implementation here, I see that it no longer raises an error when trying to unstack a dimension that does not use a MultiIndex. I'm not sure this is a desirable change -- on the one hand, it's nice to make unstacking an idempotent operation, but on the other hand, this might hide errors.

Instead, we could default `dims` to only include the entries from `self.dims` with a corresponding MultiIndex, and still raise an error when trying to unstack a MultiIndex.

jsignell: Right that was how I interpreted @fujiisoup's comment:

>I think that we can allow to unstack an object without MultiIndex, which just returns as is.
It would be useful if users want to remove any MultiIndexes from an object.

It seemed fine to me to have unstack just be a no-op if there are no MultiIndexes. But I can change it back if that would be better. What I was doing before was throwing an error if the user explicitly tries to unstack a non-MultiIndex, and otherwise just unstacking along all MultiIndexes unless there are none and throwing an error then.

jsignell: Ok so I made it so that if the user *explicitly tries to unstack a non-MultiIndex* they get an error. Otherwise it passes. So this is still valid `xr.Dataset().unstack()`.

shoyer: I was thinking something like:
```python
if dims is None:
    dims = [d for d in self.dims if isinstance(self.get_index(d), pd.MultiIndex)]
else:
    dims = [dim] if isinstance(dim, basestring) else dim
    # check for invalid dims

# call _unstack_once() repeatedly to unstack
```

jsignell:
Great! Thanks so much for all the feedback :)

PR #2384, opened 2018-08-27
Adding data kwarg to copy to create new objects with same structure  as original

jsignell:

This is a PR for the follow on work set out in https://github.com/pydata/xarray/pull/2375#issuecomment-415227870

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -951,16 +951,17 @@ def label_like(data, other):
     Returns
     -------
     out : same as other
-        New object with the same shape, and properties as other.
+        New object with the same shape and properties as other.
         Coords will be copied from other.
     """
     from .dataarray import DataArray
-    from .variable import Variable
+    from .variable import Variable, as_compatible_data
 
     if not (isinstance(other, DataArray) or isinstance(other, Variable)):
         raise TypeError("Expected DataArray, or Variable")
-    if not hasattr(data, "shape"):
-        raise TypeError("Data must have a .shape")
+
+    data = as_compatible_data(data)
+
     if other.shape != data.shape:
         raise ValueError("Data shape should match shape of object")
 
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -938,38 +938,48 @@ def ones_like(other, dtype=None):
     return full_like(other, 1, dtype)
 
 
-def structured_like(data, other):
-    """Return a new object labeled in the same way as the given object.
+def structured_like(data, template):
+    """Return a new object structured in the same way as the given object.
 
     Parameters
     ----------
     data : array_like
-        Data to fill the new object. Must have same shape as other.
-    other : DataArray, or Variable
-        The reference object from which to get labels
+        Data to fill the new object. Must have same shape as template.
+    template : DataArray, or Variable
+        The reference object from which to get labels.
 
     Returns
     -------
-    out : same as other
-        New object with the same shape and properties as other.
-        Coords will be copied from other.
+    out : same as template
+        New object with dimensions, attributes, coordinates and name
+        copied from template.
+
+    Examples
+    --------
+
+    >>> array = xr.DataArray([1, 2, 3], dims='x', coords={'x': ['a', 'b', 'c']})
+    >>> xr.structured_like([0.1, 0.2, 0.3], array)
+    <xarray.DataArray (x: 3)>
+    array([0.1, 0.2, 0.3])
+    Coordinates:
+    * x        (x) <U1 'a' 'b' 'c'
     """
     from .dataarray import DataArray
     from .variable import Variable, as_compatible_data
 
-    if not (isinstance(other, DataArray) or isinstance(other, Variable)):
+    if not (isinstance(template, DataArray) or isinstance(template, Variable)):
         raise TypeError("Expected DataArray, or Variable")
 
     data = as_compatible_data(data)
 
-    if other.shape != data.shape:
+    if template.shape != data.shape:
         raise ValueError("Data shape should match shape of object")
 
-    if isinstance(other, Variable):
-        return Variable(dims=other.dims, data=data, attrs=other.attrs)
+    if isinstance(template, Variable):
+        return Variable(dims=template.dims, data=data, attrs=template.attrs)
     return DataArray(
-        data, dims=other.dims, coords=other.coords, attrs=other.attrs,
-        name=other.name)
+        data, dims=template.dims, coords=template.coords, attrs=template.attrs,
+        name=template.name)
 
 
 def is_np_datetime_like(dtype):
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -938,50 +938,6 @@ def ones_like(other, dtype=None):
     return full_like(other, 1, dtype)
 
 
-def structured_like(data, template):
-    """Return a new object structured in the same way as the given object.
-
-    Parameters
-    ----------
-    data : array_like
-        Data to fill the new object. Must have same shape as template.
-    template : DataArray, or Variable
-        The reference object from which to get labels.
-
-    Returns
-    -------
-    out : same as template
-        New object with dimensions, attributes, coordinates and name
-        copied from template.
-
-    Examples
-    --------
-
-    >>> array = xr.DataArray([1, 2, 3], dims='x', coords={'x': ['a', 'b', 'c']})
-    >>> xr.structured_like([0.1, 0.2, 0.3], array)
-    <xarray.DataArray (x: 3)>
-    array([0.1, 0.2, 0.3])
-    Coordinates:
-    * x        (x) <U1 'a' 'b' 'c'
-    """
-    from .dataarray import DataArray
-    from .variable import Variable, as_compatible_data
-
-    if not isinstance(template, (DataArray, Variable)):
-        raise TypeError("Expected DataArray, or Variable")
-
-    data = as_compatible_data(data)
-
-    if template.shape != data.shape:
-        raise ValueError("Data shape should match shape of object")
-
-    if isinstance(template, Variable):
-        return Variable(dims=template.dims, data=data, attrs=template.attrs)
-    return DataArray(
-        data, dims=template.dims, coords=template.coords, attrs=template.attrs,
-        name=template.name)
-
-
 def is_np_datetime_like(dtype):
     """Check if a dtype is a subclass of the numpy datetime types
     """
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -967,7 +967,7 @@ def structured_like(data, template):
     from .dataarray import DataArray
     from .variable import Variable, as_compatible_data
 
-    if not (isinstance(template, DataArray) or isinstance(template, Variable)):
+    if not isinstance(template, (DataArray, Variable)):
         raise TypeError("Expected DataArray, or Variable")
 
     data = as_compatible_data(data)
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -938,6 +938,39 @@ def ones_like(other, dtype=None):
     return full_like(other, 1, dtype)
 
 
+def label_like(data, other):
+    """Return a new object labeled in the same way as the given object.
+
+    Parameters
+    ----------
+    data : array_like
+        Data to fill the new object. Must have same shape as other.
+    other : DataArray, or Variable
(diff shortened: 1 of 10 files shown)


xarray/core/common.py
```diff
@@ -938,6 +938,39 @@ def ones_like(other, dtype=None):
...
+        New object with the same shape, and properties as other.
+        Coords will be copied from other.
+    """
+    from .dataarray import DataArray
+    from .variable import Variable
+
+    if not (isinstance(other, DataArray) or isinstance(other, Variable)):
+        raise TypeError("Expected DataArray, or Variable")
+    if not hasattr(data, "shape"):
+        raise TypeError("Data must have a .shape")
```
shoyer: Use `data = xarray.core.variable.as_compatible_data(data)` here instead of checking for the `shape` attribute for a consistent wrapper.

xarray/core/common.py
```diff
@@ -938,6 +938,39 @@ def ones_like(other, dtype=None):
...
+    -------
+    out : same as other
+        New object with the same shape, and properties as other.
+        Coords will be copied from other.
+    """
+    from .dataarray import DataArray
+    from .variable import Variable
+
+    if not (isinstance(other, DataArray) or isinstance(other, Variable)):
+        raise TypeError("Expected DataArray, or Variable")
```
shoyer: I think it could make logical sense to support coercing a dict of numpy arrays into an xarray.Dataset (as the `data_vars` argument). But I'm also happy to leave that for a later improvement.

xarray/tests/test_dataarray.py
```diff
@@ -3153,6 +3153,22 @@ def test_full_like(self):
...
+                         coords={'x': [4, 3]},
+                         name='helloworld')
+        new_data = np.arange(4).reshape(2, 2)
+        new_var = label_like(new_data, orig)
+        for i in ['dims', 'attrs', 'name']:
+            assert getattr(new_var, i) == getattr(orig, i)
+        for coord in new_var.coords:
+            assert coord in orig.coords
+        with raises_regex(AssertionError, 'Arrays are not equal'):
+            assert_array_equal(new_var, orig)
```
shoyer: This could be a good use case for `xarray.testing.assert_identical()`. You would just need to prepare an object manually to compare identical to (e.g., by copying `orig` and replacing its `data`).

xarray/tests/test_dataarray.py
```diff
@@ -3153,6 +3153,22 @@ def test_full_like(self):
...
+        # For more thorough tests, see test_variable.py
+        orig = DataArray(np.random.random(size=(2, 2)),
+                         dims=('x', 'y'),
+                         attrs={'attr1': 'value1'},
+                         coords={'x': [4, 3]},
+                         name='helloworld')
+        new_data = np.arange(4).reshape(2, 2)
+        new_var = label_like(new_data, orig)
+        for i in ['dims', 'attrs', 'name']:
+            assert getattr(new_var, i) == getattr(orig, i)
```
shoyer: For future reference: for testing, we usually prefer to avoid loops when possible, even if it results in a bit of extra code. It's much easier to debug when a test failure points to the line `assert new_var.dims == orig.dims`.

xarray/core/common.py
jsignell: Yeah I wasn't sure of the best approach and I couldn't think of a time when I'd use it (I was mostly working with DataArrays) so I figured it would be ok to throw an error for now. But I am happy to add it if that is more normal.

xarray/tests/test_dataarray.py
jsignell: yeah that is a lot better.

jsignell:
Any thoughts on this one?

jsignell:
Or what about `xr.DataArray.like` and `xr.Variable.like`? My main worry is that it kind of feels like these should return `bool`s. Another option would be `xr.DataArray.from_dataarray` and `xr.Variable.from_variable`, but that is maybe also confusing...

shoyer:
What about a DataArray method called `replace_data` or `with_data`? e.g., `data_array.replace_data(new_data)` or `data_array.with_data(new_data)`.

We could also make a more general `DataArray.replace` method (inspired by `collections.namedtuple._replace`) that lets you replace any of `data`, `coords`, `dims`, `attrs` or `encoding` on a top object. Then this would look like `data_array.replace(new_data)` or `data_array.replace(attrs=new_attrs)`. Replacing `name` or `dims` could invoke the `rename()` method.

jsignell:
I don't like it being a method on an existing object. I think the syntax is clearer if the template object is an argument and the second argument is the data. I think that conveys the concept of creating a new object that is just using the structure of another object.

jsignell:
That said. I think `with_data` is better than anything replace. I would worry about confusion with the pandas replace which replaces specific values.

shoyer:
More ideas:
- `xarray.copy_structure(template, new_data)`
- `xarray.structured_like(new_data, template)`
- `xarray.copy_labels(template, new_data)`

jsignell:
👍 for  `xarray.structured_like(new_data, template)`

jsignell:
I think this is good to merge

xarray/core/common.py
```diff
@@ -938,6 +938,40 @@ def ones_like(other, dtype=None):
...
 
 
+def structured_like(data, other):
+    """Return a new object labeled in the same way as the given object.
+
+    Parameters
+    ----------
+    data : array_like
+        Data to fill the new object. Must have same shape as other.
+    other : DataArray, or Variable
```
shoyer: Maybe call this parameter `template` instead of `other`?

xarray/core/common.py
```diff
@@ -938,6 +938,40 @@ def ones_like(other, dtype=None):
...
+        The reference object from which to get labels
+
+    Returns
+    -------
+    out : same as other
+        New object with the same shape and properties as other.
+        Coords will be copied from other.
```
shoyer: Maybe: "New object with dimensions, attributes, coordinates and name copied from template."

jsignell:
I made the docs changes and added an example.

xarray/core/common.py
```diff
@@ -938,6 +938,50 @@ def ones_like(other, dtype=None):
...
+    >>> xr.structured_like([0.1, 0.2, 0.3], array)
+    <xarray.DataArray (x: 3)>
+    array([0.1, 0.2, 0.3])
+    Coordinates:
+    * x        (x) <U1 'a' 'b' 'c'
+    """
+    from .dataarray import DataArray
+    from .variable import Variable, as_compatible_data
+
+    if not (isinstance(template, DataArray) or isinstance(template, Variable)):
```
shoyer: Optional: `isinstance` can check against tuples of types all at the same time, so you could write: `if not isinstance(template, (Variable, DataArray)):

jsignell: nice!

shoyer:
`template.copy(data=new_data)`? That seems pretty clear, actually.

jsignell:
I added some examples on copy for each of the object types and I decided to let `Dataset`s do this too. I just require that the user pass a dict where keys are contained in data_vars and each value has the same shape as the corresponding data_var.

xarray/core/dataset.py
```diff
@@ -709,20 +709,104 @@ def _replace_indexes(self, indexes):
...
+                                          self._dims.copy(), self._attrs_copy(),
+                                          encoding=self.encoding)
+        if not utils.is_dict_like(data):
+            raise ValueError('Data must be dict-like')
+        elif not set(data) <= set(self._variables.keys()):
+            raise ValueError('All keys in `data` must be'
+                             'contained in the original dataset')
+
+        variables = OrderedDict((k, self._variables[k].copy(data=v))
+                                for k, v in iteritems(data))
```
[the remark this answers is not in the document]
jsignell: > What happens if you supply some but not all variables in ``data``? 

I was operating under the assumption that you never get old data if you set data. So you get a dataset containing only the vars that you included in ``data``

I can document that, or I can make it mandatory to include all the data_vars in a dataset. I don't think mixing old and new data is an expected behavior.

xarray/core/dataarray.py
```diff
@@ -677,14 +677,76 @@ def persist(self, **kwargs):
...
+        original but entirely new data. When `data` is used, `deep` is ignored.
+
+        Parameters
+        ----------
+        deep : bool, optional
+            Whether the data array is loaded into memory and copied onto
+            the new object. Default is True.
+        data : array_like, optional
+            Data to use in the new object. Must have same shape as original.
+            If `data` is set, deep is ignored.
```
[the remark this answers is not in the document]
jsignell: The coordinates are `IndexVariable`s  which are immutable so they ignore deep in their own method. Maybe I should make that more explicit?

xarray/core/dataset.py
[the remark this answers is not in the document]
shoyer: Probably requiring that the keys of `data` match `data_vars` leaves the least room for surprises

xarray/core/dataarray.py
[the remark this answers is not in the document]
shoyer: You can also have non `IndexVariable` objects in coords, if they don't correspond to dimensions, e.g., `latitude` and `longitude` in [this example](http://xarray.pydata.org/en/stable/data-structures.html#dataset) . Those variables aren't immutable.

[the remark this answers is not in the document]
jsignell: Oh! Yeah that makes sense. Hmm ok so pass ``deep`` on to coords and just clean up the docs?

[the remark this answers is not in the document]
shoyer: Yep, I think we already pass `deep` on to coordinates.

xarray/core/dataset.py
```diff
@@ -743,47 +743,59 @@ def copy(self, deep=False, data=None):
         Shallow copy versus deep copy
 
         >>> da = xr.DataArray(np.random.randn(2, 3))
```
shoyer: Tip for next time: to make a reproducible docstring example, use `np.random.RandomState(0).randn(2, 2)` instead. It's a little longer, but RandomState objects are almost always a better design choice than the global functions.

xarray/core/dataset.py
```diff
@@ -793,20 +805,28 @@ def copy(self, deep=False, data=None):
...
+        else:
+            var_keys = set(self.data_vars.keys())
+            data_keys = set(data.keys())
+            keys_not_in_vars = data_keys - var_keys
+            if keys_not_in_vars:
+                raise ValueError('Data must only contain variables in original '
+                                 'dataset. Extra variables: {}'
+                                 .format(keys_not_in_vars))
+            keys_missing_from_data = var_keys - data_keys
+            if keys_missing_from_data:
```
shoyer: you could potentially consolidate these checks into `if `data_keys != var_keys`, but this is probably a more informative error message :)

xarray/core/variable.py
```diff
@@ -1765,12 +1766,12 @@ def concat(cls, variables, dim='concat_dim', positions=None,
...
         return cls(first_var.dims, data, attrs)
 
-    def copy(self, deep=True):
+    def copy(self, deep=True, data=None):
         """Returns a copy of this object.
 
-        `deep` is ignored since data is stored in the form of pandas.Index,
-        which is already immutable. Dimensions, attributes and encodings are
-        always copied.
+        `deep` and `data` are ignored since data is stored in the form of
```
shoyer: I'm not sure it makes sense to skip the `data` argument for an IndexVariable.

`deep=True` is safe to ignore, because there's no need to worry about deeply copying data for immutable objects (the pandas.Index).

But that doesn't exclude making a new object with new data.

xarray/core/dataset.py
jsignell: yep I wrote them that way on purpose for the error message.

xarray/core/variable.py
jsignell: Yeah that is definitely right. I wasn't thinking about it properly. I made the change.

xarray/core/dataset.py
jsignell: ok sounds good. I tend to use `np.arange(size).reshape(shape)`  but I decided it was too long for this one.

shoyer:
If we're all happy with copy (which I think we are), then this should be good to go!

max-sixty:
Thank you v much @jsignell ! Both for the excellent code and for bearing with the changes!

PR #2386, opened 2018-08-28
fix typo in uri in the docs

DocOtak:
Seems I left off a trailing `>` in the `whats-new.rst`. Here is a minor fix for that.


shoyer:
thanks!

PR #2387, opened 2018-08-28
DOC: move xskillscore to 'Extend xarray capabilities'

raybellwaves:
Think my package should go in the 'extend xarray capabilities' section rather than the 'geoscience' section.


jhamman:
Cool. Thanks @raybellwaves!

PR #2391, opened 2018-08-30
WIP: wrap stored arrays in dask.delayed

shoyer:


jhamman:
@shoyer - this seems to have gone stale. Is this still worth pursuing?

shoyer:
I think we should drop this, unless we can find an example where it actually make a performance difference after the backends refactor.

PR #2397, opened 2018-09-03
add options for nondivergent and divergent cmap

### LINKED ISSUE #2394: Change default colormaps
#### Problem description

xarrays plotting module is awesome because it detects automatically if the data is divergent or not and adjusts the colormap accordingly.

But what if I do not particularly like the default colormap choice for either data type (e.g. my current boss is not a fan of matplotlibs [viridis](https://matplotlib.org/examples/color/colormaps_reference.html))?

@shoyer mentioned on [stackexchange](https://stackoverflow.com/questions/52114364/change-the-default-colorbar-in-xarray-for-non-diverging-data?noredirect=1#comment91188566_52114364) that this is not currently possible. 

I would be happy to submit a PR, if I can get some guidance on where to implement this.

What does everyone think about a syntax this shoule be implemented with?

Something like:

```
with xr.set_options(diverging_cmap='RdYlBl'):
    ...
with xr.set_options(nondiverging_cmap='magma'):
    ....
```
Or is there a better way to do this?
I would personally prefer to be able to set this at the beginning of a notebook, so that it is applied to all following plots. Maybe something like the `mpl.rcparams` is a way to go here?

jbusecke:
- [ ] Closes #2394 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -479,7 +479,7 @@ class TestDetermineCmapParams(TestCase):
     def test_cmap_sequential_explicit_option(self):
         with xr.set_options(cmap_sequential=mpl.cm.magma):
             cmap_params = _determine_cmap_params(self.data)
-            assert cmap_params['cmap'] == 'magma'
+            assert cmap_params['cmap'] == mpl.cm.magma
 
     def test_cmap_divergent_option(self):
         with xr.set_options(cmap_divergent='magma'):
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -472,7 +472,7 @@ class TestDetermineCmapParams(TestCase):
         assert cmap_params['norm'] is None
 
     def test_cmap_option(self):
-        with xr.set_options(cmap='magma'):
+        with xr.set_options(cmap_sequential='magma'):
             cmap_params = _determine_cmap_params(self.data)
             assert cmap_params['cmap'] == 'magma'
 
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -5,6 +5,7 @@ from datetime import datetime
 
 import numpy as np
 import pandas as pd
+import xarray as xr
 import pytest
 
 import xarray.plot as xplt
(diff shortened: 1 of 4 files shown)


dcherian:
Should we call it `cmap_sequential` instead of just `cmap`?

xarray/tests/test_plot.py
```diff
@@ -470,6 +471,16 @@ def test_center(self):
         assert cmap_params['levels'] is None
         assert cmap_params['norm'] is None
 
+    def test_cmap_option(self):
```
dcherian: `test_cmap_sequential_option`.

Nice test btw.

xarray/tests/test_plot.py
```diff
@@ -470,6 +471,16 @@ def test_center(self):
...
+        with xr.set_options(cmap_sequential='magma'):
+            cmap_params = _determine_cmap_params(self.data)
+            assert cmap_params['cmap'] == 'magma'
+
+    def test_cmap_divergent_option(self):
+        with xr.set_options(cmap_divergent='magma'):
```
dcherian: can you add a test where you set it to `mpl.cm.magma`. It should all just work but would be nice to explicitly test for that option.

xarray/tests/test_plot.py
jbusecke: Thanks!

xarray/tests/test_plot.py
jbusecke: Just added an explicit test.

jbusecke:
Awesome. Thanks a lot for the feedback @dcherian.

jbusecke:
Yikes. Also sorry for this merge fail commit. I am sitting in a conference :-).

xarray/tests/test_plot.py
```diff
@@ -470,6 +471,21 @@ def test_center(self):
...
 
+    def test_cmap_sequential_option(self):
+        with xr.set_options(cmap_sequential='magma'):
+            cmap_params = _determine_cmap_params(self.data)
+            assert cmap_params['cmap'] == 'magma'
+
+    def test_cmap_sequential_explicit_option(self):
+        with xr.set_options(cmap_sequential=mpl.cm.magma):
+            cmap_params = _determine_cmap_params(self.data)
+            assert cmap_params['cmap'] == 'magma'
```
dcherian: `== mpl.cm.magma`?

jbusecke:
I am unsure if that failure is due to a time out or changes in the PR. Is ther anything else that I should change before merge? Again, many thanks for the help in getting this forward.

shoyer:
I restarted the timed out test, but it does look non-related. Merging -- thanks!

Issue #1288, opened 2017-02-28
Add trapz to DataArray for mathematical integration


jeffbparker:
Since scientific data is often an approximation to a continuous function, when we write mean() or sum(), our underlying intention is often to approximate an integral.  For example, if we have temperature of a rod T(t, x) as a function of time and space, the average value Tavg(x) is the integral of T(t,x) with respect to x, divided by the length.  

I would guess that in practice, many uses of `mean()` and `sum()` are intending to approximate integrals of continuous functions.  That is typically my use, at least.  But simply adding up all values is a Riemann sum approximation to an integral which is not very accurate.  

For approximating an integral, it seems to me that the trapezoidal rule (`trapz()` in numpy) should be preferred to `sum()` or `mean()` in essentially all cases, as the trapezoidal rule is more accurate while still being efficient.

It would be very useful to have `trapz()` as a method of DataArrays, so I could write, e.g., for an average value, `Tavg = T.trapz(dim='time') / totalTime`.  Currently, I would have to use numpy's method and then rebuild the reduced-dimensional array myself:

```
TavgVal= np.trapz(y, T['time'], axis=0) / totalTime
Tavg= xr.DataArray(TavgVal, coords=T['space'], dims='space')
```

It could even be useful to have a function like `mean_trapz()` that calculates the mean value based on trapz.  More generally, one could imagine having other integration methods too.  E.g., `data.integrate(dim='x', method='simpson')`.  But `trapz` is probably good enough for many cases and a big improvement over `mean`, and `trapz` is very simple even for unequally spaced data.  And `trapz` shouldn't be much less efficient in principle, although in practice I find `np.trapz()` to be several times slower than `np.mean()`.

Quick examples demonstrating `sum()`/`mean()` vs. `trapz()`

```
x = np.linspace(0, 2, 200)
y = 1/3 * x**3
dx = x[1] - x[0]
integralRiemann =  dx * np.sum(y)  # 1.3467673375251465
integralTrapz = np.trapz(y, x)  # 1.3333670025167712
integralExact = 4/3  # 1.3333333333333333
```

This second example demonstrates the special advantages of trapz() for periodic functions because the trapezoidal rule happens to be extremely accurate for periodic functions integrated over their period.

```
x = np.linspace(0, 2*np.pi, 200)
y = cos(x)**2
meanRiemann = np.mean(y)  #  0.50249999999999995
meanTrapz = np.trapz(y, x) / (2*np.pi)  # 0.5
[...]

PR #2398, opened 2018-09-04
implement Gradient

### LINKED ISSUE #1332: Shape preserving `diff` via new keywords
Currently, an operation such as `ds.diff('x')` will result in a smaller size dimension, e.g.,

```python
In [1]: import xarray as xr

In [2]: ds = xr.Dataset({'foo': (('x',), [1, 2, 3])}, {'x': [1, 2, 3]})

In [3]: ds
Out[3]: 
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 1 2 3
Data variables:
    foo      (x) int64 1 2 3

In [4]: ds.diff('x')
Out[4]: 
<xarray.Dataset>
Dimensions:  (x: 2)
Coordinates:
  * x        (x) int64 2 3
Data variables:
    foo      (x) int64 1 1
```

However, there are cases where the same size would be beneficial to keep so that you would get
```python
In [1]: import xarray as xr

In [2]: ds = xr.Dataset({'foo': (('x',), [1, 2, 3])}, {'x': [1, 2, 3]})

In [3]: ds.diff('x', preserve_shape=True, empty_value=0)
Out[3]: 
<xarray.Dataset>
Dimensions:  (x: 3)
Coordinates:
  * x        (x) int64 1 2 3
Data variables:
    foo      (x) int64 0 1 1
```

Is there interest in addition of a `preserve_shape=True` keyword such that it results in this shape-preserving behavior?  I'm proposing you could use this with `label='upper'` and `label='lower'`. 

`empty_value` could be a value or `empty_index` could be an index for the fill value.  If `empty_value=None` and `empty_index=None`, it would produce a `nan`.  

The reason I'm asking the community is because this is at least the second time I've encountered an application where this behavior would be helpful, e.g., computing ocean layer thicknesses from bottom depths.  A previous application was computation of a time step from time slice output and the desire to use this product in an approximated integral, e.g., 
```python
y*diff(t, label='lower', preserve_shape=True)
```
 where `y` and `t` are both of size `n`, which is effectively a left-sided Riemann sum.

fujiisoup:
- [x] Closes #1332
 - [x] Tests added
 - [x] Tests passed

Added `xr.gradient`, `xr.DataArray.gradient`, and `xr.Dataset.gradient` according to #1332.


xarray/core/dask_array_ops.py
```diff
@@ -104,3 +104,49 @@ def func(x, window, axis=-1):
     index = (slice(None),) * axis + (slice(drop_size,
                                            drop_size + orig_shape[axis]), )
     return out[index]
+
+
+def gradient(a, coord, axis, edge_order):
```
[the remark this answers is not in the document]
jakirkham: Do you mean like [this]( http://dask.pydata.org/en/latest/array-api.html#dask.array.gradient )? Requires Dask 0.17.3+ though. So IDK if that works for you or not.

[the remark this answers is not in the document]
fujiisoup: Thanks @jakirkahm.
I need to compute the gradient with a non uniform coordinate, but dask Array.gradient only supports uniformly spacing coordinate.

I think this can be implemented using overlap as I do here. Is it in the scope of the dask development?

xarray/core/computation.py
```diff
@@ -1137,3 +1137,109 @@ def where(cond, x, y):
...
+    <xarray.DataArray (x: 4, y: 3)>
+    array([[30.      , 30.      , 30.      ],
+           [27.545455, 27.545455, 27.545455],
+           [27.545455, 27.545455, 27.545455],
+           [30.      , 30.      , 30.      ]])
+    Coordinates:
+      * x        (x) float64 0.0 0.1 1.1 1.2
+    Dimensions without coordinates: y
+    >>>
+    >>> xr.gradient(da, ('x', 'y'))
```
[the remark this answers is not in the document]
fujiisoup: It would be another option, but in order to store them to a dataset, we will need to define their names very heuristically...

Maybe we can drop this api as this does not speed up the computation and it's easy to do the same thing manually.

xarray/core/dask_array_ops.py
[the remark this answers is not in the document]
jakirkham: Ah sorry. Should have read more of the code.

Yeah I think that is in scope. Just [not currently supported]( https://github.com/dask/dask/blob/5fa703eaaa6679e452c1dc7cdc8c9ac039994de5/dask/array/routines.py#L437-L438 ) yet.

Could you please open an issue or perhaps PR over on the Dask repo?

xarray/core/computation.py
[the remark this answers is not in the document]
shoyer: I agree. I don't think it's particularly useful to compute to compute independent gradients with respect to multiple axes at the same time, e.g., `du/dx`, `du/dy`, etc. If anything, I would be more excited about computing higher order derivatives, e.g., `\frac{d^2 u}{dx dy}`

shoyer:
I think this will be very welcome functionality!

I wonder if we should consider calling this `xarray.differentiate` instead of `xarray.gradient`. I think the NumPy function is poorly named for differentiating along a single axis at once.

fujiisoup:
> I wonder if we should consider calling this xarray.differentiate instead of xarray.gradient. I think the NumPy function is poorly named for differentiating along a single axis at once.

Agreed. Differentiate is nicer.

Some other api questions arised during the implementation
+ Do we support `differentiate` for `Dataset`? In that case, what should we do for the variables that are independent from the target coordinate?
I thought 'keep them as is' is intuitive (and I implemented so), but mathematically, they should be zero.

+ Do we need to sort the array before computing `differentiate`? `np.gradient` implicitly assumes the array is sorted (but do nothing about this).

fujiisoup:
any thoughts for this?

fujiisoup:
Thanks, @dopplershift .

> Aren't you taking differences of values and dividing by differences between the matching coordinates?

Yes, correct. But if we have closer data points, then the estimate of the gradient becomes more precise. Sorting the array according to the coordinate provides the closest points, resulting in the most precise estimate of the gradient.

But I also think users can do it manually before taking the gradient.

xarray/core/computation.py
```diff
@@ -1137,3 +1137,81 @@ def where(cond, x, y):
...
+           [30.      , 30.      , 30.      ]])
+    Coordinates:
+      * x        (x) float64 0.0 0.1 1.1 1.2
+    Dimensions without coordinates: y
+    """
+    from .dataarray import DataArray
+
+    if not isinstance(dataarray, DataArray):
+        raise TypeError(
+            'Only DataArray is supported. Given {}.'.format(type(dataarray)))
```
shoyer: I'm a little confused here. You wrote an implementation for Dataset.differentiate, too, and here is a duplicate version of DataArray.differentiate.

shoyer: 👍 I don't think we need the separate function.

xarray/core/dataarray.py
```diff
@@ -2289,6 +2292,53 @@ def rank(self, dim, pct=False, keep_attrs=False):
         ds = self._to_temp_dataset().rank(dim, pct=pct, keep_attrs=keep_attrs)
         return self._from_temp_dataset(ds)
 
+    def differentiate(self, coord, edge_order=1):
```
dcherian: Shouldn't this be `dim` instead of `coord`? I'm thinking of multidimensional co-ordinates here.

xarray/tests/test_dataset.py
```diff
@@ -4492,3 +4492,38 @@ def test_raise_no_warning_for_nan_in_binary_ops():
     with pytest.warns(None) as record:
         Dataset(data_vars={'x': ('y', [1, 2, np.NaN])}) > 0
     assert len(record) == 0
+
+
+@pytest.mark.parametrize('dask', [True, False])
+@pytest.mark.parametrize('edge_order', [1, 2])
+def test_gradient(dask, edge_order):
```
dcherian: can you add a test for a `datetime` co-ordinate also? The `gradient` in `npcompat` seems to support it.

xarray/core/dataarray.py
fujiisoup: Currently, a 1-dimensional (non-dimensional) coordinate is also allowed. Do you think it is confusing?
For a multidimensional coordinate, ValueError will be raised.

xarray/tests/test_dataset.py
fujiisoup: It looks np.gradient does not support `datetime`. Maybe we need a unit that should be used for compute the gradient along `datetime`, such as `ns` or `date`. 
Do you have any idea about API for this?

xarray/core/dataarray.py
dcherian: OK¸ maybe add a little more to the docstring then to make it clear that we expect 1D co-ordinates.

xarray/tests/test_dataset.py
dcherian: Oh you're right; I misunderstood the code.

How about

```
da = xr.DataArray([1., 2.], dims='time', coords={'time': pd.date_range('2001-01-01', '2001-01-02')})

da.differentiate(coord='time', astype='D')
```

which would be equivalent to
```
np.diff(da.values) / (np.diff(da.time).astype('timedelta64[D]').astype('float64'))
```

`astype` isn't the best kwarg name though. `freq` or `resolution` don't seem correct either.

xarray/core/dataarray.py
```diff
@@ -2289,6 +2292,54 @@ def rank(self, dim, pct=False, keep_attrs=False):
...
+        """ Differentiate the array with the second order accurate central
+        differences.
+
+        Parameters
+        ----------
+        coord: str
+            The coordinate to be used to compute the gradient.
+        edge_order: 1 or 2. Default 1
+            N-th order accurate differences at the boundaries.
+
```
shoyer: needs `time_unit` here

xarray/core/dataset.py
```diff
@@ -3663,6 +3666,60 @@ def rank(self, dim, pct=False, keep_attrs=False):
...
+        coord_var = self[coord].variable
+        if coord_var.ndim != 1:
+            raise ValueError('Coordinate {} must be 1 dimensional but is {}'
+                             ' dimensional'.format(coord, coord_var.ndim))
+
+        dim = coord_var.dims[0]
+        coord_data = coord_var.data
+        if coord_data.dtype.kind in ['m', 'M']:
+            if time_unit is None:
+                time_unit = np.datetime_data(coord_data.dtype)[0]
```
shoyer: Use tuple unpacking here, e.g., `time_unit, _ = np.datetime_data(coord_data.dtype)`

xarray/tests/test_dataset.py
shoyer: In the long term, hopefully we'll solve this by getting real unit support working, but a `time_unit` argument seems like a good choice for now

fujiisoup: Numpy seems not to define the terminology for this unit, but *datetime_unit* is being used [here](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.datetime_data.html). I wonder if we should use it here.
Is it a little long?

fujiisoup:
I think it's ready :)

fmaussion:
This is so great! This is going to simplify my classes about derivatives on the globe even more: my only concern is that students will soon forget what a numerical derivative actually is, and that it's not a trivial to implement ;-)

shoyer:
Speaking of derivatives on the globe, we might want to include an option for periodic boundary conditions (and possibly for other functions like `interp` as well). But we can save that for another PR :).

fmaussion:
> But we can save that for another PR :)

See also https://github.com/pydata/xarray/issues/1288 : ``integrate`` is the next on my list ;) - I can try to give it a go if @fujiisoup doesn't want to do it himself

jakirkham:
@shoyer, have you seen `da.pad`?

shoyer:
> @shoyer, have you seen da.pad?

Yes, `pad` solves the hard part of periodic boundary conditions -- we should just need to do a bit of book-keeping on the xarray side.

rabernat:
I agree that the features implemented here are broadly useful and belong in xarray. The fundamental question is: how far does xarray itself want to go in supporting vector calculus in curvilinear coordinates (i.e. on the sphere). 

There is a fair bit of overlap between this new functionality and some of the things that we are trying to do in xgcm: https://xgcm.readthedocs.io/en/latest/grids.html. Xgcm supports periodic boundary conditions, as well as more complex topological connections between array edges. It allows users to reproduce precisely the sort of operations used to take gradients in finite-volume staggered-grid models.

shoyer:
@rabernat my inclination would be to draw the line at regular grids (with or without periodic boundary conditions), which are pretty commonly encountered in data analysis for any continuous physical systems. We would leave non-regular cases like staggered-grids to add-ons like xgcm -- these grids tend to be more application/PDE specific.

shoyer:
We should definitely mention specialized tools like xgcm as alternatives in the docstring for these xarray methods.

fujiisoup:
Thanks all. Updated.

Thanks, @spencerkclark 

> Eventually it would be nice if this worked on DataArrays with cftime.datetime coordinates; I think it would be relatively straightforward to modify to_numeric to enable it (we could probably enable it for interp at the same time), but I can take care of that later if you'd like.

Thanks. I added this function for something like this extension, though I do not yet fully follow your cftime update. It would be super nice if you could take care of this after merge.

@shoyer ,

> we might want to include an option for periodic boundary conditions 

Agreed. This option is nice not only `differentiate` but also `interp` and [`rolling`](#2011).
I think we can add a common logic to take care of them.

@fmaussion 

> See also #1288 : integrate is the next on my list ;) - I can try to give it a go if @fujiisoup doesn't want to do it himself

Thanks. 
Actually, I am now moving to another nation and would not have enough time in a few weeks.
I will appreciate if you could take care of this.

shoyer:
I'm going to merge this after the Appveyor tests pass, unless there are any further objections.

shoyer:
@rabernat let me know if this makes sense to you / if you agree.

doc/computation.rst
```diff
@@ -200,6 +200,26 @@ You can also use ``construct`` to compute a weighted rolling sum:
...
+    a.differentiate('x')
+
+This method can be used also for multidimensional arrays,
+
+.. ipython:: python
+    a = xr.DataArray(np.arange(8).reshape(4, 2), dims=['x', 'y'],
+                     coords=[0.1, 0.11, 0.2, 0.3])
+    a.differentiate('x')
+
+
```
rabernat: This is a place where we could mention the limitations of `differentiate` and `gradient` for non-cartesian geometry.

xarray/core/dataarray.py
```diff
@@ -2289,6 +2292,57 @@ def rank(self, dim, pct=False, keep_attrs=False):
         ds = self._to_temp_dataset().rank(dim, pct=pct, keep_attrs=keep_attrs)
         return self._from_temp_dataset(ds)
 
+    def differentiate(self, coord, edge_order=1, datetime_unit=None):
+        """ Differentiate the array with the second order accurate central
+        differences.
```
rabernat: and here

xarray/core/dataset.py
```diff
@@ -3663,6 +3666,58 @@ def rank(self, dim, pct=False, keep_attrs=False):
         attrs = self.attrs if keep_attrs else None
         return self._replace_vars_and_dims(variables, coord_names, attrs=attrs)
 
+    def differentiate(self, coord, edge_order=1, datetime_unit=None):
+        """ Differentiate with the second order accurate central
+        differences.
```
rabernat: and here

fujiisoup:
Thanks, @rabernat for the review.

Added the limitation of this method to docs.

PR #2405, opened 2018-09-06
WIP: don't create indexes on multidimensional dimensions

### LINKED ISSUE #2233: Problem opening unstructured grid ocean forecasts with 4D vertical coordinates
We can't open the IOOS New England triangular mesh ocean forecasts with Xarray because it doesn't understand their more complex CF vertical coordinate system.  

```python
import xarray as xr
url='http://www.smast.umassd.edu:8080/thredds/dodsC/FVCOM/NECOFS/Forecasts/NECOFS_GOM3_FORECAST.nc'
xr.open_dataset(url)
```
fails with:
```
MissingDimensionsError: 'siglay' has more than 1-dimension and the same name as one of its dimensions ('siglay', 'node'). xarray disallows such variables because they conflict with the coordinates used to label dimensions.
```
If you open this dataset with `nc = netCDF4.Dataset(url)` you can see what the data variables (e.g. `temp`) and coordinates (e.g. `siglay`) look like:
```
print(nc['temp'])

<class 'netCDF4._netCDF4.Variable'>
float32 temp(time, siglay, node)
    long_name: temperature
    standard_name: sea_water_potential_temperature
    units: degrees_C
    coordinates: time siglay lat lon
    type: data
    coverage_content_type: modelResult
    mesh: fvcom_mesh
    location: node
unlimited dimensions: time
current shape = (145, 40, 53087)

print(nc['siglay'])

<class 'netCDF4._netCDF4.Variable'>
float32 siglay(siglay, node)
    long_name: Sigma Layers
    standard_name: ocean_sigma_coordinate
    positive: up
    valid_min: -1.0
    valid_max: 0.0
    formula_terms: sigma: siglay eta: zeta depth: h
unlimited dimensions: 
current shape = (40, 53087)
```
So the `siglay` variable in this dataset specifies the fraction of water column contained in the layer and because this fraction changes over the grid, it has dimensions of `siglay` and `node`.   The variable `siglay` is just one of the variables used in the calculation of this CF-compliant vertical coordinate.   The actual vertical coordinate (after computation via `formula_terms`) ends up being 4D.

While we understand that there is no way to represent the vertical coordinate with a one-dimensional coordinate that xarray would like, it would be nice if there way to at least load the variable array data like `temp` into xarray.    We tried:
```
ds = xr.open_dataset(url,decode_times=False, decode_coords=False, decode_cf=False)
```
and we get the same error.

Is there any workaround for this?   

<details>

```
MissingDimensionsError                    Traceback (most recent call last)
<ipython-input-18-723c5c460db2> in <module>()
----> 1 xr.open_dataset(url)

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/backends/api.py in open_dataset(filename_or_obj, group, decode_cf, mask_and_scale, decode_times, autoclose, concat_characters, decode_coords, engine, chunks, lock, cache, drop_variables, backend_kwargs)
    344             lock = _default_lock(filename_or_obj, engine)
    345         with close_on_error(store):
--> 346             return maybe_decode_store(store, lock)
    347     else:
    348         if engine is not None and engine != 'scipy':

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/backends/api.py in maybe_decode_store(store, lock)
    256             store, mask_and_scale=mask_and_scale, decode_times=decode_times,
    257             concat_characters=concat_characters, decode_coords=decode_coords,
--> 258             drop_variables=drop_variables)
    259 
    260         _protect_dataset_variables_inplace(ds, cache)

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/conventions.py in decode_cf(obj, concat_characters, mask_and_scale, decode_times, decode_coords, drop_variables)
    428         vars, attrs, concat_characters, mask_and_scale, decode_times,
    429         decode_coords, drop_variables=drop_variables)
--> 430     ds = Dataset(vars, attrs=attrs)
    431     ds = ds.set_coords(coord_names.union(extra_coords).intersection(vars))
    432     ds._file_obj = file_obj

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/core/dataset.py in __init__(self, data_vars, coords, attrs, compat)
    363             coords = {}
    364         if data_vars is not None or coords is not None:
--> 365             self._set_init_vars_and_dims(data_vars, coords, compat)
    366         if attrs is not None:
    367             self.attrs = attrs

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/core/dataset.py in _set_init_vars_and_dims(self, data_vars, coords, compat)
    381 
    382         variables, coord_names, dims = merge_data_and_coords(
--> 383             data_vars, coords, compat=compat)
    384 
    385         self._variables = variables

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/core/merge.py in merge_data_and_coords(data, coords, compat, join)
    363     indexes = dict(extract_indexes(coords))
    364     return merge_core(objs, compat, join, explicit_coords=explicit_coords,
--> 365                       indexes=indexes)
    366 
    367 

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/core/merge.py in merge_core(objs, compat, join, priority_arg, explicit_coords, indexes)
    433     coerced = coerce_pandas_values(objs)
    434     aligned = deep_align(coerced, join=join, copy=False, indexes=indexes)
--> 435     expanded = expand_variable_dicts(aligned)
    436 
    437     coord_names, noncoord_names = determine_coords(coerced)

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/core/merge.py in expand_variable_dicts(list_of_variable_dicts)
    209                     var_dicts.append(coords)
    210 
--> 211                 var = as_variable(var, name=name)
    212                 sanitized_vars[name] = var
    213 

~/miniconda3/envs/pangeo/lib/python3.6/site-packages/xarray/core/variable.py in as_variable(obj, name)
    112                 'dimensions %r. xarray disallows such variables because they '
    113                 'conflict with the coordinates used to label '
--> 114                 'dimensions.' % (name, obj.dims))
    115         obj = obj.to_index_variable()
    116 

MissingDimensionsError: 'siglay' has more than 1-dimension and the same name as one of its dimensions ('siglay', 'node'). xarray disallows such variables because they conflict with the coordinates used to label dimensions.
```
</details>

### LINKED ISSUE #2368: Let's list all the netCDF files that xarray can't open
At the Pangeo developers meetings, I am hearing lots of reports from folks like @dopplershift and	@rsignell-usgs about netCDF datasets that xarray can't open.

My expectation is that xarray doesn't have strong requirements on the contents of datasets. (It doesn't "enforce" cf compatibility for example; that's optional.) Anything that can be written to netCDF should be readable by xarray.

I would like to collect examples of places where xarray fails. So far, I am only aware of one:

- Self-referential multidimensional coordinates (#2233). Datasets which contain variables like `siglay(siglay, node)`. Only `siglay(siglay)` would work.

__Are there other distinct cases?__

Please provide links / sample code of netCDF datasets that xarray can't read. Even better would be short code snippets to create such datasets in python using the netcdf4 interface.

rabernat:
- [x] Closes #2368, Closes #2233

This is just a start to the solution proposed in #2368. A surprisingly small number of tests broke in my local environment.


xarray/core/formatting.py
```diff
@@ -272,7 +272,8 @@ def summarize_datavar(name, var, col_width):
 
 
 def summarize_coord(name, var, col_width):
-    is_index = name in var.dims
+    from .variable import IndexVariable
```
rabernat: I tried moving this import to the top of `formatting.py` but, for reasons I can't understand, I got this error
```
    from .variable import IndexVariable
E   ImportError: cannot import name 'IndexVariable
```

Why does it work in the function but not at the top?

rabernat:
Seeking suggestions about what else needs to be tested here.

xarray/core/formatting.py
markelg: I think there is a circular import going on here, as .variable does import .common which does import .formatting Perhaps adding a self.is_index attribute to the Variable class could be a way to avoid this import?

shoyer:
I turned up a few lines that probably need to be fixed, just by grepping for `in\ \w+\.dims`:
https://github.com/pydata/xarray/blob/66a8f8dd7f5a2997ff614f3966d1951587915e7e/xarray/backends/api.py#L125-L126
https://github.com/pydata/xarray/blob/66a8f8dd7f5a2997ff614f3966d1951587915e7e/xarray/core/coordinates.py#L111-L116
https://github.com/pydata/xarray/blob/66a8f8dd7f5a2997ff614f3966d1951587915e7e/xarray/core/dataarray.py#L502

It looks like this change broke `align` somehow -- possibly we have a bug where some indexes were not being created as `IndexVariable` objects?

xarray/core/formatting.py
shoyer: Yes, we have a circular import here. We could potentially refactor to avoid but I think this is also an OK work-around.

shoyer:
It would be good to add a tests for these variables in:
- The `data_vars` argument to the Dataset constructor (these objects should become data_vars, not coords)
- The DataArray constructor.
- Explicitly adding these variables:
    - With `Dataset.__setitem__`, `Dataset.coords.__setitem__` and `DataArray.coords.__setitem__`
    - When there are no existing variables matching the dimension name, and when a 1D index variable matching the dimension name already exists (the existing index should be deleted/cleared)
- Concatenating along an *existing* dimension whose name matches existing variables (e.g., concatenate along `x` when a variable `x` with dimensions `('x', 'y')` exists)
- Concatenating along a new dimension whose name matches existing variables (e.g., concatenate along `x` when a variable `x` with dimensions `('y',)` exists)
- More generally, test variables where the name matches a dimension but the variable does not include that dimension at all, e.g., a variable `'x'` with dimensions `('y',)` in a Dataset where `x` is a dimension.

rabernat:
With 68f170c there are basically no failing tests in the original test suite. I will now work on adding more tests along the lines outlined by @shoyer.

shoyer:
I think this function mostly makes sense, but we would want to drop the stuff for scalar variables.

Unfortunately I don't know a good way to fix this stuff short of auditing a lot of code manually -- this is a hazard of data model changes. On the plus side, I've also started to do some of this for the explicit index refactor.

One thing that might turn up a few bugs is to try adding such a variable to the create_test_data() helper function which gets used in lots of places.

PR #2418, opened 2018-09-18
fixed possible typo

Schlump:
from pydata.cas.urs import setup_session

should be:
from pydap.cas.urs import setup_session


shoyer:
thanks!

PR #2420, opened 2018-09-18
Fix small typo in docs

canyon289:

I believe this is a minor fix that means I don't need the things above, but If I'm mistaken please let me know and I can update the PR.

Thanks in advance!


fujiisoup:
Thanks, @canyon289.

Issue #2422, opened 2018-09-19
Plot2D no longer sorts coordinates before plotting


fmaussion:
I have a dataset with decreasing latitude coordinates.

With ``xarray`` v0.10.8, this is what happens when plotting:

![i1](https://user-images.githubusercontent.com/10050469/45765619-b58ceb00-bc35-11e8-94b4-5e37cb9f2a0a.png)

But on latest master the image is now upside down:

![i2](https://user-images.githubusercontent.com/10050469/45765681-e53bf300-bc35-11e8-95cc-8e4fb520bfe5.png)

Sorry if I missed a change along the way, I was off for a long time.

shoyer:
The most likely source here is https://github.com/pydata/xarray/pull/2294

fmaussion:
Looks like a difficult thing to test, also...

dcherian:
Yes #2294 changed the default `xincrease, yincrease` to be `None` rather than `True`. Previously, if `xincrease` was `True` we were setting axis limits and this prevented the axis from auto-updating its limits. Now that we are using the `invert_axis()` method, this might not be an issue. (Actually, now that I look at the code for `invert_axis()`, all I needed to do earlier was set the kwarg `auto=None` in the call to `set_xlim`, so this would be an easy way to revert to the previous code.)

I switched the default to `None` because I thought it was better to make the least number of modifications to the axis by default.

That said, I cannot reproduce this with a dataset with decreasing latitude so I'm very confused about what's happening.
```
da = xr.DataArray(np.arange(24).reshape(3,8), dims=('latitude', 'longitude'),
                 coords={'latitude': [8, 0, -8],
                        'longitude': np.arange(8)})

da.plot()
```

PS: I'm on a ship in the middle of the Pacific with crappy internet access. @fmaussion I'd appreciate it if you could look at this so we can get the release out. I will respond but there'll be a large time lag.

fmaussion:
> I cannot reproduce this with a dataset with decreasing latitude 

Maybe a MPL version issue? Because our docs are also afffected, see:
- http://xarray.pydata.org/en/stable/plotting.html#two-dimensions
- http://xarray.pydata.org/en/latest/plotting.html#two-dimensions

> @fmaussion I'd appreciate it if you could look at this

I Will have a look now

shoyer:
I think this is fixed by #2425?

PR #2294, opened 2018-07-16
Additional axis kwargs

dcherian:
- [x] Closes #2224 (remove if there is no corresponding issue, which should only be the case for minor changes)
 - [x] Tests added (for all bug fixes or enhancements)
 - [x] Tests passed (for all non-documentation changes)
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

This PR adds support for `xscale, yscale, xticks, yticks, xlim, ylim` kwargs following discussion in #2224 and the Pandas API (https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.plot.html#pandas.DataFrame.plot).

Haven't added FacetGrid support yet. I'll get to that soon.

PR #2425, opened 2018-09-20
Plotting: restore xyincrease kwarg default to True

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/2422
 - [x] Tests added 
 - [x] Tests passed 

https://github.com/pydata/xarray/pull/2294 introduced the new ``xincrease`` and ``yincrease`` kwargs and a behavior documented as:

```
    xincrease : None, True, or False, optional
        Should the values on the x axes be increasing from left to right?
        if None, use the default for the matplotlib function.
```

The default was ``None``, I suggest to set it to ``True``: it was the default before https://github.com/pydata/xarray/pull/2294 and I think it is what most users expect.


dcherian:
LGTM. Thanks!

shoyer:
thanks!

PR #2426, opened 2018-09-21
Fix/misc fixes

dcherian:

Some minor fixes.
1. Don't explicitly set rotation on colorbar label. matplotlib automatically does the right thing for horizontal and vertical colorbars.

2. Adds a test for whether an object is mappable before adding it to `FacetGrid._mappables`.


shoyer:
LGTM

@fmaussion can you take a quick look?

fmaussion:
Appart from a pep8 line length problem LGTM too!

Sitckler is on strike (https://github.com/pydata/xarray/issues/2428)

PR #2432, opened 2018-09-23
switch travis language to generic

jhamman:
Following up on #2271. This switches the set language in our Travis-CI config from "python" to "generic". Since we don't use any of the Travis Python utilities, we didn't really need the python setting and the generic setting gives a few benefits:

- smaller base image which should give a bit faster spin-up time
- build matrix without reliance on python version, instead we just point to the conda environment file


shoyer:
Nice! Would it work to remove the redundant `name` field?

shoyer:
Looks great, thanks!

jhamman:
Yes, this looks good to go now.

PR #2433, opened 2018-09-23
Contour labels kwarg

dcherian:

Adds a `labels` boolean kwarg for `contour` that adds contour labels. Also adds `clabel_kwargs` that is passed on to `Axes.clabel()`

`air.isel(time=0).plot.contour(labels=True, colors='k', clabel_kwargs={'fmt': '%.1f'})`

image

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -33,6 +33,10 @@ v0.11.0 (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- :py:meth:`plot()` now accepts the kwargs ``labels, clabel_kwargs`` to add
+  labels to contours.
+  By `Deepak Cherian <https://github.com/dcherian>`_. (:issue:`2224`)
+
 Bug fixes
 ~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


dcherian:
True. 

The equivalent for facetgrid is 
```
g = air.isel(time=slice(0, 2)).plot.contour(col='time', colors='k')
for cs, ax in zip(g._mappables, g.axes.flat):
    ax.clabel(cs)
```

So adding this would make that a little more convenient. If not, I can send in a PR to document this bit (the use of `_mappables` is not very obvious; maybe we should make it public?)

fmaussion:
> If not, I can send in a PR to document this bit 

I've used the [gallery](http://xarray.pydata.org/en/latest/auto_gallery/plot_cartopy_facetgrid.html#sphx-glr-auto-gallery-plot-cartopy-facetgrid-py) to document these kind of things. It's a bit lost in the doc currently though.

> (the use of _mappables is not very obvious; maybe we should make it public?)

Looks like we need it for these kind of feature, right?

PR #2444, opened 2018-09-27
facetgrid: properly support cbar_kwargs.

### LINKED ISSUE #1504: cbar_kwargs for faceting plots
I'm using faceting and the add_colorbar=True argument along with cbar_kwargs. This triggers the following error message:

```
ValueError: cbar_ax and cbar_kwargs can't be used with add_colorbar=False.
```

I searched the facetgrid.py module and found the following lines:
```python
func_kwargs = kwargs.copy()
func_kwargs.update(cmap_params)
func_kwargs.update({'add_colorbar': False, 'add_labels': False})
```
I realized that if add_colorbar is True, then the facet function draws one colorbar for each subplot. My initial intention was providing cbar_kwargs to reduce the whitespace between the colorbar en the right edge of the plot.

Is there any way to fix this?

### LINKED ISSUE #1717: colorbars in facet grids

dcherian:
- [x] Closes #1504, closes #1717, closes #1735 

#1735 is stalled, so I jumped in. 

I've added an error if `cbar_ax` is provided as an option for `FacetGrid`. Don't think it's really needed.


doc/plotting.rst
```diff
@@ -495,7 +495,8 @@ Faceted plotting supports other arguments common to xarray 2d plots.
 
     @savefig plot_facet_robust.png
     g = hasoutliers.plot.pcolormesh('lon', 'lat', col='time', col_wrap=3,
-                                    robust=True, cmap='viridis')
+                                    robust=True, cmap='viridis',
+				    cbar_kwargs={'label': 'this has outliers'})
```
fmaussion:
```suggestion
				     cbar_kwargs={'label': 'this has outliers'})
```

xarray/plot/facetgrid.py
```diff
@@ -222,6 +222,11 @@ def map_dataarray(self, func, x, y, **kwargs):
 
         cmapkw = kwargs.get('cmap')
         colorskw = kwargs.get('colors')
+        cbar_kwargs = kwargs.pop('cbar_kwargs', {})
+        cbar_kwargs = {} if cbar_kwargs is None else dict(cbar_kwargs)
+
+        if 'cbar_ax' in kwargs and kwargs['cbar_ax'] is not None:
```
fmaussion:
```suggestion
        if kwargs.get('cbar_ax', None) is not None:
```

PR #2449, opened 2018-10-01
Add 'to_iris' and 'from_iris' to methods Dataset

jacobtomlinson:
This PR adds `to_iris` and `from_iris` methods to DataSet. I've added this because I frequently find myself writing little list and dictionary comprehensions to pack and unpack both DataSets from DataArrays and Iris CubeLists from Cubes.

### DIFF
--- a/xarray/convert.py
+++ b/xarray/convert.py
@@ -11,7 +11,7 @@ from .coding.times import CFDatetimeCoder, CFTimedeltaCoder
 from .conventions import decode_cf
 from .core import duck_array_ops
 from .core.dataarray import DataArray
-from .core.dataset import DataSet
+from .core.dataset import Dataset
 from .core.dtypes import get_fill_value
 from .core.pycompat import OrderedDict, range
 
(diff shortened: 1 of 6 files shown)


xarray/convert.py
```diff
@@ -273,3 +274,16 @@ def from_iris(cube):
...
+    """ Convert a Dataset into an Iris CubeList.
+    """
+    from iris.cube import CubeList
+    return CubeList([dataset[variable].to_iris() for variable in list(dataset.data_vars)])
+
+
+def dataset_from_iris(cubelist):
+    """ Convert an Iris CubeList into a Dataset.
+    """
+    return Dataset({cube.var_name: DataArray.from_iris(cube) for cube in cubelist})
```
shoyer: Can we use the `name` attribute already on `DataArray.from_iris(cube)`? We have some special logic already for figuring out names in `DataArray.from_iris`.

jacobtomlinson: So `cube.name` instead of `cube.var_name`?

jacobtomlinson:
Thanks for the feedback.

I think when working in iris that you expect multiple cubes to be returned as a cube list. Also when calling `iris.load` you expect a cubelist. 

I would propose that `to_iris` is fine, however if you feel strongly I would be happy to change it.

xarray/convert.py
shoyer: I would say `DataArray.from_iris(cube).name`, but it probably makes sense save it in an intermediate variable to avoid converting the cube twice.

shoyer:
I'm happy to defer to Iris users (like you) on what they would expect for converting to/from an xarray.Dataset.

jhamman:
I wonder if we can get @pelson to weigh in here. Like @shoyer said, whatever the Iris users think makes most sense for the naming of these methods is fine by me.

jhamman:
@DPeterK / @jacobtomlinson - this has grown quite stale. Any interest in finishing this up or should we close this in favor of a new contribution down the line?

trexfeathers:
@pp-mo 👀

jacobtomlinson:
@jhamman this PR is 5 years old and my need for it has long since gone away. If other folks want to pick this up that would be awesome.

PR #2462, opened 2018-10-04
pep8speaks

### LINKED ISSUE #2428: Stickler-ci
The last time stickler had a look at our PRs is 8 days ago (https://github.com/pydata/xarray/pull/2415) : https://stickler-ci.com/repositories/26661-pydata-xarray

It looks like their bot is broken: https://github.com/stickler-ci

This is not very trustworthy - we could consider switching to https://pep8speaks.com/ maybe

fujiisoup:
- [x] Closes #2428

I installed pep8speaks as suggested in #2428.
It looks they do not need a yml file, but it may be safer to add this (just renamed from `.stickler.yml`)


fujiisoup:
@pep8speaks suggest the diff

fmaussion:
Thanks! I just switched off stickler which is still broken anyway.

fmaussion:
I think we need a set-up file to set diff_only on PR. See the pandas pep8speaks template: https://github.com/pandas-dev/pandas/blob/master/.pep8speaks.yml

max-sixty:
Ref https://github.com/pydata/xarray/pull/2467#issuecomment-427535714

max-sixty:
> I think we need a set-up file to set diff_only on PR

Why do we need `diff_only`? The repo passes `flake8` as a whole on my local

fmaussion:
> Why do we need diff_only

I was worried about all the unrelated pep8 errors on this PR:  https://github.com/pydata/xarray/pull/2463

Although I'm not sure where they come from.

max-sixty:
Good point. 

That's odd - I don't get those on #2467, or locally.

Is there an issue around ignoring `# noqa`?

fmaussion:
I also have the impression that noqa are ignored by pep8speask, although I haven't seen any related issue on [their repo](https://github.com/OrkoHunter/pep8speaks/issues). Probably another reason for a diff_only option for now.

max-sixty:
OK. Though if people make changes to `# noqa` lines, we'll have false positives.

Weird that @stickler-ci broke. We're still using it at pandas-gbq and it's working well.

fujiisoup:
I just updated .pep8speaks.yml that is taken from pandas-dev/pandas.
If there are no further comments, I would merge this. Let's see how it works with this configuration.

shoyer:
👍  Let's give this is a try

PR #2465, opened 2018-10-05
Add python_requires to setup.py

djhoese:
This should assist users using PyPI/pip to install xarray when xarray drops python 2.7 support. The `python_requires` argument is described here:

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

Basically, it says what versions of python your package supports. When xarray drops python 2 support and this kwarg gets update it should stop people on python 2 from getting the wrong package.


djhoese:
What I have it set to in this PR should match the classifiers in the setup.py: 2.7 and 3.5+

shoyer:
thanks!

PR #2467, opened 2018-10-06
Replace the last of unittest with pytest

max-sixty:
This replaces (by-and-large) any remaining unittest implementations with pytest implementations. This allows us to use fixtures on any test; the existing implementation does not let us use them on `TestCase` instances. It also directs future contributors away from using the legacy methods.

Would be great to merge this fairly soon, as there's some chance of merge conflicts (which git may not catch - e.g. someone writing `self.assertEqual`). I'll try and chase down any test breaks as soon as possible.

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -2433,7 +2433,7 @@ class PydapTest(object):
             assert actual.attrs.keys() == expected.attrs.keys()
 
         with self.create_datasets() as (actual, expected):
-            assert_equal(actual.isel(l=2), expected.isel(l=2))  # noqa: E741
+            assert_equal(actual.isel(l=2), expected.isel(l=2))  # noqa
 
         with self.create_datasets() as (actual, expected):
             assert_equal(actual.isel(i=0, j=-1),
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -626,20 +626,20 @@ class CFEncodedDataTest(DatasetIOTestCases):
         encoded = create_encoded_unsigned_masked_scaled_data()
         with self.roundtrip(decoded) as actual:
             for k in decoded.variables:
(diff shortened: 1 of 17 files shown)


xarray/tests/test_backends.py
```diff
@@ -1532,14 +1529,14 @@ def test_encoding_chunksizes(self):
 
 
 @requires_zarr
-class ZarrDictStoreTest(BaseZarrTest, TestCase):
+class ZarrDictStoreTest(BaseZarrTest, object):
```
shoyer: There's no need to include `object` as a second base class for these classes :)

xarray/tests/test_plot.py
```diff
@@ -65,7 +70,7 @@ def easy_array(shape, start=0, stop=1):
 
 
 @requires_matplotlib
-class PlotTestCase(TestCase):
+class PlotTestCase(unittest.TestCase):
```
shoyer: I thought we were replacing the last of `unittest` here?

No worries on saving this for later, though :)

xarray/tests/test_backends.py
```diff
@@ -626,35 +626,35 @@ def test_unsigned_roundtrip_mask_and_scale(self):
         encoded = create_encoded_unsigned_masked_scaled_data()
         with self.roundtrip(decoded) as actual:
             for k in decoded.variables:
-                self.assertEqual(decoded.variables[k].dtype,
-                                 actual.variables[k].dtype)
+                assert decoded.variables[k].dtype == \
```
shoyer: I prefer using parentheses to explicit line continuations (per PEP8), but I'm going to try to avoid being picky about it until we have automated tools that complain about it :)

shoyer:
we can probably go ahead and safely merge this in its current state -- none of my comments are deal breakers

shoyer:
Thanks Maximilian!

PR #2469, opened 2018-10-06
isort

max-sixty:
Do we want to keep `isort` up to date? I think it's a balance between consistency vs. overhead


shoyer:
thanks @max-sixty !

Issue #2451, opened 2018-10-01
Shift changes non-float arrays to object, even for shift=0


max-sixty:
```python
In [15]: xr.DataArray(np.random.randint(2,size=(100,100)).astype(bool)).shift(dim_0=0)
Out[15]:
<xarray.DataArray (dim_0: 100, dim_1: 100)>
array([[False, True, True, ..., True, True, False],
       [False, True, False, ..., False, True, True],
       [False, True, False, ..., False, True, False],
       ...,
       [False, True, False, ..., False, True, True],
       [True, False, True, ..., False, False, False],
       [False, True, True, ..., True, True, False]], dtype=object) # <-- could be bool
Dimensions without coordinates: dim_0, dim_1

```
#### Problem description

This causes memory bloat

#### Expected Output

As above with `dtype=bool`

#### Output of ``xr.show_versions()``

<details>

In [16]: xr.show_versions()

INSTALLED VERSIONS
------------------
commit: f9c4169150286fa1aac020ab965380ed21fe1148
python: 2.7.15.final.0
python-bits: 64
OS: Darwin
OS-release: 18.0.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: None.None

xarray: 0.10.9+12.gf9c41691
pandas: 0.22.0
numpy: 1.14.2
scipy: 1.0.0
netCDF4: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: None
PseudonetCDF: None
rasterio: None
iris: None
bottleneck: 1.2.1
cyordereddict: None
dask: None
distributed: None
matplotlib: 2.1.2
cartopy: None
seaborn: 0.8.1
setuptools: 39.2.0
pip: 18.0
conda: None
pytest: 3.6.3
IPython: 5.8.0
sphinx: None
</details>

The `shift=0` is mainly theoretical. To avoid casting to `object` in practical scenarios, we could add a `fill_value` argument (e.g. `fill_value=False`) and fill with that rather than `NaN`

CC @Ivocrnkovic

shoyer:
+1 for adding a `fill_value` argument.

-1 for changing the behavior for `shift=0` in particular. It's generally not a good idea to make the type/dtype of the result of operations dependent on the values of input arguments -- that sort of things can easily lead to bugs.

max-sixty:
This was closed by #2470

PR #2470, opened 2018-10-06
fill_value in shift

max-sixty:
- [x] Closes #https://github.com/pydata/xarray/issues/2451 
 - [x] Tests added 
 - [x] Tests passed 

Should we be more defensive around which fill_values can be passed? Currently, if the array and float values have incompatible dtypes, we don't preemtively warn or cast, apart from the case of `np.nan`, which then uses the default filler


xarray/core/dataarray.py
```diff
@@ -2117,8 +2119,8 @@ def shift(self, shifts=None, **shifts_kwargs):
         Coordinates:
           * x        (x) int64 0 1 2
         """
-        ds = self._to_temp_dataset().shift(shifts=shifts, **shifts_kwargs)
-        return self._from_temp_dataset(ds)
+        return self._replace(variable=self.variable.shift(
+            shifts=shifts, fill_value=fill_value, **shifts_kwargs))
```
max-sixty: Is this a reasonable change for `DataArray` operations that only change their data?

shoyer: I think this is 100% equivalent -- I don't really care either way. But perhaps splitting this into two statements would be clearer.

xarray/core/dataset.py
```diff
@@ -3529,7 +3531,8 @@ def shift(self, shifts=None, **shifts_kwargs):
             if name in self.data_vars:
                 var_shifts = dict((k, v) for k, v in shifts.items()
                                   if k in var.dims)
-                variables[name] = var.shift(**var_shifts)
+                variables[name] = var.shift(
+                    fill_value=fill_value, **var_shifts)
```
shoyer: side note: we should probably pass `var_shifts` by position rather than as keyword arguments, so we can support non-string keys.

xarray/core/variable.py
```diff
@@ -940,7 +940,11 @@ def _shift_one_dim(self, dim, count):
             keep = slice(None)
 
         trimmed_data = self[(slice(None),) * axis + (keep,)].data
-        dtype, fill_value = dtypes.maybe_promote(self.dtype)
+
+        if fill_value is None or fill_value is np.nan:
```
shoyer: Use `dtypes.NA` as the default value for `fill_value`, and then copy these lines from `pad` to ensure that this works for arbitrary fill_values such as `None`:
https://github.com/pydata/xarray/blob/master/xarray/core/variable.py#L1012-L1015

max-sixty: I saw that, but it means that if someone passes `np.nan` (the value, not the `dtypes.NA` default) to a int dtype, I get `-9223372036854775808`. Should we raise in that case? Or I'll see what I can do to coerce.

max-sixty: > I'm mostly concerned with duplicate logic that works slightly differently
💯 

Great, I'll have a look and report back

xarray/core/variable.py
```diff
@@ -944,7 +944,11 @@ def _shift_one_dim(self, dim, count):
             keep = slice(None)
 
         trimmed_data = self[(slice(None),) * axis + (keep,)].data
-        dtype, fill_value = dtypes.maybe_promote(self.dtype)
+
+        if fill_value is dtypes.NA and True:
+            dtype, fill_value = dtypes.maybe_promote(self.dtype)
+        else:
+            dtype = self.dtype
```
fujiisoup: What happens if filler is not compatible with self.dtype?
For example, feeding np.nan to an int array.
Probably it is a part of user responsibility and we do not need to take care of this, but I am just curious of it.

xarray/core/variable.py
```diff
@@ -944,7 +944,11 @@ def _shift_one_dim(self, dim, count):
             keep = slice(None)
 
         trimmed_data = self[(slice(None),) * axis + (keep,)].data
-        dtype, fill_value = dtypes.maybe_promote(self.dtype)
+
+        if fill_value is dtypes.NA and True:
```
shoyer: We can drop `and True` here

xarray/core/variable.py
shoyer: In theory, NumPy should raise an error... But it may not.

max-sixty: (this is the issue I'm looking at ref https://github.com/pydata/xarray/pull/2470#discussion_r223372119), good foresight @fujiisoup !

max-sixty:
I think this is in a decent place. 

If `np.nan` is passed as `fill_value` into an int array, we'll get an odd result. But if they leave it to default, it'll work correctly, and I couldn't find a robust general way of solving that

Let me know thoughts!
Max

doc/whats-new.rst
```diff
@@ -180,6 +180,12 @@ Enhancements
   :py:meth:`~xarray.DataArray.interp`, and
   :py:meth:`~xarray.Dataset.interp`.
   By `Spencer Clark <https://github.com/spencerkclark>`_
+- Added support for ``fill_value`` with
```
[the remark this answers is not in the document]
max-sixty: Thanks. Shouldn't have left it so long

PR #2471, opened 2018-10-06
Tests passing in PR template

max-sixty:
Do we need this bullet? I don't generally ensure tests pass before submitting for review. (Although if I should be, then lmk)


shoyer:
indeed -- I don't bother checking this, either :)

Issue #2475, opened 2018-10-09
Please add support for GRIB files via ECMWF cfgrib / ecCodes


alexamici:
The [*cfgrib* project](https://pypi.org/project/cfgrib/) offers to open a GRIB file as an `xarray.Dataset`:
```
>>> import cfgrib
>>> cfgrib.open_dataset('era5-levels-members.grib')
<xarray.Dataset>
Dimensions:       (air_pressure: 2, latitude: 61, longitude: 120, number: 10, time: 4)
Coordinates:
  * number        (number) int64 0 1 2 3 4 5 6 7 8 9
  * time          (time) datetime64[ns] 2017-01-01 ... 2017-01-02T12:00:00
    step          timedelta64[ns] ...
  * air_pressure  (air_pressure) float64 850.0 500.0
  * latitude      (latitude) float64 90.0 87.0 84.0 81.0 ... -84.0 -87.0 -90.0
  * longitude     (longitude) float64 0.0 3.0 6.0 9.0 ... 351.0 354.0 357.0
    valid_time    (time) datetime64[ns] ...
Data variables:
    z             (number, time, air_pressure, latitude, longitude) float32 ...
    t             (number, time, air_pressure, latitude, longitude) float32 ...
Attributes:
    GRIB_edition:            1
    GRIB_centre:             ecmf
    GRIB_centreDescription:  European Centre for Medium-Range Weather Forecasts
    GRIB_subCentre:          0
    history:                 GRIB to CDM+CF via cfgrib-0.9.../ecCodes-2...
``` 
It would be very helpful to have a backend for it distributed with *xarray* to be selected via the `engine` keyword argument:
```
>>> import xarray as xr
>>> xr.open_dataset('era5-levels-members.grib', engine='cfgrib')
<xarray.Dataset>
Dimensions:       (air_pressure: 2, latitude: 61, longitude: 120, number: 10, time: 4)
Coordinates:
  * number        (number) int64 0 1 2 3 4 5 6 7 8 9
  * time          (time) datetime64[ns] 2017-01-01 ... 2017-01-02T12:00:00
    step          timedelta64[ns] ...
  * air_pressure  (air_pressure) float64 850.0 500.0
  * latitude      (latitude) float64 90.0 87.0 84.0 81.0 ... -84.0 -87.0 -90.0
  * longitude     (longitude) float64 0.0 3.0 6.0 9.0 ... 351.0 354.0 357.0
    valid_time    (time) datetime64[ns] ...
[...]

PR #2476, opened 2018-10-09
Add a GRIB backend via ECMWF cfgrib / ecCodes

alexamici:
This is currently a WIP PR for review.
 - [x] Addresses #2475 - only read support is proposed for now.
 - [x] Tests added.
 - [x] Documented, including `whats-new.rst` for all changes and `api.rst` for new API.

The implementation depends on the python module *cfgrib* and the C-library ecCodes to be installed.

Work in progress items:
 - [x] the coordinate rename doesn't really belong here, move it to *cfgrib*,
 - [x] port cfgrib backend to use the new `CachingFileManager` interface
 - [x] implement proper locking
 - [x] test *dask* support (real performance with *dask* depends on saving the external index)
 - ~~~the encoding options are needlessly different from the standard ones, try to sync them~~~ not urgent

cc @StephanSiemen @iainrussell


shoyer:
@alexamici great to see this!

I'm about to merge a refactor of xarray's backends for v0.11 in https://github.com/pydata/xarray/pull/2261. You'll need to refactor a little bit to accommodate this, but hopefully that should be straightforward. The new interface should be a little easier to use, especially for handling many files or dask-distributed support.

xarray/backends/cfgrib_.py
```diff
@@ -0,0 +1,125 @@
...
+
+        array = self.backend_array[key.tuple]
+
+        if len(np_inds.tuple) > 0:
+            array = core.indexing.NumpyIndexingAdapter(array)[np_inds]
+
+        return array
+
+
+FLAVOURS = {
```
shoyer: Is this configuration dict something that could live in cfgrid rather than xarray?

xarray/backends/cfgrib_.py
```diff
@@ -0,0 +1,125 @@
...
+}
+
+
+class CfGribDataStore(common.AbstractDataStore):
+    """
+    Implements the ``xr.AbstractDataStore`` read-only API for a GRIB file.
+    """
+    def __init__(self, ds, variable_map={}, autoclose=False):
+        self.ds = ds
+        self.variable_map = variable_map.copy()
```
shoyer: Is there a particular reason for including `variable_map` in the interface on the DataStore class, rather than letting users change variable names later with `Dataset.rename`?

xarray/backends/cfgrib_.py
alexamici: Yes, will do.

xarray/backends/cfgrib_.py
alexamici: It is the same issue as above. The whole coordinate rename doesn't belong here and will be moved in *cfgrib*.

alexamici:
I quickly sync'ed with the new backend API. I did some light testing.

shoyer:
The appropriate lock to use depends on `cfgrid`. Is the library thread-safe? If not, it's probably best to use a global (per process) lock.

pynio is probably the simplest example of how to write a new backend:
https://github.com/pydata/xarray/blob/master/xarray/backends/pynio_.py

The main difference is that you should make use of [`CachingFileManager`](https://github.com/pydata/xarray/blob/289b377129b18e7dc6da8336e958a85be868acbe/xarray/backends/file_manager.py#L34) to add support for dask and opening many files at once. File managers make it possible to pickle a datastore, which is what we need to make dask-distirbuted work.

alexamici:
The `cfgrib_.py` backend now uses the new `CachingFileManager` interface and I added a global process lock, just-in-case. The code is very simple and I mimicked the PyNIO backend so I expect it to work correctly with *dask*, but I'm not sure how to test for it.

Furthermore I expect *dask* performance to be abysmal until I implement ecmwf/cfgrib#20.

alexamici:
@shoyer BTW what timeframe do you expect for the v0.11 release? And would you consider merging this Pull Request before the release, assuming that we do a *cfgrib* release with read support declared *beta*?

xarray/backends/cfgrib_.py
```diff
@@ -50,28 +56,29 @@ class CfGribDataStore(AbstractDataStore):
...
-
-    @classmethod
-    def from_path(cls, path, lock=False, **backend_kwargs):
+    def __init__(self, filename, lock=None, **backend_kwargs):
         import cfgrib
-        return cls(ds=cfgrib.open_file(path, **backend_kwargs), lock=lock)
+        if lock is None:
+            lock = ECCODES_LOCK
+        self.lock = ensure_lock(lock)
+        backend_kwargs['filter_by_keys'] = tuple(backend_kwargs.get('filter_by_keys', {}).items())
```
shoyer: This looks a little surprising to me -- can we simply pass on `backend_kwargs` directly to cfgrib?

xarray/backends/cfgrib_.py
```diff
@@ -25,6 +25,12 @@
...
 from .common import AbstractDataStore, BackendArray
+from .file_manager import CachingFileManager
+from .locks import ensure_lock, SerializableLock
+
+# FIXME: Add a dedicated lock just in case, even if ecCodes is supposed to be thread-safe in most
+# circumstances. See: https://confluence.ecmwf.int/display/ECC/Frequently+Asked+Questions
+ECCODES_LOCK = SerializableLock()
 
 
 class CfGribArrayWrapper(BackendArray):
```
shoyer: This class will need a few updates (e.g., to make use of the `lock` from the datastore). Take a look at pynio for examples.

xarray/backends/cfgrib_.py
```diff
@@ -0,0 +1,99 @@
...
+# circumstances. See: https://confluence.ecmwf.int/display/ECC/Frequently+Asked+Questions
+ECCODES_LOCK = SerializableLock()
+
+
+class CfGribArrayWrapper(BackendArray):
+    def __init__(self, backend_array):
+        self.backend_array = backend_array
+
+    def __getattr__(self, item):
+        return getattr(self.backend_array, item)
```
shoyer: This sort of forwarding is really error prone. Let's avoid if it in favor of a explicit solution if at all possible

shoyer:
The v0.11 release still need more work (to complete deprecations), so it's
at least a week or two out (probably more). I suspect we could get this PR
ready together in time.

xarray/backends/cfgrib_.py
alexamici: This is a ugly hack araund the fact that `filter_by_keys` is a dict but `CachingFileManager` accepts only hashable `backend_kwargs` because they are passed to `_HashedSequence`.

`filter_by_keys` has a very nice dict semantics, so I'd prefer not to change it.

xarray/backends/cfgrib_.py
alexamici: Done.

xarray/backends/cfgrib_.py
alexamici: Done.

xarray/backends/cfgrib_.py
shoyer: OK, that makes please. I think the need to hash arguments used to open files with CachingFileManager is unavoidable, so this is a reasonable workaround. But please add a comment explaining this.

xarray/backends/cfgrib_.py
```diff
@@ -0,0 +1,96 @@
...
+class CfGribArrayWrapper(BackendArray):
+    def __init__(self, datastore, array):
+        self.datastore = datastore
+        self.shape = array.shape
+        self.dtype = array.dtype
+        self.array = array
+
+    def __getitem__(self, key):
+        return indexing.explicit_indexing_adapter(
+            key, self.shape, indexing.IndexingSupport.BASIC, self._getitem)
```
shoyer: Can you verify which forms of indexing cfgrib actually supports? In your previous commit this was OUTER_1VECTOR. See the docstring on IndexingSupport for details.

xarray/backends/cfgrib_.py
```diff
@@ -0,0 +1,96 @@
...
+# circumstances. See: https://confluence.ecmwf.int/display/ECC/Frequently+Asked+Questions
+ECCODES_LOCK = SerializableLock()
+
+
+class CfGribArrayWrapper(BackendArray):
+    def __init__(self, datastore, array):
+        self.datastore = datastore
+        self.shape = array.shape
+        self.dtype = array.dtype
+        self.array = array
```
shoyer: You probably do not want to store a direct reference to `array` here, because it is probably not pickleable, and this object needs to be pickleable to work with Dask distributed.

xarray/backends/cfgrib_.py
alexamici: Ok, I added a comment and made the code more explicit.

alexamici:
@shoyer thank you very much for the guidance, it is very appreciated!

I stared working on the tests, but I've been blocked immediately by something that looks trivial. I cannot get the test class `PyNioTest` to run:
```
$ python setup.py test --addopts -v | grep Test
...
xarray/tests/test_backends.py::TestCommon::test_robust_getitem PASSED    [  2%]
xarray/tests/test_backends.py::TestRasterio::test_serialization SKIPPED  [  3%]
xarray/tests/test_backends.py::TestRasterio::test_utm SKIPPED            [  4%]
xarray/tests/test_backends.py::TestRasterio::test_non_rectilinear SKIPPED [  4%]
xarray/tests/test_backends.py::TestRasterio::test_platecarree SKIPPED    [  4%]
xarray/tests/test_backends.py::TestRasterio::test_notransform SKIPPED    [  4%]
xarray/tests/test_backends.py::TestRasterio::test_indexing SKIPPED       [  4%]
xarray/tests/test_backends.py::TestRasterio::test_caching SKIPPED        [  4%]
xarray/tests/test_backends.py::TestRasterio::test_chunks SKIPPED         [  4%]
xarray/tests/test_backends.py::TestRasterio::test_pickle_rasterio SKIPPED [  4%]
xarray/tests/test_backends.py::TestRasterio::test_ENVI_tags SKIPPED      [  4%]
xarray/tests/test_backends.py::TestRasterio::test_no_mftime SKIPPED      [  4%]
xarray/tests/test_backends.py::TestRasterio::test_http_url SKIPPED       [  4%]
xarray/tests/test_backends.py::TestEncodingInvalid::test_extract_nc4_variable_encoding PASSED [  4%]
xarray/tests/test_backends.py::TestEncodingInvalid::test_extract_h5nc_encoding PASSED [  4%]
xarray/tests/test_backends.py::TestValidateAttrs::test_validating_attrs PASSED [  4%]
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_dataarray_to_netcdf_no_name PASSED [  4%]
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_dataarray_to_netcdf_with_name PASSED [  4%]
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_dataarray_to_netcdf_coord_name_clash PASSED [  4%]
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_open_dataarray_options PASSED [  4%]
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_dataarray_to_netcdf_return_bytes PASSED [  4%]
xarray/tests/test_backends.py::TestDataArrayToNetCDF::test_dataarray_to_netcdf_no_name_pathlib PASSED [  4%]
...
```
How do you make *pytest* run test classes that do not start with `Test`?

xarray/backends/cfgrib_.py
alexamici: Now that I checked, I can say that *cfgrib* supports `indexing.IndexingSupport.OUTER`. I don't remember how I ended up declaring OUTER_1VECTOR, and since it sounded wrong, in the last commit intended to play it as safe as possible.

Code updated.

xarray/backends/cfgrib_.py
shoyer: Does `OnDiskArray` store a reference to an open file?

Actually, if the object returned by `cfgrib.open_file` does not include a file object and is already pickleable, then there's no reason why we need to use `CachingFileManager` at all.

shoyer:
@alexamici oops, we accidentally disabled most of our backend unit tests -- see https://github.com/pydata/xarray/pull/2479 for the fix.

xarray/backends/cfgrib_.py
alexamici: In fact it doesn't. We `open` the file on every uncached data access. We didn't benchmark this yet so we may end up changing the implementation in the future if we discover that it improve performance. Do you want me to remove the use of `CachingFileManager`?

alexamici:
@shoyer I added the `TestCfGrib` class with basic read tests and a test GRIB file.

Questions:
 1. I failed to reach 100% coverage becasue I didn't find the way to test `CfGribDataStore.get_dimensions`. Any hint?
 2. Can /should I test `dask` and `dask.distributed` support in the unit tests? How? Any pointers?

It looks like we are very close, what do you think? Shall I move to the documentation?

xarray/backends/cfgrib_.py
alexamici: If I'll need it, I'll add an option to keep it open and maintain backward compatibility. No need to be future-proof.

alexamici:
@shoyer I'm ready to integrate more feedback (especially on the documentation), but I removed the `WIP:` prefix as I'd consider the PR good to go as soon as you like it.

Do you usually keep the history of PRs as it is or do you prefer me to rebase?

shoyer:
> Do you usually keep the history of PRs as it is or do you prefer me to rebase?

We squash commits upon merging, so please leave things as is :)

xarray/backends/cfgrib_.py
```diff
@@ -0,0 +1,97 @@
+#
+# Copyright 2017-2018 European Centre for Medium-Range Weather Forecasts.
```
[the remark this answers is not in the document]
alexamici: I'm not sure how to handle this one. ECMWF is quite sensitive to license and IPR matters so I added the licence boilerplate to all *cfgrib* files and I simply copied the existing backend code as part of the PR.

I'm the material author of the code and my name that will appear in the contributors, but I've been working fully funded by ECMWF as an external contractor, so it looks like proper attribution would be lost in this case.

I need to ask @StephanSiemen if they object to removing the copyright notice or what else they propose. To my knowledge this is the first contribution to an external Open Source project funded by ECMWF this way and we are learning how to handle these kind of details as we go.

alexamici:
@shoyer I added a test for dask.distributed and it passes but please check that it is meaningful, as I'm not completely sure to what to test.

xarray/backends/cfgrib_.py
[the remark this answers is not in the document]
shoyer: It says "Copyright 2014-2018, xarray Developers" in our README.rst file, which was modeled off projects like NumPy: https://github.com/numpy/numpy/blob/master/LICENSE.txt

I see now that our LICENSE file just has the original Apache license text. Perhaps we should add in the more specific "Copyright xarray developers" line, like a project like TensorFlow: https://github.com/tensorflow/tensorflow/blob/master/LICENSE

xarray/tests/test_distributed.py
```diff
@@ -142,6 +142,20 @@ def test_dask_distributed_rasterio_integration_test(loop):
                 assert_allclose(actual, expected)
 
 
+@requires_cfgrib
+def test_dask_distributed_cfgrib_integration_test(loop):
+    with cluster() as (s, [a, b]):
+        with Client(s['address'], loop=loop) as c:
+            with open_example_dataset('example.grib',
+                                      engine='cfgrib',
+                                      chunks={'time': 1}) as ds:
```
shoyer: This looks reasonable to me. I assume `example.grib` has multiple time points in it?

xarray/backends/cfgrib_.py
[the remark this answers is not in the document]
jhamman: @alexamici (and ECMWF by proxy I suppose) - first, I want to make sure you understand that I'm very encouraged by your recent developments of cfgrib as an open source package and your contributions to xarray. They are much appreciated and they stand to benefit a wide swath of the geoscience community. 

That said, at this point, I'm somewhat against adding this copyright/license header for these reasons:

- Xarray already has a clear copyright statement that is permissive enough to allow you (and ECMWF) to maintain copyrights over your contributions.
- We don't do this anywhere else in the xarray code base. Though many of the contributors that have developed xarray have done so as employees/contractors of various organizations, we've not adopted this level of documentation with regard to the original author or subsequent editors.

Now, I understand that it can be important for organizations to make visible their open source contributions so we want to make sure this sort of engagement can continue to happen. We've recently joined NUMFOCUS, in part to give us access to some proper legal advice when necessary. We can certainly solicit their advice here if we have technical questions.

It's probably worth pinging the rest of @pydata/xarray to get their thoughts here.

xarray/tests/test_distributed.py
alexamici: Yes, three.

xarray/backends/cfgrib_.py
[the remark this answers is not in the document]
StephanSiemen: @alexamici, @shoyer, @jhamman,
We at ECMWF are very happy for the copyright to be adjusted according to other contributions in xarray. Any acknowledgement of the contribution is much appreciated. Thanks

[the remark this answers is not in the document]
alexamici: Looks like I was being overzealous :)

I removed the copyright notice and licence boilerplate.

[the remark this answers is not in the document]
jhamman: Thanks @alexamici and @StephanSiemen!

alexamici:
@shoyer the latest AppVeyor failure is a time limit exceeded, not a real problem.

doc/io.rst
```diff
@@ -641,9 +641,13 @@ GRIB format via cfgrib
 ----------------------
 
 xarray supports reading GRIB files via ECMWF cfgrib_ python driver and ecCodes_
-C-library, if they are installed. To open GRIB file supply ``engine='cfgrib'``
+C-library, if they are installed. To open a GRIB file supply ``engine='cfgrib'``
 to :py:func:`~xarray.open_dataset`:
 
+.. ipython:: python
+
+    ds_grib = xr.open_dataset('example.grib', engine='cfgrib')
```
shoyer: If `example.grib` doesn't actually exist when we build the docs, this will give a nasty error message. It would be better to use a `:verbatim:` directive here -- scroll open to the opendap examples to see what that looks like.

alexamici: Ouch! I noticed that there was no docs testing in Travis-CI and wondered if you where testing the docs at all. I should have built the docs myself. What is the intended way do you build the docs? `python sertup.py build_sphinx` fails on my setup due to missing `numpydoc`. 

I'm fixing it as you suggest.

alexamici: I added the `:verbatim:` directive, but I was not able to test the build of the documentation due to several errors when building the gallery (a number of core dump in GEOS, maybe due to the setup on my MacOS).

The fix looks trivial enough that it may work.

shoyer: I think this would work:
```
.. ipython::
    :verbatim:

    In [3]: ds_grib = xr.open_dataset('example.grib', engine='cfgrib')
 ```

We actually do test the doc build on Travis-CI, but unfortunately there's no easy way to see generated docs and also errors in `ipython` directive blocks don't stop the build (there's a bug we filed about this somewhere).

alexamici: Fixed. I mis-read the sample.

shoyer:
Thanks @alexamici and ECMWF!

iainrussell:
Congratulations!

PR #2477, opened 2018-10-09
Inhouse LooseVersion

### LINKED ISSUE #2468: LooseVersion check on xarray in tests seems unstable
There's an elegant check against the xarray version to decide whether to run a test, so the test 'comes online' at 0.12: https://github.com/pydata/xarray/blob/638b251c622359b665208276a2cb23b0fbc5141b/xarray/tests/test_dataarray.py#L2029

But unfortunately, this seems very unstable in tests, because without a release, `LooseVersion` can't interpret the strings correctly (e.g.  `LooseVersion ('0.10.9+29.g33d9391a')`)
- A lot of the time it raises: https://travis-ci.org/max-sixty/xarray/jobs/437913418#L1036
- Occasionally it runs the test, failing to return that we're prior to 0.12: https://travis-ci.org/max-sixty/xarray/jobs/437914645#L5036

Here's the bug in the python issue tracker: https://bugs.python.org/issue14894

Is that synopsis correct? Should we attempt to take another approach? I'll disable it in my current check so tests can pass, but lmk thoughts.

fujiisoup:
- [x] Closes #2468
 - [x] Tests added
 - [N.A.] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

A fix for #2468.

### DIFF
--- a/xarray/tests/__init__.py
+++ b/xarray/tests/__init__.py
@@ -3,7 +3,7 @@ from __future__ import division
 from __future__ import print_function
 import warnings
 from contextlib import contextmanager
-from distutils.version import LooseVersion
+from distutils import version
 import re
 import importlib
 
@@ -51,6 +51,13 @@ def _importorskip(modname, minversion=None):
         has = False
     func = pytest.mark.skipif(not has, reason='requires {}'.format(modname))
     return has, func
+
+
+class LooseVersion(version.LooseVersion):
+    def __init__(self, vstring=None):
+        if '+' in vstring:
+            vstring = vstring[:vstring.find('+')]
+        super(LooseVersion, self).__init__(vstring)
 
 
 has_matplotlib, requires_matplotlib = _importorskip('matplotlib')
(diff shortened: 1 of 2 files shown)


xarray/tests/__init__.py
```diff
@@ -53,6 +53,13 @@ def _importorskip(modname, minversion=None):
     return has, func
 
 
+class LooseVersion(version.LooseVersion):
+    def __init__(self, vstring=None):
+        if '+' in vstring:
+            vstring = vstring[:vstring.find('+')]
+        super(LooseVersion, self).__init__(vstring)
```
shoyer: Rather than a subclass, can we just make an alternate constructor? I think that would be less error prone, e.g.,
```python
def LooseVersion(vstring):
    if '+' in vstring:
        vstring = vstring[:vstring.find('+')]
    return distutils.LooseVersion(vstring)
```

max-sixty: (if you're in the market for much more refine-y comments, you could use `vstring.split('+')[0]`, which works whether or not there's a `+`)

fujiisoup: Thanks. Updated

shoyer:
Thanks!

PR #2479, opened 2018-10-11
Fix backend test classes so they actually run

shoyer:
We accidentally stopped running many backend tests when we merged #2467,
because they no longer inherited from untitest.TestCase and didn't have a
name starting with "Test".

This PR renames the tests so they actually run. I also fixed a bug with the
pseudonetcdf backend that appears to have been introduced by a bad merge in
#2261. It wasn't caught because the tests weren't actually running.

cc @max-sixty


max-sixty:
Thanks @shoyer !

Issue #2473, opened 2018-10-08
Recommended way to extend xarray Datasets using accessors?


TomNicholas:
Hi,

I'm now regularly using xarray (& dask) for organising and analysing the output of the simulation code I use ([BOUT++](https://boutproject.github.io/)) and it's very helpful, thank you!. 

However my current approach is quite clunky at dealing the extra information and functionality that's specific to the simulation code I'm using, and I have questions about what the recommended way to extend the xarray Dataset class is. This seems like a general enough problem that I thought I would make an issue for it.

### Desired

What I ideally want to do is extend the xarray.Dataset class to accommodate extra attributes and methods, while retaining as much xarray functionality as possible, but avoiding reimplementing any of the API. This might not be possible, but ideally I want to make a `BoutDataset` class which contains extra attributes to hold information about the run which doesn't naturally fit into the xarray data model, extra methods to perform analysis/plotting which only users of this code would require, but also be able to use xarray-specific methods and top-level functions:

```python
bd = BoutDataset('/path/to/data')

ds = bd.data  # access the wrapped xarray dataset
extra_data = bd.extra_data  # access the BOUT-specific data

bd.isel(time=-1)  # use xarray dataset methods

bd2 = BoutDataset('/path/to/other/data')
concatenated_bd = xr.concat([bd, bd2])  # apply top-level xarray functions to the data

bd.plot_tokamak()  # methods implementing bout-specific functionality
```

### Problems with my current approach

I have read the documentation about [extending xarray](http://xarray.pydata.org/en/stable/internals.html#extending-xarray), and the issue threads about subclassing Datasets (#706) and accessors (#1080), but I wanted to check that what I'm doing is the recommended approach.

Right now I'm [trying](https://github.com/TomNicholas/xcollect/blob/master/boutdataset.py) to do something like

```python
@xr.register_dataset_accessor('bout')
class BoutDataset:
    def __init__(self, path):
        self.data = collect_data(path)  # collect all my numerical data from output files
        self.extra_data = read_extra_data(path)  # collect extra data about the simulation 

    def plot_tokamak():
        plot_in_bout_specific_way(self.data, self.extra_data)
``` 

which works in the sense that I can do

```python
bd = BoutDataset('/path/to/data')

ds = bd.bout.data  # access the wrapped xarray dataset
extra_data = bd.bout.extra_data  # access the BOUT-specific data
bd.bout.plot_tokamak()  # methods implementing bout-specific functionality
```

but not so well with

```python
[...]

Issue #2288, opened 2018-07-15
Add CRS/projection information to xarray objects


djhoese:
#### Problem description

This issue is to start the discussion for a feature that would be helpful to a lot of people. It may not necessarily be best to put it in xarray, but let's figure that out. I'll try to describe things below to the best of my knowledge. I'm typically thinking of raster/image data when it comes to this stuff, but it could probably be used for GIS-like point data.

Geographic data can be projected (uniform grid) or unprojected (nonuniform). Unprojected data typically has longitude and latitude values specified per-pixel. I don't think I've ever seen non-uniform data in a projected space. Projected data can be specified by a CRS (PROJ.4), a number of pixels (shape), and extents/bbox in CRS units (xmin, ymin, xmax, ymax). This could also be specified in different ways like origin (X, Y) and pixel size. Seeing as xarray already computes all `coords` data it makes sense for extents and array shape to be used. With this information provided in an xarray object any library could check for these properties and know where to place the data on a map.

So the question is: Should these properties be standardized in xarray Dataset/DataArray objects and how?

#### Related libraries and developers

* pyresample (me, @mraspaud, @pnuu)
* verde and gmt-python (@leouieda)
* metpy (@dopplershift)
* [geo-xarray](https://github.com/andrewdhicks/geo-xarray/wiki) (@andrewdhicks)
* rasterio
* cartopy

I know @WeatherGod also showed interest on gitter.

#### Complications and things to consider

1. Other related coordinate systems like [ECEF](https://en.wikipedia.org/wiki/ECEF) where coordinates are specified in three dimensions (X, Y, Z). Very useful for calculations like nearest neighbor of lon/lat points or for comparisons between two projected coordinate systems.
2. Specifying what coords arrays are the CRS coordinates or geographic coordinates in general.
3.  If xarray should include these properties, where is the line drawn for what functionality xarray supports? Resampling/gridding, etc?
4. How is the CRS object represented? PROJ.4 string, PROJ.4 dict, existing libraries CRS object, new CRS object, `pyproj.Proj` object?
5. Affine versus geotransforms instead of extents: https://github.com/mapbox/rasterio/blob/master/docs/topics/migrating-to-v1.rst#affineaffine-vs-gdal-style-geotransforms
6. Similar to 4, I never mentioned "rotation" parameters which some users may want and are specified in the affine/geotransform.
7. Dynamically generated extents/affine objects so that slicing operations don't have to be handled specially.
[...]

Issue #1271, opened 2017-02-15
Attrs are lost in mathematical computation


fujiisoup:
Related to #138

Why is `keep_attrs` option in reduce method set to FALSE by default?
I feel it is more natural to keep all the attrs after some computation that returns xr.DaraArray
Such as `data*100`.
(By it is not possible to set this option TRUE when using an operator.)
Is it an option to move this option to __init__ method, in case of TRUE all the attrs are tracked after computations of the object and also the object generated from this object?

shoyer:
Historically, we dropped attributes in arithmetic because attributes are often used for units and we didn't want to do computation that results in the wrong units. Blinding propagating metadata can lead to it ending up in the wrong place.

Also, if you are combining multiple DataArray objects with different `attrs`, there are a number of options for combining them, and it wasn't obvious which is the right strategy.

But in practice, this is something that a lot of people want (better to have stale metadata than no metadata at all), so maybe I made the wrong choice here. At the very least, I would be comfortable with an [option](http://xarray.pydata.org/en/stable/generated/xarray.set_options.html) to set `keep_attrs=True` for every operation and/or to specify a merge strategy for combining `attrs` in arithmetic.

fujiisoup:
Thank you for the information.
I agree that we should not strongly rely on attrs. Unit may change in arithmetic.

My sense is closest to the option 3 in #131.
Some attrs should be tracked and other should be dropped.
In the present stage, the most possible option is to add `keep_attrs` in `xr.set_options`?
Or any destructive change will be an option?
Such as to divide attrs into two kinds, `units` and `metadata`.

fmaussion:
> divide attrs into two kinds, units and metadata

Coordinate reference systems (crs) are an example of attribute which is always valid after selection, arithmetic, or even reduction since they apply to the coordinates of the DataArray. A global option to keep certain attrs (like ``crs``) would be very useful to downstream libraries like salem.

dopplershift:
Could the setting not be module-wide, but be set on `DataArray` instances themselves? That setting could then be inherited when new ones are created from arithmetic operations.

I'd like to have the option of specifying specific attributes that should be kept, or maybe dropped. It will be *really* annoying to need to keep copying the `grid_mapping` attribute.

shoyer:
> Coordinate reference systems (crs) are an example of attribute which is always valid after selection, arithmetic, or even reduction since they apply to the coordinates of the DataArray.

[...]

PR #2482, opened 2018-10-12
Global option to always keep/discard attrs on operations

TomNicholas:
- [x] Resolves wishes of some users and relevant for discussion in #138, #442, #688, #828, #988, #1009, #1271, #2288, #2473
 - [x] Tests added, both of setting the option and of attributes propagating in the expected way 
 - [x] Tests passed
 - [x] Documented

Adds a global option to either always keep or always discard attrs in method and function calls. 

The behaviour is backwards-compatible, as the logic is:
- if `keep_attrs` supplied as keyword argument then use that
- else if global option (`xarray.set_options(keep_attrs=True)`) is set then use that
- else use default value of `keep_attrs` argument for that particular function/method (kept the same as they were for backwards-compatibility).

Main use cases include users who want to store the units of their data in the attrs, users who want to always keep information about the source or history of their data, and users who want to store objects in their attributes which are needed to supplement the xarray objects (e.g. an `xgcm.grid`). It should eventually be superceded by hooks for custom attribute handling (#988), but will be useful until then. 

I have left the top-level functions like `concat` and `merge` alone. Currently concat keeps the attributes of the first object passed to it, and merge returns a dataset with no attributes. It's not clear how this should be treated though, so I left it to users to extend those functions if they need to.


xarray/core/common.py
```diff
@@ -21,12 +22,12 @@ class ImplementsArrayReduce(object):
     def _reduce_method(cls, func, include_skipna, numeric_only):
         if include_skipna:
             def wrapped_func(self, dim=None, axis=None, skipna=None,
-                             keep_attrs=False, **kwargs):
+                             keep_attrs=_set_keep_attrs(False), **kwargs):
```
shoyer: This won't work properly, because default arguments are evaluated when functions are defined instead of when the function is called.

Instead, these should probably be written something like this:
```python
def func(..., keep_attrs=None):
    if keep_attrs is None:
```

TomNicholas: You're right. Updated to be as you suggested everywhere, and tests added to check that the behaviour is as expected.

xarray/core/common.py
```diff
@@ -22,12 +22,16 @@ class ImplementsArrayReduce(object):
     def _reduce_method(cls, func, include_skipna, numeric_only):
         if include_skipna:
             def wrapped_func(self, dim=None, axis=None, skipna=None,
-                             keep_attrs=_set_keep_attrs(False), **kwargs):
+                             keep_attrs=None, **kwargs):
+                if keep_attrs is None:
+                    keep_attrs = _get_keep_attrs(default=False)
```
shoyer: Note: you probably need far fewer of these checks -- only functions that actually implement the `keep_attrs` logic need it, not functions that pass on `keep_attrs` to other functions with the same default value.

xarray/tests/test_options.py
```diff
@@ -64,3 +66,105 @@ def test_nested_options():
...
+
+
+def create_test_dataarray_attrs(seed=0, var='var1'):
+    da = create_test_data(seed)[var]
+    da.attrs = {'attr1': 5, 'attr2': 'history',
+                'attr3': {'nested': 'more_info'}}
+    return da
+
+
+class TestAttrRetention:
```
shoyer: please inherit from `object` (for PYthon 2 compat)

shoyer:
Unit tests and docs look good.

TomNicholas:
I wanted to clean up my commits using git rebase, but when I tried `git rebase -i` I only had the choice to alter my last 5 commits to this branch, not all 8. I'm not sure if you know what I'm doing wrong there?

shoyer:
@TomNicholas I'm not sure, but don't worry about cleaning up git history with rebase. We squash all commits before merging anyways, and GitHub's review tools work a little better when you don't rebase.

xarray/core/common.py
TomNicholas: I think I know what you mean, but in this case if I remove that check and just pass `keep_attrs=None` down into the `self.reduce` method, then some unit tests will fail (specifically `TestDataArray.test_reduce_keep_attrs`, `TestDataset.test_apply`, `TestVariable.test_reduce_keep_attrs`).

spencerkclark:
@TomNicholas it looks like the merge conflict in `whats-new.rst` is preventing the CI from running.  I think if you sync with the latest master branch and resolve the conflict things should work properly.

spencerkclark:
@TomNicholas thanks, things look better now.  Don't worry about the Appveyor failures; a new version of `cftime` (installed via `pip` on the Python 2 builds there) was just released and there is a bug they are still ironing out on a few platforms (including Windows): https://github.com/conda-forge/cftime-feedstock/pull/10, https://github.com/Unidata/cftime/pull/76.

doc/whats-new.rst
```diff
@@ -79,6 +79,13 @@ Enhancements
...
   :py:meth:`~xarray.DataArray.interp`, and
   :py:meth:`~xarray.Dataset.interp`.
+  By `Spencer Clark <https://github.com/spencerkclark>`_
+- There is now a global option to either always keep or always discard
+  dataset and dataarray attrs upon operations. The option is set with
+  ``xarray.set_options(keep_attrs=True)``, and the default is to use the old
+  behaviour.
+  By `Tom Nicholas <http://github.com/TomNicholas>`_.
+  
   By `Spencer Clark <https://github.com/spencerkclark>`_.
```
dcherian: this line needs to be removed.

xarray/core/options.py
```diff
@@ -41,6 +45,17 @@ def _set_file_cache_maxsize(value):
...
 
+def _get_keep_attrs(default):
+    global_choice = OPTIONS['keep_attrs']
+
+    if global_choice is 'default':
+        return default
+    elif global_choice in [True, False]:
+        return global_choice
+    else:
+        raise ValueError('The global option keep_attrs is set to an invalid value.')
```
dcherian: maybe say "The global option keep_attrs must be one of True, False or 'default'."?

xarray/core/options.py
```diff
@@ -63,6 +78,11 @@ class set_options(object):
     - ``cmap_divergent``: colormap to use for divergent data plots.
       Default: ``RdBu_r``. If string, must be matplotlib built-in colormap.
       Can also be a Colormap object (e.g. mpl.cm.magma)
+    - ``keep_attrs``: rule for whether to keep attributes on xarray
+      Datasets/dataarrays after operations. Either ``True`` to always keep
+      attrs, ``False`` to always discard them, or ``'default'`` to use original
+      logic that attrs should only be kept in unambiguous circumstances.
+      Default: ``'default'``.
 
 f    You can use ``set_options`` either as a context manager:
```
dcherian: This 'f' is a typo in current master. Can you remove it please?

xarray/core/common.py
dcherian: This should be fixed. 

For e.g. the `DataArray` methods below usually end up calling `Variable` or `Dataset` methods so you should only have to change the default to None in `dataarray.py`

TomNicholas: Ah I got it - changed in most recent commit

xarray/core/dataarray.py
```diff
@@ -1588,6 +1588,10 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=False, **kwargs):
             DataArray with this object's array replaced with an array with
             summarized data and the indicated dimension(s) removed.
         """
+
+        if keep_attrs is None:
```
dcherian: This should not be necessary since `Variable.reduce` does the same check.

xarray/core/dataarray.py
```diff
@@ -2312,11 +2316,14 @@ def quantile(self, q, dim=None, interpolation='linear', keep_attrs=False):
         numpy.nanpercentile, pandas.Series.quantile, Dataset.quantile
         """
 
+        if keep_attrs is None:
+            keep_attrs = _get_keep_attrs(default=False)
+
```
dcherian: The next two should not be necessary since `Dataset.quantile()` and `Dataset.rank()` have the same statement.

xarray/core/dataset.py
```diff
@@ -2922,7 +2925,7 @@ def apply(self, func, keep_attrs=False, args=(), **kwargs):
             transform each DataArray `x` in this dataset into another
             DataArray.
         keep_attrs : bool, optional
-            If True, the dataset's attributes (`attrs`) will be copied from
+            If True, the dataset's attributes (`attrs`) will be _getcopied from
```
dcherian: _getcopied?

xarray/core/groupby.py
```diff
@@ -590,12 +596,16 @@ def reduce_array(ar):
     def _reduce_method(cls, func, include_skipna, numeric_only):
         if include_skipna:
             def wrapped_func(self, dim=DEFAULT_DIMS, axis=None, skipna=None,
-                             keep_attrs=False, **kwargs):
+                             keep_attrs=None, **kwargs):
+                if keep_attrs is None:
```
dcherian: similarly here; doesn't `self.reduce` handle this?

shoyer:
Thanks @TomNicholas !

Issue #2300, opened 2018-07-18
zarr and xarray chunking compatibility and `to_zarr` performance


chrisbarber:
I have a situation where I build large zarr arrays based on chunks which correspond to how I am reading data off a filesystem, for best I/O performance. Then I set these as variables on an xarray dataset which I want to persist to zarr, but with different chunks more optimal for querying.

One problem I ran into is that manually selecting chunks of a dataset prior to `to_zarr` results in https://github.com/pydata/xarray/blob/66be9c5db7d86ea385c3a4cd4295bfce67e3f25b/xarray/backends/zarr.py#L83

It's difficult for me to understand exactly how to select chunks manually at the dataset level which would also make this zarr "final chunk" constraint happy. I would have been satisfied however with letting zarr [choose](https://github.com/zarr-developers/zarr/blob/07fa81a5afe906fcbafc5f81a851c438493ffc13/zarr/util.py#L49) chunks for me, but could not find a way to trigger this through the xarray API short of "unchunking" it first, which would lead to loading entire variables into memory. I came up with the following hack to trigger zarr's automatic chunking despite having differently defined chunks on my xarray dataset:

```python
    # monkey patch to get zarr to ignore dask chunks and use its own heuristics
    def copy_func(f):
        g = types.FunctionType(f.__code__, f.__globals__, name=f.__name__,
                               argdefs=f.__defaults__,
                               closure=f.__closure__)
        g = functools.update_wrapper(g, f)
        g.__kwdefaults__ = f.__kwdefaults__
        return g
    orig_determine_zarr_chunks = copy_func(xr.backends.zarr._determine_zarr_chunks)
    xr.backends.zarr._determine_zarr_chunks = lambda enc_chunks, var_chunks, ndim: orig_determine_zarr_chunks(enc_chunks, None, ndim)
```
The next problem to contend with is that `da.store` between zarr stores with differing chunks between source and destination is astronomically slow. The first thing to attempt would be to rechunk the dask arrays according to the destination zarr chunks, but xarray's [consistent chunks constraint](https://github.com/pydata/xarray/blob/59ad782f29a0f4766bac7802be6650be61f018b8/xarray/core/dataset.py#L1239) blocks this strategy as far as I can tell. Once again I took the dirty hack approach and inject a rechunking on a per-variable basis during the `to_zarr` operation, as follows:

```python
    # monkey patch to make dask arrays writable with different chunks than zarr dest
[...]

PR #2487, opened 2018-10-14
Zarr chunking (GH2300)

lilyminium:
- [x] Band-aid for #2300
 - [x] Test added to check that zarr-originating array can save
 - [x] Updated whats-new

I don't fully understand the ins-and-outs of Zarr, but it seems that if it can be serialised with a smaller end-chunk to begin with, then saving a Dataset constructed from Zarr should not have an issue with that either.

### DIFF
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -80,13 +80,13 @@ def _determine_zarr_chunks(enc_chunks, var_chunks, ndim):
         if any(len(set(chunks[:-1])) > 1 for chunks in var_chunks):
             raise ValueError(
                 "Zarr requires uniform chunk sizes except for final chunk."
-                " Variable %r has incompatible chunks. Consider "
+                " Variable dask chunks %r are incompatible. Consider "
                 "rechunking using `chunk()`." % (var_chunks,))
         if any((chunks[0] < chunks[-1]) for chunks in var_chunks):
             raise ValueError(
-                "Final chunk of Zarr array must be smaller than first. "
-                "Variable %r has incompatible chunks. Consider rechunking "
-                "using `chunk()`." % var_chunks)
+                "Final chunk of Zarr array must be the same size or smaller "
+                "than the first. Variable Dask chunks %r are incompatible. "
+                "Consider rechunking using `chunk()`." % var_chunks)
         # return the first chunk for each dimension
         return tuple(chunk[0] for chunk in var_chunks)
 
@@ -136,9 +136,11 @@ def _determine_zarr_chunks(enc_chunks, var_chunks, ndim):
                         % (enc_chunks_tuple, var_chunks))
             if dchunks[-1] > zchunk:
                 raise ValueError(
-                    "Final chunk of Zarr array must be smaller than first. "
-                    "Variable %r has incompatible chunks. Consider rechunking "
-                    "using `chunk()`." % var_chunks)
+                    "Final chunk of Zarr array must be the same size or smaller "
+                    "than the first. The specified Zarr chunk encoding is %r, "
+                    "but %r in variable Dask chunks %r is incompatible. "
+                    "Consider rechunking using `chunk()`." 
+                    % (enc_chunks_tuple, dchunks, var_chunks))
         return enc_chunks_tuple
 
     raise AssertionError(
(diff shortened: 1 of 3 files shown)


jhamman:
This looks fine to me. I actually thought we had fixed this already but I guess not. Can you add whatsnew entry in the docs as well?

xarray/backends/zarr.py
```diff
@@ -126,14 +126,19 @@ def _determine_zarr_chunks(enc_chunks, var_chunks, ndim):
...
                 if dchunk % zchunk:
                     raise NotImplementedError(
                         "Specified zarr chunks %r would overlap multiple dask "
                         "chunks %r. This is not implemented in xarray yet. "
                         " Consider rechunking the data using "
                         "`chunk()` or specifying different chunks in encoding."
                         % (enc_chunks_tuple, var_chunks))
+            if dchunks[-1] > zchunk:
+                raise ValueError(
+                    "Final chunk of Zarr array must be smaller than first. "
```
rabernat: "Final chunk of zarr array must be the same size or smaller than first."

xarray/backends/zarr.py
```diff
@@ -126,14 +126,19 @@ def _determine_zarr_chunks(enc_chunks, var_chunks, ndim):
...
+                    "Variable %r has incompatible chunks. Consider rechunking "
+                    "using `chunk()`." % var_chunks)
```
rabernat: In this error, `%r` is referring to `var_chunks`, not the variable name. I would output both `enc_chunks_tuple` and `var_chunks`, as this information is useful for debugging.

xarray/tests/test_backends.py
```diff
@@ -1389,6 +1389,10 @@ def test_chunk_encoding_with_dask(self):
         ds_chunk_irreg = ds.chunk({'x': (5, 5, 2)})
         with self.roundtrip(ds_chunk_irreg) as actual:
             assert (5,) == actual['var1'].encoding['chunks']
+        # re-save Zarr arrays
+        with self.roundtrip(ds_chunk_irreg) as original:
+            with self.roundtrip(original) as actual:
+                assert_identical(original, actual)
```
rabernat: Does this test cover #2300?

Lilyminium: I don't think it implements the NotImplementedError scenario, but the bug for my situation was that opening a dataset saved with chunks using  `to_zarr` and saving it unchanged with `to_zarr` would raise this error. The issue was that `enc_chunks_tuple` represents each chunk with an integer, while `var_chunks` represents it with a tuple. So this was the error raised in #2300 for example:

> NotImplementedError: Specified zarr chunks (63170,) would overlap multiple dask chunks ((63170, 63170, 63170, 63170, 63170, 63170, 63170, 63169),). This is not implemented in xarray yet.  Consider rechunking the data using `chunk()` or specifying different chunks in encoding.

Rechunking it, as suggested, would yield the same error. Changing the logic to allow the last element of the tuple to be less than the `zchunk` integer avoids this scenario. The test is just checking that a second roundtrip does not raise an error; it fails without this patch.

rabernat: Makes sense. Zarr's data model only allows a single integer value chunk size along each axis, while dask can have chunks of different length. 

Here is what the [zarr spec](https://zarr.readthedocs.io/en/stable/spec/v2.html#chunks) says about this topic:
>  Note that all chunks in an array have the same shape. If the length of any array dimension is not exactly divisible by the length of the corresponding chunk dimension then some chunks will overhang the edge of the array. The contents of any chunk region falling outside the array are undefined.

lilyminium: That makes sense and padding the array could be another solution -- but then why does saving the dataset with irregular arrays work the first time? I haven't poked around in Zarr's internals so I assume the difference lies in the chunks argument to  `zarr_group.create()` ?

xarray/backends/zarr.py
rabernat: @lilyminium - if you can address this comment, we can get your PR merged.

xarray/tests/test_backends.py
rabernat: I am ok with this.

xarray/backends/zarr.py
rabernat: @lilyminium - if you can address this comment, we can get your PR merged.

xarray/backends/zarr.py
lilyminium: Sorry, busy time of year -- just updated now.

rabernat:
Thanks for this contribution!

> Can you add whatsnew entry in the docs as well?

We are still missing an entry in whatsnew. Please add it to give yourself credit for your contribution! :)

shoyer:
Thanks!

PR #2491, opened 2018-10-17
Dataset.copy docstring matches behavior

### LINKED ISSUE #2490: Dataset.copy docstring has a typo
The docstring of [Dataset.copy](https://github.com/pydata/xarray/blob/9f4474d657193f1c7c9aac25bb2edf94755a8593/xarray/core/dataset.py#L715-L730) states that the default of deep is True, but in the code it is False.

georgegeddes:
- [ ] Closes #2490

This makes the docstring of Dataset.copy match the actual behavior. 

However, I noticed that DataArray.copy has the opposite behavior for the deep keyword. Is this intentional, or are they supposed to behave the same way?

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -727,7 +727,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         ----------
         deep : bool, optional
             Whether each component variable is loaded into memory and copied onto
-            the new object. Default is True.
+            the new object. Default is False.
         data : dict-like, optional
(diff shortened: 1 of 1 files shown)


shoyer:
thanks for catching this!

I'm trying to remember if there is any good reason for these methods having different defaults....

I suppose it could make sense if `Dataset.copy()` is modeled off `dict.copy()` (which does shallow copies) and `DataArray.copy()` is modeled off `numpy.ndarray.copy()` (which does deep copies).

dcherian:
Thanks!

PR #2502, opened 2018-10-23
Fix tests to new coordinate names in cfgrib>=0.9.2

alexamici:
Fix `TestCfGrib` broken by a user-visible backward-incompatible change in *cfgrib>=0.9.2*.

No other user visible coordinate name change is expected so the test can remain like it is.

Sorry for the inconvenience!

 - [x] Noted in the discussion of #2500
 - [x] Tests added
 - ~~Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API~~

### DIFF
--- a/ci/requirements-py36.yml
+++ b/ci/requirements-py36.yml
@@ -27,4 +27,4 @@ dependencies:
     - pytest-cov
     - pydap
     - lxml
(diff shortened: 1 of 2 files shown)


shoyer:
thanks!

PR #2506, opened 2018-10-24
Iterate over data_vars only

### LINKED ISSUE #884: Q: Should Dataset.keys() return only variable keys?

max-sixty:
- [x] Closes #884 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -930,32 +930,13 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         return key in self._variables
 
     def __len__(self):
-        warnings.warn('calling len() on an xarray.Dataset will change in '
-                      'xarray v0.11 to only include data variables, not '
-                      'coordinates. Call len() on the Dataset.variables '
-                      'property instead, like ``len(ds.variables)``, to '
-                      'preserve existing behavior in a forwards compatible '
-                      'manner.',
-                      FutureWarning, stacklevel=2)
-        return len(self._variables)
+        return len(self.data_vars)
 
     def __bool__(self):
-        warnings.warn('casting an xarray.Dataset to a boolean will change in '
-                      'xarray v0.11 to only include data variables, not '
-                      'coordinates. Cast the Dataset.variables property '
-                      'instead to preserve existing behavior in a forwards '
-                      'compatible manner.',
-                      FutureWarning, stacklevel=2)
-        return bool(self._variables)
+        return bool(self.data_vars)
 
     def __iter__(self):
-        warnings.warn('iteration over an xarray.Dataset will change in xarray '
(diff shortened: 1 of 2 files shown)


max-sixty:
I'm seeing (unrelated) errors on AppVeyor only: https://ci.appveyor.com/project/shoyer/xray/builds/19763584/job/44ushvsfpuw77rqd

```
Collecting netcdf4 (from -r C:\projects\xray\ci\condaenv.z95tw7.requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/eb/aa/b067f3b1a2561f29f5c282d8a0f0f4bba5b13e9bdaa5fcd29005d226c448/netCDF4-1.4.2.tar.gz (769kB)
    Complete output from command python setup.py egg_info:
    reading from setup.cfg...
    
        HDF5_DIR environment variable not set, checking some standard locations ..
    checking C:\Users\appveyor ...
    checking /usr/local ...
    checking /sw ...
    checking /opt ...
    checking /opt/local ...
    checking /usr ...
      File "<string>", line 1, in <module>
      File "c:\users\appveyor\appdata\local\temp\1\pip-install-nfiw_q\netcdf4\setup.py", line 371, in <module>
        _populate_hdf5_info(dirstosearch, inc_dirs, libs, lib_dirs)
      File "c:\users\appveyor\appdata\local\temp\1\pip-install-nfiw_q\netcdf4\setup.py", line 319, in _populate_hdf5_info
        raise ValueError('did not find HDF5 headers')
    ValueError: did not find HDF5 headers
    
    ----------------------------------------
```

PR #2509, opened 2018-10-25
Remove .T as shortcut for transpose()

### LINKED ISSUE #1232: Remove the `Dataset.T` property?

dcherian:
- [x] Closes #1232

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -33,6 +33,9 @@ v0.11.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- ``.T`` has been removed as a shortcut for :py:meth:`transpose()`.
+  Call :py:meth:`Dataset.transpose()` or :py:meth:`DataArray.transpose()` instead.
+
 - Xarray's storage backends now automatically open and close files when
   necessary, rather than requiring opening a file with ``autoclose=True``. A
   global least-recently-used cache is used to store open files; the default
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -34,7 +34,7 @@ Breaking changes
 ~~~~~~~~~~~~~~~~
 
 - ``Dataset.T`` has been removed as a shortcut for :py:meth:`Dataset.transpose`.
-  Call :py:meth:`Dataset.transpose()` instead.
+  Call :py:meth:`Dataset.transpose` directly instead.
 - Iterating over a ``Dataset`` now includes only data variables, not coordinates.
   Similarily, calling ``len`` and ``bool`` on a ``Dataset`` now  
   includes only data variables
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -33,8 +33,8 @@ v0.11.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
-- ``.T`` has been removed as a shortcut for :py:meth:`transpose()`.
-  Call :py:meth:`Dataset.transpose()` or :py:meth:`DataArray.transpose()` instead.
+- ``Dataset.T`` has been removed as a shortcut for :py:meth:`transpose()`.
+  Call :py:meth:`Dataset.transpose()` instead.
 
 - Xarray's storage backends now automatically open and close files when
   necessary, rather than requiring opening a file with ``autoclose=True``. A
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -33,7 +33,7 @@ v0.11.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
-- ``Dataset.T`` has been removed as a shortcut for :py:meth:`transpose()`.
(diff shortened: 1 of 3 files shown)


jhamman:
For posterity, what did we decide to do with the `DataArray.T` property?

dcherian:
> For posterity, what did we decide to do with the DataArray.T property?

I'm not sure but I've fixed the whats-new entry to say that only Dataset.T has been removed currently.

xarray/core/dataset.py
```diff
@@ -917,10 +917,7 @@ def _item_sources(self):
                 LevelCoordinatesSource(self)]
 
     def __dir__(self):
-        # In order to suppress a deprecation warning in Ipython autocompletion
-        # .T is explicitly removed from __dir__. GH: issue 1675
         d = super(Dataset, self).__dir__()
-        d.remove('T')
         return d
```
shoyer: You can drop this method entirely in favor of the super class method.

shoyer:
> For posterity, what did we decide to do with the `DataArray.T` property?

I think this is arguably still useful/expected since DataArray is modeled on a NumPy array. But I could probably be convinced otherwise...

dcherian:
> I think this is arguably still useful/expected since DataArray is modeled on a NumPy array.

I agree

jhamman:
Feeling like playing devils advocate here. I would expect we'll run into similar confusion with the `DataArray` transpose property. 


```Python

>>> da = xr.DataArray([1, 2, 3, 4], dims=('time', ), coords={'time': pd.date_range('2018-10-24', periods=4)})
>>> da.time

<xarray.DataArray 'time' (time: 4)>
array(['2018-10-24T00:00:00.000000000', '2018-10-25T00:00:00.000000000',
       '2018-10-26T00:00:00.000000000', '2018-10-27T00:00:00.000000000'],
      dtype='datetime64[ns]')
Coordinates:
  * time     (time) datetime64[ns] 2018-10-24 2018-10-25 2018-10-26 2018-10-27

>>> da = xr.DataArray([1, 2, 3, 4], dims=('T', ), coords={'T': pd.date_range('2018-10-24', periods=4)})
>>> da.T

<xarray.DataArray (T: 4)>
array([1, 2, 3, 4])
Coordinates:
  * T        (T) datetime64[ns] 2018-10-24 2018-10-25 2018-10-26 2018-10-27
```

max-sixty:
I'm +0.2 for merging this, but _at least_ issuing a `DeprecationWarning` before removing from `DataArray.T` (with no strong view on whether we do remove it)

jhamman:
Sure, that's a good point. I just wanted to illustrate that these are the same problems and that only removing the property from the Dataset may lead to continued confusion. @shoyer thoughts?

shoyer:
I suppose the case for removing `DataArray.T` is that you don't really need a shortcut for transposing arrays in xarray, since most functions don't rely on dimension order. And of course the real motivation: some weather/climate scientists like to use a variable named `T` for time :).

The counter-argument would be that `T` is found on most other numpy array-like types, including `pandas.DataFrame`. I've definitely used `.T` as a convenient shortcut myself in some cases.

I guess that puts me at -0 for deprecating `DataArray.T`. But we can merge this already, and leave that for a separate discussion.

max-sixty:
>  I just wanted to illustrate that these are the same problems and that only removing the property from the Dataset may lead to continued confusion. 

I think that's a good point. There are a few methods on `DataArray`s but not `Dataset`s, but not many:

```python
In [4]: {x for x in set(dir(xr.DataArray)) - set(dir(xr.Dataset)) if not x.startswith("_")}
Out[4]:
{'data',
 'dot',
 'dt',
 'dtype',
 'from_cdms2',
 'from_iris',
 'from_series',
 'get_axis_num',
 'item',
 'name',
 'ndim',
 'plot',
 'searchsorted',
 'shape',
 'size',
 'to_cdms2',
 'to_dataset',
 'to_index',
 'to_iris',
 'to_masked_array',
 'to_pandas',
 'to_series',
 'variable'}
```

Regardless, sounds like @shoyer thinks we should merge (and I'm probably still +0.1)

doc/whats-new.rst
```diff
@@ -33,6 +33,8 @@ v0.11.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- ``Dataset.T`` has been removed as a shortcut for :py:meth:`transpose()`.
```
max-sixty: Generally we don't put `()` in the docs - is this our convention or will the docs not link?

dcherian: good catch

max-sixty: It's also in the line below! 
I really don't think it matters if it's only cosmetic, though

dcherian: image

PR #2513, opened 2018-10-26
Make sure datetime object arrays are converted to datetime64

### LINKED ISSUE #2512: to_netcdf() fails because of datetime encoding
#### Simple example:

```python
import numpy as np
from datetime import datetime, timedelta
import xarray

# "time" coordinate
dt = datetime(1999, 1, 1)
dts = np.array([dt + timedelta(days=x) for x in range(10)])
coords = {'time': dts}

# simple float data
data = np.arange(10)
vrbls = {'foo': (('time',), data)}

# create the Dataset
ds = xarray.Dataset(vrbls, coords)

# encode the time coordinate
units = 'days since 1900-01-01'
ds.time.encoding['units'] = units

# write to netcdf
ds.to_netcdf('test.nc')

```
#### Problem description

When I run the above, I get the following error when executing the last line:
**`ValueError: unsupported dtype for netCDF4 variable: datetime64[ns]`**

The documentation indicates that datetime and datetime64 objects are both supported by xarray and should write to netcdf just fine when supplied "units" for encoding (this code fails with or without the encoding lines). Any Idea when is going wrong here?

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.0-8-amd64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.9
pandas: 0.20.3
numpy: 1.13.1
scipy: 0.19.1
netCDF4: 1.4.2
h5netcdf: 0.5.0
h5py: 2.8.0
Nio: None
zarr: None
cftime: 1.0.1
PseudonetCDF: None
rasterio: None
iris: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.0
distributed: 1.20.1
matplotlib: 2.1.0
cartopy: None
seaborn: 0.8.0
setuptools: 27.2.0
pip: 9.0.1
conda: 4.5.11
pytest: 3.1.3
IPython: 6.1.0
sphinx: 1.6.2
</details>

dcherian:
- [x] Closes #2512 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -95,7 +95,7 @@ def as_variable(obj, name=None):
                 'cannot set variable %r with %r-dimensional data '
                 'without explicit dimension names. Pass a tuple of '
                 '(dims, data) instead.' % (name, data.ndim))
-        obj = Variable(name, obj, fastpath=True)
+        obj = Variable(name, data, fastpath=True)
     else:
         raise TypeError('unable to convert object into a variable without an '
                         'explicit list of dimensions: %r' % obj)
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -86,6 +86,9 @@ Bug fixes
 - Addition and subtraction operators used with a CFTimeIndex now preserve the
   index's type. (:issue:`2244`).
   By `Spencer Clark <https://github.com/spencerkclark>`_.
+- We now properly handle arrays of ``datetime.datetime`` and ``datetime.timedelta``
+  provided as coordinates. (:issue:`2512`)
+  By `Deepak Cherian <https://github.com/dcherian`_.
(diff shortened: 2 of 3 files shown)


xarray/core/variable.py
```diff
@@ -95,7 +95,7 @@ def as_variable(obj, name=None):
                 'cannot set variable %r with %r-dimensional data '
                 'without explicit dimension names. Pass a tuple of '
                 '(dims, data) instead.' % (name, data.ndim))
-        obj = Variable(name, obj, fastpath=True)
+        obj = Variable(name, data, fastpath=True)
```
shoyer: good catch! I love it when fixes are this simple :)

PR #2515, opened 2018-10-26
Remove Dataset.T from api-hidden.rst

spencerkclark:
Just a minor followup to #2509 to remove `Dataset.T` from the documentation.


max-sixty:
Thanks @spencerkclark !

Issue #2505, opened 2018-10-24
xarray 0.11 release


shoyer:
We should really get a release candidate out soon for xarray 0.11, which will fix a lot of IO issues with dask (e.g., https://github.com/pydata/xarray/issues/2503).

Deprecation cycles to finish first:

- [x] Iterating over a Dataset iterates only over its data_vars #884
- [x] "in" operator does not work as expected on DataArray dimensions #1267
- [x] remove .T as an alias for .transpose() #1232
- [x] remove old resample syntax

Deprecation cycles to start (optional)

- [x] Deprecate inplace methods #1756

These were everything tagged with the "0.11" milestone.

@pydata/xarray anything else to add?

dcherian:
#2444 is a small bugfix for facetgrid that's ready for review.

These PRs seem to be almost done (it'd be nice to get some of these in):
#2470 (fill_value in shift)
#2449 (to_iris, from_iris)
#2207 (something about chunksizes)
#2415 (test for 1D object indexing)
#2413 (positional arguments for groupby apply)

rabernat:
Would be nice to get #2487 finished before release.

max-sixty:
I'll do ` Iterating over a Dataset iterates only over its data_vars`

dcherian:
Oooh we're supposed to remove `.T` as an alias for `.transpose()` #1232. finally!

jhamman:
Would it make sense to add a deprecation warning for Python2 users in this release (xref #1876). Chances are, 0.12 will come after Jan-2019 when we're slated to drop Python2 support.

shoyer:
> Would it make sense to add a deprecation warning for Python2 users in this release (xref #1876). Chances are, 0.12 will come after Jan-2019 when we're slated to drop Python2 support.

I agree, this will probably be the last major xarray release with Python 2 support. But I don't think there's much to be gained by adding a warning -- Python 2 users know the end is near.

jhamman:
> Python 2 users know the end is near.

Based on recent experience, I'm not quite as confident here 😄

fmaussion:
Did any other library have a warning for this? I think that if we follow the lead of ipython, matplotlib and https://python3statement.org/ then all we have to do is make 0.12 non-installable on py2, and py2 users will stick to 0.11 *ad vitam æternam*. 

It would be quite mean to have them listen to a warning all the time they import xarray on top of that :smile: 

[...]

Issue #2437, opened 2018-09-24
xarray potential inconstistencies with cftime


sbiner:
I am trying to use xarray with different types of calendars. I made a few tests and wonder if somebody can help me make sense of the results. In my test, I generate a DataArray `da` time series with a 365_day calendar using cftime. I then write that DataArray in a netCDF file and read it in another DataArray `da2`

#### Code Sample, a copy-pastable example if possible
```python
import xarray as xr
import cftime
import numpy as np

# generate data for 365_days calendar
#
units = 'days since 2000-01-01 00:00'
time_365 = cftime.num2date(np.arange(0, 10 * 365), units, '365_day')
da = xr.DataArray(np.arange(time_365.size), coords = [time_365], dims = 'time', name = 'data')

# write dataArray in netcdf and read it in new DataArray
da.to_netcdf('data_365.nc', 'w')
da2 = xr.open_dataset('data_365.nc').data

# try resample da
try:
    mean = da.resample(time='Y').mean()
    print(mean.values)
except TypeError:
    print('got TypeError for da')

# try resample da2
mean = da2.resample(time = 'Y').mean()
print (mean.values)
```
### Problem description
As seen in the code the resampledoes not work for `da` while it does for `da2`. The problem is related to the the type of `da.time` which is `cftime.DatetimeNoLeap` while `da2.time` is a `datetime64`. I thought that xarray is using cftime to make the conversion from time numerical values to dates but it looks to me as if it is not the case.

I wonder if this makes sense or if it is something that should eventually be corrected.

INSTALLED VERSIONS
In [6]: print (cftime.__version__)
1.0.1
<details>
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: fr_CA.UTF-8
LOCALE: fr_CA.UTF-8

xarray: 0.10.8
pandas: 0.23.0
numpy: 1.14.3
scipy: 1.1.0
netCDF4: 1.4.1
h5netcdf: None
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.5
distributed: 1.21.8
matplotlib: 2.2.2
cartopy: None
seaborn: 0.8.1
setuptools: 39.1.0
pip: 10.0.1
conda: 4.5.11
pytest: 3.5.1
IPython: 6.4.0
[...]

PR #2516, opened 2018-10-26
Switch enable_cftimeindex to True by default

spencerkclark:
As discussed in #2437 and #2505, this sets the option `enable_cftimeindex` to `True` by default.

 - [x] Fully documented, including `whats-new.rst` for all changes.

### DIFF
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -325,16 +325,16 @@ For data indexed by a :py:class:`~xarray.CFTimeIndex` xarray currently supports:
    - Resampling along the time dimension for data indexed by a
      :py:class:`~xarray.CFTimeIndex` (:issue:`2191`, :issue:`2458`)
    - Built-in plotting of data with :py:class:`cftime.datetime` coordinate axes
-     (:issue:`2164`).
-   
-   If at any time you would like to restore the old default behavior, which was
-   to attempt to decode datetimes into ``np.datetime64[ns]`` objects whenever
-   possible (regardless of calendar type), you can set ``enable_cftimeindex`` to
-   ``False`` within your context when opening a file (see
-   :py:func:`~xarray.set_options` for more information).  For some use-cases 
-   this behavior may still be useful (e.g. to allow the use of some forms
-   resample with non-standard calendars); however in this case one should use
-   caution to only perform operations which do not depend on differences
+     (:issue:`2164`).   
+
+   For some use-cases it may still be useful to convert from
+   :py:class:`cftime.datetime` objects to ``np.datetime64[ns]`` objects,
+   despite the difference in calendar types (e.g. to allow the use of some
+   forms resample with non-standard calendars).  The recommended way of doing
+   this is through...
+
+   However in this case one should
+   use caution to only perform operations which do not depend on differences
    between dates (e.g. differentiation, interpolation, or upsampling with
    resample), as these could introduce subtle and silent errors due to the
    difference in calendar types between the dates encoded in your data and the
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -71,10 +71,11 @@ One unfortunate limitation of using ``datetime64[ns]`` is that it limits the
 native representation of dates to those that fall between the years 1678 and
 2262. When a netCDF file contains dates outside of these bounds, dates will be
 returned as arrays of :py:class:`cftime.datetime` objects and a :py:class:`~xarray.CFTimeIndex`
-can be used for indexing.  The :py:class:`~xarray.CFTimeIndex` enables only a subset of
-the indexing functionality of a :py:class:`pandas.DatetimeIndex` and is only enabled
-when using the standalone version of ``cftime`` (not the version packaged with
-earlier versions ``netCDF4``).  See :ref:`CFTimeIndex` for more information.
+will be used for indexing.  :py:class:`~xarray.CFTimeIndex` enables a subset of
+the indexing functionality of a :py:class:`pandas.DatetimeIndex` and is only
+fully compatible with the standalone version of ``cftime`` (not the version
+packaged with earlier versions ``netCDF4``).  See :ref:`CFTimeIndex` for more
+information. 
 
 Datetime indexing
 -----------------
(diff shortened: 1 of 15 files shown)


dcherian:
Should we add nice error messages to resample and plot telling the user what to do if they try those two with cftimeindex?

doc/time-series.rst
```diff
@@ -223,16 +224,26 @@ Through the standalone ``cftime`` library and a custom subclass of
 functionality enabled through the standard :py:class:`pandas.DatetimeIndex` for
 dates from non-standard calendars or dates using a standard calendar, but
```
shoyer: Can we clarify "non-standard calendars" -> "non-standard calendars commonly used in climate science"

This paragraph is missing a bit of context for non-climate users :)

doc/time-series.rst
```diff
@@ -317,13 +312,32 @@ For data indexed by a :py:class:`~xarray.CFTimeIndex` xarray currently supports:
...
-   Currently resampling along the time dimension for data indexed by a
-   :py:class:`~xarray.CFTimeIndex` is not supported.
+   While much of the time series functionality that is possible for standard
+   dates has been implemented for dates from non-standard calendars, there are
+   still some remaining important features that have yet to be implemented,
+   for example:
+
+   - Resampling along the time dimension for data indexed by a
+     :py:class:`~xarray.CFTimeIndex`
+   - Built-in plotting of data with :py:class:`cftime.datetime` coordinate axes.
```
shoyer: It would be great to link to the GitHub issues for these.

doc/time-series.rst
```diff
@@ -317,13 +312,32 @@ For data indexed by a :py:class:`~xarray.CFTimeIndex` xarray currently supports:
...
+   
+   If at any time you would like to restore the old default behavior, which was
+   to attempt to decode datetimes into ``np.datetime64[ns]`` objects whenever
+   possible (regardless of calendar type), you can set ``enable_cftimeindex`` to
```
shoyer: I would rather encourage people to convert their data to a standard calendar via normal xarray/cftime APIs. That's a cleaner solution than setting global configuration.

Can we add an example of this with existing APIs? Or maybe we should augment CFTimeIndex with a new `to_datetime64` method?

doc/whats-new.rst
```diff
@@ -33,6 +33,12 @@ v0.11.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- The option ``enable_cftimeindex`` has now been set to ``True`` by default.
```
shoyer: We should think about eventually removing this option entirely, and making it simply a "no-op" for now (raising FutureWarning). In the long term I think we would prefer to avoid keeping around extraneous global options.

doc/time-series.rst
spencerkclark: > I would rather encourage people to convert their data to a standard calendar via normal xarray/cftime APIs. That's a cleaner solution than setting global configuration.

This is a really good point and would make things a lot easier to understand.  Currently we use `xarray.coding.times.cftime_to_nptime` internally; I can convert it to a better-documented public function (perhaps with friendlier error handling too).

doc/whats-new.rst
spencerkclark: Makes sense -- that will clean up the datetime decoding logic (and tests) quite a bit.

xarray/coding/times.py
```diff
@@ -291,7 +279,12 @@ def cftime_to_nptime(times):
     times = np.asarray(times)
     new = np.empty(times.shape, dtype='M8[ns]')
     for i, t in np.ndenumerate(times):
-        dt = datetime(t.year, t.month, t.day, t.hour, t.minute, t.second)
+        try:
+            dt = pd.Timestamp(t.year, t.month, t.day, t.hour, t.minute,
```
spencerkclark: I switched to using `pd.Timestamp` here, because directly converting from `datetime.datetime` to `np.datetime64[ns]` did not raise an error for dates outside 1678 to 2262:
```
In [1]: from datetime import datetime; import numpy as np

In [2]: np.datetime64(datetime(1, 1, 1), 'ns')
Out[2]: numpy.datetime64('1754-08-30T22:43:41.128654848')
```

xarray/coding/times.py
```diff
@@ -291,7 +279,12 @@ def cftime_to_nptime(times):
     times = np.asarray(times)
     new = np.empty(times.shape, dtype='M8[ns]')
     for i, t in np.ndenumerate(times):
-        dt = datetime(t.year, t.month, t.day, t.hour, t.minute, t.second)
```
spencerkclark: Was there a reason we did not include the `microsecond` attribute of the datetime here previously?

spencerkclark:
Thanks @shoyer -- that was a bit puzzling but I think I sorted it out.  I haven't done a careful diagnosis, but my hunch is that the previous build was running into this issue: https://github.com/pytest-dev/pytest/issues/2917.  A fix was merged a few weeks ago, but the Python 2.7 32-bit build is/was using a relatively old version of pytest (3.5 versus the most recent 3.9.3).

In the previous build, the FutureWarning for setting the `enable_cftimeindex` option was being emitted earlier in `test_backends.py` in the `test_open_mfdataset_manyfiles` test (before we were watching for it in the options tests).  It was being raised there, because upon calling `set_options.__exit__`, every option (not just the options that were set in a `with` statement) would be reset to its global value.  For example doing something like this would emit the warning (which even outside of this minor testing issue, I think is undesirable):
```
In [1]: import xarray

In [2]: with xarray.set_options(display_width=40):
  ...:     pass
  ...:
/Users/spencerclark/xarray-dev/xarray/xarray/core/options.py:49: FutureWarning: The enable_cftimeindex option is now a no-op and will be removed in a future version of xarray.
 FutureWarning)
```
In 6d08d3b I made changes to `set_options` to only fill the `self.old` dictionary with the old values for the keyword arguments specified in the constructor, and then had the `__exit__` method only reset the values that were temporarily set before.  The win-32 build now passes, and we no longer have the undesirable warning behavior shown above.  Do you think that was a reasonable fix?

PR #2520, opened 2018-10-28
Finish deprecation cycle for DataArray.__contains__ checking array values

### LINKED ISSUE #1267: "in" operator does not work as expected on DataArray dimensions

shoyer:
- [x] Closes #1267 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -33,11 +33,14 @@ v0.11.0 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
-- ``Dataset.T`` has been removed as a shortcut for :py:meth:`Dataset.transpose`.
-  Call :py:meth:`Dataset.transpose` directly instead.
-- Iterating over a ``Dataset`` now includes only data variables, not coordinates.
-  Similarily, calling ``len`` and ``bool`` on a ``Dataset`` now  
-  includes only data variables
+- Finished deprecation cycles:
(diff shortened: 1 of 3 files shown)


max-sixty:
Unrelated test failure around cftimes on AppVeyor

PR #2523, opened 2018-10-29
Raise more informative error when converting tuples to Variable.

### LINKED ISSUE #1016: DataArray coords: tuple vs. list
is there a reason the `coords` dictionary values cannot be a tuple? I found this fairly baffling to debug when it tripped me up. 

```
# this works
xarray.DataArray(np.random.random((3, 3, 3)),
    dims=('one', 'two', 'three'),
    coords={
         'one': ['four', 'five', 'six'],
    }
)

# this throws the following error
# ValueError: dimensions ('four',) must have the same length as the number of data dimensions, ndim=0
xarray.DataArray(np.random.random((3, 3, 3)),
    dims=('one', 'two', 'three'),
    coords={
         'one': ('four', 'five', 'six'),
    }
)
```

even if there was a clearer error it would help quite a bit. As it stands you are thinking, 'what?! four isn't a dimension!'

using 0.8.2

dcherian:
- [x] Closes #1016 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -76,11 +76,11 @@ def as_variable(obj, name=None):
     elif isinstance(obj, tuple):
         try:
             obj = Variable(*obj)
-        except TypeError:
+        except (TypeError, ValueError) as error:
             # use .format() instead of % because it handles tuples consistently
-            raise TypeError('tuples to convert into variables must be of the '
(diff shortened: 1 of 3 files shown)


shoyer:
I don't think this really needs a release note

PR #2524, opened 2018-10-29
Deprecate inplace

### LINKED ISSUE #1756: Deprecate inplace methods
The following methods have an `inplace` argument:
`DataArray.reset_coords`
`DataArray.set_index`
`DataArray.reset_index`
`DataArray.reorder_levels`
`Dataset.set_coords`
`Dataset.reset_coords`
`Dataset.rename`
`Dataset.swap_dims`
`Dataset.set_index`
`Dataset.reset_index`
`Dataset.reorder_levels`
`Dataset.update`
`Dataset.merge`

As proposed in https://github.com/pydata/xarray/issues/1755#issuecomment-348682403, let's deprecate all of these at the next major release (v0.11). They add unnecessary complexity to methods and promote confusing about xarray's data model.

Practically, we would change all of the default values to `inplace=None` and issue either a `DeprecationWarning` or `FutureWarning` (see [PEP 565](https://www.python.org/dev/peps/pep-0565/#additional-use-case-for-futurewarning) for more details on that choice).

dcherian:
- [x] Closes #1756  (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -657,7 +657,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         return cls._construct_direct(variables, coord_names, dims, attrs)
 
     def _replace_vars_and_dims(self, variables, coord_names=None, dims=None,
-                               attrs=__default_attrs, inplace=None):
+                               attrs=__default_attrs, inplace=False):
         """Fastpath constructor for internal use.
 
         Preserves coord names and attributes. If not provided explicitly,
@@ -677,7 +677,6 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         -------
         new : Dataset
         """
-        _check_inplace(inplace)
         if dims is None:
             dims = calculate_dimensions(variables)
         if inplace:
@@ -2480,7 +2479,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
             result = result._unstack_once(dim)
         return result
 
-    def update(self, other, inplace=None):
+    def update(self, other, inplace=True):
         """Update this dataset's variables with those from another dataset.
 
         Parameters
@@ -2502,7 +2501,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
             If any dimensions would have inconsistent sizes in the updated
             dataset.
         """
-        _check_inplace(inplace)
+        # _check_inplace(inplace)
         variables, coord_names, dims = dataset_update_method(self, other)
 
         return self._replace_vars_and_dims(variables, coord_names, dims,
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2480,7 +2480,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
             result = result._unstack_once(dim)
         return result
 
-    def update(self, other, inplace=True):
+    def update(self, other, inplace=None):
         """Update this dataset's variables with those from another dataset.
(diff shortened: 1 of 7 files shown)


xarray/core/dataset.py
```diff
@@ -2494,12 +2501,13 @@ def update(self, other, inplace=True):
             If any dimensions would have inconsistent sizes in the updated
             dataset.
         """
+        # _check_inplace(inplace)
```
dcherian: @shoyer What's the right thing to do here? Changing the default to None/False breaks a lot of test.

shoyer: `Dataset.update()` has always been an in-place operation for consistency with `dict.update()`. I don't think it really makes sense to change that.

I would suggest that we do the deprecation the other direction here: remove the `inplace` argument but keep the behavior as always inplace.

xarray/core/utils.py
```diff
@@ -18,6 +18,16 @@
     OrderedDict, basestring, bytes_type, dask_array_type, iteritems)
 
 
+def _check_inplace(inplace, default=False):
+    if inplace is None:
+        inplace = default
+    else:
+        warnings.warn('The inplace argument has been deprecated and will be '
+                      'removed in xarray 0.12.0.', FutureWarning)
```
shoyer: Can you add the `stacklevel` argument with an appropriate value? (you'll need to experiment a bit)

That ensures that warnings end up pointing to the specific line of code that need to be updated.

dcherian: `stacklevel=3` seems to give useful warnings with the test suite

```
xarray/tests/test_dataset.py::TestDataset::()::test_expand_dims_error
  /home/deepak/work/python/xarray/xarray/tests/test_dataset.py:2016: FutureWarning: The inplace argument has been deprecated and will be removed in xarray 0.12.0.
    original.set_coords('z', inplace=True)
```

dcherian:
Failed test is dask distributed test

xarray/tests/test_dataarray.py
```diff
@@ -23,6 +23,7 @@
     requires_scipy, source_ndarray)
 
 
+@pytest.mark.filterwarnings('ignore:The inplace argument')
```
shoyer: Can we try fixing these tests now instead? Otherwise we may be stuck with large clean-up later.

For testing deprecations, I find it  often helps to convert warnings into an error, e.g., by adding `raise Exception` on the line after issuing the warning.

dcherian: I thought we might want to keep those tests around while people are still using that code path.

shoyer: I forgot the main reason for doing this: to make sure that xarray doesn't rely upon the deprecated behavior internally.

But if you already checked that, then I suppose this is fine.

dcherian: Oh I haven't checked that closely but I can do that tonight. 
(my impression was that the tests were testing inplace behaviour, but I see that could be wrong)

shoyer:
In general, I would be OK with *not* testing inplace behavior carefully. We
can certainly add this to specific test methods but overall I think it's
better to have slightly less test coverage of deprecated behavior rather
than to raise deprecation warnings as part of unrelated operations.

dcherian:
OK so what's left is `Dataset._calculate_binary_op(g, other, inplace=True)` called by `Dataset._inplace_binary_op` at Line 3307 in dataset.py.  I don't know what the right fix is.

shoyer:
_calculate_binary_op shouldn't lose the inplace argument -- it's used for
Python inplace arithmetic like +=.

PR #2533, opened 2018-10-31
Check shapes of coordinates and data during DataArray construction

### LINKED ISSUE #2529: numpy.insert on DataArray may silently result in array inconsistent with its coordinates
```python
import numpy
import xarray

da = xarray.DataArray(numpy.arange(10*3).reshape(10, 3),
        dims=("x", "y"),
        coords={"foo": (("x", "y"), numpy.arange(3*10).reshape(10,3))})

print(da.shape == da["foo"].shape)

da2 = numpy.insert(da, 3, 0, axis=0)

print(da2.shape == da2["foo"].shape)
```

#### Problem description

Running the code snippet gives

```
True
False
```

and does not raise any exception.  In the resulting `da2`, the shape for `da2` and `da2['foo']` are different: we have changed the size of `da2` without changing the size of its corresponding `foo` coordinate.  This happens silently, no exception is thrown.  Inevitably, this is likely to result in problems at a later stage.

#### Expected Output

I would expect to get an exception, telling me that the insertion has failed because there are coordinates associated with the axis along which we are inserting values.  It would be nice to have an `xarray.insert` that can handle this, for example, by forcing us to provide corresponding insertion values for the coordinates.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.0.final.0
python-bits: 64
OS: Linux
OS-release: 2.6.32-754.el6.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8

xarray: 0.10.7
pandas: 0.23.2
numpy: 1.15.2
scipy: 1.1.0
netCDF4: 1.4.0
h5netcdf: 0.6.1
h5py: 2.8.0
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.18.1
distributed: 1.22.0
matplotlib: 3.0.0
cartopy: 0.16.0
seaborn: 0.9.0
setuptools: 39.2.0
pip: 18.0
conda: None
pytest: 3.2.2
IPython: 6.4.0
sphinx: None

</details>

lilyminium:
- [x] Closes #2529 (remove if there is no corresponding issue, which should only be the case for minor changes)

This sets DataArrayGroupBy.reduce(shortcut=False), as the shortcut first constructs a DataArray with the previous coordinates and the new mutated data before updating the coordinates; this order of events now raises a ValueError.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -433,7 +433,6 @@ class DataArray(AbstractArray, DataWithCoords, DataArrayArithmetic):
         coords=None,
         name: Union[Hashable, None, Default] = _default,
         indexes=None,
-        fastpath=True,
     ) -> "DataArray":
         if variable is None:
             variable = self.variable
@@ -441,8 +440,6 @@ class DataArray(AbstractArray, DataWithCoords, DataArrayArithmetic):
             coords = self._coords
         if name is _default:
             name = self.name
-        if not fastpath:
-            _check_shape_consistency(variable.shape, coords, variable.dims)
         return type(self)(variable, coords, name=name, fastpath=True, indexes=indexes)
 
     def _replace_maybe_drop_dims(
@@ -2984,7 +2981,9 @@ class DataArray(AbstractArray, DataWithCoords, DataArrayArithmetic):
 
     def __array_wrap__(self, obj, context=None) -> "DataArray":
         new_var = self.variable.__array_wrap__(obj, context)
-        return self._replace(new_var, fastpath=False)
+        if context is None:
+            _check_shape_consistency(new_var.shape, self._coords, new_var.dims)
+        return self._replace(new_var)
 
     def __matmul__(self, obj):
         return self.dot(obj)
@@ -3004,7 +3003,7 @@ class DataArray(AbstractArray, DataWithCoords, DataArrayArithmetic):
                 "ignore", r"Mean of empty slice", category=RuntimeWarning
             )
             with np.errstate(all="ignore"):
-                da = self.__array_wrap__(f(self.variable.data, *args, **kwargs))
+                da = self.__array_wrap__(f(self.variable.data, *args, **kwargs), -1)
             if keep_attrs:
                 da.attrs = self.attrs
             return da


xarray/core/dataarray.py
```diff
@@ -230,6 +235,9 @@ def __init__(self, data, coords=None, dims=None, name=None,
         # uncomment for a useful consistency check:
         # assert all(isinstance(v, Variable) for v in coords.values())
 
+        # check shape consistency
+        _check_shape_consistency(variable.shape, coords, variable.dims)
```
shoyer: This could be pretty expensive — it means we have to check shape consistency every time we create a DataArray, even if we know that the shape is already consistent, e.g., because it was created by another xarray operation.

shoyer:
@lilyminium thanks for looking into this! I think the specific problem here is due to our use of the shortcut `_replace` method in `DataArray.__array_wrap__` (which is called by NumPy, e.g., inside np.insert): 
https://github.com/pydata/xarray/blob/6d55f99905d664ef73cb708cfe8c52c2c651e8dc/xarray/core/dataarray.py#L193

I think the clean way to fix this would be to add error checking to `DataArray.__array_wrap__`. But we currently use this method inside `DataArray._unary_op` (and maybe elsewhere?), which is called every time you do unwary math with an xarray object (e.g., `- `). We’ll want to update these uses to use a method that uses the same shortcut to avoid redundant metadata checks.

lilyminium:
> I think the clean way to fix this would be to add error checking to `DataArray.__array_wrap__`. But we currently use this method inside `DataArray._unary_op` (and maybe elsewhere?), which is called every time you do unwary math with an xarray object (e.g., `- `). We’ll want to update these uses to use a method that uses the same shortcut to avoid redundant metadata checks.

Does adding fastpath to skip shape checking in __array_wrap__ work? fastpath=True in both DataArray.__array_wrap__ and Variable.__array_wrap__ didn't seem to break any tests.

shoyer:
> Does adding fastpath to skip shape checking in array_wrap work? fastpath=True in both DataArray.array_wrap and Variable.array_wrap didn't seem to break any tests.

I think it would be slightly better to stick with a separate method for xarray's fast-path. `__array_wrap__` is a special protocol used by NumPy, so there's always a (unlikely) risk that if we use different arguments in our version it could break in the future.

max-sixty:
This would still be valuable if @lilyminium (or someone else, if they're not interested) would want to finish this off.

lilyminium:
Ooh, this is a blast from the past. Anyone passing by, feel free to pick this up if you like, otherwise I'll see what I can do over the weekend :)

lilyminium:
Hmmmm. It's been quite some time since I've last used xarray, but I'm not sure I see an easy way to:

1. hijack np.insert (which I would guess goes directly to `DataArray.__array_wrap__ / __array_finalize__`)
2. without passing a `fastpath`-like argument to `__array_wrap__`, but
3. efficiently not checking the shape of safe operations, like `_unary_op`. 

I have for now stuck `fastpath` in `_replace` but it fails condition 3 and just runs for all calls to `__array_wrap__`.

lilyminium:
The 3.7 test looks like an installation error.

max-sixty:
> Hmmmm. It's been quite some time since I've last used xarray, but I'm not sure I see an easy way to:
> 
>     1. hijack np.insert (which I would guess goes directly to `DataArray.__array_wrap__ / __array_finalize__`)
> 
>     2. without passing a `fastpath`-like argument to `__array_wrap__`, but
> 
>     3. efficiently not checking the shape of safe operations, like `_unary_op`.

I think that's exactly correct. Without being able to tell `__array_wrap__` from `_unary_op` that we don't need to check the sizes, they all follow the same path. Is it possible to use the `context` arg for this? I don't think it's sufficiently flexible from an initial glance of the docs.

Maybe this new implementation has a low enough overhead that it's worthwhile?

If we do go ahead — what about calling `_check_shape_consistency` from within `__array_wrap__` after creating the object, instead of adding the `fastpath` kwarg?

lilyminium:
> context

I can only find references regarding `ufunc`s in the numpy docs when I look up the context. I would think that in that case the shapes should be fine, but please do correct me if not?

`_unary_op` calls `__array_wrap__(context=None)`; if the function only checks the shape when `context is None` and a dummy value gets passed in for the context, that's one way to avoid it.

max-sixty:
Nice, that seems to work!

Is anyone else more familiar with whether using `context` in this way is reasonable?

PR #2538, opened 2018-11-03
Stop loading tutorial data by default

jhamman:
- [x] Tests added 

In working on an xarray/dask tutorial, I've come to realize we eagerly load the tutorial datasets in `xarray.tutorial.load_dataset`. I'm going to just say that I don't think we should do that but I could be missing some rational. I didn't open an issue so please feel free to share thoughts here.

One option would be to create a new function (`xr.tutorial.open_dataset`) that does what I'm suggesting and then slowly deprecate `tutorial.load_dataset`. Thoughts?

xref: https://github.com/dask/dask-examples/pull/51

### DIFF
--- a/xarray/tutorial.py
+++ b/xarray/tutorial.py
@@ -102,6 +102,6 @@ def load_dataset(*args, **kwargs):
     warnings.warn(
         "load_dataset` will be removed in version 0.12. The current behavior "
         "of this function can be achived by using "
-        "`tutorial.open_dataset(...).load()`."
+        "`tutorial.open_dataset(...).load()`.",
         FutureWarning, stacklevel=2)
     return open_dataset(*args, **kwargs).load()
--- a/xarray/tutorial.py
+++ b/xarray/tutorial.py
@@ -85,6 +85,7 @@ def open_dataset(name, cache=True, cache_dir=_default_cache_dir,
     ds = _open_dataset(localfile, **kws)
 
     if not cache:
+        ds = ds.load()
         _os.remove(localfile)
 
     return ds
@@ -100,8 +101,8 @@ def load_dataset(*args, **kwargs):
     open_dataset
     """
     warnings.warn(
-        "load_dataset` will be removed in version 0.12. The current behavior "
-        "of this function can be achived by using "
+        "load_dataset` will be removed in xarray version 0.12. The current "
+        "behavior of this function can be achived by using "
         "`tutorial.open_dataset(...).load()`.",
-        FutureWarning, stacklevel=2)
+        DeprecationWarning, stacklevel=2)
     return open_dataset(*args, **kwargs).load()
--- a/xarray/tutorial.py
+++ b/xarray/tutorial.py
@@ -77,7 +77,7 @@ def load_dataset(name, cache=True, cache_dir=_default_cache_dir,
             """
             raise IOError(msg)
 
-    ds = _open_dataset(localfile, **kws).load()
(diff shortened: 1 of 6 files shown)


shoyer:
Our current tutorial datasets are 8MB and 17MB, which is pretty small. You'll definitely get better performance loading datasets of this size into NumPy arrays.

jhamman:
@shoyer - absolutely we'll get better performance with numpy arrays in this case. So I'm trying to use our tutorial datasets for some examples with dask (dask/dask-examples#51). The docstring for the `load_dataset` function states that we can pass kwargs on to the `open_dataset` function but if we pass `chunks` to the `load_dataset` call currently, we still get data back as numpy arrays. We have some other options here:

1. if chunks is a kwargs, return a dataset with data as persisted dask arrays
2. provide a second function to handle returning datasets using the same logic as `open_dataset` (caching, dask arrays, lazy loading, etc.)
3. tell people (like me) to rechunk the dataset after the fact

(3) won't require any changes but makes it a little harder to connect the typical use pattern of `open_dataset` with `tutorial.load_dataset`.

shoyer:
OK, that seems reasonable. The default behavior should cache the arrays
loaded with NumPy anyways. I would not be opposed to renaming this to
open_dataset, either.

shoyer:
> The default behavior should cache the arrays loaded with NumPy anyways.

Sorry, to be clear what I meant here is that by default arrays loaded with NumPy get cached after the first/access/operation. Not that we need to preserve the existing behavior of `load_dataset()`.

jhamman:
@shoyer - I think I was tracking with you. I've gone ahead and deprecated the current `load_dataset` in favor of the `open_dataset` name. The switch is accompanied by a change in behavior as well.

xarray/tutorial.py
```diff
@@ -77,9 +82,26 @@ def load_dataset(name, cache=True, cache_dir=_default_cache_dir,
             """
             raise IOError(msg)
 
-    ds = _open_dataset(localfile, **kws).load()
+    ds = _open_dataset(localfile, **kws)
 
     if not cache:
         _os.remove(localfile)
```
shoyer: We definitely need to load the dataset first if not caching -- otherwise you'll get errors when trying to use it.

xarray/tutorial.py
```diff
@@ -77,9 +82,26 @@ def load_dataset(name, cache=True, cache_dir=_default_cache_dir,
...
+
+    See Also
+    --------
+    open_dataset
+    """
+    warnings.warn(
+        "load_dataset` will be removed in version 0.12. The current behavior "
+        "of this function can be achived by using "
+        "`tutorial.open_dataset(...).load()`.",
+        FutureWarning, stacklevel=2)
```
shoyer: Maybe `DeprecationWarning`? Generally we save `FutureWarning` for things that are actually going to break in the future due to having different behavior.

xarray/tutorial.py
```diff
@@ -77,9 +82,26 @@ def load_dataset(name, cache=True, cache_dir=_default_cache_dir,
...
+    """
+    `load_dataset` will be removed in version 0.12. The current behavior of
+    this function can be achived by using `tutorial.open_dataset(...).load()`.
+
+    See Also
+    --------
+    open_dataset
+    """
+    warnings.warn(
+        "load_dataset` will be removed in version 0.12. The current behavior "
```
shoyer: "version 0.12" -> "xarray version 0.12"

PR #2548, opened 2018-11-07
DOC: reorganize whatsnew for xarray 0.11 release

shoyer:


doc/whats-new.rst
```diff
@@ -33,23 +33,8 @@ v0.11.0 (unreleased)
 Breaking changes
```
jhamman: just a reminder to update the release date here.

Issue #2159, opened 2018-05-18
Concatenate across multiple dimensions with open_mfdataset


TomNicholas:
#### Code Sample

```python
# Create 4 datasets containing sections of contiguous (x,y) data
for i, x in enumerate([1, 3]):
    for j, y in enumerate([10, 40]):
        ds = xr.Dataset({'foo': (('x', 'y'), np.ones((2, 3)))},
                         coords={'x': [x, x+1],
                                 'y': [y, y+10, y+20]})

        ds.to_netcdf('ds.' + str(i) + str(j) + '.nc')

# Try to open them all in one go
ds_read = xr.open_mfdataset('ds.*.nc')
print(ds_read)
```
#### Problem description

Currently ``xr.open_mfdataset`` will detect a single common dimension and concatenate DataSets along that dimension. However a common use case is a set of NetCDF files which have two or more common dimensions that need to be concatenated along simultaneously (for example collecting the output of any large-scale simulation which parallelizes in more than one dimension simultaneously). For the behaviour of ``xr.open_mfdataset`` to be n-dimensional it should automatically recognise and concatenate along all common dimensions.

#### Expected Output
```
<xarray.Dataset>
Dimensions:  (x: 4, y: 6)
Coordinates:
  * x        (x) int64 1 2 3 4
  * y        (y) int64 10 20 30 40 50 60
Data variables:
    foo      (x, y) float64 dask.array<shape=(4, 6), chunksize=(2, 3)>
```

#### Current output of ``xr.open_mfdataset()``
```
<xarray.Dataset>
Dimensions:  (x: 4, y: 12)
Coordinates:
  * x        (x) int64 1 2 3 4
  * y        (y) int64 10 20 30 40 50 60 10 20 30 40 50 60
Data variables:
    foo      (x, y) float64 dask.array<shape=(4, 12), chunksize=(4, 3)>
```

chiaral:
Thanks for opening up this issue. This would be very helpful for the forecasting community as well, where we usually concatenate along Start time and Lead time dimensions. 
Here, however, [was mentioned](https://stackoverflow.com/a/47557546/2789703) that it is quite difficult to generalize it, and he suggested a workaround. 
I know that some people did it for specific datasets, so maybe it would be helpful to add an example to the documentation that shows how this can be implemented on a case by case basis?

jhamman:
[...]

Issue #2039, opened 2018-04-05
open_mfdataset to blindly trust alignment


crusaderky:
This is a follow-up from #1521.

When invoking open_mfdataset, very frequently the user knows in advance that all of his coords that aren't
on the concat_dim are already aligned, and may be willing to blindly trust such assumption in exchange of a huge performance boost.

My production data: 200x NetCDF files on a not very performant NFS file system, concatenated on the "scenario" dimension:

```
xarray.open_mfdataset('cube.*.nc', engine='h5netcdf', concat_dim='scenario')

<xarray.Dataset>
Dimensions:      (attribute: 1, fx_id: 40, instr_id: 10765, scenario: 500001, timestep: 1)
Coordinates:
  * attribute    (attribute) object 'THEO/Value'
    currency     (instr_id) object 'ZAR' 'EUR' 'EUR' 'EUR' 'EUR' 'EUR' 'GBP' ...
  * fx_id        (fx_id) object 'GBP' 'USD' 'EUR' 'JPY' 'ARS' 'AUD' 'BRL' ...
  * instr_id     (instr_id) object 'S01626556_ZAE000204921' '537805_1275' ...
  * timestep     (timestep) datetime64[ns] 2016-12-31
    type         (instr_id) object 'American' 'Bond Future' 'Bond Future' ...
  * scenario     (scenario) object 'Base Scenario' 'SSMC_1' 'SSMC_2' ...
Data variables:
    FX           (fx_id, timestep, scenario) float64 dask.array<shape=(40, 1, 500001), chunksize=(40, 1, 2501)>
    instruments  (instr_id, attribute, timestep, scenario) float64 dask.array<shape=(10765, 1, 1, 500001), chunksize=(10765, 1, 1, 2501)>

CPU times: user 19.6 s, sys: 981 ms, total: 20.6 s
Wall time: 24.4 s
```

If I skip loading and comparing the non-index coords from all 200 files:

```
xarray.open_mfdataset('cube.*.nc'), engine='h5netcdf', concat_dim='scenario', coords='all')

<xarray.Dataset>
Dimensions:      (attribute: 1, fx_id: 40, instr_id: 10765, scenario: 500001, timestep: 1)
Coordinates:
  * attribute    (attribute) object 'THEO/Value'
  * fx_id        (fx_id) object 'GBP' 'USD' 'EUR' 'JPY' 'ARS' 'AUD' 'BRL' ...
[...]

PR #2553, opened 2018-11-10
Feature: N-dimensional auto_combine

TomNicholas:
### What I did

Generalised the `auto_combine()` function to be able to concatenate and merge datasets along any number of dimensions, instead of just one.

Provides one solution to #2159, and relevant for discussion in #2039.

Currently it cannot deduce the order in which datasets should be concatenated along any one dimension from the coordinates, so it just concatenates them in order they are supplied. This means for an N-D
concatenation the datasets have to be supplied as a list of lists, which is nested as many times as there are dimensions to be concatenated along.

### How it works

In `_infer_concat_order_from_nested_list()` the nested list of datasets is recursively traversed in order to create a dictionary of datasets, where the keys are the corresponding "tile IDs". These tile IDs are
tuples serving as multidimensional indexes for the position of the dataset within the hypercube of all datasets which are to be combined. For example four datasets which are to be combined along two dimensions would be supplied as 

```python
datasets = [[ds0, ds1], [ds2, ds3]]
```

and given tile_IDs to be stored as

```python
combined_ids = {(0, 0): ds0, (0, 1): ds1, (1, 0): ds2, (1, 1): ds3}
```

Using this unambiguous intermediate structure means that another method could be used to organise the datasets for concatenation (i.e. reading the values of their coordinates), and a new keyword argument
`infer_order_from_coords` used to choose the method. The `_combine_nd()` function concatenates along one dimension at a time, reducing the length of the tile_ID tuple by one each time `_combine_along_first_dim()` is called. After each concatenation the different variables are merged, so the new `auto_combine()` is essentially like calling the old one once for each dimension in `concat_dims`.

### Still to do

I would like people's opinions on the method I've chosen to do this, and any feedback on the code quality would be appreciated. Assuming we're happy with the method I used here, then the remaining tasks include:

- [x]  More tests of the final `auto_combine()` function

- [x] ~~Add option to deduce concatenation order from coords (or this could be a separate PR)~~

- [x] Integrate this all the way up to `open_mfdataset()`.

- [x] Unit tests for `open_mfdataset()`

- [x] More tests that the user has inputted a valid structure of datasets 

- [x] ~~Possibly parallelize the concatenation step?~~

- [x] A few other small `TODO`s which are in `combine.py`

- [x] Proper documentation showing how the input should be structured.

- [x] Fix failing unit tests on python 2.7 (though support for 2.7 is being dropped at the end of 2018?)

- [x] Fix failing unit tests on python 3.5

- [x] Update what's new

This PR was intended  to solve the common use case of collecting output from a simulation which was parallelized in multiple dimensions. I would like to write a tutorial about how to use xarray to do this, including examples of how to preprocess the data and discard processor ghost cells.


xarray/core/combine.py
```diff
@@ -369,34 +370,178 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+
+    Returns
+    -------
+    combined_tile_ids : dict[tuple(int, ...), xarray.Dataset]
+    """
+
+    from .dataset import Dataset
+
+    if isinstance(entry, list):
+        # Dive down tree and recursively open the next list
```
shoyer: I like this recursive approach, but I would suggest doing all the logic with immutable tuples. Pure functions are easier to reason about and less error prone.

Also, it's generally considered bad practice for a function to both mutate it's arguments and have a return value -- this makes it a less obvious whether a function uses mutation or not. You'll notice that all builtin functions and methods in Python either mutate their inputs (e.g., `list.append()`) or have a return value, but never do both.

I think this function could be really cleanly written as an iterator, e.g.,
```python
def _infer_tile_ids_from_nested_list(entry, current_pos):
    if isinstance(entry, list):
        for i, item in enumerate(entry):
            for result in _infer_tile_ids_from_nested_list(item, current_pos + (i,)):
                yield result
    else:
        yield current_pos, entry
``` 
Then you call it wrapping it with the `dict` constructor, e.g.,`combined_ids = dict(_infer_tile_ids_from_nested_list(datasets, ()))`

xarray/core/combine.py
```diff
@@ -369,34 +370,178 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+        return combined_tile_ids
+
+    else:
+        raise TypeError("Element at position " + str(tuple(current_pos)) +
+                        " is of type " + str(type(entry)) + ", which is "
+                        "neither a list nor an xarray.Dataset")
+
+
+def _check_shape_tile_ids(combined_tile_ids):
+    # TODO create custom exception class instead of using asserts?
```
shoyer: `ValueError` would be the usual error to raise here. And yes, this is probably a good idea -- otherwise somebody could try passing in a list with uneven depth, e.g., `[[ds1, ds2], ds3]`, which would probably result in a very difficult to debug error at some later point.

xarray/core/combine.py
```diff
@@ -369,34 +370,178 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+    tile_id, ds = single_id_ds_pair
+    return tile_id[1:]
+
+
+def _concat_along_first_dim(combined_IDs, dim, data_vars='all',
+                                coords='different'):
+    grouped = itertoolz.groupby(_new_tile_id, combined_IDs.items())
+    new_combined_IDs = {}
+
+    # TODO Would there be any point in parallelizing this concatenation step?
```
shoyer: Unlikely -- with dask, we parallelize processing the actual array data, but all this Python logic is hard to parallelize (due to Python's GIL). Unless you're dealing with a lot of very small datasets this wouldn't make a difference (and lot of small datasets will be slow regardless using Python).

shoyer:
I should have suggested this before, but the internal implementation of `numpy.block()` is a nice example of this sort of logic: https://github.com/numpy/numpy/blob/e726c045c72833d5c826e8a13f889746ee0bfdf4/numpy/core/shape_base.py#L661

xarray/core/combine.py
TomNicholas: Okay thanks, I didn't know about the mutate/return convention, but that makes sense.

xarray/core/combine.py
TomNicholas: Okay cool, I'll make sure that's covered sufficiently by the unit tests.

xarray/core/combine.py
TomNicholas: I came up with a better way to do this which involves another argument `ids` being passed the internal implementation of `auto_combine`, depending on whether or not the combined_ids have already been sorted. A prototype version is in the latest commits.

TomNicholas:
This is basically done now - I've implemented everything I wanted to, and included unit tests for the new functionality.
I'm also successfully using it in my [personal code now](https://github.com/TomNicholas/xBOUT/blob/master/xbout/load.py).

@shoyer I haven't changed the way the combine function is applied repeatedly to match the implementation in `numpy.block()`, I think the way [I've done it](https://github.com/pydata/xarray/blob/60c93ba59fd6636a9f6ebe2bb460b9924d9ebec7/xarray/core/combine.py#L478) is relatively clear, but let me know if you think it should be refactored.

I don't understand why some of the CI tests are failing - all test which run on my machine pass, and the errors in the log seem to come from dask arrays not being loaded, i.e:

```
AssertionError: 
    <xarray.Dataset>
    Dimensions:  (x: 10, y: 8)
    Data variables:
        foo      (x, y) float64 -1.19 0.4476 0.6283 ... -0.09699 -0.2311 -0.6482
    <xarray.Dataset>
    Dimensions:  (x: 10, y: 8)
    Data variables:
        foo      (x, y) float64 dask.array<shape=(10, 8), chunksize=(5, 4)>
```

TomNicholas:
I've fixed the code so it works with python 3.5 & 2.7 (tested on my local machine), but the tests on python 2.7 are still failing because for some reason it can't find the itertoolz module. I tried to fix this by adding toolz an explicit dependency in `setup.py` but that didn't work either. 

However, given that xarray is supposed to be dropping python 2 support at the end of this year (), does this particularly matter?

shoyer:
Itertoolz is a dask.array dependency that we didn't explicitly declare in
xarray (yes, this was sloppy). We should probably just copy the function we
use into some utils file inside xarray -- it has a compatible open source
license. Or we could rewrite to not need it.

TomNicholas:
Okay, I've rewritten it using `itertools.groupby` instead of `itertoolz.groupby` so that there should no longer be a dependence on `toolz`. The `itertools` groupby assumes the input is sorted, which makes the code slightly more complex, but I think that's still neater than including the functions and tests required for `itertoolz.groupby`.

TomNicholas:
Now that all the tests are passing for all python versions, and I've added an example of multidimensional concatenation to the docstring of `auto_combine`, I think this is ready to be merged.

However, the appveyor builds on python 2.7 are failing on setup when they try to install cftime, which I don't think was anything I did?

spencerkclark:
> However, the appveyor builds on python 2.7 are now failing on setup when they try to install cftime, which I don't think was anything I did?

Yes, this is unrelated.  See https://github.com/Unidata/cftime/issues/34#issuecomment-443619848; I'm sure it will be fixed soon.

xarray/core/combine.py
```diff
@@ -43,7 +45,7 @@ def concat(objs, dim=None, data_vars='all', coords='different',
           * list of str: The listed data variables will be concatenated, in
             addition to the 'minimal' data variables.
         If objects are DataArrays, data_vars must be 'all'.
-    coords : {'minimal', 'different', 'all' or list of str}, optional
+    coords : {'minimal', 'different', 'all' o list of str}, optional
```
shoyer:
```suggestion
    coords : {'minimal', 'different', 'all' or list of str}, optional
```

xarray/core/combine.py
```diff
@@ -369,37 +371,227 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+
+    return combined_ids, concat_dims
+
+
+def _infer_tile_ids_from_nested_list(entry, current_pos):
+    """
+    Given a list of lists (of lists...) of objects, returns a iterator
+    which returns a tuple containing the index of each object in the nested
+    list structure as the key, and the object. This can then be called by the
+    dict constructor to create a dictionary of the objects organised byt their
```
shoyer:
```suggestion
    dict constructor to create a dictionary of the objects organised by their
```

xarray/core/combine.py
```diff
@@ -369,37 +371,227 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+    else:
+        yield current_pos, entry
+
+
+def _check_shape_tile_ids(combined_tile_ids):
+    tile_ids = combined_tile_ids.keys()
+
+    # Check all tuples are the same length
+    # i.e. check that all lists are nested to the same depth
+    nesting_depths = [len(id) for id in tile_ids]
```
shoyer: Use another name other than `id` here to avoid masking the builtin `id` (maybe `tile_id`?)

xarray/core/combine.py
```diff
@@ -369,37 +371,227 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+
+
+def _auto_combine_1d(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
+                     compat='no_conflicts',
+                     data_vars='all', coords='different'):
+    # This is just the old auto_combine function (which only worked along 1D)
+    if concat_dim is not None:
+        dim = None if concat_dim is _CONCAT_DIM_DEFAULT else concat_dim
+        grouped = itertools.groupby(datasets,
+                                    key=lambda ds: tuple(sorted(ds.data_vars)),
```
shoyer: you can just use `tuple(sorted(ds))` now that (in xarray 0.11) iterating over a Dataset only iterates over its data variables

xarray/backends/api.py
```diff
@@ -594,6 +599,14 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     if not paths:
         raise IOError('no files to open')
 
+    # If infer_order_from_coords=True then this is unnecessary, but quick.
+    # If infer_order_from_coords=False then this creates a flat list which is
+    # easier to iterate over, while saving the originally-supplied structure
+    combined_ids_paths, concat_dims = _infer_concat_order_from_positions\
+        (paths, concat_dims)
```
shoyer: Style note: per PEP8, prefer adding extra or moving parentheses `()` rather than explicit continuations. This could be:
```python
    combined_ids_paths, concat_dims = _infer_concat_order_from_positions(
        paths, concat_dims)
```

xarray/backends/api.py
```diff
@@ -594,6 +599,14 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
+    ids, paths = list(combined_ids_paths.keys()), \
+                 list(combined_ids_paths.values())
```
shoyer: Likewise, you could add extra parentheses here.

shoyer:
Generally I like the look of this.

What happens if you have a nested list of Dataset objects with different data variables? e.g., a 2D grid of results, with different variables saved in different files. Is this supported at all?

I ask because the current version of `auto_combine()` was (for better or worse) designed to handle this automagically, but I'm not sure this makes sense for higher dimensional data.

It might be better to have a separate `nested_concat()` function rather than to squeeze this all into `auto_combine()`.

TomNicholas:
Thanks for the comments.

<br>

> What happens if you have a nested list of Dataset objects with different data variables? 

This is supported. This new `auto_combine()` simply applies the old `auto_combine()` N times along N dimensions, so if the grid of results is auto-combinable along all it's dimensions separately, then the new `auto_combine()` will auto-magically combine it all, e.g:

```python
objs = [[Dataset({'foo': ('x', [0, 1])}), Dataset({'bar': ('x', [10, 20])})],
        [Dataset({'foo': ('x', [2, 3])}), Dataset({'bar': ('x', [30, 40])})]]
expected = Dataset({'foo': ('x', [0, 1, 2, 3]), 
                    'bar': ('x', [10, 20, 30, 40])})

# This works
actual = auto_combine(objs, concat_dims=['x', None])

# Also works auto-magically
actual = auto_combine(objs)

# Proving it works symmetrically
objs = [[Dataset({'foo': ('x', [0, 1])}),   Dataset({'foo': ('x', [2, 3])})],
        [Dataset({'bar': ('x', [10, 20])}), Dataset({'bar': ('x', [30, 40])})]]
actual = auto_combine(objs, concat_dims=[None, 'x'])
``` 
(I'll add this example as another unit test)

<br>

I should point out that there is one way in which this function is not exactly as general as `auto_combine()` applied N times though. The options `compat`, `data_vars`, and `coords` are specified once for the combining along all dimensions, so you can't currently tell it to use `compat='identical'` along `dim1` and `compat='no_conflicts'` along `dim2`, it has to be the same for both. I thought about making these kwargs also accept lists, but although that would be easy to do it seemed like it would complicate the API for a very specific use case?

<br>

> It might be better to have a separate `nested_concat()` function rather than to squeeze this all into `auto_combine()`.

That was basically what I tried to do in my first attempt, but nested concatenation without merging along every dimension misses some common use cases, for example if you wanted to `auto_combine` files structured like
```bash
root
├─ time1
│   ├── density.nc
│   └── temperature.nc
└─ time2
    ├── density.nc
    └── temperature.nc
```
then you would want to merge `density.nc` and `temperature.nc`, then concat along 'time'.

<br>

You might also find it interesting to see how I've used this fork in my own code: I create the grid of datasets [here](https://github.com/TomNicholas/xBOUT/blob/615f6caf01fa3d90cdfe8a5b517b477cc0c67026/xbout/load.py#L127), so that I can combine them [here](https://github.com/TomNicholas/xBOUT/blob/615f6caf01fa3d90cdfe8a5b517b477cc0c67026/xbout/load.py#L23).

<br>

I have a question actually - currently if the concat or merge fails, then the error message won't clearly tell you which dimension it was trying to combine along when it failed. Is there a way to do that easily with `try... except...` statements? Something like
```python
for dim in concat_dims:
    try:
        _auto_combine_along_first_dim(...)
    except MergeError or ValueError as err:
        raise ValueError(f"Encoutered {err} while trying to combine along dimension {dim}")
```

<br>

(Also something else is breaking in `cftime` on the python 2.7 builds on AppVeyor now...)

shoyer:
> if the grid of results is auto-combinable along all it's dimensions separately, then the new auto_combine() will auto-magically combine it all

OK, so the rule is that a grid dimension can correspond to any of:

1. merging data variables
2. concatenating along the dimension
3. both merging and concatenating

And to specify case (2) explicitly, you can write `dim=None` along that dimension?

TomNicholas:
Specifying dim=None will specify (1) explicitly, but otherwise yes, that's exactly what it does.

That is how the current auto_combine behaves too, just for 1D obviously.

TomNicholas:
Also if you specify dim='something' when that dimension doesn't already exist, you are explicitly telling it to add a new dimension.

shoyer:
> Specifying dim=None will specify (1) explicitly, but otherwise yes, that's exactly what it does.

Yes, this was a typo on my part :).

TomNicholas:
I appreciate that this is pretty complicated, perhaps it should have its own section in the docs (I can do another PR?)

shoyer:
Do you think it would make sense to try to get rid of "both merging and concatenating" in favor of requiring another dimension in the grid? I guess we do sort of need this for this for `open_mfdataset` but it fills more complex than desired.

TomNicholas:
I think that would mean there might be some situations that 1D auto_combine() could deal with but nested_concat_and_merge() could not. Situations where you can only merge the variables once they have been individually concatenated. 

I think this is a question of whether you think that:

a) auto_combine() is a well-defined operation that is still well-defined in N-D

or

b) auto_combine() 's purpose is to do it's best to combine whatever is thrown at it, and more surgical N-D operations should be the job of another function

I personally think a), but I expect users who haven't read the source code for auto_combine might disagree.

shoyer:
> I think that would mean there might be some situations that 1D auto_combine() could deal with but nested_concat_and_merge() could not. Situations where you can only merge the variables once they have been individually concatenated. (actually I'm not sure - I would need to think about that)

I *think* multiple levels `nested_concat_and_merge()` could always replace `auto_combine()` as long as merging happens last, at the outer-most level. That's basically what `auto_combine()` already does.

There are probably some edge cases where the existing `auto_combine` would you be a little more slopping about how you put together your lists of variables. But I don't think that's really a good idea for us to support in xarray, since it removes most of the freedom from the implementation.

> I personally think a), but I expect users who haven't read the source code for auto_combine() might disagree.

Yes, this is my concern. The way this API handles nested lists makes sense from an implementation perspective, but not really from a user perspective. For users, I think it makes sense to have:
1. The precisely defined `nested_concat_and_merge()` for cases where you already data sorted. This is pretty common (e.g., with hierarchical directory structures).
2. A magical `auto_combine()` that inspects coordinates to figure out how to put everything together. We might even rename this to `combine()` (if we think it's comprehensive enough).

The original version of `auto_combine()` that I wrote was aiming towards this second use-case, but was obviously fragile and incomplete.

If you like, we could also merge this work (which is excellent progress towards user-facing APIs) but keep the changes internal to xarray for now until we figure out the public APIs.

TomNicholas:
> There are probably some edge cases where the existing auto_combine would you be a little more slopping about how you put together your lists of variables.

I think that's probably the case, but I also think that those edge cases will be so specific that maybe we don't have to explicitly support them. We could just say that anyone who has a combination of datasets that is that funky can just concatenate them themselves?

> For users, I think it makes sense to have...

I agree, two separate functions is a lot more intuitive than having `auto_combine()` behave either auto-magically or totally manually depending on the value of an optional keyword argument.

The only problem with that idea is that both of these functions should be options for `open_mfdataset()`, so you still have multiple possible argument structures for that function. I think that's okay though - it would be a lot easier to explain if you say "if the keyword argument `combine='auto'` then the function `auto_combine()` is used, if `combine='manual` then `nested_concat_and_merge()` is used, and the filepaths must be structured accordingly."

> If you like, we could also merge this work (which is excellent progress towards user-facing APIs) but keep the changes internal to xarray for now until we figure out the public APIs.

That would be great. Then I could start using the master branch of xarray again in [my code](https://github.com/boutproject/xBOUT/issues/22), while we redo the public API. If I set `concat_dims` back to `concat_dim` then it *should* be completely backwards-compatible.

shoyer:
> The only problem with that idea is that both of these functions should be options for `open_mfdataset()`, so you still have multiple possible argument structures for that function. I think that's okay though - it would be a lot easier to explain if you say "if the keyword argument `combine='auto'` then the function `auto_combine()` is used, if `combine='manual'` then `nested_concat_and_merge()` is used, and the filepaths must be structured accordingly."

This sounds pretty good to me.

> > If you like, we could also merge this work (which is excellent progress towards user-facing APIs) but keep the changes internal to xarray for now until we figure out the public APIs.
> 
> That would be great. Then I could start using the master branch of xarray again in [my code](https://github.com/boutproject/xBOUT/issues/22), while we redo the public API. If I set `concat_dims` back to `concat_dim` then it _should_ be completely backwards-compatible.

OK, do you want to go ahead and revert the documentation/public API for now?

I would even be OK supporting nested lists temporarily in xarray via APIs like `open_mfdataset` and `auto_concat` without official documentation/support -- using this feature would be "at your own risk". As long as we don't break any existing code, that is totally fine. We just shouldn't advertise an interface until we've happy with it.

TomNicholas:
> This sounds pretty good to me.

Okay good. So with this API then `auto_combine()` (and therefore the default behaviour of `open_mfdatset()`) would be to [throw a warning](https://github.com/pydata/xarray/issues/2159#issuecomment-435336049) if there isn't enough info in the dataset coords to order them?  

> OK, do you want to go ahead and revert the documentation/public API for now?

Yes, I'll do that.

> I would even be OK supporting nested lists temporarily ... "at your own risk"

If I revert the documentation and you merge this PR then that's exactly what we will have, which would be useful for me until we do the public API.

(Also it seems that whatever was wrong with cftime has been fixed now, as the CI tests are passing.)

shoyer:
> Okay good. So with this API then `auto_combine()` (and therefore the default behaviour of `open_mfdatset()`) would be to [throw a warning](https://github.com/pydata/xarray/issues/2159#issuecomment-435336049) if there isn't enough info in the dataset coords to order them?

Yes, but throwing a warning should probably only be a temporary solution. Long term we should pick a default value (e.g., `combine='auto'`) and require explicitly opting in to different behavior (e.g., `combine='manual'`).

TomNicholas:
Okay I've reverted the API, so it basically converts a `concat_dim` input to it's multidimensional version before continuing.

If we merge this then should I start a separate Pull Request for further discussion about the API?

One of the Travis CI builds failed but again I don't think that was me.

xarray/backends/api.py
```diff
@@ -483,12 +483,11 @@ def close(self):
 def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
                    compat='no_conflicts', preprocess=None, engine=None,
                    lock=None, data_vars='all', coords='different',
+                   infer_order_from_coords=False,
```
shoyer: Can we leave this this out of the public function signature for now, given that it doesn't work?

xarray/core/combine.py
```diff
@@ -369,24 +371,197 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+                           data_vars=data_vars, coords=coords)
+    return combined
+
+
 def auto_combine(datasets,
                  concat_dim=_CONCAT_DIM_DEFAULT,
                  compat='no_conflicts',
-                 data_vars='all', coords='different'):
+                 data_vars='all', coords='different',
+                 infer_order_from_coords=False):
```
shoyer: same concern as above

xarray/core/combine.py
```diff
@@ -369,24 +371,197 @@ def _auto_concat(datasets, dim=None, data_vars='all', coords='different'):
...
+
+    tile_id, ds = list(combined_ids.items())[0]
+    n_dims = len(tile_id)
+    if concat_dims == _CONCAT_DIM_DEFAULT or concat_dims is None:
+        concat_dims = [concat_dims]*n_dims
+    else:
+        if len(concat_dims) != n_dims:
+            raise ValueError("concat_dims has length " + str(len(concat_dims))
+                             + " but the datasets passed are nested in a " +
+                             str(n_dims) + "-dimensional structure")
```
shoyer: Nit: usually it's more readable to write/read these with format strings (either using `%` or `.format()`), e.g., 
```python
            raise ValueError("concat_dims has length %d " 
                             "but the datasets passed are nested in a "
                             "%d-dimensional structure" % (len(concat_dims), n_dims))
```

TomNicholas:
Removed the unnecessary argument.

shoyer:
thanks @TomNicholas! I'm looking forward to the follow-ups :)

PR #2556, opened 2018-11-16
Add transpose_coords option to DataArray.transpose

### LINKED ISSUE #1856: Option to make DataArray.transpose also transpose coords
I've noticed that `Dataset.transpose` transposes the coordinates while `DataArray.transpose` does not.

Example:

    import xarray as xr
    import numpy as np
    
    X_da = xr.DataArray(
        np.random.random((100, 10)),
        coords={'coord_1': (['sample', 'feature'], np.ones((100, 10)))},
        dims=('sample', 'feature')
    )
    
    In []: X_da.transpose().coord_1.dims
    Out[]: ('sample', 'feature')
    
    X_ds = xr.Dataset(
        {'var_1': (['sample', 'feature'], np.random.random((100, 10)))},
        coords={'coord_1': (['sample', 'feature'], np.ones((100, 10)))},
    )
    
    In []: X_ds.transpose().coord_1.dims
    Out[]: ('feature', 'sample')

This behaviour is probably intentional, but there are cases where I'd like `DataArray.transpose` to also transpose the coordinates. As a workaround, I have to convert to a Dataset, transpose and convert back.

I was thinking that `DataArray.transpose` could accept a keyword argument `transpose_coords` which would be `False` by default. I could work on a PR implementing this behavior if it's desired.

phausamann:
- [x] Closes #1856 

I've added the option `transpose_coords` to `DataArray.transpose` as described in #1856. It's `False` by default, as it breaks a couple of tests otherwise (`TestConcatDataset.test_concat`, `test_apply_output_core_dimension` and `test_dot`). I'm not sure whether fixing these tests to work with the new behavior should be part of the PR.


shoyer:
As a default, I think we want `transpose_coords=None`.

If `transpose_coords is None` and there are coordinates to transpose, we can issue a `FutureWarning` about how the default behavior of `transpose_coords` will switch to `True` in the future.

> I'm not sure whether fixing these tests to work with the new behavior should be part of the PR.

Yes, this would be ideal. It would be fine to use `transpose_coords=False` explicitly in tests, but we should silence all warnings.

phausamann:
I've updated the method to issue the warning and fixed the failing tests (which was actually a bug in my  implementation).

> It would be fine to use `transpose_coords=False` explicitly in tests, but we should silence all warnings.

I'm not exactly sure how this works, there are a lot of tests where the warning is issued, especially calls to `DataArray.T`. Should I update all of these tests to `transpose(..., transpose_coords=False)`?

xarray/core/dataarray.py
```diff
@@ -1381,8 +1383,29 @@ def transpose(self, *dims):
         numpy.transpose
         Dataset.transpose
         """
+        if dims:
+            if set(dims) ^ set(self.dims):
+                raise ValueError('arguments to transpose (%s) must be '
+                                 'permuted array dimensions (%s)'
+                                 % (dims, tuple(self.dims)))
+
+        transpose_coords = kwargs.pop('transpose_coords', None)
```
phausamann: Should we raise an error when keyword args other than `transpose_coords` are being passed? `kwargs` is necessary for Python 2 compatibility.

shoyer: if we wait 40 days before mergin this, we wouldn't have to worry about Python 2 compatibility at all :)

But yes, for now:
```
if kwargs:
    raise TypeError('unrecognized keyword arguments')
```

shoyer:
> I'm not exactly sure how this works, there are a lot of tests where the warning is issued, especially calls to `DataArray.T`. Should I update all of these tests to `transpose(..., transpose_coords=False)`?

Yes, that would probably be the right call. Or maybe even switch to `.transpose(transpose_coords=True)` to future proof the behavior?

We try to avoid issuing warnings in xarray's own test suite so we notice issues (e.g., deprecations in upstream packages) when they arise. This also gives us some estimate of how much turmoil a change will cause for users -- if we end up with lots of internal warnings, then users will probably notice, too.

xarray/core/groupby.py
```diff
@@ -473,7 +473,7 @@ def lookup_order(dimension):
             return axis
 
         new_order = sorted(stacked.dims, key=lookup_order)
-        return stacked.transpose(*new_order)
+        return stacked.transpose(*new_order, transpose_coords=True)
```
phausamann: Should this be `False`?

phausamann:
I've change all the `tranpose` calls in the tests to `transpose_coords=True`. Other than that, there's some `tranpose` call in `plot` and `groupby` that I've updated, but I'm not sure whether they should be `True`, `False` or not updated at all (see above).

xarray/core/groupby.py
shoyer: We should probably do the same thing for this transpose in groupby that we did for transpose() in general -- we should have a keyword argument to control this behavior, e.g., perhaps `ds.groupby('something', restore_coord_dims=True).mean()`?

We definitely don't want to issue warnings internally that users can't control. `groupby()` using `transpose()` internally is an implementation detail that isn't part of xarray's public API.

phausamann:
I've added a `restore_coord_dims` parameter to `groupby`, `groupby_bins` and `resample` and eliminated all warnings from the tests. The build failure seems to be inherited from the master branch.

phausamann:
@shoyer is this ready to be merged?

StanczakDominik:
Hi, just popping in - perhaps rebasing on top of the master branch would help move this along?

shoyer:
I just merged in master, let's see how that goes

phausamann:
Looks like the build failures were only temporary, all the checks are passing now. I think this is good to go.

shoyer:
OK, in it goes!

Thanks for your patience @phausamann !

PR #2557, opened 2018-11-16
add missing comma and article in error message

gerritholl:
Add missing comma and article in error message when attribute values have the wrong type.

I tihnk this change is sufficiently minor that no documentation or whatsnew changes should be necessary.


jhamman:
You get two bonus points for submitting a PR that improves both an error message and the documentation. Thanks!

PR #2559, opened 2018-11-20
Zarr consolidated

### LINKED ISSUE #2558: how to incorporate zarr's new open_consolidated method?
Zarr has a new feature called [consolidated metadata](https://zarr.readthedocs.io/en/latest/tutorial.html#consolidating-metadata). This feature will make it much faster to open certain zarr datasets, because all the metadata needed to construct the xarray dataset will live in a single .json file.

To use this new feature, the new function `zarr.open_consolidated` needs to be called. So it won't work with xarray out of the box. We need to decide how to add support for this at the xarray level.

**I am seeking feedback on what API people would like to see before starting a PR.** My proposal is to add a new keyword argument to `xarray.open_zarr` called `consolidated` (default = False). An alternative would be to automatically try `open_consolidated` and fall back on the standard `open_group` function if that fails.

I played around with this a bit and realized that https://github.com/zarr-developers/zarr/issues/336 needs to be resolved before we can do the xarray side.

cc @martindurant, who might want to weigh on what would be most convenient for intake.

rabernat:
This PR adds support for reading and writing of [consolidated metadata](https://zarr.readthedocs.io/en/latest/tutorial.html#consolidating-metadata) in zarr stores.

 - [x] Closes #2558 (remove if there is no corresponding issue, which should only be the case for minor changes)

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -650,10 +650,10 @@ opening the store. (For more information on this feature, consult the
 
 If you have zarr version 2.3 or greater, xarray can write and read stores
 with consolidated metadata. To write consolidated metadata, pass the
-``consolidate=True`` option to the
+``consolidated=True`` option to the
 :py:attr:`Dataset.to_zarr <xarray.Dataset.to_zarr>` method::
 
-    ds.to_zarr('foo.zarr', consolidate=True)
+    ds.to_zarr('foo.zarr', consolidated=True)
 
 To read a consolidated store, pass the ``consolidated=True`` option to
 :py:func:`~xarray.open_zarr`::
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -641,10 +641,10 @@ Consolidated Metadata
 Xarray needs to read all of the zarr metadata when it opens a dataset.
 In some storage mediums, such as with cloud object storage (e.g. amazon S3),
 this can introduce significant overhead, because two separate HTTP calls to the
-object store for each variable in the dataset.
-With version 2.3, zarr will support an feature called *consolidated metadata*,
+object store must be made for each variable in the dataset.
+With version 2.3, zarr will support a feature called *consolidated metadata*,
 which allows all metadata for the entire dataset to be stored with a single
-key (by default called ``.zmetadata``). This can drastically reduce the latency
+key (by default called ``.zmetadata``). This can drastically speed up
 opening the store. (For more information on this feature, consult the
 `zarr docs <https://zarr.readthedocs.io/en/latest/tutorial.html#consolidating-metadata>`_.)
 
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -635,6 +635,31 @@ For example:
     Not all native zarr compression and filtering options have been tested with
     xarray.
 
+Consolidated Metadata
+~~~~~~~~~~~~~~~~~~~~~
+
+Xarray needs to read all of the zarr metadata when it opens a dataset.
+In some storage mediums, such as with cloud object storage (e.g. amazon S3),
+this can introduce significant overhead, because two separate HTTP calls to the
+object store for each variable in the dataset.
+With version 2.3, zarr will support an feature called *consolidated metadata*,
+which allows all metadata for the entire dataset to be stored with a single
+key (by default called ``.zmetadata``). This can drastically reduce the latency
+opening the store. (For more information on this feature, consult the
+`zarr docs <https://zarr.readthedocs.io/en/latest/tutorial.html#consolidating-metadata>`_.)
+
+If you have zarr version 2.3 or greater, xarray can write and read stores
+with consolidated metadata. To write consolidated metadata, pass the
+``consolidate=True`` option to the
+:py:attr:`Dataset.to_zarr <xarray.Dataset.to_zarr>` method::
+
+    ds.to_zarr('foo.zarr', consolidate=True)
+
+To read a consolidated store, pass the ``consolidated=True`` option to
+:py:func:`~xarray.open_zarr`::
+
+    ds = xr.open_zarr('foo.zarr', consolidated=True)
+
 .. _io.cfgrib:
 
 GRIB format via cfgrib
@@ -678,7 +703,7 @@ Formats supported by PseudoNetCDF
 ---------------------------------
 
 xarray can also read CAMx, BPCH, ARL PACKED BIT, and many other file
-formats supported by PseudoNetCDF_, if PseudoNetCDF is installed. 
+formats supported by PseudoNetCDF_, if PseudoNetCDF is installed.
 PseudoNetCDF can also provide Climate Forecasting Conventions to
 CMAQ files. In addition, PseudoNetCDF can automatically register custom
 readers that subclass PseudoNetCDF.PseudoNetCDFFile. PseudoNetCDF can
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -659,6 +659,10 @@ To read a consolidated store, pass the ``consolidated=True`` option to
 :py:func:`~xarray.open_zarr`::
 
     ds = xr.open_zarr('foo.zarr', consolidated=True)
+    
+Xarray can't perform consolidation on pre-existing zarr datasets. This should
+be done directly from zarr, as described in the
+`zarr docs <https://zarr.readthedocs.io/en/latest/tutorial.html#consolidating-metadata>`_.
 
 .. _io.cfgrib:
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -36,7 +36,7 @@ Breaking changes
 Enhancements
 ~~~~~~~~~~~~
 
-- Ability to read and write consolidated metadata in zarr stores.
+- Ability to read and write consolidated metadata in zarr stores (:issue:`2558`).
   By `Ryan Abernathey <https://github.com/rabernat>`_.
 - :py:class:`CFTimeIndex` uses slicing for string indexing when possible (like
   :py:class:`pandas.DatetimeIndex`), which avoids unnecessary copies.
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -36,6 +36,9 @@ Breaking changes
 Enhancements
 ~~~~~~~~~~~~
 
+- Ability to read and write consolidated metadata in zarr stores.
+  By `Ryan Abernathey <https://github.com/rabernat>`_.
+
 Bug fixes
 ~~~~~~~~~
 
@@ -52,15 +55,15 @@ Breaking changes
   - ``Dataset.T`` has been removed as a shortcut for :py:meth:`Dataset.transpose`.
     Call :py:meth:`Dataset.transpose` directly instead.
   - Iterating over a ``Dataset`` now includes only data variables, not coordinates.
-    Similarily, calling ``len`` and ``bool`` on a ``Dataset`` now  
+    Similarily, calling ``len`` and ``bool`` on a ``Dataset`` now
     includes only data variables.
   - ``DataArray.__contains__`` (used by Python's ``in`` operator) now checks
-    array data, not coordinates. 
+    array data, not coordinates.
   - The old resample syntax from before xarray 0.10, e.g.,
     ``data.resample('1D', dim='time', how='mean')``, is no longer supported will
     raise an error in most cases. You need to use the new resample syntax
     instead, e.g., ``data.resample(time='1D').mean()`` or
-    ``data.resample({'time': '1D'}).mean()``. 
+    ``data.resample({'time': '1D'}).mean()``.
 
 
 - New deprecations (behavior will be changed in xarray 0.12):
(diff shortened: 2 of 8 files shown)


xarray/backends/api.py
```diff
@@ -876,16 +876,23 @@ def to_zarr(dataset, store=None, mode='w-', synchronizer=None, group=None,
...
-    dump_to_store(dataset, store, writer, encoding=encoding)
+    dump_to_store(dataset, zstore, writer, encoding=encoding)
     writes = writer.sync(compute=compute)
 
     if not compute:
         import dask
-        return dask.delayed(_finalize_store)(writes, store)
-    return store
+        return dask.delayed(_finalize_store)(writes, zstore)
+        # TODO: handle consolidation for delayed case
```
rabernat: This is the one major issue I don't know how to deal with. `consolidate_metadata` needs to be called on the target PutableMapping *after* the store is done writing. I don't know how to do this in a delayed context.

rabernat:
Ping @lilyminium for a review.

xarray/backends/api.py
jhamman: what does the call signature look like for the consolidation step? We can probably make that a delayed thing and have it depend on the output from `_finalize_store`.

So something like:

```Python
@delayed
def consolidate_zarr(store, finalized_store):
    # replace store with whatever you need to pass to consolidate metadata
    return zarr.consolidate_metadata(store)

...
if not compute:
    import dask
    finalized_store = dask.delayed( _finalize_store)(writes, zstore)
    return consolidate_zarr(store, finalized_store)
```

`finalized_store` is bad name here but this is just a dummy argument to add the task dependency.

xarray/backends/api.py
```diff
@@ -876,16 +876,23 @@ def to_zarr(dataset, store=None, mode='w-', synchronizer=None, group=None,
...
+
+    if consolidate:
+        import zarr
+        zarr.consolidate_metadata(store)
+        # do we need to reload ztore now that we have consolidated?
```
jhamman: would it make sense for zarr to handle this?

rabernat: What do you mean?

xarray/backends/api.py
rabernat: the call signature is exactly as you said:
```
        import zarr
        zarr.consolidate_metadata(store)
```

rabernat:
Also need to add some version checks...this will only work with zarr > 2.2.

xarray/backends/api.py
shoyer: If you look at `_finalize_store`, it's just a few lines:
```python
def _finalize_store(write, store):
    """ Finalize this store by explicitly syncing and closing"""
    del write  # ensure writing is done first
    store.close()
```

So logically, it could make sense to do this cleanup in a `ZarrStore.close()` method, e.g.,
```
def close(self):
    if self._consolidate:
        zarr.consolidate_metadata(self.ds)
```

xarray/backends/api.py
jhamman: I meant reloading the zarr store automatically.

rabernat: I think that would be hard to achieve. And I'm not sure it's necessary. Frankly I don't know why we return a store object from `to_zarr` at all.

doc/io.rst
```diff
@@ -635,6 +635,31 @@ For example:
...
+with consolidated metadata. To write consolidated metadata, pass the
+``consolidate=True`` option to the
+:py:attr:`Dataset.to_zarr <xarray.Dataset.to_zarr>` method::
+
+    ds.to_zarr('foo.zarr', consolidate=True)
+
+To read a consolidated store, pass the ``consolidated=True`` option to
+:py:func:`~xarray.open_zarr`::
+
+    ds = xr.open_zarr('foo.zarr', consolidated=True)
```
shoyer: I wonder if we should try to use the same keyword argument name here -- it seems a little confusing to write `consolidated=True` when reading but `consolidate=True` when writing. Maybe both should be `consolidated=True`?

rabernat: I see your point, and I thought about this. I decided on what I have here because `consolidate` sounds more active and lets you know that some action, consolidation, will be performed on the store.

But I don't feel strongly about this and would be happy to change it.

xarray/tests/test_backends.py
```diff
@@ -1320,6 +1321,18 @@ def roundtrip_append(self, data, save_kwargs={}, open_kwargs={},
                          allow_cleanup_failure=False):
         pytest.skip("zarr backend does not support appending")
 
+    def test_roundtrip_consolidated(self):
+        # my understanding is that pytest.mark.skipif is broken here
+        import zarr
+        if LooseVersion(zarr.__version__) <= '2.2':
+            pytest.skip('zarr version 2.3 or greater needed for consolidated')
```
rabernat: It looks like this way of skipping tests is not working on travis:
https://travis-ci.org/pydata/xarray/jobs/459866378#L7637

Any suggestions for how to do this better?

jhamman: How does `LooseVersion(zarr.__version__) <= LooseVersion('2.2')` work?

You can probably also use:

```python
zarr = pytest.importorskip("zarr", minversion="2.2")
```

rabernat: The problem is that version 2.3 is not released yet. So we need something that skips 2.2 but runs on anything > 2.2. (Current version shows up as 2.2.1.dev2.)

doc/io.rst
rabernat: Anyone else have thoughts on this? @jhamman?

jhamman: My vote would be for using the same keyword in both contexts. So +1 for `consolidated`.

xarray/backends/zarr.py
```diff
@@ -234,15 +235,27 @@ def open_group(cls, store, mode='r', synchronizer=None, group=None):
                                       "installation "
                                       "http://zarr.readthedocs.io/en/stable/"
                                       "#installation" % min_zarr)
-        zarr_group = zarr.open_group(store=store, mode=mode,
-                                     synchronizer=synchronizer, path=group)
-        return cls(zarr_group)
 
-    def __init__(self, zarr_group):
+        if consolidated or consolidate_on_close:
+            if LooseVersion(zarr.__version__) <= '2.2':  # pragma: no cover
```
jhamman: reminder to update this version check too.

xarray/tests/test_backends.py
jhamman: I see. 2.3 should be out soon right? I would vote for keeping it simple (because life is short) and skipping everything before 2.3 for these tests.

xarray/backends/zarr.py
rabernat: Being more explicit about the version seems to fix this issue here. In the tests I have used the importorskip approach.

xarray/tests/test_backends.py
rabernat: Right now I am doing `pytest.importorskip('zarr', minversion="2.2.1.dev2")`. This is nice because it allows the new tests to be run in the zarr-dev travis environment today (https://travis-ci.org/pydata/xarray/jobs/460869777).

rabernat:
Not sure I understand why there are tests failing now. The failing function is `test_basic_compute`. 

https://travis-ci.org/pydata/xarray/jobs/460873430#L7489

At first glance, this does not appear to have anything to do with my PR. The relevant error is:
```

______________________________ test_basic_compute ______________________________
    def test_basic_compute():
        ds = Dataset({'foo': ('x', range(5)),
                      'bar': ('x', range(5))}).chunk({'x': 2})
        for get in [dask.threaded.get,
                    dask.multiprocessing.get,
                    dask.local.get_sync,
                    None]:
            with (dask.config.set(scheduler=get)
                  if LooseVersion(dask.__version__) >= LooseVersion('0.19.4')
                  else dask.config.set(scheduler=get)
                  if LooseVersion(dask.__version__) >= LooseVersion('0.18.0')
                  else dask.set_options(get=get)):
>               ds.compute()
xarray/tests/test_dask.py:843: 
xarray/core/dataset.py:597: in compute
    return new.load(**kwargs)
xarray/core/dataset.py:494: in load
    evaluated_data = da.compute(*lazy_data.values(), **kwargs)
../../../miniconda/envs/test_env/lib/python3.6/site-packages/dask/base.py:390: in compute
    collections=collections)
../../../miniconda/envs/test_env/lib/python3.6/site-packages/dask/base.py:865: in get_scheduler
    return get_scheduler(scheduler=config.get('scheduler', None))
get = None, scheduler = <function get at 0x7fc31d9ae048>, collections = None
cls = None
    def get_scheduler(get=None, scheduler=None, collections=None, cls=None):
        """ Get scheduler function
    
        There are various ways to specify the scheduler to use:
    
        1.  Passing in get= parameters (deprecated)
        2.  Passing in scheduler= parameters
        3.  Passing these into global confiuration
        4.  Using defaults of a dask collection
    
        This function centralizes the logic to determine the right scheduler to use
        from those many options
        """
        if get is not None:
            if scheduler is not None:
                raise ValueError("Both get= and scheduler= provided.  Choose one")
            warn_on_get(get)
            return get
    
        if scheduler is not None:
>           if scheduler.lower() in named_schedulers:
E           AttributeError: 'function' object has no attribute 'lower'
../../../miniconda/envs/test_env/lib/python3.6/site-packages/dask/base.py:854: AttributeError
```

shoyer:
I bet this is due to the latest dask release (1.0). We can fix this in another PR.

lilyminium:
I remember dealing with this in my pull request -- if I recall correctly scheduler was pointing to the scheduler.get function instead. It was a minor bug that was either fixed in the next release of xarray (0.11.0) or Dask (0.20.1).

rabernat:
So if the test issues can be considered resolved, the only decision we need to make is about the API.

Do we prefer (the current way):
```python
ds.to_zarr(fname, consolidate=True)
xr.open_zarr(fname, consolidated=True)
```

or @shoyer's suggestion
```python
ds.to_zarr(fname, consolidated=True)
xr.open_zarr(fname, consolidated=True)
```
???

xarray/backends/api.py
martindurant: Consolidate needs to be called after the metadata files are written, which must happen before the first data write, no? The remaining data writes can happen later, and data-sets opened via consolidated remain writeabel as far as data is concerned.

xarray/backends/api.py
martindurant: `zarr.consolidate_metadata` returns the output of open_consolidated on the same store, so this is already happening

martindurant:
Will the default for both options be `False` for the time being?

xarray/backends/api.py
rabernat: Ah so you're saying we could just call consolidate before doing any writes? That would certainly be simpler.

rabernat:
> Will the default for both options be `False` for the time being?

Yes

martindurant:
Glad to see this happening, by the way. Once in, catalogs using intake-xarray can be updated and I don't thin the code will need to change.

xarray/backends/api.py
martindurant: I *think* so, creating the array hierarchy should make the metadata without the need for any data files.

alimanfoo:
Great to see this. On the API, FWIW I'd vote for using the same keyword (``consolidated``) in both, less burden on the user to remember what to use.

rabernat:
Keywords are now all `consolidated` and all tests are go.

Ready to merge?

doc/whats-new.rst
```diff
@@ -36,6 +36,8 @@ Breaking changes
 Enhancements
 ~~~~~~~~~~~~
 
+- Ability to read and write consolidated metadata in zarr stores.
+  By `Ryan Abernathey <https://github.com/rabernat>`_.
```
jhamman: Can you reference the issue this is attached to: ```(:issue:`2558`).```

xarray/backends/zarr.py
```diff
@@ -234,15 +235,27 @@ def open_group(cls, store, mode='r', synchronizer=None, group=None):
...
-    def __init__(self, zarr_group):
+        if consolidated or consolidate_on_close:
+            if LooseVersion(zarr.__version__) <= '2.2.1.dev2':  # pragma: no cover
+                raise NotImplementedError("Zarr version 2.2.1.dev2 or greater "
+                                          "is required by for consolidated "
+                                          "metadata.")
+
+        open_kwargs = dict(mode=mode, synchronizer=synchronizer, path=group)
+        if consolidated:
+            # TODO: an option to pass the metadata_key keyword
```
jhamman: do we need to consider this TODO here?

jhamman: Anything to do here now?

martindurant:
LGTM

Do you think there should be more explicit text of how to add consolidation to existing zarr/xarray data-sets, rather than creating them with consolidation turned on?

We may also need some text around updating consolidated data-sets, but that can maybe wait to see what kind of usage people try.

xarray/backends/zarr.py
rabernat: Do we feel that it's important to expose this functionality from within xarray? I don't.

martindurant: I also don't.
I think it's ok for xarray to have an opinion on what the special key is called.

rabernat:
> We may also need some text around updating consolidated data-sets, but that can maybe wait to see what kind of usage people try.

Since xarray cannot append or modify in-place existing zarr stores, this seems outside the scope of xarray for now. But maybe it is worth mentioning in the docs.

xarray/backends/zarr.py
rabernat: I propose we just leave these TODO's here as is. If anyone ever needs this feature from the xarray side, this will help guide them on how to implement it.

jhamman:
I'm happy here. ...but Appveyor is not.

shoyer:
@rabernat if you're ready, let's merge this.

The failures on Appveyor are related (an issue with int32 and cftime)

rabernat:
👍

rabernat:
If anyone wants to see how awesome consolidated metadata is, you can try it in this binder:
https://github.com/rabernat/pangeo_ecco_examples/

I did a bit of lazy profiling here:
https://gist.github.com/rabernat/ce1fb414cf53541afe2245363b06c49d

Things that used to take ~40s now take ~1s. Especially since loading the data is one of the first steps in any pangeo notebook, this is a *huge* improvement in usability.

Thanks to everyone who helped make it happen!

martindurant:
I like those timings.

Issue #2565, opened 2018-11-23
CF conventions: time_bnds and time units 


fmaussion:
### Problem

Here is the dump of a NetCDF file ([download](https://github.com/OGGM/oggm-sample-data/raw/master/test-files/cesm.TREFHT.160001-200512.selection.nc)):

```
netcdf cesm.TREFHT.160001-200512.selection {
dimensions:
        time = UNLIMITED ; // (4872 currently)
        lat = 3 ;
        lon = 3 ;
        nbnd = 2 ;
variables:
        float TREFHT(time, lat, lon) ;
                TREFHT:units = "K" ;
                TREFHT:long_name = "Reference height temperature" ;
                TREFHT:cell_methods = "time: mean" ;
        double lat(lat) ;
                lat:long_name = "latitude" ;
                lat:units = "degrees_north" ;
        double lon(lon) ;
                lon:long_name = "longitude" ;
                lon:units = "degrees_east" ;
        double time(time) ;
                time:long_name = "time" ;
                time:units = "days since 0850-01-01 00:00:00" ;
                time:calendar = "noleap" ;
                time:bounds = "time_bnds" ;
        double time_bnds(time, nbnd) ;
                time_bnds:long_name = "time interval endpoints" ;

// global attributes:
                :Conventions = "CF-1.0" ;
                :source = "CAM" ;
...
}
```

When xarray decodes the time coordinates it also deletes the `time:units` attribute (this kind of makes sense, because the unit has no meaning when the time is converted to a CFTime object):

```python
import xarray as xr
ds = xr.open_dataset(f)
ds.time
<xarray.DataArray 'time' (time: 4872)>
array([cftime.DatetimeNoLeap(1600, 2, 1, 0, 0, 0, 0, 0, 32),
       cftime.DatetimeNoLeap(1600, 3, 1, 0, 0, 0, 0, 0, 60),
       cftime.DatetimeNoLeap(1600, 4, 1, 0, 0, 0, 0, 3, 91), ...,
       cftime.DatetimeNoLeap(2005, 11, 1, 0, 0, 0, 0, 6, 305),
       cftime.DatetimeNoLeap(2005, 12, 1, 0, 0, 0, 0, 1, 335),
       cftime.DatetimeNoLeap(2006, 1, 1, 0, 0, 0, 0, 4, 1)], dtype=object)
Coordinates:
  * time     (time) object 1600-02-01 00:00:00 ... 2006-01-01 00:00:00
Attributes:
[...]

PR #2571, opened 2018-11-24
CF: also decode time bounds when available

fmaussion:
- [x] Closes https://github.com/pydata/xarray/issues/2565
 - [x] Tests added 

Not sure if this is the best way to handle it, but it seems to work


xarray/conventions.py
```diff
@@ -350,6 +350,25 @@ def stackable(dim):
...
+        for v in variables.values():
+            attrs = v.attrs
+            if ('units' in attrs and 'since' in attrs['units'] and
+                    'bounds' in attrs):
+                new_attrs = {'units': attrs['units']}
+                if 'calendar' in attrs:
+                    new_attrs['calendar'] = attrs['calendar']
+                to_update[attrs['bounds']] = new_attrs
+        # For all bound variables, update their time attribute if not present
+        for k, new_attrs in to_update.items():
```
spencerkclark: I think we could update the attributes of the potential bounds variables immediately within the `for v in variables.values()` loop rather than use another loop here (this would remove the need for the `to_update` dictionary).

xarray/conventions.py
```diff
@@ -350,6 +350,25 @@ def stackable(dim):
         drop_variables = []
     drop_variables = set(drop_variables)
 
+    # Time bounds coordinates might miss the decoding attributes
+    # https://github.com/pydata/xarray/issues/2565
```
spencerkclark: Would it also make sense to add a comment with the URL for the ["Cell Boundaries" section](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-boundaries) of the CF conventions here?

xarray/conventions.py
```diff
@@ -350,6 +350,25 @@ def stackable(dim):
...
+    if decode_times:
+        # We list all time variables with bounds
+        to_update = dict()
+        for v in variables.values():
+            attrs = v.attrs
+            if ('units' in attrs and 'since' in attrs['units'] and
```
spencerkclark: It might be cleaner to write something like:
```python
has_date_units = 'units' in attrs and 'since' in attrs['units']
if has_date_units and 'bounds' in attrs:
```

shoyer:
Could we move this logic into a separate helper function? That would make things a little better organized and easier to test.

fmaussion:
I addressed all comments but the fixtures, which seemed a bit overkill (but I simplified the tests). Thanks!

fmaussion:
Thanks! Made the change as requested. Regarding the general design:

> With this implementation, if one decodes the times, and then saves things back out to a file, a "time bounds" variable will contain units and calendar attributes even though it might not have them initially. I'm not sure if this is a big deal (in this case it is simply adding redundant metadata), but I just wanted to bring it up in case it was a concern.

I agree - I don't think it is a big deal either. It must also be noted that this might break some code in downstream libraries (including mine) which has built workarounds for this and will now error because the bounds are already decoded. Here also, I'm quite confident that this is an edge case but its worth mentioning. Should I put my what's new in "Breaking changes" instead?

spencerkclark:
> It must also be noted that this might break some code in downstream libraries (including mine) which has built workarounds for this and will now error because the bounds are already decoded. Here also, I'm quite confident that this is an edge case but its worth mentioning. Should I put my what's new in "Breaking changes" instead?

I for sure see this perspective.  I also think a plausible case could be made that this change is like a "bug fix," that is something that people may have needed to work around before in various ways, but ultimately should not have needed to.  So I think it's up to you what you think is best.  

If you do decide to shift it to the breaking changes section, I would suggest being a little more specific about under what circumstances the behavior is changing (i.e. this only impacts the behavior for time bounds variables defined via CF conventions that do not already have "units" and "calendar" attributes).

shoyer:
+1 for putting this in "Breaking changes" (we try to be pretty conservative with the definition of a strict "Enhancement"), but otherwise this looks good to me.

fmaussion:
> +1 for putting this in "Breaking changes" 

Sorry for being so slow. I just done it, will merge tomorrow once the tests pass.

PR #2572, opened 2018-11-24
DOC: remove example using Dataset.T

shoyer:
as reported on the mailing list


dcherian:
Oops, missed that. Thanks.

dcherian:
do we need a `ds.transpose()` example to show that args are unnecessary?

shoyer:
We documented `ds.transpose()` on the computation page, too, so I don't think it's really necessary here.

shoyer:
actually -- maybe that's a good idea after all.

PR #2573, opened 2018-11-25
python setup.py test now works by default

horta:
You don't need to have pytest-runner before-hand.


shoyer:
works for me -- thanks!

PR #2574, opened 2018-11-25
Remove py37 collections.abc deprecation warnings

horta:


shoyer:
Thanks @horta. We are still supporting Python 2.7 through the end of this year, so this will either need to be backwards compatible or hold off another month until 2019 to be merged.

jhamman:
I've included these changes into #2645. Tanks for the contribution @horta.

PR #2593, opened 2018-12-05
CFTimeIndex Resampling

jwenfai:
I would appreciate some feedback on why this implementation for CFTimeIndex resampling doesn't match pandas' output 100%.

- [x] Tentative attempt at addressing #2191 (resampling CFTimeIndex). Downsampling produces results that match pandas' results in all the tests done thus far. However, upsampling has trouble assigning the right values to the right bins.
- [x] Tests (`test_cftimeindex_resample.py`) created for standard calendars but not for non-standard calendars (360 days etc.). Contents from `test_cftimeindex_resample.py` will be inserted into `test_dataarray.py` and  `test_cftimeindex_resample.py` will be deleted once resampling implementation is finalized. Files found in the `tests/temp` folder is meant to highlight resampling discrepancies between pandas and this implementaion. Both the files and the folder will be removed once resampling implementation is finalized.
 - [x] Not fully documented. Docstrings have yet to be added to `resample_cftime.py`. There were no doctrings from https://github.com/pandas-dev/pandas/blob/master/pandas/core/resample.py (which is where the codes were ported from) that could be conveniently copied.


xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,134 @@
...
+
+def series(index):
+    return pd.Series(np.arange(100., 100.+index.size), index=index)
+
+
+@ pytest.mark.parametrize(('closed', 'label', 'freq'),
+                          list(itertools.product(
+                              ['left', 'right'],
+                              ['left', 'right'],
+                              ['2MS', '2M', '3MS', '3M', '7MS', '7M'])))
```
max-sixty: Brief note on the test approach - overall looks great - on these items, you can put each set of params in its own `parametrize` decorator, and avoid making the products yourself

(no need to change this time though)

xarray/tests/temp/cftime_resample_pandas_comparison.py
```diff
@@ -0,0 +1,118 @@
...
+print(ps.resample('MS').mean())  # essentially doing no resampling, should return original data
+print(ps.resample('6MS').mean())
+print(ps.resample('6MS').asfreq())  # results do not match since xarray makes asfreq = mean (see resample.py)
+print(ps.resample('6MS').sum())
+print(ps.resample('6MS').min())
+print(ps.resample('6MS').max())
+
+
+# Upsampling comparisons:
+# At seconds-resolution, xr.cftime_range is 1 second off from pd.date_range
```
spencerkclark: It would be great if you could open a separate issue for this with a minimal working example; this was not intentional and I think we would consider it a bug.

xarray/tests/test_cftimeindex_resample.py
jwenfai: Thanks for the comment, I made the change, much better than the torturous code I was writing before!

jwenfai:
@spencerkclark Thanks for the detailed review! I'll fix up my code over the next few days.

I haven't completely solved the upsampling issue yet but I think I might have some clues as to what's happening. Timedelta operations on cftime.datetime does not always return correct values. Sometimes, they are a few microseconds or one second off.

The issue can be sidestepped by shifting the the bins 1 second forward for `closed=='right'` and 1 second back for `closed=='left'` in groupby.py, but this obviously introduces issues for resampling operations at the second and microsecond resolution. This workaround doesn't pass all the tests. An extra time bin is still sometimes created. You'll see what I mean when I make a new commit sometime next week.

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,136 @@
...
+# from xarray.tests import assert_array_equal, assert_identical
+
+@pytest.fixture()
+def pd_index():
+    return pd.date_range('2000-01-01', periods=30, freq='MS', tz='UTC')
+
+
+@pytest.fixture()
+def xr_index():
+    return xr.cftime_range('2000-01-01', periods=30, freq='MS', tz='UTC')
```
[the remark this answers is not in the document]
spencerkclark: Sorry about that!  Indeed that signature is misleading; see #2613 for a fix.

xarray/plot/plot.py
```diff
@@ -14,7 +14,7 @@
 import numpy as np
 import pandas as pd
 
-from xarray.core.alignment import align
+# from xarray.core.alignment import align
```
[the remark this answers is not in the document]
jwenfai: I don't recall working on the plot.py file. It seems to be an unused import statement so that might've been why it's been commented out.

spencerkclark:
@jwenfai thanks for the updates.  It looks like there are some merge conflicts that are preventing our CI from running.  Could you please resolve those when you get chance, so we can see those results?

jwenfai:
@spencerkclark Sorry about that, must be a couple of commits behind. Should be resolved now.

spencerkclark:
Yes, I noticed this too.  I think it's related to changes made here: https://github.com/pandas-dev/pandas/pull/24347.  At least in the test cases that I've run, I've only seen it make a difference in the NaN placement at the end of the time series.

For example with pandas 0.23.4:
```
In [1]: import xarray as xr; import pandas as pd

In [2]: nptimes = pd.date_range('2000', periods=2000)

In [3]: nptime_da = xr.DataArray(range(2000), [('time', nptimes)])

In [4]: nptime_da.resample(time='4AS').mean('time')
Out[4]:
<xarray.DataArray (time: 3)>
array([ 730., 1730.,   nan])
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2004-01-01 2008-01-01
```
and with pandas master:
```
In [4]: nptime_da.resample(time='4AS').mean('time')
Out[4]:
<xarray.DataArray (time: 2)>
array([ 730., 1730.])
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2004-01-01
```
I feel like the result under pandas master makes more sense, given the input array, whose final date is 2005-06-22.  Adding an additional bin labeled 2008-01-01 to the resampled time series seems superfluous given that with the default `label='left'` it's clear no data from the original time series would fall in that bin.

spencerkclark:
So I think the way you have things written now in `_get_time_bins` makes sense, though this subtle change in behavior on the pandas side makes testing against different releases of pandas a little trickier.

jwenfai:
Glad to see that I'm not the only one getting different results. And I agree (biased as I am) that the additional bin at the end of resampled time series is superfluous.

If pandas master with the altered resampling logic will be the definitive version going forward, should development of CFTimeIndex resampling be suspended until this version of pandas master is released and xarray uses it as a dependency?

On a somewhat related note, looking over the latest pandas master resample.py made me realize that
https://github.com/pydata/xarray/blob/4317c697900c80604dee793ffc1186e5c57a03fd/xarray/core/resample_cftime.py#L114-L118
is now wrong due to changes made 16 days ago (https://github.com/pandas-dev/pandas/issues/24127) . Since non-integer offset frequencies are not supported by BaseCFTimeOffset, (`is_day and offset.n == 1` should just be `is_day`), which can be further simplified to:
```python
    if isinstance(offset, CFTIME_TICKS):
        return _adjust_dates_anchored(first, last, offset,
                                      closed=closed, base=base)
```

spencerkclark:
> If pandas master with the altered resampling logic will be the definitive version going forward, should development of CFTimeIndex resampling be suspended until this version of pandas master is released and xarray uses it as a dependency?

I think it is OK to continue working here, with the aim of implementing resample for cftime that is closest to the most up-to-date pandas version.  We can add code to skip tests that depend on this behavior if the version of pandas is not recent enough (see an example of doing this [here](https://github.com/pydata/xarray/blob/2223445905705162053756323dfd6e6f4527b270/xarray/tests/test_dask.py#L848-L849)).  As we're working before version 0.24 of pandas is released we can keep an eye on our "py36-pandas-dev" CI build listed under the "Allowed Failures" section in Travis.

xarray/core/groupby.py
```diff
@@ -234,7 +234,45 @@ def __init__(self, obj, group, squeeze=False, grouper=None, bins=None,
                 # TODO: sort instead of raising an error
                 raise ValueError('index must be monotonic for resampling')
             s = pd.Series(np.arange(index.size), index)
-            first_items = s.groupby(grouper).first()
+            if isinstance(grouper, tuple):
```
spencerkclark: I've been thinking about this logic over the last few days.  With some minor modifications to the `_get_time_bins` logic, I think there may be a simpler way of determining this `first_items` Series without distinguishing between downsampling and upsampling.  I'll try to post a gist and some more comments later today or tomorrow.

spencerkclark:
So the crux of the problem now seems to be in generating `first_items`; this is the Series that is used for both upsampling and downsampling a DataArray in xarray.  For data indexed by a DatetimeIndex, it is straightforward to generate this Series (it just takes the construction of a simple Series with `np.arange` and the reference index, the construction of a `pandas.Grouper` object, and a call to `groupby` with the method `first`).  In xarray, downsampling uses both the values (to define the groups) and index (to define the labels) of this Series, while upsampling only uses the index.

For data indexed by a CFTimeIndex, we do not have the luxury of a formal `Grouper` object; however, if we can create this `first_items` Series accurately, I think all other results of resample in xarray should follow.

I've put together a [gist](https://gist.github.com/spencerkclark/27ef5254434fc5c90e2ebbdd975930b9#file-test_first_items_current-py) which compares the `first_items` Series generated with pandas with that generated by the cftime logic (the [output](https://gist.github.com/spencerkclark/27ef5254434fc5c90e2ebbdd975930b9#file-test_first_items_current-out) of running the tests is also included).  I've tried to use a fairly challenging set of initial time indexes as well as resample frequencies (different than what are currently used in the tests); there appear to be many mismatches under the "upsampling" case, but also a few errors show up in the "downsampling" case (to some extent I think these are related to the omission of the `_adjust_bin_edges` method, which it turns out I do think we may need).  In theory though, because of how this `first_items` Series is created in the DatetimeIndex case, I don't think the way we create it in the CFTimeIndex case should depend on whether the length of the reference index is greater than or less than the length of the resample labels (upsampling or downsampling is determined instead by the resampling method used).

This inspired the alternative solution proposed in the [second part of the gist](https://gist.github.com/spencerkclark/27ef5254434fc5c90e2ebbdd975930b9#file-test_first_items_simplified-py) (I've also added back in a call to a cftime version of the `_adjust_bin_edges` method); in this case there is no dependence on the relative lengths of the reference index and resample labels, and all of the test cases I've tried so far pass.

Let me know if this alternative solution makes sense.  Digging in to the guts of the resample code in pandas/xarray is still fairly new for me too, so I could be missing something.  In the gist I'm using this branch of xarray, the development version of pandas, and the latest version of cftime.  Thanks again for your hard work on this!

jwenfai:
I'm on a break right now and I'll look more closely at the alternative solution when I'm back, but from what I've read in your comment the solution makes sense. Also, nice job catching why `_adjust_bin_edges` was needed and writing a comprehensive explanation in the code.

xarray/core/resample.py
```diff
@@ -194,14 +194,33 @@ def _interpolate(self, kind='linear'):
...
 
-        x = self._obj[self._dim].astype('float')
+        from ..coding.cftimeindex import CFTimeIndex
+        from .utils import datetime_to_numeric
+        import cftime as cf
+        import numpy as np
+        if isinstance(self._obj.indexes[self._dim], CFTimeIndex):
+            t = self._obj[self._dim]
+            x = datetime_to_numeric(t, datetime_unit='s')
+            x = x.round()
```
spencerkclark: One other minor thing -- I'm inclined not to round here.  I realize that in the climate community, with non-standard calendars, it would be pretty unusual to encounter dates with a time resolution finer than one second.  However in xarray my sense is we should avoid making assumptions about this.

Fortunately with recent updates in cftime, this is becoming less and less of an issue.  See the difference in behavior here between cftime 1.0.0 and cftime 1.0.3.4 (the latest version):

#### cftime 1.0.0

Here differences in datetime arithmetic precision lead to small differences in the interpolation results between data indexed by a CFTimeIndex and data indexed by a DatetimeIndex.

```
In [1]: import numpy as np
i
In [2]: import xarray as xr

In [3]: times = xr.cftime_range('1892-01-01T12:00:00', periods=15, freq='5256113T')

In [4]: da_cftime = xr.DataArray(range(15), [('time', times)])

In [5]: da_pandas = xr.DataArray(range(15), [('time', times.to_datetimeindex())])

In [6]: new_times = xr.cftime_range('1892', periods=10, freq='6AS-JUN')

In [7]: da_cftime_interp = da_cftime.interp(time=new_times)

In [8]: da_pandas_interp = da_pandas.interp(time=new_times.to_datetimeindex())

In [9]: np.testing.assert_array_equal(da_cftime_interp.values, da_pandas_interp.values)
<ipython-input-9-9bc960239b46> in <module>
----> 1 np.testing.assert_array_equal(da_cftime_interp.values, da_pandas_interp.values)

//anaconda/envs/xarray-dev-37/lib/python3.6/site-packages/numpy/testing/_private/utils.py in assert_array_equal(x, y, err_msg, verbose)
    863     __tracebackhide__ = True  # Hide traceback for py.test
    864     assert_array_compare(operator.__eq__, x, y, err_msg=err_msg,
--> 865                          verbose=verbose, header='Arrays are not equal')
    866
    867

//anaconda/envs/xarray-dev-37/lib/python3.6/site-packages/numpy/testing/_private/utils.py in assert_array_compare(comparison, x, y, err_msg, verbose, header, precision, equal_nan, equal_inf)
    787                                 verbose=verbose, header=header,
    788                                 names=('x', 'y'), precision=precision)
--> 789             raise AssertionError(msg)
    790     except ValueError:
    791         import traceback

AssertionError:
Arrays are not equal

(mismatch 60.0%)
 x: array([0.041506, 0.641767, 1.242028, 1.842289, 2.442824, 3.043085,
       3.64362 , 4.243881, 4.844416, 5.444677])
 y: array([0.041506, 0.641767, 1.242028, 1.842289, 2.442824, 3.043085,
       3.64362 , 4.243881, 4.844416, 5.444677])
```
Note that we can use `assert_allclose` with the default tolerances here:
```
In [10]: np.testing.assert_allclose(da_cftime_interp.values, da_pandas_interp.values)
```
#### cftime 1.0.3.4

For this example at least, the issues with datetime arithmetic precision have been resolved.

```
In [1]: import numpy as np

In [2]: import xarray as xr

In [3]: times = xr.cftime_range('1892-01-01T12:00:00', periods=15, freq='5256113T')

In [4]: da_cftime = xr.DataArray(range(15), [('time', times)])

In [5]: da_pandas = xr.DataArray(range(15), [('time', times.to_datetimeindex())])

In [6]: new_times = xr.cftime_range('1892', periods=10, freq='6AS-JUN')

In [7]: da_cftime_interp = da_cftime.interp(time=new_times)

In [8]: da_pandas_interp = da_pandas.interp(time=new_times.to_datetimeindex())

In [9]: np.testing.assert_array_equal(da_cftime_interp.values, da_pandas_interp.values)
```

Any errors in interpolation due to microsecond noise in the ordinal coordinate would be expected to be small (I think we could account for this in the tests by using [`xarray.testing.assert_allclose`](http://xarray.pydata.org/en/stable/generated/xarray.testing.assert_allclose.html), which checks that arrays are equal up to a specified tolerance).

spencerkclark:
> I'm on a break right now and I'll look more closely at the alternative solution when I'm back, but from what I've read in your comment the solution makes sense.

No worries!  I just wanted to get these thoughts down when I had some time.  No rush to make any updates.

jwenfai:
Hi @spencerkclark, sorry it took so long to get back to you. I've implemented your simplified resampling logic. Some of the logic had to be altered since pandas have made updates.

It's great not having to delineate between upsampling/downsampling cases! I ran into some issues though and I thought maybe an extra pair of eyes could help me diagnose them:

1. **cftime** : Not really important but I cannot reproduce the results you obtained for cftime 1.0.3.4. I've tried Python 2.7 and 3.6, conda packages and also building from source, Windows machine and the Windows Ubuntu shell --- datetime arithmetic precision problem persists. To work around this issue, I'm using `assert_allclose` with default tolerances on the tests as suggested.

2. **pandas** : The pandas library refuses to resample certain indices and throws a "values falls before first bin" error. The error comes from `bins=lib.generate_bins_dt64(...)` around line 1400 of `pandas/core/resample.py` and is a direct consequence of the `_adjust_bin_edges` operation adding 1 extra day minus 1 nanosecond causing the first value of sorted `bin_edges` to be larger than the first sorted `ax_values`. My current workaround is to use `pytest.mark.xfail(raises=ValueError)`.

   CFTimeIndex resampling does not encounter the same error. Nevertheless, I've changed the CFTimeIndex resampling logic so that the first bin value does not have 1 day minus 1 microsecond added to it to (hopefully) rectify the error. Testing against pandas resampling results does not show any difference between the corrected and uncorrected CFTimeIndex resampling code.

3. **xarray** : Ignoring the aforementioned issue with pandas, xarray resampling results for certain time ranges do not match pandas', specifically these two: `dict(start='1892-01-01T12:00:00', periods=15, freq='5256113T'),` labeled `XT`, and `dict(start='1892', periods=10, freq='6AS-JUN')`, labeled `6AS_JUN`. `XT` seems to be causing the most problem, which might be due to its rather challenging `freq` specification.

   Since I've rewritten `test_cftimeindex_resample.py` based on your gists, a lot more test cases are being generated. Without `XT` and `6AS_JUN`, the tests take about 40 minutes to run on my machine; including them bumps that time up to 3 hours. The number of tests should be pared down prior to merging but I think they're helpful right now for identifying problems. I've included test results in XML for you and other collaborators to compare against. One file contains the results with the 1 day minus 1 microsecond fix applied and the other is without the fix. They can be imported into PyCharm, but I'm not sure if they can be read any other way.
[Test Results - pytest_in_test_cftimeindex_resample_py.zip](https://github.com/pydata/xarray/files/2766010/Test.Results.-.pytest_in_test_cftimeindex_resample_py.zip)

xarray/core/resample.py
```diff
@@ -194,14 +194,27 @@ def _interpolate(self, kind='linear'):
                 .format(self._obj.data.name)
             )
 
-        x = self._obj[self._dim].astype('float')
+        from ..coding.cftimeindex import CFTimeIndex
```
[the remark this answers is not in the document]
jwenfai: Sorry, I'm a bit new at this. I assume you mean copy all the changes from 3 changed files https://github.com/pydata/xarray/pull/2640 using checkout --patch or something similar?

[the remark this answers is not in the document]
spencerkclark: No worries -- typically in this situation I'll do something like this.  First I'll update the master branch locally by fetching changes from the remote upstream repository (i.e. this one):
```
$ git fetch upstream
```
Then I'll merge the master branch in:
```
$ git merge upstream/master
```
It looks like you may have a few merge conflicts to sort out (which you'll need to resolve by opening each file with a conflict and editing it manually, see [here](https://help.github.com/articles/resolving-a-merge-conflict-using-the-command-line/) for more details on how to do that).

This assumes you've set up a remote called `upstream` for this repository:
```
$ git remote add upstream https://github.com/pydata/xarray.git
$ git remote -v show upstream
* remote upstream
  Fetch URL: https://github.com/pydata/xarray.git
  Push  URL: https://github.com/pydata/xarray.git
  HEAD branch: master
  Remote branches:
    0.11.2-release new (next fetch will store in remotes/upstream)
    fix-docs       new (next fetch will store in remotes/upstream)
    master         tracked
    stable         new (next fetch will store in remotes/upstream)
  Local ref configured for 'git push':
    master pushes to master (local out of date)
```

jwenfai:
> Interesting, I'm not sure what's leading to the difference in behavior between platforms (I'm on a Mac). If you can distill the precision discrepancy to a minimal example, it might be worth reporting upstream to cftime. Though I agree, from xarray's perspective this is probably not a major concern.

Could this issue be related to the linear algebra library? A minimal example would be the [earlier example you gave](https://github.com/pydata/xarray/pull/2593/#discussion_r244641645). For both cftime 1.0.0 and 1.0.3.4, I get "mismatch 60.0%".

> The logic I wrote in `CFTimeGrouper.first_items` was in fact based in part on `lib.generate_bins_dt64`. Indeed I omitted [these checks related to whether the time series values were all within the bin edges or not](https://github.com/pandas-dev/pandas/blob/512830b46e678a1c6bd86cc60f54d74ff75792db/pandas/_libs/lib.pyx#L671-L676). I think it would actually be straightforward to add these in just before the call to `np.searchsorted` (in my code `datetime_bins` corresponds to `binner` in pandas and `index` corresponds to `values`). I would be inclined to do this rather than try to deviate from pandas' behavior here for now.

So I keep pandas' logic (the first bin has 1 day and 1 microsecond added to it) and raise a Value Error when either `index[0] < datetime_bins[0]` or ` index[lenidx - 1] > datetime_bins[lenbin - 1]`.

> Which version of pandas are you testing against? I'm testing against the dev version and with your branch I can't seem to reproduce any failures that are not related to the "values falls before first bin" error in the `XT` case. I have not tried any of the `6AS_JUN` cases yet.

I'm testing against the dev version, 11 commits behind.  Could the errors for `XT` that I get but you don't be due to cftime/linear algebra library issue? There may be enough error accumulated for hourly frequencies over 140 years that cftime_range generates an extra bin compared to pandas date_range (haven't checked all manually but I believe the majority of the non-"values falls before first bin" errors are due to extra bin(s)). `6AS_JUN` only has 8 failed tests all due to "x and y nan location mismatch".

> Yeah that's too long :). I agree we'll eventually need to pare down the tests. A big reason for this is the length of the resampled time series in some of the cases (e.g. hourly frequencies over a 140 year period produce very long arrays, e.g. in the `XT` case). There's also probably a lot of overlap between the existing test cases as far as the logic they exercise (e.g. do we need to test against frequencies '2H', '5H', '7H', '12H', and '8001H' or would just '7H' suffice etc.?).

We probably don't. I forget the reason but early on in development, resampling tests failed for some time ranges when using purely odd frequencies while others failed with purely even ones. Resampling tests for12H/24H frequencies might not be needed now that `_adjust_bin_edges' is being used.

> While it's definitely good to test directly against pandas for some examples, in place of some of those tests, we might want to consider writing some targeted unit tests for the methods in `resample_cftime.py`, like `_get_time_bins`, `CFTimeGrouper.first_items`, etc. With those it would probably be easier to write some tests that use minimal data but exercise all the conditional logic.

I'll look into writing targeted unit tests.

xarray/core/resample.py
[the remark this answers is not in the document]
jwenfai: Thanks for the writeup, I was hoping that I could avoid manually inspecting merge conflicts.

spencerkclark:
> So I keep pandas' logic (the first bin has 1 day and 1 microsecond added to it) and raise a Value Error when either index[0] < datetime_bins[0] or index[lenidx - 1] > datetime_bins[lenbin - 1].

Exactly 👍

> I'm testing against the dev version, 11 commits behind. Could the errors for XT that I get but you don't be due to cftime/linear algebra library issue? There may be enough error accumulated for hourly frequencies over 140 years that cftime_range generates an extra bin compared to pandas date_range (haven't checked all manually but I believe the majority of the non-"values falls before first bin" errors are due to extra bin(s)). 6AS_JUN only has 8 failed tests all due to "x and y nan location mismatch".

I think the "values falls before first bin" errors are all from pandas, where datetime arithmetic is exact, so they could not be due to cftime, right?  I'll take a look at the 6AS-JUN tests.

spencerkclark:
> We probably don't. I forget the reason but early on in development, resampling tests failed for some time ranges when using purely odd frequencies while others failed with purely even ones. Resampling tests for12H/24H frequencies might not be needed now that `_adjust_bin_edges' is being used.

Yeah this has for sure been helpful for development -- I did not expect to encounter the "values falls before edges" error, but clearly it seems we do need to worry about it.

jwenfai:
> I think the "values falls before first bin" errors are all from pandas, where datetime arithmetic is exact, so they could not be due to cftime, right? I'll take a look at the 6AS-JUN tests.

Oh no, I meant that _except_ for all the "values falls before first bin" errors, most (if not all) of the errors are due to shape mismatch between the resampled cftime and pandas arrays. Of the ones I've inspected, the resampled cftime array always has 1 more bin than pandas, e.g.:
```
E       (shapes (175204,), (175205,) mismatch)
E        x: array([100.,  nan,  nan, ...,  nan,  nan, 114.])
E        y: array([100.,  nan,  nan, ...,  nan,  nan, 114.])
```
I was thinking that imprecise datetime arithmetic by cftime might cause an extra bin to be generated.

spencerkclark:
@jwenfai to provide some more detail for my confusion here -- my impression is that adding or subtracting a `datetime.timedelta` object from a `cftime.datetime` object to produce another `cftime.datetime` object should always be microsecond-exact.  [See comments in cftime here](https://github.com/Unidata/cftime/blob/ca6ddce6fbfddfb769de6e79947a4af4b471fc93/cftime/_cftime.pyx#L1644-L1657).  This is how a CFTimeIndex is constructed through `cftime_range`, so at least naively I would not anticipate any precision issues in constructing the bins.  

Taking the difference between two dates to produce a timedelta object takes [a different code path](https://github.com/Unidata/cftime/blob/master/cftime/_cftime.pyx#L1350) in cftime, which is not microsecond-precise.  This, as we've seen, can induce some small errors in interpolation (because in the process we need to determine the amount of time between each date in the time coordinate and a reference date).

> Of the ones I've inspected, the resampled cftime array always has 1 more bin than pandas

Could you provide more details about this example?  What were the resample parameters used (e.g. `freq`, `closed`, `label`, `base`)?  Is the extra bin added to the beginning or end of the time range?  If you convert to a DatetimeIndex, do all the other bins match exactly, or is there some error?

spencerkclark:
> The first and last values are returned by _adjust_bin_anchored when isinstance(offset, CFTIME_TICKS). Since date subtraction happens within _adjust_bin_anchored, some test cases have imprecise first and last values.

Ah indeed; this makes sense now.  Maybe we should bring this up in cftime to see what their recommendation might be?  I could imagine writing a function like this that would correct for this imprecision when taking the difference between two dates:
```python
from datetime import timedelta

def exact_cftime_datetime_difference(a, b):
    """Exact computation of b - a"""
    a_microseconds = a.microsecond
    b_microseconds = b.microsecond
    seconds = b.replace(microsecond=0) - a.replace(microsecond=0)
    seconds = int(round(seconds.total_seconds()))
    microseconds = b_microseconds - a_microseconds
    return timedelta(seconds=seconds, microseconds=microseconds)
```

Here are a couple test cases:
```python
import cftime

# Testing with cftime version 1.0.0 where datetime where I can
# reproduce precision issues.
test_cases = [
    [(2000, 1, 1, 0, 4, 0, 956321), (1892, 1, 3, 12, 0, 0, 112123)],
    [(2000, 1, 1, 0, 4, 0, 1), (1892, 1, 3, 12, 0, 0, 503432)],
    [(2000, 1, 1, 0, 4, 0, 999999), (1892, 1, 3, 12, 0, 0, 112123)],
    [(2000, 1, 1, 0, 4, 0, 11213), (1892, 1, 3, 12, 0, 0, 77777)],
]
for a_args, b_args in test_cases:
    a_cftime = cftime.DatetimeGregorian(*a_args)
    b_cftime = cftime.DatetimeGregorian(*b_args)
    a = datetime(*a_args)
    b = datetime(*b_args)

    expected = b - a
    result = exact_cftime_datetime_difference(a_cftime, b_cftime)
    assert result == expected
    
    inexact = b_cftime - a_cftime
    assert inexact != expected
    
    # Test other direction
    expected = a - b
    result = exact_cftime_datetime_difference(b_cftime, a_cftime)
    assert result == expected
```
But maybe I'm missing something important.

spencerkclark:
@jwenfai I created an issue in cftime regarding this question: Unidata/cftime#109.

jwenfai:
Thanks for raising the issue with the cftime devs. I've tested function and it produces exact datetime values but there's still a problem with extra bins and nan location mismatch. I'll report back once I have a clearer idea of what's going on.

jwenfai:
Ignore my last comment, I made a really silly mistake. I assumed it was  `a - b` instead of `b - a` for the exact cftime function. 5808 of 5920 tests now pass and the remaining 112 are ignored due to ValueError: "value falls before first bin".

I think writing targeted unit tests are the last thing on the agenda, so I'll get right on that.

spencerkclark:
> 5808 of 5920 tests now pass and the remaining 112 are ignored due to ValueError: "value falls before first bin".

That's great news!

> I think writing targeted unit tests are the last thing on the agenda, so I'll get right on that.

Before spending too much time on that just yet, see if you can resolve the merge conflicts, and if you can think about a way to reduce the length of the existing tests.  It would be helpful to see a coverage report generated by [coveralls](https://coveralls.io/github/pydata/xarray) for the new logic you've added (if you resolve the merge conflicts our CI here will run and we'll be able to see that automatically).  Maybe start by commenting out a bunch of the really long ones and see where things stand?

jwenfai:
Just saw your comment. I've actually resolved the merge conflicts on my end and haven't pushed them yet. Checks for `index[0] < datetime_bins[0]` and `index[lenidx - 1] > datetime_bins[lenbin - 1]` have also been added.

The `test_resampler` runtime has been reduced to 1 minute and it seems like it has decent coverage since `ValueError`s are still being raised. Wrote another test for `_get_range_edges` that might help cut down the number of cases that needs to be tested with the slower `test_resampler` function.

Going to push my changes and see what CI says.

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,92 @@
...
+
+@pytest.mark.parametrize('freq', [
+    '700T', '8001T',
+    '12H', '8001H',
+    '3D', '8D', '8001D',
+    '2MS', '2M', '3MS', '3M', '4MS', '4M',
+    '3AS', '3A', '4AS', '4A'])
+@pytest.mark.parametrize('closed', ['left', 'right'])
+@pytest.mark.parametrize('label', ['left', 'right'])
+@pytest.mark.parametrize('base', [17, 24])
```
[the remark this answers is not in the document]
spencerkclark: Based on the [pandas documentation](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.resample.html), I wonder if it makes sense to test these values (17 and 24) for `base` for all the frequencies listed:
> For frequencies that evenly subdivide 1 day, the “origin” of the aggregated intervals. For example, for ‘5min’ frequency, base could range from 0 through 4. Defaults to 0

Does this mean we should only test non-zero base arguments for frequencies that evenly divide one day and only use values between zero and the frequency multiple?  Is that maybe why we're running into the "value falls before first bin" error in spots?

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,92 @@
...
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+import pandas.core.resample as pdr
+import xarray.core.resample_cftime as xrr
+from pandas.tseries.frequencies import to_offset as pd_to_offset
+from xarray.coding.cftime_offsets import (to_cftime_datetime,
+                                          to_offset as xr_to_offset)
+
```
spencerkclark: Please add a couple module-level skips at the top here for these tests (which depend on both cftime and pandas >= 0.24):
```python
pytest.importorskip('cftime')
pytest.importorskip('pandas', minversion='0.24')
```
This will prevent the tests from being run on platforms where they will be expected to fail (due to either import errors or version-dependent behavior).

xarray/tests/test_dataarray.py
```diff
@@ -2314,8 +2314,8 @@ def test_resample_cftimeindex(self):
                                 calendar='noleap')
         array = DataArray(np.arange(12), [('time', times)])
 
-        with raises_regex(NotImplementedError, 'to_datetimeindex'):
-            array.resample(time='6H').mean()
+        # with raises_regex(NotImplementedError, 'to_datetimeindex'):
+        array.resample(time='6H').mean()
```
spencerkclark: Good catch -- go ahead and remove the commented-out line too.

xarray/coding/cftime_offsets.py
```diff
@@ -209,7 +209,7 @@ def _shift_months(date, months, day_option='start'):
     # the returned date object in versions of cftime between 1.0.2 and
     # 1.0.3.4.  It can be removed for versions of cftime greater than
     # 1.0.3.4.
-    return date.replace(year=year, month=month, day=day, dayofwk=-1)
+    return date.replace(year=year, month=month, day=day)
```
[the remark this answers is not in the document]
jwenfai: Not removing it means that I get a `ValueError: Replacing the dayofyr or dayofwk of a datetime is not supported.` , which is an error message you wrote for cftime. So, I'm a little confused as to what I need to do to here.

xarray/tests/test_cftimeindex_resample.py
jwenfai: Changing this to 0.24 keeps me from running the test since my pandas-dev version of pandas is  0.24.0.dev0+1555.gea8c9bf51.dirty, would 0.23.99 be an acceptable workaround?

xarray/coding/cftime_offsets.py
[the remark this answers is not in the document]
spencerkclark: You must be using the dev version of cftime locally; this will not raise an error in cftime 1.0.3.4.  I agree we'll need to fix this once the new version of cftime is released, but for now let's leave it alone.

xarray/tests/test_cftimeindex_resample.py
spencerkclark: Sorry, yes, for now that would be ok.

xarray/tests/test_cftimeindex_resample.py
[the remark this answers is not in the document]
jwenfai: Still getting "value falls before first bin" error for `base=0`.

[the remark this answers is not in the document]
spencerkclark: > Does this mean we should only test non-zero base arguments for frequencies that evenly divide one day and only use values between zero and the frequency multiple?

`_adjust_dates_anchored` also only gets called for frequencies that are part of `CFTIME_TICKS` so essentially `base` will be ignored for non-tick frequencies.  So ignore my question here (I should have thought that through more).  I think things are fine here (I still think `base=0` should be tested though).  Sorry about that.

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,92 @@
...
+@pytest.mark.parametrize('freq', [
+    '700T', '8001T',
+    '12H', '8001H',
+    '3D', '8D', '8001D',
+    '2MS', '2M', '3MS', '3M', '4MS', '4M',
+    '3AS', '3A', '4AS', '4A'])
+@pytest.mark.parametrize('closed', ['left', 'right'])
+@pytest.mark.parametrize('label', ['left', 'right'])
+@pytest.mark.parametrize('base', [17, 24])
+@pytest.mark.xfail(raises=ValueError)
```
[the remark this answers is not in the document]
jwenfai: There seems to be a pattern in the error but I can't quite see it. Different `base` values clash with different `freq` values and appears influenced by whether the values are divisible by 4, by 6, by 4 and 6, is greater than 0 is less than 6, is greater than 18 but less than 24, and is greater than 24. The index itself doesn't influence whether an `ValueError` is raised.

I tested the combinations that can be generated from the following values:
```python
@pytest.mark.parametrize('freq', ['3D', '8D', '81D', '8001D'])
@pytest.mark.parametrize('closed', ['left', 'right'])
@pytest.mark.parametrize('label', ['left', 'right'])
@pytest.mark.parametrize('base', [2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 48, 240000])
```

The code block below shows the combinations that raise error:
```python
@pytest.mark.parametrize('freq', ['3D', '8D'])
@pytest.mark.parametrize('closed', ['right'])
@pytest.mark.parametrize('label', ['left', 'right'])
@pytest.mark.parametrize('base', [24, 48, 240000])
def test_value_error_1(freq, closed, label, base):
        cftime_index = xr.cftime_range(start='1892-01-03T12:07:01',
                                       periods=10, freq='3AS-JUN')
        da_cftime = da(cftime_index).resample(
            time=freq, closed=closed, label=label, base=base).mean()
        da_cftime['time'] = da_cftime.indexes['time'].to_datetimeindex()


@pytest.mark.parametrize('freq', ['3D', '8D', '81D', '8001D'])
@pytest.mark.parametrize('closed', ['right'])
@pytest.mark.parametrize('label', ['left', 'right'])
@pytest.mark.parametrize('base', [0])
def test_value_error_2(freq, closed, label, base):
        cftime_index = xr.cftime_range(start='1892-01-03T12:07:01',
                                       periods=10, freq='3AS-JUN')
        da_cftime = da(cftime_index).resample(
            time=freq, closed=closed, label=label, base=base).mean()
        da_cftime['time'] = da_cftime.indexes['time'].to_datetimeindex()


@pytest.mark.parametrize('freq', ['3D'])
@pytest.mark.parametrize('closed', ['right'])
@pytest.mark.parametrize('label', ['left', 'right'])
@pytest.mark.parametrize('base', [12])
def test_value_error_3(freq, closed, label, base):
        cftime_index = xr.cftime_range(start='1892-01-03T12:07:01',
                                       periods=10, freq='3AS-JUN')
        da_cftime = da(cftime_index).resample(
            time=freq, closed=closed, label=label, base=base).mean()
        da_cftime['time'] = da_cftime.indexes['time'].to_datetimeindex()


@pytest.mark.parametrize('freq', ['8D'])
@pytest.mark.parametrize('closed', ['right'])
@pytest.mark.parametrize('label', ['left', 'right'])
@pytest.mark.parametrize('base', [8, 16])
def test_value_error_4(freq, closed, label, base):
        cftime_index = xr.cftime_range(start='1892-01-03T12:07:01',
                                       periods=10, freq='3AS-JUN')
        da_cftime = da(cftime_index).resample(
            time=freq, closed=closed, label=label, base=base).mean()
        da_cftime['time'] = da_cftime.indexes['time'].to_datetimeindex()


@pytest.mark.parametrize('freq', ['3D'])
@pytest.mark.parametrize('closed', ['right'])
@pytest.mark.parametrize('label', ['left', 'right'])
@pytest.mark.parametrize('base', [6, 18])
def test_value_error_5(freq, closed, label, base):
        cftime_index = xr.cftime_range(start='1892-01-03T12:07:01',
                                       periods=10, freq='3AS-JUN')
        da_cftime = da(cftime_index).resample(
            time=freq, closed=closed, label=label, base=base).mean()
        da_cftime['time'] = da_cftime.indexes['time'].to_datetimeindex()
```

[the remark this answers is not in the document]
jwenfai: That's a neat solution, all tests are passing now.

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,63 @@
...
+                        coords=[index], dims=['time'])
+
+
+@pytest.mark.parametrize('freq', [
+    '700T', '8001T',
+    '12H', '8001H',
+    '3D', '8D', '8001D',
+    '2MS', '2M', '3MS', '3M', '4MS', '4M',
+    '3AS', '3A', '4AS', '4A'])
+@pytest.mark.parametrize('closed', ['left', 'right'])
```
spencerkclark: Maybe add `None` as an option here too (to test the default logic)?

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.mark.parametrize('label', ['left', 'right'])
```
spencerkclark: Maybe add `None` as an option here too (to test the default logic)?

xarray/core/resample_cftime.py
```diff
@@ -0,0 +1,250 @@
...
+
+    datetime_bins = _adjust_bin_edges(datetime_bins, freq, closed, index)
+
+    if closed == 'right':
+        if label == 'right':
+            labels = labels[1:]
+    elif label == 'right':
+        labels = labels[1:]
+
+    if index.hasnans:  # cannot be true since CFTimeIndex does not allow NaNs
```
spencerkclark: Thanks for noting this -- maybe instead of adding this code just yet, insert a `TODO` comment:
```python
# TODO: when CFTimeIndex supports missing values, if the reference index
# contains missing values, insert the appropriate NaN value at the
# beginning of the datetime_bins and labels indexes.
```

jwenfai:
Thanks for showing me how the code could be improved @spencerkclark. I've pushed the changes.

jwenfai:
I made the changes. Did not preview the .rst files but they should render correctly.

For the non-standard calendars, I can't think of any way to test them except to compare them against output from pandas. I think it's fine as long as the original and resampled indices do not go over 28 days.

spencerkclark:
One more optional thing -- support for the `loffset` keyword was recently added for DatetimeIndex resampling (https://github.com/pydata/xarray/pull/2608).  This just involves a simple adjustment to the bin labels (the index of the `first_items` Series).  You could implement that here, or for now just raise a `NotImplementedError` for resampling with a CFTimeIndex if `loffset` is not `None`.

jwenfai:
The pandas-dev build job is still failing but everything else is passing.

xarray/core/resample_cftime.py
```diff
@@ -0,0 +1,365 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from __future__ import absolute_import, division, print_function
```
spencerkclark: Another small thing -- since xarray is no longer worried about Python 2 compatibility, we can remove these imports (xref: #2645).

xarray/tests/test_cftimeindex_resample.py
spencerkclark: pandas 0.24 is out, so let's go ahead and update the `minversion` here.  The tests should now run on our main CI builds too.  Maybe wait until #2720 is merged before pushing, so you don't hit the unrelated failures.

xarray/core/resample_cftime.py
jwenfai: Removed the import statements. Same failures as before.

shoyer:
I just merged a fix for the test failures with pandas 0.24. If you merge in master that should fix your issues here too.

xarray/tests/test_cftimeindex_resample.py
```diff
@@ -0,0 +1,89 @@
...
+
+import pytest
+
+import datetime
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+pytest.importorskip('cftime')
+pytest.importorskip('pandas', minversion='0.23.99')
```
spencerkclark: In addition to merging in the master branch, could you please also update the `minversion` to 0.24 here now that pandas 0.24 is officially out?  Thanks!

jwenfai:
> 
> 
> I just merged a fix for the test failures with pandas 0.24. If you merge in master that should fix your issues here too.

Fix works, all tests are passing now. Thanks!

xarray/core/groupby.py
```diff
@@ -259,11 +259,23 @@ def __init__(self, obj, group, squeeze=False, grouper=None, bins=None,
...
-                first_items = first_items.dropna()
+                if isinstance(grouper, CFTimeGrouper):
+                    index_dict = dict(zip(np.arange(first_items.size),
+                                          first_items.index.values))
+                    first_items.index = np.arange(first_items.size)
+                    first_items = first_items.dropna()
+                    first_items.index = [index_dict[i] for i in
+                                         first_items.index.values]
+                else:
+                    first_items = first_items.dropna()
```
shoyer: I would suggest moving all this logic into a dedicate helper function that returns `full_index` and `first_items` -- possibly with entirely separate paths for CFTimeGrouper / pandas.Grouper. That would help keep the down complexity in `GroupBy.__init__`.

jwenfai: Is this what you meant:

```python
        if grouper is not None:
            index = safe_cast_to_index(group)
            if not index.is_monotonic:
                # TODO: sort instead of raising an error
                raise ValueError('index must be monotonic for resampling')
            s = pd.Series(np.arange(index.size), index)
            full_index, first_items = self._get_index_and_items(
                index, s, grouper)
```

```python
    @staticmethod
    def _get_index_and_items(index, s, grouper):
        from .resample_cftime import CFTimeGrouper
        if isinstance(grouper, CFTimeGrouper):
            first_items = grouper.first_items(index)
            full_index = first_items.index
            if first_items.isnull().any():
                index_dict = dict(zip(np.arange(first_items.size),
                                      first_items.index.values))
                first_items.index = np.arange(first_items.size)
                first_items = first_items.dropna()
                first_items.index = [index_dict[i] for i in
                                     first_items.index.values]
        else:
            first_items = s.groupby(grouper).first()
            _apply_loffset(grouper, first_items)
            full_index = first_items.index
            if first_items.isnull().any():
                first_items = first_items.dropna()
        return full_index, first_items
```

shoyer: @jwenfai Yes, something like that looks great! But please it make just a function -- usually static methods are unnecessary and a little confusing.

jwenfai: Pushed the changes, travis-ci build is failing and I can't diagnose what's wrong from the error message (something to do with conda it seems).

fmaussion:
> I can't diagnose what's wrong from the error message (something to do with conda it seems)

Some connection error. I restarted Travis, let's see if this happens again.

xarray/core/groupby.py
```diff
@@ -234,10 +234,22 @@ def __init__(self, obj, group, squeeze=False, grouper=None, bins=None,
...
             s = pd.Series(np.arange(index.size), index)
-            first_items = s.groupby(grouper).first()
+            from .resample_cftime import CFTimeGrouper
+            if isinstance(grouper, CFTimeGrouper):
+                first_items = grouper.first_items(index)
+            else:
+                first_items = s.groupby(grouper).first()
             full_index = first_items.index
             if first_items.isnull().any():
-                first_items = first_items.dropna()
```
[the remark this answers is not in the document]
jwenfai: Now that `dropna()` for CFTimeIndex indices are working (https://github.com/pydata/xarray/pull/2734), the logic here will be simplified once https://github.com/pydata/xarray/pull/2734 is merged with master.

shoyer:
It looks like tests are passing now. I'm going to give this another look over and then (probably) merge

xarray/core/groupby.py
```diff
@@ -259,11 +259,8 @@ def __init__(self, obj, group, squeeze=False, grouper=None, bins=None,
                 # TODO: sort instead of raising an error
                 raise ValueError('index must be monotonic for resampling')
             s = pd.Series(np.arange(index.size), index)
```
shoyer: Could you make this object `s` inside the helper function instead? It's not needed outside here

xarray/tests/test_formatting.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 from textwrap import dedent
 
+from textwrap import dedent
```
shoyer: It looks like changing from another PR have leaked in here? Let's try to figure that out...

jwenfai: Seems to have accidentally crept in when I merged changes from pydata/master into my local repo 4~5 days back. Here's what I managed to trace (from latest to earliest instance of test_formatting.py being changed):
https://github.com/Ouranosinc/xarray/pull/14
https://github.com/jwenfai/xarray/commit/31ccebf984c07d2fe337e5a5f6f589000346cade
https://github.com/jwenfai/xarray/commit/9fbb0162bf30374a712e6b74cac9e73fa544ae20

shoyer: It looks like it’s just a bad merge — these tests are now duplicated twice. You can simply delete the redundant code and push a new commit.

xarray/tests/test_dataarray.py
shoyer: You should just delete this entire test method -- you already wrote methods to test for resampling in the other file, so there's not much point in just verifying that it doesn't raise an error here.

xarray/core/groupby.py
```diff
@@ -310,6 +306,27 @@ def __len__(self):
     def __iter__(self):
         return zip(self._unique_coord.values, self._iter_grouped())
 
+    def _get_index_and_items(self, index, grouper):
+        from .resample_cftime import CFTimeGrouper
+        s = pd.Series(np.arange(index.size), index)
+        if isinstance(grouper, CFTimeGrouper):
+            first_items = grouper.first_items(index)
+            full_index = first_items.index
+            if first_items.isnull().any():
```
shoyer: if you merge in master against, you could switch this block back to using `Series.dropna()`.

xarray/tests/test_formatting.py
```diff
@@ -189,6 +189,53 @@ def test_attribute_repr(self):
         assert '\n' not in newlines
         assert '\t' not in tabs
 
+    def test_diff_dataset_repr(self):
```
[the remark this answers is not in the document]
jwenfai: Sorry about that, didn't know how I missed it.

xarray/core/groupby.py
jwenfai: Done.

shoyer:
I'm going to merge this when tests pass

spencerkclark:
Sounds good @shoyer, thanks for bringing this to the finish line.

jwenfai:
All tests passed. Thanks, @spencerkclark and @shoyer, for all the help!

shoyer:
thanks @jwenfai and @spencerkclark !

PR #2603, opened 2018-12-12
Support HighLevelGraphs

### LINKED ISSUE #4291: resample function gives 0s instead of NaNs
**What happened**:
When I use `resample(time='1d').sum(dim='time')` to resample a time series with NaNs, the resampled result gives me 0s instead of NaNs, while NaNs should be the correct answer.

**What you expected to happen**:

NaNs should be the correct answer.

**Minimal Complete Verifiable Example**:

```python
import xarray as xr

dates =  pd.date_range('20200101', '20200601', freq='h')
data = np.linspace(0, 10, num=len(dates))
data[0:30*24] = np.nan

da = xr.DataArray(data, coords=[dates], dims='time')
da.plot()

# Instead of NaNs, the resampled time series in January 20202 give us 0s, which not right.
da.resample(time='1d', skipna=True).sum(dim='time', skipna=True).plot()
```

**Anything else we need to know?**:

Did I misunderstand something here? Thanks!

**Environment**:
xarray - '0.15.1' 

xarray - '0.15.1' 

</details>

mrocklin:
Fixes https://github.com/dask/dask/issues/4291

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -509,13 +509,22 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         if not graphs:
             return None
         else:
-            from dask import sharedict
-            return sharedict.merge(*graphs.values())
+            try:
+                from dask.highlevelgraph import HighLevelGraph
+                return HighLevelGraph.merge(*graphs.values())
+            except ImportError:
+                from dask import sharedict
+                return sharedict.merge(*graphs.values())
+
 
     def __dask_keys__(self):
         import dask
         return [v.__dask_keys__() for v in self.variables.values()
                 if dask.is_dask_collection(v)]
+
+    def __dask_layers__(self):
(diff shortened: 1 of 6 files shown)


xarray/core/dataset.py
```diff
@@ -509,14 +509,23 @@ def __dask_graph__(self):
...
+
 
     def __dask_keys__(self):
         import dask
         return [v.__dask_keys__() for v in self.variables.values()
                 if dask.is_dask_collection(v)]
 
+    def __dask_layers__(self):
+        return sum([v.__dask_layers__() for v in self.variables.values() if
+                    dask.is_dask_collection(v)], [])
```
shoyer: I think you need to add a local `import dask` here since dask isn't a hard dependency. This suggests there isn't any coverage for this new method -- is there any easy way to add that?

shoyer:
thanks!

PR #2612, opened 2018-12-16
Added Coarsen

### LINKED ISSUE #2525: Multi-dimensional binning/resampling/coarsening

The StackOverflow Query : https://stackoverflow.com/questions/52886703/xarray-multidimensional-binning-array-reduction-on-sample-dataset-of-4-x4-to/52981916#52981916

```python
import pandas as pd
import numpy as np
import xarray as xr
a = np.array(np.random.randint(1, 90+1,(4,4)),dtype=np.float64)
b = np.array(np.random.randint(1, 360+1,(4,4)),dtype=np.float64)
c = np.random.random_sample(16,)
c = c.reshape(4,4)
dsa = xr.Dataset()
dsa['CloudFraction'] = (('x', 'y'), c)
dsa.coords['latitude'] = (('x', 'y'), a)
dsa.coords['longitude'] = (('x', 'y'), b)
dsa


Dimensions: (x: 4, y: 4)
Coordinates:
latitude       (x, y) float64 23.0 16.0 53.0 1.0 ... 82.0 65.0 45.0 88.0
longitude      (x, y) float64 219.0 13.0 276.0 69.0 ... 156.0 277.0 16.0
Data variables:
CloudFraction  (x, y) float64 0.1599 0.05671 0.8624 ... 0.7757 0.7572
```
#### Problem description

I am trying to reduce an Xarray from 4 x 4 to 2 x 2 via both the dimensions. I haven't found any luck with the current Xarray Dataset. These are the steps I followed. I want to bin or group based on latitude and longitude both simultaneously to reduce number of steps. Currently I can achieve this by just GroupBy method which doesn't seem to perform GroupBy on both the coordinates.

To elaborate the idea I want to achieve : 
       1.  Considering a matrix of 4x4 , first we will group elements of index (0,0) with  index (0,1) as A , index of (0,2) with index (0,3) as B, index of (1,0) with index (1,1) as C , index of (1,2) with index (1,3) as D and so on so forth. Last combination being index of (3,2) with index (3,3) as H.
     2. This turns the matrix of 4x4 to 4x2 and now we combine elements A with C and B with D and so and so forth. The final matrix size should be 2x2.
     3. The combination of elements can be done with any aggregation functions like mean() 
or std() and needs to be done over the coordinate of 'Latitude' and 'Longitude in reference to the data variables 'Cloud Fraction' 
     4. Is there a way to obtain this with Xarray functions and automate it with any input matrix size .

#### Expected Output

To elaborate the idea I want to achieve :
     1.  Considering a matrix of 4x4 , first we will group elements of index (0,0) with  index (0,1) as A , index of (0,2) with index (0,3) as B, index of (1,0) with index (1,1) as C , index of (1,2) with index (1,3) as D and so on so forth. Last combination being index of (3,2) with index (3,3) as H.
     2. This turns the matrix of 4x4 to 4x2 and now we combine elements A with C and B with D and so and so forth. The final matrix size should be 2x2.
     3. The combination of elements can be done with any aggregation functions like mean() 
or std().

<details>
Numpy Version. : v1.15.1
Xarray Version : v0.10.9
Python Verison : v3.7.0
Jupyter Notebook : Locally Hosted

</details>

fujiisoup:
- [x] Closes #2525
 - [x] Tests added

Started to implement `corsen`.
The API is currently something like
```python
    actual = ds.coarsen(time=2, x=3, side='right',
                        coordinate_func={'time': np.max}).max()
```

Currently, it is not working for a datetime coordinate, since `mean` does not work for this dtype.
e.g.
```python
da = xr.DataArray(np.linspace(0, 365, num=365), 
         dims='time', coords={'time': pd.date_range('15/12/1999', periods=365)}) 
da['time'].mean()    # -> TypeError: ufunc add cannot use operands with types dtype('<M8[ns]') and dtype('<M8[ns]')
```

I am not familiar with datetime things. 
Any advice will be appreciated.


dcherian:
`(da.time - da.time[0]).mean() + da.time[0]`

Or we could default to `min()` and add an `loffset` kwarg like `resample` (https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.resample.html)

max-sixty:
Forgive me if this has an obvious answer: to what extent is this downsampling? Could this be done with `resample`?

dcherian:
@max-sixty I think it's multi-dimensional resampling / rolling. resample & rolling operate one dimension at a time.

xarray/core/variable.py
```diff
@@ -1632,6 +1632,73 @@ def rolling_window(self, dim, window, window_dim, center=False,
...
+                size = variable.shape[i]
+                shape.append(int(size / windows[d]))
+                shape.append(windows[d])
+                dims.append(d)
+                dims.append(coarsen_dimensions[d])
+            else:
+                shape.append(variable.shape[i])
+                dims.append(d)
+
+        return Variable(dims, variable.data.reshape(shape), self._attrs)
```
shoyer: Is it worth making an actual xarray.Variable object here rather than just returning `data`? With the need to come up with new dimension names, I think it might be cleaner avoiding that.

xarray/core/rolling.py
```diff
@@ -454,5 +436,121 @@ def construct(self, window_dim, stride=1, fill_value=dtypes.NA):
...
+            return DataArray(reduced, dims=self.obj.dims, coords=coords)
+
+        return wrapped_func
+
+
+class DatasetCoarsen(Coarsen):
+    @classmethod
+    def _reduce_method(cls, func):
+        """
+        Return a wrapped function for injecting numpy and bottoleneck methods.
```
shoyer: bottoleneck -> bottleneck

xarray/core/common.py
```diff
@@ -591,6 +591,55 @@ def rolling(self, dim=None, min_periods=None, center=False, **dim_kwargs):
...
+        Parameters
+        ----------
+        dim: dict, optional
+            Mapping from the dimension name to the window size.
+        side : 'left' or 'right', or dict
+            If left, coarsening windows start from 0th index. The excessed
+            entries in the most right will be removed (if trim_excess is True).
+            If right, coarsen windows ends at the most right entry, while
+            excessed entries in the most left will be removed.
+        trim_excess : boolean, default False
```
shoyer: It would be nice to have an API that lets us express at least three options:
- trim excess entries
- pad with NaN
- raise an error if the shape does not divide exactly (this is probably the safest default behavior)

Maybe a string valued keyword argument would work better here, e.g., `boundary='trim'`, `boundary='pad'` and `boundary='exact'`?

I would also suggest putting this argument before `side`, since `side` is only used for a particular (non-default) value of this argument.

max-sixty:
> I think it's multi-dimensional resampling / rolling. resample & rolling operate one dimension at a time.

Thanks. If this is multi-dimensional resampling / rolling, could we implement this functionality within those methods, enabling multiple dimensions?

Potentially the implementations are different enough that the arguments don't have enough overlap?

shoyer:
@max-sixty we discussed this a little bit: https://github.com/pydata/xarray/issues/2525#issuecomment-434477550

The main difference is that resample is coordinate based, whereas this is integer position based which makes the implementation considerably simpler.

max-sixty:
Perfect, thanks for pointing that out (I've generally been a bit out of the loop recently...)

shoyer:
I do think it would be nice for `resample()` to work with numbers and along multiple dimensions, but that's definitely a bigger project.

fujiisoup:
Updated.
The main change is
1. API updated to
```python
dataset.coarsen(self, dim=None, boundary='exact', side='left',
                            coordinate_func='mean', **dim_kwargs):
```
based on comments.

2. nanmean for DateTime was implemented.
This is a kind of backward incompatible.
Previously, we raised an Error if mean is applied to a datetime array.

fujiisoup:
During writing the test codes, I realized that the default `boundary='exact'` is a little annoying.
It would be rare that the length of the data is exactly a multiple of the window size; we almost always need to type `bounary='trim'` or the other.
I personally think that `boundary='pad' is also a good default, as it is a similar behaviour to `rolling`.

The down side of this is that it can make a uniformly spaced data to an inhomogeneously spaced one.
It is something what users may not expect.
(probably `boundary='exact` is safer)

shoyer:
@pydata/xarray any opinions on what the default value for the `boundary` argument to `coarsen()` should be?

Personally, I like `boundary='exact'`, but I also work mostly with simulated data with dimensions setup to be exact powers of 2 :).

jhamman:
+1, on `'exact'`. I like the idea of making users be explicit about when to pad or trim edge cells.

doc/computation.rst
```diff
@@ -199,6 +199,41 @@ You can also use ``construct`` to compute a weighted rolling sum:
   To avoid this, use ``skipna=False`` as the above example.
 
 
+Coarsen large arrays
+====================
+
+``DataArray`` objects include a :py:meth:`~xarray.DataArray.coarsen` method.
```
shoyer: Maybe "DataArray and Dataset"?

doc/computation.rst
```diff
@@ -199,6 +199,41 @@ You can also use ``construct`` to compute a weighted rolling sum:
...
+.. ipython:: python
+
+  da.coarsen(time=30, boundary='trim').mean()
+
+If you want to apply a specific function to coordinate, you can pass the
+function of function name to ``coordinate_func`` option,
+
+.. ipython:: python
+
+  da.coarsen(time=7, coordinate_func={'time': 'min'}).mean()
```
shoyer: Should we abbreviate this as `coord_func`? We use that elsewhere in xarray.

doc/computation.rst
```diff
@@ -199,6 +199,41 @@ You can also use ``construct`` to compute a weighted rolling sum:
...
 
+Coarsen large arrays
+====================
+
+``DataArray`` objects include a :py:meth:`~xarray.DataArray.coarsen` method.
+This method supports the block aggregation:
+
+.. ipython:: python
+
+  da = xr.DataArray(np.linspace(0, 364, num=364), dims='time',
```
shoyer: What do you think about including a 2D example in the docs, e.g., a 4x4 array to 2x2? I expect that is closer to the typical use case.

xarray/core/duck_array_ops.py
```diff
@@ -278,6 +276,22 @@ def f(values, axis=None, skipna=None, **kwargs):
...
 
 
+_mean = _create_nan_agg_method('mean')
+
+def mean(array, axis=None, skipna=None, **kwargs):
+    """ inhouse mean that can handle datatime dtype """
+    array = asarray(array)
+    if array.dtype.kind == 'M':
+        offset = min(array)
+        dtype = (array.ravel()[0] - offset).dtype  # dtype.kind == 'm'
```
shoyer: This will fail for size 0 arrays. Maybe switch to another reduce method if the array is size 0, e.g., use min instead? The only important part is that it handles reducing the shape properly.

xarray/core/duck_array_ops.py
```diff
@@ -278,6 +276,25 @@ def f(values, axis=None, skipna=None, **kwargs):
...
+_mean = _create_nan_agg_method('mean')
+
+
+def mean(array, axis=None, skipna=None, **kwargs):
+    """ inhouse mean that can handle datatime dtype """
+    array = asarray(array)
+    if array.dtype.kind == 'M':
+        offset = min(array)
+        # infer the compatible timedelta dtype
+        dtype = (np.empty((1,), dtype=array.dtype) - offset).dtype
```
fujiisoup: This is just to find the corresponding timedelta from datetime. Is there any good function to find an appropriate dtype?

xarray/core/duck_array_ops.py
```diff
@@ -278,6 +276,25 @@ def f(values, axis=None, skipna=None, **kwargs):
 cumsum_1d.numeric_only = True
 
 
+_mean = _create_nan_agg_method('mean')
+
+
+def mean(array, axis=None, skipna=None, **kwargs):
```
fujiisoup: I would like to make this compatible this function with CFTime index. @spencerkclark, could you comment for this?

doc/computation.rst
fujiisoup: Changed to 2d example

xarray/core/duck_array_ops.py
fujiisoup: Fixed, but I am not yet sure it (the updated one) is the best solution to find the appropriate dtype. See [comment](https://github.com/pydata/xarray/pull/2612/files/82c08afdaf38ab0785daca73f5c882a1ce320227#diff-b81ae485ac6616ab60b4c70334009571)

xarray/tests/test_dataset.py
```diff
@@ -4427,6 +4427,49 @@ def ds(request):
...
+    assert_equal(actual['time'],
+                 ds['time'].coarsen(time=2, x=3, boundary='trim').max())
+
+    # raise if exact
+    with pytest.raises(ValueError):
+        ds.coarsen(x=3).mean()
+    # should be no error
+    ds.isel(x=slice(0, 3 * (len(ds['x']) // 3))).coarsen(x=3).mean()
+
+    # raise if exact
```
shoyer: delete

xarray/tests/test_variable.py
```diff
@@ -1676,6 +1676,31 @@ def assert_assigned_2d(array, key_x, key_y, values):
...
+        actual = v.coarsen({'x': 2}, func='mean', boundary='pad',
+                           side='right')
+        expected = self.cls(['x'], [0, 1.5, 3.5])
+        assert_identical(actual, expected)
+
+        actual = v.coarsen({'x': 2}, func=np.mean, side='right',
+                           boundary='trim')
+        expected = self.cls(['x'], [1.5, 3.5])
+        assert_identical(actual, expected)
+
```
shoyer: I would be good to add a test case that checks the values of a higher dimensional array, e.g., to verify that coarsening a 4x4 array to 2x2 gives the right result.

doc/computation.rst
```diff
@@ -199,6 +199,45 @@ You can also use ``construct`` to compute a weighted rolling sum:
...
+.. ipython:: python
+
+  da.coarsen(time=30, x=2, boundary='trim').mean()
+
+If you want to apply a specific function to coordinate, you can pass the
+function of function name to ``coordinate_func`` option,
+
+.. ipython:: python
+
+  da.coarsen(time=7, x=2, coordinate_func={'time': 'min'}).mean()
```
shoyer: This should be `coord_func`, not `coordinate_func`.

xarray/core/duck_array_ops.py
fujiisoup: Thanks, @spencerkclark 

It would be nice if you could send a follow-up PR :)

xarray/core/duck_array_ops.py
fujiisoup: Thanks. I just realized we are always using [ns] for datetime. Updated.

xarray/core/duck_array_ops.py
spencerkclark: Sure thing, I'd be happy to take care of making this compatible with cftime dates.

xarray/tests/test_variable.py
fujiisoup: Fair comment. Thanks. Updated.

fmaussion:
This is (one more time) an extremely useful addition to xarray - thanks so much @fujiisoup !

shoyer:
Indeed, thank you @fujiisoup !

PR #2613, opened 2018-12-17
Remove tz argument in cftime_range

spencerkclark:
This was caught by @jwenfai in #2593.  I hope no one was inadvertently trying to use this argument before.

Should this need a what's new entry?


shoyer:
I don't think this even really registers as a bug worth mentioning :)

Issue #2609, opened 2018-12-15
Cannot replace xr.ufuncs.angle with np.angle


rth:
Currently one gets a deprecation warning when using `xarray.ufuncs.angle` on a DataArray with xarray 0.11 and numpy 1.15, e.g.
```
PendingDeprecationWarning: xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.13. Instead, use numpy ufuncs directly.
    X_tr = xr.ufuncs.angle(X)
```

however, `np.angle` still casts the input to `ndarray`, apparently.

#### Code Sample, a copy-pastable example if possible

```python
>>> import xarray as xr
>>> import numpy as np
>>> X = xr.DataArray(np.ones(10))
>>> X
<xarray.DataArray (dim_0: 10)>
array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])
Dimensions without coordinates: dim_0
>>> xr.ufuncs.angle(X)  # expected output
<xarray.DataArray (dim_0: 10)>
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
Dimensions without coordinates: dim_0
>>> np.angle(X)  # actual output
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
>>> np.__version__
'1.15.4'
>>> xr.__version__
'0.11.0'
>>> np.sin(X)  # other ufuns work
<xarray.DataArray (dim_0: 10)>
array([0.841471, 0.841471, 0.841471, 0.841471, 0.841471, 0.841471, 0.841471,
       0.841471, 0.841471, 0.841471])
Dimensions without coordinates: dim_0
```
#### Problem description

I have not investigated yet why happens in numpy, but opening this issue to indicate that currently I can't really do much about this DeprecationWarning sort of silencing it (unless I'm missing something).

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.1.final.0
python-bits: 64
OS: Linux
OS-release: 4.18.8-gentoo
machine: x86_64
processor: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8

xarray: 0.11.0
pandas: 0.23.4
numpy: 1.15.4
scipy: 1.1.0
netCDF4: None
h5netcdf: 0.6.2
h5py: 2.8.0
[...]

PR #2615, opened 2018-12-17
FIX Don't raise a deprecation warning for xarray.ufuncs.{angle,iscomplex}

rth:
Closes https://github.com/pydata/xarray/issues/2609

This removes PendingDeprecationWarnings for `xarray.ufuncs.{angle,iscomplex}` which appear to not be proper ufuncs and cannot be just used from numpy for now (as discussed in the parent issue).

Also adds more generic tests that numpy ufuncs do preserve the DataArray dtype.

### DIFF
--- a/xarray/tests/test_ufuncs.py
+++ b/xarray/tests/test_ufuncs.py
@@ -172,6 +172,41 @@ def test_xarray_ufuncs_deprecation():
     with pytest.warns(PendingDeprecationWarning, match='xarray.ufuncs'):
         xu.cos(xr.DataArray([0, 1]))
 
+    with pytest.warns(None) as record:
+        xu.angle(xr.DataArray([0, 1]))
+    record = [el.message for el in record
+              if el.category == PendingDeprecationWarning]
+    assert len(record) == 0
+
+
+@requires_np113
+@pytest.mark.filterwarnings('ignore::RuntimeWarning')
+@pytest.mark.parametrize(
+    'name',
+    [name for name in dir(xu)
+        if (not name.startswith('_') and hasattr(np, name)
+            and name not in ['print_function', 'absolute_import', 'division'])]
+)
+def test_numpy_ufuncs(name, request):
+    x = xr.DataArray([1, 1])
+
+    np_func = getattr(np, name)
+    if hasattr(np_func, 'nin') and np_func.nin == 2:
+        args = (x, x)
+    else:
+        args = (x,)
+
+    y = np_func(*args)
+
+    if name in ['angle', 'iscomplex']:
+        # these functions need to be handled with __array_function__ protocol
+        assert isinstance(y, np.ndarray)
+    elif name in ['frexp']:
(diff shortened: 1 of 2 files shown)


xarray/tests/test_ufuncs.py
```diff
@@ -172,6 +172,41 @@ def test_xarray_ufuncs_deprecation():
...
+    assert len(record) == 0
+
+
+@requires_np113
+@pytest.mark.filterwarnings('ignore::RuntimeWarning')
+@pytest.mark.parametrize(
+    'name',
+    [name for name in dir(xu)
+        if (not name.startswith('_') and hasattr(np, name)
+            and name not in ['print_function', 'absolute_import', 'division'])]
```
rth: Here we list all ufuncs defined in `xarray.ufuncs`, but there are some functions that need to be filtered out as a result.

xarray/tests/test_ufuncs.py
```diff
@@ -172,6 +172,41 @@ def test_xarray_ufuncs_deprecation():
...
+        args = (x,)
+
+    y = np_func(*args)
+
+    if name in ['angle', 'iscomplex']:
+        # these functions need to be handled with __array_function__ protocol
+        assert isinstance(y, np.ndarray)
+    elif name in ['frexp']:
+        # np.frexp returns a tuple
+        assert not isinstance(y, xr.DataArray)
```
rth: The output is a tuple of len 2, it's a bit of a special case, not sure if it's really worth adding specific tests for it.

xarray/ufuncs.py
```diff
@@ -44,10 +44,12 @@ def __init__(self, name):
...
-        _warnings.warn(
-            'xarray.ufuncs will be deprecated when xarray no longer supports '
-            'versions of numpy older than v1.13. Instead, use numpy ufuncs '
-            'directly.', PendingDeprecationWarning, stacklevel=2)
+        if self._name not in ['angle', 'iscomplex']:
+            _warnings.warn(
+                'xarray.ufuncs will be deprecated when xarray no longer '
+                'supports versions of numpy older than v1.13. Instead, use '
+                'numpy ufuncs directly.',
+                PendingDeprecationWarning, stacklevel=2)
```
shoyer: It might be better to say: "xarray.ufuncs will be deprecated when xarray no longer supports versions of numpy older than v1.17." We'll need `__array_function__` from 1.17 to override `angle` and `iscomplex`.

Or perhaps we should just drop the pending deprecation warning entirely given how far away this is (NumPy hasn't even released 1.16 yet, nevermind 1.17)

rth: Updated the warning message as you suggested.

I think independently of when `xarray.ufuncs` is going to be deprecated, it is good to let the user know that they can use numpy ufuncs now (for most functions except a few for which this warning is not raised)

shoyer:
thanks @rth !

PR #2616, opened 2018-12-17
API for N-dimensional combine

TomNicholas:
Continues the discussion from #2553 about how the API for loading and combining data from multiple datasets should work. (Ultimately part of the solution to #2159)

@shoyer this is for you to see how I envisaged the API would look, based on our discussion in #2553. For now you can ignore all the changes except the ones to the docstrings of `auto_combine` [here](https://github.com/pydata/xarray/compare/master...TomNicholas:feature/nd_combine_new_api?expand=1#diff-876f2fcf4679457325e8018f6d98660cR650), `manual_combine` [here](https://github.com/pydata/xarray/compare/master...TomNicholas:feature/nd_combine_new_api?expand=1#diff-876f2fcf4679457325e8018f6d98660cR550) and `open_mfdataset` [here](https://github.com/pydata/xarray/compare/master...TomNicholas:feature/nd_combine_new_api?expand=1#diff-e58ddc6340e4b5dd0a6e6b443c9a6da1R483).

Feedback from anyone else is also encouraged, as really the point of this is to make the API as clear as possible to someone who hasn't delved into the code behind `auto_combine` and `open_mfdataset`.

It makes sense to first work out the API, then change the internal implementation to match, using the internal functions developed in #2553. Therefore the tasks include:

- [x] Decide on API for 'auto_combine' and 'open_mfdataset'
- [x] Appropriate documentation
- [x] Write internal implementation of `manual_combine`
- [x] Write internal implementation of `auto-combine` 
- [x] Update `open_mfdataset` to match
- [x] Write and reorganise tests
- [x] Automatically ordering of string and datetime coords
- [x] What's new explaining changes
- [x] Make sure `auto_combine` and `manual_combine` appear on the [API page](http://xarray.pydata.org/en/stable/api.html) of the docs
- [x] PEP8 compliance
- [x] Python 3.5 compatibility
- [x] AirSpeedVelocity tests for `auto_combine`
- [x] Finish all TODOs
- [x] Backwards-compatible API to start deprecation cycle
- [x] Add examples from docstrings to main documentation pages


xarray/backends/api.py
```diff
@@ -561,6 +571,9 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     parallel : bool, optional
         If True, the open and preprocess steps of this function will be
         performed in parallel using ``dask.delayed``. Default is False.
+    combine : {'manual', 'auto'}, optional
+        Whether ``xarray.auto_combine`` or ``xarray.manual_combine`` is used to
+        combine all the data. Default is 'manual'.
```
shoyer: I think it probably makes more sense for the default to be `'auto'`, unless you explicitly provide a list of dimensions to stack along?

xarray/core/combine.py
```diff
@@ -548,31 +547,142 @@ def _auto_combine(datasets, concat_dims, compat, data_vars, coords,
...
+    parallelized in multiple dimensions, but has global coordinates saved in
+    each file specifying it's position within the domain.
+
     Parameters
     ----------
     datasets : sequence of xarray.Dataset
-        Dataset objects to merge.
-    concat_dim : str or DataArray or Index, optional
+        Dataset objects to combine.
+    concat_dim : str, DataArray, Index or None, optional
```
shoyer: I'm not sure if there are any cases in which it makes sense to insert new dimensions when doing automatic combining. Perhaps we should drop this argument?

xarray/backends/api.py
```diff
@@ -480,35 +480,45 @@ def close(self):
...
         By default, chunks will be chosen to load entire input files into
         memory at once. This has a major impact on performance: please see the
         full documentation for more details [2].
-    concat_dim : None, str, DataArray or Index, optional
-        Dimension to concatenate files along. This argument is passed on to
-        :py:func:`xarray.auto_combine` along with the dataset objects. You only
-        need to provide this argument if the dimension along which you want to
-        concatenate is not a dimension in the original datasets, e.g., if you
-        want to stack a collection of 2D arrays along a third dimension.
+    concat_dims : list of str, DataArray, Index or None, optional
```
shoyer: I'm not sure it's worth renaming this argument. We use the singular `dim` for aggregation methods like `mean`, even though you can also pass in a list of dimensions. We'll probably want to still accept scalar strings here, too.

xarray/backends/api.py
TomNicholas: I agree the default should be `'auto'`, the reason it's not here is because no-one has written the automatic-ordering logic yet.

That should be relatively simple though right? Something like:

```python
def _infer_concat_order_from_coords(datasets, concat_dims):
  
    # Deduce order along each dimension
    tile_ids = [(,) for ds in datasets]
    for dim in concat_dims:
        # Requires reading the dims and values of coordinate variables from all datasets
        unsorted_first_coord_values = np.vstack(ds[dim].values[0] for ds in datasets)

        # Sort along dim 
        # (Presumably use numpy.argsort, but might need special cases for string coords)
        order = np.argsort(unsort_first_coord_values)

        tile_ids = [tile_id + order[i] for i, tile_id in enumerate(tile_ids)]

    combined_ids = OrderedDict(zip(tile_ids, datasets))

    return combined_ids, concat_dims
```

xarray/backends/api.py
TomNicholas: Okay noted!

xarray/core/combine.py
TomNicholas: That would mean that a semi-automatic merge-then-concat (e.g. for combining simulations with different parameter choices into one large dataset) which used to be doable in one `auto_combine` now has to be done in two lines (`auto_combine` then `concat`). 

However I think I agree that restricting the scope of the function would be clearer overall though.

xarray/backends/api.py
TomNicholas: I've just realised the approach here doesn't do any auto-merging, only auto-concatenating... Maybe I need to think about this a bit more...

shoyer: This could probably be done in a two step process, like what we currently do:
- Group datasets by their unique data variables. Within each group, concatenate.
- Merge together the results.

TomNicholas: > This could probably be done in a two step process

So I think you would need to work out the ordering separately for each group of datasets, in case some of them don't contain the same dimensions, but that's okay, i.e:

```python
# Group by data vars
grouped = itertools.groupby(datasets, key=lambda ds: tuple(sorted(ds)))

concatenated = []
for var_group in grouped:
    combined_ids, concat_dims = _infer_order_from_coords(list(var_group))

    concatenated.append(_combine_nd(combined_ids, concat_dims))
    
return merge(concatenated, compat=compat)
    
```

> The alternative would be to check pandas.Index.is_monotonic to verify monotonicity on each individual dataset, and then verify that the last coordinate value of each dataset is less than the first coordinate value of the next dataset.

That's neater and clearer, I'll do that.

shoyer: > That's neater and clearer, I'll try that. Presumably the `pandas.Index` object already exists for each dimension on each dataset, is there a way to access this via some hidden part of the xarray API?

You can find indexes in `Dataset.indexes`

TomNicholas:
Okay so this is mostly done - I've completely rearranged the tests and they pass on my unmerged branch, there are just a few TODOs left.

However, when I tried to merge master in then everything failed again due to #2648, I don't understand what `ReprObject` does?

Some other questions: 

1) We didn't decide whether the ability to infer concatenation dimensions (i.e. when `concat_dim='_CONCAT_DIM_DEFAULT'`) should be retained or not. I'm not sure we need it at all any more - you're either using `auto_combine`, which will concatenate dimension coordinates and leave other dimensions alone, or you're using `manual_combine`, in which you're supposed to explicitly state all the dimensions you wish to concatenate along anyway. I think this is also related to a bug I don't properly understand (marked with an xfail in `test_manual_concat_too_many_dims_at_once`).

2) Ordering the datasets along one dimension according to the values in their dimension coordinates (done in `_infer_order_1d`) turned out to be a bit more complicated than I thought. Is the method using a `Pandas.Series` okay? Do we want to use natural sorting on strings? How should I deal with `datetime` objects? I also think it's probably easier to check that the coords can actually be arranged to be monotonic after all the concatenation has been done, not before.

3) Some input which previously wouldn't be valid now is (see the TODO in line 816 of `test_combine.py`), is that the desired behaviour in that case?

xarray/backends/api.py
```diff
@@ -570,6 +576,9 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     parallel : bool, optional
         If True, the open and preprocess steps of this function will be
         performed in parallel using ``dask.delayed``. Default is False.
+    combine : {'auto', 'manual'}, optional
+        Whether ``xarray.auto_combine`` or ``xarray.manual_combine`` is used to
+        combine all the data. Default is 'auto'.
```
WeatherGod: This seems unnecessary. If a user supplies values for `concat_dim`, then it explicitly means manual. If the user does not supply `concat_dim`, then it means auto. The logic even precludes requesting auto, and supplying concat_dim and whether the inverse would work (requesting manual and not supplying concat_dim) isn't really clear.

xarray/tests/test_backends.py
```diff
@@ -2163,11 +2164,13 @@ def test_open_mfdataset(self):
             with create_tmp_file() as tmp2:
                 original.isel(x=slice(5)).to_netcdf(tmp1)
                 original.isel(x=slice(5, 10)).to_netcdf(tmp2)
-                with open_mfdataset([tmp1, tmp2]) as actual:
+                with open_mfdataset([tmp1, tmp2], combine='manual',
+                                    concat_dim='x') as actual:
```
WeatherGod: why are you updating these tests? Is it just to increase coverage a bit, or will these previously valid calls now be invalid? If so, then this is probably going to need a deprecation cycle. It is not obvious to me why this needs changing.

xarray/core/combine.py
```diff
@@ -420,6 +66,82 @@ def _infer_tile_ids_from_nested_list(entry, current_pos):
...
+
+        # Check if dim is a coordinate dimension
+        if dim in ds0:
+
+            # Need to read coordinate values to do ordering
+            coord_vals = [ds[dim].values for ds in datasets]
+
+            # If dimension coordinate values are same on every dataset then
+            # should be leaving this dimension alone (it's just a "bystander")
+            if not _all_arrays_equal(coord_vals):
```
shoyer: I would suggest doing these checks with pandas.Index objects instead, e.g.,
```python
indexes = [ds.indexes.get(dim) for ds in datasets]
if any(index is None for index in indexes):
    raise ValueError('every dimension needs a coordinate for inferring concatation order')

# this assumes we have at least one dataset, but I think that is safe?
if not all(index.equals(indexes[0]) for index in indexes[1:]):
    concat_dims.append(dim)

    if not all(index.is_monotonic_increasing for index in indexes):
        raise ValueError
    
    first_items = indexes[0][:1].concat([index[:1] for index in indexes])

    if first_items.size < len(indexes):
        raise ValueError("can't handle size zero dimensions")

    order = first_items.to_series().rank(method='dense').astype(int).values - 1
```

TomNicholas: pandas `Index` objects don't seem to actually have a `concat()` method but apart from that I've put this in now, thanks.

This approach seems to also work for string coords and datetime coords, but I'm not sure I'm that's guaranteed pandas behaviour (we are using `.rank()` on non-numeric quantities), and it also is sorting the strings lexicographically, not naturally, which arguably isn't what we want (e.g. for a coordinate dimension like `['run8', 'run9', 'run10']`).

shoyer: Oops, it's `pandas.Index.append`, not `concat`. The virtue of using `append()` is that it should preserve the exact index type.

I think lexciographic order is probably the right choice for us. Natural sorting sort of makes sense for print output for users, but it's pretty convoluted / hard to debug so I doubt it makes sense to internally in xarray.

xarray/backends/api.py
TomNicholas: I think you make a good point here. It relates back to me saying 

>We didn't decide whether the ability to infer concatenation dimensions (i.e. when concat_dim='_CONCAT_DIM_DEFAULT') should be retained or not. 

At the moment then if `combine='manual'` and `concat_dim` is not supplied then it will try to infer the concatenation dimension for every "dimension" in the nested-list-input. If we require that `concat_dims` is supplied explicitly when `combine='manual'` then we are free to do what you suggest.

xarray/backends/api.py
```diff
@@ -487,35 +488,42 @@ def close(self):
...
+        Dimensions to concatenate files along.  You only
+        need to provide this argument if any of the dimensions along which you
+        want to concatenate is not a dimension in the original datasets, e.g.,
+        if you want to stack a collection of 2D arrays along a third dimension.
         By default, xarray attempts to infer this argument by examining
-        component files. Set ``concat_dim=None`` explicitly to disable
-        concatenation.
-    compat : {'identical', 'equals', 'broadcast_equals', 'no_conflicts'}, optional
+        component files. Set ``concat_dim=[..., None, ...]`` explicitly to
+        disable concatenation along a particular dimension.
```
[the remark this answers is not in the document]
TomNicholas: It means that the datasets along that dimension of the nested-list input are to be merged, not concatenated. I agree it's not very clear at the moment, I was going to add some more examples to the docstring to try and clarify.

xarray/core/combine.py
TomNicholas: > Oops, it's pandas.Index.append, not concat. The virtue of using append() is that it should preserve the exact index type.

I did it with by initialising a new `pandas.Index` object instead, which personally I think is a bit clearer. I guess that might not preserve the type though?

> I think lexicographic order is probably the right choice for us. Natural sorting sort of makes sense for print output for users, but it's pretty convoluted / hard to debug so I doubt it makes sense to internally in xarray.

:+1:

xarray/tests/test_backends.py
TomNicholas: > will these previously valid calls now be invalid?

If we do what you suggested in your other comment then these particular calls will still be valid (so I would revert these particular tests).

> why are you updating these tests?

Because as proposed right now this is going to be a breaking change :confused:. 

Due to what me and @shoyer decided during the discussion [here](https://github.com/pydata/xarray/pull/2553#issuecomment-445419876), then some previously-valid inputs to `auto_combine` (and hence to `open_mfdataset`) will become invalid.

As far as I know then there is only one type of input which will be go from valid to invalid (assuming we implement your comment). That's the case of wanting to combine a set of datasets which need to be both concatenated and merged, but you want to concatenate along a dimension which does not have a corresponding coordinate variable. The old `auto_combine` would have just concatenated them in the order they were given, but the new one will complain that it doesn't know what the order should be without a global coordinate to order them. The new `manual_combine` won't do it because you need to perform two operations (concat then merge), which currently requires a two-dimensional list input, i.e:

```python
datasets = [[var1_x0, var1_x1], [var2_x0, var2_x1]]
result = manual_combine(datasets, concat_dim=['x', None])
```

This is why I wrote in the what's new that it's a potentially breaking change - whether or not it justifies a deprecation cycle I guess is not up to me....

It might be possible to make `manual_combine` more flexible so that it would accept

```python
datasets = [var1_x0, var1_x1, var2_x0, var2_x1]
result = manual_combine(datasets, concat_dim=['x', None])
```

(which currently will throw an error because `len(concat_dim) =/= datasets.ndims` basically), but I'm not sure this solution is any clearer.

@shoyer what do you think?

shoyer: We can (and probably should) deprecate this, but we try really hard to avoid hard backwards compatibility breaks.

Could we implement some fall-back logic to do this? e.g., by explicitly checking for the scenario where only a single dimension to concatenate is provided, but the datasets also need to be merged. In this case we would append another dimension `None` and issue a warning that this behavior won't be supported in the future.

TomNicholas: @shoyer I've changed the API to how I think it should be finally (e.g. removed automatic inferring of `concat_dim` in `manual_combine`as discussed above), and tried to make the documentation as clear as possible. However I don't know what you want to do about the backwards-compatibility issue.

The problem is that with separate `auto_combine` and `manual_combine` functions (which I think make sense to have), then the new `auto_combine`'s behaviour is so different to the old one that creating full backwards compatibility will basically involve having both the new and the old one present in the code simultaneously. This is feasible, but also the new and old `auto_combine` will give different results for the same input, and you can't know whether a user wants the new or old version to be used. Therefore the only way I can see to slowly phase the old one out is to have some kind of default `OLD=True` flag or something. That would probably be confusing documentation-wise though.

Another idea would be to just change it so the current behaviour has a loud warning saying the default behaviour will change soon, and once it changes then raise a warning whatever you do to get people to check that their datasets are still be combined in the way they want them to.

shoyer: To recap, there are two cases whose behavior we want to change:
1. 1D lists, which require both concatenation *and* merging and for which the concentration dimension was given explicitly
2. 1D lists, which should be concentrated in order of appearance, not coordinate order.

In both cases, we eventually want to change the default behavior of `xarray.open_mfdataset(paths)`, and there's an easy way to clearly opt-in to the old behavior even with the new code (because these arguments weren't valid previously):
1. `xarray.open_mfdataset(paths, concat_dim=['time', None])`
2. `xarray.open_mfdataset(paths, combine='manual')`

And in both cases, it would also be straightforward to opt-in to the new behavior early by writing `combine='auto'`.

So I think a deprecation cycle would be relatively feasible. The logic goes roughly like this:
- If a user makes use of any new arguments that weren't previously valid, opt-in to the new behavior. Something this means setting a default value for a new argument to `None`, so we can detect whether it was explicitly provided or not.
- Otherwise, check to see behavior would change for this usage in the future. If xarray's behavior would indeed change, issue a `FutureWarning` that suggests two alternative syntax options:
   - to maintain the current behavior in a backwards compatible way, or
   - to opt-in to the new behavior early

> The problem is that with separate `auto_combine` and `manual_combine` functions (which I think make sense to have), then the new `auto_combine`'s behaviour is so different to the old one that creating full backwards compatibility will basically involve having both the new and the old one present in the code simultaneously.

Yep, this is sometimes what we have to do.

TomNicholas: I'm trying to implement this now, but I've already found one case I don't know how to deal with:

`xarray.auto_combine([1d_list])` should be concatenated in order of appearance in the old version, but in coordinate order in the new version. However there's no way to know which one the user wants, and no way to give them the option to opt straight-in to the new behaviour, because the argument signatures are identical for the old and the new version in this case. To opt-in to the new way we would have to introduce a new argument just for that purpose, which would itself then be deprecated in the next version.

TomNicholas:
Okay @shoyer so the way I've gone about this deprecation problem is to essentially introduce `manual_combine` whilst reverting `auto_combine`. That means that `auto_combine` isn't a breaking change yet, but it's inputs are checked and if they will break in future it throws a warning saying "please use `manual_combine` instead because the behaviour of `auto_combine` will soon change".

This way we don't have two publicly-accessible versions of `auto_combine` present simultaneously, and if users heed the warnings and switch to using `manual_combine` then their code won't break when we later update `auto_combine` from the old to the new behaviour.

This approach means there currently isn't a way to opt-in to the new `auto_combine` behaviour early, but as you can opt-in to future-compatible behaviour by using `manual_combine` instead, then I think that's fine?

xarray/tests/test_backends.py
shoyer: For `open_mfdataset`, we can use the trick of setting `combine=None` as a default. Then:
- `combine=None`: backwards compatible behavior, raising warnings as needed
- `combine='auto'`: opt-in to automatic combination
- `combine='manual'`: stick with manual combination

TomNicholas: Okay, but then where do I put the description of what the new `auto_combine` does? And that means it would be possible to opt-in to the new `auto_combine` behaviour from `open_mfdataset`, but not from `auto_combine`!

shoyer:
Do we want to stick with the names `manual_combine` and `auto_combine`?

Some other possibilities:

- Instead of `manual_combine`: `concat_nd`, `concat_nested`, `multi_concat`
- Instead of `auto_combine`:`assemble`, `reassemble`, `combine_sorted`, `concat_sorted`, `unify`

I guess `auto`/`manual` works well with the values of the keyword argument, and makes it clear that these are a different class of operation than the primitive `concat` and `merge`. Though if we really want to emphasize the similarity (especially to help auto-complete), we might consider even switching the order of the  words, e.g., `combine_auto` and `combine_manual`.

TomNicholas:
I think that `manual_combine` and `auto_combine` (or `combine_auto` etc.) are pretty suitable. 

Any variation on `concat` is a little misleading because these functions will also `merge`. Anything with `nested` or `multi` is kind of superfluous when the idea of the library is sort of that multi-dimensional operations are as easy as single-dimensional operations. 

`combine` (dictionary definition: to join or merge to form a single unit or substance) does accurate describe what the functions do.

If we use `combine_auto` and `combine_manual` then it simplifies the deprecation cycle because we can have the old `auto_combine` and the new `combine_auto` in the API simultaneously (though maybe that's too confusing).

dcherian:
+1 for switching to `combine_auto`, `combine_manual` but not having both `auto_combine` and `combine_auto`.

shoyer:
> If we use combine_auto and combine_manual then it simplifies the deprecation cycle because we can have the old auto_combine and the new combine_auto in the API simultaneously (though maybe that's too confusing).

I think this is a good idea. We can deprecate the old auto_combine but keep it around for now to enable a smooth transition.

TomNicholas:
> We can deprecate the old auto_combine but keep it around for now to enable a smooth transition.

This might be a little bit confusing until the deprecation cycle is complete, but it will allow us to complete the cycle in only one edition, rather than two (one to introduce `combine_manual`, and one to change the behaviour of `auto_combine`). It also means that users who want to eventually use the "coordinate-ordering" behaviour will only have to change their code once.

I'll do it this way.

TomNicholas:
@shoyer I think this is ready for a final review.

dcherian:
I haven't reviewed the PR but noticed that there's no documentation to go along with it. Is that intentional?

TomNicholas:
@dcherian the relevant docstrings have been significantly altered, including giving some examples.

However I haven't changed any of the main documentation pages, I guess I should look through to see if this will change any of the syntax in any of those examples.

Or are you suggesting there should be a new section of the documentation explaining what the difference between `combine_manual` and `combine_auto` is?

dcherian:
> @dcherian the relevant docstrings have been significantly altered, including giving some examples.

Yes I saw that but missed https://github.com/pydata/xarray/blob/513764fa4094acda67aec5fea8397bc5726afb90/xarray/core/combine.py#L307

> However I haven't changed any of the main documentation pages, I guess I should look through to see if this will change any of the syntax in any of those examples.
> 
> Or are you suggesting there should be a new section of the documentation explaining what the difference between `combine_manual` and `combine_auto` is?

Yes, I think you can add the examples from the docstrings to https://xarray.pydata.org/en/stable/combining.html since that's where users will look for such functionality.

TomNicholas:
@dcherian I added another section to the https://xarray.pydata.org/en/stable/combining.html docs page to explain what the new functions do.

One thing - this deprecation cycle means that for now warnings get thrown basically however you use `auto_combine`. The test suites are generating a lot of these warnings - is there a way to get pytest to ignore specific warnings for specific tests or is it okay leaving it?

dcherian:
Thanks @TomNicholas. You can use `@pytest.mark.filterwarnings` to filter specific warnings. I think that's what the decorator is called anyway. There are places where it's used in the testing code so you have plenty of examples to copy from.

TomNicholas:
Ah yes that worked, thanks @dcherian !

I think this PR is finally complete - ready to merge @shoyer?

doc/combining.rst
```diff
@@ -237,3 +238,64 @@ coordinates as long as any non-missing values agree or are disjoint:
...
+For combining many objects along multiple dimensions xarray provides
+``combine_manual`` and ``combine_auto``. These functions use a combination of
+``concat`` and ``merge`` across different variables to combine many objects
+into one.
+
+``combine_manual`` requires specifying the order in which the objects should be
+combined, while ``combine_auto`` attempts to infer this ordering automatically
+from the coordinates in the data.
+
+``manual_combine`` is useful when you know the spatial relationship between
```
dcherian: should this be `combine_manual`?

doc/combining.rst
```diff
@@ -237,3 +238,64 @@ coordinates as long as any non-missing values agree or are disjoint:
...
+organising the datasets into a doubly-nested list, e.g:
+
+.. ipython:: python
+
+    arr = xr.DataArray(name='temperature', data=np.random.randint(5, size=(2, 2)), dims=['x', 'y'])
+    arr
+    ds_grid = [[arr, arr], [arr, arr]]
+    xr.combine_manual(ds_grid, concat_dim=['x', 'y'])
+
+``manual_combine`` can also be used to explicitly merge datasets with
```
dcherian: same here?

doc/whats-new.rst
```diff
@@ -149,6 +149,25 @@ Other enhancements
   report showing what exactly differs between the two objects (dimensions /
   coordinates / variables / attributes)  (:issue:`1507`).
   By `Benoit Bovy <https://github.com/benbovy>`_.
+- Combining datasets along N dimensions:
+  Datasets can now be combined along any number of dimensions,
+  instead of just a one-dimensional list of datasets.
+
+  The new ``manual_combine`` will accept the datasets as a a nested
```
dcherian: same here?

doc/whats-new.rst
```diff
@@ -149,6 +149,25 @@ Other enhancements
...
+  operations.
+
+  ``open_mfdataset`` can use ``manual_combine`` to combine datasets along
+  multiple dimensions, by specifying `combine='manual'`.
+
+  Some combinations of datasets will now throw FutureWarnings. To avoid these
+  switch to using `manual_combine` (or `combine='manual'` in `open_mfdataset`).
+  (:issue:`2159`) By `Tom Nicholas <http://github.com/TomNicholas>`_.
+
+- Resampling of standard and non-standard calendars indexed by
```
dcherian: bad merge

doc/combining.rst
```diff
@@ -237,3 +238,64 @@ coordinates as long as any non-missing values agree or are disjoint:
 Note that due to the underlying representation of missing values as floating
 point numbers (``NaN``), variable data type is not always preserved when merging
 in this manner.
+
+.. _combining.multi:
```
dcherian: Can you add a warning or note box talking about `auto_combine` vs `combine_auto`? I think that's going to be confusing, so the docs should be clear.

xarray/core/combine.py
```diff
@@ -456,107 +146,369 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
-def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
-                 compat='no_conflicts', data_vars='all', coords='different'):
-    """Attempt to auto-magically combine the given datasets into one.
+def combine_manual(datasets, concat_dim, compat='no_conflicts',
+                   data_vars='all', coords='different'):
+    """
+    Explicitly combine an N-dimensional grid of datasets into one by using a
+    succession of concat and merge operations along each dimension of the grid.
+
+    Does not sort data under any circumstances, so the datasets must be passed
```
dcherian: "does not sort coordinates"?

xarray/core/combine.py
```diff
@@ -567,6 +519,9 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
     variables, and each combination of a distinct time period and set of data
     variables is saved its own dataset.
 
+    This entire function is in the process of being deprecated in favour of
```
dcherian: let's move this to just below the first line

TomNicholas:
Thanks for the review - I've fixed those mistakes, and added a note explaining why there are 3 similar-sounding functions.

doc/combining.rst
```diff
@@ -237,3 +238,75 @@ coordinates as long as any non-missing values agree or are disjoint:
...
+
+  There are currently three combining functions with similar names:
+  ``auto_combine``, ``combine_auto``, and ``combine_manual``. This is because
+  ``auto_combine`` is in the process of being deprecated in favour of the other
+  two functions, which are more general. If your code currently relies on
+  ``auto_combine``, then you will be able to get similar functionality by using
+  ``combine_manual``.
+
+For combining many objects along multiple dimensions xarray provides
+``combine_manual`` and ``combine_auto``. These functions use a combination of
```
shoyer: To turn these references into links to the API documentation, you can write something like:

    :py:func:`~xarray.combine_manual`

which will print like [`combine_manual`](http://link-to-doc).

xarray/backends/api.py
```diff
@@ -591,38 +592,51 @@ def close(self):
...
+        Dimensions to concatenate files along.  You only
+        need to provide this argument if any of the dimensions along which you
+        want to concatenate is not a dimension in the original datasets, e.g.,
+        if you want to stack a collection of 2D arrays along a third dimension.
+        Set ``concat_dim=[..., None, ...]`` explicitly to
+        disable concatenation along a particular dimension.
+    combine : {'auto', 'manual'}, optional
+        Whether ``xarray.auto_combine`` or ``xarray.manual_combine`` is used to
+        combine all the data. Default is to use ``xarray.auto_combine``, but 
+        this function has been deprecated..
```
shoyer: Maybe instead of this last sentence:

    If this argument is not provided, `xarray.auto_combine` is used, but in the future this
    behavior will switch to use `xarray.combine_auto`.

xarray/backends/api.py
```diff
@@ -715,22 +729,17 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
-
-    # If infer_order_from_coords=True then this is unnecessary, but quick.
-    # If infer_order_from_coords=False then this creates a flat list which is
-    # easier to iterate over, while saving the originally-supplied structure
-    combined_ids_paths, concat_dims = _infer_concat_order_from_positions(
-        paths, concat_dims)
+    # If combine='auto' then this is unnecessary, but quick.
+    # If combine='manual' then this creates a flat list which is easier to
+    # iterate over, while saving the originally-supplied structure as "ids"
+    if combine is 'manual':
```
shoyer: Use `==` rather than `is` for comparing to strings. The later is not guaranteed to work.

xarray/backends/api.py
```diff
@@ -758,18 +767,28 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
...
+            combined = _manual_combine(datasets, concat_dims=concat_dim,
+                                       compat=compat, data_vars=data_vars,
+                                       coords=coords, ids=ids)
+        elif combine is 'auto':
+            # Redo ordering from coordinates, ignoring how they were ordered
+            # previously
+            combined = combine_auto(datasets, compat=compat,
+                                    data_vars=data_vars, coords=coords)
+        else:
+            raise ValueError("{} is an invalid option forthe keyword argument "
```
shoyer:
```suggestion
            raise ValueError("{} is an invalid option for the keyword argument "
```

xarray/core/combine.py
```diff
@@ -421,6 +45,71 @@ def _infer_tile_ids_from_nested_list(entry, current_pos):
         yield current_pos, entry
 
 
+def _infer_concat_order_from_coords(datasets):
+
+    concat_dims = []
+    tile_ids = [() for ds in datasets]
+
+    # All datasets have same variables because they've been grouped as such
+    ds0 = datasets[0]
```
shoyer: Do we always know that `datasets` is non-empty?

Let's make sure that `combine_auto()` and `combine_manual()` do something reasonable in the case of empty input, either by returning an empty Dataset (preferred) or raising an error.

xarray/core/combine.py
```diff
@@ -421,6 +45,71 @@ def _infer_tile_ids_from_nested_list(entry, current_pos):
...
+                                     "monotonically decreasing on all datasets"
+                                     .format(dim))
+
+                # Assume that any two datasets whose coord along dim starts
+                # with the same value have the same coord values throughout.
+                try:
+                    first_items = pd.Index([index.take([0])
+                                            for index in indexes])
+                except IndexError:
+                    raise ValueError('Cannot handle size zero dimensions')
```
shoyer: This is a minor point, but it would be a little less error prone to explicitly check the size of each sub-index, e.g., `if any(index.size == 0 for index in indexes)`.

xarray/core/combine.py
```diff
@@ -456,107 +146,372 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
     """
 
-    # Perform N-D dimensional concatenation
+    tile_id, ds = list(combined_ids.items())[0]
+
+    n_dims = len(tile_id)
+    if len(concat_dims) != n_dims:
+        raise ValueError("concat_dims has length {} but the datasets "
+                         "passed are nested in a {}-dimensional structure"
+                         .format(str(len(concat_dims)), str(n_dims)))
```
shoyer: You can omit the extra calls to `str()`. `.format()` already does that by default when you write `{}`, so this is exactly equivalent: 
```suggestion
                         .format(len(concat_dims), n_dims))
```

xarray/core/combine.py
```diff
@@ -456,107 +146,372 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
+        keys in combined_ids. If the string is a dimension name then concat
+        along that dimension, if it is None then merge.
 
     Returns
     -------
     combined_ds : xarray.Dataset
     """
 
-    # Perform N-D dimensional concatenation
+    tile_id, ds = list(combined_ids.items())[0]
```
shoyer: A few minor suggestions:
- If you don't need a variable (e.g., `ds` in this case), then it's convention to assign it to the name `_`, which indicates that it isn't used. But in this case, I think it would work simply to use `.keys()` instead of `.items()`.
- Rather than constructing a list, we can get the first element just by use the builtins `iter()` and `next()`. In principle this could be much faster, because it avoids constructing a list just to look at only the first item.
- Maybe use a more descriptive name than `tile_id`?

So this could be:
```suggestion
    example_tile_id = next(iter(combined_ids.keys()))
```

xarray/core/combine.py
```diff
@@ -456,107 +146,372 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
-                                                         dim=concat_dim,
-                                                         data_vars=data_vars,
-                                                         coords=coords,
-                                                         compat=compat)
+        combined_ids = _combine_all_along_first_dim(combined_ids,
+                                                    dim=concat_dim,
+                                                    data_vars=data_vars,
+                                                    coords=coords,
+                                                    compat=compat)
     combined_ds = list(combined_ids.values())[0]
```
shoyer: At this point, we know there should only be one dataset left, right? In that case, it's often cleaner to use tuple unpacking, because it makes it clear to readers of the code that we aren't throwing anything away, e.g.,
```suggestion
    (combined_ds,) = combined_ids.values()
```

xarray/core/combine.py
```diff
@@ -456,107 +146,372 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
-                                     data_vars=data_vars, coords=coords)
-                        for id, ds_group in grouped_by_vars]
+        try:
+            combined = concat(datasets, dim=concat_dim, data_vars=data_vars,
+                              coords=coords)
+        except ValueError as err:
+            if "encountered unexpected variable" in str(err):
+                raise ValueError("These objects cannot be combined along the "
+                                 "dimension {concat_dim} using only "
+                                 "xarray.concat, you must use "
```
shoyer: I'm not sure I understand this error message. This is used inside `combine_auto` or `auto_combine` or `combine_manual`, right? It seems a little confusing to mention `xarray.concat` when users may not have mentioned it explicitly.

xarray/core/combine.py
```diff
@@ -456,107 +146,372 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
-        sorted_datasets = sorted(datasets, key=vars_as_keys)
-        grouped_by_vars = itertools.groupby(sorted_datasets, key=vars_as_keys)
-        concatenated = [_auto_concat(list(ds_group), dim=dim,
-                                     data_vars=data_vars, coords=coords)
-                        for id, ds_group in grouped_by_vars]
+        try:
+            combined = concat(datasets, dim=concat_dim, data_vars=data_vars,
+                              coords=coords)
+        except ValueError as err:
+            if "encountered unexpected variable" in str(err):
```
shoyer: This is a sign that we may want to refactor `concat`, though we can definitely save that for later :). Generally string matching of errors can be somewhat error prone -- it's better to catch a dedicated exception type.

xarray/core/combine.py
```diff
@@ -456,107 +146,372 @@ def _combine_nd(combined_ids, concat_dims, data_vars='all',
...
+
+_CONCAT_DIM_DEFAULT = '__infer_concat_dim__'
+
+
+def auto_combine(datasets, concat_dim='_not_supplied', compat='no_conflicts',
+                 data_vars='all', coords='different'):
+    """
+    Attempt to auto-magically combine the given datasets into one.
+
+    This entire function is in the process of being deprecated in favour of
```
shoyer:
```suggestion
    This entire function is deprecated in favour of
```

xarray/core/combine.py
```diff
@@ -605,21 +560,141 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
...
+
+    if concat_dim is '_not_supplied':
+        concat_dim = _CONCAT_DIM_DEFAULT
+    else:
+        message = """In xarray version 0.14 `auto_combine` will be deprecated,
+                  and `open_mfdataset` will no longer accept a `concat_dim`
+                  argument. To get equivalent behaviour from now on please use
+                  the new `combine_manual` function instead (or the
+                  `combine='manual'` option to open_mfdataset)."""
+        warnings.warn(message, FutureWarning)
```
shoyer: Please set `stacklevel` when you call `warnings.warn()` so users get an error message where they used `xarray.auto_combine()` in their code. In most cases, I think the right level is `stacklevel=2` or maybe stacklevel=3`, but you should experiment a little to make sure.

xarray/core/combine.py
```diff
@@ -605,21 +560,141 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
...
-    if concat_dim is None or concat_dim == _CONCAT_DIM_DEFAULT:
-        concat_dims = concat_dim
-    elif not isinstance(concat_dim, list):
-        concat_dims = [concat_dim]
+    """
+
+    if concat_dim is '_not_supplied':
+        concat_dim = _CONCAT_DIM_DEFAULT
+    else:
+        message = """In xarray version 0.14 `auto_combine` will be deprecated,
```
shoyer: This message will get formatted in a weird way, due to the extra indentation.  A nice way to handle multi-line strings like these while keeping the code formatted well is to use `textwrap.dedent()` from Python's standard library.

xarray/core/concat.py
```diff
@@ -0,0 +1,339 @@
+from __future__ import absolute_import, division, print_function
```
shoyer: We don't need these `__future__` imports anymore, now that we're Python 3 only :)

xarray/tests/test_backends.py
```diff
@@ -2163,6 +2163,8 @@ def skip_if_not_engine(engine):
         pytest.importorskip(engine)
 
 
+@pytest.mark.filterwarnings("ignore:In xarray version 0.14 `auto_combine` "
+                            "will be deprecated")
```
shoyer: I know it would probably be save to ignore these given that the output of these tests is probably not going to change, but generally we try to update xarray's test suite to avoid our own warnings  rather than filtering them. This helps identify any possible issues with the new warnings, because we can see how pervasive the effect is on our own code.

xarray/core/combine.py
```diff
@@ -605,21 +560,141 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
...
+                                    "neither monotonically increasing nor",
+                                    "Cannot handle size zero",
+                                    "Could not find any dimension coordinates"]
+        if any(message in str(err) for message in no_dimension_coords_errs):
+            # The ValueError just means that the datasets don't have
+            # global dimension coordinates
+            return False
+        else:
+            # There is a different problem
+            raise err
```
shoyer: Given that this is an internal check only used for determining the right deprecation warning, it might be better to swallow this error instead.

xarray/core/combine.py
```diff
@@ -605,21 +560,141 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
...
-                         ids=False)
+        message = """In xarray version 0.14 `auto_combine` will be deprecated.
+                  The datasets supplied do not have global dimension
+                  coordinates. In future, to continue concatenating without
+                  supplying dimension coordinates, please use the new
+                  `combine_manual` function (or the `combine='manual'` option
+                  to open_mfdataset)."""
+        warnings.warn(message, FutureWarning)
+
+    if _requires_concat_and_merge(datasets):
```
shoyer: It looks like up to three different FutureWarnings could be issued when a user calls this function.

I think it's fantastic how you attempt to diagnose the issue to give the user the best possible guidance, but could we consolidate all this into a single warning, e.g., by building up the warning message dynamically?

xarray/tests/test_backends.py
TomNicholas: > try to update xarray's test suite to avoid our own warnings rather than filtering them

I don't think that's possible in this case - `auto_combine` will throw a `FutureWarning` whatever arguments you give it, but we still want to test it. This is only temporary though - once `auto_combine` is fully deprecated all the tests which filter warnings will be deleted.

xarray/core/combine.py
TomNicholas: I had to special case this in `_manual_combine`, but I've added tests to ensure that an empty Dataset is always returned.

xarray/core/combine.py
TomNicholas: Sorry I'm not sure exactly what you mean... If we don't raise any remaining errors here then the code will just continue until it errors in some less explicit way?

xarray/core/combine.py
TomNicholas: This is the most subtle part of the whole PR. It's covering the case where you want to combine datasets which require both concatenation and merging. For example (from `test_manual_combine_but_need_auto_combine`):

```python
objs = [Dataset({'x': [0, 1]}), Dataset({'x': [2], 'wall': [0]})]
with raises_regex(ValueError, 'cannot be combined'):
    combine_manual(objs, concat_dim='x')
```

You can't use `combine_manual` for this case because you can't use either `concat` or `merge` along the first (and only) axis of the `objs` array, you have to use a combination of both. However `combine_manual` works by applying just one of those operations to each axis of the input object array. The error message is telling the user to use `combine_auto` instead, because that function first groups all the variables according to their dims before concatenation, then merges them back together, so could handle a case like this.

This message is covering the edge case we discussed [here](https://github.com/pydata/xarray/pull/2553#issuecomment-445419876).

> This is used inside combine_auto or auto_combine or combine_manual, right?

This internal function (`_combine_1d`) is used inside all of those, but this error should only ever be raised by `combine_manual`, because only `combine_manual` could end up passing a combination of datasets like this to `_combine_1d`.

> It seems a little confusing to mention xarray.concat when users may not have mentioned it explicitly.

The fact we're using `concat` is supposed to be understood from the description of `combine_manual`. I agree that this is confusing - perhaps the message should just say something like "can't handle this combination of variables, instead use either `combine_auto` or just do it manually with `concat` and `merge`"?

xarray/core/combine.py
TomNicholas: > could we consolidate all this into a single warning?

Done. It is potentially a pretty long warning, so I'm not sure if that's much better though.

TomNicholas:
Thanks for the detailed review @shoyer !

> I'm loving all the documentation and test coverage here (though I haven't read them all carefully yet!).

Great - the only test I think you should definitely look at is `test_manual_concat_too_many_dims_at_once`, and my diagnosis of the issue it exposes in #2975.

xarray/core/combine.py
shoyer: > You can't use `combine_manual` for this case because you can't use either `concat` or `merge` along the first (and only) axis of the `objs` array, you have to use a combination of both. However `combine_manual` works by applying just one of those operations to each axis of the input object array. The error message is telling the user to use `combine_auto` instead, because that function first groups all the variables according to their dims before concatenation, then merges them back together, so could handle a case like this.

In principle, I think you could use `align()` in this case (possibly with `exclude` to skip some dimensions), and then use `combine_manual` afterwards. 

> The fact we're using `concat` is supposed to be understood from the description of `combine_manual`.

If possible, I think it would be best to treat the usage of `concat` and `merge` inside `combine` as an implementation detail.

> I agree that this is confusing - perhaps the message should just say something like "can't handle this combination of variables, instead use either `combine_auto` or just do it manually with `concat` and `merge`"?

Yes, that would be better -- but also add `align` to that list of functions to use :)

xarray/tests/test_backends.py
shoyer: I think this test uses `open_mfdataset`, so we could probably just add the keyword argument that triggers the new code path. Or we could add an explicit `pytest.assert_warns()` check, which would fail if the warning goes away. Either way would be preferable, because it means we won't have to remember to update this later.

xarray/core/combine.py
TomNicholas: > In principle, I think you could use align() in this case (possibly with exclude to skip some dimensions), and then use combine_manual afterwards.

If this comes up in future then I guess we could add it as an example to the documentation.

> If possible, I think it would be best to treat the usage of concat and merge inside combine as an implementation detail.

That works for `combine_auto`, but `combine_manual` was always supposed to be the more explicit version, and the docs for that function do explain how it uses `concat` and `merge`.

> Yes, that would be better -- but also add align to that list of functions to use :)

Okay cool, I will do that now

xarray/tests/test_backends.py
TomNicholas: > I think this test uses open_mfdataset, so we could probably just add the keyword argument that triggers the new code path.

We could do that, but then the old code path wouldn't be tested anymore?

> Or we could add an explicit pytest.assert_warns() check, which would fail if the warning goes away.

Happy to do that too

shoyer: I think the underlying `auto_combine` has tests somewhere?

Otherwise, or if you're not sure those are sufficient, we can do the explicit `assert_warns`. It's definitely good to check that warnings are actually issued, but let's keep them in the minimum number of places that code will just get deleted later.

TomNicholas: > I think the underlying auto_combine has tests somewhere?

Actually yeah you're right, that code path will still be sufficiently tested.

> good to check that warnings are actually issued

Okay so just change the tests in `test_backends.py` to circumvent the deprecation warnings? As we are still checking that warnings are raised in `test_combine.py`.

shoyer: > Okay so just change the tests in `test_backends.py` to circumvent the deprecation warnings? As we are still checking that warnings are raised in `test_combine.py`.

Yes, but let's keep one test using `open_mfdataset()` that checks for the custom warning issued there.

xarray/core/combine.py
shoyer: This helper function is only used inside `auto_combine` as part of figuring out the proper deprecation warning. So the conservative thing to do would be to issue some sort of "default" deprecation warning if something unexpected happens, rather than to raise a new error.

In particular, I'm a little concerned that your checks above in `no_dimension_coords_errs` might not always do the right thing -- they are matching strings in error messages, which can be fragile. If `ValueError` is raised for another reason, the user could get a surprising traceback.

xarray/tests/test_backends.py
TomNicholas: > Okay so just change the tests in test_backends.py to circumvent the deprecation warnings?

Done.

> Yes, but let's keep one test using open_mfdataset() that checks for the custom warning issued there.

Turns out I already wrote these - see `TestOpenMFDataSetDeprecation`. So the only test changes needed when completing this deprecation cycle are deleting the test classes `TestOpenMFDataSetDeprecation`, `TestAutoCombineOldAPI` and `TestAutoCombineDeprecation`. (Also you could optionally remove some `combine` arguments to `open_mfdataset` as there will be a new default.)

xarray/core/combine.py
TomNicholas: > issue some sort of "default" deprecation warning if something unexpected happens

I could raise a default deprecation warning, then if no errors are raised while trying to get through `_dimension_coords_exist()` raise another warning with more info? Then the user will always get a deprecation warning regardless of what error is thrown?

> matching strings in error messages, which can be fragile

Okay, but what other way is there of doing this? We have to distinguish between these errors if we want to be able to check what type of input the user gave and still carry on going.

> If ValueError is raised for another reason, the user could get a surprising traceback.

If `ValueError` is raised for another reason then the user will see that because it will be caught and raised, no?

Do you mean that I should raise the particular error caught by `_infer_concat_order_from_coords` as a second warning?

xarray/core/combine.py
```diff
@@ -620,21 +581,142 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
...
-    # (see https://github.com/pydata/xarray/pull/2553/#issuecomment-445892746)
-    if concat_dim is None or concat_dim == _CONCAT_DIM_DEFAULT:
-        concat_dims = concat_dim
-    elif not isinstance(concat_dim, list):
-        concat_dims = [concat_dim]
+    """
+
+    message = """In xarray version 0.14 `auto_combine` will be deprecated."""
+
+    if concat_dim is '_not_supplied':
```
shoyer: I should have caught this before, but please always use `==` for string comparisons.

xarray/core/combine.py
shoyer: > I could raise a default deprecation warning, then if no errors are raised while trying to get through `_dimension_coords_exist()` raise another warning with more info? Then the user will always get a deprecation warning regardless of what error is thrown?

This sounds like a good solution to me.

> Okay, but what other way is there of doing this? We have to distinguish between these errors if we want to be able to check what type of input the user gave and still carry on going.

The alternative is either (1) making dedicated error classes or (2) refactoring into lower level functions with more specific functionality. That said, it's probably not worth bothering with either of these here.


> > issue some sort of "default" deprecation warning if something unexpected happens
> 
> I could raise a default deprecation warning, then if no errors are raised while trying to get through `_dimension_coords_exist()` raise another warning with more info? Then the user will always get a deprecation warning regardless of what error is thrown?
> 
> > matching strings in error messages, which can be fragile
> 
> Okay, but what other way is there of doing this? We have to distinguish between these errors if we want to be able to check what type of input the user gave and still carry on going.
> 
> > If ValueError is raised for another reason, the user could get a surprising traceback.
> 
> If `ValueError` is raised for another reason then the user will see that because it will be caught and raised, no?
> 
> Do you mean that I should raise the particular error caught by `_infer_concat_order_from_coords` as a second warning?

I think it's better just to silence the error.

xarray/core/combine.py
```diff
@@ -620,21 +581,142 @@ def auto_combine(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
...
     else:
-        concat_dims = concat_dim
-    infer_order_from_coords = False
-
-    # The IDs argument tells _auto_combine that the datasets are not yet sorted
-    return _auto_combine(datasets, concat_dims=concat_dims, compat=compat,
-                         data_vars=data_vars, coords=coords,
-                         infer_order_from_coords=infer_order_from_coords,
-                         ids=False, fill_value=fill_value)
+        return False
```
shoyer: This could just be `return len(list(grouped_by_vars)) > 1` now

xarray/tests/test_concat.py
```diff
@@ -0,0 +1,322 @@
+from __future__ import absolute_import, division, print_function
```
shoyer: We don't need `__future__` imports now that we're Python 3 only :)

shoyer:
I left a couple of minor comments, but I think this is pretty much ready to go! We should merge this and let users give it a try :).

TomNicholas:
Okay I've changed those things. @shoyer if you want to just double check that the new error catching behaviour in `dimension_coords_exist()` is what you wanted, then this should be good to go!

shoyer:
OK, in  it goes! Thanks @TomNicholas for your perseverance here

max-sixty:
Thanks @TomNicholas !

TomNicholas:
:tada: Thanks for your help!

shoyer:
I started thinking about names for these functions again....

I don't love the non-descriptive nature of `combine_auto` and `combine_manual`. All "auto" and "manual" tells me is that one is more automatic than the other, without saying *how* it's more automatic. This is probably even more confusing because we used "auto" previously to mean some hybrid behavior between what we now call "manual" and "auto".

What about calling these `combine_by_coords` and `combine_nested` instead? On `open_mfdataset`, these would be `combine='by_coords'` and `combine='nested'`.

TomNicholas:
@shoyer I like that suggestion a lot - not sure why we didn't think of something similar before.

How does it work if I push to a merged PR?

shoyer:
Pushing to a merged PR definitely doesn't work :). You should start a new branch after syncing master.

Issue #1736, opened 2017-11-22
Rasterio missing _FillValue in DataArray


snowman2:
#### Problem description

When xarray opens a dataset, it stores the the _FillValue in the encoding and replaces values with NaN values. However, using open_rasterio, this behavior does not occur and the _FillValue value is missing. It only has the transform, crs, is_tiled, and res attributes. Also, the encoding is empty.

#### Expected Output
It would be nice to have the _FillValue  as an attribute or in the encoding.

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.10.0-40-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0rc2
pandas: 0.21.0
numpy: 1.13.2
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: dev_20170921-05806a2
bottleneck: 1.2.1
cyordereddict: None
dask: 0.15.4
matplotlib: 2.1.0
cartopy: None
seaborn: None
setuptools: 36.6.0
pip: 9.0.1
conda: None
pytest: 3.2.5
IPython: 6.2.1
sphinx: None
</details>

fmaussion:
I agree that it would be nice to have. It seems that the corresponding rasterio attribute is `nodatavals` (https://github.com/mapbox/rasterio/blob/master/docs/topics/masks.rst). It shouldn't be too difficult to add to the existing attributes: https://github.com/pydata/xarray/blob/master/xarray/backends/rasterio_.py#L142

fmaussion:
What I'm not sure, though, is if we should rename the attribute to  _FillValue or keep its name as it is. I am also not sure if we should do the automatic conversion as it is done with the netCDF backends. Quite often geotiff files store iteger types, and filling with NaNs require floats. 

Right now I would tend to just parse the `nodatavals` attribute and leave everything else to the user.

snowman2:
[...]

PR #2618, opened 2018-12-18
Adding mask to open_rasterio

jsignell:
- [x] Discussed in #1736
 - [x] Tests added

Not sure if this is the right approach @snowman2

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -165,7 +165,7 @@ def _parse_envi(meta):
 
 
 def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
-                  lock=None, mask=False):
+                  lock=None, masked=False):
     """Open a file with rasterio (experimental).
 
     This should work with any file that rasterio can open (most often:
@@ -208,7 +208,7 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         :py:func:`dask.array.from_array`. By default, a global lock is
         used to avoid issues with concurrent access to the same file when using
         dask's multithreaded backend.
-    mask : bool, optional
+    masked : bool, optional
         If True, uses nodatavals to set values to NaN. Defaults to False.
 
     Returns
@@ -307,7 +307,7 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         nodatavals = tuple(
             np.nan if nodataval is None else nodataval
             for nodataval in riods.nodatavals)
-        if mask:
+        if masked:
             encoding['nodatavals'] = nodatavals
         else:
             attrs['nodatavals'] = nodatavals
@@ -329,7 +329,7 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
                 attrs[k] = v
 
     data = indexing.LazilyOuterIndexedArray(
-        RasterioArrayWrapper(manager, lock, vrt_params, masked=mask))
+        RasterioArrayWrapper(manager, lock, vrt_params, masked=masked))
 
     # this lets you write arrays loaded with rasterio
     data = indexing.CopyOnWriteArray(data)
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -162,7 +162,7 @@ def _parse_envi(meta):
 
 
 def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
-                  lock=None):
+                  lock=None, mask=False):
     """Open a file with rasterio (experimental).
 
     This should work with any file that rasterio can open (most often:
@@ -205,6 +205,8 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         :py:func:`dask.array.from_array`. By default, a global lock is
         used to avoid issues with concurrent access to the same file when using
         dask's multithreaded backend.
+    mask : bool, optional
+        If True, uses nodatavals to set values to NaN. Defaults to False.
 
     Returns
     -------
@@ -328,6 +330,10 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
 
     result = DataArray(data=data, dims=('band', 'y', 'x'),
                        coords=coords, attrs=attrs)
+
+    if mask:
+        for nodataval in attrs.get('nodatavals', ()):
+            result = result.where(result!=nodataval)
 
     if chunks is not None:
         from dask.base import tokenize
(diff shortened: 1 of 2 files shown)


snowman2:
@jsignell, this looks like it would work - I like the implementation. Some things I would recommend to add is checks for when the nodata value varies across bands (maybe skip the mask and raise a warning) and when the nodata value is already NaN. Also, you should probably set the nodatavals to NaN if masked to resemble the behavior of xarray elsewhere.

I think @fmaussion would be a good reviewer for this PR.

xarray/backends/rasterio_.py
```diff
@@ -298,6 +300,10 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
     result = DataArray(data=data, dims=('band', 'y', 'x'),
                        coords=coords, attrs=attrs)
 
+    if mask:
+        for nodataval in attrs.get('nodatavals', ()):
+            result = result.where(result != nodataval)
```
shoyer: The downside of this approach is that unlike how to decode metadata for netCDF files, this won't be lazy.

See `CFMaskCoder` in xarray/coding/variables.py for who we do this with netCDF files. In particular, it probably makes sense to use `lazy_elemwise_func` and to move `nodatavals` from `attrs` to `encoding`.

fmaussion:
I wonder if `nodatavals` is the only attribute for masking - in the doc (https://rasterio.readthedocs.io/en/latest/topics/masks.html) they also mention `nodata`, which in my experience is also used

snowman2:
I think the solution would be to add a masked argument and property to the class RasterioArrayWrapper [here](https://github.com/pydata/xarray/blob/778ffc49135d6f97e17b37b48304995fca72f1e0/xarray/backends/rasterio_.py#L24).

And pass in the (mased=masked) argument [here](https://github.com/pydata/xarray/blob/778ffc49135d6f97e17b37b48304995fca72f1e0/xarray/backends/rasterio_.py#L291)

That would pass riods.read( ..., masked=self.masked) when reading data from rasterio [here](https://github.com/pydata/xarray/blob/778ffc49135d6f97e17b37b48304995fca72f1e0/xarray/backends/rasterio_.py#L115)

jsignell:
I was wondering if we should just pass the mask option to rasterio like that. I'll try it out.

jsignell:
I did as @snowman2 suggested but filled the mask with nans at the end. Since it is now using rasterio directly this should address @fmaussion's concern about `nodata` (which I think was ending up in the `nodatavals` attr anyways). I also moved the `nodatavals` from `attrs` to `encoding` when `mask==True`,  but I didn't set it to true by default since I didn't want to break backwards compatibility.

xarray/backends/rasterio_.py
```diff
@@ -268,9 +274,13 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         attrs['is_tiled'] = np.uint8(riods.is_tiled)
     if hasattr(riods, 'nodatavals'):
         # The nodata values for the raster bands
-        attrs['nodatavals'] = tuple(
+        nodatavals = tuple(
             np.nan if nodataval is None else nodataval
             for nodataval in riods.nodatavals)
+        if mask:
+            encoding['nodatavals'] = nodatavals
```
snowman2: I thing setting the encoding should inline with the other if statements and should be something like:
```python
if mask and riods.nodata is not None:
    encoding['_FillValue'] = riods.nodata
```

snowman2: @shoyer is this what you intended or did you want to keep it as @jsignell modified it to be?

fmaussion:
> Do we really want a np.masked_array or do we just want an array with np.Nans in it? I prefer nans, but I guess the downside is type conversion for ints?

Masked arrays come with huge performance drops, I don't think we should use these.

shoyer:
xarray will autoconvert a masked array into a numpy array with NaNs. It
probably doesn't make a difference where this happens.

fmaussion:
I'm sorry this has stalled now - I think this is a nice addition. 

@jsignell : would you have time to finish this up?

jsignell:
Yeah I have some time to finish it up. Looking back at it now it seems like `masked` would be a better kwarg for open_rasterio. Is there anything else that should be changed?

snowman2:
I think it would be good to write the dataset to a netcdf, and read it back in with xarray.open_dataset to ensure everything encodes/decides properly.

jsignell:
> I think it would be good to write the dataset to a netcdf, and read it back in with xarray.open_dataset to ensure everything encodes/decides properly.

Do you mean write a test to do this, or just check it locally?

snowman2:
Good question, I was thinking adding tests.

xarray/backends/rasterio_.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, manager, lock, vrt_params=None):
         if not np.all(np.asarray(dtypes) == dtypes[0]):
             raise ValueError('All bands should have the same dtype')
         self._dtype = np.dtype(dtypes[0])
```
snowman2:
```suggestion
        self._dtype = np.float64 if self.masked else np.dtype(dtypes[0])
```

xarray/backends/rasterio_.py
```diff
@@ -119,7 +120,9 @@ def _getitem(self, key):
                 riods = self.manager.acquire(needs_lock=False)
                 if self.vrt_params is not None:
                     riods = WarpedVRT(riods, **self.vrt_params)
-                out = riods.read(band_key, window=window)
+                out = riods.read(band_key, window=window, masked=self.masked)
+                if self.masked:
+                    out = np.ma.filled(out, np.nan)
```
snowman2:
```suggestion
                    out = np.ma.filled(out.astype(np.float64), np.nan)
```

xarray/backends/rasterio_.py
```diff
@@ -319,15 +329,15 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
...
+        RasterioArrayWrapper(manager, lock, vrt_params, masked=masked))
 
     # this lets you write arrays loaded with rasterio
     data = indexing.CopyOnWriteArray(data)
     if cache and chunks is None:
         data = indexing.MemoryCachedArray(data)
 
     result = DataArray(data=data, dims=('band', 'y', 'x'),
-                       coords=coords, attrs=attrs)
+                       coords=coords, attrs=attrs, encoding=encoding)
```
snowman2:
```suggestion
                       coords=coords, attrs=attrs)
        result.encoding = encoding
```

snowman2:
@jsignell, the `masked` kwarg is in `rioxarray` if you need it: https://corteva.github.io/rioxarray/html/examples/clip_geom.html

keewis:
this has been unintentionally closed when renaming the branch to merge into to `main`, but given that we consider deprecating `open_rasterio` in favor of the `rasterio` engine provided by `rioxarray` (which seems to already support the `mask` kwarg) that's probably not a mistake. If you disagree feel free to reopen this PR.

jsignell:
I'm fine with this being closed. TBH I had totally forgotten about it :grimacing:

PR #2621, opened 2018-12-19
Fix multiindex selection

### LINKED ISSUE #2619: Selection of MultiIndex makes following `unstack` wrong

```python
import numpy as np 
import xarray as xr 

ds = xr.DataArray(np.arange(40).reshape(8, 5), dims=['x', 'y'],  
                  coords={'x': np.arange(8), 'y': np.arange(5)}).stack(xy=['x', 'y']) 
ds.isel(xy=ds['x'] < 4).unstack() 

Out[1]: 
<xarray.DataArray (x: 8, y: 5)>
array([[ 0.,  1.,  2.,  3.,  4.],
       [ 5.,  6.,  7.,  8.,  9.],
       [10., 11., 12., 13., 14.],
       [15., 16., 17., 18., 19.],
       [nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan]])
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7
  * y        (y) int64 0 1 2 3 4
```

#### Problem description

After unstack, there are still values that are not selected by the previous `isel`.
Probably the upstream bug?

#### Expected Output

```python
Out[1]: 
<xarray.DataArray (x: 8, y: 5)>
array([[ 0.,  1.,  2.,  3.,  4.],
       [ 5.,  6.,  7.,  8.,  9.],
       [10., 11., 12., 13., 14.],
       [15., 16., 17., 18., 19.]])
Coordinates:
  * x        (x) int64 0 1 2 3
  * y        (y) int64 0 1 2 3 4
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.1.final.0
python-bits: 64
OS: Linux
OS-release: 4.15.0-42-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.9
pandas: 0.23.4
numpy: 1.15.4
scipy: 1.1.0
netCDF4: 1.4.2
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.0.2.1
PseudonetCDF: None
rasterio: None
iris: None
bottleneck: None
cyordereddict: None
dask: 1.0.0
distributed: 1.25.0
matplotlib: 3.0.1
cartopy: None
seaborn: None
setuptools: 40.5.0
pip: 18.1
conda: None
pytest: 4.0.1
IPython: 7.1.1
sphinx: None

</details>

fujiisoup:
- [x] Closes #2619
 - [x] Tests added

Fix using ` MultiIndex.remove_unused_levels()`

### DIFF
--- a/xarray/core/pdcompat.py
+++ b/xarray/core/pdcompat.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pandas as pd
 
 
--- a/xarray/core/pdcompat.py
+++ b/xarray/core/pdcompat.py
@@ -1,0 +1,77 @@
+import pandas as pd
+
+
+# for pandas 0.19
+def remove_unused_levels(self):
+    """
+    create a new MultiIndex from the current that removing
+    unused levels, meaning that they are not expressed in the labels
+    The resulting MultiIndex will have the same outward
+    appearance, meaning the same .values and ordering. It will also
+    be .equals() to the original.
+    .. versionadded:: 0.20.0
+    Returns
+    -------
+    MultiIndex
+    Examples
+    --------
+    >>> i = pd.MultiIndex.from_product([range(2), list('ab')])
+    MultiIndex(levels=[[0, 1], ['a', 'b']],
+               labels=[[0, 0, 1, 1], [0, 1, 0, 1]])
+    >>> i[2:]
+    MultiIndex(levels=[[0, 1], ['a', 'b']],
+               labels=[[1, 1], [0, 1]])
+    The 0 from the first level is not represented
+    and can be removed
+    >>> i[2:].remove_unused_levels()
+    MultiIndex(levels=[[1], ['a', 'b']],
+               labels=[[0, 0], [0, 1]])
+    """
+
+    new_levels = []
+    new_labels = []
+
+    changed = False
+    for lev, lab in zip(self.levels, self.labels):
+
+        # Since few levels are typically unused, bincount() is more
+        # efficient than unique() - however it only accepts positive values
+        # (and drops order):
+        uniques = np.where(np.bincount(lab + 1) > 0)[0] - 1
+        has_na = int(len(uniques) and (uniques[0] == -1))
+
+        if len(uniques) != len(lev) + has_na:
+            # We have unused levels
+            changed = True
+
+            # Recalculate uniques, now preserving order.
+            # Can easily be cythonized by exploiting the already existing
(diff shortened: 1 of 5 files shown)


fujiisoup:
@shoyer 

This issue does not matter only in `unstack` but also in `sel` (reindex also?).
We can do it in both methods for efficiency, but I just prefered to do it in one place for the quick fix.

But what is the good place to do it actually?
If we invoke this in `sel`, it would be again inefficient as this is called every time we did `.sel`.

Probably the best for the efficiency is to keep a flag for it in `PandasIndexAdapter` and do it in `_get_item_with_mask` and `get_loc` only if flag is raised?

shoyer:
> This issue does not matter only in unstack but also in sel (reindex also?).

Can you explain why matters for `.sel()`? I guess it slows down repeated indexing?

My inclination was just to copy what pandas does (which is only removing unused levels in `unstack`)

fujiisoup:
But this problem of `.sel` can be simply solved by manually raising a keyerror if the result of get_loc is size 0 array.

fujiisoup:
I moved `MultiIndex.remove_unused_levels()` to the inside of `unstack`. Now it is invoked only when `unstack` is called.

xarray/core/pdcompat.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+import pandas as pd
+import pandas.core.algorithms as algos
```
shoyer: can we do a local import here instead the function?

I'm a little nervous that some future version of pandas may drop this (private) module, which would then break imports even though this isn't actually used.

xarray/core/pdcompat.py
```diff
@@ -0,0 +1,79 @@
...
+
+
```
shoyer: Can you copy the pandas license directly into this file, too? See `coding/cftimesindex.py` for an example.

xarray/core/pdcompat.py
fujiisoup: Good catch! Fixed.

xarray/core/pdcompat.py
fujiisoup: Done

PR #2625, opened 2018-12-21
Get 0d slices of ndarrays directly from indexing

### LINKED ISSUE #2622: Unnecessary copy when indexing to obtain a 0d array
#### Code Sample
```python
>>> import numpy as np
>>> import xarray as xr
>>> da = xr.DataArray(np.arange(3))
>>> da
<xarray.DataArray (dim_0: 3)>
array([0, 1, 2])
>>> da[0].values.fill(99)
>>> da
<xarray.DataArray (dim_0: 3)>
array([0, 1, 2])
```
#### Problem description
Indexing into xarray objects creates a view of the underlying data if possible. A surprising exception is when all dimensions are indexed out and the resulting object is 0d. Xarray insists on returning a 0d array rather than a scalar, which suggests (at least to me) that this is also a view whenever possible; however, it is always a copy, and modifying it will never affect the original array.

(The example above is a little contrived, since one could always call `da[0] = 99`. In my actual use case I am indexing into a Dataset in a way that creates views for all variables except the one that happens to collapse to 0d, and thus I'm unable to use the indexed Dataset to modify that variable in the original Dataset.) 

The copy happens because, internally, the 0d array is created by retrieving a scalar from the underlying numpy array and then wrapping a new array around it. However, in numpy a 0d view can be created directly by indexing with `Ellipsis`/`...`, as follows:
```python
>>> import numpy as np
>>> arr = np.arange(3)
>>> arr[0, ...]
array(0)
```
Thus, a fix that solves my immediate issues and passes all current tests is to modify the following method:
https://github.com/pydata/xarray/blob/778ffc49135d6f97e17b37b48304995fca72f1e0/xarray/core/indexing.py#L1154-L1163
to always append an ellipsis for basic and outer indexing:
```python
    def _indexing_array_and_key(self, key):
        if isinstance(key, OuterIndexer):
            array = self.array
>           key = _outer_to_numpy_indexer(key, self.array.shape) + (Ellipsis,)
        elif isinstance(key, VectorizedIndexer):
            array = nputils.NumpyVIndexAdapter(self.array)
            key = key.tuple
        elif isinstance(key, BasicIndexer):
            array = self.array
>           key = key.tuple + (Ellipsis,)
```
I'm not familiar enough with all the indexing variants in xarray to know if this covers all cases of 0d arrays that are currently copies but could be views. If someone wants to share some insight (e.g., some more advanced test cases), I could try and put together a pull request.

#### Expected Output
```python
>>> da[0].values.fill(99)
>>> da
<xarray.DataArray (dim_0: 3)>
array([99, 1, 2])
```

<details>
/home/daniel/local/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.15.0-42-lowlatency
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.11.0
pandas: 0.23.0
numpy: 1.14.3
scipy: 1.1.0
netCDF4: 1.4.0
h5netcdf: 0.6.2
h5py: 2.7.1
Nio: None
zarr: None
cftime: 1.0.0b1
PseudonetCDF: None
rasterio: None
iris: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.5
distributed: 1.21.8
matplotlib: 2.2.2
cartopy: None
seaborn: 0.8.1
setuptools: 39.1.0
pip: 10.0.1
conda: 4.5.12
pytest: 3.5.1
IPython: 6.4.0
sphinx: 1.7.4
</details>

danielwe:
- [x] Closes #2622
 - [x] Tests added

### DIFF
--- a/xarray/core/indexing.py
+++ b/xarray/core/indexing.py
@@ -1142,15 +1142,6 @@ class NumpyIndexingAdapter(ExplicitlyIndexedNDArrayMixin):
                             'Trying to wrap {}'.format(type(array)))
         self.array = array
 
-    def _ensure_ndarray(self, value):
-        # We always want the result of indexing to be a NumPy array. If it's
-        # not, then it really should be a 0d array. Doing the coercion here
-        # instead of inside variable.as_compatible_data makes it less error
-        # prone.
-        if not isinstance(value, np.ndarray):
-            value = utils.to_0d_array(value)
-        return value
-
     def _indexing_array_and_key(self, key):
         if isinstance(key, OuterIndexer):
             array = self.array
@@ -1160,7 +1151,10 @@ class NumpyIndexingAdapter(ExplicitlyIndexedNDArrayMixin):
             key = key.tuple
         elif isinstance(key, BasicIndexer):
             array = self.array
-            key = key.tuple
+            # We want 0d slices rather than scalars. This is achieved by
+            # appending an ellipsis (see
+            # https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#detailed-notes).  # noqa
+            key = key.tuple + (Ellipsis,)
         else:
             raise TypeError('unexpected key type: {}'.format(type(key)))
 
@@ -1171,7 +1165,7 @@ class NumpyIndexingAdapter(ExplicitlyIndexedNDArrayMixin):
 
     def __getitem__(self, key):
         array, key = self._indexing_array_and_key(key)
-        return self._ensure_ndarray(array[key])
+        return array[key]
 
     def __setitem__(self, key, value):
         array, key = self._indexing_array_and_key(key)
--- a/xarray/tests/test_variable.py
+++ b/xarray/tests/test_variable.py
@@ -1147,6 +1147,11 @@ class TestVariable(VariableSubclassobjects):
         assert v_new.dims == ('x', )
         assert_array_equal(v_new, v._data[:, 1])
 
+        # test that we obtain a modifiable view when taking a 0d slice
+        v_new = v[0, 0]
+        v_new[...] += 99
+        assert_array_equal(v_new, v._data[0, 0])
+
     def test_getitem_with_mask_2d_input(self):
         v = Variable(('x', 'y'), [[0, 1, 2], [3, 4, 5]])
         assert_identical(v._getitem_with_mask(([-1, 0], [1, -1])),
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -56,6 +56,9 @@ Enhancements
 - :py:meth:`DataArray.resample` and :py:meth:`Dataset.resample` now supports the
   ``loffset`` kwarg just like Pandas.
   By `Deepak Cherian <https://github.com/dcherian>`_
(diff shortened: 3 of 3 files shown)


xarray/tests/test_variable.py
```diff
@@ -139,6 +139,15 @@ def _assertIndexedLikeNDArray(self, variable, expected_value0,
...
         assert variable.identical(variable.copy())
+        # test that we obtain a view, not a copy, when taking a 0d slice
+        if (
+                isinstance(variable._data, np.ndarray)
+                and isinstance(variable[0]._data, np.ndarray)
+        ):
+            assert (
+                variable[0]._data.__array_interface__["data"][0]
+                == variable._data.__array_interface__["data"][0]
+            )
```
fujiisoup: Can you add more explicit tests that other developers easily find what we are testing?
e.g. 
```python

a = variable[0]
a = 3
assert a == variable[0]
```

xarray/core/indexing.py
```diff
@@ -1160,7 +1160,10 @@ def _indexing_array_and_key(self, key):
             key = key.tuple
         elif isinstance(key, BasicIndexer):
             array = self.array
-            key = key.tuple
+            # We're not interested in obtaining a scalar, but rather a 0d slice.
+            # This is accomplished by appending an ellipsis (see
+            # https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#detailed-notes).  # noqa
+            key = key.tuple + (Ellipsis,)
```
shoyer: Two questions:
1. does this also work in the other branches, i.e., the ones that do fancy indexing?
2. (related) can you remove the `_ensure_ndarray` method above?

xarray/tests/test_variable.py
shoyer: It might also be easier just to write a separate dedicated method for this, rather than squeezing it into the existing `_assertIndexedLikeNDArray`.

danielwe: You're right, my thinking about tests was a little muddled. The new solution should be better.

xarray/core/indexing.py
danielwe: 1. It does not affect outer indexing but breaks vectorized indexing, where some arrays come back with dimensions in a different order. I haven't looked into exactly why. It would be nice to move the change into `__getitem__`, as `return array[key + (Ellipsis,)]`, but I'm afraid that won't work. In any case, I think only basic indexing ever returns a 0d array, so the current change covers all relevant cases.

2. In fact, yes. I was convinced there were edge cases with object arrays etc., so I let it stay, but all tests pass without. Removed now.

shoyer:
wonderful, thanks @danielwe !

Issue #2627, opened 2018-12-22
Is pep8speaks working well?


max-sixty:
I returned to do some work on xarray, and looks like there are lots of linting errors. Maybe we shouldn't worry ourselves with this; though let's make a deliberate decision

Is pep8speaks working well? One example is this PR: https://github.com/pydata/xarray/pull/2553, where it looks like pep8speaks initially complained, but doesn't have a test so was potentially lost in the discussion.

Any thoughts? I'm happy to help see if there are alternatives / pep8speaks can be added as a test / add a test in travis / fix these. 

FWIW I've used `black` successfully in pandas-gbq and internally, but I think it's maybe too much change for a large project like xarray

```
./versioneer.py:943:-13592: W605 invalid escape sequence '\s'
./versioneer.py:943:-13408: W605 invalid escape sequence '\s'
./versioneer.py:943:-13228: W605 invalid escape sequence '\s'
./versioneer.py:943:-11196: W605 invalid escape sequence '\d'
./versioneer.py:943:-8162: W605 invalid escape sequence '\d'
./xarray/testing.py:148:1: W391 blank line at end of file
./xarray/core/options.py:66:80: E501 line too long (97 > 79 characters)
./xarray/core/combine.py:381:36: E226 missing whitespace around arithmetic operator
./xarray/core/combine.py:536:75: E211 whitespace before '('
./xarray/core/combine.py:537:57: E126 continuation line over-indented for hanging indent
./xarray/core/dataarray.py:19:1: F401 '.options._get_keep_attrs' imported but unused
./xarray/core/dataset.py:15:1: F401 '.computation' imported but unused
./xarray/core/dataset.py:18:1: F401 '..conventions' imported but unused
./xarray/core/dataset.py:27:1: F401 '.dtypes.is_datetime_like' imported but unused
./xarray/core/dataset.py:32:1: F401 '.pycompat.integer_types' imported but unused
./xarray/core/dataset.py:520:5: E303 too many blank lines (2)
./xarray/core/dataset.py:739:80: E501 line too long (81 > 79 characters)
./xarray/core/dataset.py:750:80: E501 line too long (80 > 79 characters)
./xarray/core/dataset.py:768:80: E501 line too long (88 > 79 characters)
./xarray/core/dataset.py:779:80: E501 line too long (84 > 79 characters)
./xarray/core/dataset.py:788:80: E501 line too long (84 > 79 characters)
[...]

PR #2629, opened 2018-12-24
Flake fixed

max-sixty:
Towards https://github.com/pydata/xarray/issues/2627

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2912,7 +2912,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
         else:
             dims = set(dim)
 
-        missing_dimensions = [dim for d in dims if d not in self.dims]
+        missing_dimensions = [d for d in dims if d not in self.dims]
         if missing_dimensions:
             raise ValueError('Dataset does not contain the dimensions: %s'
                              % missing_dimensions)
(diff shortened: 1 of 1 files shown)


xarray/backends/netcdf3.py
```diff
@@ -9,7 +9,7 @@
 
 # Special characters that are permitted in netCDF names except in the
 # 0th position of the string
-_specialchars = '_.@+- !"#$%&\()*,:;<=>?[]^`{|}~'
+_specialchars = '_.@+- !"#$%&\\()*,:;<=>?[]^`{|}~'
```
shoyer: Good catch!

xarray/core/dataset.py
```diff
@@ -2905,7 +2906,7 @@ def reduce(self, func, dim=None, keep_attrs=None, numeric_only=False,
         else:
             dims = set(dim)
 
-        missing_dimensions = [dim for dim in dims if dim not in self.dims]
+        missing_dimensions = [dim for d in dims if d not in self.dims]
```
shoyer: Here and below: This looks like a mistake -- `dim` is no longer the loop variable.

max-sixty: Thanks!

max-sixty:
Let's merge this to ensure we don't have merge errors, and then answer https://github.com/pydata/xarray/issues/2627#issuecomment-449684619 after?

PR #2632, opened 2018-12-25
Add flake check to travis

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/2627


fujiisoup:
Thanks, @max-sixty.
I personally like to have a separate CI to check pep8 violation, so that we can easily distinguish whether the tests or pep8 fail by the green check in the bottom.

Do you have any idea about the ci for pep8? Or we can simply use the general ci such as circle ci?

shoyer:
If someone else can figure out how to setup these checks to run with another CI service that would be fantastic, but this is definitely an improvement over nothing at all.

PR #2633, opened 2018-12-25
Fix dayofweek and dayofyear attributes from dates generated by cftime_range

spencerkclark:
- [x] Tests added

It turns out there was a remaining bug in cftime (https://github.com/Unidata/cftime/issues/106) that impacted the results of the `dayofwk` and `dayofyr` attributes of cftime objects generated by their `replace` method, which we use when parsing dates from strings, and in some offset arithmetic.

A workaround is to add a `dayofwk=-1` argument to each `replace` call where the `dayofwk` or `dayofyr` would be expected to change.  I've fixed this bug upstream in cftime (https://github.com/Unidata/cftime/pull/108), but it will only be available in a future version.  Would it be appropriate to use this workaround in xarray?  

This would fix [this doc page](http://xarray.pydata.org/en/latest/time-series.html#non-standard-calendars-and-dates-outside-the-timestamp-valid-range) for instance:

screen shot 2018-12-25 at 7 40 30 am


shoyer:
agreed, thanks!

PR #2639, opened 2019-01-01
ENH: switch Dataset and DataArray to use explicit indexes

shoyer:
xref https://github.com/pydata/xarray/issues/1603

This change switches Dataset.indexes and DataArray.indexes to be backed by
explicit dictionaries of indexes, instead of being implicitly defined by
the set of coordinates with names matching dimensions.

There are no changes to the public interface yet: these will come later.

My current plan:
1. (This PR) Indexes are recreated from coordinates every time a new DataArray
or Dataset is created.
2. (Follow-up PRs) Refactor indexes to be propagated explicitly in xarray operations. This will facilitate future API changes, when indexes will no longer only be associated with dimensions. I will probably add some testing decorator that can be used to mark part of a test as including no creation of default indexes.
3. Add explicit entries into `indexes` for MultiIndex levels that are checked instead of MultiIndex variables. Still no public API changes (aside from adding more entries to `.indexes`). 
4. Support arbitrary coordinates in `indexes`.


fujiisoup:
@shoyer 

It looks nice and reasonable to me.

> (This PR) Indexes are recreated from coordinates every time a new DataArray or Dataset is created.

Does it add some overhead (though I don't see any reason of it)?

xarray/core/indexes.py
```diff
@@ -0,0 +1,53 @@
...
+       Coordinate variables from which to draw default indexes.
+    dims : iterable
+        Iterable of dimension names.
+
+    Returns
+    -------
+    Mapping[Any, pandas.Index] mapping indexing keys (levels/dimension names)
+    to indexes used for indexing along that dimension.
+    """
+    return {key: coords[key].to_index() for key in dims if key in coords}
```
fujiisoup: Should we use `OrderedDict` instead of `dict` to preserve the order?
It is probably necessary for the backward compatibility.

shoyer: I guess this makes sense, for now. Eventually we are going to break this, though (when we add new entries).

shoyer:
> > (This PR) Indexes are recreated from coordinates every time a new DataArray or Dataset is created.
> 
> Does it add some overhead (though I don't see any reason of it)?

I suppose another option would be to leave `self._indexes = None` in the constructor, and only set default values for `self._indexes` when the `indexes` property is accessed.

I think we do something like for a few other attributes already.

PR #2645, opened 2019-01-03
Remove py2 compat

### LINKED ISSUE #1876: DEP: drop Python 2.7 support
The timeline for dropping Python 2.7 support for new Xarray releases is the end of 2018.

This issue can be used to track the necessary documentation and code changes to make that happen. 

xref: #1830

jhamman:
I was feeling particularly zealous today so I decided to see what it would take to strip out all the Python 2 compatibility code in xarray. I expect some will feel its too soon to merge this so I'm mostly putting this up for show-and-tell and to highlight some of the knots we've tied ourselves into over the years. 

 - [x] Closes #1876 
 - [ ] Tests added


xarray/core/formatting.py
```diff
@@ -47,22 +47,11 @@ def wrap_indent(text, start='', length=None):
...
-
-    On Python 2, this means we need to convert unicode to bytes. We won't need
-    this function once we drop Python 2.7 support.
-    """
-    if PY2 and isinstance(string, unicode_type):
-        string = string.encode('utf-8')
-    return string
-
-
 class ReprMixin(object):
```
shoyer: If you're feeling ambitious, you could drop this mixin class entirely, after replacing all `__unicode__` methods with `__repr__`.

jhamman:
Was I not ambition enough 😉? It's only January 3 after all. I was really trying not to over do it.

Anyhow, I'm happy to include the future and unicode removals here.

fmaussion:
> I expect some will feel its too soon to merge this

I never wrote py2 code myself so I don't think I'm of the "conservative" type, but I wonder if this PR will make it harder to provide bugfix cherry-picks on the hypothetical py2 compatible `0.11.x` branch? Or do we want to ditch this possibility right away and simply focus on the bright py3-only future?

jhamman:
@fmaussion - you've articulated why I expected some friction here. I'm happy to hold off on this until we're sure 0.11 is clear of nasty bugs. @shoyer just went through the back-porting process so maybe he can comment on his appetite for dealing with this sort of thing.

jhamman:
@shoyer - at this point, do you anticipate another bug-fix release for 0.11? If not, I'm now leaning toward merging this to avoid too many merge conflicts.

shoyer:
I think we should wait for a few weeks. There are two regressions that were recently fixed and that we may want to back-port: https://github.com/pydata/xarray/pull/2648 and https://github.com/pydata/xarray/pull/2651.

max-sixty:
What are our thoughts on merging this? I am sensitive that merge conflicts are growing

jhamman:
I think we should issue a back-port bug fix (0.11.3) and then merge this asap. The two issues @shoyer referenced above have been merged so, as far as I understand, there is nothing blocking the bugfix release at this point.

shoyer:
Let me try doing the backports now...

shoyer:
See https://github.com/pydata/xarray/pull/2704 for a PR with release notes for 0.11.3. But I agree, there should be nothing stopping this from being merged now.

jhamman:
I'll sort out the conflicts today and get this merged ASAP.

dcherian:
Also see #2574

jhamman:
I plan to merge this tomorrow unless I hear otherwise.

xarray/core/indexes.py
```diff
@@ -1,4 +1,3 @@
-from __future__ import absolute_import, division, print_function
 try:
     from collections.abc import Mapping
 except ImportError:
```
shoyer: we can remove this try/except now

shoyer:
I found a couple of minor issues, otherwise it looks good!

shoyer:
thanks @jhamman !

Issue #1988, opened 2018-03-14
open_mfdataset() on a single file drops the concat_dim


WeatherGod:
When calling `xr.open_mfdataset()` on a 1 element list of filenames, the concat dimension is never added.

This isn't a MWE at the moment (will make one soon enough), just wanted to get my thoughts down.

```python
from datetime import datetime
import xarray as xr

time_coord = xr.DataArray([datetime.utcnow()], name='time', dims='time')
radmax_ds = xr.open_mfdataset(['foobar.nc'], concat_dim=time_coord)
print(radmax_ds)
```
```
<xarray.Dataset>
Dimensions:    (latitude: 5650, longitude: 12050)
Coordinates:
  * latitude   (latitude) float32 13.505002 13.515002 13.525002 13.535002 ...
  * longitude  (longitude) float32 -170.495 -170.485 -170.475 -170.465 ...
Data variables:
    RadarMax   (latitude, longitude) float32 dask.array<shape=(5650, 12050), chunksize=(5650, 12050)>
Attributes:
    start_date:   03/07/2017 01:00
    end_date:     03/07/2017 01:55
    elapsed:      60
```
#### Problem description
If there are two files, then there is a `time` coordinate, and the data array becomes 3D.

#### Output of ``xr.show_versions()``
I am currently on a recent-ish master of xarray.

jhamman:
This does seem inconsistent to me. Not that the behavior of concat already works with a single object:

```Python
In [1]: import xarray as xr

In [2]: da = xr.DataArray([1, 2], dims='x', name='foo')

In [3]: xr.concat([da], dim='y')
Out[3]:
<xarray.DataArray 'foo' (y: 1, x: 2)>
array([[1, 2]])
Dimensions without coordinates: y, x
```

The offending line is here:

https://github.com/pydata/xarray/blob/9261601f89c0d3cfc54db16718c82399d95266bd/xarray/core/combine.py#L342-L344

Based on this, I'm surprised my little example with concat works the way it does. In either event, it would be great if someone could spend some time normalizing the behavior here.

shoyer:
Yes, this seems like a bug. `open_mfdataset()` should always concatenate if a `dim` argument is provided explicitly.

WeatherGod:
Could the fix be as simple as `if len(datasets) == 1 and dim is None:`?

shoyer:
@WeatherGod Possible! As usual, tests are the hard part :)

WeatherGod:
[...]

PR #2648, opened 2019-01-04
Change an `==` to an `is`. Fix tests so that this won't happen again.

### LINKED ISSUE #2647: getting a "truth value of an array" error when supplying my own `concat_dim`.
This bug was introduced sometime after v0.11.0 and has turned up in my test suite using v0.11.2. I'll pass a `DataArray()` as my `concat_dim`, and the failure will happen at line 609 in backends/api.py because of: `if concat_dim is None or concat_dim == _CONCAT_DIM_DEFAULT:`

I am not sure how this change got through. In #2048, I added a unit test that passes a DataArray as a concat_dim argument.

WeatherGod:
- [x] Closes #2647
 - [x] Tests added

Also re-affirms #1988.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -10,7 +10,8 @@ import numpy as np
 
 from .. import Dataset, backends, conventions
 from ..core import indexing
-from ..core.combine import _auto_combine, _infer_concat_order_from_positions
+from ..core.combine import (
+    _CONCAT_DIM_DEFAULT, _auto_combine, _infer_concat_order_from_positions)
 from ..core.pycompat import basestring, path_type
 from ..core.utils import close_on_error, is_grib_path, is_remote_uri
 from .common import ArrayWriter
@@ -481,9 +482,6 @@ class _MultiFileCloser(object):
     def close(self):
         for f in self.file_objs:
             f.close()
-
-
-_CONCAT_DIM_DEFAULT = '__infer_concat_dim__'
 
 
 def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -606,7 +606,7 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     # Coerce 1D input into ND to maintain backwards-compatible API until API
     # for N-D combine decided
     # (see https://github.com/pydata/xarray/pull/2553/#issuecomment-445892746)
-    if concat_dim is None or concat_dim == _CONCAT_DIM_DEFAULT:
+    if concat_dim is None or concat_dim is _CONCAT_DIM_DEFAULT:
         concat_dims = concat_dim
     elif not isinstance(concat_dim, list):
         concat_dims = [concat_dim]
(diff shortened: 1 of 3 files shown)


xarray/backends/api.py
```diff
@@ -606,7 +606,7 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     # Coerce 1D input into ND to maintain backwards-compatible API until API
     # for N-D combine decided
     # (see https://github.com/pydata/xarray/pull/2553/#issuecomment-445892746)
-    if concat_dim is None or concat_dim == _CONCAT_DIM_DEFAULT:
+    if concat_dim is None or concat_dim is _CONCAT_DIM_DEFAULT:
```
WeatherGod: Are we sure we want this to be an `is` test? Are we sure that the keyword argument will default to that string object?

shoyer: Yes, we use that as the default value here.

Possibly a better choice would be to use something like `ReprObject('<inferred>')`  for the sentinel object, e.g., as done here:
https://github.com/pydata/xarray/blob/69086b332c6c950587830b266df4e624c2106d89/xarray/core/dtypes.py#L8

WeatherGod: I guess what I am getting at is a matter of user-friendliness. A user looking up the call signature of this method will see the default as the string (and not an object), but, they'll never be able to explicitly force concat dimension inference, because passing the string themselves will not work (it'll be a different object).

shoyer:
Yes, it would be better to use the custom ReprObject.

WeatherGod:
ok, so we use the ReprObject for the default, and then test if `concat_dim` is of type `ReprObject and then test its equivalance?

shoyer:
> ok, so we use the ReprObject for the default, and then test if `concat_dim` is of type `ReprObject and then test its equivalance?

No, just check identity with the exact ReprObject used as the default value.

This is just a slightly more readable version of the common idiom of use `object()` as a default value, e.g., as shown here
http://effbot.org/zone/default-values.htm#what-to-do-instead

WeatherGod:
Is the following statement True or False: "The user should be allowed to explicitly declare that they want the concatenation dimension to be inferred by passing a keyword argument". If this is True, then you need to test equivalence. If it is False, then there is nothing more I need to do for the PR, as changing this to use a ReprObject is orthogonal to these changes.

shoyer:
> Is the following statement True or False: "The user should be allowed to explicitly declare that they want the concatenation dimension to be inferred by passing a keyword argument".

This statement is False, but it looks like we have a related bug: we really should be importing `_CONCAT_DIM_DEFAULT` from `xarray.core.combine`, because we use identity comparison in that module. Right now this only works by virtue of the accident that identical string literals points to the same variable in CPython.

WeatherGod:
I completely forgotten about that little quirk of cpython. I try to ignore implementation details like that. Heck, I still don't fully trust dictionaries to be ordered!

I removed the WIP. We can deal with the concat dim default object separately, including turning it into a ReprObject (not exactly sure what the advantage of it is over just using the string, but, meh).

shoyer:
I just pushed a commit to your branch that should fix the string identity issue (by not defining the constant multiple times).

shoyer:
Thanks

PR #2650, opened 2019-01-04
rolling_exp (nee ewm)

max-sixty:
- [x] Duplicate groupby / rolling interface
- [x] Integrate with apply_ufunc, including orientation and coords
- [x] Define com / span / alpha interface
- [x] Implement for Variable & Dataset
- [x] Docs


xarray/tests/test_dataarray.py
```diff
@@ -3709,14 +3709,14 @@ def test_to_and_from_iris(self):
             assert coord.var_name == original_coord.name
             assert_array_equal(
                 coord.points, CFDatetimeCoder().encode(original_coord).values)
-            assert (actual.coord_dims(coord) ==
-                    original.get_axis_num(
+            assert (actual.coord_dims(coord)
+                    == original.get_axis_num(
```
max-sixty: These are autoformat changes, lmk if any objections

xarray/core/ewm.py
```diff
@@ -0,0 +1,31 @@
...
+        self.span = span
+
+
+class DataArrayEWM(EWM):
+    def mean(self):
+        return apply_ufunc(
+            _ewm_nanmean,
+            self.obj,
+            input_core_dims=[[self.dim]],
+            output_core_dims=[[self.dim]],
```
max-sixty: Is this the correct approach for a function that:
- Needs to put a dimension in last place, given the gufunc interface
- Needs to retain that dimension (i.e. unlike `.sum` it shouldn't be reduced)

I looked for comparisons, but it looks like rolling functions are pre-`xarray_ufunc` era

Currently, the array
- loses the time coords
- doesn't transpose back

max-sixty: Or is this not a fit for `xarray_ufunc` and I should do it manually (which would be easy, but trying to use our tools...)

shoyer: In other cases we have achieved this by adding `.transpose(*self.dims)` after using `apply_ufunc()`. This would work at least for DataArray objects.

The other option would be to drop `input_core_dims`/`output_core_dims` entirely and pass an `axis` argument, e.g.,
```python
        axis = self.get_axis_num(self.dim)
        return apply_ufunc(
            _ewm_nanmean,
            self.obj,
            kwargs=dict(com=self.span, axis=axis)
        )
```

This works out fine because with a single argument there isn't a meaningful difference between "broadcast" and "core" dimensions -- there is no broadcasting that happens.

shoyer:
I know the name `ewm` matches pandas, but I find it rather inscrutable if you don't already know the acronym. What about something a little longer, maybe `exp_rolling()`?

xarray/core/ewm.py
shoyer: I guess the notion of a "core" dimension does matter if we want to support parallelization in dask arrays with `dask='parallelized'. But at long as you aren't explicitly setting that in `apply_ufunc()` it doesn't make a difference.

shoyer:
Before we add even an optional dependency on numbagg in xarray, we should probably do a bit of cleanup (e.g., making sure we're happy with its public interface, and putting a release up on pypi)

max-sixty:
> I know the name ewm matches pandas, but I find it rather inscrutable if you don't already know the acronym. 

👍 

> What about something a little longer, maybe exp_rolling()?

Yes, that works.
One alternative is to add a `window_type` / `weighting` argument to `Rolling` - Exponentially weighted is one of many alternatives

I'm fairly balanced between them - others' thoughts?

fujiisoup:
I like `window_type` / `weighting` argument to `Rolling`.
But a little concern is whether it matches to the current API, such as `__iter__`.

BTW, does `ewm` computes the window mean based on index or coordinate value?

shoyer:
I would lean towards a dedicated method, since there are method specific options. It's pretty awkward to reuse a single interface for that.

max-sixty:
> BTW, does ewm computes the window mean based on index or coordinate value?

Index, currently. 

Would be great to have an algo that dealt with coord value, and I think not too difficult

max-sixty:
Tests seem to be failing on a different issue? https://travis-ci.org/pydata/xarray/jobs/479042667#L7759

max-sixty:
I think this is in a reasonable state for DataArray, excluding docs. Let me know any feedback on the APi

Does anyone have a view on the canonical way to implement these for Dataset, given a subset of the variables will have the dim? 
Tests fail when naively using `apply_ufunc`; `.reduce` looks like it has some functionality for skipping those variables. Or I could do it manually in a couple of lines.

shoyer:
> Does anyone have a view on the canonical way to implement these for Dataset, given potentially only a subset of the variables will have the dimension?
> Tests fail when naively using `apply_ufunc`; `.reduce` looks like it has some functionality for skipping those variables. Or I could do it manually in a couple of lines.

You could probably copy the logic from `Dataset.reduce`, which simply applies different logic for coordinates (they either get preserved or dropped depending on if they reuse the reduced dimension). That said, if EWM preserves the dimension size/labels you probably don't nee any special logic for coordinates -- see `DatasetRolling.reduce` as well.

max-sixty:
> That said, if EWM preserves the dimension size/labels you probably don't nee any special logic for coordinates 

The error is when applying over a dimension on a dataset where only some of the variables have the dimension; e.g. applying over `time` on this:

```python
<xarray.Dataset>
Dimensions:  (time: 10, x: 8, y: 2)
Coordinates:
  * x        (x) float64 0.0 0.1429 0.2857 0.4286 0.5714 0.7143 0.8571 1.0
  * time     (time) float64 0.0 0.1111 0.2222 0.3333 ... 0.7778 0.8889 1.0
    c        (y) <U1 'a' 'b'
  * y        (y) int64 0 1
Data variables:
    z1       (y, x) float64 -0.1035 -0.8153 -1.583 ... 1.447 0.7768 -0.2699
    z2       (time, y) float64 0.968 0.7156 -1.64 ... 0.1889 -1.142 0.7172
```

...rather than any issues applying on `coords`. I think the solution is easy - only apply on variables where the dimension exists.

If it's helpful to add that functionality directly to `apply_ufunc`, lmk. I'll focus on getting this working, though

fujiisoup:
@max-sixty 

> The error is when applying over a dimension on a dataset where only some of the variables have the dimension; 

I remember I faced the same issue in implementing `differentiate`, `interp`, `trapz`, etc... and manually wrote the same logic for several times.

> If it's helpful to add that functionality directly to apply_ufunc, lmk.

I think it would make the code much cleaner at least for these methods.

max-sixty:
I made an attempt to add the "skip variables without the dimension" to `apply_ufunc`, but it's much harder than I expected - there are more cases than I expected (e.g. multiple datasets).

I may be missing something - let me know if there's an reasonable approach

Otherwise I'll do the close thing for this PR, and potentially we can have a look at the general solution later

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,36 @@
...
+        from numbagg.moving import rolling_exp_nanmean
+
+        da = apply_ufunc(
+            rolling_exp_nanmean,
+            self.obj,
+            input_core_dims=[[self.dim]],
+            output_core_dims=[[self.dim]],
+            kwargs=dict(window=self.alpha),
+        )
+        return da.transpose(*self.obj.dims)
```
shoyer: As I noted in https://github.com/pydata/xarray/pull/2674, I think we could write a generic `Dataset`/`DataArray`/`Variable` version of this with `reduce`:
```python
def rolling_exp_nanmean(array, *, axis, window):
    import numbagg
    if axis == ():
        return 1.0 * array  # cast to float
    else:
        return numbagg.rolling_exp_nanmean(array, axis=axis, window=window)

class RollingExp:
    ...
    def mean(self):
        return self.obj.reduce(rolling_exp_nanmean, dim=self.dim, window=self.alpha)
```

max-sixty: Amazing! Thank you

max-sixty: Should we promote this part to the function (in this case, to numbagg)?

```
def rolling_exp_nanmean(array, *, axis, window):
    import numbagg
    if axis == ():
        return 1.0 * array  # cast to float
    else:
        return numbagg.rolling_exp_nanmean(array, axis=axis, window=window)
```

xarray/core/common.py
```diff
@@ -243,6 +244,8 @@ def get_squeeze_dims(xarray_obj, dim, axis=None):
 class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
     """Shared base class for Dataset and DataArray."""
 
+    _rolling_exp_cls = RollingExp
```
max-sixty: I don't think we need this class attribute any longer, but leaving for the sake of conformity unless we have consensus

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,58 @@
...
+    # so use its function to get a com and then convert to alpha
+
+    com = _get_center_of_mass(com, span, halflife, alpha)
+    return 1 / (1 + com)
+
+
+def rolling_exp_nanmean(array, *, axis, window):
+    import numbagg
+    if axis == ():
+        return array.astype(np.float)
```
max-sixty: I changed this from @shoyer 's suggestion of `array * 1.0` given this should be a no-op for floats and is self-explaining, but let me know if that is better

max-sixty:
Any thoughts on the API?

@shoyer is making some updates to `numbagg` so as soon as those are complete, we could point the dependencies at that release and merge this

dcherian:
Can you add some docs to http://xarray.pydata.org/en/stable/computation.html#rolling-window-operations?

xarray/tests/test_dataset.py
```diff
@@ -4,7 +4,6 @@
 from collections import OrderedDict
 from copy import copy, deepcopy
 from io import StringIO
-import pickle
```
shoyer: I think we still need this -- there's a pickle test in this file that is failing on Travis-CI now.

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,70 @@
+
+import numpy as np
+from pandas.core.window import _get_center_of_mass
```
shoyer: We're going to want to vendor this private pandas function in xarray -- we don't want to be dependent on private APIs in pandas.

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,70 @@
...
+
+def _get_alpha(com=None, span=None, halflife=None, alpha=None):
+    # pandas defines in terms of com (converting to alpha in the algo)
+    # so use its function to get a com and then convert to alpha
+
+    com = _get_center_of_mass(com, span, halflife, alpha)
+    return 1 / (1 + com)
+
+
+def move_exp_nanmean(array, *, axis, window):
```
shoyer: Python 3... mind... blown!

xarray/core/rolling_exp.py
shoyer: I think `np.float64` would be preferrable here -- `np.float` is actually is an alias to the builtin `float` type :(.

I was originally thinking that `array * 1.0` did more minimal upcasting (i.e., sometimes to float32), but it does look like it's doing the same thing here.

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,70 @@
...
+    import numbagg
```
shoyer: Can we make sure this helper function raises if the input is a dask array?

xarray/core/rolling_exp.py
max-sixty: :) I think this was your code!

max-sixty:
This is updated! Could put an "Experimental" label on if we want (or maybe that's implicit).

Let me know any final changes. Will be good to get this merged at last.

max-sixty:
The gentlest of reminders that I think this is ready to merge (mea culpa for leaving it at 90% for so long)

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def move_exp_nanmean(array, *, axis, alpha):
+    if isinstance(array, dask_array_type):
+        raise TypeError("rolling_exp is not currently support for dask arrays")
+    import numbagg
+    if axis == ():
+        return array.astype(np.float64)
+    else:
+        return numbagg.moving.move_exp_nanmean(
```
shoyer: Let's use the top level function `numbagg.move_exp_nanmean`, that seems more likely to be stable than the submodule `numbagg.moving`.

xarray/core/common.py
```diff
@@ -613,6 +616,12 @@ def rolling(self, dim: Optional[Mapping[Hashable, int]] = None,
         return self._rolling_cls(self, dim, min_periods=min_periods,
                                  center=center)
 
+    def rolling_exp(self, dim=None, window_type='span', **dim_kwargs):
```
shoyer: Can you add a docstring here? Something short pointing to `RollingExp` for details would be fine, but we should have at least something since this is where users will look first.

max-sixty:
Great - updated! Let me know any final comments!

xarray/core/rolling_exp.py
```diff
@@ -0,0 +1,107 @@
...
+    window_type : str, one of ['span', 'com', 'halflife', 'alpha'], default 'span'
+        The format of the previously supplied window. Each is a simple
+        numerical transformation of the others. Described in detail:
+        https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.ewm.html
+
+    Returns
+    -------
+    RollingExp : type of input argument
+    """  # noqa
+    _attributes = ['alpha', 'dim']
```
shoyer: Is this used for anything?

max-sixty: No, I removed - it was on `Rolling` but actually used for something there...

max-sixty:
Updated! Let me know any final changes!

shoyer:
OK, in it goes. Thanks @max-sixty !

max-sixty:
Thanks for all the help!

PR #2653, opened 2019-01-05
Imprement integrate

### LINKED ISSUE #1288: Add trapz to DataArray for mathematical integration

fujiisoup:
- [x] Closes #1288
 - [x] Tests added

I would like to add `integrate`, which is essentially an xarray-version of `np.trapz`.
I know there was variety of discussions in #1288, but I think it would be nice to limit us within that numpy provides by `np.trapz`,
i.e.,
1. only for `trapz` not `rectangle` or `simps`
2. do not care `np.nan`
3. do not support `bounds`
Most of them (except for 1) can be solved by combining several existing methods.

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -3866,6 +3866,70 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords,
                 variables[k] = v
         return self._replace_vars_and_dims(variables)
 
+    def integrate(self, coord, datetime_unit=None):
+        """ integrate the array with the trapezoidal rule.
+
+        .. note::
+            This feature is limited to simple cartesian geometry, i.e. coord
+            must be one dimensional.
+
+        Parameters
+        ----------
+        coord: str
+            The coordinate to be used to compute the gradient.
+        datetime_unit
+            Can be specify the unit if datetime coordinate is specified.One of
+            {'Y', 'M', 'W', 'D', 'h', 'm', 's', 'ms', 'us', 'ns', 'ps', 'fs',
+             'as'}
+
+        Returns
+        -------
+        integrated: Dataset
+
+        See also
+        --------
+        DataArray.integrate
+        numpy.trapz: corresponding numpy function
+        """
+        from .variable import Variable
+
+        if coord not in self.variables and coord not in self.dims:
+            raise ValueError('Coordinate {} does not exist.'.format(coord))
+
+        coord_var = self[coord].variable
+        if coord_var.ndim != 1:
+            raise ValueError('Coordinate {} must be 1 dimensional but is {}'
+                             ' dimensional'.format(coord, coord_var.ndim))
+
+        dim = coord_var.dims[0]
+        if _contains_datetime_like_objects(coord_var):
+            if coord_var.dtype.kind in 'mM' and datetime_unit is None:
+                datetime_unit, _ = np.datetime_data(coord_var.dtype)
+            elif datetime_unit is None:
+                datetime_unit = 's'  # Default to seconds for cftime objects
+            coord_var = datetime_to_numeric(
+                coord_var, datetime_unit=datetime_unit)
+
+        variables = OrderedDict()
+        coord_names = []
+        for k, v in self.variables.items():
+            if k in self.coords:
+                if dim not in v.dims:
+                    variables[k] = v
+                    coord_names.append(k)
+            else:
+                if (k in self.data_vars and dim in v.dims):
+                    if _contains_datetime_like_objects(v):
+                        v = datetime_to_numeric(v, datetime_unit=datetime_unit)
+                    integ = duck_array_ops.trapz(
+                        v.data, coord_var, axis=v.get_axis_num(dim))
+                    v_dims = list(v.dims)
+                    v_dims.remove(dim)
+                    variables[k] = Variable(v_dims, integ)
+                else:
+                    variables[k] = v
+        return self._replace_vars_and_dims(variables, coord_names=coord_names)
+
     @property
     def real(self):
         return self._unary_op(lambda x: x.real,
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -151,6 +151,7 @@ Computation
    Dataset.diff
    Dataset.quantile
    Dataset.differentiate
+   Dataset.integrate
 
 **Aggregation**:
 :py:attr:`~Dataset.all`
@@ -319,6 +320,7 @@ Computation
    DataArray.dot
    DataArray.quantile
    DataArray.differentiate
+   DataArray.integrate
 
 **Aggregation**:
 :py:attr:`~DataArray.all`
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -199,6 +199,8 @@ You can also use ``construct`` to compute a weighted rolling sum:
   To avoid this, use ``skipna=False`` as the above example.
 
 
+.. _compute.using_coordinates:
+
 Computation using Coordinates
 =============================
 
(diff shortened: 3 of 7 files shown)


xarray/core/dataset.py
```diff
@@ -3866,6 +3866,70 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
...
+    def integrate(self, dim, datetime_unit=None):
+        """ integrate the array with the trapezoidal rule.
+
+        .. note::
+            This feature is limited to simple cartesian geometry, i.e. coord
+            must be one dimensional.
+
+        Parameters
+        ----------
+        dim: str
```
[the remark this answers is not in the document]
fujiisoup: Done :)

xarray/core/dataset.py
```diff
@@ -3867,6 +3867,79 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
                 variables[k] = v
         return self._replace_vars_and_dims(variables)
 
+    def integrate(self, dim, datetime_unit=None):
```
shoyer: Should we make the default `dim=None` do integration over all dimensions?

xarray/core/dataset.py
```diff
@@ -3867,6 +3867,79 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
...
+        result = self
+        for d in dim:
+            result = result._integrate_one(d, datetime_unit=datetime_unit)
+        return result
+
+    def _integrate_one(self, dim, datetime_unit=None):
+        from .variable import Variable
+
+        if dim not in self.variables and dim not in self.dims:
+            raise ValueError('Coordinate {} does not exist.'.format(dim))
```
shoyer: I think splitting these checks into two would be a little clearer:
- "cannot integrate over dimension {} because it does not exist" (for `dim not in self.dims`)
- "cannot integrate over dimension {} because there is no corresponding coordinate" (for `dim not in self.variables`)

xarray/core/dataset.py
```diff
@@ -3867,6 +3867,79 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
...
+
+        coord_var = self[dim].variable
+        if coord_var.ndim != 1:
+            raise ValueError('Coordinate {} must be 1 dimensional but is {}'
+                             ' dimensional'.format(dim, coord_var.ndim))
```
shoyer: Likewise, maybe:
- "cannot integrate over dimension {} because the corresponding coordinate is not a 1d array along that dimension: it has dimensions {}"

xarray/core/dataset.py
```diff
@@ -3867,6 +3867,79 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
...
+        return result
+
+    def _integrate_one(self, dim, datetime_unit=None):
+        from .variable import Variable
+
+        if dim not in self.variables and dim not in self.dims:
+            raise ValueError('Coordinate {} does not exist.'.format(dim))
+
+        coord_var = self[dim].variable
+        if coord_var.ndim != 1:
```
shoyer: I think this is currently not possible due to xarray's data model, but it's a good idea to add this anyways given that we want to change this soon (e.g., see https://github.com/pydata/xarray/pull/2405).

I would recommend adjusting this to `if coord_var.dims != (dim,)`, which is a little stricter.

xarray/core/dataset.py
```diff
@@ -3867,6 +3867,79 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
...
+                        v = datetime_to_numeric(v, datetime_unit=datetime_unit)
+                    integ = duck_array_ops.trapz(
+                        v.data, coord_var.data, axis=v.get_axis_num(dim))
+                    v_dims = list(v.dims)
+                    v_dims.remove(dim)
+                    variables[k] = Variable(v_dims, integ)
+                else:
+                    variables[k] = v
+        return self._replace_vars_and_dims(variables,
+                                           coord_names=set(coord_names))
```
shoyer: Maybe define this as a `set` instead, and use `coord_names.add(k)` instead of `append`?

xarray/core/dataset.py
fujiisoup: I first thought that it would be nice if we could integrate even along non-dimensional (1d) coordinate (as `interpolate_na`, `differential` do), but it also sounds something too much.
How do you think?

fujiisoup: Then, `coord` is a better argument rather than `dim`? 
Or we use `dim` for argument but support integration along non-dimensional coordinate with a slight avoidance of correctness, as it is more consistent with other reduction methods?

shoyer: I don't have a strong opinion here.

dcherian: Well, `differentiate` uses `coord`, so maybe `integrate` should too?

xarray/core/dataset.py
```diff
@@ -3867,7 +3867,7 @@ def differentiate(self, coord, edge_order=1, datetime_unit=None):
                 variables[k] = v
         return self._replace_vars_and_dims(variables)
 
-    def integrate(self, dim, datetime_unit=None):
+    def integrate(self, coord, datetime_unit=None):
```
shoyer: Should `coord=None` have the default behavior of integrating over all dimensions? Or would that be confusing in some way?

fujiisoup: I personally think it would be a little confusing because the result may change depending on which coordinate is used for `integrate`, e.g. if the `DataArray` has a dimension without coordinate but another one-dimensional coordinate, it is not very clear which should be used.

It would be a little convenient for 1d arrays, but aswe disallow default argument for `diff`, I like to disallow default argument here too.

dcherian:
Looks like this is ready to merge?

PR #2655, opened 2019-01-06
Type checking with mypy

shoyer:
The rest of the scientific Python stack doesn't seem to support type annotations yet, but that's OK -- we can use this incrementally in xarray when it seems appropriate, and may check a few bugs. I'm especially excited to use this for internal functions, where we don't always bother with full docstrings (e.g., what is the type of the ``variables`` argument?).

This includes:
1. various minor fixes to ensure that "mypy xarray" passes.
2. ~~adding "mypy xarray" to our lint check on Travis-CI.~~

For reference, see "Using mypy with an existing codebase":
https://mypy.readthedocs.io/en/stable/existing_code.html

Question: are we OK with (2)? This means Travis-CI will fail if your code causes mypy to error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ script:
       sphinx-build -n -j auto -b html -d _build/doctrees doc _build/html;
     elif [[ "$CONDA_ENV" == "lint" ]]; then
       pycodestyle xarray ;
+      mypy xarray ;
     elif [[ "$CONDA_ENV" == "py36-hypothesis" ]]; then
       pytest properties ;
     else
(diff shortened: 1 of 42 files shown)


dcherian:
Re (2): Perhaps it would be better to have mypy as an optional test? 

Can you add some docs to the "contributing to xarray" section on how to annotate or link to some documentation so that contributors know where to look.

shoyer:
I agree, let's experiment with keeping mypy optional for now. The error messages can definitely be a little confusing for non-experts. It should be pretty easy to fix regressions as they come up.

I've added a brief note on mypy to the contributing docs and removed the Travis-CI check.

shoyer:
I think this is good to merge now

Issue #2347, opened 2018-08-06
Serialization of just coordinates


hmaarrfk:
In the search for the perfect data storage mechanism, I find myself needing to store some of the images I am generating the metadata seperately. It is really useful for me to serialize just the coordinates of my DataArray.

My serialization method of choice is json since it allows me to read the metadata with just a text editor. For that, having the coordinates as a self contained dictionary is really important.

Currently, I convert just the coordinates to a [dataset](http://xarray.pydata.org/en/stable/data-structures.html#coordinates-methods), and serialize that. The code looks something like this:

```python
import xarray as xr
import numpy as np

# Setup an array with coordinates
n = np.zeros(3)
coords={'x': np.arange(3)}
m = xr.DataArray(n, dims=['x'], coords=coords)

coords_dataset_dict = m.coords.to_dataset().to_dict()
coords_dict = coords_dataset_dict['coords']

# Read/Write dictionary to JSON file

# This works, but I'm essentially creating an emtpy dataset for it
coords_set = xr.Dataset.from_dict(coords_dataset_dict)
coords2 = coords_set.coords  # so many `coords` :D
m2 = xr.DataArray(np.zeros(shape=m.shape), dims=m.dims, coords=coords2)
```

Would encapsulating this functionality in the `Coordinates` class be accepted as a PR?

It would add 2 functions that would look like:
```python
def to_dict(self):
    # offload the heavy lifting to the Dataset class
    return self.to_dataset().to_dict()['coords']

def from_dict(self, d):
    # Offload the heavy lifting again to the Dataset class
    d_dataset = {'dims': [], 'attrs': [], 'coords': d}
    return Dataset.from_dict(d_dataset).coords
```

rabernat:
@hmaarrfk - sorry no one replied to your issue. I would personally be fine with adding this to the Coordinates API.

hmaarrfk:
no need to be sorry. These two functions were easy enough for me to do myself in my own codebase. 

There are few issues that I've found doing this though.
Mainly, I can't find a good way to serialize numpy arrays in a round-trippable fashion.
It is difficult to get back `lists of arrays`, or `arrays of unit8`. I don't know if you have a good way to solvle this problem.

rabernat:
> I don't know if you have a good way to solvle this problem.

[...]

PR #2659, opened 2019-01-07
to_dict without data

### LINKED ISSUE #2656: dataset info in .json format
I am exploring the world of [Spatio Temporal Asset Catalogs](https://github.com/radiantearth/stac-spec) (STAC), in which all datasets are described using json/ geojson:

> The STAC specification aims to standardize the way geospatial assets are exposed online and
queried.

I am thinking about how to put the sort of datasets that xarray deals with into STAC items (see https://github.com/radiantearth/stac-spec). This would be particular valuable in the context of Pangeo and the zarr-based datasets we have been putting in cloud storage.

For this purpose, it would be very useful to have a concise summary of an xarray dataset's contents (minus the actual data) in .json format. I'm talking about the kind of info we currently get from the `.info()` method, which is designed to mirror the CDL output of [`ncdump -h`](https://www.unidata.ucar.edu/software/netcdf/netcdf-4/newdocs/netcdf/ncdump.html).

For example
```python
ds = xr.Dataset({'foo': ('x', np.ones(10, 'f8'), {'units': 'm s-1'})},
                 {'x': ('x', np.arange(10), {'units': 'm'})},
                 {'conventions': 'made up'})
ds.info()
```
```
xarray.Dataset {
dimensions:
	x = 10 ;

variables:
	float64 foo(x) ;
		foo:units = m s-1 ;
	int64 x(x) ;
		x:units = m ;

// global attributes:
	:conventions = made up ;
```

I would like to be able to do `ds.info(format='json')` and see something like this
```
{
 "coords": {
  "x": {
   "dims": [
    "x"
   ],
   "attrs": {
    "units": "m"
   }
  }
 },
 "attrs": {
  "conventions": "made up"
 },
 "dims": {
  "x": 10
 },
 "data_vars": {
  "foo": {
   "dims": [
    "x"
   ],
   "attrs": {
    "units": "m s-1"
   }
  }
 }
}
```

Which is what I get by doing `print(json.dumps(ds.to_dict(), indent=2))` and manually stripping out all the `data` fields. So an alternative api might be something like `ds.to_dict(data=False)`.

If anyone is aware of an existing spec for expressing [Common Data Language](https://www.unidata.ucar.edu/software/netcdf/workshops/2011/utilities/CDL.html) in json, we should probably use that instead of inventing our own. But I think some version of this would be a very useful addition to xarray.

rabernat:
This PR provides the ability to export Datasets and DataArrays to dictionary _without_ the actual data. This could be useful for generating indices of dataset contents to expose to search indices or other automated data discovery tools

In the process of doing this, I refactored the core dictionary export function to live in the Variable class, since the same code was duplicated in several places.

 - [x] Closes #2656
 - [x] Tests added

### DIFF
--- a/xarray/tests/test_dataarray.py
+++ b/xarray/tests/test_dataarray.py
@@ -2913,6 +2913,8 @@ class TestDataArray(object):
         expected_no_data = expected.copy()
         del expected_no_data['data']
         del expected_no_data['coords']['x']['data']
+        expected_no_data['coords']['x'].update({'dtype': '<U1', 'shape': (2,)})
+        expected_no_data.update({'dtype': 'float64', 'shape': (2, 3)})
         actual_no_data = array.to_dict(data=False)
         assert expected_no_data == actual_no_data
 
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -3047,10 +3047,15 @@ class TestDataset(object):
 
         # check the data=False option
         expected_no_data = expected.copy()
-        print(expected_no_data)
         del expected_no_data['coords']['t']['data']
         del expected_no_data['data_vars']['a']['data']
         del expected_no_data['data_vars']['b']['data']
+        expected_no_data['coords']['t'].update({'dtype': '<U1',
+                                                'shape': (10,)})
+        expected_no_data['data_vars']['a'].update({'dtype': 'float64',
+                                                   'shape': (10,)})
+        expected_no_data['data_vars']['b'].update({'dtype': 'float64',
+                                                   'shape': (10,)})
         actual_no_data = ds.to_dict(data=False)
         assert expected_no_data == actual_no_data
 
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -415,6 +415,8 @@ class Variable(common.AbstractArray, arithmetic.SupportsArithmetic,
                 'attrs': decode_numpy_dict_values(self.attrs)}
         if data:
             item['data'] = ensure_us_time_resolution(self.values).tolist()
+        else:
+            item.update({'dtype': str(self.dtype), 'shape': self.shape})
         return item
 
     @property


xarray/core/dataarray.py
```diff
@@ -1768,22 +1768,20 @@ def to_dict(self):
...
         for k in self.coords:
-            data = ensure_us_time_resolution(self[k].values).tolist()
-            d['coords'].update({
-                k: {'data': data,
-                    'dims': self[k].dims,
-                    'attrs': decode_numpy_dict_values(self[k].attrs)}})
-
-        d.update({'data': ensure_us_time_resolution(self.values).tolist(),
-                  'name': self.name})
+            d['coords'].update({k: self.coords[k].variable.to_dict(data=data)})
```
shoyer: This is only adding one item, so just using indexing for assignment seems cleaner: `d['coords'][k] = self.coords[k].variable.to_dict(data=data)`

xarray/tests/test_dataarray.py
```diff
@@ -2909,6 +2909,13 @@ def test_to_and_from_dict(self):
                 ValueError, "cannot convert dict without the key 'data'"):
             DataArray.from_dict(d)
 
+        # check the data=False option
+        expected_no_data = {**expected}
```
shoyer: I guess this is a fancy Py3 way of doing a copy? :)

I would lean towards the more explicit `expected.copy()` given that you aren't inserting any extra fields here.

rabernat:
It just occurred to me that it would be nice to have some extra info about the data, such as dtype.

xarray/tests/test_dataset.py
```diff
@@ -3045,11 +3045,20 @@ def test_to_and_from_dict(self):
         # check roundtrip
         assert_identical(ds, Dataset.from_dict(actual))
 
+        # check the data=False option
+        expected_no_data = expected.copy()
+        print(expected_no_data)
```
fmaussion: There is a forgotten `print` here!

fmaussion:
> It just occurred to me that it would be nice to have some extra info about the data, such as dtype.

II was about to ask about  it ;-). What other attribute than `dtype` were you thinking about?

shoyer:
> What other attribute than `dtype` were you thinking about?

Shape might also be a good idea.

rabernat:
Here is what the `.to_dict(data=False)` output looks like from the test example

```python
x = np.random.randn(10)
y = np.random.randn(10)
t = list('abcdefghij')
ds = Dataset(OrderedDict([('a', ('t', x)),
                          ('b', ('t', y)), ('t', ('t', t))]))
ds.to_dict(data=False)
```

```
{'attrs': {},
 'coords': {'t': {'attrs': {},
   'dims': ('t',),
   'dtype': '<U1',
   'shape': (10,)}},
 'data_vars': {'a': {'attrs': {},
   'dims': ('t',),
   'dtype': 'float64',
   'shape': (10,)},
  'b': {'attrs': {}, 'dims': ('t',), 'dtype': 'float64', 'shape': (10,)}},
 'dims': {'t': 10}}
```

The one thing I don't like about this is the empty attributes. Maybe we could change it so that `attrs` is only included if it is non-empty. Alternatively, we coiuld add a `squeeze_attrs` option.

dcherian:
Does this help with #2347?

rabernat:
> Maybe we could change it so that `attrs` is only included if it is non-empty.

How do people feel about this? If folks are fine with it as is, then LGTM.

rabernat:
Let's get this merged?

xarray/core/dataarray.py
```diff
@@ -1768,22 +1768,20 @@ def to_dict(self):
         Useful for coverting to json. To avoid datetime incompatibility
         use decode_times=False kwarg in xarrray.open_dataset.
 
+        Parameters
+        ----------
+        data : bool, optional
+            Whether to include the actual data in the dictionary. When set to
+            False, returns just the schema.
```
jhamman: It could be useful to allow including the data for the coordinates but not the data variables? 

I'm thinking something like 
```python
ds.to_dict(data='coords')
```

shoyer:
I think we should merge this -- @rabernat feel free to go ahead and do that.

We can leave `data='coords'` for a follow-up, if anyone has real use-cases for it.

rabernat:
Too late, I found this: 

https://binary-array-ld.github.io/netcdf-ld/

rabernat:
And this 
http://cf-json.org/specification

rabernat:
p.s. I learned about this from @kwilcox in https://github.com/pangeo-data/pangeo-datastore/issues/3. He might be a good person to loop into this discussion.

kwilcox:
If you are interested I could implement an `xarray -> cf-json -> xarray` round-trip. It will be a bit different from `to_dict` in terms of carrying through variable and attribute types so the file can be re-produced from the JSON. They both have their use-cases!

PR #2661, opened 2019-01-08
Remove broken Travis-CI builds

shoyer:
Remove the optional condaforge-rc, netcdf4-dev and pynio-dev builds. These
have been continuously failing (due to broken installs), so we shouldn't waste
time/energy running them.


dcherian:
Since we depend so much on netcdf4, does it make sense to spend the effort to get that build working?

shoyer:
> Since we depend so much on netcdf4, does it make sense to spend the effort to get that build working?

That would probably be a good idea.

PR #2668, opened 2019-01-11
fix datetime_to_numeric and Variable._to_numeric

### LINKED ISSUE #2667: datetime interpolation doesn't work

This code doesn't work anymore on master.

```python
a = xr.DataArray(np.arange(21).reshape(3, 7), 
                 dims=['x', 'time'], 
                 coords={'x': [1, 2, 3], 'time': pd.date_range('01-01-2001', periods=7, freq='D')})
xi = xr.DataArray(np.linspace(1, 3, 50),  
                  dims=['time'], 
                  coords={'time': pd.date_range('01-01-2001', periods=50, freq='H')})
a.interp(x=xi, time=xi.time)
```

#### Problem description

The above code now raises the error

```
<ipython-input-26-dda3a6d5725b> in <module>
      6                   dims=['time'],
      7                   coords={'time': pd.date_range('01-01-2001', periods=50, freq='H')})
----> 8 a.interp(x=xi, time=xi.time)

~/work/python/xarray/xarray/core/dataarray.py in interp(self, coords, method, assume_sorted, kwargs, **coords_kwargs)
   1032         ds = self._to_temp_dataset().interp(
   1033             coords, method=method, kwargs=kwargs, assume_sorted=assume_sorted,
-> 1034             **coords_kwargs)
   1035         return self._from_temp_dataset(ds)
   1036 

~/work/python/xarray/xarray/core/dataset.py in interp(self, coords, method, assume_sorted, kwargs, **coords_kwargs)
   2008                         in indexers.items() if k in var.dims}
   2009                     variables[name] = missing.interp(
-> 2010                         var, var_indexers, method, **kwargs)
   2011                 elif all(d not in indexers for d in var.dims):
   2012                     # keep unrelated object array

~/work/python/xarray/xarray/core/missing.py in interp(var, indexes_coords, method, **kwargs)
    468     new_dims = broadcast_dims + list(destination[0].dims)
    469     interped = interp_func(var.transpose(*original_dims).data,
--> 470                            x, destination, method, kwargs)
    471 
    472     result = Variable(new_dims, interped, attrs=var.attrs)

~/work/python/xarray/xarray/core/missing.py in interp_func(var, x, new_x, method, kwargs)
    535                              new_axis=new_axis, drop_axis=drop_axis)
    536 
--> 537     return _interpnd(var, x, new_x, func, kwargs)
    538 
    539 

~/work/python/xarray/xarray/core/missing.py in _interpnd(var, x, new_x, func, kwargs)
    558     var = var.transpose(range(-len(x), var.ndim - len(x)))
    559     # stack new_x to 1 vector, with reshape
--> 560     xi = np.stack([x1.values.ravel() for x1 in new_x], axis=-1)
    561     rslt = func(x, var, xi, **kwargs)
    562     # move back the interpolation axes to the last position

~/work/python/xarray/xarray/core/missing.py in <listcomp>(.0)
    558     var = var.transpose(range(-len(x), var.ndim - len(x)))
    559     # stack new_x to 1 vector, with reshape
--> 560     xi = np.stack([x1.values.ravel() for x1 in new_x], axis=-1)
    561     rslt = func(x, var, xi, **kwargs)
    562     # move back the interpolation axes to the last position

AttributeError: 'numpy.ndarray' object has no attribute 'values'
```

I think the issue is this line which returns a numpy array instead of a Variable. This was added in the `coarsen` PR (cc @fujiisoup)
https://github.com/pydata/xarray/blob/d4c46829b283ab7e7b7db8b86dae77861ce68f3c/xarray/core/utils.py#L636

fujiisoup:
- [x] Closes #2667
 - [x] Tests added

Started to fixing #2667


xarray/core/duck_array_ops.py
```diff
@@ -279,14 +279,49 @@ def f(values, axis=None, skipna=None, **kwargs):
...
+    if datetime_unit:
+        array = array / np.timedelta64(1, datetime_unit)
+    # convert np.NaT to np.nan
+    if hasattr(array, 'dtype'):
+        if array.dtype.kind in 'mM':
+            return np.where(isnull(array), np.nan, array.astype(dtype))
+        else:
+            print(type(array), array.dtype)
+            print(array)
+            return array.astype(dtype)
```
fujiisoup: Some tests pass a `np.ndarray` with object dtype, an entry of which is `datetime.datetime`, such as
```python
<class 'numpy.ndarray'> object
[datetime.timedelta(0) datetime.timedelta(days=1)
 datetime.timedelta(days=2)]
```
Is it an expected behavior?

shoyer: I don't think `astype()` does any conversion for object arrays.

spencerkclark: @fujiisoup this function might be helpful for casting the object arrays: https://github.com/pydata/xarray/blob/d4c46829b283ab7e7b7db8b86dae77861ce68f3c/xarray/core/variable.py#L132-L136

shoyer:
@spencerkclark what convention do you use for missing values in cftime arrays?

spencerkclark:
> what convention do you use for missing values in cftime arrays?

Currently there is nothing built in to xarray to support missing values for cftime arrays.  I think we would have to come up with a convention.

shoyer:
OK, in that case let's only worry about NaT for datetime64/timedelta64
arrays.

shoyer:
(at least for now, until we can figure out how to handle missing values)

xarray/core/duck_array_ops.py
```diff
@@ -279,14 +279,49 @@ def f(values, axis=None, skipna=None, **kwargs):
...
+        'us', 'ns', 'ps', 'fs', 'as'}
+    dtype: target dtype
+
+    Returns
+    -------
+    array
+    """
+    if offset is None:
+        offset = array.min()
+    array = array - offset
```
spencerkclark: > @fujiisoup this function might be helpful for casting the object arrays: https://github.com/pydata/xarray/blob/d4c46829b283ab7e7b7db8b86dae77861ce68f3c/xarray/core/variable.py#L132-L136

Sorry this comment was meant to refer to (if needed) casting this array from an object array to an array of type timedelta64.

fujiisoup:
I couldn't make `datetime_to_numeric` compatible both `np.ndarray` and `Variable`. I think it would be cleaner if `Variable` has a (private) method to do this, like `Variable._to_numeric(offset=None, datetime_unit=None)`.

doc/whats-new.rst
```diff
@@ -40,6 +40,8 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
+- Bug fix for interpolation with an datetime array. (:issue:`2668`)
```
spencerkclark: I think @dcherian caught this bug before it made it into a release, so maybe we do not need a bug-fix note?

xarray/tests/test_duck_array_ops.py
```diff
@@ -571,3 +572,46 @@ def test_docs():
             indicated dimension(s) removed.
         """)
     assert actual == expected
+
+
+def test_datetime_to_numeric_datetime64():
```
spencerkclark: It might make sense to modify these tests now that `duck_array_ops.datetime_to_numeric` is only meant to take NumPy or dask arrays as input (these tests are using DataArrays).

xarray/tests/test_interp.py
```diff
@@ -573,3 +573,16 @@ def test_cftime_to_non_cftime_error():
 
     with pytest.raises(TypeError):
         da.interp(time=0.5)
+
+
+@requires_cftime
+@requires_scipy
+def test_cftime_interp():
```
spencerkclark: Do you think you can find a better name for this test?  It does not seem to use cftime.

xarray/core/dataset.py
```diff
@@ -24,6 +24,7 @@
 from .coordinates import (
     DatasetCoordinates, LevelCoordinatesSource,
     assert_coordinate_consistent, remap_label_indexers)
+from .duck_array_ops import datetime_to_numeric
```
spencerkclark: Within `differentiate`, we use `datetime_to_numeric` on Variable objects; I think you should switch to the new Variable-specific method there.

xarray/core/duck_array_ops.py
```diff
@@ -279,14 +279,51 @@ def f(values, axis=None, skipna=None, **kwargs):
...
+
+    if datetime_unit:
+        array = array / np.timedelta64(1, datetime_unit)
```
spencerkclark: I think we need to move this to after we potentially convert an array of `datetime.timedelta` objects to `np.timedelta64` (otherwise this will raise an error).

fujiisoup:
Thanks, @spencerkclark 

It looks I need more work for this (probably a week). 
It is not working for cftime stuffs, but I do not yet understand how it is handled in xarray.

spencerkclark:
> It looks I need more work for this (probably a week).
> It is not working for cftime stuffs, but I do not yet understand how it is handled in xarray.

Thanks @fujiisoup, no worries.  There was only one place where I saw a potential issue for cftime dates, which I noted above.  I made a PR to your branch with the fix (https://github.com/fujiisoup/xarray/pull/9); let me know if there were other places with issues and I can sort those out too.

xarray/core/duck_array_ops.py
spencerkclark: To explain this a little more -- for cftime dates, `array - offset` (a few lines above) will return an array of `datetime.timedelta` objects.  When `datetime_to_numeric` was used with `xarray.Variable` objects, we did not have to worry about converting to `np.timedelta64`, because that would happen automatically.  In the case of pure NumPy or dask arrays, we do.

fujiisoup:
Thanks, @spencerkclark, for the PR.
It was super helpful.

I will finish it up tonight (hopefully).

fujiisoup:
@spencerkclark 

Sorry for leaving it for a long time.
I was a little stressed these days...

> Should we strive to have compatibility in `datetime_to_numeric` there? 

As I'm not a heavy user of `datetime` arrays, I personally do not mind if we do not have dask compatibility for `datetime_to_numeric`.
Probably it would be OK to leave it for the later development if someone wants this feature.

spencerkclark:
> As I'm not a heavy user of `datetime` arrays, I personally do not mind if we do not have dask compatibility for `datetime_to_numeric`.

No worries @fujiisoup, this is totally fine with me.  I think the most common use case for datetime arrays is in indexes, which currently need to be stored in memory anyway, so I don't think the dask use case would come up very often.

fujiisoup:
Thanks @spencerkclark for reviewing. Merged.

spencerkclark:
Thanks for wrapping this up @fujiisoup!

PR #2669, opened 2019-01-13
xfail cftimeindex multiindex test

shoyer:
It was a nice idea to support CFTimeIndex in a pandas.MultiIndex, but pandas
seems to have inadvertently broken this, see
https://github.com/pandas-dev/pandas/issues/24263

(This should fix our failing CI tests with pandas 0.24 rc)


spencerkclark:
Thanks @shoyer for trying to add this support to begin with, and reporting this latest issue upstream.

Issue #1739, opened 2017-11-23
Utility to restore original dimension order after apply_ufunc


shoyer:
This seems to be coming up quite a bit for wrapping functions that apply an operation along an axis, e.g., for `interpolate` in #1640 or `rank` in #1733.

We should either write a utility function to do this or consider adding an option to `apply_ufunc`.

shoyer:
I wonder if it would make sense to simply change the behavior of `apply_ufunc` so existing dimensions always get restored to their original location, and only new dimensions remain at the end. It seems like this is almost always desirable.

jhamman:
At this point, I think I'm +1 on either automatically restoring dimension order or providing some option to do so (with restoring as the default). I'm still getting used to `apply_ufunc` so I'm not sure I feel strongly enough to make this call yet though.

shoyer:
I started trying to implement this in `apply_ufunc` and then realized why I hadn't done so before: it's really not clear how to match up input/output arguments if there are more than one of each. We could do something special for there first input/output but that could be more confusing than helpful.

This is probably better reserved for the simpler `apply_raw`(https://github.com/pydata/xarray/issues/1618).

jhamman:
What if we just supplied an `output_dims` kwarg to `apply_ufunc` or `apply_raw`?

shoyer:
> What if we just supplied an output_dims kwarg to apply_ufunc or apply_raw?

We could do that, but how to do we handle specifying output dimensions for a Dataset function? Maybe use a dict?

This is part of why a helper function starts to make sense to me.

max-sixty:
What do we think about attempting to enforce a dimension order throughout the dataset? That would solve these issues for free. 

We already have _some_ order in `.dims` & `.sizes`. Could we transpose all dimensions to that ordering after any operation?

(Maybe this doesn't need to be _strictly_ enforced, but we'd at least set the expectation that dimensions could be reordered to that dimension order at any time)

Or are there uses to having different dimension order throughout a dataset?

```python
In [45]: ds
Out[45]:
<xarray.Dataset>
Dimensions:  (lat: 25, lon: 53, time: 2920)
Coordinates:
  * lat      (lat) float32 75.0 72.5 70.0 67.5 65.0 ... 25.0 22.5 20.0 17.5 15.0
  * lon      (lon) float32 200.0 202.5 205.0 207.5 ... 322.5 325.0 327.5 330.0
[...]

Issue #1251, opened 2017-02-05
Consistent naming for xarray's methods that apply functions


shoyer:
We currently have two types of methods that take a function to apply to xarray objects:
- `pipe` (on `DataArray` and `Dataset`): apply a function to this entire object (`array.pipe(func)` -> `func(array)`)
- `apply` (on `Dataset` and `GroupBy`): apply a function to each labeled object in this object (e.g., `ds.apply(func)` -> `ds({k: func(v) for k, v in ds.data_vars.items()})`).

And one more method that we want to add but isn't finalized yet -- currently named `apply_ufunc`:
- Apply a function that acts on unlabeled (i.e., numpy) arrays to each array in the object

I'd like to have three distinct names that makes it clear what these methods do and how they are different.  This has come up a few times recently, e.g., https://github.com/pydata/xarray/issues/1130

One proposal: rename `apply` to `map`, and then use `apply` only for methods that act on unlabeled arrays. This would require a deprecation cycle, but eventually it would let us add `.apply` methods for handling raw arrays to both Dataset and DataArray. (We could use a separate apply method from `apply_ufunc` to convert `dim` arguments to `axis` and not do automatic broadcasting.)

MaximilianR:
Sounds good. It breaks the consistency with pandas' `apply`, but `map` is much more logical

shoyer:
Another option is to keep `apply` as-is for Dataset and GroupBy objects, but add a separate `apply_raw` method for applying functions that act on "raw" arrays. This would be a little more similar to pandas' `apply` with `raw=True`.

We could even do the `raw=True` keyword argument like pandas, but this is a little awkward because there are some additional arguments on `apply_raw` that don't make sense on `apply` (e.g., arguments that specify that some dimensions should be dropped or added).

benbovy:
I would be +1 for `apply` and `apply_raw`.

max-sixty:
> One proposal: rename apply to map

Would we accept this? I'd be up for doing the PR to deprecate `apply` and introduce `map`. It makes a Dataset more consistent with a standard mapping interface. But it would be inconsistent with pandas and a rename of fairly widely used method

dcherian:
> One proposal: rename apply to map

-1 for pandas incompatibility.

I would like to rename `rolling.reduce` to `rolling.apply` to be consistent with pandas & groupby

max-sixty:
> I would like to rename rolling.reduce to rolling.apply to be consistent with pandas & groupby

[...]

PR #2674, opened 2019-01-15
Skipping variables in datasets that don't have the core dim

max-sixty:
ref https://github.com/pydata/xarray/pull/2650#issuecomment-454164295

This seems an ugly way of accomplishing the goal; any ideas for a better way of doing this? 

And stepping back, do others think a) it's helpful to skip variables in a dataset, and b) `apply_ufunc` should do this?

### DIFF
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -314,6 +314,11 @@ def apply_dict_of_variables_ufunc(func, *args, **kwargs):
 
     result_vars = OrderedDict()
     for name, variable_args in zip(names, grouped_by_name):
+        if len(variable_args) == 1 and signature.all_input_core_dims:
+            variable_arg, = variable_args
+            if not set(variable_arg.dims) & signature.all_core_dims:
+                result_vars[name] = variable_arg
+                continue
         result_vars[name] = func(*variable_args)
 
     if signature.num_outputs > 1:
--- a/xarray/tests/test_computation.py
+++ b/xarray/tests/test_computation.py
@@ -3,6 +3,7 @@ import operator
 import pickle
 from collections import OrderedDict
 from distutils.version import LooseVersion
+from functools import partial
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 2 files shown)


max-sixty:
Thanks a lot for the comments @shoyer , that's v clarifying 

Is there a conceptual overlap between the goals of `.reduce` and `apply_ufunc`? I had initially thought that `.reduce` strictly reduced a dimension, though that's not actually the case given `cumsum`

shoyer:
> Is there a conceptual overlap between the goals of `.reduce` and `apply_ufunc`? I had initially thought that `.reduce` strictly reduced a dimension, though that's not actually the case given `cumsum`

Yes, I think so. `reduce()` is a simpler but complementary model to `apply_ufunc`.

Maybe we should rename `reduce()` to something like `apply_axis_func()` or `apply_over_dim`? I would use the name `apply_along_axis` but the numpy function of the same name does something different.

shoyer:
It might also be clearer to split this functionality into two methods rather than the single `reduce()` method, e.g., `reduce()` and `transform()` in the model of pandas's groupby methods.

It's probably worth thinking about these APIs more systematically, see https://github.com/pydata/xarray/issues/1618, https://github.com/pydata/xarray/issues/1251, https://github.com/pydata/xarray/issues/1130

max-sixty:
> It's probably worth thinking about these APIs more systematically, see #1618, #1251, #1130

Nice, thanks for finding those

> It might also be clearer to split this functionality into two methods rather than the single reduce() method, e.g., reduce() and transform() in the model of pandas's groupby methods.

Is the split needed? I had thought that the function either
- returned a reduced array, and the output dims should be reduced
- returned an array with the original dims, and then output dims shouldn't be reduced

max-sixty:
Is there a next step for a function which:
- Is a drop-in replacement for `.reduce`
- Similar to https://github.com/pydata/xarray/issues/1618, though maybe even simpler, i.e. only operates on a single input at first
- Wraps `apply_ufunc`
- Would restore dim order https://github.com/pydata/xarray/issues/1739
- Could be used by @fujiisoup 's recent functions, ref https://github.com/pydata/xarray/pull/2650#issuecomment-454164295 (though some of these may require multiple inputs, even though there's only one output?)

I'd be happy to give this a go if we agree on the broad design

For names, it's a shame the current `apply` isn't called `map`, and this could be `apply`. I'm not a great fan of `apply_raw`, since I'm can't yet see why it's any more 'raw'.

PR #2677, opened 2019-01-15
Small typo

max-sixty:
@fujiisoup is this a small typo?

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -622,7 +622,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         Coarsen the long time series by averaging over every four days.
 
         >>> da = xr.DataArray(np.linspace(0, 364, num=364),
-        ...                   dims='time',
+        ...                   dim='time',
         ...                   coords={'time': pd.date_range(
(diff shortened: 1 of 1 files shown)


xarray/core/common.py
```diff
@@ -622,7 +622,7 @@ def coarsen(self, dim=None, boundary='exact', side='left',
         Coarsen the long time series by averaging over every four days.
 
         >>> da = xr.DataArray(np.linspace(0, 364, num=364),
-        ...                   dims='time',
+        ...                   dim='time',
```
fujiisoup: I think `dims` is the correct argument...

fujiisoup:
No problem.
I'm often bothered the variation of arguments, such as `dim`, `dims`, `coord`, `coords`, etc...
Sometimes it is better to use a pural form to indicate that the method accepts the multiple dimensions, but I think it would be nicer if we can converge them.

I know it is too late, though...

max-sixty:
FYI this one does look like a mistake: https://github.com/pydata/xarray/blob/master/xarray/core/variable.py#L1649

(btw I'm seeing these because I'm using them to do them for `rolling_exp`, because they're so good!)

PR #2678, opened 2019-01-15
Hotfix for #2662

### LINKED ISSUE #2662: open_mfdataset in v.0.11.1 is very slow
I have several repositories corresponding to different time periods of time.
Each repository contains several netCDF files with different variables. 
Here is a simplified example: `path/period1/` contains `varsABC_period1.nc` and `varsDEF_period1.nc`, and  `path/period2/` contains `varsABC_period2.nc` and `varsDEF_period2.nc`. All variables have the dimension T (time), but the other dimensions can be different (e.g., A has dimensions [T,X1,Y1], B has dimensions [T,X1,Y2], ...).
Before version v0.11.1 a was able to easily open a dataset as follow:
```python
ds=xr.open_mfdataset('path\period*\*.nc', concat_dim='T')

```
However, `open_mfdataset` now fails. It actually takes forever without returning any output, so my guess is that it's just very very slow.
I've noticed that I'm able to create the dataset concatenating the files with the same variables first, then merging them:
```python
dsABC=xr.open_mfdataset('path\period*\varsABC*.nc', concat_dim='T')
dsDEF=xr.open_mfdataset('path\period*\varsDEF*.nc', concat_dim='T')
ds=xr.merge([dsABC, dsDEF])
```
I can't find what's the update in  v0.11.1 that is causing this behavior. Any idea?

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.6 | packaged by conda-forge | (default, Oct 12 2018, 14:08:43) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)]
python-bits: 64
OS: Linux
OS-release: 3.10.0-862.3.2.el7.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None
libhdf5: 1.10.4
libnetcdf: 4.6.2

xarray: 0.11.2
pandas: 0.23.4
numpy: 1.14.3
scipy: 1.1.0
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
cyordereddict: None
dask: 1.0.0
distributed: 1.25.2
matplotlib: 3.0.0
cartopy: None
seaborn: 0.9.0
setuptools: 40.4.3
pip: 18.1
conda: 4.5.12
pytest: 3.8.2
IPython: 7.0.1
sphinx: 1.8.1

</details>

TomNicholas:
- [x] Closes #2662 
 
Explained in #2662. Also renamed some variables slightly for clarity.

Not sure how to add a test without refactoring the groupby into a separate function, as in it's current form the problem only manifests as a (huge) slowdown.

### DIFF
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -499,10 +499,12 @@ def _auto_combine_1d(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
     # This is just the old auto_combine function (which only worked along 1D)
     if concat_dim is not None:
         dim = None if concat_dim is _CONCAT_DIM_DEFAULT else concat_dim
-        grouped = itertools.groupby(datasets, key=lambda ds: tuple(sorted(ds)))
+        sorted_datasets = sorted(datasets, key=lambda ds: tuple(sorted(ds)))
+        grouped_by_vars = itertools.groupby(sorted_datasets,
+                                            key=lambda ds: tuple(sorted(ds)))
         concatenated = [_auto_concat(list(ds_group), dim=dim,
                                      data_vars=data_vars, coords=coords)
-                        for id, ds_group in grouped]
+                        for id, ds_group in grouped_by_vars]
     else:
         concatenated = datasets
     merged = merge(concatenated, compat=compat)
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -493,15 +493,18 @@ def _auto_combine_all_along_first_dim(combined_ids, dim, data_vars,
     return new_combined_ids
 
 
+def vars_as_keys(ds):
+    return tuple(sorted(ds))
+
+
 def _auto_combine_1d(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
                      compat='no_conflicts',
                      data_vars='all', coords='different'):
     # This is just the old auto_combine function (which only worked along 1D)
     if concat_dim is not None:
         dim = None if concat_dim is _CONCAT_DIM_DEFAULT else concat_dim
-        sorted_datasets = sorted(datasets, key=lambda ds: tuple(sorted(ds)))
-        grouped_by_vars = itertools.groupby(sorted_datasets,
-                                            key=lambda ds: tuple(sorted(ds)))
+        sorted_datasets = sorted(datasets, key=vars_as_keys)
+        grouped_by_vars = itertools.groupby(sorted_datasets, key=vars_as_keys)
         concatenated = [_auto_concat(list(ds_group), dim=dim,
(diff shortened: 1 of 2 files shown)


TomNicholas:
This only crept through the tests because `merge` was forgiving with what it would merge - should we change the arguments to merge in `auto_combine` to be something like `compat='no_conflicts'` so that it fails out instead?

shoyer:
> This only crept through the tests because `merge` was forgiving with what it would merge - should we change the arguments to merge in `auto_combine` to be something like `compat='no_conflicts'` so that it fails out instead?

We could need `compat='equals'` to be stricter, but I suspect that some some users do rely on this "no conflicts" behavior.

This might be a good idea for tests, though.

xarray/core/combine.py
```diff
@@ -499,10 +499,12 @@ def _auto_combine_1d(datasets, concat_dim=_CONCAT_DIM_DEFAULT,
     # This is just the old auto_combine function (which only worked along 1D)
     if concat_dim is not None:
         dim = None if concat_dim is _CONCAT_DIM_DEFAULT else concat_dim
-        grouped = itertools.groupby(datasets, key=lambda ds: tuple(sorted(ds)))
+        sorted_datasets = sorted(datasets, key=lambda ds: tuple(sorted(ds)))
```
shoyer: nit: define a single key function and reuse it here and below, e.g., `key_func = lambda ds: tuple(sorted(ds))`

xarray/tests/test_combine.py
```diff
@@ -661,7 +661,18 @@ def test_merge_one_dim_concat_another(self):
                  Dataset({'foo': ('x', [2, 3])})],
                 [Dataset({'bar': ('x', [10, 20])}),
                  Dataset({'bar': ('x', [30, 40])})]]
-        actual = auto_combine(objs, concat_dim=[None, 'x'])
+        actual = auto_combine(objs, concat_dim=[None, 'x'], compat='equals')
+        assert_identical(expected, actual)
+
+    def test_internal_odering(self):
```
shoyer: Does this test fail with the old logic?

TomNicholas: Yes, it gives a `MergeError`. 

Though maybe I need to dig more, because I thought it would pass with `compat='no_conflicts'`, and fail with `compat='equals'`...

shoyer:
thanks @TomNicholas !

Issue #2681, opened 2019-01-15
Close stale issues?


max-sixty:
As discussed with @jhamman , I've been attempting to manually close / follow up on stale issues & PRs, given we had 500+ issues

Any thoughts on doing this automatically with https://github.com/probot/stale? 

This could ask for an affirmation after an issue has been uncommented for a year, and then close a month later if there's no response

shoyer:
Thanks for taking the lead here! I'd be happy to give the bot a try.

PR #2684, opened 2019-01-16
Config for closing stale issues

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/2681

Initially set at two years, and max rate of one / hour - we can make more aggressive when we know it's working

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -5,7 +5,7 @@ daysUntilStale: 700  # start with a large number and reduce shortly
 
 # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
 # Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
-daysUntilClose: 7 # start with a large number and reduce over time
+daysUntilClose: 30
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
 exemptLabels:
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,58 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 700  # start with a large number and reduce shortly
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7 # start with a large number and reduce over time
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,58 @@
...
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7 # start with a large number and reduce over time
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+  - "[Status] Maybe Later"
+
```
dcherian: Add "feature_request" and "API_design" (not sure of syntax)? Those seem to hang around for a while.

.github/stale.yml
```diff
@@ -0,0 +1,58 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 700  # start with a large number and reduce shortly
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7 # start with a large number and reduce over time
```
fmaussion: I would rather set to stale after one year, but set daysUntilClose to 4 weeks or so.

max-sixty: I agree re one year, but let's start with longer and then reduce as we know it's working?

.github/stale.yml
max-sixty: I'll merge if you're OK with this @dcherian , unless there are any other comments?

.github/stale.yml
```diff
@@ -0,0 +1,58 @@
...
+  - "[Status] Maybe Later"
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
```
jhamman: I think it would be reasonable for this to be true

.github/stale.yml
dcherian: Fine by me. Sorry for the delay in response.

max-sixty:
This is now active, but rate limited at one per hour and with a two year tolerance. Let me know any thoughts and I'll adjust as necessary

Issue #1270, opened 2017-02-15
BUG: Resample on PeriodIndex not working?


max-sixty:
```python

import xarray as xr
import pandas as pd
da = xr.DataArray(pd.Series(1, pd.period_range('2000-1', '2000-12', freq='W')).rename_axis('date'))

da.resample('B', 'date', 'ffill')

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-eb64a66a8d1f> in <module>()
      3 da = xr.DataArray(pd.Series(1, pd.period_range('2000-1', '2000-12', freq='W')).rename_axis('date'))
      4
----> 5 da.resample('B', 'date', 'ffill')

/Users/maximilian/drive/workspace/xarray/xarray/core/common.py in resample(self, freq, dim, how, skipna, closed, label, base, keep_attrs)
    577         time_grouper = pd.TimeGrouper(freq=freq, how=how, closed=closed,
    578                                       label=label, base=base)
--> 579         gb = self.groupby_cls(self, group, grouper=time_grouper)
    580         if isinstance(how, basestring):
    581             f = getattr(gb, how)

/Users/maximilian/drive/workspace/xarray/xarray/core/groupby.py in __init__(self, obj, group, squeeze, grouper, bins, cut_kwargs)
    242                 raise ValueError('index must be monotonic for resampling')
    243             s = pd.Series(np.arange(index.size), index)
--> 244             first_items = s.groupby(grouper).first()
    245             if first_items.isnull().any():
    246                 full_index = first_items.index

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pandas/core/generic.py in groupby(self, by, axis, level, as_index, sort, group_keys, squeeze, **kwargs)
   3989         return groupby(self, by=by, axis=axis, level=level, as_index=as_index,
   3990                        sort=sort, group_keys=group_keys, squeeze=squeeze,
-> 3991                        **kwargs)
   3992
   3993     def asfreq(self, freq, method=None, how=None, normalize=False):

/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pandas/core/groupby.py in groupby(obj, by, **kwds)
   1509         raise TypeError('invalid type: %s' % type(obj))
   1510
-> 1511     return klass(obj, by, **kwds)
   1512
   1513

[...]

PR #2687, opened 2019-01-17
Enable resampling on PeriodIndex

max-sixty:
This allows resampling with `PeriodIndex` objects by keeping the `group` as an index rather than coercing to a DataArray (which coerces any non-native types to objects)

I'm still getting one failure around the name of the IndexVariable still being `__resample_dim__` after resample, but wanted to socialize the approach of allowing a `name` argument to `IndexVariable` - is this reasonable?

 - [x] Closes https://github.com/pydata/xarray/issues/1270
 - [x] Tests added

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -722,6 +722,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         # TODO support non-string indexer after removing the old API.
 
         from .dataarray import DataArray
+        from .variable import IndexVariable
         from .resample import RESAMPLE_DIM
         from ..coding.cftimeindex import CFTimeIndex
 
@@ -745,10 +746,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
             )
         dim, freq = indexer.popitem()
 
-        dim_name = dim
-        dim_coord = self[dim]
-
-        if isinstance(self.indexes[dim_name], CFTimeIndex):
+        if isinstance(self.indexes[dim], CFTimeIndex):
             raise NotImplementedError(
                 'Resample is currently not supported along a dimension '
                 'indexed by a CFTimeIndex.  For certain kinds of downsampling '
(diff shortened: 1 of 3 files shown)


max-sixty:
Let me know any thoughts / alternative approaches here. 

To emphasize - the issue with coercing the labels to a `DataArray` is that non-native dtypes, such as `PeriodIndex` get converted to objects, and pandas' resampling machinery (which xarray uses) requires the standard indexes.

xarray/core/common.py
```diff
@@ -761,12 +759,12 @@ def resample(self, indexer=None, skipna=None, closed=None, label=None,
                 'errors.'
             )
 
-        group = DataArray(dim_coord, coords=dim_coord.coords,
-                          dims=dim_coord.dims, name=RESAMPLE_DIM)
+        group = IndexVariable(
+            data=self.indexes[dim], dims=[dim], name=RESAMPLE_DIM)
```
shoyer: I *think* we could alternatively fix this by replace `dim_coord` in the `data` argument with `dim_coord.to_index()`, e.g., compare:
```
ipdb> DataArray(dim_coord, coords=dim_coord.coords, dims=dim_coord.dims, name=RESAMPLE_DIM).to_index()
Index([2000-01-01 00:00, 2000-01-01 06:00, 2000-01-01 12:00, 2000-01-01 18:00,
       2000-01-02 00:00, 2000-01-02 06:00, 2000-01-02 12:00, 2000-01-02 18:00,
       2000-01-03 00:00, 2000-01-03 06:00],
      dtype='object', name='time')
ipdb> DataArray(dim_coord.to_index(), coords=dim_coord.coords, dims=dim_coord.dims, name=RESAMPLE_DIM).to_index()
PeriodIndex(['2000-01-01 00:00', '2000-01-01 06:00', '2000-01-01 12:00',
             '2000-01-01 18:00', '2000-01-02 00:00', '2000-01-02 06:00',
             '2000-01-02 12:00', '2000-01-02 18:00', '2000-01-03 00:00',
             '2000-01-03 06:00'],
            dtype='period[6H]', name='time', freq='6H')
```

We go to some effort to preserve pandas.Index classes when passed into DataArray/Variable but apparently don't do it consistently everywhere. So there's also probably a lower level fix somewhere (perhaps in `xarray.core.variable.as_compatible_data`?) to ensure that this happens more consistently.

xarray/core/variable.py
```diff
@@ -1789,12 +1789,17 @@ class IndexVariable(Variable):
     unless another name is given.
     """
 
-    def __init__(self, dims, data, attrs=None, encoding=None, fastpath=False):
+    def __init__(self, dims, data, attrs=None, encoding=None, fastpath=False, name=None):
```
shoyer: I would rather not add an explicit `name` parameter to `IndexVariable`.

If anything, I would think about removing the `name` property. Long term, I'd like to rename `IndexVariable` -> `ImmutableVariable` and move the pandas adapter logic into class that we can pass into the `data` argument of `Variable`.

xarray/core/common.py
max-sixty: Great! I'll have a look

max-sixty:
This is so old that probably it wouldn't even be a good place to start from. Sorry I never pushed it through. I'll close.

Issue #1839, opened 2018-01-19
Add simple array creation functions for easier unit testing


nbren12:
When I am writing unit tests for routines that involve `DataArray` objects many lines of code are devoted to creating mock objects. Here is an example of a unit test I recently wrote to test some code which computes the fluid derivative of a field given the velocity. 
```python
def test_material_derivative():

    dims = ['x', 'y', 'z', 'time']
    coords = {dim: np.arange(10) for dim in dims}
    shape = [coords[dim].shape[0] for dim in coords]

    f = xr.Dataset({'f': (dims, np.ones(shape))}, coords=coords)
    f = f.f

    one = 0 *f +1
    zero = 0*f


    md = material_derivative(zero, one, zero, f.x + 0*f)
    np.testing.assert_array_almost_equal(md.values, 0)

    md = material_derivative(one, zero, zero, f.x + 0*f)
    np.testing.assert_array_almost_equal(md.isel(x=slice(1,-1)).values,
                                         one.isel(x=slice(1,-1)).values)

    md = material_derivative(zero, one, zero, f.y + 0*f)
    np.testing.assert_array_almost_equal(md.isel(y=slice(1,-1)).values,
                                         one.isel(y=slice(1,-1)).values)

    md = material_derivative(zero, zero, one, f.z + 0*f)
    np.testing.assert_array_almost_equal(md.isel(z=slice(1,-1)).values,
                                         one.isel(z=slice(1,-1)).values)
```

As you can see, I devote many lines to initializing a 4D data array of all ones, where all the coordinates are `np.arange(10)` objects. It isn't too hard to do this once, but it gets pretty annoying to do many times, especially when I forget how the DataArray and Dataset constructors work. Now, I can do something like `xr.DataArray(np.ones(...))`, but I would still have to initialize the coordinates if I use them.

In any case, having some sort of functions like `xr.ones`, `xr.zeros`, and `xr.rand` which initialize the coordinates and data would be very nice.

shoyer:
What would use suggest for the signature of a function like `xr.ones`?

I think the main reason why we don't have these helpers is that `xr.DataArray(np.ones(3, 4), dims=['x', 'y'])` is nearly as succinct 

I suppose we could possibly do better now that newer versions of Python (3.6+) guarantee order for keyword arguments, so we could have something like `xr.ones(x=3, y=4)`.

nbren12:
[...]

PR #2690, opened 2019-01-18
Add create_test_data to public testing API

### LINKED ISSUE #2686: Is `create_test_data()` public API?
We want to encourage people to use and extend xarray, and we already provide testing functions as public API to help with this.

One function I keep using when writing code which uses xarray is `xarray.tests.test_dataset.create_test_data()`. This is very useful for quickly writing tests for the same reasons that it's useful in xarray's internal tests, but it's not explicitly public API. This means that there's no guarantee it won't change/disappear, which is [not ideal](https://github.com/boutproject/xBOUT/issues/26) if you're trying to write a test suite for separate software. But so many tests in xarray rely on it that presumably it's not going to get changed. 

Is there any reason why it shouldn't be public API? Is there something I should use instead?

TomNicholas:
- [x] Closes #2686 and #1839

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -591,6 +591,7 @@ Testing
    testing.assert_equal
    testing.assert_identical
    testing.assert_allclose
+   testing.create_test_data
 
 Exceptions
 ==========
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -41,6 +41,12 @@ Enhancements
   report showing what exactly differs between the two objects (dimensions /
   coordinates / variables / attributes)  (:issue:`1507`).
   By `Benoit Bovy <https://github.com/benbovy>`_.
+- The function :py:func:`xarray.testing.create_test_data` has been added to the
+  public testing API. It creates a small example dataset, and is provided as a
+  convenience function for when you are  writing tests for code which acts on
+  xarray objects (:issue:`2686`).
+  By `Tom Nicholas <http://github.com/TomNicholas>`_.
+
 
 Bug fixes
 ~~~~~~~~~
--- a/xarray/testing.py
+++ b/xarray/testing.py
@@ -2,9 +2,11 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+from pandas import date_range
 
 from xarray.core import duck_array_ops
 from xarray.core import formatting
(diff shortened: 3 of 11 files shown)


shoyer:
I am slightly reluctant about this because it would mean effectively freezing the output of this helper function. I guess that would probably be OK, but the design was never quite fully considered!

max-sixty:
> it would mean effectively freezing the output of this helper function

We could add a `param` or `version` arg so we can make changes in the future, if this is a concern?

TomNicholas:
> it would mean effectively freezing the output of this helper function

Or we could have a public `create_example_data`, and an internal `create_test_data`? That way to begin with `create_example_data` is just an alias to the internal `create_test_data`, but if we end up needing to change the internal one (to test for some new type of common indexing edge case for example) then we can just separate them?

max-sixty:
The reason I had liked this was that I'm often creating a dataset manually, either in a repl or as a fixture in a test, and that added an overhead in both time and reduced consistency.
I recently started using the tutorial data, but that requires a network connection (and I kept forgetting the name...)

But I can empathize re compatibility issues; I defer to @shoyer on the balance, or whether `xarray.testing.dummy_dataset('some_name')` would be feasible

shoyer:
Maybe having a simple example available under just `xarray.testing.dummy_dataset()` is a good idea after all. If we do ever want to change it for xarray's own testing purposes, that would be quite easy.

That said... if we're going to emesh this in our API lets at least think a little bit to make sure we cover most of the special cases we care about. I think we cover most of the dtypes we care about, but let's also add at least:
- Some global attributes
- A scalar data variable (it's easy to forget about these)

Anything else?

We might also consider tweaking the names -- `dim1`, `dim2` and `dim3` is not very descriptive. Even `x`, `y` and `z` would be better.

max-sixty:
😁 

The only other features I could imagine adding would be a coord over multiple dimensions, and a multi-indexed dimension (but potentially we leave the latter out until we're completely confident about the conventions)

Agree on the names!

@TomNicholas if you want to split this up at all lmk

keewis:
this has been unintentionally closed while renaming `main`, probably because the feature branch was removed. Feel free to reopen if you want to work on this in the future.

PR #2693, opened 2019-01-19
Update asv.conf.json

TomAugspurger:
Is xarray 3.5+ now? Congrats, I didn't realize that.

This started failing the benchmark machine, which I was tending to last night.


shoyer:
Thanks Tom! Yes, we’ve gone to 3.5+ only on the master branch.

PR #2698, opened 2019-01-23
add line break to stale message

max-sixty:
Currently the message is interpreted as a single line

Please post any other feedback


max-sixty:
Merging given the immediate impact, but I'll check here for other feedback

PR #2706, opened 2019-01-24
Appending to zarr store

### LINKED ISSUE #2022: Enable Append/concat to existing zarr datastore
Following discussion from https://github.com/pangeo-data/pangeo/issues/19

How would we go about implementing a concat or append function for zarr data stores? I am imagining something like xr.concat here. Its not clear to me how this would work when using `open_mfdataset`.

```python
zarray_1 = xr.open_zarr(store=gcsmap)

ds2 = xr.open_dataset(path_to_netcdf)

xr.concat([zarray_1, ds2])
```
#### Problem description

If you are using cloud storage facility like gcs, `ds.to_zarr` can fail before the completion of the upload. This is a problem for multi-TB datasets as the entire process needs to be restarted without any way to resume where you left off. 

#### Expected Output

new zarr dataset with additional dataset appended along appropriate dim

jendrikjoe:
This pull request allows to append an xarray to an existing datastore.
 - [x] Closes #2022
 - [x] Tests will be added. Wanted to get an opinion if this is what is imagined by the community
To filter the data written to the array, the dimension over which the data will be appended has to be explicitly stated. If someone has an idea how to overcome this, I would be more than happy to incorporate the necessary changes into the PR.
Cheers,
Jendrik


rabernat:
Hi @jendrikjoe -- thanks for submitting a PR to address one of the most important issues in xarray (IMHO)! I am very excited about your contribution and am looking forward to getting this feature merged.

I have many questions about how this works. I think the best way to move forward is to wait until we have a test for the append feature which involves the following steps:
- Write a dataset to a zarr store
- Open the store in append mode
- Append data along a particular dimension

Seeing the code that accomplishes this will help clarify for me what is happening.

Thanks again for your contribution, and welcome to xarray!

jendrikjoe:
Hi @rabernat,

happy to help! I love using xarray. I added the test for the append mode. 
One is making sure, that it behaves like the 'w' mode, if no data exist at the target path.
The other one is testing what you described. The append_dim argument is actually the same as the dim argument for concat.
Hope that helps clarifying my code :)

rabernat:
Ok, with the example, I can see a bit better how this works.

Here is my main concern: there doesn't appear to be any alignment checking between the target dataset and the new data. The only check that happens is whether a variable with the same name already exists in the target store, if so, append is used (rather than creating a new array). What if the coordinates differ? What if the attributes differ?

I'm not sure this is a deal-breaker. But we should be very clear about this in the docs.

xarray/backends/zarr.py
```diff
@@ -312,40 +313,99 @@ def encode_variable(self, variable):
...
+                    encoded_attrs[k2] = self.encode_attribute(v2)
+
+                zarr_array = self.ds.create(name, shape=shape, dtype=dtype,
+                                            fill_value=fill_value, **encoding)
+                zarr_array.attrs.put(encoded_attrs)
+                zarr_array[...] = v.data
+            if append:
+                if self.append_dim is None:
+                    raise ValueError('The dimension on which the data is \
+                     appended has to be named.')
```
rabernat: What if we just want to add a new variable to an existing zarr store. This PR could hypothetically support that case as well, but in that case, there is no append_dim to specify.

jendrikjoe: As mentioned in the other comment, it should already work, but I will add another test for it :+1:

davidbrochart:
Hi @jendrikjoe,

Thanks for your PR, I am very interested in it because this is something I was hacking around (see [here](https://github.com/davidbrochart/pangeo_upload/blob/master/py/trmm2pangeo.py)). In my particular case, I want to append along a time dimension, but it looks like your PR currently doesn't support it. In the following example `ds2` should have a time dimension ranging from 2000-01-01 to 2000-01-06:
```python
import xarray as xr
import pandas as pd

ds0 = xr.Dataset({'temperature': (['time'],  [50, 51, 52])}, coords={'time': pd.date_range('2000-01-01', periods=3)})
ds1 = xr.Dataset({'temperature': (['time'],  [53, 54, 55])}, coords={'time': pd.date_range('2000-01-04', periods=3)})

ds0.to_zarr('temp')
ds1.to_zarr('temp', mode='a', append_dim='time')

ds2 = xr.open_zarr('temp')
```
But it's not the case:
```
ds2.time
<xarray.DataArray 'time' (time: 6)>
array(['2000-01-01T00:00:00.000000000', '2000-01-02T00:00:00.000000000',
       '2000-01-03T00:00:00.000000000', '2000-01-01T00:00:00.000000000',
       '2000-01-02T00:00:00.000000000', '2000-01-03T00:00:00.000000000'],
      dtype='datetime64[ns]')
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2000-01-02 ... 2000-01-03
```
Maybe it's not intended to work with time dimensions yet?

davidbrochart:
To make it work, time dimensions would have to be treated separately because zarr doesn't encode absolute time values but deltas relative to a reference (see https://github.com/davidbrochart/pangeo_upload/blob/master/py/trmm2pangeo.py#L108).

jendrikjoe:
Hey @davidbrochart,
thanks for all your input and as well for the resarch on how zarr stores the data. 
I would actually claim that the calculation of the accurate relative time should be handled by the zarr append function.
An exception would be of course if xarray is storing the data with deltas to a reference as well? 
Then I would try collecting the minimum and offsetting the input by this.
@rabernat can you provide input on that?

davidbrochart:
zarr stores the reference in the `.zattrs` file:
```
{
    "_ARRAY_DIMENSIONS": [
        "time"
    ],
    "calendar": "proleptic_gregorian",
    "units": "days since 2000-01-01 00:00:00"
}
```

jendrikjoe:
I will check as well how xarry stores times to check if we have to add the offset to the xarray first or if this can be resolved with a PR to zarr :)

shoyer:
We should definitely always make sure that we write data consistently (e.g., for dates), but checking for alignment of all coordinates could be expensive/slow. Potentially a keyword argument `ignore_alignment=True` would be a good way for user to opt-out of checking index coordinates for consistency.

davidbrochart:
When we use this feature e.g. to store data that is produced every day, we might start with a data set that has a small size on the time dimension, and thus the chunks will be chosen according to this initial shape. When we append to this data set, will the chunks be kept as in the initial zarr archive? If so, we might end up with a lot of small chunks on the time dimension, where ideally we would have chosen only one chunk.

rabernat:
> We should definitely always make sure that we write data consistently (e.g., for dates), but checking for alignment of all coordinates could be expensive/slow.

This implies we should be checking for attributes compatibility before calling `zarr.append`.

xarray/backends/zarr.py
```diff
@@ -312,40 +313,99 @@ def encode_variable(self, variable):
...
+                zarr_array = self.ds[name]
+                append = True
+            except KeyError:
+                encoding = _extract_zarr_variable_encoding(
+                    v, raise_on_invalid=check)
+                encoded_attrs = OrderedDict()
+                # the magic for storing the hidden dimension data
+                encoded_attrs[_DIMENSION_KEY] = dims
+                for k2, v2 in attrs.items():
+                    encoded_attrs[k2] = self.encode_attribute(v2)
```
rabernat: What if we pulled this attribute encoding out before the `try` block.  Then we check `encoded_attrs` against `zarr_array.attrs` before appending.

xarray/backends/zarr.py
```diff
@@ -312,40 +313,99 @@ def encode_variable(self, variable):
...
+            values
+        writer : ArrayWriter
+        unlimited_dims : list-like
+            List of dimension names that should be treated as unlimited
+            dimensions.
+            dimension on which the zarray will be appended
+            only needed in append mode
+        """
+
+        variables, attributes = self.encode(variables, attributes)
```
rabernat: This is where the encoding from `datetime64` to `int64` with `days since ...` units happens.

If we wanted to make sure that the encoding of the new variables is compatible with the target store, we would have to peek at the target store encodings and explicitly put them in the new variable encoding.

jendrikjoe: Will try doing that :) Will probably take a while, but I might be able to do that on Monday or Tuesday :+1:

rabernat: This is not an easy problem. Advice from @shoyer and @jhamman would be valuable.

shoyer: I would even consider opening up the zarr store (into an xarray.Dataset) before doing any appending. Then it’s easy to decode all the metadata and ensure consistency of the appended data.

jendrikjoe: I would try to avoid opening the whole zarr store for performance reasons and instead just try pulling the encodings from the array attributes. I think the only way to really solve this is adding the possibility to all CF encoders to use a specific encoding if one is passed. 
This would allow to parse the encoding to https://github.com/pydata/xarray/blob/master/xarray/conventions.py#L204 from https://github.com/pydata/xarray/blob/master/xarray/backends/zarr.py#L209 and get the correctly encoded array for the append. What do you think?

davidbrochart:
Hi @jendrikjoe, do you plan to work on this PR again in the future? I think it would be a great contribution to xarray.

davidbrochart:
May I try and take this work over?

rabernat:
@davidbrochart I would personally be happy to see anyone work on this. I'm sure @jendrikjoe would not mind if we make it a team effort!

jendrikjoe:
Sure everyone feel welcome to join in! Sorry for the long silence. Kind of a busy time right now 😉

On Apr 1, 2019 08:47, Ryan Abernathey <notifications@github.com> wrote:

@davidbrochart<https://github.com/davidbrochart> I would personally be happy to see anyone work on this. I'm sure @jendrikjoe<https://github.com/jendrikjoe> would not mind if we make it a team effort!

davidbrochart:
@jendrikjoe I think you need to give me the permission to push to the branch in your fork.

davidbrochart:
Or should I open a new PR?

jendrikjoe:
I added you to the fork :) But feel free to do whatever is easiest for you :)

davidbrochart:
Thanks @jendrikjoe, I just pushed to your fork: to make sure that the encoding of the appended variables is compatible with the target store, we explicitly put the target store encodings in the appended variable.

jendrikjoe:
Nice :+1:

On Apr 4, 2019 21:24, David Brochart <notifications@github.com> wrote:

Thanks @jendrikjoe<https://github.com/jendrikjoe>, I just pushed to your fork: to make sure that the encoding of the appended variables is compatible with the target store, we explicitly put the target store encodings in the appended variable.

xarray/backends/zarr.py
```diff
@@ -312,40 +314,111 @@ def encode_variable(self, variable):
...
+        ds = None
+        for vn, v in variables.items():
+            name = _encode_variable_name(vn)
+            if name in self.ds:
+                # existing variable, get its encoding
+                # and apply it to appended variable
+                variables_encoded[vn] = v
+                if ds is None:
+                    ds = open_zarr(self.ds.store, auto_chunk=False)
+                v.encoding = ds[vn].encoding
```
rabernat: I find the logic here a bit hard to follow. In general, the pattern of setting a variable to `None` and then assigning it on the first pass of a loop can lead to confusion (at least for me).

What about something like?

```python
existing_variables = set([vn for vn in variables if _encode_variable_name(vn) in self.ds])
if len(existing_variables) > 0:
    ds = open_zarr(self.ds.store, auto_chunk=False)
    variables_encoded = OrderedDict()
    for vn in existing_variables:
        ...
```

davidbrochart:
@rabernat you're right, I took your suggestion into account in my last commit. I also rewrote the test.

rabernat:
Where do we stand on this PR? @davidbrochart - do you feel this is ready for a final review? Or do you want advice or feedback on anything?

davidbrochart:
I don't think it's ready yet. I think I should address the chunking issue of the appended dimension, as explained in https://medium.com/pangeo/continuously-extending-zarr-datasets-c54fbad3967d. For instance if we append along a time dimension, the time coordinate (which is a 1-D array) will have very small chunks, instead of maybe only one.

davidbrochart:
I added a `chunk_dim` parameter which allows to rechunk the appended coordinate. I think it is ready for a final review now.

rabernat:
Hi @davidbrochart. I'm really sorry it takes me so long between reviews of your PR. It is very important work, and I appreciate your continued patience.

I looked at your new code, and I noticed that `chunk_dim` does not appear in the tests. I think it is important to test this parameter and verify that it works as expected. (This would also help me understand how it works, since it's not totally clear from the docs.)

davidbrochart:
No problem @rabernat, and thanks a lot for your time in reviewing this PR.

I added a test for `chunk_dim`. Please let me know if this is clearer, and if I should explain further in the docs.

shikharsg:
Hi all, not sure if I could be doing something wrong myself, but the below might be a bug in this PR.

I checked out the `jendrik_joe:append_zarr` branch. When calling `to_zarr` on a `Dataset` with the `compute=False` argument, the data array should not be populated in the target `zarr` store right? But it seems to be doing just that(I though only metadata should be initialised).
```python
>>> import zarr
>>> import xarray
>>> ds = xarray.Dataset(
...     {'arr1': (['dim1', 'dim2'], [[10,11],[12,13]])},
...     coords={'dim1': [1, 2], 'dim2': [1,2]}
... )
>>> store = dict()
>>> ds = ds.chunk(dict(dim1=1, dim2=1))
>>> ds
<xarray.Dataset>
Dimensions:  (dim1: 2, dim2: 2)
Coordinates:
  * dim1     (dim1) int64 1 2
  * dim2     (dim2) int64 1 2
Data variables:
    arr1     (dim1, dim2) int64 dask.array<shape=(2, 2), chunksize=(1, 1)>
>>> store = dict()
>>> ds.to_zarr(store, mode='w', compute=False)
Delayed('_finalize_store-72c3ec95-6ef1-473a-a0de-e81c98eb9576')
>>> xarray.open_zarr(store)['arr1'].values
array([[10, 11],
       [12, 13]])
```

Instead this does not happen in the xarray master branch:
```python
>>> import zarr
>>> import xarray
>>> ds = xarray.Dataset(
...     {'arr1': (['dim1', 'dim2'], [[10,11],[12,13]])},
...     coords={'dim1': [1, 2], 'dim2': [1,2]}
... )
>>> ds = ds.chunk(dict(dim1=1, dim2=1))
>>> ds
<xarray.Dataset>
Dimensions:  (dim1: 2, dim2: 2)
Coordinates:
  * dim1     (dim1) int64 1 2
  * dim2     (dim2) int64 1 2
Data variables:
    arr1     (dim1, dim2) int64 dask.array<shape=(2, 2), chunksize=(1, 1)>
>>> store = dict()
>>> ds.to_zarr(store, mode='w', compute=False)
Delayed('_finalize_store-965b0a89-bff6-4adc-8bab-e47a77e762f3')
>>> xarray.open_zarr(store)['arr1'].values
array([[4611686018427387904, 4611686018427387904],
       [4611686018427387904, 4611686018427387904]])
```
Python: 3.6.7
```python
>>> xarray.__version__
'0.12.1+45.g519b3986'
>>> zarr.__version__
'2.2.1.dev185'
>>> 
```
OS:
```
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.2 LTS
Release:        18.04
Codename:       bionic
```

I am also trying to figure out why this is happening(assuming this is a bug) and will post updates soon.

@davidbrochart

davidbrochart:
Thanks @shikharsg for looking into that. This PR and master have diverged quite a bit, I will need to merge the changes, I will let you know.

shikharsg:
I think I figured out the problem. Previously the `ZarrStore` defaults to the `store` method of the parent class `AbstractWritableDataStore`. The `store` method of `AbstractWritableDataStore` uses `set_variables`(which belongs to `AbstractWritableDataStore`) which adds the "source and target" variables to the `ArrayWriter` class, i.e. the array is not written until the call to `_finalize_store` which happens in the `xarray.backends.api.to_zarr` function(or not in `compute=False`).
Instead your PR implements another `store` function in `ZarrStore` which uses `set_variables`(belonging to `ZarrStore`) which directly writes the data to the target array, by  passing the `ArrayWriter` class. I think if you just use `ArrayWriter` in `set_variables` it should fix the problem.

I will be happy to push a fix, if permissions are given.

davidbrochart:
It would be great if you could fix it. @jendrikjoe can give you the permission to push to the branch in his fork.

jendrikjoe:
Gave you the permissions @shikharsg

rabernat:
Let’s make sure this new scenario is covered by tests!

shikharsg:
will do

jendrikjoe:
Hey there everyone, sorry for not working on this for so long from my side. I just picked it up again and realised that the way the encoding works, all the datatypes and the maximum string lengths in the first xarray have to be representative for all others. Otherwise the following cuts away every char after the second:

    ds0 = xr.Dataset({'temperature': (['time'],  ['ab', 'cd', 'ef'])}, coords={'time': [0, 1, 2]})
    ds1 = xr.Dataset({'temperature': (['time'],  ['abc', 'def', 'ghijk'])}, coords={'time': [0, 1, 2]})
    ds0.to_zarr('temp')
    ds1.to_zarr('temp', mode='a', append_dim='time')

It is solvable when explicitly setting the type before writing:

    ds0 = xr.Dataset({'temperature': (['time'],  ['ab', 'cd', 'ef'])}, coords={'time': [0, 1, 2]})
    ds0['temperature'] = ds0.temperature.astype(np.dtype('S5'))
    ds1 = xr.Dataset({'temperature': (['time'],  ['abc', 'def', 'ghijk'])}, coords={'time': [0, 1, 2]})
    ds0.to_zarr('temp')
    ds1.to_zarr('temp', mode='a', append_dim='time')

It becomes however worse when using non-ascii characters, as they get encoded in [zarr.py l:218](https://github.com/pydata/xarray/blob/442e938c2c5dcc0f192f0db2348cd679d07c16cb/xarray/backends/zarr.py#L218), but with the next chunk that is coming in the check in [conventions.py l:86](https://github.com/pydata/xarray/blob/442e938c2c5dcc0f192f0db2348cd679d07c16cb/xarray/conventions.py#L86) fails. So I think we actually have to resolve the the TODO in [zarr.py l:215](https://github.com/pydata/xarray/blob/442e938c2c5dcc0f192f0db2348cd679d07c16cb/xarray/backends/zarr.py#L215) before this is able to be merged. Otherwise, the following leads to multiple issues:
 
    ds0 = xr.Dataset({'temperature': (['time'],  ['ab', 'cd', 'ef'])}, coords={'time': [0, 1, 2]})
    ds1 = xr.Dataset({'temperature': (['time'],  ['üý', 'ãä', 'õö'])}, coords={'time': [0, 1, 2]})
    ds0.to_zarr('temp')
    ds1.to_zarr('temp', mode='a', append_dim='time')
    xr.open_zarr('temp').temperature.values

The only way to work around this issue is to explicitly encode the data beforehand to utf-8:
    
    from xarray.coding.variables import safe_setitem, unpack_for_encoding
    from xarray.coding.strings import encode_string_array
    from xarray.core.variable import Variable
    
    def encode_utf8(var, string_max_length):
         dims, data, attrs, encoding = unpack_for_encoding(var)
         safe_setitem(attrs, '_Encoding', 'utf-8')
         data = encode_string_array(data, 'utf-8')
         data= data.astype(np.dtype(f"S{string_max_length*2}"))
         encoding['_Encoding'] = 'utf-8'

    ds0 = xr.Dataset({'temperature': (['time'],  ['ab', 'cd', 'ef'])}, coords={'time': [0, 1, 2]})
    ds0['temperature'] = encode_utf8(ds0.temperature).astype(np.dtype('S4'))
    safe_setitem(attrs, '_Encoding', 'utf-8')
    ds1 = xr.Dataset({'temperature': (['time'],  ['üý', 'ãä', 'õö'])}, coords={'time': [0, 1, 2]})
    ds1['temperature'] = encode_utf8(ds1.temperature).astype(np.dtype('S4'))
    ds0.to_zarr('temp')
    ds1.to_zarr('temp', mode='a', append_dim='time')
    xr.open_zarr('temp').temperature.values


Even though this is doable if it is known in advance, we should definitely mention this in the documentation or fix this by fixing the encoding itself. What do you think? 

Cheers,

Jendrik

rabernat:
@jendrikjoe - thanks for digging in and finding this important issue!

This PR has been hanging around for a long time. (A lot of that is on me!) It would be good to get something merged soon. Here's what I propose.
- Identify which datatypes can easily be appended now (e.g. floats, etc.) and which cannot (variable length strings)
- Raise an error if append is called on the incompatible datatypes
- Move forward with this PR, which is otherwise very nearly ready
- Open a new issue to keep track of the outstanding incompatible types, which require upstream resolution in zarr

How does that sound to everyone?

rabernat:
@shikharsg - are the issues you found in https://github.com/pydata/xarray/pull/2706#issuecomment-498194520 now resolved and covered by tests?

jendrikjoe:
> @jendrikjoe - thanks for digging in and finding this important issue!
> 
> This PR has been hanging around for a long time. (A lot of that is on me!) It would be good to get something merged soon. Here's what I propose.
> 
>     * Identify which datatypes can easily be appended now (e.g. floats, etc.) and which cannot (variable length strings)
> 
>     * Raise an error if append is called on the incompatible datatypes
> 
>     * Move forward with this PR, which is otherwise very nearly ready
> 
>     * Open a new issue to keep track of the outstanding incompatible types, which require upstream resolution in zarr
> 
> 
> How does that sound to everyone?

This sounds like a plan. I will try to work on getting this ready tonight and tmrw.
Let us see how far I can get.

jendrikjoe:
I build a filter that is raising a value error as soon as any variable has a dtype different from any subclass of np.number or np.string_. I as well build test for that and added a function to manually convert dynamic sized string arrays to fixed sized ones.

I as well wrote a test for @shikharsg issue and can reproduce it. The test is currently commented to not fail the pipeline as I wanted to discuss if this is a blocking issue or if we should merge it and raise a new issue for it.
It seems to be originating from the fact that we moved away from using writer.add and instead are actually calling the zarr functions directly. There should be a way to change this back to do it lazily, but that will probably take time.
What do you think?

shikharsg:
Hi all, sorry for the delay. I was on break for 10 days. I have started working on it now and should be able to do this in a couple of days.

A quick thing I noticed as I started working on this: currently the [zarr.Array.append](https://zarr.readthedocs.io/en/stable/api/core.html#zarr.core.Array.append) function(which handles the resize of the array) is being used to append to the array, and as of now, it's not being done asynchronously using a dask delayed object(as I pointed out in my earlier comment). So to do it asynchronously, my plan is to resize the target array, and then the delayed object would write to the appropriate region of the resized array([dask.array.store](https://docs.dask.org/en/latest/array-api.html#dask.array.store) is currently being used for this). But if for whatever reason, the delayed object does not end up being called, we would end up with nonsense values in the resized portion of the array(or whatever is the fill value). For this reason I wonder if I should put the resize in the delayed object too. Thoughts?

shikharsg:
I have fixed the compute=False for appending to zarr store, but there are two issues that remain

- when appending to an existing array, I resize the array first, and the return a `dask.delayed` object which fills up the new region of the array when `compute` is called on it. So if the delayed object does not get computed for whatever reason, the new portion of the array will end up with nonsense values. For this reason I was wondering if the resize function should be in the delayed object itself so the array is not resized in advance.
- the `compute=False` will not work when the `chunk_dim` argument is set, i.e. instead of lazily appending when the compute method is called on the delayed object, it will directly append to the target store when the `to_zarr` method with `mode='a'` is called. The reason is because when the `chunk_dim` argument is set, it reads the original array from memory, appends to that array in memory, and overwrites the appended array to the target store. I understand that this was done because of the concern that @davidbrochart raised about doing very frequent appends to the array(for example hourly or six hourly as happens in climate modelling), and the resulting smallness of the chunk size of the dimension being appended to. But @davidbrochart, I would almost recommend removing this `chunk_dim` argument because the concern you raised can be overcome as follows: suppose you have a Dataset as follows:
```python
precip = 10 * np.random.rand(2, 2, 3)
ds = xr.Dataset({'temperature': (['x', 'y', 'time'],  temp),
                 'precipitation': (['x', 'y', 'time'], precip)},
                         'lat': (['x', 'y'], lat),
                         'time': pd.date_range('2014-09-06', periods=3),
                         'reference_time': pd.Timestamp('2014-09-05')})

```
and want to append it to very often. When calling the `to_zarr` function the first time call it like so:
```python
>>> store = dict()
>>> ds.to_zarr(store, encoding={'temperature': {'chunks':(100,100,100)}, 'precipitation': {'chunks':(100,100,100)}})
<xarray.backends.zarr.ZarrStore object at 0x7f7027778d68>
>>> import zarr
>>> zarr.open_group(store)['temperature'].info
Name               : /temperature
Type               : zarr.core.Array
Data type          : float64
Shape              : (2, 2, 3)
Chunk shape        : (100, 100, 100)
Order              : C
Read-only          : False
Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
Store type         : builtins.dict
No. bytes          : 96
No. bytes stored   : 33903 (33.1K)
Storage ratio      : 0.0
Chunks initialized : 1/1
>>> zarr.open_group(store)['precipitation'].info
Name               : /precipitation
Type               : zarr.core.Array
Data type          : float64
Shape              : (2, 2, 3)
Chunk shape        : (100, 100, 100)
Order              : C
Read-only          : False
Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
Store type         : builtins.dict
No. bytes          : 96
No. bytes stored   : 33906 (33.1K)
Storage ratio      : 0.0
Chunks initialized : 1/1
```
and then this large chunk size remains (100,100,100) (or whatever other large numbers you may want)
The `chunk_dim` functionality, as it works now, is not feasible for very large arrays, since it is essentially reading the entire array into memory(and we may not have too much memory) and then overwriting the target store, because it essentially "rechunks" the array. 

Thoughts please

davidbrochart:
You're right @shikharsg, the `chunk_dim` argument can be removed. I was not very happy with the complexity it brought as well.

shikharsg:
> @shikharsg - are the issues you found in [#2706 (comment)](https://github.com/pydata/xarray/pull/2706#issuecomment-498194520) now resolved and covered by tests?

@rabernat these are now resolved

doc/whats-new.rst
```diff
@@ -188,6 +188,7 @@ Other enhancements
   report showing what exactly differs between the two objects (dimensions /
   coordinates / variables / attributes)  (:issue:`1507`).
   By `Benoit Bovy <https://github.com/benbovy>`_.
+- Added append capability to the zarr store.
```
rabernat: Need to credit all the contributors to this PR.

xarray/backends/common.py
```diff
@@ -158,24 +158,35 @@ class ArrayWriter:
...
             self.sources.append(source)
             self.targets.append(target)
+            if region:
+                self.regions.append(region)
         else:
-            target[...] = source
+            if region:
+                target[region] = source
+            else:
+                target[...] = source
```
rabernat: :+1: This seems like a perfect (and rather general) way to handle appends at the dask level.

xarray/tests/test_backends.py
```diff
@@ -1613,11 +1614,26 @@ def test_write_persistence_modes(self):
...
+                obj1.to_zarr(store_target, mode='a')
+
+        # encode strings to fixed sized arrays
+        obj1['string_var'] = encode_utf8(obj1.string_var, 2)
+        obj2['string_var'] = encode_utf8(obj2.string_var, 2)
+        with self.create_zarr_target() as store_target:
+            obj1.to_zarr(store_target, mode='w')
+            obj2.to_zarr(store_target, mode='a', append_dim='time')
+            original = xr.concat([obj1, obj2], dim='time')
+            assert_identical(original, xr.open_zarr(store_target))
```
rabernat: :+1: love this test 😄

xarray/backends/common.py
```diff
@@ -158,26 +158,37 @@ class ArrayWriter:
...
             # targets = [dask.delayed(t) for t in self.targets]
+
+            if not self.regions:
+                regions = None
+            else:
+                regions = self.regions
             delayed_store = da.store(self.sources, self.targets,
                                      lock=self.lock, compute=compute,
-                                     flush=True)
+                                     flush=True, regions=regions)
```
shoyer: `self.regions` is always non-empty, so you should probably just use it directly (and delete the conditional above):
```suggestion
                                     flush=True, regions=self.regions)
```

rabernat:
Thanks @shoyer for your more careful review of this PR. As usual you pick up on all the important details.

shikharsg:
Thanks @shoyer  @rabernat for the detailed review. All the comments have been addresses except the removal of the `encode_utf8` function. In this last commit: https://github.com/pydata/xarray/pull/2706/commits/a6ff49436a620eb4071c0529d785370b2eb739b0, I have tried to do that. In the same commit I have also tried to take a shot at variable length strings using @shoyer's comments from here: https://github.com/pydata/xarray/issues/2724#issuecomment-458808896. Please let me know if this is acceptable or should I revert this commit or something.

shikharsg:
also any idea why all the AppVeyor builds are failing since yesterday? I did not change any build file in any of my commits.

shoyer:
+1 for saving as native variable length strings in zarr. I didn't realize that was an option when we originally wrote xarray's zarr support, but its definitely a much cleaner way to do things in most cases.

shoyer:
The Appveyor build failures are definitely not your fault, please ignore them

xarray/backends/zarr.py
```diff
@@ -313,40 +315,127 @@ def encode_variable(self, variable):
...
+                # new variable
+                encoding = _extract_zarr_variable_encoding(
+                    v, raise_on_invalid=check)
+                encoded_attrs = OrderedDict()
+                # the magic for storing the hidden dimension data
+                encoded_attrs[_DIMENSION_KEY] = dims
+                for k2, v2 in attrs.items():
+                    encoded_attrs[k2] = self.encode_attribute(v2)
+
+                if dtype == object:
```
shoyer: Does this work instead here?
```suggestion
                if coding.strings.is_vlen_dtype(dtype):
```
(This is a more robust check that should only allow true arrays of strings)

xarray/backends/api.py
```diff
@@ -1023,8 +1027,24 @@ def save_mfdataset(datasets, paths, mode='w', format=None, groups=None,
                              for w, s in zip(writes, stores)])
 
 
+def _validate_datatypes_for_zarr_append(dataset):
+    """DataArray.name and Dataset keys must be a string or None"""
+    def check_dtype(var):
+        if (not np.issubdtype(var.dtype, np.number)
+                and not np.issubdtype(var.dtype, np.string_)
+                and not np.issubdtype(var.dtype, np.unicode_)
+                and not var.dtype == object):
```
shoyer:
```suggestion
                and not coding.strings.is_unicode_dtype(var.dtype)):
```

shikharsg: this seems to work:
```python
         if (not np.issubdtype(var.dtype, np.number)
                and not coding.strings.is_unicode_dtype(var.dtype)
                and not var.dtype == object):
```
while this does not:
```python
         if (not np.issubdtype(var.dtype, np.number)
                and not np.issubdtype(var.dtype, np.string_)
                and not np.issubdtype(var.dtype, np.unicode_)
                and not coding.strings.is_unicode_dtype(var.dtype)):
```

shikharsg: or were you suggesting this?
```python
         if (not np.issubdtype(var.dtype, np.number)
                and not coding.strings.is_unicode_dtype(var.dtype)):
```
this does not work

shikharsg:
adding a new variable currently errors if we don't provide the `append_dim` argument:
```python
>>> import xarray as xr
>>> import pandas as pd
>>> ds0 = xr.Dataset({'temperature': (['time'],  [50, 51, 52])}, coords={'time': pd.date_range('2000-01-01', periods=3)})
>>> ds1 = xr.Dataset({'pressure': (['time'],  [50, 51, 52])}, coords={'time': pd.date_range('2000-01-01', periods=3)})
>>> store = dict()
>>> ds0.to_zarr(store, mode='w')
<xarray.backends.zarr.ZarrStore object at 0x7fae926505c0>
>>> ds1.to_zarr(store, mode='a')
  File "/home/shikhar/code/xarray/xarray/core/dataset.py", line 1374, in to_zarr
    consolidated=consolidated, append_dim=append_dim)
  File "/home/shikhar/code/xarray/xarray/backends/api.py", line 1071, in to_zarr
    dump_to_store(dataset, zstore, writer, encoding=encoding)
  File "/home/shikhar/code/xarray/xarray/backends/api.py", line 928, in dump_to_store
    unlimited_dims=unlimited_dims)
  File "/home/shikhar/code/xarray/xarray/backends/zarr.py", line 366, in store
    unlimited_dims=unlimited_dims)
  File "/home/shikhar/code/xarray/xarray/backends/zarr.py", line 406, in set_variables
    "was not set".format(name)
ValueError: variable 'time' already exists, but append_dim was not set
```

this works:
```python
>>> import xarray as xr
>>> import pandas as pd
>>> ds0 = xr.Dataset({'temperature': (['time'],  [50, 51, 52])}, coords={'time': pd.date_range('2000-01-01', periods=3)})
>>> ds1 = xr.Dataset({'pressure': (['time'],  [50, 51, 52])}, coords={'time': pd.date_range('2000-01-01', periods=3)})
>>> store = dict()
>>> ds0.to_zarr(store, mode='w')
<xarray.backends.zarr.ZarrStore object at 0x7fae926505c0>
>>> ds1.to_zarr(store, mode='a', append_dim='asdfasdf')
>>> xr.open_zarr(store)
<xarray.Dataset>
Dimensions:      (time: 3)
Coordinates:
  * time         (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03
Data variables:
    pressure     (time) int64 dask.array<shape=(3,), chunksize=(3,)>
    temperature  (time) int64 dask.array<shape=(3,), chunksize=(3,)>
```

will push a fix for this in a bit

shikharsg:
I have fixed the above error now and all comments have now been addressed.

@rabernat @shoyer

rabernat:
> adding a new variable currently errors if we don't provide the `append_dim` argument:

Is this scenario now covered by the tests? Sorry if the answer is obvious; it's hard for me to discern just by looking at the code.

shoyer:
Just to be clear, we do always requiring writing `append_dim` if you want to append values along a dimension, right? And we raise an informative error if you write `append_dim='not-a-valid-dimension'`?

shikharsg:
> > adding a new variable currently errors if we don't provide the `append_dim` argument:
> 
> Is this scenario now covered by the tests? Sorry if the answer is obvious; it's hard for me to discern just by looking at the code.

@rabernat , the scenario I am talking about is adding a new `DataArray` to an existing `Dataset`(in which case we do not have to specify an `append_dim` argument). Yes it is covered by tests, specifically see the `with` clause here: https://github.com/pydata/xarray/pull/2706/files#diff-df47fcb9c2f1f7dfc0c6032d97072af2R1636

> Just to be clear, we do always requiring writing `append_dim` if you want to append values along a dimension, right? And we raise an informative error if you write `append_dim='not-a-valid-dimension'`?

@shoyer We do always require `append_dim` when appending to an existing array, but I just realized that it does not raise an error when `append_dim='not-valid'`, but silently fails to append to the existing array. Let me write a test for that and push

shikharsg:
it's done. I fixed it by opening the zarr dataset beforehand using `xr.open_zarr`

xarray/backends/api.py
shoyer: OK, I understand why this doesn't work: we aren't using xarray's standard encoding pipeline when appending variables. This probably is fine for strings -- in the worst case, users will just get an error downstream from Zarr instead.

xarray/backends/zarr.py
```diff
@@ -313,40 +315,126 @@ def encode_variable(self, variable):
...
+            dtype = v.dtype
+            shape = v.shape
+
+            fill_value = attrs.pop('_FillValue', None)
+            if v.encoding == {'_FillValue': None} and fill_value is None:
+                v.encoding = {}
+            if name in self.ds:
+                zarr_array = self.ds[name]
+                """
+                If variable is a dimension of an existing array
```
shoyer: Please use `#` for comments instead of strings, which should be reserved for docstrings.

shikharsg: sorry, yes. That comment is not actually needed(it's from code I already removed). So I'll remove that.

xarray/backends/zarr.py
```diff
@@ -313,40 +315,126 @@ def encode_variable(self, variable):
...
+            variables_without_encoding, attributes)
+
+        if len(existing_variables) > 0:
+            # there are variables to append
+            # their encoding must be the same as in the store
+            ds = open_zarr(self.ds.store, chunks=None)
+            variables_with_encoding = OrderedDict()
+            for vn in existing_variables:
+                variables_with_encoding[vn] = variables[vn]
+                variables_with_encoding[vn].encoding = ds[vn].encoding
```
shoyer: This modifies an argument that was passed into the function in-place, which in general should be avoided due to unexpected side effects  in other parts of the code. It would be better to shallow copy the `Variable` before overriding `encoding`, e.g., with `variables[vn].copy(deep=False)`.

xarray/backends/zarr.py
```diff
@@ -313,40 +315,126 @@ def encode_variable(self, variable):
...
+                                 if _encode_variable_name(vn) in self.ds])
+        new_variables = set(variables) - existing_variables
+        variables_without_encoding = OrderedDict([(vn, variables[vn])
+                                                 for vn in new_variables])
+        variables_encoded, attributes = self.encode(
+            variables_without_encoding, attributes)
+
+        if len(existing_variables) > 0:
+            # there are variables to append
+            # their encoding must be the same as in the store
```
shoyer: Are there any unit tests that verify that encoding is kept consistent? This would be nice to add, if not.

Probably a good example would be dataset saved with scale/offset encoding, where the new dataset to be appended does not have any encoding provided. We could verify that probably scaled values are read back from disk.

xarray/backends/api.py
```diff
@@ -1040,11 +1078,16 @@ def to_zarr(dataset, store=None, mode='w-', synchronizer=None, group=None,
     _validate_dataset_names(dataset)
     _validate_attrs(dataset)
 
+    if mode == "a":
```
shoyer: Can we raise an error if `encoding` was passed explicitly into `to_zarr()` and specifies encoding for any existing variables? I.e., `ds.to_arr(..., mode='a', encoding={'existing_variable': ....})`. I think this would always indicate a programming mistake, given that we throw away these variable encodings anyways.

shikharsg:
I have implemented all the changes suggested and refactored the append tests as all tests were previously crammed into `test_write_persistence_modes`

I'm not sure why the build fails. In all the failed checks, it is these two tests that are failing:
```
================================== FAILURES ===================================
_________________________ test_rolling_properties[1] __________________________

da = <xarray.DataArray (a: 3, time: 21, x: 4)>
array([[[0.561926, 0.243845, 0.601879, 0.733398],
        [0.500418, 0.84942...ordinates:
  * time     (time) datetime64[ns] 2000-01-01 2000-01-02 ... 2000-01-21
Dimensions without coordinates: a, x

    def test_rolling_properties(da):
        rolling_obj = da.rolling(time=4)
    
        assert rolling_obj.obj.get_axis_num('time') == 1
    
        # catching invalid args
        with pytest.raises(ValueError) as exception:
            da.rolling(time=7, x=2)
>       assert 'exactly one dim/window should' in str(exception)
E       AssertionError: assert 'exactly one dim/window should' in '<ExceptionInfo ValueError tblen=4>'
E        +  where '<ExceptionInfo ValueError tblen=4>' = str(<ExceptionInfo ValueError tblen=4>)

xarray\tests\test_dataarray.py:3715: AssertionError
_________________________ test_rolling_properties[1] __________________________

ds = <xarray.Dataset>
Dimensions:  (time: 10, x: 8, y: 2)
Coordinates:
  * x        (x) float64 0.0 0.1429 0.2857 0.4286 0....-1.152 -0.6704 ... -0.9796 -1.884 0.4049
    z2       (time, y) float64 -1.218 -0.9627 -1.398 ... -0.3552 0.1446 0.3392

    def test_rolling_properties(ds):
        # catching invalid args
        with pytest.raises(ValueError) as exception:
            ds.rolling(time=7, x=2)
>       assert 'exactly one dim/window should' in str(exception)
E       AssertionError: assert 'exactly one dim/window should' in '<ExceptionInfo ValueError tblen=4>'
E        +  where '<ExceptionInfo ValueError tblen=4>' = str(<ExceptionInfo ValueError tblen=4>)

xarray\tests\test_dataset.py:4845: AssertionError
============================== warnings summary ===============================
```

I have no idea why as the same two tests pass on my local machine

xarray/tests/test_backends.py
```diff
@@ -1647,17 +1654,82 @@ def test_encoding_kwarg_fixed_width_string(self):
...
-        super(CFEncodedBase, self).test_append_write()
+        ds, ds_to_append, _ = create_append_test_data()
+        with self.create_zarr_target() as store_target:
+            ds.to_zarr(store_target, mode='w')
+            ds_to_append.to_zarr(store_target, mode='a', append_dim='time')
+            original = xr.concat([ds, ds_to_append], dim='time')
+            assert_identical(original, xr.open_zarr(store_target))
 
     @pytest.mark.xfail(reason="Zarr stores can not be appended to")
     def test_append_overwrite_values(self):
```
shikharsg: what is this test exactly supposed to do? Is it something like what's happening here: https://github.com/pydata/xarray/blob/master/xarray/tests/test_backends.py#L881?
If so,this kind of functionality is not implemented in this PR yet. You can only "append" to existing variables, not overwrite existing data. That would have to be done with the 'w' mode.

shoyer:
@shikharsg the test failure should be fixed on master (by https://github.com/pydata/xarray/pull/3059).

xarray/tests/test_backends.py
shoyer: Yes, I think you have  analyzed this correctly. This  is fine.

shoyer:
OK, thank you @shikharsg, @jendrikjoe and everyone else who worked on this !

PR #2715, opened 2019-01-26
Fix CRS being WKT instead of PROJ.4

djhoese:
See https://github.com/mapbox/rasterio/blob/master/CHANGES.txt#L7 for the change in rasterio 1.0.14 and my conda-forge issue where I discovered this: https://github.com/conda-forge/gdal-feedstock/issues/262

This was put in a bugfix (micro version number) release of rasterio which made it harder to discover. @sgillies @snowman2 I saw that you brought this up and how it affected xarray (https://rasterio.groups.io/g/dev/message/68). If this pull request is a duplicate of a fix one of you made let me know.

To xarray devs, if I need to make any other changes (whats-new?) also let me know.

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -282,10 +282,10 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         # CRS is a dict-like object specific to rasterio
         # If CRS is not None, we convert it back to a PROJ4 string using
         # rasterio itself
-    try:
-        attrs['crs'] = riods.crs.to_proj4()
-    except AttributeError:
-        attrs['crs'] = riods.crs.to_string()
+        try:
+            attrs['crs'] = riods.crs.to_proj4()
+        except AttributeError:
+            attrs['crs'] = riods.crs.to_string()
     if hasattr(riods, 'res'):
         # (width, height) tuple of pixels in units of CRS
         attrs['res'] = riods.res
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -282,7 +282,10 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         # CRS is a dict-like object specific to rasterio
         # If CRS is not None, we convert it back to a PROJ4 string using
         # rasterio itself
+    try:
         attrs['crs'] = riods.crs.to_proj4()
+    except AttributeError:
+        attrs['crs'] = riods.crs.to_string()
     if hasattr(riods, 'res'):
         # (width, height) tuple of pixels in units of CRS
         attrs['res'] = riods.res
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -282,7 +282,7 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         # CRS is a dict-like object specific to rasterio
         # If CRS is not None, we convert it back to a PROJ4 string using
         # rasterio itself
-        attrs['crs'] = riods.crs.to_string()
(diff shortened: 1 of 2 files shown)


xarray/backends/rasterio_.py
```diff
@@ -282,7 +282,7 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
         # CRS is a dict-like object specific to rasterio
         # If CRS is not None, we convert it back to a PROJ4 string using
         # rasterio itself
-        attrs['crs'] = riods.crs.to_string()
+        attrs['crs'] = riods.crs.to_proj4()
```
sgillies: I recommend the following:

```suggestion
        attrs['crs'] = riods.crs.to_dict()
```

It's both forward and backward compatible. As far as I can tell from scanning the xarray source, the data array's crs attribute is rarely used, except as an argument to rasterio.transform where a dict should do just fine.

djhoese: @sgillies It's not just how xarray uses the crs attribute but users. The reason I made this PR was because the xarray crs attribute changed from PROJ4 to WKT.

snowman2: Just curious, why does it need to be a PROJ.4 string in xarray?

djhoese: For now, backwards compatibility. I don't think there is any other strong reason other than what the original developer chose (it is a fairly popular CRS format). If you think it should be WKT then that should probably go in its own issue.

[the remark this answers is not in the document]
djhoese: Does `to_proj4` not exist in older versions of rasterio?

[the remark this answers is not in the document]
snowman2: It was added [here](https://github.com/mapbox/rasterio/pull/1597/files#diff-32bf07e9fdf7a58371b06e785d4766d8R323).

fmaussion: The main reason was serialization to NetCDF: attributes can't be dict or objects, and strings are a perfect representation of the CRS. Any other idea welcome though.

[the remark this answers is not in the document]
fmaussion: > Sounds like a good idea to discuss the use of WKT versus PROJ.4 format for CRS strings in another issue.

Agreed - can you start the discussion?

fmaussion:
can your merge from master? The failing tests should be solved by https://github.com/pydata/xarray/pull/2720

djhoese:
@fmaussion Done. And I merged @snowman2's suggestion and fixed the indent (I'm guessing github's editor made it difficult to see but it was off by one indentation).

fmaussion:
I'm going to merge this tomorrow unless there are further concerns

dcherian:
Does this need a whats-new entry?

shoyer:
> Does this need a whats-new entry?

Yes, I would recommend that here -- especially for cases where we're changing behavior without a deprecation cycle (which is sometime the best path forward)

shoyer:
> This PR isn't changing any functionality. It is making the same functionality available with newer versions of rasterio.

OK, good point. I do still try to lean towards adding notes about bug fixes in "what's new" but it isn't strictly required.

djhoese:
@shoyer Added something to the whats-new. Let me know if anything needs changing.

shoyer:
thanks David, that looks great

djhoese:
@shoyer Any idea when a 0.11.4 or 0.12 will be released? I'm trying to work around some other rasterio bugs and would like to remove the restriction on the rasterio version used in my CI tests, but that requires this PR.

shoyer:
Hopefully quite soon! see https://github.com/pydata/xarray/pull/2777 for
current status

PR #2716, opened 2019-01-26
ENH: resample methods with tolerance

### LINKED ISSUE #2695: Resample with limit/tolerance
#### Upsampling methods cannot be limited

It is comes very handy to limit the scope of the resample method e.g. `nearest` in time series. In pandas the `limit` argument can be given, such that:

```python
import pandas as pd
import datetime as dt

dates=[dt.datetime(2018,1,1), dt.datetime(2018,1,2)]
data=[10,20]
df=pd.DataFrame(data,index=dates)
df.resample('1H').nearest(limit=1)
```

This leads to
```
2018-01-01 00:00:00  10.0
2018-01-01 01:00:00  10.0
2018-01-01 02:00:00   NaN
2018-01-01 03:00:00   NaN
2018-01-01 04:00:00   NaN
...
2018-01-01 20:00:00   NaN
2018-01-01 21:00:00   NaN
2018-01-01 22:00:00   NaN
2018-01-01 23:00:00  20.0
2018-01-02 00:00:00  20.0
```

Currently:
```python
import xarray as xr
xdf = xr.Dataset.from_dataframe(df)
xdf.resample({'index':'1H'}).nearest(limit=1)
```
leads to
```
TypeError: nearest() got an unexpected keyword argument 'limit'
```
#### Problem description

This is very helpful, as one might not want to fill gaps with the `nearest` method indefinitely.
To my understanding the following modifications might be made by comparisions to the pandas code:

**/xarray/core/resample.py**
```python
    def _upsample(self, method, limit=None, *args, **kwargs):
        ...
        elif method in ['pad', 'ffill', 'backfill', 'bfill', 'nearest']:
            kwargs = kwargs.copy()
            kwargs.update(**{self._dim: upsampled_index})
            return self._obj.reindex(method=method, tolerance=limit, *args, **kwargs)
        ...```

and
```python
   def nearest(self, limit=None):
        """Take new values from nearest original coordinate to up-sampled
        frequency coordinates.
        """
        return self._upsample('nearest',limit=limit)
```

So I think, with the `tolerance` keyword, reindex supports already the limit, but it just hasn't been forwarded to the `_upsample` and `nearest` methods.

#### Current Output
```python
import xarray as xr
>>> xdf = xr.Dataset.from_dataframe(df)
>>> xdf.resample({'index':'1H'}).nearest()
<xarray.Dataset>
Dimensions:  (index: 25)
Coordinates:
  * index    (index) datetime64[ns] 2018-01-01 ... 2018-01-02
Data variables:
    0        (index) int64 10 10 10 10 10 10 10 10 ... 20 20 20 20 20 20 20 20
```

However, it would be nice, if the following would work:
```python
xdf.resample({'index':'1H'}).nearest(limit=1)

<xarray.Dataset>
Dimensions:  (index: 25)
Coordinates:
  * index    (index) datetime64[ns] 2018-01-01 ... 2018-01-02
Data variables:
    0        (index) float64 10.0 10.0 nan nan nan nan ... nan nan nan 20.0 20.0
```

observingClouds:
* ENH: resample methods bfill, pad, nearest accept tolerance keyword

 - [x] Closes #2695
 - [x] Tests added

### DIFF
--- a/doc/time-series.rst
+++ b/doc/time-series.rst
@@ -203,8 +203,8 @@ and supports all of its schemes. All of these resampling operations work on both
 Dataset and DataArray objects with an arbitrary number of dimensions.
 
 In order to limit the scope of the methods ``ffill``, ``bfill``, ``pad`` and
-``nearest`` the ``tolerance`` argument can be set. Data that has indices
-outside of the given ``tolerance`` are set to ``NaN``.
+``nearest`` the ``tolerance`` argument can be set in coordinate units.
+Data that has indices outside of the given ``tolerance`` are set to ``NaN``.
 
 .. ipython:: python
 


jhamman:
Thanks @observingClouds - At first glance, this seems right but we'll want some tests to confirm. You'll probably want to put these in [test_missing.py](https://github.com/pydata/xarray/blob/master/xarray/tests/test_missing.py). That's where the existing tests for bfill/pad/etc are so should be easy to extend for this feature.

observingClouds:
Sure @jhamman, I'll add some tests.
However, I thought the test should rather go into [test_dataarray.py](https://github.com/pydata/xarray/blob/master/xarray/tests/test_dataarray.py) than [test_missing.py](https://github.com/pydata/xarray/blob/master/xarray/tests/test_missing.py), because
this is an improvement to resample/_upsample?

Something like
```python
    def test_upsample_tolerance(self):
        # Test tolerance keyword for upsample methods bfill, pad, nearest
        times = pd.date_range('2000-01-01', freq='1D', periods=2)
        times_upsampeld = pd.date_range('2000-01-01', freq='6H', periods=5)
        array = DataArray(np.arange(2), [('time', times)])

        # Forward fill
        actual = array.resample(time='6H').ffill(tolerance='12H')
        expected = DataArray([[0., 0., 0., np.nan, 1.]],
                             [('time', times)])
```

jhamman:
@observingClouds - I had forgotten the upsample tests were in `test_dataarray.py`. That seems like a better place to test this functionality.

xarray/tests/test_dataarray.py
```diff
@@ -2493,6 +2493,36 @@ def test_upsample_nd(self):
...
+        actual = array.resample(time='6H').bfill(tolerance='12H')
+        expected = DataArray([0., np.nan, 1., 1., 1.],
+                             [('time', times_upsampled)])
+        assert_identical(expected, actual)
+
+        # Pad
+        actual = array.resample(time='6H').pad(tolerance='12H')
+        expected = DataArray([0., 0., 0., np.nan, 1.],
+                             [('time', times_upsampled)])
+        assert_identical(expected, actual)
```
shoyer: You can drop this case -- it's the exact same as "forward fill"

xarray/tests/test_dataarray.py
```diff
@@ -2493,6 +2493,36 @@ def test_upsample_nd(self):
...
+
+        # Nearest
+        actual = array.resample(time='6H').nearest(tolerance='12H')
```
shoyer: Maybe use `tolerance='6H'` here so we can verify that it actually works?

xarray/tests/test_dataarray.py
observingClouds: Fixed with d8acf458827893da858ebd18133669ad25d733ad

xarray/tests/test_dataarray.py
observingClouds: Fixed with d8acf458827893da858ebd18133669ad25d733ad

doc/time-series.rst
```diff
@@ -197,11 +197,20 @@ resampling group:
...
 
-For upsampling, xarray provides four methods: ``asfreq``, ``ffill``, ``bfill``,
-and ``interpolate``. ``interpolate`` extends ``scipy.interpolate.interp1d`` and
-supports all of its schemes. All of these resampling operations work on both
+For upsampling, xarray provides six methods: ``asfreq``, ``ffill``, ``bfill``, ``pad``,
+``nearest`` and ``interpolate``. ``interpolate`` extends ``scipy.interpolate.interp1d``
+and supports all of its schemes. All of these resampling operations work on both
 Dataset and DataArray objects with an arbitrary number of dimensions.
 
+In order to limit the scope of the methods ``ffill``, ``bfill``, ``pad`` and
```
dcherian: Mention that "tolerance" is in coordinate (or label) units, not index units and that the index version is limit

observingClouds: The unit for ``tolerance`` is now (a878ebc62825ea0e281960ca2e91348899f34a60) explicitly mentioned in the time-series example. However, I did not comment on the ``index`` method as this has not been implemented for the resample method, yet. The index method seems to be a bigger/other project.

PR #2726, opened 2019-01-29
Bugfix for line plot axes

### LINKED ISSUE #2725: Line plot with x=coord putting wrong variables on axes
When I try to plot the values in a 1D DataArray against the values in one of its coordinates, it does not behave at all as expected: 

```python
import numpy as np
from xarray import DataArray

current = DataArray(name='current', data=np.array([5, 8, 14, 22, 30]), dims=['time'],
                    coords={'time': (['time'], np.array([0.1, 0.2, 0.3, 0.4, 0.5])),
                            'voltage': (['time'], np.array([100, 200, 300, 400, 500]))})

print(current)

# Try to plot current against voltage
current.plot.line(x='voltage')
plt.show()
```
Output:

```
<xarray.DataArray 'current' (time: 5)>
array([ 5,  8, 14, 22, 30])
Coordinates:
  * time     (time) float64 0.1 0.2 0.3 0.4 0.5
    voltage  (time) int64 100 200 300 400 500
```
incorrect_current_plot

#### Problem description

Not only is `'voltage'` not on the x axis, but `'current'` isn't on the y axis either!

#### Expected Output

Based on the documentation (and common sense) I would have expected it to plot voltage on the x axis and current on the y axis.

(using a branch of xarray which is up-to-date with master)

TomNicholas:
- [x] Closes #2725 
 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -69,7 +69,7 @@ Bug fixes
   (e.g. '2000-01-01T00:00:00-05:00') no longer raises an error
   (:issue:`2649`).  By `Spencer Clark <https://github.com/spencerkclark>`_.
 - Fixed performance regression with ``open_mfdataset`` (:issue:`2662`).
-
+  By `Tom Nicholas <http://github.com/TomNicholas>`_.
 - Fixed supplying an explicit dimension in the ``concat_dim`` argument to
   to ``open_mfdataset`` (:issue:`2647`).
   By `Ben Root <https://github.com/WeatherGod>`_.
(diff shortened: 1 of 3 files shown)


doc/whats-new.rst
```diff
@@ -54,6 +54,8 @@ Bug fixes
   from higher frequencies to lower frequencies.  Datapoints outside the bounds
   of the original time coordinate are now filled with NaN (:issue:`2197`). By
   `Spencer Clark <https://github.com/spencerkclark>`_.
+- Line plots with the `x` argument set to a coord now plot the correct data.
```
```suggestion
- Line plots with the `x` argument set to a non-dimensional coord now plot the correct data for 1D DataArrays.
```

doc/whats-new.rst
```diff
@@ -67,7 +69,7 @@ Bug fixes
   (e.g. '2000-01-01T00:00:00-05:00') no longer raises an error
   (:issue:`2649`).  By `Spencer Clark <https://github.com/spencerkclark>`_.
 - Fixed performance regression with ``open_mfdataset`` (:issue:`2662`).
-  By `Tom Nicholas <http://github.com/TomNicholas>`_.
+
```
```suggestion
  By `Tom Nicholas <http://github.com/TomNicholas>`_.
```

Issue #2355, opened 2018-08-09
Animated plots - a suggestion for implementation


TomNicholas:
**It'd be awesome if one could animate the plots xarray creates using matplotlib just by specifying the dimension over which to animate the plot.**

This would allow for rapid visualisation of time-evolving data and could potentially be very powerful (imagine a grid of faceted 2d plots, all evolving together over time). I know that there are already some libraries which can create animated plots of xarray data (e.g. Holoviews), but I think that it's within xarray's scope (#2030) to add another dimension to its default matplotlib-style plotting capabilities.

**How?**

I saw this new package for making it easier to animate matplotlib plots using the funcanimation module: [animatplot](https://github.com/t-makaro/animatplot). It essentially works by wrapping matplotlib commands like `plt.imshow()` to instead return "blocks". These blocks can then be animated by feeding them into an `animation` class. An introductory script to plot line data can be found [here](https://animatplot.readthedocs.io/en/latest/tutorial/getting_started..html), but basically has the form

```python
import animatplot as amp
import matplotlib.pyplot as plt

X, Y = load_data_somehow
block = amp.blocks.Line(X, Y)
anim = amp.Animation([block])

anim.save_gif("animated_line") 
plt.show()
```

which creates a basic gif like this: ![animated line gif](https://user-images.githubusercontent.com/35968931/43885402-a3373002-9b6d-11e8-9b3d-f4e588a71a22.gif)

I think that it might be possible to integrate this kind of animation-plotting tool by adding an optional dimension argument to xarray's plotting methods, which if given causes the function to call the wrapped animatplot plotting command instead of the bare matplotlib one. It would then return the corresponding "block" ready to be animated. Using the resulting code might only require a few lines to create an impressive visualisation:

```python
turb2d = xr.load_dataset("turbulent_fluid_data.nc")

block = turb2d["density"].plot.imshow(animate_over='time')
anim = Animation([block])

anim.save_gif("fluid_density.gif") 
plt.show()
``` 

![n_over_time](https://user-images.githubusercontent.com/35968931/43887058-83d4161c-9b72-11e8-978d-fcb8e071a37a.gif)

**What would need changing?**

[...]

Issue #2030, opened 2018-03-30
Animate DataArrays


tomchor:
Hi, this is my first a feature request here. I've been using xarray for a while but it always comes somewhat short when I'm trying to do animations. It always ends up being a big hassle.

Is there a good way to provide some functionality to make animations easier? I really like the way xarray sets up the plots by default, but I always end up re-writing it when setting up an animation, which I always feel like is kind of a shame.

Cheers

maxim-lian:
What's an example of an animation?

tomchor:
Unfortunately I don't have any example with DataArray right now. Since I never could take advantage of DataArray's plotting capabilities for animations, I always did animations using pure Numpy. However, I'm talking about the standard matplotlib animations. Here's an example taken from [here](https://matplotlib.org/examples/animation/basic_example_writer.html):

```import numpy as np
import matplotlib
matplotlib.use("Agg")
import matplotlib.pyplot as plt
import matplotlib.animation as animation


def update_line(num, data, line):
    line.set_data(data[..., :num])
    return line,

# Set up formatting for the movie files
Writer = animation.writers['ffmpeg']
writer = Writer(fps=15, metadata=dict(artist='Me'), bitrate=1800)


fig1 = plt.figure()

data = np.random.rand(2, 25)
l, = plt.plot([], [], 'r-')
plt.xlim(0, 1)
plt.ylim(0, 1)
plt.xlabel('x')
plt.title('test')
line_ani = animation.FuncAnimation(fig1, update_line, 25, fargs=(data, l),
                                   interval=50, blit=True)
line_ani.save('lines.mp4', writer=writer)
```

benbovy:
@tomchor Note that [holoviews](http://holoviews.org/) supports xarray objects.

You can find an example of an animation made using holoviews and xarray here:
http://nbviewer.jupyter.org/gist/benbovy/e9d4f3d3be076de24b8842dae109d6e7

tomchor:
I didn't know holoviews existed, thanks for pointing that out to me. The fact that it can't create an animation file (as far as I could tell so far) does mean I can't use it, though. So I'll still try to pursue the matplotlib animation option, although maybe for most users holoviews+xarray might already be enough (I don't know).

philippjfr:
>The fact that it can't create an animation file (as far as I could tell so far) does mean I can't use it, though

You can create animation files using the matplotlib backend, as a simple example:

```python
ds = hv.Dataset(air.isel(time=range(100)))
[...]

PR #2729, opened 2019-01-30
[WIP] Feature: Animated 1D plots

TomNicholas:
This is an attempt at a proof-of-principle for making animated plots in the way I suggested in #2355. (Also relevant for #2030.)

This example code:
```python
import xarray as xr

# Load data as done in plotting tutorial
airtemps = xr.tutorial.open_dataset('air_temperature')
air = airtemps.air - 273.15
air.attrs = airtemps.air.attrs
air.attrs['units'] = 'deg C'

# Downsample to make reasonably-sized gif
data = air.isel(lat=10, time=slice(None,None,40))

# Create animated plot
anim = data.plot(animate_over='time')
anim.save('line1.gif', writer='imagemagick')
plt.show()
```
now produces this gif:
line1
~~The units on the timeline are formatted incorrectly because [this PR](https://github.com/t-makaro/animatplot/pull/21) isn't merged yet~~

I think it looks pretty good! It even animates the title properly. The actual animation creation only takes one line to do.

This currently only works for a plot with a single line, which is animated over a coordinate dimension.
~~It also required some minor modifications/bugfixes to animatplot, so it probably isn't reproducible right out of the box yet.~~ If you want to try this out then use the [develop branch](https://github.com/TomNicholas/animatplot/tree/develop) of my forked version of animatplot.

The reason I've put this up is because I wanted to

1. show people the level of complexity required, and 
2. get people's opinion on the implementation. 

I feel like although it required only ~100 lines extra to do this then the logic is very fragmented and scattered through the `plot.line` and `plot._infer_line_data` functions. In 2D this would get even more complicated, but I can't see a good way to abstract the case of animation out?

(@t-makaro I expect you will be interested in this)

EDIT: To-Do list:

- [x] Animate single line
- [x] Animated line and static line on same axes
- [x] Animate multiple lines on same axes
- [x] Multiple animated line plots on same figure
- [ ] ~~FacetGrids of multiple animated lines~~ (will leave for a later PR)
- [ ] Complete set of tests
- [x] Add animatplot as optional dependency
- [x] Add new CI tests using animatplot
- [ ] New documentation page
- [x] Fix issues with formatting of timeline label (fixed by https://github.com/t-makaro/animatplot/pull/46)


xarray/plot/plot.py
```diff
@@ -323,7 +407,15 @@ def line(darray, *args, **kwargs):
...
 
-    return primitive
+    if animate_over is None:
+        return primitive
+    else:
+        from animatplot.animation import Animation
+        anim = Animation([line_block, title_block], timeline=timeline)
+        # TODO I think ax should be passed to timeline_slider args
+        # but that just plots a single huge timeline and no line plot?!
+        anim.controls(timeline_slider_args={'text': animate_over})
```
t-makaro: You can pass ```'valfmt':'%s'``` to format datetime. It will require some code to be smarter depending on the type of data being formated. I haven't worked with datetime much, so I'm not sure the best practices about formatting them.

rabernat:
I think this is a cool example, and lots of users will probably like it.

A high-level question: why was animatplot chosen over matplotlib's animation module?

xarray/plot/plot.py
TomNicholas: > You can pass 'valfmt':'%s' to format datetime.

This gives you the date in year-month-day hour:minute:seconds.milliseconds format, but doesn't truncate all the decimal points with the milliseconds. I also haven't used `datetime64` objects much so I don't really know how to fix that in a general manner.

> If it's in the title; maybe it doesn't need to be in the slider?

I think it should be in both, although in this example the title looks good, in my experience then once you slice high-dimensional data the title becomes really complicated and it's better to definitely have it presented nicely by the timeline too.

xarray/plot/plot.py
```diff
@@ -304,10 +376,22 @@ def line(darray, *args, **kwargs):
...
 
-        ax.set_title(darray._title_for_slice())
-
-    if darray.ndim == 2 and add_legend:
-        ax.legend(handles=primitive,
+        if animate_over is None:
+            ax.set_title(darray._title_for_slice())
+        else:
+            # Would be nicer if we had something like in GH issue #266
+            frame_titles = [darray[{animate_over: i}]._title_for_slice()
```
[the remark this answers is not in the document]
TomNicholas: There's currently not much point in doing this, as the first thing animatplot will do with the generator is expand it into a list anyway.

TomNicholas:
> Re complexity: It;s probably worth extracting everything to a animate_line function in animate.py? 

I've had a go at that in the most recent commits.


> I think this is a cool example, and lots of users will probably like it.

Great - it's definitely something I personally want to do all the time.


> A high-level question: why was animatplot chosen over matplotlib's animation module?

Because animatplot already abstracts away the creation of the animation, so plotting an animation becomes almost a simple substitution `matplotlib.plot.line() -> animatplot.blocks.Line()`.

Another advantage is that animatplot's block abstraction should make it easier to compose figures made of [multiple animated plots](https://github.com/t-makaro/animatplot#animatplot), another thing people often want to do but is is awful with `matplotlib.funcanimation` directly.

The logic is explained slightly more in #2355, and also looking at [animatplot's docs](https://animatplot.readthedocs.io/en/stable/index.html) might help you see why I think it's a good idea.

We don't have to use animatplot, but if we didn't I would be suggesting reimplementing something with a similar class structure in xarray as the best approach anyway. The disadvantages of using animatplot are that it's an extra (optional) dependency, and it's not got good unit test coverage yet so someone should probably contribute tests to animatplot upstream before xarray officially relies on it. animatplot isn't particularly large, so if we wanted to reimplement something similar in xarray then that would be feasible.

xarray/plot/plot.py
TomNicholas: I've written in a special case for datetime objects, which truncates them thanks to `pandas.to_datetime()`. Unfortunately the string still gets partly obscured by the play/pause button, but I think fixing that would require either truncating the string to a fixed maximum size or changing animatplot to choose the button placement based on the size of the string. I've updated the gif in the original PR to show what it looks like now.

xarray/plot/plot.py
```diff
@@ -36,11 +36,19 @@ def _infer_line_data(darray, x, y, hue):
...
+    if ndims - animate_ndim == 1:
         huename = None
         hueplt = None
         huelabel = ''
 
+        if animate_over is not None:
+            # Set animation dimension to be along last axis of data
+            dims = list(darray.dims)
+            dims.remove(animate_over)
+            darray = darray.transpose(*dims, animate_over)
```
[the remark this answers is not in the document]
TomNicholas: You can see in my recent commits I abstracted this to a function.

xarray/plot/__init__.py
```diff
@@ -3,6 +3,8 @@
 
 from .facetgrid import FacetGrid
 
+from .animate import animate_line
```
dcherian: I have a weak preference for `xarray.plot.animate.line` instead of `xarray.plot.animate_line`. 
Similarly, the method in `animate.py` could be `line` instead of `animate_line`

xarray/plot/animate.py
```diff
@@ -0,0 +1,179 @@
...
+import numpy as np
+import pandas as pd
+
+from .plot import _infer_line_data
+from .utils import (_ensure_plottable, _interval_to_mid_points, _update_axes,
+                    _valid_other_type, get_axis, _rotate_date_xlabels,
+                    _check_animate_over, _transpose_before_animation)
+
+
+def animate_line(darray, animate_over=None, **kwargs):
```
dcherian: Weak preference for `animate` instead of  `animate_over` i.e.
`da.plot.line(x='lat', animate='time')`

xarray/plot/animate.py
```diff
@@ -0,0 +1,179 @@
...
+    xlim = kwargs.pop('xlim', None)
+    ylim = kwargs.pop('ylim', None)
+    _labels = kwargs.pop('_labels', True)
+
+    ax = get_axis(figsize, size, aspect, ax)
+    xplt, yplt, hueplt, xlabel, ylabel, huelabel = \
+        _infer_line_data(darray, x, y, hue, animate_over)
+
+    # Remove pd.Intervals if contained in xplt.values.
+    if _valid_other_type(xplt.values, [pd.Interval]):
```
dcherian: Should we just move this bit into `_infer_line_data`? I think only `xplt_val` is used in the remaining part of this function.

xarray/plot/animate.py
```diff
@@ -0,0 +1,179 @@
...
+            xplt_val = _interval_to_mid_points(xplt.values)
+            yplt_val = yplt.values
+            xlabel += '_center'
+    else:
+        xplt_val = xplt.values
+        yplt_val = yplt.values
+
+    _ensure_plottable(xplt_val, yplt_val)
+
+    fps = kwargs.pop('fps', 10)
```
dcherian: I wonder if we should just go with `animate_kwargs={'fps': 30}`. This will be tidier once we add more kwargs (`fmt` for the timeline, for example)

xarray/plot/__init__.py
TomNicholas: @dcherian I tried to implement this but couldn't work out how to have the animation functions accessible simultaneously as any of
```python
xarray.plot.animate.line(darray, animate='time')
darray.plot(animate='time')
darray.plot.line(animate='time')
darray.plot.animate(animate='time')
darray.plot.animate.line(animate='time')
```

You can see in one of the new testcases that the last two of these syntaxes don't currently work, and I can't really work out why - there's something different between `_PlotMethods` and my `_AnimateMethods` which I'm not understanding.

dcherian:
```
darray.plot.animate(animate='time')
darray.plot.animate.line(animate='time')
```
Do we really need these? They're equivalent to 
```
darray.plot(animate='time')
darray.plot.line(animate='time')
```
which is more concise and just as clear.

ping @shoyer for input.

shoyer: I agree that @dcherian's two suggestions (`darray.plot(animate='time')` and `darray.plot.line(animate='time')`) seem like enough for the user-level API.

xarray/plot/animate.py
```diff
@@ -0,0 +1,166 @@
...
+    xscale = kwargs.pop('xscale', None)  # default needs to be None
+    yscale = kwargs.pop('yscale', None)
+    xticks = kwargs.pop('xticks', None)
+    yticks = kwargs.pop('yticks', None)
+    xlim = kwargs.pop('xlim', None)
+    ylim = kwargs.pop('ylim', None)
+    _labels = kwargs.pop('_labels', True)
+
+    ax = get_axis(figsize, size, aspect, ax)
+    xplt_val, yplt_val, hueplt, xlabel, ylabel, huelabel = \
```
dcherian: Does `animate` work with `pd.Interval` step plots?

xarray/plot/animate.py
```diff
@@ -0,0 +1,166 @@
+"""
+Use this module directly:
```
dcherian: This needs an edit.

xarray/plot/animate.py
TomNicholas: I have no idea. In theory then as long as `xplt_val` and `yplt_val` get set correctly then it should? 

Have you got an example, issue, or point in the code somewhere I can look to see what a `pd.Interval` step plot consists of?

dcherian: This is the bit that tests it:
https://github.com/pydata/xarray/blob/a84ea12d34a4d1a542a34abbcf2feb3937a6a282/xarray/tests/test_plot.py#L460

dcherian: If it doesn't work; just raise an informative error...

TomNicholas:
I've done some more work on this, you can now:

### Plot multiple animated lines

```python
dat2d = air.isel(lat=[10, 15, 20])

anim = dat2d.plot(animate='time', hue='lat')
anim.save('line.gif', writer='imagemagick')
plt.show()
```
lines

### Add static lines to an animated plot

```python
# draw animated line
anim = dat1d.plot(animate='time', label='evolving')

# draw static line
# plot will be drawn on to current axes unless ax argument given
dat1d.mean(dim='time').plot(color='g', label='average')

plt.legend(loc='upper right')

anim.save('average.gif', writer='imagemagick')
plt.show()
```
average

dcherian:
There's a glitch with the legend in the first gif. Maybe the location needs to be pinned to something that is not 'best'

dcherian:
This is looking great; let me know when you're ready for another review.

I recommend saving facetgrid support for future PRs. I'm personally more excited about 2D animations and am happy to implement that once all the common infrastructure is in. With that bias in mind, I think the way to proceed would be: 

(this PR) → 2D animations → facetgrid support

rabernat:
Just pinging @jbusecke, who recently announced https://github.com/jbusecke/xmovie. He might have some insights on this problem.

jbusecke:
This looks amazing! Which problem are you referring to specifically @rabernat?

jbusecke:
Also FYI I have a PR open that will enable [xmovie](https://github.com/jbusecke/xmovie/pull/3) to write movie files (by invoking ffmpeg 'under the hood'). 
Just wanted to mention it since this might come in handy as another export option for this feature later on.

TomNicholas:
> There's a glitch with the legend in the first gif. Maybe the location needs to be pinned to something that is not 'best'

Yeah, I'm not entirely sure why that happens yet, but I guess it means the legend position should be determined automatically for only the first frame, then fixed there for the other frames.

> I recommend saving facetgrid support for future PRs. 

Agree. I think I might also leave animated step plot support for a future PR.

> I'm personally more excited about 2D animations and am happy to implement that once all the common infrastructure is in.

Help with that would be great! I also really want to get 2D animations working - I'm often trying to make animations like this:

n_over_t_visc_3 7e0

I anticipate that there might be some difficulties with cartopy integration for 2D animations, but I don't actually use cartopy in my work, so help with that would be especially appreciated.

> Just pinging jbusecke, who recently announced https://github.com/jbusecke/xmovie.

Yeah we've been talking about the relationship between this and xmovie in https://github.com/jbusecke/xmovie/issues/2. I think that movie output functionality can be tacked on later.

ahuang11:
This looks awesome; it would be really nice to have built-in xarray animation support!

>> I'm personally more excited about 2D animations and am happy to implement that once all the common infrastructure is in. 

Also wanted to point out while that's progressing along, hvplot supports 2D animations. https://hvplot.pyviz.org/
```
import xarray as xr
import hvplot.xarray
import holoviews as hv
hv.extension('matplotlib')

ds = xr.tutorial.open_dataset('air_temperature').isel(time=slice(0, 15))
hmap = ds.hvplot('lon', 'lat', dynamic=False).opts(fig_size=300, clim=(230, 300))
hv.save(hmap, 'anim.mp4', fmt='scrubber')

# ---

from IPython.core.display import display, HTML
display(HTML('anim.html'))
```
anim

spencerahill:
Just came across this PR while trying for the first time to create an animation of xarray data.  Looks like it got quite far along but then sputtered.  Did it get superseded by hvplot?

TomNicholas:
Hi @spencerahill , sorry for the slow reply. 

This functionality isn't integrated yet, but hvplot I think probably provides the better approach. You will also want to see what philipjfr came up with over on #3709 , which would allow an xarray-like syntax but with the power of holoviews.

The one remaining question for me with the holoviews approach is how easy it is to save a .gif or a video file. It seems like it [might be possible already](http://holoviews.org/user_guide/Exporting_and_Archiving.html), but if you have a go then let me know how it goes! (This is something I very very much want to see working, but isn't my priority right now)

keewis:
@TomNicholas, this has been unintentionally closed when renaming `main`. Feel free to reopen if you still plan to work on this in the future.

PR #2737, opened 2019-02-02
add h5netcdf+dask tests

### LINKED ISSUE #1571: to_netcdf fails for engine=h5netcdf when using dask-backed arrays
When using dask-backed datasets/arrays it does not seem possible to use the 'h5netcdf' engine to write to disk:

```python
import xarray as xr
ds = xr.Dataset({'a': ('x', [1, 2])}, {'x': [3, 4]}).chunk()
ds.to_netcdf("test.h5", engine='h5netcdf')
```

results in the error:

```bash
...

h5py/h5a.pyx in h5py.h5a.open()

KeyError: "Can't open attribute (can't locate attribute: 'dask')"
```

Not sure if this is a xarray or h5netcdf issue - or some inherent limitation in which case apologies!

jhamman:
- [x] Closes #1571
 - [x] Tests added


xarray/tests/test_backends.py
```diff
@@ -1916,6 +1916,46 @@ def test_dump_encodings_h5py(self):
...
+            yield ds.chunk()
+
+    def test_dataset_caching(self):
+        # caching behavior differs for dask
+        pass
+
+    @pytest.mark.xfail(reason="Failing to round trip unlimited dims")
+    def test_encoding_unlimited_dims(self):
+        # TODO: this should pass
+        super(TestH5NetCDFViaDaskData, self).test_encoding_unlimited_dims()
```
jhamman: I'm not sure why this isn't working. As far as I can tell, it passes when not using dask.

shoyer: I think the problem is that `ds.chunk()` does not preserve `encoding`. I don't think we test this behavior with netCDF4, either.

We should probably preserve `encoding` in chunks(), but also in the `roundtrip()` method we should trigger dask by passing `chunks={}` instead of calling `.chunk()` explicitly.

jhamman:
Should be good to go here once the tests finish.

xarray/tests/test_backends.py
```diff
@@ -1278,10 +1289,11 @@ class TestNetCDF4ViaDaskData(TestNetCDF4Data):
     @contextlib.contextmanager
     def roundtrip(self, data, save_kwargs={}, open_kwargs={},
                   allow_cleanup_failure=False):
+        open_kwargs['chunks'] = open_kwargs.get('chunks', 1)
```
shoyer: There's a mutating mutable default values issue here -- this is changing `open_kwargs` for every invocation.

Instead, we should make `open_kwargs=None` and then set the chunks keyword argument as is approrpriate.

Also: use `open_kwargs.setdefault('chunks', -1)` instead. `setdefault()` is a little more succinct, and `-1` means "full sized chunks" which will be faster for our test suite.

jhamman: @shoyer - good points. Addressed in my latest commit.

jhamman:
This is ready for a final review.

shoyer:
Did the changes to `dataset.py` really become entirely irrelevant with merging master?

jhamman:
@shoyer - yes, you fixed the existing bug in #2696 😄

jhamman:
@shoyer - good to merge?

shoyer:
that is a little disconcerting, but yes let's merge this

xarray/core/dataset.py
```diff
@@ -728,20 +723,26 @@ def _replace_vars_and_dims(self, variables, coord_names=None, dims=None,
...
+            if dims is None:
+                dims = self._dims.copy()
+            if attrs is self.__default:
+                attrs = copy.copy(self._attrs)
+            if indexes is self.__default:
+                indexes = copy.copy(self._indexes)
+            if encoding is self.__default:
+                encoding = copy.copy(self._encoding)
+            obj = self._construct_direct(
+                variables, coord_names, dims, attrs, indexes, encoding)
```
jhamman: @shoyer - just to close the loop on this. Your indexes refactor added the `encoding` argument here. You'll see in this merge diff that I had also added this argument in the same place.

PR #2746, opened 2019-02-05
some docs updates

### LINKED ISSUE #461: CF compliant variables.
Create a new container in variables for CF convention variables. E.g.,

```
ds.salt.cf.x
ds.salt.cf.lon  # (maybe also?)
```

which would pass through ds.salt.x_rho (the x coordinate), or ds.salt.lon_rho.

### LINKED ISSUE #2289: DataArray.to_csv()
I'm using xarray to aggregate 38 GB worth of NetCDF data into a bunch of CSV reports.
I have two problems:

1. The reports are 500,000 rows by 2,000 columns. Before somebody says "if you're using CSV for this size of data you're doing it wrong" - yes, I know, but it was the only way to make the data accessible to a bunch of people that only know how to use Excel and VBA. :tired_face:
The sheer size of the reports means that (1) it's unsavory to keep the whole thing in RAM (2) pandas to_csv will take ages to complete (as it's single-threaded). The slowness is compounded by the fact that I have to compress everything with gzip.
2. I have to produce up to 40 reports from the exact same NetCDF files. I use dask to perform the computation, and different reports share a large amount of intermediate graph nodes. So I need to do everything in a single invocation to ``dask.compute()`` to allow the dask scheduler to de-duplicate the nodes.

To solve both problems, I wrote a new function:
http://xarray-extras.readthedocs.io/en/latest/api/csv.html

And now my high level wrapper code looks like this:
```
# DataSet from 200 .nc files, with a total of 500000 points on the 'row' dimension
nc = xarray.open_mfdataset('inputs.*.nc')
reports = [
    # DataArrays with shape (500000, 2000), with the rows split in 200 chunks
    gen_report0(nc),
    gen_report1(nc),
    ....
    gen_report39(nc),
]
futures = [
    # dask.delayed objects
    to_csv(reports[0], 'report0.csv.gz', compression='gzip'),
    to_csv(reports[1], 'report1.csv.gz', compression='gzip'),
    ....
    to_csv(reports[39], 'report39.csv.gz', compression='gzip'),
]
dask.compute(*futures)
```
The function is currently production quality in xarray-extras, but it would be very easy to refactor it as a method of xarray.DataArray in the main library.

Opinions?

### LINKED ISSUE #2345: Be able to define whether file(s) on disk are opened read-only or read-write
From a testing point of view, it would be helpful to explicitly specify whether to open the source file(s) on disk as read-only or read-write in open_dataset() and open_mfdataset(). It'd also be a nice backstop for anybody concerned about data corruption from unexpected behaviour.

### LINKED ISSUE #2576: When is transpose not lazy?

```python
import xarray as xr
import dask.array as da 
from dask.diagnostics import ProgressBar
a = xr.DataArray(da.zeros((3, 5, 5), chunks=(1, 2, 2)), dims=('bands', 'y', 'x'))

with ProgressBar():
    b = a.transpose('y', 'x', 'bands')
# dask does not show a progress bar due to no computation

with ProgressBar():
    b = a.transpose('y', 'x', 'bands')
    b.compute()
# dask computes the array (since we told it to) and we see a progress bar

```

#### Question

The [documentation for transpose](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.transpose.html) says that it is not lazy. Is this only in certain situations? By not lazy does it mean that when the data is computed that the transpose task will require all data to be loaded at once (one large chunk) or does it mean that the transpose operation will immediately compute the transposed array? My test above does not seem to compute the data when `transpose` is called.

Or is the documentation just outdated?

dcherian:
- [x] Closes #2289 , closes #461, closes #2576, closes #2345


doc/data-structures.rst
```diff
@@ -549,6 +549,14 @@ binary operations that act on xarray objects. In the future, we hope to write
 more helper functions so that you can easily make your functions act like
 xarray's built-in arithmetic.
 
+CF-compliant coordinate variables
```
shoyer: I'm definitely a fan of calling out other packages that add useful functionality on top of xarray, but I'm not sure we want this in the doc page on generic data structures. Maybe a separate page for "Weather and Climate Add-ons"?

dcherian:
@shoyer I've added a new page for Weather and Climate data. Let me know if this looks good.

doc/weather-climate.rst
```diff
@@ -0,0 +1,159 @@
...
+
+.. _metpy_accessor:
+
+CF-compliant coordinate variables
+---------------------------------
+
+`MetPy`_ adds a	``metpy`` accessor that allows accessing coordinates with appropriate CF metadata using generic names ``x``, ``y``, ``vertical`` and ``time``. See `their documentation`_ for more information.
+
+.. _`MetPy`: https://unidata.github.io/MetPy/dev/index.html
+.. _`their documentation`:	https://unidata.github.io/MetPy/dev/tutorials/xarray_tutorial.html#coordinates
```
jhamman: ping @dopplershift for a quick peak at this section.

dopplershift: That seems great to me. Thanks for adding. Would it also be appropriate to mention support for converting CF projection information to Cartopy projections?

dcherian: Yes! Can you send me some text or a "suggested change" then?

doc/weather-climate.rst
```diff
@@ -0,0 +1,159 @@
...
+``xarray`` can leverage metadata that follows the `Climate and Forecast (CF) conventions`_ if present. Examples include automatic labelling of plots with descriptive names and units if proper metadata is present (see :ref:`plotting`) and support for non-standard calendars used in climate science through the ``cftime`` module (see :ref:`CFTimeIndex`). There are also a number of geosciences-focused projects that build on xarray (see :ref:`related-projects`).
+
+.. _Climate and Forecast (CF) conventions: http://cfconventions.org
+
+.. _metpy_accessor:
+
+CF-compliant coordinate variables
+---------------------------------
+
+`MetPy`_ adds a	``metpy`` accessor that allows accessing coordinates with appropriate CF metadata using generic names ``x``, ``y``, ``vertical`` and ``time``. See `their documentation`_ for more information.
```
dopplershift:
```suggestion
`MetPy`_ adds a	``metpy`` accessor that allows accessing coordinates with appropriate CF metadata using generic names ``x``, ``y``, ``vertical`` and ``time``. There is also a `cartopy_crs` attribute that provides projection information, parsed from the appropriate CF metadata, as a `Cartopy`_ projection object. See `their documentation`_ for more information.
```

doc/weather-climate.rst
dopplershift:
```suggestion
.. _`their documentation`:	https://unidata.github.io/MetPy/dev/tutorials/xarray_tutorial.html#coordinates
.. _`Cartopy`: https://scitools.org.uk/cartopy/docs/latest/crs/projections.html
```

dopplershift: Done.

dcherian: Thanks!

Issue #2748, opened 2019-02-06
Inconsistent name behavior in masking


yohai:
#### Code Sample, a copy-pastable example if possible
Consider the following 4 ways to mask an `DataArray`:

```python
import xarray as xr

ds=xr.Dataset({'a': (['x'], range(10))},
              coords={'x': range(10)})
ds['msk']=ds.a>5
msk=ds.a>5

v1=ds.a.where(ds.a>5)
v2=ds.a.where(msk)
v3=ds.where(ds.msk).a
v4=ds.a.where(ds.msk)

', '.join(str(v.name) for v in (v1, v2, v3, v4))```
 #output:
'a, a, a, None'
```
#### Problem description

All `v`s defined above are identical, except that `v4` has lost its name in masking. This can cause errors in merging (like just happened to me and was hard and annoying to catch).

Is this behavior intentional? It seems inconsistent to me that `ds.a.where(ds.msk)` is not equivalent to `ds.where(ds.msk).a`

#### Expected Output
I would expect `ds.a.where(ds.msk)`  to return a `DataArray` with the name `a`.

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.2 (default, Dec 29 2018, 00:00:04) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
python-bits: 64
OS: Darwin
OS-release: 18.2.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: None
libnetcdf: None

xarray: 0.11.1+49.g27cf53f2
pandas: 0.23.3
numpy: 1.16.0
scipy: 1.2.0
netCDF4: None
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: 2.2.0
cftime: 1.0.3.4
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
cyordereddict: None
dask: 1.1.0
distributed: 1.25.3
matplotlib: 3.0.2
cartopy: 0.17.0
seaborn: 0.9.0
setuptools: 40.8.0
pip: 19.0.1
conda: None
pytest: None
IPython: 7.2.0
sphinx: 1.8.2
</details>

dcherian:
Duplicate of #2457? PRs welcome :)

yohai:
[...]

PR #2749, opened 2019-02-07
Fix name loss when masking

### LINKED ISSUE #2457: name of array disappear when apply "where" filter with Datarray but not with np.array
I was trying to select by non dimensional coordinate, and I found that "where" works well, but it erase the name of my datarray only if filtering is done with another datarray as a non dimensional coordinate
```
import xarray as xr

precip = 10 * np.random.rand(2, 2, 3)

ds = xr.Dataset({'temperature': (['x', 'y', 'time'],  temp)},
                         'lat': (['x', 'y'], lat),
                         'precipitation': (['x', 'y', 'time'], precip),
                         'time': pd.date_range('2014-09-06', periods=3),
                         'reference_time': pd.Timestamp('2014-09-05')})

ds["temperature"].where(ds.precipitation>5).sum(axis=2).plot()
```
this line would produce a label near color bar
but only using a numpy array this happens as here
```
ds["temperature"].where(ds.precipitation.values>5).sum(axis=2).plot()
```

#### Expected Output
I think that selecting with a datarray should  not change the label of the data.
Further I was expecting that as for  "sel", "where" should change the title of the graph
so that I know from the title that I performed a selection with where with a non dimensional  coordinate

yohai:
- [x] Closes #2457 #2748
 - [x] Tests added

I took a simple-minded strategy to fix: just mask and then rename the result to `self.name`. This is a slight overkill since the renaming will be done on every masking, not just on the edge cases when it's needed, but this is really a very small computational cost and it's much easier and bulletproof.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,7 +24,7 @@ Breaking changes
 - Remove support for Python 2. This is the first version of xarray that is
   Python 3 only. (:issue:`1876`).
   By `Joe Hamman <https://github.com/jhamman>`_.
-- The `compat` argument to `Dataset` and the `encoding` argument to 
+- The `compat` argument to `Dataset` and the `encoding` argument to
   `DataArray` are deprecated and will be removed in a future release.
   (:issue:`1188`)
   By `Maximilian Roos <https://github.com/max-sixty>`_.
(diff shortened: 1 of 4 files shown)


shoyer:
This looks fine, though I'd rather fix this at a lower level, if possible.

This line in `computation.py` sets the name on result DataArray objects:
https://github.com/pydata/xarray/blob/0dfc0e63650deb3da1fb1cba48cda88a704c2d60/xarray/core/computation.py#L210

If `keep_attrs=True` (you'll need to thread the argument into `apply_dataarray_ufunc` from `apply_ufunc`), then we should just copy the name from the first argument instead, e.g., `name = args[0].name if keep_attrs else result_name(args)`.

yohai:
I agree that my solution is a bit hacky, but from a computational cost viewpoint they are identical (or almost. I don't think there's a way to avoid one if statement per call of `where`). I thought my solution is good because catching this at a high level captures all possible edge cases that we might miss at lower ones. But I can try to do what you suggested to.

yohai:
@shoyer your fix works. I think it's ready to merge.

shoyer:
thanks Yohai!

PR #2751, opened 2019-02-07
Ability to force float64 instead of float32 issue #2304

### LINKED ISSUE #2304: float32 instead of float64 when decoding int16 with scale_factor netcdf var  using xarray
#### Code Sample :
Considering a netcdf file file with the following variable:
```
short agc_40hz(time, meas_ind) ;
        agc_40hz:_FillValue = 32767s ;
        agc_40hz:units = "dB" ;
        agc_40hz:scale_factor = 0.01 ;
```

Code: 

```python
from netCDF4 import Dataset
import xarray as xr

d = Dataset("test.nc")
a = d.variables['agc_40hz'][:].flatten()[69] ## 21.940000000000001 'numpy.float64'
x = xr.open_dataset("test.nc")
b = x['agc_40hz'].values.flatten()[69] ## 21.939998626708984 'numpy.float32'
abs(a - b) # 0.000001373291017
```
## Different behaviour of xarray comparing to netCDF4 Dataset
When reading the dataset with xarray we found that the decoded type was numpy.float32 instead of numpy.float64
This netcdf variable has an int16 dtype
when the variable is read with the netCDF4 library directly, it is automatically converted to numpy.float64.
in our case we loose on precision when using xarray.
We found two solutions for this:

## First solution :

This solution aims to prevent auto_maskandscale

```python
d = Dataset("test.nc")
a = d.variables['agc_40hz'][:].flatten()[69] ## 21.940000000000001 'numpy.float64'
x = xr.open_dataset("test.nc", mask_and_scale=False, decode_times=False)
b = x['agc_40hz'].values.flatten()[69] ## 21.940000000000001 'numpy.float64'
abs(a - b) # 0.000000000000000
```

Modification in xarray/backends/netCDF4_.py line 241

```python
def _disable_auto_decode_variable(var):
    """Disable automatic decoding on a netCDF4.Variable.

    We handle these types of decoding ourselves.
    """
    pass
    # var.set_auto_maskandscale(False)
    # # only added in netCDF4-python v1.2.8
    # with suppress(AttributeError):
    #     var.set_auto_chartostring(False)
```

## Second solution :

This solution uses numpy.float64 whatever integer type provided.

```python
d = Dataset("test.nc")
a = d.variables['agc_40hz'][:].flatten()[69] ## 21.940000000000001 'numpy.float64'
x = xr.open_dataset("test.nc")
b = x['agc_40hz'].values.flatten()[69] ## 21.940000000000001 'numpy.float64'
abs(a - b) # 0.000000000000000
```

Modification in xarray/core/dtypes.py line 85

```python
def maybe_promote(dtype):
    """Simpler equivalent of pandas.core.common._maybe_promote

    Parameters
    ----------
    dtype : np.dtype

    Returns
    -------
    dtype : Promoted dtype that can hold missing values.
    fill_value : Valid missing value for the promoted dtype.
    """
    # N.B. these casting rules should match pandas
    if np.issubdtype(dtype, np.floating):
        fill_value = np.nan
    elif np.issubdtype(dtype, np.integer):
        #########################
        #OLD CODE BEGIN
        #########################
        # if dtype.itemsize <= 2:
        #     dtype = np.float32
        # else:
        #     dtype = np.float64
        #########################
        #OLD CODE END
        #########################

        #########################
        #NEW CODE BEGIN
        #########################
        dtype = np.float64 # whether it's int16 or int32 we use float64
        #########################
        #NEW CODE END
        #########################
        fill_value = np.nan
    elif np.issubdtype(dtype, np.complexfloating):
        fill_value = np.nan + np.nan * 1j
    elif np.issubdtype(dtype, np.datetime64):
        fill_value = np.datetime64('NaT')
    elif np.issubdtype(dtype, np.timedelta64):
        fill_value = np.timedelta64('NaT')
    else:
        dtype = object
        fill_value = np.nan
    return np.dtype(dtype), fill_value
```

Solution number 2 would be great for us.
At this point we don't know if this modification would introduce some side effects.
Is there another way to avoid this problem ?

#### Expected Output
In our case we expect the variable to be in numpy.float64 as it is done by netCDF4.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Linux
OS-release: 4.15.0-23-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.8
pandas: 0.23.3
numpy: 1.15.0rc2
scipy: 1.1.0
netCDF4: 1.4.0
h5netcdf: None
h5py: None
Nio: None
zarr: None
bottleneck: None
cyordereddict: None
dask: 0.18.1
distributed: 1.22.0
matplotlib: 2.2.2
cartopy: None
seaborn: None
setuptools: 40.0.0
pip: 10.0.1
conda: None
pytest: 3.6.3
IPython: 6.4.0
sphinx: None
</details>

daoudjahdou:
Ability to promote to float64 instead of float32 when dealing with int variables with scale_factor.
added parameter force_promote_float64 (False by default) to open_dataset and open_dataarray that enables this behaviour when True.

 - [x] Closes #2304 
 - [x] Tests added

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -1156,6 +1156,24 @@ class NetCDF4Base(CFEncodedBase):
                 expected = create_masked_and_scaled_data()
                 assert_identical(expected, ds)
 
+    def test_force_promote_float64_variable(self):
+        with create_tmp_file() as tmp_file:
+            with nc4.Dataset(tmp_file, mode='w') as nc:
+                nc.createDimension('t', 5)
+                nc.createVariable('x', 'int16', ('t',), fill_value=-1)
+                v = nc.variables['x']
+                v.scale_factor = 0.01
+                v[:] = np.array([-1123, -1123, 123, 1123, 2123])
+            # We read the newly created netcdf file
+            with nc4.Dataset(tmp_file, mode='r') as nc:
+                # we open the dataset with forced promotion to 64 bit
+                with open_dataset(tmp_file, force_promote_float64=True) as ds:
+                    # Both dataset values should be equal
+                    # And both of float64 array type
+                    dsv = ds['x'].values
+                    ncv = nc.variables['x'][:]
+                    np.testing.assert_array_almost_equal(dsv, ncv, 15)
+
     def test_0dimensional_variable(self):
         # This fix verifies our work-around to this netCDF4-python bug:
         # https://github.com/Unidata/netcdf4-python/pull/220
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,7 +24,7 @@ Breaking changes
 - Remove support for Python 2. This is the first version of xarray that is
   Python 3 only. (:issue:`1876`).
   By `Joe Hamman <https://github.com/jhamman>`_.
-- The `compat` argument to `Dataset` and the `encoding` argument to 
+- The `compat` argument to `Dataset` and the `encoding` argument to
   `DataArray` are deprecated and will be removed in a future release.
   (:issue:`1188`)
   By `Maximilian Roos <https://github.com/max-sixty>`_.
@@ -62,6 +62,10 @@ Enhancements
 - :py:meth:`pandas.Series.dropna` is now supported for a
   :py:class:`pandas.Series` indexed by a :py:class:`~xarray.CFTimeIndex`
   (:issue:`2688`). By `Spencer Clark <https://github.com/spencerkclark>`_.
+- Ability to promote to float64 instead of float32 when dealing int type
+  variables with scale_factor added parameter force_promote_float64
+  to :py:meth:`~xarray.open_dataset` and :py:meth:`~xarray.open_dataarray`
+  that enables this behaviour. By `Daoud Jahdou <https://github.com/daoudjahdou>`_.
 
 Bug fixes
 ~~~~~~~~~
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -62,8 +62,9 @@ Enhancements
 - :py:meth:`pandas.Series.dropna` is now supported for a
   :py:class:`pandas.Series` indexed by a :py:class:`~xarray.CFTimeIndex`
   (:issue:`2688`). By `Spencer Clark <https://github.com/spencerkclark>`_.
-- Ability to promote to float64 instead of float32 when dealing int type
-  variables with scale_factor added parameter force_promote_float64
+- Ability to promote to float64 instead of float32 when dealing with int type
+  variables with scale_factor.
+  Added parameter force_promote_float64 (False by default)
   to :py:meth:`~xarray.open_dataset` and :py:meth:`~xarray.open_dataarray`
   that enables this behaviour. By `Daoud Jahdou <https://github.com/daoudjahdou>`_.
 
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -14,6 +14,7 @@ from ..core.combine import (
 from ..core.utils import close_on_error, is_grib_path, is_remote_uri
 from .common import ArrayWriter
 from .locks import _get_scheduler
+from ..core import dtypes
 
 DATAARRAY_NAME = '__xarray_dataarray_name__'
 DATAARRAY_VARIABLE = '__xarray_dataarray_variable__'
@@ -161,6 +162,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=None, decode_times=True, autoclose=None,
                  concat_characters=True, decode_coords=True, engine=None,
                  chunks=None, lock=None, cache=None, drop_variables=None,
+                 force_promote_float64=False,
                  backend_kwargs=None):
     """Load and decode a dataset from a file or file-like object.
 
@@ -227,6 +229,9 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         A variable or list of variables to exclude from being parsed from the
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
+    force_promote_float64: bool
+        If True force int variables to be promoted to float64 instead of
+        float32
     backend_kwargs: dictionary, optional
         A dictionary of keyword arguments to pass on to the backend. This
         may be useful when backend options would improve performance or
@@ -261,6 +266,8 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
 
     if cache is None:
         cache = chunks is None
+
+    dtypes.FORCE_PROMOTE_FLOAT64 = force_promote_float64
 
     if backend_kwargs is None:
         backend_kwargs = {}
@@ -360,7 +367,7 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
                    mask_and_scale=None, decode_times=True, autoclose=None,
                    concat_characters=True, decode_coords=True, engine=None,
                    chunks=None, lock=None, cache=None, drop_variables=None,
-                   backend_kwargs=None):
+                   force_promote_float64=False, backend_kwargs=None):
     """Open an DataArray from a netCDF file containing a single data variable.
 
     This is designed to read netCDF files with only one data variable. If
@@ -424,6 +431,9 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
         A variable or list of variables to exclude from being parsed from the
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
+    force_promote_float64: bool
+        If True force int variables to be promoted to float64 instead of
+        float32
     backend_kwargs: dictionary, optional
         A dictionary of keyword arguments to pass on to the backend. This
         may be useful when backend options would improve performance or
@@ -450,6 +460,7 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
                            decode_coords=decode_coords, engine=engine,
                            chunks=chunks, lock=lock, cache=cache,
                            drop_variables=drop_variables,
+                           force_promote_float64=force_promote_float64,
                            backend_kwargs=backend_kwargs)
 
     if len(dataset.data_vars) != 1:
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -230,8 +230,8 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
     force_promote_float64: bool
-        If True force int variables to be promoted to float64 instead of
-        float32
+        If True force int variables with scale factorto be promoted to
+        float64 instead of float32
     backend_kwargs: dictionary, optional
         A dictionary of keyword arguments to pass on to the backend. This
         may be useful when backend options would improve performance or
@@ -432,8 +432,8 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
     force_promote_float64: bool
-        If True force int variables to be promoted to float64 instead of
-        float32
+        If True force int variables with scale factor to be promoted to
+        float64 instead of float32
     backend_kwargs: dictionary, optional
         A dictionary of keyword arguments to pass on to the backend. This
         may be useful when backend options would improve performance or
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -230,7 +230,7 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
         dataset. This may be useful to drop variables with problems or
         inconsistent values.
     force_promote_float64: bool
-        If True force int variables with scale factorto be promoted to
+        If True force int variables with scale factor to be promoted to
         float64 instead of float32
     backend_kwargs: dictionary, optional
         A dictionary of keyword arguments to pass on to the backend. This
--- a/xarray/core/dtypes.py
+++ b/xarray/core/dtypes.py
@@ -30,6 +30,11 @@ class AlwaysLessThan(object):
 INF = AlwaysGreaterThan()
 NINF = AlwaysLessThan()
 
+# Constant that indicates that we
+# must switch to float64 instead of
+# float32 even if dtype.itemsize <= 2
+FORCE_PROMOTE_FLOAT64 = False
+
 
 # Pairs of types that, if both found, should be promoted to object dtype
 # instead of following NumPy's own type-promotion rules. These type promotion
@@ -63,7 +68,7 @@ def maybe_promote(dtype):
         # Check np.timedelta64 before np.integer
         fill_value = np.timedelta64('NaT')
     elif np.issubdtype(dtype, np.integer):
-        if dtype.itemsize <= 2:
+        if dtype.itemsize <= 2 and not FORCE_PROMOTE_FLOAT64:
             dtype = np.float32
         else:
             dtype = np.float64
--- a/xarray/core/dtypes.py
+++ b/xarray/core/dtypes.py
@@ -30,8 +30,8 @@ class AlwaysLessThan(object):
 INF = AlwaysGreaterThan()
 NINF = AlwaysLessThan()
 
-# Constant that indicates that we
-# must switch to float64 instead of
+# Constant that indicates if we
+# want to promote to float64 instead of
 # float32 even if dtype.itemsize <= 2
 FORCE_PROMOTE_FLOAT64 = False
 


daoudjahdou:
@shoyer did you have a look at this ?

shoyer:
@daoudjahdou sorry I missed your comment over in #2304. Threading the new parameter down though other functions is definitely the preferred approach here -- mutating global variables makes it much harder to follow control flow. Though I would probably stop short of modifying `maybe_promote` and would put this logic inside `CFMaskCoder` and/or `CFScaleOffsetCoder` in xarray.coding.variables.

daoudjahdou:
@shoyer now the logic is now propagated down to _choose_float_dtype inside CFScaleOffsetCoder, please let me know what you think.

shoyer:
@daoudjahdou what do you think of @magau's proposed solution over in https://github.com/pydata/xarray/issues/2304#issuecomment-462592638? Checking the dtypes of of `add_offset` and `scale_factor` might be a better alternative than requiring setting a custom option.

daoudjahdou:
@shoyer yes sure i'll update the pull request with the mentioned modifications.

daoudjahdou:
@shoyer now scale_factor and add_offset are taken into account when encoding and decoding data.
if none of them is present or if the they're not a subtype of np.generic the old behaviour takes place.

xarray/coding/variables.py
```diff
@@ -219,7 +247,9 @@ def encode(self, variable, name=None):
         dims, data, attrs, encoding = unpack_for_encoding(variable)
 
         if 'scale_factor' in encoding or 'add_offset' in encoding:
-            dtype = _choose_float_dtype(data.dtype, 'add_offset' in encoding)
+            scale_factor = pop_to(attrs, encoding, 'scale_factor', name=name)
+            add_offset = pop_to(attrs, encoding, 'add_offset', name=name)
+            dtype = _choose_float_dtype(data.dtype, scale_factor, add_offset)
             data = data.astype(dtype=dtype, copy=True)
             if 'add_offset' in encoding:
```
shoyer: This line (and the line below for `scale_factor`) needs to be updated to use the `add_offset` variable if it isn't `None` rather than checking for it in `encoding` again. (`pop_to` removes an item from the dict in which it's found.)

xarray/coding/variables.py
```diff
@@ -189,8 +189,36 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+    # Detail:
+    # If the scale_factor and add_offset attributes are of the same
+    # data type as the
+    # associated variable, the unpacked data is assumed to be of the same
+    # data type as the packed data. However, if the scale_factor
+    # and add_offset attributes are of a different data type
+    # from the variable (containing the packed data)
+    # then the unpacked data should match
+    # the type of these attributes, which must both be of type float or both
+    # be of type double. An additional restriction in this case is that
```
shoyer: What should we do if `add_offset`/`scale_factor` are *not* floats?

Right now in your implementation, if `scale_factor` is mistakenly `np.int64(10)` then the data would get decoded as int64, which seems like a bad idea (e.g., if the data is some float type). I think we would should probably ignore the type of `scale_factor`/`add_offset` in these cases instead.

xarray/coding/variables.py
```diff
@@ -189,8 +189,36 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+    # the variable containing the packed data must
+    # be of type byte, short or int.
+    # It is not advised to unpack an int into a float as there
+    # is a potential precision loss.
+
+    # We first return scale_factor type
+    # as multiplying takes priority over
+    # adding values
+    if dtype is not type(scale_factor) and \
```
shoyer: Do we actually need to check whether the type of `scale_factor` matches `dtype`? I think the logic works the same either way.

xarray/coding/variables.py
```diff
@@ -189,8 +189,36 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+            isinstance(scale_factor, np.generic):
```
shoyer: nit: per PEP8, please prefer using parentheses `()` to split across multiple lines instead of `\`

xarray/tests/test_backends.py
```diff
@@ -1156,6 +1157,25 @@ def test_mask_and_scale(self):
                 expected = create_masked_and_scaled_data()
                 assert_identical(expected, ds)
 
+    def test_mask_and_scale_with_float64_scale_factor(self):
```
shoyer: It would be nice to test all of the edge cases just using `CFScaleOffsetCoder`, e.g.,
- what happens if `add_offset` and `scale_factor` have different types?
- what if they're integers instead of floats?

See the tests in `tests/test_coding.py` for examples.

xarray/coding/variables.py
```diff
@@ -189,8 +189,36 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+    # the type of these attributes, which must both be of type float or both
+    # be of type double. An additional restriction in this case is that
+    # the variable containing the packed data must
+    # be of type byte, short or int.
+    # It is not advised to unpack an int into a float as there
+    # is a potential precision loss.
+
+    # We first return scale_factor type
+    # as multiplying takes priority over
+    # adding values
```
shoyer: I'm a little nervous about making this assumption. The NumPy casting rules for `a * x + b` would give the result the largest dtype of *any* of `a`, `x` and `b`. Maybe we should similarly pick whichever of `scale_factor`, `data` and `add_offset` has the largest float dtype?

xarray/coding/variables.py
magau: Hi @shoyer! Sorry for my delayed intervenience... First of all I'd like to thank and congratulate you for your efficiency and say that I'm very glad to see that you've taken into account my considerations about this topic.
About the comment above I'd say that, since the _CFScaleOffsetCoder_ will always multiply the raw data by the scale_factor, in case the scale_factor is an integer the decoded values will always result into integers too. Unless the raw data dtype is float, right?
Even in these cases I'd return the decoded values with the same dtype as the scale_factor, which is the expected behavior. (data providers must be aware... and you can always call the _decode_cf_variable_ function later (after using the _values.astype_('whatever') or changing the encoding['scale_factor'] dtype...).

By the way, I've also noted that the _CFScaleOffsetCoder.encode_ changes the _values.dtype_ before applying the inverse operation (i.e the dividing by the scale_factor), this will result in truncated values. I've already faced this situation, with my own netcdf encoding processes, and would advice you to only change the dtype after applying the transformation and use the numpy.round function (for integers encoding) to avoid the truncation artifacts.

Once again, sorry for my extensive comments and not opening an issue, as I should, but my hope is that you evaluate my considerations and take your own decisions with your colleagues, and keep the nice work you all have being done.  👍 

Cheers!

daoudjahdou:
@shoyer i changed the implementation, and took into consideration your comments.
now using largest type takes place only when decoding.
i added a test with all types of scale factor, add_offset and variable.

daoudjahdou:
@shoyer did you have a look at this ?

xarray/tests/test_coding.py
```diff
@@ -50,3 +50,42 @@ def test_scaling_converts_to_float32(dtype):
...
+@pytest.mark.parametrize('add_offset', [10, 0.01,
+                                        np.float16(10),
+                                        np.float32(10),
+                                        np.float64(10),
+                                        np.int8(10),
+                                        np.int16(10), np.int32(10),
+                                        np.int64(10), np.uint8(10),
+                                        np.uint16(10), np.uint32(10),
+                                        np.uint64(10), np.uint64(10)])
+def test_scaling_according_to_cf_convention(dtype, scale_factor, add_offset):
```
shoyer: This is a great set of invariant checks! But it would still be nice to see a more specific list of unit tests verifying that the decoded dtype is exactly as expected for a set of known inputs, e.g.,
```python
@pytest.mark.parametrize(
    ('input_dtype', 'scale_factor', 'add_offset', 'expected_dtype'),
    [
        (np.float32, 1, 0, np.float32),
        (np.float32, 1, 10, np.float64),
        ...
    ]
)
def test_cfscaleoffset_decoding_dtype(input_dtype, scale_factor, add_offset, expected_dtype):
    original = xr.Variable(...)
    decoded = variables.CFScaleOffsetCoder().decode(original)
    assert decoded.dtype == expected_dtype
```
This would us confidence that it's handling all the edge cases properly, e.g., only giving a bigger dtype when truly warranted.

xarray/coding/variables.py
```diff
@@ -189,8 +189,37 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+    # be of type byte, short or int.
+    # It is not advised to unpack an int into a float as there
+    # is a potential precision loss.
+
+    if mode is 'decoding':
+        types = [np.dtype(type(scale_factor)),
+                 np.dtype(type(add_offset)),
+                 np.dtype(dtype)]
+        # scaled_type should be the largest type we find
+        scaled_dtype = max(types)
```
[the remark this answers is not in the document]
daoudjahdou: @shoyer i've tested it with all numpy dtypes pair combinations and it turns out that it always gives the largest, i can for sure modify it

xarray/tests/test_backends.py
```diff
@@ -685,7 +685,9 @@ def test_roundtrip_mask_and_scale(self, decoded_fn, encoded_fn):
         with self.roundtrip(decoded) as actual:
             for k in decoded.variables:
                 assert (decoded.variables[k].dtype
-                        == actual.variables[k].dtype)
+                        == actual.variables[k].dtype or
+                        (decoded.variables[k].dtype == np.float32 and
```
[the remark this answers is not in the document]
daoudjahdou: Here is my understanding of what xarray does :
In all cases encoding returns either np.float32 or np.float64
Encoding only cares about existence of add_offset when the variable is an integer
If the variable is a float then the encoded dtype is np.float32
If the variable is an integer with add_offset then the encoded dtype np.float64
If the variable is an integer with no add_affset the encoded dtype np.float32
In all other cases the encoded dtype is np.float64

Here is my understanding of what cf-conventions imply:
decoding is the equivalent of unpacking mentioned in the cf-convention
in all cases decoding takes the encoded dtype which is either np.float32 or np.float64
then decoded type is the largest between scale_factor, add_offset and encoded type and not original variable

That being said in that test there are cases where
original type, scale_factor, add_offset are on the following configuration:
np.float32, np.int64, np.float32
then the encoded type is np.float32 as the variable is np.float32
when decoding the largest (dtypes.result_type()) between the three (encoded, sf, ao) is np.float64

and that's why "decoded" (orignal) in the test is np.float32 and actual (roundtripped) is np.float64

In my next commit i took into consideration your remark and i wrote every possible case and what to expect as an encoded or decoded dtype.

daoudjahdou:
@shoyer tests are failing but it does'nt seem to be coming from this PR, i saw the same error on other PRs as well, my tests were working fine until i made a git pull.

shoyer:
@daoudjahdou thanks for sticking with this.

I'm concerned about how lose the "roundtrip" invariant in some edge cases, where the user did not specifically indicate the desired dtype but rather used a Python type. Right now xarray seems to use inconsistent rules for casting attribute types, based on the netCDF backend:
```
In [8]: for engine in ['scipy', 'netcdf4', 'h5netcdf']:
   ...:     ds = xarray.Dataset({'x': np.float32(0)})
   ...:     encoding = {'x': {'scale_factor': 1, 'add_offset': 1}}
   ...:     ds.to_netcdf(f'test-{engine}.nc', engine=engine, encoding=encoding)
   ...:

In [9]: ! ncdump -h test-scipy.nc
netcdf test-scipy {
variables:
	float x ;
		x:add_offset = 1 ;
		x:scale_factor = 1 ;
		x:_FillValue = NaNf ;
}

In [10]: ! ncdump -h test-netcdf4.nc
netcdf test-netcdf4 {
variables:
	float x ;
		x:_FillValue = NaNf ;
		x:add_offset = 1LL ;
		x:scale_factor = 1LL ;
}

In [11]: ! ncdump -h test-h5netcdf.nc
netcdf test-h5netcdf {
variables:
	float x ;
		x:_FillValue = NaNf ;
		x:add_offset = 1LL ;
		x:scale_factor = 1LL ;
}
```

At least for `add_offset` and `scale_factor`, I think we should be more careful when determining the attribute dtype. Python's `int` and `float` don't have any intrinsic dtype. If the attributes are specified as native Python types (not NumPy types) and can be safely cast into the dtype of the data, we should probably use that instead.

daoudjahdou:
@shoyer do you mean that we consider that by default when we deal with Python's `int` and `float` we cast them to `np.int64` and `np.float64` ?

shoyer:
> @shoyer do you mean that we consider that by default when we deal with Python's `int` and `float` we cast them to `np.int64` and `np.float64` ?

Currently we cast to `int32`/`float64` for netCDF3 files (which don't support `int64`) and `int64`/`float64` for netCDF4 files.

That default behavior doesn't really make sense if the user provided builtin Python numbers for these attributes.

For example, if I write something like the following, the data for `x` should probably decode to float32:
```python
ds = xarray.Dataset({'x': np.float32(0)})
encoding = {'x': {'scale_factor': 1, 'add_offset': 1, 'dtype': 'int8'}}
ds.to_netcdf('test.nc', encoding=encoding)
```
With the current implementation in this PR, I think the data will come back as float64.

To get something that will decode as float64 without the original data being float64, the user should need to intentionally choose dtypes for `scale_factor` and `add_offsert`, e.g.,
```python
ds = xarray.Dataset({'x': np.float32(0)})
encoding = {'x': {'scale_factor': np.float64(1), 'add_offset': np.float64(1), 'dtype': 'int8'}}
ds.to_netcdf('test.nc', encoding=encoding)
```

Does this make sense to you?  From my perspective, two goals for this change should be:
1. Adhere to CF conventions for deciding how to decode data from disk.
2. Provide a consistent interface for xarray users, so that by default data is restored in the same form that it was written.

daoudjahdou:
@shoyer sorry for the delayed response.
`dtypes.result_type(1, np.float32(1))` returns `dtype('float64')`
That's what makes this behaviour for Python's `int` and `float`
Keeping the consistency would then require testing if `scale_factor*var_dtype + add_offset` fit in `var_dtype` in case of Python's `int` and `float`. 
Correct me if i'm wrong but this is a bit hard to do without knowing max,min values to avoid overlapping
Evaluating those could break performance if only used for encoding and decoding. 
Do have any idea how this could be acheived ? or is it simpler to keep it as it is ?

shoyer:
> Correct me if i'm wrong but this is a bit hard to do without knowing max,min values to avoid overlapping
> Evaluating those could break performance if only used for encoding and decoding.

Right, we definitely don't want to inspect the data to verify if it fits.

I think we will will need explicitly checks for `isinstance(x, (int, float))` (for `scale_factor` and `add_offset`) to identify these cases in the `encode()` method. Let me try to come up with a concrete suggestion...

xarray/coding/variables.py
```diff
@@ -217,9 +258,9 @@ class CFScaleOffsetCoder(VariableCoder):
 
     def encode(self, variable, name=None):
         dims, data, attrs, encoding = unpack_for_encoding(variable)
-
         if 'scale_factor' in encoding or 'add_offset' in encoding:
```
shoyer: Alternative logic for this clause that I think might work:
```python
if 'scale_factor' in encoding or 'add_offset' in encoding:
    scale_factor = encoding.pop('scale_factor', 1)
    add_offset = encoding.pop('add_offset, 0)
    if 'dtype' in encoding:
        dtype = pop_to(encoding, attrs, 'dtype', name=name)
    else:
        dtype = _choose_encoding_float_dtype(data.dtype, add_offset != 0)

    # save scale_factor and add_offset with dtype matching the decoded
    # data, per CF conventions
    safe_setitem(attrs, 'scale_factor', data.dtype.type(scale_factor), name)
    safe_setitem(attrs, 'add_offset', data.dtype.type(add_offset), name)

    # apply scale/offset encoding
    data = data.astype(dtype=dtype, copy=True)
    if add_offset:
        data -= add_offset
    if scale_factor != 1:
        # multiplication is faster than division
        data *= 1/scale_factor
```

The idea here is to always write the `scale_factor` and `add_offset` attributes according to CF conventions, which ensures that data always gets read out the right way. This does remove some user flexibility, but xarray is OK with being opinionated about how it writes metadata -- it is not a general purpose tool for writing completely flexible netCDF files.

xarray/coding/variables.py
```diff
@@ -189,8 +189,49 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+
+    if scale_factor or add_offset:
+
+        types = (np.dtype(type(scale_factor)),
+                 np.dtype(type(add_offset)),
+                 np.dtype(dtype))
+
+        if add_offset is None:
+            types = (np.dtype(type(scale_factor)),
+                     np.dtype(dtype))
```
shoyer: It would be nice if these five lines creating `types` handled add_offset and scale_factor in a symmetric way, e.g.,
```python
types = [np.dtype(dtype)]
if scale_factor is not None:
    types.append(np.dtype(type(scale_offset))
if add_offset is not None:
    types.append(np.dtype(type(add_offset))
```

xarray/coding/variables.py
```diff
@@ -189,8 +189,49 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+    # from the variable (containing the packed data)
+    # then the unpacked data should match
+    # the type of these attributes, which must both be of type float or both
+    # be of type double. An additional restriction in this case is that
+    # the variable containing the packed data must
+    # be of type byte, short or int.
+    # It is not advised to unpack an int into a float as there
+    # is a potential precision loss.
+
+    if scale_factor or add_offset:
```
shoyer: It's safer to use `is not None` for checking default values:
```suggestion
    if scale_factor is not None or add_offset is not None:
```

xarray/coding/variables.py
```diff
@@ -189,8 +189,49 @@ def _scale_offset_decoding(data, scale_factor, add_offset, dtype):
...
+
+        # scaled_type should be the largest type we find
+        scaled_dtype = dtypes.result_type(*types)
+
+        # We return it only if it's a float32 or a float64
+        if (scaled_dtype.itemsize >= 4
+                and np.issubdtype(scaled_dtype, np.floating)):
+            return scaled_dtype
+
+    return _choose_encoding_float_dtype(dtype, add_offset is not None)
```
shoyer: Potentially we could use the `add_offset` value here -- 0 and None are equivalent as far as picking an automatic float dtype is concerned, indicating "no offset":
```suggestion
    return _choose_encoding_float_dtype(dtype, add_offset)
```

daoudjahdou:
@shoyer i've tested the solution provided, it works like a charm with my tests however many tests are broken on test_backends.py with cases where we lose precision i'll give you more detail.

fmaussion:
I've noted that the very recent ERA5 data is also cast to float32 at decoding, although the scale and offset params are float64. I think it would be a very valuable addition to xarray to do this properly: @daoudjahdou are you still willing to work on this PR? Otherwise I'll try to pick this up when time permits.

kmuehlbauer:
This should work as implemented in #8713 by setting `scale_factor` and/or `add_offset` to the wanted dtype.

Issue #2710, opened 2019-01-25
xarray.DataArray.expand_dims() can only expand dimension for a point coordinate 


pletchm:
#### Current `expand_dims` functionality
Apparently, `expand_dims` can only create a dimension for a point coordinate, i.e. it promotes a scalar coordinate into 1D coordinate. Here is an example:
```python
>>> coords = {"b": range(5), "c": range(3)}
>>> da = xr.DataArray(np.ones([5, 3]), coords=coords, dims=list(coords.keys()))
>>> da
<xarray.DataArray (b: 5, c: 3)>
array([[1., 1., 1.],
       [1., 1., 1.],
       [1., 1., 1.],
       [1., 1., 1.],
       [1., 1., 1.]])
Coordinates:
  * b        (b) int64 0 1 2 3 4
  * c        (c) int64 0 1 2
>>> da["a"] = 0  # create a point coordinate
>>> da
<xarray.DataArray (b: 5, c: 3)>
array([[1., 1., 1.],
       [1., 1., 1.],
       [1., 1., 1.],
       [1., 1., 1.],
       [1., 1., 1.]])
Coordinates:
  * b        (b) int64 0 1 2 3 4
  * c        (c) int64 0 1 2
    a        int64 0
>>> da.expand_dims("a")  # create a new dimension "a" for the point coordinated
<xarray.DataArray (a: 1, b: 5, c: 3)>
array([[[1., 1., 1.],
        [1., 1., 1.],
        [1., 1., 1.],
        [1., 1., 1.],
        [1., 1., 1.]]])
Coordinates:
  * b        (b) int64 0 1 2 3 4
  * c        (c) int64 0 1 2
  * a        (a) int64 0
>>>
```
#### Problem description
I want to be able to do 2 more things with `expand_dims` or maybe a related/similar method:
1) broadcast the data across 1 or more new dimensions
2) expand an existing dimension to include 1 or more new coordinates

#### Here is the code I currently use to accomplish this
```
from collections import OrderedDict

import xarray as xr


def expand_dimensions(data, fill_value=np.nan, **new_coords):
    """Expand (or add if it doesn't yet exist) the data array to fill in new
    coordinates across multiple dimensions.

    If a dimension doesn't exist in the dataarray yet, then the result will be
    `data`, broadcasted across this dimension.

    >>> da = xr.DataArray([1, 2, 3], dims="a", coords=[[0, 1, 2]])
[...]

PR #2757, opened 2019-02-08
Allow expand_dims() method to support inserting/broadcasting dimensions with size>1

pletchm:
This pull request enhances the `expand_dims` method for both `Dataset` and `DataArray` objects to support inserting/broadcasting dimensions with size > 1. It corresponds to this issue https://github.com/pydata/xarray/issues/2710.

## Changes:
1. dataset.expand_dims() method take dict like object where values represent length of dimensions or coordinates of dimesnsions
3. dataarray.expand_dims() method take dict like object where values represent length of dimensions or coordinates of dimesnsions
5. Add alternative option to passing a dict to the dim argument, which is now an optional kwarg, passing in each new dimension as its own kwarg
7. Add expand_dims enhancement from issue 2710 to whats-new.rst

## Included:
 - [ ] Tests added

## What's new:
All of the old functionality is still there, so it shouldn't break anyone's existing code that uses it. 

You can now pass a dim as a dict, where the keys are the new dimensions and the values are either integers (giving the length of the new dimensions) or iterables (giving the coordinates of the new dimensions).
```
import numpy as np
import xarray as xr

>>> original = xr.Dataset({'x': ('a', np.random.randn(3)),
                                  'y': (['b', 'a'], np.random.randn(4, 3))},
                                  coords={'a': np.linspace(0, 1, 3),
                                                'b': np.linspace(0, 1, 4),
                                                'c': np.linspace(0, 1, 5)},
                                  attrs={'key': 'entry'})
>>> original
<xarray.Dataset>
Dimensions:  (a: 3, b: 4, c: 5)
Coordinates:
  * a        (a) float64 0.0 0.5 1.0
  * b        (b) float64 0.0 0.3333 0.6667 1.0
  * c        (c) float64 0.0 0.25 0.5 0.75 1.0
Data variables:
    x        (a) float64 -1.556 0.2178 0.6319
    y        (b, a) float64 0.5273 0.6652 0.3418 1.858 ... -0.3519 0.8088 0.8753
Attributes:
    key:      entry
>>> original.expand_dims({"d": 4, "e": ["l", "m", "n"]})
<xarray.Dataset>
Dimensions:  (a: 3, b: 4, c: 5, d: 4, e: 3)
Coordinates:
  * e        (e) <U1 'l' 'm' 'n'
  * a        (a) float64 0.0 0.5 1.0
  * b        (b) float64 0.0 0.3333 0.6667 1.0
  * c        (c) float64 0.0 0.25 0.5 0.75 1.0
Dimensions without coordinates: d
Data variables:
    x        (d, e, a) float64 -1.556 0.2178 0.6319 ... -1.556 0.2178 0.6319
    y        (d, e, b, a) float64 0.5273 0.6652 0.3418 ... -0.3519 0.8088 0.8753
Attributes:
    key:      entry
```
Or, equivalently, you can pass the new dimensions as kwargs instead of a dictionary.
```
original.expand_dims(d=4, e=["l", "m", "n"])
```


xarray/core/dataset.py
```diff
@@ -2328,7 +2328,7 @@ def swap_dims(self, dims_dict, inplace=None):
         return self._replace_with_new_dims(variables, coord_names,
                                            indexes=indexes, inplace=inplace)
 
-    def expand_dims(self, dim, axis=None):
+    def expand_dims(self, dim=None, axis=None, **new_dims):
```
fujiisoup: It would be nice to follow an argument naming convention,
`expand_dims(self, dim=None, axis=None, **dim_kwargs):`

xarray/core/dataset.py
```diff
@@ -2379,6 +2394,24 @@ def expand_dims(self, dim, axis=None):
...
+            for k, v in dim.items():
+                if hasattr(v, "__iter__"):
+                    # If the value for the new dimension is an iterable, then
+                    # save the coordinates to the variables dict, and set the
+                    # value within the dim dict to the length of the iterable
+                    # for later use.
+                    variables[k] = xr.IndexVariable(k, list(v))
+                    self._coord_names.add(k)
+                    dim[k] = len(list(v))
+                elif isinstance(v, int):
```
fujiisoup: Probably better to take care of 0d-array of an integer type?

xarray/core/dataset.py
```diff
@@ -2379,6 +2394,24 @@ def expand_dims(self, dim, axis=None):
...
+        if utils.is_dict_like(dim):
+            # If dim is a dict, then ensure that the values are either integers
+            # or iterables.
+            for k, v in dim.items():
+                if hasattr(v, "__iter__"):
+                    # If the value for the new dimension is an iterable, then
+                    # save the coordinates to the variables dict, and set the
+                    # value within the dim dict to the length of the iterable
+                    # for later use.
+                    variables[k] = xr.IndexVariable(k, list(v))
```
fujiisoup: I think `variables[k] = Variable((k, ), v)` is OK for here.

xarray/core/dataset.py
```diff
@@ -2337,21 +2337,36 @@ def expand_dims(self, dim, axis=None):
...
         Returns
         -------
         expanded : same type as caller
             This object, but with an additional dimension(s).
         """
+        if dim is None:
+            dim = new_dims
+        elif dim and new_dims:
+            raise ValueError('cannot specify both keyword and positional '
+                             'arguments to expand_dims')
```
fujiisoup: Probably it would be a little cleaner if we can make `dim` always a dict
```python
if dim is not None and not utils.is_dict_like(dim):
    dim = {dim: 1}
dim = either_dict_or_kwargs(dim, dim_kwargs, 'expand_dims')
```

xarray/tests/test_dataset.py
```diff
@@ -2046,6 +2059,31 @@ def test_expand_dims(self):
         roundtripped = actual.squeeze('z')
         assert_identical(original, roundtripped)
 
+        # Test expanding one dimension to have size > 1 that doesn't have
+        # coordinates, and also expanding another dimension to have size > 1
+        # that DOES have coordinates.
+        actual = original.expand_dims({"d": 4, "e": ["l", "m", "n"]})
```
fujiisoup: Probably we may need to use an OrderedDict also here for python 3.5 compatibility

xarray/core/dataset.py
shoyer: We do currently try to use `IndexVariable` for every variable that gets an associated index. But I would definitely avoid casting to `list()`, since IndexVariable/Variable will internally convert it into a numpy.ndarray. So just `IndexVariable((k,), v)` would work.

xarray/tests/test_dataarray.py
```diff
@@ -1392,6 +1402,45 @@ def test_expand_dims_with_scalar_coordinate(self):
...
+                             attrs={'key': 'entry'}
+                             ).drop(['y', 'dim_0'])
+        assert_identical(expected, actual)
+
+        # Test with kwargs instead of passing dict to dim arg.
+        other_way = array.expand_dims(y=2, z=1, dim_1=['a', 'b', 'c'])
+        # Unfortunately, there is no way to maintain insertion order with
+        # kwargs in python 3.5 and earlier, so for now we have to ensure the
+        # dimensions of the expected result are in the same order as the actual
+        # result to allow the test to pass.
```
shoyer: As long as we have reuse the standard `either_dict_or_kwargs` utility, I think it is reasonable to simplify this test to only check inserting a single dimension in a way that is both Python 3.5 and 3.6+ compatible.

xarray/core/dataset.py
```diff
@@ -2337,21 +2337,36 @@ def expand_dims(self, dim, axis=None):
...
+            dimensions and the values are either integers (giving the length of
+            the new dimensions) or iterables (giving the coordinates of the new
+            dimensions).
         axis : integer, list (or tuple) of integers, or None
             Axis position(s) where new axis is to be inserted (position(s) on
             the result array). If a list (or tuple) of integers is passed,
             multiple axes are inserted. In this case, dim arguments should be
             the same length list. If axis=None is passed, all the axes will
             be inserted to the start of the result array.
+        **new_dims : int or  iterable
```
shoyer: Maybe `int or sequence/ndarray`?

Iterables actually are not valid inputs into `np.array()`.

xarray/tests/test_dataarray.py
```diff
@@ -1392,6 +1402,45 @@ def test_expand_dims_with_scalar_coordinate(self):
...
+                             coords=expected_coords,
+                             dims=list(expected_coords.keys()),
+                             attrs={'key': 'entry'}
+                             ).drop(['y', 'dim_0'])
+        assert_identical(expected, actual)
+
+        # Test with kwargs instead of passing dict to dim arg.
+        other_way = array.expand_dims(y=2, z=1, dim_1=['a', 'b', 'c'])
+        # Unfortunately, there is no way to maintain insertion order with
+        # kwargs in python 3.5 and earlier, so for now we have to ensure the
```
shoyer: Could we make specifying sizes for multiple dimensions with a non-ordered dictionary an error on Python 3.5? I don't want users to rely on behavior that will break in subtle ways when they rerun their code in a new Python session -- that is a very poor user experience.

pletchm: That's a good point and a good solution, thanks!

pletchm: Would if I always sort `dim`/`dim_kwargs` to ensure the order will be the same between python 3.5 and python 3.6. This would of course be clearly documented in the doc string.

So I think it would just be
```
        if isinstance(dim, int):
            raise TypeError('dim should be str or sequence of strs or dict')
        elif isinstance(dim, str):
            dim = {dim: 1}
        elif isinstance(dim, (list, tuple)):
            if len(dim) != len(set(dim)):
                raise ValueError('dims should not contain duplicate values.')
            dim = {d: 1 for d in dim}
        dim = either_dict_or_kwargs(dim, dim_kwargs, 'expand_dims')

        dim = sorted(dim)
```

pletchm: I guess the problem with the above is that will change the output for current users.

shoyer: I would rather not sort. It's more intuitive for users if dimensions appear in the provided order.

At this point Python 3.5 is the oldest Python version we support (there is also 3.6 and 3.7). I don't think we should go out of our way to make things more convenient for Python 3.5 users at the cost of inconveniencing the majority of our users who use newer Python versions. See https://pypistats.org/packages/xarray if you're curious about download numbers.

pletchm:
Hi @shoyer and @fujiisoup, I believe I've made all of the updates that you suggested in your comments. How is the PR looking now? 
Thank you!

xarray/core/dataarray.py
```diff
@@ -1147,21 +1148,54 @@ def expand_dims(self, dim, axis=None):
...
+        elif isinstance(dim, (list, tuple)):
+            if len(dim) != len(set(dim)):
+                raise ValueError('dims should not contain duplicate values.')
+            dim = OrderedDict(((d, 1) for d in dim))
+
+        # TODO: get rid of the below code block when python 3.5 is no longer
+        #   supported.
+        python36_plus = sys.version_info[0] == 3 and sys.version_info[1] > 5
+        not_ordereddict = dim is not None and not isinstance(dim, OrderedDict)
+        if not python36_plus and not_ordereddict:
```
shoyer: I would suggest a slight variation on this: on Python 3.5, you aren't allowed to use `dim_kwargs` at all.

pletchm: That sounds like a great idea. Thanks!

pletchm: @shoyer , I changed it so that `dim_kwargs` isn't allowed for Python 3.5.

pletchm:
@shoyer, `dim_kwargs` is no longer allowed to be used for python 3.5. Does anything else need to change in this PR?

xarray/core/dataset.py
```diff
@@ -2376,10 +2403,24 @@ def expand_dims(self, dim, axis=None):
...
+        # or iterables.
+        for k, v in dim.items():
+            if hasattr(v, "__iter__"):
+                # If the value for the new dimension is an iterable, then
+                # save the coordinates to the variables dict, and set the
+                # value within the dim dict to the length of the iterable
+                # for later use.
+                variables[k] = xr.IndexVariable((k,), v)
+                self._coord_names.add(k)
+                dim[k] = len(list(v))
```
shoyer: Can we use `variables[k].size` here?

`len(list(v))` will convert `v` into a Python list, which can be slow for large iterables.

pletchm: Sure thing! Changed it.

doc/whats-new.rst
```diff
@@ -113,6 +113,9 @@ Other enhancements
   By `Keisuke Fujii <https://github.com/fujiisoup>`_.
 - Added :py:meth:`~xarray.Dataset.drop_dims` (:issue:`1949`).
   By `Kevin Squire <https://github.com/kmsquire>`_.
+- Allow ``expand_dims`` method to support inserting/broadcasting dimensions
```
shoyer: sorry one other thing -- this should be moved up to 0.12.1

pletchm: No problem. Thanks

shoyer:
Thanks @pletchm!

PR #2759, opened 2019-02-11
Add use_cftime option to open_dataset

### LINKED ISSUE #1263: xarray.open_mfdataset returns inconsistent times
## Problem

I am running into inconsistent time coordinates with a long climate model experiment that exceeds the limits of `pandas.tslib.Timestamp` (covers roughly 17th to 23rd century).

Currently, `xarray.open_mfdataset` delegates decoding of the time axis to `xarray.open_dataset` which decodes either to pandas time stamps or, of this fails, to `netcdftime.datetime` objects. `xarray.open_mfdataset` later combines the single-file datasets and just concatenates all the time axes.

## Solution

1. Let auto_combine check for consistency and repair the time axis if necessary.
2. Let `xarray.open_mfdataset` prevent `xarray.open_dataset` from decoding the times for each file and only decode times after everything is combined.

The latter is equivalent to a workaround I use for the moment: Pass `decode_times=False` to `xarray.open_mfdataset` and then explicitly call `xarray.decode_cf` on the dataset.

### LINKED ISSUE #2754: silence warning for decode_cf_datetime?

```python
import xarray as xr
import numpy as np

x = np.arange(100) * 365
x = xr.coding.times.decode_cf_datetime(x, 'days since 2400-01-01', 
        calendar='proleptic_gregorian')

```
#### Problem description

xarray still throws an error when decoding out-of-bounds dates for proleptic_gregorian - should this be silenced?

<details>
NSTALLED VERSIONS
------------------
commit: None
python: 3.7.1 | packaged by conda-forge | (default, Nov 13 2018, 18:33:04) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.4.172-86-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.2

xarray: 0.11.3
pandas: 0.24.1
numpy: 1.16.1
scipy: 1.2.0
netCDF4: 1.4.2
pydap: None
h5netcdf: 0.6.2
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
PseudonetCDF: None
rasterio: 1.0.17
cfgrib: None
iris: None
bottleneck: 1.2.1
cyordereddict: None
dask: 1.1.1
distributed: 1.25.3
matplotlib: 3.0.2
cartopy: 0.17.0
seaborn: 0.9.0
setuptools: 40.7.3
pip: 19.0.1
conda: None
pytest: 4.2.0
IPython: 7.2.0
sphinx: 1.8.4
</details>

spencerkclark:
Based on @shoyer's suggestion in https://github.com/pydata/xarray/issues/2754#issuecomment-461983092.

 - [x] Closes #1263; Closes #2754
 - [x] Tests added


xarray/coding/times.py
```diff
@@ -149,41 +170,30 @@ def decode_cf_datetime(num_dates, units, calendar=None):
...
+        except (OutOfBoundsDatetime, OverflowError):
+            dates = _decode_datetime_with_cftime(
+                flat_num_dates.astype(np.float), units, calendar)
+
+            if (dates[np.nanargmin(num_dates)].year < 1678 or
+               dates[np.nanargmax(num_dates)].year >= 2262):
+                if calendar in _STANDARD_CALENDARS:
+                    warnings.warn(
+                        'Unable to decode time axis into full '
+                        'numpy.datetime64 objects, continuing using dummy '
```
mathause: maybe not call them 'dummy'?
```suggestion
                        'numpy.datetime64 objects, continuing using '
```

spencerkclark: Yes, that's a good idea.

shoyer:
Let me suggest a slight variation: let's make the behavior of `use_cftime=None` dependent on the calendar attribute, e.g.,
- `gregorian`, `standard`, `proleptic_gregorian` or not found: try to decode into numpy.datetime64 
-  everything else: only use cftime -- numpy.datetime64 is not correct!

spencerkclark:
As of #2516 this is already the default behavior :)

shoyer:
> As of #2516 this is already the default behavior :)

OK, great! Maybe we should just update the description of `use_cftime`, to clarify that it only is relevant with standard calendars?

spencerkclark:
> Maybe we should just update the description of `use_cftime`, to clarify that it only is relevant with standard calendars?

Good idea!  See my update in 52866a8.

xarray/coding/times.py
```diff
@@ -397,15 +407,17 @@ def encode(self, variable, name=None):
 
         return Variable(dims, data, attrs, encoding)
 
-    def decode(self, variable, name=None):
+    def decode(self, variable, name=None, use_cftime=None):
```
shoyer: Should we keep this option on the `__init__` method instead? That’s what we’ve done for other coder objects.

I think that is a little nicer since it makes it more obvious how to use coders.

spencerkclark: Agreed.  That is nicer.

spencerkclark:
Ok merging now!  Thanks @shoyer for your initial design suggestion.

PR #2766, opened 2019-02-13
ENH - Adding Pseudo-Inverse to computation.py

erbian:
- [x] Tests added

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -25,6 +25,7 @@ Top-level functions
    zeros_like
    ones_like
    dot
+   pinv
 
 Dataset
 =======
--- a/xarray/__init__.py
+++ b/xarray/__init__.py
@@ -7,7 +7,7 @@ del get_versions
 from .core.alignment import align, broadcast, broadcast_arrays
 from .core.common import full_like, zeros_like, ones_like
 from .core.combine import concat, auto_combine
-from .core.computation import apply_ufunc, dot, where
+from .core.computation import apply_ufunc, dot, where, pinv
 from .core.extensions import (register_dataarray_accessor,
                               register_dataset_accessor)
 from .core.variable import as_variable, Variable, IndexVariable, Coordinate
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -1114,3 +1114,38 @@ def where(cond, x, y):
                        join='exact',
                        dataset_join='exact',
                        dask='allowed')
+
+
+def pinv(array, dims=None, **kwargs):
+    """Computes vectorized Moore-Penrose pseudo-inverse
+
+    Parameters
+    ----------
+    array: DataArray
+        Matrices from which we compute generalized inverses.
+    dims: pair of strings, optional
+        (row, column) dimensions of matrix to invert.
+        If not specified, computes over last 2 dimensions
+    **kwargs: dict
+        Additional keyword arguments passed to numpy.linalg.pinv
+
+    Returns
+    -------
+    pinv: DataArray
+
+    Examples
+    --------
+
+    >>> da = xr.DataArray(np.arange(3 * 4 * 5).reshape(3, 4, 5),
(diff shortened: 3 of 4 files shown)


max-sixty:
Unfortunately this requires numpy 1.14. That's a year old - too soon to bump the minimum version?

shoyer:
It's not entirely clear to me where we should draw the line between included wrappers in xarray and leaving this to external packages, e.g., see xr-scipy.readthedocs.io

If we do include this in xarray, I would prefer to have both (1) a clear policy around what we include (e.g., only direct wrappers for numpy/scipy functions) and (2) a separate namespace for wrapped functions, e.g., `xarray.linalg` or something more generic like `xarray.wrapped`.

Note that with [NEP-18](http://www.numpy.org/neps/nep-0018-array-function-protocol.html) we'll be able to directly everything in numpy.linalg on xarray objects, without requiring the use of a separate wrapper.

cc @pydata/xarray  for discussion

max-sixty:
> Note that with [NEP-18](http://www.numpy.org/neps/nep-0018-array-function-protocol.html) we'll be able to directly everything in numpy.linalg on xarray objects, without requiring the use of a separate wrapper.

It would be awesome for those numpy functions to just work! Could we start on this now, making it conditional on numpy 1.16 installed? 

Reading the NEP, I _think_ you could literally pass `dim=['x','y']` to the numpy function, and it'll pass that onto xarray. Lmk if I'm mistaken.

Speaking for @erbian (my colleague) , no rush to merge this - we thought it would be helpful but have it internally. Even better if we can get a general solution.

shoyer:
> Reading the NEP, I _think_ you could literally pass `dim=['x','y']` to the numpy function, and it'll pass that onto xarray. Lmk if I'm mistaken.

No, that would give an error message about passing in a bad argument name. We'll need our own wrappers to switch `axis` to `dim`.

dcherian:
Since this is basically a one line call to apply_ufunc, I propose that we add it as an example  of using apply_ufunc in the docs. Maybe even emphasize by adding a section on "wrapping numpy functions"?

max-sixty:
> No, that would give an error message about passing in a bad argument name. We'll need our own wrappers to switch `axis` to `dim`.

I understand we'd need to define `__array_function__` on xarray objects managed the conversion. 

Conditional on that, could someone call `np.pinv(da, dim=['x', 'y']`? And then `DataArray.__array_function__` would receive the `dim` kwarg. 

Or would `np.pinv` raise if passed a `dim` kwarg? (if so, we'd still need to declare n wrappers, IIUC?)

shoyer:
> Or would `np.pinv` raise if passed a `dim` kwarg? (if so, we'd still need to declare n wrappers, IIUC?)

Yes, `np.pinv` will raise if passed a `dim` kwargs. This was an intentional choice: http://www.numpy.org/neps/nep-0018-array-function-protocol.html#support-for-implementation-specific-arguments

max-sixty:
> Yes, `np.pinv` will raise if passed a `dim` kwargs. This was an intentional choice: http://www.numpy.org/neps/nep-0018-array-function-protocol.html#support-for-implementation-specific-arguments

Thanks.

With the current state, am I correct in thinking we still need wrappers for any `numpy.linalg` function that requires a `dim` kwarg?
(I guess we _could_ use the `axis` kwarg instead, but that might be confusing things)

shoyer:
Yes, we'll want to add our own wrapper functions. (Or possibly a new
`apply()` method could do the dim -> axis renaming)

max-sixty:
OK, thanks. Being able to call `np.pinv(da, dim='x')` would be awesome, but ack the downsides you outlined in the PEP

dcherian:
> Since this is basically a one line call to apply_ufunc, I propose that we add it as an example of using apply_ufunc in the docs. Maybe even emphasize by adding a section on "wrapping numpy functions"?

@erbian can you change this PR to improve the documention this way?

PR #2767, opened 2019-02-13
Add `Dataset.drop_dims`

### LINKED ISSUE #1949: Removing dimensions from Dataset objects
```python
test_dataset = xr.Dataset(dict(
    empty_array=xr.DataArray([], dims='a'), 
    populated_array=xr.DataArray([1], {'b':['1']}, 'b')
))
```

I have a dataset that is produced programatically and can at times end up producing an empty DataArray. This isn't an issue, per-se, because I can remove those empty DataArrays later. However, I cannot find any way in which to remove the unused, empty dimension! I've tried deleting, dropping, resetting indeces, etc., and have had no luck in purging this empty dimension. It causes issues down the line as existence of entries in the dimensions list triggers certain events.

Is there a way to remove a dimension (and possibly then all data variables which depend on it)?

kmsquire:
- [x] Closes #1949 
 - [x] Tests added

### DIFF
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -229,8 +229,8 @@ arrays). However, you can do normal indexing with dimension names:
 Using indexing to *assign* values to a subset of dataset (e.g.,
 ``ds[dict(space=0)] = 1``) is not yet supported.
 
-Dropping labels
----------------
+Dropping labels and dimensions
+------------------------------
 
 The :py:meth:`~xarray.Dataset.drop` method returns a new object with the listed
 index labels along a dimension dropped:
@@ -240,6 +240,13 @@ index labels along a dimension dropped:
     ds.drop(['IN', 'IL'], dim='space')
 
 ``drop`` is both a ``Dataset`` and ``DataArray`` method.
+
+Use :py:meth:`~xarray.Dataset.drop_dims` to drop a full dimension from a Dataset.
+Any variables with these dimensions are also dropped:
+
+.. ipython:: python
+
+    ds.drop_dims('time')
 
 
 .. _masking with where:
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2797,6 +2797,37 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         coord_names = set(k for k in self._coord_names if k in variables)
         return self._replace_vars_and_dims(variables, coord_names)
 
+    def drop_dims(self, drop_dims):
+        """Drop dimensions and associated variables from this dataset.
+
+        Parameters
+        ----------
+        drop_dims : str or list
+            Dimension or dimensions to drop.
+
+        Returns
+        -------
+        obj : Dataset
+            The dataset without the given dimensions (or any variables
+            containing those dimensions)
+        """
+        if utils.is_scalar(drop_dims):
+            drop_dims = [drop_dims]
+
+        missing_dimensions = [d for d in drop_dims if d not in self.dims]
+        if missing_dimensions:
+            raise ValueError('Dataset does not contain the dimensions: %s'
+                             % missing_dimensions)
+
+        drop_vars = set(k for k, v in self._variables.items()
+                        for d in v.dims if d in drop_dims)
+
+        variables = OrderedDict((k, v) for k, v in self._variables.items()
+                                if k not in drop_vars)
+        coord_names = set(k for k in self._coord_names if k in variables)
+
+        return self._replace_with_new_dims(variables, coord_names)
+
     def transpose(self, *dims):
         """Return a new Dataset object with all array dimensions transposed.
 
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -87,6 +87,7 @@ Dataset contents
    Dataset.swap_dims
    Dataset.expand_dims
    Dataset.drop
+   Dataset.drop_dims
    Dataset.set_coords
    Dataset.reset_coords
 
--- a/doc/data-structures.rst
+++ b/doc/data-structures.rst
@@ -408,6 +408,13 @@ operations keep around coordinates:
     list(ds[['x']])
     list(ds.drop('temperature'))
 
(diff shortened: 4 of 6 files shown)


kmsquire:
I was looking for this functionality and ran across #1949.

There was one bit of behavior that was unclear to me.  Say I have

```python
In [6]: data = xr.Dataset({'A': (['x', 'y'], np.random.randn(2, 3)),
   ...:                    'B': ('x', np.random.randn(2)),
   ...:                    'x': ['a', 'b'], 'z': np.pi})

In [7]: data
Out[7]:
<xarray.Dataset>
Dimensions:  (x: 2, y: 3)
Coordinates:
  * x        (x) <U1 'a' 'b'
Data variables:
    A        (x, y) float64 -0.662 -0.705 -0.1866 -0.8655 0.5816 -0.2308
    B        (x) float64 -0.1002 2.402
    z        float64 3.142

In [8]: data.drop_dims('x')
Out[8]:
<xarray.Dataset>
Dimensions:  ()
Data variables:
    z        float64 3.142
```

In this case, I dropped `x`, but because that eliminated variable `A`, I also dropped my `y` dimension, since `A` was the only variable using `y`.

At one point, I implemented this slightly differently, which instead gave

```python
In [8]: data.drop_dims('x')
Out[8]:
<xarray.Dataset>
Dimensions:  (y: 3)
Data variables:
    z        float64 3.142
```

In some ways, this seems preferable, especially if `y` has coordinates.  The only downside I can see is that this version called `Dataset._replace_vars_and_dims()`, the docstring for which says it's deprecated (although it's still widely used).

kmsquire:
@shoyer I'm sure you're busy... wondering if you might have time to review this (or direct it to someone else to review)? 
Other than my question above, it should be pretty straightforward.

shoyer:
It's OK to use `_replace_with_new_dims` still. I think we'll eventually replace it in favor of `_replace_with_new_dims` but we aren't there yet.

I agree that it makes sense to recompute dimensions from variables. There's not much point in keeping around extra dimensions that don't have corresponding variables, and I'm pretty sure that this is consistent with how `Dataset.drop()` works.

kmsquire:
> I agree that it makes sense to recompute dimensions from variables.

Okay, that's what this PR already did (via the call to `_replace_with_new_dims`).

Rebased on master.  From my perspective, this should be good to go.

xarray/core/dataset.py
```diff
@@ -2797,6 +2797,42 @@ def _drop_vars(self, names):
         coord_names = set(k for k in self._coord_names if k in variables)
         return self._replace_vars_and_dims(variables, coord_names)
 
+    def drop_dims(self, drop_dims, inplace=False):
```
shoyer: We've deprecated the `inplace` option for other Dataset methods, so I would rather not add it here.

xarray/core/dataset.py
```diff
@@ -2797,6 +2797,42 @@ def _drop_vars(self, names):
...
+        """Drop dimensions from a Dataset.  Any data variables containing
```
shoyer: For something that fits on one line, maybe:
`"""Drop dimensions and associated variables from this dataset."""`

doc/indexing.rst
```diff
@@ -229,7 +229,7 @@ arrays). However, you can do normal indexing with dimension names:
 Using indexing to *assign* values to a subset of dataset (e.g.,
 ``ds[dict(space=0)] = 1``) is not yet supported.
 
-Dropping labels
+Dropping labels and dimensions
 ---------------
```
shoyer: I think you need to increase the width of the `----` line to match the new label.

doc/indexing.rst
```diff
@@ -241,6 +241,15 @@ index labels along a dimension dropped:
...
 ``drop`` is both a ``Dataset`` and ``DataArray`` method.
 
+Use :py:meth:`~xarray.Dataset.drop_dims` to drop a full dimension.  Any variables
+with these dimensions are also dropped:
+
+.. ipython:: python
+
+    ds.drop_dims('time')
+
+``drop_dims`` is a ``Dataset`` method.
```
shoyer: Maybe delete this, and above say "to drop a full dimension from a dataset."?

xarray/core/dataset.py
kmsquire: Okay.

xarray/core/dataset.py
kmsquire: Okay.

doc/indexing.rst
kmsquire: Done.

doc/indexing.rst
kmsquire: Done.

kmsquire:
@shoyer I force pushed the requested changes.  Please let me know if there is anything else.  Thanks!

xarray/core/dataset.py
```diff
@@ -2797,6 +2797,38 @@ def _drop_vars(self, names):
...
+
+        drop_vars = set(k for k, v in self._variables.items()
+                        for d in v.dims if d in drop_dims)
+
+        variables = OrderedDict((k, v) for k, v in self._variables.items()
+                                if k not in drop_vars)
+        coord_names = set(k for k in self._coord_names if k in variables)
+
+        return self._replace_with_new_dims(
+            variables, coord_names, inplace=inplace)
```
shoyer: `inplace` is now an undefined variable, so your tests are failing

kmsquire: Whoops!  I took that out of the tests, but forgot to take it out of `dataset.py`.  Should be fixed now.

kmsquire:
Rebased to fix the NEWS conflict.

shoyer:
thanks @kmsquire, especially for your patience here!

kmsquire:
No worries--thanks for merging!

naught101:
Is there currently a way to drop unused coordinates, like this?

naught101:
I guess `ds = ds.drop([c for c in ds.coords if c not in ds.dims])` works. Might be nice to have a convenience function for that though.

PR #2769, opened 2019-02-13
Bugfix/reduce no axis

### LINKED ISSUE #2768: [Bug] Reduce fails when no axis given
`DataArray.reduce()` fails if you try to reduce using a function which doesn't accept any axis arguments.

```python
import numpy as np
from xarray import DataArray

da = DataArray(np.array([[1, 3, 3], [2, 1, 5]]))

def total_sum(data):
    return np.sum(data.flatten())

sum = da.reduce(total_sum)
print(sum)
```

This should print a dataarray with just the number 15 in it, but instead it throws the error
```
  File "mwe.py", line 9, in <module>
    sum = da.reduce(total_sum)
  File "/home/tegn500/Documents/Work/Code/xarray/xarray/core/dataarray.py", line 1605, in reduce
    var = self.variable.reduce(func, dim, axis, keep_attrs, **kwargs)
  File "/home/tegn500/Documents/Work/Code/xarray/xarray/core/variable.py", line 1365, in reduce
    axis=axis, **kwargs)
TypeError: total_sum() got an unexpected keyword argument 'axis'
```

This contradicts what the docstring of `.reduce()` says:
```
axis: int or sequence of int, optional
    Axis(es) over which to repeatedly apply func. Only one of the ‘dim’ and ‘axis’ 
    arguments can be supplied. If neither are supplied, then the reduction is 
    calculated over the flattened array (by calling f(x) without an axis argument).
```

The problem is that in `variable.py` an `axis=None` kwarg is always passed to func, even if no axis argument is given by the user in `reduce`.

TomNicholas:
- [x] Closes #2768
 - [x] Tests added

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -1361,11 +1361,11 @@ class Variable(common.AbstractArray, arithmetic.SupportsArithmetic,
 
         if dim is not None:
             axis = self.get_axis_num(dim)
+        input_data = self.data if allow_lazy else self.values
         if axis is not None:
-            data = func(self.data if allow_lazy else self.values,
-                        axis=axis, **kwargs)
+            data = func(input_data, axis=axis, **kwargs)
         else:
-            data = func(self.data if allow_lazy else self.values, **kwargs)
+            data = func(input_data, **kwargs)
 
         if getattr(data, 'shape', ()) == self.shape:
             dims = self.dims
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -1361,8 +1361,11 @@ class Variable(common.AbstractArray, arithmetic.SupportsArithmetic,
 
         if dim is not None:
             axis = self.get_axis_num(dim)
-        data = func(self.data if allow_lazy else self.values,
-                    axis=axis, **kwargs)
+        if axis is not None:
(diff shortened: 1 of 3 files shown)


xarray/core/variable.py
```diff
@@ -1361,8 +1361,11 @@ def reduce(self, func, dim=None, axis=None,
 
         if dim is not None:
             axis = self.get_axis_num(dim)
-        data = func(self.data if allow_lazy else self.values,
-                    axis=axis, **kwargs)
+        if axis is not None:
+            data = func(self.data if allow_lazy else self.values,
+                        axis=axis, **kwargs)
+        else:
+            data = func(self.data if allow_lazy else self.values, **kwargs)
```
max-sixty: This no longer passes `axis=None` into `func`, which I think is reasonable - it's generally the default. (but someone should confirm my thinking)

shoyer: This looks good to me, though I would rewrite this to an intermediate variable for `self.data if allow_lazy else self.values` rather than writing it out twice

TomNicholas: Done, should be ready for merging now.

shoyer:
Thanks!

Issue #2382, opened 2018-08-27
Add option to choose the source of global attributes in mfdataset.


juseg:
#### Code Sample

```python
import numpy as np
import xarray as xr

# prepare fake data
time = np.arange(1000)
data = time**2

# write to multiple netcdf files
for i in range(10):
    filename = 'ds{:d}.nc'.format(i)
    ds = xr.Dataset({'data': (['time'],  data[100*i:100*i+100])},
                    coords={'time': time[100*i:100*i+100]},
                    attrs={'history': 'Created ' + filename + '.'})
    ds.to_netcdf(filename)

# open as mfdataset
with xr.open_mfdataset('ds?.nc') as ds:
    print ds.history
```
#### Problem description

Currently, global attributes of multi-file datasets are taken from the first file in the list.

https://github.com/pydata/xarray/blob/b87b684b36cf5adbe4dca208aed0c69c44fc44c4/xarray/backends/api.py#L649

I think this is a problem in the context of consecutive model runs where history is appended in each subsequent run. When opening the results as mfdataset, history is taken from the first run file all subsequent history is lost.

NetCDF4 has a new keyword argument to set the `master_file` in a `MFDataset` (Unidata/netcdf4-python#835). Would it be possible to add a similar option in xarray?

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.15.final.0
python-bits: 64
OS: Linux
OS-release: 4.14.65-1-MANJARO
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: fr_CH.utf8
LOCALE: None.None

xarray: 0.10.8
pandas: 0.23.4
numpy: 1.15.0
scipy: 1.1.0
netCDF4: 1.4.2
h5netcdf: None
h5py: None
Nio: None
zarr: None
bottleneck: None
cyordereddict: None
dask: 0.18.2
distributed: None
matplotlib: 2.2.3
cartopy: 0.15.1
seaborn: None
setuptools: 40.0.0
pip: 18.0
conda: None
pytest: None
IPython: 5.8.0
sphinx: 1.7.6
</details>

Issue #1646, opened 2017-10-23
Make passing a DataArray for the xarray.concat dim argument equivalent to passing a pandas Index


ceridwen:
Extending from #839, if I'm concatenating some DataArrays using concat,

```
print(xarray.concat(data, xarray.DataArray(['foo1', 'foo2', 'foo3', 'foo4', 'foo5'], name='stat'))
```

I get an unnamed dimension without coordinates.

```
<xarray.DataArray (dim_0: 5, index: 2)>
array([[ 24.841064,   0.750451],
       [ 24.841064,   0.750451],
       [ 19.062874,   0.796722],
       [ 14.9631  ,   0.354273],
       [ 14.9631  ,   0.354273]])
Coordinates:
  * index    (index) object 'Intercept' 'Lvl'
         (dim_0) <U3 'foo1' foo2' 'foo3' 'foo4' 'foo5'
Dimensions without coordinates: dim_0
```

Using a pandas.Index,

```
print(xarray.concat(data, pandas.Index(['foo1', 'foo2', 'foo3', 'foo4', 'foo5'], name='stat'))
```

```
<xarray.DataArray (stat: 5, index: 2)>
array([[ 14.9631  ,   0.354273],
       [ 19.982272,   0.555708],
       [ 14.974026,   0.60658 ],
       [ 24.841064,   0.750451],
       [ 24.841064,   0.750451]])
Coordinates:
  * index    (index) object 'Intercept' 'Lvl'
  * stat     (stat) object 'foo1' 'foo2' 'foo3' 'foo4' 'foo5'
```

I want the latter, not the former, but I expected the latter when using a DataArray.

shoyer:
Agreed, this should definitely work! (I think the fact that it doesn't is probably related to the relatively recent change that made coordinate labels along dimensions optional.)

shoyer:
This is the location of the helper function that parsing the `dim` argument: https://github.com/pydata/xarray/blob/9763a66e0e4675e7adc3fff3830c62f0e31a2bb3/xarray/core/combine.py#L123

dcherian:
I have a simple fix in #1812 .

With that change, this works.
> print(xarray.concat(data, xarray.DataArray(['foo1', 'foo2', 'foo3', 'foo4', 'foo5'], name='stat'))

But if you provided a dimension name like
[...]

PR #2777, opened 2019-02-19
Improved default behavior when concatenating DataArrays

### LINKED ISSUE #2775: Improved inference of names when concatenating arrays
#### Problem description

Using the name of the first element to concatenate as the name of the concatenated array is only correct if all names are identical.  When names vary, using a clear placeholder name or the name of the new dimension would avoid misleading data users.

This came up for me recently when stacking several bands of a satellite image to produce a faceted plot - the resulting colorbar was labelled "blue", even though that was clearly incorrect.

A similar process is probably also desirable for aggregation of units across concatenated arrays - use first if identical, otherwise discard or error depending on the `compat` argument.

```python
ds = xr.Dataset({
    k: xr.DataArray(np.random.random((2, 2)), dims="x y".split(), name=k) 
    for k in "blue green red".split()
})
# arr.name == "blue", could be "band" or "concat_dim"
arr = xr.concat([ds.blue, ds.green, ds.red], dim="band")
# label of colorbar is "blue", which is meaningless
arr.plot.imshow(col="band")
```

image

One implementation that would certainly be nice for this use-case (though perhaps not generally) is that concatenating `DataArray`s along an entirely new dimension with unique array names and `dim` passed a string could create a new Index as well, as `pd.Index([a.name for a in objs], name=dim)`.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5 |Anaconda, Inc.| (default, Mar 29 2018, 13:32:41) [MSC v.1900 64 bit (AMD64)]
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 9, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None
libhdf5: 1.10.3
libnetcdf: 4.4.1.1

xarray: 0.11.2
pandas: 0.23.1
numpy: 1.14.5
scipy: 1.2.1
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.0.3.4
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
cyordereddict: None
dask: None
distributed: None
matplotlib: 3.0.2
cartopy: None
seaborn: 0.9.0
setuptools: 40.6.2
pip: 10.0.1
conda: None
pytest: 4.2.0
IPython: 6.4.0
sphinx: 1.8.0

</details>

I'd be happy to write a PR for this if it would be accepted.

Zac-HD:
- [x] Closes #2775
 - [x] Tests added

This is really nice to have when producing faceted plots of satellite observations in various bands, and should be somewhere between useful and harmless in other cases.

Example code:

```python
ds = xr.Dataset({
    k: xr.DataArray(np.random.random((2, 2)), dims="x y".split(), name=k) 
    for k in "blue green red".split()
})
xr.concat([ds.blue, ds.green, ds.red], dim="band").plot.imshow(col="band")
```

Before - facets have an index, colorbar has misleading label: 

image

After - facets have meaningful labels, colorbar has no label:

image

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -40,6 +40,9 @@ Enhancements
   `Spencer Clark <https://github.com/spencerkclark>`_.
 - Add ``data=False`` option to ``to_dict()`` methods. (:issue:`2656`)
   By `Ryan Abernathey <https://github.com/rabernat>`_
+- Use new dimension name and unique array names to create a new coordinate
+  when concatenating arrays, if no coordinates are given.
+  (:issue:`2775`). By `Zac Hatfield-Dodds <https://github.com/Zac-HD>`_.
 - :py:meth:`~xarray.DataArray.coarsen` and
   :py:meth:`~xarray.Dataset.coarsen` are newly added.
   See :ref:`comput.coarsen` for details.
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -82,6 +82,10 @@ Bug fixes
 
 - Silenced warnings that appear when using pandas 0.24.
   By `Stephan Hoyer <https://github.com/shoyer>`_
+- Concatenating a sequence of :py:class:`~xarray.DataArray` with varying names
+  sets the name of the output array to ``None``, instead of the name of the
+  first input array.
+  (:issue:`2775`). By `Zac Hatfield-Dodds <https://github.com/Zac-HD>`_.
 - Interpolating via resample now internally specifies ``bounds_error=False``
   as an argument to ``scipy.interpolate.interp1d``, allowing for interpolation
   from higher frequencies to lower frequencies.  Datapoints outside the bounds
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -6,6 +6,7 @@ import pandas as pd
 
 from . import utils
 from .alignment import align
+from .computation import result_name
 from .merge import merge
 from .variable import IndexVariable, Variable, as_variable
 from .variable import concat as concat_vars
@@ -323,16 +324,10 @@ def _dataarray_concat(arrays, dim, data_vars, coords, compat,
         raise ValueError('data_vars is not a valid argument when '
                          'concatenating DataArray objects')
 
-    datasets = []
-    for n, arr in enumerate(arrays):
-        if n == 0:
-            name = arr.name
-        elif name != arr.name:
-            if compat == 'identical':
-                raise ValueError('array names not identical')
-            else:
-                arr = arr.rename(name)
-        datasets.append(arr._to_temp_dataset())
+    name = result_name(arrays)
+    if name is None and compat == 'identical':
+        raise ValueError('array names not identical')
+    datasets = [arr.rename(name)._to_temp_dataset() for arr in arrays]
 
     ds = _dataset_concat(datasets, dim, data_vars, coords, compat,
                          positions)
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -325,9 +325,22 @@ def _dataarray_concat(arrays, dim, data_vars, coords, compat,
                          'concatenating DataArray objects')
 
     name = result_name(arrays)
-    if name is None and compat == 'identical':
-        raise ValueError('array names not identical')
+    names = [arr.name for arr in arrays]
+    if compat == 'identical' and len(set(names)) != 1:
+        raise ValueError(
+            "compat='identical', but array names {!r} are not identical"
+            .format(names if len(names) <= 10 else sorted(set(names)))
+        )
     datasets = [arr.rename(name)._to_temp_dataset() for arr in arrays]
+
+    if (
+        isinstance(dim, str)
+        and len(set(names) - {None}) == len(names)
+        and not any(dim in a.dims or dim in a.coords for a in arrays)
+    ):
+        # We're concatenating arrays with unique non-None names along
+        # a new dimension, so we use the existing names as coordinates.
+        dim = pd.Index(names, name=dim)
 
     ds = _dataset_concat(datasets, dim, data_vars, coords, compat,
                          positions)
--- a/xarray/tests/test_combine.py
+++ b/xarray/tests/test_combine.py
@@ -248,7 +248,8 @@ class TestConcatDataArray(object):
 
         # from dataset array:
         expected = DataArray(np.array([foo.values, bar.values]),
-                             dims=['w', 'x', 'y'], coords={'x': [0, 1]})
+                             dims=['w', 'x', 'y'],
+                             coords={'x': [0, 1], 'w': ['foo', 'bar']})
         actual = concat([foo, bar], 'w')
         assert_equal(expected, actual)
         # from iteration:
@@ -297,15 +298,24 @@ class TestConcatDataArray(object):
         assert combined.shape == (2, 3, 3)
         assert combined.dims == ('z', 'x', 'y')
 
-    def test_concat_names(self):
+    def test_concat_names_and_coords(self):
         ds = Dataset({'foo': (['x', 'y'], np.random.random((2, 2))),
                       'bar': (['x', 'y'], np.random.random((2, 2)))})
         # Concat arrays with different names, new name is None
+        # and unique array names are used as coordinates
         new = concat([ds.foo, ds.bar], dim='new')
         assert new.name is None
+        assert (new.coords['new'] == ['foo', 'bar']).values.all()
+        # Get a useful error message for unexpectedly different names
+        with pytest.raises(ValueError) as err:
+            concat([ds.foo, ds.bar], dim='new', compat='identical')
+        assert err.value.args[0] == "compat='identical', " + \
+            "but array names ['foo', 'bar'] are not identical"
         # Concat arrays with same name, name is preserved
+        # and non-unique names are not used as coords
         foobar = ds.foo.rename('bar')
         assert concat([foobar, ds.bar], dim='new').name == 'bar'
+        assert 'new' not in concat([foobar, ds.bar], dim='new').coords
 
 
 class TestAutoCombine(object):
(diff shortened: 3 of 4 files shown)


xarray/core/combine.py
```diff
@@ -323,16 +324,17 @@ def _dataarray_concat(arrays, dim, data_vars, coords, compat,
...
-            name = arr.name
-        elif name != arr.name:
-            if compat == 'identical':
-                raise ValueError('array names not identical')
-            else:
-                arr = arr.rename(name)
-        datasets.append(arr._to_temp_dataset())
+    name = result_name(arrays)
+    if name is None and compat == 'identical':
+        raise ValueError('array names not identical')
```
shoyer: I think it’s also possible to get a name of None if none of the arguments had names at all. So this logic probably needs to go in result_name()

xarray/core/combine.py
```diff
@@ -323,16 +324,17 @@ def _dataarray_concat(arrays, dim, data_vars, coords, compat,
...
+    datasets = [arr.rename(name)._to_temp_dataset() for arr in arrays]
+
+    if name is None and isinstance(dim, str) and dim not in arrays[0].dims:
```
shoyer: This should verify that there are no coordinates with the dimension name, either — and probably  should check every argument, too. I think we support some cases with broadcasting where dimensions/coordinates can differ across arguments.

xarray/core/combine.py
Zac-HD: None of the other `result_name` calls would use this logic, so I've implemented it but in `_dataarray_combine`.

xarray/core/combine.py
Zac-HD: We now check that the string `dim` is not in the coords or dims of any of the arrays to concatenate.

Zac-HD:
Thanks for the support and quick review @shoyer!

Any idea when Xarray 0.12 might be out?  I'm teaching some remote sensing workshops in mid-March and would love to have this merged, as a colleague's review of those notebooks prompted this PR 😄

xarray/core/combine.py
```diff
@@ -323,16 +324,17 @@ def _dataarray_concat(arrays, dim, data_vars, coords, compat,
...
-        elif name != arr.name:
-            if compat == 'identical':
-                raise ValueError('array names not identical')
-            else:
-                arr = arr.rename(name)
-        datasets.append(arr._to_temp_dataset())
+    name = result_name(arrays)
+    names = [arr.name for arr in arrays]
+    if compat == 'identical' and len(set(names)) != 1:
+        raise ValueError('array names not identical')
```
shoyer: Can you add a test for this error message? Also, please include the conflicting names in the error message -- that makes the error messages easier to debug.

Zac-HD: Done - the new message includes all names if ten or less, but just the unique names over ten to avoid incredibly long messages with large data.  Plus a test of course!

Zac-HD:
The docs build failed due to an http error when loading tutorial data for the docs, so I've also finalised the planned conversion from `xarray.tutorial.load_dataset` to `xarray.tutorial.open_dataset`.

Zac-HD:
Hmm, it looks like the failure to download [the naturalearth coastlines.zip](http://naciscdn.org/naturalearth/10m/physical/ne_10m_coastline.zip) wasn't so transient after all - but it *does* work on my machine!

Zac-HD:
OK!  @shoyer, I've got everything passing and it's ready for review.

Even the accidental tutorial/docs fixes :smile:

doc/whats-new.rst
```diff
@@ -30,6 +30,10 @@ Breaking changes
   By `Maximilian Roos <https://github.com/max-sixty>`_.
 - `cyordereddict` is no longer used as an optional dependency (:issue:`2744`).
   By `Joe Hamman <https://github.com/jhamman>`_.
+- ``xarray.tutorial.load_dataset`` has been removed as scheduled in Xarray 0.11.
```
dcherian: 0.12 not 0.11

xarray/core/combine.py
```diff
@@ -323,16 +324,20 @@ def _dataarray_concat(arrays, dim, data_vars, coords, compat,
...
+    name = result_name(arrays)
+    names = [arr.name for arr in arrays]
+    if compat == 'identical' and len(set(names)) != 1:
+        raise ValueError(
+            "compat='identical', but array names {!r} are not identical"
+            .format(names if len(names) <= 10 else sorted(set(names)))
+        )
+    datasets = [arr.rename(name)._to_temp_dataset() for arr in arrays]
+
+    if isinstance(dim, str) and len(set(names) - {None}) == len(names) \
```
dcherian: can we not do `None in set(names)`?

Zac-HD: Not for this - we're checking that names is a list of unique strings.

shoyer:
I'll check on the doc build failure, but I would rather not remove
load_dataset yet -- we didn't deprecate it that long ago.

Zac-HD:
@shoyer - don't worry about the docs build, I'm pretty sure that was just a flaky network from Travis and it's working now in any case.

I've left `tutorial.load_dataset` in, just changed "removed in 0.12" to "removed in a future version".

xarray/core/combine.py
shoyer: nit: per PEP8, I try to avoid explicitly using `\`, in preference to explicit parentheses

Zac-HD: These days I just feed everything through `black`!

I'd even recommend it for Xarray, but that's a separate pull request 😄

shoyer:
@pep8speaks seems to have gone hay-wire -- maybe you have a syntax error?

Thinking about this a little more, one hazard of converting names into index labels is that we lose the invariant that you get the same result regardless of order in which you call concat, e.g., something like these expressions could now give different results:
```
xarray.concat([a, b], dim='x')
```
vs
```
xarray.concat([xarray.concat([a], dim='x'), xarray.concat([b], dim='x')], dim='x')
```
If `a` and `b` have the same name, then you'd get an index with the two duplicate entries in the second case but not the first case.

I'm not entirely sure this is a deal-breaker but it makes me a little nervous reluctant. In particular, it might break some the invariants we're relying upon for the next version of `open_mfdataset` (https://github.com/pydata/xarray/pull/2616, cc @TomNicholas )

Zac-HD:
> maybe you have a syntax error?

...yep, an unmatched paren.  😥 

> If `a` and `b` have the same name, then you'd get an index with the two duplicate entries in the second case but not the first case.  [which would be bad]

I think it's impossible to avoid this when using inference in the general case.  Two options I think would be decent-if-unsatisfying:

1. Explicitly manage this in the new combining functions, e.g. clear the concat dim coords if they are not unique and the input arrays did not have coords in that dimension.
2. Add an argument to `xr.concat` to enable or disable this, e.g. `infer_coords=True`, and disable it when calling `xr.concat` from other combining functions.

TomNicholas:
@Zac-HD forgive me for this but I think what you need basically already exists in the API.

Going back to your original example, you could have got the same indexing by creating a DataArray to use as a coordinate to concatenate over:
```python
colors = "blue green red".split()
ds = xr.Dataset({
    k: xr.DataArray(np.random.random((2, 2)), dims="x y".split(), name=k)
    for k in colors
})

band = xr.DataArray(colors, name="band", dims=["band"])
xr.concat([ds.blue, ds.green, ds.red], dim=band).plot.imshow(col="band")

```

figure_1

This still leaves the wrong label on the colorbar, but that could be fixed separately and has to do with concat using the attrs of the first dataset in the list for the final dataset (a similar problem to #2382). I think it would be easier to change that behaviour instead (perhaps to `if all names the same, use that name, else name of result = None`, but this also relates to #1614).  

Creating a new coordinate using a DataArray is in the docstring for `xr.concat`: 
> If dimension is provided as a DataArray or Index, its name is used as the dimension to concatenate along and the values are added as a coordinate.

but I think there should be an example there too. (Also I think this is relevant to #1646)

> I'm not entirely sure this is a deal-breaker but it makes me a little nervous

@shoyer I agree, although I like the idea then I think this could introduce all sorts of complex concatentation edge cases.

At the very least the new API should have symmetry properties something like:

```python
da1 = DataArray(name='a': data=[[0]], dims=['x', 'y'])
da2 = DataArray(name='b': data=[[1]], dims=['x', 'y'])
da3 = DataArray(name='a': data=[[2]], dims=['x', 'y'])
da4 = DataArray(name='b': data=[[3]], dims=['x', 'y'])

xr.manual_combine([[da1, da2], [da3, da4]], concat_dim=['x', 'y'])
# should give the same result as 
xr.manual_combine([[da1, da3], [da2, da4]], concat_dim=['y', 'x'])
```
but with this PR I don't think it would. In the first case the x coord would be created with values `['a', 'b']`, and no y coord would be created, while in the second case no y coord would be created, and the intermediate DataSet would be nameless, so then no x coord would be created either. 

I think my suggestion for naming would pass this test because the result would be nameless and have no coords either way.

I might have got that wrong but I definitely think this kind of change should be carefully considered :confused:

TomNicholas:
I've just submitted a PR which solves this issue in the way I just suggested instead.

TomNicholas:
@Zac-HD there's actually another way to get the indexing behaviour you wanted with the current API:

```python
colors = "blue green red".split()
das = [xr.DataArray(np.random.random((2, 2)), dims="x y".split(),
                    coords={"band": k})
       for k in colors]
xr.concat(das, dim="band").plot.imshow(col="band")
```
Here instead of using the name attribute to label each band I've used a scalar coordinate called `"band"`, so that when you concat along `"band"` it will just stack along that coordinate.

This never touches the names so actually gives the desired output without the need for #2792:
figure_2

shoyer:
I guess we should probably roll back the "name to scalar coordinates" part of this change.

@Zac-HD do you want to do that here or should we go with @TomNicholas's PR?

Zac-HD:
No objection to going with #2792; I'm just happy to have the change merged 😄 

It would be nice for someone to cherry-pick 63da214d697345ebdd0ecc0967c72eafc70bcb0d before releasing 0.12 though, just to fix that warning.

PR #2782, opened 2019-02-20
enable loading remote hdf5 files

### LINKED ISSUE #2781: enable reading of file-like HDF5 objects
#### xarray 11.3 currently won't read HDF5 file-like objects

```python
import xarray as xr
import gcsfs
fs = gcsfs.GCSFileSystem()
images = fs.ls('pangeo-data/grfn-v2/137/')
fileObj = fs.open('pangeo-data/grfn-v2/137/S1-GUNW-A-R-137-tops-20181129_20181123-020010-43220N_41518N-PP-e2c7-v2_0_0.nc')
# but, can we open this w/ xarray anyway? Yes! with modifications to xarray and h5netcdf
da = xr.open_dataset(fileObj, group='/science/grids/data', engine='h5netcdf')
da 
```

```pytb
<ipython-input-3-22e0010de1f2> in <module>()
      1 # but, can we open this w/ xarray anyway? Yes! with modifications to xarray and h5netcdf
----> 2 da = xr.open_dataset(fileObj, group='/science/grids/data', engine='h5netcdf')
      3 da

/srv/conda/lib/python3.6/site-packages/xarray/backends/api.py in open_dataset(filename_or_obj, group, decode_cf, mask_and_scale, decode_times, autoclose, concat_characters, decode_coords, engine, chunks, lock, cache, drop_variables, backend_kwargs)
    347     else:
    348         if engine is not None and engine != 'scipy':
--> 349             raise ValueError('can only read file-like objects with '
    350                              "default engine or engine='scipy'")
    351         # assume filename_or_obj is a file-like object

ValueError: can only read file-like objects with default engine or engine='scipy'
```

#### Problem description

It is now possible to do this with h5py >2.9.0. see https://github.com/h5py/h5py/pull/1105. This would be a useful feature because there is a lot of NASA data out there in HDF5. This functionality could open up reading without first writing to disk (to translate to Zarr or other formats possibly). There seem to be many issues related to this:
https://github.com/dask/s3fs/issues/144
https://github.com/pydata/xarray/issues/2535

I'm guessing adding this functionality doesn't fix many of the performance issues related to HDF5 and Dask
https://github.com/dask/dask/issues/2488
https://github.com/dask/distributed/issues/2319

#### Expected Output
```
<xarray.Dataset>
Dimensions:              (latitude: 2045, longitude: 4158)
Coordinates:
  * longitude            (longitude) float64 -123.1 -123.1 ... -119.6 -119.6
  * latitude             (latitude) float64 43.22 43.22 43.22 ... 41.52 41.52
Data variables:
    crs                  int32 ...
    unwrappedPhase       (latitude, longitude) float32 ...
    coherence            (latitude, longitude) float32 ...
    connectedComponents  (latitude, longitude) float32 ...
    amplitude            (latitude, longitude) float32 ...
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.7 | packaged by conda-forge | (default, Nov 21 2018, 03:09:43) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.14.65+
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.2

xarray: 0.11.3
pandas: 0.24.1
numpy: 1.16.1
scipy: 1.2.0
netCDF4: 1.4.2
pydap: None
h5netcdf: 0.6.2
h5py: 2.9.0
Nio: None
zarr: 2.2.0
cftime: 1.0.3.4
PseudonetCDF: None
rasterio: 1.0.18
cfgrib: None
iris: None
bottleneck: None
cyordereddict: None
dask: 1.1.0
distributed: 1.25.2
matplotlib: 3.0.2
cartopy: 0.17.0
seaborn: 0.9.0
setuptools: 40.7.1
pip: 19.0.2
conda: 4.6.3
pytest: None
IPython: 7.1.1
sphinx: None
</details>

scottyhq:
Enable loading remote hdf5 files. Will require h5py>2.9.0 and some changes to https://github.com/shoyer/h5netcdf. I've current just made a quick hack change to backends/api.py, so further tests are needed. Pinging @jhamman, @mrocklin, and @rabernat for thoughts on this. 

Here is a short notebook demonstrating how this works:
https://gist.github.com/scottyhq/790bf19c7811b5c6243ce37aae252ca1

 - [x] Closes #2781
 - [x] Tests added


mrocklin:
I'm glad to see this.  I'll also be curious to see what the performance will look like.

cc @llllllllll

shoyer:
This looks great!

I'll note one minor extension: you could look at the first few bytes of the file (the "magic number") to determine if it's a netCDF3 or netCDF4 file, and hence whether it can be opened with scipy or h5netcdf:

- `CDF\001` or `CDF\002` would indicate [netCDF3](https://en.wikipedia.org/wiki/NetCDF) (use scipy)
- `\211HDF\r\n\032\n` would indicate [netCDF4/HDF5](https://en.wikipedia.org/wiki/Hierarchical_Data_Format) (use h5netcdf)

scottyhq:
@shoyer , it would be great to have your feedback on these recent changes now that h5netcdf 0.7 is out. There's a bit more logic required in `api.py` now that scipy isn't the only backend that is able to read file-like objects (and people may not specify `engine=` when opening datasets)

`test_backends.py` passes locally for me except for `TestValidateAttrs.test_validating_attrs`... not sure why. 

Also, per your comment here: https://github.com/shoyer/h5netcdf/pull/51#issuecomment-467591446, I think it would be great to get a few small netcdf4/hdf test files in https://github.com/pydata/xarray-data.

xarray/backends/api.py
```diff
@@ -352,11 +344,24 @@ def maybe_decode_store(store, lock=False):
...
+            filename_or_obj.seek(0)
+            if magic_number.startswith(b'CDF'):
+                store = backends.ScipyDataStore(filename_or_obj,
+                                                **backend_kwargs)
+            elif magic_number.startswith(b'\211HDF\r\n\032\n'):
+                store = backends.H5NetCDFStore(filename_or_obj, group=group,
+                                               lock=lock, **backend_kwargs)
+            else:
+                raise ValueError("byte header doesn't match netCDF3 or "
+                                 "netCDF4/HDF5: {}".format(magic_number))
```
shoyer: I suspect this is one of those rare cases where it's best *not* to report all the details -- most users probably don't know about magic numbers. Maybe something like:
- `"file-like object is not a netCDF file: {}"`.format(filename_or_obj)`, or
- `"bytes do not represent in-memory netCDF file: {}. (Pass a string or pathlib.Path object to read a filename from disk.)".format(filename_or_obj[:80] + b'...' if len(filename_or_obj) > 80 else b'')`

xarray/tests/test_backends.py
```diff
@@ -1955,6 +1955,38 @@ def test_dump_encodings_h5py(self):
             assert actual.x.encoding['compression_opts'] is None
 
 
+# Requires h5py>2.9.0
```
shoyer: Can you add a `pytest.mark.skipif` based on the version number? (The test on Travis-CI is failing on Python 3.5 because it has an old version of h5py installed)

xarray/tests/test_backends.py
```diff
@@ -1955,6 +1955,38 @@ def test_dump_encodings_h5py(self):
...
+
+    def test_h5bytes(self):
+        import h5py
+        bio = BytesIO()
+        with h5py.File(bio) as ds:
+            v = np.array(2.0)
+            ds['scalar'] = v
+        bio.seek(0)
+        with xr.open_dataset(bio) as ds:
+            v = ds['scalar']
```
shoyer: prefer using `assert_identical` and comparing to another expected dataset object.

xarray/tests/test_backends.py
```diff
@@ -1955,6 +1955,38 @@ def test_dump_encodings_h5py(self):
...
+        print(url)
+        bytes = requests.get(url).content
+        with xr.open_dataset(bytes) as ds:
+            assert len(ds['UTC_time']) == 74
+            assert ds['UTC_time'].attrs['name'] == 'time'
+
+    def test_h5bytes(self):
+        import h5py
+        bio = BytesIO()
+        with h5py.File(bio) as ds:
```
shoyer: Wouldn't it be nice if we supported writing to file-like objects, too? :)

(But don't do that now, this PR is a nice logical size already.)

scottyhq: agreed. hopefully someone else could pick that up!

xarray/tests/test_backends.py
scottyhq: i've changed that test quite a bit

xarray/backends/api.py
scottyhq: went with the first more-concise option

xarray/tests/test_backends.py
scottyhq: i think i did this correctly (added some lines to `tests/__init__.py`)

xarray/backends/api.py
```diff
@@ -352,11 +344,25 @@ def maybe_decode_store(store, lock=False):
...
+            magic_number = filename_or_obj.read(8)
+            filename_or_obj.seek(0)
+            if magic_number.startswith(b'CDF'):
+                store = backends.ScipyDataStore(filename_or_obj,
+                                                **backend_kwargs)
+            elif magic_number.startswith(b'\211HDF\r\n\032\n'):
+                store = backends.H5NetCDFStore(filename_or_obj, group=group,
+                                               lock=lock, **backend_kwargs)
+            else:
+                print(magic_number)
```
shoyer: should be deleted

xarray/tests/test_backends.py
```diff
@@ -1955,6 +1955,39 @@ def test_dump_encodings_h5py(self):
...
+class TestH5NetCDFFileObject(TestH5NetCDFData):
+    h5py = pytest.importorskip('h5py', minversion='2.9.0')
+    engine = 'h5netcdf'
+
+    @network
+    def test_h5remote(self):
+        # alternative: http://era5-pds.s3.amazonaws.com/2008/01/main.nc
+        import requests
+        url = ('https://www.unidata.ucar.edu/'
+               'software/netcdf/examples/test_hgroups.nc')
```
rabernat: Rather than going over the network, it might be quite easy to fire up a [`http.server.SimpleHTTPRequestHandler`](https://docs.python.org/3/library/http.server.html#http.server.SimpleHTTPRequestHandler) as part of a fixture. This would allow us to test the remote capability without internet (and without depending on a third party to host a file.)

xarray/backends/api.py
```diff
@@ -352,11 +344,25 @@ def maybe_decode_store(store, lock=False):
...
+                raise ValueError("file-like object is not a netCDF file: {}"
+                                 .format(filename_or_obj))
```
shoyer: I'm a little concerned about giving users an error message about file-like objects if they passed in a `bytes` object, e.g., consider `xarray.open_dataset(b'garbage')`.

Ideally this should give a useful error message, something like: `ValueError:  b'garbage' is not a valid netCDF file, did you mean to pass a string for a path instead?`, not `ValueError: file-like object is not a netCDF file: <_io.BytesIO at 0x105663888>`.

xarray/backends/api.py
```diff
@@ -352,11 +344,25 @@ def maybe_decode_store(store, lock=False):
...
-                             "default engine or engine='scipy'")
-        # assume filename_or_obj is a file-like object
-        store = backends.ScipyDataStore(filename_or_obj)
+        if engine not in [None, 'scipy', 'h5netcdf']:
+            raise ValueError('can only read bytes or file-like objects with '
+                             "engine = None, 'scipy', or 'h5netcdf'")
+        else:
+            if isinstance(filename_or_obj, bytes):
+                filename_or_obj = BytesIO(filename_or_obj)
+            # read first bytes of file-like object to determine engine
```
shoyer: It would be nice to factor this logic (for file-type detection) out into a separate helper function, just to keep `open_dataset()` a little tidier.

shoyer:
I don't think it's essential to have an integration test doing real network
access in xarray, so I would consider just dropping that part instead.

xarray/tests/test_backends.py
scottyhq: thanks for the suggestion @rabernat, i've gone with shoyers suggestion below to scrap the network test for now

xarray/backends/api.py
scottyhq: i've done this now and added some new error messages and tests. one of the things i did is not cast to BytesIO() behind the scenes, because this was leading to confusing errors if open_dataset() was called more than once and the file/buffer was then no longer at the initial read position. i guess the alternative is adding a line that forces `filename_or_obj.seek(0)`...

scottyhq:
thanks for the input @shoyer, i attempted to tidy up a bit and in the process re-ordered some things such as adding an 'engine' check at the top of open_dataset(). backend tests are passing locally on my machine. hopefully i didn't add too much here or overstep!

jhamman:
@scottyhq - can you add note to the what's new page? 

From what I can tell, I don't think the failing tests are related to this PR.

xarray/backends/api.py
```diff
@@ -75,6 +75,34 @@ def _get_default_engine_netcdf():
...
+                             "manager")
+        magic_number = filename_or_obj.read(8)
+        filename_or_obj.seek(0)
+
+    if magic_number.startswith(b'CDF'):
+        engine = 'scipy'
+    elif magic_number.startswith(b'\211HDF\r\n\032\n'):
+        engine = 'h5netcdf'
+        if isinstance(filename_or_obj, bytes):
+            raise ValueError("can't open netCDF4/HDF5 as bytes "
```
shoyer: Just a note: we *could* support this in the future, by wrapping bytes in a `io.BytesIO` object (like we do for the scipy backend). But no need to add it now -- I like explicitly providing file objects.

xarray/tests/test_backends.py
```diff
@@ -1955,6 +1955,52 @@ def test_dump_encodings_h5py(self):
...
+            with open_dataset(BytesIO(b'garbage'), engine='h5netcdf'):
+                pass
+
+    def test_open_twice(self):
+        expected = create_test_data()
+        expected.attrs['foo'] = 'bar'
+        with raises_regex(ValueError, 'read/write pointer not at zero'):
+            with create_tmp_file() as tmp_file:
+                expected.to_netcdf(tmp_file, engine='h5netcdf')
+                f = open(tmp_file, 'rb')
```
shoyer: There is a real test failure on Window (see the Appveyor CI results), likely because this file never get closed. You should use a context manager here instead.

scottyhq: that test was for the case where the file isn't closed before reopening, but it looks like on windows the error is different compared to linux (` PermissionError versus ValueError`), so I added a check for the windows error: ` PermissionError: [WinError 32] The process cannot access the file because it is being used by another process:`

jhamman:
I think we're good here. I made one minor tweak to the windows fix @scottyhq implemented. I plan to merge this on Monday if I don't hear any objections.

shoyer:
thanks @scottyhq !

Issue #2776, opened 2019-02-19
0.12.0 release


shoyer:
We have quite a few nice new features merged into master.

Is anything holding up making a 0.12 release shortly?

cc @pydata/xarray

dcherian:
#2759 looks ready and complements all the cftime stuff.

#2769 seems like a nice bugfix to merge.

Otherwise lgtm.

spencerkclark:
I think #2721 looks ready too, and @jwenfai and @tlogan2000 would probably appreciate it if we merged it in time for the 0.12 release (sorry for letting that sit for a couple weeks).

Zac-HD:
I'd love to get #2777 merged - it's a small patch, but really useful in a remote sensing niche. 

My students are going to be graded on how well their figures are labelled, and I really don't want to have to teach them matplotlib 😅

TomNicholas:
Potentially completing #2616 would start the deprecation cycle for `auto_combine` etc? I would need at least a few days to do that though...

shoyer:
I'm going to cherry-pick https://github.com/pydata/xarray/commit/63da214d697345ebdd0ecc0967c72eafc70bcb0d and then try to get a release out.

We can start the deprecation cycle for auto-combine later.

shoyer:
I guess we should probably fix and/or remove the failing test in https://github.com/pydata/xarray/issues/2803, too.

max-sixty:
(I have let some of my PRs stay almost-finished. I will endeavor to follow my own advice and complete soon. Thanks for everyone's patience)

shoyer:
Any objections to pushing back finalizing deprecations currently marked for 0.12? See https://github.com/pydata/xarray/pull/2809

I'm a little concerned that most of these changes (introduced in 0.11) haven't been around long enough to safely finalize the deprecations.

shoyer:
I finally tagged the 0.12 release... probably should have done this a while ago!

jhamman:
Thanks @shoyer. I merged the conda-forge build about an hour ago.

PR #2809, opened 2019-03-14
Push back finalizing deprecations for 0.12

shoyer:
0.12 will already have a big change in dropping Python 2.7 support. I'd rather
wait a bit longer to finalize these deprecations to minimize the impact on
users.

xref https://github.com/pydata/xarray/issues/2776


fujiisoup:
Agreed. 
This part needs to be extended to somewhat 0.13
https://github.com/pydata/xarray/blob/81cedd6e6d111c3ac58655585ffa6945fcc0e39b/xarray/tests/test_dataarray.py#L2040-L2041

shoyer:
> This part needs to be extended to somewhat 0.13

Yes, I already did this, in my original commit :)

PR #2813, opened 2019-03-14
added protect_dataset_variables_inplace to open_zarr

### LINKED ISSUE #2812: expose zarr caching from xarray
Zarr has its own internal mechanism for caching, described here:
- https://zarr.readthedocs.io/en/stable/tutorial.html#distributed-cloud-storage
- https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.LRUStoreCache

However, this capability is currently inaccessible from xarray.

I propose to add a new keyword `cache=True/False` to `open_zarr` which wraps the store in an `LRUStoreCache`.

rabernat:
This adds the same call to `_protect_dataset_variables_inplace` to `open_zarr` which we find in `open_dataset`. It wraps the arrays with `indexing.MemoryCachedArray`.

As far as I can tell, it *does not work*, in the sense that nothing is cached.

 - [ ] One possible way to close #2812 
 - [ ] Tests added

### DIFF
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -8,6 +8,7 @@ from ..core import indexing
 from ..core.pycompat import integer_types
 from ..core.utils import FrozenOrderedDict, HiddenKeyDict
 from .common import AbstractWritableDataStore, BackendArray
+from .api import _protect_dataset_variables_inplace
 
 # need some special secret attributes to tell us the dimensions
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@rabernat Given the age of this, is this still in the works?

kmuehlbauer:
Thanks @rabernat for the quick reply. I was skimming old issues/pr to see if there are some which might be closed.

PR #2814, opened 2019-03-14
[WIP] Use zarr internal LRU caching

### LINKED ISSUE #2812: expose zarr caching from xarray

rabernat:
Alternative way to close #2812. This uses zarr's own caching.

In contrast to #2813, this *does work*.

 - [ ] Closes #2812 
 - [ ] Tests added

### DIFF
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -8,6 +8,7 @@ from ..core import indexing
 from ..core.pycompat import integer_types
 from ..core.utils import FrozenOrderedDict, HiddenKeyDict
 from .common import AbstractWritableDataStore, BackendArray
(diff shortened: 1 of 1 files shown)


rabernat:
This was more of an experiment to see if it would work. Not sure if it is worth pursuing further. Thoughts @dcherian @jhamman?

PR #2828, opened 2019-03-20
Add quantile method to GroupBy

### LINKED ISSUE #3018: Add quantile method to groupby object
Dataset and DataArray objects have a quantile method, but not GroupBy. This would be useful for climatological analyses.

huard:
- [x] Tests added

Fixes #3018

Note that I've added an unrelated test that exposes an issue with grouping when there is only one element per group.


huard:
I just realized that I could do 
```da.groupby('time.month').apply(xr.DataArray.quantile, q=.5, dim='time')```
which seems to yield the same results.

**Comment by shoyer** [comment]
> I just realized that I could do
> `da.groupby('time.month').apply(xr.DataArray.quantile, q=.5, dim='time')`
> which seems to yield the same results.

Yes, I think this would be the preferred way to implement this :)

**Comment by huard** [comment]
Made the change. 
I had to drop the quantile dimension for scalar `q` after the apply to maintain the same behavior as `DataArray.quantile`, but it got me wondering if keeping the quantile dimension even for scalar `q` should not be the default.

File: xarray/core/groupby.py [hunk 1]
```diff
@@ -595,6 +595,55 @@ def _combine(self, applied, restore_coord_dims=False, shortcut=False):
...
+            the result corresponds to the quantile and a quantile dimension
+            is added to the return array. The other dimensions are the
+            dimensions that remain after the reduction of the array.
+
+        See Also
+        --------
+        numpy.nanpercentile, pandas.Series.quantile, Dataset.quantile,
+        DataArray.quantile
+        """
+        if dim is None:
```
dcherian: You'll have to copy the `DEFAULT_DIMS` and the warning from `reduce` below otherwise this wont be consistent with the other GroupBy methods.

xarray/tests/test_groupby.py
```diff
@@ -94,6 +94,18 @@ def func(arg1, arg2, arg3=0):
...
+@pytest.mark.xfail
+def test_da_groupby_single_value_per_dim():
+
+    array = xr.DataArray([[1, 1, 1], [2, 2, 2]],
+                         [('x', [1, 2]), ('y', [0, 1, 2])])
+
+    # This raises an error.
+    # I think the issue is that gr._group_indices is [0, 1]
+    # instead of [[0,], [1,]]
+    array.groupby('x').mean(dim='x')
```
dcherian: Don't you just want `array.groupby('x').mean()`? You're grouping over dimension 'x' and then attempting to take a mean along `x` which isn't a meaningful operation.

xarray/core/groupby.py
huard: Done.

xarray/tests/test_groupby.py
huard: In a case where you have multiple realization of a variable for the same coordinate, doesn't it make sense to compute the mean along the same dimension you grouped on ? I've added a few tests where we group by month, then compute the quantile over time (same dimension as grouping).

dcherian: Yeah but when you group by `x` in this example, each group does not have an `x` dimension. `groupby_bins` on `x` would be meaningful.

We should probably raise a more informative error.

huard: Should I remove this test from the PR ?

dcherian: I think so. You could open an issue for further discussion.

dcherian:
LGTM.

shoyer:
Thanks @huard !

Issue #2830, opened 2019-03-20
xarray 0.12.0 Fails To Import On Python 3.5.2


ancantus:
#### Error Code Example

Problem was discovered when xarray was pulled in via a pip dependency.

```python
import xarray
```
#### Problem description

xarray seems to now use `typing.ClassVar` which is unavailable in python 3.5.2

#### Output of ``import xarray``
<details>
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "$VENV_PATH/lib/python3.5/site-packages/xarray/__init__.py", line 26, in <module>
    from .coding.cftime_offsets import cftime_range
  File "$VENV_PATH/lib/python3.5/site-packages/xarray/coding/cftime_offsets.py", line 46, in <module>
    from typing import ClassVar, Optional
ImportError: cannot import name 'ClassVar'
</details>

#### Suggested Solution

- Remove ``typing.ClassVar`` from xarray
- Updated ``setup.py`` for 0.12.0 to reflect new python version requirement for the package.

max-sixty:
Thanks @ancantus! Sorry for that.

Python 3.5.3 does [support `ClassVar`](https://docs.python.org/3/library/typing.html#typing.ClassVar), which is how it passed our tests. That was [released](https://docs.python.org/3.5/whatsnew/changelog.html#python-3-5-3-final) 2017-01-17. So this shouldn't be too common.

We should have specified in `setup.py`. I'll do a PR now.

@shoyer do you think we need to do anything immediately re PyPI?

shoyer:
Thanks for the report. We should fix xarray on Python 3.5.0.

Probably the simplest thing to do would be to guard the import in a `typing.TYPE_CHECKING` block and replace `ClassVar[something]` in the code with `'ClassVar[something]'`. So you'll need at least Python 3.5.3 to do type checking of xarray, but it won't be required for users.

max-sixty:
Great, made that change. (my search didn't turn up any references to `ClassVar` outside the comments, so only did the guard)

ancantus:
Thanks! Came up for me cause the Ubuntu LTS Xenial uses Python 3.5.2 as its default installed version.

dcherian:
Closed by #2831

PR #2831, opened 2019-03-20
Enable python 3.5.0-3.5.2

max-sixty:
- [x] Closes some of https://github.com/pydata/xarray/issues/2830

### DIFF
--- a/xarray/coding/cftime_offsets.py
+++ b/xarray/coding/cftime_offsets.py
@@ -41,14 +41,17 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import re
+import typing
 from datetime import timedelta
 from functools import partial
-from typing import ClassVar, Optional
 
 import numpy as np
 
 from .cftimeindex import CFTimeIndex, _parse_iso8601_with_reso
 from .times import format_cftime_datetime
+
+if typing.TYPE_CHECKING:
+    from typing import ClassVar, Optional
 
 
 def get_date_type(calendar):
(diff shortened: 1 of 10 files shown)


max-sixty:
TBC, I don't think this would allow `3.5.2` to work from PyPI even if we did a point release - `3.5.2` would still run `0.12.0`

max-sixty:
From #2830:

> Probably the simplest thing to do would be to guard the import in a typing.TYPE_CHECKING block

Unfortunately that's only [supported from 3.5.2](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING)! 

I'll add a version check. I think there are only two places, and it won't last long

max-sixty:
I added a fix; it may have some mistakes which I'll correct

Is the approach OK? It's not pretty

xarray/coding/cftime_offsets.py
```diff
@@ -41,15 +41,20 @@
...
-from typing import ClassVar, Optional
 
 import numpy as np
 
 from .cftimeindex import CFTimeIndex, _parse_iso8601_with_reso
+from .core.pycompat import USE_TYPING
 from .times import format_cftime_datetime
 
+if USE_TYPING:
+    if typing.TYPE_CHECKING:
```
shoyer: Can mypy interpret this if we it this on one line? e.g., `if USE_TYPING and typing.TYPE_CHECKING` ?

max-sixty:
Green! Good idea re the single line, thanks @shoyer

max-sixty:
Please could someone else take a final check and merge? @shoyer reviewed but not since some changes. Thanks!

shoyer:
This looks good to me. I tested it locally and it works with mypy.

I pushed a small change to put a single `TYPE_CHECKING` variable in pycompat, so we can simply the check that we put in each file.

max-sixty:
> I pushed a small change to put a single TYPE_CHECKING variable in pycompat, so we can simply the check that we put in each file.

Much nicer!

PR #2642, opened 2019-01-02
Use pycodestyle for lint checks.

shoyer:
flake8 includes a few more useful checks, but it's annoying to only see it's
output in Travis-CI results.

This keeps Travis-CI and pep8speaks in sync.

PR #2832, opened 2019-03-20
Small fix: use == to compare strings

max-sixty:
`is` relies on interning, which is almost always OK for small strings, but reliant on language implementations

(unless there's something I'm missing about `.kind`?)

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -1597,7 +1597,7 @@ class Variable(common.AbstractArray, arithmetic.SupportsArithmetic,
                             "prior to calling this method.")
 
         axis = self.get_axis_num(dim)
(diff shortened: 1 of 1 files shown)


shoyer:
thanks, this is definitely the right thing to do

max-sixty:
I picked this and https://github.com/pydata/xarray/issues/2833 up b/c I have flake8 checking in my editor. 

I think we moved to restricting our automated checks to pycodestyle https://github.com/pydata/xarray/pull/2642 rather than the fuller flake8. I'm generally keen on more automated checks and less reliance on humans, but defer to @shoyer as ever

shoyer:
I think pep8speaks supports flake8 checks now, so that is a viable option. Previously it only did pycodestyle, and I didn't want to force people to read the build logs to get lint errors.

PR #2839, opened 2019-03-22
Fixing deepcopy for Dataset.attrs

kefirbandi:
- [x] Closes #2835
 - [X] Tests added

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -709,6 +709,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         indexes: 'Optional[OrderedDict[Any, pd.Index]]' = __default,
         encoding: Optional[dict] = __default,
         inplace: bool = False,
+        deep: bool = False,
     ) -> T:
         """Fastpath constructor for internal use.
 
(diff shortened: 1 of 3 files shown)


xarray/tests/test_dataset.py
```diff
@@ -1905,6 +1906,8 @@ def test_copy(self):
                 v1 = copied.variables[k]
                 assert v0 is not v1
 
+            assert data.attrs['Test'] is not copied.attrs['Test']
```
shoyer: can you also add a test for shallow copying?

xarray/core/dataset.py
```diff
@@ -709,6 +709,7 @@ def _replace(  # type: ignore
         indexes: 'Optional[OrderedDict[Any, pd.Index]]' = __default,
         encoding: Optional[dict] = __default,
         inplace: bool = False,
+        deep: bool = False,
```
shoyer: Rather than modifying `_replace`, could you use pass in `attrs` explicitly from `Dataset.copy()` itself?

I'd like to keep this method relatively simple -- it already has a large number of keyword arguments.

xarray/core/dataset.py
```diff
@@ -915,7 +915,9 @@ def copy(self: T, deep: bool = False, data: Mapping = None) -> T:
             variables = OrderedDict((k, v.copy(deep=deep, data=data.get(k)))
                                     for k, v in self._variables.items())
 
-        return self._replace(variables)
+        attrs = copy.deepcopy(self._attrs) if deep else copy.copy(self._attrs)
```
shoyer: I made one small fix, renaming `_attrs` to `attrs`.

As a general rule, we use the underscore prefix in Python as a convention for "private" but technically public variables. This isn't necessary for local variables defined within the scope of a function, since these are already private.

kefirbandi: Right. Actually I first called it "new_attrs" but it was 3 characters above PEP8 line-length limit, so I removed the first 3... :)
Thanks for the fix.

Issue #1598, opened 2017-09-28
Need better user control of _FillValue attribute in NetCDF files


mmartini-usgs:
This issue is under discussion here:  https://github.com/pydata/xarray/pull/1165

It is not desirable for us to have _FillValue = NaN for dimensions and coordinate variables.

In trying to use xarray, _FillValue was carefully kept from these variables and dimensions during the creation of the un-resampled file and then were found to appear during the to_netcdf operation. This happens in spite of mask_and_scale=False is being used with xr.open_dataset

I would hope that downstream code would have trouble with coordinates that don't make logical sense (time or place being NaN, for instance). We would prefer NOT to instantiate coordinate variable data with any fill value. Keeping NaNs out of coordinate variables, dimensions and minima and maxima is part of our QA/QC process to avoid downstream issues.

shoyer:
cc @thenaomig @laliberte 

There are at least two ways to fix this:
1. Support a flag of some sort in encoding (e.g., `_FillValue = False`) to indicate that fill value shouldn't be added. This would be easy to add, but is somewhat inelegant.
2. Check  for the presence of NaNs before setting `_FillValue = NaN`. This would be easy to add for dimension coordinates because they are already guaranteed to be in memory, but could cause performance trouble if any inputs are loaded as dask arrays. I don't know a satisfactory way to handle dask arrays with our current design, since we don't want to add another pass over the data to check for NaNs. I suppose one option would be to refactor our backend classes to write data before writing attributes and then make some sort of dask array operation that checks for NaNs as the data is written. But I'm not even sure this would work with the standard dask task schedulers.

mmartini-usgs:
There is also the philosophical problem of fill values for coordinate variables.  To be true to reality, one really would want to add an interpolated value that fills whatever gap or bad value exists.  That seems to be out of the scope of xarray though.  

I'm fine with a flag that controls only the coordinate data.  That said, for the rest of the variables, we avoid NaN in _FillValue.  We use 1E35.  So there you could give the user a choice in default fill value.  It seems pythonic to give the user flexibility.  And the minute you satisfy us, there will be another use case that comes along with conflicting requirements.  So you could use a flag and make it the user's choice, and not xarray's concern.

It also depends on where in the process one cleans up one's data - reduce first, then QA/QC, or QA/QC first, then reduce.  We do both, it depends on the instrument.

shoyer:
> There is also the philosophical problem of fill values for coordinate variables.

[...]

PR #2843, opened 2019-03-22
Allow passing _FillValue=False in encoding for vlen str variables.

DWesl:
The documentation seems to imply that passing _FillValue=False in the encoding works to set no fill value for any value.  These changes to `netCDF4_.py` and `h5netcdf_.py` seem to allow this for variables that are vlen strings (dtype `str` rather than `"S1"`): I have used the code-path in `netCDF4_.py` in real code and know that it at least allows the save to complete.

Allowing _FillValue=False makes it easier to explicitly exclude `_FillValue` from being written for coordinate variables, which some CF-compliance checkers complain about.

 - [ ] Closes #xxxx
 - [X] Tests added

### DIFF
--- a/xarray/backends/h5netcdf_.py
+++ b/xarray/backends/h5netcdf_.py
@@ -170,6 +170,8 @@ class H5NetCDFStore(WritableCFDataStore):
             variable, raise_on_invalid_encoding=check_encoding)
 
         fillvalue = attrs.pop('_FillValue', None)
+        if dtype is str and fillvalue is False:
+            fillvalue = None
         if dtype is str and fillvalue is not None:
             raise NotImplementedError(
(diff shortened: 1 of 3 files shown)


shoyer:
@DWesl thanks for your interest in improving xarray.

Where do we mention passing `_FillValue  = False` in the docs? This should probably be updated.

We made an intentional choice to support `None` instead of `False` to indicate "don't write a fill value": https://github.com/pydata/xarray/issues/1598

DWesl:
I just checked and can't find that section of the documentation now, so that seems to be consistent.

I suppose that's a vote for "be sure to check current behavior before submitting old packages".
I'll change my code to this new method then.

Thanks

Issue #1475, opened 2017-07-11
Allow DataArray to hold cell boundaries as coordinate variables


JiaweiZhuang:
Cell boundaries can be either N+1 sized arrays as suggested by xgcm/xmitgcm#15, or (N,2) sized arrays as suggested by [the CF convention](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-boundaries). However, a DataArray cannot hold both kinds of coordinate variables because they contain a new dimension.

If you try to assign a new coordinate to a DataArray by `dr.assign_coords()`, you will get
`ValueError: cannot add coordinates with new dimensions to a DataArray`

On the other hand, if your DataSet contains cell boundary variables (for example, #667), the bounds will be dropped when you extract a single variable into a DataArray.

Having cell bounds available in a DataArray is important for a couple of applications:

- Pass cell bounds to DataArray's plotting methods (xgcm/xmitgcm#15). I am aware of the discussion about inferring boundaries (#781). However, for the [Cube-Sphere grid](http://wiki.seas.harvard.edu/geos-chem/index.php/GEOS-Chem_HP#Cubed-sphere_Grid_Geometry) or the Lat-Lon-Cap grid ([reference](http://www.geosci-model-dev.net/8/3071/2015/)) which have tiles covering the poles, I have to explicitly pass cell bounds to the original plt.pcolormesh() to get a good-looking plot. (see [this comment](https://github.com/xgcm/xmitgcm/issues/15#issuecomment-314326358) for details)

- For conservative (i.e. area-weighted) regridding (mentioned in #486).  Cell centers are enough for bilinear interpolation or other simple resamping, but for any Finite-Volume meshes, knowing the boundaries is crucial if you want to conserve the total amount of mass or flux.

Plotting or regridding will work fine if you pass cell bounds as an additional argument to a wrapper function. However, having a single DataArray object containing boundary information seems like a more elegant solution. Is it possible to **let DataArray accept N+1 sized coordinate variables, and be able to inherit them from the parent DataSet**? If that's too drastic, is it possible to write an [accessor](http://xarray.pydata.org/en/stable/internals.html#extending-xarray) to extend DataArray's capability? Say, a "bound" accessor for a new attribute `ds.bnd['lat_b']`, which and be kept when a DataArray get extracted (`ds['data_var'].bnd['lat_b']` )? Does this make sense?

fmaussion:
[...]

PR #2844, opened 2019-03-22
Read grid mapping and bounds as coords

DWesl:
I prefer having these as coordinates rather than data variables.

This does not cooperate with slicing/pulling out individual variables.
`grid_mapping` should only be associated with variables that have
horizontal dimensions or coordinates.
`bounds` should stay associated despite having more dimensions.

I have not implemented similar functionality for the iris conversions.

An alternate approach to dealing with bounds (not used here) is to use a
`pandas.IntervalIndex`
http://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.IntervalIndex.html#pandas.IntervalIndex
and use where the coordinate is within its cell to determine on which
side the intervals are closed (`x_dim == x_dim_bnds[:, 0]` corresponds
to "left", `x_dim == x_dim_bnds[:, 1]` corresponds to "right", and
anything else is "neither").  This would stay through slicing and
might already be used for `.groupby_bins()`, but would not generalize
to boundaries of multidimensional coordinates unless someone
implements a multidimensional generalization of `pd.IntervalIndex`

 - [ ] Closes #xxxx
 - [X] Tests added


DWesl:
Related to #1475 and #2288 , but this is just keeping the metadata consistent where already present, not extending the data model to include bounds, cells, or projections.  I should add a test to ensure saving still works if the bounds are lost when pulling out variables.

xarray/conventions.py
```diff
@@ -543,6 +556,12 @@ def _encode_coordinates(variables, attributes, non_dim_coord_names):
                 variable_coordinates[k].add(coord_name)
                 global_coordinates.discard(coord_name)
 
+            att_val = v.attrs.get
+            if ((att_val("bounds", None) == coord_name or
+                 att_val("grid_mapping", None) == coord_name)):
```
shoyer: I'm also concerned that this is inconsistent with our other rules for CF encoding. In general, we only look at `encoding` for how to encode data on disk, not `attrs`. This suggests using `encoding={'bounds': 'something'}` as an indication that a coordinate should not be serialized as a  coordinate.

DWesl:
I can shift this to use encoding only, but I'm having trouble figuring out where that code would go.

Would the preferred path be to create VariableCoder classes for each and add them to encode_cf_variable, then add tests to xarray.tests.test_coding?

shoyer:
The current VariableCoder interface only works for coders that work at the level of individual variables. But coordinates only make sense on a dataset, so we'll need a different abstraction for that, e.g., a DatasetCoder?

For now, it's probably fine to keep this logic where you have it currently.

xarray/conventions.py
```diff
@@ -408,6 +408,18 @@ def stackable(dim):
                     new_vars[k].encoding['coordinates'] = coord_str
                     del var_attrs['coordinates']
                     coord_names.update(var_coord_names)
+            if 'bounds' in var_attrs:
+                bounds_str = var_attrs['bounds']
+                var_bounds_names = [bounds_str]
+                if all(k in variables for k in var_bounds_names):
+                    new_vars[k].encoding['bounds'] = bounds_str
```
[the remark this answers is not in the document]
DWesl: Shifted to be ``encoding`` only.

andreas-h:
I'd like this one to be merged very much.  Is there anything holding this back?


Also, it might be nice to update the documentation with info on this.

xarray/conventions.py
```diff
@@ -235,6 +235,11 @@ def encode_cf_variable(var, needs_copy=True, name=None):
     var = maybe_default_fill_value(var)
     var = maybe_encode_bools(var)
     var = ensure_dtype_not_object(var, name=name)
+
+    if var.encoding.get('grid_mapping', None) is not None:
```
shoyer: There's no need to supply a default value of `None` with `.get()`, that is already the default:
```suggestion
    if var.encoding.get('grid_mapping') is not None:
```

shoyer: But actually: are there any cases where someone would explicitly have `grid_mapping=None` in `encoding`? If not, then let's just check:
```suggestion
    if 'grid_mapping' in var.encoding:
```

xarray/conventions.py
```diff
@@ -543,6 +563,12 @@ def _encode_coordinates(variables, attributes, non_dim_coord_names):
                 variable_coordinates[k].add(coord_name)
                 global_coordinates.discard(coord_name)
 
+            encoding_val = v.encoding.get
+            if ((encoding_val('bounds', None) == coord_name or
+                 encoding_val('grid_mapping', None) == coord_name)):
```
shoyer: please fix as above

xarray/conventions.py
DWesl: Probably not.  That detail should go in documentation somewhere.  Where would you suggest?

dcherian:
I'm a little confused on why these are in encoding and not in attrs.

DWesl:
This is briefly mentioned above, in
https://github.com/pydata/xarray/pull/2844#discussion_r270595609
The rationale was that everywhere else xarray uses CF attributes for something, the original values of those attributes are recorded in `var.encoding`, not `var.attrs`, and consistency across a code base is a good thing.  Since I'm doing this primarily to get `grid_mapping` and `bounds` variables out of `ds.data_vars`, I don't have strong opinions on the subject.

If you feel strongly to the contrary, there's an idea at the top of this thread for getting `bounds` information encoded in terms xarray already uses in some cases (`Dataset.groupby_bins()`), and the diffs for this PR should help you figure out what needs changing to support this.  

I don't know how to get `grid_mapping` information into `Dataset`s cleanly, especially in cases with multiple `grid_mapping`s per `xarray.Dataset`

DWesl:
Switched to use `in` rather than `is not None`.

Re: `grid_mapping` in `.encoding` not `.attrs`
[MetPy assumes `grid_mapping` will be in `.attrs`](https://github.com/Unidata/MetPy/blob/master/metpy/xarray.py#L200).  Since the [xarray documentation mentions this capability](http://xarray.pydata.org/en/latest/weather-climate.html#cf-compliant-coordinate-variables), should I be making concurrent changes to MetPy to allow this to continue?

If so, would it be sufficient to change their `.attrs` references to `.encoding` and mentioning in both sets of documentation that the user should call `ds.metpy.parse_cf()` immediately after loading to ensure the information is available for MetPy to use?  I don't entirely understand the accessor API.

dcherian:
It isn't just MetPy though. I'm sure there's existing code relying on adding `grid_mapping` and `bounds` to `attrs` in order to write CF-compliant files. So there's a (potentially big) backward compatibility issue. This becomes worse if in the future we keep interpreting more CF attributes and moving them to `encoding` :/.

> Since I'm doing this primarily to get grid_mapping and bounds variables out of ds.data_vars.

I'm +1 on this but I wonder whether saving them in `attrs` and using that information when encoding coordinates would be the more pragmatic choice. 

We could define `encoding` as containing a specified set of CF attributes that control on-disk representation such as `units`, `scale_factor`, `contiguous` etc. and leaving everything else in `attrs`. A full list of attributes that belong in `encoding` could be in the docs so that downstream packages can fully depend on this behaviour.

Currently I see `coordinates` is interpreted and moved to `encoding`. In the above proposal, this would be left in `attrs` but its value would still be interpreted if `decode_coords=True`.

What do you think?

DWesl:
On 5/31/2019 11:50 AM, Deepak Cherian wrote:
> It isn't just MetPy though. I'm sure there's existing code relying on 
> adding |grid_mapping| and |bounds| to |attrs| in order to write 
> CF-compliant files. So there's a (potentially big) backward 
> compatibility issue. This becomes worse if in the future we keep 
> interpreting more CF attributes and moving them to |encoding| :/.

At present, the proper, CF-compliant way to do this is to have both 
|grid_mapping| and |bounds| variables in |data_vars|, and maintain the 
attributes yourself, including making sure the variables get copied into 
the result after relevant |ds[var_name]| and |ds.sel(axis=bounds)| 
operations.  If you decide to move these variables to |coords|, the 
|bounds| variables will still get dropped on any subsetting operation, 
including those where the relevant axis was retained, and the 
|grid_mapping| variables will be included in the result of all 
subsetting operations, and both will be included in some |coordinates| 
attribute when written to disk, breaking CF compliance.
> 
>     Since I'm doing this primarily to get grid_mapping and bounds
>     variables out of ds.data_vars.
> 
> I'm +1 on this but I wonder whether saving them in |attrs| and using 
> that information when encoding coordinates would be the more pragmatic 
> choice.

You have a point about |grid_mapping|, but applying the MetPy approach 
of saving the information in another, more directly useful format 
(|cartopy.Projection| instances) immediately after loading the file 
would be a way around that.

For |bounds|, I think |pd.PeriodIndex| would be the most natural 
representation in most 1D cases, but that still leaves |bounds| for 
two-or-more-dimensional coordinates.

That's a design choice I'll leave to the maintainers.
> 
> We could define |encoding| as containing a specified set of CF 
> attributes that control on-disk representation such as |units|, 
> |scale_factor|, |contiguous| etc. and leaving everything else in 
> |attrs|. A full list of attributes that belong in |encoding| could be in 
> the docs so that downstream packages can fully depend on this behaviour.
> 
> Currently I see |coordinates| is interpreted and moved to |encoding|. In 
> the above proposal, this would be left in |attrs| but its value would 
> still be interpreted if |decode_coords=True|.
> 
> What do you think?

At present, |set(ds[var_name].attrs["coordinates"].split())| and
|set(ds[var_name].coords) - set(ds[var_name].indexes[dim_name])|
would be identical, since the |coordinates| attribute is essentially 
computed from the second expression on write.

Do you have a use case in mind where you need specifically the list of 
CF auxiliary coordinates, or is that just an example of something that 
would change under the new proposal?  I assume |units| would be moved to 
|encoding| only for |datetime64[ns]| and |timedelta64[ns]| variables.

xarray/conventions.py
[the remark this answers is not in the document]
DWesl: On the other hand, `attrs` seems to stick with the data through more operations, while `encoding` is dropped quickly; this would indicate that `bounds` in `encoding` is for how we save to disk only, while `bounds` in `attrs` is for remembering the relationship while manipulating the data in memory.

Alternately, the mess around where `bounds` should end up could indicate that `bounds` should be dealt with be decoding to a `pd.IntervalIndex` rather than trying to associate the bounds with the coordinate.  Out of curiosity, does `xarray` have a multidimensional generalization of `pd.IntervalIndex` for multidimensional bounds?

DWesl:
I just noticed [pandas.PeriodIndex](https://pandas.pydata.org/docs/user_guide/timeseries.html#time-span-representation) would be an alternative to [pandas.IntervalIndex](https://pandas.pydata.org/docs/reference/api/pandas.IntervalIndex.html#pandas.IntervalIndex) for time data if which side the interval is closed on is largely irrelevant for such data.

Is there an interest in using these for 1D coordinates with bounds?  I think `ds.groupby_bins()` already returns an `IntervalIndex`.

DWesl:
The test failures seem to all be due to recent changes in `cftime`/`CFTimeIndex`, which I haven't touched.  

Is sticking the `grid_mapping` and `bounds` attributes in `encoding` good, or should I put them back in `attrs`?

dcherian:
My preference is for attrs but someone else should weigh in. cc @pydata/xarray 

Maybe @dopplershift , as MetPy maintainer has thoughts on this matter?

dopplershift:
As a downstream user, I just want to be told what to do (assuming `encoding` is part of the public API for xarray). I'd love not to have to modify our code, but that's not essential necessarily.

So to clarify: is this about whether they should be in one spot or the other? Or is it about having `grid_mapping` and `bounds` in both?

DWesl:
I think the choice is between `attrs` and `encoding`, not both.  

If it helps lean your decision one way or the other, `attrs` tends to stay associated with `Dataset`s through more operations than `encoding`, so `parse_cf()` would have to be called fairly soon after opening if the information ends up in `encoding`, while putting it in `attrs` gives users a bit more time for that.

dopplershift:
Thanks for the info. Based on that, I lean towards `attrs`.

I think a better rationale, though, would be to formalize the role of `encoding` in xarray.

shoyer:
We could probably make it a rule that encoding gets preserved in exactly
the same way as attrs. I think there are some old issues about formalizing
metadata conventions...

DWesl:
Should I change this to put `grid_mapping` and `bounds` back in `attrs`, or should I leave them in `encoding`?

xarray/conventions.py
DWesl: Done.

xarray/conventions.py
DWesl: Code change is done.  Docs change is not.  What page should that go on? IO/NetCDF?

xarray/conventions.py
DWesl: Moved to `encoding`

xarray/conventions.py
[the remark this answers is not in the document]
DWesl: The CF attributes are now in `encoding` only.

DWesl:
This PR currently puts `grid_mapping` and `bounds` in `encoding` once it is done with them.  Is that where XArray wants to put them, or should they be somewhere else?

dcherian:
Hi @DWesl . I sincerely apologize for not responding earlier.

Reading over this thread again, I think I agree with @shoyer these should live in `encoding`, because we've technically "decoded" the attribute and converted these variables to coords.

xarray/conventions.py
```diff
@@ -502,6 +507,20 @@ def stackable(dim):
...
+                bounds_str = var_attrs["bounds"]
+                var_bounds_names = [bounds_str]
+                if all(k in variables for k in var_bounds_names):
+                    new_vars[k].encoding["bounds"] = bounds_str
+                    coord_names.update(var_bounds_names)
+                del var_attrs["bounds"]
+            if "grid_mapping" in var_attrs:
+                proj_str = var_attrs["grid_mapping"]
+                var_proj_names = proj_str.split()
+                if all(k in variables for k in var_proj_names):
```
dcherian, line 520: if the attribute mentions variables that doesn't exist, shall we raise a warning about the nonexistent variables and convert the others to coords?

xarray/conventions.py
```diff
@@ -502,6 +507,20 @@ def stackable(dim):
                     new_vars[k].encoding["coordinates"] = coord_str
                     del var_attrs["coordinates"]
                     coord_names.update(var_coord_names)
+            if "bounds" in var_attrs:
+                bounds_str = var_attrs["bounds"]
+                var_bounds_names = [bounds_str]
+                if all(k in variables for k in var_bounds_names):
```
dcherian, line 513: this is always a 1 element list. Can we just say `if bounds_str in variables`?

xarray/conventions.py
```diff
@@ -502,6 +507,20 @@ def stackable(dim):
                     new_vars[k].encoding["coordinates"] = coord_str
                     del var_attrs["coordinates"]
                     coord_names.update(var_coord_names)
+            if "bounds" in var_attrs:
```
dcherian, line 510: This new behaviour is controlled by `decode_coords` which is True by default. To deal with backward compatibility, should we add `decode_aux_vars` that controls whether these auxiliary variables are set as coords?

xarray/conventions.py
```diff
@@ -661,6 +681,14 @@ def _encode_coordinates(variables, attributes, non_dim_coord_names):
             ):
                 variable_coordinates[k].add(coord_name)
 
+            encoding_val = v.encoding.get
+            if (
+                encoding_val("bounds") == coord_name
```
dcherian, line 686: nit: Mild preference for writing out `v.encoding.get("bounds")`

xarray/conventions.py
```diff
@@ -254,6 +254,11 @@ def encode_cf_variable(var, needs_copy=True, name=None):
     var = maybe_default_fill_value(var)
     var = maybe_encode_bools(var)
     var = ensure_dtype_not_object(var, name=name)
+
+    if "grid_mapping" in var.encoding:
```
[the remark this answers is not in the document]
DWesl: Done.

xarray/conventions.py
DWesl: Added the warning.

xarray/conventions.py
DWesl: Done.

xarray/conventions.py
DWesl: Done.

xarray/conventions.py
DWesl: I feel like that question is above my pay grade.

Would that also cover `ancillary_variables` and `cell_measures` if someone decides they want those to be coords instead of data vars?

DWesl:
The example the doc build doesn't like:
```python
ds = xr.tutorial.load_dataset("rasm")
ds.to_zarr("rasm.zarr", mode="w")
import zarr

zgroup = zarr.open("rasm.zarr")
print(zgroup.tree())
dict(zgroup["Tair"].attrs)
```
I'll need to look into the `rasm` dataset to figure out why there is a warning now.

DWesl:
The `rasm` dataset has coordinates `xc` and `yc`, which reference bounds `xv` and `yv`, which I do not see in the variable list with `decode_coords=False`.  It would appear that pydata/xarray-data#4 did not include the bounds in the updated dataset when adding coordinates to `rasm.nc`, so this warning is correct.  I do not know that file, so I'm probably not the best person to add bounds.

xarray/conventions.py
DWesl: [There's also a `decode_cf` argument to `open_dataset`](https://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html#xarray.open_dataset), which is another possibility for the argument controlling the behavior.

I have the loop over attributes I think are lists of variables, including `ancillary_variables` and `cell_measures`.

dcherian:
@DWesl can you put delete the "bounds" attribute and send in a PR to xarray-data?

DWesl:
pydata/xarray-data#19

xarray/conventions.py
dcherian: I agree that `decode_cf` should control this since we are decoding CF attributes.

DWesl:
The MinimumVersionsPolicy error appears to be a series of internal `conda` errors, and is probably unrelated.

xarray/conventions.py
DWesl: On further inspection, [`decode_cf` appears to be a quick way to control `mask_and_scale`, `decode_coords`, `decode_times`, `decode_timedelta`, and `concat_characters`](https://github.com/DWesl/xarray/blob/ca0f80523ef84f9159c0eebf2580354dd7459d75/xarray/backends/api.py#L437), so, in a sense, the new behavior can already be turned off by `decode_cf`.

Whether the new behavior should remain under the control of the existing `decode_coords` or a new `decode_aux_vars`.
This is primarily sticking things in `ds.coords` if the file says they should be, the difference is whether to look only at the `coordinates` attribute or at more attributes from the CF standard.

DocOtak:
I decided to try out this PR on some of the data files we are working with at my data office. In our datasets we have per variable quality flag information per variable uncertainty information. The CF way of tying all these together is via the [ancillary_variables attribute](http://cfconventions.org/cf-conventions/cf-conventions.html#ancillary-data). This PR pulls all these out into the Dataset coordinates. Since in the xarray data model (right now) the coordinates apply to an entire dataset, this feels inappropriate and maybe even breaking. The ancillary_variables attribute is not used in CF grid mapping or bounds as far as I can tell.

Here is an example using this PR (note that all the varN type names will be replaced with better variable names before we publish these):

```python
In [1]: import xarray as xr

In [2]: ds = xr.open_dataset("examples/converted/06AQ19840719.nc")

In [3]: ds
Out[3]:
<xarray.Dataset>
Dimensions:    (N_LEVELS: 24, N_PROF: 38)
Coordinates:
    var1_qc    (N_PROF, N_LEVELS) float32 ...
    var4_qc    (N_PROF, N_LEVELS) float32 ...
    var5_qc    (N_PROF, N_LEVELS) float32 ...
    var6_qc    (N_PROF, N_LEVELS) float32 ...
    var7_qc    (N_PROF, N_LEVELS) float32 ...
    var8_qc    (N_PROF, N_LEVELS) float32 ...
    var9_qc    (N_PROF, N_LEVELS) float32 ...
    var10_qc   (N_PROF, N_LEVELS) float32 ...
    var11_qc   (N_PROF, N_LEVELS) float32 ...
    var12_qc   (N_PROF, N_LEVELS) float32 ...
    var13_qc   (N_PROF, N_LEVELS) float32 ...
    var14_qc   (N_PROF, N_LEVELS) float32 ...
    var15_qc   (N_PROF, N_LEVELS) float32 ...
    pressure   (N_PROF, N_LEVELS) float64 ...
    latitude   (N_PROF) float64 ...
    longitude  (N_PROF) float64 ...
    time       (N_PROF) datetime64[ns] ...
    expocode   (N_PROF) object ...
    station    (N_PROF) object ...
    cast       (N_PROF) int8 ...
    sample     (N_PROF, N_LEVELS) object ...
Dimensions without coordinates: N_LEVELS, N_PROF
Data variables:
    var0       (N_PROF) object ...
    var1       (N_PROF, N_LEVELS) object ...
    var2       (N_PROF) float32 ...
    var3       (N_PROF, N_LEVELS) float32 ...
    var4       (N_PROF, N_LEVELS) float32 ...
    var5       (N_PROF, N_LEVELS) float32 ...
    var6       (N_PROF, N_LEVELS) float32 ...
    var7       (N_PROF, N_LEVELS) float32 ...
    var8       (N_PROF, N_LEVELS) float32 ...
    var9       (N_PROF, N_LEVELS) float32 ...
    var10      (N_PROF, N_LEVELS) float32 ...
    var11      (N_PROF, N_LEVELS) float32 ...
    var12      (N_PROF, N_LEVELS) float32 ...
    var13      (N_PROF, N_LEVELS) float32 ...
    var14      (N_PROF, N_LEVELS) float32 ...
    var15      (N_PROF, N_LEVELS) float32 ...
Attributes:
    Conventions:  CF-1.8 CCHDO-0.1.dev157+g52933e0.d20200707
```

This looks especially confusing when you ask for one specific variable:

```python
In [15]: ds.var6
Out[15]:
<xarray.DataArray 'var6' (N_PROF: 38, N_LEVELS: 24)>
array([[33.3965, 33.5742, 34.8769, ..., 34.9858, 34.9852,     nan],
       [33.1129, 34.0742, 34.6595, ...,     nan,     nan,     nan],
       [32.5328, 33.2687, 34.2262, ...,     nan,     nan,     nan],
       ...,
       [35.0686, 35.09  , 35.1415, ...,     nan,     nan,     nan],
       [35.0303, 35.0295, 35.0715, ...,     nan,     nan,     nan],
       [35.0682, 35.0756, 35.0622, ...,     nan,     nan,     nan]],
      dtype=float32)
Coordinates:
    var1_qc    (N_PROF, N_LEVELS) float32 ...
    var4_qc    (N_PROF, N_LEVELS) float32 ...
    var5_qc    (N_PROF, N_LEVELS) float32 ...
    var6_qc    (N_PROF, N_LEVELS) float32 ...
    var7_qc    (N_PROF, N_LEVELS) float32 ...
    var8_qc    (N_PROF, N_LEVELS) float32 ...
    var9_qc    (N_PROF, N_LEVELS) float32 ...
    var10_qc   (N_PROF, N_LEVELS) float32 ...
    var11_qc   (N_PROF, N_LEVELS) float32 ...
    var12_qc   (N_PROF, N_LEVELS) float32 ...
    var13_qc   (N_PROF, N_LEVELS) float32 ...
    var14_qc   (N_PROF, N_LEVELS) float32 ...
    var15_qc   (N_PROF, N_LEVELS) float32 ...
    pressure   (N_PROF, N_LEVELS) float64 ...
    latitude   (N_PROF) float64 ...
    longitude  (N_PROF) float64 ...
    time       (N_PROF) datetime64[ns] ...
    expocode   (N_PROF) object ...
    station    (N_PROF) object ...
    cast       (N_PROF) int8 ...
    sample     (N_PROF, N_LEVELS) object ...
Dimensions without coordinates: N_PROF, N_LEVELS
Attributes:
    whp_name:         CTDSAL
    whp_unit:         PSS-78
    standard_name:    sea_water_practical_salinity
    units:            1
    reference_scale:  PSS-78
```

DWesl:
`ancillary_variables` is neither `grid_mapping` or `bounds`.  

My personal view is that the quality information should stay with the variable it describes unless explicitly dropped; I think your view is that quality information can always be extracted from the original dataset, and that no variable should carry quality information for a different variable.  At this point it would be simple to remove `ancillary_variables` from the attributes processed by this PR.

I should point out that a similar situation arises for `grid_mapping`; `ds.coords["time"]` will include the `grid_mapping` variable in its coordinates.
In contrast, `ds.coords["x"]` will not include the bounds for the `x` variable, since it has more dimensions than `ds.coords["x"]`

DocOtak:
Yes, my view is that things in `ancillary_variables` should stay in the attrs of their variable (DataArray) and not moved to coords. Currently this PR will remove the `ancillary_variables` from the attrs of the variables in a file which have it. This appears to break CF defined connection between associated variables (like uncertainty and QC). While the information isn't lost, I would need to look in encoding to get it. It looks like the [first reply](https://github.com/pydata/xarray/issues/4215#issuecomment-656789017) in #4215 also didn't like putting `ancillary_variables` in the coords.

What would be really awesome is some sort of variable proxy I could replace the string names with actual references/pointers to the correct DataArray in the Dataset.

DWesl:
That's two people with that view so I made the change.

Again, I feel that the quality flags are essentially meaningless on their own, useful primarily in the context of their associated variables, like the items currently put in the XArray `coords` attribute, which, admittedly, is only those variables identified by CF as dimension or auxiliary coordinates at the moment, and should remain associated with the relevant variable even if it is extracted into a `DataArray`.  Since all of the other people who have opinions on the matter seem to disagree with me, I changed the code to preserve the present behavior with regards to `ancillary_variables`.  I can always monkey-patch it back in if it really bothers me, or add a `__getitem__` wrapper to `xarray-contrib/cf-xarray` so that the `ancillary_variables` stay associated when I pull variables out, or move back to `SciTools/iris`.

On a related note, I should probably check whether this breaks conversion to an `iris.Cube`.

jthielen:
For what it's worth, here's a bit of an elaboration on my view from https://github.com/pydata/xarray/issues/4215#issuecomment-656789017 and https://github.com/pydata/xarray/issues/4121#issuecomment-639758583:

`ancillary_variables` really fit a role of "non-coordinate linked variables," which doesn't have a good place in xarray's current data model. Both the current behavior (any links are completely lost if not manually handled) and linking as coordinates (conceptually confusing and [causing spurious proliferation](https://github.com/pydata/xarray/pull/2844#issuecomment-670808763)) are problematic. Short of a "linked" or "subsidiary" variable concept being added to xarray's data model (could be worth discussing on a new issue?), `cf-xarray` seems like the best place for managing `ancillary_variables`, either through
- A method that extracts `ancilliary_variables` from a Dataset given a variable name (https://github.com/xarray-contrib/cf-xarray/issues/5#issue-628569748)
- Parsing the `ancilliary_variables` attribute and replacing the string with a Dataset or list of DataArrays or Variables containing the actual ancillary variables (https://github.com/pydata/xarray/issues/4121#issue-630573329 and https://github.com/pydata/xarray/pull/2844#issuecomment-670991386)
- A method or addition to `__getitem__` on the accessor returning a Dataset with the linked variables included (https://github.com/pydata/xarray/pull/2844#issuecomment-670996109)

For those interested, the issue for `ancillary_variables` in cf-xarray is https://github.com/xarray-contrib/cf-xarray/issues/5.

dcherian:
I think it's OK if we skip `ancillary_variables`, the CF data model does not map perfectly to xarray's data model so this kind of annoyance will always be present. Also thanks @DocOtak for that pathological example :D (now I know what those mathematicians talk about)

> A method or addition to __getitem__ on the accessor returning a Dataset with the linked variables include

This has already been implemented though it could use a lot more testing: https://cf-xarray.readthedocs.io/en/latest/examples/introduction.html#Feature:-Accessing-variables-by-standard-names

doc/whats-new.rst
```diff
@@ -33,6 +33,10 @@ New Features
   of :py:class:`DataArray` and :py:class:`Dataset` objects and
   document the new method in :doc:`internals`. (:pull:`4248`).
   By `Justus Magin <https://github.com/keewis>`_.
+- Decode the CF attributes 'grid_mapping' and 'bounds' on file read
```
dcherian, line 36: ~should update this list.~ Let's add a small para with the whole list to https://xarray.pydata.org/en/stable/weather-climate.html and link to that from here.

Also please take credit by adding your name and github ID if you would like.

xarray/conventions.py
```diff
@@ -506,6 +526,31 @@ def stackable(dim):
                     new_vars[k].encoding["coordinates"] = coord_str
                     del var_attrs["coordinates"]
                     coord_names.update(var_coord_names)
+            for attr_name in CF_RELATED_DATA:
+                if attr_name in var_attrs:
```
dcherian, line 530: Should we raise a warning here saying that we've started parsing these attributes and the folowing variables ... have been set as coords. ?

I'm thinking about backwards compatibility...

DWesl: I added a warning about variables moving from `data_vars` to `coords`

doc/whats-new.rst
DWesl: I added a paragraph, and I think I added a link.

xarray/conventions.py
DWesl: I just changed the `stacklevel` so the warning would point to the `xarray.open_dataset` call, not the `cf_decode` call in `decode_cf`.

doc/weather-climate.rst
```diff
@@ -10,6 +10,28 @@ Weather and climate data
...
+- `node_coordinates`
+- `node_count`
+- `part_node_count`
+- `interior_ring`
+- `cell_measures`
+- `formula_terms`
+
+The CF attribute `ancillary_variables` was not included in the list
+due to the variables listed there being associated primarily with the
+variable with the attribute, rather than with the dimensions.
```
DWesl, line 33:
```suggestion
variable with the attribute, rather than with the dimensions 
associated with that variable.
```

doc/whats-new.rst
```diff
@@ -276,7 +280,6 @@ New Features
 - Expose ``use_cftime`` option in :py:func:`~xarray.open_zarr` (:issue:`2886`, :pull:`3229`)
   By `Samnan Rahee <https://github.com/Geektrovert>`_ and `Anderson Banihirwe <https://github.com/andersy005>`_.
 
-
```
DWesl, line 279: I didn't mean to delete the extra newline here.  Should I put it back in?
```suggestion

```

xarray/tests/test_backends.py
```diff
@@ -857,6 +858,105 @@ def test_roundtrip_mask_and_scale(self, decoded_fn, encoded_fn):
...
+    @requires_iris
+    def test_grid_mapping_and_bounds_are_coordinates_after_iris_roundtrip(self):
+        original = self._create_cf_dataset()
+        iris_cube = original["variable"].to_iris()
+        actual = DataArray.from_iris(iris_cube)
+        # Bounds will be missing (xfail)
+        del original.coords["latitude_bnds"], original.coords["longitude_bnds"]
+        # Ancillary vars will be missing
+        # Those are data_vars, and will be dropped when grabbing the variable
+        assert_identical(actual, original["variable"])
```
DWesl, line 958: Should I create more tests to see whether converting a DataArray to an iris Cube produces the same results as writing to disk and reading with iris?  Or test creating an iris Cube with bounds on coordinates to check whether conversion works the same as writing to disk and reading with XArray?

DWesl (commented):
I updated the PR to work with current master.  I don't remember what else needed to be done.

doc/whats-new.rst
```diff
@@ -35,6 +35,10 @@ New Features
 ~~~~~~~~~~~~
 - Performance improvement when constructing DataArrays. Significantly speeds up repr for Datasets with large number of variables.
   By `Deepak Cherian <https://github.com/dcherian>`_
+- Decode more CF attributes on file read (put the referenced variables
```
dcherian, line 38: Let's move this to "breaking changes" and add a little more detail. Maybe something like

```
Variables referred to in attributes like `bounds` and `grid_mapping` are now automatically set as coordinate variables. These attributes are moved to :py:attr:`DataArray.encoding` from :py:attr:`DataArray.attrs`. This behaviour is controlled by the `decode_coords` kwarg to :py:func:`open_dataset` and :py:func:`open_mfdataset`
```

doc/whats-new.rst
dcherian: It's OK

doc/weather-climate.rst
```diff
@@ -10,6 +10,28 @@ Weather and climate data
...
+- `climatology`
+- `geometry`
+- `node_coordinates`
+- `node_count`
+- `part_node_count`
+- `interior_ring`
+- `cell_measures`
+- `formula_terms`
+
+The CF attribute `ancillary_variables` was not included in the list
```
dcherian, line 31:
```suggestion
This decoding is controlled by the `decode_coords` kwarg to :py:func:`open_dataset` and :py:func:`open_mfdataset`. 
The CF attribute `ancillary_variables` was not included in the list
```

xarray/conventions.py
```diff
@@ -508,6 +528,39 @@ def stackable(dim):
...
+                    attr_val = var_attrs[attr_name]
+                    var_names = attr_val.split()
+                    if attr_name in CF_RELATED_DATA_NEEDS_PARSING:
+                        var_names = [
+                            name
+                            for name in var_names
+                            if not name.endswith(":") and not name == k
```
dcherian, line 539: This would break if `attrs["cell_measures"] = "area : cell_area volume : cell_volume"`. I don't know if that version is CF-compliant, so maybe we shouldn't worry about it.

PS: cf-xarray does this: https://github.com/xarray-contrib/cf-xarray/blob/5133a4dada23b18221f2136e1f763cda8e65f249/cf_xarray/utils.py#L52-L69

xarray/conventions.py
```diff
@@ -508,6 +528,39 @@ def stackable(dim):
...
+                        ]
+                    if all(k in variables for k in var_names):
+                        new_vars[k].encoding[attr_name] = attr_val
+                        coord_names.update(var_names)
+                        # Warn that some things will be coords rather
+                        # than data_vars.
+                        warnings.warn(
```
dcherian, line 546: I think we can avoid this warning.

xarray/conventions.py
```diff
@@ -508,6 +528,39 @@ def stackable(dim):
...
+                            "Variable(s) {0!s} moved from data_vars to coords\n"
+                            "based on {1:s} attribute".format(var_names, attr_name),
+                            stacklevel=5,
+                        )
+                    else:
+                        warnings.warn(
+                            "Variable(s) referenced in {0:s} not in variables: {1!s}".format(
```
dcherian, line 553: Can we use f-strings here please?

xarray/tests/test_backends.py
```diff
@@ -857,6 +858,109 @@ def test_roundtrip_mask_and_scale(self, decoded_fn, encoded_fn):
...
+        original["variable"].encoding.update(
+            {"cell_measures": "area: areas", "grid_mapping": "latlon"},
+        )
+        original.coords["latitude"].encoding.update(
+            dict(grid_mapping="latlon", bounds="latitude_bnds")
+        )
+        original.coords["longitude"].encoding.update(
+            dict(grid_mapping="latlon", bounds="longitude_bnds")
+        )
+        original.coords["ln_p"].encoding.update({"formula_terms": "p0: P0 lev: ln_p"})
```
dcherian, line 912: Let's also test that `areas`, `P0` and `ln_p` are in `.coords` when `decode_coords=True`.

xarray/conventions.py
```diff
@@ -508,6 +528,39 @@ def stackable(dim):
                     new_vars[k].encoding["coordinates"] = coord_str
                     del var_attrs["coordinates"]
                     coord_names.update(var_coord_names)
+            for attr_name in CF_RELATED_DATA:
```
dcherian, line 531: We should update the docstring for `decode_coords` saying that it controls this behaviour.

doc/conf.py
```diff
@@ -34,7 +34,7 @@
     subprocess.run(["conda", "list"])
 else:
     print("pip environment:")
-    subprocess.run(["pip", "list"])
+    subprocess.run([sys.executable, "-m", "pip", "list"])
```
dcherian, line 37: Is this needed? cc @keewis

dcherian (commented):
Thanks @DWesl, just some minor comments

xarray/tests/test_backends.py
dcherian: That would be nice but could be done in a future PR.

xarray/conventions.py
DWesl: The examples do not have a space between the measure and the colon, but it isn't clear if that is allowed.
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#cell-measures
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#parametric-v-coord

I added a test for that case in `test_backends` (`formula_terms` set to `p0: P0 lev : ln_p` to test both cases at once).

doc/conf.py
keewis: for the CI this is not necessary, but it might help when building locally. For the difference between `python -m pip` and `pip` see this [blog post](https://snarky.ca/why-you-should-use-python-m-pip/).

DWesl: If you have python2.7 installed alongside pythons 3.6 through 3.9, `pip` will likely default to python 2.7 or python 3.6 unless you're in a virtual environment.  In a virtual environment, there is no difference.  When using system python, with more than one version of python available, this ensures the packages available for import in the current python session are listed.

xarray/conventions.py
DWesl: Somewhat updated.  I'm not sure if I should list the attributes or point to the list of attributes.

xarray/tests/test_backends.py
DWesl: Does `assert_identical` not already take care of this?  The actual tests are the two methods below, this just returns a dataset that can be used for the tests.

xarray/conventions.py
DWesl: Changed to f-string.  It's a long line.

xarray/conventions.py
DWesl: Removed.

doc/whats-new.rst
DWesl: Moved to breaking changes.

doc/weather-climate.rst
DWesl: Added.

xarray/tests/test_backends.py
DWesl: Okay, I'll leave that for now.

doc/whats-new.rst
```diff
@@ -29,6 +29,13 @@ Breaking changes
   values were required, which would lead to inaccurate I/O round-trips.
 - remove deprecated ``autoclose`` kwargs from :py:func:`open_dataset` (:pull: `4725`).
   By `Aureliana Barghini <https://github.com/aurghs>`_
+- Variables referred to in attributes like `bounds` and `grid_mapping`
```
dcherian, line 32:
```suggestion
- Variables referred to in attributes like ``bounds`` and ``grid_mapping``
```

xarray/conventions.py
```diff
@@ -543,8 +593,8 @@ def decode_cf(
         Decode cf times (e.g., integers since "hours since 2000-01-01") to
         np.datetime64.
     decode_coords : bool, optional
-        Use the 'coordinates' attribute on variable (or the dataset itself) to
-        identify coordinates.
+        Use the 'coordinates', 'grid_mapping', and 'bounds' attributes
```
dcherian, line 596:
```suggestion
        Use the 'coordinates', 'grid_mapping', 'bounds' and other attributes
```

xarray/conventions.py
```diff
@@ -543,8 +593,8 @@ def decode_cf(
...
+        on variables (or the dataset itself) to identify coordinates.
```
dcherian, line 597:
```suggestion
        on variables (or the dataset itself) to identify coordinate variables.
```

Let's also link to the full list if we can.

xarray/tests/test_backends.py
```diff
@@ -857,6 +858,106 @@ def test_roundtrip_mask_and_scale(self, decoded_fn, encoded_fn):
...
+        original = self._create_cf_dataset()
+        with create_tmp_file() as tmp_file:
+            original.to_netcdf(tmp_file)
+            with open_dataset(tmp_file, decode_coords=False) as ds:
+                assert ds.coords["latitude"].attrs["bounds"] == "latitude_bnds"
+                assert ds.coords["longitude"].attrs["bounds"] == "longitude_bnds"
+                assert "latlon" not in ds["variable"].attrs["coordinates"]
+                assert "coordinates" not in ds.attrs
+
+    def test_grid_mapping_and_bounds_are_coordinates_after_dataset_roundtrip(self):
```
dcherian, line 925:
```suggestion
    def test_coordinate_variables_after_dataset_roundtrip(self):
```

xarray/conventions.py
dcherian: Let's link to the list if we can.

xarray/tests/test_backends.py
dcherian: Yes you are correct! Sorry for the noise.

dcherian (approved):
Thanks @DWesl just pushed a small docs commit. I think this is ready to merge. Thank you for your patience.

DWesl:
Looks good to me.  I was wondering where those docstrings were.

jthielen:
I unfortunately have not been following along with the recent developments in this PR, so these may have already been previously covered. Sorry if that is the case!

However, earlier on in the development of this PR, there were some substantial concerns about backwards compatibility (e.g., for libraries like MetPy that currently rely on `grid_mapping` and the like being in `attrs`) and the improved propagation of `encoding` through operations (so that moving these to `encoding` doesn't mean they are unnecessarily lost). What is the current status with regards to these?

dcherian:
> there were some substantial concerns about backwards compatibility 

:( yes this is currently backward incompatible (so the next release will bump major version). There's reluctance to add yet another `decode_*` kwarg to `open_dataset`, and also reluctance to issue a warning saying that the behaviour of `decode_coords` will change in a future version (since this would be very common).

One option is to change `decode_coords` from `bool` to `Union[bool, str]` and allow `decode_coords` to be either `"all"` (this PR) or `"coordinates"`, or `True` ( == `"coordinates"`, backwards compatible). I can bring this up at the next dev meeting.

dcherian:
> One option is to change decode_coords from bool to Union[bool, str] and allow decode_coords to be either "all" (this PR) or "coordinates", or True ( == "coordinates", backwards compatible). I can bring this up at the next dev meeting.

Updated to implement this proposal after getting OK at the previous meeting.

@DWesl can you take a look at the most recent changes please?

DWesl:
I think this looks good.

dcherian:
Great I'll merge before the next release if no one else has comments. Thanks @DWesl

doc/weather-climate.rst
andersy005: @DWesl, do you still want to commit this suggestion?

andersy005 (approved):
This looks great! 

~~Does anyone know why the `xr.open_dataset(....)` call is echoed in the warning message. Is this intentional? Cc @dcherian @DWesl~~

```python
In [4]: ds = xr.open_dataset('/home/abanihi/Downloads/pr_Amon_ACCESS-ESM1-5_historical_r1i1p1f1_gn_201001-201412.nc', decode_c
   ...: oords="all")
<ipython-input-4-ccf90ed4a433>:1: UserWarning: Variable(s) referenced in cell_measures not in variables: ['areacella']
  ds = xr.open_dataset('/home/abanihi/Downloads/pr_Amon_ACCESS-ESM1-5_historical_r1i1p1f1_gn_201001-201412.nc', decode_coords="all")

In [5]: ds
Out[5]: 
<xarray.Dataset>
Dimensions:    (bnds: 2, lat: 145, lon: 192, time: 60)
Coordinates:
  * time       (time) datetime64[ns] 2010-01-16T12:00:00 ... 2014-12-16T12:00:00
    time_bnds  (time, bnds) datetime64[ns] ...
  * lon        (lon) float64 0.0 1.875 3.75 5.625 ... 352.5 354.4 356.2 358.1
    lon_bnds   (lon, bnds) float64 ...
  * lat        (lat) float64 -90.0 -88.75 -87.5 -86.25 ... 86.25 87.5 88.75 90.0
    lat_bnds   (lat, bnds) float64 ...
Dimensions without coordinates: bnds
Data variables:
    pr         (time, lat, lon) float32 ...

```

DWesl:
> ~Does anyone know why the `xr.open_dataset(....)` call is echoed in the warning message. Is this intentional? Cc @dcherian @DWesl~
> 

It seems you've already figured this out, but for anyone else with this question, the repeat of the call on that file is part of the warning that the file does not have all the variables the attributes refer to.  You can fix this by recreating the file with the listed variables added (`areacella`, or by deleting the attribute from the variables (`cell_measures`).  You can also ignore the warning using the machinery in the `warnings` module.

PR #2846, opened 2019-03-22
Fix formatting of engine parameter in docstrings

spencermathews:
The formatting of the `engine` parameter is wonky in the docs of [`open_dataset()`](https://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html), [`open_mfdataset()`](https://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html), [`open_dataarray()`](https://xarray.pydata.org/en/stable/generated/xarray.open_dataarray.html).

To maintain consistent format with the other parameters once the docs are build, text up through the word `optional` should be on the same line as the parameter name. Currently it's pushed down into the parameter description.

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -230,7 +230,8 @@ def open_dataset(filename_or_obj, group=None, decode_cf=True,
     decode_coords : bool, optional
         If True, decode the 'coordinates' attribute to identify coordinates in
         the resulting dataset.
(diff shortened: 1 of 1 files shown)


shoyer:
Hmm. I wish there was a way to do this so both sphinx and lint checkers were happy.

spencermathews:
Yes, it seemed like such a simple fix, until the PEP 8 bit. I came across numpy/numpydoc#87, which seems apropos. I pushed an update that should fix the linter error and it looks good on a local build of the docs, hoping it passes the other checks.

shoyer:
thanks for looking into this @spencermathews !

PR #2849, opened 2019-03-24
Remove py37 collections.abc deprecation warning from lru_cache module.

douglatornell:
Looks like this occurrence was missing in #2574.


max-sixty:
Thanks @douglatornell !

douglatornell:
You're welcome @max-sixty !  And I congratulate you and the rest of the xarray maintainers on the https://xarray.pydata.org/en/latest/contributing.html docs. They made it super easy for me to jump in and do this. Well done!

PR #2855, opened 2019-03-28
Add note about integer compression precision

naught101:
16 bit integer compression is pretty lossy, by my calculations. Probably good enough for a lot of cases, but worth warning about.

 - [x] documentation changes only.

### DIFF
--- a/doc/io.rst
+++ b/doc/io.rst
@@ -286,6 +286,15 @@ converting ``NaN`` to ``-9999``, we would use
 ``encoding={'foo': {'dtype': 'int16', 'scale_factor': 0.1, '_FillValue': -9999}}``.
 Compression and decompression with such discretization is extremely fast.
 
+.. note::
+
+    This form of compression is lossy - the maximum precision available is 
+    `1 / (scaling factor * 2 ** 16)` - slightly better than 4 significant 
+    digits when using a scaling factor of 1. 
+    To achieve the maximum precision for your data, calculate the scaling factor 
+    using `max(data) - min(data)`, and use `min(data)` as the offset.
+
+
 .. _io.string-encoding:
 
 String encoding
--- a/README.rst
+++ b/README.rst
@@ -97,7 +97,7 @@ to supporting the open source scientific computing community. If you like
 Xarray and want to support our mission, please consider making a donation_
 to support our efforts.
 
-.. _donation: https://www.flipcause.com/secure/cause_pdetails/NDE2NTU=
+.. _donation: https://numfocus.salsalabs.org/donate-to-xarray/
 
 History
 -------
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -140,7 +140,7 @@ to supporting the open source scientific computing community. If you like
 Xarray and want to support our mission, please consider making a donation_
 to support our efforts.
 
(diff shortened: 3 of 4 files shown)


doc/io.rst
```diff
@@ -286,6 +286,15 @@ converting ``NaN`` to ``-9999``, we would use
 ``encoding={'foo': {'dtype': 'int16', 'scale_factor': 0.1, '_FillValue': -9999}}``.
 Compression and decompression with such discretization is extremely fast.
 
+.. note::
+
+    This form of compression is lossy - the maximum precision available is 
+    `1 / (scaling factor * 2 ** 16)` - slightly better than 4 significant
```
shoyer: I'm not sure if you're using "precision" correctly here. Shouldn't precision just be equal to scale_factor?

It might more helpful to specifically state the value ranges supported by typical signed integer types, e.g., something like:

> This form of compression is lossy. When converting to integer dtype, be sure that your scaled data will fit  inside the valid integer range, e.g., -128 to 127 for int8 and -32,768 to 32,767 for int16.

naught101: Hrm, maybe I'm understanding it wrong.

With uint16, and data ranging between 0-100, and a scale factor of 0.01, you would get slightly better than 2 decimal points of (mathematical) precision, right? And with uint 8, you'd get about half a point of precision.

With data 0-1000, and a scale factor of 0.001, you'd get a bit better than 1 decimal point for uint16, and ~4 _units_ of precision for uint8.

So is the precision actually log10(2**<int_size>) + log10(scale_factor)


.. Or do I have the scale_factor inverted?

shoyer: > With uint16, and data ranging between 0-100, and a scale factor of 0.01, you would get slightly better than 2 decimal points of (mathematical) precision, right? And with uint 8, you'd get about half a point of precision.

Either way, if you set `scale_factor` to 0.01, you would only have 2 decimal points of precision in the decoded data (e.g., what you have direct access to in xarray).

With uint8, you would definitely have overflow -- you could only represent the numbers in the range 0 to 2.55. I don't know if we currently issue a warning when this happens, but we probably should!

shoyer: We silently overflow right now :(
```python
ds = xr.Dataset({'x': ('y', [1, 2, 3])})
ds.to_netcdf('test3.nc', encoding={'x': {'scale_factor': 0.01, 'dtype': 'uint8'}})
xr.open_dataset('test3.nc').load()
```
Prints:
```
/usr/local/lib/python3.6/dist-packages/ipykernel_launcher.py:2: SerializationWarning: saving variable x with floating point data as an integer dtype without any _FillValue to use for NaNs
  
<xarray.Dataset>
Dimensions:  (y: 3)
Data variables:
    x        (y) float32 1.0 2.0 0.44
```

naught101: Ah.. I wasn't really thinking of always using powers of 10 as the scaling factor, but I guess that makes sense - at least the precision is much clearer that way.

rabernat:
Those interested in lossy compression should check out [ZFP](https://zfp.readthedocs.io/en/release0.5.4/). 

We are working on adding ZFP support in zarr (see https://github.com/zarr-developers/numcodecs/pull/160).

PR #2877, opened 2019-04-08
WIP: type annotations

### LINKED ISSUE #2869: xarray.concat docstring confuses PyCharm type detection
PyCharm inspects docstrings to figure out the return type of functions. In the case of xarray.concat(), it believes it is 'type'; this causes it to prominently highlight the user code as an error:

pycharm

crusaderky:
Fixes #2869

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1,8 +1,8 @@
 from collections import OrderedDict
 from contextlib import suppress
 from textwrap import dedent
-from typing import (Any, Callable, Hashable, Iterable, List, Optional, Tuple,
-                    TypeVar, Union)
+from typing import (Any, Callable, Hashable, Iterable, Iterator, List, Mapping,
+                    MutableMapping, Optional, Tuple, TypeVar, Union)
 
 import numpy as np
 import pandas as pd
@@ -13,11 +13,6 @@ from .options import _get_keep_attrs
 from .pycompat import dask_array_type
 from .utils import Frozen, ReprObject, SortedKeysDict, either_dict_or_kwargs
 
-from typing import Mapping, MutableMapping
-try:
-    from .pycompat import Mapping, MutableMapping  # noqa: F811
-except ImportError:
-    pass
 
 # Used as a sentinel value to indicate a all dimensions
 ALL_DIMS = ReprObject('<all-dims>')
@@ -28,7 +23,8 @@ T = TypeVar('T')
 
 class ImplementsArrayReduce:
     @classmethod
-    def _reduce_method(cls, func, include_skipna: bool, numeric_only):
+    def _reduce_method(cls, func: Callable, include_skipna: bool,
+                       numeric_only):
         if include_skipna:
             def wrapped_func(self, dim=None, axis=None, skipna=None,
                              **kwargs):
@@ -59,7 +55,8 @@ class ImplementsArrayReduce:
 
 class ImplementsDatasetReduce(object):
     @classmethod
-    def _reduce_method(cls, func, include_skipna, numeric_only):
+    def _reduce_method(cls, func: Callable, include_skipna: bool,
+                       numeric_only: bool):
         if include_skipna:
             def wrapped_func(self, dim=None, skipna=None,
                              **kwargs):
@@ -89,38 +86,36 @@ class ImplementsDatasetReduce(object):
 class AbstractArray(ImplementsArrayReduce):
     """Shared base class for DataArray and Variable."""
 
-    def __bool__(self):
+    def __bool__(self: Any) -> bool:
         return bool(self.values)
 
-    # Python 3 uses __bool__, Python 2 uses __nonzero__
-    __nonzero__ = __bool__
-
-    def __float__(self):
+    def __float__(self: Any) -> float:
         return float(self.values)
 
-    def __int__(self):
+    def __int__(self: Any) -> int:
         return int(self.values)
 
-    def __complex__(self):
+    def __complex__(self: Any) -> complex:
         return complex(self.values)
 
-    def __array__(self, dtype=None):
+    def __array__(self: Any, dtype: Union[str, np.dtype, None] = None
+                  ) -> np.ndarray:
         return np.asarray(self.values, dtype=dtype)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return formatting.array_repr(self)
 
-    def _iter(self):
+    def _iter(self: Any) -> Iterator[Any]:
         for n in range(len(self)):
             yield self[n]
 
-    def __iter__(self):
+    def __iter__(self: Any):
         if self.ndim == 0:
             raise TypeError('iteration over a 0-d array')
         return self._iter()
 
     @property
-    def T(self):
+    def T(self: Any):
         return self.transpose()
 
     def get_axis_num(self, dim: Union[Hashable, Iterable[Hashable]]
@@ -142,7 +137,7 @@ class AbstractArray(ImplementsArrayReduce):
         else:
             return self._get_axis_num(dim)
 
-    def _get_axis_num(self, dim: Hashable) -> int:
+    def _get_axis_num(self: Any, dim: Hashable) -> int:
         try:
             return self.dims.index(dim)
         except ValueError:
@@ -150,7 +145,7 @@ class AbstractArray(ImplementsArrayReduce):
                              (dim, self.dims))
 
     @property
-    def sizes(self) -> Mapping[Hashable, int]:
+    def sizes(self: Any) -> Mapping[Hashable, int]:
         """Ordered mapping from dimension names to lengths.
 
         Immutable.
@@ -885,7 +880,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
 
         return ops.where_method(self, cond, other)
 
-    def close(self) -> None:
+    def close(self: Any) -> None:
         """Close any files linked to this object
         """
         if self._file_obj is not None:
@@ -941,10 +936,10 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
             dask='allowed',
         )
 
-    def __enter__(self):
+    def __enter__(self: T) -> T:
         return self
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
         self.close()
 
 
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1,8 +1,8 @@
 from collections import OrderedDict
 from contextlib import suppress
 from textwrap import dedent
-from typing import (Callable, Iterable, List, Mapping, MutableMapping,
-                    Optional, Tuple, TypeVar, Union, overload)
+from typing import (Callable, Iterable, List, Mapping, Optional, Tuple,
+                    TypeVar, Union)
 
 import numpy as np
 import pandas as pd
@@ -120,15 +120,8 @@ class AbstractArray(ImplementsArrayReduce):
     def T(self):
         return self.transpose()
 
-    @overload
-    def get_axis_num(self, dim: str) -> int:
-        ...
-
-    @overload  # noqa:F811
-    def get_axis_num(self, dim: Iterable[str]) -> Tuple[int, ...]:
-        ...
-
-    def get_axis_num(self, dim):  # noqa:F811
+    def get_axis_num(self, dim: Union[str, Iterable[str]]
+                     ) -> Union[int, Tuple[int, ...]]:
         """Return axis number(s) corresponding to dimension(s) in this array.
 
         Parameters
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -109,14 +109,10 @@ class AbstractArray(ImplementsArrayReduce):
         for n in range(len(self)):
             yield self[n]
 
-    def __iter__(self: Any):
+    def __iter__(self: Any) -> Iterator[Any]:
         if self.ndim == 0:
             raise TypeError('iteration over a 0-d array')
         return self._iter()
-
-    @property
-    def T(self: Any):
-        return self.transpose()
 
     def get_axis_num(self, dim: Union[Hashable, Iterable[Hashable]]
                      ) -> Union[int, Tuple[int, ...]]:
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -24,7 +24,7 @@ T = TypeVar('T')
 class ImplementsArrayReduce:
     @classmethod
     def _reduce_method(cls, func: Callable, include_skipna: bool,
-                       numeric_only):
+                       numeric_only: bool):
         if include_skipna:
             def wrapped_func(self, dim=None, axis=None, skipna=None,
                              **kwargs):
@@ -53,7 +53,7 @@ class ImplementsArrayReduce:
             and 'axis' arguments can be supplied.""")
 
 
-class ImplementsDatasetReduce(object):
+class ImplementsDatasetReduce:
     @classmethod
     def _reduce_method(cls, func: Callable, include_skipna: bool,
                        numeric_only: bool):
@@ -84,8 +84,8 @@ class ImplementsDatasetReduce(object):
 
 
 class AbstractArray(ImplementsArrayReduce):
-    """Shared base class for DataArray and Variable."""
-
+    """Shared base class for DataArray and Variable.
+    """
     def __bool__(self: Any) -> bool:
         return bool(self.values)
 
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1,8 +1,8 @@
 from collections import OrderedDict
 from contextlib import suppress
 from textwrap import dedent
-from typing import (Callable, Iterable, Mapping, Optional, Tuple, TypeVar,
-                    Sequence, Union, overload)
+from typing import (Callable, Iterable, List, Mapping, MutableMapping,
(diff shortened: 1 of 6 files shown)


xarray/core/common.py
```diff
@@ -115,7 +120,15 @@ def __iter__(self):
     def T(self):
         return self.transpose()
 
-    def get_axis_num(self, dim):
+    @overload
+    def get_axis_num(self, dim: str) -> int:
+        pass
+
+    @overload  # noqa:F811
+    def get_axis_num(self, dim: Iterable[str]) -> Tuple[int]:
```
shoyer: Dimensions are actually allowed to be non-strings, so it might be better to keep this overload more generic, given that that reflects how the code actually works:
```suggestion
    def get_axis_num(self, dim: Iterable[Hashable]) -> Tuple[int, ...]:
```

(Also `Tuple[int]`  should be `Tuple[int, ...]`)

xarray/core/common.py
```diff
@@ -115,7 +120,15 @@ def __iter__(self):
     def T(self):
         return self.transpose()
 
-    def get_axis_num(self, dim):
+    @overload
+    def get_axis_num(self, dim: str) -> int:
+        pass
```
shoyer: I think it's more idiomatic to use `...` rather than `pass` when writing overloads?

xarray/core/common.py
```diff
@@ -133,15 +146,15 @@ def get_axis_num(self, dim):
         else:
             return tuple(self._get_axis_num(d) for d in dim)
 
-    def _get_axis_num(self, dim):
+    def _get_axis_num(self, dim: str) -> int:
```
shoyer:
```suggestion
    def _get_axis_num(self, dim: Hashable) -> int:
```

xarray/core/common.py
```diff
@@ -133,15 +146,15 @@ def get_axis_num(self, dim):
...
         try:
             return self.dims.index(dim)
         except ValueError:
             raise ValueError("%r not found in array dimensions %r" %
                              (dim, self.dims))
 
     @property
-    def sizes(self):
+    def sizes(self) -> Frozen[str, int]:
```
shoyer: I don't think `Frozen` is a valid type annotation. Maybe better to stick with `Mapping[str, int]`?

xarray/core/common.py
```diff
@@ -214,36 +227,38 @@ def _ipython_key_completions_(self):
         return list(set(item_lists))
 
 
-def get_squeeze_dims(xarray_obj, dim, axis=None):
+def get_squeeze_dims(xarray_obj, dim: Union[str, Tuple[str, ...], None] = None,
```
shoyer: Probably should be something more relaxed, perhaps `dim: Union[Hashable, Iterable[Hashable]]`

xarray/core/common.py
```diff
@@ -271,7 +286,7 @@ def squeeze(self, dim=None, drop=False, axis=None):
         dims = get_squeeze_dims(self, dim, axis)
         return self.isel(drop=drop, **{d: 0 for d in dims})
 
-    def get_index(self, key):
+    def get_index(self, key: str) -> pd.Index:
```
shoyer: `key` can actually be any hashable item here.

xarray/core/common.py
crusaderky: If they were intended to be allowed to be non-strings, sure they aren't tested for it!
```
>>> xarray.DataArray([1,2], dims=[123])
TypeError: dimension 123 is not a string
```
Also I've never seen it mentioned in the documentation...

xarray/core/common.py
crusaderky: it is valid as I changed the Frozen class to be a subclass of typing.Mapping

xarray/core/common.py
crusaderky: trouble with Iterable is that it doesn't necessarily allow you to iterate upon it twice (e.g. a file handler). I've tweaked the code to allow for it.

crusaderky:
@shoyer how do you feel about raising the minimum Python version to 3.5.2? There's quite a bit of important stuff in it, some of which undocumented. Namely ``@overload`` doesn't work with 3.5.0 (although I could conditionally replace it with a no-op thorough a backport module).

crusaderky:
From the 3.5.2 release notes:

> A new version of typing.py provides several new classes and features: @overload outside stubs, Reversible, DefaultDict, Text, ContextManager, Type[], NewType(), TYPE_CHECKING, and numerous bug fixes (note that some of the new features are not yet implemented in mypy or other static analyzers). Also classes for PEP 492 (Awaitable, AsyncIterable, AsyncIterator) have been added (in fact they made it into 3.5.1 but were never mentioned).

crusaderky:
@shoyer I am using mypy. But unlike flake8 I don't think I'll be able to integrate it in CI.
This error doesn't make any sense to me:
```
xarray/core/common.py:124: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
```

These I think are unfixable in Python 3.5, as I can't declare these under the class definition:
```
xarray/core/utils.py:445: error: "NdimSizeLenMixin" has no attribute "shape"
xarray/core/utils.py:450: error: "NdimSizeLenMixin" has no attribute "shape"
xarray/core/utils.py:454: error: "NdimSizeLenMixin" has no attribute "shape"
xarray/core/utils.py:468: error: "NDArrayMixin" has no attribute "array"
xarray/core/utils.py:472: error: "NDArrayMixin" has no attribute "array"
xarray/core/utils.py:478: error: "NDArrayMixin" has no attribute "array"
xarray/core/common.py:151: error: "AbstractArray" has no attribute "dims"
xarray/core/common.py:154: error: "AbstractArray" has no attribute "dims"
xarray/core/common.py:166: error: "AbstractArray" has no attribute "dims"
xarray/core/common.py:166: error: "AbstractArray" has no attribute "shape"
xarray/core/common.py:896: error: Cannot determine type of '_file_obj'
xarray/core/common.py:897: error: Cannot determine type of '_file_obj'
```
(additionally, typing doesn't offer anything for "file-like" objects)

crusaderky:
Mmh it looks like if I run mypy in Python 3.7 I get a *much* terser output. Still some horrible broken things to care about, like the fact that in 3.5 ``typing.Mapping`` did not inherit from ``collections.abc.Mapping``, but they can be worked around somehow.

crusaderky:
@shoyer I have a problem.
In Python 3.5.0-3.5.2, to create a custom typed collection you have to write
```
class Frozen(collections.abc.Mapping, typing.Mapping[K, V]):
```
Whereas in Python >= 3.5.3, you must just write
```
class Frozen(typing.Mapping[K, V]):
```
Things won't work if you use the < 3.5.3 syntax.

As an unrelated note, I'm observing that if I set python=3.5.3 in ``ci/requirements-py35.yml``, a ton of unit tests fall apart. The test suite ``ci/requirements-py35-min.yml`` ensures compatibility with Python 3.5.0... but only as long as you don't use any of the non-mandatory dependencies.
Can we 1) increase the minimum requirements to 3.5.4 and 2) set python=3.5.4 in ``ci/requirements-py35.yml``?
Is there a roadmap for dropping 3.5 entirely? (conda-forge already did...)

xarray/core/common.py
crusaderky: fixed

xarray/core/common.py
crusaderky: fixed ``Tuple[int, ...]``

xarray/core/common.py
crusaderky: changed to ``Mapping[str, int]`` for the sake of user friendliness, as Frozen isn't really part of the public API

crusaderky:
Opened https://bugs.python.org/issue36555

xarray/core/common.py
shoyer: Hmm. It seems that you can only construct these sorts of names/dimensions with the Dataset API:
```
In [8]: xarray.Dataset({123: ((456,), [1, 2, 3])})
Out[8]:
<xarray.Dataset>
Dimensions:  (456: 3)
Dimensions without coordinates: 456
Data variables:
    123      (456) int64 1 2 3
```

max-sixty:
FYI we added some code here to deal with python 3.5.0-3.5.2: https://github.com/pydata/xarray/pull/2831

shoyer:
If it's not too much trouble, I'd like to keep the xarray (with type checking) working on Python 3.5.0 for now. The type checks themselves only need to work on Python 3.7.

This should be relatively straightforward if you put type annotations in strings as needed and guard typing imports in `if TYPE_CHECKING` blocks.

crusaderky:
@shoyer I changed dimensions from str to Hashable as requested. In my opinion however it's not great, because
1. it's currently thoroughly broken due to lack of unit tests
2. it weakens type checking a lot, since Tuple is a subclass of Hashable however the vast majority of xarray functions deal with tuples with a specific code branch

Please review again - in theory, common.py and util.py are now 100% done. They pass flake8 and mypy (on python 3.7). If you like them please go on and merge into Master; since the work that needs to be done is going to touch literally everything I think it's best to break it down into more manageable chunks.

xarray/core/utils.py
```diff
@@ -266,7 +292,10 @@ def dict_equiv(first, second, compat=equivalent):
     return True
 
 
-def ordered_dict_intersection(first_dict, second_dict, compat=equivalent):
+def ordered_dict_intersection(first_dict: Mapping[K, V],
+                              second_dict: Mapping[K, V],
+                              compat: Callable[[V, V], bool] = equivalent
+                              ) -> MutableMapping[K, V]:
```
shoyer: This should probably be `'OrderedDict[K, V]'`

xarray/core/utils.py
```diff
@@ -290,172 +319,130 @@ def ordered_dict_intersection(first_dict, second_dict, compat=equivalent):
...
     """A simple ordered set.
 
     The API matches the builtin set, but it preserves insertion order of
     elements, like an OrderedDict.
     """
-
-    def __init__(self, values=None):
-        self._ordered_dict = OrderedDict()
+    def __init__(self, values: Optional[AbstractSet[T]] = None) -> None:
+        self._ordered_dict = OrderedDict()  # type: MutableMapping[T, None]
```
shoyer: You can use `type: OrderedDict[T, None]`

xarray/core/utils.py
```diff
@@ -290,172 +319,130 @@ def ordered_dict_intersection(first_dict, second_dict, compat=equivalent):
...
         if values is not None:
-            self |= values
+            self.__ior__(values)
```
shoyer: Does mypy not recognize `|=` as valid? That's usually preferred since that's the operator `__ior__` corresponds to.

If not, that's OK but please leave a comment.

xarray/core/utils.py
```diff
@@ -290,172 +319,130 @@ def ordered_dict_intersection(first_dict, second_dict, compat=equivalent):
...
-    def discard(self, value):
+    def discard(self, value: T) -> None:
         del self._ordered_dict[value]
 
     # Additional methods
 
-    def update(self, values):
-        self |= values
+    def update(self, values: AbstractSet[T]) -> None:
+        self.__ior__(values)
```
shoyer: same as above

xarray/core/utils.py
```diff
@@ -470,27 +457,27 @@ class NDArrayMixin(NdimSizeLenMixin):
...
 
 
-class ReprObject(object):
-    """Object that prints as the given value, for use with sentinel values."""
-
-    def __init__(self, value: str):
+class ReprObject:
+    """Object that prints as the given value, for use with sentinel values.
+    """
+    def __init__(self, value: str) -> None:
```
shoyer: note: I think mypy is OK with skipping `-> None` for `__init__` methods.

xarray/core/common.py
```diff
@@ -590,8 +602,11 @@ def rolling(self, dim=None, min_periods=None, center=False, **dim_kwargs):
         return self._rolling_cls(self, dim, min_periods=min_periods,
                                  center=center)
 
-    def coarsen(self, dim=None, boundary='exact', side='left',
-                coord_func='mean', **dim_kwargs):
+    def coarsen(self, dim: Optional[Mapping[Hashable, int]] = None,
+                boundary: str = 'exact',
+                side: Union[str, Mapping[Hashable, str]] = 'left',
+                coord_func: str = 'mean',
```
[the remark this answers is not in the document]
crusaderky: didn't you say that all dims should be Hashable?

xarray/core/utils.py
crusaderky: That's painful to achieve, because typing.OrderedDict was added in python 3.7.2, and I'm scared of hacking about what is such a fundamental class for xarray.
Also, are there any methods that OrderedDict offers that MutableMapping doesn't?

shoyer: If you put it in quotes, it's only checked when running mypy. That seems to work without any hacking, e.g., see 
https://github.com/pydata/xarray/blob/742ed3984f437982057fd46ecfb0bce214563cb8/xarray/core/merge.py#L28

xarray/core/common.py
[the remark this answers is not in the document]
shoyer: I think `Mapping[Any, str]` and `Mapping[Hashable, str]` are equivalent -- every mapping key must be hashable.

shoyer:
> @shoyer I changed dimensions from str to Hashable as requested. In my opinion however it's not great, because
> 
> 1. it's currently thoroughly broken due to lack of unit tests
> 2. it weakens type checking a lot, since Tuple is a subclass of Hashable however the vast majority of xarray functions deal with tuples with a specific code branch

I agree it's not great. But some of our APIs are going to be hard to type check regardless, e.g., in cases where we have different behavior for strings vs sequences of strings.

xarray/core/utils.py
crusaderky: Neither mypy nor PyCharm are happy with it.
It has to do with the signature of ``__ior__``:
```
def __ior__(self: MutableSet[T1], other: AbstractSet[T2]) -> MutableSet[Union[T1, T2]]:
    ...
```
When you do ``self |= other``, both PyCharm and mypy understand that you're changing the type of self in place, in other word they treat it to be the same as ``self = self | other``.
Calling ``__ior__`` directly bypasses that. 
This seems to be like a limitation of both tools like in our specific cases we explicitly specified that other is an ``AbstractSet[T1]``. I'll revert to |= and just disable the type checking.

xarray/core/utils.py
crusaderky: same as above

xarray/core/utils.py
crusaderky: fixed

xarray/core/utils.py
shoyer: OK, I'm happy with a comment, too. It's also not a big deal to disable type checking on particular lines when we're sure the overall method/function signature is right.

xarray/core/utils.py
crusaderky: I was going with PEP 484:
> (Note that the return type of __init__ ought to be annotated with -> None. The reason for this is subtle. If __init__ assumed a return annotation of -> None, would that mean that an argument-less, un-annotated __init__ method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that __init__ ought to have a return annotation; the default behavior is thus the same as for other methods.)

but I've now read https://github.com/python/mypy/issues/3358, which likely means a revision of the PEP in the opposite sense is incoming. I'll remove the ``-> None`` from ``__init__``.

crusaderky:
@shoyer implemented all change requests except OrderedDict

xarray/core/common.py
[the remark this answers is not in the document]
crusaderky: No, there's absolutely nothing in Mapping saying that the keys must be hashable.

This is perfectly legit, and pretty useful too:
```
class AnyKeyDict(MutableMapping[K, V]):
    def __init__(self, *args: Tuple[K, V], **kwargs: V)::
       self.mapping = {}
       for k, v in args:
           self[k] = v
       for k, v in kwargs.items():
           self[k] = v

    def __setitem__(self, k: K, v: V) -> None:
        try:
            self.mapping[k] = v
        except TypeError:
            self.mapping[pickle.dumps(k)] = v

    ...
```
This can be useful in a few borderline scenarios; for example when de-serialising msgpack data from a non-Python producer, as in msgpack map keys can be anything. Also I'd love to see something like this:
```
class DataArray:
    def __setitem__(self, k, v):
        if isinstance(k, Mapping):
            view = self.sel(**k)
        else:
            view = self.isel(**k)
        view[:] = v
```

xarray/core/utils.py
crusaderky: that would mean that anybody with less than Python 3.7.2 will get errors when running mypy...

shoyer: Personally, I'm OK with that. But it may be worth clarifying some cohesive policy.

shoyer:
thanks @crusaderky !

xarray/core/utils.py
shoyer: I'm not entirely sure, but It might help to annotate `self` here

xarray/core/common.py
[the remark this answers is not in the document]
shoyer: OK, good point.

It turns out we actually do already support this for `DataArray` :)
```
>>> data = xarray.DataArray([0, 0, 0], dims='x')
>>> data[{'x': 1}] = 1
>>> data
<xarray.DataArray (x: 3)>
array([0, 1, 0])
```

PR #2892, opened 2019-04-13
Return correct count for scalar datetime64 arrays

### LINKED ISSUE #2770: xarray .count() gives -1 when it should be 1?
Data array `da` I'm working with looks like: 
```
<xarray.DataArray 'T' (time: 365, latitude: 256, longitude: 512)>
[47841280 values with dtype=float32]
Coordinates:
  * time         (time) datetime64[ns] 1979-01-01T09:00:00 ... 1979-12-31T09:00:00
  * latitude     (latitude) float32 89.462944 88.766945 ... -89.462944
  * longitude    (longitude) float32 0.0 0.703125 ... 358.59375 359.29688
```

Now `int(da.time.count())` gives 365 which is correct. However when I select a time coordinate `da2 = da.isel(time=0)` the resulting array has one time coordinate:

```
<xarray.DataArray 'T' (latitude: 256, longitude: 512)>
[131072 values with dtype=float32]
Coordinates:
    time         datetime64[ns] 1979-01-01T09:00:00
  * latitude     (latitude) float32 89.462944 88.766945 ... -89.462944
  * longitude    (longitude) float32 0.0 0.703125 ... 358.59375 359.29688
```

and `int(da2.time.count())` gives me -1 instead of 1. I wasn't sure if this was a bug or I wasn't selecting the coordinate properly - if this is the case how should I call the time coordinate to count it correctly?

dnowacki-usgs:
BUG: Fix #2770 by changing ~ to np.logical_not()
TST: Add test for scalar datetime64 value
DOC: Add to whats-new.rst

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -26,6 +26,8 @@ Bug fixes
 
 - indexing with an empty list creates an object with zero-length axis (:issue:`2882`)
   By `Mayeul d'Avezac <https://github.com/mdavezac>`_.
+- Return correct count for scalar datetime64 arrays (:issue:`2770`)
+  By `Dan Nowacki <https://github.com/dnowacki-usgs>`_.
 
 .. _whats-new.0.12.1:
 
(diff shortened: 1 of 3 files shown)


shoyer:
thanks!

PR #2894, opened 2019-04-15
Added docs example for `xarray.Dataset.get()`

### LINKED ISSUE #2884: drop all but specified data_variables/coordinates as a convenience function
I often work with datasets that consist out of a lot of data_variables and coordinates.
Often I am only concerned about a subset of variables, and for convenience drop all but a selected list of variables with a little snippet like this:

```python
def xr_keep(obj, varlist):
    """drop all data_vars exept the ones provided in `varlist` """
    obj = obj.copy()
    drop_vars = [a for a in obj.data_vars if a not in varlist]
    return obj.drop(drop_vars)
```

I would love to have this functionality available as a DataArray/Dataset function. It could look something like `da_slim = da.drop_all_but(['var1, 'var3'])`. Would this be of interest to people here? Then I could try to put in a PR.

jbusecke:
Added example of `xarray.Dataset.get` method and alternative `ds[var_list]` syntax.
 - [ ] Closes #2884

### DIFF
--- a/doc/data-structures.rst
+++ b/doc/data-structures.rst
@@ -363,6 +363,7 @@ example, to create this example dataset from scratch, we could have written:
 
     ds = xr.Dataset()
     ds['temperature'] = (('x', 'y', 'time'), temp)
+    ds['temperature_double'] = (('x', 'y', 'time'), temp * 2 )
     ds['precipitation'] = (('x', 'y', 'time'), precip)
     ds.coords['lat'] = (('x', 'y'), lat)
     ds.coords['lon'] = (('x', 'y'), lon)
@@ -397,23 +398,9 @@ operations keep around coordinates:
 
 .. ipython:: python
 
-    list(ds[['temperature']])
-    list(ds[['x']])
-    list(ds.drop('temperature'))
-
-When working with a dataset containing many variables, it can be useful to
-select a subset of variables. Passing a list of variables to the
-:py:meth:`~xarray.Dataset.get` method returns a ``Dataset`` with only those
-variables:
-
-.. ipython:: python
-
-    ds['temperature_double'] = (('x', 'y', 'time'), temp * 2)
-    ds['precipitation_double'] = (('x', 'y', 'time'), precip * 2)
-
-    list(ds.get(['temperature', 'temperature_double']))
-    # alternative syntax
-    list(ds[['temperature', 'temperature_double']])
+    ds[['temperature']]
+    ds[['temperature', 'temperature_double']]
+    ds.drop('temperature')
 
 To remove a dimension, you can use :py:meth:`~xarray.Dataset.drop_dims` method.
 Any variables using that dimension are dropped:
--- a/doc/data-structures.rst
+++ b/doc/data-structures.rst
@@ -401,6 +401,20 @@ operations keep around coordinates:
     list(ds[['x']])
     list(ds.drop('temperature'))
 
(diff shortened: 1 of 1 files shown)


doc/data-structures.rst
```diff
@@ -401,6 +401,20 @@ operations keep around coordinates:
...
+select a subset of variables. Passing a list of variables to the
+:py:meth:`~xarray.Dataset.get` method returns a ``Dataset`` with only those
+variables:
+
+.. ipython:: python
+
+    ds['temperature_double'] = (('x', 'y', 'time'), temp * 2)
+    ds['precipitation_double'] = (('x', 'y', 'time'), precip * 2)
+
+    list(ds.get(['temperature', 'temperature_double']))
```
dcherian: Why `list`?

doc/data-structures.rst
```diff
@@ -401,6 +401,20 @@ operations keep around coordinates:
     list(ds[['x']])
```
dcherian: Not sure why there is a `list` here either :) but @jbusecke I think you just need to add `ds[['x', 'temperature']]` as an example here since the text already talks about passing a list of variable names.

jbusecke: Oh I see. Honestly I find the `ds[['x']]` line quite confusing without context. Why should I chose one of the dimensions?
Should I remove the `list(...)` calls alltoghether?

doc/data-structures.rst
jbusecke: See above. I wasnt sure what to make of this either but kept if for consistency.

doc/data-structures.rst
dcherian: I think the `list` calls need to be removed.

And I think changing `ds[['x']]` to `ds[['x', 'temperature']]` is probably all that is needed.

dcherian: Oh the `.get` example is great too, let's keep that.

shoyer:
For context: we didn't explicitly implement a `.get()` method on Dataset, we get it implemented for free because `xarray.Dataset` inherits from `collections.abc.Mapping`. So the only practical difference between `.get()` and `[]` on Dataset is the same as the difference for a general Python mapping: `.get()` returns a default value (which defaults to `None)` instead of raising `KeyError`.

I'm not sure it's particularly useful to show examples using `.get()` here, unless we want to show this difference in behavior if a key is not found.

jbusecke:
Oh I see. It returns None if *any* of the keys is not found. That might indeed lead to confusion. 
So should I just add an example with multiple variables using `ds[['var1', 'var2']]`?

shoyer:
> So should I just add an example with multiple variables using `ds[['var1', 'var2']]`?

I think this is probably a better idea.

To be honest, I don't know if `.get()` if useful for anyone with a list -- it's implementation here was really an accident (based on how `Mapping.get` works) rather than an intentional choice.

jbusecke:
I am still a bit puzzled over the `ds[['x', 'temperature']]` suggestion. Maybe I am not getting somthing, but `x` is a dimension and the output of ``ds[['x', 'temperature']]` is the same as ``ds[['temperature']]`. I think it would be clearer to add a third variable (maybe in the code block above) and then select two out of three data variables?

shoyer:
Yes, I think it would make more sense to add an example with another variable.

jbusecke:
How about this one?

dcherian:
I missed that `x` is a coordinate. 

Thanks @jbusecke

Issue #2895, opened 2019-04-15
Set/preserve the character array dimension name


jmccreight:
This is a new feature proposal not a bug. I'll open a PR against this issue momentarily, it consists of 4 lines of new code.

I've found it highly annoying that one can not set the name of the character array dimension. Looking at the code,  I basically found what I expected, except for what I added. Summary: Using a variable's variable.encoding one can decode the name into `variable.encoding['char_dim_name']` or one can simply set it when creating data from scratch. The "char_dim_name"  can be applied upon encoding. It's simple. All the new code is the same code that already handled character arrays, so there may not be any nasty edge cases. 

This shows how it works and the behavoir it changes:

```
# # Using the proposed changes.... 
# user@machine-session-1[1]:~/Downloads> ipython

import xarray as xa
char_arr = ['abc', 'def', 'ghi']
ds = xa.Dataset(data_vars={'char_arr': char_arr})
ds.char_arr.encoding.update({"dtype": "S1"})

# Default/current behavior
ds.to_netcdf('char_arr_string.nc')

# New functionality - name the character dimension.
ds.char_arr.encoding.update({"char_dim_name": "char_dim"})
ds.to_netcdf('char_arr_named.nc')

# user@machine-session-2[1]:~/Downloads> ncdump -h char_arr_string.nc
# netcdf char_arr_string {
# dimensions:
# 	char_arr = 3 ;
# 	string3 = 3 ;
# variables:
# 	char char_arr(char_arr, string3) ;
# 		char_arr:_Encoding = "utf-8" ;
# }
#
# user@machine-session-2[2]:~/Downloads> ncdump -h char_arr_named.nc 
# netcdf char_arr_named {
# dimensions:
# 	char_arr = 3 ;
# 	char_dim = 3 ;
# variables:
# 	char char_arr(char_arr, char_dim) ;
# 		char_arr:_Encoding = "utf-8" ;
# }


# New functionality - when decoding, preserve the character dimension name in the variable encoding for... encoding.
ds_read = xa.open_dataset('char_arr_named.nc')
ds_read.char_arr.encoding
# Out[4]: 
# {'_Encoding': 'utf-8',
#  'char_dim_name': 'char_dim',
#  'chunksizes': None,
#  'complevel': 0,
#  'contiguous': True,
#  'dtype': dtype('S1'),
#  'fletcher32': False,
#  'original_shape': (3, 3),
#  'shuffle': False,
[...]

PR #2896, opened 2019-04-15
Handle the character array dim name 

jmccreight:
See #2895

### DIFF
--- a/xarray/coding/strings.py
+++ b/xarray/coding/strings.py
@@ -104,11 +104,10 @@ class CharacterArrayCoder(VariableCoder):
         if data.dtype.kind == 'S' and encoding.get('dtype') is not str:
             data = bytes_to_char(data)
             if 'char_dim_name' in encoding.keys():
-                dims = dims + (encoding['char_dim_name'],)
+                dims = dims + (encoding.pop('char_dim_name'),)
             else:
                 default_char_dim_name = 'string%s' % data.shape[-1]
                 dims = dims + (default_char_dim_name,)
-                encoding['char_dim_name'] = default_char_dim_name
         return Variable(dims, data, attrs, encoding)
 
     def decode(self, variable, name=None):
--- a/xarray/coding/strings.py
+++ b/xarray/coding/strings.py
@@ -104,10 +104,10 @@ class CharacterArrayCoder(VariableCoder):
         if data.dtype.kind == 'S' and encoding.get('dtype') is not str:
             data = bytes_to_char(data)
             if 'char_dim_name' in encoding.keys():
-                dims = dims + (encoding.pop('char_dim_name'),)
+                char_dim_name = encoding.pop('char_dim_name')
             else:
-                default_char_dim_name = 'string%s' % data.shape[-1]
-                dims = dims + (default_char_dim_name,)
+                char_dim_name = 'string%s' % data.shape[-1]
+            dims = dims + (char_dim_name,)
         return Variable(dims, data, attrs, encoding)
 
     def decode(self, variable, name=None):
--- a/xarray/coding/strings.py
+++ b/xarray/coding/strings.py
@@ -113,7 +113,7 @@ class CharacterArrayCoder(VariableCoder):
         dims, data, attrs, encoding = unpack_for_decoding(variable)
 
         if data.dtype == 'S1' and dims:
-            encoding.update({'char_dim_name': dims[-1]})
+            encoding['char_dim_name'] = dims[-1]
             dims = dims[:-1]
             data = char_to_bytes(data)
         return Variable(dims, data, attrs, encoding)
--- a/xarray/coding/strings.py
+++ b/xarray/coding/strings.py
@@ -103,16 +103,19 @@ class CharacterArrayCoder(VariableCoder):
         dims, data, attrs, encoding = unpack_for_encoding(variable)
         if data.dtype.kind == 'S' and encoding.get('dtype') is not str:
             data = bytes_to_char(data)
-            dims = dims + ('string%s' % data.shape[-1],)
+            if 'char_dim_name' in encoding.keys():
+                dims = dims + (encoding['char_dim_name'],)
+            else:
+                dims = dims + ('string%s' % data.shape[-1],)
         return Variable(dims, data, attrs, encoding)
 
     def decode(self, variable, name=None):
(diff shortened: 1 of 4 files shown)


xarray/coding/strings.py
```diff
@@ -103,16 +103,19 @@ def encode(self, variable, name=None):
...
+                dims = dims + (encoding['char_dim_name'],)
+            else:
+                dims = dims + ('string%s' % data.shape[-1],)
         return Variable(dims, data, attrs, encoding)
 
     def decode(self, variable, name=None):
         dims, data, attrs, encoding = unpack_for_decoding(variable)
 
         if data.dtype == 'S1' and dims:
+            encoding.update({'char_dim_name': dims[-1]})
```
shoyer: nit: maybe `encoding['char_dim_name'] = dims[-1]` would be clearer, given that there is only one entry in `encoding`?

jmccreight:
thanks, @shoyer. I will add the documentation and tests now that the first hurdle is cleared and update the PR.

xarray/coding/strings.py
jmccreight: fair enough.

xarray/tests/test_coding_strings.py
```diff
@@ -92,6 +93,14 @@ def test_CharacterArrayCoder_roundtrip(original):
     coder = strings.CharacterArrayCoder()
     roundtripped = coder.decode(coder.encode(original))
     assert_identical(original, roundtripped)
+    if original.encoding == {}:
```
jmccreight: This is a little funky, but fairly clear. If there is no original char_dim_name, the decoded version has the default "stringN", otherwise the char_dim_names match.

jmccreight:
@shoyer Added test and documentation. I did not build documentation, wasnt sure if that was necessary. 
The history should be squashed when the time comes...

xarray/tests/test_coding_strings.py
shoyer: We try to avoid writing any logic in tests because it makes things harder to follow. Would you mind writing a separate test function for just this case?

It would also be good to explicitly test the name of the new dimension on the encoded variable.

xarray/tests/test_coding_strings.py
```diff
@@ -107,6 +107,35 @@ def test_CharacterArrayCoder_encode(data):
...
 
 
+@pytest.mark.parametrize(
+    'original',
+    [
+        Variable(('x',), [b'ab', b'cdef']),
+        Variable(('x',), [b'ab', b'cdef'], encoding={'char_dim_name': 'foo'})
+    ]
+)
+def test_CharacterArrayCoder_char_dim_name(original):
```
jmccreight: I think this is better and getting warmer. The test improved the underlying code [here](https://github.com/pydata/xarray/pull/2896/files#diff-64dc8cc049658cbab07c1615c1580689R111). 

But I still dont see how to eliminate this logic. Putting an `or` in the assert does not seem like a better alternative, nor does only testing just one of the two possible input scenarios.... let me know if you have a better idea that I'm too blind to see.

Thanks,

shoyer: The clean way to do this would be to add a second parametrized argument for the expected value, e.g.,
```
@pytest.mark.parametrize(
    ['original', 'expected_char_dim_name'],
    [
        (Variable(('x',), [b'ab', b'cdef']), 'string4'),
        (Variable(('x',), [b'ab', b'cdef'], encoding={'char_dim_name': 'foo'}), 'char_dim_name')
    ]
)
def test_CharacterArrayCoder_char_dim_name(original, expected_char_dim_name):
     ...
```

xarray/coding/strings.py
```diff
@@ -103,16 +103,21 @@ def encode(self, variable, name=None):
         dims, data, attrs, encoding = unpack_for_encoding(variable)
         if data.dtype.kind == 'S' and encoding.get('dtype') is not str:
             data = bytes_to_char(data)
-            dims = dims + ('string%s' % data.shape[-1],)
+            if 'char_dim_name' in encoding.keys():
+                dims = dims + (encoding['char_dim_name'],)
+            else:
+                default_char_dim_name = 'string%s' % data.shape[-1]
+                dims = dims + (default_char_dim_name,)
+                encoding['char_dim_name'] = default_char_dim_name
```
jmccreight: I think this line is causing a set of failures. I think this line is desirable. I could be wrong.

xarray/tests/test_coding_strings.py
jmccreight: That is the first case of 2-variable usage in that test file... thanks for pointing it out.

xarray/coding/strings.py
jmccreight: line 111, to be clear.

shoyer: The convention we have (which is apparently enforced by tests) is that when you use an encoding you remove it from the `encoding` dict. That way we catch cases where you have a typo, e.g., `char_dimension_name` instead of `char_dim_name`.

So this section should instead probably be something like:
```python
char_dim_name = encoding.pop(
    'char_dim_name', 'string%s' % data.shape[-1])
dims = dims + (char_dim_name,)
```

jmccreight:
I'm uncertain why travis is failing. Two of them look http-related and the other maybe be docs-related (but dont trust me). Running pytest locallin in the xarray/tests/ dir
```
============== 7007 passed, 1170 skipped, 25 xfailed, 1 xpassed, 30 warnings in 62.12 seconds ==============
```

xarray/coding/strings.py
jmccreight: Great reason, this solves all the errors I was seeing locally. THanks

dcherian:
One is a lint failure: https://github.com/pydata/xarray/pull/2896#issuecomment-484267839

The docs failure looks like a cartopy install problem

jmccreight:
🎊

shoyer:
Could you please add a brief note to `whats-new.rst`? Otherwise this looks good to go.

jmccreight:
🤦‍♂️ with that formatting in the whats-new.rst.
I think this is complete. 
thanks for the mini tour of xarray internals, I learned some useful things!

dcherian:
Thanks @jmccreight

castelao:
Perfect timing, I just needed that! Thanks @jmccreight et al

Issue #2841, opened 2019-03-22
Automatic naming causes trouble


randolf-scholz:
#### Code Sample, a copy-pastable example if possible

A "Minimal, Complete and Verifiable Example" will make it much easier for maintainers to help you:
http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports

```python
from xarray import Dataset, DataArray
ds = Dataset(coords={'foo':[1,2,3], 'bar':[1,2,3]})
da = DataArray([0], dims='test')
da.name = 'abc'
ds[(1,2)] = da
print(da.name)
print(ds[(1,2)].name)
print(type(ds[(1,2)].name))
print(ds[(1,2)])              # throws error
```
#### Problem description

The name gets overwritten with the tuple key, but is should stay intact?!?
Secondly the print statement throws an error because internally `name_str = '%r ' % arr.name` is called. But here `arr.name` is a `tuple` !! 

Instead, it should be `repr(arr.name)` or `F"{arr.name}"`

#### Expected Output

The name should not change. 

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.18.0-16-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.1

xarray: 0.12.0
pandas: 0.24.2
numpy: 1.16.2
scipy: 1.2.1
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: None
distributed: None
matplotlib: 3.0.3
cartopy: None
seaborn: None
setuptools: 40.8.0
pip: 19.0.3
conda: 4.6.8
pytest: None
IPython: 7.3.0
sphinx: 1.8.5

</details>

shoyer:
Thanks for the report.

Yes, this looks like a bug. A pull request to fix this would be very welcome.

randolf-scholz:
[...]

PR #2906, opened 2019-04-18
Partial fix for #2841 to improve formatting.

dnowacki-usgs:
Updates formatting to use .format() instead of % operator. Changed all instances of % to .format() and added test for using tuple as key, which errored using % operator.

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -116,7 +116,7 @@ def format_timestamp(t):
         if time_str == '00:00:00':
             return date_str
         else:
-            return '%sT%s' % (date_str, time_str)
+            return '{}T{}'.format(date_str, time_str)
 
 
 def format_timedelta(t, timedelta_format=None):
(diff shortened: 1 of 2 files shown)


shoyer:
Thanks @dnowacki-usgs !

One minor note -- it's valid (since Python 2.7) to write string formatting with explicit positional indexes, e.g., you could just write `'{} {}'.format('one', 'two')` rather than `'{0} {1}'.format('one', 'two')`.

I think it's usually a little cleaner to omit the numeric indices, unless you're reusing arguments multiple times or printing the names out of order. (But even in these cases it's usually nicer to use keyword names.)

max-sixty:
Also come Py3.6, f strings 🙏

dnowacki-usgs:
Thanks @shoyer. I think numeric indexes are clearer, though maybe not cleaner. Removed them and fixed the dask CI test I didn't catch the first time through. Hopefully everything passes now. 🤞

shoyer:
Looks great! You can ignore the failing doc build in the tests, that is unrelated.

If you like, feel free to give yourself a credit in `whats-new.rst` for fixing this. Otherwise I'll merge this in a day or two.

dnowacki-usgs:
Good to go as far as I'm concerned. I'm not even sure how I'd phrase these changes in whats-new, so let's go with it as-is.

max-sixty:
OK, thanks @dnowacki-usgs !

PR #2917, opened 2019-04-25
Implement load_dataset() and load_dataarray()

### LINKED ISSUE #2887: Safely open / close netCDF files without resource locking

(essentially the same to #1629)

Opening netCDF file via `xr.open_dataset` locks a resource, preventing to write a file with the same name (as pointed out and answered as an expected behavior in #1629).
```python
import xarray as xr
ds = xr.Dataset({'var': ('x', [0, 1, 2])})
ds.to_netcdf('test.nc')

ds_read = xr.open_dataset('test.nc')
ds.to_netcdf('test.nc')   # -> PermissionError
```
```python
ds_read = xr.open_dataset('test.nc').load()
ds.to_netcdf('test.nc')   # -> PermissionError
```
```python
ds_read = xr.open_dataset('test.nc').load()
ds_read.close()
ds.to_netcdf('test.nc')  # no error
```

#### Problem description

Another program cannot write the same netCDF file that xarray has opened, unless `close` method is not called.

-----
-- EDIT --

`close()` method does not return the object, thus it cannot be put in the chain call, such as
```python
some_function(xr.open_dataset('test.nc').close())
```
-----

It is understandable when we do not want to load the entire file into the memory.
However, sometimes I want to read the file that will be updated soon by another program.
Also, I think that many users who are not accustomed to netCDF may expect this behavior (as `np.loadtxt` does) and will be surprised after getting `PermissionError`.

I think it would be nice to have an option such as `load_all=True` or even make it a default?

#### Expected Output
No error

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.1 (default, Oct 23 2018, 19:19:42) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.15.0-1035-oem
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.2
libnetcdf: 4.6.1

xarray: 0.12.0+11.g7d0e895f.dirty
pandas: 0.23.4
numpy: 1.15.4
scipy: 1.2.0
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: 2.8.0
Nio: None
zarr: None
cftime: 1.0.2.1
nc_time_axis: None
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 1.0.0
distributed: 1.25.0
matplotlib: 2.2.2
cartopy: None
seaborn: 0.9.0
setuptools: 40.5.0
pip: 18.1
conda: None
pytest: 4.0.1
IPython: 7.1.1
sphinx: 1.8.2

</details>

dnowacki-usgs:
- [x] Closes #2887
 - [x] Tests added

Implement load_dataset() and load_dataarray()

BUG: Fixes #2887 by adding @shoyer solution for load_dataset and load_dataarray, wrappers around open_dataset and open_dataarray which open, load, and close the file and return the Dataset/DataArray
TST: Add tests for sequentially opening and writing to files using new functions
DOC: Add to whats-new.rst. Also a tiny change to the open_dataset docstring


xarray/backends/api.py
```diff
@@ -185,12 +185,28 @@ def _finalize_store(write, store):
     store.close()
 
 
+def load_dataset(*args, **kwargs):
+    """Open, load into memory, and close a dataset from a file or file-like
+    object.
+    """
```
shoyer: It would be good to document that arguments are passed on to `open_dataset`, and include a reference to `open_dataset` under a "See also" section. That gets turned into a link in the online API docs.

Also: there are probably a couple of arguments that should be invalid with `load_dataset`, e.g., `cache` is meaningless if you're loading the data into memory anyways.

xarray/backends/api.py
```diff
@@ -185,12 +185,28 @@ def _finalize_store(write, store):
     store.close()
 
 
+def load_dataset(*args, **kwargs):
```
shoyer: Maybe this should be: `def load_dataset(filename_or_obj, **kwargs)`? That would be a little more explicit, and ensure that users don't rely on the argument order of keyword arguments to `open_dataset`.

xarray/backends/api.py
shoyer: The docstring should also spell out exactly how this differs from `open_dataset`.

shoyer:
What do you think un-deprecating `xarray.tutoral.load_dataset` as part of this change? It might be useful to have that stick around for consistency.

xarray/backends/api.py
dnowacki-usgs: Good, idea, changed as suggested

dnowacki-usgs:
Seems like undeprecating it makes sense as a function with identical functionality is being added here. What if we just imported the new `load_dataset` as `xarray.tutorial.load_dataset` for the purposes of the tutorial?

shoyer:
> What if we just imported the new `load_dataset` as `xarray.tutorial.load_dataset` for the purposes of the tutorial?

`tutorial.load_dataset()` is a little different. It downloads the indicated file from GitHub to open instead of opening it directly.

dnowacki-usgs:
Forgive me, I don't understand what you mean by download vs open. I see that `tutorial.load_dataset()` is different from this PR in that it keeps the file handle open instead of closing it. Is that the distinction you are referring to? Otherwise all the gallery/examples I have checked call `tutorial.load_dataset()` at the beginning and that's it. Unless I'm missing something?

https://github.com/pydata/xarray/blob/6d93a95d05bdbfc33fff24064f67d29dd891ab58/xarray/tutorial.py#L92-L107

shoyer:
> Forgive me, I don't understand what you mean by download vs open. I see that `tutorial.load_dataset()` is different from this PR in that it keeps the file handle open instead of closing it. Is that the distinction you are referring to?

The difference I was referring to is that `xarray.tutorial.load_dataset()` needs to call `xarray.tutorial.open_dataset()` internally, not `xarray.open_dataset()`. But otherwise it can look exactly like `xarray.load_dataset`, e.g.,

```python
# in tutorial.py
def load_dataset(*args, **kwargs):
    # this calls tutorial.open_dataset, not xarray.open_dataset
    with open_dataset(*args, **kwargs) as ds:
        return ds.load() 
```

does that make sense now? :)

dnowacki-usgs:
Got it, thanks. I can make this change, assuming the dask argument used for deprecating it in the first place isn't stronger than maintaining this consistency.

shoyer:
I think the dask argument is really an argument that `tutorial.open_dataset()` needs to exist. But if we're also including a `xarray.load_dataset()` function, then for consistency it makes sense to have both tutorial opens, too.

dnowacki-usgs:
Sounds good. This latest commit un-deprecates `tutorial.load_dataset()` and makes it parallel to `xarray.load_dataset()` in terms of functionality.

xarray/backends/api.py
```diff
@@ -185,12 +185,68 @@ def _finalize_store(write, store):
...
+    -------
+    dataset : Dataset
+        The newly created Dataset.
+
+    See Also
+    --------
+    open_dataset
+    """
+    if 'cache' in kwargs:
+        warnings.warn(
```
shoyer: It's probably better to just raise an error here (`TypeError`), rather than to issue a warning.

doc/whats-new.rst
```diff
@@ -24,7 +24,13 @@ Enhancements
 - Character arrays' character dimension name decoding and encoding handled by
   ``var.encoding['char_dim_name']`` (:issue:`2895`)
   By `James McCreight <https://github.com/jmccreight>`_.
-   
+- Implement ``load_dataset()`` and ``load_dataarray()`` as alternatives to ``open_dataset()`` and
+``open_dataarray()` to open, load into memory,
```
shoyer: there's a missing backtick here after `open_dataarray`

dcherian:
LGTM. Thanks @dnowacki-usgs

shoyer:
LGTM, too. Thanks @dnowacki-usgs !

poplarShift:
Thanks a lot! @shoyer @dnowacki-usgs Could you clarify how this is different from `xr.open_dataset(..., autoclose=True)`?

PR #2919, opened 2019-04-25
DOC: avoid downloading .tif file

grinsted:
Simplify by using that rasterio can read directly from http. 

This also removes imports of requests and os.

### DIFF
--- a/doc/gallery/plot_rasterio.py
+++ b/doc/gallery/plot_rasterio.py
@@ -16,9 +16,6 @@ likely be irregular in lon/lat. It is often recommended to work in the data's
 original map projection (see :ref:`recipes.rasterio_rgb`).
 """
 
-import os
-import urllib.request
-
 import cartopy.crs as ccrs
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks! Can you do the same for the other example as well?

grinsted:
Done.

fmaussion:
@grinsted what OS are you running this on? I'm a bit concerned about this here: https://github.com/mapbox/rasterio#ssl-certs

I came into this while trying it locally, and even after setting the env variable I still get this error:

```
RasterioIOError: '/vsicurl/https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif' does not exist in the file system, and is not recognized as a supported dataset name.
```

I wonder if I'm the only one in that case, or if other users might be affected as well

grinsted:
I wrote this code on Windows 10. I also use the rasterio feature on a Linux
box (but not via xarray). I use anaconda on both machines.

grinsted:
I have just tested on a linux box and it works. I have access to a geo-notebook on a cloud service via my university . sys.version says:

3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 09:07:38) 
[GCC 7.3.0]

fmaussion:
Thanks! Yes, I've tested on a conda env and it works. So my take is that it probably won't work out-of-the box on debian/ubuntu/mint where rasterio is installed via pip.

So my question to @shoyer or @jhamman : should we change this example for the nicer code above, even if a minority of people (pip users like me) might find that this doesn't work for them?

grinsted:
I wonder why it desn't work with pip. It is a feature inherited from [GDAL](https://gdal.org/gdal_virtual_file_systems.html). I guess it has something to do with how that is compiled.

fmaussion:
> I wonder why it desn't work with pip.

Because of that: https://github.com/mapbox/rasterio#ssl-certs . It will work, but not out-of-the box. I personally don't mind, but we might get a few "your example doesn't work" requests here although it has nothing to do with xarray, this is why I'm asking others devs before merging.

dcherian:
How about adding a comment saying that this may/may not work and suggest a workaround

fmaussion:
> How about adding a comment saying that this may/may not work and suggest a workaround

Thanks for the suggestion! I've decided against it because it will be too wordy - let's see if this is really affecting that many people first.

Thanks @grinsted for the PR!

PR #2920, opened 2019-04-25
[WIP] Custom fill value for reindex, align, and merge operations

### LINKED ISSUE #2876: Custom fill value for align, reindex and reindex_like
It would be nice to be able to specify a custom fill value other than NaN for alignment/reindexing, e.g.,
```
>>> xr.DataArray([0, 0], [('x', [1, 2])]).reindex(x=[0, 1, 2, 3], fill_value=-1)
<xarray.DataArray (x: 4)>
array([-1, 0, 0, -1])
Coordinates:
  * x        (x) int64 1 2 3 4
```

This should be pretty straightforward, simplify a matter of adding a `fill_value` keyword argument to the various interfaces and passing it on to `Variable.__getitem_with_mask` inside `xarray.core.alignment.reindex_variables()`.

zdgriffith:
This PR adds an optional `fill_value` parameter for the following functions/methods:
```
xarray.core.alignment.align()
xarray.core.alignment.deep_align()

xarray.core.dataset.reindex()
xarray.core.dataset.reindex_like()
xarray.core.dataset.merge()

xarray.core.dataarray.DataArray.reindex()
xarray.core.dataarray.DataArray.reindex_like()

xarray.core.merge.merge_coords()
xarray.core.merge.merge_core()
xarray.core.merge.merge()
xarray.core.merge.dataset_merge_method()
```
The value given to `fill_value` is used to fill missing values in the new object after the operation.  By default `fill_value` is NaN.

 - [x] Closes #2876
 - [x] Tests added

### DIFF
--- a/xarray/core/alignment.py
+++ b/xarray/core/alignment.py
@@ -8,7 +8,7 @@ from typing import Any, Mapping, Optional, Tuple
 import numpy as np
 import pandas as pd
 
-from . import utils
+from . import utils, dtypes
 from .indexing import get_indexer_nd
 from .utils import is_dict_like, is_full_slice
 from .variable import IndexVariable, Variable
@@ -35,8 +35,8 @@ _DEFAULT_EXCLUDE = frozenset()  # type: frozenset
 
 
 def align(*objects, **kwargs):
-    """align(*objects, join='inner', copy=True, indexes=None,
-             exclude=frozenset())
+    """align(*objects, join='inner', copy=True, fill_value=dtypes.NA,
+             indexes=None, exclude=frozenset())
 
     Given any number of Dataset and/or DataArray objects, returns new
     objects with aligned indexes and dimension sizes.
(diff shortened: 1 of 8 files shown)


xarray/core/alignment.py
```diff
@@ -35,16 +35,17 @@ def _get_joiner(join):
 
 
 def align(*objects, **kwargs):
-    """align(*objects, join='inner', copy=True, indexes=None,
-             exclude=frozenset())
+    """align(*objects, join='inner', copy=True, fill_value=dtypes.NA,
+             indexes=None, exclude=frozenset())
```
shoyer: One small additional clean-up you could do here: since we no longer support Python 2, these function signature could be moved from the docstring onto the function itself, e.g.,
```
def align(*objects, join='inner', copy=True, fill_value=dtypes.NA,
             indexes=None, exclude=frozenset())"
```

This would let you drop the calls to `kwargs.pop()` below, and would be a small win for efficiency/clean code.

zdgriffith:
Thanks for the review @shoyer!  Would edits to the docs be in-scope for this PR?  For example where fill values are mentioned in http://xarray.pydata.org/en/latest/indexing.html#align-and-reindex.

shoyer:
> Would edits to the docs be in-scope for this PR?

Up to you. Would be happy to include any edits you think are appropriate but they aren't required. The API docs are the canonical reference for various function options but if you think there are good places to mention this option in the narrative docs that would also be welcome.

dcherian:
+1 for a line in the docs, I don't think an example is needed

shoyer:
thanks @zdgriffith!

PR #2924, opened 2019-04-27
fix multiIndex min max issue #2923

### LINKED ISSUE #2923: Calling max() on a stacked dimension raises exception

A "Minimal, Complete and Verifiable Example" will make it much easier for maintainers to help you:
http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports

```python
import numpy as np
import xarray as xr
data2 = np.random.randn(3,100, 4)
da = xr.DataArray(
    data2,
    name='value',
    dims=['replicate', 'rsample', 'exp'],
    coords=dict(replicate=[0,1,2],
        exp=['a', 'b', 'c', 'd'], rsample=list(range(100))))
da2 = da.stack(sample=('replicate', 'rsample'))
da2.sample.max()
```

#### Problem description
Calling  `max()` on a stacked dimension raises the exception
`ValueError: dimensions () must have the same length as the number of data dimensions, ndim=1`

#### Expected Output
Maximum value of selected dimension. i.e an index I can use with loc
e.g. ` da2.loc['a', da2.sample.max()]` that is `da2.loc['a', (2,99)]`

So I guess I'm expecting something like:
```
<xarray.DataArray 'sample' ()>
array((2, 99), dtype=object)
```
This works: `da2.loc['a', xr.DataArray((2,99), dims=(), name='sample')]`

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.15.0-47-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_AU.UTF-8
LOCALE: en_AU.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.1

xarray: 0.12.1
pandas: 0.24.2
numpy: 1.15.4
scipy: 1.2.1
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 1.2.0
distributed: 1.27.0
matplotlib: 3.0.3
cartopy: None
seaborn: 0.9.0
setuptools: 40.6.3
pip: 18.1
conda: 4.6.14
pytest: None
IPython: 7.4.0
sphinx: 2.0.1

arabidopsis:
- [x] Fixes #2923
 - [x] Test added

I ran pytest and it didn't seem to produce any problems.


arabidopsis:
Oops... I usually use black -- which I turned off since it would really have stuffed the code :)

shoyer:
Awesome, thank you @arabidopsis!

If you don't mind, it would be great if you could add a brief note about the bug fix to `whats-new.rst`. (if not, don't worry, I'll merge this in a few days either way)

You can ignore the test failure on Python 3.5. Something else seems to be going wrong there.

arabidopsis:
No problem. Thanks for xarray! 

Just a question. I put a guard in to test just for tuples (because that's what was breaking for me) but it seems to me that `_nan_minmax_object` should just return `utils.to_0d_object_array(data)`
unconditionally at that point no? It's a max or a min right.

Wasn't sure.

shoyer:
> Just a question. I put a guard in to test just for tuples (because that's what was breaking for me) but it seems to me that `_nan_minmax_object` should just return `utils.to_0d_object_array(data)`
> unconditionally at that point no? It's a max or a min right.
> 
> Wasn't sure.

Yes, it does seem like a good idea to just return `utils.to_0d_object_array()` unilaterally in this case.

dcherian:
Thanks @arabidopsis. Welcome to xarray!

PR #2925, opened 2019-04-28
Attempt to fix py35 build on Travis

shoyer:
This build is currently install NumPy 1.11, which isn't supported by xarray.
Maybe adding minimum numpy and pandas versions will help.


ocefpaf:
@shoyer conda-forge dropped support for Python 3.5 a while back and the old packages may present a problem.

Issue #2292, opened 2018-07-16
Support non-string dimension/variable names


joshburkart:
#### Problem description

Currently, it appears that "dimension"/"coordinate" labels must be strings. However, in more rigorous software engineering applications it is often desirable to use something more organized/structured for labels, e.g. enums. I think it would be great if `xarray` supported this.

Obviously storing to e.g. NetCDF necessitates string-valued field names, so I would think calling `str` could be appropriate when performing this sort of serialization. This is what `pandas` seems to do (see below). But I imagine there might be other issues that would need to be resolved to do what I'm suggesting...?

#### Code sample

```python
import enum

import numpy as np
import pandas as pd
import xarray as xr

class CoordId(enum.Enum):
    LAT = 'lat'
    LON = 'lon'

pd.DataFrame({CoordId.LAT: [1,2,3]}).to_csv()
# Returns: ',CoordId.LAT\n0,1\n1,2\n2,3\n'

xr.DataArray(
    data=np.arange(3 * 2).reshape(3, 2),
    coords={CoordId.LAT: [1, 2, 3], CoordId.LON: [7, 8]},
    dims=[CoordId.LAT, CoordId.LON],
)
# Fails: TypeError: dimension CoordId.LAT is not a string
```

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.15.0-1010-gcp
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: C.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.7
pandas: 0.23.1
numpy: 1.14.5
scipy: 1.1.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
h5py: 2.7.1
Nio: None
zarr: None
bottleneck: None
cyordereddict: None
dask: None
distributed: None
matplotlib: 2.1.1
cartopy: 0.16.0
seaborn: None
setuptools: 39.2.0
pip: 9.0.1
conda: None
pytest: 3.6.1
IPython: 6.4.0
sphinx: None
</details>

shoyer:
Hi @joshburkart -- thanks for raising this concern.

I agree, it would be nice to support enums (really any hashable value) as dimension names. Our current checks for strings are somewhat inconsistent, e.g., you can actually use these in an xarray.Dataset if you use the fully explicit API for constructing a dataset:
[...]

PR #2929, opened 2019-04-29
WIP: typing for DataArray/Dataset

crusaderky:
Status:
* I'm generally not pleased with the amount of added verbosity. Happy to accept suggestions on how to improve.
* Switching all variable names from str to Hashable. Without proper unit tests however (out of scope) non-string hashables are expected not to work most of the times. My preference would still be to stay limited on str...
* DataArray done.
* Dataset not done (except where it was hindering DataArray).
* mypy passes with the only error ``"Mapping[...]" has no attribute "copy"``. This is due to the fact that I can't see a way to use ``typing.OrderedDict`` without breaking compatibility with python < 3.7.2.
* py.test should be successful

@shoyer any early feedback is appreciated


crusaderky:
One of the many reasons why I'm strongly against generic Hashables instead of strings is that I expect user code all over the world to have string-specific functions and methods applied to variable names and dimensions.
e.g.
```
[d for d in arr.dims if d.startswith('foo')]
```
The above will produce an error in mypy, and the user will be forced to either disable the test or verbosely convince mypy with a ``typing.cast`` that actually in his project he never uses non-string dims.

max-sixty:
Thanks @crusaderky , this looks great!

Re: 

>Switching all variable names from str to Hashable. Without proper unit tests however (out of scope) non-string hashables are expected not work most of the times. My preference would still be to stay limited on str...

Curious on others' thoughts. I would vote to allow `Hashable` for data variable names in Datasets - similar to other `Mapping`s - but less so for dimension names in arrays. (Not opposed to it per se, open to instances where that would be helpful)

crusaderky:
DataArray finished; unit tests (hopefully) successful. Will now move to Dataset.
Would appreciate an official, final decision from @shoyer et. al. on Hashable vs. str before I continue - also in light of all the hacks Hashable forced me to go through.

shoyer:
@crusaderky let's discuss that over in https://github.com/pydata/xarray/issues/2292. For now, I would try to defer making a decision on non-string dimension/variable names, even though that means we will have less informative type annotations.

max-sixty:
@crusaderky I finished this off as best I could. Tests pass though needs a review. Here's the code: https://github.com/max-sixty/xarray/tree/annotations

(I tried pasting a patch, but I my git-fu isn't enough to exclude all the master code given there's a merge commit in there)

crusaderky:
@max-sixty thanks for the work! I'll go through it ASAP

crusaderky:
@max-sixty I merged your branch in and did some tweaks to the OrderedDict's.
Now moving to add more typing info to dataset.py...

xarray/core/dataarray.py
```diff
@@ -178,13 +207,23 @@ def __init__(self, data, coords=None, dims=None, name=None,
...
+            - sequence of tuples (dimension name, array-like)
+
+            Additionally, it is possible to define a coord whose name
+            does not match the dimension name, or a coord based on multiple
+            dimensions, with one of the following notations:
+
+            - mapping {coord name: DataArray}
+            - mapping {coord name: Variable}
+            - mapping {coord name: (dimension name, array-like)}
+            - mapping {coord name: (tuple of dimension names, array-like)}
```
max-sixty: 👍

max-sixty:
Great, thanks @crusaderky ! Let me know if I can do anything to help!

crusaderky:
@max-sixty you can keep adding type info to all functions in dataset.py. I don't think I'll have time to touch it next week anyway...

max-sixty:
@crusaderky I updated my branch slightly:
- Couple of merge conflicts
- Couple of issues around dict / OrderedDict. I tried to stay as close as possible while making the tests pass (I don't think I undid any of your most recent changes - if I did it was a mistake - feel free to modify)
- I wonder whether we should merge this and then continue; I worry we'll get a constant tax of merge conflicts given how many places this touches

max-sixty:
Just fixed some more merge conflicts over at https://github.com/max-sixty/xarray/tree/annotations. 

Thoughts on merging this now and continuing off master?

dcherian:
> Thoughts on merging this now and continuing off master?

+1 else the merge conflicts will just pile up.

crusaderky:
Sorry I didn't have time last week - I'll try merging into the pr ASAP. I'm in favour of merging to mainline too

crusaderky:
Sorry I didn't have time last week - I'll try merging into the pr ASAP. I'm in favour of merging to mainline too

max-sixty:
Cool, I think it should be a straight fast-forward merge from my branch to yours, and then an easy merge to master

xarray/core/dataarray.py
```diff
@@ -178,13 +207,23 @@ def __init__(self, data, coords=None, dims=None, name=None,
...
-            corresponding coordinates. If sequence-like, should be a sequence
-            of tuples where the first element is the dimension name and the
-            second element is the corresponding coordinate array_like object.
-        dims : str or sequence of str, optional
-            Name(s) of the data dimension(s). Must be either a string (only
-            for 1D data) or a sequence of strings with length equal to the
+            The following notations are accepted:
+
+            - mapping {dimension name: array-like}
+            - sequence of tuples (dimension name, array-like)
```
shoyer: The actual format for these tuples are any valid arguments for `xarray.Variable()`. So any of these would work: `(dims, data)`, `(dims, data, attrs)` and `(dims, data, attrs, encoding)`.

For now, it might be better to only enforce `tuple` as the type.

xarray/core/dataarray.py
```diff
@@ -234,30 +271,37 @@ def __init__(self, data, coords=None, dims=None, name=None,
...
         # public interface.
         self._indexes = indexes
 
         self._file_obj = None
 
-        self._initialized = True
+        self._initialized = True  # type: bool
 
-    __default = object()
+    __default = utils.ReprObject('__default')
```
shoyer: Maybe call this `'<default>'` to make it clear what it can be?

xarray/core/dataarray.py
```diff
@@ -234,30 +271,37 @@ def __init__(self, data, coords=None, dims=None, name=None,
...
 
-    def _replace(self, variable=None, coords=None, name=__default):
+    def _replace(self, variable: Optional[Variable] = None,
+                 coords=None,
+                 name: Union[str, None, utils.ReprObject] = __default
```
shoyer: Probably best to just stick with `Hashable` for now for `DataArray.name`. I think this falls in the same category as Dataset keys and dimensions (that is, we currently aren't entirely consistent with supporting non-strings but we definitely do, at least in some cases).

xarray/core/dataarray.py
```diff
@@ -234,30 +271,37 @@ def __init__(self, data, coords=None, dims=None, name=None,
...
+                 ) -> 'DataArray':
         if variable is None:
             variable = self.variable
         if coords is None:
             coords = self._coords
-        if name is self.__default:
+        if isinstance(name, utils.ReprObject):
```
shoyer: I would rather keep this as recognizing the sentinel value.

xarray/core/dataset.py
```diff
@@ -369,24 +400,24 @@ def __init__(self, data_vars=None, coords=None, attrs=None,
         else:
             compat = 'broadcast_equals'
 
-        self._variables = OrderedDict()  # type: OrderedDict[Any, Variable]
-        self._coord_names = set()
-        self._dims = {}  # type: Dict[Any, int]
+        self._variables = \
+            OrderedDict()  # type: OrderedDict[Hashable, Variable]
```
shoyer: I think anything used as a dict key already has to be `Hashable`, which is why I used `Any` previously. But maybe type checkers handle these differently?

xarray/core/dataset.py
```diff
@@ -1839,8 +1871,9 @@ def relevant_keys(mapping):
             dim_name = dim
             dim_coord = None  # type: ignore
 
+        _ = list(non_indexed_dims)
```
shoyer: This line looks like a no-op?

xarray/core/dataset.py
```diff
@@ -1839,8 +1871,9 @@ def relevant_keys(mapping):
...
         reordered = self.transpose(
-            *(list(indexer_dims) + list(non_indexed_dims)))
+            *list(indexer_dims), *list(non_indexed_dims))
```
shoyer: Nice :)

xarray/core/dataset.py
```diff
@@ -2086,6 +2119,8 @@ def interp(self, coords=None, method='linear', assume_sorted=False,
         """
         from . import missing
 
+        if kwargs is None:
+            kwargs = {}
```
shoyer: Good catch!

xarray/core/dataset.py
max-sixty: I think they'll be equivalent. Which do you think we should use?

shoyer: I like `Any` just because it's shorter. But not a strong preference.

max-sixty:
Great, I've resolved those comments @shoyer . 

@crusaderky we should be ready to go - maybe take a glance over to ensure you're happy.

I added a whatsnew with both of us - hope that's OK

shoyer:
@max-sixty do you want to push those changes to @crusaderky's branch? As a maintainer for xarray, I think you should have push permissions.

max-sixty:
Ah it works! Didn't know permissions transferred to forks.

@crusaderky hope that's OK, not wanting to step on your toes here...

xarray/core/dataset.py
```diff
@@ -333,7 +338,13 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
     _coarsen_cls = rolling.DatasetCoarsen
     _resample_cls = resample.DatasetResample
 
-    def __init__(self, data_vars=None, coords=None, attrs=None,
+    def __init__(self,
+                 data_vars: Optional[Mapping[Hashable, Union[
+                     'DataArray', Variable,
+                     Tuple[Hashable, Any],
+                     Tuple[Tuple[Hashable, ...], Any]]]] = None,
```
shoyer: This one also can handle any tuples `args` that you can use like `Variable(*args)` (maybe we could define a `VariableArgs` alias of some sort?)

shoyer:
By the way, I think I can write a script based on pep8speaks that writes/updates GitHub comments from within from Travis or Azure Pipelines. We could use that for surfacing mypy results in pull requests. I will probably give that a try over the next few days....

max-sixty:
> By the way, I think I can write a script based on pep8speaks that writes/updates GitHub comments from within from Travis or Azure Pipelines. We could use that for surfacing mypy results in pull requests. I will probably give that a try over the next few days....

That would be v cool. I'm kinda surprised something like pep8speaks doesn't do this already tbh...

shoyer:
> > By the way, I think I can write a script based on pep8speaks that writes/updates GitHub comments from within from Travis or Azure Pipelines. We could use that for surfacing mypy results in pull requests. I will probably give that a try over the next few days....
> 
> That would be v cool. I'm kinda surprised something like pep8speaks doesn't do this already tbh...

I was considering adding this into pep8speaks, but I think this fundamentally incompatible with pep8speak's design. It runs a handful of pre-specified linters, without actually installing/running user code. But something like mypy needs to actually have xarray and its dependencies installed to work.

xarray/core/dataarray.py
```diff
@@ -163,10 +182,20 @@ class DataArray(AbstractArray, DataWithCoords):
     _resample_cls = resample.DataArrayResample
 
     dt = property(DatetimeAccessor)
-    str = property(StringAccessor)
 
-    def __init__(self, data, coords=None, dims=None, name=None,
-                 attrs=None, encoding=None, indexes=None, fastpath=False):
+    def __init__(self, data: Any,
+                 coords: Union[
+                     Sequence[Tuple[Hashable, Any]],
```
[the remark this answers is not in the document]
max-sixty: Erm, in the latest it's just `Tuple`, is that OK? (I think maybe I pushed an older one first and you looked over that)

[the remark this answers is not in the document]
shoyer: `Tuple` should be fine, but it looks like that commit isn't on this branch yet

max-sixty:
> I was considering adding this into pep8speaks, but I think this fundamentally incompatible with pep8speak's design. It runs a handful of pre-specified linters, without actually installing/running user code. But something like mypy needs to actually have xarray and its dependencies installed to work.

Yes, good point. 

I haven't spent much time with the GitHub Checks, but potentially could be a good fit - one issue with the simpler "Update the comment" model is that it doesn't encode to GH whether it's safe to merge, so projects need to duplicate the test in CI

xarray/core/dataset.py
max-sixty: Yes I added a comment re VariableArgs, def would be good to move some of these long definitions into a single place (but if you're OK with leaving until the next PR let's do that)

xarray/core/dataarray.py
```diff
@@ -163,11 +181,27 @@ class DataArray(AbstractArray, DataWithCoords):
...
     dt = property(DatetimeAccessor)
-    str = property(StringAccessor)
 
-    def __init__(self, data, coords=None, dims=None, name=None,
-                 attrs=None, encoding=None, indexes=None, fastpath=False):
+    def __init__(
+        self,
+        data: Any,
+        coords: Union[
+            Sequence[Tuple],
```
max-sixty: @shoyer updated here, is this correct?

shoyer: Yes looks good here

max-sixty:
(AppVeyor fail looks unrelated)

shoyer:
Thanks @max-sixty and @crusaderky !

max-sixty:
Thanks @crusaderky !

xarray/core/dataset.py
gimperiale: It is true that a key of an ordereddict is always hashable. However, mypy doesn't know it, and the moment you pull it out of the dict and you pass it to some helper function it's nice to track what it is. Also, a Mapping does not require its keys to be hashable, but when xarray accepts a Mapping input (such as what is accepted by DataArray.attrs.setter, DataArray.coords.setter, etc.)  it should always state Mapping[Hashable, ...] since it's going to be used to feed an OrderedDict.

Issue #1463, opened 2017-06-22
copy / deepcopy not deepcopying coords?


robintibor:
I don't know if this is intentional, I thought that ```arr.copy(deep=True)``` or ```deepcopy(arr)``` would give me completely independent copies of a DateArray, but this seems not be the case?

```python
>>> import xarray as xr
>>> xarr1 = xr.DataArray([1,2], coords=dict(x=[0,1]), dims=('x',))
>>> xarr1.x.data[0]
0
>>> xarr2 = xarr1.copy(deep=True) #xarr2 = deepcopy(xarr1) -> leads to same result
>>> xarr2.x.data[0] = -1
>>> xarr1.x.data[0]
-1
```

How can I create completely independent copies of a DateArray? I wrote a function for this, but don't know if this really always does what I expect and if there is a more elegant way?
```python
def deepcopy_xarr(xarr):
    """
    Deepcopy for xarray that makes sure coords and attrs
    are properly deepcopied.
    With normal copy method from xarray, when i mutated
    xarr.coords[coord].data it would also mutate in the copy
    and vice versa.
    Parameters
    ----------
    xarr: DateArray

    Returns
    -------
    xcopy: DateArray
        Deep copy of xarr
    """
    xcopy = xarr.copy(deep=True)

    for dim in xcopy.coords:
        xcopy.coords[dim].data = np.copy(xcopy.coords[dim].data)
    xcopy.attrs = deepcopy(xcopy.attrs)
    for attr in xcopy.attrs:
        xcopy.attrs[attr] = deepcopy(xcopy.attrs[attr])
    return xcopy
```

shoyer:
This seems like a bug.

I suspect the problem is in `Variable.copy` on [these lines](https://github.com/pydata/xarray/blob/22ff955d53e253071f6e4fa849e5291d0005282a/xarray/core/variable.py#L467-L469) (which should probably just be removed). Coordinates store their data in `pandas.Index` objects, which are supposed to be immutable. But apparently that's not necessarily the case.

fujiisoup:
We do not allow to assign value in `IndexVariable` as `pandas.Index` is immutable (assigning value raises a `TypeError`),
but we can actually do this from `.data` attribute ([this line](https://github.com/pydata/xarray/blob/master/xarray/core/variable.py#L1422)).
(Our `IndexVariable` is not immutable...)

I think we should copy also `IndexVariable` not keeping the original reference.

pletchm:
[...]

PR #2930, opened 2019-04-29
Bugfix/coords not deep copy

pletchm:
This pull request fixes a bug that prevented making a complete deep copy of a `DataArray` or `Dataset`, because the `coords` weren't being deep copied. It took a small fix in the `IndexVariable.copy` method. This method now allows both deep and shallow copies of coords to be made.

This pull request corresponds to this issue https://github.com/pydata/xarray/issues/1463.

 - [x] Tests added


pletchm:
I'm perplexed by why the build and tests are failing on this. I'm going to try 3 things:
1) Make a new branch from master and PR (making sure my fork master is synced up with the upstream master) to see if the tests/build pass without any additional commits
2) Commit my non-test changes and see if the build/tests pass
3) Commit my tests and see if the build/tests pass
Then I'll rebase my 2 commits into 1 commit.

mgaspary:
seems like a same issue again:
```python
import xarray as xr

xarr1 = xr.DataArray(
    np.zeros([2]), 
    coords=dict(x=[0.0, 1.0]), # important to use 'float' here! with 'int' it is working fine
    dims=("x")
)
print(xarr1.x.data[0]) # 0.0

xarr2 = xarr1.copy(deep=True)
xarr2.x.data[0] = 45
print(xarr1.x.data[0]) # gives 45
```

PR #2935, opened 2019-05-02
plot: If provided with colormap do not modify it.

### LINKED ISSUE #2932: Facetgrid: colors beyond range (extend) not saturated

Screen Shot 2019-04-30 at 11 59 30

Minimal example here:
https://github.com/lvankampenhout/bug-reports/blob/master/Facetgrid_cmap_extend.ipynb

#### Problem description

The extreme colors of neither the pcolormesh or colorbar (using extend='both') are not saturated as they should when faceting.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.8 |Anaconda custom (x86_64)| (default, Dec 29 2018, 19:04:46) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
python-bits: 64
OS: Darwin
OS-release: 17.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: C
LANG: None
LOCALE: None.None
libhdf5: 1.10.1
libnetcdf: 4.4.1.1

xarray: 0.12.1
pandas: 0.23.4
numpy: 1.14.2
scipy: 0.18.1
netCDF4: 1.3.1
pydap: None
h5netcdf: None
h5py: 2.7.1
Nio: None
zarr: None
cftime: 1.0.0b1
nc_time_axis: None
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.0
dask: 0.13.0
distributed: None
matplotlib: 3.0.2
cartopy: 0.16.0
seaborn: 0.7.1
setuptools: 38.5.1
pip: 9.0.1
conda: 4.6.14
pytest: 3.0.5
IPython: 5.1.0
sphinx: 1.5.1
</details>

dcherian:
- [x] Closes #2932
 - [x] Tests added

If provided a `Colormap` we don't override it anymore. facetgrid determines a colormap first and passes that down, so this prevents overwriting.


xarray/tests/test_plot.py
```diff
@@ -4,12 +4,10 @@
 import numpy as np
 import pandas as pd
 import pytest
-from numpy.testing import assert_array_equal
```
dcherian: lint fixes

mathause:
Thanks for the PR, however this did not work for the minimal example @lvankampenhout provided because `isinstance(mpl.colors.ListedColormap, mpl.colors.Colormap)` returns `False` (and you get a ListedColormap when specifying levels).

dcherian:
:man_facepalming: Thanks @mathause. I was too sleepy I guess.

The example now works; but the test doesn't work. It should fail for cmap='RdBu_r' but doesn't locally. Any suggestions would be helpful

dcherian:
Never mind; should be all good now.

xarray/plot/utils.py
```diff
@@ -265,7 +265,13 @@ def _determine_cmap_params(plot_data, vmin=None, vmax=None, cmap=None,
     if extend is None:
         extend = _determine_extend(calc_data, vmin, vmax)
 
-    if levels is not None or isinstance(norm, mpl.colors.BoundaryNorm):
+    cmap_is_colormap = (
+        isinstance(cmap, mpl.colors.LinearSegmentedColormap)
+        or isinstance(cmap, mpl.colors.ListedColormap)
+    )
```
fmaussion: Both ihnerit from `mpl.colors.Colormap`, and after a quick check it seems these are the only two in mpl, so it might be possible to test for this only

dcherian: OK. done. This is what I tried first but it didn't work; perhaps there was some other bug. The example works now and the tests pass locally.

dcherian:
Failed test is #2940. I'll merge in a few days if no one else has any comments

PR #2942, opened 2019-05-06
Fix rolling operation with dask and bottleneck

### LINKED ISSUE #2940: test_rolling_wrapped_dask is failing with dask-master
The `test_rolling_wrapped_dask` tests in `test_dataarray.py` are failing with dask master, e.g., as seen here:
https://travis-ci.org/pydata/xarray/jobs/527936531

I reproduced this locally. `git bisect` identified the culprit as https://github.com/dask/dask/pull/4756.

The source of this issue on the xarray side appears to be these lines:
https://github.com/pydata/xarray/blob/dd99b7d7d8576eefcef4507ae9eb36a144b60adf/xarray/core/rolling.py#L287-L291

In particular, we are currently `padded` as an xarray.DataArray object, not a dask array. Changing this to `padded.data` shows that passing an actual dask array to `dask_array_ops.rolling_window` results in failing tests.

@fujiisoup @jhamman any idea what's going on here?

fujiisoup:
- [x] Closes #2940
 - [x] Tests added

Fix for #2940 
It looks that there was a bug in the previous logic, but I am not sure why it was working...

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -27,10 +27,13 @@ Enhancements
 - Character arrays' character dimension name decoding and encoding handled by
   ``var.encoding['char_dim_name']`` (:issue:`2895`)
   By `James McCreight <https://github.com/jmccreight>`_.
-   
+
 Bug fixes
 ~~~~~~~~~
 
+- a bug fix in rolling with dask and bottleneck
+  (:issue:`2940`)
+  By `Keisuke Fujii <https://github.com/fujiisoup>`_.
 - indexing with an empty list creates an object with zero-length axis (:issue:`2882`)
   By `Mayeul d'Avezac <https://github.com/mdavezac>`_.
 - Return correct count for scalar datetime64 arrays (:issue:`2770`)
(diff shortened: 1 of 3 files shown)


shoyer:
@fujiisoup, thanks for looking into this! Unfortunately these tests still seem to still be failing on Travis.

max-sixty:
@fujiisoup I mistakenly removed the dask-master tests (I moved the dask test from the test matrix to the ignored failures, rather than adding it to ignored failures)

Here's a tiny patch which restores it, if you want to add this to your PR:

```patch

diff --git a/.travis.yml b/.travis.yml
index 913c5e1c..8249b134 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,12 @@ matrix:
   - env: CONDA_ENV=py36-pandas-dev
   - env: CONDA_ENV=py36-rasterio
   - env: CONDA_ENV=py36-zarr-dev
+  - env: CONDA_ENV=py36-dask-dev
   - env: CONDA_ENV=docs
   - env: CONDA_ENV=lint
   - env: CONDA_ENV=py36-hypothesis
 
   allow_failures:
-  - env: CONDA_ENV=py36-dask-dev
   - env:
     - CONDA_ENV=py36
     - EXTRA_FLAGS="--run-flaky --run-network-tests"
```

zbarry:
Hey just wanted to chime in and say it appears that commit https://github.com/fujiisoup/xarray/commit/098daf3020e817d03a17a81963ad3cf831fbb48c is still losing chunking for me, as far as I can tell when running it with dask distributed / dask jobqueue. I can do some extending testing & look further into this if people have some suggestions for how to go about it.

shoyer:
I did a little more looking into this, and I unfortunately I think the necessary fix is a little more involved. For now, I think we should disable `rolling()` with dask arrays to avoid computing incorrect results: https://github.com/pydata/xarray/issues/2940#issuecomment-504058279

fujiisoup:
Hi. Sorry for my slow responce...

lf I understood correctly, it happens with dask and bottleneck.
So, probably we can skip using bottleneck if the backend array is dask array?

shoyer:
@fujiisoup yes, indeed, that's a much better temporary fix! See https://github.com/pydata/xarray/pull/3040.

shoyer:
This is subsumed by https://github.com/pydata/xarray/pull/3040

PR #2953, opened 2019-05-09
Mark test for copying coords of dataarray and dataset with xfail

pletchm:
Mark test for copying coords of dataarray and dataset with `xfail`. It looks like the test fails for the shallow copy, and apparently only on Windows for some reason. In Windows coords seem to be immutable unless it's one dataarray deep copied from another (which is why only the deep=False test fails). So I decided to just mark the tests as xfail for now (but I'd be happy to create an issue and look into it more in the future).


shoyer:
thanks!

Issue #2436, opened 2018-09-24
allow access to time attributes from netcdf files


sbiner:
I like the automatic decoding of the time variable when reading netcdf files but I often need to keep the calendar attribute of the time variable.

Could it be possible to keep those attributes in the DataSet/DataArray return by open_dataset?

sbiner:
It would be ok but it is (or looks) empty when I use open_dataset()

spencerkclark:
@sbiner are you looking at the encoding attribute of the full Dataset or the time variable?  The time variable should retain the calendar encoding (the Dataset will not).  E.g.:
```
In [1]: import cftime

In [2]: import numpy as np

In [3]: import xarray as xr

In [4]: units = 'days since 2000-02-25'

In [5]: times = cftime.num2date(np.arange(7), units=units, calendar='365_day')

In [6]: da = xr.DataArray(np.arange(7), coords=[times], dims=['time'], name='a')

In [7]: da.to_netcdf('data-noleap.nc')

In [8]: ds = xr.open_dataset('data-noleap.nc')

In [9]: ds.encoding['calendar']
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-38-677c245c7bb8> in <module>()
----> 1 default.encoding['calendar']

KeyError: 'calendar'

In [10]: ds.time.encoding['calendar']
Out[10]: u'noleap'
```

sbiner:
@spencerkclark Yes I was looking at  time.encoding. Following you example I did some tests and the problem is related to the fact that I am opening multiple netCDF files with open_mfdataset. Doing so  time.encoding is empty while it is as expected when opening any of the files with open_dataset instead.

TomNicholas:
If `open_mfdataset()` is actually dropping the encoding, then this is an issue related to #1614. That's because in `open_mfdataset()` while the attrs are [explicitly set](https://github.com/pydata/xarray/blob/c2ce5ea83b5924302653c8dfba7de68c7d98c942/xarray/backends/api.py#L656) to those of the first supplied dataset, I don't see any similar explicit treatment of the encoding. I think that means the encoding is being set by what happens inside the core of `auto_combine()`, and is presumably being lost upon some of the `concat` or `merge` operations which happen inside `auto_combine()`.

So I think to fix this then either `open_mfdataset()` should contain explicit treatment of the encoding, or the rules for propagating the encoding through the `auto_combine()` should be solidified.

[...]

PR #2965, opened 2019-05-15
Don't set encoding attributes on bounds variables.

### LINKED ISSUE #2921: to_netcdf with decoded time can create file with inconsistent time:units and time_bounds:units

```python
import numpy as np
import xarray as xr

# create time and time_bounds DataArrays for Jan-1850 and Feb-1850
time_bounds_vals = np.array([[0.0, 31.0], [31.0, 59.0]])
time_vals = time_bounds_vals.mean(axis=1)

time_var = xr.DataArray(time_vals, dims='time',
                        coords={'time':time_vals})
time_bounds_var = xr.DataArray(time_bounds_vals, dims=('time', 'd2'),
                               coords={'time':time_vals})

# create Dataset of time and time_bounds
ds = xr.Dataset(coords={'time':time_var}, data_vars={'time_bounds':time_bounds_var})
ds.time.attrs = {'bounds':'time_bounds', 'calendar':'noleap',
                 'units':'days since 1850-01-01'} 

# write Jan-1850 values to file 
ds.isel(time=slice(0,1)).to_netcdf('Jan-1850.nc', unlimited_dims='time')

# write Feb-1850 values to file
ds.isel(time=slice(1,2)).to_netcdf('Feb-1850.nc', unlimited_dims='time')

# use open_mfdataset to read in files, combining into 1 Dataset
decode_times = True
decode_cf = True
ds = xr.open_mfdataset(['Jan-1850.nc', 'Feb-1850.nc'],
                       decode_cf=decode_cf, decode_times=decode_times)

# write combined Dataset out
ds.to_netcdf('JanFeb-1850.nc', unlimited_dims='time')
```

#### Problem description

The above code initially creates 2 netCDF files, for Jan-1850 and Feb-1850, that have the variables `time` and `time_bounds`, and `time:bounds='time_bounds'`. It then reads the 2 files back in as a single Dataset, using `open_mfdataset`, and this Dataset is written back out to a netCDF file. ncdump of this final file is
```
netcdf JanFeb-1850 {
dimensions:
	time = UNLIMITED ; // (2 currently)
	d2 = 2 ;
variables:
	int64 time(time) ;
		time:bounds = "time_bounds" ;
		time:units = "hours since 1850-01-16 12:00:00.000000" ;
		time:calendar = "noleap" ;
	double time_bounds(time, d2) ;
		time_bounds:_FillValue = NaN ;
		time_bounds:units = "days since 1850-01-01" ;
		time_bounds:calendar = "noleap" ;
data:

 time = 0, 708 ;

 time_bounds =
  0, 31,
  31, 59 ;
}
```
The problem is that the units attribute for `time` and `time_bounds` are different in this file, contrary to what [CF conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries) requires.

The final call to `to_netcdf` is creating a file where `time`'s units (and type) differ from what they are in the intermediate files. These transformations are not being applied to `time_bounds`.

While the change to `time`'s type is not necessarily an issue, I do find it surprising.

This inconsistency goes away if either of `decode_times` or `decode_cf` is set to `False` in the python code above. In particular, the transformations to `time`'s units and type do not happen.

The inconsistency also goes away if `open_mfdataset` opens a single file. In this scenario also, the transformations to `time`'s units and type do not happen.

I think that the desired behavior is to either not apply the units and type transformations to `time`, or to also apply them to `time_bounds`. The first option would be consistent with the current single-file behavior.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 3.12.62-60.64.8-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.2

xarray: 0.12.1
pandas: 0.24.2
numpy: 1.16.2
scipy: 1.2.1
netCDF4: 1.4.2
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudonetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 1.1.5
distributed: 1.26.1
matplotlib: 3.0.3
cartopy: None
seaborn: None
setuptools: 40.8.0
pip: 19.0.3
conda: None
pytest: 4.3.1
IPython: 7.4.0
sphinx: None
</details>

dcherian:
Here's a proposed fix for #2436 and #2921. Ping @spencerkclark @mathause @klindsay28

1. Removes certain attributes from bounds variables on encode.
2. open_mfdataset: Sets encoding on variables based on encoding in first file.

 - [x] Closes #2921
 - [x] Tests added

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -724,6 +724,8 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
 
     combined._file_obj = _MultiFileCloser(file_objs)
     combined.attrs = datasets[0].attrs
+    for v in combined:
+        combined[v].encoding = datasets[0][v].attrs
     return combined
 
 
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -185,12 +185,64 @@ def _finalize_store(write, store):
     store.close()
 
 
+def load_dataset(filename_or_obj, **kwargs):
+    """Open, load into memory, and close a Dataset from a file or file-like
+    object.
+
+    This is a thin wrapper around :py:meth:`~xarray.open_dataset`. It differs
+    from `open_dataset` in that it loads the Dataset into memory, closes the
+    file, and returns the Dataset. In contrast, `open_dataset` keeps the file
+    handle open and lazy loads its contents. All parameters are passed directly
+    to `open_dataset`. See that documentation for further details.
+
+    Returns
+    -------
+    dataset : Dataset
+        The newly created Dataset.
+
+    See Also
+    --------
+    open_dataset
+    """
+    if 'cache' in kwargs:
+        raise TypeError('cache has no effect in this context')
+
+    with open_dataset(filename_or_obj, **kwargs) as ds:
+        return ds.load()
+
+
+def load_dataarray(filename_or_obj, **kwargs):
+    """Open, load into memory, and close a DataArray from a file or file-like
+    object containing a single data variable.
+
+    This is a thin wrapper around :py:meth:`~xarray.open_dataarray`. It differs
+    from `open_dataarray` in that it loads the Dataset into memory, closes the
+    file, and returns the Dataset. In contrast, `open_dataarray` keeps the file
+    handle open and lazy loads its contents. All parameters are passed directly
+    to `open_dataarray`. See that documentation for further details.
+
+    Returns
+    -------
+    datarray : DataArray
+        The newly created DataArray.
+
+    See Also
+    --------
+    open_dataarray
+    """
+    if 'cache' in kwargs:
+        raise TypeError('cache has no effect in this context')
+
+    with open_dataarray(filename_or_obj, **kwargs) as da:
+        return da.load()
+
+
 def open_dataset(filename_or_obj, group=None, decode_cf=True,
                  mask_and_scale=None, decode_times=True, autoclose=None,
                  concat_characters=True, decode_coords=True, engine=None,
                  chunks=None, lock=None, cache=None, drop_variables=None,
                  backend_kwargs=None, use_cftime=None):
-    """Load and decode a dataset from a file or file-like object.
+    """Open and decode a dataset from a file or file-like object.
 
     Parameters
     ----------
@@ -406,7 +458,8 @@ def open_dataarray(filename_or_obj, group=None, decode_cf=True,
                    concat_characters=True, decode_coords=True, engine=None,
                    chunks=None, lock=None, cache=None, drop_variables=None,
                    backend_kwargs=None, use_cftime=None):
-    """Open an DataArray from a netCDF file containing a single data variable.
+    """Open an DataArray from a file or file-like object containing a single
+    data variable.
 
     This is designed to read netCDF files with only one data variable. If
     multiple variables are present then a ValueError is raised.
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -725,7 +725,8 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
     combined._file_obj = _MultiFileCloser(file_objs)
     combined.attrs = datasets[0].attrs
     for v in combined:
-        combined[v].encoding = datasets[0][v].attrs
+        if v in datasets[0]:
+            combined[v].encoding = datasets[0][v].encoding
     return combined
 
 
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -724,7 +724,7 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
 
     combined._file_obj = _MultiFileCloser(file_objs)
     combined.attrs = datasets[0].attrs
-    for v in combined:
+    for v in combined.variables:
         if v in datasets[0]:
             combined[v].encoding = datasets[0][v].encoding
     return combined
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -724,9 +724,6 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
 
     combined._file_obj = _MultiFileCloser(file_objs)
     combined.attrs = datasets[0].attrs
-    for v in combined.variables:
-        if v in datasets[0]:
-            combined[v].encoding = datasets[0][v].encoding
     return combined
 
 
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -354,6 +354,7 @@ def _update_bounds_attributes(variables):
                 if 'calendar' in attrs:
                     bounds_attrs.setdefault('calendar', attrs['calendar'])
 
+
 def _update_bounds_encoding(variables):
     """Adds time encoding to time bounds variables.
 
@@ -395,7 +396,9 @@ def _update_bounds_encoding(variables):
                 bounds_encoding = variables[attrs['bounds']].encoding
                 bounds_encoding.setdefault('units', encoding['units'])
                 if 'calendar' in encoding:
-                    bounds_encoding.setdefault('calendar', encoding['calendar'])
+                    bounds_encoding.setdefault('calendar',
+                                               encoding['calendar'])
+
 
 def decode_cf_variables(variables, attributes, concat_characters=True,
                         mask_and_scale=True, decode_times=True,
@@ -665,7 +668,7 @@ def cf_encoder(variables, attributes):
     decode_cf_variable, encode_cf_variable
     """
 
-    # add encoding for time bounds variables so that they are encoded correctly.
+    # add encoding for time bounds variables if present.
     _update_bounds_encoding(variables)
 
     new_vars = OrderedDict((k, encode_cf_variable(v, name=k))
--- a/xarray/conventions.py
+++ b/xarray/conventions.py
@@ -235,6 +235,7 @@ def encode_cf_variable(var, needs_copy=True, name=None):
     var = maybe_default_fill_value(var)
     var = maybe_encode_bools(var)
     var = ensure_dtype_not_object(var, name=name)
+
     return var
 
 
@@ -599,7 +600,7 @@ def cf_encoder(variables, attributes):
     as possible.  This includes masking, scaling, character
     array handling, and CF-time encoding.
 
-    Decode a set of CF encoded variables and attributes.
+    Encode a set of CF encoded variables and attributes.
 
     See Also, decode_cf_variable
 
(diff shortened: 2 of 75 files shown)


mathause:
Thanks for putting this together.

I think this can work as long as `time` has an encoding - however, if it doesn't  we might end up with different `units` for `time` and `time_bounds` again. I am not sure, actually, just something to check.

dcherian:
Yeah, this still doesn't work as @klindsay28 just pointed out to me. 

I still don't understanding how the `units` attribute/encoding is being treated. I'll get back to this soon.

xarray/conventions.py
```diff
@@ -621,4 +621,17 @@ def cf_encoder(variables, attributes):
...
+    # Remove attrs from bounds variables before encoding
+    # See issue #2921
+    for var in new_vars.values():
+        bounds = var.attrs['bounds'] if 'bounds' in var.attrs else None
+        if bounds and bounds in new_vars:
+            # see http://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries # noqa
+            for attr in ['units', 'standard_name', 'axis', 'positive',
+                         'calendar', 'long_name', 'leap_month', 'leap_year',
+                         'month_lengths']:
+                if attr in new_vars[bounds].attrs:
```
spencerkclark: Should we be a little more careful about which attributes we drop here?
```python
if attr in new_vars[bounds].attrs and attr in var.attrs:
    if new_vars[bounds].attrs[attr] == var[attr]:
        new_vars[bounds].attrs.pop(attr)
```
I'm thinking of the case where we might read in a file that is slightly CF-incompliant in this aspect, e.g. if the `units` encoding were different for the time and time bounds coordinates.  I think in that case if we dropped the `units` attribute from the time bounds coordinate, we would no longer be able to accurately round trip the file.

dcherian: Good idea! I've implemented this now.

dcherian:
Thanks @spencerkclark, I'm going by the example in #2921. I've added that now as a test. I also test for the slight non-compliance case where time_bounds has different units already set. 

@mathause would this test catch the issue you were encountering?

mathause:
Great - this does indeed solve my problem (read multiple files with `open_mfdataset` and write as one). However, it still 'fails' - is not cf-compiant to be more precise - for a new dataset created in xarray, see:

``` python
import xarray as xr
import pandas as pd
time = pd.date_range('2000-01-16', periods=1)
time_bounds = pd.date_range('2000-01-01', periods=2, freq='MS')

ds = xr.Dataset(dict(time=time, time_bounds=time_bounds))
ds.time.attrs['bounds'] = 'time_bounds'

xr.conventions.cf_encoder(ds.variables, ds.attrs)
```
(compare the units). This might be difficult to solve, as xarray currently assumes all variables can be encoded independently.

spencerkclark:
> This might be difficult to solve, as xarray currently assumes all variables can be encoded independently.

I was hoping someone wouldn't bring that case up :).  But I agree I think it's something we should discuss.

I'm sort of torn on whether it is worth the complexity.  Part of me feels like if a user is concerned/clever enough to explicitly create such linking attributes (e.g. `'bounds'` pointing to `'time_bounds'`) in their Datasets that we could leave it as their responsibility to make sure the `'units'` and `'calendar'` encodings match for each of the variables as well.  For example, while a bit verbose, this would work as desired:
```python
import xarray as xr
import pandas as pd
time = pd.date_range('2000-01-16', periods=1)
time_bounds = pd.date_range('2000-01-01', periods=2, freq='MS')

ds = xr.Dataset(dict(time=time, time_bounds=time_bounds))
ds.time.encoding['bounds'] = 'time_bounds'
ds.time.encoding['units'] = 'days since 2000-01-01'
ds.time.encoding['calendar'] = 'proleptic_gregorian'

ds.time_bounds.encoding['units'] = ds.time.encoding['units']
ds.time_bounds.encoding['calendar'] = ds.time.encoding['calendar']
```
That said, that case is something that in principle we *could* address, and maybe it is worth thinking about if we ever consider increasing the functionality related to cell bounds coordinates (e.g. https://github.com/pydata/xarray/issues/1475).

xarray/conventions.py
```diff
@@ -617,8 +617,23 @@ def cf_encoder(variables, attributes):
...
+        bounds = var.attrs['bounds'] if 'bounds' in var.attrs else None
+        if bounds and bounds in new_vars:
+            # see http://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries # noqa
+            for attr in ['units', 'standard_name', 'axis', 'positive',
+                         'calendar', 'long_name', 'leap_month', 'leap_year',
+                         'month_lengths']:
+                if attr in new_vars[bounds].attrs and attr in var.attrs:
+                    if new_vars[bounds].attrs[attr] == var.attrs[attr]:
+                        new_vars[bounds].attrs.pop(attr)
+
```
mathause: Could we issue a warning here?

``` python

else:
    warning.warn("The attribute 'time' is not the same in 'units' and it's associated bounds "
"'time_bnds', which is not cf-compliant.")

```

mathause:
Yes, I agree - I don't intend to actually create such a dataset by hand anytime soon ;) but it would be super hard to track this issue down - can we issue a warning instead? See my inline comment.

dcherian:
This case might actually be really easy to fix since we already have `xr.conventions._update_bounds_attributes()` from Fabien's previous PR (see below).

I think we would just need to call this prior to calling `encode_cf_variable()` in https://github.com/pydata/xarray/blob/612d390f925e5490314c363e5e368b2a8bd5daf0/xarray/conventions.py#L622

``` python
def _update_bounds_attributes(variables):
    """Adds time attributes to time bounds variables.

    Variables handling time bounds ("Cell boundaries" in the CF
    conventions) do not necessarily carry the necessary attributes to be
    decoded. This copies the attributes from the time variable to the
    associated boundaries.

    See Also:

    http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/
         cf-conventions.html#cell-boundaries

    https://github.com/pydata/xarray/issues/2565
    """


    # For all time variables with bounds
    for v in variables.values():
        attrs = v.attrs
        has_date_units = 'units' in attrs and 'since' in attrs['units']
        if has_date_units and 'bounds' in attrs:
            if attrs['bounds'] in variables:
                bounds_attrs = variables[attrs['bounds']].attrs
                bounds_attrs.setdefault('units', attrs['units'])
                if 'calendar' in attrs:
                    bounds_attrs.setdefault('calendar', attrs['calendar'])
```

spencerkclark:
> I think we would just need to call this prior to calling `encode_cf_variable()`

I'm afraid it's not quite that simple :).  

In cases where someone creates a datetime-like variable in memory (e.g. with `pd.date_range` in @mathause's example), unless they explicitly add encoding attributes, `'units'` and `'calendar'` will need to be computed on the fly.  This happens inside of `encode_cf_variable`, currently on a per-variable basis, so trying to make sure the attributes are equal before that step will not help, because they might not exist.

dcherian:
Thanks @spencerkclark & @mathause. I now understand the issue better. #2921 was confusing in that @klindsay28 was trying to write an encoded dataset.

I've updated the tests to use @mathuse's example and now throw a warning when `encoding.units` is not specified for variables with a `bounds` attribute.

Do you think this is a good approach?

xarray/conventions.py
```diff
@@ -353,6 +354,43 @@ def _update_bounds_attributes(variables):
...
+    http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/
+         cf-conventions.html#cell-boundaries
+
+    https://github.com/pydata/xarray/issues/2565
+    """
+
+    # For all time variables with bounds
+    for v in variables.values():
+        attrs = v.attrs
```
spencerkclark: A general question -- would we consider `'bounds'` to be an encoding parameter (like `'units'` or `'calendar'`)?  In other words should we expect it to be in the `encoding` dictionary or `attrs` dictionary at this stage?  I feel like it may be more intuitive as part of `encoding`, but currently I know that we don't treat it that way when decoding files.

xarray/tests/test_coding_times.py
```diff
@@ -671,6 +672,48 @@ def test_decode_cf_time_bounds():
     _update_bounds_attributes(ds.variables)
 
 
+def test_encode_time_bounds():
```
spencerkclark: It looks like you'll need a `@requires_cftime_or_netCDF4` decorator here to make sure this test gets skipped without one of those installed.

xarray/conventions.py
```diff
@@ -353,6 +354,43 @@ def _update_bounds_attributes(variables):
...
+        encoding = v.encoding
+        has_date_units = 'units' in encoding and 'since' in encoding['units']
+        is_datetime_type = (np.issubdtype(v.dtype, np.datetime64) or
+                            contains_cftime_datetimes(v))
+
+        if is_datetime_type and not has_date_units and 'bounds' in attrs:
+            warnings.warn("Variable '{0}' has datetime units and a "
```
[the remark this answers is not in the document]
dcherian: Thanks!

xarray/conventions.py
dcherian: In my mental model, encoding attributes are those that control on-disk representation of the data. `bounds` counts as an attr to my mind since it's an attribute that links the variable to another variable. 

A definition or list of what goes in `encoding` and what goes in `attrs` would make a good addition to the docs.

xarray/conventions.py
dcherian: Do we need to? xarray allows writing CF-non-compliant files anyway...

xarray/conventions.py
spencerkclark: > In my mental model, encoding attributes are those that control on-disk representation of the data.

I think this is fair; I guess I was going off of the mental model of encoding parameters defined as "attributes that are potentially required for decoding all the variables in a file," in which case `'bounds'` could qualify.  I think your definition is probably cleaner, because it requires that encoding parameters control how the variable they are attached to is represented on disk (as opposed to another variable).

xarray/conventions.py
spencerkclark: I think the warning you added is the perfect approach.  It will still be issued in the case of @mathause's example, but will still allow a user to write a non-CF-compliant file without a warning if the encoding attributes do not need to be computed on the fly.

xarray/backends/api.py
```diff
@@ -777,6 +777,9 @@ def open_mfdataset(paths, chunks=None, concat_dim=_CONCAT_DIM_DEFAULT,
 
     combined._file_obj = _MultiFileCloser(file_objs)
     combined.attrs = datasets[0].attrs
+    for v in combined.variables:
+        if v in datasets[0]:
+            combined[v].encoding = datasets[0][v].encoding
```
shoyer: Are we sure this is a good idea in general? What if data is legitimately written with a different encoding?

It also looks a bit unrelated to the rest of this change.

dcherian: #2436 is about mfdataset losing all encoding information. So this is a band-aid that makes the arbitrary choice of using the first file (but the same choice as is made for attrs).

> "What if data is legitimately written with a different encoding?"

If encoding changes between different files read with mfdataset, then I guess we should drop it in that case. It seems like this problem should be solved at the merge/concat level than up here.

What do you think? Remove this change and merge?

dcherian: It's also a small part of #2921 (current behaviour is called surprising) that's why this change is in this PR.

shoyer: I also think it makes more sense to handle this at the merge/concat level rather than here. So I would lean towards removing this from this change (and then merging it).

PR #2967, opened 2019-05-16
Fix rolling.constuct() example

kmsquire:
The example was using the wrong name for the function (to_datarray), and used the wrong dimension for the window.

Cc: @fujiisoup, who added this in #1837


fujiisoup:
Thanks, @kmsquire

I didn't notice that this still survives. Merging.

PR #2968, opened 2019-05-16
Add examples for `DataArrayRolling.reduce()` 

kmsquire:
* ~~When reducing over a rolling window, `DataArrayRolling.reduce()`
  only returns values where the complete rolling window is fully
  conatined within the base array; the remaining locations where
  the window overlaps the array border are filled with `NaN`~~
* ~~However, in some situations, it is desirable to obtain values from the
  reducer function for all locations.  This commit adds a keyword
  argument, `include_border`, to `DataArrayRolling.reduce()` and
  `DataSetRolling.reduce()` to allow this behavior.~~

~~If this seems okay, I have some questions:~~

~~1. Should this behavior be added to specific reducing functions (sum, mean, etc.)?  As an alternative, all of the specialized functions for `DataArrayRolling` have corresponding numpy `nan*` functions, so support could be added for those.  (The current PR is still useful for other reducing functions, which is my use case.)~~
~~2. I don't use dask, so I don't know if this is needed/desired for the dask versions of these functions.~~

As @mathause pointed out, the functionality desired already exists by passing `min_periods=1` to `DataArray.rolling()`.  So this MR has been updated to simply include some examples for `DataArrayRolling.reduce()`

### DIFF
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -143,20 +143,35 @@ name of the dimension as a key (e.g. ``y``) and the window size as the value
 
     arr.rolling(y=3)
 
-The label position and minimum number of periods in the rolling window are
-controlled by the ``center`` and ``min_periods`` arguments:
-
-.. ipython:: python
-
-    arr.rolling(y=3, min_periods=2, center=True)
-
-Aggregation and summary methods can be applied directly to the ``Rolling`` object:
+Aggregation and summary methods can be applied directly to the ``Rolling``
+object:
 
 .. ipython:: python
 
     r = arr.rolling(y=3)
-    r.mean()
     r.reduce(np.std)
+    r.mean()
+
+Aggregation results are assigned the coordinate at the end of each window by
+default, but can be centered by passing ``center=True`` when constructing the
+ ``Rolling`` object:
+
+.. ipython:: python
+
+    r = arr.rolling(y=3, center=True)
+    r.mean()
+
+As can be seen above, aggregations of windows which overlap the border of the
+array produce ``nan``s.  Setting ``min_periods`` in the call to ``rolling``
+changes the minimum number of observations within the window required to have
+a value when aggregating:
+
+.. ipython:: python
+
+    r = arr.rolling(y=3, min_periods=2)
+    r.mean()
+    r = arr.rolling(y=3, center=True, min_periods=2)
+    r.mean()
 
 Note that rolling window aggregations are faster when bottleneck_ is installed.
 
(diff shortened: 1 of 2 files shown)


mathause:
I think you can keep the function signature of `reduce` but rather allow to set  `min_periods=0` in `__init__` (This is currently disallowed on [Line 77](https://github.com/pydata/xarray/blob/master/xarray/core/rolling.py#L77)). 

Then you change your code in `reduce` to
``` python
if self.min_periods == 0:
    return result

```

btw: If I understand correctly, this does not need to be added for the other functions `ds.rolling(...).mean()` calls `reduce` on [Line 238](https://github.com/pydata/xarray/blob/master/xarray/core/rolling.py#L238).

kmsquire:
@mathause, thank you, I didn't realize that.  I'll make those changes.

kmsquire:
Actually, every window has at least 1 valid value, so this change isn't really needed at all.

I'll remove the code changes, and change the added example to show a working example of what I wanted.

I will say that `min_periods` isn't the most obvious name for this parameter--I'm struggling to figure out how the number of valid observations in a window is related to periods or periodicity.

Would it be reasonable to deprecate this name and change it to something like `min_observations`?

kmsquire:
Changed to simply add examples to `DataArrayRolling.reduce()`.  I changed the original description accordingly.

dcherian:
@kmsquire thanks. Can you also add some text to https://xarray.pydata.org/en/stable/computation.html#rolling-window-operations (i.e. computation.rst) describing what `min_periods` does.

kmsquire:
@dcherian added.  Let me know if you want any changes.

doc/computation.rst
```diff
@@ -143,20 +143,34 @@ name of the dimension as a key (e.g. ``y``) and the window size as the value
...
 
 .. ipython:: python
 
-    arr.rolling(y=3, min_periods=2, center=True)
+    r = arr.rolling(y=3)
+    r.reduce(np.std)
+    r.mean()
 
-Aggregation and summary methods can be applied directly to the ``Rolling`` object:
+Aggregation results are at the end of each window by default, but can be
```
dcherian: "Aggregation results are assigned the coordinate at the end of each window by default"?

kmsquire: Sorry, saw this, but got busy and forgot to come back to it until now.  Just force-pushed this change.

dcherian:
Thanks @kmsquire . The tests failed but they seem unrelated (the docs build passed).

PR #2972, opened 2019-05-18
ENH: Preserve monotonic descending index order when merging

### LINKED ISSUE #2947: xr.merge always sorts indexes ascending

```python
import xarray as xr
import numpy as np


nrow, ncol = (4, 5)
dx, dy = (1.0, -1.0)
xmins = (0.0, 3.0, 3.0, 0.0)
xmaxs = (5.0, 8.0, 8.0, 5.0)
ymins = (0.0, 2.0, 0.0, 2.0)
ymaxs = (4.0, 6.0, 4.0, 6.0)
data = np.ones((nrow, ncol), dtype=np.float64)

das = []
for xmin, xmax, ymin, ymax in zip(xmins, xmaxs, ymins, ymaxs):
    kwargs = dict(
            name="example",
            dims=("y", "x"),
            coords={"y": np.arange(ymax, ymin, dy),
            "x": np.arange(xmin, xmax, dx)},
    )
    das.append(xr.DataArray(data, **kwargs))
    
xr.merge(das)

# This won't flip the coordinate:
xr.merge([das[0]))

```
#### Problem description
Let's say I have a number of geospatial grids that I'd like to merge (for example, loaded with `xr.open_rasterio`). To quote [https://www.perrygeo.com/python-affine-transforms.html](url)

> The typical geospatial coordinate reference system is defined on a cartesian plane with the 0,0 origin in the bottom left and X and Y increasing as you go up and to the right. But raster data, coming from its image processing origins, uses a different referencing system to access pixels. We refer to rows and columns with the 0,0 origin in the upper left and rows increase and you move down while the columns increase as you go right. Still a cartesian plane but not the same one.

`xr.merge` will alway return the result with ascending coordinates, which creates some issues / confusion later on if you try to write it back to a GDAL format, for example (I've been scratching my head for some time looking at upside-down .tifs for example).

#### Expected Output

I think the expected output for these geospatial grids is that; if you provide only DataArrays with positive dx, negative dy; that the merged result comes out with a positive dx and a negative dy as well.

When the DataArrays to merge are mixed in coordinate direction (some with ascending, some with descending coordinate values), defaulting to an ascending sort seems sensible.

#### A suggestion

I saw that the sort is occurring [here, in pandas](https://github.com/pandas-dev/pandas/blob/2bbc0c2c198374546408cb15fff447c1e306f99f/pandas/core/indexes/base.py#L2260-L2265); and that there's a `is_monotonic_decreasing` property in [pandas.core.indexes.base.Index](https://github.com/pandas-dev/pandas/blob/2bbc0c2c198374546408cb15fff447c1e306f99f/pandas/core/indexes/base.py#L1601)

I think this could work (it solves my issue at least), in [xarray.core.alignment](https://github.com/pydata/xarray/blob/5aaa6547cd14a713f89dfc7c22643d86fce87916/xarray/core/alignment.py#L125)
```python
                index = joiner(matching_indexes)
                if all(
                    (matching_index.is_monotonic_decreasing
                        for matching_index in matching_indexes)
                ):
                    index = index[::-1]
                joined_indexes[dim] = index
```
But I lack the knowledge to say whether this plays nice in all cases. And does `index[::-1]` return a view or a copy? (And does it matter?)

For reference this is what it looks like now:
```python
            if (any(not matching_indexes[0].equals(other)
                    for other in matching_indexes[1:]) or
                    dim in unlabeled_dim_sizes):
                if join == 'exact':
                    raise ValueError(
                        'indexes along dimension {!r} are not equal'
                        .format(dim))
                index = joiner(matching_indexes)
                joined_indexes[dim] = index
            else:
                index = matching_indexes[0]
```
It's also worth highlighting that the `else` causes, arguably, some inconsistency. If the indexes are equal, no reversion occurs.

Huite:
* Addresses GH2947

* When indexes were joined in a dataset merge, they would always get sorted in ascending order. This is awkward for geospatial grids, which are nearly always descending in the "y" coordinate.

* This also caused an inconsistency: when a merge is called on datasets with identical descending indexes, the resulting index is descending. When a merge is called with non-identical descending indexes, the resulting index is ascending.

* When indexes are mixed ascending and descending, or non-monotonic, the resulting index is still sorted in ascending order.

 - [x] Closes #2947
 - [x] Tests added

## Comments
I was doing some work and I kept running into the issue described at #2947, so I had a try at a fix. It was somewhat of a hassle to understand the issue because I kept running into seeming inconsistencies. This is caused by the fact that the joiner doesn't sort with a single index:

```python
def _get_joiner(join):
    if join == 'outer':
        return functools.partial(functools.reduce, operator.or_)
```
That makes sense, since I'm guessing `pandas.Index.union` isn't get called at all. (I still find the workings of  `functools` a little hard to infer.)

I also noticed that an outer join gets called with e.g. an `.isel` operation, even though there's only one index (so there's not really anything to join). However, skipping the join completely in that case makes several tests fail since dimension labels end up missing (I guess the `joiner` call takes care of it). 

It's just checking for the specific case now, but it feels like an very specific issue anyway...

The merge behavior is slightly different now, which is reflected in the updated test outcomes in `test_dataset.py`. These tests were turning monotonic decreasing indexes into an increasing index; now the decreasing order is maintained.

### DIFF
--- a/xarray/core/alignment.py
+++ b/xarray/core/alignment.py
@@ -123,6 +123,14 @@ def align(*objects, join='inner', copy=True, indexes=None, exclude=frozenset(),
                         'indexes along dimension {!r} are not equal'
                         .format(dim))
                 index = joiner(matching_indexes)
+                # Pandas.Index.union gets called for an outer join, and the index
+                # is sorted in ascending order. When all indexes are monotonic
+                # descending, it's desirable to keep maintain descending order.
+                # The ascending sort doesn't happen if only one index is joined.
+                if join == 'outer' and len(matching_indexes) > 1:
+                    if all((matching_index.is_monotonic_decreasing
+                            for matching_index in matching_indexes)):
+                        index = index[::-1]
                 joined_indexes[dim] = index
             else:
                 index = matching_indexes[0]
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -1851,13 +1851,13 @@ class TestDataset:
                                       coords={'y': [1, 2], 'x': [10, -3]})})
         x2, y2 = broadcast(x, y)
         expected_x2 = Dataset(
-            {'foo': DataArray([[3, 3], [2, 2], [1, 1], [np.nan, np.nan]],
+            {'foo': DataArray([[np.nan, np.nan], [1, 1], [2, 2], [3, 3]],
                               dims=['x', 'y'],
-                              coords={'y': [1, 2], 'x': [-3, -2, -1, 10]})})
+                              coords={'y': [1, 2], 'x': [10, -1, -2, -3]})})
         expected_y2 = Dataset(
             {'bar': DataArray(
-                [[2, 4], [np.nan, np.nan], [np.nan, np.nan], [1, 3]],
-                dims=['x', 'y'], coords={'y': [1, 2], 'x': [-3, -2, -1, 10]})})
+                [[1, 3], [np.nan, np.nan], [np.nan, np.nan], [2, 4]],
+                dims=['x', 'y'], coords={'y': [1, 2], 'x': [10, -1, -2, -3]})})
         assert_identical(expected_x2, x2)
         assert_identical(expected_y2, y2)
 
--- a/xarray/tests/test_merge.py
+++ b/xarray/tests/test_merge.py
@@ -67,6 +67,13 @@ class TestMergeFunction:
         with raises_regex(ValueError, 'indexes .* not equal'):
             xr.merge([ds, other], join='exact')
 
+    def test_merge_monotonic_descending_indexes(self):
+        ds = xr.Dataset(coords={'x': [2, 1]})
+        other = xr.Dataset(coords={'x': [3, 2]})
+        actual = xr.merge([ds, other])
+        expected = xr.Dataset(coords={'x': [3, 2, 1]})
+        assert expected.identical(actual)
+
     def test_merge_no_conflicts_single_var(self):
         ds1 = xr.Dataset({'a': ('x', [1, 2]), 'x': [0, 1]})
         ds2 = xr.Dataset({'a': ('x', [2, 3]), 'x': [1, 2]})


shoyer:
I'm not entirely sure this is a good idea -- a discrepancy in the behavior of an outer join between pandas seems non-ideal. Perhaps somebody else has opinions about whether this would be worthwhile?

If we do want to do this, let's put the change inside the `_get_joiner` function. Specifically, instead of 
```python
def _get_joiner(join):
    if join == 'outer':
        return functools.partial(functools.reduce, operator.or_)
```
you could write something like:
```python
def _outer_join(indexes):
    index = functools.reduce(indexes, operator.or_)
    # your logic for reverse sorting the result
    return index

def _get_joiner(join):
    if join == 'outer':
        return _outer_join
```

Note that `operator.or_` on a pandas.Index is the same thing as `pandas.Index.union`, so this could be equivalently written (maybe more clearly) as:
```python
def _outer_join(indexes):
    index = indexes[0]
    for other in indexes[1:]
        index = index.union(other)
    # your logic for reverse sorting the result
    return index
```

Huite:
I'm definitely not convinced it's a great idea either; a pull request is hopefully the best way to get some discussion going! Putting in the `_get_joiner` is definitely the sensible choice, and the last suggestion is definitely more clear to me.

I've had a bit more thought and maybe it's useful to consider what xarray's philosophy about these things is. I think flexibility is a primary concern and generally, things \*just work\*. The reason I'm running into issues here is because I'm writing some code which operates on DataArrays, isn't nearly as robust or flexible, and **doesn't** just work.

The answer in this case is might be that I should be using xarray's powerful alignment machinery to do the work for me, rather than to assume/demand certain features of the data. Of course, that requires some digging into how xarray does alignment. But I'd end up with a more flexible tool in the end.

Perhaps that should be the general rule: if you're extending xarray, like xarray, *don't* rely too much about your coordinates staying the way they are. Maybe such a description could belong in the [xarray Internals page](http://xarray.pydata.org/en/stable/internals.html) to make it a bit more transparent what's going on during merges.

dcherian:
From a user perspective, I think it's plain weird to merge a bunch of datasets with montonically decreasing coordinates and land up with a new dataset that has monotonically increasing coordinates.

shoyer:
Yes, I do think this deviation from pandas would make sense -- though it would be even better if we could convince pandas to make the same change!

We already have low-level tools for controlling how alignment works (construct the index yourself and use `.reindex`), but we should make the default behavior as user friendly as possible.

PR #2985, opened 2019-05-23
Default initialize Variable._attrs to OrderedDict

yifanmai:
When Variable is initialized with the default parameters, `_attr` is set to `None` by default. If `_attr` is none and the `attr` property is accessed, `_attr` is mutated to `OrderedDict()`. It is not good practice for a getter to be able to cause a mutation. This change initializes `_attr` to `OrderedDict()` by default instead.

This is an internal fix, so tests and documentation are omitted.

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -263,7 +263,7 @@ class Variable(common.AbstractArray, arithmetic.SupportsArithmetic,
         """
         self._data = as_compatible_data(data, fastpath=fastpath)
         self._dims = self._parse_dimensions(dims)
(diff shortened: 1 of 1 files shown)


shoyer:
I would consider this lazy initialization, not mutation. This is a pretty common pattern in Python, e.g., it's used all over the place in pandas (note use of the `@cache_readonly` decorator). Given that lazily initialization yields measurable code performance here, I am reluctant to remove it.

yifanmai:
Sounds reasonable to me. Closing the PR.

Issue #1053, opened 2016-10-20
Support __matmul__ operator (@)


chris-b1:
xref https://github.com/pandas-dev/pandas/issues/10259

Presumably deferring to the semantics of `np.matmul` - not sure if that API is stable yet?

shoyer:
For xarray, probably the right choice is for `@` to be an alias for `.dot()`:
http://xarray.pydata.org/en/stable/generated/xarray.DataArray.dot.html

The broadcasting semantics of `np.matmul` don't quite make sense because it broadcasts based on axis position, not name.

dhimmel:
Would love support for [PEP 465](http://legacy.python.org/dev/peps/pep-0465/) `@` notation.

Recently, `@`  [came in handy](https://github.com/greenelab/hetmech/issues/13#issuecomment-290780424) when multiplying `numpy.ndarray` with `scipy.sparse` matrices. We're [considering](https://github.com/greenelab/hetmech/issues/12) `xarray` for our project and compatibility with this unified operator would be a real plus!

dhimmel:
More specifically, I'd like to be able to do matrix multiplication between numpy ndarrays / matrices, scipy sparse matrices, and xarray DataArrays. `@` seems like the most natural operator to enable this cross-package compatibility.

shoyer:
> More specifically, I'd like to be able to do matrix multiplication between numpy ndarrays / matrices, scipy sparse matrices, and xarray DataArrays.

I'm intrigued, but how would this work? `data_array + numpy_array` yields a result with well-defined labels as long as `numpy_array` broadcasts against `data_array.data`, but `data_array @ numpy_array` does not if `numpy_array` has 2 or more dimensions.

I guess we could prohibit `@` with non-vector other arguments, but I still am concerned that the suggested meaning of `@` per [PEP 465](http://legacy.python.org/dev/peps/pep-0465/) and numpy depends on the order of array dimensions. Basically, the last dimension of the left-hand-side argument should be matched against the second-to-last (or last, for 1D) dimension of the right-hand-side for the tensor contraction. In xarray terms, we could match the last dimension of the left-hand-side with any matching dimensions (by name) of the right-hand-side, but it's still messily inconsistent with other xarray operations, which are generally agnostic to to dimension order.

It also gets messy on `Dataset` objects, because the order of dimensions now becomes a bit more ambiguous: there's the order of dimensions on the `Dataset` itself, and the order on each `DataArray` in the dataset.

[...]

PR #2987, opened 2019-05-24
Implement @ operator for DataArray

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/1053
 - [x] Tests added

Is it really this easy? What am I missing?

### DIFF
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -45,6 +45,12 @@ Use :py:func:`~xarray.where` to conditionally switch between values:
 
     xr.where(arr > 0, 'positive', 'negative')
 
+Use `@` to perform matrix multiplication:
+
+.. ipython:: python
+
+    arr @ arr
+
 Data arrays also implement many :py:class:`numpy.ndarray` methods:
 
 .. ipython:: python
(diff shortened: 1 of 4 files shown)


max-sixty:
Any thoughts, team?

shoyer:
This looks good to me!

In general, it's a good idea to also define reflexive methods, e.g., `__rmatmul__` in this case. But I'm not sure we actually need that, given that `dot` only supports operations with xarray.DataArray at present.

I guess this could be useful for people writing subclasses, or other objects that override arithmetic with xarray objects. So you might as well add something like:
```
def __rmatmul__(self, other):
    return xarray.dot(other, dot)
```

dcherian:
Do we need a line here? https://xarray.pydata.org/en/stable/computation.html#basic-array-math

shoyer:
It might be worth using @ instead of dot() in the docs in a few places, but
I don't think it's essential to mention it specifically.

max-sixty:
Great - I added `rmatmul` and a v short example in the basic-array-math section. LMK if good to merge

max-sixty:
LMK if good to merge! 

(coming from the person with a terrible record of finishing PRs... 😉 )

PR #2988, opened 2019-05-24
Remove deprecated pytest.config usages

max-sixty:
- [x] Tests added

I need to confirm whether this works with an installed version of xarray, because of https://github.com/pytest-dev/pytest/issues/1596. 
If pytest doesn't pick up this config, it will run all tests, because the default is to not skip.

I know I've generally been the point person for pytest - let me know if anyone has an immediate solution for this though

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,7 @@
 """Configuration for pytest."""
 
+import pytest
+
 
 def pytest_addoption(parser):
     """Add command-line flags for pytest."""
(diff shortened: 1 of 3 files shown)


max-sixty:
After some testing, this doesn't change behavior. As before, it _mostly_ works: 

- Running `pytest --pyargs xarray` to test against an installed version doesn't allow passing command line options such as `--run-flaky` (either with the existing or proposed version)
- With the proposed version you can pass `-m flaky`, because these are now marks

shoyer:
This looks fine to me. I don't think it's important to retain the same command for running flakey tests.

max-sixty:
OK great; I'll merge

We don't test-the-tests so anyone who encounters any problems: post here and I'll prioritize

PR #2991, opened 2019-05-25
ENH: str accessor

### LINKED ISSUE #2983: string accessor
Hello,

I have written a small wrapper around pandas internal methods
https://gist.github.com/0x0L/ef78c80a42892c0f832c91357914a5a4

Missing methods are those involving list or list of arrays (join, split, partition, ...)

Let me know if there's enough interest to turn this into a full commit

0x0L:
Hello,

Some of the pandas str functionalities. Instead of wrapping pandas internal as in #2983 I copy/pasted the code since it's simple and tiny.

Currently it's a bit more restrictive than pandas since it expects all elements to be string like.

 - [x] Closes #2983 
 - [x] Tests added


xarray/core/accessors.py
```diff
@@ -256,3 +260,888 @@ def round(self, freq):
...
+
+        Returns
+        -------
+        sliced strings : same type as values
+        '''
+        s = slice(start, stop, step)
+        f = lambda x: x[s]
+        return self._apply(f)
+
+    def slice_replace(self, start=None, stop=None, repl=None):
```
shoyer: Should this just be `repl=''`?

xarray/core/accessors.py
```diff
@@ -256,3 +260,888 @@ def round(self, freq):
...
+        Dimensions without coordinates: dim_0
+
+    """
+
+    def __init__(self, xarray_obj):
+        self._obj = xarray_obj
+
+    def _apply(self, f):
+        # TODO handling of na values ?
+        return apply_ufunc(np.vectorize(f), self._obj)
```
shoyer: Can you add `dask='parallelized` to add dask support? I'm pretty sure that should "just work" with these functions.

0x0L:
@shoyer 
The api reference does not generate doc for the accessor methods. It's also missing for `.dt`
Do you know how to do this ?

shoyer:
You will need to reference the underlying class, e.g., `xarray.core.str_accessor.StringAccessor`. We do something similar for GroupBy objects: http://xarray.pydata.org/en/stable/api.html#groupby-objects

shoyer:
@0x0L did you copy any of this code or tests from pandas? If so, that's totally fine, but we should include the original pandas copyright notice. See what we did in `cftimeindex.py` for an example: https://github.com/pydata/xarray/blob/master/xarray/coding/cftimeindex.py

0x0L:
@shoyer it should be all good now

shoyer:
thanks @0x0L !

PR #3001, opened 2019-06-05
BUG: fix safe_cast_to_index

### LINKED ISSUE #3000: Slowness when cftime is installed
With `cftime` installed

```python
import numpy as np
import pandas as pd
import xarray as xr

da = xr.DataArray(np.random.randn(5000, 500))
df = da.to_pandas()

# with pandas
%time df_stacked = df.stack()
# Wall time: 48.3 ms
%time df_unstacked = df_stacked.unstack()
# Wall time: 368 ms

# with xarray
%time da_stacked = da.stack(stacked_dim=('dim_0', 'dim_1'))
# Wall time: 1.03 s
%time da_unstacked = da_stacked.unstack('stacked_dim')
# Wall time: 78.2 ms
```

prun points to `CFTimeIndex(index)` in `_maybe_cast_to_cftimeindex`

The behaviour is also incorrect for empty indexes:
```python
da[:0].stack(dim=['dim_0', 'dim_1']).dim.to_index()
# CFTimeIndex([], dtype='object', name='dim')
```

0x0L:
- [x] Closes #3000
 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -61,8 +61,10 @@ Bug fixes
   By `Deepak Cherian <https://github.com/dcherian`_.
 - A deep copy deep-copies the coords (:issue:`1463`)
   By `Martin Pletcher <https://github.com/pletchm>`_.
-- Removed usages of `pytest.config`, which is deprecated (:issue:`2988`:)
+- Removed usages of `pytest.config`, which is deprecated (:issue:`2988`)
   By `Maximilian Roos <https://github.com/max-sixty>`_.
+- Fixed performance issues with cftime installed (:issue:`3000`)
+  By `0x0L <https://github.com/0x0L>`_.
 
 .. _whats-new.0.12.1:
 
--- a/xarray/core/utils.py
+++ b/xarray/core/utils.py
@@ -62,7 +62,7 @@ def alias(obj: Callable[..., T], old_name: str) -> Callable[..., T]:
 def _maybe_cast_to_cftimeindex(index: pd.Index) -> pd.Index:
     from ..coding.cftimeindex import CFTimeIndex
 
-    if index.dtype == 'O':
+    if len(index) > 0 and index.dtype == 'O':
         try:
             return CFTimeIndex(index)
         except (ImportError, TypeError):
--- a/xarray/coding/cftimeindex.py
+++ b/xarray/coding/cftimeindex.py
@@ -184,7 +184,7 @@ def get_date_type(self):
 def assert_all_valid_date_type(data):
     import cftime
 
-    if data.size:
+    if len(data) > 0:
         sample = data[0]
         date_type = type(sample)
         if not isinstance(sample, cftime.datetime):
@@ -229,12 +229,12 @@ class CFTimeIndex(pd.Index):
     date_type = property(get_date_type)
 
     def __new__(cls, data, name=None):
+        assert_all_valid_date_type(data)
         if name is None and hasattr(data, 'name'):
             name = data.name
 
         result = object.__new__(cls)
         result._data = np.array(data, dtype='O')
-        assert_all_valid_date_type(result._data)
         result.name = name
         return result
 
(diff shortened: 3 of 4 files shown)


doc/whats-new.rst
```diff
@@ -63,6 +63,9 @@ Bug fixes
   By `Martin Pletcher <https://github.com/pletchm>`_.
 - Removed usages of `pytest.config`, which is deprecated (:issue:`2988`:)
   By `Maximilian Roos <https://github.com/max-sixty>`_.
+- Fixed a bug in `utils.safe_cast_to_index` and performance issues when
+  `cftime` is installed (:issue:`3000`:)
```
spencerkclark: nit: I think it might be helpful to be a little more specific here (also the trailing colon for the issue link is not needed -- looks like we may have missed that in the entry above)
```suggestion
  using :py:meth:`~xarray.DataArray.stack` when cftime is installed (:issue:`3000`)
```

xarray/core/utils.py
```diff
@@ -62,8 +62,11 @@ def wrapper(*args, **kwargs):
 def _maybe_cast_to_cftimeindex(index: pd.Index) -> pd.Index:
     from ..coding.cftimeindex import CFTimeIndex
 
-    if index.dtype == 'O':
+    if len(index) > 0 and index.dtype == 'O':
```
spencerkclark: It might make more sense to instead check that `index` is not an instance of a `MultiIndex`.  I think the fact that `index.dtype` returns `object` in your example in #3000 is what trips things up originally.
```suggestion
    if not isinstance(index, pd.MultiIndex) and index.dtype == 'O':
```

xarray/core/utils.py
```diff
@@ -62,8 +62,11 @@ def wrapper(*args, **kwargs):
...
         try:
+            import cftime
+            if not isinstance(index[0], cftime.datetime):
```
spencerkclark: We have a check similar to this that gets called in the `CFTimeIndex` constructor, so I think we may not need this check here:
https://github.com/pydata/xarray/blob/5343ccc4bef77020a8181c9da0fdee6bcae35b5f/xarray/coding/cftimeindex.py#L188-L193

doc/whats-new.rst
0x0L: Actually `safe_cast_to_index` is called in from multiple files (alignment, groupby, indexing, utils), I'm not sure only `stack` was impacted

xarray/core/utils.py
0x0L: It looks like I was really tired... The name of the function was `assert_all_valid_date_type` so I somehow assume this would test all elements are valid not just the first :)

doc/whats-new.rst
spencerkclark: Good point -- maybe let's say something like "performance issues when using a MultiIndex with cftime installed."  I'm reasonably confident this is MultiIndex-centric.

xarray/core/utils.py
0x0L: For the same reasons as above I wanted to avoid defaulting to CFTimeIndex([]) for other yet unknown cases

0x0L:
@spencerkclark 
Are you ok with keeping the len > 0 test and then just moving up the assert at the top of the constructor of CFTimeIndex so that we raise before allocating the memory with `result._data = np.array(data, dtype='O')`

spencerkclark:
> Are you ok with keeping the len > 0 test and then just moving up the assert at the top of the constructor of CFTimeIndex so that we raise before allocating the memory with result._data = np.array(data, dtype='O')

Yes, sorry, it seems I'm a bit sleepy today as well.  This is probably the best approach 👍.

doc/whats-new.rst
spencerkclark: > I'm reasonably confident this is MultiIndex-centric.

Performance will also be impacted for other object type indexes (like an index of strings), so ignore this too :).  @0x0L your suggested approach, https://github.com/pydata/xarray/pull/3001#issuecomment-499598244, fixes this too.

Issue #2990, opened 2019-05-24
Some minor errors in repo / flake8


max-sixty:
Currently we use pycodestyle: https://github.com/pydata/xarray/blob/ccd0b047ea8ca89c68ab6cfa942557e676e7d402/.travis.yml#L63

I think we used to use flake8. I can't find / remember the reason we moved to pycodestyle. 

master has some non-trivial issues that flake would catch, including a test overwritting another and undefined variables:
```
flake8 xarray --ignore=I,W503,W504,F401,E265,E402

xarray/core/options.py:62:8: F632 use ==/!= to compare str, bytes, and int literals
xarray/core/dataset.py:4148:69: F821 undefined name 'dim'
xarray/backends/netCDF4_.py:177:12: F632 use ==/!= to compare str, bytes, and int literals
xarray/tests/test_dataarray.py:1264:9: F841 local variable 'foo' is assigned to but never used
xarray/tests/test_dataarray.py:1270:18: F821 undefined name 'x'
xarray/tests/test_dataarray.py:1301:5: F811 redefinition of unused 'test_reindex_fill_value' from line 1262
xarray/tests/test_dataarray.py:1647:16: F632 use ==/!= to compare str, bytes, and int literals
xarray/tests/test_dataarray.py:1648:16: F632 use ==/!= to compare str, bytes, and int literals
xarray/tests/test_dataset.py:4759:8: F632 use ==/!= to compare str, bytes, and int literals
xarray/tests/test_dataset.py:4761:10: F632 use ==/!= to compare str, bytes, and int literals
xarray/tests/test_distributed.py:62:9: F811 redefinition of unused 'loop' from line 12
xarray/tests/test_distributed.py:92:9: F811 redefinition of unused 'loop' from line 12
xarray/tests/test_distributed.py:117:49: F811 redefinition of unused 'loop' from line 12
xarray/tests/test_distributed.py:141:53: F811 redefinition of unused 'loop' from line 12
xarray/tests/test_distributed.py:152:51: F811 redefinition of unused 'loop' from line 12
```

Happy to fix these in a PR. For ensuring these don't crop up again, any objection to flake8?

shoyer:
I would be very happy to switch back to flake8, but we should also switch `pep8speaks` to use flake8, too, which it now supports!

max-sixty:
Closed by #3010

Issue #2833, opened 2019-03-20
Integrate has undefined name 'dim'


max-sixty:
https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L4085

Should that be called `coord` or `dim`? Currently there's a variable that's undefined:

```python
            raise ValueError('Coordinate {} does not exist.'.format(dim))
```
I would have made a quick fix but not sure the correct name

shoyer:
I think it should be `coord`. Also we should add a unit test :)

max-sixty:
Closed by #3010

PR #3010, opened 2019-06-10
Use flake8 rather than pycodestyle

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/2990

I also ran `isort`, which created more changes that I would have hoped. Lmk if that's OK. 
We could also enable `isort` by default. It's a balance: it reduces the chance of merge conflicts and adds automatic consistency, but at the cost of making first-time contributions a bit more difficult.

### DIFF
--- a/ci/requirements-py37.yml
+++ b/ci/requirements-py37.yml
@@ -15,7 +15,7 @@ dependencies:
   - pytest-cov
   - pytest-env
   - coveralls
-  - pycodestyle
+  - flake8
   - numpy>=1.12
   - pandas>=0.19
   - scipy
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -5,6 +5,7 @@
 
 scanner:
     diff_only: False  # If True, errors caused by only the patch are shown
+    linter: flake8
 
 pycodestyle:
     max-line-length: 79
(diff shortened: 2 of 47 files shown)


max-sixty:
Also fixes https://github.com/pydata/xarray/issues/2833

max-sixty:
Will wait for a couple of days to see if there's a response on the `pep8speaks` config issue; otherwise will revert to the existing config.

Tests pass apart from `dask-dev` which breaks on master

max-sixty:
This is ready to merge after a review. `pep8speaks` [checks the config in `master`](https://github.com/OrkoHunter/pep8speaks/issues/130#issuecomment-500784182), and so this needs to be merged in order for the config to be updated

ci/requirements-py37.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - pytest-cov
   - pytest-env
   - coveralls
-  - pycodestyle
+  - flake8
```
shoyer: Can you also replace pycodestyle with flake8 in requirements-py36.yml? Our contributing guide is still pointing users to that for setup.

shoyer:
Thanks!

PR #3011, opened 2019-06-10
Pytest capture uses match, not message

max-sixty:

`message` was always going being ignored, and with the newer pytest version raises a warning that an unknown kwarg is supplied


max-sixty:
Tests pass apart from py36-dask-dev, which fail on master: https://travis-ci.org/pydata/xarray/builds/543887984

dcherian:
LGTM and I didn't see any pytest warnings in the test logs :)

PR #3013, opened 2019-06-10
Add `scales` attributes to Dataset created in open_rasterio (#3012)

### LINKED ISSUE #3012: rasterio.open exposes the GeoTIFF scales attribute, but xarray.open_rasterio not

```python
>> import xarray as xr
>> from pprint import pprint
>> arr = xr.open_rasterio('example.tif')
>> pprint(arr)
<xarray.DataArray (band: 2, y: 7201, x: 5896)>
[84914192 values with dtype=uint16]
Coordinates:
  * band     (band) int64 1 2
  * y        (y) float64 -17.18 -17.18 -17.18 -17.19 ... -23.58 -23.58 -23.58
  * x        (x) float64 -52.23 -52.23 -52.23 -52.23 ... -46.99 -46.99 -46.99
Attributes:
    transform:   (0.0008888888888890009, 0.0, -52.22933333332605, 0.0, -0.000...
    crs:         +init=epsg:4326
    res:         (0.0008888888888890009, 0.0008888888888890009)
    is_tiled:    1
    nodatavals:  (65535.0, 65535.0)
```

Would be nice if it had the `scales` attribute.

#### Problem description

rasterio.open exposes the GeoTIFF scales attribute, but xarray not.

```python
>> import rasterio
>> rarr = rasterio.open('example.tif')
>> print(rarr.scales)
(0.001, 0.001)
```

#### Expected Output

```python
>> import xarray as xr
>> from pprint import pprint
>> arr = xr.open_rasterio('example.tif')
>> pprint(arr)
<xarray.DataArray (band: 2, y: 7201, x: 5896)>
[84914192 values with dtype=uint16]
Coordinates:
  * band     (band) int64 1 2
  * y        (y) float64 -17.18 -17.18 -17.18 -17.19 ... -23.58 -23.58 -23.58
  * x        (x) float64 -52.23 -52.23 -52.23 -52.23 ... -46.99 -46.99 -46.99
Attributes:
    transform:   (0.0008888888888890009, 0.0, -52.22933333332605, 0.0, -0.000...
    crs:         +init=epsg:4326
    res:         (0.0008888888888890009, 0.0008888888888890009)
    is_tiled:    1
    nodatavals:  (65535.0, 65535.0)
    scales: (0.001, 0.001)
```

#### xarray version

```python
xr.__version__ == '0.12.1'
```

ecarrara:
- [X] Closes #3012
 - [X] Tests ~added~ edited


dcherian:
Does rasterio define a set of valid attributes or can you assign arbitrary attributes? If so, should we read and assign all attributes as we do for netCDF files?

ecarrara:
> Does rasterio define a set of valid attributes or can you assign arbitrary attributes? If so, should we read and assign all attributes as we do for netCDF files?

@dcherian the attributes are pre-defined. Looking at the source code[1] of rasterio I found this list:

```property nodatavals             # Nodata values for each band
property nodata                 # The dataset's single nodata value
property mask_flag_enums        # Sets of flags describing the sources of band masks.
property crs                    # The dataset's coordinate reference system
property descriptions           # Descriptions for each dataset band
property transform              # The dataset's georeferencing transformation matrix
property offsets                # Raster offset for each dataset band
property scales                 # Raster scale for each dataset band
property units                  # A list of str: one units string for each dataset band
property bounds                 # Returns the lower left and upper right bounds of the dataset
property res                    # Returns the (width, height) of pixels in the units of its
```

I belive that the attributes not handled yet by xarray are: **scales, offsets and units**.

[1] - https://github.com/mapbox/rasterio/blob/1.0.24/rasterio/_base.pyx

dcherian:
I think we should read and expose all those attributes. 

@ecarrara are you up for expanding the scope of your PR? Also, I see you're a "first-time contributor". Welcome! and thanks for contributing!

ping @fmaussion for feedback.

ecarrara:
thank you @dcherian!

I added `offsets`, `units`, `descriptions` attributes. The other attributes (`res`, `nodata`, `crs`,  `transform`, `bounds` and `res`) were already available.

`mask_flag_enums` is a list of tuples, I have tried to add this attribute too but the `TestRasterio::test_serialization` fails with the message:

```python
>   ???
E   ValueError: multi-dimensional array attributes not supported

netCDF4/_netCDF4.pyx:1529: ValueError
```

@fmaussion any feedback is welcome!

xarray/tests/test_backends.py
```diff
@@ -3108,6 +3108,8 @@ def test_utm(self):
         with create_tmp_geotiff() as (tmp_file, expected):
             with xr.open_rasterio(tmp_file) as rioda:
                 assert_allclose(rioda, expected)
+                assert rioda.attrs['scales'] == (1.0, 1.0, 1.0)
+                assert rioda.attrs['offsets'] == (0.0, 0.0, 0.0)
```
dcherian: can we test for `description` and `units` too?

fmaussion:
Thanks @ecarrara !

>  but the TestRasterio::test_serialization fails

This is not the case anymore, right? This test is here to ensure that the dataarrays we are creating with `open_rasterio` are compatible with the NetCDF model. I think it is good to try to keep it that way.

dcherian:
Yeah I think @ecarrara found a workaround. 

@ecarrara You should also add a note to whats-new and give yourself credit.

xarray/backends/rasterio_.py
```diff
@@ -313,6 +313,11 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
     if hasattr(riods, 'units') and any(riods.units):
         # A list of units string for each dataset band
         attrs['units'] = riods.units
+    if hasattr(riods, 'mask_flag_enums'):
+        # Sets of flags describing the sources of band masks
+        attrs['mask_flags'] = [
```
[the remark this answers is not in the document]
dcherian: Let's see what @fmaussion thinks about the string approach. I don't use rasterio so I don't know what that community would find useful.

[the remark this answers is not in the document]
fmaussion: I don't use these attributes either, unfortunately. The string trick doesn't look very good though...

What is the expected dimensions of this attribute? Can it be anything, or is it bound to the dimension of the bands? Because this looks like one of the limitation of the DataArray model for rasterio objects (another possibility would have been a Dataset, it was discussed in earlier stages but then discarded).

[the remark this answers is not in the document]
ecarrara: > The string trick doesn't look very good though...
:+1: 

I prefer to use the int bit mask as defined in GDAL documentation (rasterio expose these ints too), but it can be a little puzzling for people who dont known gdal internals.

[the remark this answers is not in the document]
dcherian: Thanks. Let's just not deal with `mask_flag_enums` unless someone really needs it. None of these approaches seem that great unfortunately.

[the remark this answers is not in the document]
fmaussion: Instead of incrementally adding the attributes, could there be a way to expose the rasterio object *from* the DataArray object? I guess there is no clean way to do so, but it would allow everyone to access whatever they want...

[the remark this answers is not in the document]
ecarrara: > Thanks. Let's just not deal with `mask_flag_enums` unless someone really needs it. None of these approaches seem that great unfortunately.

@dcherian I agree with you. Today `mask_flag_enums` is useless for me, and probably someone that really uses it can find a better way to expose the attribute.

dcherian:
OK let's remove the mask_flag_enums bits then. I noticed that there's a binary .idx file that was added to. Is that needed?

xarray/tests/test_backends.py
```diff
@@ -3223,6 +3231,10 @@ def test_utm(self):
         with create_tmp_geotiff() as (tmp_file, expected):
             with xr.open_rasterio(tmp_file) as rioda:
                 assert_allclose(rioda, expected)
+                assert rioda.attrs['scales'] == (1.0, 1.0, 1.0)
```
dcherian: I don't see where these attributes are being set. How is this working?

ecarrara: I added scales attribute in the first commit of this PR.

https://github.com/pydata/xarray/pull/3013/files#diff-6364b203943c799516f9ecba19e1b119R306

dcherian: No I mean for the tests. Where is (1.0, 1.0, 1.0) being set for scales? I see descriptions and units being set.

ecarrara: Geotiff bands have a default value 1.0 for scale.

dcherian: Ah ok. Can you either add a comment saying that or set custom values for the test so that people looking over the code in the future are not confused? (I think I prefer the latter)

Apologies for the delay. With that, I think this should be good to merge.

doc/whats-new.rst
```diff
@@ -96,6 +96,14 @@ Enhancements to existing functionality
 
 - Add ``keepdims`` argument for reduce operations (:issue:`2170`)
   By `Scott Wales <https://github.com/ScottWales>`_.
+- Add ``scales``, ``offsets``, ``units`` and ``descriptions``
+  attributes to :py:class:`~xarray.DataArray` returned by
+  :py:func:`~xarray.open_rasterio`. (:issue:`3013`)
+  By `Erle Carrara <https://github.com/ecarrara>`_.
+- netCDF chunksizes are now only dropped when original_shape is different,
```
dcherian: Bad merge. This needs to move up to 0.12.3 and include jsut your contribution

shoyer:
I fixed up the merge. Let's just make sure tests pass...

dcherian:
Thanks @ecarrara !

PR #3016, opened 2019-06-11
Pandas labels deprecation

max-sixty:
Should I still add a `whatsnew`? It's very small and very internal


shoyer:
No need for what's new, but we do still want this to work on older versions of pandas.

max-sixty:
Great, and found a few examples in docs too

spencerkclark:
Thanks for catching this @max-sixty.

PR #3019, opened 2019-06-12
Update issue templates

max-sixty:
This: 
- Updates to the [newer GitHub format](https://help.github.com/en/articles/about-issue-and-pull-request-templates)
- Strengthens the language around MCVE. It doesn't say people *have* to, but makes it less of a suggestion
- Only contains 'Bug Report'. Should we have others? I don't immediately see how they'd be different

(I realize because I did this from the GH website, it created from this repo rather than my own fork. That's a mistake.)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+#### MCVE Code Sample
+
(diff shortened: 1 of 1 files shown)


max-sixty:
Some ideation re MCVE:

An app that would be generally useful: an easy way for people to post an runnable example. e.g. Rust has [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018). That would force people to have an example that runs; no risk of copy-paste errors.

Has anyone tried with a Colab link? 

Stack Overflow would have the scale more than xarray-on-GH, so maybe we're not the ones to forge new ground...

shoyer:
> Has anyone tried with a Colab link?

I link this idea, we'd just have to ask users to update xarray first, since Colab updates xarray on their own schedule.

But let's go ahead and merge this for now anyways.

PR #3024, opened 2019-06-15
Check types in travis

max-sixty:
As part of https://github.com/pydata/xarray/pull/2929#issuecomment-502302695, I looked at enabling travis to check types. 

What do people think? As a more cautious option I could add to `allowed_failures` for the moment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,27 +10,27 @@ branches:
 matrix:
   fast_finish: true
   include:
-    - env: CONDA_ENV=py35-min
-    - env: CONDA_ENV=py36
-    - env: CONDA_ENV=py37
-    - env:
-        - CONDA_ENV=py36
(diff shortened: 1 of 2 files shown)


shoyer:
My original reason for not enabling this was to impose the burden of hunting down type errors on Travis-CI. But maybe we can figure out how to run `mypy` with pep8speaks?

max-sixty:
pep8speaks doesn't run mypy yet, unfortunately.

Shall I add to `allow_failures` and we can see whether it's failing too often for normal PRs?

shoyer:
Sure, sounds great to add this as an allowed failure for now

max-sixty:
Great, moved to `allow_failures` - could someone give this a final check (or feel free to merge directly)

shoyer:
thanks!

PR #3027, opened 2019-06-17
Ensure explicitly indexed arrays are preserved

shoyer:
Fixes https://github.com/pydata/xarray/issues/3009

Previously, indexing an ImplicitToExplicitIndexingAdapter object could directly return an ExplicitlyIndexed object, which could not be indexed normally, e.g., `x[index]` could result in an object that could not be indexed properly. This resulted in broken behavior with dask's new `_meta` attribute.

I'm pretty sure this fix is appropriate, but it does introduce two failing tests with xarray on dask master. In particular, there are now errors raised inside two tests from dask's `blockwise_meta` helper function:
```
>       return meta.astype(dtype)
E       AttributeError: 'ImplicitToExplicitIndexingAdapter' object has no attribute 'astype'
```

cc @mrocklin  @pentschev

 - [x] Tests added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
   - env:
     - CONDA_ENV=py36
     - EXTRA_FLAGS="--run-flaky --run-network-tests"
+  - env: CONDA_ENV=py36-dask-dev
   - env: CONDA_ENV=py36-pandas-dev
   - env: CONDA_ENV=py36-rasterio
   - env: CONDA_ENV=py36-zarr-dev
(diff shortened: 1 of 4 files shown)


mrocklin:
I think that relaxing the astype constrain seems quite reasonable.  I'll clean this up on the Dask side.

mrocklin:
Does the green check mark here mean that we're all good @shoyer ?

shoyer:
@mrocklin Yes, this appears to work :)

PR #3028, opened 2019-06-17
Add "errors" keyword argument to drop() and drop_dims() (#2994)

### LINKED ISSUE #2994: xr.Dataset.drop
Currently, `drop` throws an error if one of the labels doesn't exist. It would be nice to have a parameter in the drop method for optionally ignoring errors like in the pandas.DataFrame.
From the pandas [documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.drop.html):

> errors : {‘ignore’, ‘raise’}, default ‘raise’
>     If ‘ignore’, suppress error and only existing labels are dropped.

andrew-c-ross:
- [x] Closes #2994 
 - [x] Tests added

This addresses #2994 by adding an "errors" keyword argument to `Dataset.drop()`, `Dataset.drop_dims()`, and `DataArray.drop()`. 

I stuck with pandas' convention of using either `errors='raise'`, now the default that maintains previous behavior by raising an error if any passed label is not found in the dataset/array, or `errors='ignore'` in which case any missing labels are silently ignored. 

This seems like a pretty straightforward change; mainly it is just skipping checks for missing labels when `errors == 'ignore'` and passing the errors keyword over to the pandas method when using `index.drop()`. Hopefully there are no subtleties that I've missed. 

I added documentation to the appropriate methods, although I have been struggling to build the docs locally and am unsure if they look right.

Also this is my first attempt to contribute to any project, so suggestions and feedback are welcome.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1461,7 +1461,7 @@ class DataArray(AbstractArray, DataWithCoords):
     def T(self) -> 'DataArray':
         return self.transpose()
 
-    def drop(self, labels, dim=None):
+    def drop(self, labels, dim=None, errors='raise'):
         """Drop coordinates or index labels from this DataArray.
 
         Parameters
@@ -1471,14 +1471,18 @@ class DataArray(AbstractArray, DataWithCoords):
         dim : str, optional
             Dimension along which to drop index labels. By default (if
             ``dim is None``), drops coordinates rather than index labels.
-
+        errors: {'raise', 'ignore'}, optional
+            If 'raise' (default), raises a ValueError error if
+            any of the variable or index labels passed are not
+            in the dataset. If ``ignore'', any given labels that are in the
+            dataset are dropped and no error is raised.
         Returns
         -------
         dropped : DataArray
         """
         if utils.is_scalar(labels):
             labels = [labels]
-        ds = self._to_temp_dataset().drop(labels, dim)
+        ds = self._to_temp_dataset().drop(labels, dim, errors=errors)
         return self._from_temp_dataset(ds)
 
     def dropna(self, dim, how='any', thresh=None):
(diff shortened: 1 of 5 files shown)


shoyer:
This looks great, thank you!

One design question: should we stick with `errors='ignore'` like pandas, or would it be better to use a more self-descriptive argument name like `missing='ignore'`/`missing='raise'`?

max-sixty:
> One design question: should we stick with errors='ignore' like pandas, or would it be better to use a more self-descriptive argument name like missing='ignore'/missing='raise'?

I'm ambivalent. I prefer the `missing=`, but I'm more 'in it' than the average user, who would probably prefer consistency. If people / @shoyer vote that `missing=` is materially better, let's change

andrew-c-ross:
I have to say as someone who is probably an average user, inconsistencies between related projects, like xarray/pandas or matplotlib/seaborn, drive me nuts. But I also agree that `missing=` is more descriptive, and if we were starting from scratch I would totally support that. So I will defer to the maintainers here.

xarray/core/dataarray.py
```diff
@@ -1461,7 +1461,7 @@ def transpose(self, *dims, transpose_coords=None) -> 'DataArray':
     def T(self) -> 'DataArray':
         return self.transpose()
 
-    def drop(self, labels, dim=None):
+    def drop(self, labels, dim=None, errors='raise'):
```
shoyer: Let's make this new argument require using a keyword argument: 
```suggestion
    def drop(self, labels, dim=None, *, errors='raise'):
```

shoyer: Yes, let's make that keyword argument only, too.

shoyer:
OK, I'm going to go ahead and merge after tests pass!

andrew-c-ross:
Sounds great. Thanks for making this a smooth and well documented process for new contributors

PR #3029, opened 2019-06-17
Fix pandas-dev tests 

max-sixty:
Currently pandas-dev tests get 'stuck' on the conda install. The last instruction to run is the standard install:

```
$ if [[ "$CONDA_ENV" == "docs" ]]; then conda env create -n test_env --file doc/environment.yml; elif [[ "$CONDA_ENV" == "lint" ]]; then conda env create -n test_env --file ci/requirements-py37.yml; else conda env create -n test_env --file ci/requirements-$CONDA_ENV.yml; fi
```

And after installing the libraries, [it prints this and then stops](https://travis-ci.org/max-sixty/xarray/jobs/546491330):

```
Preparing transaction: - - done
Verifying transaction: | / \ | / - \ | / / done
Executing transaction: \ | / - \ | / - \ | / - \ | / - \ | / - \ | / / - \ | / - \ done
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I'm not that familiar with conda. Anyone have any ideas as to why this would fail while the other builds would succeed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
   - env:
     - CONDA_ENV=py36
     - EXTRA_FLAGS="--run-flaky --run-network-tests"
-  - env: CONDA_ENV=py36-pandas-dev
+  # - env: CONDA_ENV=py36-pandas-dev
   - env: CONDA_ENV=py36-zarr-dev
 
 before_install:
--- a/ci/requirements-py36-pandas-dev.yml
+++ b/ci/requirements-py36-pandas-dev.yml
@@ -56,5 +56,6 @@ dependencies:
   - pydap
   - lxml
   - pip:
+    - cython
     - git+https://github.com/pydata/pandas.git
     - mypy==0.660
--- a/ci/requirements-py36-pandas-dev.yml
+++ b/ci/requirements-py36-pandas-dev.yml
@@ -19,5 +19,6 @@ dependencies:
   - numpy>=1.12
   - scipy
   - toolz
+  - pip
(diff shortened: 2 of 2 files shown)


max-sixty:
We're not alone: https://github.com/conda/conda/issues/6986

max-sixty:
This is how dask installs the 'nightlies' (I'm guessing these are similar to the master branch): https://github.com/dask/dask/blob/master/continuous_integration/travis/install.sh

...maybe we need to do this; ideally we could make a smaller tweak...

max-sixty:
I had this passing a couple of times (e.g. https://travis-ci.org/max-sixty/xarray/builds/546932430), I think this those didn't fix some fundamental problem rather
- the end of the conda install stalls for around ten minutes
- travis will error the build if it's silent for ten minutes
- those builds were _slightly_ less than ten minutes

I'll pause here for the moment; let me know if anyone has any bright ideas

shoyer:
It probably makes sense to use a separate `pip install` afterwards to install pandas. That lets us leverage `conda` for all the binary dependencies, and on the whole should go much faster.

max-sixty:
> It probably makes sense to use a separate pip install afterwards to install pandas

To confirm, does this mean running a separate `pip install git+https://github.com/pandas-dev/pandas.git` after the install? We already have a separate pip section in the conda environment (I generally don't use conda so am not well-versed in this)

shoyer:
This is working on Azure, which has a longer build timeout.

PR #3033, opened 2019-06-19
ENH: keepdims=True for xarray reductions

### LINKED ISSUE #2170: keepdims=True for xarray reductions
For operations where arrays are aggregated but then combined, the `keepdims=True` option for NumPy aggregations is convenient.

We should consider supporting this in xarray as well. Aggregating a DataArray/Dataset with `keepdims=True` (or maybe `keep_dims=True`) would remove all original coordinates along aggregated dimensions and return a result with a dimension of size 1 without any coordinates, e.g.,
```
>>> array = xr.DataArray([1, 2, 3], dims='x', coords={'x': ['a', 'b', 'c']})
>>> array.mean(keepdims=True)
<xarray.DataArray (x: 1)>
array([2.])
```

In case, `array.mean(keepdims=True()` is equivalent to `array.mean().expand_dims('x')` but in general this equivalent does not hold, because the location of the original dimension is lost.

Implementation-wise, we have two options:
1. Pass on `keepdims=True` to NumPy functions like `numpy.mean()`, or
2. Implement `keepdims=True` ourselves, in `Variable.reduce()`.

I think I like option 2 a little better, because it places fewer requirements on aggregation functions. For example, functions like `bottleneck.nanmean()` don't accept a `keepdims` argument.

ScottWales:
Add new option `keepdims` to xarray reduce operations, following the behaviour of Numpy.

`keepdims` may be passed to reductions on either Datasets or DataArrays, and will result in the reduced dimensions being still present in the output with size 1.

Coordinates that depend on the reduced dimensions will be removed from the Dataset/DataArray

The name `keepdims` is used here to be consistent with Numpy, `keep_dims` was an alternative name proposed in #2170.

The functionality has only been added to `Variable.reduce()`, `Dataarray.reduce()` and `DataSet.reduce()` to start off with, it is not implemented for `GroupBy`, `Rolling` or `Resample`.

 - [x] Closes #2170
 - [x] Tests added

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -258,8 +258,14 @@ class DataArray(AbstractArray, DataWithCoords):
         return type(self)(variable, coords, name=name, fastpath=True)
 
     def _replace_maybe_drop_dims(self, variable, name=__default):
-        if variable.dims == self.dims:
+        if variable.dims == self.dims and variable.shape == self.shape:
             coords = self._coords.copy()
+        elif variable.dims == self.dims:
+            # Shape has changed (e.g. from reduce(..., keepdims=True)
+            new_sizes = dict(zip(self.dims, variable.shape))
+            coords = OrderedDict((k, v) for k, v in self._coords.items()
+                                 if v.shape == tuple(new_sizes[d]
+                                                     for d in v.dims))
         else:
             allowed_dims = set(variable.dims)
             coords = OrderedDict((k, v) for k, v in self._coords.items()
@@ -1637,7 +1643,8 @@ class DataArray(AbstractArray, DataWithCoords):
         """
         return ops.fillna(self, other, join="outer")
 
-    def reduce(self, func, dim=None, axis=None, keep_attrs=None, **kwargs):
+    def reduce(self, func, dim=None, axis=None, keep_attrs=None, keepdims=None,
+               **kwargs):
         """Reduce this array by applying `func` along some dimension(s).
 
         Parameters
@@ -1657,6 +1664,10 @@ class DataArray(AbstractArray, DataWithCoords):
             If True, the variable's attributes (`attrs`) will be copied from
             the original object to the new one.  If False (default), the new
             object will be returned without attributes.
+        keepdims : bool, optional
+            If True, the dimensions which are reduced are left in the result
+            as dimensions of size one. Coordinates that use these dimensions
+            are removed.
         **kwargs : dict
             Additional keyword arguments passed on to `func`.
 
@@ -1667,7 +1678,8 @@ class DataArray(AbstractArray, DataWithCoords):
             summarized data and the indicated dimension(s) removed.
         """
 
-        var = self.variable.reduce(func, dim, axis, keep_attrs, **kwargs)
+        var = self.variable.reduce(func, dim, axis, keep_attrs, keepdims,
+                                   **kwargs)
         return self._replace_maybe_drop_dims(var)
 
     def to_pandas(self):
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1643,8 +1643,8 @@ class DataArray(AbstractArray, DataWithCoords):
         """
         return ops.fillna(self, other, join="outer")
 
-    def reduce(self, func, dim=None, axis=None, keep_attrs=None, keepdims=False,
-               **kwargs):
+    def reduce(self, func, dim=None, axis=None, keep_attrs=None,
+               keepdims=False, **kwargs):
         """Reduce this array by applying `func` along some dimension(s).
 
         Parameters
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1643,7 +1643,7 @@ class DataArray(AbstractArray, DataWithCoords):
         """
         return ops.fillna(self, other, join="outer")
 
-    def reduce(self, func, dim=None, axis=None, keep_attrs=None, keepdims=None,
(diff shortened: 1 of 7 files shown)


xarray/core/variable.py
```diff
@@ -1334,7 +1334,7 @@ def where(self, cond, other=dtypes.NA):
         return ops.where_method(self, cond, other)
 
     def reduce(self, func, dim=None, axis=None,
-               keep_attrs=None, allow_lazy=False, **kwargs):
```
max-sixty: Is there a reason this isn't default `False`?
(I'm not sure why `keep_attrs` is default `None`, but looks like it's calling into `_get_keep_attrs` at least)

ScottWales: Yes it was to follow the behaviour of `keep_attrs`, happy to change the default to be `False`

xarray/core/dataarray.py
```diff
@@ -1657,6 +1664,10 @@ def reduce(self, func, dim=None, axis=None, keep_attrs=None, **kwargs):
             If True, the variable's attributes (`attrs`) will be copied from
             the original object to the new one.  If False (default), the new
             object will be returned without attributes.
+        keepdims : bool, optional
```
```suggestion
        keepdims : bool, default False
```

xarray/core/dataarray.py
```diff
@@ -1637,7 +1643,8 @@ def combine_first(self, other):
         """
         return ops.fillna(self, other, join="outer")
 
-    def reduce(self, func, dim=None, axis=None, keep_attrs=None, **kwargs):
+    def reduce(self, func, dim=None, axis=None, keep_attrs=None,
```
```suggestion
    def reduce(self, func, dim=None, axis=None, keep_attrs=False,
```

xarray/core/dataset.py
```diff
@@ -3149,6 +3149,10 @@ def reduce(self, func, dim=None, keep_attrs=None, numeric_only=False,
             If True, the dataset's attributes (`attrs`) will be copied from
             the original object to the new one.  If False (default), the new
             object will be returned without attributes.
+        keepdims : bool, optional
```
```suggestion
        keepdims : bool, default False
```

xarray/core/variable.py
```diff
@@ -1354,6 +1354,9 @@ def reduce(self, func, dim=None, axis=None,
             If True, the variable's attributes (`attrs`) will be copied from
             the original object to the new one.  If False (default), the new
             object will be returned without attributes.
+        keepdims : bool, optional
```
max-sixty:
```suggestion
        keepdims : bool, default False
```

xarray/core/dataarray.py
ScottWales: This seems outside the scope of the current change

ScottWales:
Thanks for the review @max-sixty, happy to help out

xarray/core/variable.py
```diff
@@ -1388,6 +1391,12 @@ def reduce(self, func, dim=None, axis=None,
             keep_attrs = _get_keep_attrs(default=False)
         attrs = self._attrs if keep_attrs else None
 
+        if keepdims:
+            for i, d in enumerate(self.dims):
+                if d not in dims:
+                    dims.insert(i, d)
+                    data = np.expand_dims(data, axis=i)
```
shoyer: This works on NumPy arrays, but won't work on dask arrays (or other "duck array" types that we want to support in the future).

Two alternatives that would work here:
1. define `expand_dims` in `duck_array_ops.py` using `_dask_or_eager_func`.
2. expand dimensions by indexing with a tuple built-up out of `np.newaxis`/`slice(None)` instead.

xarray/core/dataarray.py
max-sixty: Ah, wrong one. Thanks

xarray/core/variable.py
ScottWales: Thanks @shoyer, I've added option 2 and a test for dask variables

xarray/core/variable.py
```diff
@@ -1388,6 +1391,12 @@ def reduce(self, func, dim=None, axis=None,
             keep_attrs = _get_keep_attrs(default=False)
         attrs = self._attrs if keep_attrs else None
 
+        if keepdims:
```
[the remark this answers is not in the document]
ScottWales: This has been moved as suggested

xarray/core/variable.py
```diff
@@ -1381,8 +1384,19 @@ def reduce(self, func, dim=None, axis=None,
         else:
             removed_axes = (range(self.ndim) if axis is None
                             else np.atleast_1d(axis) % self.ndim)
-            dims = [adim for n, adim in enumerate(self.dims)
-                    if n not in removed_axes]
+            if keepdims:
+                # Insert np.newaxis for removed dims
+                slices = tuple(np.newaxis if i in removed_axes else
+                               slice(None, None) for i in range(self.ndim))
+                if getattr(data, 'shape', None) is None:
```
shoyer: I think this would be equivalent to just `hasattr(data, 'shape')`?

xarray/core/variable.py
```diff
@@ -1381,8 +1384,19 @@ def reduce(self, func, dim=None, axis=None,
         else:
             removed_axes = (range(self.ndim) if axis is None
                             else np.atleast_1d(axis) % self.ndim)
-            dims = [adim for n, adim in enumerate(self.dims)
-                    if n not in removed_axes]
+            if keepdims:
+                # Insert np.newaxis for removed dims
+                slices = tuple(np.newaxis if i in removed_axes else
+                               slice(None, None) for i in range(self.ndim))
```
shoyer: `slice(None)` is equivalent to `slice(None, None)`

shoyer:
Thanks @ScottWales!

PR #3038, opened 2019-06-21
Revert cmap fix

dcherian:
Unfortunately my fix in #2935 broke some major functionality. A proper fix would involve some facetgrid refactoring I think; so that'll take some time. This reverts that commit and adds a test.

```
xr.DataArray(np.random.randn(10, 20)).plot(levels=10, cmap=mpl.cm.RdBu)
```

image


shoyer:
It looks like there's a limit failure -- could you fix that?

dcherian:
Done.

PR #3039, opened 2019-06-23
Set up CI with Azure Pipelines (and remove Appveyor)

shoyer:
xref https://github.com/astropy/astropy/pull/8445


shoyer:
OK, this is actually working pretty well now.

Azure Pipelines is has loads of concurrent builds with a longer time-out and lets us use a single CI platform for everything (Linux, Mac, Window). The output is also pretty reasonable, and well integrated with GitHub.

The only feature we're missing here is integration with Coveralls for coverage statistics. I haven't looked hard for a way to do that yet, though.

At the very last, I am tempted to merge this and turn off our use of Appveyor.

jhamman:
Thanks for doing this @shoyer! This has been on my wish list for some time.

max-sixty:
Should we turn AppVeyor off now? Or wait until this is merged? I think the constant fails are a net negative. Happy to do it if I have permissions.

shoyer:
I think this is ready to merge, and probably worth doing soon given the broken state of our Appveyor config.

Switching away from Travis can wait until we figure out how to report coverage and are sure we've worked out the kinks. But it already seems a bit more usable -- Azure surfaces test failures much more prominently.

@max-sixty would you mind looking it over for any obvious issues?

azure-pipelines.yml
```diff
@@ -0,0 +1,109 @@
+trigger:
+  - master
```
max-sixty: If people fork this repo, I think this would only trigger CI runs on their `master` branch; ideally we would default to all branches (which I think would happen if we removed these lines)

azure-pipelines.yml
```diff
@@ -0,0 +1,109 @@
...
+
+pr:
+  - master
```
max-sixty: I'm guessing the branch name refers to the target branch, not the source branch, or this wouldn't have triggered? I couldn't find it in the docs.

Could add an autoCancel
```yaml
pr:
  branches:
    include: "*"
  autoCancel: true
```

azure-pipelines.yml
```diff
@@ -0,0 +1,109 @@
...
+  steps:
+  - template: ci/azure/install.yml
+  - bash: |
+      source activate test_env
+      flake8
+    displayName: flake8 lint checks
+  - bash: |
+      source activate test_env
+      mypy .
+    displayName: mypy type checks
```
max-sixty: GitHub incorrectly interprets this as part of the `bash` string; confirmed that other parsers correctly treat this as a new item

azure-pipelines.yml
```diff
@@ -0,0 +1,109 @@
...
+      py36-pandas-dev:
+        conda_env: py36-pandas-dev
+      py36-flakey:
+        conda_env: py36
+        pytest_extra_flags: --run-flaky --run-network-tests
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - template: ci/azure/unit-tests.yml
+  continueOnError: 'true'
```
max-sixty: Did you see whether this means a failure means a) it continues the other tasks but still results in a failure b) allows the failure

There's another setting `ignoreLASTEXITCODE: boolean` but I think that's only on Powershell (!) and I can't find something comparable

max-sixty: I looked at whether this could be merged with the `matrix` above, but unfortunately it looks like the [matrix](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#matrix) can only vary the variables

azure-pipelines.yml
```diff
@@ -0,0 +1,109 @@
...
+      source activate test_env
+      flake8
+    displayName: flake8 lint checks
+  - bash: |
+      source activate test_env
+      mypy .
+    displayName: mypy type checks
+    # allow mypy failures until we have a nice way of displaying results in
+    # pull requests
+    continueOnError: 'true'
```
max-sixty: Passing `'true'` in yaml means it's a string of `true`, not an actual boolean value

It's possible Azure will coerce to a bool though. 

`true`, `True`, `Yes` will all work

azure-pipelines.yml
shoyer: Good check, this seems to be entirely unnecessary.

azure-pipelines.yml
shoyer: Your suggestion appears to be the default value if we don't include a `pr` section at all, so I removed this.

azure-pipelines.yml
shoyer: also a good catch!

azure-pipelines.yml
shoyer: I tried defining this with a variable, let's see if that works.

shoyer:
@max-sixty thanks for the review! I've gone ahead and removed the Appveyor hook, so it should be entirely gone now.

max-sixty:
Amazing - thanks!

shoyer:
For anyone else following along, this needed one more minor follow-on: https://github.com/pydata/xarray/pull/3052

Issue #3026, opened 2019-06-17
Rename dims independently from coords?


nedclimaterisk:
I have a dataset that looks like this:

```python
<xarray.Dataset>
Dimensions:  (lat: 226, lon: 261, time: 7300)
Coordinates:
  * lat      (lat) float32 -32.0 -31.9 -31.8 -31.7 ... -9.700001 -9.6 -9.5
  * lon      (lon) float32 132.0 132.1 132.2 132.3 ... 157.7 157.8 157.9 158.0
  * time     (time) object 1980-01-01 15:00:00 ... 1999-12-31 15:00:00
Data variables:
    rnd24    (time, lat, lon) float32 ...
```

#### Problem description

I would like to be able to rename the dataset dimensions, *without renaming the coordinates*.

#### Expected Output

```python
<xarray.Dataset>
Dimensions:  (y: 226, x: 261, time: 7300)
Coordinates:
  * lat      (y) float32 -32.0 -31.9 -31.8 -31.7 ... -9.700001 -9.6 -9.5
  * lon      (x) float32 132.0 132.1 132.2 132.3 ... 157.7 157.8 157.9 158.0
  * time     (time) object 1980-01-01 15:00:00 ... 1999-12-31 15:00:00
Data variables:
    rnd24    (time, y, x) float32 ...
```

As far as I can tell, there is no way to do this. I can `rename` the existing dims/coords to x/y, and then manually create new coordinates that are copies of x and y, which gets me to:

```python
<xarray.Dataset>
Dimensions:  (time: 7300, x: 261, y: 226)
Coordinates:
  * y        (y) float32 -32.0 -31.9 -31.8 -31.7 ... -9.700001 -9.6 -9.5
  * x        (x) float32 132.0 132.1 132.2 132.3 ... 157.7 157.8 157.9 158.0
  * time     (time) object 1980-01-01 15:00:00 ... 1999-12-31 15:00:00
    lat      (y) float32 -32.0 -31.9 -31.8 -31.7 ... -9.700001 -9.6 -9.5
    lon      (x) float32 132.0 132.1 132.2 132.3 ... 157.7 157.8 157.9 158.0
Data variables:
    rnd24    (time, y, x) float32 ...
```

But it doesn't seem to be possible to re-assign the new coordinates as the indexes for the existing dims.

In this case, it may seem a bit redundant, because the coordinates are equal to the grid. But I'm trying to get this output to work with code that also deals with other datasets that have non- rectilinear grids.

[...]

PR #3042, opened 2019-06-25
Added rename_coords and rename_dims

jukent:
- [ ] Closes #xxxx
 - [ ] Tests added


xarray/core/dataset.py
```diff
@@ -2285,10 +2287,88 @@ def rename(self, name_dict=None, inplace=None, **names):
...
+            if k not in self.dims:
+                raise ValueError("cannot rename %r because it is not a "
+                                 "dimension in this dataset" % k)
+
+        dims = self._rename_dims(name_dict)
+        indexes = self._rename_indexes(name_dict)
+        return self._replace(dims=dims, indexes=indexes, inplace=inplace)
+
+
+    def rename_coords(self, name_dict=None, inplace=None, **names):
```
shoyer: I think it probably makes sense to call this `rename_vars()` (though we could also define a method for renaming coordinates only)

jukent: I was wondering about that. rename_vars seemed to be more consistent with the code.

jukent:
pydata#3026

There is some unexpected behavior. The new rename_dims function does not change the dimensions of the variables. Working on this now.

jukent:
> #3026
> 
> There is some unexpected behavior. The new rename_dims function does not change the dimensions of the variables. Working on this now.

Similarly rename_vars will rename the variable dimensions and variables, but not the dataset dimensions. I thought I was dealing with this with the _rename_indexes. But will have to keep working.

xarray/core/dataset.py
```diff
@@ -2285,10 +2287,88 @@ def rename(self, name_dict=None, inplace=None, **names):
...
+            if k not in self.dims:
+                raise ValueError("cannot rename %r because it is not a "
+                                 "dimension in this dataset" % k)
+
+        dims = self._rename_dims(name_dict)
+        indexes = self._rename_indexes(name_dict)
+        return self._replace(dims=dims, indexes=indexes, inplace=inplace)
+
+
+    def rename_vars(self, name_dict=None, inplace=None, **names):
```
max-sixty: What do people think about `vars` vs `variables` here?

I'm +0.2 on `variables` for consistency with `merge_variables` etc (but only did a quick search, I would update if there were more public `vars` methods around)

xarray/core/dataset.py
```diff
@@ -2285,10 +2287,88 @@ def rename(self, name_dict=None, inplace=None, **names):
...
+        """Returns a new object with renamed variables and coordinates
+
+        Parameters
+        ----------
+        name_dict : dict-like, optional
+            Dictionary whose keys are current variable or coordinate names and
+            whose values are the desired names.
+        inplace : bool, optional
```
max-sixty: Given inplace has been deprecated, I don't think we should add it here

xarray/core/dataset.py
```diff
@@ -2285,10 +2287,88 @@ def rename(self, name_dict=None, inplace=None, **names):
...
+                raise ValueError("cannot rename %r because it is not a "
+                                 "dimension in this dataset" % k)
+
+        dims = self._rename_dims(name_dict)
+        indexes = self._rename_indexes(name_dict)
+        return self._replace(dims=dims, indexes=indexes, inplace=inplace)
+
+
+    def rename_vars(self, name_dict=None, inplace=None, **names):
+        """Returns a new object with renamed variables and coordinates
```
max-sixty: Maybe technically this should say "variables including coordinates", but this isn't the only place where it's written like this (and let's not slow down this PR with these)

max-sixty:
This is great! Thanks @jukent ! Ping back here if there's anything we can do to help

max-sixty:
And tbc, we should add tests here. That will help clarify the issues above re exactly which items on the dataset each method should rename vs leave.

xarray/core/dataset.py
jukent: Done

xarray/core/dataset.py
jukent: Makes sense. I removed it, but it is still floating around in the other functions in the script (swap_dims for example)

jukent:
> And tbc, we should add tests here. That will help clarify the issues above re exactly which items on the dataset each method should rename vs leave.

Yes I need to add tests, but I don't have much experience with this yet. I will ask my team here to help me.

jukent:
> > #3026
> > There is some unexpected behavior. The new rename_dims function does not change the dimensions of the variables. Working on this now.
> 
> Similarly rename_vars will rename the variable dimensions and variables, but not the dataset dimensions. I thought I was dealing with this with the _rename_indexes. But will have to keep working.

This works now! But to do so I created three different renaming variable helper functions (var name only, var and dims, var dims only). I will clean this up since the majority of these three functions is the same.

max-sixty:
@jukent check out the existing rename tests here (they're a bit over-complicated at the beginning): https://github.com/pydata/xarray/blob/master/xarray/tests/test_dataset.py#L2046-L2130

You could do something as simple as this: https://github.com/pydata/xarray/blob/master/xarray/tests/test_dataset.py#L2114-L2119 - i.e. start with a dataset, call a rename method, and compare the dataset with a manually constructed version of the expected result.

Ping back with any issues

jukent:
> @jukent check out the existing rename tests here (they're a bit over-complicated at the beginning): https://github.com/pydata/xarray/blob/master/xarray/tests/test_dataset.py#L2046-L2130
> 
> You could do something as simple as this: https://github.com/pydata/xarray/blob/master/xarray/tests/test_dataset.py#L2114-L2119 - i.e. start with a dataset, call a rename method, and compare the dataset with a manually constructed version of the expected result.
> 
> Ping back with any issues

Thanks @max-sixty. I added some very simple tests. I might copy more of the test_rename functions (if I tried to rename a variable to the same name for example)

max-sixty:
Those look great!

max-sixty:
@jukent I don't think that's our bot

PR #3051, opened 2019-06-27
pytest deprecation

max-sixty:
We weren't trying to do this in the first place, so somewhat coincidental:

```
xarray/tests/test_combine.py::TestManualCombine::test_manual_concat_too_many_dims_at_once
  /home/vsts/work/1/s/xarray/tests/test_combine.py:347: PytestDeprecationWarning: raises(..., 'code(as_a_string)') is deprecated, use the context manager form or use `exec()` directly
```


shoyer:
thanks!

PR #3052, opened 2019-06-27
Replace Appveyor with Azure Pipelines in README and contributor guide

shoyer:


shoyer:
The Coveralls upload requires a secret variable, which by default is not available on forks. So for now I've just deleted it, given that we're still using Travis for calculating test coverage.

Issue #907, opened 2016-07-19
unstack() treats string coords as objects


crusaderky:
unstack() should be smart enough to recognise that all labels in a coord are strings, and convert them to numpy strings.
This is particularly relevant e.g. if you want to dump the xarray to netcdf and then read it with a non-python library.

``` python
import xarray

a = xarray.DataArray([[1,2],[3,4]], dims=['x', 'y'], coords={'x': ['x1', 'x2'], 'y': ['y1', 'y2']})
a
```

```
<xarray.DataArray (x: 2, y: 2)>
array([[1, 2],
       [3, 4]])
Coordinates:
  * y        (y) <U2 'y1' 'y2'
  * x        (x) <U2 'x1' 'x2'
```

``` python
a.stack(s=['x', 'y']).unstack('s')
```

```
<xarray.DataArray (x: 2, y: 2)>
array([[1, 2],
       [3, 4]])
Coordinates:
  * x        (x) object 'x1' 'x2'
  * y        (y) object 'y1' 'y2'
```

shoyer:
This is a tricky one to fix. Stack is not the only operation that can convert string arrays to dtype=object -- reindex and align also do this.

The challenge is that NumPy does not support variable width strings or strings with missing values. Hence, we end up with dtype=object arrays.

In this particular case, we could preserve the string dtype by keeping track of the original dtype of the arrays as metadata, similar to how we currently use [`PandasIndexAdapter`](https://github.com/pydata/xarray/blob/884b247742af2480e269bb54c9102f08c867fdaa/xarray/core/indexing.py#L425) to preserve the original dtype on arrays converted into non-multi indexes.

crusaderky:
Crude workaround:

```python
# Convert object coords to string
for dim in array.dims:
    if array.coords[dim].dtype == numpy.dtype('O'):
        array.coords[dim] = array.coords[dim].values.tolist()

return array
```
The above works, but if you have more than a handful elements in the coord the performance hit will be severe.

shoyer:
Does `array.coords[dim] = array.coords[dim].values.astype(str)` work? In theory, I think it should...

crusaderky:
Yes that works as well

crusaderky:
This issue is still valid as of xarray 0.11.0

benbovy:
This is fixed in v2022.6.0. Xarray's `PandasMultiIndex` wrapper keeps track of the level coordinate dtypes.

PR #3054, opened 2019-06-28
Flat iteration over DataArray

yohai:
- [ ] Tests added

A very simple way to iterate over `DataArray`s, imitating numpy's flat iterator. 

For a DataArray `da`, I implemented a flat iterator that works exactly like `da.values.flat`, but returns a singelton `DataArray` that has all the coordinates and meta data and whatnot.

I wrote it because I needed it for something and thought it would be useful for others too. I'd love to hear your opinions - if you think it's useful I'll write up some unittests.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2686,6 +2686,13 @@ class DataArray(AbstractArray, DataWithCoords):
         ds = self._to_temp_dataset().integrate(dim, datetime_unit)
         return self._from_temp_dataset(ds)
 
+    @property
+    def flat(self):
+        import itertools
+        gen = itertools.product(*(range(n) for n in self.shape))
+        for indices in gen:
+            yield self.__getitem__(indices)
+
     # this needs to be at the end, or mypy will confuse with `str`
     # https://mypy.readthedocs.io/en/latest/common_issues.html#dealing-with-conflicting-names  # noqa
     str = property(StringAccessor)


coroa:
@yohai It's a lot more efficient to simply iterate over the underlying array, ie. `da.values.flat`, if you can afford to hold everything in memory.

If you are instead using streaming computation based on dask, then you would have to do something similar on per-chunk basis.

crusaderky:
Indeed this is extremely inefficient

You can get the same with a much faster one-liner: ``a.stack(__flat=a.dims).reset_index('__flat')``.
On related notes,
- stack() could use a set_index=True optional parameter that avoids you from going through a MultiIndex if you don't need one
- stack() should accept non-string hashables; this would allow avoiding potential collisions (e.g. ``flat = object()``)
- there is an issue with the stack -> reset_index round-trip where it converts unicode variables to object #907

I'm afraid it's a -1 from me.

yohai:
@crusaderky @corora Thanks for your comments, glad to see that there's a more efficient way to do it. The question is do you think it's useful enough to justify adding it as a built in function. I end up using my solution quite often

crusaderky:
@yohai Iterating point by point in pure python over numpy data is horribly slow. ``numpy.ndarray.flat`` is there mostly to be used within cython/numba code. In a DataArray it's much worse than in a plain ndarray, because every time you invoke the slice operator to fetch a single element it's being applied to all coordinates too.

If you just need to iterate over the values of a DataArray, then ``DataArray.values.ravel().tolist()`` is the fastest option that I know of (``ndarray.tolist()`` is much faster than ``list(ndarray)``!).
If you need the coords as well, then I suspect you may be doing it wrong - could you show a simple example of your use case?

crusaderky:
Mh. Actually it looks like ``ndarray.flat`` is actually the fastest way to iterate over numpy. Still considerably slower than a CPython iterator though

```python
import numpy

N = 1000000
a = numpy.arange(N)

def exhaust(it):
    for _ in it:
        pass

%timeit exhaust(a)
24.8 ms ± 723 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit exhaust(a.flat)
20.4 ms ± 701 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit exhaust(a.tolist())
27.2 ms ± 1.16 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit exhaust(range(N))
10.5 ms ± 234 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

coroa:
@yohai : In short, no. It does not make sense to add a built-in function for iteration, if it is unable to augment the low-level functionality.

PR #3055, opened 2019-06-28
Better add-conda-to-path template for azure pipelines

shoyer:
Template time expansion should reduce the noise in the CI output.

Follow on to https://github.com/pydata/xarray/pull/3039


shoyer:
Nevermind, variables seem not to be available at template  expansion time -- see https://github.com/Microsoft/azure-pipelines-yaml/issues/30

PR #3058, opened 2019-06-28
Ensure xarray imports (and docs build) even without pandas.Panel

### LINKED ISSUE #3057: Remove remaining usage of pandas Panel
There is still some usage of `pd.Panel` (just accessing the name in the pandas namespace for isinstance check), eg:

https://github.com/pydata/xarray/blob/a78e0f6ce7c0cc48f746a122c79f1bcf903efd51/xarray/core/merge.py#L18

We are currently removing Panel from pandas (https://github.com/pandas-dev/pandas/pull/27101), so that breaks xarray (and statsmodels). For that reason, we are going to keep an empty class so that the above use case doesn't directly break. But it would still be good to remove that usage (or only add it for pandas versions below 0.25.0)

shoyer:
It's being removed in the next pandas 0.25 release
(see https://github.com/pandas-dev/pandas/pull/27101).

 - [x] Closes #3057

### DIFF
--- a/doc/pandas.rst
+++ b/doc/pandas.rst
@@ -151,10 +151,11 @@ However, you will need to set dimension names explicitly, either with the
 Transitioning from pandas.Panel to xarray
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-:py:class:`~pandas.Panel`, pandas' data structure for 3D arrays, has always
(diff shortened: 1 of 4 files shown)


shoyer:
@max-sixty your perspective on my doc edits would be appreciated

max-sixty:
Looks great! Thanks!

PR #3064, opened 2019-06-30
Upload coverage from Azure to codecov.io

shoyer:


shoyer:
I'm guessing codecov will start commenting on new PRs after it gets a configuration from master.

It's already updating reports on this pull request: https://codecov.io/gh/pydata/xarray/pull/3064

PR #3065, opened 2019-06-30
kwargs.pop() cleanup

crusaderky:
- Clean up everywhere the pattern
```
def my_func(*args, **kwargs):
    my_optional_arg = kwargs.pop('my_optional_arg', None)
```
which was inherited from not being able to put named keyword arguments after ``*args`` in Python 2.

- Fix bug in SplineInterpolator where the ``__init__`` method would write to the class attributes of BaseInterpolator.
- ``map_dataarray`` was unintentionally and subtly relying on ``_process_cmap_cbar_kwargs`` to modify the kwargs in place. ``_process_cmap_cbar_kwargs`` is now strictly read-only and the modifications in kwargs have been made explicit in the caller function.
- Rename all 'kwds' to 'kwargs' for sake of coherency

### DIFF
--- a/xarray/core/missing.py
+++ b/xarray/core/missing.py
@@ -1,7 +1,6 @@
 import warnings
-from collections.abc import Iterable
 from functools import partial
-from typing import Any, Callable, Dict
+from typing import Any, Callable, Dict, Sequence
 
 import numpy as np
 import pandas as pd
@@ -15,9 +14,10 @@ from .variable import Variable, broadcast_variables
 
 
 class BaseInterpolator:
-    '''gerneric interpolator class for normalizing interpolation methods'''
-    cons_kwargs = {}  # type: Dict[str, Any]
-    call_kwargs = {}  # type: Dict[str, Any]
+    """Generic interpolator class for normalizing interpolation methods
+    """
+    cons_kwargs = None  # type: Dict[str, Any]
+    call_kwargs = None  # type: Dict[str, Any]
     f = None  # type: Callable
     method = None  # type: str
 
@@ -30,13 +30,12 @@ class BaseInterpolator:
 
 
 class NumpyInterpolator(BaseInterpolator):
-    '''One-dimensional linear interpolation.
+    """One-dimensional linear interpolation.
 
     See Also
     --------
     numpy.interp
-    '''
-
+    """
     def __init__(self, xi, yi, method='linear', fill_value=None, period=None):
 
         if method != 'linear':
@@ -53,7 +52,7 @@ class NumpyInterpolator(BaseInterpolator):
         if fill_value is None:
             self._left = np.nan
             self._right = np.nan
-        elif isinstance(fill_value, Iterable) and len(fill_value) == 2:
+        elif isinstance(fill_value, Sequence) and len(fill_value) == 2:
             self._left = fill_value[0]
             self._right = fill_value[1]
         elif is_scalar(fill_value):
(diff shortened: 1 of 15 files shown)


crusaderky:
@shoyer ready for review and merge

xarray/core/missing.py
```diff
@@ -30,35 +30,30 @@ def __repr__(self):
...
+    """One-dimensional linear interpolation.
 
     See Also
     --------
     numpy.interp
-    '''
-
-    def __init__(self, xi, yi, method='linear', fill_value=None, **kwargs):
+    """
+    def __init__(self, xi, yi, method='linear', fill_value=None, period=None):
```
shoyer: You lost `**kwargs` here.

xarray/core/missing.py
```diff
@@ -30,35 +30,30 @@ def __repr__(self):
...
 
         if method != 'linear':
             raise ValueError(
                 'only method `linear` is valid for the NumpyInterpolator')
 
         self.method = method
         self.f = np.interp
-        self.cons_kwargs = kwargs
-        self.call_kwargs = {'period': self.cons_kwargs.pop('period', None)}
+        self.cons_kwargs = {}
```
shoyer: Should still be `kwargs`

xarray/plot/plot.py
```diff
@@ -241,23 +240,6 @@ def line(darray, *args, **kwargs):
                          'dimensions'.format(ndims=ndims))
 
     # Ensures consistency with .plot method
```
shoyer: This comment can be deleted now

xarray/core/missing.py
crusaderky: Yes, and I also lost the lines that were checking that kwargs could contain at most the period key :)

xarray/core/missing.py
crusaderky: No, because it's always empty after you remove period

xarray/core/missing.py
```diff
@@ -30,35 +30,30 @@ def __repr__(self):
...
+        self.call_kwargs = {'period': period}
 
         self._xi = xi
         self._yi = yi
 
-        if self.cons_kwargs:
-            raise ValueError(
-                'received invalid kwargs: %r' % self.cons_kwargs.keys())
-
```
crusaderky: ...as pointed out by the old code here

xarray/plot/plot.py
crusaderky: Done

xarray/core/missing.py
shoyer: yes, indeed

xarray/plot/plot.py
```diff
@@ -239,10 +239,6 @@ def line(darray, *args, row=None, col=None, figsize=None, aspect=None,
                          'Passed DataArray has {ndims} '
                          'dimensions'.format(ndims=ndims))
 
-    # Ensures consistency with .plot method
-    if args is ():
-        args = kwargs.pop('args', ())
```
shoyer: I'm not quite sure why, but this seems to be important. Now five matplotlib tests are failing.

crusaderky: Ah. Because you can have
```
>>> def f(*args, **kwargs):
...    return args, kwargs
    
>>> f(1, 2, args=(3, 4))
(1, 2) {'args': (3, 4)}
```

crusaderky: I figured it out, it's because line() calls itself
line() -> _easy_facetgrid() -> map_dataarray_line() -> line()
I tried changing it to something clearer but a lot more stuff broke. I suggest leaving it there for now and taking care of it separately. For now I just clarified.

crusaderky:
@shoyer all done. The worsened coverage is caused by removing a wealth of lines, so the lines that were already being missed weight more >__>

shoyer:
Thanks @crusaderky. I’m going to disable the codecov status check shortly. It’s nice to see the numbers, but it definitely is not helpful to block PRs merely because they are deleting lines of code that had test coverage!

max-sixty:
Much cleaner! Thanks @crusaderky!

crusaderky:
@shoyer maybe instead of percentage you can configure it to count the absolute number of lines not covered?

PR #3071, opened 2019-07-02
Try installing pytest-azurepipelines

shoyer:
See https://github.com/tonybaloney/pytest-azurepipelines

No user facing changes


shoyer:
Hmm... "0 errors / 5,173 warnings"

We should probably work on that :)

max-sixty:
> 5,173 warnings

I think most of those are matplotlib-related.... (possibly a couple of lines are generating almost all of them)

PR #3079, opened 2019-07-04
Fix printing summaries of multiindex coords

coroa:
Since merging #2293 summarize_variable displays the first and last few entries,
so we have to pass all along.

MWE:
```python
import xarray as xr
import numpy as np
x = np.arange(0, 90, 0.5)
y = np.arange(0, 90, 0.5)
da = xr.DataArray(np.random.random((len(x), len(y))), [('x', x), ('y', y)])
da.stack(z=('x', 'y'))
```
gives on master:
```
<xarray.DataArray (z: 32400)>
array([0.497724, 0.793955, 0.140193, ..., 0.05734 , 0.884105, 0.479093])
Coordinates:
  * z        (z) MultiIndex
  - x        (z) float64 0.0 0.0 0.0 0.0 0.0 0.0 0.0 ... 0.0 0.0 0.0 0.0 0.0 0.0
  - y        (z) float64 0.0 0.5 1.0 1.5 2.0 2.5 ... 12.5 13.0 13.5 14.0 14.5
```
and on this branch
```
<xarray.DataArray (z: 32400)>
array([0.497724, 0.793955, 0.140193, ..., 0.05734 , 0.884105, 0.479093])
Coordinates:
  * z     (z) MultiIndex
  - x     (z) float64 0.0 0.0 0.0 0.0 0.0 0.0 ... 89.5 89.5 89.5 89.5 89.5 89.5
  - y     (z) float64 0.0 0.5 1.0 1.5 2.0 2.5 ... 87.0 87.5 88.0 88.5 89.0 89.5
```

 - [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/xarray/tests/test_dataarray.py
+++ b/xarray/tests/test_dataarray.py
@@ -64,6 +64,22 @@ class TestDataArray:
           - level_2  (x) int64 1 2 1 2""")
         assert expected == repr(self.mda)
 
+    def test_repr_multiindex_long(self):
+        mindex_long = pd.MultiIndex.from_product(
+            [['a', 'b', 'c', 'd'], [1, 2, 3, 4, 5, 6, 7, 8]],
+            names=('level_1', 'level_2'))
+        mda_long = DataArray(list(range(32)),
+                             coords={'x': mindex_long}, dims='x')
+        expected = dedent("""\
+        <xarray.DataArray (x: 32)>
+        array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16, 17,
+               18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31])
+        Coordinates:
+          * x        (x) MultiIndex
+          - level_1  (x) object 'a' 'a' 'a' 'a' 'a' 'a' 'a' ... 'd' 'd' 'd' 'd' 'd' 'd'
+          - level_2  (x) int64 1 2 3 4 5 6 7 8 1 2 3 4 5 6 ... 4 5 6 7 8 1 2 3 4 5 6 7 8""")  # noqa: E501
+        assert expected == repr(mda_long)
+
     def test_properties(self):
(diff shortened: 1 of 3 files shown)


shoyer:
Thanks @coroa. Would you mind adding a test to verify your fix? It could probably go alongside `test_repr_multiindex`:
https://github.com/pydata/xarray/blob/23fc0d6f8a5affefb84e5f20dc0bef7ace4542e2/xarray/tests/test_dataarray.py#L57-L65

shoyer:
The test failures on pandas-dev are unrelated. If you merge in master, I believe they should be fixed.

coroa:
Ok, rebased on master, added tests. Was not sooo easy to make linter and pytest happy :), but we're there. Hope skipping `E501 line too long (... > 79 characters)` for the multiline test string is allowed.

https://github.com/pydata/xarray/blob/e4716a4c2b9df681eb9e43dd31a3c9ec6e104e44/xarray/tests/test_dataarray.py#L74-L81

xarray/tests/test_dataarray.py
```diff
@@ -37,6 +37,12 @@ def setup(self):
                                                  names=('level_1', 'level_2'))
         self.mda = DataArray([0, 1, 2, 3], coords={'x': self.mindex}, dims='x')
 
+        self.mindex_long = pd.MultiIndex.from_product(
```
shoyer: Would you mind only defining these variables in the `test_repr_multiindex_long` method? 

We try to only define variables in `setup()` that are used in multiple methods (and even then, it's often better to just repeat setup code)

coroa: Sure enough, was my initial thought as well, but since `self.{mindex,mda}` were only used once and defined in the fixture, i had arranged it like that.

coroa: Done

dcherian:
@coroa you should give yourself credit in `whats-new.rst`

shoyer:
thanks @coroa! this will be in the next xarray release

Issue #2954, opened 2019-05-10
Segmentation fault reading many groups from many files


gerritholl:
This is probably the wrong place to report it, but I haven't been able to reproduce this without using xarray.  Repeatedly opening NetCDF4/HDF5 files and reading a group from them, triggers a Segmentation Fault after about 130–150 openings.  See details below.

#### Code Sample, a copy-pastable example if possible

```python
from itertools import count, product
import netCDF4
import glob
import xarray

files = sorted(glob.glob("/media/nas/x21308/2019_05_Testdata/MTG/FCI/FDHSI/uncompressed/20170410_RC70/*BODY*.nc"))

# get all groups
def get_groups(ds, pre=""):
    for g in ds.groups.keys():
        nm = pre + "/" + g
        yield from get_groups(ds[g], nm)
        yield nm
with netCDF4.Dataset(files[0]) as ds:
    groups = sorted(list(get_groups(ds)))
print("total groups", len(groups), "total files", len(files))
ds_all = []
ng = 20
nf = 20
print("using groups", ng, "using files", nf)
for (i, (g, f)) in zip(count(), product(groups[:ng], files[:nf])):
    print("attempting", i, "group", g, "from", f)
    ds = xarray.open_dataset(
            f, group=g, decode_cf=False)
    ds_all.append(ds)
```
#### Problem description

I have 70 NetCDF-4 files with 70 groups each.  When I cycle through the files and read one group from them at the time, after about 130–150 times, the next opening fails with a Segmentation Fault.  If I try to read one group from one file at the time, that would require a total of 70*70=4900 openings.  If I limit to 20 groups from 20 files in total, it would require 400 openings.  In either case, it fails after about 130–150 times.  I'm using the Python xarray interface, but the error occurs in the HDF5 library.  The message belows includes the traceback in Python:

```
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 140107218855616:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      [9/1985]
  #000: H5D.c line 485 in H5Dget_create_plist(): Can't get creation plist
    major: Dataset
    minor: Can't get value
  #001: H5Dint.c line 3159 in H5D__get_create_plist(): can't get dataset's creation property list
    major: Dataset
    minor: Can't get value
  #002: H5Dint.c line 3296 in H5D_get_create_plist(): datatype conversion failed
    major: Dataset                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
[...]

PR #3082, opened 2019-07-05
Cache root netCDF4.Dataset objects instead of groups

shoyer:
NetCDF-C and HDF5 are not threads-safe, so it's likely that closing the file
object associated with one group could invalidate other open groups from the
same file.

Now, we cache a single object corresponding to the root group for each file,
and access sub-groups on the fly as needed.

 - [x] Closes https://github.com/pydata/xarray/issues/2954
 - [x] Tests added


gerritholl:
As the original reported for #2954 I can confirm that both of my test scripts that were previously segfaulting are with this PR running as expected.

shoyer:
Awesome, thanks for checking @gerritholl !

It would be great if someone (maybe you?) could review this PR for any potential issues. If not, I will probably merge this in a day or two.

gerritholl:
@shoyer I'm afraid I don't understand well enough what is going to say much usefully...

PR #3085, opened 2019-07-05
Fix codecov reports

shoyer:
The location of the XML output was being changed by pytest-azurepipelines.


shoyer:
For now I've removed pytest-azurepipelines altogether. That seems to the only way to get codecov.io working.

PR #3086, opened 2019-07-06
Add broadcast_like.

### LINKED ISSUE #2885: Add broadcast_like?
What do you think of adding `da.broadcast_like(a)`?

This would be equivalent to `xr.broadcast(a, da)[1]`

dcherian:
- [x] Closes #2885
 - [x] Tests added


shoyer:
In principle, it feels like it would be better to have both `xarray.broadcast()` and `broadcast_like` written in terms of a lower level helper method that broadcasts only one object. This is what we do for `align`/`reindex_like`, for example. But in practice this is certainly way easier to write and probably good enough...

xarray/core/alignment.py
```diff
@@ -391,6 +391,58 @@ def reindex_variables(
...
+                # ignore dim not in var.dims
+                var_dims_map[dim] = var.shape[var.dims.index(dim)]
+
+        return var.set_dims(var_dims_map)
+
+    def _broadcast_array(array):
+        data = _set_dims(array.variable)
+        coords = OrderedDict(array.coords)
+        coords.update(common_coords)
+        return DataArray(data, coords, data.dims, name=array.name,
```
max-sixty: Could potentially use `._replace`, and then only supply the changed items (but not a big deal)

dcherian: Not sure what you mean

max-sixty: i.e. `return array._replace(data, coords)`
So `array` keeps its type if someone inherited from `DataArray` and we're not coupled to specific DataArray properties if those ever change

shoyer:
I'm merging this now, so @DavidMertz can put his doc/test improvements on top of it.

xarray/core/alignment.py
shoyer: I agree, but let's save this for a later refactor.

shoyer:
thanks @dcherian !

dcherian:
Sounds good. @DavidMertz's docs/tests are much better!

PR #3087, opened 2019-07-06
Suppress warnings and add test coverage

shoyer:
- Suppressed various warnings.
- PsuedoNetCDF was not getting tested. Now it is.
- Added test coverage for `print_versions.py`

### DIFF
--- a/ci/requirements/py36.yml
+++ b/ci/requirements/py36.yml
@@ -8,7 +8,9 @@ dependencies:
   - distributed
   - h5py
   - h5netcdf
-  - matplotlib=3.0  # for https://github.com/barronh/pseudonetcdf/issues/69
+  # pin matplotlib for now, to ensure that we have at least one CI job
+  # testing pseudonetcdf: https://github.com/barronh/pseudonetcdf/issues/69
+  - matplotlib=3.0
   - netcdf4
   - pytest
   - pytest-cov
--- a/ci/requirements/py36.yml
+++ b/ci/requirements/py36.yml
@@ -8,7 +8,7 @@ dependencies:
   - distributed
   - h5py
   - h5netcdf
-  - matplotlib
+  - matplotlib=3.0  # for https://github.com/barronh/pseudonetcdf/issues/69
   - netcdf4
   - pytest
   - pytest-cov
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,4 @@
 [run]
-branch = True
 omit =
     xarray/tests/*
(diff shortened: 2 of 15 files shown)


ci/requirements/py36.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - distributed
   - h5py
   - h5netcdf
-  - matplotlib
+  - matplotlib=3.0  # for https://github.com/barronh/pseudonetcdf/issues/69
```
crusaderky: Why only py36.yml? What about the other requirements files?
Also, will this restriction appear in the conda recipe?

shoyer: This is just to ensure that we have at least one test job testing xarray with psuedonetcdf, which doesn't import in matplotlib 3.1 due to the linked issue. Xarray itself works perfectly fine on matplotlib 3.1.

dcherian:
I'm also seeing
```
xarray/core/dataset.py:3666
  /home/deepak/work/python/xarray/xarray/core/dataset.py:3666: FutureWarning: Converting timezone-aware DatetimeArray to timezone-naive ndarray with 'datetime64[ns]' dtype. In the future, this will return an ndarray with 'object' dtype where each element is a 'pandas.Timestamp' with the correct 'tz'.
  	To accept the future behavior, pass 'dtype=object'.
  	To keep the old behavior, pass 'dtype="datetime64[ns]"'.
    data = np.asarray(series).reshape(shape)

/home/deepak/miniconda3/envs/dcpy/lib/python3.7/site-packages/pandas/core/apply.py:286
  /home/deepak/miniconda3/envs/dcpy/lib/python3.7/site-packages/pandas/core/apply.py:286: FutureWarning: Converting timezone-aware DatetimeArray to timezone-naive ndarray with 'datetime64[ns]' dtype. In the future, this will return an ndarray with 'object' dtype where each element is a 'pandas.Timestamp' with the correct 'tz'.
  	To accept the future behavior, pass 'dtype=object'.
  	To keep the old behavior, pass 'dtype="datetime64[ns]"'.
    results[i] = self.f(v)
```

locally.

shoyer:
Yeah, I'm not quite sure how we want to handle the datetime64 with timezones warnings. Possibly we just want to silence them in our test suite, and enocourage users to explicitly cast their pandas data to a dtype supported by numpy before loading it into xarray. Doing that properly would require the equivalent of "re-raising" a warning.

shoyer:
I'm going to merge this shortly, and leave further clean-up for later.

PR #3088, opened 2019-07-07
More annotations

crusaderky:
A little incremental addition to type annotations. By no means complete, but it should be ready for merge in its own right nonetheless.

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -161,13 +161,15 @@ class AttrAccessMixin:
     _initialized = False
 
     @property
-    def _attr_sources(self):
-        """List of places to look-up items for attribute-style access"""
+    def _attr_sources(self) -> List[Mapping[Hashable, Any]]:
+        """List of places to look-up items for attribute-style access
+        """
         return []
 
     @property
-    def _item_sources(self):
-        """List of places to look-up items for key-autocompletion """
+    def _item_sources(self) -> List[Mapping[Hashable, Any]]:
+        """List of places to look-up items for key-autocompletion
+        """
         return []
 
     def __getattr__(self, name: str) -> Any:
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -6,7 +6,7 @@ from collections import OrderedDict, defaultdict
 from distutils.version import LooseVersion
 from numbers import Number
 from typing import (Any, Dict, Hashable, Iterable, Iterator, List,
-                    Mapping, Optional, Sequence, Set, Tuple, Union)
+                    Mapping, Optional, Sequence, Set, Tuple, Union, cast)
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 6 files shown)


xarray/core/dataset.py
```diff
@@ -1205,24 +1209,28 @@ def indexes(self) -> 'Mapping[Any, pd.Index]':
...
-    def data_vars(self):
+    def data_vars(self) -> DataVariables:
         """Dictionary of DataArray objects corresponding to data variables
         """
         return DataVariables(self)
 
-    def set_coords(self, names, inplace=None):
+    def set_coords(
+        self,
+        names: 'Union[Hashable, Iterable[Hashable]]',
```
crusaderky: It would have been nicer to use Collection[Hashable], but Collection requires Python 3.6

crusaderky: Note: Sequence is NOT ok, because sets are not sequences

xarray/core/dataset.py
```diff
@@ -1241,19 +1249,26 @@ def set_coords(self, names, inplace=None):
         # nb. check in self._variables, not self.data_vars to insure that the
         # operation is idempotent
         inplace = _check_inplace(inplace)
-        if isinstance(names, str):
+        if isinstance(names, str) or not isinstance(names, Iterable):
             names = [names]
+        else:
+            names = list(names)
```
crusaderky: This forced casting could have been avoided with Collection :(

xarray/core/dataset.py
```diff
@@ -2974,7 +2992,11 @@ def unstack(self, dim=None):
             result = result._unstack_once(dim)
         return result
 
-    def update(self, other, inplace=None):
+    def update(
+        self,
+        other: 'Union[Dataset, Mapping[Hashable, DataArray]]',
```
crusaderky: unsure about this - are there more options available for 'other'?

shoyer: This goes through the `merge` machinery (same as `Dataset.__init__`), so in principle I think you can put `Variable` objects or tuples of Variable arguments in the mapping values, too.

xarray/core/common.py
```diff
@@ -161,13 +161,15 @@ class AttrAccessMixin:
     _initialized = False
 
     @property
-    def _attr_sources(self):
-        """List of places to look-up items for attribute-style access"""
+    def _attr_sources(self) -> List[Mapping[Hashable, Any]]:
```
max-sixty: Is `Mapping[Hashable, Any]` identical to `Mapping`?

crusaderky: There is no requirement for the keys of a Mapping to be hashable. This is relevant for the various setter methods Dataset.attrs which cast any Mapping to an OrderedDict.

shoyer:
thanks @crusaderky !

PR #3090, opened 2019-07-09
WIP: more annotations

crusaderky:


xarray/core/dataarray.py
```diff
@@ -871,13 +871,19 @@ def chunks(self) -> Optional[Tuple[Tuple[int, ...], ...]]:
...
-        Mapping[Hashable, Union[None, int, Tuple[int, ...]]],
-    ] = None,
-            name_prefix: str = 'xarray-',
-            token: Optional[str] = None,
-            lock: bool = False) -> 'DataArray':
+    def chunk(
+        self,
+        chunks: Union[
+            None,
+            Number,
```
shoyer: Last I checked, type checkers were much happier with `int` rather than `numbers.Number`. Maybe mypy has been updated since then, though....

crusaderky: dask will happily accept float, numpy.int64, etc. and transparently cast them to int.
Should we decide that dask is excessively nice and the _intended_ input must strictly be int?

max-sixty: I vote to lean towards wider input bounds - it much more annoying for users to have false negatives than false positives; so in this case `Number` assuming mypy understands that

shoyer: OK, let's give this a try. We can always change later, that's not a big deal.

One other minor note -- it would be great to define a few type aliases for these longer types. The line wrapping you've done here helps, but this is still very long to write in a function signature.

shoyer:
Thanks @crusaderky. This is great, please keep them coming!

By the way, codecov updates its coverage reports as the tests on Azure complete. So usually in the first comment you'll see absurdly low coverage, but if you wait another ten minutes it will edit with the complete results. I don't know how to fix this easily, but fortunately it's only a minor annoyance.

crusaderky:
@shoyer On the first run the pytests themselves (not codecov) were showing half of the tests as "skipped". As if all optional dependencies randomly and silently failed to install. Not sure why. As a direct result codecov was showing a drop in coverage.

PR #3101, opened 2019-07-12
Improve open_mfdataset deprecation warnings

### LINKED ISSUE #3091: auto_combine warnings with open_mfdataset
I'm seeing this warning when using `open_mfdataset(list_of_files)` on master.
```
/gpfs/u/home/dcherian/python/xarray/xarray/backends/api.py:783: FutureWarning: In xarray version 0.13 `auto_combine` will be deprecated.
/gpfs/u/home/dcherian/python/xarray/xarray/backends/api.py:783: FutureWarning: The datasets supplied have global dimension coordinates. You may want
to use the new `combine_by_coords` function (or the
`combine='by_coords'` option to `open_mfdataset` to order the datasets
before concatenation. Alternatively, to continue concatenating based
on the order the datasets are supplied in in future, please use the
new `combine_nested` function (or the `combine='nested'` option to
open_mfdataset).
```

I don't think we should be issuing this warning when using `open_mfdataset` with default kwargs. Also there's a typo: "in in future"

TomNicholas:
- [x] Closes #3091
 - [x] Tests added

### DIFF
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -523,7 +523,8 @@ _CONCAT_DIM_DEFAULT = '__infer_concat_dim__'
 
 
 def auto_combine(datasets, concat_dim='_not_supplied', compat='no_conflicts',
-                 data_vars='all', coords='different', fill_value=dtypes.NA):
+                 data_vars='all', coords='different', fill_value=dtypes.NA,
+                 from_openmfds=False):
     """
     Attempt to auto-magically combine the given datasets into one.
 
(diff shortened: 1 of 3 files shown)


dcherian:
Awesome. I think this looks great. Sorry for not catching it earlier.

max-sixty:
That's perfect! Thanks @TomNicholas

xarray/core/combine.py
```diff
@@ -523,7 +523,8 @@ def combine_by_coords(datasets, compat='no_conflicts', data_vars='all',
 
 
 def auto_combine(datasets, concat_dim='_not_supplied', compat='no_conflicts',
-                 data_vars='all', coords='different', fill_value=dtypes.NA):
+                 data_vars='all', coords='different', fill_value=dtypes.NA,
```
shoyer: Don't worry about it here -- it's not worth the trouble for a deprecated function -- but in the future when you want something like this the clean way to do it is probably to make a private helper function `_auto_combine` that adds the extra non-user-facing arguments.

shoyer:
thanks @TomNicholas !

PR #3102, opened 2019-07-12
mfdataset, concat now support the 'join' kwarg.

### LINKED ISSUE #1354: concat automagically outer-joins coordinates
I would like to concatenate two netCDF files that have `float64` coordinate variables. I thought the coordinate values were the same, but in fact they differ by something in the order of `1e-14`. 

Using `open_mfdataset` or `concat` to merge the datasets, I get a completely different output shape than the two input files have.

This is because `concat` is somewhere performing an outer join on the coordinates. Now I am wondering where else in my workflows this might happen without my notice...

It would be awesome if there was an option to change this behaviour on `concat`, `open_mfdataset`, and `auto_combine`. I would actually rather make these functions **fail if any dimension other than the concatenation dimension differs**.

Note: This could also be a special case, because
```python
>>> (ds1.lon == ds2.lon).all()
<xarray.DataArray 'lon' ()>
array(True, dtype=bool)
```
while
```python
>>> (ds1.lon.values == ds2.lon.values).all()
False
```

Maybe an interface change could be considered together with that discussed in #1340?

dcherian:
- [x] Closes #1354
 - [x] Tests added

I won't work on it for the next few days if someone else wants to take over...

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -609,7 +609,7 @@ def open_mfdataset(paths, chunks=None, concat_dim='_not_supplied',
                    compat='no_conflicts', preprocess=None, engine=None,
                    lock=None, data_vars='all', coords='different',
                    combine='_old_auto', autoclose=None, parallel=False,
-                   **kwargs):
+                   join='outer', **kwargs):
     """Open multiple files as a single dataset.
 
     If combine='by_coords' then the function ``combine_by_coords`` is used to 
@@ -704,6 +704,8 @@ def open_mfdataset(paths, chunks=None, concat_dim='_not_supplied',
     parallel : bool, optional
         If True, the open and preprocess steps of this function will be
         performed in parallel using ``dask.delayed``. Default is False.
+    join : {'outer', 'inner', 'left', 'right', 'exact'}, optional
+        Passed on to align.
     **kwargs : optional
         Additional arguments passed on to :py:func:`xarray.open_dataset`.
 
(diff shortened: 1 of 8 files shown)


xarray/backends/api.py
```diff
@@ -704,6 +704,8 @@ def open_mfdataset(paths, chunks=None, concat_dim='_not_supplied',
     parallel : bool, optional
         If True, the open and preprocess steps of this function will be
         performed in parallel using ``dask.delayed``. Default is False.
+    join : {'outer', 'inner', 'left', 'right', 'exact'}, optional
+        How to combine objects with different indexes.
```
TomNicholas: This seems a bit vague - should it at least direct the user towards the documentation for `align`?

xarray/tests/test_backends.py
```diff
@@ -2368,8 +2368,12 @@ class TestOpenMFDatasetWithDataVarsAndCoordsKw:
     var_name = 'v1'
 
     @contextlib.contextmanager
-    def setup_files_and_datasets(self):
+    def setup_files_and_datasets(self, fuzz=0):
         ds1, ds2 = self.gen_datasets_with_common_coord_and_time()
+
+        # to test join='exact'
+        ds1['x'] = ds1.x + fuzz
```
TomNicholas: This is a pretty neat and self-explanatory way of doing this

xarray/backends/api.py
dcherian: Updated docstrings

dcherian:
Thanks for the review @TomNicholas. I've updated the docstrings and added tests to `test_combine.py`. I think this should be good to go.

PR #3106, opened 2019-07-13
Replace sphinx_gallery with notebook

rabernat:
Today @jhamman and I discussed how to refactor our somewhat fragmented "examples". We decided to basically copy the approach of the [dask-examples](https://github.com/dask/dask-examples) repo, but have it live here in the main xarray repo. Basically this approach is:
- all examples are notebooks
- examples are rendered during doc build by nbsphinx
- we will eventually have a binder that works with all of the same examples

This PR removes the dependency on sphinx_gallery and replaces the existing gallery with a standalone notebook called `visualization_gallery.ipynb`. However, not all of the links that worked in the gallery work here, since we are now using nbsphinx to render the notebooks (see https://github.com/spatialaudio/nbsphinx/issues/308).

Really important to get @dcherian's feedback on this, as he was the one who originally introduced the gallery. My view is that having everything as notebooks makes examples easier to maintain. But I'm curious to hear other views.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -25,6 +25,9 @@ dependencies:
   - sphinx_rtd_theme=0.4.2
   - mock=2.0.0
   - nbsphinx=0.4.2
+<<<<<<< HEAD
   - jupyter_client=5.3.1
   - ipykernel=5.1.1
+=======
+>>>>>>> switching out examples to use nbsphinx
   - pip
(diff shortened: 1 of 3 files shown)


dcherian:
It wasn't me!

rabernat:
I'm gonna merge this so we can use it at the SciPy sprint. Can always revert later.

PR #3112, opened 2019-07-13
More annotations in Dataset

crusaderky:


xarray/core/alignment.py
```diff
@@ -169,8 +182,8 @@ def deep_align(objects, join='inner', copy=True, indexes=None,
 
     This function is not public API.
     """
-    from .dataarray import DataArray
-    from .dataset import Dataset
+    from .dataarray import DataArray  # noqa: F811
+    from .dataset import Dataset  # noqa: F811
```
crusaderky: This is a workaround to a flake8 limitation https://gitlab.com/pycqa/flake8/issues/553

xarray/core/dataset.py
```diff
@@ -2542,39 +2609,49 @@ def expand_dims(self, dim=None, axis=None, **dim_kwargs):
...
         """
-        if isinstance(dim, int):
-            raise TypeError('dim should be str or sequence of strs or dict')
-        elif isinstance(dim, str):
+        if dim is None:
+            pass
+        elif isinstance(dim, Mapping):
+            # We're later going to modify dim in place; don't tamper with
+            # the input
+            dim = OrderedDict(dim)
```
crusaderky: Fixed bug where dim was modified in place, tainting the input variable. Also, fixed crash if the input dim is a read-only mapping.

xarray/core/dataset.py
```diff
@@ -2635,8 +2711,13 @@ def expand_dims(self, dim=None, axis=None, **dim_kwargs):
...
 
-    def set_index(self, indexes=None, append=False, inplace=None,
-                  **indexes_kwargs):
+    def set_index(
+        self,
+        indexes: Mapping[Hashable, Union[Hashable, Sequence[Hashable]]] = None,
+        append: bool = False,
+        inplace: bool = None,
+        **indexes_kwargs: Union[Hashable, Sequence[Hashable]]
+    ) -> 'Dataset':
```
crusaderky: Worth noting the behaviour is inconsistent with DataArray and pandas, which return None when inplace=True. Changing it however would benefit nobody and would break stuff.

crusaderky:
Ready for review and merge

crusaderky:
@shoyer @max-sixty ready for review and merge

xarray/core/dataarray.py
```diff
@@ -1329,9 +1334,13 @@ def expand_dims(self, dim: Union[None, Hashable, Sequence[Hashable],
...
 
-    def set_index(self, indexes: Optional[Mapping[Hashable, Any]] = None,
-                  append: bool = False, inplace: Optional[bool] = None,
-                  **indexes_kwargs: Any) -> Optional['DataArray']:
+    def set_index(
+        self,
+        indexes: Mapping[Hashable, Union[Hashable, Sequence[Hashable]]] = None,
+        append: bool = False,
+        inplace: bool = None,
+        **indexes_kwargs: Union[Hashable, Sequence[Hashable]]
```
max-sixty: I'm guessing this is fine because mypy passes - my original thought was whether this should be `Optional[Mapping]` given it was either `None` or a dict?

xarray/core/dataset.py
```diff
@@ -2674,7 +2755,12 @@ def set_index(self, indexes=None, append=False, inplace=None,
         return self._replace_vars_and_dims(variables, coord_names=coord_names,
                                            inplace=inplace)
 
-    def reset_index(self, dims_or_levels, drop=False, inplace=None):
+    def reset_index(
+        self,
+        dims_or_levels: Union[Hashable, Sequence[Hashable]],
+        drop: bool = False,
+        inplace: bool = None,
```
max-sixty: `inplace` should go away soon so shouldn't matter... but should this be `Optional[bool]` given the default is `None`?

xarray/core/merge.py
```diff
@@ -14,6 +27,21 @@
     from .dataarray import DataArray
     from .dataset import Dataset
 
+    DatasetLikeValue = Union[
+        DataArray,
+        Variable,
+        Tuple[Hashable, Any],
+        Tuple[Sequence[Hashable], Any],
+    ]
```
max-sixty: 🙌

xarray/core/dataarray.py
crusaderky: As of mypy 0.701, ``x: MyType = None`` and ``x: Optional[MyType] = None`` are the same.

```python
from typing import Optional


def f1(x: int = None) -> None:
    pass

def f2(x: Optional[int] = None) -> None:
    pass

def f3(x: int = 1) -> None:
    pass

def f4(x: Optional[int] = 1) -> None:
    pass


f1()
f2()
f3()
f4()
f1(1)
f2(1)
f3(1)
f4(1)
f1(None)
f2(None)
f3(None)
f4(None)
```
mypy output:
```
test_mypy.py:27: error: Argument 1 to "f3" has incompatible type "None"; expected "int"
```

xarray/core/dataset.py
crusaderky: See above

xarray/core/dataarray.py
crusaderky: @max-sixty I did a generalised cleanup of all already existing cases; I find it's much more readable now

max-sixty:
Great! Agree re `Optional`, thanks for clearing up. 
Good to go from my end

crusaderky:
@shoyer this is ready for merge unless there are comments

doc/whats-new.rst
```diff
@@ -63,6 +63,14 @@ New functions/methods
...
   By `Julia Kent <https://github.com/jukent>`_.
 
+- The xarray package is now discoverably by mypy (although typing hints
+  coverage is not complete yet). mypy users can now remove from their setup.cfg
+  the lines::
+
+    [mypy-xarray]
+    ignore_missing_imports = True
+
+   By `Guido Imperiale <https://github.com/crusaderky>`_
```
max-sixty: Let's move this up to 0.12.4?

max-sixty:
Thanks @crusaderky ! I think fine to merge given we've both reviewed

shoyer:
+1 please go ahead without me!

doc/whats-new.rst
crusaderky: Done

crusaderky:
@max-sixty @shoyer can we merge please?

max-sixty:
Sorry @crusaderky - we were all waiting on each other! My mistake

max-sixty:
Thanks again for the PR

Issue #2916, opened 2019-04-24
xr.open_dataset no groups info


ThetomekK:
#### Code Sample, a copy-pastable example if possible

I have to write some data to disk using xr.to_netcdf() method. The data must be structured in groups so i use the group key word. Reading the .nc file from disk leads to empty data variables if group not supplied. Here some samples
```python
data_ds.to_netcdf(path=savepath,mode='w',format='NETCDF4',group='Audio',engine='netcdf4')

datafromdisk = xr.open_dataset(savepath)
datafromdisk

<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*

datafromdisk = xr.open_dataset(savepath,group='Audio')

<xarray.Dataset>
Dimensions:  (time: 15360000)
Coordinates:
  * time     (time) datetime64[ns] 2017-05-30T07:40:00 ... 2017-05-30T07:49:59.992280938
Data variables:
    audio    (time) float32 ...
Attributes:
    unit:     Pa

```
#### Problem description
Actually this is not a real problem, if you keep in mind what data groups are strored in a .nc file. 
At the moment, i work arround with netCDF4 to get infos about possible groups within a .nc file.

```python

from netCDF4 import Dataset
rootgrp = Dataset(savepath)
rootgrp

<class 'netCDF4._netCDF4.Dataset'>
root group (NETCDF4 data model, file format HDF5):
    dimensions(sizes): 
    variables(dimensions): 
    groups: Audio

# or

rootgrp.groups

OrderedDict([('Audio', <class 'netCDF4._netCDF4.Group'>
              group /Audio:
                  unit: Pa
                  dimensions(sizes): time(15360000)
                  variables(dimensions): float32 audio(time), float64 time(time)
                  groups: )])
```

#### Expected Output
Well, at least something like this i would appreciate:
```python
datafromdisk = xr.open_dataset(savepath)
datafromdisk

<class 'netCDF4._netCDF4.Dataset'>
root group (NETCDF4 data model, file format HDF5):
    dimensions(sizes): 
    variables(dimensions): 
    groups: Audio
```

#### Output of ``xr.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.1 | packaged by conda-forge | (default, Mar 13 2019, 13:32:59) [MSC v.1900 64 bit (AMD64)]
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
[...]

PR #3113, opened 2019-07-13
[DOC] Be more explicit about netCDF groups and xarray.Dataset relationship.

zdgriffith:
This PR is related to the discussion in #2916.  In https://xarray.pydata.org/en/stable/io.html#netcdf there are statements like

>  netCDF files on disk directly correspond to Dataset objects.

This is useful conceptually but can be confusing when working with netCDF groups which are not supported as part of the data model.  I thought it would be clearer if the description of how to work with groups explicitly said this.


shoyer:
Thanks @zdgriffith, this is great!

Issue #1375, opened 2017-04-14
Sparse arrays


mrocklin:
I would like to have an XArray that has scipy.sparse arrays rather than numpy arrays.  Is this in scope?  

What would need to happen within XArray to support this?

shoyer:
Currently, we define most of our operations on duck arrays in https://github.com/pydata/xarray/blob/master/xarray/core/duck_array_ops.py

There are a few other hacks throughout the codebase, which can find by searching for "dask_array_type": https://github.com/pydata/xarray/search?p=1&q=dask_array_type&type=&utf8=%E2%9C%93

It's pretty crude, but basically this would need to be extended to implement many of these methods on for sparse arrays, too. Ideally we would define xarray's adapter logic into more cleanly separated submodules, perhaps using multiple dispatch. Even better, we would make this public API, so you can write something like `xarray.register_data_type(MySparseArray)` to register a type as valid for xarray's `.data` attribute.

It looks like `__array_ufunc__` will actually finally land in NumPy 1.13, which might make this easier.

See also https://github.com/pydata/xarray/pull/1118

rabernat:
👍  to the scipy.sparse array suggestion

[While we are discussing supporting other array types, we should keep gpu arrays on the radar:
https://documen.tician.de/pycuda/array.html]

benbovy:
Although I don't know much about [SciDB](http://scidb-py.readthedocs.io/en/stable/), it seems to be another possible application for  `xarray.register_data_type`.

mrocklin:
Here is a brief attempt at a multi-dimensional sparse array: https://github.com/mrocklin/sparse

It depends on numpy and scipy.sparse and, with the exception of a bit of in-memory data movement and copies, should run at scipy speeds (though I haven't done any benchmarking). 

@rabernat do you have an application that we could use to drive this?

rabernat:
> @rabernat do you have an application that we could use to drive this?

Nothing comes to mind quickly. My data are unfortunately quite dense! 😜

olgabot:
In case you're still looking for an application, [gene expression from single cells](https://github.com/olgabot/macosko2015) (see `data/00_original/GSM162679$i_P14Retina_$j.digital_expression.txt.gz`) is very sparse due to high gene dropout. The shape is `expression.shape (49300, 24760)` and it's mostly zeros or nans. A plain csv from this data was 2.5 gigs, which gzipped to 300 megs.

[...]

PR #3117, opened 2019-07-13
Support for __array_function__ implementers (sparse arrays) [WIP]

nvictus:
Doing a SciPy sprint. Working towards #1375

Together with https://github.com/pydata/sparse/pull/261, it seems to work.

 - [ ] Closes #xxxx
 - [x] Tests added


andersy005:
@nvictus, thank you for your work! 

I just tried this on CuPy arrays, and it seems to be working during array creation: 

```python
In [1]: import cupy as cp                                                           

In [2]: import xarray as xr                                                         

In [3]: x = cp.arange(6).reshape(2, 3).astype('f')                                  

In [4]: y = cp.ones((2, 3), dtype='int')                           

In [5]: x                                                          
Out[5]: 
array([[0., 1., 2.],
       [3., 4., 5.]], dtype=float32)

In [6]: y                                                          
Out[6]: 
array([[1, 1, 1],
       [1, 1, 1]])

In [7]: y.device                                                   
Out[7]: <CUDA Device 0>

In [8]: x.device                                                   
Out[8]: <CUDA Device 0>

In [9]: ds = xr.Dataset()                                          

In [10]: ds['x'] = xr.DataArray(x, dims=['lat', 'lon'])            

In [11]: ds['y'] = xr.DataArray(y, dims=['lat', 'lon'])            

In [12]: ds                                                        
Out[12]: 
<xarray.Dataset>
Dimensions:  (lat: 2, lon: 3)
Dimensions without coordinates: lat, lon
Data variables:
    x        (lat, lon) float32 ...
    y        (lat, lon) int64 ...

In [13]: ds.x.data.device                                          
Out[13]: <CUDA Device 0>

In [14]: ds.y.data.device                                          
Out[14]: <CUDA Device 0>
```

 

Even though it failed when I tried applying an operation on the dataset, this is still awesome! I am very excited to see this feature in xarray:

```python
In [15]: m = ds.mean(dim='lat')                                    
-------------------------------------------------------------------
TypeError                         Traceback (most recent call last)
<ipython-input-15-8e4d5e7d5ee3> in <module>
----> 1 m = ds.mean(dim='lat')

/glade/work/abanihi/devel/pangeo/xarray/xarray/core/common.py in wrapped_func(self, dim, skipna, **kwargs)
     65                 return self.reduce(func, dim, skipna=skipna,
     66                                    numeric_only=numeric_only, allow_lazy=True,
---> 67                                    **kwargs)
     68         else:
     69             def wrapped_func(self, dim=None, **kwargs):  # type: ignore

/glade/work/abanihi/devel/pangeo/xarray/xarray/core/dataset.py in reduce(self, func, dim, keep_attrs, keepdims, numeric_only, allow_lazy, **kwargs)
   3532                                                  keepdims=keepdims,
   3533                                                  allow_lazy=allow_lazy,
-> 3534                                                  **kwargs)
   3535 
   3536         coord_names = set(k for k in self.coords if k in variables)

/glade/work/abanihi/devel/pangeo/xarray/xarray/core/variable.py in reduce(self, func, dim, axis, keep_attrs, keepdims, allow_lazy, **kwargs)
   1392         input_data = self.data if allow_lazy else self.values
   1393         if axis is not None:
-> 1394             data = func(input_data, axis=axis, **kwargs)
   1395         else:
   1396             data = func(input_data, **kwargs)

/glade/work/abanihi/devel/pangeo/xarray/xarray/core/duck_array_ops.py in mean(array, axis, skipna, **kwargs)
    370         return _to_pytimedelta(mean_timedeltas, unit='us') + offset
    371     else:
--> 372         return _mean(array, axis=axis, skipna=skipna, **kwargs)
    373 
    374 

/glade/work/abanihi/devel/pangeo/xarray/xarray/core/duck_array_ops.py in f(values, axis, skipna, **kwargs)
    257 
    258         try:
--> 259             return func(values, axis=axis, **kwargs)
    260         except AttributeError:
    261             if isinstance(values, dask_array_type):

/glade/work/abanihi/devel/pangeo/xarray/xarray/core/nanops.py in nanmean(a, axis, dtype, out)
    158         return dask_array.nanmean(a, axis=axis, dtype=dtype)
    159 
--> 160     return np.nanmean(a, axis=axis, dtype=dtype)
    161 
    162 

/glade/work/abanihi/softwares/miniconda3/envs/xarray-tests/lib/python3.7/site-packages/numpy/core/overrides.py in public_api(*args, **kwargs)
    163             relevant_args = dispatcher(*args, **kwargs)
    164             return implement_array_function(
--> 165                 implementation, public_api, relevant_args, args, kwargs)
    166 
    167         if module is not None:

TypeError: no implementation found for 'numpy.nanmean' on types that implement __array_function__: [<class 'cupy.core.core.ndarray'>]
```

xarray/core/variable.py
```diff
@@ -207,6 +210,8 @@ def _as_array_or_item(data):
 
     TODO: remove this (replace with np.asarray) once these issues are fixed
     """
+    if hasattr(data, '__array_function__'):
```
shoyer: This function is called inside `Variable.values` for coercing into NumPy arrays.

Instead, I would suggest changing the conditional check inside the `data` property, so it return `self._data` directly for types that aren't just dask arrays:
https://github.com/pydata/xarray/blob/82fd8320b1be7a434effd247baf07004b66b802f/xarray/core/variable.py#L289-L294

xarray/core/variable.py
```diff
@@ -179,6 +179,9 @@ def as_compatible_data(data, fastpath=False):
         else:
             data = np.asarray(data)
 
+    if hasattr(data, '__array_function__'):
+        return data
```
shoyer: `numpy.ndarray` has a `__array_function__` attribute, so this won't quite do -- we still want to do the conversions on the lines below on NumPy arrays.

Also, it's at least possible to have a `__array_function__` attribute even if overrides with NEP-18 are not enabled (e.g., with NumPy version 1.16 without the environment variable set). We might want to raise an error in this case? Dask has a little utility for detecting this situation which might be worth borrowing: https://github.com/dask/dask/blob/c792364d1528c9c688a66c4d116e5f3b49776bb8/dask/array/utils.py#L352

shoyer:
> Even though it failed when I tried applying an operation on the dataset, this is still awesome!

Yes, it really is!

For this specific failure, we should think about adding an option for the default `skipna` value, or maybe making the semantics depend on the array type.

If someone is using xarray to wrap a computation oriented library like CuPy, they probably almost always want to set `skipna=False` (along with `join='exact'`). I don't think I've seen any deep library that has bothered to implement `nanmean`.

xarray/core/variable.py
nvictus: Sounds good. See 1b62f6d1. I added checks for ndarray-ness and NEP18.

xarray/core/duck_array_ops.py
```diff
@@ -76,6 +76,11 @@ def isnull(data):
         return _isnull(data)
     except TypeError:
         return np.zeros(data.shape, dtype=bool)
+    except Exception:
```
shoyer: I'm a little nervous about relying on an exception for checking for null data, because xarray uses this check all over the place, and I'm pretty sure that `_isnull` (which calls `pandas.isnull`) will first try to convert unknown array types into NumPy arrays.

For now, I think this is OK, but I'm going to try to write another helper function for doing this entirely in xarray.

shoyer: See https://github.com/pydata/xarray/pull/3132 for a cleaner fix here.

nvictus:
After writing more tests, turns out `sparse`-backed xarrays were generating strange coordinates. On closer inspection, this is because `sparse.COO` objects have their own `coords` attribute which stores the indices of its nonzeros.

With a serendipitous shape and density of a sparse array, there were the right number of `coords` such that they got converted into xarray `coords`:

```
>>> S = sparse.random((100,100)) 
>>> S.coords                                                                                                                                                                                                          
array([[ 0,  0,  3,  3,  4,  5,  6,  7,  7,  7,  9, 10, 11, 14, 14, 16,
        17, 18, 19, 19, 21, 21, 21, 21, 22, 23, 23, 24, 24, 25, 26, 28,
        29, 31, 35, 35, 36, 36, 38, 39, 41, 42, 42, 43, 44, 46, 46, 47,
        48, 48, 49, 49, 49, 49, 50, 52, 53, 54, 55, 56, 57, 57, 58, 60,
        60, 61, 61, 62, 64, 64, 68, 70, 72, 73, 76, 77, 78, 79, 79, 80,
        81, 83, 84, 85, 88, 89, 90, 90, 90, 92, 92, 93, 93, 94, 94, 96,
        97, 98, 98, 99],
       [14, 28, 58, 67, 66, 37, 50,  9, 66, 67,  6, 22, 44, 51, 64, 26,
        53, 91, 45, 81, 13, 25, 42, 86, 47, 22, 67, 77, 81, 22, 96, 18,
        23, 96, 34, 57,  6, 96, 91, 86, 75, 86, 88, 91, 30,  6, 65, 47,
        61, 74, 14, 73, 82, 83, 32, 42, 53, 68, 21, 38, 48, 50, 87,  8,
        89, 22, 57, 60, 92, 94, 19, 79, 38, 53, 32, 95, 69, 22, 46, 17,
        17, 86, 36,  7, 71, 35,  9, 58, 79, 22, 68, 10, 47, 48, 54, 72,
        24, 47, 63, 86]])
>>> xr.DataArray(S)
<xarray.DataArray (dim_0: 100, dim_1: 100)>
<COO: shape=(100, 100), dtype=float64, nnz=100, fill_value=0.0>
Coordinates:
  * dim_0    (dim_0) int64 0 0 3 3 4 5 6 7 7 7 ... 92 93 93 94 94 96 97 98 98 99
  * dim_1    (dim_1) int64 14 28 58 67 66 37 50 9 66 ... 47 48 54 72 24 47 63 86
```

A simple fix is to special-case `SparseArray`s in the `DataArray` [constructor](https://github.com/pydata/xarray/blob/master/xarray/core/dataarray.py#L262-L271) and not extract the sparse `coords` attribute in that case, but that would require importing `sparse`. 

Would it make sense to just assume that all non-DataArray NEP-18 compliant arrays do not contain an xarray-compliant `coords` attribute?

nvictus:
Hmm, looks like the `DataWithCoords` base type might come in handy here.

xarray/tests/test_nep18.py
```diff
@@ -0,0 +1,186 @@
...
+    x = xr.Variable('x', sparse.COO.from_numpy([1, 0, 2, 0, 0, 3, 0, 4]))
+    y = xr.Variable('y', sparse.COO.from_numpy([0, 4, 0, 3, 2, 0, 0, 1]))
+    data = sparse.random((8, 8), random_state=0, density=0.2)
+    A = xr.DataArray(data, dims=['x', 'y'])
+    A.coords['xx'] = x
+    A.coords['yy'] = y
+    assert isinstance(func(A).data, sparse.SparseArray)
+
+
+class TestSparseVariable:
```
max-sixty: An aside (don't let this slow you down): these are tests that we could run across any eligible array, including the current numpy & dask arrays

You could pull tests from the current dask tests if that would be a shortcut to building up this test suite from scratch

We could do that by inheriting this class, or having `var` as a fixture

shoyer:
> Would it make sense to just assume that all non-DataArray NEP-18 compliant arrays do not contain an xarray-compliant `coords` attribute?

Yes, let's switch:
```
coords = getattr(data, 'coords', None) 
```
to
```
if isinstance(data, DataArray):
    coords = data.coords
```

xarray/core/dataarray.py
```diff
@@ -259,7 +259,10 @@ def __init__(
         else:
             # try to fill in arguments from data if they weren't supplied
             if coords is None:
-                coords = getattr(data, 'coords', None)
+
+                if isinstance(data, DataWithCoords):
```
shoyer: `DataWithCoords` is only used for Dataset and DataArray. But putting a `Dataset` into the `DataArray` constructor doesn't do anything useful, so let's just stick with `DataArray`: 
```suggestion
                if isinstance(data, DataArray):
                    coords = data.coords
```

Example of what passing in a Dataset currently gives you:
```
In [25]: xr.DataArray(xr.Dataset(coords={'x': 1}))
Out[25]:
<xarray.DataArray ()>
array(<bound method Mapping.values of <xarray.Dataset>
Dimensions:  ()
Coordinates:
    x        int64 1
Data variables:
    *empty*>, dtype=object)
Coordinates:
    x        int64 1
```
Yes, `coords` is copied, but the `data` attribute is the bound `values()` method from the Dataset object!

xarray/core/dataarray.py
```diff
@@ -260,7 +260,8 @@ def __init__(
             # try to fill in arguments from data if they weren't supplied
             if coords is None:
 
-                if isinstance(data, DataWithCoords):
+                if isinstance(data, DataArray):
+                    coords = data.coords
                     coords = getattr(data, 'coords', None)
```
shoyer: Sorry, this line should be removed -- suggestions in GitHub comments can't replace multiple lines.

nvictus: Yeah, Github's diffs give often me a headache. This is what we want, correct?
```python
            if coords is None:

                if isinstance(data, DataArray):
                    coords = data.coords
                if isinstance(data, pd.Series):
                    coords = [data.index]
                elif isinstance(data, pd.DataFrame):
                    coords = [data.index, data.columns]
                elif isinstance(data, (pd.Index, IndexVariable)):
                    coords = [data]
                elif isinstance(data, pdcompat.Panel):
                    coords = [data.items, data.major_axis, data.minor_axis]
```

shoyer: yes... probably could convert `if isinstance(data, pd.Series):` to use an `elif`

xarray/tests/test_nep18.py
nvictus: Thanks @max-sixty. I did copy them over from the dask tests. I'm still figuring out which ones are transferable to this use case since I'm not a regular xarray user myself (yet) :)

max-sixty: Great! LMK if I can be helpful at all. Excited about the PR!

mrocklin:
Checking in here.  This was a fun project during SciPy Sprints that both showed a lot of potential and generated a lot of excitement.  But of course as we all returned home other things came up and this has lingered for a while.

How can we best preserve this work?  Two specific questions:

1.  @nvictus can you summarize, from your perspective, what still needs to be done here?  If you aren't likely to have time to finish this up (which would not be surprising) where should someone else start if they wanted to push this forward?
2.  Xarray devs, are there parts of this PR that could go in quickly, even without the full implementation, just so that this doesn't grow stale?

nvictus:
Thanks for bumping this @mrocklin! I've put in some extra work on my free time, which hasn't been pushed yet. I'll try to write up a summary of my findings today. Briefly though, it seems like the two limiting factors for NEP18 duck array support are:

1. Operations which ultimately coerce duck arrays to ndarrays (e.g. via `np.asarray`). Many, but maybe not all of these operations can fixed to dispatch to the duck array's implementation. But that leads to:

2. Operations not supported by the duck type. This happens in a few cases with pydata/sparse, and would have to be solved upstream, unless it's a special case where it might be okay to coerce. e.g. what happens with binary operations that mix array types?

I think NEP18-backed xarray structures can be supported in principle, but it won't prevent some operations from simply failing in some contexts. So maybe xarray will need to define a minimum required implementation subset of the array API for duck arrays.

shoyer:
> 2\. Operations not supported by the duck type. This happens in a few cases with pydata/sparse, and would have to be solved upstream, unless it's a special case where it might be okay to coerce. e.g. what happens with binary operations that mix array types?

This is totally fine for now, as long as there are clear errors when attempting to do an unsupported operation. We can write unit tests with expected failures, which should provide a clear roadmap for things to fix upstream in sparse.

We could attempt to define a minimum required implementation, but in practice I suspect this will be hard to nail down definitively. The ultimate determinant of what works will be xarray's implementation.

xarray/core/duck_array_ops.py
nvictus: merged into this PR

shoyer: #3132 isn't quite working yet, as evidenced by the failing tests. It would probably be best to leave that for a follow-up :)

nvictus:
So, tests are passing now and I've documented the expected failures on sparse arrays. :)

As mentioned before, most fall into the categories of (1) implicit coercion to dense and (2) missing operations on sparse arrays.

Turns out a lot of the implicit coercion is due to the use of routines from `bottleneck`. A few other cases of using `np.asarray` remain. Sometimes this is through the use of the `.values` attribute. At the moment, the behavior of Variables and DataArrays is such that `.data` provides the duck array and `.values` coerces to numpy, following the original behavior for dask arrays -- which made me realize, we never asked if this behavior is desired in general? 

I also modified `Variable.load()` to be a no-op for duck arrays (unless it is a dask array), so now `compute()` works. To make equality comparisons work between dense and sparse, I modified `as_like_arrays()` to coerce all arguments to **sparse** if at least one is sparse. As a side-effect, `a.identical(b)` can return True if one is dense and the other sparse. Again, here we're just copying existing behavior when the duck array is a dask array, but this may seem questionable to extend to other duck arrays.

If `bottleneck` is turned off, most remaining issues are due to missing implementations in `sparse`. Some of these should be easy to add: PR pydata/sparse/pull/261 already exposes the `result_type` function.

xarray/tests/test_duck_array_ops.py
```diff
@@ -178,14 +178,18 @@ def test_wrong_shape(self):
         assert not array_notnull_equiv(a, b)
 
     @pytest.mark.parametrize("val1, val2, val3, null", [
-        (1, 2, 3, None),
+        (np.datetime64('2000'),
```
shoyer: This part looks left over from https://github.com/pydata/xarray/pull/3132. You could remove it, or since I just merged that PR, you could just merge in the latest from master.

xarray/tests/test_sparse.py
```diff
@@ -0,0 +1,647 @@
...
+
+@pytest.mark.parametrize("func,sparse_output", [
+    (do('all'), False),
+    (do('any'), False),
+    (do('assign_attrs', {'foo': 'bar'}), True),
+    (do('assign_coords', x=make_xrarray({'x': 10}).x + 1), True),
+    (do('astype', int), True),
+    (do('broadcast_equals', make_xrarray({'x': 10, 'y': 5})), False),
+    (do('clip', min=0, max=1), True),
+    # (do('close'), False),
```
shoyer: I'm guessing the commented out cases don't work yet? :)

xarray/tests/test_sparse.py
```diff
@@ -0,0 +1,647 @@
...
+        s = COO.from_numpy(a)
+        cond = s > 3
+        xr.DataArray(s).where(cond)
+
+        x = xr.DataArray(s)
+        cond = x > 3
+        x.where(cond)
+
+
+# Do we even want this?
```
shoyer: Right now, we load every "dimension coordinate" into memory as a `pandas.Index`. But for other coordinates, sure why not?

shoyer:
> At the moment, the behavior of Variables and DataArrays is such that `.data` provides the duck array and `.values` coerces to numpy, following the original behavior for dask arrays -- which made me realize, we never asked if this behavior is desired in general?

I think the right behavior is probably for `.values` to be implemented by calling `np.asarray()` on `.data`. That means it should raise on sparse arrays.

xarray/tests/test_sparse.py
nvictus: Oops, forgot to finish those! :) 
I've added tests for all except those I couldn't figure out how to use like `swap_dims`, `sel_points` and `isel_points`.

xarray/tests/test_sparse.py
```diff
@@ -0,0 +1,694 @@
...
+
+    def test_unary_op(self):
+        assert_sparse_eq(-self.var.data, -self.data)
+        assert_sparse_eq(abs(self.var).data, abs(self.data))
+        assert_sparse_eq(self.var.round().data, self.data.round())
+
+    def test_binary_op(self):
+        assert_sparse_eq((2 * self.var).data, 2 * self.data)
+        assert_sparse_eq((self.var + self.var).data, self.data + self.data)
+        # assert_eq((self.var[0] + self.var).data, self.data[0] + self.data)
```
shoyer: remove?

shoyer:
@nvictus we are good to go ahead and merge, and do follow-ups in other PRs?

nvictus:
Sounds good!

dcherian:
Thanks a lot, @nvictus

mrocklin:
Woot!  Thanks @nvictus !

ci/requirements/py37.yml
```diff
@@ -21,6 +21,7 @@ dependencies:
   - pip
   - scipy
   - seaborn
+  - sparse
   - toolz
```
crusaderky: Why did you exclude py36?

max-sixty:
Thanks @nvictus !

Issue #2078, opened 2018-04-24
apply_ufunc should include variable names in error messages


shoyer:
This would make it easier to debug issues with dimensions.

For example, in [this example](https://stackoverflow.com/questions/49959449/how-to-use-xr-apply-ufunc-with-changing-dimensions) from StackOverflow, the error message was "ValueError: operand to apply_ufunc has required core dimensions ['time', 'lat', 'lon'], but some of these are missing on the input variable:  ['lat', 'lon']".

A better error message would be: "ValueError: operand to apply_ufunc has required core dimensions ['time', 'lat', 'lon'], but some of these are missing on input variable "status":  ['lat', 'lon']"

rdrussotto:
I am working on this. The error listed here is thrown by the "broadcast_compat_data" function, which is called by "apply_variable_ufunc", which is in turn called by "apply_ufunc". There are other similar error messages that can happen where a variable name would also be helpful.

rdrussotto:
This is much harder than I thought. The function that throws this error doesn't have access to the variable name data, since it takes Variable objects as input which don't contain name information. apply_ufunc and its dependent functions are designed to take in a Dataset, DataArray, Variable, GroupBy or other object that might contain one or many variables, strip out the individual variable objects and run the computation on each of them. Including the variable name in the error message would require separately specifying the variable name as an input argument to "broadcast_compat_data" and "apply_variable_ufunc" and refactoring the code that calls these functions, which seems like adding too much complexity just for an error message. For now I think I'm going to change "the input variable" to "an input variable", which is at least a bit clearer.

shoyer:
> Including the variable name in the error message would require separately specifying the variable name as an input argument to "broadcast_compat_data" and "apply_variable_ufunc" and refactoring the code that calls these functions, which seems like adding too much complexity just for an error message.

I agree that this is more complex, but it still might be worth doing. Good error messages are important!

We do internally in a few other places already.

PR #3119, opened 2019-07-13
Clarify apply_ufunc error message (Issue #2078)

rdrussotto:

Changed an error message thrown by a dependent function of apply_ufunc to specify that "an input variable" rather than "the input variable" did not have the correct dimensions. This error occurs when apply_ufunc is used on a Dataset with multiple variables, at least one of which does not have correct dimensions matching the other input arguments. Wanted to add variable name to error message to specify which variable is causing the error, but this isn't possible without significantly refactoring the code.


xarray/core/computation.py
```diff
@@ -482,9 +482,9 @@ def broadcast_compat_data(variable, broadcast_dims, core_dims):
     missing_core_dims = [d for d in core_dims if d not in set_old_dims]
     if missing_core_dims:
         raise ValueError(
-            'operand to apply_ufunc has required core dimensions %r, but '
-            'some of these are missing on the input variable:  %r'
-            % (list(core_dims), missing_core_dims))
+            'operand to apply_ufunc has required core dimensions {}, but '
+            'some of these are missing on an input variable: {}'.format(
```
dcherian:
```suggestion
            'some of these dimensions are absent on an input variable: {}'.format(
```

dcherian:
Thanks @rdrussotto. Do you like the suggested change? If so, commit that and I'll merge.

dcherian:
The test failure should be fixed by merging master.

rdrussotto:
I agree with the suggested change, just committed it.

max-sixty:
Thanks @rdrussotto - there are now a couple of small merge conflicts - could you resolve? Then we can merge

rdrussotto:
I tried to resolve the merge conflicts but given the tests don't pass I might have messed something up...

dcherian:
You just need to run `black xarray/core/computation.py`

dcherian:
Yay! Thanks @rdrussotto

PR #3121, opened 2019-07-13
Allow other tutorial filename extensions

### LINKED ISSUE #3118: Allow GRIB data as an example dataset
Need to add support for other files than netCDF in `load_dataset`, namely for GRIB data, which a PR will be coming to fix shortly.

rabernat:
- [x] Closes #3118
 - [ ] Tests added

Together with https://github.com/pydata/xarray-data/pull/15, this allows us to generalize out tutorial datasets to non netCDF files. But it is backwards compatible--if there is no file suffix, it will append `.nc`.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -25,6 +25,9 @@ dependencies:
   - sphinx_rtd_theme=0.4.2
   - mock=2.0.0
   - nbsphinx=0.4.2
+<<<<<<< HEAD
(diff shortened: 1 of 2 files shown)


rabernat:
I will be going offline soon when I board a plane. If any other @pydata/xarray devs want to merge this, it might unblock some ongoing work.

shoyer:
thanks!

PR #3123, opened 2019-07-14
to_zarr(append_dim='dim0') doesn't need mode='a'

### LINKED ISSUE #3100: Append to zarr redundant with mode='a'
When appending to a zarr store, we need to set `append_dim=dim_name` but also `mode='a'`. Any reason to also specify the writing mode? I think it should automatically be set to `'a'` if `append_dim is not None`.

davidbrochart:
- [x] Closes #3100
 - [x] Tests updated

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -1118,7 +1118,8 @@ def to_zarr(dataset, store=None, mode='w-', synchronizer=None, group=None,
     _validate_dataset_names(dataset)
     _validate_attrs(dataset)
 
-    if mode == "a":
+    if (mode == 'a') or (append_dim is not None):
+        mode = 'a'
         _validate_datatypes_for_zarr_append(dataset)
         _validate_append_dim_and_encoding(dataset, store, append_dim,
                                           group=group,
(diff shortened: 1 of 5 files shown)


xarray/backends/api.py
```diff
@@ -1118,7 +1118,8 @@ def to_zarr(dataset, store=None, mode='w-', synchronizer=None, group=None,
     _validate_dataset_names(dataset)
     _validate_attrs(dataset)
 
-    if mode == "a":
+    if (mode == 'a') or (append_dim is not None):
+        mode = 'a'
```
shoyer: Normally the preferred way to handle cases like this (intellligent default argument values) is to make them default to None. Thus, if you write `mode=‘w’` and `append_dim=‘time’`, you would get an error message rather than having your choice of mode overwritten.

davidbrochart: Thanks @shoyer, I've made mode default to None, which allows for checking inconsistency if append_dim is set and mode!='a' (I added a test for this). If mode and append_dim are not set, mode is internally set to `w-`, which is equivalent to the previous behavior.

davidbrochart:
@shoyer do you think this PR can be merged?

shoyer:
thanks!

PR #3126, opened 2019-07-14
Return immutable view of Pandas indexes

### LINKED ISSUE #2949: xr.DataArray.to_series returns a (mutable) view
```python
import numpy as np
import pandas as pd
import xarray as xr

dates = pd.date_range('01-Jan-2019', '01-Jan-2020', name='forward_date')[:-1]
s = pd.Series(np.random.randn(dates.size), dates)
array = xr.DataArray.from_series(s)
```
```python
In [52]: array
Out[52]: 
<xarray.DataArray (forward_date: 365)>
array([ 0.511149, -1.009054, -0.098397, ...,  0.496596,  1.435346,  0.21206 ])
Coordinates:
  * forward_date  (forward_date) datetime64[ns] 2019-01-01 ... 2019-12-31

In [53]: array.to_series().index.name
Out[53]: 'forward_date'

In [54]: array.to_series().index.name = 'peristent!!!'

In [55]: array.to_series().index.name
Out[55]: 'peristent!!!'

In [56]: array.get_index('forward_date')
Out[56]: 
DatetimeIndex(['2019-01-01', '2019-01-02', '2019-01-03', '2019-01-04',
               '2019-01-05', '2019-01-06', '2019-01-07', '2019-01-08',
               '2019-01-09', '2019-01-10',
               ...
               '2019-12-22', '2019-12-23', '2019-12-24', '2019-12-25',
               '2019-12-26', '2019-12-27', '2019-12-28', '2019-12-29',
               '2019-12-30', '2019-12-31'],
              dtype='datetime64[ns]', name='peristent!!!', length=365, freq='D')
```

#### Problem description
IMHO Setting `index.name` on the returned series shouldn't affect the original `DataArray` - i.e. every call to `array.to_series().index.name` should return `forward_date` irrespective of any mutating operations performed on the returned series.

<details>

```python
In [48]: xr.show_versions()

  File "<ipython-input-48-6f391305f2fe>", line 1, in <module>
    xr.show_versions()

  File "C:\Users\dhirschf\envs\dev\lib\site-packages\xarray\util\print_versions.py", line 84, in show_versions
    sys_info.extend(netcdf_and_hdf5_versions())

  File "C:\Users\dhirschf\envs\dev\lib\site-packages\xarray\util\print_versions.py", line 75, in netcdf_and_hdf5_versions
    libhdf5_version = h5py.__hdf5libversion__

AttributeError: module 'h5py' has no attribute '__hdf5libversion__'
```
</details>

gwgundersen:
- This closes #2949. We return a shallow copy of Pandas indexes to ensure immutability.
- Added `test_index_immutability` in `tests/test_dataarray.py`.
- Amended `whats-new.rst`.


max-sixty:
I don't quite understand why the previous constructor was returning a mutable view, but the tests look great.

@gwgundersen do you know if we should instead be changing the `to_index` method which the `to_series` calls?

Any other comments from anyone?

gwgundersen:
@max-sixty, `name` is just a mutable property on a Pandas index:

```python
>>> dates = pd.date_range('01-Jan-2019', '31-Jan-2019', name='mutable?')
>>> series = pd.Series(np.random.randn(dates.size), dates)
>>> series.index.name
'mutable?'
>>> series.index.name = 'yes'
>>> series.index.name
'yes'
```

But you're right that `to_series` is the wrong place. It looks like `to_series` calls `to_index`, which calls [`get_index`](https://github.com/pydata/xarray/blob/master/xarray/core/common.py#L288), which is inherited by other classes such as `Dataset`. For example, the bug persists with `Dataset` despite my fix:

```python
>>> dates = pd.date_range('01-Jan-2019', '31-Jan-2019', name='mutable?')
>>> series = pd.Series(np.random.randn(dates.size), dates)
>>> dataset = xr.Dataset({'foo': series})
>>> dataset.indexes['mutable?'].name = 'yes'
>>> dataset.indexes['mutable?'].name
'yes'
```

Does anyone know if `get_index` is the best place for this? Maybe the tests should be somewhere more generic as well.

max-sixty:
>  It looks like to_series calls to_index, which calls get_index,

Thanks for finding that. That sounds right - if we make these changes there then I think we've solved the problem at its root and we should fix all these cases?

gwgundersen:
Locally, I've moved my changes from `DataArray.to_series` to `DataWithCoords.get_index`, and my tests still pass. After work, I'll write some more tests, e.g. for for `DataArray.to_index`, and push.

@shoyer, it seems like this issue happens anywhere the user can modify a Pandas index. For example, my fix the Dataset bug in [my comment yesterday](https://github.com/pydata/xarray/pull/3126#issuecomment-511506569). Do we want all views of indexes to be immutable or do we trust the user to not do stuff like this:

```python
>>> dataset.indexes['mutable?'].name = 'yes'
>>> dataset.indexes['mutable?'].name
'yes'
```

max-sixty:
> Do we want all views of indexes to be immutable or do we trust the user to not do stuff like this:

Is there a reasonable way of making that immutable? Given that the index allows its name to be mutated, I'm not sure.

(we can also separate that issue from this change, which is already 👍 )

gwgundersen:
Okay, a couple changes to the original PR: we now perform a shallow copy of the returned index in `DataWithCoords.get_index` and `Indexes.__getitem__`. The latter is needed because both `Dataset.indexes` and `DataArray.indexes` returns an `Indexes` object. Finally, I added [a bunch of tests for various index accessors](https://github.com/pydata/xarray/pull/3126/files#diff-02d824ae7c673c0c4bf2f30dff024753R3055).

This seems pretty robust:

```python
>>> import pandas as pd
>>> import xarray as xr
>>> index = pd.Index(list('abcd'), name='foo')
>>> series = pd.Series(range(4), index)
>>> array = xr.DataArray.from_series(series)
>>> array.get_index('foo').name = 'bar'
>>> array.get_index('foo').name
'foo'
>>> array['foo'].to_index().name = 'bar'
>>> array['foo'].to_index().name
'foo'
>>> array.to_index().name = 'bar'
>>> array.to_index().name
'foo'
>>> array.indexes['foo'].name = 'bar'
>>> array.indexes['foo'].name
'foo'
```

Since a `Dataset` is just a collection of `DataArray` objects, the behavior propagates nicely, e.g.

```python
>>> dataset = xr.Dataset({'myvar': array})
>>> dataset['myvar'].to_index().name = 'bar'
>>> dataset['myvar'].to_index().name 
'foo'
>>> dataset.indexes['foo'].name = 'bar'
>>> dataset.indexes['foo'].name
'foo'
```

max-sixty:
Do you have an accidental binary file in the commit? https://github.com/pydata/xarray/pull/3126/files#diff-337006a573a2971594e829114137f91b

xarray/core/indexes.py
```diff
@@ -30,7 +30,10 @@ def __contains__(self, key):
         return key in self._indexes
 
     def __getitem__(self, key):
-        return self._indexes[key]
+        index = self._indexes[key]
+        # Issue 2949: Pandas indexes have mutable names, but we want to return
+        # immutable views.
+        return index.copy(deep=False)
```
max-sixty: Does anyone know if we may face performance issues by making lots of views? (we're never copying the underlying data, but we'll have a `Series` instance on each call)

xarray/tests/test_dataarray.py
```diff
@@ -3024,6 +3024,47 @@ def test_to_pandas(self):
...
+        # Utility function that allows us to consistently many different paths
+        # to accessing the index.
+        def test_index(get_index):
+            # Test `index.data`
+            with pytest.raises(Exception):
+                get_index().data = np.random.randn(10)
+
+            # Test `index.dtype`
+            with pytest.raises(Exception):
+                get_index().dtype = np.float64
```
shoyer: These first two tests (for `index.data` and `index.dtype`) looks like they're testing implementation details of `pandas.Index`, rather than how xarray wraps it. I think we should probably leave this sort of testing to pandas itself.

xarray/tests/test_dataarray.py
```diff
@@ -3024,6 +3024,47 @@ def test_to_pandas(self):
...
+
+            # Test `index.copy`
+            assert callable(get_index().copy)
+            get_index().copy = 'mutable?'
```
shoyer: I'm not sure we should be testing overriding a method. I don't think pandas would consider the ability to override `Index.copy` a guarantee of their public API, so this might break in the future, e.g., if pandas rewrote `Index` to be a CPython extension class rather than a class written in Python.

xarray/tests/test_dataarray.py
```diff
@@ -3024,6 +3024,47 @@ def test_to_pandas(self):
...
+            assert callable(get_index().copy)
+
+            # Test `index.name`
+            assert get_index().name == 'persistent'
+            get_index().name = 'mutable?'
+            assert get_index().name == 'persistent'
```
shoyer: This test looks very sensible 👍

xarray/core/indexes.py
shoyer: Shallow copies should be pretty cheap. But the real question is whether pandas copies the underlying hash-table. If not, this could cause the performance of repeated indexing to fall off a cliff.

shoyer: Unfortunately, it looks like the hash-table *does not* get copied by `copy(deep=False)`, even though it probably could. Compare:
```
In [11]: import numpy as np

In [12]: import pandas as pd

In [13]: x = np.random.RandomState(0).randint(int(1e6), size=int(1e7))

In [14]: index = pd.Index(x)

# hash-table is not built yet
In [15]: %time index.get_loc(10)
CPU times: user 475 ms, sys: 9.79 ms, total: 485 ms
Wall time: 486 ms
Out[15]: array([False, False, False, ..., False, False, False])

# hash table is cached, so look-up is fast
In [16]: %time index.get_loc(10)
CPU times: user 10.8 ms, sys: 4.75 ms, total: 15.6 ms
Wall time: 14.1 ms
Out[16]: array([False, False, False, ..., False, False, False])

# copied index is slow -- it's rebuilding the hash table!
In [17]: %time index.copy(deep=False).get_loc(10)
CPU times: user 461 ms, sys: 12.8 ms, total: 474 ms
Wall time: 474 ms
Out[17]: array([False, False, False, ..., False, False, False])

# a shallow copy itself is pretty fast
In [18]: %timeit index.copy(deep=False)
16 µs ± 456 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

gwgundersen: Ah, good catch. What does this mean for us? Sometimes returning an immutable index (e.g. in `to_series`) and sometimes not (e.g. in `indexes`) seems confusing to me.

xarray/tests/test_dataarray.py
gwgundersen: My rationale was: every test uses a `get_index()` function, so every test tests how Xarray wraps Pandas. Given that we're changing how Pandas treats `Index.name`, it's not a given that we're not changing how Pandas treats `Index.dtype`.

xarray/tests/test_dataarray.py
gwgundersen: I thinking was the same as [here](https://github.com/pydata/xarray/pull/3126#discussion_r304272667). If you'd rather limit the scope of this PR to guarantee that only `name` behaves as expected, I can remove these.

xarray/core/indexes.py
gwgundersen: @shoyer: any more thoughts on this? Happy to see this through or just close.

shoyer: The situation here is a little disappointing, but unfortunately I don't think this fix is workable. I don't sacrificing performance here is worthwhile. It would largely defeat the purpose of indexing in the first place.

Issue #2910, opened 2019-04-20
Keyword argument support for drop()


shoyer:
Currently, to drop labels along an existing dimension, you need to write something like: `ds.drop(['a', 'b'], dim='x)`.

It would be nice if keyword arguments were supported, e.g., `ds.drop(x=['a', 'b'])`. This would make `drop()` more symmetric with `sel()`.

gwgundersen:
Created PR https://github.com/pydata/xarray/pull/3128 to add support for this feature.

PR #3128, opened 2019-07-14
Support keyword API for `Dataset.drop`

gwgundersen:
- Closes https://github.com/pydata/xarray/issues/2910.
- In `tests/test_dataset`, added `test_drop_labels_by_keyword`
- Added a utility function `is_list_like`.
- Documented in `whats-new.rst` and confirmed API docs built correctly.


shoyer:
Awesome, thanks for working on this @gwgundersen !

In the long term, we might want to deprecate the current function signature, but I'm happy to leave that for later.

xarray/core/dataset.py
```diff
@@ -3150,11 +3150,25 @@ def drop(self, labels, dim=None, *, errors='raise'):
...
             dataset are dropped and no error is raised.
+        **dim_kwargs : {dim: label, ...}, optional
+            The keyword arguments form of ``dim`` and ``labels``.
+            Either ``dim`` and ``labels`` or ``dim_kwargs`` must be provided.
 
         Returns
         -------
         dropped : Dataset
         """
+        if len(dim_kwargs) > 0:
```
shoyer: Just using a dict directly in a condition is the preferred way to check if it's empty:
```suggestion
        if dim_kwargs:
```

xarray/core/dataset.py
```diff
@@ -3150,11 +3150,25 @@ def drop(self, labels, dim=None, *, errors='raise'):
             any of the variable or index labels passed are not
             in the dataset. If 'ignore', any given labels that are in the
             dataset are dropped and no error is raised.
+        **dim_kwargs : {dim: label, ...}, optional
```
shoyer: Everywhere else in xarray where we support keyword arguments, we try to use the `xarray.core.utils. either_dict_or_kwargs()` helper function, to allow writing either `dataset.method({k: v})` or `dataset.method(k=v)` inter-changeably.

It would be nice to do the same here, if possible, though it get a little bit tricky due to the need to support the original signature. I think something like this would work:
```python
if dim_kwargs or utils.is_dict_like(labels):
    dim_kwargs = utils.either_dict_or_kwargs(labels, dim_kwargs, 'drop')
    if dim is not None:
        raise ValueError(...)
    ...
```

xarray/core/dataset.py
```diff
@@ -3150,11 +3150,25 @@ def drop(self, labels, dim=None, *, errors='raise'):
...
+            The keyword arguments form of ``dim`` and ``labels``.
+            Either ``dim`` and ``labels`` or ``dim_kwargs`` must be provided.
 
         Returns
         -------
         dropped : Dataset
```
shoyer: I would be awesome if you're up for adding a few examples here inline in the docstring.

xarray/core/dataset.py
```diff
@@ -3150,11 +3150,25 @@ def drop(self, labels, dim=None, *, errors='raise'):
...
         """
+        if len(dim_kwargs) > 0:
+            if labels is not None or dim is not None:
+                raise ValueError('cannot specify both `dim` and `labels` along'
+                                 'with keyword arguments')
+            if len(dim_kwargs) > 1:
+                raise ValueError('cannot specify multiple dimensions')
+            dim = list(dim_kwargs.keys())[0]
+            labels = list(dim_kwargs.values())
+            if utils.is_list_like(labels[0]):
```
shoyer: I don't think we need this check -- we already handle both scalars and lists in the logic below.

gwgundersen: Since we convert `dim_kwargs`'s values to a list, if the user passes in a list, we get a list with a single element which is also a list. The logic below won't catch this list-in-list scenario. Happy to change this if you know a simpler way to handle `dict_values`. My problem was

```python
>>> x = {'foo': 1}
>>> x.values()
dict_values([1])
>>> x.values()[0] # TypeError
```

which is why I do `list(x.values())` instead.

xarray/core/dataset.py
gwgundersen: Isn't `labels` supposed to be `list`-like, not `dict`-like? That is why I did not try to use `either_dict_or_kwargs`.

xarray/core/dataset.py
gwgundersen: Good idea. Done.

xarray/core/dataset.py
shoyer: Ah, yes. I think the cleanest way to do this would be with tuple unpacking, e.g.,
```
((dim, labels),) = dim_kwargs.items()
```
or
```
(dim, labels) = list(dim_kwargs.items())[0]
```

xarray/core/dataset.py
shoyer: I'm proposing allowing labels to be *either* `list`-like or `dict`-like.

Though perhaps that's more confusing...

xarray/core/dataset.py
gwgundersen: That works great. Thanks.

xarray/core/dataset.py
gwgundersen: Ah, I see. This means we would also want to support the user dropping labels along multiple dimensions? The docstring would be complicated. You also mentioned above that "we might want to deprecate the current function signature", but the `dict`-like approach does not allow `dim=None`. This means we would need to support four different ways of calling the function:

- `labels` is `list`-like but no `dim`: drop variables.
- `labels` is `list`-like and `dim`: drop labels along `dim`.
- `labels` is a `dict`: drop values for each `dim` key along that dimension.
- `labels_kwargs` are used: drop values for each `dim` keyword argument.

As a first-time contributor, I don't think this is my call, but I'd vote for deprecating the "drop variables" approach in favor of `dict`-like long-term to be more consistent with other functions.

Maybe we add a new function for dropping variables or an argument `variables`?

dcherian: In the long term, can we deprecate 
> labels is list-like and dim: drop labels along dim.

and then have 3 modes of operation? i.e. when `labels` is list-like, they must be variable names.

gwgundersen: Do you think that's more clear for a user than `drop(..., variables=True)`? IMO explicit is better than implicit, and conditioning on the type of the input seems confusing, especially when that input type was previously valid. In the worst case, this would fail silently by dropping variables when a user meant dimensions.

gwgundersen: Sorry, wrote a confused reply. Yeah, I like your idea. Shall I implement?

shoyer: > In the long term, can we deprecate
> 
> > labels is list-like and dim: drop labels along dim.
> 
> and then have 3 modes of operation? i.e. when `labels` is list-like, they must be variable names.

I think this would be make sense. We'll need to support this through a deprecation cycle, but eventually we can remove it.

max-sixty:
@gwgundersen I see you trying to merge re the recent black changes. Here are the instructions for making this (relatively) seamless: https://github.com/pydata/xarray/blob/d3f1ced1aa84cb0a0f3e47e7ee67492e203fcce5/doc/contributing.rst#code-formatting

gwgundersen:
@max-sixty, thanks! It's useful to be able to do in advance.

After merging from master, I get this error:

```
flake8
./xarray/core/dataset.py:3466:5: F811 redefinition of unused 'drop' from line 3460
```

I assume this is due to https://github.com/pydata/xarray/pull/3177 cc @crusaderky. Is there a way to avoid this?

max-sixty:
@gwgundersen Could you force-push your latest code? 

If your current latest code is pushed, you may have to `git reset --hard` and start again - unfortunately it looks like your PR is proposing ` +24,439 −17,647 ` changes

gwgundersen:
Sorry about that. I'm not sure what happened. I've reverted branch and cherry picked my changes. The issue with `drop` being a duplicated name still stands.

xarray/core/dataset.py
```diff
@@ -3463,7 +3463,7 @@ def drop(
     ) -> "Dataset":
         ...
 
-    def drop(self, labels, dim=None, *, errors="raise"):  # noqa: F811
+    def drop(self, labels=None, dim=None, *, errors="raise", **labels_kwargs):
```
```suggestion
    def drop(self, labels=None, dim=None, *, errors="raise", **labels_kwargs):  # noqa: F811
```

max-sixty:
Thanks @gwgundersen - good idea re the cherry pick for more discrete changes. 

Still getting a test failure though, check it out and lmk if you need any help

xarray/core/dataset.py
max-sixty: Hopefully this doesn't break black - may need to ensure the `noqa` is in the right place if black reformats

xarray/core/dataset.py
gwgundersen: Great. I followed another bit of code that just used `warnings.warn` with a `DeprecationWarning`.

xarray/core/dataset.py
gwgundersen: Thanks. I had to move this comment slightly after using `black`, but it worked.

gwgundersen:
Now the docs don't build. When I try to build locally, I get a vague error:

```bash
  File "/Users/gwg/miniconda3/envs/xarray-docs/lib/python3.7/site-packages/IPython/sphinxext/ipython_directive.py", line 583, in process_input
    raise RuntimeError('Non Expected warning in `{}` line {}'.format(filename, lineno))
RuntimeError: Non Expected warning in `/Users/gwg/projects/xarray/doc/indexing.rst` line 241
```

Do you see anything wrong with this?

```rest
The :py:meth:`~xarray.Dataset.drop` method returns a new object with the listed
index labels along a dimension dropped:

.. ipython:: python

    ds.drop(['IN', 'IL'], dim='space')
<-- line 241
``drop`` is both a ``Dataset`` and ``DataArray`` method.

Use :py:meth:`~xarray.Dataset.drop_dims` to drop a full dimension from a Dataset.
Any variables with these dimensions are also dropped:
```

dcherian:
It looks like you've added a warming for this case. So you'll need to specify :okwarning: .

max-sixty:
I see the docs timing out after 60 min. I also see that message, but is that causing the failure?

I restarted the docs build to see whether that might help

gwgundersen:
Sorry for the back-and-forth. CI tests and checks are all passing, and this is ready for another review.

xarray/core/dataset.py
```diff
@@ -3479,34 +3482,75 @@ def drop(self, labels, dim=None, *, errors="raise"):  # noqa: F811
...
         """
         if errors not in ["raise", "ignore"]:
             raise ValueError('errors must be either "raise" or "ignore"')
 
-        if dim is None:
+        labels_are_coords = isinstance(labels, DataArrayCoordinates)
+        if labels_kwargs or (utils.is_dict_like(labels) and not labels_are_coords):
+            labels_kwargs = utils.either_dict_or_kwargs(labels, labels_kwargs, "drop")
+            if dim is not None:
+                raise ValueError("cannot specify dim amd dict-like " "arguments.")
```
```suggestion
                raise ValueError("cannot specify dim and dict-like arguments.")
```

max-sixty:
@gwgundersen could we add a whatsnew too?

gwgundersen:
@max-sixty, done.

max-sixty:
Great @gwgundersen !

xarray/core/dataset.py
```diff
@@ -3479,34 +3482,75 @@ def drop(self, labels, dim=None, *, errors="raise"):  # noqa: F811
...
+        Dimensions without coordinates: x
+        Data variables:
+            A        (x, y) float64 -0.3944 -1.418 1.423 -1.041
         """
         if errors not in ["raise", "ignore"]:
             raise ValueError('errors must be either "raise" or "ignore"')
 
-        if dim is None:
+        labels_are_coords = isinstance(labels, DataArrayCoordinates)
+        if labels_kwargs or (utils.is_dict_like(labels) and not labels_are_coords):
```
shoyer: I wonder if it would be more predictable to special case the `dict` type rather than `not labels_are_coords`, which is  just one of many internal xarray mapping types, e.g.,
```suggestion
        if labels_kwargs or isinstance(labels, dict):
```

More generally, I get that writing `array.drop(array.coords)` is convenient, but in the long term it seems like a bad idea to treat different mapping types differently. This suggests that such uses should be deprecated. Either `array.drop_vars(array.coords)` or `array.drop(list(array.coords))` would be less ambiguous.

doc/indexing.rst
```diff
@@ -236,6 +236,7 @@ The :py:meth:`~xarray.Dataset.drop` method returns a new object with the listed
 index labels along a dimension dropped:
 
 .. ipython:: python
+    :okwarning:
 
     ds.drop(['IN', 'IL'], dim='space')
```
shoyer: Since this is deprecated, we should fix the docs to use `ds.drop(space=['IN', 'IL']` as an example of the preferred syntax going forward.

max-sixty:
@shoyer thanks for the follow-up comments. I probably should have spotted those myself. 
But given there'll be points I miss, lmk if I should be waiting longer / raising specific points before hitting the big green button, or these post-merge reviews are OK. I can ensure they're complete before a release.

shoyer:
Post merge reviews are totally fine. It is not a big deal to revert a merge if needed.

gwgundersen:
Happy to make these changes, but what's the protocol here? Seems like reverting the merge and letting me amend is cleaner than me creating a new PR.

xarray/core/dataset.py
gwgundersen: Yeah, I thought needing the check for `labels_are_coords` was odd, but I didn't expect passing in coordinates to take a separate code path. Special-casing the `dict` type makes sense. And the API that is being deprecated is treating `DataArrayCoordinates` as non-dict-like, right?

shoyer:
My (limited) understanding of how github handles reverting is that you'll still need to make a new pull request to apply your edits either way. There is no more going to back to amend after PR is  merges. The only difference is what the state of "master" looks like when you make your new pull request (whether it  includes your partial changes or not).

So I don't think it particularly matters if you revert first  (as long as the forward fix happens relatively  soon), but if you'd prefer to do it that way that's totally fine -- let me know and I will do it.

xarray/core/dataset.py
shoyer: > And the API that is being deprecated is treating `DataArrayCoordinates` as non-dict-like, right?

That's right, we would rather *always* use this  new rule for dict-like inputs.

max-sixty:
I think the easiest way forward is to do a follow-up PR, no need to revert.

Reverting would be helpful if we wanted to _undo_ these changes - in this case we have some v marginal additional changes.

PR #3130, opened 2019-07-14
Merge broadcast_like docstrings, analyze implementation problem

### LINKED ISSUE #3129: New .broadcast_like() incorrect
Unfortunately, more robust tests show there is some logic problem in the @dcherian solution to #2885.  I was merging my work in #3127 with his more-principled refactoring in #3086, and my tests showed an error in the refactoring.

DavidMertz:
This PR fixed the forked code connected to #3129.  

 - [x] Closes #3129
 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -25,7 +25,8 @@ New functions/methods
 ~~~~~~~~~~~~~~~~~~~~~
 
 - Added :py:meth:`DataArray.broadcast_like` and :py:meth:`Dataset.broadcast_like`.
-  By `Deepak Cherian <https://github.com/dcherian>`_.
+  By `Deepak Cherian <https://github.com/dcherian>`_ and `David Mertz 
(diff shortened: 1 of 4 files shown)


dcherian:
Thanks @DavidMertz 

Line 2042 of dataset.py should be 
`return _broadcast_helper(args[1], exclude, dims_map, common_coords)` i.e. `args[1]` instead of `self`.

Can you make that change (and the same change in dataarray.py, Line 1019)?

DavidMertz:
Made changes suggested by @dcherian. This resolves the detected bug in the test.  If CI passes, this should be ready to merge.

dcherian:
Thanks for catching that @DavidMertz. Can you add back the original, simpler test too? This one is really just testing align. You could use `@pytest.mark.parametrize` to provide two different sets of `arr1, arr2` arguments.

shoyer:
thanks @DavidMertz and @dcherian !

PR #3111, opened 2019-07-13
A new example Notebook that plots Discrete Sampling Geometry Data

MBARIMike:
In the spirit of providing more examples of using xarray here's a Notebook that makes some simple plots of oceanographic measurement data. I expect that I'll come back to this myself when I need to be reminded of some of the special incantations needed to quickly get some nice plots.

PR #3131, opened 2019-07-15
WIP: tutorial on merging datasets

### LINKED ISSUE #1391: Adding Example/Tutorial of importing data to Xarray (Merge/conact/etc)
I love xarray for analysis but getting my data into xarray often takes a lot more time than I think it should. I am a hydrologist and very often hydro data is poorly stored/formatted, which means I need to do multiple merge/conact/combine_first operations etc. to get to a nice xarray dataset format. I think having more examples for importing different types of data would be helpful (for me and possibly others), instead of my current approach, which often entails trial and error. 

I can start off by providing an example of importing funky hydrology data that hopefully would be general enough for others to use. Maybe we can compile other examples as well. With the end goal of adding to the readthedocs.

rabernat:
- [x] Closes #1391 

This is a start on a tutorial about merging / combining datasets.


shoyer:
I think "by_coords" is probably the most user friendly default for open_mfdataset? But I'm not entirely sure...

TomNicholas:
Nor am I. I originally thought it should be 'nested' because the
concat-in-order behaviour is more similar to the original auto_combine, but
I don't know.

It seems to me like it depends on the quality of user's data: if they have
high-quality datasets which already have sensible coordinates for each
dimension then' by_coords' is best, but if their data is more primitive
without coordinates (like mine happens to be) then 'nested' is the most
natural default. So do we have a sense of what the largest number of users
would find most natural? Or is this not a good way to think about it?

shoyer:
It is possible that we do actually need a third combine mode that works like the old `auto_combine`.

rabernat:
I am still hoping to finish this one day. Any reason it needs to be closed?

keewis:
err, sorry, no. That happened because I deleted the branch you tried to merge into. Let me try to fix that.

keewis:
we should rebase this and #3111 onto `master` so we don't depend on the old `scipy19-docs` branch. If instead we want to continue having a separate development branch for documentation, I think we should use one that is kept in sync with current master.

keewis:
@rabernat, I did the rebase for this and #3111, so when you eventually pick this up again, a simple merge should get this up-to-date with `master`

dcherian:
Thanks @keewis !

andersy005:
@rabernat, the gentlest of bumps on this :)... How much work (content) is left to bring this to completion? I'm asking because I'd be happy to help if there's still more work and/or follow-up PR needed.

PR #3132, opened 2019-07-15
Internal clean-up of isnull() to avoid relying on pandas

shoyer:
This version should be much more compatible out of the box with duck typing.

No user facing changes.


shoyer:
I plan to merge this soon. Tests are passing, aside from the flakey docs build.

Issue #2042, opened 2018-04-07
Anyone working on a to_tiff?  Alternatively, how do you write an xarray to a geotiff?  


ebo:
Matthew Rocklin wrote a gist <https://gist.github.com/mrocklin/3df315e93d4bdeccf76db93caca2a9bd> to demonstrate using XArray to read tiled GeoTIFF datasets, but I am still confused as to how to write them to a GeoTIFF.  I can easily create a tiff with "rasterio.open(out, 'w', **src.profile)", but the following does not seem like the best/cleanest way to do this:

```
ds = xr.open_rasterio('myfile.tif',  chunks={'band': 1, 'x': 2048, 'y': 2048})
with rasterio.open('myfile.tif', 'r') as src:
    with rasterio.open('new_myfile.tif', 'w', **src.profile) as dst:
            for i in range(1, src.count + 1):
                dst.write(ds.variable.data[i-1].compute(), i)

```
Also, if the profile and tags were propagated through open_rasterio, then the second open would not be necessary and would be generally useful.

Schlump:
https://github.com/robintw/XArrayAndRasterio/blob/master/rasterio_to_xarray.py

fmaussion:
> if the profile and tags were propagated through open_rasterio, then the second open would not be necessary and would be generally useful.

We have been adding new attributes like this recently (https://github.com/pydata/xarray/pull/1583 and https://github.com/pydata/xarray/pull/1740), so I don't see much trouble in adding a few more. Note that the rasterio object is available via the (undocumented) ``_file_obj`` attribute. So a quick workaround for you in the mean time would be to access the info you need directly via this object.

As for the ``to_rasterio`` method, I'm currently against it. I'm already starting to think that these kind of domain specific tools should exist in dedicated projects, not in the main xarray codebase. For rasterio in particular, it turns out that the geotiff/GDAL data model is fairly different from the xarray/NetCDF model. The rasterio folks have also shown only limited interest in our endeavor (https://github.com/mapbox/rasterio/issues/920), which is understandable. I don't have a strong opinion though, and I am curious if the @pydata/xarray crew sees it differently.

ebo:
On Apr 8 2018 12:45 PM, Fabien Maussion wrote:
>> if the profile and tags were propagated through open_rasterio, then 
>> the second open would not be necessary and would be generally useful.
>
> We have been adding new attributes like this recently
> (https://github.com/pydata/xarray/pull/1583 and
[...]

PR #3136, opened 2019-07-17
Added suppport for geotiff colormap() colorinterp to be loaded into attrs

alando46:
This is tangentially related to the [`to_geotiff()` discussion](https://github.com/pydata/xarray/issues/2042).

We're using [intake-xarray](https://github.com/intake/intake-xarray) to read a year's worth of Sentinel-2 Scene Classification Layers into a DataArray, doing some processing, and then writing the output to a single band GeoTIFF. When writing the output, we'd like to preserve the colormap of the original underlying data. `xarray-intake` uses `xarray` to read the dataset, so thought it might be useful to add support in `open_rasterio()` to read and load `src.colormap(1)` and `src.colorinterp` into the attributes of the DataArray. My thinking is that if in the future the community ends up wanting to add a `to_geotiff()` method, reading the `src` colormap and colorinterp might be useful.

Before adding tests, I thought I'd submit the PR to see what others thought about adding this functionality. If there's support for this, I'm happy to add any needed tests or documentation.

 - [ ] Tests added

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -313,12 +313,12 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None,
     if hasattr(riods, 'units') and any(riods.units):
         # A list of units string for each dataset band
         attrs['units'] = riods.units
-    if hasattr(riods, 'colormap') and any(riods.colormap):
+    if hasattr(riods, 'colormap'):
         # A dict containing the colormap for a band
         attrs['colormap'] = riods.colormap(1)
     if hasattr(riods, 'colorinterp') and any(riods.colorinterp):
         # A tuple of the band's colorinterp property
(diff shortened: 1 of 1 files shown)


dcherian:
I think we'd like to load all possible attrs and make them accessible if it's easy to do so. 

Is there a way to do that without having to explicitly add each one?

alando46:
@dcherian that makes sense to me and shouldn't be difficult to do. I basically just tried to follow the existing design pattern for the other attributes.

keewis:
this is related to #4697: now that we consider deprecating `xr.open_rasterio` in favor of `xr.open_dataset(..., engine="rasterio")` (implemented by `rioxarray`), should this be moved there?

snowman2:
Related: https://github.com/corteva/rioxarray/issues/191

dcherian:
Should be fixed in rioxarray since we've now deprecated open_raserio.

Issue #3092, opened 2019-07-11
black formatting


crusaderky:
I, like many others, have irreversibly fallen in love with black.
Can we apply it to the existing codebase and as an enforced CI test?
The only (big) problem is that developers will need to manually apply it to any open branches and then merge from master - and even then, merging likely won't be trivial.
How did the dask project tackle the issue?

max-sixty:
I would be v keen. I find it _great_. 
But I'm generally too keen on a) tools and b) changing everything at once, so I discount myself a bit here!

I don't think it's possible to apply incrementally, unfortunately (e.g. only on changed lines). So it would require a single sweeping change. Interested to know how other projects are handling. IIRC they're applying it to some parts of stdlib too.

jhamman:
If we do this, which I am +1 on, we should add a pre-commit hook so we don't have to think about the manual applications. We should probably just copy the dask approach: https://github.com/dask/dask/pull/4983.

shoyer:
+1 from me, too.

It's definitely going to cause some nasty merge conflicts for any work in progress, but hopefully it's straightforward to fix those by running black on the PRs and then merging in master.

We should be able to simply add another job on Azure for this. The `lint` task using flake8 runs very quickly (less than 30 seconds!) and I would expect to see the same from `black`.

andersy005:
Since Black requires Python >=3.6 only, which would set the minimum Python version to 3.6 for contributing to xarray (which requires Python >= 3.5), are there any plans of dropping Python 3.5 support in the near future?

shoyer:
> Since Black requires Python >=3.6, which would set the minimum Python version to 3.6 for contributing to xarray (which currently requires Python >= 3.5), are there any plans of dropping Python 3.5 support in the near future?

I don't think these decisions need to be coupled. It's OK to make contributors upgrade to a newer version of Python before all our users.

There are no immediate plans to drop Python 3.5 support, but we haven't really discussed it either. I think it would definitely make sense to drop Python 3.5 after Python 3.8 is available, maybe sooner. Four major Python versions would definitely be too many.

max-sixty:
I'm happy to do this today

shoyer:
Let's wait a little while for the burst of activity from SciPy to die down
first.

max-sixty:
@shoyer great

TODOs (please add where I've missed)

- [ ] Reformat code
- [ ] CI checks
[...]

PR #3142, opened 2019-07-18
Black

max-sixty:
From https://github.com/pydata/xarray/issues/3092

- [x] Reformat code
- [x] CI checks
- [x] Short instructions for how to merge an existing PR (i.e. avoid manual merge resolution)
  - Not sure if there's magic here - I think people would just have to format their code and then hope git can resolve (i.e. because there would still be no common parent)?
- [x] Black badge
- [x] Do we want to keep `flake8` checks? Black is mostly stricter but not always, e.g. on lines at the end of files. (+0.3 from me to use only `black` and stop using `flake8`)

~- [ ] Do we want to include `isort`? (-0.1 from me, even though I like the tool)~


doc/contributing.rst
```diff
@@ -340,23 +340,24 @@ do not make sudden changes to the code that could have the potential to break
 a lot of user code as a result, that is, we need it to be as *backwards compatible*
 as possible to avoid mass breakages.
 
-Python (PEP8)
-~~~~~~~~~~~~~
+Code Formatting
```
max-sixty: This was predominantly copied from dask's

azure-pipelines.yml
```diff
@@ -55,6 +55,10 @@ jobs:
     displayName: Install flake8
   - bash: flake8
     displayName: flake8 lint checks
+  - bash: python -m pip install black
```
shoyer: Can we just make this a separate job? The lint job takes like 20 seconds to warm up, so the difference in run-time will not be noticeable and it will be nice to see the status separately.

max-sixty:
FYI, unfortunately `pep8speaks` doesn't yet support black

max-sixty:
I think we can suggest something like this to existing PRs to eliminate manual conflicts, and I'm attempting to find something more official. Let me know if anyone knows of something:

- Merge master prior to the `black` commit (any conflicts are real conflicts and require resolving)
- Apply `black .` to this branch
- Merge master at the `black` commit, resolving in favor of our changes: `git merge [black-commit] -X ours`
- Merge current master (any conflicts here are real and again require resolving)

max-sixty:
What are people's thoughts on doing this now? There are a few big merges happening atm, so could do this after those.

I tested the approach [above](https://github.com/pydata/xarray/pull/3142#issuecomment-514325568) and it worked well. I don't have any more external validation than my test though. If anyone has experience here, please let us know.

shoyer:
Let's do it after merging #2277

Zac-HD:
Sadly if there's a way to automatically fix up merge conflicts I don't know of it 😭.  However I'd be happy to help anyone who is having trouble rebasing their open PR.

On a different note, have you considered adding the `pyupgrade` and `autoflake8` autoformatters?  They go further than Black in that they do change the AST, but it's a nice way to ensure consistency of style, automatically remove unused imports and variables for flake8, and so on.

crusaderky:
I did a manual pass of pyupgrade (#3190), and it looks good.
+1, but let's avoid bloating and leave it for a later PR

max-sixty:
> On a different note, have you considered adding the pyupgrade and autoflake8 autoformatters? 

Nice, I didn't know those. I had use `autopep8` before. `isort` is another I use a lot.

>  +1, but let's avoid bloating and leave it for a later PR

Yeah - not sure of the balance here. 
I think the focus should be on easy initial contributions. Consistent formatting is helpful in reading and reviewing code, and eliminating any formatting iterations.
Each extra tool's benefit needs to be weighed against the startup cost of another install - unless there's a way of eliminating that (e.g. I've seen some bots add the code changes from GH)

An alternative is to run those additional tools across the whole codebase on occasion, assuming this doesn't create too many merge conflicts.

Interested in others' thoughts... (but probably for another PR)

max-sixty:
@pydata/xarray this is ready to go! 

Any final comments / thoughts / feedback?

dcherian:
image 

:exclamation:

doc/gallery/plot_cartopy_facetgrid.py
```diff
@@ -22,23 +22,26 @@
...
-             col='time', col_wrap=1,  # multiplot settings
-             aspect=ds.dims['lon'] / ds.dims['lat'],  # for a sensible figsize
-             subplot_kws={'projection': map_proj})  # the plot's projection
+p = air.plot(
+    transform=ccrs.PlateCarree(),  # the data's projection
+    col="time",
+    col_wrap=1,  # multiplot settings
+    aspect=ds.dims["lon"] / ds.dims["lat"],  # for a sensible figsize
+    subplot_kws={"projection": map_proj},
+)  # the plot's projection
```
shoyer: move this up one line?

max-sixty:
OK - going to hit the button as soon as tests pass. 

🙏

max-sixty:
Please ping here / tag me on the relevant PR with any issues, and I'll help!

PR #3143, opened 2019-07-18
Remove the matplotlib=3.0 constraint from py36.yml

shoyer:
The upstream issue that required the constraint seems to have been fixed:
https://github.com/barronh/pseudonetcdf/issues/69


shoyer:
pseudonetcdf has test coverage now with matplotlib 3.1 installed, so we can go ahead here!

Issue #3144, opened 2019-07-18
h5py raising on xr.show_versions()


max-sixty:
Any ideas why `__hdf5libversion__` wouldn't be available? Shall I put a try / except around it?

```python
In [4]: import xarray as xr

In [5]: xr.show_versions()
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/usr/local/lib/python3.7/site-packages/xarray/util/print_versions.py in netcdf_and_hdf5_versions()
     64     try:
---> 65         import netCDF4
     66         libhdf5_version = netCDF4.__hdf5libversion__

ModuleNotFoundError: No module named 'netCDF4'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-5-6f391305f2fe> in <module>
----> 1 xr.show_versions()

/usr/local/lib/python3.7/site-packages/xarray/util/print_versions.py in show_versions(file)
     78     sys_info = get_sys_info()
     79
---> 80     sys_info.extend(netcdf_and_hdf5_versions())
     81
     82     deps = [

/usr/local/lib/python3.7/site-packages/xarray/util/print_versions.py in netcdf_and_hdf5_versions()
     69         try:
     70             import h5py
---> 71             libhdf5_version = h5py.__hdf5libversion__
     72         except ImportError:
     73             pass

AttributeError: module 'h5py' has no attribute '__hdf5libversion__'
```

I check I'm on the latest h5py:
```
pip install h5py -U                                                                                                   Thu Jul 18 16:47:29 2019
Requirement already up-to-date: h5py in ./Library/Python/3.7/lib/python/site-packages (2.9.0)
Requirement already satisfied, skipping upgrade: numpy>=1.7 in /usr/local/lib/python3.7/site-packages (from h5py) (1.16.4)
Requirement already satisfied, skipping upgrade: six in ./Library/Python/3.7/lib/python/site-packages (from h5py) (1.12.0)
```

shoyer:
Hmm, now I'm not sure that line ever worked. I may just copied the variable from netCDF4.

It think it should be `h5py.version.hdf5_version`, which has been around in h5py for quite a while, e.g.,
https://github.com/h5py/h5py/blob/2.0.0/h5py/version.py#L8
https://github.com/h5py/h5py/blob/2.9.0/h5py/version.py#L39

But yes, let's catch `AttributeError`, too, just in case that changes in the future.

[...]

PR #3145, opened 2019-07-18
Fix h5py version printing

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/3144

I surrounded the whole check for h5py & netcdf with a blanket try / except, consistent with the other modules (but maybe too broad?

### DIFF
--- a/xarray/util/print_versions.py
+++ b/xarray/util/print_versions.py
@@ -68,7 +68,7 @@ def netcdf_and_hdf5_versions():
     except ImportError:
         try:
             import h5py
-            libhdf5_version = h5py.__hdf5libversion__
+            libhdf5_version = h5py.version.hdf5_version
         except ImportError:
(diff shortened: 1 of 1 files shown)


xarray/util/print_versions.py
```diff
@@ -77,7 +77,10 @@ def netcdf_and_hdf5_versions():
 def show_versions(file=sys.stdout):
     sys_info = get_sys_info()
 
-    sys_info.extend(netcdf_and_hdf5_versions())
+    try:
+        sys_info.extend(netcdf_and_hdf5_versions())
+    except Exception as e:
+        print(f"Error collecting netcdf / hdf5 version: {e}")
```
shoyer: sorry, we still support Python 3.5

shoyer:
thanks!

PR #3153, opened 2019-07-20
BUG fix +test .sel method gives error with float32 values

### LINKED ISSUE #3137: .sel method gives error with float32 values
`.sel` method gives error when it is used to select `float32` values however it works fine for  `float64`
Example:
```python
import xarray as xr
import numpy as np

a = np.asarray([0.    , 0.111, 0.222, 0.333], dtype='float32')
ds = xr.Dataset(coords={'a': (['a'],a )})
ds.a.sel({'a': 0.111})
```
Error Traceback:
```python
~/anaconda3/envs/check3/lib/python3.7/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2656             try:
-> 2657                 return self._engine.get_loc(key)
   2658             except KeyError:

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Float64HashTable.get_item()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Float64HashTable.get_item()

KeyError: 0.111

<ipython-input-7-b4190c1d9580> in <module>
      4 a = np.asarray([0.    , 0.111, 0.222, 0.333], dtype='float32')
      5 ds = xr.Dataset(coords={'a': (['a'],a )})
----> 6 ds.a.sel({'a': 0.111})
      7 ds.a.sel({'a': 0.111}, method= 'nearest')

~/anaconda3/envs/check3/lib/python3.7/site-packages/xarray/core/dataarray.py in sel(self, indexers, method, tolerance, drop, **indexers_kwargs)
    853         ds = self._to_temp_dataset().sel(
    854             indexers=indexers, drop=drop, method=method, tolerance=tolerance,
--> 855             **indexers_kwargs)
    856         return self._from_temp_dataset(ds)
    857 

~/anaconda3/envs/check3/lib/python3.7/site-packages/xarray/core/dataset.py in sel(self, indexers, method, tolerance, drop, **indexers_kwargs)
   1729         indexers = either_dict_or_kwargs(indexers, indexers_kwargs, 'sel')
   1730         pos_indexers, new_indexes = remap_label_indexers(
-> 1731             self, indexers=indexers, method=method, tolerance=tolerance)
   1732         result = self.isel(indexers=pos_indexers, drop=drop)
   1733         return result._overwrite_indexes(new_indexes)

~/anaconda3/envs/check3/lib/python3.7/site-packages/xarray/core/coordinates.py in remap_label_indexers(obj, indexers, method, tolerance, **indexers_kwargs)
    315 
    316     pos_indexers, new_indexes = indexing.remap_label_indexers(
--> 317         obj, v_indexers, method=method, tolerance=tolerance
    318     )
    319     # attach indexer's coordinate to pos_indexers

~/anaconda3/envs/check3/lib/python3.7/site-packages/xarray/core/indexing.py in remap_label_indexers(data_obj, indexers, method, tolerance)
    250         else:
    251             idxr, new_idx = convert_label_indexer(index, label,
--> 252                                                   dim, method, tolerance)
    253             pos_indexers[dim] = idxr
    254             if new_idx is not None:

~/anaconda3/envs/check3/lib/python3.7/site-packages/xarray/core/indexing.py in convert_label_indexer(index, label, index_name, method, tolerance)
    179                 indexer, new_index = index.get_loc_level(label.item(), level=0)
    180             else:
--> 181                 indexer = get_loc(index, label.item(), method, tolerance)
    182         elif label.dtype.kind == 'b':
    183             indexer = label

~/anaconda3/envs/check3/lib/python3.7/site-packages/xarray/core/indexing.py in get_loc(index, label, method, tolerance)
    106 def get_loc(index, label, method=None, tolerance=None):
    107     kwargs = _index_method_kwargs(method, tolerance)
--> 108     return index.get_loc(label, **kwargs)
    109 
    110 

~/anaconda3/envs/check3/lib/python3.7/site-packages/pandas/core/indexes/numeric.py in get_loc(self, key, method, tolerance)
    434             pass
    435         return super(Float64Index, self).get_loc(key, method=method,
--> 436                                                  tolerance=tolerance)
    437 
    438     @cache_readonly

~/anaconda3/envs/check3/lib/python3.7/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2657                 return self._engine.get_loc(key)
   2658             except KeyError:
-> 2659                 return self._engine.get_loc(self._maybe_cast_indexer(key))
   2660         indexer = self.get_indexer([key], method=method, tolerance=tolerance)
   2661         if indexer.ndim > 1 or indexer.size > 1:

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Float64HashTable.get_item()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Float64HashTable.get_item()

KeyError: 0.111
```
xarray version: 0.12.1
numpy version: 1.16.3

HasanAhmadQ7:
.sel method gives error when it is used to select float32 values

 - [x] Closes #3137
 - [x] Tests added


max-sixty:
Hi @HasanAhmadQ7 thanks for the PR - I saw your comment here: https://github.com/pydata/xarray/issues/3137#issuecomment-515626172

I see the tests are failing, but the build fails before it tests these changes - could you try merging master and pushing? That'll give them another chance to run.

xarray/core/dataset.py
```diff
@@ -2322,6 +2323,22 @@ def interp_like(self, other, method='linear', assume_sorted=False,
             ds = self.reindex(object_coords)
         return ds.interp(numeric_coords, method, assume_sorted, kwargs)
 
+    # Helper method for sel()
+    def _maybe_cast_floats(self, indexers):
```
max-sixty: Does anyone have a prior on the likely performance impact of this? We're creating a copy of the indexers on each `.sel` call. 

If a) the requirement to coerce is rare enough, and b) it's expensive to coerce, we could loop through and check whether there are any incongruencies, and only then coerce.

Is there a viable alternative in having the indexer do these casts?

HasanAhmadQ7: @max-sixty Thank you for your time and note. I modified it to avoid unnecessary looping or copying.

In my understanding, the root of the problem is that pandas index is float 64. Despite that the issue only mentions float32, I tried to generalize the solution, by casting to the coords type for any kind of float.

max-sixty: Thanks @HasanAhmadQ7 !

I'll let someone who knows the indexing machinery review this next (@pydata/xarray), given how central this code path is and my relative knowledge of this. Thanks for your patience

shoyer:
@HasanAhmadQ7 thanks for looking into this!

I think this could be solved a little more cleanly at a lower level of xarray's indexing logic. In particular, all of our indexing calls go through [`convert_label_indexer`](https://github.com/pydata/xarray/blob/1d7bcbdc75b6d556c04e2c7d7a042e4379e15303/xarray/core/indexing.py#L121), which in turn uses `get_loc` and `get_indexer_nd` helper functions to wrap the calls into `pandas.Index` methods:
https://github.com/pydata/xarray/blob/1d7bcbdc75b6d556c04e2c7d7a042e4379e15303/xarray/core/indexing.py#L106-L118

These helper function would be a good place to implement this casting logic -- and you might even consider trying to fix it upstream in pandas as well.

xarray/tests/test_dataarray.py
```diff
@@ -817,6 +817,16 @@ def test_sel_dataarray_datetime(self):
...
         assert_equal(result, array)
 
+    def test_sel_float32(self):
+        float_values = [0., 0.111, 0.222, 0.333]
+        coord_values = np.asarray(float_values, dtype='float32')
+        data_values = np.arange(4)
+        array = DataArray(data_values, [('float32_coord', coord_values)])
+        expected = DataArray(data_values[1:3], [('float32_coord',
+                                                 coord_values[1:3])])
+        actual = array.sel(float32_coord=float_values[1:3])
```
shoyer: It would also be good to add a test case where you're only indexing by a scalar.

HasanAhmadQ7:
@shoyer @max-sixty really appreciate your time and feedback. I will work on it within a week.

xarray/tests/test_dataarray.py
HasanAhmadQ7: A fix in pandas that I can imagine is to modify the pd.Index to retain the dtype similar to the PandasIndexAdapter in xarray/core/indexing. However, it seems to me that their design it to coerce any float to float64, so I did not feel they would consider this to be a bug. I would try to dig into such solution more if you recommend so.

I added a test case in which the coords are float16 to show that casting to the coords type is required. 

The lowest level in which I can access the coords type is in the indexing/remap_label_indexers where I could get the coords from the data_obj. In the latest code, casting is done just before calling convert_label_indexer

Sorry if I am missing the point, and thank you in advance for any further feedback.

HasanAhmadQ7:
@shoyer A fix in pandas that I can imagine is to modify the pd.Index to retain the dtype similar to the PandasIndexAdapter in xarray/core/indexing. However, it seems to me that their design it to coerce any float to float64, so I did not feel they would consider this to be a bug. I would try to dig into such solution more if you recommend so.

I added a test case in which the coords are float16 to show that casting to the coords type is required.

The lowest level in which I can access the coords type is in the indexing/remap_label_indexers where I could get the coords from the data_obj. In the latest code, casting is done just before calling convert_label_indexer

Sorry if I am missing the point, and thank you in advance for any further feedback.

xarray/core/indexing.py
```diff
@@ -249,6 +249,8 @@ def remap_label_indexers(data_obj, indexers, method=None, tolerance=None):
                                  'an associated coordinate.')
             pos_indexers[dim] = label
         else:
+            coords_dtype = data_obj.coords[dim].values.dtype
```
shoyer: Could we simplify this as just:
```suggestion
            coords_dtype = data_obj.coords[dim].dtype
```

xarray/core/utils.py
```diff
@@ -66,6 +66,14 @@ def _maybe_cast_to_cftimeindex(index: pd.Index) -> pd.Index:
         return index
 
 
+def maybe_cast_to_coords_dtype(label, coords_dtype):
+    if isinstance(label, float):
+        label = coords_dtype.type(label)
+    elif isinstance(label, list) and coords_dtype.kind == 'f':
```
shoyer: We also want to convert arrays like `np.array([0, 1/3, 2/3, 1], dtype=np.float32)`.

Actually, maybe we could actually convert any indexing argument to a float array of the appropriate dtype? e.g.,
```python
def maybe_cast_to_coords_dtype(label, coords_dtype):
    if coords_dtype.kind == 'f':
        label = np.asarray(label, dtype=coords_dtype)
    return label
```

shoyer:
> A fix in pandas that I can imagine is to modify the pd.Index to retain the dtype similar to the PandasIndexAdapter in xarray/core/indexing. However, it seems to me that their design it to coerce any float to float64, so I did not feel they would consider this to be a bug. I would try to dig into such solution more if you recommend so.

Yes, after thinking a little bit more about this I think you're totally right.

Pandas only has `Float64Index`. When xarray makes an index for float32 data, we use a Float64Index under the hood for look-ups, but the fact that index represents float32 data is only known to xarray, not pandas.

xarray/core/utils.py
HasanAhmadQ7: This is definitely cleaner. Thank you. However, we still need to check label is not of type slice.

HasanAhmadQ7: I adopted the change in the latest code, but added a check for the case of label is slice

shoyer:
I'm going to merge this shortly assuming tests pass...

shoyer:
thanks @HasanAhmadQ7 !

max-sixty:
Thanks @HasanAhmadQ7 !

PR #3166, opened 2019-07-28
[Feature] Backend entrypoint

### LINKED ISSUE #1970: API Design for Xarray Backends
It has come time to formalize the API for Xarray backends. We now have the following bakends implemented in xarray:

| Backend        | Read | Write |
|----------------|------|-------|
| netcdf4-python | x    | x     |
| h5netcdf       | x    | x     |
| scipy          | x    | x     |
| rasterio       | x    |       |
| zarr           | x    | x     |

And there are conversations about adding additional backends, for example:

- TileDB: https://github.com/pangeo-data/storage-benchmarks/issues/6
- PseudoNetCDF: #1905 

However, as anyone who has worked on implementing or optimizing any of our current backends can attest, the existing DataStore API is not particularly user/developer friendly. @shoyer asked me to open an issue to discuss what a more user friendly backend API would look like so that is what this issue will be. I have left out a thorough description of the current API because, well, I don't think it can done in a  succinct manner (thats the problem).

Note that @shoyer started down a API refactor some time ago in #1087 but that effort has stalled, presumably because we don't have a well defined set of development goals here.

cc @pydata/xarray

jhamman:
In this PR, I'm experimenting with using the entrypoints package to support 3rd party backends. This does not attempt to solidify the API for what the store is, I feel like that should happen in a second PR. Here's how it would work...

In @rabernat's `xmitgcm` package, there is a `_MDSDataStore` that inherits from `xarray.backends.common.AbstractDataStore`. To allow reading `mds` datasets directly in `xarray.open_dataset`, `xmitgcm` would add the following lines to its `setup.py` file:

```python
setup(
    ...
    entry_points={
        'xarray.backends': [
            'mds = xmitgcm.mds_store:_MDSDataStore',
            ...
            ]
        }
    )
```

Xarray would then be able to discover this backend at runtime and users could use the store directly in `open_dataset` calls like this:

```python
ds = xr.open_dataset('./path/to/file.mds', engine='mds', backend_kwargs={...})
```

*Note: I recognize that `xmitgcm.open_mdsdataset` has a bunch of other user options that I'm likely ignoring here but this is meant just as an illustration.*

Now a list of caveats and things to consider:

1. I have only done this for `open_dataset`, not for `to_netcdf`. We may want to consider more generic serialization method that allows for plug-able writers. 
1. `open_dataset` has some special handling for some readers (lock and group selection, file-like objects, etc.). We should work toward moving as much of that logic into the Store objects as possible.
1. We should decide what to do when a 3rd party plugin conflicts with an existing backend. For example, someone could include an entrypoint with the key of `netcdf4`.

---------------------------------
 - [x] Partially closes #1970
 - [ ] Tests added


alexamici:
@jhamman & @shoyer the refactor of the big `if`-chain to a dictionary  in the form `{backend_name: backend_open}` is useful independently of the using the plugin interface, would make this PR smaller and may be useful in the bigger backend refactort.

If you don't mind I'll ask @TheRed86 to open a PR for this refactor alone.

jhamman:
closed by #4577

PR #3173, opened 2019-08-01
Fix distributed.Client.compute applied to DataArray

### LINKED ISSUE #3171: distributed.Client.compute fails on DataArray
As of
- dask 2.1.0
- distributed 2.1.0
- xarray 0.12.1 or git head (didn't try older versions):

```python
>>> import xarray
>>> import distributed
>>> client = distributed.Client(set_as_default=False)
>>> ds = xarray.Dataset({'d': ('x', [1, 2])}).chunk(1)
>>> client.compute(ds).result()
<xarray.Dataset>
Dimensions:  (x: 2)
Data variables:
    d        (x) int64 1 2

>>> client.compute(ds.d).result()
distributed.worker - WARNING -  Compute Failed
Function:  _dask_finalize
args:      ([[array([1]), array([2])]], <function Dataset._dask_postcompute at 0x316a1db70>, ([(True, <this-array>, (<function Variable._dask_finalize at 0x3168f7f28>, (<function finalize at 0x1166bb8c8>, (), ('x',), OrderedDict(), None)))], set(), {'x': 2}, None, None, None, None), 'd')
kwargs:    {}
Exception: KeyError(<this-array>)

<ipython-input-8-2dbfe1b2ff17> in <module>
----> 1 client.compute(ds.d).result()

/anaconda3/lib/python3.7/site-packages/distributed/client.py in result(self, timeout)
    226         result = self.client.sync(self._result, callback_timeout=timeout, raiseit=False)
    227         if self.status == "error":
--> 228             six.reraise(*result)
    229         elif self.status == "cancelled":
    230             raise result

/anaconda3/lib/python3.7/site-packages/six.py in reraise(tp, value, tb)
    690                 value = tp()
    691             if value.__traceback__ is not tb:
--> 692                 raise value.with_traceback(tb)
    693             raise value
    694         finally:

~/PycharmProjects/xarray/xarray/core/dataarray.py in _dask_finalize()
    706     def _dask_finalize(results, func, args, name):
    707         ds = func(results, *args)
--> 708         variable = ds._variables.pop(_THIS_ARRAY)
    709         coords = ds._variables
    710         return DataArray(variable, coords, name=name, fastpath=True)
```

crusaderky:
- [x] Closes #3171
 - [x] Tests added

### DIFF
--- a/xarray/core/utils.py
+++ b/xarray/core/utils.py
@@ -462,11 +462,21 @@ class NDArrayMixin(NdimSizeLenMixin):
 class ReprObject:
     """Object that prints as the given value, for use with sentinel values.
     """
+    __slots__ = ('_value', )
+
     def __init__(self, value: str):
         self._value = value
 
     def __repr__(self) -> str:
         return self._value
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, ReprObject):
+            return self._value == other._value
+        return False
+
+    def __hash__(self) -> int:
+        return hash((ReprObject, self._value))
 
 
 @contextlib.contextmanager
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -55,8 +55,11 @@ Bug fixes
   due to a ``datetime`` issue in NumPy (:issue:`2334`).
   By `Graham Inggs <https://github.com/ginggs>`_.
 - Fixed bug in ``combine_by_coords()`` causing a `ValueError` if the input had
-  an unused dimension with coordinates which were not monotonic (:issue`3150`).
+  an unused dimension with coordinates which were not monotonic (:issue:`3150`).
   By `Tom Nicholas <http://github.com/TomNicholas>`_.
+- Fixed crash when applying ``distributed.Client.compute()`` to a DataArray
+  (:issue:`3171`). By `Guido Imperiale <https://github.com/crusaderky>`_.
+
 
 .. _whats-new.0.12.3:
 
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -300,7 +300,7 @@ class DataArray(AbstractArray, DataWithCoords):
         self,
         variable: Variable = None,
         coords=None,
-        name: Union[Hashable, None, ReprObject] = __default,
+        name: Optional[Hashable] = __default,
     ) -> 'DataArray':
         if variable is None:
             variable = self.variable
@@ -313,7 +313,7 @@ class DataArray(AbstractArray, DataWithCoords):
     def _replace_maybe_drop_dims(
             self,
             variable: Variable,
-            name: Union[str, None, utils.ReprObject] = __default
+            name: Optional[Hashable] = __default
     ) -> 'DataArray':
         if variable.dims == self.dims and variable.shape == self.shape:
             coords = self._coords.copy()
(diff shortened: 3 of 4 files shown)


crusaderky:
Ready for review and merge.
The CI failures are caused by #3174

xarray/core/utils.py
```diff
@@ -462,12 +462,22 @@ def __repr__(self: Any) -> str:
...
     def __repr__(self) -> str:
         return self._value
 
+    def __eq__(self, other) -> bool:
+        if isinstance(other, ReprObject):
+            return self._value == other._value
+        return False
+
+    def __hash__(self) -> int:
+        return hash((ReprObject, self._value))
```
max-sixty: I don't think this will be stable across processes - I just tried this in two different processes:

```python

In [5]: hash(ReprObject)
Out[5]: 8774331846827

(new process)


In [2]: hash(ReprObject)
Out[2]: 8795055479397
```

Does that matter here? If so, could we change to `hash(('ReprObject', self._value))`?

max-sixty: Ah, the hash is never stable across processes, I just tried the same for `'hello'`.

Does dask rely on the hash, or an equality (or something else)?

max-sixty: I think it's here: `--> 708         variable = ds._variables.pop(_THIS_ARRAY)` from https://github.com/pydata/xarray/issues/3171

So then both, so this looks good - doesn't need to be consistent across processes

crusaderky: xarray (actually not dask) relies on the fact that a deep copy of a key object that has gone through a round-trip on the network produces the same hash as the original. The fact that the hash changes (by design; it's an anti-DoS measure) every time you restart the interpteter is irrelevant because the ``__hash__`` method is always invoked locally. It would be, on the other hand, a grave mistake to cache it and then allow the cache to be serialised.

crusaderky: To elaborate:

BAD - the hash defaults to the id, which will break as soon as the object is deep-copied, e.g. by a pickle round-trip
```python
class C:
    pass
```

GOOD:
```python
class C:
    def __hash__(self):
        return 123  # skip: actual calculation
```
BAD - it works as long as you do everything locally, but it will break as soon as you open the object in a different interpreter
```python
class C:
    _hash_cache = None

    def __hash__(self):
        if self._hash_cache is None:
            self._hash_cache = 123  # skip: actual, CPU-intensive, calculation
        return self._hash_cache
```
GOOD AGAIN:
```python
class C:
    _hash_cache = None

    def __hash__(self):
        if self._hash_cache is None:
            self._hash_cache = 123  # skip: actual, CPU-intensive, calculation
        return self._hash_cache

    def __getstate__(self):
        state = self.__dict__.copy()
        state.pop("_hash_cache", None)
        return state
```

max-sixty: Thanks for the explanation! Out of interest, why is this the case: 

> BAD - it works as long as you do everything locally, but it will break as soon as you pickle the object and unpickle it in a different interpreter. 

...assuming...

>The fact that the hash changes every time you restart the interpteter is irrelevant because the __hash__ method is always invoked locally.

shoyer: I'm a little surprised that these default values end up in pickles, but I guess it can happen.

max-sixty: I'm doubtful the default values would - an attribute on the class wouldn't be in the instance dict until it's assigned to

Issue #2217, opened 2018-06-05
tolerance for alignment


naomi-henderson:
When using open_mfdataset on files which 'should' share a grid, there is often a small mismatch which results in the grid not aligning properly. This happens frequently when trying to read data from large climate models from multiple files of the same variable, same lon,lat grid and different time intervals. This silent behavior means that I always have to check the sizes of the lon,lat grids whenever I rely on mfdataset to concatenate the data in time.

Here is an example in which I create two 1d DataArrays which have slightly different coordinates:

```python
import xarray as xr
import numpy as np
from glob import glob

tol=1e-14
x1 = np.arange(1,6)+ tol*np.random.rand(5)
da1 = xr.DataArray([9, 0, 2, 1, 0], dims=['x'], coords={'x': x1})

x2 = np.arange(1,6) + tol*np.random.rand(5)
da2 = da1.copy()
da2['x'] = x2

print(da1.x,'\n', da2.x)
```
```
<xarray.DataArray 'x' (x: 5)>
array([1., 2., 3., 4., 5.])
Coordinates:
  * x        (x) float64 1.0 2.0 3.0 4.0 5.0 
 <xarray.DataArray 'x' (x: 5)>
array([1., 2., 3., 4., 5.])
Coordinates:
  * x        (x) float64 1.0 2.0 3.0 4.0 5.0
```
First I save both DataArrays as netcdf files and then use open_mfdataset to load them:
```
da1.to_netcdf('da1.nc',encoding={'x':{'dtype':'float64'}})
da2.to_netcdf('da2.nc',encoding={'x':{'dtype':'float64'}})

db = xr.open_mfdataset(glob('da?.nc'))

db
```
```
<xarray.Dataset>
Dimensions:                        (x: 10)
Coordinates:
  * x                              (x) float64 1.0 2.0 3.0 4.0 5.0 1.0 2.0 ...
Data variables:
    __xarray_dataarray_variable__  (x) int64 dask.array<shape=(10,), chunksize=(5,)>
```
So the x grid is now twice the size.  This behavior is the same if I just use align with join='outer':
```
xr.align(da1,da2,join='outer')
```
```
(<xarray.DataArray (x: 10)>
 array([nan,  9., nan,  0.,  2., nan, nan,  1.,  0., nan])
 Coordinates:
   * x        (x) float64 1.0 1.0 2.0 2.0 3.0 3.0 4.0 4.0 5.0 5.0,
 <xarray.DataArray (x: 10)>
[...]

PR #3175, opened 2019-08-01
Add join='override'

### LINKED ISSUE #2023: quantile method returns quantile coordinates which can raise issues
I browsed the issues to check whether there was an open issue on this, but didn't find anything.
Apologies if it is a duplicate.

I am an happy user of the new quantile method in xarray.  I think though that creating a coordinate that has the same name as the method can raise issues.

While updating my code to switch entirely to xarray and the quantile method, I realized that this could be an issue. In the old code I had some datasets created using np.nanpercentiles, in which I had generated the coordinate and named it "q". 
At some point in my code I have to retrieve the list of the q values, I used to do that by extracting the values:

```python
x = ds.q.values
```
updating it to what would be the coordinate automatically generated by xarray, it should be

```python
x = ds.quantile.values
```

#### Problem description

But of course, when trying to call the coordinate "quantile" from the dataset, you get an Attribute Error:

```python
AttributeError: 'function' object has no attribute 'values'
```
This can be easily solved by accessing the coordinate as:

```python
x = ds['quantile'].values
```

However, I thought to point out that using the same name for the coordinate and the method can raise this type of issues.

dcherian:
This adds `join='override'` which checks that indexes along a dimension are of the same size and overwrites those indices with indices from the first object.

Definitely need help, feedback.

 - [x] With #3102 this partially helps with #2039, #2217
 - [x] Tests added


xarray/core/dataarray.py
```diff
@@ -329,7 +329,7 @@ def _replace_maybe_drop_dims(
                                  if set(v.dims) <= allowed_dims)
         return self._replace(variable, coords, name)
 
-    def _replace_indexes(
+    def _overwrite_indexes(
```
dcherian: This wasn't used anywhere in the codebase so I renamed it to be consistent with `Dataset._overwrite_indexes`.

xarray/core/alignment.py
```diff
@@ -102,6 +108,20 @@ def align(*objects, join='inner', copy=True, indexes=None, exclude=frozenset(),
...
+        for index, obj in enumerate(objects[1:]):
+            for dim in objects[0].indexes:
+                if dim not in exclude:
+                    if obj.indexes[dim].size != objects[0].indexes[dim].size:
+                        raise ValueError("Cannot use join='override' when "
+                                         "all indexes don't have the same "
+                                         "shape.")
+                    new_indexes[dim] = objects[0].indexes[dim]
+
+            objects[index + 1] = obj._overwrite_indexes(new_indexes)
```
dcherian: Is this the right way to do this?

crusaderky:
Hi,

This doesn't close #2039 because

- open_mfdataset does not offer a 'join' parameter to be propagated to align
- even if it did, by when open_mfdataset invokes align time has already been wasted because indices, as of today, are eagerly loaded by open_dataset
- it does nothing about non-index variables whose dims don't include the concat_dim

dcherian:
Ah, good point. Do you have any thoughts on the implementation? or on the `compat='override'` proposal in https://github.com/pydata/xarray/issues/2064#issuecomment-519149757

crusaderky:
Lazy-loading indices, plus this and #3102, would probably solve #2039.
Without lazy-loading, an ad-hoc needs to be implemented in open_mfdataset directly.

dcherian:
Ready for review.

xarray/core/alignment.py
shoyer: It looks like this should work, yes.

xarray/core/alignment.py
```diff
@@ -99,6 +105,21 @@ def align(
         obj, = objects
         return (obj.copy(deep=copy),)
 
+    if join == "override":
+        objects = list(objects)
+        new_indexes = dict()
+        for index, obj in enumerate(objects[1:]):
+            for dim in objects[0].indexes:
```
shoyer: There is a potential edge case here, where the first object does not have one of the dimensions found on multiple later objects, e.g.,
`align(DataArray(0), DataArray([1], [('x', [1])]), DataArray([2, 3], [('x', [2, 3])])`

This should be an error, but I think the current code would let this pass through, possibly creating an invalid object. Instead, I think the right strategy is to iterate everything, building up the set of indexes to match against as you go.

Another edge case that I think you might get wrong currently is something like this, where the last DataArray doesn't have one of the indexes at all:
`align(DataArray([1], [('x', [1])]), DataArray([2], [('x', [2])], DataArray(3))`

xarray/core/alignment.py
```diff
@@ -99,6 +105,21 @@ def align(
         obj, = objects
         return (obj.copy(deep=copy),)
 
+    if join == "override":
```
shoyer: Could the logic from this block be moved to a helper function? `align` is already getting pretty long.

xarray/core/alignment.py
dcherian: Ah thanks. Fixed now. I've added these two examples as a test case.

xarray/core/alignment.py
```diff
@@ -31,10 +31,34 @@ def _get_joiner(join):
...
+        # We rewrite all indexes and then use join='left'
+        return operator.itemgetter(0)
     else:
         raise ValueError("invalid value for join: %s" % join)
 
 
+def _override_indexes(objects, all_indexes, exclude):
+    for dim, dim_indexes in all_indexes.items():
+        if dim not in exclude:
+            lengths = [dd.size for dd in dim_indexes]
```
shoyer: Could we just use a set comprehension here? Also I have a minor distaste for abbreviations like `dd`.

e.g., maybe
``python
            lengths = {index.size for index in dim_indexes}
            if len(lengths) != 1:
                raise ValueError(...)
```

File: xarray/core/alignment.py [hunk 6]
```diff
@@ -31,10 +31,34 @@ def _get_joiner(join):
...
+            if not all([ll == lengths[0] for ll in lengths[1:]]):
+                raise ValueError(
+                    "Indexes along dimension %r are not equal."
```
shoyer:
```suggestion
                    "Indexes along dimension %r don't have the same length."
```

xarray/core/alignment.py
dcherian: I like that.

shoyer:
LGTM -- anyone else have opinions here?

max-sixty:
Yes agree! Thanks @dcherian !

dcherian:
Great, thanks!

PR #3177, opened 2019-08-02
More annotations

crusaderky:


xarray/core/dataset.py
```diff
@@ -3291,11 +3298,18 @@ def drop(self, labels, dim=None, *, errors='raise'):
...
+
         if dim is None:
+            if isinstance(labels, str) or not isinstance(labels, Iterable):
+                labels = {labels}
+            else:
+                labels = set(labels)
+
             return self._drop_vars(labels, errors=errors)
         else:
+            if utils.is_scalar(labels):
```
crusaderky: I don't think I fully understand the use case where dims is not None, where apparently you are allowed to feed into labels a whole DataArray. It also makes me nervous that there's a single unit test, not together with the other tests for Dataset.drop, that triggers the use case.

dcherian:
Like a machine!

xarray/core/dataset.py
max-sixty: Isn't this for the case you pass a single label; what's the case with passing a whole DataArray?

```python
In [12]: da = xr.DataArray(data=np.ones((2, 3)), dims=['x', 'y'], coords=dict(x=list('ab')))

In [13]: da
Out[13]:
<xarray.DataArray (x: 2, y: 3)>
array([[1., 1., 1.],
       [1., 1., 1.]])
Coordinates:
  * x        (x) <U1 'a' 'b'

In [14]: da.drop('a', dim='x')
Out[14]:
<xarray.DataArray (x: 1, y: 3)>
array([[1., 1., 1.]])
Coordinates:
  * x        (x) <U1 'b'
```

max-sixty:
> Like a machine!

Living up to the headshot; a rampage of typing...

Thanks @crusaderky !

xarray/core/dataset.py
crusaderky: This is the problem. I'm unsure if it's supposed to be allowed?
I've never seen anywhere else, where dataset variable names are required, the option to pass named DataArray's instead.

https://github.com/pydata/xarray/blob/dba85bdfc1ac311349bbfcabccb02a71747a97b2/xarray/tests/test_dataset.py#L4245

crusaderky: The fact is that ``iter(DataArray)`` yields scalar DataArrays, not the elements of the DataArray (unlike ``iter(numpy.ndarray)``)

So the lines
```python
            if isinstance(labels, str) or not isinstance(labels, Iterable):
                labels = {labels}
            else:
                labels = set(labels)
```
fall apart when labels is a DataArray.
To make it work I should change it to something like:
```python
            if isinstance(labels, str) or not isinstance(labels, Iterable):
                labels = {labels}
            elif isinstance(labels, (DataArray, Variable):
                labels = set(labels.values)
            else:
                labels = set(labels)
```

The problem is that the logic above is EVERYWHERE. None of the many, many functions that expect "one or more dimension names" as an input cope (or are tested against) a DataArray input.

I lean towards removing the option entirely, and changing the unit test from
```python
ds.drop(ds.y, dim='y')
```
to
```python
ds.drop(ds.y.values, dim='y')
```

This is chiefly due to the fact that passing a DataArray is incredibly counter-intuitive; it took me a *long* time to figure out that ``ds.drop(ds.y, dim='y')`` is not trying to drop the y variable.

crusaderky: Note that I wrote ``labels = labels.values`` and not ``labels = set(labels.values)``. That would cause a huge performance hit when values has many elements.

crusaderky: I think the problem is due to the fact that the same method drop() performs two fundamentally different things - drop variables or elements from the variables - and in the latter case there are plenty of similar methods that DO accept a DataArray, e.g. sel(), where(), etc.

crusaderky: I think I cracked it - please review

crusaderky:
Ready for review and merge

xarray/tests/test_dataarray.py
```diff
@@ -1904,9 +1904,9 @@ def test_drop_coordinates(self):
         assert_identical(actual, expected)
 
         with raises_regex(ValueError, 'cannot be found'):
-            arr.drop(None)
+            arr.drop('w')
```
max-sixty: What happens if `None` is passed? My quick test suggest this test should still pass, is that right?

xarray/tests/test_dataset.py
```diff
@@ -2000,6 +2000,12 @@ def test_drop_index_labels(self):
         expected = data.isel(x=slice(0, 0))
         assert_identical(expected, actual)
 
+        # DataArrays as labels are a nasty corner case as they are not
+        # Iterable[Hashable] - DataArray.__iter__ yields scalar DataArrays.
+        actual = data.drop(DataArray(['a', 'b', 'c']), 'x', errors='ignore')
```
max-sixty: Great, it's good this now works

xarray/core/dataset.py
shoyer: Yes, it would be ideal to have separate functions for dropping variables rather than elements along an array dimension.

We are also considering refactoring this function over in https://github.com/pydata/xarray/pull/3128, I would be very interested in your thoughts!

xarray/core/dataset.py
```diff
@@ -3338,8 +3385,10 @@ def drop_dims(self, drop_dims, *, errors='raise'):
         if errors not in ['raise', 'ignore']:
             raise ValueError('errors must be either "raise" or "ignore"')
 
-        if utils.is_scalar(drop_dims):
+        if isinstance(drop_dims, str) or not isinstance(drop_dims, Iterable):
```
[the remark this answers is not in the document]
crusaderky: ``is_scalar`` doesn't cope with non-string Hashables, e.g. ``enum.Enum``. I'm uncomfortable touching it without going through the exercise of thoroughly reviewing everything that invokes it.

xarray/core/dataset.py
```diff
@@ -3291,11 +3315,21 @@ def drop(self, labels, dim=None, *, errors='raise'):
...
+                labels = {labels}
+            else:
+                labels = set(labels)
+
             return self._drop_vars(labels, errors=errors)
         else:
+            # Don't cast to set, as it would harm performance when labels
+            # is a large numpy array
+            if utils.is_scalar(labels):
+                labels = [labels]
```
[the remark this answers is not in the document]
crusaderky: Done

xarray/core/dataset.py
```diff
@@ -3351,15 +3400,15 @@ def drop_dims(self, drop_dims, *, errors='raise'):
                         for d in v.dims if d in drop_dims)
         return self._drop_vars(drop_vars)
 
-    def transpose(self, *dims):
+    def transpose(self, *dims: Hashable) -> 'Dataset':
```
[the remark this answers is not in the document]
crusaderky: It is possible, yes, but IMHO detracts from readibility. The question here is, do people actually subclass from DataArray and Dataset? I've always found xarray rather hostile to subclassing, chiefly due to the heavy interaction between DataArray and Dataset.

xarray/tests/test_dataarray.py
crusaderky: None _technically_ works, but it's not a valid use case. ``hash(None)`` works, but ``isinstance(None, Hashable)`` returns False. Also, xarray uses None everywhere to mark optional parameters, and in this case it asks the user to explicitly provide a value.

xarray/core/dataset.py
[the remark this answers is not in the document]
crusaderky: Actually, rereading it, it does cope with Enum. However, most functions that today use ``if isinstance(drop_dims, str) or not isinstance(drop_dims, Iterable):`` won't transparently cope with 0-dimensional numpy arrays or, even worse, 0-dimensional DataArrays.

crusaderky:
Ready for merge if everybody's happy with the current state

max-sixty:
OK, let's merge and we can iterate if needed from here. Thanks as ever @crusaderky 

@shoyer - do you have any thoughts on `is_scalar` from [this thread](https://github.com/pydata/xarray/pull/3177#discussion_r310813393)?

PR #3179, opened 2019-08-02
remove type annotations from autodoc method signatures

### LINKED ISSUE #3178: type annotations make docs confusing
The annotations make this signature basically unreadable to anyone not familiar with them. 

image

Is there a way to hide them in the documentation? Or at least change formatting so it's clearer what the function arguments are?

DocOtak:
This PR removes all the type hints from method signatures generated by sphinx.ext.autodoc. See http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints

The sphinx documentation doesn't say which version this was added in, but I imagine it is quite recent.

 - [ ] Closes #3178

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -83,6 +83,7 @@ sphinx_gallery_conf = {'examples_dirs': 'gallery',
                        }
 
 autosummary_generate = True
+autodoc_typehints = 'none'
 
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks @DocOtak !

PR #3180, opened 2019-08-02
enable sphinx.ext.napoleon

### LINKED ISSUE #3056: Argument and its type joined in docs
This issue has nearly no effect on users, but (at least for me) it is quite bothering aesthetically. It also looks like something simple to solve even though I am not an sphinx expert so I am not sure. 

In the docstring, there is a colon between the name of one argument and its type, however, in the online documentation, the name and the type are smashed together. In the case of `xr.align`:

image
image
image

I think it would be more readable and appealing if there was some kind of separation between the two.

DocOtak:
Enables the napoleon extension in sphinx. This will interpret the numpydoc style parameters and types and convert them to sphinx `:param type name:` Note that `sphinx.ext.napoleon` must come before the `numpydoc` extension.

Eventually the numpydoc dependency might be able to be removed but currently removing it makes the wrapped `ufunc` documentation omit the "parameters" and "returns", see the attached screenshot for what one of these ufuncs looks like with `numpydoc` removed.

 - [ ] Closes #3056

napoleon_no_numpydoc

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -66,6 +66,7 @@ extensions = [
     'sphinx.ext.intersphinx',
     'sphinx.ext.extlinks',
     'sphinx.ext.mathjax',
+    'sphinx.ext.napoleon',
     'numpydoc',
     'IPython.sphinxext.ipython_directive',
     'IPython.sphinxext.ipython_console_highlighting',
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks @DocOtak ! 

It's difficult to see the results without merging - are you comfortable enough with this that we can merge and assess the result? Or would you like someone to build the docs locally and check first?

DocOtak:
I was diffing the directory outputs when testing locally and nothing really breaking stood out to me... I think this is safe enough that it should be able to merge and asses the results. IMO the results were quite pretty and definitely addressed #3056 in most places.

dcherian:
Thanks @DocOtak

PR #3183, opened 2019-08-03
Call darray.compute() in plot()

### LINKED ISSUE #2390: Why are there two compute calls for plot?
Anecdotally I find that when I call `.plot()` on a dataset object that holds dask arrays `compute` gets called twice.  Why is this?  I'm curious if this is something that should be resolved.

dcherian:
- [x] Closes #2390
 - [ ] Tests added

Still don't know how to test this.

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -10,7 +10,6 @@ import functools
 import numpy as np
 import pandas as pd
 
-from ..core.pycompat import dask_array_type
 from .facetgrid import _easy_facetgrid
 from .utils import (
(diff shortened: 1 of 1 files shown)


xarray/plot/plot.py
```diff
@@ -141,6 +142,9 @@ def plot(darray, row=None, col=None, col_wrap=None, ax=None, hue=None,
     """
     darray = darray.squeeze()
 
+    if isinstance(darray.data, dask_array_type):
+        darray = darray.compute()
```
shoyer: `compute()` works even on numpy arrays, so let's just write `darray = darray.compute()` here without the `if` check.

Issue #3182, opened 2019-08-03
RTD build failing


dcherian:
Looks like our RTD builds have been failing for the last 3 days.

```
The HTML pages are in _build/html.
/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/sphinx_rtd_theme/search.html:20: RemovedInSphinx30Warning: To modify script_files in the theme is deprecated. Please insert a <script> tag directly in your theme instead.
  {{ super() }}

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/sphinx/cmd/build.py", line 284, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/sphinx/application.py", line 380, in build
    self.events.emit('build-finished', None)
  File "/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/sphinx/events.py", line 103, in emit
    results.append(callback(self.app, *args))
  File "/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/sphinx_gallery/gen_gallery.py", line 313, in sumarize_failing_examples
    "\n" + "-" * 79)
ValueError: Here is a summary of the problems encountered when running the examples

Examples expected to fail, but not failing:
Please remove these examples from
sphinx_gallery_conf['expected_failing_examples']
/home/docs/checkouts/readthedocs.org/user_builds/xray/checkouts/latest/doc/gallery/plot_rasterio.pyin your conf.py file
/home/docs/checkouts/readthedocs.org/user_builds/xray/checkouts/latest/doc/gallery/plot_rasterio_rgb.py
-------------------------------------------------------------------------------

Exception occurred:
  File "/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/sphinx_gallery/gen_gallery.py", line 313, in sumarize_failing_examples
    "\n" + "-" * 79)
ValueError: Here is a summary of the problems encountered when running the examples

Examples expected to fail, but not failing:
Please remove these examples from
sphinx_gallery_conf['expected_failing_examples']
[...]

PR #3186, opened 2019-08-05
bump rasterio to 1.0.24 in doc building environment

DocOtak:
This is hopefully a fix for #3182 but I wasn't sure how to really test this on read the docs (RTD) itself.

There may be a few things going on:
* Local testing showed removing the "auto_gallery" dir would cause the failing gallery examples to actually be recognized as failing (i.e. bust the cache)
* https://github.com/conda-forge/rasterio-feedstock/issues/118 and some examining of the `conda list` output on RTD for led me to think this was a deeper problem. Local testing with a bumped version of rasterio resulted in conda-forge being the channel source (rather than pip?), it also fixed the failing builds which relied on rasterio.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -12,7 +12,7 @@ dependencies:
   - ipython=7.2.0
   - netCDF4=1.4.2
   - cartopy=0.17.0
-  - rasterio=1.0.13
+  - rasterio=1.0.24
   - zarr=2.2.0
   - iris=2.2.0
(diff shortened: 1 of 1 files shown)


DocOtak:
Examining the output of `conda list` in the CI build shows most of the packages are coming from conda-forge now, including rasterio. This should mean the docs will build successfully on RTD.

max-sixty:
Great, let's try it! Thanks @DocOtak

PR #3187, opened 2019-08-06
reduce the size of example dataset in dask docs

DocOtak:
Another attempt at getting the docs to build again on RTD (#3182).

The current failure is due to high memory usage in the dask examples. I've converted the two most memory expensive code blocks into `:verbatim:` blocks so they don't use any additional memory. The contents of these blocks were taken from the last time the docs successfully built on RTD


dcherian:
Thanks @DocOtak. How about we reduce the size of the dataset to 180 in longitude and 30 in time?

DocOtak:
Sure, that seems to work as well, want a second PR or just update this one (with some forcing)?

dcherian:
yes, let's force update this PR

dcherian:
Thanks. let's try this.

PR #3193, opened 2019-08-08
Small updates to contributing.rst

max-sixty:
I'm not sure if people really use this? If they do, it could use another run through. Potentially some of our explanations could become links to better maintained docs; e.g. how to use git / pytest

This makes a couple of small tweaks, in pursuit of incrementalism :)


crusaderky:
Looks good to me, thanks

PR #3190, opened 2019-08-07
pyupgrade one-off run

crusaderky:
A one-off, manually vetted and tweaked run of pyupgrade

PR #3194, opened 2019-08-08
Remove future statements

max-sixty:
Not sure why these weren't picked up in some of the previous python3-ifying

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
     rev: v2.2.3
     hooks:
       - id: flake8
-  # run occasionally, ref discussion https://github.com/pydata/xarray/pull/3194
+  # run these occasionally, ref discussion https://github.com/pydata/xarray/pull/3194
   # - repo: https://github.com/asottile/pyupgrade
   #   rev: v1.22.1
   #   hooks:
@@ -19,3 +19,7 @@ repos:
   #         - "--py3-only"
   #         # remove on f-strings in Py3.7
   #         - "--keep-percent-format"
+  # - repo: https://github.com/timothycrosley/isort
+  #   rev: 4.3.21-2
+  #   hooks:
+  #     - id: isort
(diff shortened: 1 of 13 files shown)


Zac-HD:
In all seriousness, pyupgrade will do this and more automatically with the --py3-only option.  I'd add it to CI!

max-sixty:
We did run it through! https://github.com/pydata/xarray/pull/3190

But not with the `--py3-only` it seems... We could do that in this PR

One reason not to enforce it - I can't seem to work out how to apply it to all files in a repo - am I missing something basic? Or does it need another process to loop over files?

Zac-HD:
The author pushes pre-commit pretty hard, but that shouldn't be a problem for Xarray as we're already using it. Then `pre-commit run --all-files` will execute it.

max-sixty:
I think `pre-commit` is great, but that it's too much burden to compel new contributors to use by adding its effects to CI checks. 
I think we've been fairly successful recently at having some less technical folks contribute; keen not to erect even small barriers.

We could add this to `pre-commit` but _not_ to CI; there'd be some unrelated change though.

If anyone is up for a side project - a tool that submits PRs whenever someone commits non-idiomatic code!

Open minded as ever though

max-sixty:
One addendum: if `pre-commit` gained widespread use, and new contributors were installing and learning about it early, I would definitely be up for loading that up with checks

Zac-HD:
[My config](https://github.com/Zac-HD/hypothesmith/blob/master/tox.ini) uses `tox` and `bash` instead of pre-commit.  Doesn't work well for Windows, but it's a tiny Python script to glob for .py files and `subprocess.run` the pyupgrade command.  Balancing powerful CI against new-contributor-friendliness is hard!

```ini
[testenv:check]
deps =
    -r requirements.txt
whitelist_externals = bash
commands =
    autoflake --recursive --in-place --remove-all-unused-imports --remove-duplicate-keys --remove-unused-variables .
    bash -c \"pyupgrade --py36-plus **.py\"
    isort --recursive --apply .
    black .
    flake8
    mypy --config-file=tox.ini . src/
    bandit --recursive --exclude=./.tox/** --skip=B101,B110,B310 --quiet .
```

max-sixty:
Nice! That's quite a list. Between us there's an xkcd or narkoz reference...

PR #3195, opened 2019-08-08
Update black instructions

max-sixty:
This needed updating with the relevant commit post-black change

I also added a line to apply the patch of manual changes we made on top of the black changes. It makes the list of steps a bit burdensome. But I've tested them a couple of times and, where people have lots of code changes, it's still going to be much easier than resolving manually.

I generally wouldn't want to suggest people curl data from the internet (even if we trust the individual). I think it's probably OK in this instance: the address contains a hash of the contents, and it's only being fed into `git apply -`, not executed. But lmk if that's still a security concern.

I'll update the issue I opened on `black` with the above issue; we should have done one commit with only the `black` changes, and then another with any manual changes.

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -389,14 +389,14 @@ each time you commit changes. You can skip these checks with
   conflicts. Because they have had limited testing, please reach out to the core devs
   on your pull request if you face any issues, and we'll help with the merge:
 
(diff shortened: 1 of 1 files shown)


dcherian:
These instructions worked quite well on #3175

max-sixty:
Great - thanks for the update @dcherian !

PR #3196, opened 2019-08-09
One-off isort run

crusaderky:
A one-off, manually vetted and tweaked isort run

### DIFF
--- a/asv_bench/benchmarks/combine.py
+++ b/asv_bench/benchmarks/combine.py
@@ -1,4 +1,5 @@
 import numpy as np
+
 import xarray as xr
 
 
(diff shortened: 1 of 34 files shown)


crusaderky:
Ready for review and merge

max-sixty:
I was thinking about doing this (and had done it a couple of times in the past). My only concern with doing it now is whether it's likely to exacerbate any merge conflicts from the black changes. I've found that merge conflicts scale super-linearly with changes, at least in frustration if not lines - if so potentially we could wait for the outstanding PRs to update and then rerun?

crusaderky:
Fine for me - is a week enough to allow everybody to adapt?

crusaderky:
Updated, re-run everything, and removed the blanket F401 (unused import) suppression.
Ready for final review and merge.

PR #3202, opened 2019-08-11
chunk sparse arrays

### LINKED ISSUE #3191: DataArray.chunk() from sparse array produces malformed dask array
#3117 by @nvictus introduces support for sparse in plain xarray.
dask already supports it.

Running with:
- xarray git head
- dask 2.2.0
- numpy 1.16.4
- sparse 0.7.0
- NUMPY_EXPERIMENTAL_ARRAY_FUNCTION=1

```python
>>> import numpy, sparse, xarray, dask.array
>>> s = sparse.COO(numpy.array([1, 2]))         
>>> da1 =  dask.array.from_array(s)
>>> da1._meta
<COO: shape=(0,), dtype=int64, nnz=0, fill_value=0>
>>> da1.compute()
<COO: shape=(2,), dtype=int64, nnz=2, fill_value=0>
>>> da2 = xarray.DataArray(s).chunk().data
>>> da2._meta                                                                                                                                                                                                      
array([], dtype=int64)  # Wrong
>>> da2.compute()
RuntimeError: Cannot convert a sparse array to dense automatically. To manually densify, use the todense method.
```

crusaderky:
Closes #3191

@shoyer I completely disabled wrapping in ImplicitToExplicitIndexingAdapter for sparse arrays, cupy arrays, etc.
I'm not sure if it's desirable; the chief problem is that I don't think I understand the purpose of ImplicitToExplicitIndexingAdapter to begin with... some enlightenment would be appreciated.

### DIFF
--- a/xarray/core/duck_array_ops.py
+++ b/xarray/core/duck_array_ops.py
@@ -186,7 +186,12 @@ def as_like_arrays(*data):
     elif any(isinstance(d, sparse_array_type) for d in data):
         from sparse import COO
 
-        return tuple(COO(d) for d in data)
+        out = []
+        for d in data:
+            if isinstance(d, dask_array_type):
+                d = d.compute()
+            out.append(COO(d))
+        return tuple(out)
     else:
         return tuple(np.asarray(d) for d in data)
 
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -941,7 +941,7 @@ class Variable(
                 if LooseVersion(dask.__version__) < "2.0.0":
                     kwargs = {}
                 else:
-                    # All of our lazily loaded backend array classes) should use NumPy
+                    # All of our lazily loaded backend array classes should use NumPy
                     # array operations.
                     kwargs = {"meta": np.ndarray}
             else:
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -926,28 +926,29 @@ class Variable(
         if isinstance(data, da.Array):
             data = data.rechunk(chunks)
         else:
-            if utils.is_dict_like(chunks):
-                chunks = tuple(chunks.get(n, s) for n, s in enumerate(self.shape))
-
-            if not isinstance(data, np.ndarray) and hasattr(data, "__array_function__"):
-                # This should always be true after Variable.as_compatible_data
-                assert IS_NEP18_ACTIVE
-                kwargs = {}
-            else:
-                # da.from_array works by using lazily indexing with a tuple of
-                # slices. Using OuterIndexer is a pragmatic choice: dask does not
-                # yet handle different indexing types in an explicit way:
+            if isinstance(data, indexing.ExplicitlyIndexed):
+                # Unambiguously handle array storage backends (like NetCDF4 and h5py)
+                # that can't handle general array indexing. For example, in netCDF4 you
+                # can do "outer" indexing along two dimensions independent, which works
+                # differently from how NumPy handles it.
+                # da.from_array works by using lazy indexing with a tuple of slices.
+                # Using OuterIndexer is a pragmatic choice: dask does not yet handle
+                # different indexing types in an explicit way:
                 # https://github.com/dask/dask/issues/2883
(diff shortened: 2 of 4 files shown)


xarray/core/variable.py
```diff
@@ -928,21 +928,26 @@ def chunk(self, chunks=None, name=None, lock=False):
...
+                # This should always be true after Variable.as_compatible_data
+                assert IS_NEP18_ACTIVE
+                kwargs = {}
             else:
-                kwargs = dict()
+                # da.from_array works by using lazily indexing with a tuple of
+                # slices. Using OuterIndexer is a pragmatic choice: dask does not
+                # yet handle different indexing types in an explicit way:
+                # https://github.com/dask/dask/issues/2883
+                data = indexing.ImplicitToExplicitIndexingAdapter(
```
shoyer: We should probably only do this wrapping if `data` is a subclass of `xarray.core.indexing.ExplicitlyIndexed`, which is a base class used for various backend arrays. For example, we should just let NumPy arrays through without wrapping. So this bit could just be:
```python
if isinstance(data, ExplicitlyIndexed):
    data = indexing.ImplicitToExplicitIndexingAdapter(
        data, indexing.OuterIndexer
    )
```

The reason why we have explicitly indexed arrays in the first place is so we can unambiguously handle array storage backends (like netCDF4 and h5py) that can't handle general array indexing. For example, in netCDF4 you can do "outer" indexing along two dimensions independent, which works differently from how NumPy handles it.

Xarray uses explicit indexing classes (`BasicIndexer`, `OuterIndexer` and `VectorizedIndexer`) which keep track of and canonicalize the desired indexing operation, without forcing everything into the [form desired by NumPy](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html) -- which has lots of edge cases (especially mixed `slice`/`ndarray` indexing) and isn't always easy to convert into other forms.

xarray/core/duck_array_ops.py
```diff
@@ -186,7 +186,12 @@ def as_like_arrays(*data):
     elif any(isinstance(d, sparse_array_type) for d in data):
         from sparse import COO
 
-        return tuple(COO(d) for d in data)
+        out = []
+        for d in data:
+            if isinstance(d, dask_array_type):
+                d = d.compute()
```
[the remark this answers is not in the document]
crusaderky: I agree it's not a great design due to e.g. distributed.Client global registration.
However, this is already what happens on the line after that for non-sparse mixed arrays:

``return tuple(np.asarray(d) for d in data)`` invokes ``dask.array.Array.__array__``, which on the first line invokes ``self.compute()``.

xarray/tests/test_sparse.py
```diff
@@ -879,3 +864,17 @@ def test_sparse_coords(self):
...
         )
+
+
+def test_chunk():
+    s = sparse.COO.from_numpy(np.array([0, 0, 1, 2]))
+    a = DataArray(s)
+    ac = a.chunk(2)
+    assert ac.chunks == ((2, 2),)
+    assert isinstance(ac.data._meta, sparse.COO)
+    assert_identical(ac, a)
```
crusaderky: @shoyer After #3204, ``assert_identical(a, ac)`` does not work anymore - not even with the latest sparse git head

crusaderky:
Ready for final review and merge

xarray/tests/test_sparse.py
shoyer: I'm guessing this is due to https://github.com/dask/dask/issues/5259

crusaderky:
@shoyer NEP18 and type annotations feel like two groundbreaking changes to me. Shall we call the next version to 0.13.0 instead of just 0.12.4?

shoyer:
> @shoyer NEP18 and type annotations feel like two groundbreaking changes to me. Shall we call the next version to 0.13.0 instead of just 0.12.4?

Sounds pretty reasonable to me!

shoyer:
I don't think we're quite ready to release 0.13 -- NEP-18 support could probably use a bit more polish. But we're pretty close.

PR #3204, opened 2019-08-11
Remove duck_array_ops.as_like_arrays()

shoyer:
It has some questionable coercion logic that no longer seems to be necessary.

Not a user facing change.

### DIFF
--- a/xarray/tests/test_sparse.py
+++ b/xarray/tests/test_sparse.py
@@ -99,16 +99,13 @@ def test_variable_property(prop):
         (do("all"), False),
         (do("any"), False),
         (do("astype", dtype=int), True),
-        (do("broadcast_equals", make_xrvar({"x": 10, "y": 5})), False),
         (do("clip", min=0, max=1), True),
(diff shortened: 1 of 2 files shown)


shoyer:
I did have to disable a few of the sparse tests that entailed element-wise comparison between dense and sparse arrays. This currently isn't supported by sparse (but perhaps it should be): https://github.com/pydata/sparse/issues/270

xarray/tests/test_sparse.py
```diff
@@ -380,10 +390,18 @@ def test_dataarray_property(prop):
...
         (do("transpose"), True),
         # TODO
-        # isel_points
-        # sel_points
         # set_index
         # swap_dims
+        param(
+            do("broadcast_equals", make_xrvar({"x": 10, "y": 5})),
+            False,
+            marks=xfail(reason="https://github.com/pydata/sparse/issues/270"),
```
shoyer: This issue is fixed, so we should be able to re-enable these tests after the next sparse release.

PR #3207, opened 2019-08-12
Annotations for .data_vars() and .coords()

crusaderky:


xarray/core/coordinates.py
```diff
@@ -23,49 +34,58 @@
...
+    def __setitem__(self, key: Hashable, value: Any) -> None:
         self.update({key: value})
 
     @property
-    def indexes(self):
+    def _names(self) -> Set[Hashable]:
+        raise NotImplementedError()
+
+    @property
+    def dims(self) -> Union[Mapping[Hashable, int], Tuple[Hashable, ...]]:
```
max-sixty: Why the change here from `return self._data.dims`?

xarray/core/coordinates.py
```diff
@@ -23,49 +34,58 @@
...
-    def __setitem__(self, key, value):
+    def __getitem__(self, key: Hashable) -> "DataArray":
+        raise NotImplementedError()
+
+    def __setitem__(self, key: Hashable, value: Any) -> None:
         self.update({key: value})
 
     @property
-    def indexes(self):
+    def _names(self) -> Set[Hashable]:
```
max-sixty: Is this new? Should this be on the abstract class?

xarray/core/coordinates.py
```diff
@@ -259,19 +291,15 @@ def _update_coords(self, coords):
     def variables(self):
         return Frozen(self._data._coords)
 
-    def _to_dataset(self, shallow_copy=True):
```
max-sixty: I think removing this should be fine...

xarray/core/coordinates.py
crusaderky: It's a property invoked by the methods of the abstract class, defined in both the implementations.

xarray/core/coordinates.py
crusaderky: Because ``DataArray.dims`` is a ``Tuple[Hashable, ...]`` whereas ``Dataset.dims`` is a ``Mapping[Hashable, int]``

crusaderky: See below for the two implementations, which are identical in code but different in signature.
The method needs stay in the abstract class because other methods of the abstract class invoke it; this signature with NotImplementedError() guarantees that the methods of the abstract class can only use dims as the common denominator of the two output types, which is a ``Collection[Hashable]``.

xarray/core/coordinates.py
crusaderky: I searched and there's nothing invoking it across the whole codebase

max-sixty: Nice

xarray/core/coordinates.py
max-sixty: OK great

xarray/core/coordinates.py
max-sixty: Here, I'm not so sure - do all concrete classes need to define this? Generally we wouldn't define private methods on an abstract class (even if all currently concrete classes implement it), because the implementation would be up to the concrete class?

crusaderky: What I basically did was turn AbstractCoordinates from a mixin (that is, a class that invokes own methods that are only defined in a subclass) into a C++-style abstract class.
mypy, justifiably, doesn't like mixins - short of hacking your way through with ``# type`` tags.

This, in synthesis, is how it was before:
```
class C:
    def g(self) -> int:
        return self.f()


class D(C):
    def f(self) -> int:
        return 123
```
mypy correctly states:
```3: error: "C" has no attribute "f"```

max-sixty: Right - I agree this should either be an abstract class without implementations or a mixin with implementations - not a bit of each. It's a bit odd we call this class `Abstract` but then enforce a `_names` property.

That a nice upside of mypy - we get to find those things out

I agree we're in a reasonable local maximum here.

shoyer: Yes, thanks for fixing this up!

PR #3209, opened 2019-08-12
Add note about accessor instance creation

### LINKED ISSUE #3205: Accessors are recreated on every access
#### MCVE Code Sample

1. Create `test_accessor.py` in current directory with:

```python
import xarray as xr

@xr.register_dataarray_accessor('test')
class TestDataArrayAccessor(object):
    def __init__(self, obj):
        self._obj = obj
        print("DataArray accessor created")

@xr.register_dataset_accessor('test')
class TestDatasetAccessor(object):
    def __init__(self, obj):
        self._obj = obj
        print("Dataset accessor created")
```

2. Run the following code:

```python
import xarray as xr
import numpy as np
import test_accesor

ds = xr.Dataset({'a': xr.DataArray(np.array([1, 2, 3])), 'b': xr.DataArray(np.array([4, 5, 6]))})
ds.test
# Dataset accessor created <accessor created on first access - expected>

ds.test
# <no output - expected>

ds['a'].test
# DataArray accessor created <accessor created on first access - expected>

ds['a'].test
# DataArray accessor created <accessor created on second access - unexpected>

var = ds['a']
var.test
# DataArray accessor created <accessor created on direct instance access - expected/unexpected>

var.test
# <no output - expected>

```

#### Expected Output

Based on the xarray [accessor documentation](http://xarray.pydata.org/en/stable/internals.html#extending-xarray) I would have assumed that the accessor would stick around on the same DataArray object for the life of the data. My guess is that `Dataset.__getitem__` is recreating the `DataArray` every time from the underlying `Variable` object which means `ds['a'] is not ds['a']`?

#### Problem Description

I'm currently working on an accessor for a new package called `geoxarray` to address the issues talked about in #2288. One of the cases I'm trying to handle is a NetCDF file with CR standard `grid_mapping` variables. This means that the easiest way to set a CRS object for all variables in a Dataset is to have the Dataset accessor use the accessor of every DataArray (to cache a `_crs` property). However, with the way things are working doing something like the below won't work:

```python
ds.geo.apply_grid_mapping()
ds['Rad'].geo.crs  # this is current `None` because the DataArray was recreated
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 14:38:56) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
python-bits: 64
OS: Darwin
OS-release: 18.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.6.2

xarray: 0.12.3
pandas: 0.25.0
numpy: 1.17.0
scipy: None
netCDF4: 1.5.1.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.2.0
distributed: 2.2.0
matplotlib: None
cartopy: None
seaborn: None
numbagg: None
setuptools: 41.0.1
pip: 19.2.2
conda: None
pytest: None
IPython: 7.7.0
sphinx: None

</details>

djhoese:
This is an attempt at clarifying accessor docs to help with the confusion I had in #3205. I'm not sure how many other xarray doc pages use `.. note::` but it seemed useable here. I'm not sure this is worth the explanation or if it is worded the best, but thought this could start the discussion.

 - [ ] Closes #3205 
 - [ ] Tests added

### DIFF
--- a/doc/internals.rst
+++ b/doc/internals.rst
@@ -106,6 +106,13 @@ reasons:
    functionality that clearly identifies it as separate from built-in xarray
    methods.
 
+.. note::
+
+   Accessors are created once per DataArray and Dataset instance. New
+   instances, like those created from arithmetic operations or when accessing
+   a DataArray from a Dataset (ex. ``ds[var_name]``), will have new
+   accessors created.
+
 Back in an interactive IPython session, we can use these properties:
 
 .. ipython:: python


shoyer:
Thanks @djhoese, this is definitely worth noting!

PR #3210, opened 2019-08-13
sparse=True option for from_dataframe and from_series

### LINKED ISSUE #2139: From pandas to xarray without blowing up memory
I have a billion rows of data, but really it's just two categorical variables, time, lat, lon and some data variables.

Thinking it would somehow help me get the data into xarray, I created a five level pandas MultiIndex array out of the data, but thus far this has not been successful. xarray tries to create a product and that's just not going to work..

Trying to write a NetCDF file has presented its own issues, and I'm left wondering if there isn't a much simpler way to go about this?

### LINKED ISSUE #3206: DataFrame with MultiIndex -> xarray with sparse array
Now that we have preliminary support for [sparse](https://sparse.pydata.org/en/latest/) arrays in xarray, one really cool feature we could explore is creating sparse arrays from MultiIndexed pandas DataFrames.

Right now, xarray's methods for creating objects from pandas always create dense arrays, but the size of these dense arrays can get big really quickly if the MultiIndex is sparsely populated, e.g.,
```python
import pandas as pd
import numpy as np
import xarray
df = pd.DataFrame({
    'w': range(10),
    'x': list('abcdefghij'),
    'y': np.arange(0, 100, 10),
    'z': np.ones(10),
}).set_index(['w', 'x', 'y'])
print(xarray.Dataset.from_dataframe(df))
```
This length 10 DataFrame turned into a dense array with 1000 elements (only 10 of which are not NaN):
```
<xarray.Dataset>
Dimensions:  (w: 10, x: 10, y: 10)
Coordinates:
  * w        (w) int64 0 1 2 3 4 5 6 7 8 9
  * x        (x) object 'a' 'b' 'c' 'd' 'e' 'f' 'g' 'h' 'i' 'j'
  * y        (y) int64 0 10 20 30 40 50 60 70 80 90
Data variables:
    z        (w, x, y) float64 1.0 nan nan nan nan nan ... nan nan nan nan 1.0
```

We can imagine `xarray.Dataset.from_dataframe(df, sparse=True)` would make the same Dataset, but with sparse array (with a `NaN` fill value) instead of dense arrays.

Once sparse arrays work pretty well, this could actually obviate most of the use cases for `MultiIndex` in arrays. Arguably the model is quite a bit cleaner.

shoyer:
Fixes https://github.com/pydata/xarray/issues/3206

Example usage:

    In [3]: import pandas as pd
       ...: import numpy as np
       ...: import xarray
       ...: df = pd.DataFrame({
       ...:     'w': range(10),
       ...:     'x': list('abcdefghij'),
       ...:     'y': np.arange(0, 100, 10),
       ...:     'z': np.ones(10),
       ...: }).set_index(['w', 'x', 'y'])
       ...:

    In [4]: ds = xarray.Dataset.from_dataframe(df, sparse=True)

    In [5]: ds.z.data
    Out[5]: <COO: shape=(10, 10, 10), dtype=float64, nnz=10, fill_value=nan>

 - [x] Closes #3206, Closes #2139
 - [x] Tests added

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2295,7 +2295,7 @@ class DataArray(AbstractArray, DataWithCoords):
         return obj
 
     @classmethod
-    def from_series(cls, series: pd.Series) -> "DataArray":
+    def from_series(cls, series: pd.Series, sparse: bool = False) -> "DataArray":
         """Convert a pandas.Series into an xarray.DataArray.
 
         If the series's index is a MultiIndex, it will be expanded into a
@@ -2303,10 +2303,10 @@ class DataArray(AbstractArray, DataWithCoords):
         values with NaN). Thus this operation should be the inverse of the
         `to_series` method.
         """
-        # TODO: add a 'name' parameter
+        # TODO: add a 'name' parameter?
         name = series.name
         df = pd.DataFrame({name: series})
-        ds = Dataset.from_dataframe(df)
+        ds = Dataset.from_dataframe(df, sparse=sparse)
         return ds[name]
 
     def to_cdms2(self) -> "cdms2_Variable":
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2307,7 +2307,7 @@ class DataArray(AbstractArray, DataWithCoords):
         name = series.name
         df = pd.DataFrame({name: series})
         ds = Dataset.from_dataframe(df, sparse=sparse)
-        return ds[name]
+        return ds[name]  # type: ignore
 
     def to_cdms2(self) -> "cdms2_Variable":
         """Convert this array into a cdms2.Variable
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2302,12 +2302,19 @@ class DataArray(AbstractArray, DataWithCoords):
         tensor product of one-dimensional coordinates (filling in missing
         values with NaN). Thus this operation should be the inverse of the
         `to_series` method.
+
+        If sparse=True, creates a sparse array instead of a dense NumPy array.
+        Requires the pydata/sparse package.
+
+        See also
+        --------
+        xarray.Dataset.from_dataframe
         """
-        # TODO: add a 'name' parameter?
-        name = series.name
-        df = pd.DataFrame({name: series})
+        df = pd.DataFrame({_THIS_ARRAY: series})
         ds = Dataset.from_dataframe(df, sparse=sparse)
-        return ds[name]  # type: ignore
+        result = cast(DataArray, ds[_THIS_ARRAY])
+        result.name = series.name
+        return result
 
     def to_cdms2(self) -> "cdms2_Variable":
         """Convert this array into a cdms2.Variable
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -727,7 +727,7 @@ class DataArray(AbstractArray, DataWithCoords):
         else:
             if self.name is None:
                 raise ValueError(
-                    "cannot reset_coords with drop=False " "on an unnamed DataArrray"
+                    "cannot reset_coords with drop=False on an unnamed DataArrray"
                 )
             dataset[self.name] = self.variable
             return dataset
@@ -1468,9 +1468,7 @@ class DataArray(AbstractArray, DataWithCoords):
             This object, but with an additional dimension(s).
         """
         if isinstance(dim, int):
-            raise TypeError(
-                "dim should be hashable or sequence/mapping of " "hashables"
-            )
+            raise TypeError("dim should be hashable or sequence/mapping of hashables")
         elif isinstance(dim, Sequence) and not isinstance(dim, str):
             if len(dim) != len(set(dim)):
                 raise ValueError("dims should not contain duplicate values.")
@@ -2310,9 +2308,10 @@ class DataArray(AbstractArray, DataWithCoords):
         --------
         xarray.Dataset.from_dataframe
         """
-        df = pd.DataFrame({_THIS_ARRAY: series})
+        temp_name = "__temporary_name"
+        df = pd.DataFrame({temp_name: series})
         ds = Dataset.from_dataframe(df, sparse=sparse)
-        result = cast(DataArray, ds[_THIS_ARRAY])
+        result = cast(DataArray, ds[temp_name])
         result.name = series.name
         return result
 
@@ -2729,7 +2728,7 @@ class DataArray(AbstractArray, DataWithCoords):
         """
         if isinstance(other, Dataset):
             raise NotImplementedError(
-                "dot products are not yet supported " "with Dataset objects."
+                "dot products are not yet supported with Dataset objects."
             )
         if not isinstance(other, DataArray):
             raise TypeError("dot only operates on DataArrays.")
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -4072,7 +4072,9 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         """
         return self._to_dataframe(self.dims)
 
-    def _set_sparse_data_from_dataframe(self, dataframe, dims, shape):
+    def _set_sparse_data_from_dataframe(
+        self, dataframe: pd.DataFrame, dims: tuple, shape: Tuple[int, ...]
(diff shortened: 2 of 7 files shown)


max-sixty:
Amazing! This is going to allow way more usage of xarray for sparse representations of data currently confined to pandas

shoyer:
It would be great if someone could review this. Maybe @crusaderky ?

xarray/core/dataarray.py
```diff
@@ -2295,19 +2295,19 @@ def from_dict(cls, d: dict) -> "DataArray":
...
     @classmethod
-    def from_series(cls, series: pd.Series) -> "DataArray":
+    def from_series(cls, series: pd.Series, sparse: bool = False) -> "DataArray":
         """Convert a pandas.Series into an xarray.DataArray.
 
         If the series's index is a MultiIndex, it will be expanded into a
         tensor product of one-dimensional coordinates (filling in missing
         values with NaN). Thus this operation should be the inverse of the
         `to_series` method.
         """
```
crusaderky: Missing docstring for sparse parameter

xarray/core/dataarray.py
```diff
@@ -2295,19 +2295,19 @@ def from_dict(cls, d: dict) -> "DataArray":
...
-        # TODO: add a 'name' parameter
+        # TODO: add a 'name' parameter?
         name = series.name
         df = pd.DataFrame({name: series})
-        ds = Dataset.from_dataframe(df)
-        return ds[name]
+        ds = Dataset.from_dataframe(df, sparse=sparse)
+        return ds[name]  # type: ignore
```
crusaderky: When a Series has no name, Series.name is None. There is a ton of logic in xarray that treats None as a special case. And while ``hash(None)`` works and ``isinstance(None, Hashable)`` returns True, mypy doesn't consider None as a valid input for Hashable. So this is bound to trigger a mypy failure as soon as annotations are added to pandas, and a ``Dataset.__getitem__`` failure as soon as its implementation for any reason starts handling None explicitly.

Could you please change to
```python
# Don't use series.name, in case it is None
df = pd.DataFrame({"__temp": series})
ds = Dataset.from_dataframe(df, sparse=sparse)
out = ds["__temp"]
out.name = series.name
return out
```
The  ``# TODO: add a 'name' parameter?`` can go away IMHO.

xarray/core/dataset.py
```diff
@@ -4071,8 +4072,55 @@ def to_dataframe(self):
         """
         return self._to_dataframe(self.dims)
 
+    def _set_sparse_data_from_dataframe(self, dataframe, dims, shape):
```
crusaderky: Albeit boring, I find that putting type annotations in the internal functions helps a lot in weeding out subtle bugs.

xarray/core/dataset.py
```diff
@@ -4071,8 +4072,55 @@ def to_dataframe(self):
...
+        from sparse import COO
+
+        idx = dataframe.index
+        if isinstance(idx, pd.MultiIndex):
+            try:
+                codes = idx.codes
+            except AttributeError:
+                # deprecated since pandas 0.24
+                codes = idx.labels
+            coords = np.stack([np.asarray(code) for code in codes], axis=0)
```
crusaderky: You can change the whole paragraph to
```python
            try:
                coords = np.asarray(idx.codes)
            except AttributeError:
                # deprecated since pandas 0.24
                coords = np.asarray(idx.labels)

File: xarray/core/dataset.py [hunk 5]
```diff
@@ -4071,8 +4072,55 @@ def to_dataframe(self):
...
+
+        for name, series in dataframe.items():
+            values = np.asarray(series)
+
+            # In virtually all real use cases, the sparse array will now have
+            # missing values and needs a fill_value. For consistency, don't
+            # special case the rare exceptions (e.g., dtype=int without a
+            # MultiIndex).
+            dtype, fill_value = dtypes.maybe_promote(values.dtype)
+            values = np.asarray(values, dtype=dtype)
```
crusaderky: You can remove line 4092 and change 4098:4099 to
```python
dtype, fill_value = dtypes.maybe_promote(series.dtype)
values = np.asarray(series, dtype=dtype)``
```

xarray/tests/test_dataarray.py
```diff
@@ -3374,6 +3374,18 @@ def test_to_and_from_series(self):
         expected_da = self.dv.rename(None)
         assert_identical(expected_da, DataArray.from_series(actual).drop(["x", "y"]))
 
+    def test_from_series_sparse(self):
+        sparse = pytest.importorskip("sparse")
```
crusaderky: I find it would be more readable to write
```python
@requires_sparse
def test_from_series_sparse(self):
    import sparse
```

xarray/tests/test_dataarray.py
```diff
@@ -3374,6 +3374,18 @@ def test_to_and_from_series(self):
...
+
+        series = pd.Series([1, 2], index=[("a", 1), ("b", 2)])
+
+        actual_sparse = DataArray.from_series(series, sparse=True)
+        actual_dense = DataArray.from_series(series, sparse=True)
```
crusaderky: isn't it sparse=False?

xarray/tests/test_dataset.py
```diff
@@ -3674,6 +3674,27 @@ def test_to_and_from_dataframe(self):
         expected = pd.DataFrame([[]], index=idx)
         assert expected.equals(actual), (expected, actual)
 
+    def test_from_dataframe_sparse(self):
+        sparse = pytest.importorskip("sparse")
```
crusaderky: same @requires_sparse

xarray/tests/test_dataarray.py
crusaderky: Also it could use testing the output DataArray name when the input Series has and hasn't got a name.

xarray/core/dataarray.py
shoyer: Good catch. I switched to using the `_THIS_ARRAY` sentinel value which we already use elsewhere for this purpose.

xarray/core/dataset.py
shoyer: done

xarray/core/dataset.py
shoyer: We need to cast to numpy arrays first in case it's an extension array. Added a comment to explain.

xarray/core/dataset.py
shoyer: Pandas uses a weird `FrozenList` type for codes. I'd rather explicitly iterate over it than rely upon NumPy properly converting the whole thing at once.

xarray/tests/test_dataarray.py
shoyer: Oops, fixed.

There's already a case that checks `name=None` in the test above.

xarray/tests/test_dataarray.py
shoyer: Agreed (and changed here), though I'm not sure it's worth enforcing this in general since it goes slightly against the norms of pytest.

xarray/core/dataarray.py
```diff
@@ -2295,19 +2293,27 @@ def from_dict(cls, d: dict) -> "DataArray":
...
         """
-        # TODO: add a 'name' parameter
-        name = series.name
-        df = pd.DataFrame({name: series})
-        ds = Dataset.from_dataframe(df)
-        return ds[name]
+        temp_name = "__temporary_name"
+        df = pd.DataFrame({temp_name: series})
+        ds = Dataset.from_dataframe(df, sparse=sparse)
+        result = cast(DataArray, ds[temp_name])
```
crusaderky: I think I can get rid of the cast by changing ``Dataset.__getitem__``. Let's merge this one first though.

crusaderky:
Looks good to me

xarray/core/dataarray.py
crusaderky: ... or not. Unlike ``@functools.single_dispatch``, ``@typing.overload`` does not allow for a more specific specific signature followed by a more generic one :(

```python

    @overload
    def __getitem__(self, key: Hashable) -> "DataArray":
        ...

    @overload
    def __getitem__(self, key: Any) -> "Dataset":
        ...

    def __getitem__(self, key):
```
mypy complains:
```
xarray/core/dataset.py:1222: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader
```

Issue #2528, opened 2018-10-30
display_width doesn't apply to dask-backed arrays


rabernat:
The representation of dask-backed arrays in xarray's `__repr__` methods results in very long lines which often overflow the desired line width. Unfortunately, this can't be controlled or overridden with `xr.set_options(display_width=...)`. 

#### Code Sample, a copy-pastable example if possible

```python
import xarray as xr
xr.set_options(display_width=20)
ds = (xr.DataArray(range(100))
      .chunk({'dim_0': 10})
      .to_dataset(name='really_long_long_name'))
ds
```

```
<xarray.Dataset>
Dimensions:                (dim_0: 100)
Dimensions without coordinates: dim_0
Data variables:
    really_long_long_name  (dim_0) int64 dask.array<shape=(100,), chunksize=(10,)>
```
#### Problem description

[this should explain **why** the current behavior is a problem and why the expected output is a better solution.]

#### Expected Output

We need to decide how to abbreviate dask arrays with something more concise. I'm not sure the best way to do this. Maybe
```
   really_long_long_name  (dim_0) int64 dask chunks=(10,)
```

shoyer:
As of #3211, we no longer include the redundant `shape` in short reprs for dask arrays.

So I'm going to close this, though of course further improvements to the short repr would be welcome!

PR #3211, opened 2019-08-13
Array formatting fixes for sparse and NEP-18 arrays.

shoyer:
I also did a bit of cleanup (e.g., renaming methods) in xarray.core.formatting.

Sparse arrays were previously not shown in the Dataset repr:

    <xarray.Dataset>
    Dimensions:  (x: 4)
    Coordinates:
        y        (x) int64 ...
    Data variables:
        a        (x) float64 ..."""

Now they are:

    <xarray.Dataset>
    Dimensions:  (x: 4)
    Coordinates:
        y        (x) int64 <COO: shape=(4,), nnz=3, fill_value=0>
    Data variables:
        a        (x) float64 <COO: shape=(4,), nnz=4, fill_value=0.0>"""

 - [x] Tests added

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -10,7 +10,7 @@ import pandas as pd
 
 from .duck_array_ops import array_equiv
 from .options import OPTIONS
-from .pycompat import dask_array_type
+from .pycompat import dask_array_type, sparse_array_type
 
 try:
     from pandas.errors import OutOfBoundsDatetime
@@ -168,7 +168,7 @@ def format_items(x):
     return formatted
 
 
-def format_array_flat(array, max_width):
+def inline_ndarray_repr(array, max_width):
     """Return a formatted string for as many items in the flattened version of
     array that will fit within max_width characters.
     """
@@ -211,9 +211,41 @@ def format_array_flat(array, max_width):
     return pprint_str
 
 
-def summarize_variable(
-    name, var, col_width, show_values=True, marker=" ", max_width=None
-):
+def inline_dask_repr(array):
+    """Similar to dask.array.DataArray.__repr__, but without
+    redundant information that's already printed by the repr
+    function of the xarray wrapper.
+    """
+    assert isinstance(array, dask_array_type)
+    chunksize = tuple(c[0] for c in array.chunks)
+    return "dask.array<shape={}, chunksize={}>".format(array.shape, chunksize)
+
+
+def inline_sparse_repr(array):
+    """Similar to sparse.COO.__repr__, but without the redundant shape/dtype."""
+    assert isinstance(array, sparse_array_type)
+    return "<COO: shape={!s}, nnz={:d}, fill_value={!s}>".format(
+        array.shape, array.nnz, array.fill_value
+    )
+
+
+def inline_variable_array_repr(var, max_width):
+    """Build a one-line summary of a variable's data."""
+    if var._in_memory:
+        return inline_ndarray_repr(var, max_width)
+    elif isinstance(var._data, dask_array_type):
+        return inline_dask_repr(var.data)
+    elif isinstance(var._data, sparse_array_type):
+        return inline_sparse_repr(var.data)
+    elif hasattr(var._data, "__array_function__"):
+        return maybe_truncate(repr(var._data).replace("\n", " "), max_width)
+    else:
+        # internal xarray array type
+        return "..."
+
+
+def summarize_variable(name, var, col_width, marker=" ", max_width=None):
+    """Summarize a variable in one line, e.g., for the Dataset.__repr__."""
     if max_width is None:
         max_width = OPTIONS["display_width"]
     first_col = pretty_print("  {} {} ".format(marker, name), col_width)
@@ -222,12 +254,9 @@ def summarize_variable(
     else:
         dims_str = ""
     front_str = "{}{}{} ".format(first_col, dims_str, var.dtype)
-    if show_values:
-        values_str = format_array_flat(var, max_width - len(front_str))
-    elif isinstance(var._data, dask_array_type):
-        values_str = short_dask_repr(var, show_dtype=False)
-    else:
-        values_str = "..."
+
+    values_width = max_width - len(front_str)
+    values_str = inline_variable_array_repr(var, values_width)
 
     return front_str + values_str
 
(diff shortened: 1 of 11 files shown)


xarray/tests/test_dataset.py
```diff
@@ -178,24 +178,22 @@ def test_repr(self):
         data = create_test_data(seed=123)
         data.attrs["foo"] = "bar"
         # need to insert str dtype at runtime to handle both Python 2 & 3
```
max-sixty:
```suggestion
```

shoyer: we still need  this for supporting different endiannesses

xarray/core/formatting.py
```diff
@@ -211,9 +211,41 @@ def format_array_flat(array, max_width):
...
-    name, var, col_width, show_values=True, marker=" ", max_width=None
-):
+def inline_dask_repr(array):
+    """Similar to dask.array.DataArray.__repr__, but without
+    redundant information that's already printed by the repr
+    function of the xarray wrapper.
+    """
+    assert isinstance(array, dask_array_type), array
+    chunksize = tuple(c[0] for c in array.chunks)
+    return "dask.array<shape={}, chunksize={}>".format(array.shape, chunksize)
```
dcherian: Should we remove shape here and in the sparse repr? https://github.com/pydata/xarray/issues/2528#issuecomment-434456201

Also what about adding support for `max_width` (and closing #2528)?

shoyer: Sure, we could drop `shape` if we think it's redundant in general, or potentially only drop it if the repr is already too long to fit into `max_width`.

I don't know a good way to truncate to `max_width` in general, beyond blinding truncating the repr to a certain maximum width and adding `...`.

crusaderky: +1 for dropping shape

xarray/tests/test_dask.py
```diff
@@ -149,8 +149,8 @@ def test_binary_op(self):
     def test_repr(self):
         expected = dedent(
             """\
-        <xarray.Variable (x: 4, y: 6)>
-        dask.array<shape=(4, 6), dtype=float64, chunksize=(2, 2)>"""
+            <xarray.Variable (x: 4, y: 6)>
+            dask.array<array, shape=(4, 6), dtype=float64, chunksize=(2, 2)>"""
```
crusaderky: Was it a conscious decision to add the array name? Shouldn't it be coherent with inline_dask_repr?

xarray/tests/test_dask.py
```diff
@@ -499,11 +499,11 @@ def test_dataarray_repr(self):
...
-        <xarray.DataArray 'data' (x: 1)>
-        dask.array<shape=(1,), dtype=int64, chunksize=(1,)>
-        Coordinates:
-            y        (x) int64 dask.array<shape=(1,), chunksize=(1,)>
-        Dimensions without coordinates: x"""
+            <xarray.DataArray 'data' (x: 1)>
+            dask.array<data, shape=(1,), dtype=int64, chunksize=(1,)>
+            Coordinates:
+                y        (x) int64 dask.array<shape=(1,), chunksize=(1,)>
+            Dimensions without coordinates: x"""
```
crusaderky: Why would I want to know the dask name for the data, but not for the coords? for me it's either both or neither.
I'm in favour of neither because actual real-life dask names can get VERY long!

crusaderky:
Possible improvements:

- Take a slice from the front of sparse arrays, densify it, and show the array contents just like you can see them for dense arrays
- Represent dask arrays differently when they have a sparse _meta

xarray/tests/test_dask.py
shoyer: In Variable and DataArray, we are now using the repr of the underlying data directly (which includes `name` for dask arrays), rather than additional processing with xarray. That seemed like an improvement to me, given that we aren't as strictly space constrainted as we are in the inline Dataset repr.

shoyer:
> * Take a slice from the front of sparse arrays, densify it, and show the array contents just like you can see them for dense arrays

This is a good idea, but it will require slightly more complex logic to pull items only from the front of an array. I don't have a lot of time to iterate on this right now, so I'm going to leave it for later.

> Represent dask arrays differently when they have a sparse _meta

I have now added in `meta` to the repr for dask arrays. PTAL.

xarray/tests/test_dask.py
```diff
@@ -491,6 +491,7 @@ def test_dot(self):
         lazy = self.lazy_array.dot(self.lazy_array[0])
         self.assertLazyAndAllClose(eager, lazy)
 
+    @pytest.mark.skipif(LooseVersion(dask.__version__) < "2.0", reason="needs meta")
```
crusaderky: if anybody changes ``inline_dask_repr`` to expect _meta to always be there, dask < 2.0 will start crashing and no test will notice.

xarray/tests/test_dask.py
```diff
@@ -502,12 +503,13 @@ def test_dataarray_repr(self):
...
             dask.array<data, shape=(1,), dtype=int64, chunksize=(1,)>
             Coordinates:
-                y        (x) int64 dask.array<shape=(1,), chunksize=(1,)>
+                y        (x) int64 dask.array<chunksize=(1,), meta=np.ndarray>
             Dimensions without coordinates: x"""
         )
         assert expected == repr(a)
         assert kernel_call_count == 0
 
+    @pytest.mark.skipif(LooseVersion(dask.__version__) < "2.0", reason="needs meta")
```
crusaderky: same

xarray/tests/test_sparse.py
```diff
@@ -707,7 +707,7 @@ def test_dataarray_repr(self):
             <xarray.DataArray (x: 4)>
             <COO: shape=(4,), dtype=float64, nnz=4, fill_value=0.0>
             Coordinates:
-                y        (x) int64 <COO: shape=(4,), nnz=3, fill_value=0>
+                y        (x) int64 <COO: nnz=3, fill_value=0>
```
crusaderky: isn't it sparse.COO now?

shoyer: I was sticking with a simplified version of the original repr here, only putting `sparse.COO` inside `meta`.

xarray/tests/test_dask.py
shoyer: added a test for this, though I'm if we have a CI configuration currently that tests old versions of dask (that would be good to add)

crusaderky:
Re legacy dask: #3222

PR #3217, opened 2019-08-14
Refactor concat tests.

dcherian:
small cleanup in `test_concat.py`

### DIFF
--- a/xarray/tests/test_concat.py
+++ b/xarray/tests/test_concat.py
@@ -18,61 +18,62 @@ from .test_dataset import create_test_data
 
 
 class TestConcatDataset:
-    def test_concat(self):
-        # TODO: simplify and split this test case
+    @pytest.fixture(autouse=True)
+    def setUp(self):
+        self.data = create_test_data().drop_dims("dim3")
 
-        # drop the third dimension to keep things relatively understandable
-        data = create_test_data()
-        for k in list(data.variables):
-            if "dim3" in data[k].dims:
-                del data[k]
+    def rectify_dim_order(self, dataset):
+        # return a new dataset with all variable dimensions transposed into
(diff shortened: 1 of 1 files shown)


xarray/tests/test_concat.py
```diff
@@ -18,61 +18,62 @@
...
-            # return a new dataset with all variable dimensions transposed into
-            # the order in which they are found in `data`
-            return Dataset(
-                {k: v.transpose(*data[k].dims) for k, v in dataset.data_vars.items()},
-                dataset.coords,
-                attrs=dataset.attrs,
-            )
+    @pytest.fixture(autouse=True)
+    def setUp(self):
+        self.data = create_test_data().drop_dims("dim3")
```
max-sixty: Obv no need to change but a hopefully helpful comment - this could very easily be a fixture called `data`, and then no need for using instance attributes

dcherian: How would I do that?

max-sixty:
```python

    @pytest.fixture
    def data(self):
        return create_test_data().drop_dims("dim3")
```

and then in the methods, e.g.

```python
    def test_concat(self, data):  # <- added 

        # data = self.data  <- no longer needed
        split_data = [data.isel(dim1=slice(3)), data.isel(dim1=slice(3, None))]
        assert_identical(data, concat(split_data, "dim1"))
```

dcherian: OK. done.

max-sixty:
Oh nice! I didn't mean you should to change it for this PR, hope that was clear. Thank you!

dcherian:
Yeah it was. I just thought it was a good idea :)

PR #3228, opened 2019-08-19
Add examples of set_index and provide more clear error message

### LINKED ISSUE #3176: DataArray.set_index throws error on documented input
#### Problem Description

[Docs for `DataArray.set_index`](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.set_index.html) describe the main `indexes` argument as:

> Mapping from names matching dimensions and values given by (lists of) the names of existing coordinates or variables to set as new (multi-)index.

This suggests that one can set a `DataArray` instance's coordinates by passing in a dimension and a list-like object of coordinates.

#### MCVE

```python
In [1]: import numpy as np

In [2]: import xarray as xr

In [3]: arr = xr.DataArray(data=np.ones((2, 3)), dims=['x', 'y'])

In [4]: arr.dims
Out[4]: ('x', 'y')

In [5]: arr.set_index({'x': range(2)})
KeyError   
...
    144         for n in var_names:
--> 145             var = variables[n]
    146             if (current_index_variable is not None and
    147                     var.dims != current_index_variable.dims):

KeyError: 0
```

At first, I thought it might be because `coords` and `_coords` were not being set in this case:

```python
In [18]: arr.coords
Out[18]: 
Coordinates:
    *empty*

In [19]: arr._coords
Out[19]: OrderedDict()
```

but even if I set the coordinates first and then try to re-index, it fails:

```python
In [20]: arr = xr.DataArray(data=np.ones((2, 3)), dims=['x', 'y'], coords={'x': range(2), 'y': range(3)})
In [21]: arr.set_index({'x': ['a', 'b', 'c']})
...
    144         for n in var_names:
--> 145             var = variables[n]
    146             if (current_index_variable is not None and
    147                     var.dims != current_index_variable.dims):
```

#### Expected Output

I expect my MCVE to work based on the documentation.

#### Problem Solution

My guess is that the issue is Xarray is using the `merge_indexes` function ([see here](https://github.com/pydata/xarray/blob/master/xarray/core/dataarray.py#L1444)) from the `Dataset` module, and there is no concept of a `variable` in a `DataArray`.

gwgundersen:
- [x] Closes #3176 by adding examples of `set_index` and raising an error with a more clear error message.
- [x] Added unit test for error message.
- [x] Passes `black . ` and `flake8`.
- [x] Documented in `whats-new.rst`.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -65,8 +65,14 @@ Enhancements
   ``append_dim`` is set, as it will automatically be set to ``'a'`` internally.
   By `David Brochart <https://github.com/davidbrochart>`_.
 
-- :py:meth:`~xarray.Dataset.drop` now supports keyword arguments; dropping index labels by specifying both ``dim`` and ``labels`` is deprecated (:issue:`2910`).
+- :py:meth:`~xarray.Dataset.drop` now supports keyword arguments; dropping index
+  labels by specifying both ``dim`` and ``labels`` is deprecated (:issue:`2910`).
   By `Gregory Gundersen <https://github.com/gwgundersen/>`_.
+
+- Added examples of :py:meth:`Dataset.set_index` and
+  :py:meth:`DataArray.set_index`, as well are more specific error messages
+  when the user passes invalid arguments (:issue:`3176`).
+  By `Gregory Gundersen <https://github.com/gwgundersen>`_.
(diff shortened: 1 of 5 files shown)


max-sixty:
This is great! Thanks @gwgundersen !

PR #3233, opened 2019-08-19
Drop keyword support: small fixes

gwgundersen:
- Addresses issue https://github.com/pydata/xarray/issues/2910 and completes PR https://github.com/pydata/xarray/pull/3128.
 - Passes `black .` and `flake8`.
 - Documented in `whats-new.rst` and `indexing.rst`.

### DIFF
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -236,7 +236,6 @@ The :py:meth:`~xarray.Dataset.drop` method returns a new object with the listed
 index labels along a dimension dropped:
 
 .. ipython:: python
-    :okwarning:
 
     ds.drop(space=['IN', 'IL'])
 
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -238,7 +238,7 @@ index labels along a dimension dropped:
 .. ipython:: python
     :okwarning:
 
-    ds.drop(['IN', 'IL'], dim='space')
+    ds.drop(space=['IN', 'IL'])
 
 ``drop`` is both a ``Dataset`` and ``DataArray`` method.
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -55,7 +55,6 @@ from .common import (
     _contains_datetime_like_objects,
 )
 from .coordinates import (
-    DataArrayCoordinates,
     DatasetCoordinates,
     LevelCoordinatesSource,
     assert_coordinate_consistent,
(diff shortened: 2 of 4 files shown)


gwgundersen:
Strange. This passes all my tests locally.

doc/indexing.rst
```diff
@@ -238,7 +238,7 @@ index labels along a dimension dropped:
 .. ipython:: python
     :okwarning:
```
max-sixty: We can remove this now

max-sixty:
The failure is b/c of an upstream package; we can ignore

xarray/core/dataset.py
```diff
@@ -3558,6 +3557,13 @@ def drop(  # noqa: F811
                 ds = ds._drop_labels(labels, dim, errors=errors)
             return ds
         elif dim is None:
+            if isinstance(labels, DataArrayCoordinates):
```
shoyer: I think this should probably be even more generic:
```suggestion
            if is_dict_like(labels) and not isinstance(labels, dict):
```
and probably moved up to ensure it gets triggered in all situations, not just if `dim is None`.

xarray/core/dataset.py
```diff
@@ -3558,6 +3557,13 @@ def drop(  # noqa: F811
...
+                warnings.warn(
+                    "dropping coordinates using a DataArrayCoordinates object "
+                    "is deprecated; use drop_vars or a list of coordinates",
+                    DeprecationWarning,
```
shoyer: `FutureWarning` is more appropriate here, since we will probably break existing code in the future by giving this different behavior:
```suggestion
                    FutureWarning,
```

shoyer: It also would be great to add a test that verifies that the warning is raised :)

gwgundersen: Should the second warning for `list`-like `labels` also be a `FutureWarning`?

shoyer: I think `DeprecationWarning` might be more appropriate there, since continuing to rely on that will result in errors rather than different behavior.

xarray/core/dataset.py
gwgundersen: Sure, makes sense. I tried to write an error message that captured the generic case as well.

gwgundersen:
@shoyer, I've addressed your comments by adding more helpful warnings and testing that said warnings are raised properly. I've also moved both warnings to the top of the function as I think it's a bit easier to reason about the code branches this way.

PR #3234, opened 2019-08-20
Explicitly keep track of indexes with merging

shoyer:
Part of the explicit indexes refactor (https://github.com/pydata/xarray/issues/1603)

No user facing changes.


shoyer:
Test failure should be fixed by https://github.com/pydata/xarray/pull/3235

shoyer:
@pydata/xarray if anyone has time, it would be great to get a review here.

(I think this is pretty low risk, so I will probably self merge this in a week or so if there aren't any reviews)

max-sixty:
I had a read through—looks good, better design (and lots of nice idioms!). 
But I confess that I don't know the code well enough to make a more helpful judgement, and while I'd normally approach from the tests, no changes to those.

shoyer:
@max-sixty thanks for taking a look!

I agree, this was hard to review without test changes.

PR #3235, opened 2019-08-20
Fix xarray's test suite with the dask master

shoyer:
We shouldn't be checking the details of dask's repr.

 - [x] Tests added

### DIFF
--- a/xarray/tests/test_dask.py
+++ b/xarray/tests/test_dask.py
@@ -150,7 +150,9 @@ class TestVariable(DaskTestCase):
         expected = dedent(
             """\
             <xarray.Variable (x: 4, y: 6)>
-            dask.array<array, shape=(4, 6), dtype=float64, chunksize=(2, 2)>"""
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks

xarray/tests/test_dask.py
```diff
@@ -150,7 +150,9 @@ def test_repr(self):
         expected = dedent(
             """\
             <xarray.Variable (x: 4, y: 6)>
-            dask.array<array, shape=(4, 6), dtype=float64, chunksize=(2, 2)>"""
+            {!r}""".format(
+                self.lazy_var.data
+            )
```
max-sixty: If these get any more complex, I've had great results using https://gitlab.com/uweschmitt/pytest-regtest - a snapshot / expect test framework - for string outputs that require checking but not explicitly constructing.

shoyer:
Merging this now to keep our tests green...

PR #3238, opened 2019-08-22
tests for arrays with units

keewis:
This PR adds tests that verify that xarray properly wraps arrays with units.

       and `api.rst` for new API

They are currently restricted to using `pint` but may easily be extended to
use other unit libraries. However, most of the libraries I know use
subclassing instead of wrapping, so using them together with `dask` might
not be ideal as unit conflicts will only be detected on `compute()`.

As a reference for myself, here are the operations I plan on adding tests for:
 * top-level functions:
   - [x] replication: `full_like`, `ones_like`, `zeros_like`
 * dataarray methods
   - [x] creating: `DataArray()`
   - [x] representation: `__repr__`, `__str__`
   - [x] content manipulation: `assign_coords`, `assign_attrs`, `pipe`, `rename`, `swap_dims`, `expand_dims`, `drop`, `reset_coords`, `copy`, `astype`, `item`
   - [x] indexing: `loc`, `isel`, `sel`, `squeeze`, `interp`, `interp_like`, `reindex`, `reindex_like`
   - [x] missing value handling: `isnull`, `notnull`, `combine_first`, `count`, `dropna`, `fillna`, `ffill`, `bfill`, `interpolate_na`, `where`, `isin`
   - [x] comparisons: `equals`, `identical`, `broadcast_equals`
   - [x] aggregation: `all`, `any`, `argmax`, `argmin`, `max`, `mean`, `median`, `min`, `prod`, `sum`, `std`, `var`
   - [x] ndarray methods: `argsort`, `clip`, `conj`, `conjugate`, `imag`, `searchsorted`, `round`, `real`,     `T`, `cumsum`, `cumprod`, `rank`
   - [x] reshaping and reorganizing: `transpose`, `stack`, `unstack`, `set_index`, `reset_index`,     `reorder_levels`, `shift`, `roll`, `sortby`, `to_unstacked_dataset`
   - [x] arithmetic: unary (-, abs, round), binary (+, *, @)
   - [x] ufuncs (univariate: `sin`, `cos`,  bivariate: `minimum`, `maximum`)
   - [x] grouped operations: `assign_coords`, `first`, `last`, `fillna`, `where`, `quantile`
   - [x] computation: `reduce`, `groupby`, `groupby_bins`, `rolling`, `rolling_exp`, `coarsen`, `resample`, `diff`, `dot`, `quantile`, `differentiate`, `integrate`

The parts of the API that are currently not covered by this PR are:
* top level:
   - combining: `align`, `concat`, `merge`, `auto_combine`, `combine_by_coords`, `combine_nested`
   - computation: `broadcast`, `dot`, `apply_ufunc`
* dataarray:
   - computation: `get_axis_num`
* dataset

From what is already done, `mypy .` fails because of `pint` (this the same as with `quantities` in #2956). What would I do to fix that?


max-sixty:
Great! This is going to be exciting. Ping here for any questions!

> From what is already done, `mypy .` fails because of `pint` (this the same as with `quantities` in #2956). What would I do to fix that?

V easy - add to `setup.cfg` like other libraries listed there

jthielen:
@keewis: In case it helps, I've added a bunch of additional `__array_function__` implementations in pint in https://github.com/andrewgsavage/pint/pull/6, which will hopefully get merged soon into the main https://github.com/hgrecco/pint/pull/764 PR. If through these tests with xarray you find something missing on pint's end, let me know and I'd be glad to help.

keewis:
@jthielen: thanks, I will look at these when I have the time.

Could you take a look at the failing test for the `DataArray`-constructor? It fails because `pint` emits a `UnitStrippedWarning` on calling `hasattr(data, "__array_function__")`, which I believe should not happen (correct me if I'm wrong, though).

jthielen:
I noticed you have `@pytest.mark.filterwarnings("error::pint.errors.UnitStrippedWarning")`, which I believe is what is raising that warning to the error level. What happens when you remove that?

As to why the warning is happening in the first place, I think that is because the `__array_function__` implementation in pint right now only exists in a PR. So, when xarray looks for an `__array_function__` attribute on the pint Quantity right now, it falls back to this:

https://github.com/hgrecco/pint/blob/2afdc4bf5c5727ed9cef6fdaccb00b88813c4a24/pint/quantity.py#L1438-L1449

Hence, the `UnitStrippedWarning`. However, `hasattr(data, "__array_function__")` should still work, albeit for the wrong reason since pint doesn't yet have a true `__array_function__` method.

keewis:
ah, okay, that makes sense. I will leave the `filterwarning` since I believe the warning should not be raised, but mark that test as `xfail` until a released pint version implements `__array_function__`.

keewis:
@jthielen: I tried my tests (those that are currently written) with your PR to hgrecco/pint#764. From the ones I have been using, `np.prod`, `np.all` and `np.any` (and also `__matmul__`, but you seem to be working on that) are missing. The question is, how many of these should be supported?
Also, there might be better ways to compare quantities, but if not: would it be possible to implement `np.allclose`? I have been using a workaround, but it seems pretty fragile to me.
You may be aware of this, but `np.minimum` and `np.maximum` and probably others similar to these are working strangely, even with plain pint: if both arguments are quantities, equal units are required, but the result is not a quantity. It correctly fails if the second argument is not a quantity, but works silently if only the second argument is a quantity (still, the result is a plain array).

As I'm not too content with it: could someone have a look at the way they are written?  If there are any serious problems, it might be easier to correct them now and I can keep them in mind for new tests.

Lastly, a status report: Assuming there are no bugs in my tests and using andrewgsavage/pint#6, there are some issues with `np.argmax`/`np.argmin`, `np.median`, `np.sum` (and probably `np.prod`, but that is currently hidden). Indexing (`sel`, `isel` and `loc`) seems to work except if a single value (e.g. `1 * ureg.m`) / index is passed.

jthielen:
Thank you for the update! Here are responses to each issue brought up:

`np.prod` (along with `np.product` and `np.nanprod`) was not implemented yet since it seems non-trivial to determine the number of unit multiplications that occur given both `axis` and `where` arguments in a performant way. I can take a closer look at it though, and hopefully will be able to have an implementation soon. (If anyone has any ideas on this, let me know!)

I passed up `np.all` and `np.any` since I associated those with boolean arrays, rather than arrays with units. However, since it sounds like they are needed here, I can add those to the list. (Looking at it though, behavior with non-multiplicative units is unclear right now...see https://github.com/hgrecco/pint/issues/866)

`np.allclose` should be easy to implement, and I agree it would be nice to not have to have that workaround.

`np.maximum` and `np.minimum` appear to be routed through `__array_ufunc__` instead of `__array_function__`. I wasn't aware of that issue until you pointed it out, but looking into the implementation, "maximum" and "minimum" (and any other ufunc not explicitly listed by pint in `BaseQuantity.__set_units`, `BaseQuantity.__copy_units`, or `BaseQuantity.__prod_units`) will be behave in this way since checking input compatibility but not wrapping output is the fallback behavior. This seems like a bug to me that it doesn't raise a `UnitStrippedWarning`, and at the very least, it is inconsistent with the fail-to-NotImplemented behavior of the `__array_function__` implementation. I'll bring up this point at https://github.com/hgrecco/pint/pull/764.

`np.argmin`/`np.argmax`/`np.sum`: there seems to be a lot going on with these at least when I ran the tests locally. First, when called on `data_array`, they appear to be going deep into `xarray/core/duck_array_ops.py` and `xarray/core/nanops.py` rather than ever calling `np.nanargmin`/`np.nanargmax`/`np.nansum` on `data_array.data`. Second, even if that is the proper pathway, it ends up failing because it ends up calling `np.result_type` on mixed ndarray and `QuantitySequence` arguments. This is still unresolved on pint's end (see https://github.com/hgrecco/pint/pull/764#issuecomment-523272038 for the specific issue, https://github.com/hgrecco/pint/issues/845 for the more general discussion). All this also only happens on float dtype...it works just fine with int dtype, which is puzzling to me.

`np.median`: This one has been driving me crazy. In my environment, when running in pytest, it is erroring out because somewhere `np.median` being called on `data_array` is trying to cast the result (a pint QuantityScalar) to a `float`/`int`, which is not allowed by pint when the quantity is not dimensionless. But, I have no idea why it is ending up there. However, when replicating the test outside of pytest, it doesn't error, but instead the result is missing the magnitude (with a UnitStrippedWarning raised when going through the `__array_struct__` attribute on the quantity. `data_array.median()`, however, works just fine. I'm not really sure what to do here.

For indexing, when you say "working," would you be able to clarify what your expected behavior is for indexing? Based on https://github.com/pydata/xarray/issues/525#issuecomment-514880353 and the preceding discussion, right now indices will have units stripped, so for me at least, I would expect any attempt at unit-aware indexing to either not work or raise a `UnitStrippedWarning`.

keewis:
**`np.prod`**: seems fair, and I don't especially need that function as I have been copying it's use from `test_sparse.py` (though I realize now that there only the method is used). However, using `array.prod()` works on quantities, so I would expect `np.prod` to delegate to it. Same with `any()` and `all()`.

`np.median`: this also happens in my environment, so you can rule out the environment. Calling it on quantities instead of data arrays works, so the issue is within xarray. Interestingly, outside of `pytest`:
```python
>>> xr.DataArray(data=np.arange(10).astype(int) * ureg.m).median()
<xarray.DataArray ()>
<Quantity(4.5, 'meter')>
>>> xr.DataArray(data=np.arange(10).astype(int) * ureg.m).median()
<xarray.DataArray ()>
array(4.5)
```
which seems to be a manifestation of #3241 (which is the reason I now always try different dtypes). I guess that one is not really fixed?
Also, would it be better to have a test to check the methods in addition to the numpy functions?

 **Indexing**: when I wrote "indexing works" I meant that my tests for `sel()`, `isel()` and `loc` mostly pass (except from the "calling `int` / `float` on scalar results" issue). So when calling `sel()` and `loc` with either no units, dimensionless or incompatible units (but existing magnitudes), a `KeyError` is raised. In the case of both compatible units and existing magnitudes the results are also correct, but scalars trigger the error above.

keewis:
It seems the reason for both the `np.median` weirdness and the indexing unexpectedly passing is the module level `filterwarnings` mark. I'm not sure why that happens, but for now I'll remove the module level mark. Also, that means that now all aggregation functions fail for data arrays, so I'll definitely have to add method tests.

jthielen:
After digging a bit more into `np.prod` based on your comment, the situation is more complicated than I thought it was. I forgot that `array.prod()` has an implementation because pint has a ufunc implementation for it (so, actually `np.prod(array)` gives a result by way of `__array_ufunc__` before https://github.com/hgrecco/pint/pull/764, but it no longer functions that way when `__array_function__` is used). I hesitate to delegate to that implementation though, since it gives incorrect units when the `axis` and/or `where` arguments are used (see https://github.com/hgrecco/pint/issues/867).

`array.any()` and `array.all()` work now because they fall back to `.any()` and `.all()` on the magnitude; they aren't explicitly handled (or tested) by pint. Falling back to the magnitude here seems mostly fine, [except possibly for offset units](https://github.com/hgrecco/pint/issues/867). Once a decision is made about the expected behavior in that case, I can add them to the `__array_function__` list and add tests in pint.

Based on your tests and examples, I would agree with you that https://github.com/pydata/xarray/issues/3241 isn't really fixed. Since I would presume that both method calls and numpy function calls are supported APIs, yes, I think it would be best to test it both ways (but, please correct me if I'm wrong about that).

Thank you for clarifying about the tests. For some reason, I'm unfortunately getting results more consistent with the stripped unit behavior, rather than what your test results are showing. For example:

```python
import xarray as xr
import numpy as np
import pint

unit_registry = pint.UnitRegistry()

array = np.linspace(5, 10, 20).astype(int) * unit_registry.m
x = np.arange(len(array)) * unit_registry.s
data_array = xr.DataArray(data=array, coords={"x": x}, dims=["x"])

print(data_array.sel(x = [15, 16] * unit_registry.volts))
```
outputs
```
<xarray.DataArray (x: 2)>
<Quantity([8 9], 'meter')>
Coordinates:
  * x        (x) int64 15 16
```
whereas it should be giving a `KeyError` by your tests, if I'm interpreting them correctly. What happens in your environment when you run the above snippet?

keewis:
After digging a little bit more, I think the reason why `filterwarnings()` made me think the indexing tests pass is that when a `UnitStrippedWarning` is raised (mostly after accessing `__array_struct__`), the filter replaces it with a `DimensionalityError`, complaining that it can't convert to `dimensionless`. This confuses me: when accessing an attribute starting with `__array_`, `pytest` correctly points out the raised `UnitStrippedWarning`.

Also, it seems that because `DimensionalityError` inherits from `ValueError`, the type of the parameter `index` of `convert_label_indexer` is changed from `Int64Index` to `Index`, which does not cast to `ndarray`, thus keeping the unit information and causing `sel()` correctly raise a `KeyError`. I can reproduce both behaviors using
```python
import numpy as np
import pytest
import xarray as xr
from xarray.core.npcompat import IS_NEP18_ACTIVE

ureg = pint.UnitRegistry()

@pytest.mark.xfail
def test_indexing_fail():
    array = np.linspace(1, 2, 20).astype(float) * ureg.m
    x = np.arange(len(array)) * ureg.s
    data_array = xr.DataArray(data=array, coords={"x": x}, dims=["x"])
    with pytest.raises(KeyError):
        data_array.sel(x=np.array([13, 5, 9]) * ureg.m)

@pytest.mark.filterwarnings("error:::pint[.*]")
def test_indexing_pass():
    array = np.linspace(1, 2, 20).astype(float)	* ureg.m
    x = np.arange(len(array)) *	ureg.s
    data_array = xr.DataArray(data=array, coords={"x": x}, dims=["x"])
    with pytest.raises(KeyError):
        data_array.sel(x=np.array([13, 5, 9]) * ureg.m)
```

keewis:
I tried adding tests for `zeros_like`, `ones_like` and `full_like`, but I'm not sure what the expected result should look like, at least for the last one. I would assume `zeros_like` and `ones_like` return a quantity with the same unit as `other`, but what is the unit for `full_like`? The unit of `other`, or the unit of `fill_value`? Also, what happens if either of them has no unit, or both have units that are incompatible with each other?

jthielen:
@keewis In https://github.com/andrewgsavage/pint/pull/6, I implemented `zeros_like`, `ones_like` and `empty_like` to return base ndarrays rather than quantities, and `full_like` to have the unit of `fill_value`. This seemed like the most sensible behavior to me, since it avoids the ambiguities you mention in the `full_like` case if they were based on the unit of `other`, and for many of my own use cases, I've often wanted a zeros/ones array with a different unit than the quantity whose shape I am basing it off of.

Does this behavior seem reasonable to you? Also, would this be something that should be cleared up with an issue on pint's end?

keewis:
I like your approach because it avoids all those issues, but stripping the units might be a surprise. Then again, what does a function named `*_like` promise? From the docs, it seems that only shape and dtype are guaranteed, so the question is whether or not the unit is part of the data or part of the type?

keewis:
About the indexing: this might have been accidental (or a bug in pandas) and is probably a bad idea, but do you think it might be worth to somehow actively trigger this behavior in `pd.Index` for pint arrays (and possibly others) until the index refactoring from https://github.com/pydata/xarray/issues/1603#issuecomment-511126208 is done?

jthielen:
@keewis My inclination is to think of the units as part of the data, and that, for example, `zeros_like` returning an array of bare zeros is reasonably intuitive. But since this is likely not everyone's perspective, I think raising an issue on pint's end would be good.

I'm not sure about the indexing behavior. From the sounds of your prior comment, it works because it functions as an "object-type" index? If so, it may cause a decent hit on performance. I'd definitely want to hear others' thoughts on it too.

keewis:
yeah, well, I can create that index outside of pytest with
```python
pd.Index(list(np.arange(10) * ureg.m))
```
so the performance is that of an index on a python list. How that happens is beyond me, though.

keewis:
I'm not sure whether that belongs into hgrecco/pint#882, so I'll ask this here: In the case of `full_like` with `other` being a `ndarray` and `fill_value` having a unit, numpy complains about `np.copyto` not being implemented for this type of operation:
```python
>>> np.full_like(np.arange(10), fill_value=1 * ureg.m)
  File "<__array_function__ internals>", line 6, in full_like
  File ".../lib/python3.7/site-packages/numpy/core/numeric.py", line 401, in full_like
    multiarray.copyto(res, fill_value, casting='unsafe')
  File "<__array_function__ internals>", line 6, in copyto
TypeError: no implementation found for 'numpy.copyto' on types that implement __array_function__: [<class 'numpy.ndarray'>, <class 'pint.quantity.build_quantity_class.<locals>.QuantityScalar'>]

**Comment by jthielen** [comment]
@keewis Thank you for pointing that out, I forgot to mention that right now mixed types are not handled by https://github.com/hgrecco/pint/pull/764 (see https://github.com/hgrecco/pint/pull/764#issuecomment-523272038).

**Comment by keewis** [comment]
me hitting that issue is probably a bug in how `xr.Variable` handles `np.full_like` (the unit gets stripped), so no worries.

**Comment by keewis** [comment]
I'm not sure where to put this, so I'll post this here.

While working on the numpy methods part and using any of the pint versions implement `__array_function__` (hgrecco/pint#764 and andrewgsavage/pint#6) , I encountered the following:
```python
>>> array = np.array([1.3, 3.1, 7.5])
>>> array.searchsorted(2.4)
1
>>> (array * ureg.m).searchsorted(2.4 * ureg.m)
  File ".../pint/quantity.py", line 1548, in searchsorted
    raise DimensionalityError('dimensionless', self._units)
pint.errors.DimensionalityError: Cannot convert from 'dimensionless' to 'meter'
>>> (array * ureg.m).searchsorted(2.4)
<Traceback>
pint.errors.DimensionalityError: Cannot convert from 'dimensionless' to 'meter'
>>> (array * ureg.m).searchsorted(2.4 * ureg.s)
<Traceback>
pint.errors.DimensionalityError: Cannot convert from 'dimensionless' to 'meter'
```
I believe that is because in `BaseQuantity.searchsorted`, `self.__class__` is used to check the type, which may be either `QuantitySequence` or `QuantityScalar`. If `self` is a sequence and a scalar is passed to the method, this fails. To fix this, replacing `self.__class__` with `BaseQuantity` seems to work.
```python
>>> np.searchsorted(array * ureg.m, 2.4 * ureg.m)
<Quantity(1, 'meter')>
>>> np.searchsorted(array * ureg.m, 2.4)
<Quantity(1, 'meter')>
>>> np.searchsorted(array * ureg.m, 2.4 * ureg.s)
<Traceback>
pint.errors.DimensionalityError: Cannot convert from 'second' ([time]) to 'meter' ([length])
```
I don't know what to do here (using the code from andrewgsavage/pint#6 I tried moving `searchsorted` into the same category as `isclose` -- this seems to work partially, but passing non-quantities is still allowed), but the result should be indices and incompatible units (or no unit at all) should raise an error.

Both the method and the function work fine with current master.

jthielen:
@keewis Thank you for catching this! I've pushed an update to https://github.com/andrewgsavage/pint/pull/6 with your suggested changes. Raising an error on incompatible/missing units is definitely something that will be need to be added, but it may take some re-thinking the current implementations and how `convert_to_consistent_units` is used. My instinct is to auto-wrap non-Quantities (of known types on the type casting hierarchy) as dimensionless (as in https://github.com/hgrecco/pint/pull/764#issuecomment-523272038), and let pint's usual unit comparison checks handle the rest. But, knowing exactly what to do here and when may be something that would have to wait until some other discussions with pint are cleared up a little more (like https://github.com/hgrecco/pint/pull/764 vs. https://github.com/hgrecco/pint/pull/875 and https://github.com/hgrecco/pint/issues/845 / https://github.com/hgrecco/pint/issues/878)?

I'm thinking that discussion regarding bugs with pint's upcoming `__array_function__` may fit best on https://github.com/andrewgsavage/pint/pull/6 right now, since https://github.com/hgrecco/pint/pull/764 seems to have halted in favor of https://github.com/hgrecco/pint/pull/875? (That is, until a new PR is made that follows up on https://github.com/hgrecco/pint/pull/875 to replicate what was done in https://github.com/hgrecco/pint/pull/764 and https://github.com/andrewgsavage/pint/pull/6.)

keewis:
I'm currently working on the tests for the comparison methods (`DataArray.equals`, `DataArray.identical` and `DataArray.broadcast_equals`) and I'm trying to figure out what `equals` and `identical` return. `equals` is easy: just make sure the wrapped quantities are equal. `identical` is a stricter version of `equals`, so I was wondering whether it should also return False if the values are equal, but in compatible units (like `10 cm` and `0.1 m`)? Any thoughts on this?

max-sixty:
In general we've tried to make `identical` as exactly identical, including types etc, so I think different units would return `False` there

jhamman:
@keewis - I'm wondering if it would be worth merging this in its current state and leaving the rest of the checklist for a follow up PR? No harm in working incrementally on this. We just need to add a note in `whats-new.rst` and resolve a few merge conflicts. Thoughts?

keewis:
that's fine with me. I resolved the merge conflicts (along with other stuff), so only `whats-new.rst` and sorting out what is to be done in new PRs are left to do.

xarray/tests/test_units.py
```diff
@@ -8,6 +8,9 @@
 from xarray.core.npcompat import IS_NEP18_ACTIVE
 
 pint = pytest.importorskip("pint")
+DimensionalityError = pint.errors.DimensionalityError
```
max-sixty: This will raise if pint isn't installed, can we put it [here](https://github.com/pydata/xarray/pull/3238/files/bc85c3b64ace6d5ba8d8ff8dce2f5406953f3e47..4b4605d05bffeb2d4d449e2be4c87c50786e769a#diff-45289dbef3724b51d0b667552f57592cR33) ?

crusaderky: Won't raise because of the line just before it
``python
```

**Comment by crusaderky** [comment]
Fixed CI, What's New, and Installing doc

File: xarray/tests/test_units.py (hunk 1 above)
Comment (max-sixty) [reply xarray/tests/test_units.py]: Mea culpa. I had thought that sent `pint` to `None`, but it skips the whole module.

File: xarray/tests/test_units.py [hunk 2]
```diff
@@ -0,0 +1,1515 @@
...
+pytestmark = [
+    pytest.mark.skipif(
+        not IS_NEP18_ACTIVE, reason="NUMPY_EXPERIMENTAL_ARRAY_FUNCTION is not enabled"
+    ),
+    # pytest.mark.filterwarnings("ignore:::pint[.*]"),
+]
+
+
+def use_pint_version_or_xfail(*, version, reason):
+    return pytest.mark.xfail(LooseVersion(pint.__version__) < version, reason=reason)
```
crusaderky: Also unused, and in #3358 we went with the philosophy of not polluting the codebase with minimum version tests

xarray/tests/test_units.py
```diff
@@ -0,0 +1,1515 @@
...
+pint = pytest.importorskip("pint")
+DimensionalityError = pint.errors.DimensionalityError
+
+
+pytestmark = [
+    pytest.mark.skipif(
+        not IS_NEP18_ACTIVE, reason="NUMPY_EXPERIMENTAL_ARRAY_FUNCTION is not enabled"
+    ),
+    # pytest.mark.filterwarnings("ignore:::pint[.*]"),
+]
```
crusaderky: This is unused at the moment. If NEP18 is off, you should skip the whole test suite with a line right after import_or_skip("pint").

xarray/tests/test_units.py
crusaderky: Also unused, and in #3358 we went with the philosophy of not polluting the codebase with minimum version tests

xarray/tests/test_units.py
```diff
@@ -0,0 +1,1515 @@
...
+
+def use_pint_version_or_xfail(*, version, reason):
+    return pytest.mark.xfail(LooseVersion(pint.__version__) < version, reason=reason)
+
+
+unit_registry = pint.UnitRegistry()
+require_pint_array_function = pytest.mark.skipif(
+    not hasattr(unit_registry.Quantity, "__array_function__"),
+    reason="pint does not implement __array_function__ yet",
+)
```
crusaderky: Same as above - the minimum version of pint supported by our policy - which means, the latest available - does support 0.9. So this whole check should go.

xarray/tests/test_units.py
```diff
@@ -0,0 +1,1515 @@
...
+
+try:
+    from pint.quantity import BaseQuantity
+except ImportError:
+    BaseQuantity = unit_registry.Quantity
```
crusaderky: This is not merged into master, and I'm arguing on the pint board that it is likely a bad design.
https://github.com/hgrecco/pint/pull/880 is an alternative, cleaner approach.
For compatibility and until either PR is merged and a release is published off it, I think this whole module should just use
```python
Quantity = pint.UnitRegistry().Quantity
```

xarray/tests/test_units.py
```diff
@@ -0,0 +1,1515 @@
...
+        new_obj = xr.DataArray(
+            name=obj.name, data=data, coords=coords, attrs=attrs, dims=dims
+        )
+
+    return new_obj
+
+
+def assert_equal_with_units(a, b):
+    # works like xr.testing.assert_equal, but also explicitly checks units
+    # so, it is more like assert_identical
```
crusaderky: Wouldn't it be better to add pint awareness to xarray.testing.assert_identical?

xarray/tests/test_units.py
keewis: that is true, I have been using this for `require_pint_array_function`, but forgot to remove it

xarray/tests/test_units.py
keewis: the problem here is that the current pint master does not have `__array_function__` yet, so I think this is necessary to avoid failing too much.

xarray/tests/test_units.py
keewis: of course, that would be ideal (and I have been trying to replace it with that). But at the moment, `DataArray.identical` passes if a quantity is equal but in different units (like `10 cm` and `0.1 m`). So this is a workaround until `identical` works as expected. Also, I recently ran into an issue with converting: it returns a float, which seems to throw at least `identical` off (so maybe a combination of `allclose` and `identical`?)

xarray/tests/test_units.py
keewis: I have been developing all this against hgrecco/pint#764, so there may be more references to it. That said, replacing with
```python
```
seems to work

xarray/tests/test_units.py
keewis: so the question is whether the whole module should be skipped if pint does not implement `__array_function__`, or if each individual test should be skipped (I think this makes a difference?)

crusaderky: @keewis I think it's inconsequential from the moment we decided that we don't care about versions of pint that are older than the very latest available... if somebody runs the tests with an obsolete, unsupported version he'll get test failures.

keewis: yeah, but the most recent version produces these failures right now (even with pint master). So until pint has a released version with `__array_function__`, I'd like to keep this

xarray/tests/test_units.py
crusaderky: Ah sorry, my bad. I didn't know that ``pytestmark`` is a special pytest keyword.

keewis:
I updated the PR so that all the DataArray methods I found in the docs have a test, the only exception is `to_unstacked_dataset` where I don't know what the result should look like: Consider
```python
>>> data = [[0, 10, 20], [30, 40, 50]]
>>> x = [0, 1]
>>> y = [0, 1, 2]
>>> arr = xr.DataArray(data=data, coords={"x": x, "y": y}, dims=("x", "y"))
>>> arr.stack(z=("x", "y")).to_unstacked_dataset(dim="z")
<xarray.Dataset>
Dimensions:  (y: 3)
Coordinates:
  * y        (y) int64 0 1 2
Data variables:
    0        (y) int64 0 10 20
    1        (y) int64 30 40 50
```
and supposing that indexes keep the units, what should the result of the above, but with units on `x`, look like?

xarray/tests/test_units.py
keewis: @crusaderky I changed this to skipping the whole module, which should be removed once pint has at least one released version. Does this address your concern?

max-sixty:
> I assumed this to be the same as datetime, which means object labels instead of strings. So now only the top-level functions and dataset are missing.

👍

xarray/tests/test_units.py
max-sixty: I think we should try and merge this soon so let's not get hung up. But why not add this logic into `.identical`, or at least `assert_identical`? 
i.e. `.identical` returns `False` if units are different?

keewis: in the end having `identical` return the correct result is the goal (which we need to do anyway to have the corresponding test pass), but I would try to keep the code as general and independent from pint as possible -- though I don't know how that would look like.

To get `assert_identical` to work with quantities would require modifying `identical` and `formatting.diff_*_repr`. However, I agree that this is not urgent, so let's do that in a new PR

max-sixty:
OK great! @crusaderky any final thoughts? Anyone else?

xarray/tests/test_units.py
crusaderky: fine for me

crusaderky:
@keewis could you tweak whatsnew.rst and installing.rst to set the expectations (very low) for people using the current latest available release of pint?

keewis:
what do you think of this for `whats-new.rst`?
> Added integration tests against `pint <https://pint.readthedocs.io/>`_.  Note that these tests currently require a special version of pint to run. Also note that at the moment of writing, there are issues when pint interacts with non-numpy array libraries, e.g. dask or sparse.

crusaderky:
@keewis maybe you could clarify where to find this special version and how to get it (``pip install https+git://...``)? Also I think readers will be much more interested in what they can actually do with pint and what they can't, as opposed to what CI tests we're running.

max-sixty:
> Added integration tests against `pint <https://pint.readthedocs.io/>`_.  Note that these tests currently require a special version of pint to run *and are otherwise skipped*. Also note that at the moment of writing, there are issues when pint interacts with non-numpy array libraries, e.g. dask or sparse.

(remove emphasis in actual version)

> readers will be much more interested in what they can actually do with pint

👍 , let's add a sentence

keewis:
I have been using a PR to hgrecco/pint#764 because that has been the version that had the most tests pass. Is that okay, or should I use the original PR instead?

So something like this?
> Added integration tests against `pint <https://pint.readthedocs.io/>`_. Note that these tests currently require a `special version of pint <https://github.com/andrewgsavage/pint/pull/6>` (install with `pip install https+git://github.com/andrewgsavage/pint/pull/6`) to run and are otherwise skipped. Also note that at the moment of writing, there are issues when pint interacts with non-numpy array libraries, e.g. dask or sparse.

I don't know whether the install actually works, I'll experiment on this once my environment is done setting up.

crusaderky:
@keewis looks good. Also something equivalent in installing.rst on the line that mentions pint, please

keewis:
I don't really know what to write about
> readers will be much more interested in what they can actually do with pint

because a feature list seems to be a bit too long to me.

re `installing.rst`: is adding `(experimental)`  enough, or should the line with pip also appear there?

crusaderky:
whatsnew.rst:
```
- Added integration tests against pint <https://pint.readthedocs.io/>_.

  ..note::

    At the moment of writing, these tests *as well as the ability to use pint in general*
    require `a highly experimental version of pint
    <https://github.com/andrewgsavage/pint/pull/6>`_ (install with 
    ``pip install git+https://github.com/andrewgsavage/pint.git@refs/pull/6/head)``.
    Even with it, interaction with non-numpy array libraries, e.g. dask or sparse, is broken.
```
installing.rst
```
- `pint <https://pint.readthedocs.io/>`_: for units of measure.

  ..note::

    At the moment of writing, xarray requires a `highly experimental version of pint 
    <https://github.com/andrewgsavage/pint/pull/6>`_ (install with
    ``pip install git+https://github.com/andrewgsavage/pint.git@refs/pull/6/head)``.
    Even with it, interaction with non-numpy array libraries, e.g. dask or sparse, is broken.
```

crusaderky:
Also please move the what's new blob from 0.14.0 to 0.14.1

keewis:
done

max-sixty:
I think these tests are unrelated and we can merge?

```python


_________________________ test_datetime_reduce[False] __________________________

dask = False

    @arm_xfail
    def test_datetime_reduce(dask):
        time = np.array(pd.date_range("15/12/1999", periods=11))
        time[8:11] = np.nan
        da = DataArray(np.linspace(0, 365, num=11), dims="time", coords={"time": time})
    
        if dask and has_dask:
            chunks = {"time": 5}
            da = da.chunk(chunks)
    
        actual = da["time"].mean()
>       assert not pd.isnull(actual)
E       AssertionError: assert not True
E        +  where True = <function isna at 0x7fcbdefdbae8>(<xarray.DataArray 'time' ()>\narray('NaT', dtype='datetime64[ns]'))
E        +    where <function isna at 0x7fcbdefdbae8> = pd.isnull

xarray/tests/test_duck_array_ops.py:288: AssertionError
__________________________ test_datetime_reduce[True] __________________________

dask = True

    @arm_xfail
    def test_datetime_reduce(dask):
        time = np.array(pd.date_range("15/12/1999", periods=11))
        time[8:11] = np.nan
        da = DataArray(np.linspace(0, 365, num=11), dims="time", coords={"time": time})
    
        if dask and has_dask:
            chunks = {"time": 5}
            da = da.chunk(chunks)
    
        actual = da["time"].mean()
>       assert not pd.isnull(actual)
E       AssertionError: assert not True
E        +  where True = <function isna at 0x7fcbdefdbae8>(<xarray.DataArray 'time' ()>\narray('NaT', dtype='datetime64[ns]'))
E        +    where <function isna at 0x7fcbdefdbae8> = pd.isnull
```

keewis:
I would think so, too. Go ahead.

max-sixty:
Great! Quite an effort @keewis —thank you!

dcherian:
Phew thanks a lot @keewis!

shoyer:
awesome work @keewis !

Issue #3231, opened 2019-08-19
Dict-like argument support for assign_coords()


gwgundersen:
Often the xarray API supports [either a `dict` or `kwargs`](https://github.com/pydata/xarray/blob/master/xarray/core/utils.py#L257), and it seems like this should be consistent as often as possible. With `assign_coords`, this works as expected:

```python
>>> dims = ["x", "y"]
>>> coords = {"fruit": ("x", ["apple", "banana"])}
>>> arr = xr.DataArray([[1,2,3],[4,5,6]], dims=dims, coords=coords)
>>> arr.assign_coords(color=("x", ["red", "yellow"]))
<xarray.DataArray (x: 2, y: 3)>
array([[1, 2, 3],
       [4, 5, 6]])
Coordinates:
    fruit    (x) <U6 'apple' 'banana'
    color    (x) <U6 'red' 'yellow'
Dimensions without coordinates: x, y
```

Since `assign_coords` works with keyword arguments, I would expect it to also work with a `dict`, similar to `sel`, `drop`, etc. But it does not:

```python
>>> arr.assign_coords({"color": ("x", ["red", "yellow"])})
*** TypeError: assign_coords() takes 1 positional argument but 2 were given
```

max-sixty:
Definitely we could add this. There's an issue somewhere for rolling this functionality out - I think we've done 90% of the methods.

shoyer:
Yes, we should definitely do this.

gwgundersen:
Okay, I'll finish https://github.com/pydata/xarray/pull/3128 first and then tackle this.

PR #3243, opened 2019-08-22
Support for dict-like arguments for assign_coords

gwgundersen:
- [x] Closes https://github.com/pydata/xarray/issues/3231.
 - [x] Added tests for `Dataset` and `groupby`. Do we also need tests for `DataArray`? I assumed not since the actual code is in `common.py`, but I can add more.
 - [x] Passes `black . && mypy . && flake8`.
 - [x] Fully documented, including `whats-new.rst` and new example in the docstring.

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -341,7 +341,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
                 results[k] = v
         return results
 
-    def assign_coords(self, **kwargs):
+    def assign_coords(self, coords=None, **coords_kwargs):
         """Assign new coordinates to this object.
 
         Returns a new object with all the original data in addition to the new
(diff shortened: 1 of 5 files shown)


xarray/core/common.py
```diff
@@ -341,15 +341,21 @@ def _calc_assign_results(
...
+    def assign_coords(self, coords=None, **coords_kwargs):
         """Assign new coordinates to this object.
 
         Returns a new object with all the original data in addition to the new
         coordinates.
 
         Parameters
         ----------
-        kwargs : keyword, value pairs
+        coords : dict, optional
```
max-sixty: Great, I don't think you need to repeat the description - you could use the same format as https://github.com/pydata/xarray/blob/master/xarray/core/dataset.py#L1864

gwgundersen: Thanks for the feedback. Done.

max-sixty:
Great, thanks @gwgundersen !

PR #3247, opened 2019-08-22
Update filter_by_attrs to use 'variables' instead of 'data_vars'

cspencerjones:
This will allow `filter_by_attrs` to filter coordinates as well as variables. @ocefpaf

### DIFF
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -4938,7 +4938,7 @@ class TestDataset:
         new_ds = ds.filter_by_attrs(long_name="time_in_seconds")
         assert new_ds["time"].long_name == "time_in_seconds"
         assert not bool(new_ds.data_vars)
- 
+
         # Test return more than one DataArray.
         new_ds = ds.filter_by_attrs(standard_name="air_potential_temperature")
         assert len(new_ds.data_vars) == 2
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -4920,7 +4920,7 @@ class TestDataset:
                 "temperature_10": (["t"], [0], temp10),
                 "precipitation": (["t"], [0], precip),
             },
-            coords={"time": (["t"], [0], dict(axis="T"))},
+            coords={"time": (["t"], [0], dict(axis="T",standard_name="time_in_seconds"))},
         )
 
         # Test return empty Dataset.
@@ -4934,6 +4934,11 @@ class TestDataset:
 
         assert_equal(new_ds["precipitation"], ds["precipitation"])
 
+        #Test filter coordinates
+        new_ds = ds.filter_by_attrs(standard_name="time_in_seconds")
+        assert new_ds["time"].standard_name == "time_in_seconds"
+        assert not bool(new_ds.data_vars)
+        
         # Test return more than one DataArray.
         new_ds = ds.filter_by_attrs(standard_name="air_potential_temperature")
         assert len(new_ds.data_vars) == 2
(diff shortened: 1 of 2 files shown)


dcherian:
Thanks @cspencerjones . Can you add a test that tests this?

ocefpaf:
> Thanks @cspencerjones . Can you add a test that tests this?

I guess we can just extend the existing test and check if we can find a coordinate by its attributes. For example, add a standard name [for time here](https://github.com/pydata/xarray/blob/52a16a6517dfc56094a20c3f7484f62afca2bda7/xarray/tests/test_dataset.py#L4923) and see if we can find `time` using it.

dcherian:
Exactly.

dopplershift:
cc: @jthielen Might make some things in MetPy easier...

xarray/tests/test_dataset.py
```diff
@@ -4920,7 +4920,7 @@ def test_filter_by_attrs(self):
                 "temperature_10": (["t"], [0], temp10),
                 "precipitation": (["t"], [0], precip),
             },
-            coords={"time": (["t"], [0], dict(axis="T"))},
+            coords={"time": (["t"], [0], dict(axis="T", standard_name="time_in_seconds"))},
```
jthielen:
```suggestion
            coords={"time": (["t"], [0], dict(axis="T", standard_name="time", units="s"))},
```

If using standard_name, I think it would be best to stay CF-compliant? Based on [the guidelines](http://cfconventions.org/Data/cf-standard-names/docs/guidelines.html) I don't think units should be given in the standard_name, but rather as its own attribute.

xarray/tests/test_dataset.py
```diff
@@ -4934,6 +4934,11 @@ def test_filter_by_attrs(self):
 
         assert_equal(new_ds["precipitation"], ds["precipitation"])
 
+        # Test filter coordinates
+        new_ds = ds.filter_by_attrs(standard_name="time_in_seconds")
```
jthielen:
```suggestion
        new_ds = ds.filter_by_attrs(standard_name="time")
```

xarray/tests/test_dataset.py
```diff
@@ -4934,6 +4934,11 @@ def test_filter_by_attrs(self):
...
+        assert new_ds["time"].standard_name == "time_in_seconds"
```
jthielen:
```suggestion
        assert new_ds["time"].standard_name == "time"
```

xarray/tests/test_dataset.py
cspencerjones: I think this is a good idea, but I don't want to set the standard name to be the same as the name of the coordinate since it will make the code confusing to read. Perhaps i can change it to "time_after_1990" or something similar? Do you have a suggestion?

jthielen: I think if you want to use `standard_name`, it has to be "time" to be CF compliant. However, if you want "time_in_seconds" or "time_after_1990", I believe you can use the much-more lenient `long_name` attribute (see http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#long-name).

cspencerjones: Ahh, ok - I will switch to long_name then. Thanks for your help!

max-sixty:
Thanks @cspencerjones ! Great to have you as a contributor

dcherian:
Thanks @cspencerjones

dcherian:
Oh no! I forgot to ask you to add a note in `whats-new.rst`. Can you send in a new PR with a small note in that file please?

PR #3249, opened 2019-08-23
added support for rasterio geotiff tags

normanb:
- [x] Tests added

GeoTIFF and TIFF include tags that are useful in geo-processing. In particular the `AREA_OR_POINT` tag is commonly used. Although GDAL corrects the geo-referencing it is important to know which scheme is being used. An example of when this is useful is when aligning digital elevation models (which can be geotiff) with a raster data source.

### DIFF
--- a/xarray/backends/rasterio_.py
+++ b/xarray/backends/rasterio_.py
@@ -322,11 +322,14 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None, loc
         attrs["units"] = riods.units
 
     # Parse extra metadata from tags, if supported
-    parsers = {"ENVI": _parse_envi}
+    parsers = {"ENVI": _parse_envi, "GTiff": lambda m: m}
 
     driver = riods.driver
     if driver in parsers:
-        meta = parsers[driver](riods.tags(ns=driver))
+        if driver == "GTiff":
+            meta = parsers[driver](riods.tags())
+        else:
+            meta = parsers[driver](riods.tags(ns=driver))
 
         for k, v in meta.items():
             # Add values as coordinates if they match the band count,
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -3923,6 +3923,12 @@ class TestRasterio:
                 assert isinstance(rioda.attrs["map_info"], str)
                 assert isinstance(rioda.attrs["samples"], str)
 
+    def test_geotiff_tags(self):
+        # Create a geotiff file with some tags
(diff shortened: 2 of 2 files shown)


xarray/backends/rasterio_.py
```diff
@@ -322,11 +322,14 @@ def open_rasterio(filename, parse_coordinates=None, chunks=None, cache=None, loc
         attrs["units"] = riods.units
 
     # Parse extra metadata from tags, if supported
-    parsers = {"ENVI": _parse_envi}
+    parsers = {"ENVI": _parse_envi, "GTiff": lambda m: m}
 
     driver = riods.driver
     if driver in parsers:
-        meta = parsers[driver](riods.tags(ns=driver))
+        if driver == "GTiff":
```
dcherian: So the metadata that is read back depends on the driver used?

normanb: yes, metadata can be driver dependent. there was already support for ENVI drivers and I followed this approach for geotiff. As it happens geotiff is a special case within GDAL and doesn't require specifying a namespace.

xarray/tests/test_backends.py
```diff
@@ -3923,6 +3923,12 @@ def test_ENVI_tags(self):
                 assert isinstance(rioda.attrs["map_info"], str)
                 assert isinstance(rioda.attrs["samples"], str)
 
+    def test_geotiff_tags(self):
+        # Create a geotiff file with some tags
+        with create_tmp_geotiff() as (tmp_file, _):
+            with xr.open_rasterio(tmp_file) as rioda:
+                assert isinstance(rioda.attrs["AREA_OR_POINT"], str)
```
dcherian: Is `"AREA_OR_POINT"` set by default? I don't see it in `create_tmp_geotiff()`

dcherian: Should we check for a specific value instead of just the type?

normanb: Yes, `AREA_OR_POINT` is set by default in a geotiff http://web.archive.org/web/20160326194152/http://remotesensing.org/geotiff/spec/geotiff2.5.html#2.5.2

normanb: > Should we check for a specific value instead of just the type?

I don't think that is needed, whatever is there needs to be reported for interpretation. XArray doesn't act on that value.

dcherian:
OK thanks @normanb . Welcome to xarray! I'll merge in a couple of days if no one else has comments (e.g. @fmaussion).

dcherian:
Going in. Thanks @normanb

PR #3253, opened 2019-08-23
Updater to testing environment name

cspencerjones:
The testing environment name has been updated to `xarray-tests` in the package and we should do this in the documentation as well.


dcherian:
Sweet! Thanks.

Issue #3245, opened 2019-08-22
sparse and other duck array issues


dcherian:
Issues I've run into working with @friedrichknuth at the Pangeo meeting trying to use sparse

1. `.plot()` doesn't work. We need to convert to dense before plotting
2. `.values` doesn't work. Raises an error about explicitly calling `todense()`
3. `align` doesn't seem to work.

shoyer:
I think it is intentional for automatic coercion to NumPy arrays to fail. Making `plot()` and `.values` work should probably require explicitly writing `datarray.to_dense()` first.

dcherian:
The better way to phrase this is:

1. Should we formalize the convention that `.values` always return a numpy array i.e. it will call `todense()` implicitly and that `.data` will always return the underlying container: sparse/dask etc.?
2. Should we call `todense()` automatically in `plot()` the same way we do `compute()`?

shoyer:
> * Should we formalize the convention that `.values` always return a numpy array i.e. it will call `todense()` implicitly and that `.data` will always return the underlying container: sparse/dask etc.?

I think `.values` should either return a NumPy array or raise an exception, based upon whether or not the underlying duck array supports coercion with `np.array()`.

We should have a separate API (maybe `.to_dense()` or `.to_numpy_data()`?) for explicitly converting into NumPy arrays. This should not be called automatically inside xarray.

Basically, we should leave the  decision about whether automatic coercion is safe up to the authors of duck array libraries.

dcherian:
:+1: I think `to_numpy()` would be good.

shoyer:
The main downside of  `to_numpy()` is that it isn't obvious whether the result should be a DataArray containing a NumPy array or a raw NumPy array. In particular, pandas uses `to_numpy()` for converting into a raw NumPy, which is the opposite of our intent here.

max-sixty:
I do think the pandas approach is pretty good here - don't break anything relying on `.values` - instead have `.array` to get the backing array (in whatever form it might be), and `to_numpy()` coercing to a raw numpy array

ref https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.24.0.html#accessing-the-values-in-a-series-or-index

shoyer:
Xarray's .data is basically the equivalent of pandas's .array.

khaeru:
As far as I can tell, the proposal here will require either
```python
s = pd.Series(...)
xr.DataArray.from_series(s).to_series()
```
or:
```python
[...]

PR #3255, opened 2019-08-23
Sparse fixes for reindex

dcherian:
- [x] Tests added

### DIFF
--- a/xarray/core/indexing.py
+++ b/xarray/core/indexing.py
@@ -10,7 +10,7 @@ import pandas as pd
 
 from . import duck_array_ops, nputils, utils
 from .npcompat import DTypeLike
-from .pycompat import dask_array_type, integer_types
+from .pycompat import dask_array_type, integer_types, sparse_array_type
 from .utils import is_dict_like, maybe_cast_to_coords_dtype
 
 
@@ -1076,7 +1076,13 @@ def _logical_any(args):
     return functools.reduce(operator.or_, args)
 
 
-def _masked_result_drop_slice(key, chunks_hint=None):
+def _masked_result_drop_slice(key, data=None):
+
+    if data is not None:
+        chunks_hint = getattr(data, "chunks", None)
+    else:
+        chunks_hint = None
+
     key = (k for k in key if not isinstance(k, slice))
     if chunks_hint is not None:
         key = [
@@ -1085,10 +1091,17 @@ def _masked_result_drop_slice(key, chunks_hint=None):
             else k
             for k in key
         ]
-    return _logical_any(k == -1 for k in key)
+
+    mask = _logical_any(k == -1 for k in key)
+    if isinstance(data, sparse_array_type):
+        import sparse
+
+        mask = sparse.COO.from_numpy(mask)
+
+    return mask
 
 
-def create_mask(indexer, shape, chunks_hint=None):
+def create_mask(indexer, shape, data=None):
     """Create a mask for indexing with a fill-value.
 
     Parameters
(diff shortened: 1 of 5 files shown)


xarray/core/variable.py
```diff
@@ -712,6 +720,8 @@ def _getitem_with_mask(self, key, fill_value=dtypes.NA):
             data = as_indexable(self._data)[actual_indexer]
             chunks_hint = getattr(data, "chunks", None)
             mask = indexing.create_mask(indexer, self.shape, chunks_hint)
+            if isinstance(self.data, sparse.COO):
+                mask = sparse.COO.from_numpy(mask)
             data = duck_array_ops.where(mask, fill_value, data)
```
dcherian: The underlying issue is that `sparse.where` expects mask to be `SparseArray` so `create_mask` needs to know to return that. @shoyer What would be the best way to do this properly?

dcherian: Another fix is that `fill_value` should be using `self.data.fill_value` by default for a sparse array.

xarray/core/variable.py
```diff
@@ -235,7 +240,10 @@ def _as_array_or_item(data):
 
     TODO: remove this (replace with np.asarray) once these issues are fixed
     """
+    if isinstance(data, sparse.COO):
+        data = data.todense()
```
shoyer: If I  understand correctly, this will automatically coerce sparse -> dense arrays with `DataArray.values`. I don't think this is a good idea -- see https://github.com/pydata/xarray/issues/3245

xarray/core/variable.py
shoyer: It seems like we should probably pipe some information about the indexed data into `indexing.create_mask`. Right  now we do that with `chunks_hint` specifically for dask array, but that could be made more generic by passing in  `data` directly. In particular we should coerce ndarrays in the indexing key to sparse arrays inside `indexing._masked_result_drop_slice`.

xarray/core/variable.py
dcherian: Agreed. it was a temp thing to make development easier.

xarray/core/variable.py
```diff
@@ -33,6 +33,11 @@
 except ImportError:
     pass
 
+try:
+    import sparse
```
shoyer: use `pycompat.sparse_array_type` here

xarray/core/variable.py
dcherian: OK done. Let me know if you want more changes.

xarray/core/indexing.py
```diff
@@ -1085,10 +1091,17 @@ def _masked_result_drop_slice(key, chunks_hint=None):
             else k
             for k in key
         ]
-    return _logical_any(k == -1 for k in key)
+
+    mask = _logical_any(k == -1 for k in key)
+    if isinstance(data, sparse_array_type):
+        import sparse
+
+        mask = sparse.COO.from_numpy(mask)
```
shoyer: I think it's probably better to convert the indexer before logical_any above. If I understand correctly, the indexer keys will generally be 1d but the mask may be higher dimensional, like the data itself.

xarray/core/indexing.py
```diff
@@ -1076,7 +1076,13 @@ def _logical_any(args):
     return functools.reduce(operator.or_, args)
 
 
-def _masked_result_drop_slice(key, chunks_hint=None):
+def _masked_result_drop_slice(key, data=None):
+
+    if data is not None:
+        chunks_hint = getattr(data, "chunks", None)
+    else:
+        chunks_hint = None
```
shoyer: The getattr line above already does this. I don't think you need this if statement at all.

xarray/core/indexing.py
dcherian: OK the latest commit does that.

dcherian:
Test fails because we need the git master version of sparse. I opened an issue asking about a release there (https://github.com/pydata/sparse/issues/279)

Issue #3147, opened 2019-07-18
Implementing map_blocks and map_overlap


jakirkham:
Just as there are [`map_blocks`]( https://docs.dask.org/en/latest/array-api.html#dask.array.Array.map_blocks ) and [`map_overlap`]( https://docs.dask.org/en/latest/array-api.html#dask.array.Array.map_overlap ) methods for Dask Array, it would be useful to have equivalent methods for Xarray objects. This would make it easier to leverage duck typing to work with both Dask Arrays and Xarray objects.

dcherian:
+1. The `split_by_chunks` method in this comment (https://github.com/pydata/xarray/issues/1093#issuecomment-259213382) would also be useful for more general per-chunk manipulation.

jakirkham:
That sounds somewhat similar to `.blocks` accessor in Dask Array. ( https://github.com/dask/dask/pull/3689 ) Maybe we should align on that as well?

jakirkham:
Another approach for the `split_by_chunks` implementation would be...

```python
def split_by_chunks(a):
    for sl in da.core.slices_from_chunks(a.chunks): 
        yield (sl, a[sl])
```

While a little bit more cumbersome to write, this could be implemented with `.blocks` and may be a bit more performant.

```python
def split_by_chunks(a):
    for i, sl in zip(np.ndindex(a.numblocks), da.core.slices_from_chunks(a.chunks)):
        yield (sl, a.blocks[i])
```

If the slices are not strictly needed, this could be simplified a bit more.

```python
def split_by_chunks(a):
    for i in np.ndindex(a.numblocks):
        yield a.blocks[i]
```

Admittedly `slices_from_chunks` is an internal utility function. Though it is unlikely to change. We could consider exposing it as part of the API if that is useful.

We could consider other things like making `.blocks` iterable, which could make this more friendly as well.

jhamman:
`map_blocks` went in as of #3276. We'll leave this open for the future work implementing `map_overlap`.

TomAugspurger:
I'm thinking through a `map_overlap` API right now. In dask, map_overlap requires a few different argument

```
    depth: int, tuple, dict or list
        The number of elements that each block should share with its neighbors
        If a tuple or dict then this can be different per axis.
        If a list then each element of that list must be an int, tuple or dict
        defining depth for the corresponding array in `args`.
        Asymmetric depths may be specified using a dict value of (-/+) tuples.
        Note that asymmetric depths are currently only supported when
[...]

PR #3258, opened 2019-08-24
[WIP] Add map_blocks.

dcherian:
- Related #3147 
 - [ ] Tests added

ping @mrocklin @sofroniewn @shanaxel42

### DIFF
--- a/xarray/core/parallel.py
+++ b/xarray/core/parallel.py
@@ -1,0 +1,86 @@
+try:
+    import dask
+    import dask.array
+    from dask.highlevelgraph import HighLevelGraph
+
+except ImportError:
+    pass
+
+from .dataarray import DataArray
+
+
+def get_chunk_slices(dataset):
+    chunk_slices = {}
+    for dim, chunks in dataset.chunks.items():
+        slices = []
+        start = 0
+        for chunk in chunks:
+            stop = start + chunk
+            slices.append(slice(start, stop))
+            start = stop
+        chunk_slices[dim] = slices
+
+    return chunk_slices
+
+
+def map_blocks(func, darray):
+    """
+    A version of dask's map_blocks for DataArrays.
+
+    Parameters
+    ----------
+    func: callable
+        User-provided function that should accept DataArrays corresponding to one chunk.
+    darray: DataArray
+        Chunks of this array will be provided to 'func'. The function must not change
+        shape of the provided DataArray.
+
+    Returns
+    -------
+    DataArray
+
+    See Also
+    --------
+    dask.array.map_blocks
+    """
+
+    def _wrapper(darray):
+        result = func(darray)
+        if not isinstance(result, type(darray)):
+            raise ValueError("Result is not the same type as input.")
+        if result.shape != darray.shape:
+            raise ValueError("Result does not have the same shape as input.")
+        return result
+
+    meta_array = DataArray(darray.data._meta, dims=darray.dims)
+    result_meta = func(meta_array)
+
+    name = "%s-%s" % (darray.name or func.__name__, dask.base.tokenize(darray))
+
+    slicers = get_chunk_slices(darray._to_temp_dataset())
+    dask_keys = list(dask.core.flatten(darray.__dask_keys__()))
+
+    graph = {
+        (name,)
+        + (*key[1:],): (
+            _wrapper,
+            (
+                DataArray,
+                key,
+                {
+                    dim_name: darray[dim_name][slicers[dim_name][index]]
+                    for dim_name, index in zip(darray.dims, key[1:])
+                },
+                darray.dims,
+            ),
+        )
+        for key in dask_keys
+    }
+
+    graph = HighLevelGraph.from_collections(name, graph, dependencies=[darray])
+
+    return DataArray(
+        dask.array.Array(graph, name, chunks=darray.chunks, meta=result_meta),
+        dims=darray.dims,
+        coords=darray.coords,
+    )
--- a/xarray/__init__.py
+++ b/xarray/__init__.py
@@ -34,6 +34,8 @@ from .conventions import decode_cf, SerializationWarning
 from .coding.cftime_offsets import cftime_range
 from .coding.cftimeindex import CFTimeIndex
 
+from .core.parallel import map_blocks
+
 from .util.print_versions import show_versions
 
 from . import tutorial
--- a/xarray/tests/test_dask.py
+++ b/xarray/tests/test_dask.py
@@ -878,3 +878,25 @@ def test_dask_layers_and_dependencies():
     assert set(x.foo.__dask_graph__().dependencies).issuperset(
         ds.__dask_graph__().dependencies
     )
+
+
+def test_map_blocks():
+    darray = xr.DataArray(
+        dask.array.ones((10, 20), chunks=[4, 5]),
+        dims=["x", "y"],
+        coords={"x": np.arange(10), "y": np.arange(100, 120)},
+    )
+    darray.name = None
+
+    def good_func(darray):
+        return darray * darray.x + 5 * darray.y
+
+    def bad_func(darray):
+        return (darray * darray.x + 5 * darray.y)[:1, :1]
+
+    actual = xr.map_blocks(good_func, darray)
+    expected = good_func(darray)
+    xr.testing.assert_equal(expected, actual)
+
+    with raises_regex(ValueError, "not have the same shape"):
+        xr.map_blocks(bad_func, darray).compute()


xarray/core/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+            slices.append(slice(start, stop))
+            start = stop
+        chunk_slices[dim] = slices
+
+    return chunk_slices
+
+
+def map_blocks(func, darray):
+    """
+    A version of dask's map_blocks for DataArrays.
```
mrocklin: We should maybe not assume that the user is familiar with Dask?  I don't know how common this is among Xarray users.

```suggestion
    Apply a function across each chunk of the DataArray in parallel
    
    This is similar to Dask Array's map_blocks function, but your function 
    receives a DataArray rather than a Numpy Array.
```

xarray/core/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+        if not isinstance(result, type(darray)):
+            raise ValueError("Result is not the same type as input.")
+        if result.shape != darray.shape:
+            raise ValueError("Result does not have the same shape as input.")
+        return result
+
+    meta_array = DataArray(darray.data._meta, dims=darray.dims)
+    result_meta = func(meta_array)
+
+    name = "%s-%s" % (darray.name or func.__name__, dask.base.tokenize(darray))
```
mrocklin: You might want to use `dask.utils.funcname` in order to get the function name.  This handles some additional cases we've seen, like `functools.partial`.

mrocklin: What is `darray.name`?  It might be best to use the function name regardless.  This is the name that shows up in things like the progress bars in the dashboard.

xarray/core/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+        for chunk in chunks:
+            stop = start + chunk
+            slices.append(slice(start, stop))
+            start = stop
+        chunk_slices[dim] = slices
+
+    return chunk_slices
+
+
+def map_blocks(func, darray):
```
mrocklin: We might consider accepting `*args, **kwargs` here, at least in the simple case when they aren't other Dask/Xarray collections.  That would allow for simple things like ...

```python
darray.map_blocks(operator.add, 10)
```

So maybe something like ...

```python
for arg in args + tuple(kwargs.values()):
    if dask.base.is_dask_collection(arg):
        raise NotImplementedError(...)
```

And then use the task`(_wrapper, func, (DataArray, key, ...), args, kwargs)`

xarray/core/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+    def _wrapper(darray):
+        result = func(darray)
+        if not isinstance(result, type(darray)):
+            raise ValueError("Result is not the same type as input.")
+        if result.shape != darray.shape:
+            raise ValueError("Result does not have the same shape as input.")
+        return result
+
+    meta_array = DataArray(darray.data._meta, dims=darray.dims)
+    result_meta = func(meta_array)
```
mrocklin: We might want to have a try-except around this, and raise some informative error or warning if it doesn't work.

crusaderky:
Hi,

A few design opinions:

1. could we call it just "map"? It makes sense as this thing would be very useful for non-dask based arrays too. Working routinely with scipy (chiefly with scipy.stats transforms), I tire a lot of writing very verbose ``xarray.apply_ufunc`` calls.

2. could we have it as a method of DataArray and Dataset, to allow for method chaining?

e.g.
```python
myarray.map(func1).chunk().map(func2).sum().compute()
```

xarray/core/parallel.py
crusaderky: +1 for ``*args, **kwargs``

crusaderky: Can we expose dtype=None as in the underlying dask function? It would save some CPU when

xarray/core/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+                {
+                    dim_name: darray[dim_name][slicers[dim_name][index]]
+                    for dim_name, index in zip(darray.dims, key[1:])
+                },
+                darray.dims,
+            ),
+        )
+        for key in dask_keys
+    }
+
```
crusaderky: Instead of building a graph by hand, why aren't you simply calling ``dask.array.map_blocks``?

crusaderky:
@shoyer let me rephrase it - apply_ufunc is extremely powerful, and when you need to cope with all possible shape transformations, I suspect its verbosity is quite necessary.
It's just that, when all you need to do is apply an elementwise, embarassingly parallel function (80% of the times in my real life experience), apply_ufunc is overkill.

The thing I have against the name map_blocks is that backends other than dask have no notion of blocks...

dcherian:
I started prototyping a Dataset version. Here's what I have:

``` python
import dask
import xarray as xr

darray = xr.DataArray(np.ones((10, 20)), 
                  dims=['x', 'y'], 
                  coords={'x': np.arange(10), 'y': np.arange(100, 120)})
dset = darray.to_dataset(name='a')
dset['b'] = dset.a + 50
dset['c'] = (dset.x + 20)
dset = dset.chunk({'x': 4, 'y': 5})
```

The function I'm applying takes a dataset and returns a DataArray because that's easy to test without figuring out how to assemble everything back into a dataset.
``` python
import itertools

# function takes dataset and returns dataarray so that I can check that things work without reconstructing a dataset
def function(ds):
    return ds.a + 10

dataset_dims = list(dset.dims)

graph = {}
gname = 'dsnew'

# map dims to list of chunk indexes
# If different variables have different chunking along the same dim
# the call to .chunks will raise an error.
ichunk = {dim: range(len(dset.chunks[dim])) for dim in dataset_dims}

# iterate over all possible chunk combinations
for v in itertools.product(*ichunk.values()):
    chunk_index_dict = dict(zip(dataset_dims, v))
    data_vars = {}
    for name, variable in dset.data_vars.items():
        # why do does dask_keys have an extra level?
        # the [0] is not required for dataarrays
        var_dask_keys = variable.__dask_keys__()[0]
        
        # recursively index into dask_keys nested list
        chunk = var_dask_keys
        for dim in variable.dims:
            chunk = chunk[chunk_index_dict[dim]]
            
        # I have key corresponding to chunk
        # this tuple is in a dictionary passed to xr.Dataset()
        # dask doesn't seem to replace this with a numpy array at execution time.
        data_vars[name] = (variable.dims, chunk)
        
    graph[(gname, ) + v] = (function, (xr.Dataset, data_vars))

final_graph = dask.highlevelgraph.HighLevelGraph.from_collections(name, graph, dependencies=[dset])
```

Elements of the graph look like
```
('dsnew', 0, 0): (<function __main__.function(ds)>,
  (xarray.core.dataset.Dataset,
   {'a': (('x', 'y'), ('xarray-a-f178df193efafa67203f3862b3f9f0f4', 0, 0)),
    'b': (('x', 'y'), ('xarray-b-e2d8d06bb9e5c1f351671a94816bd331', 0, 0)),
    'c': (('x',), ('xarray-c-d90f8b2af715b53f4c170be391239655', 0))}))
```

This doesn't work because dask doesn't replace the keys by numpy arrays when the `xr.Dataset` call is executed.  The graph is "disconnected": 
image


I'm not sure what I'm doing wrong here. An equivalent version for DataArrays works perfectly.

mrocklin:
Dask doesn't traverse through tuples to find possible keys, so the keys here are hidden from view:

```python
   {'a': (('x', 'y'), ('xarray-a-f178df193efafa67203f3862b3f9f0f4', 0, 0)),
```

I recommend changing wrapping tuples with lists:

```diff
-   {'a': (('x', 'y'), ('xarray-a-f178df193efafa67203f3862b3f9f0f4', 0, 0)),
+   {'a': [('x', 'y'), ('xarray-a-f178df193efafa67203f3862b3f9f0f4', 0, 0)],
```

dcherian:
Thanks @mrocklin. Unfortunately that doesn't work with the Dataset constructor. With a list it treats it as array-like

```
    The following notations are accepted:

    - mapping {var name: DataArray}
    - mapping {var name: Variable}
    - mapping {var name: (dimension name, array-like)}
    - mapping {var name: (tuple of dimension names, array-like)}
    - mapping {dimension name: array-like}
      (it will be automatically moved to coords, see below)
```

Unless @shoyer has another idea, I guess I can insert creating a DataArray into the graph and then refer to those keys in the Dataset constructor.

mrocklin:
Then you can construct a tuple as a task `((1, 2, 3),)` -> `(tuple, [1, 2, 3])`

dcherian:
Thanks. That worked. I have a new version up in #3276 that works with both DataArrays and Datasets.

mrocklin:
I'm glad to see progress here.  FWIW, I think that many people would be quite happy with a version that just worked for DataArrays, in case that's faster to get in than the full solution with DataSets.

dcherian:
Closing in favour of #3276

PR #3271, opened 2019-08-28
Raise proper error for scalar array when coords is a dict

DangoMelon:
As explained here https://github.com/pydata/xarray/pull/3159#discussion_r316230281 , when a user uses a scalar array to build a `DataArray` with `coords` given as a dictionary the error is not self explanatory.
```python
>>> xr.DataArray(np.array(1), coords={'x': np.arange(4), 'y': 'a'}, dims=['x'])
...
KeyError: 'x'
```

This PR makes sure that when `data` is a scalar array and `dims` is not empty, it sets the shape to `(0,)` to make it fail with the proper raise message 

```python
>>> xr.DataArray(np.array(1), coords={'x': np.arange(4), 'y': 'a'}, dims=['x'])
...
ValueError: conflicting sizes for dimension 'x': length 0 on the data but length 4 on coordinate 'x'
```

 - [x] Test updated
 - [ ] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API (is this needed for a change like this?)

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -113,6 +113,11 @@ def _infer_coords_and_dims(
                 coord = as_variable(coord, name=dims[n]).to_index_variable()
                 dims[n] = coord.name
         dims = tuple(dims)
+    elif len(dims) != len(shape):
+        raise ValueError(
+            "different number of dimensions on data "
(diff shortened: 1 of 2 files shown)


shoyer:
Looks great, thank you!

(We don't really need release notes for every minor change, especially if nothing is changing from a user perspective.)

PR #3276, opened 2019-09-02
map_blocks

### LINKED ISSUE #3323: arithmetic resulting in inconsistent chunks
``` python

import xarray as xr
import numpy as np

def make_da():
    return xr.DataArray(
        np.ones((10, 20)),
        dims=["x", "y"],
        coords={"x": np.arange(10), "y": np.arange(100, 120),},
        name="a",
    ).chunk({"x": 4, "y": 5})

map_da = make_da()
map_ds = xr.Dataset()
map_ds["a"] = make_da()
map_ds["c"] = map_ds.x + 20
map_ds = map_ds.chunk({"x": 4, "y": 5})
```

```
<xarray.Dataset>
Dimensions:  (x: 10, y: 20)
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9
  * y        (y) int64 100 101 102 103 104 105 106 ... 114 115 116 117 118 119
Data variables:
    a        (x, y) float64 dask.array<chunksize=(4, 5), meta=np.ndarray>
    c        (x) int64 dask.array<chunksize=(4,), meta=np.ndarray>
```

`(map_ds + map_ds.y)` gives `y` chunksize of 20 for `c` and 5 for `a`

```
(map_ds + map_ds.y)
<xarray.Dataset>
Dimensions:  (x: 10, y: 20)
Coordinates:
  * x        (x) int64 0 1 2 3 4 5 6 7 8 9
  * y        (y) int64 100 101 102 103 104 105 106 ... 114 115 116 117 118 119
Data variables:
    a        (x, y) float64 dask.array<chunksize=(4, 5), meta=np.ndarray>
    c        (x, y) int64 dask.array<chunksize=(4, 20), meta=np.ndarray>
```

This seems reasonable except `(map_ds + map_ds.y).chunks)` raises an "Inconsistent chunks" error.

I ran into this writing tests for map_blocks.

dcherian:
- [x] xref #3147, closes #3323 
 - [x] Tests added

This is a second attempt. It works with Datasets also in that it can pass Dataset chunks to the user's function. I haven't figured out how to reconstruct a Dataset from things returned from the user's function. For DataArrays, I currently assume that the user function returns chunks of the same shape as the provided chunk.

A few questions:
1. I wonder if we should introduce a meta xarray object that describes the final result that user expects back i.e. a template DataArray or Dataset with chunks, dtype etc. specified. We could have a utility function that makes it easy to create these templates from existing objects.

2. If the user function doesn't return anything, (e.g. it writes to a file), how do I "compute" the graph? In this case, we could rename this function to `apply_blocks` and release that till we figure out point 1?

3. Should we also add `Dataset.to_delayed` and `DataArray.to_delayed`? This would make use-cases like the one in point 2 easier.

ping @mrocklin


xarray/core/parallel.py
```diff
@@ -0,0 +1,141 @@
...
+                coords.append([name, task_name])
+
+        graph[(gname,) + v] = (
+            _wrapper,
+            func,
+            (Dataset, (dict, data_vars), (dict, coords), dataset.attrs),
+            to_array,
+            args,
+            kwargs,
+        )
```
shoyer: To make this work on a `Dataset`, we need one more layer to map a dask graph of (list of array) tasks into a list of dask arrays.

Suppose you change `_wrapper` to return a list of arrays Numpy arrays, as I suggested above. Then this would look something like:
```
for i in range(num_data_vars):
    gname_i = gname + '-out' + str(i)
    graph[(gname_i,) + v] = (operator.getitem, (gname,) + v, i)
```
Now you should simply construct the desired dask arrays from each of `gname_i` instead of `gname`.

dcherian: Yes, I got that to work. Though I use  dict to keep track of variable names.

dcherian:
I can reconstruct Datasets now. This version requires a `template` kwarg to know what the returned object looks like. I'll add some code to infer that next. Suggestions are appreciated.

dcherian:
Ready for review.

@crusaderky I'd like to add `DataArray.map_blocks` and `Dataset.map_blocks` but don't know how to work around the circular import issue.

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            func,
+            (Dataset, (dict, data_vars), (dict, coords), dataset.attrs),
+            input_is_array,
+            args,
+            kwargs,
+        )
+
+        # mapping from variable name to dask graph key
+        var_key_map = {}
+        for name, variable in template.variables.items():
```
crusaderky: You could add ``if name in indexes: continue`` to save some graph nodes.

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+                else:
+                    # unchunked dimensions in the input have one chunk in the result
+                    key += (0,)
+
+            graph[key] = (operator.getitem, from_wrapper, name)
+
+    graph = HighLevelGraph.from_collections(name, graph, dependencies=[dataset])
+
+    result = Dataset()
+    for var, key in var_key_map.items():
```
crusaderky: Inconsistent - earlier you called variable names ``name``

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            result[var] = indexes[var]
+            continue
+
+        dims = template[var].dims
+        var_chunks = []
+        for dim in dims:
+            if dim in input_chunks:
+                var_chunks.append(input_chunks[dim])
+            else:
+                if dim in indexes:
```
crusaderky: elif

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+                    var_chunks.append((len(indexes[dim]),))
+
+        data = dask.array.Array(
+            graph, name=key, chunks=var_chunks, dtype=template[var].dtype
+        )
+        result[var] = DataArray(data=data, dims=dims, name=var)
```
crusaderky: ``result[var] = dims, data``, and again you're accidentally moving non-index coords to data_vars

xarray/tests/test_dask.py
```diff
@@ -878,3 +878,77 @@ def test_dask_layers_and_dependencies():
...
     )
+
+
+def make_da():
+    return xr.DataArray(
+        np.ones((10, 20)),
+        dims=["x", "y"],
+        coords={"x": np.arange(10), "y": np.arange(100, 120)},
+        name="a",
+    ).chunk({"x": 4, "y": 5})
```
crusaderky: missing use case for non-index coords

xarray/tests/test_dask.py
```diff
@@ -878,3 +878,77 @@ def test_dask_layers_and_dependencies():
...
+    xr.testing.assert_equal(expected, actual)
+
+
+@pytest.mark.parametrize("obj", [map_da, map_ds])
+def test_map_blocks_args(obj):
+    import operator
+
+    expected = obj + 10
+    actual = xr.map_blocks(operator.add, obj, 10)
+    xr.testing.assert_equal(expected, actual)
```
crusaderky: Missing use cases: 
- func(DataArray) -> Dataset 
- func(Dataset) -> DataArray
- func output contains less variables
- func output contains new variables
- func changes dtypes
- *args, **kwargs are passed through
- IndexVariables don't accidentally cause the whole graph to be computed (the logic you wrote in the main function is quite subtle!)

xarray/tests/test_dask.py
```diff
@@ -878,3 +878,77 @@ def test_dask_layers_and_dependencies():
...
+
+@pytest.mark.parametrize(
+    "func, obj",
+    [[simple_func, map_da], [simple_func, map_ds], [complicated_func, map_ds]],
+)
+def test_map_blocks(func, obj):
+
+    actual = xr.map_blocks(func, obj)
+    expected = func(obj)
+    xr.testing.assert_equal(expected, actual)
```
crusaderky: assert_equal does not test chunks. You should.

xarray/tests/test_dask.py
crusaderky: Could you add convenience methods?
```python
class Dataset:
    ...
    def map_blocks(self, func, *args, **kwargs):
        from .parallel import map_blocks
        return map_blocks(func, self, *args, **kwargs)
```

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+    ----------
+    func: callable
+        User-provided function that should accept DataArrays corresponding to one chunk.
+        The function will be run on a small piece of data that looks like 'obj' to determine
+        properties of the returned object such as dtype, variable names,
+        new dimensions and new indexes (if any).
+
+        This function cannot
+        - change size of existing dimensions.
+        - add new chunked dimensions.
```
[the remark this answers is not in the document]
crusaderky: It should also be emphasized that, if you don't actually need coords or multiple Dataset variables in parallel, and everything you need is access the underlying numpy array, then apply_ufunc() is going to be _much_ faster.

xarray/tests/test_dask.py
dcherian: I'd like to but i was running into circular import issues when i tried this a while back. Any thoughts on fixing that?

dcherian:
Thanks for the extremely thorough review @crusaderky. I'll work on this slowly over the next few days

xarray/tests/test_dask.py
crusaderky: I wrote the whole thing above...

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+                    )
+
+        to_return = _make_dict(result)
+
+        return to_return
+
+    if not dask.is_dask_collection(obj):
+        raise ValueError(
+            "map_blocks can only be used with dask-backed DataArrays. Use .chunk() to convert to a Dask array."
+        )
```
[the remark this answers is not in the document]
shoyer: What is the right behavior for non-dask objects? Just computing the function eagerly, as if it has one block?

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+
+        This function cannot
+        - change size of existing dimensions.
+        - add new chunked dimensions.
+
+    obj: DataArray, Dataset
+        Chunks of this object will be provided to 'func'. The function must not change
+        shape of the provided DataArray.
+    args:
+        Passed on to func.
```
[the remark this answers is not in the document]
shoyer: I would suggest suggest dedicated keyword arguments for `args` and `kwargs` rather than using `*args, **kwargs`. The problem `*args, **kwargs` is that it limits our ability to add new arguments later, because we've already promised we're passing on everything.

xarray/tests/test_dask.py
shoyer: I agree that these are interesting use cases, but let's not block this PR on all this functionality. For now, I think it would be useful even if this function only works on function that accept/return objects with all the same metadata.

For completeness, we could also add one more use-case:
- func output contains less (or more) dimensions

dcherian: > I wrote the whole thing above...

=) sorry! didn't look too closely.

crusaderky: For

> IndexVariables don't accidentally cause the whole graph to be computed (the logic you wrote in the main function is quite subtle!)

take a look at https://stackoverflow.com/questions/53289286/determine-how-many-times-dask-computed-something

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            if dask.is_dask_collection(variable.data):
+                var_dask_keys = variable.__dask_keys__()
+
+                # recursively index into dask_keys nested list to get chunk
+                chunk = var_dask_keys
+                for dim in variable.dims:
+                    chunk = chunk[chunk_index_dict[dim]]
+
+                task_name = ("tuple-" + dask.base.tokenize(chunk),) + v
+                graph[task_name] = (tuple, [variable.dims, chunk])
```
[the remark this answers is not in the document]
dcherian: no that is considered array-like by the constructor so it doesn't work.

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+            input_is_array,
+            args,
+            kwargs,
+        )
+
+        # mapping from variable name to dask graph key
+        var_key_map = {}
+        for name, variable in template.variables.items():
+            var_dims = variable.dims
+            # cannot tokenize "name" because the hash of <this-array> is not invariant!
```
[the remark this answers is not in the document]
dcherian: I think this is the issue (`_value` in `__hash__()`)

```
class ReprObject:
    """Object that prints as the given value, for use with sentinel values.
    """

    __slots__ = ("_value",)

    def __init__(self, value: str):
        self._value = value

    def __repr__(self) -> str:
        return self._value

    def __eq__(self, other) -> bool:
        if isinstance(other, ReprObject):
            return self._value == other._value
        return False

    def __hash__(self) -> int:
        return hash((ReprObject, self._value))
```

xarray/tests/test_dask.py
dcherian: Yeah we have `CountingScheduler` in `test_dask.py` that does what we want.

dcherian:
Ready for review. I failed hard at adding types, so some help there would be appreciated.

Otherwise I think I've made all changes.

xarray/core/parallel.py
```diff
@@ -0,0 +1,275 @@
...
+    da.name = name
+    return da
+
+
+def make_meta(obj):
+
+    from dask.array.utils import meta_from_array
+
+    if isinstance(obj, DataArray):
+        meta = DataArray(obj.data._meta, dims=obj.dims)
```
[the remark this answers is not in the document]
dcherian: OK haven't done this yet.

xarray/core/parallel.py
```diff
@@ -0,0 +1,310 @@
...
+                chunk = variable.__dask_keys__()
+                for dim in variable.dims:
+                    chunk = chunk[chunk_index_dict[dim]]
+
+                chunk_variable_task = ("tuple-" + dask.base.tokenize(chunk),) + v
+                graph[chunk_variable_task] = (tuple, [variable.dims, chunk])
+            else:
+                # non-dask array with possibly chunked dimensions
+                # index into variable appropriately
+                subsetter = dict()
```
crusaderky: ``subsetter = {}``

xarray/core/parallel.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+
+        # mapping from variable name to dask graph key
+        var_key_map = {}
+        for name, variable in template.variables.items():
+            if name in indexes:
+                continue
+            # cannot tokenize "name" because the hash of ReprObject (<this-array>)
+            # is a function of its value. This happens when the user function does not
+            # set a name on the returned DataArray
```
crusaderky: I really don't understand your point here...

xarray/core/parallel.py
```diff
@@ -0,0 +1,310 @@
...
+            graph[key] = (operator.getitem, from_wrapper, name)
+
+    graph = HighLevelGraph.from_collections(name, graph, dependencies=[dataset])
+
+    result = Dataset()
+    # a quicker way to assign indexes?
+    # indexes need to be known
+    # otherwise compute is called when DataArray is created
+    for name in template.dims:
+        result[name] = indexes[name]
```
crusaderky: ``result = Dataset(coords=indexes)``

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
     assert set(x.foo.__dask_graph__().dependencies).issuperset(
         ds.__dask_graph__().dependencies
     )
+
+
+def make_da():
+    da = xr.DataArray(
+        np.ones((10, 20)),
+        dims=["x", "y"],
+        coords={"x": np.arange(10), "y": np.arange(100, 120)},
```
crusaderky: missing use case of non-index coords (chunked and not chunked)

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+    map_ds.coords["cxy"] = (("x", "y"), map_da.x * map_da.y)
+    map_ds.coords["cxy"] = map_ds.cxy.chunk({"y": 10})
+    map_ds.attrs["test"] = "test"
+    map_ds["xx"] = (map_ds["a"] * map_ds.y).chunk({"y": 20})
+
+    return map_ds
+
+
+# work around mypy error
+# xarray/tests/test_dask.py:888: error: Dict entry 0 has incompatible type "str": "int"; expected "Hashable": "Union[None, Number, Tuple[Number, ...]]"
```
crusaderky: ? what dict?

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+map_ds = make_ds()
+
+
+def test_unify_chunks():
+    with raises_regex(ValueError, "inconsistent chunks"):
+        map_ds.chunks
+
+    expected = {"x": (4, 4, 2), "y": (5, 5, 5, 5)}
+    actual = map_ds.unify_chunks().chunks
+    assert expected == actual
```
crusaderky: Please add ``assert_identical(map_ds, map_ds.unify_chunks())``

Missing use cases:
- DataArray.unify_chunks
- invoke unify_chunks when chunks are already unified (returned object must be a shallow copy)
- invoke unify_chunks when there are no chunks to begin with (returned object must be a shallow copy)

xarray/core/dataset.py
```diff
@@ -5146,5 +5149,106 @@ def filter_by_attrs(self, **kwargs):
...
+        See Also
+        --------
+
+        dask.array.core.unify_chunks
+        """
+        import dask.array
+
+        try:
+            if self.chunks:
+                return self
```
crusaderky: This will subtly break user code if the user relies on this method to always be a shallow copy.
Replace with ``self.copy()``

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+
+
+# TODO: DataArray.chunks is not a dict but Dataset.chunks is!
+def assert_chunks_equal(a, b):
```
crusaderky: I think this should be moved to xarray.testing.

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+    def at_least_one_chunk(obj):
+        new_chunks = {}
+        for dim in obj.dims:
+            if dim not in obj.chunks:
+                new_chunks[dim] = len(obj[dim])
+            else:
+                # must preserve old chunks too
+                new_chunks[dim] = obj.chunks[dim][0]
+        return obj.chunk(new_chunks)
+
```
crusaderky:
```python
if isinstance(a, DataArray) != isinstance(b, DataArray):
    raise TypeError("a and b have mismatched types")
```

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+    if isinstance(a, DataArray):
+        a = a._to_temp_dataset()
+
+    if isinstance(b, DataArray):
+        b = b._to_temp_dataset()
+
+    a = at_least_one_chunk(a.unify_chunks())
+    b = at_least_one_chunk(b.unify_chunks())
```
crusaderky: With this you're stating that the test should pass if a non-dask object is compared against a dask object with whole chunks only. Is it intentional? It doesn't feel right to me.
e.g.
```python
a = DataArray([1, 2])
assert_chunks_equal(a, a.chunk())  # passes
```

xarray/core/dataset.py
```diff
@@ -5146,5 +5149,106 @@ def filter_by_attrs(self, **kwargs):
                 selection.append(var_name)
         return self[selection]
 
+    def unify_chunks(self):
+        """ Unifies chunksize along all chunked dimensions of this Dataset.
```
crusaderky: This could greatly benefit from being moved to top-level with thin wrappers in DataArray and Dataset, so that one can unify chunks across multiple DataArray's and Dataset's.

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+    # do this manually before comparing
+    def at_least_one_chunk(obj):
+        new_chunks = {}
+        for dim in obj.dims:
+            if dim not in obj.chunks:
+                new_chunks[dim] = len(obj[dim])
+            else:
+                # must preserve old chunks too
+                new_chunks[dim] = obj.chunks[dim][0]
+        return obj.chunk(new_chunks)
```
crusaderky: This whole function can be replaced with ``obj.chunk()``.

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+    def bad_func(darray):
+        return (darray * darray.x + 5 * darray.y)[:1, :1]
+
+    with raises_regex(ValueError, "Length of the.* has changed."):
+        xr.map_blocks(bad_func, map_da).compute()
+
+    def returns_numpy(darray):
+        return (darray * darray.x + 5 * darray.y).values
+
+    with raises_regex(ValueError, "Function must return an xarray DataArray"):
```
crusaderky: or Dataset

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+
+
+@pytest.mark.parametrize("obj", [map_da, map_ds])
+def test_map_blocks_convert_args_to_list(obj):
+    expected = obj.unify_chunks() + 10
+    with raise_if_dask_computes():
+        actual = xr.map_blocks(operator.add, obj, [10])
+    assert_chunks_equal(expected, actual)
+    # why is compute needed?
+    xr.testing.assert_equal(expected.compute(), actual.compute())
```
crusaderky: indeed it should not be needed! please investigate

xarray/tests/test_dask.py
```diff
@@ -917,3 +918,176 @@ def test_dask_layers_and_dependencies():
...
+        [lambda x: x.isel(x=1), map_ds],
+        [lambda x: x.isel(x=1).drop("x"), map_da],
+        [lambda x: x.assign_coords(new_coord=("y", x.y * 2)), map_da],
+        [lambda x: x.astype(np.int32), map_da],
+        [lambda x: x.rename({"a": "new1", "b": "new2"}), map_ds],
+    ],
+)
+def map_blocks_transformations(func, obj, expected):
+    with raise_if_dask_computes():
+        assert_equal(xr.map_blocks(func, obj), func(obj))
```
crusaderky: How is it possible that assert_equal doesn't compute anything if it is supposed to compare the contents?

xarray/core/parallel.py
```diff
@@ -0,0 +1,310 @@
...
+def infer_template(func, obj, *args, **kwargs):
+    """ Infer return object by running the function on meta objects. """
+    meta_args = []
+    for arg in (obj,) + args:
+        meta_args.append(make_meta(arg))
+
+    try:
+        template = func(*meta_args, **kwargs)
+    except ValueError:
+        raise ValueError("Cannot infer object returned by user-provided function.")
```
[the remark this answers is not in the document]
dcherian: Good point. I'd like to communicate this error ("Cannot infer object returned by user-provided function.") in addition to the stack trace. How do I do that?

xarray/tests/test_dask.py
dcherian: Yes I agree, forgot to add a comment asking for opinions about that.

xarray/core/parallel.py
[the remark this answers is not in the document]
crusaderky:
```python
except Exception as e:
    raise Exception("User-provided function failed when run with stub data") from e
```

xarray/core/dataarray.py
```diff
@@ -3034,6 +3034,58 @@ def integrate(
         ds = self._to_temp_dataset().integrate(dim, datetime_unit)
         return self._from_temp_dataset(ds)
 
+    def unify_chunks(self):
+        ds = self.copy()._to_temp_dataset().unify_chunks()
+        return self._from_temp_dataset(ds)
+
+    def map_blocks(self, func, args=[], kwargs={}):
```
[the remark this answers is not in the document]
dcherian: How do I define `T`?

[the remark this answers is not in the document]
crusaderky:
```python
from typing import TypeVar
T = TypeVar("T") 
```

xarray/tests/test_dask.py
dcherian: `assert_equal(actual, expected)` passes but `assert_equal(expected, actual)` fails! See #3350 for some weird behaviour.

xarray/tests/test_dask.py
dcherian: How do I test for shallow copies of xarray objects?

crusaderky: assert_identical(in, out) and out is not in

xarray/core/parallel.py
```diff
@@ -0,0 +1,310 @@
...
+        result_is_array = False
+    else:
+        raise ValueError(
+            "Function must return an xarray DataArray or Dataset. Instead it returned %r"
+            % type(template)
+        )
+
+    # If two different variables have different chunking along the same dim
+    # fix that by "unifying chunks"
+    dataset = dataset.unify_chunks()
```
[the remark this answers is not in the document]
dcherian: Agreed. won't apply this automatically.

xarray/tests/test_dask.py
dcherian: OK I call `.chunk()` on objects in the test functions now before sending to `assert_chunks_equal`

xarray/core/dataset.py
dcherian: Yeah that would be good. Let's leave that to a future PR.

xarray/core/parallel.py
```diff
@@ -0,0 +1,310 @@
...
+        obj_array = obj.copy()
+        obj = obj._to_temp_dataset()
+    else:
+        to_array = False
+
+    if isinstance(obj, Dataset):
+        meta = Dataset()
+        for name, variable in obj.variables.items():
+            if dask.is_dask_collection(variable):
+                meta_obj = obj[name].data._meta
```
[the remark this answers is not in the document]
dcherian: copied meta_from_array over

xarray/core/dataarray.py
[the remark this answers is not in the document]
dcherian: `xarray/core/dataarray.py:3060: error: Function "xarray.core.dataarray.DataArray.T" is not valid as a type
`

Can I trouble you to push a commit with the type annotations to this branch?

dcherian:
Thanks for the type annotations, @crusaderky.

`map_blocks` now handles `attrs` and `name` changes so I've switched almost all tests to `assert_identical`. I'm calling `compute` manually till we can figure out what's wrong with `assert_equal` (e.g. #3350). The only exception is `Dataset.to_array()` which doesn't preserve `name`.

I copied over `meta_to_array` to account for `dask < 2.0`

xarray/tests/test_dask.py
```diff
@@ -917,3 +919,217 @@ def test_dask_layers_and_dependencies():
...
+    map_ds.coords["cx"] = ("x", np.arange(len(map_da.x)))
+    map_ds.coords["cx"].attrs["test2"] = "test2"
+    map_ds.attrs["test"] = "test"
+    map_ds.coords["xx"] = map_ds["a"] * map_ds.y
+
+    return map_ds
+
+
+map_da = make_da()
+map_ds = make_ds()
```
jhamman: I think you could wrap `make_da` and `make_ds` in [`pytest.fixture`](https://docs.pytest.org/en/latest/fixture.html) objects and you wouldn't have to do this step at the module level.

xarray/core/dataarray.py
```diff
@@ -3053,6 +3056,75 @@ def integrate(
         ds = self._to_temp_dataset().integrate(dim, datetime_unit)
         return self._from_temp_dataset(ds)
 
+    def unify_chunks(self) -> "DataArray":
+        """ Unifies chunksize along all chunked dimensions of this DataArray.
```
jhamman: Unifies --> Unify (maybe?)

xarray/core/dataset.py
```diff
@@ -5396,5 +5402,105 @@ def filter_by_attrs(self, **kwargs):
...
+        if not dask.is_dask_collection(self):
+            return self.copy()
+
+        try:
+            self.chunks
+            return self.copy()
+        except ValueError:  # "inconsistent chunks"
+            pass
+
+        import dask.array
```
jhamman: can we move this import to the top of the method?

xarray/core/dataarray.py
```diff
@@ -64,6 +65,8 @@
 )
 
 if TYPE_CHECKING:
+    T_DSorDA = TypeVar("T_DSorDA", "DataArray", Dataset)
```
mrocklin: Should this be a Union?

xarray/core/parallel.py
```diff
@@ -0,0 +1,333 @@
...
+
+    template_indexes = set(template.indexes)
+    dataset_indexes = set(dataset.indexes)
+    preserved_indexes = template_indexes.intersection(dataset_indexes)
+    new_indexes = set(template_indexes) - set(dataset_indexes)
+    indexes = {dim: dataset.indexes[dim] for dim in preserved_indexes}
+    indexes.update({k: template.indexes[k] for k in new_indexes})
+
+    graph = {}  # type: Dict[Any, Any]
+    gname = "%s-%s" % (dask.utils.funcname(func), dask.base.tokenize(dataset))
```
mrocklin: This should maybe depend both on the dataset, and on all inputs to this function.  For example the following results should have different names.:

```python
ds.map_blocks(func, kwargs={"foo": 1})
ds.map_blocks(func, kwargs={"foo": 2})
```

mrocklin: Also, has xarray implemented dask.base.normalize_token for DataSets and DataArrays?

xarray/core/parallel.py
```diff
@@ -0,0 +1,333 @@
...
+
+        for name, variable in dataset.variables.items():
+            # make a task that creates tuple of (dims, chunk)
+            if dask.is_dask_collection(variable.data):
+                # recursively index into dask_keys nested list to get chunk
+                chunk = variable.__dask_keys__()
+                for dim in variable.dims:
+                    chunk = chunk[chunk_index_dict[dim]]
+
+                chunk_variable_task = ("tuple-" + dask.base.tokenize(chunk),) + v
```
mrocklin: You might keep the same name as above.  It's probably not necessary to re-tokenize everytihng.  There is also some benefit while debugging in all of them having the same hash, just so that we know what parts of the computation are related.

xarray/core/parallel.py
```diff
@@ -0,0 +1,333 @@
...
+            for dim in variable.dims:
+                if dim in chunk_index_dict:
+                    key += (chunk_index_dict[dim],)
+                else:
+                    # unchunked dimensions in the input have one chunk in the result
+                    key += (0,)
+
+            graph[key] = (operator.getitem, from_wrapper, name)
+
+    graph = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
```
mrocklin: Just verifying here.  Are there likely to be any dask objects in any of `args` or `kwargs`?  People might do something like put a `ds.mean()` value in there somewhere or a dask.delayed value.  Maybe we save this for a future PR, but it might be good to sift through args and kwargs and either verify that they aren't dask collections or else include them in the computation.

xarray/core/dataarray.py
mrocklin: (Disclaimer, I don't know the type system that you all are using, and I don't have much expeience using Python type annotations).

xarray/core/parallel.py
dcherian: good catch! Raising an error now.

xarray/core/parallel.py
dcherian: > This should maybe depend both on the dataset, and on all inputs to this function.

Done.


> Also, has xarray implemented dask.base.normalize_token for DataSets and DataArrays?

Doesn't look like it. (cc @shoyer , @crusaderky )

xarray/core/dataarray.py
dcherian: This is a @crusaderky question.

xarray/tests/test_dask.py
dcherian: pytest doesn;t support fixtures in parametrize so i had to add a workaround but I do think it's nicer.

xarray/core/dataset.py
dcherian: @crusaderky wanted it here to save time I think.

xarray/core/parallel.py
dcherian: thanks. now using the same name and replaced tokenize call with chunk[0]

xarray/core/dataarray.py
crusaderky: No. The whole point of this contrived syntax (which I hate, but I don't think there's any other sensible way) is to allow to define

```python
def f(x: T_DSorDA) -> T_DSorDA
```
which is the same as writing
```python
@overload
def f(x: DataArray) -> DataArray:
    ...

@overload
def f(x: Dataset) -> Dataset:
    ...

def f(x):
    # implementation
```
except that it's a lot more compact, particularly when other arguments start being involved (like in this case), and doesn't explode cartesian-ly (is it a word? :P) when two or more separate overloading combinations enter need to be represented.

In constrast,
```python
def f(x: Union[DataArray, Dataset]) -> Union[DataArray, Dataset]
```
means that the output of the function is unrelated with the input, which causes problems down the line when you use the function; e.g.
```python
x = DataArray()
y = f(x)
y[0] # mypy error: Dataset is not subscriptable
```

If instead you were to use a simpler, unbound TypeVar, then mypy would complain in the implementation of the function:
```python
T = TypeVar("T")

def f(x: T) -> T:
    return x.copy()  # mypy error: object has no attribute "copy"

x = DataArray()
y = f(x)
y[0] # this one instead is ok; mypy knows that y is a DataArray
```

xarray/core/parallel.py
crusaderky: Nope.
```python
>>> a = xarray.DataArray(da.ones(2))
>>> normalize_token(a)
'6743063447a54e87a8bd15ef07733093'
>>> normalize_token(a)
'b2e634be86fb49f6ab7f5deac862f4fe'
```
This looks like a showstopper - I'm afraid you'll have to implement it...

xarray/core/dataarray.py
crusaderky: Forgot to mention that the last syntax completely disables validation on the parameters;
``y = f(123)`` won't raise any errors.

xarray/core/dataset.py
crusaderky: No, I wanted to allow using the function when dask is not installed and there is no dask backend.
But that's nullified by 
```python

        import dask

        if not dask.is_dask_collection(self):
            return self.copy()
```
right above (it was not there when I made the comment)!

Don't we have a wrapper around ``dask.is_dask_collection`` that defaults to a dummy False if dask is not installed?

crusaderky: Correction - that whole paragraph can be flat out removed.
Dataset.chunks returns an empty ``Frozen(SortedKeysDict({}))`` when there are no dask variables already.

xarray/core/parallel.py
crusaderky: In order not to make this PR balloon even more, it may be a good idea to create a new PR for normalize_token, review & merge it, and then go back to work on this one.

mrocklin: It's fine to not implement `normalize_token`.  It is safe to not have it implemented, it just means that you won't get deduplication of equivalent computations.  This is slightly inefficient, but doesn't produce wrong results.  I would recommend merging in this PR, and raising an issue for `normalize_token` some other time.

xarray/core/dataarray.py
mrocklin: That makes sense.  Thanks for the explanation @crusaderky !

xarray/core/parallel.py
dcherian: Opened #3378

xarray/core/dataset.py
dcherian: 🤦‍♂ Now changed to

``` python
        try:
            # if pure numpy
            if len(self.chunks) == 0:
                return self.copy()
        except ValueError:  # "inconsistent chunks"
            pass
```

crusaderky: Wrong again - if you have chunks and they are already consistent, the self.chunks property will return a non-empty mapping.

```python
        try:
            self.chunks
        except ValueError:  # "inconsistent chunks"
            pass
        else:
            # No variables with dask backend, or all chunks are already aligned
            return self.copy()

        # import dask is placed after the quick exit test above to allow
        # running this method if dask isn't installed and there are no chunks
        import dask.array
```

dcherian: ugh, forgot about that.

jhamman:
@crusaderky - are you happy with the state of things here?

I plan to merge this tomorrow evening unless additional reviews come in.

crusaderky:
@jhamman I'd like to give it a final read.
@dcherian merge's failing right now

dcherian:
hahaha the conflict was that you deleted `dask_array_compat`! I added it back in for `meta_from_array`

crusaderky:
@dcherian make_meta is failing on dask 1.2

jhamman:
@crusaderky - are we happy here?

crusaderky:
@dcherian I gave it a polish pass, please review my changes.

dcherian:
Added one line to docstring. Looks good otherwise. Thanks @crusaderky

crusaderky:
Thank you

mrocklin:
Woo!

jhamman:
Fantastic work all! This is going to be a killer addition to the project. 🎉

jakirkham:
Congratulations! 🎉

shoyer:
yes, nice work @dcherian and @crusaderky !

dcherian:
!

PR #3278, opened 2019-09-03
Add head, tail and thin methods

### LINKED ISSUE #319: Add head(), tail() and thin() methods?
These would be shortcuts for `isel`/slice syntax:
- `ds.head(time=5)` -> `ds.isel(time=slice(5))`: select the first five time values
- `ds.tail(time=5)` -> `ds.isel(time=slice(-5, None))`: select the last five time values
- `ds.thin(time=5)` -> `ds.isel(time=slice(None, None, 5))`: select every 5th time value

DangoMelon:
I feel like there's room for improvement in the docstrings, any change or suggestion is welcome!

 - [x] Closes #319 
 - [x] Tests added

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2019,7 +2019,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         indexers : dict, optional
             A dict with keys matching dimensions and integer values `n`.
             One of indexers or indexers_kwargs must be provided.
-        **indexers_kwarg : {dim: n, ...}, optional
+        **indexers_kwargs : {dim: n, ...}, optional
             The keyword arguments form of ``indexers``.
             One of indexers or indexers_kwargs must be provided.
 
@@ -2045,7 +2045,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         indexers : dict, optional
             A dict with keys matching dimensions and integer values `n`.
             One of indexers or indexers_kwargs must be provided.
-        **indexers_kwarg : {dim: n, ...}, optional
+        **indexers_kwargs : {dim: n, ...}, optional
             The keyword arguments form of ``indexers``.
             One of indexers or indexers_kwargs must be provided.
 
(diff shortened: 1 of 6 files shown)


xarray/core/dataset.py
```diff
@@ -2008,6 +2008,90 @@ def sel(
...
+    ) -> "Dataset":
+        """Returns a new dataset with the first `n` values of each array
+        for the specified dimension(s).
+
+        Parameters
+        ----------
+        indexers : dict, optional
+            A dict with keys matching dimensions and integer values `n`.
+            One of indexers or indexers_kwargs must be provided.
+        **indexers_kwarg : {dim: n, ...}, optional
```
```suggestion
        **indexers_kwargs : {dim: n, ...}, optional
```

max-sixty:
Looks great! V good first effort @DangoMelon 

Any thoughts from others? Do we want a default (e.g. 5)?

xarray/core/dataset.py
DangoMelon: Thanks for the correction! I found multiple places where it's written in singular form. I'll see what else I can find to contribute to the docs.

shoyer:
Do we want to support integer values, indicating that an operation should be applied to *all* dimensions? e.g., supposing `ds` has dimensions `x` and `y`, then `ds.head(5)` would be equivalent to `ds.head(x=5, y=5)`.

This would be a natural way to support a default value (if we want that) -- we could simply set the default value to 5, so `ds.head()` -> `ds.head(5)` -> `ds.head(x=5, y=5)`.

shoyer:
But yes, I agree with @max-sixty this looks great, and already ready to merge!

If you don't have time to work on default values and/or integers now, we can always merge this PR now and leave those backwards compatible API changes for later :)

max-sixty:
Great! 

In the spirit of faster iteration, I'll merge this @DangoMelon and then if you want to add default values (I also like the idea of supporting a single value), please go ahead!

DangoMelon:
Thanks @max-sixty @shoyer !! I also agree with your suggestions and will be working on adding the default values during these days.

PR #3283, opened 2019-09-06
Add hypothesis test for netCDF4 roundtrip

takluyver:
Part of #1846: add a property-based test for reading & writing netCDF4 files.

This is the first time I've played with Hypothesis, but it seems to be working - e.g. I got an error with float16, and the [netCDF docs show](https://www.unidata.ucar.edu/software/netcdf/docs/data_type.html) that 16-bit floats are not a supported data type.

However:

- This currently only tests a dataset with a single variable - it could be extended to multiple variables if that's useful.
- It [looks like](https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html#Permitted) netCDF4 should support unicode characters, but it failed when I didn't have `max_codepoint=255` in there. I don't know if that's an expected limitation I'm not aware of, or a bug somewhere. But I thought I'd make the test pass for now.

### DIFF
--- a/properties/test_netcdf_roundtrip.py
+++ b/properties/test_netcdf_roundtrip.py
@@ -14,34 +14,36 @@ settings.load_profile("ci")
 
 an_array = npst.arrays(
     dtype=st.one_of(
-        npst.unsigned_integer_dtypes(), npst.integer_dtypes(),
+        npst.unsigned_integer_dtypes(),
+        npst.integer_dtypes(),
         # NetCDF does not support float16
         # https://www.unidata.ucar.edu/software/netcdf/docs/data_type.html
-        npst.floating_dtypes(sizes=(32, 64))
+        npst.floating_dtypes(sizes=(32, 64)),
     ),
     shape=npst.array_shapes(max_side=3),  # max_side specified for performance
 )
 
 compatible_names = st.text(
     alphabet=st.characters(
-        whitelist_categories=('Ll', 'Lu', 'Nd'),
+        whitelist_categories=("Ll", "Lu", "Nd"),
         # It looks like netCDF should allow unicode names, but removing
         # this causes a failure with 'ά'
-        max_codepoint=255
+        max_codepoint=255,
     ),
-    min_size=1
+    min_size=1,
 )
+
 
 @given(st.data(), an_array)
 def test_netcdf_roundtrip(tmp_path, data, arr):
     names = data.draw(
-        st.lists(compatible_names, min_size=arr.ndim, max_size=arr.ndim, unique=True).map(
-            tuple
-        )
+        st.lists(
+            compatible_names, min_size=arr.ndim, max_size=arr.ndim, unique=True
+        ).map(tuple)
     )
     var = xr.Variable(names, arr)
-    original = xr.Dataset({'data': var})
-    original.to_netcdf(tmp_path / 'test.nc')
+    original = xr.Dataset({"data": var})
+    original.to_netcdf(tmp_path / "test.nc")
 
-    roundtripped = xr.open_dataset(tmp_path / 'test.nc')
+    roundtripped = xr.open_dataset(tmp_path / "test.nc")
     xr.testing.assert_identical(original, roundtripped)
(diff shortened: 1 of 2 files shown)


dcherian:
Thanks @takluyver. LGTM but maybe @Zac-HD can review this and #3285

properties/test_netcdf_roundtrip.py
```diff
@@ -0,0 +1,49 @@
...
+        # https://www.unidata.ucar.edu/software/netcdf/docs/data_type.html
+        npst.floating_dtypes(sizes=(32, 64)),
+    ),
+    shape=npst.array_shapes(max_side=3),  # max_side specified for performance
+)
+
+compatible_names = st.text(
+    alphabet=st.characters(
+        whitelist_categories=("Ll", "Lu", "Nd"),
+        # It looks like netCDF should allow unicode names, but removing
```
Zac-HD: Where is this specified?  It would be useful to provide a link and prehaps explain what's going on with the categories too :smile:

properties/test_netcdf_roundtrip.py
```diff
@@ -0,0 +1,49 @@
...
+settings.load_profile("ci")
+
+
+an_array = npst.arrays(
+    dtype=st.one_of(
+        npst.unsigned_integer_dtypes(),
+        npst.integer_dtypes(),
+        # NetCDF does not support float16
+        # https://www.unidata.ucar.edu/software/netcdf/docs/data_type.html
+        npst.floating_dtypes(sizes=(32, 64)),
```
shoyer: I think we could add `byte_string_dtypes`, `unicode_string_dtypes`, `datetime64_dtypes` and `timedelta64_dtypes` to this list.

properties/test_netcdf_roundtrip.py
```diff
@@ -0,0 +1,49 @@
...
+    names = data.draw(
+        st.lists(
+            compatible_names, min_size=arr.ndim, max_size=arr.ndim, unique=True
+        ).map(tuple)
+    )
+    var = xr.Variable(names, arr)
+    original = xr.Dataset({"data": var})
+    original.to_netcdf(tmp_path / "test.nc")
+
+    roundtripped = xr.open_dataset(tmp_path / "test.nc")
```
shoyer: It it good practice to use `open_dataset` as a context manager so the file gets explicitly closed.

(Garbage collection should also close the file, but that can't always be relied upon.)

jhamman:
@takluyver -

takluyver:
Yes, sorry, it dropped off my radar. I've implemented all the changes suggested above.

The datetime64 & timedelta64 cases don't roundtrip completely, so there are test failures. I've added these in a separate commit at the end, so it's easy to cut that back out if you prefer.

Zac-HD:
> The datetime64 & timedelta64 cases don't roundtrip completely, so there are test failures. I've added these in a separate commit at the end, so it's easy to cut that back out if you prefer.

The basic problem seems to be that Hypothesis is generating `timedelta64[Y]`, but Xarray is dealing with `timedelta64[ns]`:

```
____________________________ test_netcdf_roundtrip _____________________________
tmp_path = PosixPath('/tmp/pytest-of-vsts/pytest-0/test_netcdf_roundtrip0')
data = data(...), arr = array([293], dtype='timedelta64[Y]')
        ...
        with xr.open_dataset(tmp_path / "test.nc") as roundtripped:
>           xr.testing.assert_identical(original, roundtripped)
E           AssertionError: Left and right Dataset objects are not identical
E           
E           Differing data variables:
E           L   data     (0) timedelta64[ns] -106488 days +01:41:02.290448
E           R   data     (0) timedelta64[ns] -106488 days +01:41:02.290449

properties/test_netcdf_roundtrip.py:51: AssertionError
---------------------------------- Hypothesis ----------------------------------
Falsifying example: test_netcdf_roundtrip(
    tmp_path=PosixPath('/tmp/pytest-of-vsts/pytest-0/test_netcdf_roundtrip0'), 
    data=data(...), 
    arr=array([293], dtype='timedelta64[Y]')
)
```

So either that's a pretty serious bug, or you should specify [the `max_period` of the timedelta dtype strategy](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.timedelta64_dtypes).

PR #3285, opened 2019-09-06
Hypothesis tests for roundtrip to & from pandas

takluyver:
Part of #1846: test roundtripping between xarray DataArray & Dataset and pandas Series & DataFrame.

I haven't particularly tried to hunt down corner cases (e.g. dataframes with 0 columns), in favour of adding tests that currently pass. But these tests probably form a useful platform if you do want to ensure corner cases like that behave nicely - just modify the limits and see what fails.

### DIFF
--- a/properties/conftest.py
+++ b/properties/conftest.py
@@ -1,0 +1,5 @@
+from hypothesis import settings
+
+# Run for a while - arrays are a bigger search space than usual
+settings.register_profile("ci", deadline=None)
+settings.load_profile("ci")
--- a/properties/test_encode_decode.py
+++ b/properties/test_encode_decode.py
@@ -6,14 +6,9 @@ These ones pass, just as you'd hope!
 """
 import hypothesis.extra.numpy as npst
 import hypothesis.strategies as st
-from hypothesis import given, settings
+from hypothesis import given
 
 import xarray as xr
-
-# Run for a while - arrays are a bigger search space than usual
-settings.register_profile("ci", deadline=None)
-settings.load_profile("ci")
-
 
 an_array = npst.arrays(
     dtype=st.one_of(
--- a/properties/test_pandas_roundtrip.py
+++ b/properties/test_pandas_roundtrip.py
@@ -29,7 +29,7 @@ def datasets_1d_vars(draw):
     Suitable for converting to pandas dataframes.
     """
     n_vars = draw(st.integers(min_value=1, max_value=3))
-    n_entries = draw(st.integers(min_value=1, max_value=100))
+    n_entries = draw(st.integers(min_value=0, max_value=100))
     dims = ("rows",)
     vars = {}
     for _ in range(n_vars):
--- a/properties/test_pandas_roundtrip.py
+++ b/properties/test_pandas_roundtrip.py
@@ -33,7 +33,7 @@ def datasets_1d_vars(draw):
     dims = ("rows",)
     vars = {}
     for _ in range(n_vars):
-        name = draw(st.text(min_size=1))
+        name = draw(st.text(min_size=0))
         dt = draw(numeric_dtypes)
         arr = draw(npst.arrays(dtype=dt, shape=(n_entries,)))
         vars[name] = xr.Variable(dims, arr)
--- a/properties/test_pandas_roundtrip.py
+++ b/properties/test_pandas_roundtrip.py
@@ -34,10 +34,26 @@ def test_roundtrip_dataarray(data, arr):
     roundtripped = xr.DataArray(original.to_pandas())
     xr.testing.assert_identical(original, roundtripped)
 
+
 @given(numeric_series, st.text())
-def test_roundtrip_pandas_series(ser, name):
+def test_roundtrip_pandas_series(ser, ix_name):
     # Need to name the index, otherwise Xarray calls it 'dim_0'.
-    ser.index.name = name
+    ser.index.name = ix_name
     arr = xr.DataArray(ser)
     roundtripped = arr.to_pandas()
     pd.testing.assert_series_equal(ser, roundtripped)
+
+
+numeric_homogeneous_dataframe = numeric_dtypes.flatmap(
+    lambda dt: pdst.data_frames(columns=pdst.columns(["a", "b", "c"], dtype=dt))
+)
+
+
+@given(numeric_homogeneous_dataframe)
+def test_roundtrip_pandas_dataframe(df):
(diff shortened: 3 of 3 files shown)


max-sixty:
This looks great! I'll let someone who knows hypothesis better do a full review. Thanks for submitting @takluyver !

properties/conftest.py
```diff
@@ -0,0 +1,5 @@
+from hypothesis import settings
+
+# Run for a while - arrays are a bigger search space than usual
+settings.register_profile("ci", deadline=None)
```
Zac-HD: :+1: on moving the settings profile to `conftest.py`, I was about to suggest that on the NetCDF PR.

I'd add `print_blob=True` to this, which prints more details to allow local reproduction of failures in CI.  It's a bit more verbose, so we leave it disabled by default, but if you ever need it you'll be very glad it exists.

properties/test_pandas_roundtrip.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@given(numeric_homogeneous_dataframe)
+def test_roundtrip_pandas_dataframe(df):
+    # Need to name the indexes, otherwise Xarray names them 'dim_0', 'dim_1'.
+    df.index.name = "rows"
+    df.columns.name = "cols"
+    arr = xr.DataArray(df)
+    roundtripped = arr.to_pandas()
+    pd.testing.assert_frame_equal(df, roundtripped)
```
Zac-HD: I would add another half-circuit to each of these tests, in this case `xr.testing.assert_identical(arr, roundtripped.to_xarray())`.

It seems redundant, but the pattern ocasionally picks up on odd cases where e.g. `.to_xarray()` can output a variety of objects which are all converted back to the same Pandas object.  I don't actually expect this to fail if the existing tests are passing, but it could in future.

properties/test_pandas_roundtrip.py
```diff
@@ -0,0 +1,91 @@
...
+def datasets_1d_vars(draw):
+    """Generate datasets with only 1D variables
+
+    Suitable for converting to pandas dataframes.
+    """
+    n_vars = draw(st.integers(min_value=1, max_value=3))
+    n_entries = draw(st.integers(min_value=0, max_value=100))
+    dims = ("rows",)
+    vars = {}
+    for _ in range(n_vars):
```
Zac-HD: This pattern - draw a number, then draw that many elements - is tempting but [tends to be inefficient](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/strategies-that-shrink.rst) when Hypothesis tries to minimse any failures.

The alternative, which we recommend, is to generate collections using the `st.lists()` strategy - that way Hypothesis will be able to operate in terms of elements of the list.

In this case it's probably only worth doing so for either the vars or entries dimension, and keep the other as-is.  If you're keen to do both, it's complicated enough that I'd just fall back on the Hypothesis pandas extension and `.map(pd.DataFrame.to_xarray)` over the result :sweat_smile:

takluyver:
As in my other PR, one suggested addition causes a test failure, and I've put that in the last commit.

properties/test_pandas_roundtrip.py
Zac-HD: You could draw `indices` before the loop, then draw a list of (name, array) tuples.  Or in this case you could use `st.dictionaries()` to, well, generate a list of key-value tuples internally.

The other nice trick would be to draw your index first, and use it's length - deleting elements from that will be *slightly* more efficient than shrinking the `n_elements` parameter.

Putting it all togther, I'd write 

```python
idx = draw(pdst.indexes(dtype="u8", min_size=0, max_size=100))
vars_strat = st.dictionaries(
    keys=st.text(),
    values=npst.arrays(dtype=numeric_dtypes, shape=len(idx)).map(partial(xr.Variable, ("rows",))),
    min_size=1,
    max_size=3,
)
return xr.Dataset(draw(vars_strat), coords={"rows": idx})
```

takluyver: Thanks, that does look neater!

max-sixty:
This seems so close—could we fix the test (maybe that's just a merging of master?) and merge?

max-sixty:
This seems so close—could we fix the test (maybe that's just a merging of master?) and merge?

takluyver:
Merged master, crossing fingers that fixes it.

takluyver:
Nope. I don't understand the error, though it looks like astropy has had something similar: https://github.com/astropy/astropy/issues/6424

Also black is now failing on a number of files not affected here.

properties/test_encode_decode.py
```diff
@@ -10,15 +10,10 @@
 
 import hypothesis.extra.numpy as npst
```
dcherian: These may need to be guarded too using `pytest.importorskip` perhaps? @max-sixty  what do you think?

max-sixty: Yes same as here! https://github.com/max-sixty/xarray/blob/black/properties/test_encode_decode.py#L9

takluyver: Aha, I was being distracted by the other errors around the real one. Let's see if the latest commit helps.

takluyver:
OK, looks like the test failure now is real. Let me know if you want me to comment out the relevant line so the tests pass.

max-sixty:
You're right @takluyver 

It looks like hypothesis tests are running in the normal test suites. Anyone know offhand why that is? e.g. https://dev.azure.com/xarray/xarray/_build/results?buildId=1284

(that doesn't solve the test failure, though)

max-sixty:
> OK, looks like the test failure now is real. Let me know if you want me to comment out the relevant line so the tests pass.

If we want to merge a subset of the tests then that's fine. Ofc even better if we can use these tests to find & fix the errors

Zac-HD:
>> OK, looks like the test failure now is real. Let me know if you want me to comment out the relevant line so the tests pass.

> If we want to merge a subset of the tests then that's fine. Ofc even better if we can use these tests to find & fix the errors

In my experience it's better to open an issue, add an xfail decorator to the test, and merge the tests PR.  Otherwise the initial PR can take a very long time and no other property-based tests get added. 

It's also likely that the person who found the bug is not the best person to fix it, and requiring that they do so in order to merge a useful test just disincentives testing!

shoyer:
> In my experience it's better to open an issue, add an xfail decorator to the test, and merge the tests PR. Otherwise the initial PR can take a very long time and no other property-based tests get added.

+1 let's do that!

takluyver:
OK, I've xfailed it.

Issue #3280, opened 2019-09-04
Deprecation cycles to finish for xarray 0.13


shoyer:
Clean-ups we should definitely do:
- [x] remove deprecated options from `xarray.concat` (deprecated back in July 2015!): https://github.com/pydata/xarray/blob/79dc7dc461c7540cc0b84a98543c6f7796c05268/xarray/core/concat.py#L114-L144 (edit by @max-sixty )
- [x] argument order in `DataArray.to_dataset` (also from July 2015) https://github.com/pydata/xarray/blob/41fecd8658ba50ddda0a52e04c21cec5e53415ac/xarray/core/dataarray.py#L491 (edit by @max-sixty )
- [x] remove the warning back reindex with variables with different dimensions (from 2017). This could either be replaced by replacing dimensions like `sel` or by simply raising an error for now and leaving replacing dimensions for later: https://github.com/pydata/xarray/pull/1639): https://github.com/pydata/xarray/blob/79dc7dc461c7540cc0b84a98543c6f7796c05268/xarray/core/alignment.py#L389-L398 (edit by @max-sixty )
- [x] remove `xarray.broadcast_array`, deprecated back in 2016 in https://github.com/pydata/xarray/commit/52ee95f8ae6b9631ac381b5b889de47e41f2440e (edit by @max-sixty )
- [x] remove `Variable.expand_dims` (deprecated back in August 2017), whose implementation actually looks like it's already broken: https://github.com/pydata/xarray/blob/41fecd8658ba50ddda0a52e04c21cec5e53415ac/xarray/core/variable.py#L1232-L1237 (edit by @max-sixty )
- [x] stop supporting a list of colors in the `cmap` argument (dating back to at least v0.7.0): https://github.com/pydata/xarray/blob/d089df385e737f71067309ff7abae15994d581ec/xarray/plot/utils.py#L737-L745 (edit by @max-sixty )
- [x] push the removal of the `compat` and `encoding` arguments from `Dataset`/`DataArray` back to 0.14. These were only deprecated 7 months ago in https://github.com/pydata/xarray/pull/2703. (edit by @max-sixty )

Clean-ups to consider:
- [x] switch the default reduction dimension of groupby and resample? (https://github.com/pydata/xarray/pull/2366) This has been giving a FutureWarning since v0.11.0, released back in November 2018. We could also potentially push this back to 0.14, but these warnings are a little annoying...
[...]

PR #3288, opened 2019-09-06
Remove deprecated concat kwargs.

dcherian:
- [x] Passes `black . && mypy . && flake8`
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

PR #3294, opened 2019-09-09
Compat and encoding deprecation to 0.14

max-sixty:
Per https://github.com/pydata/xarray/issues/3280

### DIFF
--- a/xarray/core/concat.py
+++ b/xarray/core/concat.py
@@ -112,14 +112,7 @@ def concat(
         raise ValueError("must supply at least one object to concatenate")
 
     if dim is None:
-        warnings.warn(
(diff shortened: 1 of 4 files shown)


xarray/core/concat.py
```diff
@@ -112,14 +112,7 @@ def concat(
         raise ValueError("must supply at least one object to concatenate")
 
     if dim is None:
```
dcherian: See #3288 (I changed `dim=None` to `dim`) in the signature. LGTM otherwise. Feel free to merge

max-sixty: Ah great, thanks

PR #3298, opened 2019-09-09
Accept int value in head, thin and tail

DangoMelon:
Related #3278 
This PR makes the methods `head`, `thin` and `tail` for both `DataArray` and `Dataset` accept a single integer value as a parameter. If no parameter is given, then it defaults to 5.
 - [x] Tests added


xarray/core/dataset.py
```diff
@@ -2065,15 +2084,18 @@ def tail(
         return self.isel(indexers)
 
     def thin(
-        self, indexers: Mapping[Hashable, Any] = None, **indexers_kwargs: Any
+        self,
+        indexers: Union[Mapping[Hashable, Any], int] = None,
```
max-sixty: Is this right? 

```suggestion
        indexers: Union[Mapping[Hashable, int], int] = None,
```

xarray/core/dataset.py
```diff
@@ -2086,6 +2108,13 @@ def thin(
         Dataset.tail
         DataArray.thin
         """
+        if not indexers_kwargs:
```
max-sixty: Good spot

xarray/core/dataset.py
```diff
@@ -2030,20 +2033,30 @@ def head(
         Dataset.thin
         DataArray.head
         """
+        if not indexers_kwargs:
+            if indexers is None:
+                indexers = 5
+            if not isinstance(indexers, int) and not is_dict_like(indexers):
```
max-sixty: Should we do anything with negative numbers?  (Galaxy brain is to call the opposite head / tail method, but I think that's too clever)

xarray/core/dataarray.py
```diff
@@ -1035,52 +1035,54 @@ def sel(
         return self._from_temp_dataset(ds)
 
     def head(
-        self, indexers: Mapping[Hashable, Any] = None, **indexers_kwargs: Any
+        self,
+        indexers: Union[Mapping[Hashable, Any], int] = None,
```
shoyer: While not actually use the default value of 5 in the method signatures, rather than setting it to None?

xarray/core/dataset.py
```diff
@@ -2065,15 +2084,18 @@ def tail(
...
+        **indexers_kwargs: Any
     ) -> "Dataset":
         """Returns a new dataset with each array indexed along every `n`th
         value for the specified dimension(s)
 
         Parameters
         ----------
-        indexers : dict, optional
-            A dict with keys matching dimensions and integer values `n`.
+        indexers : dict or int, default: 5
```
shoyer: Does it make sense to thin by a factor of five by default? Or should we not have a default value? The use case for a default thinning value are less clear to me than defaults for head/tail.

max-sixty: Agree!

dcherian: I think picking a default makes it very convenient to use. And this is a convenience method...

xarray/core/dataarray.py
DangoMelon: I tried to set it there but it would raise `only one must be given` when providing only keyword arguments. This happened because whenever a call was made like this `ds.head(time=3)`, it would find the default value set for `indexers`, cast it into values of a dictionary, and then throw an error because `indexers` and `indexers_kwargs` are both set.  If I leave it to `None`, then I can easily avoid this kind of error by only setting the default value when no parameters are found.

The code below is basically the core of this PR
```python
if not indexers_kwargs:
    if indexers is None:
        indexers = 5
    if not isinstance(indexers, int) and not is_dict_like(indexers):
        raise TypeError("indexers must be a dict or a single integer")
if isinstance(indexers, int):
    indexers = {dim: indexers for dim in self.dims}
```

xarray/core/dataset.py
DangoMelon: I'm kind of new to this typing thing but it seems logical to me. Thanks for the suggestion!

xarray/core/dataset.py
dcherian: I think we should raise an error.

xarray/core/dataarray.py
shoyer: Right, that makes sense. This seems to be about the best we can do then!

DangoMelon:
> 
> 
> Any other feedback before we merge? (Errors are unrelated)

@max-sixty I think the typing errors are due to changing the `Any` to `int` and actually finding `slice`.

xarray/core/dataset.py
```diff
@@ -2030,20 +2033,35 @@ def head(
...
+            if not isinstance(indexers, int) and not is_dict_like(indexers):
+                raise TypeError("indexers must be a dict or a single integer")
+        if isinstance(indexers, int):
+            indexers = {dim: indexers for dim in self.dims}
         indexers = either_dict_or_kwargs(indexers, indexers_kwargs, "head")
+        for v in indexers.values():
+            if not isinstance(v, int):
+                raise TypeError("indexer value must be an integer")
+            elif v < 0:
+                raise ValueError("indexer value must be positive")
```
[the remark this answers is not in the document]
DangoMelon: Hmmm, Something along these lines maybe?
```python
"expected integer as indexer value, found type %r for dim %r" % (type(v), k)
```
and

```python
"expected positive integer as indexer value for dim %r" % k
```

The `k` and `v` come from iterating over `indexers.items()`

shoyer:
I would try to include value in the message, too, which can often be a
useful hint about what went wrong, e.g., “expected positive integer for
dimension ‘x’, got -1”

max-sixty:
Here's a solution to the mypy issue:

```diff
diff --git a/xarray/core/dataset.py b/xarray/core/dataset.py
index 1d762114..ee4ebf4d 100644
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2046,8 +2046,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
                 raise TypeError("indexer value must be an integer")
             elif v < 0:
                 raise ValueError("indexer value must be positive")
-        indexers = {k: slice(val) for k, val in indexers.items()}
-        return self.isel(indexers)
+        indexers_slices = {k: slice(val) for k, val in indexers.items()}
+        return self.isel(indexers_slices)
 
     def tail(
         self,
@@ -2087,11 +2087,11 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
                 raise TypeError("indexer value must be an integer")
             elif v < 0:
                 raise ValueError("indexer value must be positive")
-        indexers = {
+        indexers_slices = {
             k: slice(-val, None) if val != 0 else slice(val)
             for k, val in indexers.items()
         }
-        return self.isel(indexers)
+        return self.isel(indexers_slices)
 
     def thin(
         self,
@@ -2134,8 +2134,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
                 raise ValueError("indexer value must be positive")
             elif v == 0:
                 raise ValueError("step cannot be zero")
-        indexers = {k: slice(None, None, val) for k, val in indexers.items()}
-        return self.isel(indexers)
+        indexers_slices = {k: slice(None, None, val) for k, val in indexers.items()}
+        return self.isel(indexers_slices)
 
     def broadcast_like(
         self, other: Union["Dataset", "DataArray"], exclude: Iterable[Hashable] = None
```

xarray/core/dataset.py
DangoMelon: Thanks!

DangoMelon:
@shoyer I hope it's much clear now, I tried to phrase what you suggested.
@max-sixty That did the trick!
Thank you both for your suggestions.

shoyer:
thanks @DangoMelon !

jhamman:
@DangoMelon - thanks for your contribution. This will get used a lot!

PR #3309, opened 2019-09-15
Fix DataArray api doc

DangoMelon:
Seems like I forgot to point `head`, `tail` and `thin` to the right direction in the `DataArray` api documentation


jhamman:
Looks good. Thanks @DangoMelon!

max-sixty:
Thanks @DangoMelon !

PR #3314, opened 2019-09-17
move auto_combine deprecation to 0.14

### LINKED ISSUE #3280: Deprecation cycles to finish for xarray 0.13

dcherian:
- [x] Closes #3280

This undoes the `auto_combine` deprecation until we figure out the best way to proceed.

(I am not very familiar with `auto_combine` so someone else should look over this. The tests all pass though...)

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -915,7 +915,7 @@ def open_mfdataset(
             # Remove this after deprecation cycle from #2616 is complete
             basic_msg = dedent(
                 """\
-            In xarray version 0.13 the default behaviour of `open_mfdataset`
+            In xarray version 0.14 the default behaviour of `open_mfdataset`
             will change. To retain the existing behavior, pass
             combine='nested'. To use future default behavior, pass
             combine='by_coords'. See
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -716,7 +716,7 @@ def auto_combine(
     if not from_openmfds:
         basic_msg = dedent(
             """\
-        In xarray version 0.13 `auto_combine` will be deprecated. See
+        In xarray version 0.14 `auto_combine` will be deprecated. See
         http://xarray.pydata.org/en/stable/combining.html#combining-multi"""
         )
         warnings.warn(basic_msg, FutureWarning, stacklevel=2)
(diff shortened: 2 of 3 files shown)


shoyer:
Can we actually keep the warnings, but just change 0.13 to 0.14?

It's true that `combine_nested` and `combine_by_coords` only solve *most* of the use cases for `auto_combine`, but I'm still not sure it's worth keeping around the later.

PR #3316, opened 2019-09-18
Clarify that "scatter" is a plotting method in what's new.

shoyer:
When I read this, I thought it was referring to scattering data somehow :).


max-sixty:
Yes, shall we push to stable now?

dcherian:
Haha yeah let's push to stable

shoyer:
I pushed it into the "stable" branch, too

PR #3325, opened 2019-09-19
Fix DataArray.to_netcdf type annotation

ttung:
It calls DataSet.to_netcdf, which returns Union[bytes, "Delayed", None].  So this should as well.


shoyer:
Thanks!

PR #3045, opened 2019-06-25
Add rename_vars and rename_dims

jukent:
- [x] Closes #3026
 - [x] Tests added
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

Issue #2838, opened 2019-03-21
What does `swap_dims` do?


tomchor:
Basically I'm trying to swap the names (and just the names) of two dimensions in my `DataArray` and I was looking for options that could do that. 

So I thought `swap_dims` would be a good candidate for that given its description but it doesn't work. In fact, based on the docs, I have no idea how to use it and what it does. Based off of issues like [this](https://github.com/pydata/xarray/issues/1081) it seems I'm not alone.

Maybe it would be good to add example of usage of this function?

shoyer:
`swap_dims` would probably be better called `rename_dims`. It replaces one dimension with another on all variables in a Dataset.

PR #3329, opened 2019-09-21
fix the name of the return value in the swap_dims docstrings

keewis:
- [ ] Tests added

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1453,7 +1453,7 @@ class DataArray(AbstractArray, DataWithCoords):
 
         Returns
         -------
-        renamed : Dataset
+        swapped : Dataset
             DataArray with swapped dimensions.
 
         See Also
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2676,7 +2676,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
 
         Returns
         -------
-        renamed : Dataset
+        swapped : Dataset
             Dataset with swapped dimensions.
 
         See Also


dcherian:
Thanks @keewis I just remembered issue https://github.com/pydata/xarray/issues/2838. I am sorry I didn't point this out earlier. It looks like we want to change this from `swap_dims` to `rename_dims`. Are you up for doing that?

I think the way to proceed would be to change the name of this method to `rename_dims`; add a new `swap_dims` method that raises a DeprecationWarning and calls the new `rename_dims`

keewis:
from what I understand of how `swap_dims` works, that name change does not really reflect what it does:
```python
>>> a = xr.DataArray(data=[0, 1], coords={"x": ["a", "b"], "x2": ("x", [0, 1])}, dims="x")
>>> a
<xarray.DataArray (x: 2)>
array([0, 1])
Coordinates:
  * x        (x) <U1 'a' 'b'
    x2       (x) int64 0 1
>>> a.swap_dims({"x": "x2"})
<xarray.DataArray (x2: 2)>
array([0, 1])
Coordinates:
    x        (x2) <U1 'a' 'b'
  * x2       (x2) int64 0 1
```
I think that rather than renaming `swap_dims`, an entirely new function named `rename_dims` (which Dataset got in #3045) would be better? Also, from #2838 it seems that adding examples to `swap_dims` should help?

keewis:
oh, and I would certainly like to extend this PR to fix this (but only once #3238 is done)

dcherian:
:flushed:

keewis:
say, would it help to add my example to the docstring? If so, should I open a new PR for that?

dcherian:
Yes and yes.

PR #3331, opened 2019-09-21
Improve the documentation of swap_dims

keewis:
This adds examples to both `Dataset.swap_dims` and `DataArray.swap_dims` (and fixes an oversight on my part in #3329).

From #2838, it seems that there was the intention to rename it, but as far as I can tell, that decision is not official yet (and `Dataset` even got its own new `rename_dims`).

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -25,6 +25,10 @@ Bug fixes
   objects remain unaddressable by weakref in order to save memory.
   (:issue:`3317`) by `Guido Imperiale <https://github.com/crusaderky>`_.
 
+Documentation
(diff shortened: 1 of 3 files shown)


keewis:
done.

The error in building the docs seems to have something to do with plotting and "tight_layout" (the reason is the warning from `plotting.rst`, line 528: `Tight layout not applied. tight_layout cannot make axes height small enough to accommodate all axes decorations`), so I guess this is unrelated to my changes?

max-sixty:
Thanks @keewis! 

Yes that failure looks unrelated.

dcherian:
Thanks @keewis!

PR #3333, opened 2019-09-22
More doc fixes

keewis:
While going through the docs, I noticed several issues:
* in `reshaping.rst`, a code block gets hidden because of a missing newline
* in `api.rst`, the section `What parts of xarray are considered public API?` is referenced (I think), but the link is broken.
* the code samples in `DataArray.plot` are not recognized as such because of missing newlines
* the documentation of `Dataset.plot.scatter` in the API reference is broken

This PR tries to fix all of those, but while the first three issues are easy to fix, I'm not sure whether my proposed fix for the one involving `scatter` is correct: I've added `plot.dataset_plot.scatter` to `plot.__all__` (exposing `xr.plot.scatter`) and changed `api.rst` to reference `plot.scatter` instead of `Dataset.plot.scatter`. It seems harmless to me, but the information that `scatter` only exists for `Dataset` is less visible. Is that the way to fix this?

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -8,7 +8,7 @@ This page provides an auto-generated summary of xarray's API. For more details
 and examples, refer to the relevant chapters in the main part of the
 documentation.
 
-See also: :ref:`public api`_.
+See also: :ref:`public api`
 
 Top-level functions
 ===================
--- a/doc/reshaping.rst
+++ b/doc/reshaping.rst
@@ -156,6 +156,7 @@ represented by a :py:class:`pandas.MultiIndex` object. These methods are used
 like this:
 
 .. ipython:: python
+
         data = xr.Dataset(
             data_vars={'a': (('x', 'y'), [[0, 1, 2], [3, 4, 5]]),
                       'b': ('x', [6, 7])},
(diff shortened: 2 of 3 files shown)


dcherian:
> I've added plot.dataset_plot.scatter to plot.__all__ (exposing xr.plot.scatter) and changed api.rst to reference plot.scatter instead of Dataset.plot.scatter. It seems harmless to me, but the information that scatter only exists for Dataset is less visible. Is that the way to fix this?

I don't know about this. We don't test for `xr.plot.scatter`. I don't know why adding `Dataset.plot.scatter` doesn't work. It would be good to track that down.

dcherian:
The rest looks good though. Thanks @keewis

keewis:
Then either we have to start testing for it or modify the docs of [`Dataset.plot`](https://xarray.pydata.org/en/latest/generated/xarray.Dataset.plot.html), because there that method is mentioned.

keewis:
I tried tracking it down, but it uses the same mechanism as `DataArray.plot` where a new instance of `_PlotMethods` is assigned on each `getattr` on the property. The documentation on `DataArray.plot` skips this issue entirely by pointing to `xr.plot`.

dcherian:
Yes we have similar issues with `GroupBy`, `Resample` and `rolling` objects. methods like `mean` etc. aren't shown in the docs. There must be some way to tell sphinx to look for these "injected" methods (I think that's what they're called).

How about you remove the `scatter` changes for now and we can merge the rest?

keewis:
done

dcherian:
Thanks!

Issue #3324, opened 2019-09-19
`sel` fails confusingly or silently when a dimension name matches an optional argument


gwgundersen:
Given that many Xarray methods accept `dict` or keyword arguments, this may effect other methods. That said, here is a minimal example with `sel`.

#### Minimal example

If I want to select based on a dimension named `method`, I cannot because Xarray thinks `method` is the `method` argument to `sel`:

```python
>>> da1 = xr.DataArray(range(3), dims=['method'], coords={'method': range(3)})
>>> da1
<xarray.DataArray (method: 3)>
array([0, 1, 2])
Coordinates:
  * method   (method) int64 0 1 2
>>> da1.sel(method=0)
...
TypeError: ``method`` must be a string
```

And if the `method` dimension has string labels, this fails silently:

```python
>>> da2 = xr.DataArray(range(3), dims=['method'], coords={'method': list('abc')})
>>> da2.sel(method='a')
<xarray.DataArray (method: 3)>
array([0, 1, 2])
Coordinates:
  * method   (method) <U1 'a' 'b' 'c'
```

#### Expected Output

I think raising a `ValueError` and providing a clarifying error message is the right call, but maybe the maintainers have a different opinion. At the very least, it seems like Xarray could note that the `DataArray` instance has a dimension that matches one of the function arguments and ask the user to use `dict`-like arguments if required.

I imagine a general error handling function could be written to check this for any function and `DataArray` pair.

max-sixty:
Yes, agree this isn't ideal. One of the great qualities of xarray is the code we write in exploration also works in production, and this behavior violates some of that.
I'd vote for a warning as getting some of the way to a solution

FWIW passing a dict to these methods will work, if you want complete dependability

gwgundersen:
What if I create a PR with a function in `utils` that takes an array and a list of keyword arguments, and warns the user if the array has any of the keyword arguments as dimensions. The warning can suggest using a `dict`.

dcherian:
+1. Though I prefer raising a more informative over raising a warning.

max-sixty:
> +1. Though I prefer raising a more informative over raising a warning.

Raising an error?

dcherian:
Yes. fat fingers on mobile :)

max-sixty:
Yes I agree, on reflection

shoyer:
[...]

PR #3336, opened 2019-09-24
Provide better error message when dimension name matches argument

gwgundersen:
WIP PR for https://github.com/pydata/xarray/issues/3324. Since `either_dict_or_kwargs` is called 57 times, I don't want to modify every function call without a preliminary review. My main question is if using `locals()` at the top of each function is considered acceptable. Of course, if `locals()` is called _after_ some local variables are created, this code could raise an error when no conflict exists. But explicitly passing in the function argument names in all 57 functions seems brittle. I'd appreciate any thoughts on the PR before I modify all the functions and write some tests.

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1999,7 +1999,11 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         Dataset.isel
         DataArray.sel
         """
-        indexers = either_dict_or_kwargs(indexers, indexers_kwargs, "sel")
+        func_args = set(locals().keys())
+        dims = set(self.dims)
+        indexers = either_dict_or_kwargs(
+            indexers, indexers_kwargs, "sel", func_args=func_args, dims=dims
+        )
         pos_indexers, new_indexes = remap_label_indexers(
             self, indexers=indexers, method=method, tolerance=tolerance
         )
--- a/xarray/core/utils.py
+++ b/xarray/core/utils.py
@@ -249,7 +249,16 @@ def either_dict_or_kwargs(
     pos_kwargs: Optional[Mapping[Hashable, T]],
     kw_kwargs: Mapping[str, T],
     func_name: str,
+    func_args: Any = None,
+    dims: Any = None,
 ) -> Mapping[Hashable, T]:
+    if func_args is not None:
+        inter = func_args.intersection(dims)
+        if inter:
+            raise ValueError(
+                "the dimension name '%s' matches an argument "
+                "to .%s" % (inter.pop(), func_name)
+            )
     if pos_kwargs is not None:
         if not is_dict_like(pos_kwargs):
             raise ValueError(


max-sixty:
That's a clever approach. It's a bit magic, but maybe the right tradeoff. Are there any performance implications?

I agree with @gwgundersen re writing each item out being impractical, unless we're only doing for `sel` & `isel`. 

Any broader thoughts re the need to apply to all methods, vs this sort of magic?

xarray/core/utils.py
```diff
@@ -249,7 +249,16 @@ def either_dict_or_kwargs(
...
     func_name: str,
+    func_args: Any = None,
+    dims: Any = None,
 ) -> Mapping[Hashable, T]:
+    if func_args is not None:
+        inter = func_args.intersection(dims)
+        if inter:
+            raise ValueError(
+                "the dimension name '%s' matches an argument "
+                "to .%s" % (inter.pop(), func_name)
```
dcherian: "Please pass a dictionary to avoid this conflict." or something similar?

xarray/core/dataset.py
```diff
@@ -1999,7 +1999,11 @@ def sel(
         Dataset.isel
         DataArray.sel
         """
-        indexers = either_dict_or_kwargs(indexers, indexers_kwargs, "sel")
+        func_args = set(locals().keys())
+        dims = set(self.dims)
+        indexers = either_dict_or_kwargs(
```
dcherian: I think the magic is OK for the convenience; but is there a better way than `locals()`? Like `inspect`?

xarray/core/utils.py
gwgundersen: Yes, will do.

xarray/core/dataset.py
gwgundersen: So this works

```python
func_args = set(inspect.signature(self.sel).parameters)
```

and is less brittle than `locals()` without much function-specific overhead. `either_dict_or_kwargs` could even take `self.sel` as an argument, and call `inspect` itself. I did a basic performance test, and `locals()` is 1/100th of a second faster on average.

---

That said, do we want to _not_ raise an error on something like this?

```python
arr = xr.DataArray(np.random.random((100, 10)),
                   dims=['foo', 'method'],
                   coords={
                       'foo': range(100),
                       'method': range(10)
                   })
 arr.sel(foo=range(0, 100, 2))
```

My current implementation will raise an error because while the user did not pass in a value for `method`, it is still a function parameter. One way to address this is something like

```python
locals_ = locals()
params = inspect.signature(self.sel).parameters
func_args = {p for p in params if locals_[p] != params[p].default}
```

but now we're getting even more complicated.

max-sixty:
>  I did a basic performance test, and locals() is 1/100th of a second faster on average.

Do you know what the absolute numbers are? i.e. how much performance penalty does the faster routine introduce?

max-sixty:
Another alternative is a decorator that can encapsulate some of this meta-programming (could also do the work of `either_dict_or_kwargs`)

But if we're only doing this for a couple of methods, I would advocate we just write out the keywords. I know it's a bit of a inelegant defeat, but it's pythonic. 

`sel` & `isel` are probably more important than the others

@shoyer do you have thoughts re encapsulation vs magic?

gwgundersen:
I'm almost done with classes and am happy to return to this if there is any interest. I don't want to design something brittle, though, and would love some guidance.

max-sixty:
@gwgundersen I'm really sorry to have kept you waiting, especially after this PR and the others strong ones you've recently done. I temporarily paused my xarray oversight while I came up to speed with a new job and let this drop.

@dcherian any chance you could take a glance? Otherwise I'll look properly this weekend. Thanks.

xarray/core/dataset.py
dcherian: I don't know what to do here. Perhaps @crusaderky has some ideas?

crusaderky: It seems like it's seriously overengineered to me. You really shouldn't be using inspect.signature when you're dealing with a single, hardcoded function...

dcherian: Well the idea is to use a helper function anytime we allow dict or kwargs. `sel` is the test case for now

gwgundersen:
The more I think about this PR, the more I dislike this approach. The solution must either be brittle or over-engineered. I discussed this issue with a friend, and other approaches don't seem better: currying the function—`arr.sel(method="nearest")(method="foo")`—or adding a decorator that caches the results of a stripped down `inspect.signature`. I think the best approach is to just raise a `ValueError` if certain string arguments are not in a predefined set of choices:

```
if method not in ["nearest", ...]:
    raise ValueError(...)
```

@max-sixty, @dcherian, @shoyer thoughts?

---

Not to open a can of worms, but the root cause of this issue is that the Xarray API accepts both `**kwargs` and ordinary named args. This PR—and methods like `either_dict_or_kwargs`—seems like workarounds to accommodate this model. Is there any interest in moving away from this API long-term?

max-sixty:
It's a shame this stalled. 

> I think the best approach is to just raise a `ValueError` if certain string arguments are not in a predefined set of choices:

I think this would be a good solution without any tradeoffs. If anyone is up for implementing that, I would be keen to merge...

Issue #3222, opened 2019-08-15
Minimum versions for optional libraries


crusaderky:
In CI there are:

- tests for all the latest versions of all libraries, mandatory and optional (py36, py37, py37-windows)
- tests for the minimum versions of the mandatory libraries only (py35-min)

There are no tests for legacy versions of the optional libraries.

Today I tried downgrading dask in the py37 environment to dask=1.1.2, which is 6 months old...

**...it's a bloodbath.** 383 errors of the most diverse kind.

In the codebase I found mentions to much older minimum versions: installing.rst mentions dask >=0.16.1, and Dataset.chunk() even asks for dask>=0.9.

It think we should add CI tests for old versions of the optional dependencies.
What policy should we adopt when we find an incompatibility? How old a library should be not to bother fixing bugs and just require a newer version? I personally would go for an aggressive 6 months worth' of backwards compatibility; less if the time it takes to fix the issues is excessive.
The tests should run on py36 because py35 builds are becoming very scarce in anaconda.

This has the outlook of being an exercise in extreme frustration. I'm afraid I personally hold zero interest towards packages older than the latest available in the anaconda official repo, so I'm not volunteering for this one (sorry).

I'd like to hear other people's opinions and/or offers of self-immolation... :)

shoyer:
> Today I tried downgrading dask in the py37 environment to dask=1.1.2, which is 6 months old...

Yikes, we should fix this.

It has not been hard to ensure that older version of NumPy and pandas work. The key thing is CI tests that verify things as PR goes in. If work-arounds for old versions are hard, then we bump minimum required dependencies.

Probably the best place to start would be adding a CI build with pinned versions of the oldest required versions that works today, and then looking if there are simple fixes that will let us downgrade project like dask. It's not worth a huge effort, but it's not very user friendly for everything to broken with old versions of optional dependencies either.

crusaderky:
How does this sound as a rolling policy?
- **Python**: latest 2 major versions published in official anaconda releases (as opposed to just the conda repository), plus the latest major version available version in conda-forge. Latest available minimum versions. This today translates to =3.6 and =3.7 (with no minor versions, so actually 3.6.7 and 3.7.4); 3.8 to be added as soon as conda-forge rebuilds its own stack with it.
- **numpy, pandas**: latest major version at least 1 year old; latest available minor version (which may be newer than 1 year old). Today this translates to numpy=1.15, pandas=0.23.
[...]

Issue #3154, opened 2019-07-20
pynio causes dependency conflicts in py36 CI build


crusaderky:
On Saturday night, all Python 3.6 CI builds started failing. Python 3.7 is unaffected.
See https://dev.azure.com/xarray/xarray/_build/results?buildId=362&view=logs

MacOSX py36:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - pynio
  - python=3.6
  - rasterio
```

Linux py36:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - cfgrib[version='>=0.9.2']
  - h5netcdf
  - pynio
```

HasanAhmadQ7:
I am getting a different error for linux py36

Linux py36
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - cfgrib[version='>=0.9.2']
  - h5py
  - pynio
```

shoyer:
I can reproduce this issue locally:
```
$ conda create -n xarray-tmp -c conda-forge python=3 pynio h5py
Collecting package metadata (current_repodata.json): done
Solving environment: failed with current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:



Package hdfeos5 conflicts for:
pynio -> hdfeos5==5.1.15=0
Package zlib conflicts for:
pynio -> zlib
python=3 -> zlib[version='1.2.*|1.2.11|1.2.8|>=1.2.11,<1.3.0a0']
Package hdfeos2 conflicts for:
pynio -> hdfeos2
Package libgfortran conflicts for:
pynio -> libgfortran[version='>=3.0,<4.0.0.a0|>=3.0.1,<4.0.0.a0']
Package libuuid conflicts for:
pynio -> libuuid[version='>=2.32.1,<3.0a0']
Package hdf5 conflicts for:
pynio -> hdf5[version='1.10.1|1.8.15.1|>=1.10.2,<1.10.3.0a0|>=1.10.4,<1.10.5.0a0|>=1.10.5,<1.10.6.0a0']
[...]

PR #3199, opened 2019-08-09
Add an example of ERA5 and GRIB data & visualization to the gallery

zbruick:
This was from the SciPy sprint, but failed to get submitted as a PR yet. This was developed by myself and @stephansiemen to demonstrate visualizing ERA5 GRIB data. The data was already added in https://github.com/pydata/xarray-data/pull/14.

PR #3340, opened 2019-09-24
CI environments overhaul

crusaderky:
Propaedeutic CI work to #3222.

- py36 and py37 are now identical
- Many optional dependencies were missing in one test suite or another (see details below)
- Tests that require hypothesis now always run if hypothesis is installed
- py37-windows.yml requirements file has been rebuilt starting from py37.yml
- Sorted requirements files alphabetically for better maintainability
- Added black. This is not needed by CI, but I personally use these yaml files to deploy my dev environment and I would expect many more developers to do the same.
  Alternatively, we could go the other way around and remove flake8 from everywhere and mypy from py36 and py37-windows. IMHO the marginal speedup would not be worth the complication.

Added packages to py36.yml (net of changes in order):
+ black
+ hypothesis
+ nc-time-axis
+ numba
+ numbagg
+ pynio (https://github.com/pydata/xarray/issues/3154 seems to be now fixed upstream)
+ sparse

Added packages to py37.yml (net of changes in order):

+ black
+ cdms2
+ hypothesis
+ iris>=1.10
+ numba (previously implicitly installed from pip by numbagg; now installed from conda)
+ pynio

Added packages to py37-windows.yml (net of changes in order):

+ black
+ bottleneck
+ flake8
+ hypothesis
+ iris>=1.10
+ lxml
+ mypy==0.720
+ numba
+ numbagg
+ pseudonetcdf>=3.0.1
+ pydap
+ sparse


crusaderky:
Differences between the requirements files after the changes:

```patch
--- py36.yml
+++ py37.yml
@@ -5 +5 @@
-  - python=3.6
+  - python=3.7

--- py37.yml
+++ py37-windows.yml
@@ -9 +9 @@
-  - cdms2
+  # - cdms2  # Not available on Windows
@@ -31 +31 @@
-  - pynio
+  # - pynio  # Not available on Windows
```

rabernat:
For this PR, it could be worth considering the doc built challenges we are discussing over in https://github.com/pydata/xarray/pull/3199#issuecomment-534702846.

In that case, the doc build succeeded on CI but still failed on RTD. In theory, that shouldn't be possible.

setup.cfg
```diff
@@ -3,7 +3,7 @@ universal = 1
 
 [tool:pytest]
 python_files=test_*.py
-testpaths=xarray/tests
+testpaths=xarray/tests properties
```
max-sixty: IIUC, do we want pytest running hypothesis tests?

crusaderky: Why not? Is there a risk that they'll randomly start failing?

Note the added line to the test script that disables the whole thing if the hypothesis package is not installed.

crusaderky:
@rabernat I'll look into it but on first inspection it feels like scope creep to me and I think it should be dealt with separately. Note how I didn't touch ``doc/environment.yml``.

setup.cfg
crusaderky: Also the whole test suite runs in 900ms for me

crusaderky:
Ready for review

setup.cfg
max-sixty: Um, OK, then I take that back. I thought hypothesis took much longer!

crusaderky: I've added ``@pytest.mark.slow``; that should settle the argument I believe?

shoyer:
This looks great to me!

@crusaderky when you're ready, feel free to merge

setup.cfg
max-sixty: Great!

crusaderky:
Thanks all!

shoyer:
thank you, for tackling this!

PR #3346, opened 2019-09-25
CI test suites with pinned minimum dependencies

crusaderky:
Second step towards resolving #3222.
Added two suites of CI tests:
- Pinned minimum versions for all optional dependencies, except NEP18-dependant ones
- Pinned minimum versions for NEP18 optional dependencies - at the moment only sparse; soon also pint (#3238)

**All versions are the frozen snapshot of what py36.yml deploys today**.
This PR ensures that we won't have accidental breakages _from this moment on_. 
I made no effort  to try downgrading to sensible obsolete versions, as that would require a completely different order of magnitude of work. 
I would suggest to proceed  with the downgrades (and consequent bugfixes) over several small, iterative future PRs that build upon this framework.


crusaderky:
Ready for review

azure-pipelines.yml
```diff
@@ -8,8 +8,12 @@ jobs:
 - job: Linux
   strategy:
     matrix:
-      py35-min:
-        conda_env: py35-min
+      py35-bare-minimum:
+        conda_env: py35-bare-minimum
+      py36-min:
+        conda_env: py36-min
```
max-sixty: Do we need a py36-min as well as a py35-min?

shoyer: maybe `py36-min-all-deps` would be a more descriptive alternative name.

ci/requirements/py36-min.yml
```diff
@@ -0,0 +1,43 @@
...
+  - coveralls
+  - dask=2.4.0
+  - distributed=2.4.0
+  - flake8
+  - h5netcdf=0.7.4
+  - h5py=2.10.0
+  - hdf5=1.10.5
+  - hypothesis
+  - iris=2.2.0
+  - lxml=4.4.1  # FIXME what's the purpose? (added in #1776)
```
shoyer: I *think* this may be an undeclared dependency of pydap.

azure-pipelines.yml
crusaderky: done

crusaderky:
Thanks all

PR #3353, opened 2019-09-28
Remove `complex.nc` from built docs

gwgundersen:
I assume this is related to https://github.com/pydata/xarray/issues/3297 or an associated issue. Basically, a file, `complex.nc`, is being added to `doc/` after running `make clean && make html`. Would be nice to suppress this.


max-sixty:
Many thanks @gwgundersen !

PR #3362, opened 2019-10-01
Fix concat bug when concatenating unlabeled dimensions.

dcherian:
This fixes the following behaviour. (downstream issue https://github.com/xgcm/xgcm/issues/154)

```
def test_concat(self, data):
        split_data = [
            data.isel(dim1=slice(3)),
            data.isel(dim1=3), # this wouldn't work on master
            data.isel(dim1=slice(4, None)),
        ]
        assert_identical(data, concat(split_data, "dim1"))
```

### DIFF
--- a/xarray/tests/test_concat.py
+++ b/xarray/tests/test_concat.py
@@ -41,6 +41,9 @@ def test_concat_compat():
     for var in ["has_x", "no_x_y"]:
         assert "y" not in result[var]
 
+    with raises_regex(ValueError, "One or more of the specified variables"):
+        concat([ds1, ds2], dim="q")
+
 
 class TestConcatDataset:
     @pytest.fixture
--- a/xarray/tests/test_concat.py
+++ b/xarray/tests/test_concat.py
@@ -41,8 +41,10 @@ def test_concat_compat():
     for var in ["has_x", "no_x_y"]:
         assert "y" not in result[var]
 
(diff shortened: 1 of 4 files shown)


xarray/tests/test_concat.py
```diff
@@ -41,9 +41,6 @@ def test_concat_compat():
     for var in ["has_x", "no_x_y"]:
         assert "y" not in result[var]
 
-    with raises_regex(ValueError, "'q' is not present in all datasets"):
```
dcherian: This no longer raises a nice error message. I'm not sure what to do in this case: one dataset has a labeled coordinate "q" and the other doesn't have "q" at all. Currently this will error with "ValueError: One or more of the specified variables cannot be found in this dataset" which is fine ( I think)

jbusecke:
Thanks for this quick implementation @dcherian. I will work on implementing testing with the xarray master downstream so we can catch these earlier.

xarray/tests/test_concat.py
max-sixty: Agree—naming it is ideal but that's fine if we can fix it quickly

max-sixty: Can we keep the test which asserts it raises an error though?

dcherian: Done.

max-sixty:
🚀

jhamman:
Thanks @dcherian!

PR #3367, opened 2019-10-02
Remove setting of universal wheels

dopplershift:
Universal wheels indicate that one wheel supports Python 2 and 3. This
is no longer the case for xarray. This causes builds to generate files
with names like xarray-0.13.0-py2.py3-none-any.whl, which can cause
pip to incorrectly install the wheel on Python 2 when installing from a list of
wheel files.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[wheel]
-universal = 1
-
 [tool:pytest]
 python_files=test_*.py
 testpaths=xarray/tests properties


max-sixty:
Thanks @dopplershift , good spot

Do you know whether there any problems with having published these? Or it's just the name that was incorrect?

dopplershift:
You don’t run into a problem when doing `pip install xarray` on Python 2. We have run into problems on our CI builds where we install from our own cache on S3, which has a list of wheels in index.html, which we install from using `pip install -f <index_url> xarray`. When we have an xarray wheel in there with the incorrect name, python 2.7 tries to install from it.

max-sixty:
Ah OK, thanks. Do you need anything else from our end to fix that?

dopplershift:
All good, thanks.

Issue #2799, opened 2019-03-04
Performance: numpy indexes small amounts of data 1000 faster than xarray


nbren12:
Machine learning applications often require iterating over every index along some of the dimensions of a dataset. For instance, iterating over all the `(lat, lon)` pairs in a 4D dataset with dimensions `(time, level, lat, lon)`. Unfortunately, this is very slow with xarray objects compared to numpy (or h5py) arrays. When the Pangeo machine learning working group met [today](https://github.com/pangeo-data/ml-workflow-examples/issues/1), we found that several of us have struggled with this.

I made some simplified [benchmarks](https://gist.github.com/nbren12/e781c5a8fe03ee170628194c4b3c3160), which show that xarray is about 1000 times slower than numpy when repeatedly grabbing a small amount of data from an array. This is a problem with both `isel` or `[]` indexing. After doing some profiling, the main culprits seem to be xarray routines like `_validate_indexers` and `_broadcast_indexes`.

While python will always be slower than C when iterating over an array in this fashion, I would hope that xarray could be nearly as fast as numpy. I am not sure what the best way to improve this is though.

nbren12:
cc @rabernat

shoyer:
> While python will always be slower than C when iterating over an array in this fashion, I would hope that xarray could be nearly as fast as numpy. I am not sure what the best way to improve this is though.

I'm sure it's possible to optimize this significantly, but short of rewriting this logic in a lower level language it's pretty much impossible to match the speed of NumPy.

This benchmark might give some useful context:
```
def dummy_isel(*args, **kwargs):
  pass

def index_dummy(named_indices, arr):
    for named_index in named_indices:
        dummy_isel(arr, **named_index)
```
```
%%timeit -n 10
index_dummy(named_indices, arr)
```

On my machine, this is already twice as slow as your NumPy benchmark (497 µs vs 251 µs) , and all it's doing is parsing `*args` and **kwargs`! Every Python function/method call involving keyword arguments adds about 0.5 ns of overhead, because the highly optimized `dict` is (relatively) slow compared to positional arguments. In my experience it is almost impossible to get the overhead of a Python function call below a few microseconds.

Right now we're at about 130 µs per indexing operation. In the best case, we might make this 10x fatser but even that would be quite challenging, e.g., consider that even creating a DataArray takes about 20 µs.

nbren12:
[...]

PR #3375, opened 2019-10-06
Speed up isel and __getitem__

crusaderky:
First iterative improvement for #2799.

Speed up Dataset.isel up to 33% and DataArray.isel up to 25% (when there are no indices and the numpy array is small). 15% speedup when there are indices.

Benchmarks can be found in #2799.


jhamman:
Thanks @crusaderky. Do you think the indexing benchmarks we have in https://github.com/pydata/xarray/blob/master/asv_bench/benchmarks/indexing.py are sufficient? Anything you think would be worth adding to cover performance regressions here?

doc/whats-new.rst
```diff
@@ -31,6 +31,9 @@ Enhancements
       DataArrayGroupBy, grouped over 'season'
       4 groups with labels 'DJF', 'JJA', 'MAM', 'SON'
 
+- Speed up :meth:`Dataset.isel` up to 33% and :meth:`DataArray.isel` up to 25%
```
max-sixty:
```suggestion
- Speed up :meth:`Dataset.isel` up to 33% and :meth:`DataArray.isel` up to 25% for small arrays
```
What do you think @crusaderky ?

xarray/core/variable.py
```diff
@@ -957,7 +969,11 @@ def chunk(self, chunks=None, name=None, lock=False):
 
         return type(self)(self.dims, data, self._attrs, self._encoding, fastpath=True)
 
-    def isel(self, indexers=None, drop=False, **indexers_kwargs):
+    def isel(
+        self: VariableType,
+        indexers: Mapping[Hashable, Any] = None,
+        **indexers_kwargs: Any
+    ) -> VariableType:
```
max-sixty: Did we lose the `drop` keyword? Is that intentional?

jhamman:
> On first sight, it's a good start, but it's missing some important use cases:
> 
> * DataArray slicing
> * slicing when there are no IndexVariables
> * no benchmarks for `.sel()`

Would you be willing to add a few of these cases to the benchmarks?

xarray/core/variable.py
crusaderky: Yes, it was completely useless and untested. I'll add it to the What's New as a breaking change unless somebody can come up with reasoning as to why it was there...

crusaderky:
> Would you be willing to add a few of these cases to the benchmarks?

Yes, in due time. It's out of scope for this PR though.

xarray/core/variable.py
shoyer: I think this was a bad copy/paste. I agree, this argument only makes sense on Dataset/DataArray.

crusaderky: Added to What's New

xarray/core/dataset.py
```diff
@@ -1765,50 +1765,61 @@ def _validate_indexers(
...
-                    v = Variable((), v)
-                elif v.ndim == 1:
-                    v = Variable((k,), v)
-                else:
+                if v.ndim > 1:
                     raise IndexError(
                         "Unlabeled multi-dimensional array cannot be "
                         "used for indexing: {}".format(k)
                     )
+                yield k, v
```
shoyer: Does it actually meaningfully speed things up to not create a Variable here? I'm OK making things a little more complicated if need be, but it's nice to have slightly simpler logic and Variable objects can be pretty lightweight. Maybe we could use `Variable(..., fastpath=True)`, which skips most of the validation logic?

xarray/core/variable.py
```diff
@@ -41,6 +41,18 @@
 # https://github.com/python/mypy/issues/224
 BASIC_INDEXING_TYPES = integer_types + (slice,)  # type: ignore
 
+VariableType = TypeVar("VariableType", bound="Variable")
```
shoyer: This looks in fine to me, though I think mypy would also be equally happy with the `bound="Variable"` part as long as you're doing it on a method.

xarray/core/dataset.py
crusaderky: Yes, it's definitely worth it. Check out https://github.com/pydata/xarray/issues/2799#issuecomment-538570946 vs. https://github.com/pydata/xarray/issues/2799#issuecomment-538791352.

This PR reduces the runtime from 1.88s / 10k loops to 1.37s (-27%). Of the original time, 0.25s (13%) is spent on ``Variable.__init__``.

Didn't check fastpath=True. However, consider that Variable.isel *needs to accept any of these inputs anyway*.

xarray/core/variable.py
crusaderky: @shoyer I don't think I understand - could you elaborate?

shoyer: From mypy's perspective, it would be equivalent to write:
```
T = TypeVar('T')

class Variable:
  ...
  def some_method(self: T) -> T:
    ...
```

xarray/core/dataset.py
```diff
@@ -1904,19 +1915,21 @@ def isel(
                 )
                 if new_index is not None:
                     indexes[name] = new_index
-            else:
+            elif var_indexers:
                 new_var = var.isel(indexers=var_indexers)
+            else:
+                new_var = var
```
shoyer: Can we make this:
```suggestion
                new_var = var.copy(deep=False)
```
Otherwise, this would mean that you could modify attributes on variables on the resulting dataset and change the input dataset.

xarray/core/variable.py
crusaderky: Nope, wish it would.
```python
from typing import TypeVar

T = TypeVar("T")


class C:
    x: int

    def __init__(self, x: int):
        self.x = x

    def copy(self: T) -> T:
        return type(self)(self.x)
```
mypy 0.730 output:
```
13: error: Too many arguments for "object"
13: error: "T" has no attribute "x"
```

xarray/core/dataset.py
crusaderky: Fixed. Good catch!

xarray/core/variable.py
shoyer: Interesting. I've mostly been using Google's pytype for type checking, which handles this case.

Issue #3380, opened 2019-10-07
[Release] 0.14


jhamman:
#3358 is going to make some fairly major changes to the minimum supported versions of required and optional dependencies. We also have a few bug fixes that have landed since releasing 0.13 that would be good to get out.

From what I can tell, the following pending PRs are close enough to get into this release. 
- [ ] ~tests for arrays with units #3238~
- [x] map_blocks #3276
- [x] Rolling minimum dependency versions policy #3358
- [x] Remove all OrderedDict's (#3389)
- [x] Speed up isel and \_\_getitem\_\_ #3375
- [x] Fix concat bug when concatenating unlabeled dimensions. #3362
- [ ] ~Add hypothesis test for netCDF4 roundtrip #3283~
- [x] Fix groupby reduce for dataarray #3338 
- [x] Need a fix for https://github.com/pydata/xarray/issues/3377

Am I missing anything else that needs to get in?

I think we should aim to wrap this release up soon (this week). I can volunteer to go through the release steps once we're ready.

dcherian:
Added #3338

crusaderky:
Nuke away all OrderedDicts! (blocked by #3358)

dcherian:
This looks like it may have fallen out of the indexing refactor; definitely should be fixed: https://github.com/pydata/xarray/issues/3377

keewis:
I don't know if I will have the time to finish #3238 this week, even if I limit it to only testing `DataArray`.

jhamman:
> Nuke away all OrderedDicts! (blocked by #3358)

@crusaderky - have you started on this yet?

crusaderky:
@jhamman no - I should have it ready by the end of the weekend.

jhamman:
> @jhamman no - I should have it ready by the end of the weekend.

I plan to make the release on Friday. I can do this task before then unless you really want to do it :wink:.

shoyer:
Should we also make `assign` and `assign_coords` no longer sort their arguments? We can do that now that dictionary are always ordered

crusaderky:
@jhamman never release to production on a Friday! 🏖 
Jokes aside, I'm happy to leave the burden to you

jhamman:
See #3389 for the PR removing OrderedDicts.

jhamman:
>  Need a fix for #3377

See #3393

jhamman:
[...]

PR #3389, opened 2019-10-10
OrderedDict --> dict, some python3.5 cleanup too

jhamman:
- [x] Toward https://github.com/pydata/xarray/issues/3380#issuecomment-539224341

See below for inline comments where I could use some input from @shoyer and @crusaderky


xarray/backends/netCDF4_.py
```diff
@@ -495,6 +494,7 @@ def prepare_variable(
             )
 
         for k, v in attrs.items():
+            # TODO: update now that we dont support this old version of netCDF4
             # set attributes one-by-one since netCDF4<1.0.10 can't handle
             # OrderedDict as the input to setncatts
             _set_nc_attribute(nc4_var, k, v)
```
jhamman: I think we can update this now. @shoyer - do you recall what the setter syntax is for a dict of attributes?

xarray/core/combine.py
```diff
@@ -207,14 +207,14 @@ def _combine_all_along_first_dim(
 
     # Group into lines of datasets which must be combined along dim
     # need to sort by _new_tile_id first for groupby to work
-    # TODO remove all these sorted OrderedDicts once python >= 3.6 only
-    combined_ids = OrderedDict(sorted(combined_ids.items(), key=_new_tile_id))
+    # TODO: ask for help here
+    combined_ids = dict(sorted(combined_ids.items(), key=_new_tile_id))
```
jhamman: Do we need this `sorted` call here still? Its possible this line can be removed.

xarray/core/combine.py
```diff
@@ -207,14 +207,14 @@ def _combine_all_along_first_dim(
...
     grouped = itertools.groupby(combined_ids.items(), key=_new_tile_id)
 
     # Combine all of these datasets along dim
     new_combined_ids = {}
     for new_id, group in grouped:
-        combined_ids = OrderedDict(sorted(group))
+        combined_ids = dict(sorted(group))
```
jhamman: This may not be necessary anymore.

jhamman:
I'm also not sure where/how to do @shoyer's [comment](https://github.com/pydata/xarray/issues/3380#issuecomment-540069770):

> Should we also make `assign` and `assign_coords` no longer sort their arguments? We can do that now that dictionary are always ordered

xarray/core/combine.py
shoyer: I think we still need to canonicalize the order of variables between different datasets. @TomNicholas does that sound right to you?

shoyer:
> I'm also not sure where/how to do @shoyer's [comment](https://github.com/pydata/xarray/issues/3380#issuecomment-540069770):
> 
> > Should we also make `assign` and `assign_coords` no longer sort their arguments? We can do that now that dictionary are always ordered

I think this usage of `SortedKeysDict` could just become `dict`:
https://github.com/pydata/xarray/blob/a4c0210d6cbc6d28859ac21a3af8808e2660b0c0/xarray/core/common.py#L384

jhamman:
All green here. @crusaderky - I'd appreciate it if you could give this a quick look over.

xarray/core/combine.py
jhamman: I think we can leave this for now.

xarray/core/dataset.py
```diff
@@ -291,7 +289,7 @@ def split_indexes(
             dims.append(k)
 
     vars_to_replace = {}
-    vars_to_create = OrderedDict()  # type: OrderedDict[Any, Variable]
+    vars_to_create = {}  # type: Dict[Any, Variable]
```
dcherian:
```suggestion
    vars_to_create : Dict[Any, Variable] = {}
```

xarray/core/alignment.py
```diff
@@ -502,19 +502,19 @@ def reindex_variables(
...
         Dict of reindexed variables.
-    new_indexes : OrderedDict
+    new_indexes : dict
         Dict of indexes associated with the reindexed variables.
     """
     from .dataarray import DataArray
 
     # create variables for the new dataset
-    reindexed = OrderedDict()  # type: OrderedDict[Any, Variable]
+    reindexed = {}  # type: Dict[Any, Variable]
```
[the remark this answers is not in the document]
jhamman: This does work. But we don't seem to use this syntax elsewhere. For example: 

https://github.com/pydata/xarray/blob/4c05d38459dcf20ac4dc7fb32bb8cef4d7627c8b/xarray/core/dataset.py#L502

I feel like @crusaderky will have opinions here but for now, I'd like to leave the annotations with in their current format.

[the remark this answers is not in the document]
shoyer: This syntax is only Python 3.6+. So at this point might as well switch over type comments to variable annotations. But I agree, it isn't urgent.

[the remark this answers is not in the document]
dcherian: Yes it was just a suggestion. Merge away!

crusaderky:
I'll try to go through it later today

xarray/core/alignment.py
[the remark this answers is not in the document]
crusaderky: In my PR that migrated from 3.5 to 3.6 I got rid of almost all the ``# type: ...`` annotations, but I couldn't remove the ``# type: OrderedDict[...]`` because typing.OrderedDict was added in 3.7.2

crusaderky:
@jhamman I did a polish pass. Please review (particularly the substantial changes to ``DataArray.__init__`` and ``Dataset.__init__``)

Once you're happy, you can merge

jhamman:
Thanks @crusaderky - fixes to the `__init__` methods look good!

xarray/core/combine.py
TomNicholas: Yes you need the `sorted` call. 

Each object to be combined is assigned a "tile id" which specifies its position relative to the other objects. Sorted lines them all up in the correct order before the concatenation happens, which is [required](https://docs.python.org/3/library/itertools.html#itertools.groupby) by `itertools.groupby`. The `OrderedDict` was originally needed otherwise in python 3.5 this order was not guaranteed to be preserved before concatenation.

There isn't anything left to do here now, the Todo can be removed.

ttung:
FYI, python 3.6 does *not* guarantee that dicts are ordered.  CPython 3.6 happens to have that implementation but pypy, for instance, is not required to follow that.  Python 3.7, however, mandates that dicts are ordered.

shoyer:
Pypy already has ordered dicts on all versions of Python it supports. So technically you're right, but in practice this holds for every existing implementation of Python that could run xarray. Unless there's another one that I'm missing?

ttung:
No idea if there are any other ones; you're probably right.

PR #3352, opened 2019-09-28
Add glossary to documentation

gwgundersen:
This is my attempt at closing https://github.com/pydata/xarray/issues/2410. I've borrowed a bit of text from that PR (thanks @horta), from https://github.com/pydata/xarray/issues/1295, and from the xarray docs. I avoided verbose examples—it makes quickly searching and reading the glossary difficult—and put most of the effort into explaining dimension vs non-dimension coordinates, which in my experience opens the door to true xarray enlightenment.

Corrections, comments, or suggestions for additional terminology welcome.

PR #3410, opened 2019-10-17
Update Terminology page to account for multidimensional coordinates

jthielen:
As discussed in https://github.com/pydata/xarray/pull/3352, this PR modifies the Terminology page in the docs to briefly address multidimensional coordinates. Sorry for the delay in getting this in!

Also, when attempting to test the doc build, I found that the `doc/environment.yml` file was no longer present, so I updated it to `ci/requirements/doc.yml`.

### DIFF
--- a/doc/terminology.rst
+++ b/doc/terminology.rst
@@ -27,7 +27,7 @@ Terminology
 
 ----
 
-**Coordinate:** An array that labels a dimension or set of dimensions of another ``DataArray``. In the one-dimensional case, the coordinate array's values can loosely be thought of as tick labels along a dimension, whereas :doc:`multidimensional coordinates are often used when the data's physical coordinates differ from their logical coordinates <examples/multidimensional-coords>`. There are two types of coordinate arrays: *dimension coordinates* and *non-dimension coordinates* (see below). A coordinate named ``x`` can be retrieved from ``arr.coords[x]``. A ``DataArray`` can have more coordinates than dimensions because a single dimension can be assigned multiple coordinate arrays. However, only one coordinate array can be a assigned as a particular dimension's dimension coordinate array. As a consequence, ``len(arr.dims) <= len(arr.coords)`` in general.
+**Coordinate:** An array that labels a dimension or set of dimensions of another ``DataArray``. In the usual one-dimensional case, the coordinate array's values can loosely be thought of as tick labels along a dimension. There are two types of coordinate arrays: *dimension coordinates* and *non-dimension coordinates* (see below). A coordinate named ``x`` can be retrieved from ``arr.coords[x]``. A ``DataArray`` can have more coordinates than dimensions because a single dimension can be labeled by multiple coordinate arrays. However, only one coordinate array can be a assigned as a particular dimension's dimension coordinate array. As a consequence, ``len(arr.dims) <= len(arr.coords)`` in general.
 
 ----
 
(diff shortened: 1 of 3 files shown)


doc/terminology.rst
```diff
@@ -27,15 +27,15 @@ Terminology
...
+**Coordinate:** An array that labels a dimension or set of dimensions of another ``DataArray``. In the one-dimensional case, the coordinate array's values can loosely be thought of as tick labels along a dimension, whereas :doc:`multidimensional coordinates are often used when the data's physical coordinates differ from their logical coordinates <examples/multidimensional-coords>`. There are two types of coordinate arrays: *dimension coordinates* and *non-dimension coordinates* (see below). A coordinate named ``x`` can be retrieved from ``arr.coords[x]``. A ``DataArray`` can have more coordinates than dimensions because a single dimension can be assigned multiple coordinate arrays. However, only one coordinate array can be a assigned as a particular dimension's dimension coordinate array. As a consequence, ``len(arr.dims) <= len(arr.coords)`` in general.
```
max-sixty: > are often used when the data's physical coordinates differ from their logical coordinates

What does this mean? Is it climate-focused? (I'm not sure I have something better in mind yet, though)

doc/terminology.rst
```diff
@@ -27,15 +27,15 @@ Terminology
...
 
 ----
 
-**Dimension coordinate:** A coordinate array assigned to ``arr`` with both a name and dimension name in ``arr.dims``. Dimension coordinates are used for label-based indexing and alignment, like the index found on a :py:class:`pandas.DataFrame` or :py:class:`pandas.Series`. In fact, dimension coordinates use :py:class:`pandas.Index` objects under the hood for efficient computation. Dimension coordinates are marked by ``*`` when printing a ``DataArray`` or ``Dataset``.
+**Dimension coordinate:** A one-dimensional coordinate array assigned to ``arr`` with both a name and dimension name in ``arr.dims``. Dimension coordinates are used for label-based indexing and alignment, like the index found on a :py:class:`pandas.DataFrame` or :py:class:`pandas.Series`. In fact, dimension coordinates use :py:class:`pandas.Index` objects under the hood for efficient computation. Dimension coordinates are marked by ``*`` when printing a ``DataArray`` or ``Dataset``.
```
max-sixty: 👍

doc/terminology.rst
jthielen: This was taken from https://xarray.pydata.org/en/latest/examples/multidimensional-coords.html, and it was the best I could think of for a concise explanation. I interpreted it to refer to physical coordinates (like latitude and longitude) that do not always line up with the axes ("logical coordinates") of one's data (which is often the case when working with Earth-based data that are on some grid other than latitude/longitude). I'd be glad to change it to something else though if there are any suggestions!

max-sixty: That makes sense. I think it's probably too climate focused atm. I think we could either add `for example, in climate datasets...`, or remove that clause, or if anyone has suggestions?

jthielen: I think `in geoscience datasets` may be more appropriate/general than `in climate datasets` (speaking as a meteorologist :grin:), but I can easily make that change if no other suggestions come up.

jthielen: Since no other suggestions came in, I made this update. It also seemed to work better moving this to the **Non-dimension coordinate** definition (if I should move it back, just let me know).

max-sixty:
Looks great! Any other suggestions before we merge?

PR #3424, opened 2019-10-21
enable xr.ALL_DIMS in xr.dot

### LINKED ISSUE #3423: add `ALL_DIMS` in `xr.dot`
#### MCVE Code Sample

```python
import numpy as np
import xarray as xr
da_a = xr.DataArray(np.arange(3 * 2).reshape(3, 2), dims=['a', 'b'])
da_b = xr.DataArray(np.arange(3 * 2 * 2).reshape(3, 2, 2), dims=['a', 'b', 'c'])

xr.dot(da_a, da_b, dims=None)

```
This sums over the common dimensions:
```python
<xarray.DataArray (c: 2)>
array([110, 125])
Dimensions without coordinates: c
```

To sum over all dimensions:
```python
xr.dot(da_a, da_b, dims=['a', 'b', 'c'])

```
```python
<xarray.DataArray (c: 2)>
array([110, 125])
Dimensions without coordinates: c
```
#### Problem Description
`xr.dot` with `dims=None` currently sums over `all the common dimensions`. However, there are cases when a sum over all dimensions is desired. E.g. `xr.dot(da_a, da_b, dims=['a', 'b', 'c'])
` is a a memory efficient way to compute `(da_a * da_b).sum()` (if  `a` and `b` don't share the same dimensions). This is currently used in #2922 ([example](https://github.com/pydata/xarray/blob/dc7f6057c1a2533569ff6b5995d9a3c3e4c1fd85/xarray/core/weighted.py#L159)).

Therefore I suggest to allow `xr.dot(da_a, da_b, dims=xr.ALL_DIMS)` as shortcut to sum over all dimensions.

I assume there is no intent to change the behavior of `xr.dot(..., dims=None)`? (As it is a bit in contrast to other functions that are mostly applied over all dimensions.)

@fujiisoup @max-sixty 

<details>

</details>

mathause:
- [x] Closes #3423 
 - [x] Tests added


mathause:
Would you like me to replace `xr.ALL_DIMS` with `...`? (Although here I kind of like the explicit `xr.ALL_DIMS`).

shoyer:
It's fine to stick with `ALL_DIMS` for now. We'll remap `ALL_DIMS = ...` later, but every existing usage of ALL_DIMS will still work.

xarray/core/computation.py
```diff
@@ -1055,7 +1056,7 @@ def dot(*arrays, dims=None, **kwargs):
     ----------
     arrays: DataArray (or Variable) objects
         Arrays to compute.
-    dims: str or tuple of strings, optional
+    dims: xarray.ALL_DIMS, str or tuple of strings, optional
```
max-sixty:
```suggestion
    dims: '...', str or tuple of strings, optional
```

xarray/core/dataarray.py
```diff
@@ -2747,7 +2747,7 @@ def dot(
         ----------
         other : DataArray
             The other array with which the dot product is performed.
-        dims: hashable or sequence of hashables, optional
+        dims: xarray.ALL_DIMS, hashable or sequence of hashables, optional
```
```suggestion
        dims: '...', hashable or sequence of hashables, optional
```

max-sixty:
@mathause I did change the docstrings to `...`. I tried to commit but don't think I have permissions to your branch

Could we resolve the whatsnew conflicts and then we can merge?

Thank you!

max-sixty:
Super, thanks @mathause !

PR #3425, opened 2019-10-21
Html repr

### LINKED ISSUE #1627: html repr of xarray object (for the notebook)

jsignell:
This PR supersedes #1820 - see that PR for original discussion. See [this gist](https://nbviewer.jupyter.org/gist/jsignell/2b7843a6e4852198bf963fbd299e8d46) to try out the new MultiIndex and options functionality. 
 
 - [x] Closes #1627, closes #1820 
 - [x] Tests added

TODO:

 - [x] Add support for Multi-indexes
 - [x] Probably good to have some opt-in or fail back system in case where we (or users) know that the rendering will not work
 - [x] Add some tests


jsignell:
I don't really like how the ``attrs`` look without the alignment of keys and values. I will think about how to make that more legible

dcherian:
Honestly this is looking super good and ready to merge already!

Thanks @jsignell 

The only thing that looks funny to me is the alternating black-gray text coloring below:
image

jsignell:
Yeah I agree that is a strange part of the original PR. The version in the fiddle doesn't do that, so I'd really like to hear back from @benbovy about whether the fiddle css is more uptodate.

rabernat:
Julia, it's fantastic to see you working on this! Thanks so much!

shoyer:
This looks fantastic! Really great to have you working on this!

Some specific feedback:

1. Yes, I agree it would be nice to align the Attributes text, maybe this could make use of an HTML table?
2. Maybe bold would look better than underline for coordinates that are dimensions? I find the underline makes text a little hard to read. Happy to leave this up to you, though.
3. On my Ubuntu desktop, the "+" icon is bright yellow, which I'm guessing is not what was intended. Maybe it would better to stick with the three circle "data" icon?
image
4. What happens if there is a really long variable name? I had a version of the jsfiddle that expanded names when you hovered over them.

xarray/core/formatting_html.py
```diff
@@ -0,0 +1,313 @@
...
+
+    if len(var.attrs):
+        d['disabled'] = ''
+        d['attrs'] = summarize_attrs(var.attrs)
+    else:
+        d['disabled'] = 'disabled'
+        d['attrs'] = ''
+
+    # TODO: no value preview if not in memory
+    d['preview'] = d.get('preview', format_values_preview(var))
```
shoyer: We should probably implement this TODO before merging. Otherwise printing the repr() will probably cause computers to crash when using large datasets with dask-distributed.

shoyer: Also: this could be a nice place to recursively check for an `_html_repr_` attribute, e.g., on dask arrays!

xarray/core/formatting_html.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+def array_section(obj):
+    d = {}
+
+    # "unique" id to expand/collapse the section
+    d['id'] = 'section-' + str(uuid.uuid4())
+
+    # TODO: no value preview if not in memory
+    d['preview'] = format_values_preview(obj.values, max_char=70)
```
shoyer: same concern as above

xarray/core/formatting_html.py
jsignell: oh I didn't look closely and read that TODO as a DONE. Yes I agree that we should make sure the data are loaded.

rabernat: Yes, would be amazing to see dask's new fancy array html repr here!

shoyer: I would suggest using `inline_variable_array_repr` from `formatting.py`.

xarray/core/formatting_html.py
```diff
@@ -0,0 +1,313 @@
...
+        d['disabled'] = ''
+        d['attrs'] = summarize_attrs(var.attrs)
+    else:
+        d['disabled'] = 'disabled'
+        d['attrs'] = ''
+
+    # TODO: no value preview if not in memory
+    d['preview'] = d.get('preview', format_values_preview(var))
+    d['attrs_ul'] = summarize_attrs(var.attrs)
+    d['data_repr'] = repr(var.data)
```
shoyer: You could use  `short_data_repr` from `formatting.py` here.

benbovy:
This is really nice @jsignell !

> I'd really like to hear back from @benbovy about whether the fiddle css is more uptodate.

We did a lot of experiments, even some after starting the implementation in #1820, so I admit this is quite messy. I also did some clean-up and tweaks directly in that PR, so unfortunately there is no fiddle of reference that exactly corresponds to what's in here. Sorry for that!

The html/css in this PR is the cleanest one, I think. 

The latest fiddles (for `Dataset`) are these two ones:

- https://jsfiddle.net/Lmqq7yzz/9/
- https://jsfiddle.net/1g04ykum/  (@shoyer's  try for handling long variable names)

> the only thing that looks funny to me is the alternating black-gray text coloring

In the HTML repr, the width of the space character is less large that for the monospace font used in the text repr, so I originally used this trick to better distinguish between the values in the inline data repr. Now that each value is encapsulated in it's own `span`, we could tweak the padding instead (or both... I don't have strong opinion on this).

benbovy:
Just one small issue in the gist with the vertical alignment within a variable row, and the height of the icon container (I using Firefox):

Screenshot 2019-10-22 at 09 33 04

xarray/core/formatting_html.py
benbovy: > I would suggest using inline_variable_array_repr from formatting.py

The issue with just reusing `inline_variable_array_repr` here is that it would be then more tricky to use html templates to tweak the rendering of the values (e.g., encapsulate each value in its own `span` with custom padding). See [my comment below](https://github.com/pydata/xarray/pull/3425#issuecomment-544839504).

benbovy: Mmm there is apparently no custom padding in the CSS, so maybe using `span` for each value is not really needed?

Here is how it looks without alternating grey coloring:

Screenshot 2019-10-22 at 09 52 26

What do you think about the spacing between the values? Is it enough?

jsignell: I think it's enough.

jsignell:
Yeah the data preview is also shifted vertically.

jsignell:
I just updated the gist with these changes:
 - got rid of alternating data colors in preview
 - fixed icon alignment in row, and rest of data in row
 - got rid of some not-useful spans
 - changed from underline to bold for index coords

xarray/core/formatting_html.py
```diff
@@ -0,0 +1,312 @@
...
+
+import uuid
+import pkg_resources
+from functools import partial
+from collections import OrderedDict
+
+from .formatting import format_array_flat
+
+
+CSS_FILE_PATH = '/'.join(('static', 'css', 'style-jupyterlab.css'))
```
shoyer: We could probably just call this `style.css`, since we're going to use it everywhere, even outside of jupyterlab.

jsignell:
Is there ever a situation in which dimensions would be expandable? It seems like it is hard-coded to collapsed. 

I am trying to generalize the css to reduce the length.

shoyer:
> Is there ever a situation in which dimensions would be expandable? It seems like it is hard-coded to collapsed.

I don't think it would ever be expanded. The collapsed icon is just there for visual consistency.

jsignell:
Ok I think I've addressed all the comments. I just updated the gist with the latest output. Tomorrow I'll start adding tests. The css is now ~300 lines and it gets injected everytime the dataset is rendered. I'm not sure if that is considered too long and if it is I don't quite know what to do about it.

shoyer:
> The css is now ~300 lines and it gets injected everytime the dataset is rendered. I'm not sure if that is considered too long and if it is I don't quite know what to do about it.

You could consider running the CSS through some sort of preprocessing tool to minify it, e.g., to remove all the spaces. But my guess is that it's *probably* fine.

It would be informative to measure the size of the output HTML encoded as UTF-8 bytes for some example datasets. For context, a simple line chart from matplotlib is ~20 KB when saved as a PNG image. Anything in that ball-park would be OK, up to perhaps a few hundred KB for really big datasets with lots of variables.

xarray/core/common.py
```diff
@@ -134,6 +134,12 @@ def __array__(self: Any, dtype: DTypeLike = None) -> np.ndarray:
     def __repr__(self) -> str:
         return formatting.array_repr(self)
 
+    def _repr_html_(self):
+        if OPTIONS["display_style"] == "classic":
+            classic = repr(self).replace("<", "&lt;").replace(">", "&gt;")
```
shoyer: Use `html.escape` here: https://docs.python.org/3/library/html.html#html.escape

xarray/core/common.py
```diff
@@ -134,6 +134,12 @@ def __array__(self: Any, dtype: DTypeLike = None) -> np.ndarray:
...
+            return "<pre>{repr}</pre>".format(repr=classic)
```
shoyer: Could use an f-string here:
```suggestion
            return f"<pre>{classic}</pre>"
```

setup.py
```diff
@@ -104,5 +104,5 @@
     tests_require=TESTS_REQUIRE,
     url=URL,
     packages=find_packages(),
-    package_data={"xarray": ["py.typed", "tests/data/*"]},
+    package_data={"xarray": ["py.typed", "tests/data/*", "static/*"]},
```
shoyer: This might need `static/**` to pick up subdirectories. When I tried installing with pip it didn't work otherwise.

shoyer:
I just gave this a try in Chrome / Google Colab. It's looking great!

Here's the `rasm` dataset. It weighs in at exactly 20 KB!
image

The alignment of "Dimensions" line does looks a little off -- would that be easy to fix?

There's also a strange alignment thing with the three circle icon in array repr, which maybe could be moved to the right side?
image

shoyer:
We probably should be careful to use `html.escape()` pervasively to protect against malformed data. We can trust user code, but probably should not trust all data -- it should be safe to load and view arbitrary netCDF files in xarray.

xarray/core/common.py
jsignell: I didn't realize that xarray was already supporting >=3.6 :+1:

shoyer:
Here's what we're currently looking at!
image

jsignell:
I'm still worried about size, but minifying the css and even minifying the html did not make a big difference. It looks like the html_repr for rasm is 38kb .

shoyer:
> It looks like the html_repr for rasm is 38kb .

How are you measuring this? 

I see:
```
>>> len(rasm._repr_html_().encode('utf8'))
19624
```

jsignell:
Oh I was looking at bytes not len.

jsignell:
Ok I am going to write some tests now. I updated the gist. Do people like the names "classic" and "html" for the display_style options? Another option would be "classic" and "rich".

benbovy:
Or maybe "text" and "html"?

Hopefully "html" will eventually become the classic display :)

jsignell:
I added some tests. I can add more if anyone has a test case that they want to be sure is covered.

jsignell:
Otherwise I think all that is left is adding docs. Also I want to confirm that this should be an experimental feature (off by default).

jsignell:
Oh. The dask version in ci doesn't have `_repr_html_`. I'll add a version checker to the test.

jsignell:
I am 90% sure [this failure](https://github.com/pydata/xarray/pull/3425/checks?check_run_id=272331486) is unrelated. Maybe I should rebase now that #3436 has landed?

jsignell:
Well, I rebased. Hopefully it'll pass now. I consider this done unless someone wants more docs or more tests.

keewis:
nope, it still fails. However, that failure is definitely not related to this PR (it fails to install because the [referenced PR in cftime](https://github.com/pydata/xarray/blob/3fa3cae7fe48de84e676e91a9d328aa00457e5da/ci/azure/install.yml#L28-L29) was merged in the meantime).

jsignell:
So maybe one more rebase will do it?

jsignell:
@benbovy just to make double sure, you are happy with this new feature being off by default right?

xarray/core/options.py
```diff
@@ -98,6 +102,9 @@ class set_options:
       attrs, ``False`` to always discard them, or ``'default'`` to use original
       logic that attrs should only be kept in unambiguous circumstances.
       Default: ``'default'``.
+    - ``display_style``: display style to use in jupyter for xarray objects.
+      Default: ``classic``. Other options are ``html``.
```
benbovy: classic -> text

jhamman:
I would be fine having this be on by default now. Or, we could keep it off until the next major release (0.15).

benbovy:
I'd would rather wait before turning on html display by default. I'd be fine doing so on the next major release if until then we make sure that the rendering will look reasonably good in most front-end situations.

For example, it will really need better integration with jupyterlab dark themes (see below) and would potentially cause issues elsewhere too (nbsphinx, etc.). But we should probably figure that out after this PR is merged.

Screenshot 2019-10-24 at 17 12 21

I've noticed an issue with the alignment of attributes in jupyterlab (1.1.4) (Firefox 69.0.3):

Screenshot 2019-10-24 at 17 11 46

It looks good in the classic notebook.

xarray/core/options.py
dcherian:
```suggestion
      Default: ``text``. Other options are ``html``.
```

dcherian:
```suggestion
      Default: ``'text'``. Other options are ``'html'``.
```

jsignell:
> I've noticed an issue with the alignment of attributes in jupyterlab (1.1.4) (Firefox 69.0.3):

I'm taking a look at the jlab css.

shoyer:
+1 for merging this turned off by default, with a plan to switch it on once
we're tested it on more datasets and platforms

jsignell:
Fixed the jlab alignment issue. The issue was that jlab has different default css.

Screen Shot 2019-10-24 at 11 37 45 AM

jsignell:
Yeah to make it work with the dark theme we'll need to use the jlab css variables I imagine. I think it can be merged with dark mode not supported.

benbovy:
Mmm this still looks weird on my side:

Screenshot 2019-10-24 at 17 39 44

Which version of jupyterlab are you using?

jsignell:
Ok I think it's fixed now

benbovy:
Yes it looks nice!

jsignell:
It isn't too hard to get it working with dark theme, but requires slight tweaks to the colors to align with those provided as vars by jlab.

dcherian:
Awesome. Thanks @jsignell Let's leave dark theme compatibility for a future PR.

shoyer:
Should we expand the `.data` section of the DataArray HTML repr by default? That would look a little more consistent with the text repr.

benbovy:
Thanks a lot @jsignell ! It is great to see this merged finally!

mrocklin:
Thanks @jsignell (and all).  I'm really jazzed about this.

choldgraf:
this is awesome! I was so excited about it that I updated my little blog post with the latest xarray master https://predictablynoisy.com/xarray-explore-ieeg

I was pretty annoyed at the gigantic repr that came out of the DataArray, and now it is nice and tidy html!

benbovy:
Cool!

I guess [this post](https://predictablynoisy.com/jekyll-markdown-nbconvert) about how you use notebooks in your blog is still up to date?

I looks like we'll need to fix a couple of issues (the `channel` coordinate looks weird), but maybe it is not related to the front-end.

Besides variable names, we probably need to somehow handle overflow for variables that have a long list of dimension labels.

choldgraf:
Actually, no that's totally out of date now haha. Thanks for reminding me I should update that

jsignell:
> I looks like we'll need to fix a couple of issues (the channel coordinate looks weird), but maybe it is not related to the front-end. 

I am opening a PR. I didn't escape dtypes and this dtype is `<U5` which html does **terribly** with

shoyer:
To verify proper escaping, we could validate the generated HTML as XML:
https://stackoverflow.com/questions/13742538/how-to-validate-xml-using-python-without-third-party-libs

This could be a good use case for Hypothesis. Or you could manually build a dataset with variables of every dtype.

jsignell:
I opened [a PR](#3444) and included a small test. It can probably go in while we work on a more complete testing infrastructure.

jsignell:
> Besides variable names, we probably need to somehow handle overflow for variables that have a long list of dimension labels.

These are handled in the same way as long variable names, they truncate with an ellipsis and then expand on hover.

benbovy:
> These are handled in the same way as long variable names, they truncate with an ellipsis and then expand on hover.

Oh yes I missed it.

Issue #3427, opened 2019-10-22
another cftime failure


dcherian:
```
________ TestNetCDF3ViaNetCDF4Data.test_roundtrip_cftime_datetime_data _________

dates = array([cftime.DatetimeProlepticGregorian(0001-01-01 00:00:00),
       cftime.DatetimeProlepticGregorian(0001-01-02 00:00:00)],
      dtype=object)
units = 'days since 0001-01-01 00:00:00.000000'
calendar = 'proleptic_gregorian'

    def encode_cf_datetime(dates, units=None, calendar=None):
        """Given an array of datetime objects, returns the tuple `(num, units,
        calendar)` suitable for a CF compliant time variable.
    
        Unlike `date2num`, this function can handle datetime64 arrays.
    
        See also
        --------
        cftime.date2num
        """
        dates = np.asarray(dates)
    
        if units is None:
            units = infer_datetime_units(dates)
        else:
            units = _cleanup_netcdf_time_units(units)
    
        if calendar is None:
            calendar = infer_calendar_name(dates)
    
        delta, ref_date = _unpack_netcdf_time_units(units)
        try:
            if calendar not in _STANDARD_CALENDARS or dates.dtype.kind == "O":
                # parse with cftime instead
>               raise OutOfBoundsDatetime
E               pandas._libs.tslibs.np_datetime.OutOfBoundsDatetime

xarray/coding/times.py:400: OutOfBoundsDatetime

During handling of the above exception, another exception occurred:

self = <xarray.tests.test_backends.TestNetCDF3ViaNetCDF4Data object at 0x7fd54f4cb780>

    @requires_cftime
    def test_roundtrip_cftime_datetime_data(self):
        from .test_coding_times import _all_cftime_date_types
    
        date_types = _all_cftime_date_types()
        for date_type in date_types.values():
            times = [date_type(1, 1, 1), date_type(1, 1, 2)]
            expected = Dataset({"t": ("t", times), "t0": times[0]})
            kwargs = {"encoding": {"t0": {"units": "days since 0001-01-01"}}}
            expected_decoded_t = np.array(times)
            expected_decoded_t0 = np.array([date_type(1, 1, 1)])
            expected_calendar = times[0].calendar
    
            with warnings.catch_warnings():
                if expected_calendar in {"proleptic_gregorian", "gregorian"}:
                    warnings.filterwarnings("ignore", "Unable to decode time axis")
    
[...]

Issue #3426, opened 2019-10-21
`.sel()` failures when using latest cftime release (v1.0.4)


andersy005:
I just updated to the latest `cftime`[ release](https://github.com/Unidata/cftime/releases/tag/v1.0.4rel), and all of a sudden `sel()` appears to be broken:

```python
In [1]: import xarray as xr                                                                                                                        

In [2]: import cftime                                                                                                                              

In [3]: ds = xr.tutorial.load_dataset('rasm')                                                                                                      

In [4]: ds                                                                                                                                         
Out[4]: 
<xarray.Dataset>
Dimensions:  (time: 36, x: 275, y: 205)
Coordinates:
  * time     (time) object 1980-09-16 12:00:00 ... 1983-08-17 00:00:00
    xc       (y, x) float64 189.2 189.4 189.6 189.7 ... 17.65 17.4 17.15 16.91
    yc       (y, x) float64 16.53 16.78 17.02 17.27 ... 28.26 28.01 27.76 27.51
Dimensions without coordinates: x, y
Data variables:
    Tair     (time, y, x) float64 nan nan nan nan nan ... 29.8 28.66 28.19 28.21
Attributes:
    title:                     /workspace/jhamman/processed/R1002RBRxaaa01a/l...
    institution:               U.W.
    source:                    RACM R1002RBRxaaa01a
    output_frequency:          daily
    output_mode:               averaged
    convention:                CF-1.4
    references:                Based on the initial model of Liang et al., 19...
    comment:                   Output from the Variable Infiltration Capacity...
    nco_openmp_thread_number:  1
    NCO:                       "4.6.0"
    history:                   Tue Dec 27 14:15:22 2016: ncatted -a dimension...

In [5]: ds.sel(time=slice("1980", "1982"))                                                                                                         
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-2c26e36a673a> in <module>
----> 1 ds.sel(time=slice("1980", "1982"))

~/opt/miniconda3/envs/intake-esm-dev/lib/python3.7/site-packages/xarray/core/dataset.py in sel(self, indexers, method, tolerance, drop, **indexers_kwargs)
   1998         indexers = either_dict_or_kwargs(indexers, indexers_kwargs, "sel")
   1999         pos_indexers, new_indexes = remap_label_indexers(
-> 2000             self, indexers=indexers, method=method, tolerance=tolerance
   2001         )
[...]

PR #3429, opened 2019-10-22
minor lint tweaks

crusaderky:
- Ran pyflakes 2.1.1
- Some f-string tweaks
- Ran black -t py36
- Ran mypy 0.740. We'll need to skip it and jump directly to 0.750 once it's released because of https://github.com/python/mypy/issues/7735


crusaderky:
Blocked by #3426 and #3427

crusaderky:
@max-sixty the cftime issue is causing 10,440 test failures. Are you volunteering to go through them one by one to verify there aren't any new ones? :)

dcherian:
only upstream-dev and the minimum versions thing should fail if you merge master now

PR #3431, opened 2019-10-22
Add cftime git tip to upstream-dev + temporarily pin cftime

dcherian:
temporarily pin cftime so that we can keep merging.

### DIFF
--- a/ci/requirements/py36-min-all-deps.yml
+++ b/ci/requirements/py36-min-all-deps.yml
@@ -13,7 +13,7 @@ dependencies:
   - cartopy=0.17
   - cdms2=3.1
   - cfgrib=0.9
-  - cftime=1.0.0
+  - cftime=1.0.3
   - coveralls
   - dask=1.2
   - distributed=1.27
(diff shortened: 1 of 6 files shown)


ci/requirements/py36-min-all-deps.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - cartopy=0.17
   - cdms2=3.1
   - cfgrib=0.9
-  - cftime=1.0
+  - cftime=1.0.0
```
dcherian: @spencerkclark do we want to bump this up to 1.0.3?

spencerkclark: Thanks for doing this @dcherian.  Yes, I think that is appropriate, given the [new minimum dependency rules](http://xarray.pydata.org/en/stable/installing.html#minimum-dependency-versions); version 1.0.3 was released about 10 months ago.  

It would be nice at some point to remove some of the cruft that is leftover from supporting both `netcdftime` and `cftime` in places.  Maybe when I get a free moment I'll look into that.

dcherian: :+1: done.

crusaderky:
My MinimumVersionsPolicy script is not designed to cope with patch versions in the requirement file. I'll amend it tonight.

dcherian:
Thanks @crusaderky. I'll merge this now.

PR #3436, opened 2019-10-22
MAGA (Make Azure Green Again)

crusaderky:
Let all CI tests become green again to avoid hindering developers who are working on PRs unrelated to the present incompatibilities (numpy=1.18, cftime=1.0.4, pseudonetcdf=3.1.0).

PR #3439, opened 2019-10-24
Use cftime master for upstream-dev build

shoyer:
Follow-up on #3436, needed now that https://github.com/Unidata/cftime/pull/127 has been merged.


shoyer:
There's another failure with pandas-master (see https://github.com/pydata/xarray/issues/3440), but I'm going to merge this as a step in the right direction.

PR #3443, opened 2019-10-24
jupyterlab dark theme

jsignell:
- [ ] Closes #xxxx
 - [ ] Tests added

Follow on to #3425 to include support for jupyterlab dark theme. Note that this includes slight color changes. The most striking of which is that in jupyterlab light and regular notebook the even rows are white like the background. 

## Jlab dark
Screen Shot 2019-10-24 at 12 57 56 PM

## Jlab light
Screen Shot 2019-10-24 at 12 58 34 PM

## notebook
Screen Shot 2019-10-24 at 1 04 13 PM

### DIFF
--- a/xarray/static/css/style.css
+++ b/xarray/static/css/style.css
@@ -2,6 +2,17 @@
  *
  */
 
+:root {
+  --xr-font-color0: var(--jp-content-font-color0, rgba(0, 0, 0, 1));
+  --xr-font-color2: var(--jp-content-font-color2, rgba(0, 0, 0, 0.54));
+  --xr-font-color3: var(--jp-content-font-color3, rgba(0, 0, 0, 0.38));
+  --xr-border-color: var(--jp-border-color2, #e0e0e0);
+  --xr-disabled-color: var(--jp-layout-color3, #bdbdbd);
+  --xr-background-color: var(--jp-layout-color0, white);
+  --xr-background-color-row-even: var(--jp-layout-color1, white);
+  --xr-background-color-row-odd: var(--jp-layout-color2, #eeeeee);
+}
+
 .xr-wrap {
   min-width: 300px;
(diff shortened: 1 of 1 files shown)


jsignell:
Note that this won't work in IE https://caniuse.com/#feat=css-variables

jsignell:
@benbovy in case you have ideas about which colors to use.

shoyer:
IE is pretty niche these days -- not even Microsoft is maintaining it. I expect that this is fine for most of our users, as long as the degradation is not too terrible.

benbovy:
This looks great, I like it!

Slightly related: I've tried to add hover background color on variable items, using `--jp-rendermime-table-row-hover-background`, to be consistent with other rich outputs in jupyterlab, but unfortunately I couldn't make it work nicely.

If I remember well, this would need wrapping each variable item (name, dims, values, etc.) in a container, but then this wouldn't work well with `display: contents`. I might be wrong, though.

jsignell:
I don't really like the hover-background anyways if those rows aren't clickable and other things are.

benbovy:
For tables with many columns it might be useful. That’s not the case here, although I’ve found that it helps to track the list of variables when variable data and/or attributes reprs are unfolded. But yeah, it’s a detail.

dcherian:
Is this ready to merge?

Also if jupyter is using css variables, i assume that it too won't work well in IE?

jsignell:
Right I don't think jlab works in IE. But this won't work in regular notebook in IE either.

jsignell:
> Is this ready to merge?

Yeah I think so.

max-sixty:
Can we add a whatsnew with attribution?

shoyer:
> Can we add a whatsnew with attribution?

I think this is  covered by the previous PR adding the HTML repr?

PR #3450, opened 2019-10-26
Remove outdated code related to compatibility with netcdftime

spencerkclark:
Per https://github.com/pydata/xarray/pull/3431#discussion_r337620810, this removes outdated code leftover from the netcdftime -> cftime transition.  Currently the [minimum version of netCDF4 that xarray tests against](https://github.com/dcherian/xarray/blob/fa9b644dd3d41d5bedb4b040d71f101590e48d11/ci/requirements/py36-min-all-deps.yml#L30) is 1.4, which does not include netcdftime, and instead specifies cftime as a required dependency.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -21,7 +21,7 @@ v0.14.1 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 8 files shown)


dcherian:
Thanks @spencerkclark

Issue #3266, opened 2019-08-26
Warnings in the test suite


max-sixty:
If anyone is looking for any bite-size contributions, the test suite is throwing off many warnings. Most of these indicate that something will break in the future without code changes; thought mostly the code changes are small.

```

=============================== warnings summary ===============================
/usr/share/miniconda/envs/xarray-tests/lib/python3.7/site-packages/heapdict.py:11
  /usr/share/miniconda/envs/xarray-tests/lib/python3.7/site-packages/heapdict.py:11: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    class heapdict(collections.MutableMapping):

/usr/share/miniconda/envs/xarray-tests/lib/python3.7/site-packages/pydap/model.py:175
  /usr/share/miniconda/envs/xarray-tests/lib/python3.7/site-packages/pydap/model.py:175: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import OrderedDict, Mapping

/usr/share/miniconda/envs/xarray-tests/lib/python3.7/site-packages/pydap/responses/das.py:14
  /usr/share/miniconda/envs/xarray-tests/lib/python3.7/site-packages/pydap/responses/das.py:14: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Iterable

xarray/tests/test_accessor_dt.py::test_cftime_strftime_access[365_day]
  /home/vsts/work/1/s/xarray/tests/test_accessor_dt.py:226: RuntimeWarning: Converting a CFTimeIndex with dates from a non-standard calendar, 'noleap', to a pandas.DatetimeIndex, which uses dates from the standard calendar.  This may lead to subtle errors in operations that depend on the length of time between dates.
    xr.coding.cftimeindex.CFTimeIndex(data.time.values).to_datetimeindex(),

xarray/tests/test_accessor_dt.py::test_cftime_strftime_access[360_day]
  /home/vsts/work/1/s/xarray/tests/test_accessor_dt.py:226: RuntimeWarning: Converting a CFTimeIndex with dates from a non-standard calendar, '360_day', to a pandas.DatetimeIndex, which uses dates from the standard calendar.  This may lead to subtle errors in operations that depend on the length of time between dates.
[...]

PR #3451, opened 2019-10-26
Remove deprecated behavior from dataset.drop docstring

max-sixty:
I'm less up to speed on this behavior, but IIUC this part of the docstring refers to [deprecated behavior](https://github.com/pydata/xarray/compare/master...max-sixty:drop-docstring?expand=1#diff-921db548d18a549f6381818ed08298c9R3586-R3592)—is that correct or am I missing something?

xref https://github.com/pydata/xarray/issues/3266


dcherian:
Thanks @max-sixty

PR #3453, opened 2019-10-28
Optimize dask array equality checks.

### LINKED ISSUE #3068: Multidimensional dask coordinates unexpectedly computed
#### MCVE Code Sample

```python
from dask.diagnostics import ProgressBar
import xarray as xr
import numpy as np
import dask.array as da

a = xr.DataArray(da.zeros((10, 10), chunks=2), dims=('y', 'x'), coords={'y': np.arange(10), 'x': np.arange(10), 'lons': (('y', 'x'), da.zeros((10, 10), chunks=2))}) 
b = xr.DataArray(da.zeros((10, 10), chunks=2), dims=('y', 'x'), coords={'y': np.arange(10), 'x': np.arange(10), 'lons': (('y', 'x'), da.zeros((10, 10), chunks=2))}) 

with ProgressBar():
    c = a + b

```

Output:

```
[########################################] | 100% Completed |  0.1s
```

#### Problem Description

Using arrays with 2D dask array coordinates results in the coordinates being computed for any binary operations (anything combining two or more DataArrays). I use `ProgressBar` in the above example to show when coordinates are being computed.

In my own work, when I learned that 2D dask coordinates were possible, I started adding `longitude` and `latitude` coordinates. These are rather large and can take a while to load/compute so I was surprised that simple operations (ex. `a.fillna(b)`) were causing things to be computed and taking a long time.

Is this computation by design or a possible bug?

#### Expected Output

No output from the `ProgressBar`, hoping that no coordinates would be computed/loaded.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.7 | packaged by conda-forge | (default, Feb 28 2019, 02:16:08) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
python-bits: 64
OS: Darwin
OS-release: 18.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.2

xarray: 0.12.1
pandas: 0.24.2
numpy: 1.14.3
scipy: 1.3.0
netCDF4: 1.5.1.2
pydap: None
h5netcdf: 0.7.4
h5py: 2.9.0
Nio: None
zarr: 2.3.2
cftime: 1.0.3.4
nc_time_axis: None
PseudonetCDF: None
rasterio: 1.0.22
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 2.0.0
distributed: 2.0.0
matplotlib: 3.1.0
cartopy: 0.17.1.dev147+HEAD.detached.at.5e624fe
seaborn: None
setuptools: 41.0.1
pip: 19.1.1
conda: None
pytest: 4.6.3
IPython: 7.5.0
sphinx: 2.1.2

</details>

### LINKED ISSUE #3311: optimize compatibility checks in merge.unique_variable
Currently `merge.unique_variable` does this

```
    if equals is None:
        out = out.compute()
        for var in variables[1:]:
            equals = getattr(out, compat)(var)
            if not equals:
                break
```

`out` (`=variables[0]`) is always computed though it may not be necessary. 

One solution would be to loop through once checking attrs, shapes and `_data`. If these checks were satisfied, then we execute the above code

### LINKED ISSUE #3454: Large coordinate arrays trigger computation
I want to bring up an issue that has tripped up my workflow with large climate models many times. I am dealing with large data arrays of vertical cell thickness. These are 4d arrays (x, y, z, time) but I would define them as coordinates, not data_variables in the xarrays data model (e.g. they should not be multiplied by a value if a dataset is multiplied).
These sort of coordinates might become more prevalent with newer ocean models like [MOM6](http://www.cesm.ucar.edu/events/wg-meetings/2017/presentations/omwg/adcroft.pdf)

Whenever I assign these arrays as coordinates operations on the arrays seem to trigger computation, whereas they don't if I set them up as data_variables. The example below shows this behavior. 
Is this a bug or done on purpose? Is there a workaround to keep these vertical thicknesses as coordinates?

```
import xarray as xr
import numpy as np
import dask.array as dsa

# create dataset with with vertical thickness `dz` as data variable
data = xr.DataArray(dsa.random.random([30, 50, 200, 1000]), dims=['x','y', 'z', 't'])
dz = xr.DataArray(dsa.random.random([30, 50, 200, 1000]), dims=['x','y', 'z', 't'])
ds = xr.Dataset({'data':data, 'dz':dz})

#another dataset with `dz` as coordinate
ds_new = xr.Dataset({'data':data})
ds_new.coords['dz'] = dz
```

```
%%time
test = ds['data'] * ds['dz']
```
`CPU times: user 1.94 ms, sys: 19.1 ms, total: 21 ms
Wall time: 21.6 ms`

```
%%time
test = ds_new['data'] * ds_new['dz']
```
`CPU times: user 17.4 s, sys: 1.98 s, total: 19.4 s
Wall time: 12.5 s`

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.4 (default, Aug 13 2019, 15:17:50) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
python-bits: 64
OS: Darwin
OS-release: 18.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: None
libnetcdf: None

xarray: 0.13.0+24.g4254b4af
pandas: 0.25.1
numpy: 1.17.2
scipy: None
netCDF4: None
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: None
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.5.0
distributed: 2.5.1
matplotlib: None
cartopy: None
seaborn: None
numbagg: None
setuptools: 41.2.0
pip: 19.2.3
conda: None
pytest: 5.2.0
IPython: 7.8.0
sphinx: None

</details>

dcherian:
Dask arrays with the same graph have the same name. We can use this to quickly
compare dask-backed variables without computing. (see https://github.com/pydata/xarray/issues/3068#issuecomment-508853564)

I will work on adding extra tests but could use feedback on the approach.

 - [x] Closes #3068, closes #3311, closes #3454 
 - [x] Tests added

@djhoese, thanks for the great example code!


xarray/core/duck_array_ops.py
```diff
@@ -181,16 +181,49 @@ def allclose_or_equiv(arr1, arr2, rtol=1e-5, atol=1e-8):
...
+        return False
+    if (
+        dask_array
+        and isinstance(arr1, dask_array.Array)
+        and isinstance(arr2, dask_array.Array)
+    ):
+        # GH3068
+        if arr1.name == arr2.name:
+            return True
+
```
shoyer: It's a good practice to either always or never have a return value. So I would explicitly write `return None` here to indicate that you didn't forget that.

xarray/core/duck_array_ops.py
```diff
@@ -181,16 +181,49 @@ def allclose_or_equiv(arr1, arr2, rtol=1e-5, atol=1e-8):
...
     if arr1.shape != arr2.shape:
         return False
+    if (
+        dask_array
+        and isinstance(arr1, dask_array.Array)
+        and isinstance(arr2, dask_array.Array)
+    ):
+        # GH3068
+        if arr1.name == arr2.name:
+            return True
```
shoyer: This is looking a little repetitive. Can you make a helper function for doing these checks?

Also note the similar logic (checking object identity) inside `Variable.equals`

dcherian: Done. We can just call `lazy_array_equiv` in all these before doing the actual value comparison. Then only the `asarray` cast is duplicated.

Re `Variable.equals`:  
https://github.com/pydata/xarray/blob/fb0cf7b5fe56519a933ffcecbce9e9327fe236a6/xarray/core/variable.py#L1560-L1576

`equiv=duck_array_ops.array_equiv` by default which now calls `duck_array_ops.lazy_array_equiv` first so things are good?

dcherian: OK I've added the identity check to `lazy_array_equiv` too

xarray/core/merge.py
```diff
@@ -123,16 +124,24 @@ def unique_variable(
         combine_method = "fillna"
 
     if equals is None:
-        out = out.compute()
+        # first check without comparing values i.e. no computes
         for var in variables[1:]:
-            equals = getattr(out, compat)(var)
+            equals = getattr(out, compat)(var, equiv=lazy_array_equiv)
             if not equals:
```
shoyer: Should this be `if equals is not None`?

xarray/core/concat.py
```diff
@@ -189,6 +190,21 @@ def process_subset_opt(opt, subset):
...
                 for k in getattr(datasets[0], subset):
                     if k not in concat_over:
+                        equals[k] = None
+                        variables = [ds.variables[k] for ds in datasets]
+                        # first check without comparing values i.e. no computes
+                        for var in variables[1:]:
+                            equals[k] = getattr(variables[0], compat)(
+                                var, equiv=lazy_array_equiv
+                            )
+                            if not equals[k]:
```
shoyer: same concern as below -- should this be checking against `None` explicitly, which I believe `lazy_array_equiv` can return?

dcherian: I think this is right. We are exiting early because `equals[k] is not True` i.e. either the shapes are not equal, or one (or both) of the arrays is a numpy array or the dask names are not equal

xarray/core/merge.py
dcherian: I think this is right (as above) but the next one is wrong. I've fixed that and added a test.

dcherian:
Thanks for the review @shoyer 

> Thanks for adding the explicit checks is not True, etc. It's definitely good to be explicit 

I myself was confused when I went through it the second time :)

PR #3456, opened 2019-10-29
upgrade black verison to 19.10b0

max-sixty:
﻿Black version 19.10b0 recently came out. Looks like the biggest change is assigning to a one-item-tuple

### DIFF
--- a/xarray/core/merge.py
+++ b/xarray/core/merge.py
@@ -277,7 +277,7 @@ def collect_variables_and_indexes(
 
 
 def collect_from_coordinates(
-    list_of_coords: "List[Coordinates]"
+    list_of_coords: "List[Coordinates]",
 ) -> Dict[Hashable, List[MergeElement]]:
     """Collect variables and indexes to be merged from Coordinate objects."""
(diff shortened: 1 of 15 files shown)


max-sixty:
Self-merged given the CI breaks, hope that's OK

xarray/core/merge.py
```diff
@@ -320,7 +320,7 @@ def merge_coordinates_without_align(
 
 
 def determine_coords(
-    list_of_mappings: Iterable["DatasetLike"]
+    list_of_mappings: Iterable["DatasetLike"],
```
crusaderky: We should run ``black --t py36`` to fix all of these at once. Otherwise they'll continue randomly cropping every time a module loses compatibility with py35 for unrelated reasons.

max-sixty: I think I've done this—do you get any changes on running that?

crusaderky: mmh no, it looks already clean :+1:

max-sixty: Cheers!

PR #3460, opened 2019-10-29
fix test suite warnings re `drop`

max-sixty:
Some more warnings silenced

### DIFF
--- a/xarray/tests/test_duck_array_ops.py
+++ b/xarray/tests/test_duck_array_ops.py
@@ -440,7 +440,9 @@ def test_argmin_max(dim_num, dtype, contains_nan, dask, func, skipna, aggdim):
             **{aggdim: getattr(da, "arg" + func)(dim=aggdim, skipna=skipna).compute()}
         )
         expected = getattr(da, func)(dim=aggdim, skipna=skipna)
-        assert_allclose(actual.drop(actual.coords), expected.drop(expected.coords))
+        assert_allclose(
+            actual.drop(list(actual.coords)), expected.drop(list(expected.coords))
+        )
 
 
 def test_argmin_max_error():


max-sixty:
ref https://github.com/pydata/xarray/issues/3266. This was causing about a quarter of the total, because of the test's parameterization.

xarray/tests/test_duck_array_ops.py
```diff
@@ -440,7 +440,9 @@ def test_argmin_max(dim_num, dtype, contains_nan, dask, func, skipna, aggdim):
             **{aggdim: getattr(da, "arg" + func)(dim=aggdim, skipna=skipna).compute()}
         )
         expected = getattr(da, func)(dim=aggdim, skipna=skipna)
-        assert_allclose(actual.drop(actual.coords), expected.drop(expected.coords))
+        assert_allclose(
+            actual.drop(list(actual.coords)), expected.drop(list(expected.coords))
```
dcherian: do we want to special case the passing of coordinate objects for  convenience?

max-sixty: It's a classic special case question! https://github.com/pydata/xarray/pull/3128/files#r315000922

max-sixty: I hear you, but I can also see: https://github.com/pydata/xarray/pull/3128/files#r315000922

max-sixty: Fair point, but I can also see this POV: https://github.com/pydata/xarray/pull/3128/files#r315000922 ; I defer to @shoyer

max-sixty:
One other issue: it suggests using `drop_vars` but that doesn't exist!

shoyer:
> One other issue: it suggests using `drop_vars` but that doesn't exist!

Yes, this was a suggestion for a  new method

xarray/tests/test_duck_array_ops.py
dcherian: Ah wasn't aware. That makes sense.

max-sixty:
> Yes, this was a suggestion for a new method

How would this work differently from `.drop(var_name)`? Or would we deprecate that use of `drop`?

shoyer:
I think the idea is just to provide a more explicit option for dropping
variables instead of labels. It's a little messy all in one API.

max-sixty:
> I think the idea is just to provide a more explicit option for dropping variables instead of labels. It's a little messy all in one API.

Yeah, agree it's messy. I think we could have `drop_vars`, completely deprecate any use of `drop` apart from lables, and then we're in a decent place? Or is that too much churn?

shoyer:
Yeah, that was my thought

Issue #3434, opened 2019-10-22
v0.14.1 Release


crusaderky:
I think with the multiple recent breakages we've just had due to dependency upgrades, we should push out a patch release with some haste.

Please comment/add/object

Must have
--------------
- [x] numpy 1.18 support #3409
- [x] pseudonetcdf 3.1.0 support #3409, #3420 
- [x] require cftime != 1.0.4 #3463 
- [x] groupby reduce regression fix #3403
- [x] pandas master support #3440 
 
Nice to have
-----------------
- [x] ellipsis (...) work #1081, #3414, #3418, #3421, #3423, #3424
- [x] HTML repr #3425 (really mouth-watering, but I'm unsure about how far it is from completion)
- [x] groupby drop nan groups #3406 
- [x] deprecate `allow_lazy` #3435 
- [x] \_\_dask_tokenize\_\_ #3446 
- [x] dask name equality #3453
- [x] Leave empty slot when not using accessors #3531

dcherian:
#3406 needs a review

crusaderky:
@mathause I'm personally really uncomfortable pushing out a release when there are multiple known breakages against dependencies - see Must have list. Help is welcome!

crusaderky:
@barronh any update re pseudonetcdf 3.1?

barronh:
I believe there are two issues here.

First, the 3.1 release was not 2.7 compliant. It mixed named keyword arguments with `**kwds`. That could be fixed easily.

Second, the camx reader changed subclasses, which affects the xarray test for CF variables. I could update the test that would fix xarray test-case.

crusaderky:
@barronh, we don't care about Python 2.7 compliance. We support >=3.6.

barronh:
Got it. This is a test update to a test that is backward and forward compatible. I'll get something checked in.

mathause:
xref for numpy 1.18 numpy/numpy#14831

dcherian:
@mathause Thanks for filing these upstream issues!

dcherian:
Once #3491 is merged, we should be ready to release since numpy will fix itself (https://github.com/numpy/numpy/pull/14841)

mathause:
Don't we have to figure out whether to use `np.mean` or `np.nanmean` for datetime vectors?

dcherian:
I'm in favour of releasing even if we are currently broken with numpy master. 

Our current release is also broken with numpy master, so the situation is no worse...

[...]

PR #3463, opened 2019-10-30
unpin cftime

mathause:
I think the cftime problems should be fixed after the release of [v1.0.4.2](https://github.com/Unidata/cftime/releases/tag/v1.0.4.2rel)

 #3434

### DIFF
--- a/ci/requirements/py36-min-all-deps.yml
+++ b/ci/requirements/py36-min-all-deps.yml
@@ -13,7 +13,7 @@ dependencies:
   - cartopy=0.17
   - cdms2=3.1
   - cfgrib=0.9
-  - cftime=1.0.3
+  - cftime=1.0
   - coveralls
   - dask=1.2
   - distributed=1.27
--- a/ci/requirements/py36-min-all-deps.yml
+++ b/ci/requirements/py36-min-all-deps.yml
@@ -13,7 +13,7 @@ dependencies:
   - cartopy=0.17
   - cdms2=3.1
   - cfgrib=0.9
-  - cftime=1.0.3  # FIXME need 1.0.5 (not released yet); 1.0.4 is broken
+  - cftime
   - coveralls
   - dask=1.2
   - distributed=1.27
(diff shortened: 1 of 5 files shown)


ci/requirements/py36-min-all-deps.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - cartopy=0.17
   - cdms2=3.1
   - cfgrib=0.9
-  - cftime=1.0.3  # FIXME need 1.0.5 (not released yet); 1.0.4 is broken
+  - cftime
```
spencerkclark: I think we may want to keep the version in this particular file pinned at `1.0.3` as it represents the new minimum version of cftime we will test against for the time being; xref: https://github.com/pydata/xarray/pull/3431#discussion_r337606766.
```suggestion
  - cftime=1.0.3
```

mathause:
Ahh, yes of course... it's getting late.

ci/requirements/py36-min-all-deps.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - cartopy=0.17
   - cdms2=3.1
   - cfgrib=0.9
-  - cftime=1.0.3  # FIXME need 1.0.5 (not released yet); 1.0.4 is broken
+  - cftime=1.0.3
```
crusaderky: Could you change to 1.0, as per our policy of not using patch versions?

ci/requirements/py36-min-all-deps.yml
crusaderky: As per our versions policy, we only support the latest available patch version
http://xarray.pydata.org/en/stable/installing.html#minimum-dependency-versions

mathause:
Are the release notes fine, I wasn't sure what you are expecting.

crusaderky:
Looks good to me

max-sixty:
Thanks @mathause !

PR #3469, opened 2019-10-30
Fix integrate docs

keewis:
- [ ] Tests added

This updates the name in the parameter list of `Dataset.integrate` and adds an example. Also, the note in `DataArray.integrate` referred to `coord` which does not exist. I guess in the end to make it match `Dataset.integrate` it would be better to rename that parameter from `dim` to `coord`, but that would mean changing the API.


keewis:
done

PR #3471, opened 2019-10-30
Restrict pytest to 'unit' tests (i.e. not hypothesis)

max-sixty:
Currently Azure is running hypothesis tests in any test job that has hypothesis installed (found in https://github.com/pydata/xarray/pull/3285). This isn't required, since we have a separate job for hypothesis, and slows down the normal tests.

We can either exclude it in the test job or in the tests that pytest runs as specified in `setup.cfg`

I think adjusting pytest's test paths in `setup.cfg` is a more logical place: pytest is responsible for running normal / unit tests in `xarray/tests`, and hypothesis can run the tests in `properties`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [tool:pytest]
 python_files=test_*.py
-testpaths=xarray/tests properties
+testpaths=xarray/tests
 # Fixed upstream in https://github.com/kwgoodman/bottleneck/pull/199
 filterwarnings =
     ignore:Using a non-tuple sequence for multidimensional indexing is deprecated:FutureWarning


crusaderky:
That was a conscious decision based on the observation that the whole lot of the hypothesis tests was taking something like 3 seconds. Has this changed now?

max-sixty:
Ah. Am I thinking there's a separate hypothesis job when actually there isn't?

crusaderky:
@max-sixty there used to. I took it away.

crusaderky:
End-to-end runtime for properties, as of the git tip, is 1.65 seconds

max-sixty:
Great. Thanks for being ahead of me on some of these recent issues!

PR #3472, opened 2019-10-31
Type check sentinel values

### LINKED ISSUE #1603: Future of MultiIndex

max-sixty:
- [ ] Closes #xxxx

I did this before I realized it only works on mypy master! But posting so no one else does it. We can wait to merge until the next version of mypy is out

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -73,7 +73,7 @@ from .utils import (
     Default,
     Frozen,
     SortedKeysDict,
-    __default,
+    _default,
     _check_inplace,
     decode_numpy_dict_values,
     either_dict_or_kwargs,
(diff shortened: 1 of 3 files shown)


xarray/core/utils.py
```diff
@@ -701,3 +702,10 @@ def get_temp_dimname(dims: Container[Hashable], new_dim: Hashable) -> Hashable:
     while new_dim in dims:
         new_dim = "_" + str(new_dim)
     return new_dim
+
+
+class Default(Enum):
+    sentinel = 0
+
+
+__default = Default.sentinel
```
crusaderky: I think this would be just as good?
```python
class Default:
    pass

__default = Default()
```

crusaderky:
@max-sixty have a look at my changes and then feel free to merge

xarray/core/dataset.py
```diff
@@ -890,11 +887,11 @@ def _replace(  # type: ignore
                 self._coord_names = coord_names
             if dims is not None:
                 self._dims = dims
-            if attrs is not self.__default:
+            if not isinstance(attrs, Default):
```
max-sixty: I feel like there's some Chesterton's fence here: is there some reason we had `self.` here?

max-sixty:
Great, thanks @crusaderky 

I followed the Enum pattern because that was suggested in [python docs](https://github.com/python/typing/pull/240). But I agree your suggestion is a bit simpler. Maybe have a quick glance at those docs and unless you change your mind, hit the green button?

xarray/core/dataset.py
crusaderky: I don't think so. ``self.__default`` is misleading - it was a class attribute to begin with (you could rewrite it as ``Dataset.__default``).

crusaderky:
Ah now I understand what you meant with mypy master

Issue #2227, opened 2018-06-12
Slow performance of isel


JohnMrziglod:
Hi,

I get a very slow performance of Dataset.isel or DataArray.isel in comparison with the native numpy approach. Do you know where this comes from?

```python
ds = xr.Dataset(
    {
        "a": ("time", np.arange(55_000_000))
    }, coords={
        "time": np.arange(55_000_000)
    }
)
time_filter = ds.time > 50_000
```

Select some values with DataArray.isel:
```python
%timeit ds.a.isel(time=time_filter)
```
`2.22 s ± 375 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

Use the native numpy approach:
```python
%timeit ds.a.values[time_filter]
```
`163 ms ± 12.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 3.16.0-4-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: en_US.utf8
LOCALE: en_US.UTF-8

xarray: 0.10.4
pandas: 0.23.0
numpy: 1.14.2
scipy: 1.1.0
netCDF4: 1.4.0
h5netcdf: 0.5.1
h5py: 2.8.0
Nio: None
zarr: None
bottleneck: 1.2.1
cyordereddict: None
dask: 0.17.5
distributed: 1.21.8
matplotlib: 2.2.2
cartopy: 0.16.0
seaborn: 0.8.1
setuptools: 39.1.0
pip: 9.0.3
conda: None
pytest: 3.5.1
IPython: 6.4.0
sphinx: 1.7.4
</details>

rabernat:
I don't have experience using `isel` with boolean indexing. (Although the docs on [positional indexing](http://xarray.pydata.org/en/latest/indexing.html#positional-indexing) claim it is supported.) My guess is that that the time is being spent aligning the indexer with the array, which is unnecessary since you know they are already aligned. Probably not the most efficient pattern for xarray.

Here's how I would recommend writing the query using label-based selection:
```python
%timeit ds.a.sel(time=slice(50_001, None))
117 ms ± 5.29 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

maxim-lian:
@rabernat that's a good solution where it's a slice

[...]

PR #3481, opened 2019-11-04
Propagate indexes in DataArray binary operations.

dcherian:
Works by propagating indexes in DataArray._replace.

xref #2227. Tests pass!

 - [x] Tests added

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -386,6 +386,7 @@ class DataArray(AbstractArray, DataWithCoords):
         variable: Variable = None,
         coords=None,
         name: Union[Hashable, None, Default] = _default,
+        indexes=None,
     ) -> "DataArray":
         if variable is None:
             variable = self.variable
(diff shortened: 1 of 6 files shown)


xarray/core/indexes.py
```diff
@@ -35,6 +35,9 @@ def __contains__(self, key):
     def __getitem__(self, key):
         return self._indexes[key]
 
+    def __delitem__(self, key):
```
dcherian: needed for the groupby change above.

xarray/core/dataarray.py
```diff
@@ -386,14 +386,15 @@ def _replace(
...
+        indexes=None,
     ) -> "DataArray":
         if variable is None:
             variable = self.variable
         if coords is None:
             coords = self._coords
         if name is _default:
             name = self.name
-        return type(self)(variable, coords, name=name, fastpath=True)
+        return type(self)(variable, coords, name=name, fastpath=True, indexes=indexes)
```
max-sixty: Do we need to do this in `.copy` (which I think calls this method)?

dcherian: I can do that in a followup.

max-sixty: Great, yes only if needed, 👍  re small PRs in

PR #3420, opened 2019-10-20
Restore crashing CI tests on pseudonetcdf-3.1

crusaderky:
Related to #3409 

The crashes caused by pseudonetcdf-3.1 are blocking all PRs.
Sorry I don't know anything about pseudonetcdf. This PR takes the issue out of the critical path so that whoever knows about the library can deal with it in due time.

Issue #3409, opened 2019-10-17
test failure with upstream-dev


dcherian:
https://dev.azure.com/xarray/xarray/_build/results?buildId=1101&view=logs

```
=================================== FAILURES ===================================
_________________________ test_datetime_reduce[False] __________________________

dask = False

    @arm_xfail
    @pytest.mark.parametrize("dask", [False, True])
    def test_datetime_reduce(dask):
        time = np.array(pd.date_range("15/12/1999", periods=11))
        time[8:11] = np.nan
        da = DataArray(np.linspace(0, 365, num=11), dims="time", coords={"time": time})
    
        if dask and has_dask:
            chunks = {"time": 5}
            da = da.chunk(chunks)
    
        actual = da["time"].mean()
>       assert not pd.isnull(actual)
E       AssertionError: assert not True
E        +  where True = <function isna at 0x7f2475449ae8>(<xarray.DataArray 'time' ()>\narray('NaT', dtype='datetime64[ns]'))
E        +    where <function isna at 0x7f2475449ae8> = pd.isnull

xarray/tests/test_duck_array_ops.py:288: AssertionError
__________________________ test_datetime_reduce[True] __________________________

dask = True

    @arm_xfail
    @pytest.mark.parametrize("dask", [False, True])
    def test_datetime_reduce(dask):
        time = np.array(pd.date_range("15/12/1999", periods=11))
        time[8:11] = np.nan
        da = DataArray(np.linspace(0, 365, num=11), dims="time", coords={"time": time})
    
        if dask and has_dask:
            chunks = {"time": 5}
            da = da.chunk(chunks)
    
        actual = da["time"].mean()
>       assert not pd.isnull(actual)
E       AssertionError: assert not True
E        +  where True = <function isna at 0x7f2475449ae8>(<xarray.DataArray 'time' ()>\narray('NaT', dtype='datetime64[ns]'))
E        +    where <function isna at 0x7f2475449ae8> = pd.isnull

xarray/tests/test_duck_array_ops.py:288: AssertionError
=============================== warnings summary 
```

crusaderky:
This has now spread to all tests

crusaderky:
Delta of conda list (py36) from before and after the breakage:
```
< boto3                     1.9.252                    py_0    conda-forge
> boto3                     1.9.253                    py_0    conda-forge
< botocore                  1.12.252                   py_0    conda-forge
> botocore                  1.12.253                   py_0    conda-forge
[...]

PR #3485, opened 2019-11-05
uamiv test using only raw uamiv variables

barronh:
Previous test relied on CF generated metadata, but this test is more robust.

 - [x] Addresses #3409 #3420 #3434
 - [x] Tests fixed

Development machine is missing python3.6 environment, using azure to check. This should help me get this fix in faster.


dcherian:
Thanks @barronh !

PR #3487, opened 2019-11-05
Respect user-specified coordinates attribute.

### LINKED ISSUE #3351: custom "coordinates" attribute
This set of lines prevents users from writing files with a handcrafted `coordinates` attribute. This is useful when not all non-dimensional coordinates in a Dataset are appropriate for every data variable in a dataset.

https://github.com/pydata/xarray/blob/16fdac93c09036b44f90201d9078b17808f819e0/xarray/conventions.py#L667-L675

It seems to me like we should only automatically set "coordinates" when `"coordinates" not in attrs`. We could raise a warning saying that setting `attrs["coordinates"]` prevents full roundtripping of Datasets.

What do you think?

dcherian:
A minimally invasive solution to #3351. If `variable.encoding["coordinates"]` is specified, we write that attribute to disk and warn the user that roundtripping may not work.

 - [x] Closes #3351
 - [x] Tests added


dcherian:
@jhamman your input would be appreciated here if you have the time.

xarray/tests/test_backends.py
```diff
@@ -810,6 +810,17 @@ def equals_latlon(obj):
...
+        original["temp"].encoding["coordinates"] = "lat"
+        with self.roundtrip(original) as actual:
+            assert_identical(actual, original)
+        with create_tmp_file() as tmp_file:
+            original.to_netcdf(tmp_file)
+            with open_dataset(tmp_file, decode_coords=False) as ds:
+                assert equals_latlon(ds.precip.attrs["coordinates"])
+                assert "lon" not in ds["temp"].attrs["coordinates"]
+                assert "coordinates" not in ds["lat"].attrs
+                assert "coordinates" not in ds["lon"].attrs
```
jhamman: Am I correct in my interpretation that this should also work?

```python
assert 'lon' not in ds['temp'].coords
```

doc/io.rst
```diff
@@ -440,6 +440,17 @@ supported by netCDF4-python: 'standard', 'gregorian', 'proleptic_gregorian' 'nol
 By default, xarray uses the 'proleptic_gregorian' calendar and units of the smallest time
 difference between values, with a reference time of the first time value.
 
+Coordinates
+...........
+
+You can control the ``coordinates`` attribute written to disk by specifying ``encoding["coordinates"]``.
```
jhamman: I think it would be good to explicitly state that the we're talking about `DataArray.encoding`. The notion of the _global coordinate_ makes me think of a `Dataset.encoding['coordinate']` attribute. Perhaps you can clarify here (and in the whats new note).

xarray/tests/test_backends.py
dcherian: no. `"lon"` is still a coord in xarray-land because `ds["precip"].attrs["coordinates"} = "lat lon"` when the dataset is read. Other software might do the right thing.

doc/io.rst
dcherian: :+1:

dcherian:
updated tests and fixed whats-new

max-sixty:
Shall we merge?

dcherian:
I'm going to merge this. I'm happy to address any comments in future PRs.

This is a step forward because:
1. This change now lets xarray roundtrip datasets that have "coordinates" attributes specified.
2. It also lets users specify a coordinates attribute and writes it to disk (this raised an error earlier). 
3. Xarray should now set not-CF-compliant "global" coordinates attribute in fewer cases than earlier.

PR #3491, opened 2019-11-07
fix pandas-dev tests

### LINKED ISSUE #3440: Build failure with pandas master
See https://dev.azure.com/xarray/d5e7a686-a114-4b8c-a2d8-4b5b11efd896/_build/results?buildId=1218&view=logs&jobId=41d90575-019f-5cfd-d78e-c2adebf9a30b for a log.

Appears to be due to https://github.com/pandas-dev/pandas/pull/29062, which adds a `.attrs` attribute to pandas objects. We copy this attribute in the DataArray constructor.

dcherian:
- [x] Closes #3440

This PR makes `DataArray.__init__` ignore the pandas `attrs`. 

Not sure what we want to do about these attributes in the long term. One option would be to pop the `name` attribute, assign to `DataArray.name` and keep the rest as `DataArray.attrs`? But what if `name` clashes with the provided `name`?

### DIFF
--- a/ci/azure/install.yml
+++ b/ci/azure/install.yml
@@ -16,7 +16,7 @@ steps:
         --pre \
         --upgrade \
         matplotlib \
-        pandas=0.26.0.dev0+628.g03c1a3db2 \  # FIXME https://github.com/pydata/xarray/issues/3440
+        pandas \
         scipy
         # numpy \  # FIXME https://github.com/pydata/xarray/issues/3409
     pip install \
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -52,6 +52,7 @@ from .coordinates import (
 from .dataset import Dataset, merge_indexes, split_indexes
 from .formatting import format_item
 from .indexes import Indexes, default_indexes
+from .merge import PANDAS_TYPES
 from .options import OPTIONS
 from .utils import Default, ReprObject, _default, _check_inplace, either_dict_or_kwargs
 from .variable import (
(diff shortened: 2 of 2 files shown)


dcherian:
flake is complaining about
``` python
# Drop variables
    @overload  # noqa: F811
    def drop(
        self, labels: Union[Hashable, Iterable[Hashable]], *, errors: str = "raise"
    ) -> "Dataset":
        ...

    # Drop index labels along dimension
    @overload  # noqa: F811
    def drop(
        self, labels: Any, dim: Hashable, *, errors: str = "raise"  # array-like
    ) -> "Dataset":
        ...
```

```
./xarray/core/dataarray.py:1903:5: F811 redefinition of unused 'drop' from line 1896
./xarray/core/dataset.py:3531:5: F811 redefinition of unused 'drop' from line 3524
```

max-sixty:
@dcherian you can `# noqa` those, they're going away anyway; I can sort in the drop PR

crusaderky:
The F811 errors are a known issue that pyflakes has with @overload, already fixed in their git tip. For now just add ``# noqa: F811`` wherever you get it

max-sixty:
I merged the drop PR, so those overloads are now removed

max-sixty:
> Not sure what we want to do about these attributes in the long term. One option would be to pop the `name` attribute, assign to `DataArray.name` and keep the rest as `DataArray.attrs`? But what if `name` clashes with the provided `name`?

Agree! I think we could prioritize the supplied name above that in `attrs`. Another option would be raising an error if both were supplied.

dcherian:
I'm merging this so we can continue testing against pandas master. Discussion can continue in #3497

PR #3498, opened 2019-11-08
Add option to choose mfdataset attributes source.

### LINKED ISSUE #2382: Add option to choose the source of global attributes in mfdataset.

juseg:
Add a `master_file` keyword arguments to `open_mfdataset` to choose the source of global attributes in a multi-file dataset.

 - [x] Closes #2382 
 - [x] Tests added


juseg:
I think I'm done. Can someone look at it? The `master_file` keyword argument is borrowed from NetCDF (see #2382 and Unidata/netcdf4-python#835), although `xarray`'s mechanism is independant.

The default is `0`, which is consistent with current `xarray` behaviour. When concatenating model output in time it would be more logical to use `-1`, i.e. preserve history of the last file.

xarray/backends/api.py
```diff
@@ -825,6 +826,10 @@ def open_mfdataset(
         - 'override': if indexes are of same size, rewrite indexes to be
           those of the first object with that dimension. Indexes for the same
           dimension must have the same size in all objects.
+    master_file : int or str, optional
```
dcherian: This is netcDF4's documentation for `master_file`:

>  file to use as "master file", defining all the variables with an aggregation dimension and all global attributes.

let's make it clear that unlike netCDF4 we are only using this for attributes

dcherian:
Thanks @juseg . I've left a few comments.

I see that this is your first PR. Welcome to xarray! and thanks for contributing :clap:

xarray/backends/api.py
```diff
@@ -718,6 +718,7 @@ def open_mfdataset(
     autoclose=None,
     parallel=False,
     join="outer",
+    master_file=0,
```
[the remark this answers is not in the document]
juseg: I am in favor of integers for flexibility. This is familiar syntax to Python users, and we don't need to worry about a file called `first` or `last`.

xarray/tests/test_backends.py
```diff
@@ -2799,6 +2799,42 @@ def test_attrs_mfdataset(self):
...
+            ds2.attrs["test2"] = "bar"
+            ds1.to_netcdf(tmp1)
+            ds2.to_netcdf(tmp2)
+            with open_mfdataset(
+                [tmp1, tmp2], concat_dim="x", combine="nested", master_file=tmp2
+            ) as actual:
+                # attributes are inherited from the master file
+                assert actual.test2 == ds2.test2
+                # attributes from ds1 are not retained, e.g.,
+                with raises_regex(AttributeError, "no attribute"):
```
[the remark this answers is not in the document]
juseg: Done in 8cc3b34.

xarray/tests/test_backends.py
```diff
@@ -2799,6 +2799,42 @@ def test_attrs_mfdataset(self):
...
+            ds2 = original.isel(x=slice(5, 10))
+            ds1.attrs["test1"] = "foo"
+            ds2.attrs["test2"] = "bar"
+            ds1.to_netcdf(tmp1)
+            ds2.to_netcdf(tmp2)
+            with open_mfdataset(
+                [tmp1, tmp2], concat_dim="x", combine="nested", master_file=-1
+            ) as actual:
+                # attributes are inherited from the master file
+                assert actual.test2 == ds2.test2
```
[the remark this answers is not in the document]
juseg: Done in 8cc3b34.

xarray/backends/api.py
juseg: Do you suggest to use a different keyword, maybe `attrs_file`?
Or just clarify the difference in the docs? I don't mind.
@dcherian Thanks for the review!

xarray/backends/api.py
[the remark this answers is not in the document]
dcherian: That is a good point.

xarray/backends/api.py
dcherian: I was initially thinking of just adding a line to the docstring but we should think about renaming this to something like `attrs_from`?

juseg: So I've renamed it to `attrs_file` to avoid confusion with netCDF4. Thanks for pointing that. I am open to any name as long as the option is here.

juseg:
This will add a new kwarg in `open_mfdataset`. The current name is `attrs_file`, defaulting to 0 (current behaviour). Or? Suggestions welcome.

TomNicholas:
Thanks for this @juseg. The only problem I see is that a scalar number to specify the file only makes sense if it's a 1D list, but `open_mfdataset` can also accept nested list-of-lists (with `combine='nested'`), or ignore the order of the input entirely (with `combine='manual'`). What happens if you pass a list-of-lists of datasets? 

On the other hand specifying the particular filepath or object makes sense in all cases, so perhaps the easiest way to avoid ambiguity would be to restrict to that option? (The default would just be left as-is.)

juseg:
@TomNicholas Thanks for bringing the discussion live again! I'm not sure what happens in those cases, but I'm confident the default behaviour is unchanged, i.e. the attributes file is 0, whatever that 0 means (see my first commit).

If this is an issue I would suggest to discuss in a separate thread, as I think it is independent from my changes. On the other hand I am eager to keep the file number option because (1) `attrs_file=-1` is the behaviour that I need (to ensure that history is always preserved) and (2) `attrs_file=0` is the current behaviour (again, whatever that means for nested lists).

TomNicholas:
I'm not sure we should merge changes if we're unsure how they will behave in certain circumstances.

> On the other hand I am eager to keep the file number option because (1) attrs_file=-1 is the behaviour that I need

If we kept just the string specifier, you could still solve the problem of preserving the history:
```python
files_to_open = ['filepath1', 'filepath2']

ds = open_mfdataset(files_to_open, attrs_file=files_to_open[-1])
```
But then the option would always have clear and well-defined behaviour, even in more complex cases like `combine_by_coords`or `combine='nested'` with a >1D input file list.

juseg:
@TomNicholas I've had a closer look at the code. Nested lists of file paths are processed by:

https://github.com/pydata/xarray/blob/f2b2f9f62ea0f1020262a7ff563bfe74258ffaa1/xarray/backends/api.py#L881-L882

Using the method defined in:

https://github.com/pydata/xarray/blob/f2b2f9f62ea0f1020262a7ff563bfe74258ffaa1/xarray/core/combine.py#L15-L46

In Python 3.7+ the list of `paths` is essentially flattened list e.g.:

```
>>> import xarray as xr
>>> paths = [list('abc'), list('def')]
>>> list(xr.core.combine._infer_concat_order_from_positions(paths).values())
['a', 'b', 'c', 'd', 'e', 'f']
```
Unfortunately the current code uses a dictionary which means that in Python 3.6- the order is not guaranteed preserved. This also implies that the current default `0` is not well defined in case of nested lists.

https://github.com/pydata/xarray/blob/f2b2f9f62ea0f1020262a7ff563bfe74258ffaa1/xarray/backends/api.py#L900
https://github.com/pydata/xarray/blob/f2b2f9f62ea0f1020262a7ff563bfe74258ffaa1/xarray/backends/api.py#L964

On the other hands the `ids` are essentially ND indexes that could perhaps be used...

```
>>> list(xr.core.combine._infer_concat_order_from_positions(paths).keys())
[(0, 0), (0, 1), (0, 2), (1, 0), (1, 1), (1, 2)]
```

Or should we just stick to file paths as you suggest? And leave the default as is (e.g. ambiguous for Python 3.6-)?

TomNicholas:
Thanks @juseg .

> in Python 3.6- the order is not guaranteed preserved.

I think for python 3.6 and above the [order is preserved](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6) isn't it?

> the current default 0 is not well defined in case of nested lists.

Yes, this is what I was thinking of.

> the ids are essentially ND indexes that could perhaps be used...

We *could* do this, and that's how we would solve it in general, but I don't really think it's worth the effort/complexity.

> Or should we just stick to file paths as you suggest? And leave the default as is

I think so - if we do this then users can still easily pick the attributes from the file of their choosing (solving the original issue), and if someone wants to be able to choose the `attrs_file` in another way later then we can worry about that then.

xarray/backends/api.py
```diff
@@ -827,6 +828,14 @@ def open_mfdataset(
...
           those of the first object with that dimension. Indexes for the same
           dimension must have the same size in all objects.
+    attrs_file : str, optional
+        Path of the file used to read global attributes from.
+        By default global attributes are read from the first file provided.
+        Wildcard matches are sorted by filename. Nested lists are flattened
+        using a dictionary. For Python 3.6 and higher, global attributes are
+        read from the first file of the first list (of the first list etc.).
+        For Python 3.5 or lower, global attributes are read from an
+        arbitrary file.
```
TomNicholas: Actually it looks like we only currently [explicitly support](https://github.com/pydata/xarray/blob/f2b2f9f62ea0f1020262a7ff563bfe74258ffaa1/setup.py#L24) python 3.6 & 3.7, so this can be shortened to simply 
```
By default global attributes are read from the first file provided, with wildcard matches sorted by filename.
```

juseg: So... In this case should we just keep the file index option after all? For nested list that would be the index in the flattened list.

TomNicholas: No, the reason we wanted to avoid the integer index option wasn't because unordered dictionaries made it non-deterministic, but because it's a less clear API for users who aren't required to understand how the list flattening works under the hood. Also, it's not clear what passing an integer index should mean if `combine='by_coords'`.

By keeping only the string specifier, the behaviour is totally clear for users without any knowledge of what happens within either of the `combine` functions.

TomNicholas: No, the reason we wanted to avoid the integer index option wasn't because unordered dictionaries made it non-deterministic, but because it's a less clear API for users who aren't required to understand how the list flattening works under the hood. Also, it's not clear what passing an integer index should mean if `combine='by_coords'`.

By keeping only the string specifier, the behaviour is totally clear for users without any knowledge of what happens within either of the combine functions.

juseg: @TomNicholas Agreed. I simplified the docstring as suggested above (aa62e18) and mark this conversation resolved.

xarray/backends/api.py
juseg: @dcherian can we mark this as resolved? The `attrs_file` now only accept a file name (see other conversation below).

xarray/backends/api.py
```diff
@@ -827,6 +828,10 @@ def open_mfdataset(
         - 'override': if indexes are of same size, rewrite indexes to be
           those of the first object with that dimension. Indexes for the same
           dimension must have the same size in all objects.
+    attrs_file : str, optional
```
keewis: since `paths` may be a sequence of `pathlib.Path` objects, we should allow passing one here, too:
```suggestion
    attrs_file : os.PathLike, optional
```
or
```suggestion
    attrs_file : str or pathlib.Path, optional
```

xarray/backends/api.py
```diff
@@ -961,7 +966,13 @@ def open_mfdataset(
         raise
 
     combined._file_obj = _MultiFileCloser(file_objs)
-    combined.attrs = datasets[0].attrs
+
+    # read global attributes from the attrs_file or from the first dataset
+    if attrs_file is not None:
+        combined.attrs = datasets[paths.index(attrs_file)].attrs
```
keewis: since `Path` objects get converted to `str`, we need
```python
        if isinstance(attrs_file, Path):
            attrs_file = str(attrs_file)
```
somewhere to allow passing a `Path`. Could you also add a test for that?

juseg: Thank you for the review. I did not know about pathlib. I've added a test (aa72572).

TomNicholas:
I think there's nothing left to do here, thanks @juseg!

PR #3500, opened 2019-11-08
make coarsen reductions consistent with reductions on other classes

dcherian:
- [x] Tests added

This PR uses `inject_reduce_methods` to inject reduction methods into the Coarsen classes. So now we can do `coarsen.count()` and pass `skipna` down to the reduction methods.


max-sixty:
Great idea, there's lots of room to centralize some of these!

dcherian:
@fujisoup this could use a review if you have time.

xarray/core/variable.py
```diff
@@ -1843,11 +1843,11 @@ def coarsen(self, windows, func, boundary="exact", side="left"):
             func = getattr(duck_array_ops, name, None)
             if func is None:
                 raise NameError(f"{name} is not a valid method.")
-        return type(self)(self.dims, func(reshaped, axis=axes), self._attrs)
+        return type(self)(self.dims, func(reshaped, axis=axes, **kwargs), self._attrs)
```
fujiisoup: Probably now we can use `_replace` method?

dcherian:
> I don't remember why I added a separate injection function...

:+1:

> Can you add some tests for coarsen.count() and skipna in the reduction methods?

There is a `count` test in `test_variable`. I've added `sum(skipna=...)` tests there too now.

doc/whats-new.rst
```diff
@@ -85,6 +85,9 @@ New Features
   invoked. (:issue:`3378`, :pull:`3446`, :pull:`3515`)
   By `Deepak Cherian <https://github.com/dcherian>`_ and
   `Guido Imperiale <https://github.com/crusaderky>`_.
+- Added the :py:meth:`count` reduction method to both :py:class:`DatasetCoarsen`
+  and :py:class:`DataArrayCoarsen` objects. (:pull:`3500`)
+  By `Deepak Cherian <https://github.com/dcherian/>`_
```
keewis: now that #3516 is merged, this should be
```suggestion
  By `Deepak Cherian <https://github.com/dcherian>`_
```

dcherian: thanks!

dcherian:
> Probably I wondered if bottleneck works with multiple reduction axes.

I'll check this before merging.

dcherian:
In the case of multiple reduction axes i.e. `isinstance(axis, tuple) is True` we don't use the bottleneck method so things should be fine.

https://github.com/pydata/xarray/blob/ed05f9862622b00f40f7b9b99ccdb0ab3766ff0f/xarray/core/nputils.py#L210-L218

PR #3502, opened 2019-11-09
Harmonize `FillValue` and `missing_value` during encoding and decoding steps

andersy005:
As pointed out in https://github.com/jbusecke/cmip6_preprocessing/issues/5, xarray appears to be very strict during the *encoding* and *decoding* steps even when there are (harmless) discrepancies between `missing_value` and `_FillValue`. For instance, when dtypes of `missing_value` and `_FillValue` are different, xarray gives up: 

```python
In [74]: from xarray.coding import variables                                                    

In [75]: import numpy as np                                                                     

In [76]: import xarray as xr                                                                    

In [77]: original = xr.Variable( 
    ...:         ("x",), 
    ...:         [0.0, -1.0, 1.0], 
    ...:         encoding={"_FillValue": np.float32(1e20), "missing_value": np.float64(1e20)}, 
    ...:     )                                                                                  

In [78]: coder = variables.CFMaskCoder()                                                        

In [79]: encoded = coder.encode(original)                                                       
<ipython-input-79-9fbc3632e28b> in <module>
----> 1 encoded = coder.encode(original)

/glade/work/abanihi/devel/pangeo/xarray/xarray/coding/variables.py in encode(self, variable, name)
    156             raise ValueError(
    157                 "Variable {!r} has multiple fill values {}. "
--> 158                 "Cannot encode data. ".format(name, [fv, mv])
    159             )
    160 

ValueError: Variable None has multiple fill values [1e+20, 1e+20]. Cannot encode data. 
```

 - [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/xarray/coding/variables.py
+++ b/xarray/coding/variables.py
@@ -162,17 +162,15 @@ class CFMaskCoder(VariableCoder):
             )
 
         if fv is not None:
-            encoding["_FillValue"] = np.asarray(encoding["_FillValue"]).astype(
-                data.dtype
-            )
+            # Ensure _FillValue is cast to same dtype as data's
+            encoding["_FillValue"] = data.dtype.type(encoding["_FillValue"])
             fill_value = pop_to(encoding, attrs, "_FillValue", name=name)
             if not pd.isnull(fill_value):
                 data = duck_array_ops.fillna(data, fill_value)
 
         if mv is not None:
-            encoding["missing_value"] = np.asarray(encoding["missing_value"]).astype(
-                data.dtype
-            )
+            # Ensure missing_value is cast to same dtype as data's
+            encoding["missing_value"] = data.dtype.type(encoding["missing_value"])
             fill_value = pop_to(encoding, attrs, "missing_value", name=name)
             if not pd.isnull(fill_value) and fv is None:
                 data = duck_array_ops.fillna(data, fill_value)
--- a/xarray/coding/variables.py
+++ b/xarray/coding/variables.py
@@ -162,13 +162,17 @@ class CFMaskCoder(VariableCoder):
             )
 
         if fv is not None:
-            # TODO: Cast _FillValue to the same dtype as the data
+            encoding["_FillValue"] = np.asarray(encoding["_FillValue"]).astype(
+                data.dtype
+            )
             fill_value = pop_to(encoding, attrs, "_FillValue", name=name)
             if not pd.isnull(fill_value):
                 data = duck_array_ops.fillna(data, fill_value)
 
         if mv is not None:
-            # TODO: Cast missing_value to the same dtype as the data
+            encoding["missing_value"] = np.asarray(encoding["missing_value"]).astype(
+                data.dtype
+            )
             fill_value = pop_to(encoding, attrs, "missing_value", name=name)
             if not pd.isnull(fill_value) and fv is None:
                 data = duck_array_ops.fillna(data, fill_value)
--- a/xarray/coding/variables.py
+++ b/xarray/coding/variables.py
@@ -8,7 +8,6 @@ import pandas as pd
 
 from ..core import dtypes, duck_array_ops, indexing
 from ..core.pycompat import dask_array_type
-from ..core.utils import equivalent
 from ..core.variable import Variable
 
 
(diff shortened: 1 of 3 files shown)


andersy005:
@dcherian, what is the right way to do the type casting in `encode()`?

I thought of trying something along these lines:

```python
encoding["missing_value"] = encoding["missing_value"].astype(data.dtype)
```
However, I quickly realized that this breaks when `encoding["missing_value"]` is not of a numpy object.

dcherian:
Thanks for taking this on, Anderson.

You'll have to convert back from the numpy array before exiting that function. It may be simpler if you avoid the cast and just keep the change from `equivalent` to `allclose_or_equiv`.

For reference, here's how xarray determines `dtype` for netCDF4
https://github.com/pydata/xarray/blob/4e9240a2087ffbf119919e1ac98046bbf164f94d/xarray/backends/netCDF4_.py#L131-L141

(I found this by working looking for the `createVariable` statement here)
https://github.com/pydata/xarray/blob/4e9240a2087ffbf119919e1ac98046bbf164f94d/xarray/backends/netCDF4_.py#L439-L485

xarray/coding/variables.py
```diff
@@ -152,18 +151,28 @@ def encode(self, variable, name=None):
...
+            and mv is not None
+            and not duck_array_ops.allclose_or_equiv(fv, mv)
+        ):
             raise ValueError(
                 "Variable {!r} has multiple fill values {}. "
                 "Cannot encode data. ".format(name, [fv, mv])
             )
 
         if fv is not None:
+            encoding["_FillValue"] = np.asarray(encoding["_FillValue"]).astype(
```
shoyer: Maybe using a NumPy scalar would work here? e.g., `encoding["_FillValue"] = data.dtype.type(encoding["_FillValue"])`

Actually I think you'll need to modify the `pop_to` line below, which overrides `_FillValue` in encoding.

andersy005: Thank you for the hint, @shoyer! 

> Actually I think you'll need to modify the pop_to line below, which overrides _FillValue in encoding.

I took a look at `pop_to()` function, and it appears that it is meant to just move a key/value from one dict to another dict. If I understand this correctly, we need to do type casting prior the `pop_to()` call, right?

andersy005:
@dcherian, 

> You'll have to convert back from the numpy array before exiting that function.

I was able to address this using @shoyer's suggestion above.

xarray/coding/variables.py
```diff
@@ -152,18 +151,26 @@ def encode(self, variable, name=None):
...
         mv = encoding.get("missing_value")
 
-        if fv is not None and mv is not None and not equivalent(fv, mv):
+        if (
+            fv is not None
+            and mv is not None
+            and not duck_array_ops.allclose_or_equiv(fv, mv)
+        ):
             raise ValueError(
                 "Variable {!r} has multiple fill values {}. "
```
dcherian: Let's make this nicer. Maybe something like 
```
f"Variable {name!r} has conflicting _FillValue ({fv}) and missing_value ({mv}). Cannot encode data."
```

xarray/coding/variables.py
```diff
@@ -152,18 +151,26 @@ def encode(self, variable, name=None):
...
                 "Cannot encode data. ".format(name, [fv, mv])
             )
 
         if fv is not None:
+            # Ensure _FillValue is cast to same dtype as data's
+            encoding["_FillValue"] = data.dtype.type(encoding["_FillValue"])
```
dcherian: nit:
```suggestion
            encoding["_FillValue"] = data.dtype.type(fv)
```

xarray/coding/variables.py
```diff
@@ -152,18 +151,26 @@ def encode(self, variable, name=None):
...
             fill_value = pop_to(encoding, attrs, "_FillValue", name=name)
             if not pd.isnull(fill_value):
                 data = duck_array_ops.fillna(data, fill_value)
 
         if mv is not None:
+            # Ensure missing_value is cast to same dtype as data's
+            encoding["missing_value"] = data.dtype.type(encoding["missing_value"])
```
dcherian: nit:
```suggestion
            encoding["missing_value"] = data.dtype.type(mv)
```

andersy005:
@dcherian & @shoyer, thank you both for your help! 

In which section(bug fixes? enhancements?) in `whats-new.rst` should I document these changes?

dcherian:
Let's do bug fixes

max-sixty:
Thanks @andersy005 !

PR #3505, opened 2019-11-09
Remove syntax warning

max-sixty:
Currently raises a `SyntaxWarning` (though I don't think causing any issues)

```
/home/vsts/work/1/s/xarray/plot/plot.py:291: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if args is ():
```

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -288,7 +288,7 @@ def line(
         )
 
     # The allargs dict passed to _easy_facetgrid above contains args
-    if args is ():
+    if args == ():
         args = kwargs.pop("args", ())
     else:
         assert "args" not in kwargs


crusaderky:
Yes, ``is ()`` instead of ``== ()`` accidentally works on CPython because in order to save RAM CPython has a bunch of hardcoded objects that are defined at interpreter level. That's however an implementation detail, and there's no guarantee it won't break on an alternative implementation, or even on a future version of CPython.

```python
>>> a = ()                                                                 
>>> b = ()                                                                 
>>> a is b                                                                 
True
>>> a = (1, 2)                                                             
>>> b = (1, 2)                                                             
>>> a is b                                                                 
False
>>> x = 1                                                                  
>>> y = 1                                                                  
>>> x is y                                                                 
True
>>> x = 1000                                                          
>>> y = 1000                                                      
>>> x is y                                                                 
False
>>> for x in range(1000): 
...     y = int(str((x))) 
...     if x is not y: 
...         print(x) 
...         break                                                                         
257

```

PR #3507, opened 2019-11-09
DOC: update bottleneck repo url

qwhelan:
The bottleneck repo url has changed as part of a maintainership handover, so updating in the documentation here.


max-sixty:
Thanks @qwhelan !

PR #3511, opened 2019-11-12
Format some of the docs code

max-sixty:
I struggled to parse some of the code in the docs, and so formatted some of the hardest-to-read code (e.g. `     mask = (ds.coords['lat']>20)&(ds.coords['lat']<60)&(ds.coords['lon']>220)&(ds.coords['lon']<260) `!) with black.

I'm not sure whether there's a more automated solution for this?

On a related point, I'm not sure why some of the code examples are colorized and some are B&W? I couldn't see anything obvious in the rst.

image

### DIFF
--- a/doc/indexing.rst
+++ b/doc/indexing.rst
@@ -209,19 +209,22 @@ simultaneously, returning a new dataset:
 
 .. ipython:: python
 
-    da = xr.DataArray(np.random.rand(4, 3),
-                      [('time', pd.date_range('2000-01-01', periods=4)),
-                       ('space', ['IA', 'IL', 'IN'])])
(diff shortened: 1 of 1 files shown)


shoyer:
Thanks @max-sixty !

That style different is really puzzling, I wonder if there is a bug somewhere in the relevant sphinx extensions.

max-sixty:
Where can I see what the sphinx extensions are? I can put an issue in for those (but don't know that code well at all)

dcherian:
I think they are here:
https://github.com/pydata/xarray/blob/b74f80ca2df4920f711f9fe5762458c53ce3c2c6/doc/conf.py#L69-L80

PR #3515, opened 2019-11-12
Recursive tokenization

crusaderky:
After misreading the dask documentation  <https://docs.dask.org/en/latest/custom-collections.html#deterministic-hashing>, I was under the impression that the output of ``__dask_tokenize__`` would be recursively parsed, like it happens for ``__getstate__`` or ``__reduce__``. That's not the case - the output of ``__dask_tokenize__`` is just fed into a str() function so it has to be made explicitly recursive!

### DIFF
--- a/xarray/tests/test_dask.py
+++ b/xarray/tests/test_dask.py
@@ -1283,6 +1283,20 @@ def test_token_identical(obj, transform):
     )
 
 
+def test_recursive_token():
+    """Test that tokenization is invoked recursively, and doesn't just rely on the
+    output of str()
+    """
+    a = np.ones(10000)
+    b = np.ones(10000)
+    b[5000] = 2
+    assert str(a) == str(b)
+    assert dask.base.tokenize(a) != dask.base.tokenize(b)
+    xa = DataArray(a)
+    xb = DataArray(b)
+    assert dask.base.tokenize(xa) != dask.base.tokenize(xb)
+
+
 @requires_scipy_or_netCDF4
 def test_normalize_token_with_backend(map_ds):
     with create_tmp_file(allow_cleanup_failure=ON_WINDOWS) as tmp_file:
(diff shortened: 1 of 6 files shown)


crusaderky:
xref https://github.com/pydata/sparse/issues/300

xarray/tests/test_dask.py
```diff
@@ -1283,6 +1283,32 @@ def test_token_identical(obj, transform):
...
+    a = np.ones(10000)
+    b = np.ones(10000)
+    b[5000] = 2
+    assert str(a) == str(b)
+    assert dask.base.tokenize(a) != dask.base.tokenize(b)
+
+    # Test DataArray and Variable
+    da_a = DataArray(a)
+    da_b = DataArray(b)
+    assert dask.base.tokenize(da_a) != dask.base.tokenize(da_b)
```
dcherian: Just so I'm clear, these are two different (but equal) xarray objects, so we don't want dask to think that they are the same?

xarray/tests/test_sparse.py
```diff
@@ -856,6 +856,10 @@ def test_dask_token():
     import dask
 
     s = sparse.COO.from_numpy(np.array([0, 0, 1, 2]))
+
```
dcherian: xfail this instead?

xarray/tests/test_dask.py
crusaderky: They aren't equal. da_b[5000] == 2, while da_a[5000] == 1.

xarray/tests/test_sparse.py
crusaderky: Yeah that was an option. But it would have completely disabled the test, and the test is really about not accidentally invoking ``self.values`` (which in turn invoked ``self._variable._data.__array__()``) on NEP18-compatible backends. It is not really about testing sparse specifically; it just conveniently relies on the fact that ``COO.__array__`` raises an exception. A more correct, but also more laborious, test could have created a NEP18-compatible dummy class on the fly.

Issue #3370, opened 2019-10-03
Hundreds of Sphinx errors


crusaderky:
sphinx-build emits a ton of errors that need to be polished out:

https://readthedocs.org/projects/xray/builds/ -> latest -> open last step

Options for the long term:
- Change the "Docs" azure pipelines job to crash if there are new failures. From past experience though, this should come together with a sensible way to whitelist errors that can't be fixed. This will severely slow down development as PRs will systematically fail on such a check.
- Add a task in the release process where, immediately before closing a release, the maintainer needs to manually go through the sphinx-build log and fix any new issues. This would be a major extra piece of work for the maintainer.

I am honestly not excited by either of the above. Alternative suggestions are welcome.

dcherian:
Is there a sphinx or restructured text linter that we can run? e.g. https://github.com/PyCQA/doc8

max-sixty:
Do these matter?

Do we know of examples of bad outcomes (e.g. hard-to-read docs)?

crusaderky:
Yes. For example

http://xarray.pydata.org/en/stable/generated/xarray.Dataset.integrate.html#xarray.Dataset.integrate

The sphinx log states:

/home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/xarray-0.14.0+37.g96cc2bc6-py3.7.egg/xarray/core/dataset.py:docstring of xarray.Dataset.integrate:12: WARNING: Unexpected indentation.

What I just now noticed is that several docstrings may be imported from numpy - which complicates everything.

max-sixty:
> The sphinx log states:
> 
> /home/docs/checkouts/readthedocs.org/user_builds/xray/conda/latest/lib/python3.7/site-packages/xarray-0.14.0+37.g96cc2bc6-py3.7.egg/xarray/core/dataset.py:docstring of xarray.Dataset.integrate:12: WARNING: Unexpected indentation.

That's a good case...

max-sixty:
I ran https://github.com/myint/docformatter on our code `docformatter --in-place -r --wrap-summaries=88 . `; results [here](https://github.com/pydata/xarray/compare/master...max-sixty:docformatter?expand=1)

But it doesn't solve many of our issues, which are around the sphinx docstrings 

Happy to do a PR, but worry it's more churn than help?

keewis:
[...]

PR #3516, opened 2019-11-13
Silence sphinx warnings

keewis:
This PR tries to silence most sphinx warnings. 5d9d263 is a collection of changes that silence warnings, but they also reduce the readability of the plain text docstrings, so I'd especially appreciate feedback on these.

There are still some warnings left, due to various reasons:
* the ufunc warnings are inherited from numpy
* some are due to duplicate target names in `whats-new.rst` (e.g. `CF Conventions` but also developer names).
* two files are not included in any toctree (`README.rst` and `api-hidden.rst`)
* references in `whats-new.rst` to labels that were either renamed or removed
* conflicting definitions for properties
* the CI also lists a lot of broken references to other functions / methods, which do not show for me

To fix these, I would
* update the numpy docstrings upstream or find a way to ignore these warnings (probably both)
* reference both files from somewhere
* track down the labels the references refer to and add them back, or disable the links

but I have no idea how to deal with the duplicate target names, duplicate definitions of properties, where to put the references to both files or if it would be better to leave the broken references broken. I did not look at the long list of broken references the CI shows since I can't reproduce it.

 - [x] short-term fix for #3370

### DIFF
--- a/xarray/ufuncs.py
+++ b/xarray/ufuncs.py
@@ -13,6 +13,7 @@ priority order:
 Once NumPy 1.10 comes out with support for overriding ufuncs, this module will
 hopefully no longer be necessary.
 """
+import textwrap
 import warnings as _warnings
 
 import numpy as _np
@@ -78,10 +79,43 @@ class _UFuncDispatcher:
         return res
 
 
+def _skip_signature(doc):
+    if doc.startswith(name):
+        signature_end = doc.find("\n\n")
+        doc = doc[signature_end + 2 :]
+
+    return doc
+
+
+def _remove_unused_reference_labels(doc):
+    max_references = 5
+    for num in range(max_references):
+        label = f".. [{num}]"
+        reference = f"[{num}]_"
+        index = f"{num}.    "
+
+        if label not in doc or reference in doc:
+            continue
+
+        doc = doc.replace(label, index)
+
+    return doc
+
+
+def _dedent(doc):
+    if not isinstance(doc, str):
+        return doc
+
+    return textwrap.dedent(doc)
+
+
 def _create_op(name):
     func = _UFuncDispatcher(name)
     func.__name__ = name
-    doc = getattr(_np, name).__doc__
+    doc = _remove_unused_reference_labels(
+        _skip_signature(_dedent(getattr(_np, name).__doc__))
+    )
+
     func.__doc__ = (
         "xarray specific variant of numpy.%s. Handles "
         "xarray.Dataset, xarray.DataArray, xarray.Variable, "
--- a/xarray/ufuncs.py
+++ b/xarray/ufuncs.py
@@ -80,6 +80,9 @@ class _UFuncDispatcher:
 
 
 def _skip_signature(doc):
+    if not isinstance(doc, str):
+        return doc
+
     if doc.startswith(name):
         signature_end = doc.find("\n\n")
         doc = doc[signature_end + 2 :]
@@ -88,6 +91,9 @@ def _skip_signature(doc):
 
 
 def _remove_unused_reference_labels(doc):
+    if not isinstance(doc, str):
+        return doc
+
     max_references = 5
     for num in range(max_references):
         label = f".. [{num}]"
@@ -112,9 +118,9 @@ def _dedent(doc):
 def _create_op(name):
     func = _UFuncDispatcher(name)
     func.__name__ = name
-    doc = _remove_unused_reference_labels(
-        _skip_signature(_dedent(getattr(_np, name).__doc__))
-    )
+    doc = getattr(_np, name).__doc__
+
+    doc = _remove_unused_reference_labels(_skip_signature(_dedent(doc)))
 
     func.__doc__ = (
         "xarray specific variant of numpy.%s. Handles "
--- a/xarray/ufuncs.py
+++ b/xarray/ufuncs.py
@@ -79,7 +79,7 @@ class _UFuncDispatcher:
         return res
 
 
-def _skip_signature(doc):
+def _skip_signature(doc, name):
     if not isinstance(doc, str):
         return doc
 
@@ -120,7 +120,7 @@ def _create_op(name):
     func.__name__ = name
     doc = getattr(_np, name).__doc__
 
-    doc = _remove_unused_reference_labels(_skip_signature(_dedent(doc)))
+    doc = _remove_unused_reference_labels(_skip_signature(_dedent(doc), name))
 
     func.__doc__ = (
         "xarray specific variant of numpy.%s. Handles "
--- a/doc/README.rst
+++ b/doc/README.rst
@@ -1,3 +1,5 @@
+:orphan:
+
 xarray
 ------
 
--- a/doc/api-hidden.rst
+++ b/doc/api-hidden.rst
@@ -32,9 +32,11 @@
    core.groupby.DatasetGroupBy.first
    core.groupby.DatasetGroupBy.last
    core.groupby.DatasetGroupBy.fillna
+   core.groupby.DatasetGroupBy.quantile
    core.groupby.DatasetGroupBy.where
 
    Dataset.argsort
+   Dataset.astype
    Dataset.clip
    Dataset.conj
    Dataset.conjugate
@@ -73,6 +75,7 @@
    core.groupby.DataArrayGroupBy.first
    core.groupby.DataArrayGroupBy.last
    core.groupby.DataArrayGroupBy.fillna
+   core.groupby.DataArrayGroupBy.quantile
    core.groupby.DataArrayGroupBy.where
 
    DataArray.argsort
--- a/doc/api-hidden.rst
+++ b/doc/api-hidden.rst
@@ -2,6 +2,8 @@
 .. This extra page is a work around for sphinx not having any support for
 .. hiding an autosummary table.
 
+:orphan:
+
 .. currentmodule:: xarray
 
 .. autosummary::
--- a/doc/combining.rst
+++ b/doc/combining.rst
@@ -255,11 +255,11 @@ Combining along multiple dimensions
   ``combine_nested``.
 
 For combining many objects along multiple dimensions xarray provides
-:py:func:`~xarray.combine_nested`` and :py:func:`~xarray.combine_by_coords`. These
+:py:func:`~xarray.combine_nested` and :py:func:`~xarray.combine_by_coords`. These
 functions use a combination of ``concat`` and ``merge`` across different
 variables to combine many objects into one.
 
-:py:func:`~xarray.combine_nested`` requires specifying the order in which the
+:py:func:`~xarray.combine_nested` requires specifying the order in which the
 objects should be combined, while :py:func:`~xarray.combine_by_coords` attempts to
 infer this ordering automatically from the coordinates in the data.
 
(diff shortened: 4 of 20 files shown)


max-sixty:
Awesome, thanks @keewis !!

I think the changes in https://github.com/pydata/xarray/commit/5d9d263e40e1f67910cbefaf96d46a91c560b8b5 are good (and having one standard is good for long-term readability)

For old whatsnew entries, I think it's fine that the references have decayed if we can silence the warnings (maybe easiest way is your suggestion to just remove the links, though)

Should we put an issue in for the numpy docstrings? 

As a follow-up, is there a way to add a check in CI so these don't stack up again as people make changes? Or maybe difficult without stamping them all out?

keewis:
CI: I think it's possible using `sphinx-build -W --keep-going`, which checks all and returns a non-zero error code, but first we need to stamp out all warnings (otherwise we have a lot of failing checks on unrelated PRs)

Numpy: sure, but updating upstream is only a long term fix, we will still have to somehow ignore these warnings if we want to check for new warnings using CI.

keewis:
I took a shot at using `:orphan:`, but [`doc/README.rst`](https://xarray.pydata.org/en/latest/README.html) seems like it should be excluded from the documentation?

Also, while trying to fix the nit-picky warnings, I found this:
```python
>>> import xarray as xr
>>> ds = xr.Dataset(data_vars={"a": ("x", list("abcd"))}, coords={"x": range(4)})
>>> ds.a.groupby(ds.x % 2 == 0).quantile
<bound method DataArrayGroupBy.quantile of DataArrayGroupBy, grouped over 'x' 
2 groups with labels False, True.>
>>> ds.groupby(ds.x % 2 == 0).quantile
AttributeError: 'DatasetGroupBy' object has no attribute 'quantile'
```
despite the documentation claiming otherwise (and indeed, `DataArrayGroupBy` explicitly implements `quantile` while `DatasetGroupBy` doesn't). Does this deserve its own issue?

Implementing `DatasetGroupBy.quantile` probably won't silence the nit-picky warning, though, since `DataArrayGroupBy.quantile` has the same problem. I'll investigate more on this.

max-sixty:
> I took a shot at using `:orphan:`, but [`doc/README.rst`](https://xarray.pydata.org/en/latest/README.html) seems like it should be excluded from the documentation?

Is `doc/README.rst` some special page we have to have? It seems to have no content value at the moment? Could we remove?

> despite the documentation claiming otherwise (and indeed, `DataArrayGroupBy` explicitly implements `quantile` while `DatasetGroupBy` doesn't). Does this deserve its own issue?

Definitely, good spot

keewis:
does anyone have an idea about what to do about broken links in `whats-new.rst`? Since we don't want to update them, the easiest way to silence these warnings would be to remove any prefixes like `:func:` or `:py:meth:`. I can't seem to find anything on this, but ideally there would be a way to mark links as intentionally broken. Thoughts?

max-sixty:
> There are also a lot of those that reference `xray` but apart from that would work fine. 

I think it'd be good to do a find/replace on those, if that solves our issues?

Or are there lots that remain despite that?

keewis:
unfortunately, those are not the majority of warnings and this would also cause warnings for deprecated methods (like `xray.Dataset.isel_points`). Unless someone has a better idea, I'd vote for converting `:py:meth:...` links to double backtick-quoted text

I also created a [preview page](https://xarray-keewis.readthedocs.io/en/silence-sphinx-warnings/), but that currently fails. Can someone help me with the [build errors](https://readthedocs.org/projects/xarray-keewis/builds/9969807/) (excessive memory consumption)?

max-sixty:
Yes good idea @keewis 

Where do you see the excessive memory consumption? When building locally?

I see the tests [failing](
https://dev.azure.com/xarray/xarray/_build/results?buildId=1514&view=logs&jobId=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36&taskId=68484831-0a19-5145-bfe9-6309e5f7691d&lineStart=271&lineEnd=274&colStart=1&colEnd=1) from:

```
Sphinx parallel build error:
RuntimeError: Non Expected warning in `/home/vsts/work/1/s/doc/plotting.rst` line 570
##[error]The operation was canceled.
```

keewis:
the CI will probably work when pushing new commits (or one might try to remove `-j auto` so sphinx runs sequentially: this might take longer but should avoid timeouts).

No, I was trying to set up readthedocs for this branch which I thought could make discussing the changes a bit easier. However, the [current build](https://readthedocs.org/projects/xarray-keewis/builds/9969807/) fails because conda ends up consuming too much memory. My impression was that this is a common problem that was somehow fixed for master.

keewis:
for author names, would it be reasonable to always require a format of `Name <https://github.com/user>`? The warnings regarding these was due to inconsistencies with the link location (mostly a trailing slash or using `http` instead of `https`).

max-sixty:
> would it be reasonable to always require a format of `Name <https://github.com/user>`? The warnings regarding these was due to inconsistencies with the link location (mostly a trailing slash or using `http` instead of `https`).

For sure. How's your regex-fu for find-replacing? I can try and have a go otherwise

keewis:
there were not enough entries to actually require regex-replace, so I did it by hand.

How would you handle links to `h5netcdf` and `cftime`? Convert them to double-backtick quoted? I don't think either of them has working intersphinx entries?

dcherian:
> How would you handle links to h5netcdf and cftime? Convert them to double-backtick quoted? I don't think either of them has working intersphinx entries?

Sounds good!

keewis:
now the only warnings left in non-nitpicky mode are the ones in the numpy docstrings and these:
```
whats-new.rst:1268: WARNING: undefined label: interpolating values with interp (if the link has no caption the label must precede a section header)
whats-new.rst:1387: WARNING: undefined label: faq.other_projects (if the link has no caption the label must precede a section header)
whats-new.rst:3253: WARNING: undefined label: copies vs views (if the link has no caption the label must precede a section header)
whats-new.rst:4035: WARNING: unknown document: tutorial
```
which are all related to the documentation. What should we do with these? I think `interpolating values with interp` never had an existing label, `copies vs views` is still there but with underscores instead of spaces and `faq.other_projects` got moved/renamed to `related-projects`. `tutorial` got removed a long time ago. I'm not sure about the first three, but the entry for `tutorial` actually looks fine, I'd simply put it in double-backticks.

dcherian:
> I think interpolating values with interp never had an existing label, 

Lets do :ref:`interp` or whatever works for references to `interpolation.rst`

> copies vs views is still there but with underscores instead of spaces and faq.other_projects got moved/renamed to related-projects

Let's change these appropriately. 

> the entry for tutorial actually looks fine, I'd simply put it in double-backticks.

:+1:

keewis:
looking at the links to `h5netcdf` and `cftime` again: while `h5netcdf` is only referenced once (`h5netcdf.File` in `whats-new.rst`), `cftime.datetime` is referenced quite often and not just in `whats-new.rst`. So I think it might be better to ask the `cftime` project to fix their intersphinx:
```bash
$ python -m sphinx.ext.intersphinx https://unidata.github.io/cftime/objects.inv
std:doc
	api                                      API                                     : api.html
	index                                    cftime                                  : index.html
	installing                               Installation                            : installing.html
std:label
	genindex                                 Index                                   : genindex.html
	modindex                                 Module Index                            : py-modindex.html
	search                                   Search Page                             : search.html
```

keewis:
I think I found the reasons for the numpy docstring warnings:
1. some of the numpy functions include signatures. I think that is because they are written in C? In the signature there is an asterisk which sphinx misinterprets.
2. the numpy docstrings are indented which somehow confuses sphinx, but only for some of the functions
3. some references are labeled, but not used

all of these can be silenced by rewriting the docstrings:
```python
def _create_op(name):
    func = _UFuncDispatcher(name)
    func.__name__ = name
    doc = textwrap.dedent(getattr(_np, name).__doc__)
    if doc.strip().startswith(name):
        start_location = doc.strip().find("\n\n")
        doc = doc[start_location + 2:]

    for ref_num in range(5):
        if f".. [{ref_num}]" not in doc or f"[{ref_num}]_" in doc:
            continue

        doc = doc.replace(f".. [{ref_num}]", f"{ref_num}.    ")

    func.__doc__ = (
        "xarray specific variant of numpy.%s. Handles "
        "xarray.Dataset, xarray.DataArray, xarray.Variable, "
        "numpy.ndarray and dask.array.Array objects with "
        "automatic dispatching.\n\n"
        "Documentation from numpy:\n\n%s" % (name, doc)
    )
    return func
```
However, I'm not sure this is a good idea.

max-sixty:
Great! That's cool we can start checking warnings.

I think wrapping the numpy docstrings seems pretty reasonable—what are you concerns? Any thoughts @dcherian ?

dcherian:
I think so. Though would a decorator be cleaner? I'm pretty sure dask uses a decorator when they wrap numpy functions

keewis:
re warnings checking: I'd start with something like `sphinx-build -WT --keep-going` and wait until we fixed all the nitpicky warnings to enable `-n`

dcherian:
Will it still fail of there are errors?

keewis:
it will print all warnings, then return a non-zero error code (or zero if there was no warning). Of course, if there are actual errors, it will still fail (at least, that is what the help on `sphinx-build` says).

dcherian:
Ok sounds good then

xarray/ufuncs.py
```diff
@@ -78,10 +79,49 @@ def __call__(self, *args, **kwargs):
         return res
 
 
+def _skip_signature(doc):
+    if not isinstance(doc, str):
+        return doc
+
+    if doc.startswith(name):
```
max-sixty: Where's `name` from? (Am I missing something very obvious?)

keewis: no, you're not, that's a bug

keewis:
there are three warnings (locally for me five) of the same kind left before we could switch on `-W --keep-going`, but I don't know how to fix these:
autosummary (or autodoc?) seems to put labels for name, values, attrs (locally also coords and dims) in both `xarray.DataArray.rst` and `xarray.DataArray.name` (or similar). The warnings recommends to put `:noindex:` in one of these, but both are generated so I don't know what to edit. Does anyone know more?

dcherian:
:man_shrugging: I think it's fine to merge right now. This is great work!

max-sixty:
It would be awesome to be able to enable that @keewis, so we don't regress in future. We can merge this for now and come back to that? 

I may have some time later this week to look more directly.

Great work, again!

keewis:
Enabling is blocked by these warnings, so we can only do that once we fixed that issue. I would be fine with a merge once we fix these warnings.

dcherian:
I'm going to merge so we can get the docstring improvements out. Let's have a follow-up PR to fix the remaining warnings.

Thanks for the great work @keewis  🍻

keewis:
okay, go ahead

max-sixty:
@keewis do you have a twitter username and first name? We're going to start calling out big new contributors on our twitter and release notes. (@keewis is also OK if you prefer!)

keewis:
no, sorry, I don't have a twitter account. Can you use the github handle?

max-sixty:
Done! https://twitter.com/xarray_dev/status/1196963951340392448?s=20

(And in case others see this: we're going to be calling out a few others who have made big contributions recently too; we're really excited and appreciative of those who've recently made such prolific contributions)

PR #3519, opened 2019-11-13
propagate indexes in to_dataset, from_dataset

dcherian:
happy to make changes!

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -61,9 +61,9 @@ from .coordinates import (
 from .duck_array_ops import datetime_to_numeric
 from .indexes import (
     Indexes,
-    copy_indexes,
     default_indexes,
     isel_variable_and_index,
+    propagate_indexes,
     roll_index,
 )
 from .merge import (
@@ -4357,7 +4357,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
 
         coords = dict(self.coords)
         coords[dim] = list(self.data_vars)
-        indexes = copy_indexes(self._indexes)
+        indexes = propagate_indexes(self._indexes)
 
         dims = (dim,) + broadcast_vars[0].dims
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -59,7 +59,13 @@ from .coordinates import (
     remap_label_indexers,
 )
 from .duck_array_ops import datetime_to_numeric
-from .indexes import Indexes, default_indexes, isel_variable_and_index, roll_index
+from .indexes import (
+    Indexes,
+    copy_indexes,
+    default_indexes,
+    isel_variable_and_index,
+    roll_index,
+)
 from .merge import (
     dataset_merge_method,
     dataset_update_method,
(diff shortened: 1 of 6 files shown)


xarray/core/indexes.py
```diff
@@ -5,11 +5,12 @@
 import pandas as pd
 
 from . import formatting
+from .utils import is_scalar
 from .variable import Variable
 
 
 class Indexes(collections.abc.Mapping):
-    """Immutable proxy for Dataset or DataArrary indexes."""
+    """Mutable proxy for Dataset or DataArrary indexes."""
```
dcherian:
```suggestion
    """Immutable proxy for Dataset or DataArrary indexes."""
```

xarray/core/indexes.py
```diff
@@ -35,9 +36,6 @@ def __contains__(self, key):
     def __getitem__(self, key):
         return self._indexes[key]
 
-    def __delitem__(self, key):
```
dcherian: undoes my previous change.

 I think I understand the indexes model better now. We only use `._indexes` dict internally and `Indexes` is intended for external use. Is that right?

xarray/core/indexes.py
```diff
@@ -100,3 +98,24 @@ def roll_index(index: pd.Index, count: int, axis: int = 0) -> pd.Index:
         return index[-count:].append(index[:-count])
     else:
         return index[:]
+
+
+def copy_indexes(
```
dcherian: Should this be a DataArray & Dataset method or just a utility function?

max-sixty: Forgive me if this is naive (I haven't been following this as closely as others have), but why do we need to copy the indexes? They're immutable and (IIUC) the hash tables are not copied, so copying may mean we need to recreate the hash table after each copy.

xarray/core/dataset.py
```diff
@@ -862,8 +868,12 @@ def _construct_direct(
...
 
     @classmethod
-    def _from_vars_and_coord_names(cls, variables, coord_names, attrs=None):
-        return cls._construct_direct(variables, coord_names, attrs=attrs)
+    def _from_vars_and_coord_names(
+        cls, variables, coord_names, indexes=None, attrs=None
+    ):
+        return cls._construct_direct(
+            variables, coord_names, indexes=indexes, attrs=attrs
+        )
```
max-sixty: I realize this isn't part of this PR, so maybe for another day; does calling `_from_vars_and_coord_names` do anything different from calling `_construct_direct`?

xarray/core/indexes.py
dcherian: No you're not naive! I just don't know what I'm doing :)

I've removed the `.copy` below, so we are just moving the actual Index around. Basically I needed a utility function that could create new `_indexes` dictionaries excluding certain dimensions as needed and that accounted for `self._indexes = None`

What do you think would be a better name for this function? `propagate_indexes`?

max-sixty: That makes two of us! :)

`propagate_indexes` is great, I think

max-sixty: There's one issue about not copying indexes—while the index's values are immutable, its attributes (e.g. its name) aren't. As long as every other downstream operation doesn't mutate the index (for example it uses `idx.rename("foo")` rather than `idx.name = "foo"`, then we're fine. But potentially that's not robust.

@shoyer let us know if you have guidance!

dcherian: :+1: The current version does this:

```
new_indexes = {k: v for k, v in indexes.items() if k not in exclude}
```
which should be fine?

shoyer:
Yes, looks good to me!

PR #3523, opened 2019-11-13
FUNDING.yml

Dr-G:
add NumFOCUS github sponsors button (recurring donations only)

This feature launched today at GitHub Universe!

Also add the custom link to point to the donation form for xarray.

cc @shoyer


shoyer:
Thank you!

PR #3526, opened 2019-11-13
Allow nested dictionaries in the Zarr backend (#3517)

### LINKED ISSUE #3517: Support for nested dictionaries in atttributes for Zarr
Similar to #2868, but for Zarr. In this case Zarr should support saving nested dictionaries.

```
import xarray as xr
ds = xr.Dataset()
ds.attrs['foo'] = {'bar': 1}
ds.to_zarr('foo.zarr', mode='w')
```

Error message:

```
<ipython-input-11-4e7ad4f0a599> in <module>
----> 1 ds.to_zarr('ll.zarr', mode='w')

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/dataset.py in to_zarr(self, store, mode, synchronizer, group, encoding, compute, consolidated, append_dim)
   1614             compute=compute,
   1615             consolidated=consolidated,
-> 1616             append_dim=append_dim,
   1617         )
   1618 

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/backends/api.py in to_zarr(dataset, store, mode, synchronizer, group, encoding, compute, consolidated, append_dim)
   1299     # validate Dataset keys, DataArray names, and attr keys/values
   1300     _validate_dataset_names(dataset)
-> 1301     _validate_attrs(dataset)
   1302 
   1303     if mode == "a":

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/backends/api.py in _validate_attrs(dataset)
    209     # Check attrs on the dataset itself
    210     for k, v in dataset.attrs.items():
--> 211         check_attr(k, v)
    212 
    213     # Check attrs on each variable within the dataset

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/backends/api.py in check_attr(name, value)
    204                 "a string, an ndarray or a list/tuple of "
    205                 "numbers/strings for serialization to netCDF "
--> 206                 "files".format(value)
    207             )
    208 

TypeError: Invalid value for attr: {'bar': 1} must be a number, a string, an ndarray or a list/tuple of numbers/strings for serialization to netCDF files
```

Additionally the error message mentions netCDF -- even if I am writing a Zarr file.

Workaround is to save as a list, i.e.

```
ds.attrs['foo'] = [{'bar': 1}]
```

eddienko:
Closes #3517.

I have tried to touch the minimum code as possible while leaving the defaults as they are now. The code works now by redefining to value to save as a list containing the dictionary. However when writing to disk the the dictionary is saved correctly, without the list (I do not understand 100% how it works but it does so someone double check!).

```
$ cat test.zarr/.zattrs
{
    "foo": {
        "bar": {
            "val": 3
        }
    }
}
```

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -179,7 +179,7 @@ def _validate_dataset_names(dataset):
         check_name(k)
 
 
-def _validate_attrs(dataset):
+def _validate_attrs(dataset, backend="netCDF"):
     """`attrs` must have a string key and a value which is either: a number,
     a string, an ndarray or a list/tuple of numbers/strings.
     """
@@ -197,6 +197,9 @@ def _validate_attrs(dataset):
                 "Invalid name for attr: {} must be a string for "
                 "serialization to netCDF files".format(name)
             )
+
+        if isinstance(value, dict) and ("zarr" in backend):
+            value = [value]
 
         if not isinstance(value, (str, Number, np.ndarray, np.number, list, tuple)):
             raise TypeError(
@@ -1299,7 +1302,7 @@ def to_zarr(
 
     # validate Dataset keys, DataArray names, and attr keys/values
     _validate_dataset_names(dataset)
-    _validate_attrs(dataset)
+    _validate_attrs(dataset, backend="zarr")
 
     if mode == "a":
         _validate_datatypes_for_zarr_append(dataset)
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -96,6 +96,8 @@ Bug fixes
   (:issue:`3402`). By `Deepak Cherian <https://github.com/dcherian/>`_
 - Allow appending datetime and bool data variables to zarr stores.
   (:issue:`3480`). By `Akihiro Matsukawa <https://github.com/amatsukawa/>`_.
+- Allow nested dictionaries in zarr attributes.
+  (:issue:`3517`). By `Eduardo Gonzalez <https://github.com/eddienko>`_.
 
 Documentation
 ~~~~~~~~~~~~~
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -1754,6 +1754,13 @@ class ZarrBase(CFEncodedBase):
     def test_dataset_caching(self):
         super().test_dataset_caching()
 
+    def test_nested_dictionary(self):
+        original = create_test_data()
+        original.attrs["foo"] = {"bar": {"baz": 3}}
+        with self.create_zarr_target() as store_target:
+            original.to_zarr(store_target, mode="w")
+            assert_identical(original, xr.open_zarr(store_target))
+
     def test_append_write(self):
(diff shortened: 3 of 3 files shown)


xarray/backends/api.py
```diff
@@ -198,6 +198,9 @@ def check_attr(name, value):
                 "serialization to netCDF files".format(name)
             )
 
+        if isinstance(value, dict) and ("zarr" in backend):
```
dcherian: Hmm.. This looks like we are inventing a convention to store nested dictionaries; is that right?

eddienko: I thought the same, but it is not the case. The nested dictionary is stored correctly in the .zattrs file, and by correctly I mean in the same way that using Zarr would store them. If we use Zarr directly to read the attributes I get a nested dictionary.

rabernat:
I am slammed right now with a proposal and can't really spend time on this, so I'll tag @jakirkham and @alimanfoo who can provide a zarr perspective on what is going on here.

jakirkham:
Yeah this probably works as these are just JSON files. That said, IDK that we are making any attempt to ensure this works. IOW I don't think this is tested or in the spec.

Additionally IDK that we do the same decoding on nested dictionaries as would be done on a flat dictionary. Meaning non-JSON values like `datetime64`/`timedelta64` might not be handled correctly in this case.

Could be wrong about these things. Those are just my immediate thoughts.

alimanfoo:
FWIW in the Zarr Python implementation I don't think we do any special encoding or decoding of attribute values. Whatever value is given then gets serialised using the built-in `json.dumps`. This means I believe that if someone provides a `dict` as an attribute value then that will get serialised as a JSON object, and get deserialised back to a `dict`, although this is not something we test for currently.

From the zarr v2 spec point of view I think anything goes in the `.zattrs` file, as long as `.zattrs` is a JSON object at the root.

Hth.

dcherian:
> Whatever value is given then gets serialised using the built-in json.dumps. This means I believe that if someone provides a dict as an attribute value then that will get serialised as a JSON object,

@eddienko I think this means we can avoid converting to a list?

eddienko:
mmm, I am not sure I understand. Currently

```
ds = xr.Dataset()
ds.attrs['foo'] = {'bar': 1}
ds.to_zarr('foo.zarr', mode='w')
```

does not work.  This is because dict is not allowed in the list of types that the attributes are validated against. My feeling is this should not produce an error.

This PR just addresses that so the code above works. The following works as well:

```
ds = xr.open_zarr('foo.zarr')
ds.attrs['foo']
{'bar': 1}
```

jhamman:
Bring back this old issue since I think there is value in getting something like this in. My suggestion would be to write a zarr-specific `_validate_attrs` function and use that, rather than stubbing in a special case for zarr in the generic one. @eddienko , I know this has sat for a long time but would you like to try to finish this up?

jhamman:
closing as stale. Folks should feel free to open a new PR to address the issue referenced above.

PR #3529, opened 2019-11-14
Deprecation warning in _replace_vars_and_dims

max-sixty:
& removed internal usages

I needed to pass `indexes=None` from some of the existing methods that were using the deprecated method. I spent some time debugging but couldn't work out the causality. I added TODOs as part of the index work.


xarray/core/dataset.py
```diff
@@ -962,6 +962,11 @@ def _replace_vars_and_dims(
         Unlike _replace_with_new_dims(), this method always recalculates
         indexes from variables.
         """
+        warnings.warn(
+            "`_replace_vars_and_dims` is deprecated in favor of `_replace_with_new_dims` and will be removed in 0.16",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
shoyer: Do we really need to deprecate this? It's an internal API, so I think we should feel free to remove it without warning.

max-sixty: Yeah, I wasn't sure, and thought it safer to deprecate, but I was probably being overly cautious. I'll remove.

max-sixty:
Updated!

max-sixty:
(test failure unrelated)

xarray/backends/zarr.py
```diff
@@ -656,4 +656,4 @@ def maybe_chunk(name, var, chunks):
             return var
 
     variables = {k: maybe_chunk(k, v, chunks) for k, v in ds.variables.items()}
-    return ds._replace_vars_and_dims(variables)
+    return ds._replace_with_new_dims(variables)
```
dcherian: can this just be `_replace`? I find it confusing to "replace with new dims" but not specify "new dims".

xarray/core/dataset.py
```diff
@@ -941,25 +941,6 @@ def _replace_with_new_dims(
...
-    ) -> "Dataset":
-        """Deprecated version of _replace_with_new_dims().
-
-        Unlike _replace_with_new_dims(), this method always recalculates
-        indexes from variables.
-        """
-        if dims is None:
-            dims = calculate_dimensions(variables)
-        return self._replace(
-            variables, coord_names, dims, attrs, indexes=None, inplace=inplace
```
dcherian: looks like this old function set `indexes=None` so your later changes make sense.

Also see docstring: 
`Unlike _replace_with_new_dims(), this method always recalculates indexes from variables.`

xarray/backends/zarr.py
max-sixty: Hmmm, I think in this case `dims` need to be recomputed, since `variables` changed?

shoyer:
Looking more carefully at this, I'm remembering now why I called this "deprecated." Eventually, we always want to keep track of `indexes` explicitly. Currently we don't do that, and in particular `_replace_vars_and_dims` does not do that.

My main reason for keeping it around was simply as a reminder that the functions that use it still need to be updated. It would be great to do that, but each usage should explicitly set `indexes` rather than setting `indexes=None` which we want to eventually stop doing.

Does that make sense?

xarray/core/dataset.py
```diff
@@ -3024,11 +3005,11 @@ def expand_dims(
...
-        new_dims = self._dims.copy()
-        new_dims.update(dim)
-
-        return self._replace_vars_and_dims(
-            variables, dims=new_dims, coord_names=coord_names
+        return self._replace_with_new_dims(
+            # TODO: fix indexes, I think by passing nothing (i.e. default)
+            variables=variables,
+            coord_names=coord_names,
+            indexes=None,
```
shoyer: I think the right way to set `indexes` here would be to copy them from the original dataset and then add/modify indexes as needed for any expanded dimensions.

Issue #3514, opened 2019-11-12
Should we cache some small properties?


max-sixty:
I was doing some profiling on `isel`, and see there are some properties that (I think) never change, but are called frequently. Should we cache these on their object? 

Pandas uses [cache_readonly](https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/properties.pyx#L7) for these cases.

Here's a case: we call [`LazilyOuterIndexedArray.shape`](https://github.com/pydata/xarray/blob/af28c6b02fac08494f5d9ae2718d68a084d93949/xarray/core/indexing.py#L535) frequently when doing a simple indexing operation. Each call takes ~150µs. An attribute lookup on a python object takes ~50ns (i.e. 3000x faster). IIUC the result on that property should never change.

I don't think this is the solution to performance issues, and there's some additional complexity. 
Could they be easy & small wins, though?

shoyer:
I think this would be totally fine to add. A variant on `cache_readonly` appears in Python 3.8 standard library as `functools.cached_property`. I think we could backport that pretty easily:
https://github.com/python/cpython/blob/d360346640e19231032b072216195484fa2450b4/Lib/functools.py#L924

max-sixty:
Great, didn't know about that.

It does remind me, though, that I'm not sure it's possible given we're using `__slots__`, at least without adding an entry to `__slots__`...
https://github.com/python/cpython/blob/d360346640e19231032b072216195484fa2450b4/Lib/functools.py#L954-L958

crusaderky:
By reading the implementation of cachedproperty, it needs a ``__dict__``. It should be straightforward to write a variant that uses slots though.

max-sixty:
> By reading the implementation of cachedproperty, it needs a `__dict__`. It should be straightforward to write a variant that uses slots though.

Is there a way of doing that which doesn't nullify the benefits of slots? i.e. if we add a `_cache_dict` to the class, is that the same overhead as having a `__dict__`?

crusaderky:
@max-sixty afraid so. But as I said it should be straightforward to use a variant that instead of ``self.__dict__[self.name]`` uses getattr/setattr.

crusaderky:
I prefer ``pandas.utils.cache_readonly`` to the Python 3.8 ``cachedproperty``. Besides not needing a dict, it is also written in cython, which should make a substantial difference.

[...]

PR #3531, opened 2019-11-14
Leave empty slot when not using accessors

crusaderky:
Save a few bytes and nanoseconds for the overwhelming majority of the users that don't use accessors.
Lay the groundwork for potential future use of ``@pandas.utils.cache_readonly``.

xref https://github.com/pydata/xarray/issues/3514


crusaderky:
Ready for review and merge

PR #3532, opened 2019-11-14
ensure rename does not change index type

### LINKED ISSUE #3522: CFTimeIndex changes to normal Index after renaming
#### MCVE Code Sample
Since this code crashes for me and I don't have xarray master, I can't create a tested MCVE Code sample at the moment, but I think something along the lines of:
```python
import xarray as xr
ds = xr.Dataset(coords={'time': xr.cftime_range(start='2000', periods=6, freq='2MS', calendar='noleap'), 'something': [0, 1, 2, 3]})
print(ds.indexes['time'])
ds = ds.rename({'something': 'something_else'})
print(ds.indexes['time'])
```

#### Expected Output

```
CFTimeIndex([2000-05-01 12:00:00, 2000-05-02 12:00:00,
             ...
             2001-04-01 12:00:00],
            dtype='object', name='time', length=425)
```
After renaming
```
Index([2000-05-01 12:00:00, 2000-05-02 12:00:00,
             ...
             2001-04-01 12:00:00],
            dtype='object', name='time', length=425)
```

#### Problem Description
CFTimeIndex changes to normal Index after renaming

<details>

xarray: 0.14.0
pandas: 0.25.2
numpy: 1.17.2
scipy: 1.3.1
netCDF4: 1.5.3
pydap: None
h5netcdf: 0.7.4
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.4.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: 1.1.0
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 2.6.0
distributed: 2.6.0
matplotlib: 3.1.1
cartopy: 0.17.0
seaborn: None
numbagg: None
setuptools: 41.4.0
pip: 19.3.1
conda: None
pytest: 5.2.1
IPython: 7.9.0
sphinx: None
</details>

mathause:
- [x] Closes #3522
 - [x] Tests added

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2667,8 +2667,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
                     verify_integrity=False,
                 )
             else:
-                index = v
-                index.name = new_name
+                index = v.rename(new_name)
             indexes[new_name] = index
         return indexes
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2667,7 +2667,8 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
                     verify_integrity=False,
                 )
             else:
-                index = pd.Index(v, name=new_name)
+                index = v
+                index.name = new_name
             indexes[new_name] = index
         return indexes
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -79,6 +79,10 @@ New Features
 
 Bug fixes
 ~~~~~~~~~
+- Ensure an index of type ``CFTimeIndex`` is not converted to a ``DatetimeIndex`` when 
+  calling :py:meth:`Dataset.rename` (also :py:meth:`Dataset.rename_dims`
+  and :py:meth:`xr.Dataset.rename_vars`). By `Mathias Hauser <https://github.com/mathause>`_
+  (:issue:`3522`).
 - Fix a bug in `set_index` in case that an existing dimension becomes a level variable of MultiIndex. (:pull:`3520`)
   By `Keisuke Fujii <https://github.com/fujiisoup>`_.
 - Harmonize `_FillValue`, `missing_value` during encoding and decoding steps. (:pull:`3502`)
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -2461,28 +2461,51 @@ class TestDataset:
             original.rename_vars(names_dict_bad)
 
     @requires_cftime
-    def test_rename_does_not_change_index_type(self):
+    def test_rename_does_not_change_CFTimeIndex_type(self):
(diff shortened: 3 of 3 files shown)


xarray/core/dataset.py
```diff
@@ -2667,7 +2667,8 @@ def _rename_indexes(self, name_dict, dims_set):
                     verify_integrity=False,
                 )
             else:
-                index = pd.Index(v, name=new_name)
+                index = v
+                index.name = new_name
```
mathause: Alternatively there is a `index = v._shallow_copy(name=new_name)` method. It
> *[] is an internal non-public method*
in pandas. Should I use it anyway?

max-sixty: While I think it's fairly stable, you're right—best practice is indeed to avoid. 

I think the current code will mutate the existing index name, is that correct? We should at least test to ensure that doesn't happen. I think `.rename` rather than direct assignment will avoid this (and calls `_shallow_copy` from within pandas so we're safe)

FWIW I think we can replace https://github.com/pydata/xarray/pull/3532/files#diff-921db548d18a549f6381818ed08298c9R2662-R2667 also with `.rename`, but it's minor and doesn't need to be in this PR

mathause: Oh, there is even `rename` - I did not see that, will switch to it.

mathause:
> FWIW I think we can replace https://github.com/pydata/xarray/pull/3532/files#diff-921db548d18a549f6381818ed08298c9R2662-R2667 also with .rename, but it's minor and doesn't need to be in this PR

I have to look how this is done in the  `MultiIndex`.

mathause:
> I think the current code will mutate the existing index name, is that correct?

Currently the index is recreated with the same name (even if it does not change it's name). We could change the following lines

https://github.com/pydata/xarray/blob/40588dc38ddc2d573e3dc8c63b2e6533eb978656/xarray/core/dataset.py#L2649-L2651


``` python
 new_name = name_dict.get(k, None) 
 if new_name not in dims_set or new_name is None: 
     continue
```
to skip doing this.

mathause:
No, this probably does not work if only one of the names of a MultiIndex changes...

max-sixty:
> Currently the index is recreated with the same name (even if it does not change it's name). We could change the following lines

I was unclear when I wrote "current code"; I was referring to proposed code in this PR. I don't think the code on master mutates the existing index, because it's making a copy. But I think this does: 

```python
                index = v
                index.name = new_name
```

mathause:
Thanks, yes my approach was indeed wrong. I updated the code and the tests. I was confused for a while because `orig.rename_dims(time="time_new")` drops `indexes` while `orig.rename_dims()` does not.

mathause:
Fine for me. I think `MultiIndex` could be treated the same. However, it does conserve `CFTimeIndex` already, and the test would require a bit of thinking, so let's leave that for now.

max-sixty:
Perfect, thanks again @mathause

Issue #3382, opened 2019-10-08
Improve indexing performance benchmarks


crusaderky:
As discussed in #3375 - FYI @jhamman 

``asv_bench/benchmarks/indexing.py`` is currently missing some key use cases:

- All tests the above module use arrays with 2~6 million points.
While this is important to spot any case where the numpy underlying functions start being unnecessarily called more than once, it also means any performance improvement or degradation in any of the pure-Python code will be completely drowned out.
All tests should be run twice, once with the current ``nx = 3000; ny = 2000; nt = 1000`` and again with ``nx = 15; ny = 10; nt = 5``.
- DataArray slicing (isel and square brackets)
- Slicing when there are no IndexVariables (verify that we're not creating dummy variables, doing a full scan on them, and then discarding them)
- DataArray.sel and DataArray.isel
- other?

PR #3533, opened 2019-11-14
WIP: Speed up isel in most cases

crusaderky:
Yet another major improvement for #2799.

Achieve a 2x to 5x boost in isel performance when slicing by int, list of int, scalar ndarray, or 1-dimensional ndarray.

```python
import xarray

da = xarray.DataArray([[1, 2], [3, 4]], dims=['x', 'y'])
v = da.variable
a = da.variable.values
ds = da.to_dataset(name="d")

ds_with_idx = xarray.Dataset({'x': [10, 20], 'y': [100, 200], 'd': (('x', 'y'), [[1, 2], [3, 4]])})
da_with_idx = ds_with_idx.d

# before -> after
%timeit a[0] # 121 ns
%timeit v[0] # 7 µs
%timeit v.isel(x=0) # 10 µs
%timeit da[0]  # 65 µs -> 15 µs
%timeit da.isel(x=0)  # 63 µs -> 13 µs
%timeit ds.isel(x=0)  # 48 µs -> 24 µs
%timeit da_with_idx[0]  # 209 µs -> 82 µs
%timeit da_with_idx.isel(x=0, drop=False)  # 135 µs -> 34 µs
%timeit da_with_idx.isel(x=0, drop=True)  # 101 µs -> 34 µs
%timeit ds_with_idx.isel(x=0, drop=False)  # 90 µs -> 49 µs
%timeit ds_with_idx.isel(x=0, drop=True)  # 65 µs -> 49 µs
```

Marked as WIP because this commands running the asv suite to verify there are no regressions for large arrays.
(on a separate note, we really need to add the small size cases to asv - as discussed in #3382).

This profoundly alters one of the most important methods in xarray and I must confess it makes me nervous, particularly as I am unsure if the test coverage of  DataArray.isel() is as through as Dataset.isel().

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -51,6 +51,7 @@ from .coordinates import (
 from .dataset import Dataset, split_indexes
 from .formatting import format_item
 from .indexes import Indexes, default_indexes
+from .indexing import is_fancy_indexer
 from .merge import PANDAS_TYPES
 from .options import OPTIONS
 from .utils import Default, ReprObject, _check_inplace, _default, either_dict_or_kwargs
@@ -1012,8 +1013,27 @@ class DataArray(AbstractArray, DataWithCoords):
         DataArray.sel
         """
         indexers = either_dict_or_kwargs(indexers, indexers_kwargs, "isel")
-        ds = self._to_temp_dataset().isel(drop=drop, indexers=indexers)
-        return self._from_temp_dataset(ds)
+        if any(is_fancy_indexer(idx) for idx in indexers.values()):
+            ds = self._to_temp_dataset()._isel_fancy(indexers, drop=drop)
+            return self._from_temp_dataset(ds)
+
+        # Much faster algorithm for when all indexers are ints, slices, one-dimensional
+        # lists, or zero or one-dimensional np.ndarray's
+
+        data = self._variable.isel(indexers)
+
+        coords = {}
+        for coord_name, coord_value in self._coords.items():
+            coord_indexers = {
+                k: v for k, v in indexers.items() if k in coord_value.dims
+            }
+            if coord_indexers:
+                coord_value = coord_value.isel(coord_indexers)
+                if drop and coord_value.ndim == 0:
+                    continue
+            coords[coord_name] = coord_value
+
+        return DataArray(data=data, coords=coords, name=self.name, fastpath=True)
 
     def sel(
         self,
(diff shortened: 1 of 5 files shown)


crusaderky:
@shoyer I struggle to understand the meaning of ``Dataset.indexes``. How does it  differ from
```python
{k: v.to_index() for k, v in self._variables.items() if isinstance(v, IndexVariable)
```
?

xarray/core/dataarray.py
```diff
@@ -1012,8 +1013,27 @@ def isel(
...
+            coord_indexers = {
+                k: v for k, v in indexers.items() if k in coord_value.dims
+            }
+            if coord_indexers:
+                coord_value = coord_value.isel(coord_indexers)
+                if drop and coord_value.ndim == 0:
+                    continue
+            coords[coord_name] = coord_value
+
+        return DataArray(data=data, coords=coords, name=self.name, fastpath=True)
```
```suggestion
        return self._replace(data=data, coords=coords)
```
?
Then we retain any attrs or subtypes

crusaderky: Good point

shoyer:
Dataset.indexes is currently equivalent to just calling to_index() on each appropriate variable. But eventually we want to separate the notion of an index from variables with names matching dimensions, and this is a prerequisite for that.

crusaderky:
I would *very* much appreciate if everybody could play around with this PR in their projects and confirm that nothing breaks!

crusaderky:
Merging soon if nobody speaks up!

xarray/core/dataset.py
```diff
@@ -1907,6 +1908,51 @@ def isel(
...
+                    if indexes:
+                        indexes.pop(var_name, None)
+                    continue
+                if indexes and var_name in indexes:
+                    if var_value.ndim == 1:
+                        indexes[var_name] = var_value.to_index()
+                    else:
+                        del indexes[var_name]
+            variables[var_name] = var_value
+            dims.update(dict(zip(var_value.dims, var_value.shape)))
```
shoyer: I think this is making one more temporary dict than necessary:
```suggestion
            dims.update(zip(var_value.dims, var_value.shape))
```

xarray/core/dataset.py
```diff
@@ -1907,6 +1908,51 @@ def isel(
...
+
+        # Restore order
+        dims = {k: dims[k] for k in self._dims if k in dims}
```
shoyer: I'm not sure this is necessary. Right now, we always sort dimensions in Dataset.dims (though we have considered changing this going forward).

Even if we preserved dimension order, we generally don't go out of our way to preserve variable order in dicts. I think this is probably fine simply being omitted.

crusaderky:
Implemented @shoyer 's suggestions and aligned to master.
Merging into master tomorrow!

crusaderky:
In it goes \*ducks for cover\*

dcherian:
=) thanks @crusaderky

PR #3537, opened 2019-11-15
Numpy 1.18 support

### LINKED ISSUE #3409: test failure with upstream-dev

crusaderky:
Fix mean() and nanmean() for datetime64 arrays on numpy backend when upgrading from numpy 1.17 to 1.18.
All other nan-reductions on datetime64s were broken before and remain broken.
mean() on datetime64 and dask was broken before and remains broken.

 - [x] Closes #3409

### DIFF
--- a/xarray/core/duck_array_ops.py
+++ b/xarray/core/duck_array_ops.py
@@ -10,7 +10,6 @@ from functools import partial
 
 import numpy as np
 import pandas as pd
-from distutils.version import LooseVersion
 
 from . import dask_array_ops, dtypes, npcompat, nputils
 from .nputils import nanfirst, nanlast
--- a/xarray/core/duck_array_ops.py
+++ b/xarray/core/duck_array_ops.py
@@ -357,6 +357,7 @@ def _datetime_crude_nanmin(array):
 
     - can't accept an axis parameter
     - will return incorrect results if the array exclusively contains NaT
+    - doesn't work with dask!
     """
     if LooseVersion(np.__version__) < "1.18":
         # numpy.min < 1.18 incorrectly skips NaT - which we exploit here
@@ -371,10 +372,8 @@ def _datetime_crude_nanmin(array):
     array = array.ravel()
     array = array[~pandas_isnull(array)]
 
-    assert array.dtype.kind in "Mm"
     assert array.dtype.itemsize == 8
     initial = np.array(2 ** 63 - 1, dtype=array.dtype)
-
     return min(array, initial=initial, skipna=False)
 
 
--- a/xarray/core/duck_array_ops.py
+++ b/xarray/core/duck_array_ops.py
@@ -360,12 +360,6 @@ def _datetime_nanmin(array):
     - numpy nanmin() don't work on datetime64 (all versions at the moment of writing)
     - dask min() does not work on datetime64 (all versions at the moment of writing)
     """
-    from .dataarray import DataArray
-    from .variable import Variable
-
-    if isinstance(array, (DataArray, Variable)):
-        array = array.data
-
     assert array.dtype.kind in "mM"
     dtype = array.dtype
     # (NaT).astype(float) does not produce NaN...
--- a/xarray/core/duck_array_ops.py
+++ b/xarray/core/duck_array_ops.py
@@ -352,29 +352,30 @@ cumsum_1d.numeric_only = True
 _mean = _create_nan_agg_method("mean")
 
 
-def _datetime_crude_nanmin(array):
-    """Implement nanmin for datetime64 arrays, with caveats:
+def _datetime_nanmin(array):
+    """nanmin() function for datetime64.
 
-    - can't accept an axis parameter
-    - will return incorrect results if the array exclusively contains NaT
-    - doesn't work with dask!
+    Caveats that this function deals with:
+
+    - In numpy < 1.18, min() on datetime64 incorrectly ignores NaT
+    - numpy nanmin() don't work on datetime64 (all versions at the moment of writing)
+    - dask min() does not work on datetime64 (all versions at the moment of writing)
     """
-    if LooseVersion(np.__version__) < "1.18":
-        # numpy.min < 1.18 incorrectly skips NaT - which we exploit here
-        return min(array, skipna=False)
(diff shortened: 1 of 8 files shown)


crusaderky:
xref #3434

crusaderky:
Ready for review and merge

mathause:
Given https://github.com/numpy/numpy/pull/14841 - is this worth the extra complexity?

Currently xarray is only broken wrt numpy master and there is a chance https://github.com/numpy/numpy/pull/14841 gets into 1.18 - then we could just use `np.nanmin`?

Also - is this the only place of `min` for datetimes? Should this also go into the operator dispatcher?

dopplershift:
IMO, it's always best to release code that you *know* will work rather than rely on upstream to get something into the next release in order for you not to be broken. I say that both as a downstream user of xarray and a library maintainers.

dcherian:
LGTM. I think we should merge so we can get a release out.

crusaderky:
@dopplershift fully agree.

> is this the only place of min for datetimes? 

This PR *does not* fix the nan-reductions for datetime arrays with NaT.

It *very narrowly* fixes the use case of **mean() and nanmean() with a numpy backend**, so that they work on numpy 1.18 as they did on 1.17. As a matter of fact they never worked with dask, were never tested, and still won't work after the PR.

To emphasize: I am not introducing any new feature or fixing bugs on numpy 1.17; I'm just getting rid of regressions on the upgrade from 1.17 to 1.18.

Specifically, the PR fixes this failing test:
https://github.com/pydata/xarray/blob/aa876cfd6b3b97ee5028d089ec741d057e3ed688/xarray/tests/test_duck_array_ops.py#L276-L293

which was not well formulated - despite its name, it doesn't test all reductions, only mean(), and in fact never runs on dask (the index of a dask array is not chunked).
I have now rewritten it  for the sake of clarity.

*If* numpy/numpy#14841 makes into master before 1.18 is published, and *if* it actually delivers on making reductions work with NaT (I did not test it yet), I'll be happy to write a second PR and rework everything.
Even then, a switch of this sort will still be needed:

```python
def datetime_nanmin(array, **kwargs):
    if LooseVersion(numpy.__version__) < "1.18":
        return np.min(array, **kwargs)
    else:
        return np.nanmin(array, **kwargs)
```
The actual code will be in fact more complicated because you also need to think about dask.

max-sixty:
Agree—let's merge?

mathause:
Ok, thanks for the clarification.

crusaderky:
Updated PR description:

Fix mean() and nanmean() for datetime64 arrays on numpy backend when upgrading from numpy 1.17 to 1.18.
All other nan-reductions on datetime64s were broken before and remain broken.
All reductions on datetime64s, including mean() and nanmean(), were broken on dask and remain broken.

crusaderky:
I just realised there are no tests at all for ``datetime_to_numeric(offset=None) `` on a dask backend. I suspect I broke something - I need to look at it better.

crusaderky:
``datetime_to_numeric(offset=None)`` never worked with dask to begin with - there's a big fat TODO in it about it. I rewrote most of the PR; now mean() works with dask.

Pre-empting an observation: *yes*, the nanmin function I wrote *could* be used to work around the shortcomings of both numpy and dask on datetime64. I don't think xarray should do it and IMHO the workaround currently implemented for mean() should never have been done to begin with.

Again, this PR only fixes the regression on numpy 1.18.

crusaderky:
Ready for review and merge

xarray/core/duck_array_ops.py
```diff
@@ -351,6 +351,32 @@ def f(values, axis=None, skipna=None, **kwargs):
...
+    Caveats that this function deals with:
+
+    - In numpy < 1.18, min() on datetime64 incorrectly ignores NaT
+    - numpy nanmin() don't work on datetime64 (all versions at the moment of writing)
+    - dask min() does not work on datetime64 (all versions at the moment of writing)
+    """
+    from .dataarray import DataArray
+    from .variable import Variable
+
+    if isinstance(array, (DataArray, Variable)):
```
dcherian: does this ever happen? All the other functions in this file work at the numpy, dask level (I think) so maybe this shouldn't be necessary.

crusaderky: Yes, I was surprised myself. I didn't investigate, but I agree it's not clean.

crusaderky: Fixed it

PR #3542, opened 2019-11-16
sparse option to reindex and unstack

### LINKED ISSUE #3518: Have "unstack" return a boolean mask?
#### MCVE Code Sample
```python
arr = xr.DataArray(np.arange(6).reshape(2, 3),
                  coords=[('x', ['a', 'b']), ('y', [0, 1, 2])])
arr
stacked = arr.stack(z=('x', 'y'))
stacked[:4].unstack().dtype
```

#### Expected Output
```python
>>> arr = xr.DataArray(np.arange(6).reshape(2, 3),
...                  coords=[('x', ['a', 'b']), ('y', [0, 1, 2])])
>>> arr
<xarray.DataArray (x: 2, y: 3)>
array([[0, 1, 2],
       [3, 4, 5]])
Coordinates:
  * x        (x) <U1 'a' 'b'
  * y        (y) int64 0 1 2
>>> stacked = arr.stack(z=('x', 'y'))
>>> stacked[:4].unstack().dtype
dtype('float64')
```

#### Problem Description
Unstacking changes the data type to float for `NaN`'s.
Are there thoughts on alternative options, e.g. `fill_value=0` or `return_boolean_mask`, in order to retain the original data type?

Currently, I obtain a boolean `missing` array by checking for `isnan`.
Then I call `fillnan(0)` and convert the data type back to integer.
However, this is quite inefficient.

<details>
INSTALLED VERSIONS
------------------
commit: None
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 3.10.0-957.10.1.el7.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.1

xarray: 0.14.0
pandas: 0.25.1
numpy: 1.17.2
scipy: 1.3.1
netCDF4: 1.4.2
pydap: None
h5netcdf: 0.7.4
h5py: 2.9.0
Nio: None
zarr: 2.3.2
cftime: 1.0.3.4
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.5.2
distributed: 2.5.2
matplotlib: 3.1.1
cartopy: None
seaborn: 0.9.0
numbagg: None
setuptools: 41.4.0
pip: 19.2.3
conda: None
pytest: 5.0.1
IPython: 7.8.0
sphinx: None

</details>

fujiisoup:
- [x] Closes #3518
 - [x] Tests added

Added `sparse` option to `reindex` and `unstack`.
I just added a minimal set of codes necessary to `unstack` and `reindex`.

There is still a lot of space to complete the sparse support as discussed in #3245.

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2254,7 +2254,6 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         tolerance: Number = None,
         copy: bool = True,
         fill_value: Any = dtypes.NA,
-        sparse: bool = False,
         **indexers_kwargs: Any,
     ) -> "Dataset":
         """Conform this object onto a new set of indexes, filling in
@@ -2430,6 +2429,29 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         the original and desired indexes. If you do want to fill in the `NaN` values present in the
         original dataset, use the :py:meth:`~Dataset.fillna()` method.
 
+        """
+        return self._reindex(
+            indexers,
+            method,
+            tolerance,
+            copy,
+            fill_value,
+            sparse=False,
+            **indexers_kwargs,
+        )
+
+    def _reindex(
+        self,
+        indexers: Mapping[Hashable, Any] = None,
+        method: str = None,
+        tolerance: Number = None,
+        copy: bool = True,
+        fill_value: Any = dtypes.NA,
+        sparse: bool = False,
+        **indexers_kwargs: Any,
+    ) -> "Dataset":
+        """
+        same to _reindex but support sparse option
         """
         indexers = utils.either_dict_or_kwargs(indexers, indexers_kwargs, "reindex")
 
@@ -3345,7 +3367,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         if index.equals(full_idx):
             obj = self
         else:
-            obj = self.reindex(
+            obj = self._reindex(
                 {dim: full_idx}, copy=False, fill_value=fill_value, sparse=sparse
             )
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -2254,6 +2254,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         tolerance: Number = None,
         copy: bool = True,
         fill_value: Any = dtypes.NA,
+        sparse: bool = False,
         **indexers_kwargs: Any,
     ) -> "Dataset":
         """Conform this object onto a new set of indexes, filling in
@@ -2286,6 +2287,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
             the input. In either case, a new xarray object is always returned.
         fill_value : scalar, optional
             Value to use for newly missing values
+        sparse: use sparse-array. By default, False
         **indexers_kwarg : {dim: indexer, ...}, optional
             Keyword arguments in the same form as ``indexers``.
             One of indexers or indexers_kwargs must be provided.
@@ -2444,6 +2446,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
             tolerance,
             copy=copy,
             fill_value=fill_value,
+            sparse=sparse
         )
         coord_names = set(self._coord_names)
         coord_names.update(indexers)
@@ -3333,7 +3336,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
 
         return data_array
 
-    def _unstack_once(self, dim: Hashable, fill_value) -> "Dataset":
+    def _unstack_once(self, dim: Hashable, fill_value, sparse) -> "Dataset":
         index = self.get_index(dim)
         index = index.remove_unused_levels()
         full_idx = pd.MultiIndex.from_product(index.levels, names=index.names)
@@ -3342,7 +3345,9 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         if index.equals(full_idx):
             obj = self
         else:
-            obj = self.reindex({dim: full_idx}, copy=False, fill_value=fill_value)
+            obj = self.reindex(
+                {dim: full_idx}, copy=False, fill_value=fill_value, sparse=sparse
+            )
 
         new_dim_names = index.names
         new_dim_sizes = [lev.size for lev in index.levels]
@@ -3372,6 +3377,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         self,
         dim: Union[Hashable, Iterable[Hashable]] = None,
         fill_value: Any = dtypes.NA,
+        sparse: bool = False
     ) -> "Dataset":
         """
         Unstack existing dimensions corresponding to MultiIndexes into
@@ -3385,6 +3391,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
             Dimension(s) over which to unstack. By default unstacks all
             MultiIndexes.
         fill_value: value to be filled. By default, np.nan
+        sparse: use sparse-array if True
 
         Returns
         -------
@@ -3422,7 +3429,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
 
         result = self.copy(deep=False)
         for dim in dims:
-            result = result._unstack_once(dim, fill_value)
+            result = result._unstack_once(dim, fill_value, sparse)
         return result
 
     def update(self, other: "CoercibleMapping", inplace: bool = None) -> "Dataset":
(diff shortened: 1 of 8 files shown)


xarray/core/variable.py
```diff
@@ -993,6 +993,32 @@ def chunk(self, chunks=None, name=None, lock=False):
 
         return type(self)(self.dims, data, self._attrs, self._encoding, fastpath=True)
 
+    def _as_sparse(self, sparse_format=_default, fill_value=dtypes.NA):
```
fujiisoup: Currently, this is a private method.
Probably we can expose it to the public and add the same method to DataArray and Dataset as well in the future.

xarray/tests/test_variable.py
```diff
@@ -1862,6 +1863,17 @@ def test_getitem_with_mask_nd_indexer(self):
         )
 
 
+@requires_sparse
+class TestVariableWithSparse:
+    # TODO inherit VariableSubclassobjects to cover more tests
```
max-sixty: 👍

xarray/core/variable.py
```diff
@@ -993,6 +993,32 @@ def chunk(self, chunks=None, name=None, lock=False):
...
+        else:
+            dtype = self.dtype
+
+        if sparse_format is _default:
+            sparse_format = "coo"
+        as_sparse = getattr(sparse, "as_{}".format(sparse_format.lower()))
+        data = as_sparse(self.data.astype(dtype), fill_value=fill_value)
+        return self._replace(data=data)
+
+    def _to_dense(self):
```
fujiisoup: Also private, as is `_as_sparse`.

xarray/tests/test_dataset.py
```diff
@@ -2811,6 +2826,23 @@ def test_unstack_fill_value(self):
...
+        # make ds incomplete
+        ds = ds.isel(x=[0, 2, 3, 4]).set_index(index=["x", "y"])
+        # test fill_value
+        actual = ds.unstack("index", sparse=True)
+        expected = ds.unstack("index")
+        assert actual["var"].variable._to_dense().equals(expected["var"].variable)
+
+        actual = ds["var"].unstack("index", sparse=True)
+        expected = ds["var"].unstack("index")
+        assert actual.variable._to_dense().equals(expected.variable)
```
max-sixty: Do we test whether `actual.variable` is actually sparse?

xarray/tests/test_dataset.py
```diff
@@ -1748,6 +1748,23 @@ def test_reindex(self):
...
         assert_identical(expected, actual)
 
+    @requires_sparse
+    def test_reindex_sparse(self):
+        data = create_test_data()
+        dim3 = list("abdeghijk")
+        actual = data.reindex(dim3=dim3, sparse=True)
+        expected = data.reindex(dim3=dim3, sparse=False)
+        for k, v in data.data_vars.items():
+            np.testing.assert_equal(actual[k].data.todense(), expected[k].data)
```
fujiisoup: Currently, `assert_equal` cannot be used as we need to explicitly densify the array for the comparison.

xarray/tests/test_dataset.py
```diff
@@ -1748,6 +1748,23 @@ def test_reindex(self):
...
+        assert actual['var3'].data.density < 1.0
+
+        data["var3"] = data["var3"].astype(int)
+        actual = data.reindex(dim3=dim3, sparse=True, fill_value=-10)
+        expected = data.reindex(dim3=dim3, sparse=False, fill_value=-10)
+        for k, v in data.data_vars.items():
+            np.testing.assert_equal(actual[k].data.todense(), expected[k].data)
```
max-sixty: I think we generally use `assert_array_equal` for numpy arrays (but I can't immediately recall the difference...)

xarray/core/variable.py
```diff
@@ -993,6 +993,32 @@ def chunk(self, chunks=None, name=None, lock=False):
...
+        """
+        use sparse-array as backend.
+        """
+        import sparse
+
+        # TODO  what to do if dask-backended?
+        if fill_value is dtypes.NA:
+            dtype, fill_value = dtypes.maybe_promote(self.dtype)
+        else:
+            dtype = self.dtype
```
shoyer: I think it would be a nice idea to use result_type here, in case the fill_value and array have different types.
```suggestion
            dtype = dtypes.result_type(self.dtype, fill_value)
```

xarray/tests/test_dataset.py
shoyer: I think these actually end up doing the exact same checks.

xarray/core/variable.py
```diff
@@ -993,6 +993,32 @@ def chunk(self, chunks=None, name=None, lock=False):
...
+
+        if sparse_format is _default:
+            sparse_format = "coo"
+        as_sparse = getattr(sparse, "as_{}".format(sparse_format.lower()))
```
shoyer: I like the idea of not hard-coding supported sparse formats, but I wonder if we could be a little more careful here if AttributeError is raised. We should probably catch and re-raise Attribute error with a more informative message if this fails.

Otherwise, I expect we might see bug reports from confused users, e.g., when `sparse_format='csr'` raises a confusing message.

xarray/core/duck_array_ops.py
```diff
@@ -251,6 +253,9 @@ def count(data, axis=None):
 
 def where(condition, x, y):
     """Three argument where() with better dtype promotion rules."""
+    # sparse support
+    if isinstance(x, sparse_array_type) or isinstance(y, sparse_array_type):
```
shoyer: I am a little surprised this is necessary. Does `sparse` not support `__array_function__` for `np.where`?

xarray/core/variable.py
```diff
@@ -993,6 +993,32 @@ def chunk(self, chunks=None, name=None, lock=False):
 
         return type(self)(self.dims, data, self._attrs, self._encoding, fastpath=True)
 
+    def _as_sparse(self, sparse_format=_default, fill_value=dtypes.NA):
+        """
+        use sparse-array as backend.
+        """
+        import sparse
+
+        # TODO  what to do if dask-backended?
```
shoyer: Hopefully sparse will raise an error if you try to convert a dask array into a sparse array! If not, we should do that ourselves.

Long term, the best solution would be to convert a dask array from dense chunks to sparse chunks.

xarray/core/dataset.py
```diff
@@ -2254,6 +2254,7 @@ def reindex(
         tolerance: Number = None,
         copy: bool = True,
         fill_value: Any = dtypes.NA,
+        sparse: bool = False,
```
shoyer: This is a very clean solution, but I'm not sure it is a good idea to add a `sparse` argument to `reindex`. Can we think of independent uses for sparse reindexing, outside of `unstack`? If not, let's try to avoid expanding the user facing API of reindex. Maybe we could only expose this method in a private `_reindex` function, or convert to sparse arrays inside `_unstack_once`?

My general perspective is that it makes the most sense to add explicit `sparse` arguments on methods that could easily greatly enlarge the size of arrays (like `unstack`). In other cases, it would be better to add an explicit method for converting all data variables into sparse arrays (e.g., `to_sparse()`) which could be composed with existing methods like reindex, rather than adding `sparse=True` options to everything.

xarray/tests/test_dataset.py
fujiisoup: `values` property does not work for the sparse-backed Variable, resulting in the failure of `assert_array_equal`.
I'll add TODO comment for this.

xarray/core/duck_array_ops.py
fujiisoup: Well, yes. sparse looks not working with `np.result_type` and `astype(copy=False)`.
I'll add a TODO here.

xarray/core/variable.py
dcherian: We should make these public in `DataArray` and `Dataset`. See discussion here: https://github.com/pydata/xarray/issues/3245#issuecomment-524407909

Can be left for a future PR though :)

xarray/core/duck_array_ops.py
shoyer: do you have the latest version of sparse installed?

when I test this on my machine, it works:
```
In [13]: import sparse

In [14]: import numpy as np

In [15]: import xarray

In [16]: x = sparse.COO(np.arange(3))

In [17]: xarray.core.duck_array_ops.where(x > 1, x, x)
Out[17]: <COO: shape=(3,), dtype=int64, nnz=2, fill_value=0>
```

fujiisoup: Thanks. You are right. 
I was running with sparse 0.7.0. With 0.8.0, it is running.

xarray/core/variable.py
fujiisoup: Thanks, @dcherian. 
I would like to expose them to public, but what is the best name of these functions? 
#3245

xarray/core/dataset.py
fujiisoup: Done. Now, there is no sparse option in `reindex`.

shoyer:
Thank you @fujiisoup !

I think it could be a little cleaner to entirely avoid `sparse` inside any of the `reindex` functions, but this is fine for now.

PR #3550, opened 2019-11-20
cov() and corr() - finalization

r-beer:
Dear maintainers, 

this is my first PR and contribution to xarray. I have read the [contributing guidelines](xarray.pydata.org/en/stable/contributing.html) and did my best to conform to your [code of conduct](). However, I would be happy if you have a closer look whether everything is correct and give feedback for improvement!

As I don't have push access to @hrishikeshac's fork, I created this PR to add the final changes for PR #2652.

 - [x] Closes #2652
 - [x] Tests had been added in #2652
   - [x] black
   - [x] flake8
   - [ ] mypy
   - [x]  `whats-new.rst`
   - [x] `api.rst`
- [x] example notebook for usage of `corr()` and `cov()`
  - [x] where and how to write the examples?
  - [x] covariance
  - [x] correlation
- [x] moved implementation of cov() and corr() to computation
- [x] add test_cov()
- [x] remove da.cov and da.corr


r-beer:
@max-sixty, all three checks fail because `test_open_mfdataset_manyfiles` requires either netcdf4 or scipy.

Shall I, therefore, add the `@requires_scipy_or_netCDF4` decorator?

keewis:
There seems to be a problem with your email address (github can't match it to your account). Does it match the one you used for signing up on github?

r-beer:
> 
> 
> There seems to be a problem with your email address (github can't match it to your account). Does it match the one you used for signing up on github?

Dear @keewis, I just set my real email address also as public email address and set my local git user name and email address accordingly. Does this solve the problem?

keewis:
for the new commit, yes. The older commits could be rewritten if you don't like the other email address being public (I guide you through the process if you want) but since PRs get squashed after the merge you don't have to.

r-beer:
> 
> 
> for the new commit, yes. The older commits could be rewritten if you don't like the other email address being public (I guide you through the process if you want) but since PRs get squashed after the merge you don't have to.

OK, no need for the extra effort at this point, the squashing at the end of PR is fine for me. Thank you for the offer, anyway! 🙂 

Are any other changes necessary for the PR?

ideas:
 - examples for corr and cov usage
 - whats-new.rst

keewis:
yes, `whats-new.rst` and `api.rst` will have to eventually be updated before the merge. Also, examples would be really good.

keewis:
since the APIs of `DataArray` and `Dataset` are usually kept in sync as much as possible, it might be good to add the new methods to `Dataset`, too. This doesn't have to be in this PR, though, and I'd wait for confirmation first.

r-beer:
> 
> 
> since the APIs of `DataArray` and `Dataset` are usually kept in sync as much as possible, it might be good to add the new methods to `Dataset`, too. This doesn't have to be in this PR, though, and I'd wait for confirmation first.

Yes, the functionality that I would be very interested in would be to calculate `corr()` also between different variables of a dataset as in [pandas.DataFrame.corr](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.corr.html) as well as correlations between different locations within the same `DataArray` (e.g. correlations of time series with different latitudes).

xarray/tests/test_dataarray.py
```diff
@@ -4079,6 +4080,49 @@ def test_rank(self):
...
+            valid_values = ts1.notnull() & ts2.notnull()
+
+            ts1 = ts1.where(valid_values, drop=True)
+            ts2 = ts2.where(valid_values, drop=True)
+
+            return ts1.to_series().corr(ts2.to_series())
+
+        expected = pd_corr(ts1, ts2)
+        actual = ts1.corr(ts2)
+        np.allclose(expected, actual)
```
max-sixty: You can use xarray's `assert_allclose` and you get some additional checks for free

xarray/core/dataarray.py
```diff
@@ -3182,6 +3183,64 @@ def map_blocks(
...
+        valid_values = self.notnull() & other.notnull()
+        self = self.where(valid_values, drop=True)
+        other = other.where(valid_values, drop=True)
+        valid_count = valid_values.sum(dim)
+
+        # 3. Compute mean and standard deviation along the given dim
+        demeaned_self = self - self.mean(dim=dim)
+        demeaned_other = other - other.mean(dim=dim)
+
+        # 4. Compute  covariance along the given dim
```
```suggestion
        # 4. Compute covariance along the given dim
```

xarray/tests/test_dataarray.py
```diff
@@ -4079,6 +4080,49 @@ def test_rank(self):
         y = DataArray([0.75, 0.25, np.nan, 0.5, 1.0], dims=("z",))
         assert_equal(y.rank("z", pct=True), y)
 
+    # TODO: use fixture or randomly created array instead of load_dataset
```
max-sixty: 👍

xarray/tests/test_dataarray.py
```diff
@@ -4079,6 +4080,49 @@ def test_rank(self):
...
+    @requires_scipy_or_netCDF4
```
max-sixty: Why do we need scipy here? Because of the dataset loading?

max-sixty:
Great start! 

Thanks for the review @keewis 

I think it's fine to leave `Dataset` for another PR; it'd be great to get this in, and given it's your first PR (welcome!) let's not bloat it

Test looks good. Let's make those adjustments you put in the `TODO` and add one for the other function? We could also add some variants which test along different dimensions / all NaNs / other special cases.

xarray/tests/test_dataarray.py
r-beer: Yes, exactly. Until now, I was just updating the code and trying to get all tests passed. Therefore, I also added the TODO, because I think loading the dataset here is not necessary, random data or similar would be more elegant, wouldn't it?

xarray/tests/test_dataarray.py
keewis: I think this (or using `assert np.allclose(...)`) actually *has* to be done, `np.allclose` only compares its arguments and returns a boolean: the tests would still pass even if `expected` and `actual` have different values.

r-beer:
> 
> 
> Great start!
> 
> Thanks for the review @keewis
> 
> I think it's fine to leave `Dataset` for another PR; it'd be great to get this in, and given it's your first PR (welcome!) let's not bloat it
> 
> Test looks good. Let's make those adjustments you put in the `TODO` and add one for the other function? We could also add some variants which test along different dimensions / all NaNs / other special cases.

Thank you two, @max-sixty and @keewis, for the constructive feedback that made the start easier! 
I would also be happy to go once through the whole process and then implement the PR for the ds separately. This will probably cut the learning curve and accelerate the process.

Additionally, I wonder whether I have to configure black for the spaces between colons?

I had autoformatting running with autopep8 previously but since one of the last VS code updates something broke so I will set this up properly again on Friday or tomorrow.

xarray/tests/test_dataarray.py
r-beer: So, I mark this as resolved, please correct me if I am wrong!

r-beer:
> Test looks good. Let's make those adjustments you put in the `TODO` and add one for the other function?

Do you mean to add a TODO to the function mentioned in this commit: https://github.com/pydata/xarray/pull/3550/commits/d57eb7cfaf1407179080f0ca95e7951e5622b493?

Here, I think the `@requires_scipy_or_netCDF4` is actually necessary:
```python
@requires_scipy_or_netCDF4
@requires_dask
@pytest.mark.filterwarnings("ignore:use make_scale(name) instead")
def test_open_mfdataset_manyfiles(
```

xarray/tests/test_dataarray.py
keewis: I don't think this is resolved yet: `np.allclose` is a function that compares its inputs, in this case `expected` and `actual`  and returns a boolean (like `==`, but using a threshold to compensate numerical errors). As we don't do anything with that boolean, it goes out of scope and the tests won't be able to detect bugs in the code they should be testing.

So the minimal fix I offered as an alternative is
```suggestion
        assert np.allclose(expected, actual)
```
which would raise an `AssertionError` if `np.allclose` returns `False`, allowing `pytest` to detect that the test failed. This works, but I think in this case is not enough: `actual` is a `DataArray` (which is consistent with other `xarray` operations) whereas `expected` is a `np.float64`.

What @max-sixty suggested would be
```suggestion
        assert_allclose(expected, actual)
```
`assert_allclose` expects `expected` and `actual` to be the same type of `xarray` objects (i.e. `DataArray` or `Dataset` objects) and compares like `np.allclose`, but raises an `AssertionError` if the comparison would return `False` (along with other checks and providing a nice error report). That means that the suggested change won't be enough, we need to wrap the result of `pandas.Series.corr` (a `np.float64` instance) into a `DataArray` while making sure it has the correct coords.

xarray/core/dataarray.py
```diff
@@ -3182,6 +3183,64 @@ def map_blocks(
...
+        return cov
+
+    def corr(self, other, dim=None):
+        """Compute correlation between two DataArray objects along a shared dimension.
+
+        Parameters
+        ----------
+        other: DataArray
+            The other array with which the correlation will be computed
+        dim: The dimension along which the correlation will be computed
```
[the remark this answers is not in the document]
keewis: this seems to have been lost somehow, can you add it?

r-beer:
Concerning the examples, I have started to create a `Jupyter Notebook`  in the examples folder. 
Is this the right way to create the examples?

I have seen that the existing examples contain html code. Why is that? Why not only markdown?

r-beer:
> We could also add some variants which test along different dimensions / all NaNs / other special cases.

@max-sixty, I have found the `da(request)` fixture. Do I correctly understand that the tests you mentioned above can be implemented using this fixture and parametrize decorators as in `@pytest.mark.parametrize("da", (1, 2), indirect=True)`?

keewis:
I'm not sure about the examples in the `examples` folder, I think those are real-world examples?

What I meant was examples like those in [`DataArray.integrate`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.integrate.html#xarray.Dataset.integrate) ([source](https://github.com/pydata/xarray/blob/6b70107ab3063187b663290538c0d5a4107dab6e/xarray/core/dataarray.py#L3063-L3110)) that simply demonstrate how to use that function.

r-beer:
> 
> 
> I'm not sure about the examples in the `examples` folder, I think those are real-world examples?
> 
> What I meant was examples like those in [`DataArray.integrate`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.integrate.html#xarray.Dataset.integrate) ([source](https://github.com/pydata/xarray/blob/6b70107ab3063187b663290538c0d5a4107dab6e/xarray/core/dataarray.py#L3063-L3110)) that simply demonstrate how to use that function.

OK! That simplifies the tasks quite a bit 🙂 

I will implement this and the special cases for the tests, I might add Jupyter Notebook examples later to the documentation, when the `dataset.corr` functionality is also implemented.

r-beer:
> @max-sixty, I have found the `da(request)` fixture. Do I correctly understand that the tests you mentioned above can be implemented using this fixture and parametrize decorators as in `@pytest.mark.parametrize("da", (1, 2), indirect=True)`?

I have tried to wrap my head around the elegant use of fixtures and parametrize but didn't get there yet. If you have some time to show a more elegant solution, I would be happy to learn! If not, I guess this should work, too.

Are the examples OK like this?

max-sixty:
One question we started discussing on the old PR which we should resolve ASAP given @r-beer 's work: do we want `xr.cov(da1, da2)` or `da1.cov(da2)`?

CC @fujiisoup @shoyer

max-sixty:
@r-beer your current test case looks fine, no stress on the fixtures!

shoyer:
> One question we started discussing on the old PR which we should resolve ASAP given @r-beer 's work: do we want `xr.cov(da1, da2)` or `da1.cov(da2)`?

In my opinion the function form is clearer, because `cor` and `corr` are symmetric in both arguments.

r-beer:
> 
> 
> @r-beer your current test case looks fine, no stress on the fixtures!

Alright! 

I am now looking into the generation of the documentation. However, it seems like the contributing guidelines are not up-to-date, as it's not possible to create the `xarray-docs` environment (see #3556).

max-sixty:
> I am now looking into the generation of the documentation. 

What do you mean by this? You shouldn't have to generate anything locally?

max-sixty:
@r-beer let's go with @shoyer 's suggestion to change this from a method to a function `xr.cov` & `xr.corr`. Is that OK?

r-beer:
> 
> 
> > I am now looking into the generation of the documentation.
> 
> What do you mean by this? You shouldn't have to generate anything locally?

Oh, ok. I thought I have to create `xarray-docs` to update `whats-new.rst` and `api.rst`?

max-sixty:
> Oh, ok. I thought I have to create `xarray-docs` to update `whats-new.rst` and `api.rst`?

You can just edit the files directly! No need to build all the docs. It should be much simpler than the work you've already done!

(lmk if the docs were unclear and we can update)

r-beer:
> 
> 
> @r-beer let's go with @shoyer 's suggestion to change this from a method to a function `xr.cov` & `xr.corr`. Is that OK?

Sure! I actually asked myself a similar software design question in another project. 

Shouldn't be too much effort I guess? I will tackle it tomorrow. 😴

r-beer:
For the sake of completeness (and to already learn the procedure) I just added some information to `whats-new.rst` and `api.rst` for the methods `da.cov` and `da.corr`. 

I would be happy for feedback! 

Tomorrow, I will rewrite the methods to functions and go through all the steps again. Good night from Germany!

doc/whats-new.rst
```diff
@@ -25,6 +25,8 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+- Added `~xarray.DataArray.cov` and `~xarray.DataArray.corr` (:pull:`3550`).
```
max-sixty: Good, though these should have `:py:meth:` before them (some other examples below)

r-beer:
> (lmk if the docs were unclear and we can update)

I think it already gives a good overview, but then it might be unclear how to proceed. 

Maybe it might make sense to add a subsection `How to contribute to the xarray documentation?` in which the different options are shortly described examplary:
```
1. change docstring, e.g. ....
2. update api.rst, e.g. ....
3. update whats-new.rst, e.g. ...
```
Then, underline that building the documentation locally is optional:

```
Optionally, you can build the docs locally as described in the next section.
```

doc/whats-new.rst
r-beer: Thanks! That's already enough?

max-sixty: What do you mean by already enough?

r-beer: If changes have been made only for `xarray.dataarray.cov` and `.corr`, the above-mentioned changes in `whats-new.rst` are sufficient?

r-beer:
> @r-beer let's go with @shoyer 's suggestion to change this from a method to a function `xr.cov` & `xr.corr`. Is that OK?

@shoyer, @max-sixty, @keewis, where should these functions be placed?

my first guesses:
- computation
- dataarray
-

r-beer:
PS: accidently closed and commented instead of commenting

Hoeze:
Awesome, thanks a lot @r-beer!

dcherian:
Computation? Where is `xr.dot` documented?

r-beer:
> 
> 
> Computation? Where is `xr.dot` documented?

`xr.dot` is implemented in computation, what do you mean by "documented"?

doc/whats-new.rst
max-sixty: Yes! You could maybe say what the functions do, but they're already fairly self-explanatory (and we'll need to change this a bit given the updated location)

r-beer: Alright 🙂 

Yes, I wanted to already understand it completely. Will update the merge request later today. With the moved functions.

dcherian:
> where should these functions be placed?

Sorry, I misunderstood this question. I thought you were asking about documentation but I see that you're asking about which `.py` file to stick these functions in. Let's do `computation` just like `xr.dot`.


PS: It looks like `xr.dot` is only present in the "API reference" and not in https://xarray.pydata.org/en/stable/computation.html. We should add `dot`, `cov` and `corr` to `computation.rst` in a future PR.

r-beer:
@dcherian, @max-sixty, @keewis, @Hoeze, this is it, is it?  

Let me know if there are some adjustments to be made. 🙂

And I still get `trailing whitespace errors`, despite having set up pre-commit to run black and flake8 automatically. Do I have to set some parameters to delete the trailing whitespaces?

r-beer:
PS: I personally would like to use cov and corr also with dask. Does it also make sense to put this in a future PR, together with the above-mentioned other improvements?

Further PRs:
- Add `dot`, `cov` and `corr` to `computation.rst`
   > PS: It looks like `xr.dot` is only present in the "API reference" and not in https://xarray.pydata.org/en/stable/computation.html. We should add `dot`, `cov` and `corr` to `computation.rst` in a future PR.
- Test (and if necessary implement) `dask` compatibility to `cov` and `corr`
- Improve documentation on `How to contribute to the xarray documentation?` 
   > I think it already gives a good overview, but then it might be unclear how to proceed.
   > 
   > Maybe it might make sense to add a subsection `How to contribute to the xarray documentation?` in which the different options are shortly described examplary:
   > 
   > ```
   > 1. change docstring, e.g. ....
   > 2. update api.rst, e.g. ....
   > 3. update whats-new.rst, e.g. ...
   > ```
   > 
   > Then, underline that building the documentation locally is optional:
   > 
   > ```
   > Optionally, you can build the docs locally as described in the next section.
   > ```

xarray/core/computation.py
```diff
@@ -1047,6 +1048,162 @@ def earth_mover_distance(first_samples,
...
+    * space    (space) <U2 'IA' 'IL' 'IN'
+    """
+
+    # 1. Broadcast the two arrays
+    da_a, da_b = broadcast(da_a, da_b)
+
+    # 2. Ignore the nans
+    valid_values = da_a.notnull() & da_b.notnull()
+    da_a = da_a.where(valid_values, drop=True)
+    da_b = da_b.where(valid_values, drop=True)
```
max-sixty: Could we add a `TODO` with a reference to @shoyer 's comment on the efficiency of `drop=True`, from the previous PR (lmk if you can't find it).

xarray/core/dataarray.py
```diff
@@ -2853,6 +2861,135 @@ def dot(
 
         return computation.dot(self, other, dims=dims)
 
+    def cov(self, other: "DataArray", dim=None) -> "DataArray":
```
max-sixty: We can remove these in favor of the `xr.cov` functions, i.e. we don't need both. Sorry if this was unclear before and that lost you some time

xarray/tests/test_computation.py
```diff
@@ -789,6 +813,45 @@ def func(x):
...
+        valid_values = ts1.notnull() & ts2.notnull()
+
+        ts1 = ts1.where(valid_values, drop=True)
+        ts2 = ts2.where(valid_values, drop=True)
+
+        return ts1.to_series().corr(ts2.to_series())
+
+    expected = pd_corr(ts1, ts2)
+    actual = xr.corr(ts1, ts2)
+    np.allclose(expected, actual)
```
max-sixty:
```suggestion
    assert np.allclose(expected, actual)
```
(and the others)

xarray/tests/test_computation.py
```diff
@@ -789,6 +813,45 @@ def func(x):
     assert_identical(expected, actual)
 
 
+def test_corr(da):
```
max-sixty: Can we do the same thing for `cov`? Copy and paste is OK (parameterizing is another option)

doc/whats-new.rst
```diff
@@ -25,6 +25,9 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+:py:func:`xarray.dot`, and :py:meth:`DataArray.dot`
```
max-sixty: Looks like an odd artifact here

r-beer:
> I'm not sure what's going on with the formatting. If you run ` black --version` and `black .` from the root of the repo, what do you get?

` black --version`:
```bash
black, version 19.10b0
```
`black .`:
```bash
reformatted ...\xarray\xarray\core\dataarray.py
All done! ✨ �🍰✨✨
1 file reformatted, 143 files left unchanged.
```

xarray/core/computation.py
r-beer: added it 🙂

xarray/core/dataarray.py
r-beer: OK, I just imitated the dot implementation enterily and thought that it could be a useful method.

Why making a difference between dot and cov? 

I invested some time but learned from it, so no time lost! 🙂

xarray/tests/test_computation.py
r-beer: Sure, I added a TODO asking exactly that question: 
```
# TODO: up to now, cov is only implicitly tested via corr(), is that sufficient?
def cov(da_a, da_b, dim=None):
```

So, we always want explicit testing, right?

r-beer:
> 
> 
> > I'm not sure what's going on with the formatting. If you run ` black --version` and `black .` from the root of the repo, what do you get?
> 
> ` black --version`:
> 
> ```shell
> black, version 19.10b0
> ```
> 
> `black .`:
> 
> ```shell
> reformatted ...\xarray\xarray\core\dataarray.py
> All done! ✨ �🍰✨✨
> 1 file reformatted, 143 files left unchanged.
> ```

Now, I additionally get errors from pre-commit claiming that `GIT: unstaged files detected`.

Guess, the pre-commit installation was not completely correct?

r-beer:
@dcherian, @max-sixty, @keewis, @Hoeze, second try: any more improvements required?

r-beer:
> 
> 
> PS: I personally would like to use cov and corr also with dask. Does it also make sense to put this in a future PR, together with the above-mentioned other improvements?
> 
> Further PRs:
> 
>     * Add `dot`, `cov` and `corr` to `computation.rst`
>       > PS: It looks like `xr.dot` is only present in the "API reference" and not in https://xarray.pydata.org/en/stable/computation.html. We should add `dot`, `cov` and `corr` to `computation.rst` in a future PR.
> 
>     * Test (and if necessary implement) `dask` compatibility to `cov` and `corr`
> 
>     * Improve documentation on `How to contribute to the xarray documentation?`
>       > I think it already gives a good overview, but then it might be unclear how to proceed.
>       > Maybe it might make sense to add a subsection `How to contribute to the xarray documentation?` in which the different options are shortly described examplary:
>       > ```
>       > 1. change docstring, e.g. ....
>       > 2. update api.rst, e.g. ....
>       > 3. update whats-new.rst, e.g. ...
>       > ```
>       > 
>       > 
>       > Then, underline that building the documentation locally is optional:
>       > ```
>       > Optionally, you can build the docs locally as described in the next section.
>       > ```

Further PRs:

- Add dot, cov and corr to computation.rst

        PS: It looks like xr.dot is only present in the "API reference" and not in https://xarray.pydata.org/en/stable/computation.html. We should add dot, cov and corr to computation.rst in a future PR.

- Test (and if necessary implement) dask compatibility to cov and corr
- Improve documentation on How to contribute to the xarray documentation?

        I think it already gives a good overview, but then it might be unclear how to proceed.

        Maybe it might make sense to add a subsection How to contribute to the xarray documentation? in which the different options are shortly described examplary:

        1. change docstring, e.g. ....
        2. update api.rst, e.g. ....
        3. update whats-new.rst, e.g. ...

        Then, underline that building the documentation locally is optional:

        Optionally, you can build the docs locally as described in the next section.

- Extend xarray functionality, such that pairwise correlations and covariances can be easily calculated.
    Either write a wrapper function or extend xr.corr and xr.cov, such that the following computation is supported with ease:
    ```python
    pairwise_dim = 'space'
    for region in da_a[pairwise_dim]:
          print(xr.corr(da_a.sel(space=region.item()), da_a, dim='time').values)
    ```
   Target output would be an xarray with pairwise correlations, e.g.: 
   ```bash
   [ 1.         -0.43942827 -0.22425502]
   [-0.43942827  1.          0.30940081]
   [-0.22425502  0.30940081  1.        ]
   ```

xarray/core/computation.py
```diff
@@ -1047,6 +1046,163 @@ def earth_mover_distance(first_samples,
...
+    """
+
+    # 1. Broadcast the two arrays
+    da_a, da_b = broadcast(da_a, da_b)
+
+    # 2. Ignore the nans
+    valid_values = da_a.notnull() & da_b.notnull()
+    da_a = da_a.where(
+        valid_values, drop=True
+    )  # TODO: avoid drop as explained in https://github.com/pydata/xarray/pull/2652#discussion_r245492002
```
max-sixty: 👍

xarray/core/computation.py
```diff
@@ -1047,6 +1046,163 @@ def earth_mover_distance(first_samples,
         return apply_array_ufunc(func, *args, dask=dask)
 
 
+def cov(da_a, da_b, dim=None):
+    """Compute covariance between two DataArray objects along a shared dimension.
+
+    Parameters
+    ----------
+    arrays: DataArray (or Variable) objects
```
max-sixty: I realize these are mostly positional, but we should probably match the parameters here to the kwargs. e.g. if someone passes `arrays=`, they're going to be confused...

xarray/core/computation.py
```diff
@@ -1047,6 +1046,163 @@ def earth_mover_distance(first_samples,
...
+        Arrays to compute.
+    dim : str, optional
```
max-sixty: I think this can be a list too, but we can leave for another PR

xarray/core/computation.py
```diff
@@ -1047,6 +1046,163 @@ def earth_mover_distance(first_samples,
...
+
+    return cov
+
+
+def corr(da_a, da_b, dim=None):
+    """Compute correlation between two DataArray objects along a shared dimension.
+
+    Parameters
+    ----------
+    da_b : DataArray
```
max-sixty: Ditto re matching

xarray/core/computation.py
```diff
@@ -1047,6 +1046,163 @@ def earth_mover_distance(first_samples,
...
+    dim: str, optional
+        The dimension along which the correlation will be computed
+
+    Returns
+    -------
+    correlation: DataArray
+
+    See also
+    --------
+    pandas.Series.corr: corresponding pandas function
```
max-sixty: Could add `xr.cov` and vice versa to these

xarray/tests/test_computation.py
```diff
@@ -789,6 +811,84 @@ def func(x):
...
+        return ts1.to_series().corr(ts2.to_series())
+
+    expected = pd_corr(ts1, ts2)
+    actual = xr.corr(ts1, ts2)
+    assert np.allclose(expected, actual)
+
+    # Test #2: Misaligned N-D dataarrays with missing values
+    actual_ND = xr.corr(da, da_smooth, dim="time")
+    actual = select_pts(actual_ND)
+    np.allclose(expected, actual)
```
max-sixty: Same for all these `np.allclose`: it returns a bool, so we need an assert

max-sixty:
@shoyer take a glance if you get a second. I'm deliberately pushing this through even though https://github.com/pydata/xarray/pull/3550/files#diff-293280a1ed0154c9fb35b842f69e7e33R1112 isn't resolved, since I think these are important features which have dragged for years, and we can iterate on the performance (and @r-beer has done a great job on his first PR!)

max-sixty:
@r-beer great list of future PRs!

xarray/core/computation.py
r-beer: Yes, thanks for the precise control. I thought of also using arrays as parameter at one point, but guess explicit `da_a` and `da_b` is more explicit and better readable.

xarray/core/computation.py
r-beer: Yes, good point!

xarray/core/computation.py
r-beer: Done. 🙂

xarray/core/computation.py
r-beer: Ok, I added those.

xarray/tests/test_computation.py
r-beer: Sure, thanks! Added `assert statements` to all of those tests.

xarray/core/computation.py
r-beer: This is fine as preparation for future PR right? 

If there is an easy way to drop the drop, I can implement it of course!

r-beer:
Turns out the `` statements actually failed.
```bash
        actual = xr.cov(ts1, ts2)
>       assert np.allclose(expected, actual)
E       assert False
E        +  where False = <function allclose at 0x7ff5dbaf0e18>(0.004632253406505826, <xarray.DataArray ()>\narray(0.004323))
E        +    where <function allclose at 0x7ff5dbaf0e18> = np.allclose
```
0.00463 vs. 0.00432

xarray/core/computation.py
```diff
@@ -1047,6 +1046,163 @@ def earth_mover_distance(first_samples,
...
+    da_b = da_b.where(valid_values, drop=True)
+    valid_count = valid_values.sum(dim)
+
+    # 3. Compute mean and standard deviation along the given dim
+    demeaned_da_a = da_a - da_a.mean(dim=dim)
+    demeaned_da_b = da_b - da_b.mean(dim=dim)
+
+    # 4. Compute covariance along the given dim
+    cov = (demeaned_da_a * demeaned_da_b).sum(dim=dim) / (valid_count)
+
```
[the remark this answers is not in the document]
r-beer: @shoyer, @max-sixty how shall the covariance be normalized?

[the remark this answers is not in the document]
r-beer: For reference, the [numpy.cov documentation](https://docs.scipy.org/doc/numpy/reference/generated/numpy.cov.html):

>bias : bool, optional
>    Default normalization (False) is by (N - 1), where N is the number of observations given (unbiased estimate). If bias is True, then normalization is by N. These values can be overridden by using the keyword ddof in numpy versions >= 1.5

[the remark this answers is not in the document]
max-sixty: Yes I think let's default to numpy's default. 

We can add a `ddof=1` as a kwarg or leave that for another PR.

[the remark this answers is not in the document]
shoyer: +1 for copying NumPy's behavior. That convention is consistent with how we define `.var()`.

xarray/core/computation.py
```diff
@@ -1047,6 +1046,174 @@ def earth_mover_distance(first_samples,
...
+    return cov
+
+
+def corr(da_a, da_b, dim=None):
+    """Compute correlation between two DataArray objects along a shared dimension.
```
shoyer: This should be clearly documented as the "Pearson correlation coefficient"

xarray/core/computation.py
shoyer: I don't `drop=True` is needed here at all, and I think it actually introduces a bug. In particular, consider the following case:
```
>>> array = xarray.DataArray([[1, 2], [np.nan, np.nan]], dims=['x', 'y'])
>>> xarray.corr(array, array, dim='y')
<xarray.DataArray (x: 1)>
array([2.])
```
This doesn't look right to me at all. It has the wrong shape, size 1 instead of size 2, and also somehow the correlation is greater than 1.

xarray/core/computation.py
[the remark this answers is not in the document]
r-beer: OK, thanks for the feedback!

xarray/core/computation.py
r-beer: Updated the docstring accordingly

max-sixty:
@shoyer good spot, thanks

@r-beer we could add that as a test case

r-beer:
> 
> 
> @shoyer good spot, thanks
> 
> @r-beer we could add that as a test case

Yes, I will add it. 

Thinking about test arrays in general: 
Wouldn't it be good to define some data array fixtures in the `conftest.py` which can then be used in all different `test_function.py` tests?

In view of the additional test I started to restructure the `test_cov` and `test_corr` to a single `test_func`. I am now asking myself how much of the data preprocessing shall go to the parametrization and how much remains in the test function...

Any guidelines or suggestions?

Either way, it would be good to have a clear list of `da_a` and `da_b` and respective `expexted` results. I am a bit confused about what we actually would expect for `xarray.DataArray([[1, 2], [np.nan, np.nan]], dims=['x', 'y'])`.

r-beer:
> Either way, it would be good to have a clear list of `da_a` and `da_b` and respective `expected` results. I am a bit confused about what we actually would expect for `xarray.DataArray([[1, 2], [np.nan, np.nan]], dims=['x', 'y'])`.

Maybe, as a first step some type checking would be enough, to identify such special cases as the one above and give a `ValueError` or `NotImplementedError`?

max-sixty:
> > @shoyer good spot, thanks
> > @r-beer we could add that as a test case
> 
> Yes, I will add it.
> 
> Thinking about test arrays in general:
> Wouldn't it be good to define some data array fixtures in the `conftest.py` which can then be used in all different `test_function.py` tests?

I think we could do more of this, yes. It's a tradeoff between (a) better sharing & performance vs. (b) customization for each function & localization of effects. 
We have `def create_test_data` in a few files, before we used pytest, so could convert some of the usages to fixtures.
@shoyer and I discussed this briefly with @keewis 's work on tests for units. I think @shoyer is keener on having fixture-like code close to the test functions.

> In view of the additional test I started to restructure the `test_cov` and `test_corr` to a single `test_func`. I am now asking myself how much of the data preprocessing shall go to the parametrization and how much remains in the test function...
> 
> Any guidelines or suggestions?

I would start with having the code in the function; i.e. don't abstract too early unless we're really confident on the eventual state. When we find ourselves something multiple times, then we can start pulling it out into more abstractions (e.g. fixtures).

max-sixty:
> Either way, it would be good to have a clear list of `da_a` and `da_b` and respective `expected` results. I am a bit confused about what we actually would expect for `xarray.DataArray([[1, 2], [np.nan, np.nan]], dims=['x', 'y'])`.

For `xarray.corr(array, array, dim='y')`, I think we'd expect an array of the same dimensionality as below, with one point as the correlation of 1.0 (because `{1,2}` is 100% correlated to `{1,2}`) and another `NaN`. Does that make sense?

```
In [17]: xr.dot(da,da, dims='y')
Out[17]:
<xarray.DataArray (x: 2)>
array([ 5., nan])
```

r-beer:
> I would start with having the code in the function; i.e. don't abstract too early unless we're really confident on the eventual state. When we find ourselves something multiple times, then we can start pulling it out into more abstractions (e.g. fixtures).

Alright! I found myself having very similar tests and therefore abstracted it to test_func. I completely agree with not abstracting too early and but when having the same structure several times.

r-beer:
> For `xarray.corr(array, array, dim='y')`, I think we'd expect an array of the same dimensionality as below, with one point as the correlation of 1.0 (because `{1,2}` is 100% correlated to `{1,2}`) and another `NaN`. Does that make sense?

On the one side, I am with you in terms of "What You Put In Is What You Get Out", on the other hand `pandas.Series.cov` and `np.cov` have other behaviors that both seem plausible:

<details>

grafik

grafik
</details>

So for the moment, we might stick with `pandas.Series.cov`?

r-beer:
I have added several test cases, almost all pass. The ones that don't pass are related to `dim="time"` and incompletely implemented `pandas_cov` and `pandas_corr`, that currently don't have a `dim` parameter.

I had a look at `np.cov`, `np.corrcoef`, `pd.Series.cov`, `pd.frame.cov`, `pd.Series.corr`, and `pd.frame.corr`: Computation of `cov` and `corr` is implemented in many different ways. 

`xr.cov` and `xr.corr` now imitate the behavior of the respective `pd.Series` functions.

I actually prefer the numpy behavior, resulting in covariance and correlation matrices. However I feel that efficient implementation of this behavior is above my current understanding of xarray. So, I would highly appreciate your support on this implementation!

Otherwise, I added the `ddof` parameter to `xr.cov` to imitate the behavior of the respective `pd.Series` functions. I now get the results that `pd.Series` produces. However, I had to set `ddof=1` for `xr.cov` and `ddof=0` for `xr.corr`. Does that make sense?

xarray/tests/test_computation.py
```diff
@@ -789,6 +789,70 @@ def func(x):
...
+@pytest.mark.parametrize("da_a, da_b", array_tuples)
+@pytest.mark.parametrize("dim", [None, "time"])
+def test_cov(da_a, da_b, dim):
+    def pandas_cov(ts1, ts2):
+        """Ensure the ts are aligned and missing values ignored"""
+        ts1, ts2 = xr.align(ts1, ts2)
+        valid_values = ts1.notnull() & ts2.notnull()
+
+        ts1 = ts1.where(valid_values, drop=True)
+        ts2 = ts2.where(valid_values, drop=True)
```
max-sixty: I would have thought that we should be testing that we get a similar result from pandas' `cov` and xarray `cov` _without_ much preparation. Can you help me understand why we need to drop values here? What's the test case that fails when we don't have these lines?

xarray/tests/test_computation.py
```diff
@@ -789,6 +789,70 @@ def func(x):
     assert_identical(expected, actual)
 
 
+da = xr.DataArray(
+    np.random.random((3, 21, 4)),
+    coords={"time": pd.date_range("2000-01-01", freq="1D", periods=21)},
+    dims=("a", "time", "x"),
+)
+
+arrays = [
```
max-sixty: The issue with putting these at the module level (and one of the benefits of fixtures) is that each time this file is imported—even when no relevant test is being run—this will execute, and so it'll slow down rapid test collection. Could we put this in a fixture? I can help show you how if it's not clear. Thanks @r-beer

xarray/tests/test_computation.py
```diff
@@ -789,6 +789,70 @@ def func(x):
...
+def test_cov(da_a, da_b, dim):
+    def pandas_cov(ts1, ts2):
+        """Ensure the ts are aligned and missing values ignored"""
+        ts1, ts2 = xr.align(ts1, ts2)
+        valid_values = ts1.notnull() & ts2.notnull()
+
+        ts1 = ts1.where(valid_values, drop=True)
+        ts2 = ts2.where(valid_values, drop=True)
+
+        return ts1.to_series().cov(ts2.to_series())
```
keewis: The assert works as it is right now, but let's keep in mind that `actual` is a `DataArray` and `expected` is a float (`np.float64`, I think).

I wrote a lengthy explanation on this in https://github.com/pydata/xarray/pull/3550#discussion_r348652667, but the short version is that I think it might make sense to construct a `DataArray` using the result from `pandas_cov` and to then use `assert_allclose` to compare.

Something like this, maybe?
```suggestion
        cov = ts1.to_series().cov(ts2.to_series())
        return xr.DataArray(cov)
```
though that won't solve the issues with the covariance along a dimension (which pandas does not support, I think), so I suspect you will have to rewrite `pandas_cov` and `pandas_corr` using numpy.

xarray/tests/test_computation.py
```diff
@@ -789,6 +789,70 @@ def func(x):
...
+    (arrays[0], arrays[0]),
+    (arrays[0], arrays[1]),
+    (arrays[1], arrays[1]),
+    (arrays[2], arrays[2]),
+    (arrays[2], arrays[3]),
+    (arrays[3], arrays[3]),
+    (arrays[4], arrays[4]),
+    (arrays[4], arrays[5]),
+    (arrays[5], arrays[5]),
+]
```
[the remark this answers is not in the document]
max-sixty: If you want all permutations of the two variables, just pass them as separate parameters with pytest

xarray/tests/test_computation.py
max-sixty: You can also construct the result yourself—doing the 'compare with a known implementation' often saves time, but constructing the result is also fine!

xarray/tests/test_computation.py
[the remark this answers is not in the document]
keewis: sorry, I mistook `combinations` with `permutations`. Looking at it again, it seems that the pattern is some sort of "combine with the same index and the following one".

So never mind.

xarray/tests/test_computation.py
r-beer: Yes, I planned to make a fixture out of it. (How) Can I then easily use this fixture within parametrize?

The following code does not work, as the function is not iterable:

<details>

```python

@pytest.fixture()
def array_tuples():
    da = xr.DataArray(np.random.random((3, 21, 4)),
                      coords={"time": pd.date_range("2000-01-01", freq="1D", periods=21)},
                      dims=("a", "time", "x"),)

    arrays = [
        da.isel(time=range(0, 18)),
        da.isel(time=range(2, 20)).rolling(time=3, center=True).mean(dim="time"),
        xr.DataArray([0, np.nan, 1, 2, np.nan, 3, 4, 5, np.nan, 6, 7], dims="time"),
        xr.DataArray([1, 1, np.nan, 2, np.nan, 3, 5, 4, 6, np.nan, 7], dims="time"),
        xr.DataArray([[1, 2], [1, np.nan]], dims=["x", "time"]),
        xr.DataArray([[1, 2], [np.nan, np.nan]], dims=["x", "time"]),
    ]

    array_tuples = [
        (arrays[0], arrays[0]),
        (arrays[0], arrays[1]),
        (arrays[1], arrays[1]),
        (arrays[2], arrays[2]),
        (arrays[2], arrays[3]),
        (arrays[3], arrays[3]),
        (arrays[4], arrays[4]),
        (arrays[4], arrays[5]),
        (arrays[5], arrays[5]),
    ]

    return array_tuples


@pytest.mark.parametrize("da_a, da_b", array_tuples)
@pytest.mark.parametrize("dim", [None, "time", "x"])
def test_cov(da_a, da_b, dim):
    def pandas_cov(ts1, ts2):
```

</details>

Errors in:

```bash
E   TypeError: 'function' object is not iterable
```

xarray/tests/test_computation.py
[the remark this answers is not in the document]
r-beer: Thanks for the feedback. I have a questions
Should all possible data arrays be allowed? 

If so,  instead of using `array_tuples` we can use what I previously tested (but didn't commit):

```python

@pytest.mark.parametrize("da_b", arrays)
@pytest.mark.parametrize("da_a", arrays)
@pytest.mark.parametrize("dim", [None, "time", "x"])
def test_cov(da_a, da_b, dim):

```
This is, what @max-sixty means, right?

However, doesn't using all combinations might blow up the number of test functions unnecessarily? On the other hand not using `xarray_tuples` would be much more readable.

xarray/tests/test_computation.py
r-beer: I changed to `xr.DataArray` and `assert_allclose`, but I will first stick with comparing to `pandas.Series` results.

I will rewrite `pandas_cov` and `pandas_corr`, do you have suggestions how to implement the `dim` parameter?

keewis: :+1: for the `assert`, but let's take a look at the changes regarding `xr.DataArray`: I think you applied it in the wrong place. Since `cov` is a dataarray anyways, the result of `xr.cov` does not change (because `xr.DataArray(da).identical(da) == True` for any dataarray). The result of `pandas_cov` should be wrapped instead.

keewis: For the result you could use `np.cov` and `np.corr`, but you would probably have to first transpose the second object so `dim` is in the same axis for both `DataArray` objects. Since `numpy` uses `axis` instead of `dim` you'd then determine the axis for `dim` (`DataArray.get_axis_num`, I think).

keewis: However, that might be too much effort for a test and maybe not a good idea since you would end up reimplementing the function you want to test. So I think it probably would be better to follow @max-sixty's suggestion to construct the result yourself instead?

xarray/tests/test_computation.py
max-sixty: If you want to use `array_tuples` as a fixture, you need for the test function to take it as an argument, not to pass it as a argument to `parametrize`. (I know the multiple ways of parametrizing can be confusing)

5P2 combinations is fine for performance, particularly if the inputs are differentiated and so each test has marginal value. It does limit you to tests with a known correct implementation, though (you're not going to want to write out the results for all of those!)

r-beer: > If you want to use `array_tuples` as a fixture, you need for the test function to take it as an argument, not to pass it as a argument to `parametrize`. (I know the multiple ways of parametrizing can be confusing)
That means, that all the permutation is done within the fixture? It's not possible to put the fixture into the parametrize at all?

 > 5P2 combinations is fine for performance, particularly if the inputs are differentiated and so each test has marginal value. It does limit you to tests with a known correct implementation, though (you're not going to want to write out the results for all of those!)
Haha, yeah, don't want to write all those by hand. 🙂 
I did not completely get the part about "differentiation" and "marginal test value". Could you write one or two more sentences about it?

Thanks a lot for all the explanations!

xarray/tests/test_computation.py
r-beer: Thanks for the feedback! 

Unfortunately, I am very busy these days, so I just added the xr.Dataarray() in a haste, hence the wrong placement...

I hope I can take some time the weekend on 7./8.12. Cheers!

xarray/core/computation.py
```diff
@@ -1047,6 +1047,175 @@ def earth_mover_distance(first_samples,
...
+    da_a = da_a.where(valid_values, drop=True)
+    da_b = da_b.where(
+        valid_values, drop=True
+    )  # TODO: avoid drop as explained in https://github.com/pydata/xarray/pull/2652#discussion_r245492002
+
+    # 3. Compute correlation based on standard deviations and cov()
+    da_a_std = da_a.std(dim=dim)
+    da_b_std = da_b.std(dim=dim)
+
+    corr = cov(da_a, da_b, dim=dim, ddof=ddof) / (da_a_std * da_b_std)
```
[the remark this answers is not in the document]
max-sixty: I think we're more likely to have a working performant feature by make it work first, and then making it performant...

max-sixty:
@r-beer in case you come back to see this: thank you for taking it so far; your code was helpful to eventually getting this feature in. And we'd of course appreciate any additional contributions.

Issue #3552, opened 2019-11-20
Dataset.drop() no longer works with a set as an argument


xylar:
#### MCVE Code Sample

```python
# Your code here
#!/usr/bin/env python

import xarray
import numpy

ds = xarray.Dataset()
ds['x'] = ('x', numpy.linspace(0., 1., 100))
ds['y'] = ('y', numpy.linspace(0., 1., 50))
print(ds)
ds = ds.drop(set('x'))
print(ds)
```

#### Expected Output
```
<xarray.Dataset>
Dimensions:  (x: 100, y: 50)
Coordinates:
  * x        (x) float64 0.0 0.0101 0.0202 0.0303 ... 0.9697 0.9798 0.9899 1.0
  * y        (y) float64 0.0 0.02041 0.04082 0.06122 ... 0.9592 0.9796 1.0
Data variables:
    *empty*
<xarray.Dataset>
Dimensions:  (y: 50)
Coordinates:
  * y        (y) float64 0.0 0.02041 0.04082 0.06122 ... 0.9592 0.9796 1.0
Data variables:
    *empty*
```

#### Problem Description

In versions before xarray 0.14.1, the code above involving calls to `Dataset.drop()` with a `set` as an argument worked as expected.  With the new release, the result is an error as shown below.  This is breaking backwards compatibility with our software ([MPAS-Analysis](https://github.com/MPAS-Dev/MPAS-Analysis)).

```
<xarray.Dataset>
Dimensions:  (x: 100, y: 50)
Coordinates:
  * x        (x) float64 0.0 0.0101 0.0202 0.0303 ... 0.9697 0.9798 0.9899 1.0
  * y        (y) float64 0.0 0.02041 0.04082 0.06122 ... 0.9592 0.9796 1.0
Data variables:
    *empty*
Traceback (most recent call last):
  File "./drop_issue.py", line 10, in <module>
    ds = ds.drop(set('x'))
  File "/home/xylar/miniconda3/envs/test/lib/python3.7/site-packages/xarray/core/dataset.py", line 3643, in drop
    return self.drop_sel(labels, errors=errors)
  File "/home/xylar/miniconda3/envs/test/lib/python3.7/site-packages/xarray/core/dataset.py", line 3689, in drop_sel
    labels = either_dict_or_kwargs(labels, labels_kwargs, "drop")
[...]

Issue #3369, opened 2019-10-03
Define a process to test the readthedocs CI before merging into master


crusaderky:
This is an offshoot of #3358.

The readthedocs CI has a bad habit of failing even after the Azure Pipelines job "Docs" has succeeded.

After major changes that impact the documentation, and before merging everything into master, it would be advisable to explicitly verify that RTD builds correctly.

So far I tried to
1. create my own readthedocs project, https://readthedocs.org/projects/crusaderky-xarray/
2. point it to my fork https://github.com/crusaderky/xarray/
3. enable build for the branch I want to merge

This is currently failing because of an issue with versioneer, which incorrectly sets ``xarray.__version__`` to ``0+untagged.111.g6d60700``.
This in turn causes a failure in a minimum version check in ``pandas.DataFrame.to_xarray()`` on pandas>=0.25.

In the master RTD project https://readthedocs.org/projects/xray/, I can instead read ``xarray: 0.13.0+20.gdd2b803a``.

So far the only workaround I could find was to downgrade pandas to 0.24 in ``ci/requirements/doc.yml``.

dcherian:
Maybe switch away from versioneer? https://github.com/pydata/xarray/issues/2853

max-sixty:
We've had success with `setuptools_scm` internally, though not without any issues—in particular it has to be built in a git repo. Assuming that's the case, it requires very little maintainence. 

Anything I'm missing? I haven't used it in an OSS project

crusaderky:
I've been using setuptools-scm in multiple other projects and it's great!

PR #3554, opened 2019-11-20
Fix documentation

keewis:
It seems we introduced a warning in `whats-new.rst` and we dropped `Dataset.drop` and `DataArray.drop` from the documentation (which were only deprecated).

xref #3552 


keewis:
the docs fail because it cannot find a rasterio / tif file. Does anyone know if that is a random / network error or if something broke?

keewis:
@max-sixty, what do you think about re-adding the parameter / returns sections of `drop`?

max-sixty:
My thought behind removing `drop` was to only list methods that we recommend people use. We're keeping `drop` for back-compat reasons, but I wouldn't recommend anyone who is creating new code use it.

But maybe that's too aggressive. Any thoughts anyone?

max-sixty:
> the docs fail because it cannot find a rasterio / tif file. 

Hmmm, I haven't seen that before. Do you know whether it fails on master? (I can check too)

dcherian:
Here's the error 
```
rasterio.errors.RasterioIOError: '/vsicurl/https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif' does not exist in the file system, and is not recognized as a supported dataset name.
```

That URL does work. The `/vsicurl/` is screwing it up? 

I think our efforts may be better spent getting the `scip19-docs` branch to work because that migrates the gallery .py scripts to notebooks.

keewis:
Is the problem there still that readthedocs fails but CI works, isn't it?

@max-sixty, I think it should be fine as long as the deprecation is visible enough, but we could also add it to `api-hidden.rst` which would make the documentation page generate without adding it to `api.rst`

dcherian:
> Is the problem there still that readthedocs fails but CI works, isn't it?

:+1:

jthielen:
> But maybe that's too aggressive. Any thoughts anyone?

I'd like to have documentation available for anything in the public API, so that it can be a useful reference for both previously existing and new code. I think for deprecated functionality it makes the most sense to keep the docs, but warn prominently that it should not be used in new code (e.g. https://unidata.github.io/MetPy/dev/api/generated/metpy.calc.html#deprecated)?

keewis:
@dcherian: should I prepare a merge PR so we can make progress on that?

dcherian:
@keewis Yes please! We might have to iterate on RTD. Do you have an RTD build set up already? If not, see https://github.com/pydata/xarray/issues/3369

keewis:
I do, but the builds currently choke on conda taking too much memory. Let's discuss that in #3557.

max-sixty:
> I'd like to have documentation available for anything in the public API, so that it can be a useful reference for both previously existing and new code. I think for deprecated functionality it makes the most sense to keep the docs, but warn prominently that it should not be used in new code (e.g. https://unidata.github.io/MetPy/dev/api/generated/metpy.calc.html#deprecated)?

I think this is a pretty good balance, i.e. a section with `Deprecated / Pending Deprecation`. If we can do that as part of this then great; otherwise I'm happy to have a go

keewis:
@max-sixty, I'm probably too occupied with #3557, so go ahead if you want to (I'll remove the `drop` related changes).

keewis:
I removed the `drop` related changes, so we can merge?

PR #3555, opened 2019-11-20
Tweaks to release instructions

max-sixty:
A few small tweaks, including a script for getting all recent contributors (feel free to edit if my bash is bad).

### DIFF
--- a/HOW_TO_RELEASE.md
+++ b/HOW_TO_RELEASE.md
@@ -82,7 +82,7 @@ Time required: about an hour.
     You can get a list of contributors with script (NB: this has limited testing),
     replacing `v0.X.Y` with the _previous_ release
     ```
-    git log v0.X.Y.. --format="%aN" --reverse | perl -e 'my %dedupe; while (<STDIN>) { print unless $dedupe{$_}++}' | sort
+    git log v0.X.Y.. --format="%aN" --reverse | sort -u
     ```
 
 Note on version numbering:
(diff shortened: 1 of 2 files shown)


HOW_TO_RELEASE.md
```diff
@@ -53,11 +71,19 @@ Time required: about an hour.
...
+      - numpy-discussion@scipy.org
+      - scipy-user@scipy.org
+      - pyaos@lists.johnny-lin.com
+
     Google search will turn up examples of prior release announcements (look for
     "ANN xarray").
+    You can get a list of contributors with script (NB: this has limited testing),
+    replacing `v0.X.Y` with the _previous_ release
+    ```
+    git log v0.X.Y.. --format="%aN" --reverse | perl -e 'my %dedupe; while (<STDIN>) { print unless $dedupe{$_}++}' | sort
```
keewis: I'm not sure if that option is always available (could be a GNU extension), but this could be
```suggestion
git log v0.X.Y.. --format="%aN" --reverse | sort -u
```

max-sixty:
Great thanks @keewis !

keewis:
there is also a way to automatically select the version with the second-biggest release number:
```bash
git tag | sort -V | sed -n 'x;$p'
```

use it like this :
```bash
git log "$(git tag | sort -V | sed -n 'x;$p').." --format="%aN" --reverse | sort -u
```
but maybe that's too much automation?

max-sixty:
Nice bash-fu! I added it as an option while retaining the old one, until we've used it a couple of times

keewis:
That's what stackoverflow told me to use, I would have written `tail -2 | head -1`

I noticed `git tag` supports sorting, so this could also be
```bash
git log "$(git tag --sort="v:refname" | sed -n 'x;$p').." --format="%aN" | sort -u 
```
which is probably more reliable

PR #3464, opened 2019-10-30
Error in leap year?

mickaellalande:
I've tried this script; however, it adds +1 to all months of the leap years. It sounds like an error, or I am wrong? So I wrote the condition "and month == 2" line 86 so that only the month of February gets +1.

 - [ ] Closes #xxxx
 - [ ] Tests added
 - [ ] Passes `black . && mypy . && flake8`
 - [ ] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

PR #3557, opened 2019-11-20
merge scipy19 docs

keewis:
This PR tries to merge the scipy19-docs branch into master while also making sure that the readthedocs build does not fail (see #3199 and #3369).

There is also a [preview page](https://xarray-keewis.readthedocs.io/en/scipy19-docs/), but it currently [fails](https://readthedocs.org/projects/xarray-keewis/builds/) because conda is taking too much memory (which the official builds do not). Does anyone know why that is?


keewis:
something that could in the long run reduce the memory usage might be [conda-metachannel](https://metachannel.conda-forge.org/) ([blog post](https://medium.com/@marius.v.niekerk/conda-metachannel-f962241c9437)), but I think the integration into conda is in its early stages?

keewis:
it looks like it would be easier to do this by rebasing / cherry-picking onto master: there would not be as many conflicts as using merge would have (look at the list: 146 changed files). I'll cherry-pick and force push afterwards.

dcherian:
:+1:

.pre-commit-config.yaml
```diff
@@ -11,7 +11,7 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.720  # Must match ci/requirements/*.yml
+    rev: v0.730  # Must match ci/requirements/*.yml
```
max-sixty: IIUC `0.730` breaks some stuff so we are waiting for the next release

rabernat:
Sorry for creating this mess! 🤦‍♂️ Thank you for trying to clean it up.

.pre-commit-config.yaml
keewis: I'll try to match master.

keewis:
okay, now the cherry-picking is done and the number of touched files is down to a more reasonable 14, though I dropped the non-PR commits. I'm not sure which of these was really necessary, but I guess that's something that we will notice once we try to get CI and readthedocs to work?

The original branch is still available for comparison on [scipy19-docs](https://github.com/pydata/xarray/tree/scipy19-docs)

@dcherian (or anyone else who knows a bit more about RTD than me), can you help me fix the readthedocs [builds](https://readthedocs.org/projects/xarray-keewis/builds/)? I can't test anything as long as the environment creation fails.

dcherian:
Does anyhting on this page help? It's likely we have requested extra resources for xarray.

https://docs.readthedocs.io/en/stable/guides/build-using-too-many-resources.html#use-pip-when-possible

keewis:
and now I ran into the shadowing issue we had in #3199. I think the culprit is `cfgrib` which uses `extras_require` to pull in `xarray`. Is there any way to tell conda not to use that? We could also try to pull it using pip, since that works (at least for me) and bypasses the download of the packaged xarray.

dcherian:
maybe pip is the solution to all of these problems

keewis:
probably, but the only issue I found is this: https://pypi.org/project/iris/, which means `iris` is only available on conda-forge?

dcherian:
How about we pip install iris master just to see if we can get things to work?

keewis:
I'm currently trying to move everything except python and iris into the pip section, let's see what kind of memory usage that gives us.

keewis:
iris from github would work, but cartopy has no wheels, so installing using pip fails. I'll stop investigating this for now and try to make CI pass.

dcherian:
image

keewis:
now that CI passes we only need to get the readthedocs build to work (and maybe re-apply #3464?).

keewis:
oh, and we have some warnings regarding missing document labels. Is there a way to reference labels in notebooks?

keewis:
I added the changes from #3464 to the notebook. @rabernat, do you have the time to check whether the cherry-picks were correctly applied?

@dcherian, do you think you can try out this branch on the official RTD setup that (apparently) has more resources? I don't think I can make my setup work without a lot of effort.

keewis:
also, there is one issue with the labels:
this
```/examples/visualization_gallery.ipynb#parsing-rasterio%u2019s-geocoordinates```
does not work (because of the unicode character I think).

**Comment by dcherian** [comment]
OK it's going here: https://readthedocs.org/projects/xray/builds/10000789/

I renamed the existing `scipy19-docs` branch to `scipy19-docs-backup` and force pushed your branch to `scipy19-docs`

**Comment by dcherian** [comment]
My recollection is that we had some kind of version shadowing problem so maybe now is the time to replace versioneer: https://github.com/pydata/xarray/issues/2853

**Comment by keewis** [comment]
the shadowing was avoided by using pip for installing `cfgrib` (otherwise it would have pulled a packaged version of xarray). Anyways, the build seems to have been successfully completed

**Comment by dcherian** [comment]
is it time to party? 
image


https://xarray.pydata.org/en/scipy19-docs/examples/ROMS_ocean_model.html

**Comment by dcherian** [comment]
:tada: :tada: :tada: :tada:

**Comment by dcherian** [comment]
Let's just rename this heading (/examples/visualization_gallery.ipynb#parsing-rasterio%u2019s-geocoordinates) and merge?

**Comment by keewis** [comment]
okay. How about using `Parsing the geocoordinates of rasterio`?

**Comment by keewis** [comment]
oh, and is there a reason why we merge into stable instead of master?

**Comment by dcherian** [comment]
I just switched it so we can tweet about binderable examples ;)

**Comment by dcherian** [comment]
`Parsing rasterio geocoordinates`? :man_shrugging: 

Is that the last thing to fix?

**Comment by keewis** [comment]
I would think so, but I think we can fix anything that pops up later in new PRs.

**Comment by keewis** [comment]
one thing that might be done in a future PR is enabling the HTML reprs?

**Comment by max-sixty** [comment]
Feel free to ignore since it's not a big deal relative to the benefits, but looks like this adds some recent changes on `master` to `stable` (e.g. https://github.com/pydata/xarray/pull/3557/files#diff-99289912975ea19afc02f51c240ca451R18)

**Comment by keewis** [comment]
yeah, should we merge to stable we should probably rebase first

**Comment by keewis** [comment]
@dcherian, I'd merge into master and then cherry-pick into stable?

**Comment by max-sixty** [comment]
Should we change the base of this PR to master? Then we can choose when we push to stable? (But let's not get stuck on this, fine to just go ahead if we can't resolve)

**Comment by dcherian** [comment]
OK I jsut got too excited.

**Comment by keewis** [comment]
ah, maybe we should move the `whats-new.rst` entries?

**Comment by dcherian** [comment]
Yes! there seems to be one new example entry

**Comment by keewis** [comment]
would that be documentation or enhancements?

**Comment by dcherian** [comment]
Documentation

**Comment by keewis** [comment]
while we have an entry for #3116 the one for #3199 is missing (also, @rabernat, do you want an entry for the switching to `nbsphinx`?)

**Comment by dcherian** [comment]
Yes let's add appropriate entries for all those.

**Comment by keewis** [comment]
how about
```
- Switch doc examples to use nbsphinx and replace sphinx_gallery with notebook.
  (:pull:`3105`, :pull:`3106`, :pull:`3121`)
  By `Ryan Abernathey <https://github.com/rabernat>`
- Added example notebook demonstrating the visualization of ERA5 GRIB data.
  (:pull:`3199`).
  By `Zach Bruick <https://github.com/zbruick>` and `Stephan Siemen <https://github.com/StephanSiemen>`
```

dcherian:
:+1:

keewis:
done

keewis:
I think we can merge after making sure RTD still builds?

dcherian:
Great work as usual @keewis !

PR #3559, opened 2019-11-22
Reimplement quantile with apply_ufunc

### LINKED ISSUE #3326: quantile with Dask arrays
Currently the `quantile` method [raises an exception](https://github.com/pydata/xarray/blob/master/xarray/core/variable.py#L1637) when it encounters a Dask array.  

```python
        if isinstance(self.data, dask_array_type):
            raise TypeError(
                "quantile does not work for arrays stored as dask "
                "arrays. Load the data via .compute() or .load() "
                "prior to calling this method."
            )
```
I think it's because taking a quantile needs to see all the data in the dimension it's quantile-ing, or blocked/approximate methods weren't on hand when the feature was added.  Dask arrays where the dimension being quantile-ed was exactly one chunk in extent seem like a special case where no blocked algorithm is needed.  

The problem with following the suggestion of the exception (loading the array into memory) is that "wide and shallow" arrays are too big to load into memory, yet each chunk is statistically independent if the quantile dimension is the "shallow" dimension.

I'm not necessarily proposing delegating to Dask's quantile (unless it's super easy), but wanted to explore this special case described above.

Related links:  
* https://github.com/pydata/xarray/issues/2999 
* https://stackoverflow.com/a/47103407/745557 

Thank you!

EDIT: added stackoverflow link

dcherian:
Adds support for dask arrays.

 - [x] Closes #3326
 - [x] Tests added


xarray/core/variable.py
```diff
@@ -1716,40 +1716,45 @@ def quantile(self, q, dim=None, interpolation="linear", keep_attrs=None):
...
+            output_dtypes=[np.float64],
+            output_sizes={"quantile": len(q)},
+            dask="parallelized",
+            kwargs={"q": q * 100, "axis": axis, "interpolation": interpolation},
+        )
+
+        # for backward compatibility
+        result = result.transpose("quantile", ...)
+        if scalar:
+            result = result.squeeze()
```
shoyer: Always specify an axis with `squeeze` (at least in library code!)
```suggestion
            result = result.squeeze("quantile")
```

dcherian: Thanks for the review @shoyer. I always learn something!

doc/whats-new.rst
```diff
@@ -25,6 +25,9 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+- :py:meth:``Dataset.quantile``, :py:meth:``DataArray.quantile`` and ``GroupBy.quantile``
```
```suggestion
- :py:meth:`Dataset.quantile`, :py:meth:`DataArray.quantile` and ``GroupBy.quantile``
```

dcherian:
cc @huard in case something breaks ;)

PR #3566, opened 2019-11-23
Make 0d-DataArray compatible for indexing.

### LINKED ISSUE #3562: Minimize `.item()` call
#### MCVE Code Sample

I want to minimize the number of calls `.item()` within my data analysis.
It often happens 

1. when putting a 0d-DataArray into a slice
```python
da = xr.DataArray([0.5, 4.5, 2.5], dims=['x'], coords={'x': [0, 1, 2]})
da[: da.argmax()]
```
-> 
`TypeError: 'DataArray' object cannot be interpreted as an integer`

2. when using a 0d-DataArray for selecting
```python
da = xr.DataArray([0.5, 4.5, 2.5], dims=['x'], coords={'x': [0, 0, 2]})
da.sel(x=da['x'][0])
```
-> 
`IndexError: arrays used as indices must be of integer (or boolean) type`

Both cases, I need to call '.item()'.
It does not a big issue, but I think it would be nice if xarray becomes more self-contained.

fujiisoup:
- [x] Closes #3562
 - [x] Tests added

Now 0d-DataArray can be used for indexing.


xarray/core/coordinates.py
```diff
@@ -399,5 +399,7 @@ def remap_label_indexers(
             # drop coordinates found in indexers since .sel() already
             # ensures alignments
             coords = {k: var for k, var in v._coords.items() if k not in indexers}
-            pos_indexers[k] = DataArray(pos_indexers[k], coords=coords, dims=v.dims)
+            pos = DataArray(pos_indexers[k], coords=coords, dims=v.dims)
+            if pos.dtype.kind == "i":
+                pos_indexers[k] = pos
```
shoyer: Why should this only be done for integer dtypes? What happens with non-integer positional indexers?

I'm a little puzzled by how this is related to the other changes in this PR. It might make sense to add some specific tests related to whatever is being fixed here?

fujiisoup: Thanks for reviewing.
This meant to avoid wrapping a slice by a DataArray.
Updated to a cleaner logic.

sjvrijn:
Apart from a trivial conflict in `xarray\core\indexing.py`, this PR still looks fine to merge.

Illviljan:
Maybe @benbovy can take a look and see if this make sense with regards to #5692?

keewis:
> But we can also merging this now and then I'll fix it in #5692

that one is already way too big (>3500 added lines), so I'd vote to mark this as blocked by #5692

benbovy:
> that one is already way too big

True 😄

benbovy:
FYI @fujiisoup (or anyone else interested by taking over this PR), label selection using a 0d-DataArray has now been implemented as part of the index refactor.

PR #3576, opened 2019-11-27
Examples for quantile

keewis:
Following up to #3527, this adds examples for `DataArray.quantile`, `Dataset.quantile` and `GroupBy.quantile` (which is inherited by both `DatasetGroupBy` and `DataArrayGroupBy`).


keewis:
there seems to be something wrong with one of the rasterio tests. Due to a new version of rasterio, maybe?

max-sixty:
These are great! Thanks as ever @keewis !

max-sixty:
Do you want to add to `whatsnew` in Documentation? I hit the green button so would need to be another PR.

keewis:
see #3581.

PR #3577, opened 2019-11-27
use cftime's intersphinx

keewis:
Now that Unidata/cftime#133 is merged, we can link to the `cftime` documentation using intersphinx (silencing a few more sphinx warnings).


dcherian:
Thanks @keewis

Issue #3573, opened 2019-11-25
rasterio test failure


dcherian:
version 
```
rasterio                  1.1.1            py36h900e953_0    conda-forge
```

```
=================================== FAILURES ===================================
________________________ TestRasterio.test_rasterio_vrt ________________________

self = <xarray.tests.test_backends.TestRasterio object at 0x7fc8355c8f60>

    def test_rasterio_vrt(self):
        import rasterio
    
        # tmp_file default crs is UTM: CRS({'init': 'epsg:32618'}
        with create_tmp_geotiff() as (tmp_file, expected):
            with rasterio.open(tmp_file) as src:
                with rasterio.vrt.WarpedVRT(src, crs="epsg:4326") as vrt:
                    expected_shape = (vrt.width, vrt.height)
                    expected_crs = vrt.crs
                    expected_res = vrt.res
                    # Value of single pixel in center of image
                    lon, lat = vrt.xy(vrt.width // 2, vrt.height // 2)
>                   expected_val = next(vrt.sample([(lon, lat)]))

xarray/tests/test_backends.py:3966: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/rasterio/sample.py:43: in sample_gen
    data = read(indexes, window=window, masked=masked, boundless=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   ValueError: WarpedVRT does not permit boundless reads

rasterio/_warp.pyx:978: ValueError
```

keewis:
I'm fairly certain this is mapbox/rasterio#1833

PR #3585, opened 2019-12-01
Add bottleneck & rasterio git tip to upstream-dev CI

dcherian:


dcherian:
Also xfails the failing rasterio test (#3573)

PR #3587, opened 2019-12-02
boundary options for rolling.construct

fujiisoup:
- [x] Closes #2003, #2011
 - [x] Tests added

Added some boundary options for rolling.construct.
Currently, the option names are inherited from `np.pad`, `['edge' | 'reflect' | 'symmetric' | 'wrap']`.
Do we want a more intuitive name, such as `periodic`?

### DIFF
--- a/xarray/core/dask_array_ops.py
+++ b/xarray/core/dask_array_ops.py
@@ -27,7 +27,7 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
     return result
 
 
-def rolling_window(a, axis, window, center, fill_value):
+def rolling_window(a, axis, window, center, fill_value, mode):
     """Dask's equivalence to np.utils.rolling_window
     """
     import dask.array as da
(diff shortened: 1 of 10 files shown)


xarray/core/dask_array_ops.py
```diff
@@ -27,7 +28,109 @@ def dask_rolling_wrapper(moving_func, a, window, min_count=None, axis=-1):
...
 
 
-def rolling_window(a, axis, window, center, fill_value):
+def rolling_window(a, axis, window, center, fill_value, mode):
+    import dask
+
+    if version.LooseVersion(dask.__version__) < "1.7":
+        if mode is not None:
+            raise NotImplementedError(
+                "dask >= 1.7 is required for mode={}.".format(mode)
```
shoyer: Is this the right version number for dask? I can't find a 1.7 release.

Also note that we only need to support versions of dask < 6 months old, per http://xarray.pydata.org/en/stable/installing.html. If we can just delete the old code entirely, that might be cleaner.

fujiisoup: [Dask API page](https://docs.dask.org/en/latest/array-api.html#dask.array.pad) says it came at the version 1.7, but I just noticed this doc is just copied from numpy...
I'll fix it.

xarray/core/rolling.py
```diff
@@ -206,6 +206,11 @@ def construct(self, window_dim, stride=1, fill_value=dtypes.NA):
             Size of stride for the rolling window.
         fill_value: optional. Default dtypes.NA
             Filling value to match the dimension size.
+        mode: optional. Default None
```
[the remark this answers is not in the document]
dcherian: `pad_mode` seems better since we should eventually have `da.pad(mode="constant")`.

mathause:
Thanks for picking this up - excited to get this in! Are you able to get it to work with dask arrays? (That was the stopper in my PR.) Or maybe you can require that dask arrays are not chunked along the rolling dimension?

mathause:
@fujiisoup I would eventually like to pick this one up. Can you remember what was missing? Mostly the tests?

fujiisoup:
I was thinking to wait for the pad method implemented but forgot until now.
I am not sure this is easily merginable, as the rolling.py has been updated for a while...

The first motivation was to implement the rolling operation for the periodic coordinate, but it is not yet implemented.

PR #3588, opened 2019-12-03
Optimize isel for lazy array equality checking

### LINKED ISSUE #3558: optimizing xarray operations for lazy array equality test
TLDR: I think we want `A.sel(x=A.x).equals(A)` to pass lazily. It doesn't do so currently.

Currently if I do `A.sel(x=A.x)`, this sticks in a `getitem` call in the dask graph which breaks our lazy array equality optimization. Here's an example

``` python
>>> A = xr.DataArray(np.arange(100), dims="x", coords={"x": np.arange(100)}).chunk({"x": 1})
>>> A.sel(x=A.x).variable.equals(A, equiv=xr.core.duck_array_ops.lazy_array_equiv)
None
```

**Questions:**

1. Where is the best place to do this? In `sel` or `isel`? Both? 
    Sticking the following in `sel` makes the above check return `True` which is what we want:
    ``` python
        if self._indexes:
            equals = []
            for index in indexers:
                equals.append(indexers[index].to_index().equals(self._indexes[index]))

            if all(equals):
                return self
    ```

    This doesn't handle slice objects though so that makes me think we'd want to add something similar to `isel` too.
2. What is the behaviour we want? `A.sel(x=A.x).equals(A)` or `A.sel(x=A.x) is A`? 
3. Doing the latter will mean changing `_to_temp_dataset` and `_from_temp_dataset` which suggests the constraint `A._from_temp_dataset(A._to_temp_dataset()) is A`? But this seems too strong to me. Do we only want to lazily satisfy an `equals` constraint rather than an `identical` constraint?
4. It seems like we'll want to add such short-circuits in many places (I have not checked all of these): `sortby`, `broadcast`, `align`, `reindex` (`transpose` does this now).

dcherian:
- [x] Closes #3558
 - [x] Tests added

This mostly adds tests that already pass (yay!).


xarray/core/variable.py
```diff
@@ -540,6 +540,20 @@ def _broadcast_indexes(self, key):
...
+                # let da.sel(x=da.x) pass but skip if Variable has different dimensions
+                # e.g. da.sel(x=Variable(("points",), [0, 1, 2]))
+                if isinstance(k, Variable) and k.dims != (dim,):
+                    continue
+                if duck_array_ops.array_equiv(k, np.arange(self.sizes[dim])):
+                    # short-circuit when keys are effectively slice(None)
+                    # This preserves dask name and passes lazy array equivalence checks
+                    # (see duck_array_ops.lazy_array_equiv)
+                    key_dict[dim] = slice(None)
+        key = tuple(key_dict.values())
```
dcherian: This is the big change. It rewrites things like `da.sel(x=da.x)` to `da.sel(x=slice(None))` which seems to do what we want i.e. not insert a `getitem` call in the dask graph.

A nice benefit is that `sortby` on a sorted object just returns the same object.

Is this the right place to make this change? (I wasn't comfortable going any deeper into the indexing code).

shoyer: My concern is that this changes the semantics of xarray indexing on NumPy arrays, because we now return a view of NumPy arrays rather than a copy.

dcherian: hmmm.... How about we special case this section and have it run only on dask arrays. The view-vs-copy distinction does not apply to dask arrays, does it?

shoyer: I agree, this would be safe on dask arrays. But it probably makes sense to do that at a lower level, e.g., inside `DaskIndexingAdapter`. Or maybe even inside dask itself?

dcherian: Now in `DaskIndexingAdapter`

xarray/core/variable.py
```diff
@@ -1054,7 +1054,9 @@ def isel(
 
         invalid = indexers.keys() - set(self.dims)
         if invalid:
-            raise ValueError("dimensions %r do not exist" % invalid)
+            raise ValueError(
+                f"dimensions {invalid} do not exist. Expected one or more of {self.dims} "
```
crusaderky:
```suggestion
                f"dimensions {invalid} do not exist. Expected one or more of {self.dims}"
```

dcherian:
Thanks for the review @crusaderky

max-sixty:
Great, thanks @dcherian !

PR #3589, opened 2019-12-03
Resolve the version issues on RTD

### LINKED ISSUE #3567: version collisions on readthedocs
In #3199 the documentation on `readthedocs` did not build because the newest packaged version of xarray was pulled in by a dependency -- in this case `cfgrib`. I'm not sure why, but this shadows the versions installed by at least `python setup.py install --force` and `python -m pip install -e .` (not `python -m pip install .`, I think, but `readthedocs` uses `--update-strategy=eager` for pip-installing, which deactivates version pinning).

Fortunately, `cfgrib` does not have a hard dependency on `xarray` (it's in the `extras_require` section), so this issue was bypassed in #3557 by using pip to install it. Should we ever want to introduce a dependency that requires `xarray` (or to use conda to install `cfgrib`), this is bound to resurface.

It might be that this is a problem with how versioneer constructs versions based on git commit hashes and a fix to #2853 is a fix to this, but it certainly needs more investigation.

#3369 is related, but more about catching these sort of issues before merging.

keewis:
The trick mentioned in https://github.com/pydata/xarray/issues/3567#issuecomment-561270215 works locally, but I don't know whether that's also the case for RTD. Since my own RTD setup cannot go past the environment creation stage: can someone with the appropriate rights on the official setup try this out for me?

Adding `xarray` to `doc.yml` is something temporary that has to be removed before the merge, but I think it is necessary to make sure the version collision occurs.

 - [x] Closes #3567 

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,9 +15,14 @@
 
 import datetime
 import os
+import pathlib
 import subprocess
 import sys
 from contextlib import suppress
+
+root = pathlib.Path(__file__).absolute().parent.parent
+os.environ["PYTHONPATH"] = str(root)
+sys.path.insert(0, str(root))
 
 import xarray
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -20,6 +20,7 @@ import subprocess
 import sys
 from contextlib import suppress
 
+# make sure the source version is preferred (#3567)
 root = pathlib.Path(__file__).absolute().parent.parent
 os.environ["PYTHONPATH"] = str(root)
 sys.path.insert(0, str(root))
--- a/ci/requirements/doc.yml
+++ b/ci/requirements/doc.yml
@@ -6,7 +6,7 @@ dependencies:
   - python=3.7
   - bottleneck
   - cartopy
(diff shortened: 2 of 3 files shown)


doc/conf.py
```diff
@@ -15,10 +15,15 @@
 
 import datetime
 import os
+import pathlib
 import subprocess
 import sys
 from contextlib import suppress
 
+root = pathlib.Path(__file__).absolute().parent.parent
```
max-sixty: Do you want to add a comment here referencing this / the original issue, so it's not removed in future?

keewis: done?

keewis:
okay, then we will need to create a new PR that removes the `xarray` dependency from `doc.yml` again?

For future reference I think you can push the PR branch to the main xarray repository, then activate it in https://readthedocs.org/projects/xray/versions/. Deactivating is possible by disabling the checkbox in the `edit` menu of the branch.

keewis:
it works without breaking :tada:
I'll create a new PR to update `doc.yml`.

PR #3591, opened 2019-12-03
update doc.yml

### LINKED ISSUE #3567: version collisions on readthedocs

keewis:
Remove `xarray` from `doc.yml` which got added in order to test whether #3589 truly fixes the issue mentioned in #3567.

### DIFF
--- a/ci/requirements/doc.yml
+++ b/ci/requirements/doc.yml
@@ -21,5 +21,4 @@ dependencies:
   - seaborn
   - sphinx
   - sphinx_rtd_theme
-  - xarray
   - zarr


keewis:
in the `Replicate readthedocs CI environment` step, conda complains that `cfgrib` is causing the environment to be inconsistent. Do we need to update it?

keewis:
also, this should probably be a different PR, but we should update `readthedocs.yml` to use the new format.

dcherian:
> but we should update readthedocs.yml to use the new format.

yes let's open an issue for this.

PR #3592, opened 2019-12-04
Silence sphinx warnings: Round 2

keewis:
As a follow-up to #3516, this silences the last warnings without `-n`, so once the doctest issue below is fixed we could replace `-n` with `-W --keep-going` and have the CI fail should a PR introduce new warnings.

There is, however, a problem with the doctest output for the 3D result of `GroupBy.quantile`: since the repr of a DataArray with at least 3 dimensions contains blank lines, it complains about a block quote ending without a blank line.

Also, the complaints from #3516 about duplicate definitions of attributes were caused by the `Attributes` section in the docstring of `DataArray`. I fixed this by removing the whole section, but we probably want to first make sure this does not remove any information (or maybe someone has a different idea?).


max-sixty:
Great!! Will be good to get CI up & running. Should we add that as part of this PR?

doc/whats-new.rst
```diff
@@ -30,7 +30,7 @@ New Features
   By `Deepak Cherian <https://github.com/dcherian>`_.
 - Added the :py:meth:`count` reduction method to both :py:class:`DatasetCoarsen`
   and :py:class:`DataArrayCoarsen` objects. (:pull:`3500`)
-  By `Deepak Cherian <https://github.com/dcherian/>`_
+  By `Deepak Cherian <https://github.com/dcherian>`_
```
dcherian: :) thanks!

Can you fix the github links in the Documentation section below too?

keewis: which links?

keewis:
do we agree on dropping `-n` and adding `-W --keep-going` to the CI? If so, I think I can do that in this PR.

doc/whats-new.rst
dcherian: image

I guess the quantile links need to be fixed too.

keewis: oh, yes, they are missing the trailing underscore and the `:py:meth:` prefix. Done.

dcherian:
> do we agree on dropping -n and adding -W --keep-going to the CI?

I think this is OK. If it gets too annoying, we can revert

>  I fixed this by removing the whole section

I think this is OK too. The attributes seem to get added to the html documentation automatically. The list you removed is formatted kind of badly at the beginning of the docstring. see the end of https://xarray.pydata.org/en/stable/generated/xarray.DataArray.html

keewis:
I'll update the CI, then.

We could also have CI run doctest so we know which examples need updates, but we'd first need fix all outdated examples.

keewis:
now CI should fail on new warnings (no nit-picks, so no complaints about broken links)

max-sixty:
Looks great!! Really good that we can enforce the warnings from here 

(I'm on vacation so forgive the lack of engagement)

dcherian:
Thanks @keewis

PR #3596, opened 2019-12-04
Add DataArray.pad, Dataset.pad, Variable.pad

### LINKED ISSUE #2605: Pad method
It would be nice to have a generic `.pad()` method to xarray objects based on numpy.pad and dask.array.pad.

In particular,`pad` with `mode='wrap'` could solve several use-cases related to periodic boundary conditions: https://github.com/pydata/xarray/issues/1005 , https://github.com/pydata/xarray/issues/2007. For example, `ds.pad(longitude=(0, 1), mode='wrap')` to add an extra point with periodic boundary conditions along the `longitude` dimension.

It probably makes sense to linearly extrapolate coordinates along padded dimensions, as long as they are regularly spaced. This might use heuristics and/or a keyword argument.

I don't have a plans to work on this in the near term. It could be a good project of moderate complexity for a new contributor.

mark-boer:
Hello all,

This is my first PR to a pydata project. This pull request is still very much a work in progress and I could really use your input on a couple of things.

 - [x] Closes #2605 
 - [x] Tests added

 1. I moved the custom dask pad method into dask_array_compat to ensure backwards compatability to Dask versions that do not have dask.pad yet. We could chose to drop this support if we wanted to.
 2. I'm still in doubt about the function signature, numpy as dask use optional kwargs, but that kinda interferes with the `**pad_width_kwargs`. I chose the signature that I thought looked least awkward.
 3. The default behaviour of pad with `mode=constant` pads with NaN's converting the array to float in the process. This goes against the default behaviour of numpy as dask.
 4. How should the coordinates of a DataArray be padded? I chose default padding except for modes "edge", "reflect", "symmetric", "wrap".
 5. How should we handle inconsistencies between numpy.pad and Dask.pad, it turns out there are a couple [5303](https://github.com/dask/dask/issues/5303)

Dataset.pad is coming up.


xarray/core/variable.py
```diff
@@ -1099,66 +1099,85 @@ def shift(self, shifts=None, fill_value=dtypes.NA, **shifts_kwargs):
...
+
+        # workaround for bug in Dask's default value of stat_length  https://github.com/dask/dask/issues/5303
+        if stat_length is None and mode in ["maximum", "mean", "median", "minimum"]:
+            stat_length = [(n, n) for n in self.data.shape]
+
+        pads = [(0, 0) if d not in pad_widths else pad_widths[d] for d in self.dims]
+
+        # numpy/dask work with optional kwargs
+        pad_option_kwargs = {}
+        if stat_length is not None:
```
dcherian: are these if conditions needed? just pass these kwargs directly in the `duck_array_ops.pad` call below even if they are None?

xarray/tests/test_variable.py
```diff
@@ -782,14 +784,37 @@ def test_pad(self):
         data = np.arange(4 * 3 * 2).reshape(4, 3, 2)
         v = self.cls(["x", "y", "z"], data)
 
+        # TODO "mean", "median" and "reflect" have issues
+        modes = ["edge", "linear_ramp", "maximum", "minimum", "symmetric", "wrap"]
```
dcherian: I would do this as
```python

@pytest.mark.parametrize("modes", [
pytest.param("mean", marks=pytest.mark.xfail),
pytest.param("median", marks=pytest.mark.xfail),
pytest.param("reflect", marks=pytest.mark.xfail),
"edge", "linear_ramp", "maximum", "minimum", "symmetric", "wrap"])
@pytest.mark.parametrize("xr_args, np_args", 
[
  [{"x": (2, 1)}, ((2,1), (0,0), (0,0))],
...
])
def test_pad(modes, xr_args, np_args):
```

instead of the for loop on line 795

xarray/tests/test_variable.py
```diff
@@ -1848,6 +1873,11 @@ def test_getitem_with_mask_nd_indexer(self):
             self.cls(("x", "y"), [[0, -1], [-1, 2]]),
         )
 
+    def test_pad(self):
```
dcherian: This can be removed since our minimum version is dask 1.2

xarray/core/dataarray.py
```diff
@@ -3110,6 +3110,158 @@ def map_blocks(
...
+            One of pad_widths or pad_widths_kwarg must be provided.
+
+        Returns
+        -------
+        padded : DataArray
+            DataArray with the padded coordinates and data.
+
+        See also
+        --------
+        shift
```
dcherian: DataArray.shift, DataArray.roll, numpy.pad, dask.array.pad

xarray/tests/test_variable.py
```diff
@@ -782,14 +784,37 @@ def test_pad(self):
...
+            ((2, 1), (0, 0), (0, 0)),
+            ((0, 0), (0, 3), (0, 0)),
+            ((3, 1), (0, 0), (2, 0)),
+        ]
+        for (xr_arg, np_arg), mode in itertools.product(zip(xr_args, np_args), modes):
+            print(mode)
+            actual = v.pad(mode=mode, **xr_arg)
+            expected = np.pad(
+                np_arg,
+                mode=mode,
```
dcherian: Looks like you need `pad_width` here for tests to run?

max-sixty:
Thanks for this @mark-boer ! 

We try and review PRs quickly. I'm out on vacation but hopefully others can have a look soon. Thanks for your patience.

fujiisoup:
Hi, @mark-boer.
In #3587, I tried using dask's pad method but noticed a few bugs in older (but newer than 1.2) dask.
For me, it would be very welcome if you add this method to dask_array_compat.
Then, I would wait for merging #3587 until this PR is completed.

Thanks for your contribution :)

xarray/tests/test_variable.py
mark-boer: Cool, didn't know that was possible!

xarray/core/variable.py
mark-boer: Unfortunately, yes.

xarray/core/variable.py
```diff
@@ -1099,66 +1099,85 @@ def shift(self, shifts=None, fill_value=dtypes.NA, **shifts_kwargs):
...
+        padded : Variable
+            Variable with the same dimensions and attributes but padded data.
         """
         pad_widths = either_dict_or_kwargs(pad_widths, pad_widths_kwargs, "pad")
 
-        if fill_value is dtypes.NA:
-            dtype, fill_value = dtypes.maybe_promote(self.dtype)
+        # change default behaviour of pad with mode constant
+        if mode == "constant" and constant_values is None:
+            dtype, constant_values = dtypes.maybe_promote(self.dtype)
```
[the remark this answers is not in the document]
mark-boer: We need some kind of default behaviour, there is currently no way to add the fill_value for the coords of a data array. This behaviour was already implemented by the `variable.pad_with_fill_value` method as it existed in xarray, I thought I would leave it as is.

We could chose to pad with 0's as numpy does it, but I don't think I would prefer this.

xarray/tests/test_variable.py
mark-boer: Unfortunately, decorators are not inherited, so I'm not sure about this change. It feels kind of awkward in `TestIndexVariable.test_pad`

dcherian: OK. Leave it for now. I can see if it works and push a change later

dcherian:
It seems like we have some value mismatches on `dask==1.2`. @fujiisoup is this the error you found?

```
__________ TestVariableWithDask.test_pad[xr_arg0-np_arg0-linear_ramp] __________

self = <xarray.tests.test_variable.TestVariableWithDask object at 0x7fb69b629b00>
mode = 'linear_ramp', xr_arg = {'x': (2, 1)}, np_arg = ((2, 1), (0, 0), (0, 0))

    @pytest.mark.parametrize(
        "mode",
        [
            pytest.param("mean", marks=pytest.mark.xfail),
            pytest.param("median", marks=pytest.mark.xfail),
            pytest.param("reflect", marks=pytest.mark.xfail),
            "edge",
            "linear_ramp",
            "maximum",
            "minimum",
            "symmetric",
            "wrap",
        ],
    )
    @pytest.mark.parametrize(
        "xr_arg, np_arg",
        [
            [{"x": (2, 1)}, ((2, 1), (0, 0), (0, 0))],
            [{"y": (0, 3)}, ((0, 0), (0, 3), (0, 0))],
            [{"x": (3, 1), "z": (2, 0)}, ((3, 1), (0, 0), (2, 0))],
        ],
    )
    def test_pad(self, mode, xr_arg, np_arg):
        data = np.arange(4 * 3 * 2).reshape(4, 3, 2)
        v = self.cls(["x", "y", "z"], data)
    
        actual = v.pad(mode=mode, **xr_arg)
        expected = np.pad(data, np_arg, mode=mode,)
    
>       assert_array_equal(actual, expected)
E       AssertionError: 
E       Arrays are not equal
E       
E       (mismatch 2.3809523809523796%)
E        x: array([[[ 0,  0],
E               [ 0,  0],
E               [ 0,  0]],...
E        y: array([[[ 0,  0],
E               [ 0,  0],
E               [ 0,  0]],...

xarray/tests/test_variable.py:821: AssertionError
__________ TestVariableWithDask.test_pad[xr_arg1-np_arg1-linear_ramp] __________

self = <xarray.tests.test_variable.TestVariableWithDask object at 0x7fb69df9eb00>
mode = 'linear_ramp', xr_arg = {'y': (0, 3)}, np_arg = ((0, 0), (0, 3), (0, 0))

    @pytest.mark.parametrize(
        "mode",
        [
            pytest.param("mean", marks=pytest.mark.xfail),
            pytest.param("median", marks=pytest.mark.xfail),
            pytest.param("reflect", marks=pytest.mark.xfail),
            "edge",
            "linear_ramp",
            "maximum",
            "minimum",
            "symmetric",
            "wrap",
        ],
    )
    @pytest.mark.parametrize(
        "xr_arg, np_arg",
        [
            [{"x": (2, 1)}, ((2, 1), (0, 0), (0, 0))],
            [{"y": (0, 3)}, ((0, 0), (0, 3), (0, 0))],
            [{"x": (3, 1), "z": (2, 0)}, ((3, 1), (0, 0), (2, 0))],
        ],
    )
    def test_pad(self, mode, xr_arg, np_arg):
        data = np.arange(4 * 3 * 2).reshape(4, 3, 2)
        v = self.cls(["x", "y", "z"], data)
    
        actual = v.pad(mode=mode, **xr_arg)
        expected = np.pad(data, np_arg, mode=mode,)
    
>       assert_array_equal(actual, expected)
E       AssertionError: 
E       Arrays are not equal
E       
E       (mismatch 16.66666666666667%)
E        x: array([[[ 0,  1],
E               [ 2,  3],
E               [ 4,  5],...
E        y: array([[[ 0,  1],
E               [ 2,  3],
E               [ 4,  5],...

xarray/tests/test_variable.py:821: AssertionError
__________ TestVariableWithDask.test_pad[xr_arg2-np_arg2-linear_ramp] __________

self = <xarray.tests.test_variable.TestVariableWithDask object at 0x7fb69c609860>
mode = 'linear_ramp', xr_arg = {'x': (3, 1), 'z': (2, 0)}
np_arg = ((3, 1), (0, 0), (2, 0))

    @pytest.mark.parametrize(
        "mode",
        [
            pytest.param("mean", marks=pytest.mark.xfail),
            pytest.param("median", marks=pytest.mark.xfail),
            pytest.param("reflect", marks=pytest.mark.xfail),
            "edge",
            "linear_ramp",
            "maximum",
            "minimum",
            "symmetric",
            "wrap",
        ],
    )
    @pytest.mark.parametrize(
        "xr_arg, np_arg",
        [
            [{"x": (2, 1)}, ((2, 1), (0, 0), (0, 0))],
            [{"y": (0, 3)}, ((0, 0), (0, 3), (0, 0))],
            [{"x": (3, 1), "z": (2, 0)}, ((3, 1), (0, 0), (2, 0))],
        ],
    )
    def test_pad(self, mode, xr_arg, np_arg):
        data = np.arange(4 * 3 * 2).reshape(4, 3, 2)
        v = self.cls(["x", "y", "z"], data)
    
        actual = v.pad(mode=mode, **xr_arg)
        expected = np.pad(data, np_arg, mode=mode,)
    
>       assert_array_equal(actual, expected)
E       AssertionError: 
E       Arrays are not equal
E       
E       (mismatch 5.208333333333329%)
E        x: array([[[ 0,  0,  0,  0],
E               [ 0,  0,  0,  0],
E               [ 0,  0,  0,  0]],...
E        y: array([[[ 0,  0,  0,  0],
E               [ 0,  0,  0,  0],
E               [ 0,  0,  0,  0]],...
```

mark-boer:
> It seems like we have some value mismatches on `dask==1.2`. 

The reason that `mean` and `reflect` are marked as xfail are because there too there are value mismatches. I'll see if I can find a workaround for this failing test.

mark-boer:
Hi everyone, happy new year :tada:

I feel like this PR is slowly getting into a descent shape. But I could use your input on a couple of subjects:

**How should we handle the default padding?**
I currently pad with `dtypes.NaN` this is similar behaviour as used by the replaced  variable.pad_with_fill_value` and is consistent with the `shift` method. 

We could chose to pad with 0's as numpy does it, or we could force the user to choose a constantt_value,  but I don't think I would prefer this.

**How should the coordinates of a DataArray/set be padded?** 
I chose default padding except for modes "edge", "reflect", "symmetric", "wrap". @dcherian noted that this could make sense in some situations, but that I would require some further discussion.

Personally I think it more often than not it makes sense, but padding with NaN's should also work fine.

**dask_array_compat**
Also I created a workaround for the changing of dtype by Dask.pad mode=mean and added an additional check to validate the output shape of dask.pad. Let me know what you think about this, I'm not 100% convinced...

I used a couple of `# type: ignore` as any work around the mypy errors lead to uglier code. But if any of you have some suggestions, I'm happy to hear them

mark-boer:
Hi @dcherian and @fujiisoup, appart from the issues I raised in my last comment, I think this PR is close to done.

Thx in advance

xarray/tests/test_variable.py
```diff
@@ -785,36 +785,69 @@ def test_getitem_error(self):
         with raises_regex(IndexError, "Dimensions of indexers mis"):
             v[:, ind]
 
-    def test_pad(self):
+    @pytest.mark.parametrize(
+        "mode",
+        [
+            "mean",
+            pytest.param("median", marks=pytest.mark.xfail),
```
dcherian: let's do
```
            pytest.param("median", marks=pytest.mark.xfail(reason="not supported by dask")),
            pytest.param("reflect", marks=pytest.mark.xfail(reason="dask.array.pad bug")),
```

xarray/tests/test_variable.py
```diff
@@ -785,36 +785,69 @@ def test_getitem_error(self):
...
+            pytest.param("reflect", marks=pytest.mark.xfail),
+            "edge",
+            pytest.param("linear_ramp", marks=pytest.mark.xfail),
```
dcherian: `linear_ramp` passes locally with dask 2.9.1. Is the xfail needed for an older dask version? If so, let's mention that

xarray/tests/test_variable.py
```diff
@@ -2034,8 +2067,42 @@ def test_getitem_uint(self):
         super().test_getitem_fancy()
 
     @pytest.mark.xfail
-    def test_pad(self):
-        super().test_rolling_window()
+    @pytest.mark.parametrize(
+        "mode",
+        [
+            "mean",
+            pytest.param("median", marks=pytest.mark.xfail),
```
dcherian: Same as above.

xarray/core/dask_array_compat.py
```diff
@@ -97,6 +99,95 @@ def meta_from_array(x, ndim=None, dtype=None):
...
+    elif len(pad_width) == 2 and all(map(isint, pad_width)):
+        pad_width = sum(pad_width)
+    elif (
+        len(pad_width) == len(input_shape)
+        and all(map(lambda x: len(x) == 2, pad_width))
+        and all((isint(i) for p in pad_width for i in p))
+    ):
+        pad_width = np.sum(pad_width, axis=1)
+    else:
+        return  # should be impossible
```
dcherian: raise an error here in case it isn't impossible ;)

xarray/core/dask_array_compat.py
```diff
@@ -97,6 +99,95 @@ def meta_from_array(x, ndim=None, dtype=None):
...
+
+    if not np.array_equal(np.array(input_shape) + pad_width, output_shape):
+        raise RuntimeError(
+            "There seems to be something wrong with the shape of the output of dask.array.pad, "
+            "try upgrading Dask, use a different pad mode e.g. mode='constant' or first convert "
+            "your DataArray/Dataset to one backed by a numpy array by calling the `compute()` method."
```
dcherian: Let's link to the dask github issue

xarray/core/dask_array_compat.py
```diff
@@ -97,6 +99,95 @@ def meta_from_array(x, ndim=None, dtype=None):
...
+
+
+if LooseVersion(dask_version) >= LooseVersion("0.18.1"):
+
+    def pad(array, pad_width, mode="constant", **kwargs):
+        padded = da.pad(array, pad_width, mode=mode, **kwargs)
+        # workaround for inconsistency between numpy and dask: https://github.com/dask/dask/issues/5303
+        if mode == "mean" and issubclass(array.dtype.type, np.integer):
+            return da.round(padded).astype(array.dtype)
+        _validate_pad_output_shape(array.shape, pad_width, padded.shape)
```
dcherian: This is only needed if `mode == "reflect"`?

xarray/core/dask_array_compat.py
```diff
@@ -97,6 +99,95 @@ def meta_from_array(x, ndim=None, dtype=None):
...
+        )
+
+
+if LooseVersion(dask_version) >= LooseVersion("0.18.1"):
+
+    def pad(array, pad_width, mode="constant", **kwargs):
+        padded = da.pad(array, pad_width, mode=mode, **kwargs)
+        # workaround for inconsistency between numpy and dask: https://github.com/dask/dask/issues/5303
+        if mode == "mean" and issubclass(array.dtype.type, np.integer):
+            return da.round(padded).astype(array.dtype)
```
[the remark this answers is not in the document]
mark-boer: Thanks for the review. I'm not sure, when using numpy it does that implicitly. Although there is a chance that the casting from int to float to int will lead to rounding errors.

xarray/core/dataset.py
```diff
@@ -5686,5 +5686,158 @@ def map_blocks(
...
+        >>> ds.pad(x=(1,2))
+        <xarray.Dataset>
+        Dimensions:  (x: 8)
+        Dimensions without coordinates: x
+        Data variables:
+            foo      (x) float64 nan 0.0 1.0 2.0 3.0 4.0 nan nan
+        """
+        pad_width = either_dict_or_kwargs(pad_width, pad_width_kwargs, "pad")
+
+        if mode in ("edge", "reflect", "symmetric", "wrap"):
```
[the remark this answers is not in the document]
mark-boer: If the points are roughly linear one could use `reflect` with `reflect_type='odd'`. That will more or less do an extrapolation.

I could see that something like coords_mode being useful, but I'm not sure how much use this would get.

xarray/core/dask_array_compat.py
mark-boer: Yes, I have only seen the validation gone wrong when using the `mode=="reflect"`, but in principle the validation should always hold.

xarray/tests/test_variable.py
mark-boer: I totally agree that it looks weird using the `np.nan`, but that is the way numpy ints handle NaN's. `xr.shift` has  the same problem
```
>>> arr = xr.DataArray([1,2,3])
>>> arr.shift(dim_0=1, fill_value=np.nan)
<xarray.DataArray (dim_0: 3)>
array([-2147483648,           1,           2])
```

The xfails with the reason looks much better I agree :-)

dcherian: right that's how nan is represented with int. My point was that the output was cast to float in one case and not the other; and it's not very obvious why. Sorry, I should've been clearer.

mark-boer: It's the difference between `xarray.dtypes.NA` and `np.nan`.

mark-boer:
Hi @dcherian 

Once again an excellent code review :-)

I adressed most points in your review, except for the function signature of `DataArray.pad`. I think I prefer the default values of None. Now numpy/Dask will throw an error when you set invalid combinations e.g. `constant_values=0`, and set the `mode="mean"`. The np.nan is unfortunate, but this is the difference between np.nan and xarray.dtypes.NaN. I could set the default to `dtypes.NaN`, but then you lose the automatic error when setting an explicit value with `mode!="constant"`.

I can also see that something like coords_mode being useful, but I think it would be wise to wait and see how pad gets used and what issues we run into along the way.

I also ran into a small issue with pint (https://github.com/hgrecco/pint/issues/1026), but I'll get to that once this PR is merged.

dcherian:
I pushed some minor changes. 

I think this is ready to go in.

The big outstanding issue is what to do about dimension coordinates or indexes. Currently this PR treats all variables in `coords` different from those in `data_vars`.  I think this is confusing.

I am thinking that we want to use linear extrapolation for IndexVariables by default and apply the same padding mode to all other variables. The reasoning being that IndexVariables with NaNs are hard to deal with and it's hard to fill them in: `padded["x"] = padded.x.drop_vars("x").interpolate_na("x", fill_value="extrapolate")`

xarray/core/dataarray.py
```diff
@@ -3230,6 +3230,170 @@ def map_blocks(
 
         return map_blocks(func, self, args, kwargs)
 
+    def pad(
+        self,
+        pad_width: Mapping[Hashable, Union[int, Tuple[int, int]]] = None,
+        mode: str = "constant",
+        stat_length: Union[
+            None, int, Tuple[int, int], Mapping[Hashable, Tuple[int, int]]
+        ] = None,
```
max-sixty: FWIW you don't need to supply `None` in the union if the default is `None` (e.g. https://github.com/pydata/xarray/pull/3596/files#diff-ffd3597671590bab245b3193affa62b8R3246)

xarray/core/dataarray.py
```diff
@@ -3230,6 +3230,170 @@ def map_blocks(
...
+               [ 0.,  1.,  2.,  3.],
+               [10., 11., 12., 13.],
+               [nan, nan, nan, nan]])
+        Coordinates:
+          * x        (x) float64 nan 0.0 1.0 nan
+          * y        (y) int64 10 20 30 40
+            z        (x) float64 nan 100.0 200.0 nan
+        >>> da.pad(x=1, constant_values=np.nan)
+        <xarray.DataArray (x: 4, y: 4)>
+        array([[-9223372036854775808, -9223372036854775808, -9223372036854775808,
```
max-sixty: Is this intended?

max-sixty:
This looks excellent @mark-boer , thank you! 

I will try and have a proper look through today (but don't wait for me)

xarray/core/variable.py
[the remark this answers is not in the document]
max-sixty: FYI I hit similar issues, though less complex in https://github.com/pydata/xarray/pull/2470

xarray/core/dataset.py
```diff
@@ -5686,5 +5686,165 @@ def map_blocks(
...
+            One of ``pad_width`` or ``pad_width_kwargs`` must be provided.
+
+        Returns
+        -------
+        padded : Dataset
+            Dataset with the padded coordinates and data.
+
+        See also
+        --------
+        Dataset.shift, Dataset.roll, numpy.pad, dask.array.pad
```
max-sixty: We could add `ffill` & `bfill` here

xarray/core/dataset.py
```diff
@@ -5686,5 +5686,165 @@ def map_blocks(
...
+        mode: str = "constant",
+        stat_length: Union[
+            None, int, Tuple[int, int], Mapping[Hashable, Tuple[int, int]]
+        ] = None,
+        constant_values: Union[
+            None, int, Tuple[int, int], Mapping[Hashable, Tuple[int, int]]
+        ] = None,
+        end_values: Union[
+            None, int, Tuple[int, int], Mapping[Hashable, Tuple[int, int]]
+        ] = None,
```
max-sixty: I'm trying to think whether we can simplify the signature of the function; it's currently really complicated and may be difficult for people to easily read a call to `pad`. (tbc, not the literal python type signature; the number of different ways to call the function).

Any thoughts? We could remove supplying `int` alone and requie a tuple? But then doesn't agree with numpy. Or require a mapping for each argument, but that seems really verbose if people want the same values for each dim (is that a common case)?

OK if the answer is "it's as complicated as it needs to be"

mark-boer:
> The big outstanding issue is what to do about dimension coordinates or indexes. Currently this PR treats all variables in `coords` different from those in `data_vars`. I think this is confusing.

I do agree it can be confusing, but it is not unique in xarray. Dataset.shift only shifts `data_vars`, `bfill` and `ffill` only fill `data_vars`, etc.

Personally I think that extrapolating data coordinates without specifically settings a keyword or flag could also be confusing. I occasionally have data in my `coords` that make no sense to extrapolate. I do agree that adding an extrapolate_coord option or keyword of some sorts would be cool in the future. 

Both defaults could be really useful, I'm still in a bit of doubt. Are there any other people that might want to weigh in?

xarray/core/dataset.py
max-sixty: Yes I think you're right...

max-sixty:
> I do agree it can be confusing, but it is not unique in xarray. Dataset.shift only shifts `data_vars`, `bfill` and `ffill` only fill `data_vars`, etc.

I agree, I wouldn't have expected coords to be included given existing behavior. People can switch coords <> data_vars as needed, so there's an escape hatch

xarray/core/dataarray.py
max-sixty: Yes. I think it currently works well now, but we should probably have a different example: passing `np.nan` is a mistake, and instead users can rely on the default and it'll coerce to float with `dtypes.NA` implied. 

Is that right?

mark-boer: Yes. That looks to me like a reasonable default.

mark-boer:
> Edit: But it's more awkward for _indexes_ than non-index coords. The index becomes less useful with non-unique values, and generally indexes don't have nulls. I'm not sure what the other options would be: to some extent it's the intersection of `pad` with xarray's data model.

Hmm, I don't really see a solution. What do you suggest?

dcherian:
> Hmm, I don't really see a solution. What do you suggest?

:) I think we need to extrapolate indexes by default. It seems like the most sensible option.

mark-boer:
> :) I think we need to extrapolate indexes by default. It seems like the most sensible option.

Sorry, obviously that is a solution ;-). But I do have some concerns:

In some instances extrapolating all coords, can lead to some unwanted behaviour. Would you suggest we only interpolate the indexes?

How would we handle unsorted indexes? 

How would we extrapolate all the different kind of indexes, like the MultiIndex or CategoricalIndex?

max-sixty:
> In some instances extrapolating all coords, can lead to some unwanted behaviour. Would you suggest we only interpolate the indexes?
> 
> How would we handle unsorted indexes?
> 
> How would we extrapolate all the different kind of indexes, like the MultiIndex or CategoricalIndex?

I agree; I can't see easy solutions to these. If there are evenly spaced indexes (e.g. dates, grids), then it's easy to know what to do. But there are plenty of times it's difficult, if not intractable.

One option to merge something useful without resolving these questions is to return Variables only and label this method experimental. I think this is immediately useful for the cases where these difficult questions don't need to be answered. 

Of course if there are good answers to the questions, then even better!

dcherian:
Extremely good points @mark-boer 

I propose we merge and open an issue to decide what to do with IndexVariables.

max-sixty:
Agree!

We could add an "Experimental" label and then worry less about future changes

xarray/tests/test_units.py
```diff
@@ -2089,28 +2079,28 @@ def test_pad(self, dtype):
             pytest.param(unit_registry.m, None, id="identical_unit"),
         ),
     )
-    def test_pad_with_fill_value(self, unit, error, dtype):
+    def test_pad_unit_constant_value(self, unit, error, dtype):
```
keewis: I'm not sure we should do that, the only difference between both `pad` tests is the tested data. I realize that that's something that was already the case before this PR, so instead of touching the test above it might make sense to rename this test to `test_pad` and temporarily disable the other test by marking it with `skip`. I'll think a bit more about both tests and try to combine them in a new PR.

dcherian:
Merging. I've added an experimental warning to the docstrings and we can discuss the IndexVariable situation here: https://github.com/pydata/xarray/issues/3868

Thanks @mark-boer this is a major contribution for your first PR!

max-sixty:
Thanks @mark-boer ! Must be one of the largest first contributions...

+1 re merge + experimental warning; maybe we should do this more often. Cheers @dcherian

mark-boer:
Wow, didn't look for a week. Very happy 🎉! Thank you @dcherian  and @max-sixty  for all your input.

> Must be one of the largest first contributions...
I was inspired by the hacktoberfest last year. Took a little bit over a month though 😛

PR #3597, opened 2019-12-05
Add Facetgrid.row_labels & Facetgrid.col_labels

dcherian:
This allows labels to be changed later using `Facetgrid.set_titles`. We now save handles in `Facetgrid.row_labels` and `Facetgrid.col_labels`. I also added some API docs for facetgrid.

Example
```
g = darray.plot.imshow(row="row", col="col")
g.set_titles("abc={value}")
```

**before** (see row labels; the column labels work because those are set using `set_title`. )
before

**after**
after

 - [x] Tests added

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -625,7 +625,25 @@ Plotting
    plot.imshow
    plot.line
    plot.pcolormesh
+
+Faceting
+--------
+.. autosummary::
+   :toctree: generated/
+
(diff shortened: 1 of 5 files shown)


dcherian:
I'll merge tomorrow if there are no comments. It's a pretty safe change.

max-sixty:
LGTM!

PR #3584, opened 2019-12-01
Make dask names change when chunking Variables by different amounts.

dcherian:
When rechunking by the current chunk size, name should not change.
Add a `__dask_tokenize__` method for ReprObject so that this behaviour is present
when DataArrays are converted to temporary Datasets and back.

 - [x] Closes #3350
 - [x] Tests added
 - [x] Passes `black . && mypy . && flake8`
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

PR #3598, opened 2019-12-05
Fix map_blocks HLG layering

### LINKED ISSUE #3599: map_blocks graph construction bug
Just making a new issue for #3598 

---------------

The tests for https://github.com/pydata/xarray/pull/3584  fail on `dask == 2.8.1` with this interesting bug. Here's a reproducible example.

``` python
import dask
import xarray as xr

ds = xr.Dataset({'x': (('y',), dask.array.ones(10, chunks=(3,)))})
mapped = ds.map_blocks(lambda x: x)
mapped.compute()  # works

xr.testing.assert_equal(mapped, ds)  # does not work
xr.testing.assert_equal(mapped, ds.compute()) # works
xr.testing.assert_equal(mapped.compute(), ds)  # works
xr.testing.assert_equal(mapped.compute(), ds.compute())  # works
```

The traceback is
```
~/miniconda3/envs/dcpy/lib/python3.7/site-packages/dask/array/optimization.py in optimize(dsk, keys, fuse_keys, fast_functions, inline_functions_fast_functions, rename_fused_keys, **kwargs)
     41     if isinstance(dsk, HighLevelGraph):
     42         dsk = optimize_blockwise(dsk, keys=keys)
---> 43         dsk = fuse_roots(dsk, keys=keys)
     44 
     45     # Low level task optimizations

~/miniconda3/envs/dcpy/lib/python3.7/site-packages/dask/blockwise.py in fuse_roots(graph, keys)
    819             isinstance(layer, Blockwise)
    820             and len(deps) > 1
--> 821             and not any(dependencies[dep] for dep in deps)  # no need to fuse if 0 or 1
    822             and all(len(dependents[dep]) == 1 for dep in deps)
    823         ):

~/miniconda3/envs/dcpy/lib/python3.7/site-packages/dask/blockwise.py in <genexpr>(.0)
    819             isinstance(layer, Blockwise)
    820             and len(deps) > 1
--> 821             and not any(dependencies[dep] for dep in deps)  # no need to fuse if 0 or 1
    822             and all(len(dependents[dep]) == 1 for dep in deps)
    823         ):

KeyError: 'lambda-6720ab0e3639d5c63fc06dfc66a3ce47-x'
```

This key is not in `dependencies`. From https://github.com/dask/dask/blob/67fb5363009c583c175cb577776a4f2f4e811410/dask/blockwise.py#L816-L826 

``` python
    for name, layer in graph.layers.items():
        deps = graph.dependencies[name]
        if (
            isinstance(layer, Blockwise)
            and len(deps) > 1
            and not any(dependencies[dep] for dep in deps)  # no need to fuse if 0 or 1
            and all(len(dependents[dep]) == 1 for dep in deps)
        ):
            new = toolz.merge(layer, *[layers[dep] for dep in deps])
            new, _ = fuse(new, keys, ave_width=len(deps))
```

I'm not sure whether this is a bug in `fuse_roots`, `HighLevelGraph.from_collections` or in how `map_blocks` calls `HighLevelGraph.from_collections` here:
https://github.com/pydata/xarray/blob/69c85b85a1ef5d13b6ca51b6b655a1f719cc5abf/xarray/core/parallel.py#L315

TomAugspurger:
[x] closes #3599

This fixes an issue with the HighLevelGraph noted in
https://github.com/pydata/xarray/pull/3584, and exposed by a recent
change in Dask to do more HLG fusion.

cc @dcherian.

### DIFF
--- a/xarray/core/parallel.py
+++ b/xarray/core/parallel.py
@@ -7,12 +7,14 @@ try:
 except ImportError:
     pass
 
+import collections
 import itertools
 import operator
 from typing import (
     Any,
     Callable,
     Dict,
+    DefaultDict,
     Hashable,
     Mapping,
     Sequence,
@@ -222,6 +224,7 @@ def map_blocks(
     indexes.update({k: template.indexes[k] for k in new_indexes})
 
     graph: Dict[Any, Any] = {}
+    new_layers: DefaultDict[str, Dict[Any, Any]] = collections.defaultdict(dict)
     gname = "{}-{}".format(
         dask.utils.funcname(func), dask.base.tokenize(dataset, args, kwargs)
     )
@@ -310,9 +313,13 @@ def map_blocks(
                     # unchunked dimensions in the input have one chunk in the result
                     key += (0,)
 
-            graph[key] = (operator.getitem, from_wrapper, name)
+            new_layers[gname_l][key] = (operator.getitem, from_wrapper, name)
 
-    graph = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
+    hlg = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
+
+    for gname_l, layer in new_layers.items():
+        hlg.dependencies[gname_l] = {gname}
+        hlg.layers[gname_l] = layer
 
     result = Dataset(coords=indexes, attrs=template.attrs)
     for name, gname_l in var_key_map.items():
@@ -325,7 +332,7 @@ def map_blocks(
                 var_chunks.append((len(indexes[dim]),))
 
         data = dask.array.Array(
-            graph, name=gname_l, chunks=var_chunks, dtype=template[name].dtype
+            hlg, name=gname_l, chunks=var_chunks, dtype=template[name].dtype
         )
         result[name] = (dims, data, template[name].attrs)
 
--- a/xarray/core/parallel.py
+++ b/xarray/core/parallel.py
@@ -317,14 +317,18 @@ def map_blocks(
                     # unchunked dimensions in the input have one chunk in the result
                     key += (0,)
 
+            # We're adding multiple new layers to the graph:
+            # The first new layer is the result of the computation on
+            # the array.
+            # Then we add one layer per variable, which extracts the
+            # result for that variable, and depends on just the first new
+            # layer.
             new_layers[gname_l][key] = (operator.getitem, from_wrapper, name)
 
     hlg = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
 
     for gname_l, layer in new_layers.items():
-        # Ensure we have a valid HighLevelGraph.
         # This adds in the getitems for each variable in the dataset.
-        # This just depends on the layer we created earlier ("graph")
         hlg.dependencies[gname_l] = {gname}
         hlg.layers[gname_l] = layer
 
--- a/xarray/core/parallel.py
+++ b/xarray/core/parallel.py
@@ -223,6 +223,10 @@ def map_blocks(
     indexes = {dim: dataset.indexes[dim] for dim in preserved_indexes}
     indexes.update({k: template.indexes[k] for k in new_indexes})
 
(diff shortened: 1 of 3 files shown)


xarray/core/parallel.py
```diff
@@ -310,9 +313,13 @@ def _wrapper(func, obj, to_array, args, kwargs):
                     # unchunked dimensions in the input have one chunk in the result
                     key += (0,)
 
-            graph[key] = (operator.getitem, from_wrapper, name)
+            new_layers[gname_l][key] = (operator.getitem, from_wrapper, name)
 
-    graph = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
+    hlg = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
```
TomAugspurger: Changed to `hlg` primarily to please mypy, since we declare `graph` to be a Dict.

xarray/core/parallel.py
```diff
@@ -310,9 +313,13 @@ def _wrapper(func, obj, to_array, args, kwargs):
                     # unchunked dimensions in the input have one chunk in the result
                     key += (0,)
 
-            graph[key] = (operator.getitem, from_wrapper, name)
+            new_layers[gname_l][key] = (operator.getitem, from_wrapper, name)
```
dcherian: Can you add a comment about why these tasks should be in `new_layers` and not `hlg`?

xarray/core/parallel.py
```diff
@@ -310,9 +313,13 @@ def _wrapper(func, obj, to_array, args, kwargs):
...
 
-    graph = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
+    hlg = HighLevelGraph.from_collections(gname, graph, dependencies=[dataset])
+
+    for gname_l, layer in new_layers.items():
```
dcherian: A comment here would be useful.

xarray/tests/test_dask.py
```diff
@@ -1189,6 +1189,13 @@ def func(obj):
     assert_identical(expected.compute(), actual.compute())
 
 
+def test_map_blocks_hlg_layers():
+    ds = xr.Dataset({"x": (("y",), dask.array.ones(10, chunks=(5,)))})
```
dcherian: Let's add a comment saying "regression test for #3599"

xarray/tests/test_dask.py
```diff
@@ -1189,6 +1189,13 @@ def func(obj):
...
+    mapped = ds.map_blocks(lambda x: x)
+
+    xr.testing.assert_equal(mapped, ds)  # does not work
```
dcherian:
```suggestion
    xr.testing.assert_equal(mapped, ds) 
```

TomAugspurger:
Hopefully the new comments make sense. I'm struggling a bit to explain things since I don't fully understand them myself :)

> So it was a graph construction issue.

I *think* so. Dask doesn't actually validate arguments passed to HighLevelGraph. But I believe we assume that when all the values in `dependencies` are themselves keys of `layers`. We didn't have that before with things like

```
(Pdb) pp collections[0].dask.dependencies
{'all-84bc51ac43a9275b3662b0089710eab9': {'or_-64f95b81b2f8001b4c61f2023ac4c223'},
 ...
 'eq-abac622d95ce5055d3e7b7dea944ec37': {'lambda-e79de3edfa267f41111057d26471bce3-x',
                                         'ones-c4a83f4b990021618d55e0fa61a351d6'},
...
}
 ```

The `'lambda-e79de3edfa267f41111057d26471bce3-x'` wasn't a layer of the graph. It was previously nested under the single new layer we were creating `gname` or `lambda-e79de3edfa267f41111057d26471bce3` in this case.

dcherian:
> But I believe we assume that when all the values in dependencies are themselves keys of layers. 

I came to the same conclusion when I looked at it but didn't know if this was by accident or design. The docs don't say anything.

anyway the fix looks good. Thanks for tackling this.

I'll merge tomorrow if there are no other comments.

Issue #3594, opened 2019-12-04
support for units with pint


keewis:
`pint`'s implementation of NEP-18 (see hgrecco/pint#905) is close enough so we can finally start working on the `pint` support (i.e. make the integration tests pass).

Failing tests:
* top level functions
  - [ ] `align`: needs more investigation
  - [ ] `combine_by_coords`: needs more investigation
  - [ ] `combine_nested`: needs more investigation
  - [ ] `concat`: needs more investigation
  - [ ] `merge`: needs more investigation
  - [ ] `full_like`: `pint` currently does not implement `copyto`
  - [ ] `where`: needs more investigation
  - [ ] `dot`: `pint` currently does not implement `einsum`
* DataArray methods
  - [ ] `__init__`: indexes strip units, blocked by #1603?
  - [ ] aggregation: `pint` currently does not implement `all`, `any`, `median`, `prod`, `sum`. `cumprod` works but is not allowed for quantities with dimensions (update the test?)
  - [ ] binary operations: matrix multiply (not implemented by `pint`, yet: hgrecco/pint#865)
  - [ ] bivariate ufuncs: `pint` does not implement `maximum`
  - [ ] numpy methods: `pint` does not implement `rank`
  - [ ] `ffill`, `bfill`: needs more investigation
  - [ ] `fillna`: calls `astype` on fill value which may be a python scalar wrapped by `pint.Quantity` and needs more investigation
  - [ ] `isin`: `pint` does not implement `isin`
  - [ ] `where`: bug in the test code and needs more investigation
  - [ ] `interpolate_na`: needs more investigation
  - [ ] `combine_first`: needs more investigation
  - [ ] `equals`, `identical`: needs more investigation
  - [ ] `broadcast_like`: needs more investigation
  - [ ] `drop_sel`: test bug / wrong method (use `drop_vars`)
  - [ ] `astype`: needs more investigation
  - [ ] `item`: needs more investigation
  - [ ] `drop_sel`: needs more investigation
  - [ ] `sel`, `loc`: indexes strip units (#1603)
  - [ ] `squeeze`: needs more investigation
  - [ ] `head`, `tail`, `thin`: quantity not accepted as indexer
  - [ ] `interp_like`: needs more investigation
  - [ ] `reindex_like`: needs more investigation
  - [ ] `to_unstacked_dataset`: needs more investigation
  - [ ] `shift`, `roll`: needs more investigation
[...]

PR #3600, opened 2019-12-05
Fix pint integration tests

keewis:
This cleans up the tests which became a bit of a mess. Once that is done, we can actively work on fixing #3594.

 - [x] Tests added

### DIFF
--- a/xarray/tests/test_units.py
+++ b/xarray/tests/test_units.py
@@ -2210,17 +2210,17 @@ class TestDataArray:
             isinstance(raw_values, (int, float)) and array.check(unit)
         ):
             with pytest.raises(error):
-                data_array.loc[values]
+                data_array.loc[{"x": values}]
 
             return
 
         expected = attach_units(
             strip_units(data_array).loc[
-                strip_units(convert_units(values, {None: array.units}))
+                {"x": strip_units(convert_units(values, {None: array.units}))}
             ],
             extract_units(data_array),
         )
-        result = data_array.loc[values]
+        result = data_array.loc[{"x": values}]
         assert_equal_with_units(expected, result)
 
     @pytest.mark.parametrize(
--- a/xarray/tests/test_units.py
+++ b/xarray/tests/test_units.py
@@ -2260,27 +2260,12 @@ class TestDataArray:
             result = data_array.squeeze(dim=name)
             assert_equal_with_units(expected, result)
 
-    @pytest.mark.xfail(
-        reason="indexes strip units and head / tail / thin only support integers"
-    )
-    @pytest.mark.parametrize(
-        "unit,error",
-        (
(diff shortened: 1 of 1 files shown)


jthielen:
In general, this looks good!

After running through this, a couple thoughts did come to mind:

- Do you have tests of `align`/`combine_*`/`concat` that have Quantity data but no Quantity coordinates? I think these could help isolate the coordinate/indexing issue from the any other operation-related issue.
- Would you be able to alter the `cumprod` tests to test a dimensionless Quantity? `np.cumprod` is only supported (and only really makes sense) for dimensionless Quantities
- Could you change the xfail reason for matrix multiplication to pint not supporting `einsum`? `__matmul__` has been implemented in https://github.com/hgrecco/pint/pull/905, but xarray's `__matmul__` appears to route through `np.einsum`.
- Would it make sense to remove the `np.rank` test as discussed here: https://github.com/hgrecco/pint/pull/905#issuecomment-562109451?

keewis:
`np.rank` is removed in one of the yet-to-be-pushed commits.

`align`/ etc.: not yet, but I can change the `variants` dict (replace `original_unit` with `1` if you want to play with it)

`cumprod`, `matmul`: sure, will do

max-sixty:
Looks like a good set of improvements, thanks @keewis 

Shall we merge this and do flups in another PR?

keewis:
I'd first like to go through the Dataset tests which probably won't need as much updates as the DataArray tests since I had more experience when I wrote them.

What do you mean with "flups"?

max-sixty:
> What do you mean with "flups"?

Sorry, that's probably not widely used! "flups"="follow-ups"

keewis:
I'm done with revisiting the tests so this is ready for review & merge.

@jthielen I think I applied most of the changes you requested. `cumprod` fails using xarray's structures (works fine with `Quantity`), so the xfail is still necessary. I also didn't isolate the dims from any tests other than `align` and I think we still need at least one more iteration to get full isolation (I'd like to leave that to a new PR, though -- maybe when we attempt to silence the `UnitStrippedWarning`s?).

jthielen:
>  `cumprod` fails using xarray's structures (works fine with `Quantity`), so the xfail is still necessary. I also didn't isolate the dims from any tests other than `align` and I think we still need at least one more iteration to get full isolation (I'd like to leave that to a new PR, though -- maybe when we attempt to silence the `UnitStrippedWarning`s?).

Sounds like a good plan.

max-sixty:
Looks good! (as ever with these, I had a browse through and picked a couple of sections to look at more deeply, but didn't go through each line in detail)

max-sixty:
And test failure is unrelated (and think there's a PR to fix)

keewis:
thanks for reviewing and merging, @max-sixty

max-sixty:
Thank you @keewis !

PR #3601, opened 2019-12-06
Fix contourf set under

### LINKED ISSUE #3590: cmap.set_under() does not work as expected
When using matplotlib, the `set_under()` method can be used to set values below the range of a colormap to a certain color, for example
```
import matplotlib
from matplotlib import pyplot
import numpy

dat = numpy.linspace(0, 1)[numpy.newaxis, :]*numpy.linspace(0, 1)[:, numpy.newaxis]

cmap = matplotlib.cm.viridis
#cmap.set_under('w')

pyplot.contourf(dat, vmin=.3, cmap=cmap)
pyplot.colorbar()
pyplot.show()
```
produces
![matplotlib_no-set_under](https://user-images.githubusercontent.com/3958036/70076161-195ec200-15f6-11ea-94ab-59f1b38e9d25.png)

while uncommenting the `cmap.set_under()` call produces
![matplotlib](https://user-images.githubusercontent.com/3958036/70076218-36939080-15f6-11ea-9601-93d3a52b6e20.png)

However, using `xarray` to do the same thing,
```
import matplotlib
from matplotlib import pyplot
from xarray import DataArray
import numpy

da = DataArray(numpy.linspace(0, 1)[numpy.newaxis, :]*numpy.linspace(0, 1)[:, numpy.newaxis])

cmap = matplotlib.cm.viridis
cmap.set_under('w')

da.plot.contourf(vmin=.3, cmap=cmap)
pyplot.show()
```
produces
![xarray](https://user-images.githubusercontent.com/3958036/70076465-bc174080-15f6-11ea-9f59-37f20569cf6e.png)
where it seems the call to `cmap.set_under('w')` had no effect. Expected behaviour would be output like the second plot.

<details>

```
In [2]: xarray.show_versions()                                                                                                                                                                                                                

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
python-bits: 64
OS: Linux
OS-release: 5.0.0-37-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8
libhdf5: 1.10.0
libnetcdf: 4.6.0

xarray: 0.14.1
pandas: 0.24.2
numpy: 1.16.3
scipy: 1.2.1
netCDF4: 1.3.1
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: None
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.1.0
distributed: None
matplotlib: 3.1.1
cartopy: None
seaborn: None
numbagg: None
setuptools: 41.0.1
pip: 19.3.1
conda: None
pytest: 4.4.1
IPython: 7.6.1
sphinx: None
```

</details>

johnomotani:
Copies the `cmap._rgba_bad`, `cmap._rgba_under`, and `cmap._rgba_over` values to `new_cmap`, in case they have been set to non-default values. Allows the user to customize plots more by using matplotlib methods on a `cmap` before passing as an argument to xarray's plotting methods. Previously these settings were overridden by defaults when creating the `cmap` actually used to make the plot.

I'm not a fan of copying attributes one-by-one like this, but I guess this is an issue with matplotlib's API, unless there's a nicer way to convert a `cmap` to a discrete `cmap` than `mpl.colors.from_levels_and_colors()`.

 - [x] Closes #3590
 - [x] Tests added


xarray/plot/utils.py
```diff
@@ -98,6 +98,12 @@ def _build_discrete_cmap(cmap, levels, extend, filled):
     # copy the old cmap name, for easier testing
     new_cmap.name = getattr(cmap, "name", cmap)
 
+    # copy colors to use for bad, under, and over values in case they have been set to
+    # non-default values
+    new_cmap._rgba_bad = getattr(cmap, "_rgba_bad")
```
keewis: this fails if `cmap` is a string (same for the two lines directly below). Try adding a default:
```suggestion
    new_cmap._rgba_bad = getattr(cmap, "_rgba_bad", None)
```

johnomotani: Good point. I tested using `None` and it changes the default behaviour slightly, so I've added a fix getting the default values from `new_cmap._rgba_bad`, etc., instead. Then if `cmap` is a string, we continue to do exactly the same thing as was done before this PR.

xarray/tests/test_plot.py
```diff
@@ -267,6 +268,51 @@ def test2d_1d_2d_coordinates_contourf(self):
         a.plot.contourf(x="time", y="depth")
         a.plot.contourf(x="depth", y="time")
 
+    def test_contourf_cmap_set_bad(self):
```
dcherian: can we consolidate these into one test where you change all three parameters?

xarray/plot/utils.py
```diff
@@ -98,6 +98,12 @@ def _build_discrete_cmap(cmap, levels, extend, filled):
     # copy the old cmap name, for easier testing
     new_cmap.name = getattr(cmap, "name", cmap)
 
+    # copy colors to use for bad, under, and over values in case they have been set to
+    # non-default values
+    new_cmap._rgba_bad = getattr(cmap, "_rgba_bad", new_cmap._rgba_bad)
```
dcherian: can you raise an issue on the matplotlib repo and ask if there is a better way of doing this?

johnomotani:
Not sure why some tests are failing... The ones I can see are to do with `CFTime`. I can't see a relation to the changes in this PR??

keewis:
that's #3673, you can ignore these `upstream-dev` failures

xarray/plot/utils.py
dcherian: See https://github.com/mwaskom/seaborn/pull/1948

johnomotani: The implementation in mwaskom/seaborn#1948 of the same work-around that we are adding here looks nicer and more robust than the way I have implemented it. Will update to follow the same pattern. Not sure if the issue with under/over colors needing to change if they were not explicitly set by the user applies to this case, but in any case I think it is better to handle it in case we ever need to use the same fix somewhere else.

johnomotani: Done in de6ffeb and 47a6fc0

xarray/plot/utils.py
```diff
@@ -98,6 +98,30 @@ def _build_discrete_cmap(cmap, levels, extend, filled):
     # copy the old cmap name, for easier testing
     new_cmap.name = getattr(cmap, "name", cmap)
 
+    # copy colors to use for bad, under, and over values in case they have been
+    # set to non-default values
+    try:
+        # matplotlib<3.2 only uses bad color for masked values
+        bad = cmap(np.ma.masked_invalid([np.nan]))[0]
```
mathause: Can't you do

```suggestion
        bad = cmap(np.ma.masked_invalid(np.nan))
```
?

xarray/tests/test_plot.py
```diff
@@ -267,6 +268,35 @@ def test2d_1d_2d_coordinates_contourf(self):
         a.plot.contourf(x="time", y="depth")
         a.plot.contourf(x="depth", y="time")
 
+    def test_contourf_cmap_set(self):
+        sz = (4, 4)
```
mathause:
```suggestion
        shape = (4, 4)
```
I find this easier to read

xarray/tests/test_plot.py
```diff
@@ -267,6 +268,35 @@ def test2d_1d_2d_coordinates_contourf(self):
...
+        a = DataArray(easy_array(sz), dims=["z", "time"])
```
mathause:
```suggestion
        a = DataArray(easy_array(shape), dims=["z", "time"])
```

xarray/tests/test_plot.py
```diff
@@ -267,6 +268,35 @@ def test2d_1d_2d_coordinates_contourf(self):
...
+
+        cmap.set_under("w")
+        # check we actually changed the set_under color
+        assert cmap._rgba_under != mpl.cm.viridis._rgba_under
+
+        cmap.set_over("w")
+        # check we actually changed the set_over color
+        assert cmap._rgba_over != mpl.cm.viridis._rgba_over
+
+        pl = a.plot.contourf(cmap=copy(cmap))
```
mathause: Why do you apply `copy` here? I think it is unnecessary.

xarray/plot/utils.py
johnomotani: It seems like I can. I'd just used the same form as mwaskom/seaborn#1948, but don't know why that form was used originally...

xarray/tests/test_plot.py
johnomotani: I'd copied the `sz = (4,4)` from the test just above this one... Actually the variable is not really necessary as it's only used once. Probably better to just change here to
```suggestion
        a = DataArray(easy_array((4, 4)), dims=["z", "time"])
```

xarray/plot/utils.py
mathause: Maybe because they use `npt.assert_array_equal` to test for equality - `cmap(np.ma.masked_invalid([np.nan]))` returns an numpy array and `cmap(np.ma.masked_invalid(np.nan))` a tuple.

mathause:
I think (some of) the test failures are unrelated (`test_seaborn_palette_as_cmap`), see #3747

xarray/plot/utils.py
johnomotani: Checking this, I get an error from this
```
>>> cmap = copy(mpl.cm.viridis)
>>> cmap(numpy.ma.masked_invalid(numpy.nan))                               
/home/john/.local/lib/python3.6/site-packages/matplotlib/colors.py:509: UserWarning: Warning: converting a masked element to nan.
  xa = np.array([X])
```
and I get different results from the original and the changed version:
```
In [11]: cmap(numpy.ma.masked_invalid(numpy.nan))                               
/home/john/.local/lib/python3.6/site-packages/matplotlib/colors.py:509: UserWarning: Warning: converting a masked element to nan.
  xa = np.array([X])
Out[11]: (0.267004, 0.004874, 0.329415, 1.0)
```
```
In [13]: cmap(numpy.ma.masked_invalid([numpy.nan]))[0]                          
Out[13]: array([0., 0., 0., 0.])
```
And the second one here (the original form from mwaskom/seaborn#1948) is correct
```
In [10]: cmap._rgba_bad                                                         
Out[10]: (0.0, 0.0, 0.0, 0.0)
```
so I will get rid of the commit changing this.

mathause: Ok, weird - sorry for sending you down the wrong path!

johnomotani: Slightly worrying: the incorrect change didn't cause a test fail. I'll add another test case that would have caught this.

xarray/tests/test_plot.py
johnomotani: The idea is that since we test the output against `cmap` we want to be totally sure `cmap` isn't changed by `a.plot.contourf()` (even though it should not be changed anyway). Actually, I think we should `deepcopy` instead of `copy` to achieve this.

xarray/tests/test_plot.py
```diff
@@ -267,6 +268,50 @@ def test2d_1d_2d_coordinates_contourf(self):
         a.plot.contourf(x="time", y="depth")
         a.plot.contourf(x="depth", y="time")
 
+    def test_contourf_cmap_set(self):
+        a = DataArray(easy_array((4, 4)), dims=["z", "time"])
+
+        cmap = deepcopy(mpl.cm.viridis)
```
dcherian: let's add a comment saying why we are deepcopying.

xarray/tests/test_plot.py
```diff
@@ -267,6 +268,50 @@ def test2d_1d_2d_coordinates_contourf(self):
...
+
+        pl = a.plot.contourf(cmap=deepcopy(cmap))
+
+        # check the set_bad color
+        assert pl.cmap._rgba_bad == cmap._rgba_bad
```
dcherian: let's do `assert pl.cmap(0) == cmap(0)` and so on. That way we are not relying on private attributes.

johnomotani: I think that may actually fix (some of?) the current test failures - apparently the private attributes can sometimes be `None`, and the test fails in that case at the moment.

johnomotani:
I've added an extra test that checks the right thing happens when we do *not* change the bad, under or over colors. That test needs to set `vmin` and `vmax` to get sensible defaults for the under and over colors: without `vmin` or `vmax`, `mpl.colors.from_levels_and_colors()` is called with `extend = 'neither'`; then it sets `_rgba_under` and `_rgba_over` in its result to `(0.0, 0.0, 0.0, 0.0)` (but these values can never be used, because with no `vmin` or `vmax` there can be no data outside of the plotted range).

I think this is ready now. The unit test failures on my local computer appear to be unrelated to this PR (see earlier references to #3673 by @keewis and #3747 by @mathause).

doc/whats-new.rst
```diff
@@ -41,6 +41,9 @@ Bug fixes
 - :py:func:`concat` can now handle coordinate variables only present in one of
   the objects to be concatenated when ``coords="different"``.
   By `Deepak Cherian <https://github.com/dcherian>`_.
+- xarray now respects the over, under and bad colors if set on a provided colormap.
+  (:issue:`3590`, :pull:`3601`)
+  By `John Motani <https://github.com/johnomotani>`_.
```
```suggestion
  By `johnomotani <https://github.com/johnomotani>`_.
```

dcherian:
Thanks @johnomotani

PR #3602, opened 2019-12-07
silence sphinx warnings round 3

### LINKED ISSUE #3370: Hundreds of Sphinx errors

keewis:
In this last "sphinx warnings" PR the goal is to silence all nit-picky warnings that are not related to napoleon's interpretation of parameter types. ~In https://github.com/pydata/xarray/issues/3370#issuecomment-562582121 I posted ways to define type aliases (so `dict-like` points to the term `mapping` on https://docs.python.org/3/ and `array-like` to the appropriate page in the numpy docs) or to ignore words like `of`. This PR applies these to silence all the nit-picky warnings (which mostly means broken links).~

~As a reference for myself, the documentation of numpydoc's sphinx config options is here: https://numpydoc.readthedocs.io/en/latest/install.html#sphinx-config-options~ `numpydoc` does not have anything to do with this, we are blocked by a bug in `napoleon` (see #3370).

At the moment only the autodoc / autosummary / ~numpydoc~ napoleon warnings remain, with a few exceptions in `whats-new.rst`.

In theory we could also re-enable `-n` and use the `nitpick_ignore` settings to ignore any unfixable warnings, but I'm undecided about whether that would be a good idea. Thoughts?

 - [x] Closes #3370 


keewis:
I added a lot of new entries to `api-hidden.rst`, mostly the methods of `CFTimeIndex` that were inherited from `pandas.Index`. Is that okay? `sphinx` also complains about the attributes (properties?) so I would probably need to add more.

running `sphinx-build` with `-w warnings` and executing `grep -e "doc/[^/]*.rst" warnings` returns
```
.../xarray/doc/whats-new.rst:336: WARNING: py:meth reference target not found: Dataset.plot.scatter
.../xarray/doc/whats-new.rst:421: WARNING: py:meth reference target not found: Dataset.plot.scatter
.../xarray/doc/whats-new.rst:468: WARNING: py:class reference target not found: xarray.core.coordinates.DataArrayCoordinates
.../xarray/doc/whats-new.rst:1059: WARNING: py:meth reference target not found: xarray.DataArray.plot.line
.../xarray/doc/whats-new.rst:1091: WARNING: py:meth reference target not found: xarray.tutorial.open_dataset
.../xarray/doc/whats-new.rst:1091: WARNING: py:meth reference target not found: xarray.tutorial.load_dataset
.../xarray/doc/whats-new.rst:1778: WARNING: py:meth reference target not found: DataArray.__dask_scheduler__
.../xarray/doc/whats-new.rst:1780: WARNING: py:meth reference target not found: DataArray.plot.imshow
.../xarray/doc/whats-new.rst:1957: WARNING: py:func reference target not found: xarray.show_versions
.../xarray/doc/whats-new.rst:2029: WARNING: py:meth reference target not found: xarray.backends.PydapDataStore.open
```
these are the broken links in the non-autodoc / autosummary documentation:
* if I remember correctly, `scatter`, `line` and `imshow` not having a documentation is known and the link will probably be fixed once we got sphinx to document what I think @dcherian called "injection method".
* I'd say `__dask_scheduler__` is not part of the public API so we can convert to double-backtick quoted
* `DataArrayCoordinates` and `DatasetCoordinates` still exist, but they are never referenced by the docs. Are they part of the public API?
* `show_versions` does not have a docstring (which could be really small) and is not referenced elsewhere in the docs. What should I do with it?
* the tutorial functions do have docstrings. Should we mention them somewhere else?
* all the datastore documentation pages have broken links to their methods. Should I also add them to `api-hidden.rst`?

doc/api-hidden.rst
```diff
@@ -27,13 +27,66 @@
    Dataset.std
    Dataset.var
 
+   core.rolling.DatasetCoarsen
```
dcherian: Can you add this to `api.rst` with a top-level heading "Coarsen objects" like "GroupBy objects" etc.

doc/api-hidden.rst
```diff
@@ -27,13 +27,66 @@
...
+
    core.groupby.DatasetGroupBy.assign
    core.groupby.DatasetGroupBy.assign_coords
    core.groupby.DatasetGroupBy.first
    core.groupby.DatasetGroupBy.last
    core.groupby.DatasetGroupBy.fillna
    core.groupby.DatasetGroupBy.quantile
    core.groupby.DatasetGroupBy.where
+   core.groupby.DatasetGroupBy.all
```
dcherian: I actually think these methods should not be hidden. They are visible if you navigates to `DataArrayGroupBy` and scroll down to find methods but that's totally not obvious.

doc/api-hidden.rst
```diff
@@ -91,6 +196,80 @@
    DataArray.cumprod
    DataArray.rank
 
+   Variable.all
```
dcherian: Hiding Variable methods is a good idea I think

doc/api-hidden.rst
```diff
@@ -156,6 +335,94 @@
    plot.FacetGrid.set_ticks
    plot.FacetGrid.map
 
+   CFTimeIndex.all
```
dcherian: Not sure about these but perhaps they should be hidden until our Index API is finalized and public.

doc/interpolation.rst
```diff
@@ -128,7 +128,7 @@ It is now possible to safely compute the difference ``other - interpolated``.
 Interpolation methods
 ---------------------
 
-We use :py:func:`scipy.interpolate.interp1d` for 1-dimensional interpolation and
+We use :py:class:`scipy.interpolate.interp1d` for 1-dimensional interpolation and
```
dcherian:
```suggestion
We use :py:meth:`scipy.interpolate.interp1d` for 1-dimensional interpolation and
```
?

doc/io.rst
```diff
@@ -47,7 +47,7 @@ read/write netCDF V4 files and use the compression options described below).
 __ https://github.com/Unidata/netcdf4-python
 
 We can save a Dataset to disk using the
-:py:meth:`~Dataset.to_netcdf` method:
+:py:meth:`~xarray.Dataset.to_netcdf` method:
```
dcherian: I think this and the remaining "xarray" changes can be avoided by adding `.. currentmodule:: xarray` at the top of the file. We should just add this at the top of every .rst file I think

doc/pandas.rst
```diff
@@ -65,7 +65,7 @@ For datasets containing dask arrays where the data should be lazily loaded, see
 
 To create a ``Dataset`` from a ``DataFrame``, use the
 :py:meth:`~xarray.Dataset.from_dataframe` class method or the equivalent
-:py:meth:`pandas.DataFrame.to_xarray <DataFrame.to_xarray>` method:
+:py:meth:`~pandas.DataFrame.to_xarray` method:
```
dcherian:
```suggestion
:py:meth:`pandas.DataFrame.to_xarray` method:
```

so that it's obvious that this is a pandas method, not a xarray method.

doc/plotting.rst
```diff
@@ -227,7 +227,7 @@ It is required to explicitly specify either
 
 Thus, we could have made the previous plot by specifying ``hue='lat'`` instead of ``x='time'``.
 If required, the automatic legend can be turned off using ``add_legend=False``. Alternatively,
-``hue`` can be passed directly to :py:func:`xarray.plot` as `air.isel(lon=10, lat=[19,21,22]).plot(hue='lat')`.
+``hue`` can be passed directly to :py:func:`xarray.plot.plot` as `air.isel(lon=10, lat=[19,21,22]).plot(hue='lat')`.
```
dcherian:
```suggestion
``hue`` can be passed directly to :py:meth:`xarray.plot.line` as `air.isel(lon=10, lat=[19,21,22]).plot.line(hue='lat')`.
```

doc/terminology.rst
```diff
@@ -19,7 +19,7 @@ Terminology
 
 .. note::
 
-    The :py:class:`Variable` class is low-level interface and can typically be ignored. However, the word "variable" appears often enough in the code and documentation that is useful to understand.
+    The :py:class:`~xarray.Variable` class is low-level interface and can typically be ignored. However, the word "variable" appears often enough in the code and documentation that is useful to understand.
```
dcherian: Let's stick `.. currentmodule:: xarray` at the top instead

doc/interpolation.rst
keewis: `interp1d` is a class pretending to be a function:
```python
>>> type(scipy.interpolate.interp1d)
<class 'scipy.interpolate.interpolate.interp1d'>
```
so it is in the `py:class` section in `objects.inv`. To investigate, using
```bash
python -m sphinx.ext.intersphinx https://docs.scipy.org/doc/scipy/reference/objects.inv
```
helps a lot.

doc/api-hidden.rst
keewis: the other methods except from `map` and `reduce` are listed under `grouped operations`. I'd say increasing the visibility of these is out of scope and I'd like to leave that to a new PR / issue where we can have a focused discussion. The current changes just make sure a documentation page is created, so navigating to the `DataArrayGroupBy` page does not expose broken links.

doc/api-hidden.rst
keewis: without these, the methods are still visible from the `CFTimeIndex` page but the links are broken. We can probably push the decision of where to put them back until the index refactoring is done.

doc/api-hidden.rst
dcherian: :+1:

doc/api-hidden.rst
dcherian: Yes they can stay in api-hidden

keewis:
thanks for the review, @dcherian.

The `Data*Coordinates` objects are returned by the `coords` attributes of `DataArray` and `Dataset`. They are mostly treated as `dict` but they have a few public (i.e. not prefixed with an underscore) properties (e.g. `dims`), all of which don't have docstrings. I'll treat both classes as implementation detail.

Other than that, only the injection methods and the datastore methods remain for the manually written part. Unfortunately, the autodoc / autosummary warnings are due to a sphinx bug (see https://github.com/pydata/xarray/issues/3370#issuecomment-564124368), so our only option right now is to use `nitpicky_ignore` to silence them. I don't think we should, though. Instead, I'll try to silence all warnings that are not due to the bug and we can leave the rest to a potential round 4 or a sphinx / napoleon update.

dcherian:
> The Data*Coordinates objects are returned by the coords attributes of DataArray and Dataset

Ah! I was wrong. These are public and have public methods that should be documented: see https://xarray.pydata.org/en/stable/data-structures.html#coordinates-methods

dcherian:
Is this ready to merge?

keewis:
no, not yet. There are a lot of warnings left even if I leave out the ones due to the parameter type.

keewis:
this is fairly close now:
```bash
grep -e "doc/[^/]*.rst" -e "<autosummary>" warnings
```
gives us
```
.../xarray/doc/generated/xarray.Dataset.rst:132:<autosummary>:1: WARNING: py:obj reference target not found: xarray.Dataset.dump_to_store
.../xarray/doc/generated/xarray.Dataset.rst:132:<autosummary>:1: WARNING: py:obj reference target not found: xarray.Dataset.load_store
.../xarray/doc/whats-new.rst:346: WARNING: py:meth reference target not found: Dataset.plot.scatter
.../xarray/doc/whats-new.rst:431: WARNING: py:meth reference target not found: Dataset.plot.scatter
.../xarray/doc/whats-new.rst:1068: WARNING: py:meth reference target not found: xarray.DataArray.plot.line
.../xarray/doc/whats-new.rst:1789: WARNING: py:meth reference target not found: DataArray.plot.imshow
```
which means only the injected methods and the `DataStore` related `Dataset` methods are left. If they are still supported, where should I put them?

keewis:
I think once we decided what to do with the `DataStore` methods and how to treat the injected methods, we can merge?

dcherian:
> A grep over the code tells me that load_store is never used, tested or mentioned (from_store is mentioned but never defined or used): do we still support it? 

I guess `from_store` mentions should be removed; let's open an issue for removing `load_store`. I don't know the backend code but it'd gone through multiple refactors so things may have been missed.

> dump_to_store is mentioned and tested, so we definitely should put it somewhere (even if it is just api-hidden.rst).

Let's stick this in `api-hidden.rst`

I think you should merge this after making these changes. It's a large step forward. We can revisit once someone comes up with a solution for the injected methods.

keewis:
I renamed `from_store` to `load_store`, let's discuss both in a new issue. For the injected / accessor methods there is #3625.

dcherian:
:+1: Wanna merge?

keewis:
on it

dcherian:
:clap: Welcome to xarray!

PR #3607, opened 2019-12-10
Strided rolling

niowniow:
This PR adds a stride parameter to the rolling function of DataArray and Dataset . It basically extends the stride functionality being available for `core.rolling.DataArrayRolling.construct`  and `core.rolling.DatasetRolling.construct` to the other methods of `DataArrayRolling` and `DatasetRolling`.

Note: it makes the arguments of `DataArrayRolling` and `DatasetRolling` inconsistent with the respective `rolling` arguments of pandas Series and DataFrame (They do not support stride).
Moreover, it does not solve the issue addressed in [this pandas issue](https://github.com/pandas-dev/pandas/issues/15354) (Efficient stride computation).

 - [x] Tests added


fujiisoup:
@niowniow 
Thank you for your contribution!

I think `stride` option is a good idea.
One thing is how to implement this efficient nan-reduction method.

Currently, we use 'bottleneck' if it is installed for speeding up nan-ops, but bottleneck does not support `stride` option.
Another problem is inefficiency of nan-ops of numpy for strided arrays; he copies the strided array into full array and replace np.nan by zero before the reduction.

One way we could do is 
1. skip using 'bottleneck' if `stride` is other than 1
2. implement our nan-ops for rolling.
For example, for `nansum`, we can replace `np.nan` by 0 *before* creating the strided arrays and apply usual `sum` for the strided array.

In `rolling.count`, we did a similar thing.

headtr1ck:
Is that still wanted?
It seems like a useful feature to have.

Issue #3483, opened 2019-11-04
assign_coords with mixed DataArray / array args removes coords


keewis:
I'm not sure if using `assign_coords` to overwrite the data of coords is the best way to do so, but using mixed args (on current master) turns out to have surprising results:
```python
>>> obj = xr.DataArray(
...     data=[6, 3, 4, 6],
...     coords={"x": list("abcd"), "y": ("x", range(4))},
...     dims="x",
... )
>>> obj
<xarray.DataArray 'obj' (x: 4)>
array([6, 3, 4, 6])
Coordinates:
  * x        (x) <U1 'a' 'b' 'c' 'd'
    y        (x) int64 0 1 2 3
>>> # works as expected
>>> obj.assign_coords(coords={"x": list("efgh"), "y": ("x", [0, 2, 4, 6])})
<xarray.DataArray 'obj' (x: 4)>
array([6, 3, 4, 6])
Coordinates:
  * x        (x) <U1 'e' 'f' 'g' 'h'
    y        (x) int64 0 2 4 6
>>> # works, too (same as .data / .values)
>>> obj.assign_coords(coords={
...     "x": obj.x.copy(data=list("efgh")).variable,
...     "y": ("x", [0, 2, 4, 6]),
... })
>>> # this drops "y"
>>> obj.assign_coords(coords={
...     "x": obj.x.copy(data=list("efgh")),
...     "y": ("x", [0, 2, 4, 6]),
... })
<xarray.DataArray 'obj' (x: 4)>
array([6, 3, 4, 6])
Coordinates:
  * x        (x) <U1 'e' 'f' 'g' 'h'
```
Passing a `DataArray` for `y`, like `obj.y * 2` while also changing `x` (the type does not matter) always results in a `MergeError`:
```python
>>> obj.assign_coords(x=list("efgh"), y=obj.y * 2)
xarray.core.merge.MergeError: conflicting values for index 'x' on objects to be combined:
first value: Index(['e', 'f', 'g', 'h'], dtype='object', name='x')
second value: Index(['a', 'b', 'c', 'd'], dtype='object', name='x')
```

I would expect the result to be the same regardless of the type of the new coords.

dcherian:
The `x=DataArray` example looks like a bug.

[...]

PR #3611, opened 2019-12-11
Pint support for top-level functions

keewis:
This PR tries to get the pint integration tests (see #3594) to pass. To make sure this does not become a giant PR, it is limited to the top-level functions and the work on the tests for `DataArray` and `Dataset` will be done in other PRs.

I added pint master to the `upstream-dev` CI since that way CI finally helps finding bugs. Should I have edited `py36-min-nep18` instead?

Also, I finally ran into a situation where I have to use `assert_allclose` instead of `assert_equal_with_units`. However, `assert_allclose` does not check for identical units, I think? Would it make sense to copy the code of `assert_allclose` and add the strict unit checking or would it be better to just use `assert_allclose` and something like `assert extract_units(expected) == extract_units(actual)`?

Failing tests list from #3594:
  - `align`: ~needs more investigation~ ~partially blocked by `where`, `IndexVariable` and a bug in `assert_equal_with_units` and / or the tests, also a behaviour decision on align~ works (but units in `IndexVariable` still fail)
  - `combine_by_coords`: ~needs more investigation~ ~partially blocked by `IndexVariable` and a test bug (due to #3483)~ works (but units in `IndexVariable` still fail)
  - `combine_nested`: ~needs more investigation~ ~blocked by `reindex`~ works (but units in `IndexVariable` still fail)
  - `concat`: ~needs more investigation~ works (but units in `IndexVariable` still fail)
  - `merge`: ~needs more investigation~ works (but units in `IndexVariable` still fail)
  - `full_like`: ~`pint` currently does not implement `copyto`~ blocked by hgrecco/pint#882
  - `where`: ~needs more investigation~ ~unit is stripped because `pint` does not implement `astype` and test bug in `test_where_dataset`~ works
  - `dot`: ~`pint` currently does not implement `einsum`~ works


keewis:
it seems the changes to `Variable._getitem_with_mask` will make other tests fail. The reason for the changes is that pint will convert everything to the first value's units, which would convert the array to the fill value's units and would be quite unexpected, I think?. Does anyone know what to do here?

keewis:
the `_getitem_with_mask` failures are due to `bool` not defining `__invert__`, which means it falls back to `int.__invert__`. I special-cased it, but we could also modify `indexing.create_mask` to always return a duck array.

xarray/tests/test_units.py
```diff
@@ -527,12 +550,17 @@ def test_align_dataset(fill_value, unit, variant, error, dtype):
...
+        expected_b = attach_units(expected_b, units_b)
 
     actual_a, actual_b = func(ds1, ds2)
 
-    assert_equal_with_units(expected_a, actual_a)
-    assert_equal_with_units(expected_b, actual_b)
+    assert_allclose(expected_a, actual_a)
+    assert extract_units(expected_a) == extract_units(actual_a)
+    assert_allclose(expected_b, actual_b)
+    assert extract_units(expected_b) == extract_units(actual_b)
```
keewis: I'm using a combination of `assert_allclose` + `extract_units` and `assert_identical` now. The output of `assert_allclose` is not really ideal because only the values (no labels and no units) are printed. The output of `assert_identical` is also not ideal because the `repr` of quantities prints the units after the values. Since long strings get truncated, the unit is almost always not visible.

xarray/tests/test_units.py
```diff
@@ -542,28 +570,49 @@ def test_broadcast_dataarray(dtype):
...
-    (actual,) = xr.broadcast(ds)
 
-    assert_equal_with_units(expected, actual)
+    units_a = extract_units(ds)
+    units_b = extract_units(other)
+    expected_a, expected_b = xr.broadcast(strip_units(ds), strip_units(other))
+    expected_a = attach_units(expected_a, units_a)
+    expected_b = attach_units(expected_b, units_b)
+
+    actual_a, actual_b = xr.broadcast(ds, other)
```
keewis: this aligns the test with the `DataArray` version (i.e. two arguments to `broadcast`) but might not change the coverage.

xarray/tests/test_units.py
```diff
@@ -915,24 +958,39 @@ def test_merge_dataarray(variant, unit, error, dtype):
     )
 
     func = function(xr.merge)
-    if error is not None:
+    if error is not None and variant != "data":
```
keewis: this is necessary because `merge` requires named `DataArray` objects and puts them into a `Dataset`, which means the units don't have to be compatible.

xarray/tests/test_units.py
```diff
@@ -355,7 +365,7 @@ def test_apply_ufunc_dataarray(dtype):
     expected = attach_units(func(strip_units(data_array)), extract_units(data_array))
     actual = func(data_array)
 
-    assert_equal_with_units(expected, actual)
+    assert_identical(expected, actual)
```
dcherian: So `assert_identical` checks units too? Let's add that to the docstring.

xarray/tests/test_units.py
dcherian: > assert_allclose is not really ideal because only the values (no labels and no units) are printed. 

Yeah this needs to be updated to pretty-print the diffs like assert_equal.

> The output of assert_identical is also not ideal because the repr of quantities prints the units after the values. Since long strings get truncated, the unit is almost always not visible.

There was some talk about adding units explicitly to the repr. Seems like this would be a good thing to do early and will help you debug these tests more easily. Can you open an issue or PR for that?

xarray/tests/test_units.py
dcherian: :+1:

xarray/tests/test_units.py
```diff
@@ -915,24 +958,39 @@ def test_merge_dataarray(variant, unit, error, dtype):
     )
 
     func = function(xr.merge)
```
dcherian: Is this wrapping with `function` needed here?

xarray/tests/test_units.py
```diff
@@ -1021,10 +1091,11 @@ def test_replication_dataarray(func, dtype):
     data_array = xr.DataArray(data=array, dims="x")
 
     numpy_func = getattr(np, func.__name__)
-    expected = xr.DataArray(data=numpy_func(array), dims="x")
+    units = extract_units(numpy_func(data_array))
```
dcherian: would it help to abstract this away to a function that
1. extracts units
2. applies a function to unit-less objects
3. reattaches units

xarray/core/variable.py
```diff
@@ -739,7 +739,11 @@ def _getitem_with_mask(self, key, fill_value=dtypes.NA):
 
             data = as_indexable(self._data)[actual_indexer]
             mask = indexing.create_mask(indexer, self.shape, data)
-            data = duck_array_ops.where(mask, fill_value, data)
+            if isinstance(mask, bool):
```
dcherian: seems ok to me. @shoyer ?

xarray/tests/test_units.py
```diff
@@ -7,14 +8,18 @@
...
 pint = pytest.importorskip("pint")
 DimensionalityError = pint.errors.DimensionalityError
 
-
 unit_registry = pint.UnitRegistry()
 Quantity = unit_registry.Quantity
 
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    Quantity([])
```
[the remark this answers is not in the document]
keewis: `pint` warns about the changed behaviour for quantities with `numpy` (due to the implementation of NEP-18). This fragment is the recommended way to ignore this warning.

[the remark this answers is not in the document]
dcherian: Let's add a comment saying that .

xarray/tests/test_units.py
keewis: well, `identical` seems to check the units, so `assert_identical` does, too. I didn't change anything, using a `pint` version with the changes from hgrecco/pint#905 was enough.

xarray/tests/test_units.py
keewis: no, I think not. I'll remove it.

xarray/tests/test_units.py
keewis: in general, yes, but there are three variations of this pattern:
1. one where we have to precompute using pint quantities to extract units
2. two where we have to convert before stripping and
  a. then reattach the extracted units
  b. then convert back to the old units

I think I would probably create a huge mess with that function, so it might be better to leave that to a new PR?

xarray/tests/test_units.py
keewis: I think I can do the formatting for `assert_allclose`, but I don't know where to start for the repr for units. Maybe I should open a new issue for that?

xarray/tests/test_units.py
dcherian: :+1: to new PR but only if you think it will help.

xarray/tests/test_units.py
dcherian: >  Maybe I should open a new issue for that?

let's talk about it on #2773

max-sixty:
Thanks for the review @dcherian !

@keewis FYI I am starting a new job so will be lying low on xarray for a couple of months, forgive the temporary pause in speedy reviews. It was great to work together for the past few months! The whole team is looking forward to working more closely with you in the future. We'd like to invite you to be a maintainer: could you drop me an email at `m` at `maxroos` period `com`?

max-sixty:
Thanks for the review @dcherian !

@keewis FYI I am starting a new job so will be lying low on xarray for a couple of months, forgive the temporary pause in speedy reviews. It was great to work together for the past few months! 

The whole team is looking forward to working more closely with you in the future. We'd like to invite you to be a maintainer: could you drop me an email at `m` at `maxroos` period `com`?

keewis:
I added a `whats-new.rst` entry, but I'm not sure if it is too minimal.

If this is okay for now, we could merge?

keewis:
This should be ready for review / merge once again.

It seems that after the merge of #3706 all that was left to get the top-level function tests to pass was to clean up those tests.

max-sixty:
Looks good! 
(as ever with me & this section of code, without being that confident)

dcherian:
Yeah, thanks @keewis

PR #3615, opened 2019-12-12
Minor docstring fixes

DangoMelon:
Really minor docstring fixes, just added 's' at the end of `kwarg` and deleted ` Default n = 5` from `thin` method's docstring since it doesn't have a default value.


dcherian:
Thanks @DangoMelon

PR #3618, opened 2019-12-13
ENH: enable `H5NetCDFStore` to work with already open h5netcdf.File a…

kmuehlbauer:
enable `H5NetCDFStore` to work with open h5netcdf.File and h5netcdf.Group objects, add test

 - [x] Tests added

### DIFF
--- a/xarray/backends/h5netcdf_.py
+++ b/xarray/backends/h5netcdf_.py
@@ -4,9 +4,9 @@ import numpy as np
 
 from .. import Variable
 from ..core import indexing
-from ..core.utils import FrozenDict
+from ..core.utils import FrozenDict, is_remote_uri
 from .common import WritableCFDataStore
-from .file_manager import CachingFileManager
+from .file_manager import CachingFileManager, DummyFileManager
 from .locks import HDF5_LOCK, combine_locks, ensure_lock, get_write_lock
 from .netCDF4_ import (
     BaseNetCDF4Array,
@@ -69,8 +69,42 @@ class H5NetCDFStore(WritableCFDataStore):
     """Store for reading and writing data via h5netcdf
     """
 
+    __slots__ = (
+        "autoclose",
+        "format",
+        "is_remote",
+        "lock",
+        "_filename",
+        "_group",
+        "_manager",
+        "_mode",
+    )
+
     def __init__(
-        self,
+            self, manager, group=None, mode=None, lock=HDF5_LOCK, autoclose=False
+    ):
+
+        import h5netcdf
+
+        if isinstance(manager, (h5netcdf.File, h5netcdf.Group)):
+            if group is None:
+                root, group = manager._root, manager.name
+            else:
+                root = manager
+            manager = DummyFileManager(root)
+
+        self._manager = manager
+        self._group = group
+        self._mode = mode
+        self.format = None
+        self._filename = self.ds._root.filename
+        self.is_remote = is_remote_uri(self._filename)
+        self.lock = ensure_lock(lock)
+        self.autoclose = autoclose
+
+    @classmethod
+    def open(
+        cls,
         filename,
         mode="r",
         format=None,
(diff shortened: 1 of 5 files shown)


kmuehlbauer:
I'm unsure how this should be referenced in the docs/news because the netCDF4 counterpart doesn't have any mention about that feature too.

It seems that everything runs smoothly, the one error looks like it is unrelated. Please let me know, if the added test is enough and if you want me to do anything else. I'll leave the `WIP` mark in place so far.

Ping @shoyer, he might see if there are immediate problems with the code.

kmuehlbauer:
Happy New Year to everyone especially at pydata.

I'd be happy to actively follow any reviews to sort things out and to get this addition pulled in. Many thanks!

xarray/backends/h5netcdf_.py
```diff
@@ -69,8 +69,40 @@ class H5NetCDFStore(WritableCFDataStore):
...
+        "_mode",
+    )
+
+    def __init__(self, manager, group=None, mode=None, lock=HDF5_LOCK, autoclose=False):
+
+        import h5netcdf
+
+        if isinstance(manager, (h5netcdf.File, h5netcdf.Group)):
+            if group is None:
+                root, group = manager._root, manager.name
```
shoyer: I think it's probably a better idea to use the `find_root_and_group` helper function that we use for netCDF4. Accessing private attributes like `_root` is a little unsafe.

xarray/backends/h5netcdf_.py
```diff
@@ -69,8 +69,40 @@ class H5NetCDFStore(WritableCFDataStore):
...
+            else:
+                root = manager
+            manager = DummyFileManager(root)
+
+        self._manager = manager
+        self._group = group
+        self._mode = mode
+        self.format = None
+        self._filename = self.ds._root.filename
```
shoyer: Likewise, we don't want to rely on a private API here

xarray/backends/h5netcdf_.py
kmuehlbauer: #3680 needs to be resolved first. I appreciated any guidance on that blocking issue.

xarray/backends/h5netcdf_.py
kmuehlbauer: Unfortunately `h5netcdf.Group` doesn't seem to have a `filename` property like `netCDF4.Dataset`. So there is currently no direct means of extracting the filename from `h5netcdf.Group` using public API. Trying to to facilitate `find_root_and_group` now.

kmuehlbauer:
@TomNicholas After commit #1689db493f10262555196f658c52e370aacb4a33 I get mypy TypCheck error:

xarray/core/dataset.py:3610: error: Item "Mapping[...]" of "Union[Dataset, Mapping[...]]" has no attribute "to_dataset"

Should I create an issue on that?

kmuehlbauer:
@shoyer I tried to address your review comments. I've added a temporary (backwards compatible) fix for #3680 to get along. Also using `find_root_and_group` to get hold of the filename is some kind of workaround until there is a better option.

I've added an item to `whats-new.rst`. For the API docs I did not find a fitting position because `H5NetCDFStore` and `NetCDF4DataStore` are nowhere mentioned there. So this might better be added in another PR by experienced xarray devs.

Unfortunately mypy checks are breaking due to some recent change (PR #3677).

kmuehlbauer:
@shoyer Please let me know if this needs rebasing after #3690 gets merged, or if I can anything do to finalize this.

kmuehlbauer:
Took the chance to rebase on latest master.

kmuehlbauer:
@dcherian Following up on #3702 it would be really great if this enhancement could make it into 0.15.0 If I can be of any further assistance please ping me.

shoyer:
This looks great, thanks for figuring this out!

PR #3619, opened 2019-12-13
Provide shape info in shape mismatch error.

danielballan:
While it is easy enough to obtain this info from `%debug`, putting it
in the error messages is helpful when those messages come from user
reports (e.g. emails).

I am guessing that this change does not call for a new test or a
documentation update, but I'm happy to add one if asked.

### DIFF
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -346,7 +346,10 @@ class Variable(
     def data(self, data):
         data = as_compatible_data(data)
         if data.shape != self.shape:
-            raise ValueError("replacement data must match the Variable's shape")
+            raise ValueError(
+                f"replacement data must match the Variable's shape. "
+                f"data.shape is {data.shape}; Variable.shape is {self.shape}"
+            )
         self._data = data
 
     def load(self, **kwargs):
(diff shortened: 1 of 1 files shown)


xarray/core/variable.py
```diff
@@ -346,7 +346,10 @@ def data(self):
     def data(self, data):
         data = as_compatible_data(data)
         if data.shape != self.shape:
-            raise ValueError("replacement data must match the Variable's shape")
+            raise ValueError(
+                f"replacement data must match the Variable's shape. "
+                f"data.shape is {data.shape}; Variable.shape is {self.shape}"
```
dcherian:
```suggestion
                f"replacement data has shape {data.shape}; Variable has shape {self.shape}"
```

dcherian:
Thanks @danielballan. Nicer error messages are always welcome! I've made one small suggestion.

Also I see that this is your first xarray PR. Welcome! We hope to see you contribute often.

danielballan:
Thanks! I agree that your proposed wording is better and I have committed it.

Thanks for the friendly welcome. I learned OSS as a minor pandas contributor back in my grad schools days, now glad to finally have a contribution in xarray. I love this project!

dcherian:
:) It's great to have you contribute.

PR #3636, opened 2019-12-17
Support swap_dims to dimension names that are not existing variables

### LINKED ISSUE #1855: swap_dims should support dimension names that are not existing variables

```python
input_ds = xarray.Dataset({'foo': ('x', [1, 2])}, {'x': [0, 1]})
input_ds.swap_dims({'x': 'z'})
```
#### Problem description

Currently this results in the error `KeyError: 'z'`

#### Expected Output

We now support dimensions without associated coordinate variables. So `swap_dims()` should be able to create new dimensions (e.g., `z` in this example) even if there isn't already a coordinate variable.

keewis:
- [x] Closes #1855
 - [x] Tests added


dcherian:
Thanks @keewis

Issue #789, opened 2016-03-10
Time limitation (between years 1678 and 2262) restrictive to climate community


pwolfram:
The restriction of

>    One unfortunate limitation of using datetime64[ns] is that it limits the native representation of dates to
>    those that fall between the years 1678 and 2262. When a netCDF file contains dates outside of these
>    bounds, dates will be returned as arrays of netcdftime.datetime objects.

is a potential roadblock inhibiting easy adoption of this library in the climate community.

pwolfram:
@jhamman and @shoyer, do you know if more general support for times will be provided?

I would say this is a potential roadblock inhibiting easy adoption of this library within the climate community. For example, this is creating problems when we setup and analyze climate models (cc @akturner, @douglasjacobsen, @milenaveneziani). There are obvious work arounds, but they are hacks, e.g., I just add 1700 or some arbitrary year within my wrapper to MPAS at https://github.com/pwolfram/mpas_xarray_wrapper. However, this does not appear to be a viable long-term solution.

Does anyone have some advice on how to better deal with this problem?

pwolfram:
See also https://github.com/pydata/xarray/issues/531 (cc @jsbj, @darothen, @jhamman); https://github.com/pydata/xarray/issues/521 (cc @rabernat, @jhamman)

It appears this is an issue with the numpy `datetime64` API (https://github.com/numpy/numpy/issues/6207).

shoyer:
There are two issues here:

1. Support for years outside 1678-2262 -- blocked by pandas standardizing on nanosecond precision.
2. Support for custom calendars -- blocked by limitations of numpy's datetime64.

Unfortunately, I don't see easy fixes to either of these, though if I had to guess, adding support for other datetime precisions (perhaps only sub-second resolution) to pandas would be easier than fixing up NumPy's datetime64 itself (which is already pretty hacky).

milenaveneziani:
Thank you, @pwolfram, for bringing this up. I just wanted to add that this limitation is important for analyzing so-called fixed-conditions climate model experiments, in which the model is run with fixed greenhouse gases conditions for a particular year (say, fixed levels of CO2 concentration representative of levels for the year 1850, for example). In these cases, time is simply measured with respect to start of simulation (year 0, or year 1) and each year is a no-leap year of 365 days.

pwolfram:
[...]

PR #3642, opened 2019-12-18
Make datetime_to_numeric more robust to overflow errors

huard:
- [x] Closes #3641 
 - [x] Tests added

This is likely only safe with NumPy>=1.17 though.

### DIFF
--- a/xarray/core/duck_array_ops.py
+++ b/xarray/core/duck_array_ops.py
@@ -372,43 +372,62 @@ def _datetime_nanmin(array):
 
 
 def datetime_to_numeric(array, offset=None, datetime_unit=None, dtype=float):
-    """Convert an array containing datetime-like data to an array of floats.
+    """Return an array containing datetime-like data to numerical values.
+
+    Convert the datetime array to a timedelta relative to an offset.
 
     Parameters
     ----------
-    da : np.array
-        Input data
-    offset: Scalar with the same type of array or None
-        If None, subtract minimum values to reduce round off error
-    datetime_unit: None or any of {'Y', 'M', 'W', 'D', 'h', 'm', 's', 'ms',
-        'us', 'ns', 'ps', 'fs', 'as'}
-    dtype: target dtype
+    da : array-like
+      Input data
+    offset: None, datetime or cftime.datetime
+      Datetime offset. If None, this is set by default to the array's minimum
+      value to reduce round off errors.
+    datetime_unit: {None, Y, M, W, D, h, m, s, ms, us, ns, ps, fs, as}
+      If not None, convert output to a given datetime unit. Note that some
+      conversions are not allowed due to non-linear relationships between units.
+    dtype: dtype
+      Output dtype.
 
     Returns
     -------
     array
+      Numerical representation of datetime object relative to an offset.
+
+    Notes
+    -----
+    Some datetime unit conversions won't work, for example from days to years, even
+    though some calendars would allow for them (e.g. no_leap). This is because there
+    is no `cftime.timedelta` object.
     """
     # TODO: make this function dask-compatible?
+    # Set offset to minimum if not given
     if offset is None:
         if array.dtype.kind in "Mm":
             offset = _datetime_nanmin(array)
         else:
             offset = min(array)
+
+    # Compute timedelta object.
+    # For np.datetime64, this can silently yield garbage due to overflow.
     array = array - offset
 
-    if not hasattr(array, "dtype"):  # scalar is converted to 0d-array
+    # Scalar is converted to 0d-array
+    if not hasattr(array, "dtype"):
         array = np.array(array)
 
+    # Convert timedelta objects to timedelta64[ms] dtype.
     if array.dtype.kind in "O":
-        # possibly convert object array containing datetime.timedelta
-        array = np.asarray(pd.Series(array.ravel())).reshape(array.shape)
+        array = array.astype("timedelta64[ms]")
 
+    # Convert to specified timedelta units.
     if datetime_unit:
         array = array / np.timedelta64(1, datetime_unit)
 
-    # convert np.NaT to np.nan
+    # Convert np.NaT to np.nan
     if array.dtype.kind in "mM":
         return np.where(isnull(array), np.nan, array.astype(dtype))
+
     return array.astype(dtype)
 
 
--- a/xarray/tests/test_duck_array_ops.py
+++ b/xarray/tests/test_duck_array_ops.py
@@ -669,10 +669,22 @@ def test_datetime_to_numeric_datetime64():
     expected = 24 * np.arange(0, 35, 7).astype(dtype)
     np.testing.assert_array_equal(result, expected)
 
+    # Test overflow. Need to convert to ms first otherwise we get garbage
+    # TODO: make datetime_to_numeric more robust to overflow errors.
+    offset = np.datetime64("0001-01-01")
+    ms = times.astype("datetime64[ms]")
+    result = duck_array_ops.datetime_to_numeric(
+        ms, offset=offset, datetime_unit="D", dtype=int
+    )
+    expected = 730119 + np.arange(0, 35, 7)
+    np.testing.assert_array_equal(result, expected)
+
 
 @requires_cftime
 def test_datetime_to_numeric_cftime():
-    times = cftime_range("2000", periods=5, freq="7D").values
+    import cftime
+
+    times = cftime_range("2000", periods=5, freq="7D", calendar="standard").values
     result = duck_array_ops.datetime_to_numeric(times, datetime_unit="h")
     expected = 24 * np.arange(0, 35, 7)
     np.testing.assert_array_equal(result, expected)
@@ -685,4 +697,12 @@ def test_datetime_to_numeric_cftime():
     dtype = np.float32
     result = duck_array_ops.datetime_to_numeric(times, datetime_unit="h", dtype=dtype)
     expected = 24 * np.arange(0, 35, 7).astype(dtype)
+    np.testing.assert_array_equal(result, expected)
+
+    # Test potential overflow (outputs are different from timedelta64. Expected ?)
+    offset = cftime.DatetimeGregorian(1, 1, 1)
+    result = duck_array_ops.datetime_to_numeric(
+        times, offset=offset, datetime_unit="D", dtype=int
+    )
+    expected = 730121 + np.arange(0, 35, 7)
     np.testing.assert_array_equal(result, expected)
--- a/xarray/tests/test_interp.py
+++ b/xarray/tests/test_interp.py
@@ -662,3 +662,10 @@ def test_datetime_interp_noerror():
         coords={"time": pd.date_range("01-01-2001", periods=50, freq="H")},
     )
     a.interp(x=xi, time=xi.time)  # should not raise an error
+
+
+@requires_cftime
+def test_3641():
+    times = xr.cftime_range("0001", periods=3, freq="500Y")
+    da = xr.DataArray(range(3), dims=["time"], coords=[times])
+    da.interp(time=["0002-05-01"])


xarray/core/duck_array_ops.py
```diff
@@ -372,43 +372,62 @@ def _datetime_nanmin(array):
...
     # TODO: make this function dask-compatible?
+    # Set offset to minimum if not given
     if offset is None:
         if array.dtype.kind in "Mm":
             offset = _datetime_nanmin(array)
         else:
             offset = min(array)
+
+    # Compute timedelta object.
+    # For np.datetime64, this can silently yield garbage due to overflow.
```
spencerkclark: Ah good catch; this is another bug (we end up with a NaN in this example):
```
In [1]: import numpy as np; import xarray as xr

In [2]: times = [np.datetime64('1690-01-01', 'ns'), np.datetime64('2200-01-01', 'ns')]

In [3]: da = xr.DataArray(range(2), dims=['time'], coords=[times])

In [4]: da
Out[4]:
<xarray.DataArray (time: 2)>
array([0, 1])
Coordinates:
  * time     (time) datetime64[ns] 1690-01-01 2200-01-01

In [5]: da.interp(time=['1700-01-01'])
Out[5]:
<xarray.DataArray (time: 1)>
array([nan])
Coordinates:
  * time     (time) datetime64[ns] 1700-01-01
```
This is another potential plus of using a universal offset.  In the case of NumPy dates, I think if we always use 1970-01-01T00:00:00 as the offset we are guaranteed this will never return garbage.

xarray/core/duck_array_ops.py
```diff
@@ -372,43 +372,62 @@ def _datetime_nanmin(array):
...
-    if not hasattr(array, "dtype"):  # scalar is converted to 0d-array
+    # Scalar is converted to 0d-array
+    if not hasattr(array, "dtype"):
         array = np.array(array)
 
+    # Convert timedelta objects to timedelta64[ms] dtype.
     if array.dtype.kind in "O":
-        # possibly convert object array containing datetime.timedelta
-        array = np.asarray(pd.Series(array.ravel())).reshape(array.shape)
+        array = array.astype("timedelta64[ms]")
```
spencerkclark: Here we may want to use microsecond units (`"timedelta64[us]"`); using millisecond resolution may lead to unnecessary truncation of precision.

huard: Will do. In `get_clean_interp_index`, should we also use `datetime_to_numeric` for pd.DatetimeIndex with the us resolution ?

```
    if isinstance(index, (CFTimeIndex, pd.DatetimeIndex)):
        offset = type(index[0])(1970, 1, 1)
        index = Variable(
            data=datetime_to_numeric(index, offset=offset, datetime_unit="us"),
            dims=(dim,),
        )
```

xarray/core/duck_array_ops.py
huard: We'd also have to enforce a microsecond resolution, which could break some code.

spencerkclark: For the issue I raised in the comment above, I think switching to 1970-01-01T00:00:00 as the offset would work for NumPy dates at nanosecond resolution (maybe there is something I'm missing?).  Let me think a little more about the implications of casting the `datetime.timedelta` objects to `timedelta64[us]` objects, though.  I think you're right that we should think carefully about that.

xarray/core/duck_array_ops.py
huard: Switching to "us" breaks `interpolate_na` because max_gap is ns by default. 

Throughout xarray, "ns" is the default resolution. I'm a bit worried that switching this for "us" in a couple of places it a recipe for future bugs.

spencerkclark: It occurs to me that we *could* use nanosecond units (or units even finer) for cftime dates, so long as we do the unit conversion step after converting to floats.  E.g. one could write a function like the following to aid in that:
```python
SUB_MICROSECOND_UNIT_CONVERSION_FACTORS = {
    "ns": 1e3,
    "ps": 1e6,
    "fs": 1e9,
    "as": 1e12
}

def _py_timedelta_to_float(array, datetime_unit="ns"):
    array = array.astype("timedelta64[us]")
    if datetime_unit in SUB_MICROSECOND_UNIT_CONVERSION_FACTORS:
        factor = SUB_MICROSECOND_UNIT_CONVERSION_FACTORS[datetime_unit]
        return factor * array.astype(np.float64)
    else:
        return array / np.timedelta64(1, datetime_unit)
```
I have not thought carefully yet about how to handle `max_gap` in `interpolate_na` in the case of cftime dates, but this seems like a possible compromise to allow us to continue to default to nanosecond resolution internally for these functions.  @huard do you think that could help?

spencerkclark: Here's perhaps a cleaner version:
```python
def _py_timedelta_to_float(array, datetime_unit="ns"):
    array = array.astype("timedelta64[us]").astype(np.float64)
    conversion_factor = np.timedelta64(1, "us") / np.timedelta64(1, datetime_unit)
    return conversion_factor * array
```

dcherian: > max_gap in interpolate_na in the case of cftime dates,

It should just work if you transform `max_gap` in the same way you transform the coordinate.

huard: I don't have a complete enough picture to have a strong opinion on this. My feeling is that on the long term, xarray would probably want to let users pick their own time resolution. So for example when you call `xr.cftime_range`, you would have a `dtype` or `precision` argument that would be used every time objects need to be converted to numerical values. 

One potential issue with `_py_timedelta_to_float` is that we won't be able to convert back the float to the original timedelta without first casting to int, and I would suppose a loss of precision. Also you'd have to remember the original resolution. My sense is that using datetime64 and timedelta64 as the default numerical dtypes would be less error prone. 

I have a naive fix for max_gap in PR #3631 simply switching the resolution to us. Want me to merge both PRs into one ?

spencerkclark: > My feeling is that on the long term, xarray would probably want to let users pick their own time resolution. 

Can you elaborate on what the example use-case of this would be?  Fundamentally, without upstream changes, the underlying date objects will always be of either nanosecond or microsecond resolution.

While not exactly a "resolution" option, we do already expose `datetime_unit` arguments in relevant functions, e.g. in [differentiate](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.differentiate.html) and [integrate](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.integrate.html).  Is that along the lines of what you meant?

> One potential issue with _py_timedelta_to_float is that we won't be able to convert back the float to the original timedelta without first casting to int, and I would suppose a loss of precision. Also you'd have to remember the original resolution. 

I believe one could say the same thing about the current implementation of `datetime_to_numeric` (it ultimately produces unitless floats too).  How would you suggest rewriting that?

huard: > Can you elaborate on what the example use-case of this would be? Fundamentally, without upstream changes, the underlying date objects will always be of either nanosecond (mainly a constraint of pandas) or microsecond (cftime) resolution.

Agreed, but I assume upstream libraries will eventually leverage datetime64 and timedelta64 objects and become more flexible in terms of their internal resolution. This would let people use xarray to look at femtoseconds experiments in optics for example. 

> While not exactly a "resolution" option, we do already expose `datetime_unit` arguments in relevant functions, e.g. in [differentiate](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.differentiate.html) and [integrate](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.integrate.html). Is that along the lines of what you meant?

Yes, but I feel like this resolution should be a property of the underlying object. 

> I believe one could say the same thing about the current implementation of `datetime_to_numeric` (it ultimately produces unitless floats too). How would you suggest rewriting that?

Correct. I'm wondering if using timedelta64 and datetime64 objects would be cleaner. I'm not saying this should be done now though.

spencerkclark: Those are all good big-picture discussion points @huard.  I'll briefly add my two cents, but I think we should probably get back to the matter at hand.  It would be great to fix these interpolation bugs.

> Agreed, but I assume upstream libraries will eventually leverage datetime64 and timedelta64 objects and become more flexible in terms of their internal resolution. This would let people use xarray to look at femtoseconds experiments in optics for example.

Yes, that would be nice.  We've been down this road to some extent before; [we've discussed](https://github.com/pydata/xarray/issues/789) (see also [this mailing list thread](https://groups.google.com/forum/#!topic/pydata/kk04maBGw1U)) potentially making changes upstream in pandas to allow us to use coarser resolution datetime64 types in DatetimeIndexes for long time series, but the situation there is complicated: https://github.com/pandas-dev/pandas/issues/7307.  I'm afraid I'm not aware of any ongoing effort to change things.

> Yes, but I feel like this resolution should be a property of the underlying object.

It sounds like in this context by "resolution" you are referring to the units (as opposed to a truncation in precision), which could be more tractable.  This perhaps could be something we could discuss within the context of making xarray data structures/operations units-aware (e.g. through `pint`).  There is a fair bit of ongoing work in this area; see #3594 for a jumping off point.

> Correct. I'm wondering if using timedelta64 and datetime64 objects would be cleaner. I'm not saying this should be done now though.

I see.  Eventually for operations like interpolation and differentiation I think we need to convert to floats.  We're for sure open to ideas for how to improve things though.

huard: So happy to get back on target and fix this bug. What both PRs do (and I could merge them into one) is enforce the "us" precision in places where we need a numerical representation resolution in `duck_array_ops.datetime_to_numeric`, the converting to float: 
```
if array.dtype.kind in "O":
     array = array.astype("timedelta64[us]")
...
return array.astype(dtype) 
```
where dtype is `float` by default. 

I leverage that in `missing.get_clean_interp_index` with a new if clause:
```
    if isinstance(index, (CFTimeIndex, pd.DatetimeIndex)):
        offset = type(index[0])(1970, 1, 1)
        index = Variable(
            data=datetime_to_numeric(index, offset=offset, datetime_unit="us"),
            dims=(dim,),
        )
...
index = index.values.astype(np.float64)
```
This however break `interpolate_na`, so I convert `max_gap` in `missing.interp_na` to "us" before converting to float. 
```
max_gap = np.timedelta64(max_gap, "us").astype(np.float64)
```
So this change would break people's code that assume `datetime_to_numeric` and `get_clean_interp_index` return "ns" based floats. If that's unacceptable, then we could use your suggested `_py_timedelta_to_float`.

dcherian: @huard These are private functions so that's not a concern. We do want to maintain compatibility with numpy>= 0.14 though

spencerkclark: Yes.  Keep in mind though the default units for `differentiate` and `integrate` have been nanoseconds, and that is user-facing; should we preserve that?  Also should we preserve the ability to specify sub-microsecond `max_gap`'s in `interpolate_na`?

dcherian: ah I see. Thanks @spencerkclark 

> Keep in mind though the default units for differentiate and integrate have been nanoseconds, and that is user-facing; should we preserve that?

Yes I think we should. Or we have to go through a deprecation cycle.

> Also should we preserve the ability to specify sub-microsecond max_gap's in interpolate_na

From a user perspective, it seems like we should allow all units that may be used for the time coordinate. I admit I don't understand the subtleties of precision though.

spencerkclark: Thanks @dcherian -- I agree on all points.   

So @huard, I think it makes sense to do the following (feel free to merge your PRs or not as you see fit):

1. Maintain the ability to output floats with nanosecond units (or finer) from `datetime_to_numeric` by converting timedelta-like objects to floats before doing the unit conversion.
2. Use a similar unit conversion approach when converting `max_gap` to a float, so that we again get the best of both worlds.  I.e. continue to support nanosecond-resolution `max_gap` timedeltas, but at the same time allow for > 292-year `max_gap` timedeltas (specified, e.g., using `datetime.timedelta` objects or `np.timedelta64` objects of microsecond or coarser resolution).

(2) might require some careful logic, but in principle I think it would be doable.  Does that make sense?  Thanks already for your help in identifying/sorting all these issues out.  Sorry for taking a while to think all this through.

huard: I've merged my two PRs into #3631 and implemented the `py_timedelta_to_float` function. 
It seems to work except for earlier numpy versions. I've special-cased earlier numpy versions since I don't think I can fix that.

spencerkclark: Thanks a lot @huard -- I'll have a closer look over the coming days.

huard:
Merged with #3631.

PR #3646, opened 2019-12-19
GroupBy plotting

dcherian:
- [ ] Tests added

This adds plotting methods to GroupBy objects so that it's easy to plot each group as a facet. I'm finding this super helpful in my current research project. 

It's pretty self-contained, mostly just adding `map_groupby*` methods to `FacetGrid`. But that's because I make `GroupBy` mimic the underlying `DataArray` by adding `coords`, `attrs` and `__getitem__`. This still needs more tests but I was wondering what people thought of the feature and the implementation.

## Example

``` python
import numpy as np
import xarray as xr

time = np.arange(80)
da = xr.DataArray(5 * np.sin(2*np.pi*time/10), coords={"time": time}, dims="time")
da["period"] = da.time.where((time % 10) == 0).ffill("time")/10
da.plot()
```
image

``` python
da.groupby("period").plot(col="period", col_wrap=4)
```
image

``` python
da = da.expand_dims(y=10)
da.groupby("period").plot(col="period", col_wrap=4, sharey=True, robust=True)
```
image

### DIFF
--- a/xarray/plot/plot.py
+++ b/xarray/plot/plot.py
@@ -163,19 +163,26 @@ def plot(
         Additional keyword arguments to matplotlib
 
     """
-    darray = darray.squeeze().compute()
+    from ..core.groupby import GroupBy
+
+    if isinstance(darray, GroupBy):
+        if row is None and col is None:
+            raise ValueError(
+                f"Both 'row' and 'col' are None. Expected one of 'row' or 'col' to be '{darray._unique_coord.name}'"
+            )
 
     plot_dims = set(darray.dims)
     plot_dims.discard(row)
     plot_dims.discard(col)
     plot_dims.discard(hue)
 
-    ndims = len(plot_dims)
+    # if we are grouping over a 1D uniquely valued coordinate, then plotting should work.
+    if isinstance(darray, GroupBy) and darray._unique_coord.equals(
+        darray._obj.coords[darray._group_dim]
+    ):
+        plot_dims.discard(darray._group_dim)
 
-    error_msg = (
-        "Only 1d and 2d plots are supported for facets in xarray. "
-        "See the package `Seaborn` for more options."
-    )
+    ndims = len(plot_dims)
 
     if ndims in [1, 2]:
         if row or col:
@@ -194,12 +201,26 @@ def plot(
                 plotfunc = pcolormesh
     else:
         if row or col or hue:
-            raise ValueError(error_msg)
+            raise ValueError(
+                f"Only 1d and 2d plots are supported for facets in xarray. "
+                f"Provided DataArray has {ndims} dimension(s). "
+                f"See the package `Seaborn` for more options."
+            )
         plotfunc = hist
 
     kwargs["ax"] = ax
 
     return plotfunc(darray, **kwargs)
+
+
+def _sanity_check_groupby_row_col(grouped, row, col):
+    grouped_name = grouped._unique_coord.name
+    row_or_col = row if row is not None else col
+    if row_or_col != grouped_name:
+        raise ValueError(
+            "Expected grouped variable %r for 'row' or 'col', received %r instead."
+            % (grouped_name, row_or_col)
+        )
 
 
 # This function signature should not change so that it can use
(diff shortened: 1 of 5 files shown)


Illviljan:
This is looking great! Being able to easily plot your groupings like this is so useful. Especially when grouping results isn't trivial and requires a bit of tuning.

When I've been doing these kinds of plots I've mostly just changed the colors for each group:
image
Would it be easy to do this with this implementation?

dcherian:
It should be possible but is not implemented yet.

dcherian:
That is possible but hasn't been implemented yet.

dcherian:
@Illviljan would you mind testing this out and possibly reviewing it?

xarray/plot/plot.py
```diff
@@ -404,12 +428,27 @@ def line(
     *args, **kwargs : optional
         Additional arguments to :py:func:`matplotlib:matplotlib.pyplot.plot`.
     """
+
+    from ..core.groupby import GroupBy
+
+    if isinstance(darray, GroupBy):
+        if row is None and col is None:
+            raise ValueError(
+                f"Both 'row' and 'col' are None. Expected one of 'row' or 'col' to be '{darray._unique_coord.name}'"
```
Illviljan, line 437: Not a fan of this error. I was expecting `ds.variable.groupby(ds.id).plot.line()` to work.

If no facetgrid options have been defined it should just loop through the plots and append to the current plot with  different colors, for example:
```python
ds = xr.Dataset(
            {
                "variable": (
                    ("lat", "lon", "time"),
                    np.arange(72.0).reshape((4, 3, 6)),
                ),
                "id": (("lat", "lon"), np.arange(12.0).reshape((4, 3))),
            },
            coords={
                "lat": np.arange(4),
                "lon": np.arange(3),
                "time": pd.date_range(start="2001-01-01", freq="12H", periods=6),
            },
        )

for id_val, var in ds.variable.groupby(ds.id):
    var.plot.line(x="time")
```
image

Compared to facetgrid version:
```python
ds.variable.groupby(ds.id).plot(row="id")
```
image

Not sure how difficult this is to implement but this can also be saved for a future PR.

xarray/plot/plot.py
```diff
@@ -428,6 +467,9 @@ def line(
     ax = get_axis(figsize, size, aspect, ax)
     xplt, yplt, hueplt, hue_label = _infer_line_data(darray, x, y, hue)
 
+    if not isinstance(darray, GroupBy):
+        darray = darray.squeeze().compute()
```
Illviljan, line 471: Is this compute necessary? Because the ax.plot uses `xplt_val`, `yplt_val` as inputs and those does not appear to be directly dependent on `darray`?

xarray/plot/plot.py
```diff
@@ -404,12 +428,27 @@ def line(
...
     # Handle facetgrids first
     if row or col:
         allargs = locals().copy()
         allargs.update(allargs.pop("kwargs"))
-        allargs.pop("darray")
-        return _easy_facetgrid(darray, line, kind="line", **allargs)
+        del allargs["darray"]
+        del allargs["GroupBy"]
+
+        if not isinstance(darray, GroupBy):
```
Illviljan, line 447: Why the `not isinstance(darray, GroupBy)`? Is it faster?

Illviljan (commented):
This is exciting! I noticed the what's new isn't mentioning the plotting functionality.

Illviljan:
```python
ds = xr.tutorial.scatter_example_dataset()
ds.A.groupby(ds.w).plot.contour()
ds.A.groupby(ds.w).plot.imshow()
```
These crashes with `AttributeError: 'DataArrayGroupBy' object has no attribute 'ndim'`. 
Can `NdimSizeLenMixin` be used here?

Illviljan:
A few notes before I forget them:
* Docs crashed because UncachedAccessor wasn't used.
* _plot2d stuff didn't work because of the missing ndim.
* `ds.A.groupby(ds.w).plot.contour(col="w", row="z")` didn't work even though `ds.A.plot.contour(col="w", row="z")` did. Made that work. The facegrids aren't an exact copy though, the curves are the same but may be in other subplots, there's some sorting going on somewhere.
* I think we should avoid adding DataArrayGroupby to the mix in `plot.py,` I think it'll look cleaner if we always send in pure dataarrays there and do the filtering in DataArrayGroupby.plot instead. Might avoid circular imports then as well.

PR #3654, opened 2019-12-28
Tests for variables with units

keewis:
As promised in #3493, this adds integration tests for units. I'm doing this now rather than later since I encountered a few cases in #3643 where a increased test coverage for variables would have been helpful.

 - [x] Tests added

### DIFF
--- a/xarray/tests/test_units.py
+++ b/xarray/tests/test_units.py
@@ -1495,6 +1495,62 @@ class TestVariable(VariableSubclassobjects):
         xr.testing.assert_identical(expected, actual)
 
     @pytest.mark.parametrize(
+        "unit,error",
+        (
+            pytest.param(
+                1,
+                DimensionalityError,
+                id="no_unit",
+                marks=pytest.mark.xfail(reason="uses 0 as a replacement"),
+            ),
+            pytest.param(
+                unit_registry.dimensionless, DimensionalityError, id="dimensionless"
+            ),
+            pytest.param(unit_registry.s, DimensionalityError, id="incompatible_unit"),
+            pytest.param(
+                unit_registry.cm,
+                None,
+                id="compatible_unit",
+                marks=pytest.mark.xfail(reason="converts to fill value's unit"),
+            ),
+            pytest.param(unit_registry.m, None, id="identical_unit"),
+        ),
+    )
+    def test_missing_value_fillna(self, unit, error):
+        value = 0
+        array = (
+            np.array(
+                [
+                    [1.4, 2.3, np.nan, 7.2],
+                    [np.nan, 9.7, np.nan, np.nan],
+                    [2.1, np.nan, np.nan, 4.6],
+                    [9.9, np.nan, 7.2, 9.1],
+                ]
+            )
+            * unit_registry.m
+        )
+        variable = xr.Variable(("x", "y"), array)
+
+        fill_value = value * unit
+
+        if error is not None:
+            with pytest.raises(error):
+                print(variable.fillna(value=fill_value))
+
+            return
+
+        expected = attach_units(
+            strip_units(variable).fillna(
+                value=fill_value.to(unit_registry.m).magnitude
+            ),
+            extract_units(variable),
+        )
+        actual = variable.fillna(value=fill_value)
+
+        assert extract_units(expected) == extract_units(actual)
+        xr.testing.assert_identical(expected, actual)
+
+    @pytest.mark.parametrize(
         "unit",
         (
             pytest.param(1, id="no_unit"),
--- a/xarray/tests/test_units.py
+++ b/xarray/tests/test_units.py
@@ -1552,7 +1552,7 @@ class TestVariable(VariableSubclassobjects):
 
         if error is not None:
             with pytest.raises(error):
-                print(variable.fillna(value=fill_value))
+                variable.fillna(value=fill_value)
 
             return
 
--- a/xarray/tests/test_units.py
+++ b/xarray/tests/test_units.py
@@ -2030,7 +2030,7 @@ class TestVariable(VariableSubclassobjects):
         array = np.linspace(0, 5, 3 * 10).reshape(3, 10).astype(dtype) * unit_registry.m
         variable = xr.Variable(("x", "y"), array)
 
-        fill_value = np.array(-100) * unit
+        fill_value = -100 * unit
 
         func = method("pad_with_fill_value", x=(2, 3), y=(1, 4))
         if error is not None:
--- a/xarray/tests/test_units.py
+++ b/xarray/tests/test_units.py
@@ -1519,9 +1519,16 @@ class TestVariable(VariableSubclassobjects):
     )
     @pytest.mark.parametrize("func", (method("equals"), method("identical")), ids=repr)
     def test_comparisons(self, func, unit, convert_data, dtype):
-        def is_compatible(first, second):
+        def is_compatible(unit1, unit2):
+            return (
+                isinstance(unit1, unit_registry.Unit)
+                and isinstance(unit2, unit_registry.Unit)
+                and unit1.dimensionality == unit2.dimensionality
+            )
+
+        def compatible_mappings(first, second):
             return {
-                key: unit_registry.Quantity(0, unit1).check(unit2)
+                key: is_compatible(unit1, unit2)
                 for key, (unit1, unit2) in merge_mappings(first, second)
             }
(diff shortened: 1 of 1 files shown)


xarray/tests/test_units.py
```diff
@@ -1245,6 +1263,146 @@ def test_dot_dataarray(dtype):
...
+            setattr(cls, item, None)
+
+        return cls
+
+    return wrapper
+
+
+@delete_attrs(
+    "test_getitem_with_mask",
+    "test_getitem_with_mask_nd_indexer",
```
keewis: are there better ways to exclude inherited tests? I'd like to avoid marking them as skipped since they will probably never be passing.

keewis:
this should be ready for review and merge.

Also, I found a few issues that I'm posting here so I won't forget about them:
* documentation:
  - the examples of `Variable.rolling_window` pass a variable `x` that is not set anywhere and would change the results
  - the docstring of `set_dims` is quite confusing (to me, at least), it does not mention the `shape` parameter (which would be the same as passing a dictionary with values `dict(zip(dims, shape))` to `dims`) or that broadcasting is used to expand if necessary (so no reshaping).
* code: `pad_with_fill_value` seems to implement pad for dask arrays for all versions of dask, but at least for recent versions we should rely dask's implementation instead

I can update the documentation, but I'd leave the revisiting of the dask issues (there are a lot of them) to someone else.

keewis:
the test failures are due to the pint update which means they should not be related. I'll mark them as xfail and fix them in the appropriate PR (#3611 for the `apply_ufunc` failures and the `Dataset` PR I will open once all `Variable` and `DataArray` tests are passing for the `to_stacked_array` failures)

keewis:
this should probably be merged before any of the other related PRs

xarray/tests/test_units.py
```diff
@@ -1245,6 +1295,766 @@ def test_dot_dataarray(dtype):
...
+            )
+            * unit_registry.m
+        )
+        variable = xr.Variable(("x", "y"), array)
+
+        fill_value = value * unit
+
+        if error is not None:
+            with pytest.raises(error):
+                print(variable.fillna(value=fill_value))
```
dcherian:
```suggestion
                variable.fillna(value=fill_value)
```

xarray/tests/test_units.py
```diff
@@ -1245,6 +1295,766 @@ def test_dot_dataarray(dtype):
...
+
+            return
```
dcherian: Use `else:` instead of `return`?

xarray/tests/test_units.py
```diff
@@ -1245,6 +1295,766 @@ def test_dot_dataarray(dtype):
...
+                compatible_mappings(
+                    extract_units(variable), extract_units(other)
+                ).values()
+            )
+
+        actual = func(variable, other)
+
+        assert expected == actual
+
+    @pytest.mark.parametrize(
```
dcherian: I stopped here. Will look at the rest tomorrow.

xarray/tests/test_units.py
```diff
@@ -159,8 +182,9 @@ def strip_units(obj):
 
 
 def attach_units(obj, units):
-    if not isinstance(obj, (xr.DataArray, xr.Dataset)):
-        return array_attach_units(obj, units.get("data", 1))
+    if not isinstance(obj, (xr.DataArray, xr.Dataset, xr.Variable)):
+        units = units.get("data", None) or units.get(None, None) or 1
```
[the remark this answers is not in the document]
keewis: that's a fast path for passing non-xarray objects like raw quantities. I wanted to make sure attaching to raw `numpy.arrays` works, but didn't want to check for them below.

xarray/tests/test_units.py
keewis: I could use `else:`, but `return` saves me a level of indentation

xarray/tests/test_units.py
```diff
@@ -7,12 +7,13 @@
...
 from xarray.core import formatting
 from xarray.core.npcompat import IS_NEP18_ACTIVE
+from .test_variable import VariableSubclassobjects
 
 pint = pytest.importorskip("pint")
 DimensionalityError = pint.errors.DimensionalityError
 
 
-unit_registry = pint.UnitRegistry()
+unit_registry = pint.UnitRegistry(force_ndarray=True)
```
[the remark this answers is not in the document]
keewis: Will do. Here's a more detailed explanation:

Because `Quantity` implements `__array_function__`, `xarray` treats `pint.Quantity` instances as arrays. This is mostly correct, but `pint` also allows wrapping python scalars which don't implement `ndim`, `shape` and `dtype`. Setting `force_ndarray` means `Quantity` instances always wrap `numpy.array`s and we don't get `AttributeError`s.

See hgrecco/pint#950.

xarray/tests/test_units.py
```diff
@@ -28,6 +29,21 @@
 ]
 
 
+def is_compatible(unit1, unit2):
```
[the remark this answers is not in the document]
keewis: :+1:, I'll ask

xarray/tests/test_units.py
```diff
@@ -318,12 +362,16 @@ def __repr__(self):
 
 
 class function:
```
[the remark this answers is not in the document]
keewis: They have the same purpose as the `do` class from `test_sparse.py`: pass methods and functions via `parametrize` to the tests while also working a bit like `functools.partial`.

xarray/tests/test_units.py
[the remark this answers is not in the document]
dcherian: :+1: we should stick this in the docs somewhere.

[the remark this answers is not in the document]
keewis: that's true. I will add that to the documentation todo item in #3594

xarray/tests/test_units.py
```diff
@@ -1245,6 +1311,766 @@ def test_dot_dataarray(dtype):
...
+        ) & is_compatible(unit, base_unit)
+        actual = left.broadcast_equals(right)
+
+        assert expected == actual
+
+    @pytest.mark.parametrize(
+        "indices",
+        (
+            pytest.param(4, id="single index"),
+            pytest.param([5, 2, 9, 1], id="multiple indices"),
```
dcherian: I think `Variable` supports more complicated indexing. This could be added later

xarray/tests/test_units.py
```diff
@@ -1245,6 +1311,766 @@ def test_dot_dataarray(dtype):
...
+            converted_variable = convert_units(variable, units)
+        else:
+            converted_variable = variable
+
+        expected = attach_units(
+            func(strip_units(converted_variable), strip_units(converted_y)), units
+        )
+        actual = func(variable, y)
+
+        assert extract_units(expected) == extract_units(actual)
```
dcherian: we have `assert_chunks_equal` so it makes sense (to me) that we add `assert_units_compatible`. Thoughts?

xarray/tests/test_units.py
```diff
@@ -1245,6 +1311,766 @@ def test_dot_dataarray(dtype):
...
+
+        expected = attach_units(
+            strip_units(variable).squeeze(), extract_units(variable)
+        )
+        actual = variable.squeeze()
+
+        assert extract_units(expected) == extract_units(actual)
+        xr.testing.assert_identical(expected, actual)
+
+        for dim in names:
```
dcherian: this loop doesn't seem to be doing anything.

keewis: that's true, I wanted to pass `dim` to `squeeze`

xarray/tests/test_units.py
keewis: probably, though we just want to make sure the units survive through the indexing.

xarray/tests/test_units.py
keewis: this is more of a band-aid until `assert_allclose` and `assert_identical` check unit equality (compatibility is not enough for this). It is true, though, that it is used often enough to put it into a function.

dcherian: ah this looks fine

keewis: I changed it since for new tests (or when I switch the `Dataset` tests to `assert_allclose` / `assert_identical`) I don't have to type as much

keewis:
I'll merge once the tests pass.

keewis:
thanks for the review, @dcherian

PR #3660, opened 2020-01-02
apply_ufunc: Add meta kwarg + bump dask to 2.2

### LINKED ISSUE #3574: apply_ufunc with dask='parallelized' and vectorize=True fails on compute_meta
#### MCVE Code Sample

```python
import numpy as np
import xarray as xr

ds = xr.Dataset({
    'signal': (['das_time', 'das', 'record'], np.empty((1000, 120, 45))),
    'min_height': (['das'], np.empty((120,)))   # each DAS has a different resolution
})

def some_peak_finding_func(data1d, min_height):
    """process data1d with contraints by min_height"""
    result = np.zeros((4,2))  # summary matrix with 2 peak characteristics
    return result

ds_dask = ds.chunk({'record':3})

xr.apply_ufunc(some_peak_finding_func, ds_dask['signal'], ds_dask['min_height'],
               input_core_dims=[['das_time'], []],  # apply peak finding along trace
               output_core_dims=[['peak_pos', 'pulse']],
               vectorize=True, # up to here works without dask!
               dask='parallelized',
               output_sizes={'peak_pos': 4, 'pulse':2},
               output_dtypes=[np.float],
              )
```

fails with `ValueError: cannot call `vectorize` with a signature including new output dimensions on size 0 inputs` because `dask.array.utils.compute_meta()` passes it 0-sized arrays.

#### Expected Output

This should work and works well on the non-chunked ds, without `dask='parallelized'` and the associated `output*` parameters.

#### Problem Description

I'm trying to parallelize a peak finding routine with dask (works well without it) and I hoped that `dask='parallelized` would make that simple. However, the peak finding needs to be vectorized  and it works well with vectorize=True`, but `np.vectorize` appears to have issues in `compute_meta` which is internally issued by dask in blockwise application as indicated in the source code: 

https://github.com/dask/dask/blob/e6ba8f5de1c56afeaed05c39c2384cd473d7c893/dask/array/utils.py#L118

A possible solution might be for `apply_ufunc` to pass `meta` directly to dask if it would be possible to foresee what `meta` should be. I suppose we are aiming for `np.nadarray` most of the time, though `sparse` might change that in the future.

I know I could use groupby-apply as an alternative, but there are several issues that made us use `apply_ufunc` instead:
- groupby-apply seems to have much larger overhead
- the non-core dimensions would have to be stacked into a new dimension over which to groupby, but some of the dimensions to be stacked are already a MutliIndex and cannot be easily stacked.
    - we could unstack the MultiIndex dimensions first at the risk of introducing quite a number of NaNs
    - extra coords might lose dimension infromation (will depend on all) after unstacking after application

<details>

commit: None
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.9.0-11-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.1

xarray: 0.14.0
pandas: 0.25.1
numpy: 1.17.2
scipy: 1.3.1
netCDF4: 1.4.2
pydap: None
h5netcdf: 0.7.4
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.4.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 2.5.2
distributed: 2.5.2
matplotlib: 3.1.1
cartopy: None
seaborn: 0.9.0
numbagg: None
setuptools: 41.4.0
pip: 19.2.3
conda: 4.7.12
pytest: 5.2.1
IPython: 7.8.0
sphinx: 2.2.0

</details>

dcherian:
- [x] Closes #3574
 - [x] Tests added

This makes `vectorize=True` works with functions that add new dimensions.

cc @smartass101 @jbusecke

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -1723,6 +1723,7 @@ class ZarrBase(CFEncodedBase):
                 with xr.decode_cf(store):
                     pass
 
+    @pytest.mark.xfail(reason="fails for old dask versions.")
     def test_write_persistence_modes(self):
         original = create_test_data()
 
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -1723,7 +1723,7 @@ class ZarrBase(CFEncodedBase):
                 with xr.decode_cf(store):
                     pass
 
-    @pytest.mark.xfail(reason="fails for old dask versions.")
+    @pytest.mark.xfail(reason="fails for dask <2.4.0. dask GH5334")
     def test_write_persistence_modes(self):
         original = create_test_data()
 
@@ -1788,7 +1788,7 @@ class ZarrBase(CFEncodedBase):
     def test_dataset_caching(self):
         super().test_dataset_caching()
 
-    @pytest.mark.xfail(reason="fails for old dask versions.")
+    @pytest.mark.xfail(reason="fails for dask <2.4.0. dask GH5334")
     def test_append_write(self):
         ds, ds_to_append, _ = create_append_test_data()
         with self.create_zarr_target() as store_target:
@@ -1865,7 +1865,7 @@ class ZarrBase(CFEncodedBase):
                 xr.concat([ds, ds_to_append], dim="time"),
             )
 
-    @pytest.mark.xfail(reason="fails for old dask versions.")
+    @pytest.mark.xfail(reason="fails for dask <2.4.0. dask GH5334")
     def test_append_with_new_variable(self):
 
         ds, ds_to_append, ds_with_new_var = create_append_test_data()
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -1788,6 +1788,7 @@ class ZarrBase(CFEncodedBase):
     def test_dataset_caching(self):
         super().test_dataset_caching()
 
+    @pytest.mark.xfail(reason="fails for old dask versions.")
     def test_append_write(self):
         ds, ds_to_append, _ = create_append_test_data()
         with self.create_zarr_target() as store_target:
@@ -1864,6 +1865,7 @@ class ZarrBase(CFEncodedBase):
                 xr.concat([ds, ds_to_append], dim="time"),
             )
 
+    @pytest.mark.xfail(reason="fails for old dask versions.")
     def test_append_with_new_variable(self):
 
         ds, ds_to_append, ds_with_new_var = create_append_test_data()
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -37,7 +37,7 @@ from xarray.conventions import encode_dataset_coordinates
 from xarray.core import indexing
 from xarray.core.options import set_options
 from xarray.core.pycompat import dask_array_type
-from xarray.tests import mock
+from xarray.tests import LooseVersion, mock
 
 from . import (
     arm_xfail,
@@ -76,9 +76,14 @@ except ImportError:
     pass
 
 try:
+    import dask
     import dask.array as da
+
+    dask_version = dask.__version__
 except ImportError:
-    pass
+    # needed for xfailed tests when dask < 2.4.0
+    # remove when min dask > 2.4.0
+    dask_version = "10.0"
 
 ON_WINDOWS = sys.platform == "win32"
 
@@ -1723,7 +1728,7 @@ class ZarrBase(CFEncodedBase):
                 with xr.decode_cf(store):
                     pass
 
-    @pytest.mark.xfail(reason="fails for dask <2.4.0. dask GH5334")
+    @pytest.mark.skipif(LooseVersion(dask_version) < "2.4", reason="dask GH5334")
     def test_write_persistence_modes(self):
         original = create_test_data()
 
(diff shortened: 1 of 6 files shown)


dcherian:
The test failed because `meta` was not valid for `dask==1.2`

Our minimum versions policy says we can bump dask to 2.1.0. 
`dask          1.2     (2019-04-13) 2.1     (2019-07-08) <`

So I've bumped the min dask version and added `meta` as a kwarg to `apply_ufunc`.

dcherian:
`min-all-deps` fails with this weird error:

```

    def test_append_with_new_variable(self):
    
        ds, ds_to_append, ds_with_new_var = create_append_test_data()
    
        # check append mode for new variable
        with self.create_zarr_target() as store_target:
            xr.concat([ds, ds_to_append], dim="time").to_zarr(store_target, mode="w")
            ds_with_new_var.to_zarr(store_target, mode="a")
            combined = xr.concat([ds, ds_to_append], dim="time")
            combined["new_var"] = ds_with_new_var["new_var"]
>           assert_identical(combined, xr.open_zarr(store_target))

xarray/tests/test_backends.py:1876: 
xarray/core/dataset.py:1354: in identical
    other, "identical"
xarray/core/dataset.py:1302: in _all_compat
    self._variables, other._variables, compat=compat
xarray/core/utils.py:338: in dict_equiv
    if k not in second or not compat(first[k], second[k]):
xarray/core/dataset.py:1299: in compat
    return getattr(x, compat_str)(y)
xarray/core/variable.py:1670: in identical
    other, equiv=equiv
xarray/core/variable.py:1647: in equals
    self._data is other._data or equiv(self.data, other.data)
xarray/core/duck_array_ops.py:224: in array_equiv
    flag_array = (arr1 == arr2) | (isnull(arr1) & isnull(arr2))
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:1739: in __eq__
    return elemwise(operator.eq, self, other)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:3749: in elemwise
    **blockwise_kwargs
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/blockwise.py:145: in blockwise
    chunkss, arrays = unify_chunks(*args)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:3020: in unify_chunks
    (asanyarray(a) if ind is not None else a, ind) for a, ind in partition(2, args)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:3020: in <listcomp>
    (asanyarray(a) if ind is not None else a, ind) for a, ind in partition(2, args)

a = array(['2019-01-01T00:00:00.000000000', '2019-01-02T00:00:00.000000000',
       '2019-01-03T00:00:00.000000000', '2019-01-04T00:00:00.000000000',
       '2019-01-05T00:00:00.000000000'], dtype='datetime64[ns]')

    def asanyarray(a):
        """Convert the input to a dask array.
    
        Subclasses of ``np.ndarray`` will be passed through as chunks unchanged.
    
        Parameters
        ----------
        a : array-like
            Input data, in any form that can be converted to a dask array.
    
        Returns
        -------
        out : dask array
            Dask array interpretation of a.
    
        Examples
        --------
        >>> import dask.array as da
        >>> import numpy as np
        >>> x = np.arange(3)
        >>> da.asanyarray(x)
        dask.array<array, shape=(3,), dtype=int64, chunksize=(3,)>
    
        >>> y = [[1, 2, 3], [4, 5, 6]]
        >>> da.asanyarray(y)
        dask.array<array, shape=(2, 3), dtype=int64, chunksize=(2, 3)>
        """
        if isinstance(a, Array):
            return a
        elif hasattr(a, "to_dask_array"):
            return a.to_dask_array()
>       elif hasattr(a, "data") and type(a).__module__.startswith("xarray."):
E       ValueError: cannot include dtype 'M' in a buffer

```

dcherian:
Does anyone have thoughts on what might be happening here?

shoyer:
This looks like an dask bug that we fixed a somewhat recently (6 months ago?).

I wouldn’t worry about it too much about it. Supporting old versions of dependencies is a best effort, not a mandate. I would just skip it on old versions of dask and move on.

dcherian:
Ah found it: https://github.com/dask/dask/pull/5334. Missed it when I looked through the changelog earlier.

dcherian:
ready for review and merge.

xarray/tests/test_sparse.py
```diff
@@ -873,3 +873,16 @@ def test_dask_token():
     t5 = dask.base.tokenize(ac + 1)
     assert t4 != t5
     assert isinstance(ac.data._meta, sparse.COO)
+
+
+@requires_dask
```
shoyer: Should this also have `@requires_sparse`?

xarray/tests/test_backends.py
```diff
@@ -1723,6 +1723,7 @@ def test_hidden_zarr_keys(self):
                 with xr.decode_cf(store):
                     pass
 
+    @pytest.mark.xfail(reason="fails for dask <2.4.0. dask GH5334")
```
shoyer: can we make this xfail conditional on the old dask version?

xarray/tests/test_sparse.py
dcherian: There's a `importorskip("sparse")` on top so I think this is fine

shoyer:
OK great! Looks good to me!

dcherian:
Thanks for the review. I'll merge when tests pass.

PR #3667, opened 2020-01-07
Add map_blocks example to docs

bradyrx:
This PR adds an example to the `xr.map_blocks()` docs. 

It is modeled off of @rabernat's example here: https://nbviewer.jupyter.org/gist/rabernat/30e7b747f0e3583b5b776e4093266114 and was encouraged to be opened via the gitter.

It might be slightly large for an example, but I think it shows an applied case to use this in and clears up some common pitfalls (e.g. how to pass args/kwargs).

### DIFF
--- a/xarray/core/parallel.py
+++ b/xarray/core/parallel.py
@@ -154,6 +154,48 @@ def map_blocks(
     --------
     dask.array.map_blocks, xarray.apply_ufunc, xarray.Dataset.map_blocks,
     xarray.DataArray.map_blocks
+
+    Examples
+    --------
+
+    Calculate an anomaly from climatology using ``.groupby()``. Using
+    ``xr.map_blocks()`` allows for parallel operations with knowledge of ``xarray``,
+    its indices, and its methods like ``.groupby()``.
+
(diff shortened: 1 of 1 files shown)


bradyrx:
What's going on here? I use travis on my repos so I'm not familiar with the Azure setup. I only modified a docstring so I'm not sure why it would break the testing suite? Unless it's testing my code snippet in the docs?

keewis:
We use the upstream-dev check as a early warning system for changes in the unreleased versions of our dependencies. It being the only failure usually means upstream (I guess pandas in this case) changed something, breaking some of our tests.

Since this also happens in other PRs and is completely unrelated (I think we don't run doctest in CI), I'd say you can safely ignore the failure.

dcherian:
Thanks @bradyrx

bradyrx:
Oops, forgot to add to `whats-new`, but this is a pretty minor addition.

dcherian:
Send in a new PR with an entry! We have a documentation section...

bradyrx:
Thanks @dcherian -- done in https://github.com/pydata/xarray/pull/3682.

PR #3679, opened 2020-01-09
How do I add a new variable to dataset.

### LINKED ISSUE #3678: Feature request: How Do I add a new variable to a data set?
A common thing to want to do is to create a new `DataArray` and add it to an existing `Dataset`, but the answer does not seem to appear in the "How Do I?" list.

Thanks!

rpgoldman:
Closes #3678 

I don't have the full build environment: someone who does will have to make sure that I got the sphinx formatting right.

### DIFF
--- a/doc/howdoi.rst
+++ b/doc/howdoi.rst
@@ -12,7 +12,7 @@ How do I ...
    * - How do I...
      - Solution
    * - add a DataArray to my dataset as a new variable
-     - ``my_dataset[varname] = my_dataArray``
+     - ``my_dataset[varname] = my_dataArray`` or :py:meth:`Dataset.assign`
    * - add variables from other datasets to my dataset
      - :py:meth:`Dataset.merge`
    * - add a new dimension and/or coordinate
--- a/doc/howdoi.rst
+++ b/doc/howdoi.rst
@@ -62,4 +62,4 @@ How do I ...
 
 .. rubric:: Footnotes
 
-.. [#assign_dict] See also :ref:`_dictionary_like_methods`:.
+.. [#assign_dict] See also :ref:`dictionary_like_methods`:.
--- a/doc/howdoi.rst
+++ b/doc/howdoi.rst
@@ -12,7 +12,7 @@ How do I ...
    * - How do I...
      - Solution
    * - add a DataArray to my dataset as a new variable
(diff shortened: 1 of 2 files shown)


doc/howdoi.rst
```diff
@@ -11,6 +11,8 @@ How do I ...
 
    * - How do I...
      - Solution
+   * - add a DataArray to my dataset as a new variable
+     - ``my_dataset[varname] = my_dataArray``
```
dcherian:
```suggestion
     - ``my_dataset[varname] = my_dataArray`` or :py:meth:`Dataset.assign`
```

We should also link to https://xarray.pydata.org/en/stable/data-structures.html#dictionary-like-methods. We have docs on building the documentation here: https://xarray.pydata.org/en/stable/contributing.html#contributing-to-the-documentation

rpgoldman:
OK, I have tried to do as you suggested.

I looked at the directions for building the docs as you suggested, and I'm afraid it's too big an ask for me.  I don't use conda, and I can't spare the time to learn it.  Similarly, I know enough RST to write numpy-style docstrings, but I don't have the time to learn Sphinx.  With all due respect, if xarray wants contributions from people like me, something has to be done to make this easier.

I'm just doing this because I'm contributing to Arviz, which depends on xarray.  I can try to fix things if I see something wrong in the "Docs" check, but that's all I can do, I'm afraid.

rpgoldman:
I didn't get the cross-reference syntax right. I've tried to fix it, and am waiting for a new check.

rpgoldman:
OK, I'm not trying to be difficult, but I think someone else who knows how RST works could fix this in 2.4 seconds, or I could bang my head on it for hours.

Would someone please just fix my syntax error and merge this?

doc/howdoi.rst
```diff
@@ -11,6 +11,8 @@ How do I ...
 
    * - How do I...
      - Solution
+   * - add a DataArray to my dataset as a new variable
+     - ``my_dataset[varname] = my_dataArray`` or :py:meth:`Dataset.assign`[#assign_dict]_
```
keewis: I think you could use
```suggestion
     - ``my_dataset[varname] = my_dataArray`` or :py:meth:`Dataset.assign` (see also :ref:`dictionary_like_methods`)
```
then you wouldn't need the footnote.

dcherian:
Great. Thanks @rpgoldman Welcome to xarray!

PR #3683, opened 2020-01-10
Fix docstring for combine_first: returns a Dataset

davidcaron:
The return type in the docstring should be a Dataset, not a DataArray.


dcherian:
Thanks @davidcaron. I see this is your first PR here. Welcome to xarray!

Issue #3673, opened 2020-01-08
Test failures with pandas master


dcherian:
https://dev.azure.com/xarray/xarray/_build/results?buildId=1859&view=logs&jobId=41d90575-019f-5cfd-d78e-c2adebf9a30b&j=41d90575-019f-5cfd-d78e-c2adebf9a30b

```
2020-01-07T19:19:39.0297443Z =================================== FAILURES ===================================
2020-01-07T19:19:39.0495451Z _________________ test_timedeltaindex_add_cftimeindex[365_day] _________________
2020-01-07T19:19:39.0496702Z 
2020-01-07T19:19:39.0498391Z calendar = '365_day'
2020-01-07T19:19:39.0499052Z 
2020-01-07T19:19:39.0500036Z     @requires_cftime
2020-01-07T19:19:39.0500669Z     @pytest.mark.parametrize("calendar", _CFTIME_CALENDARS)
2020-01-07T19:19:39.0501986Z     def test_timedeltaindex_add_cftimeindex(calendar):
2020-01-07T19:19:39.0502556Z         a = xr.cftime_range("2000", periods=5, calendar=calendar)
2020-01-07T19:19:39.0503384Z         deltas = pd.TimedeltaIndex([timedelta(days=2) for _ in range(5)])
2020-01-07T19:19:39.0503635Z         result = deltas + a
2020-01-07T19:19:39.0504330Z         expected = a.shift(2, "D")
2020-01-07T19:19:39.0505283Z         assert result.equals(expected)
2020-01-07T19:19:39.0506265Z >       assert isinstance(result, CFTimeIndex)
2020-01-07T19:19:39.0507151Z E       AssertionError: assert False
2020-01-07T19:19:39.0508329Z E        +  where False = isinstance(Index([2000-01-03 00:00:00, 2000-01-04 00:00:00, 2000-01-05 00:00:00,\n       2000-01-06 00:00:00, 2000-01-07 00:00:00],\n      dtype='object'), CFTimeIndex)
2020-01-07T19:19:39.0509344Z 
2020-01-07T19:19:39.0510351Z xarray/tests/test_cftimeindex.py:724: AssertionError
2020-01-07T19:19:39.0510935Z _________________ test_timedeltaindex_add_cftimeindex[360_day] _________________
2020-01-07T19:19:39.0511748Z 
2020-01-07T19:19:39.0513201Z calendar = '360_day'
2020-01-07T19:19:39.0513958Z 
[...]

PR #3688, opened 2020-01-12
Fix test_cf_datetime_nan under pandas master

spencerkclark:
This fixes `test_cf_datetime_nan` for upcoming releases of pandas.  See failure class (2) reported in #3673.

 - [x] Tests added


dcherian:
Thanks @spencerkclark

PR #3703, opened 2020-01-17
hardcoded xarray.__all__

### LINKED ISSUE #3695: mypy --strict fails on scripts/packages depending on xarray; __all__ required
Checked this with both 0.14.1 and master branch.

Create `foo.py`:

```python
from xarray import DataArray
```

and run:

```sh
$ mypy --strict foo.py
```

which results in

```
foo.py:1: error: Module 'xarray' has no attribute 'DataArray'
Found 1 error in 1 file (checked 1 source file)
```

I did a bit of digging trying to make it work, it looks like what makes the above script work with mypy is adding

```python
__all__ = ('DataArray',)
```

to `xarray/__init__.py`, otherwise mypy treats those imports as "private" (and is correct in doing so).

Should `__all__` be added to the root `__init__.py`? To any `__init__.py` in subpackages as well?

crusaderky:
Closes #3695

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.730  # Must match ci/requirements/*.yml
+    rev: v0.761  # Must match ci/requirements/*.yml
     hooks:
       - id: mypy
   # run these occasionally, ref discussion https://github.com/pydata/xarray/pull/3194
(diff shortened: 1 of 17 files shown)


max-sixty:
Great! Thanks @crusaderky

aldanor:
Thanks a million!

aldanor:
Wondering, would it be possible to release a minor version with this stuff anytime soon, or is the plan to wait for the next big 0.15?

crusaderky:
I think we're very close to 0.15...

PR #3706, opened 2020-01-17
Pint support for variables

### LINKED ISSUE #3783: New TestVariable.test_pad failure with pint 0.11
Since y-day, `pint` 0.11 was released and it results in a new test failure (in addition to the ones in #3778 and #3779):
```
____________________________ TestVariable.test_pad _____________________________

self = <xarray.tests.test_units.TestVariable object at 0x7f6bfd8c1b20>

    def test_pad(self):
        data = np.arange(4 * 3 * 2).reshape(4, 3, 2)
        v = self.cls(["x", "y", "z"], data)
    
        xr_args = [{"x": (2, 1)}, {"y": (0, 3)}, {"x": (3, 1), "z": (2, 0)}]
        np_args = [
            ((2, 1), (0, 0), (0, 0)),
            ((0, 0), (0, 3), (0, 0)),
            ((3, 1), (0, 0), (2, 0)),
        ]
        for xr_arg, np_arg in zip(xr_args, np_args):
            actual = v.pad_with_fill_value(**xr_arg)
            expected = np.pad(
                np.array(v.data.astype(float)),
                np_arg,
                mode="constant",
                constant_values=np.nan,
            )
            assert_array_equal(actual, expected)
            assert isinstance(actual._data, type(v._data))
    
        # for the boolean array, we pad False
        data = np.full_like(data, False, dtype=bool).reshape(4, 3, 2)
        v = self.cls(["x", "y", "z"], data)
        for xr_arg, np_arg in zip(xr_args, np_args):
>           actual = v.pad_with_fill_value(fill_value=False, **xr_arg)

xarray/tests/test_variable.py:813: 
xarray/core/variable.py:1210: in pad_with_fill_value
    array = np.pad(
<__array_function__ internals>:5: in pad
    ???
/usr/lib/python3.8/site-packages/pint/quantity.py:1543: in __array_function__
    return numpy_wrap("function", func, args, kwargs, types)
/usr/lib/python3.8/site-packages/pint/numpy_func.py:894: in numpy_wrap
    return handled[name](*args, **kwargs)
/usr/lib/python3.8/site-packages/pint/numpy_func.py:671: in _pad
    kwargs["constant_values"] = _recursive_convert(kwargs["constant_values"], units)
/usr/lib/python3.8/site-packages/pint/numpy_func.py:659: in _recursive_convert
    arg = unit._REGISTRY.Quantity(arg, unit)
/usr/lib/python3.8/site-packages/pint/quantity.py:200: in __new__
    inst._magnitude = _to_magnitude(

value = False, force_ndarray = True, force_ndarray_like = False

    def _to_magnitude(value, force_ndarray=False, force_ndarray_like=False):
        if isinstance(value, (dict, bool)) or value is None:
>           raise TypeError("Invalid magnitude for Quantity: {0!r}".format(value))
E           TypeError: Invalid magnitude for Quantity: False

/usr/lib/python3.8/site-packages/pint/compat.py:49: TypeError
```

keewis:
I realized that most of the operations depend on `Variable`, which means that this should be the first step to making the other tests pass. This PR copies from #3611, making that PR a work-in-progress again until I remove those parts.

 - [x] Closes #3783

These are the current failures:
* `np.prod`: not implemented by `pint` yet, but should work once it is
* comparisons: `identical` fails to detect same value, but in different units (like `1 m` and `100 cm`) as different. This is hard to implement in a way that does not include `isinstance` or `hasattr` checks.
* `rank`: not implemented for non-ndarrays, so maybe we should mark this as `skip`?
* `rolling_window`: `nputils._rolling_window` uses `np.lib.stride_tricks.as_strided`, which cannot be overridden by `pint`. We probably have to use something different?
* `shift`: ~this tries to trim, then concatenate a filled array, but I think this should use `np.pad` after trimming? Are there any disadvantages to that?~ it uses `numpy.pad` now which is supported since `dask==0.18` or `dask==0.19`
* `concat`: ~this was a misconception on my part, I didn't realize this was a classmethod. After fixing that this still fails because I assumed `Variable.concat` used the dimension names to reshape the arrays. It does not, so this fails:~ I rewrote the test to pass arrays that don't cause the failure
```python
xr.Variable.concat([xr.Variable(("x", "y"), ...), xr.Variable(("y", "z"), ...)], dim="y")
```
* `pad_with_fill_value`: ~I think this is a bug in `pint` (see hgrecco/pint#992)~ fixed on pint master

Does anyone have any comments on these before I start fixing them? @dcherian?


keewis:
I fixed the tests that were easier, but I have no idea about `rolling_window` / `as_strided` because I don't understand how it works, and `identical` would be easy to fix if we special-cased `pint`, but I'd like to avoid that

keewis:
I'm not sure why, but `dask==1.2` fails where later versions work:
```pytb
xarray/tests/test_dask.py:95: in assertLazyAndIdentical
    self.assertLazyAnd(expected, actual, assert_identical)
xarray/tests/test_dask.py:72: in assertLazyAnd
    test(actual, expected)
xarray/core/variable.py:1671: in identical
    other, equiv=equiv
xarray/core/variable.py:1648: in equals
    self._data is other._data or equiv(self.data, other.data)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:1080: in __bool__
    return bool(self.compute())
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/base.py:156: in compute
    (result,) = compute(self, traverse=False, **kwargs)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/base.py:398: in compute
    results = schedule(dsk, keys, **kwargs)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:503: in get_sync
    return get_async(apply_sync, 1, dsk, keys, **kwargs)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:470: in get_async
    fire_task()
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:445: in fire_task
    callback=queue.put)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:492: in apply_sync
    res = func(*args, **kwds)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:235: in execute_task
    result = pack_exception(e, dumps)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:230: in execute_task
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/core.py:119: in _execute_task
    return func(*args2)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/optimization.py:942: in __call__
    dict(zip(self.inkeys, args)))
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/core.py:149: in get
    result = _execute_task(task, cache)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:470: in get_async
    fire_task()
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:445: in fire_task
    callback=queue.put)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:492: in apply_sync
    res = func(*args, **kwds)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:235: in execute_task
    result = pack_exception(e, dumps)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/local.py:230: in execute_task
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/core.py:119: in _execute_task
    return func(*args2)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/optimization.py:942: in __call__
    dict(zip(self.inkeys, args)))
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/core.py:149: in get
    result = _execute_task(task, cache)
 
arg = (<built-in function and_>, 'isnan-453fb8344a03bec317cacd3c62a597be', '_0')
cache = {'_0': array([[ True,  True],
       [ True,  True]]), '_1': array([], shape=(2, 0), dtype=float64), '_2': array([[nan, nan],
       [nan, nan]]), 'eq-09349d4a881a8a69cae20e46abf35b77': False, ...}
        2
        >>> _execute_task((add, (inc, 'x'), 1), cache)  # Support nested computation
        3
    
        Also grab data from cache
        >>> _execute_task('x', cache)
        1
    
        Support nested lists
        >>> list(_execute_task(['x', 'y'], cache))
        [1, 2]
    
        >>> list(map(list, _execute_task([['x', 'y'], ['y', 'x']], cache)))
        [[1, 2], [2, 1]]
    
        >>> _execute_task('foo', cache)  # Passes through on non-keys
        'foo'
        """
        if isinstance(arg, list):
        elif istask(arg):
            func, args = arg[0], arg[1:]
>           return func(*args2)
E           ValueError: operands could not be broadcast together with shapes (2,0) (2,2)
```

keewis:
the easiest way to fix `identical` would be to pass a `equiv` function that forwards to `duck_array_ops.array_equiv` but also checks the units:
```python
def units_equiv(a, b):
    if hasattr(a, "units") or hasattr(b, "units"):
        units_a = getattr(a, "units", None)
        units_b = getattr(b, "units", None)

	registry = getattr(units_a, "_REGISTRY", None) or getattr(b, "_REGISTRY", None)

	units_a = units_a or registry.dimensionless
	units_b = units_b or registry.dimensionless

	return units_a == units_b
    else:
	# no units at all
	return True

def equiv_with_units(a, b):
    return duck_array_ops.array_equiv(a, b) and units_equiv(a, b)
```
However, I'm out of ideas on how to automatically use that (I'd like to avoid having to call `a.identical(b, equiv=equiv_with_units)`).

---

There is also `rolling_window` that strips the units due to calling `numpy.lib.stride_tricks.as_strided`, but I don't know how to fix that unless we rewrite `_rolling_window` or dispatch differently in `rolling_window`.

There seems to have been some work to add a function named `numpy.rolling_window` or `numpy.sliding_window` that could then be overridden by `pint`, but I think that effort has stalled?

dcherian:
I think your `units_equiv` thing could go in `duck_array_ops.lazy_array_equiv` which tries to check everything but the actual values themselves.

>  but I don't know how to fix that unless we rewrite _rolling_window or dispatch differently in rolling_window.

This may be necessary but perhaps not the most important thing right now?

keewis:
> I think your `units_equiv` thing could go in `duck_array_ops.lazy_array_equiv`

that could work for `Dataset.identical` but not `Variable.identical`, which I believe by default directly calls `array_equiv`

re `rolling_window`: should I leave it as `xfail` for now?

dcherian:
`array_equiv` calls `lazy_array_equiv` before doing much else:

https://github.com/pydata/xarray/blob/6d1434e9b9e9766d6fcafceb05e81734b29994ea/xarray/core/duck_array_ops.py#L215-L227

dcherian:
> re rolling_window: should I leave it as xfail for now?

Fine by me. :)

keewis:
hrmm... I had investigated this before and thought I remembered correctly. I can't put it in `lazy_array_equiv`, though, since `array_equiv` is also used by `equals` (which should not check the units, only the dimensionality)

dcherian:
I guess we need to either pass a kwarg regarding checking units/dimensionality down to `lazy_array_equiv` or add a `units_equiv(only_dimensionality=True/False)` check to https://github.com/pydata/xarray/blob/6d1434e9b9e9766d6fcafceb05e81734b29994ea/xarray/core/variable.py#L1645-L1648

I think a separate `units_equiv` function may be cleaner?

Note that we explicitly use `lazy_array_equiv` in `concat` so it'd be nice to have something that could be easily used there too:
https://github.com/pydata/xarray/blob/6d1434e9b9e9766d6fcafceb05e81734b29994ea/xarray/core/concat.py#L194-L208

keewis:
> I think a separate `units_equiv` function may be cleaner?

I agree. However, I'm reluctant to add that function since that would be the first `pint` dependent code we have except from the unit tests (but do tell me if that's not an issue).

I'm inclined to provide some kind of hook for wrapped libraries instead (allow them to define a method like `metadata_identical` or something that does the check for us) so the code in `identical` would become something like
```python
def metadata_identical(arr1, arr2):
    if hasattr(arr1, "metadata_identical"):
        return arr1.metadata_identical(arr2)
    elif hasattr(arr2, "metadata_identical"):
        return arr2.metadata_identical(arr1)
    else:
        return True

return (
    self.dims == other.dims
    and (self._data is other._data or equiv(self.data, other.data))
    and metadata_identical(self.data, other.data)
)
```

> Note that we explicitly use `lazy_array_equiv` in `concat` so it'd be nice to have something that could be easily used there too

I don't think that's a problem since what calls `lazy_array_equiv` is `Variable.identical` and if `identical` works correctly this should, too.

dcherian:
OK let's see what @shoyer thinks

keewis:
gentle ping, @shoyer

keewis:
these issues (`rolling_window` and `identical`) seem too big to discuss here, so in order to keep moving forward let's mark these tests as xfail and discuss this in their own issues / PRs.

dcherian:
> let's mark these tests as xfail and discuss this in their own issues / PRs.

:+1: to smaller PRs!

keewis:
this should be ready for review and merge once the tests pass.

keewis:
it should be ready now

xarray/core/variable.py
```diff
@@ -1099,24 +1103,21 @@ def _shift_one_dim(self, dim, count, fill_value=dtypes.NA):
...
         if isinstance(trimmed_data, dask_array_type):
-            chunks = list(trimmed_data.chunks)
-            chunks[axis] = (shape[axis],)
-            full = functools.partial(da.full, chunks=chunks)
+            pad_func = da.pad
         else:
-            full = np.full
-
-        filler = full(shape, fill_value, dtype=dtype)
+            pad_func = np.pad
```
shoyer: We try to consolidate all our duck typing aliases into duck_array_ops.py. Can you make a `pad` function there instead?

xarray/core/duck_array_ops.py
```diff
@@ -261,7 +261,7 @@ def where_method(data, cond, other=dtypes.NA):
 
 
 def fillna(data, other):
-    return where(isnull(data), other, data)
+    return where(notnull(data), data, other)
```
shoyer: Is there a reason why you had to flip this?

xarray/core/variable.py
```diff
@@ -742,7 +742,11 @@ def _getitem_with_mask(self, key, fill_value=dtypes.NA):
 
             data = as_indexable(self._data)[actual_indexer]
             mask = indexing.create_mask(indexer, self.shape, data)
-            data = duck_array_ops.where(mask, fill_value, data)
+            if isinstance(mask, bool):
+                mask = not mask
+            else:
+                mask = ~mask
```
shoyer: Could you flip the argument order rather than adding this? I’m a little puzzles here.

shoyer:
Thanks for pinging me again here (I get a lot of GitHub notifications). `identical` is an interesting case!

I think the current behavior (1 meter is identical to 100 centimeters) is arguably consistent with how `identical` currently works, which only check equality between array elements.

Right now, `identical` considers numbers of different data types equal, e.g., int `1` is identical to float 1.0``. I think units arguably have a similar to role to data types -- hopefully eventually libraries like `pint` could be implemented via custom NumPy dtypes, rather than needing to reimplement all of NumPy.

Did this come up in the context of some other downstream use-case, or is this just something that occurred to you for the sake of consistency?

xarray/core/variable.py
shoyer: If the concern here is about consistency when applying `~` to `bool` objects and boolean dtype arrays, explicitly calling `np.logical_not` is a good alternative.

But it does feel a little weird to me to see this here. Maybe changing `duck_array_ops.notnull` would have the same effect?

xarray/core/variable.py
keewis: sure

xarray/core/variable.py
keewis: same as the `fillna` issue above, in order to get the results in the units of `data`, we need to flip the arguments and for that I need invert the mask (if there is a different way to flip the arguments without inverting, please do tell me).

I tried to use `mask = ~mask`, but `~` does not work as expected for `bool`. I'll use `np.logical_not` instead.

keewis:
when writing the unit tests, we decided the definition of `equals` to be different from `identical` (see https://github.com/pydata/xarray/pull/3238#issuecomment-533145925 and https://github.com/pydata/xarray/pull/3238#issuecomment-533181017) which would be consistent with the behaviour of `identical` when a `"units"` attribute is set.

As far as I know there was no real use case (except from being able to use `assert_identical` in the unit tests), so we can change that.

cc @jthielen

xarray/core/duck_array_ops.py
shoyer: Is this worth fixing in pint? It seems like it would make sense to determine units in `where` a symmetrical way, similar to operations like `+`.

(We can keep this change for now -- I don't see much downside -- but please add a comment documenting why, so we will know when or if this could be changed in the future)

keewis: actually, the current implementation of `where` is consistent with the behaviour of `+`:
```python
In [1]: import pint 
   ...: import numpy as np 
   ...: ureg = pint.UnitRegistry() 
   ...:  
   ...: a = [12, 10, 9] * ureg.m 
   ...: b = [15, 9, 11] * ureg.mm 
   ...: a + b, b + a
Out[1]: 
(array([12.015, 10.009,  9.011]) <Unit('meter')>,
 array([12015., 10009.,  9011.]) <Unit('millimeter')>)
```

comment: done

jthielen: Pint's implementation of `where` indeed behaves like `+` by using the units of the first argument:

```python
a = [[1, 2], [3, 4]] * ureg.m
b = [[-100, -200], [-300, -400]] * ureg.cm
m = np.array([[True, False], [False, True]])
print(a + b)
print(b + a)
print(np.where(m, a, b))
print(np.where(m, b, a))
```
```
[[0.0 0.0] [0.0 0.0]] meter
[[0.0 0.0] [0.0 0.0]] centimeter
[[1.0 -2.0] [-3.0 4.0]] meter
[[-100.0 200.0] [300.0 -400.0]] centimeter
```

(this only really matters though if the `identical` check is particular about exact unit as discussed below)

shoyer: I'm a little surprised that arithmetic is not commutative in pint, but at least it's internally consistent! I suspect this is likely to cause issues in other places where we use duck typing.

I guess this is a bigger issue, but I think it would be worth considering adding some sort of unit casting hierarchy into pint's unit registries. It almost doesn't matter what exactly the rules are as long as it exists.

For a library like xarray that rely upon duck typing, it is nice to be able to rely upon properties like arithmetic being commutative to reduce complexity. When really needed we can add special case logic, but it means untested code paths are likely to break in unexpected ways.

jthielen:
@keewis For what it is worth, as far as `identical` goes, I think it makes the most sense to treat unit matching like dtype matching as @shoyer mentioned. Although, I had interpreted @max-sixty's comment https://github.com/pydata/xarray/pull/3238#issuecomment-533181017 to mean that dtypes are compared, it appears from @shoyer's comment https://github.com/pydata/xarray/pull/3706#issuecomment-583259053 that this not the case. If strict unit checking is required, I think that may be better served by an additional `assert unit == "meter"` type statement.

xarray/core/duck_array_ops.py
jthielen: @shoyer That's a good point! Having
```python
(a + b).units == (b + a).units
```
(and likewise for other commutative bin-ops) would be a good invariant. I'll raise an issue on Pint.

keewis:
> If strict unit checking is required, I think that may be better served by an additional `assert unit == "meter"` type statement.

which is what I've been doing with `assert_units_equal`. I'll change the tests for `identical`, then.

Also, concerning the commutative operations: should we wait for hgrecco/pint#1019 and remove the flipped parameters or should we merge as is and possibly revert once `pint` implements a type casting hierarchy?

shoyer:
> Also, concerning the commutative operations: should we wait for [hgrecco/pint#1019](https://github.com/hgrecco/pint/issues/1019) and remove the flipped parameters or should we merge as is and possibly revert once `pint` implements a type casting hierarchy?

I don't anticipate any performance cost to this, just a small decrease in readability. So I think this is fine to merge for now with comments in the relevant sections and we can revert it later. My only suggestion is to add a note like `TODO: revert after https://github.com/hgrecco/pint/issues/1019 is fixed` to each comment.

max-sixty:
> Although, I had interpreted @max-sixty's comment [#3238 (comment)](https://github.com/pydata/xarray/pull/3238#issuecomment-533181017) to mean that dtypes are compared, it appears from @shoyer's comment [#3706 (comment)](https://github.com/pydata/xarray/pull/3706#issuecomment-583259053) that this not the case. 

I was wrong; I should have at least realized I didn't know.  Apologies if that caused wasted time @jthielen 

Separately: should `assert_identical` assert that the dtypes are the same? I'd have thought there should be some way of testing whether dtypes are consistent with expectations, and I'd have thought `assert_identical` would be it?

xarray/tests/test_units.py
```diff
@@ -19,6 +21,13 @@
 unit_registry = pint.UnitRegistry(force_ndarray=True)
 Quantity = unit_registry.Quantity
 
+
+# silence pint's BehaviorChangeWarning
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    Quantity([])
```
keewis: this is not necessary now that `pint=0.11` has been released which does not warn

keewis:
@shoyer, I added the notes

@max-sixty, @jthielen: the `identical` tests will be skipped for now. At the moment that does not make any difference since `identical` is the same as `equals` with additionally checking the attributes (any changes to it are not limited to `pint`, so I guess we should open a new issue for discussion).

dcherian:
Test failures look unrelated. Thanks @keewis

PR #3707, opened 2020-01-18
remove PR pre-black instructions

max-sixty:
I don't think these should be needed any longer


max-sixty:
No real rush, but we could merge pre-0.15?

crusaderky:
agree

PR #3714, opened 2020-01-22
setuptools-scm and one-liner setup.py

crusaderky:
- Closes #3369
- Replace versioneer with setuptools-scm
- Replace setup.py with setup.cfg
- Drop pytest-runner as instructed by deprecation notice on the project webpage

PR #3720, opened 2020-01-24
setuptools-scm and isort tweaks

crusaderky:
Follow-up on https://github.com/pydata/xarray/pull/3714

- Fix regression in mypy if pip creates a zipped archive
- Avoid breakage in the extremely unlikely event that setuptools is not installed
- Guarantee ``xarray.__version__`` to be always PEP440-compatible. This prevents a breakage if you run pandas without xarray installed and with the xarray sources folder in PYTHONPATH.
- Apply isort to ``xarray.__init__``

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -75,7 +75,7 @@ classifiers =
 
 [options]
 packages = xarray
-zip_safe = True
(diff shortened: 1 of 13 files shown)


max-sixty:
Great!

dcherian:
nice.

PR #3721, opened 2020-01-24
Add isort to CI

crusaderky:


crusaderky:
Demo isort CI in action: https://github.com/pydata/xarray/pull/3721/checks?check_run_id=406904833

max-sixty:
+0.2, and we can revert if people really complain.

If there were a tool to autocorrect PRs (i.e. not just fail CI, but offer to add commits to correct the problems), that would be great

crusaderky:
@max-sixty I wouldn't 100% advise using such level of automation for isort. Have a look at the ``# isort:skip`` tags in our codebase; they're all cases where isort would otherwise break the code.

max-sixty:
Yes; I'm particularly thinking of someone making their first contribution, maybe without all the linters and tools locally, and getting stymied by a slightly incorrect import. 

Ofc the tool would have to respect `isort:skip`...

crusaderky:
@max-sixty that's what PULL_REQUEST_TEMPLATE.md is for...

max-sixty:
I know, I know :)

Potentially we have different priors on whether people might get stuck on all the setup. I hope there's no one, and if there is we can decide what to do!

max-sixty:
Just in case my previous comments were unclear, I think we should merge this! 
(and then reassess if we get complaints, which I give ~20% likelihood)

max-sixty:
Thank you @crusaderky

dcherian:
OK. thanks @crusaderky

Issue #3711, opened 2020-01-21
PseudoNetCDF tests failing randomly


keewis:
The `py37-windows` CI seems to fail for newer PRs:
```pytb
_______________ TestPseudoNetCDFFormat.test_uamiv_format_write ________________

self = <xarray.tests.test_backends.TestPseudoNetCDFFormat object at 0x000002E11FF2DC08>

    def	test_uamiv_format_write(self):
	fmtkw =	{"format": "uamiv"}

	expected = open_example_dataset(
            "example.uamiv", engine="pseudonetcdf", backend_kwargs=fmtkw
	)
	with self.roundtrip(
	    expected,
            save_kwargs=fmtkw,
	    open_kwargs={"backend_kwargs": fmtkw},
            allow_cleanup_failure=True,
	) as actual:
>           assert_identical(expected, actual)

xarray\tests\test_backends.py:3532:
_ _ _ _	_ _ _ _	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _
xarray\core\formatting.py:628: in diff_dataset_repr
    summary.append(diff_attrs_repr(a.attrs, b.attrs, compat))
_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _	_ _ _ _

a_mapping = {'CPROJ': 0, 'FILEDESC': 'CAMx            ', 'FTYPE': 1, 'GDNAM': 'CAMx            ', ...}
b_mapping = {'CPROJ': 0, 'FILEDESC': 'CAMx            ', 'FTYPE': 1, 'GDNAM': 'CAMx            ', ...}
compat = 'identical', title = 'Attributes'
summarizer = <function summarize_attr at 0x000002E1156813A8>, col_width = None

    def _diff_mapping_repr(a_mapping, b_mapping, compat, title, summarizer, col_width=None):
        def extra_items_repr(extra_keys, mapping, ab_side):
            extra_repr = [summarizer(k, mapping[k], col_width) for k in extra_keys]
            if extra_repr:
                header = f"{title} only on the {ab_side} object:"
                return [header] + extra_repr
            else:
                return []

	a_keys = set(a_mapping)
        b_keys = set(b_mapping)

        summary = []

	diff_items = []

        for k in a_keys & b_keys:
	    try:
                # compare xarray variable
[...]

Issue #3697, opened 2020-01-15
documentation build issues on RTD


keewis:
It seems we have (seemingly) random failures on RTD.

Some of these are the known memory issue: recreating my doc environment used about 1.4 GB of RAM, which might be too much for RTD, even with the extended memory.

Much more often is a timeout when building the docs but I can't reproduce them locally. Any ideas? 
Edit: This really is random, I tried rerunning and the build passed.

Also, a warning:
```
proj_create: init=epsg:/init=IGNF: syntax not supported in non-PROJ4 emulation mode
```

rabernat:
Many projects have moved away from RTD for this reason. We can easily build the docs in travis and then use [doctr](https://github.com/drdoctr/doctr) to deploy them.

Is there a downside to this?

keewis:
not sure on downsides, but I think we could use this to provide a documentation preview for PRs?

crusaderky:
Very glad to upvote anything that rids us of the RTD CI!

crusaderky:
The obvious downside is that anybody with a link to one of the internal pages of our documentation will have the link broken. Also I'm unsure how straightforward it will be to rebuild all of our historical versions.

keewis:
it may be that the timeouts are not caused by RTD: I have been trying to build the documentation several times and it sometimes pauses while trying to read / build (?) `dask.rst`. This is the traceback if I send a `KeyboardInterrupt`:
```pytb
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-4-2ef53683336b> in <module>
----> 1 ds.to_netcdf('manipulated-example-data.nc')

.../xarray/core/dataset.py in to_netcdf(self, path, mode, format, group, engine, encoding, unlimited_dims, compute, invalid_netcdf)
   1543             unlimited_dims=unlimited_dims,
   1544	            compute=compute,
-> 1545		    invalid_netcdf=invalid_netcdf,
   1546		)
   1547

.../xarray/backends/api.py in to_netcdf(dataset, path_or_file, mode, format, group, engine, encoding, unlimited_dims, compute, multifile, invalid_netcdf)
   1095             return writer, store
   1096
-> 1097		writes = writer.sync(compute=compute)
   1098
   1099	       	if path_or_file is None:

.../xarray/backends/common.py in sync(self, compute)
    202			compute=compute,
    203                 flush=True,
--> 204			regions=self.regions,
    205             )
    206		    self.sources = []

[...]

PR #3730, opened 2020-01-30
Remove extra && in PR template

max-sixty:


keewis:
I never realized we had that, probably because I never tried to copy / paste the command. Looks good to me.

The build failures are fine examples for both #3711 and #3697 which means we should probably try to fix both as soon as possible.

PR #3737, opened 2020-01-31
Fix/rtd

dcherian:
1. python 3.8 is not allowed on RTD (yet)
2. I  pinned a few versions (based on the env solution obtained locally). This seems to have fixed the memory problem.

### DIFF
--- a/ci/requirements/doc.yml
+++ b/ci/requirements/doc.yml
@@ -6,20 +6,21 @@ dependencies:
   - python=3.7
   - bottleneck
   - cartopy
-  - cfgrib
-  - h5netcdf
+  - cfgrib>=0.9
+  - dask>=2.10
+  - h5netcdf>=0.7.4
   - ipykernel
   - ipython
-  - iris
+  - iris>=2.3
   - jupyter_client
   - nbsphinx
-  - netcdf4
+  - netcdf4>=1.5
   - numba
-  - numpy
+  - numpy>=1.17
   - numpydoc
-  - pandas
-  - rasterio
+  - pandas>=1.0
+  - rasterio>=1.1
   - seaborn
-  - sphinx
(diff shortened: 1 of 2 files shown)


keewis:
you could also pip install everything except `cartopy` and `cf-units` (which is a dependency of `scitools-iris`, the package name for `iris` on PyPI). The memory usage on my machine is at about 700MB, which should help

keewis:
also, you can just remove the whole `python` section in `readthedocs.yml`, it is not required and since we use `conda` we use our own python version anyway.

dcherian:
The docs build with this branch did work: https://readthedocs.org/projects/xray/builds/10362278/ so shall we merge this to stable and then work to pip install things on master?

keewis:
I'd think so. What do we do about the build for `v0.15.0`?

dcherian:
> What do we do about the build for v0.15.0?

Not sure. We could figure out the `pip install` bit and release 0.15.1. In the meantime, once this PR is merged to stable, the docs will still be accurate for 0.15.0.

dcherian:
I'll merge when this build succeeds: https://readthedocs.org/projects/xray/builds/10362524/

keewis:
looking at the way you pinned the versions (using `>=`) this should be fine as is, no need for switching to pip.

I have also been experimenting with using channels like
```
https://metachannel.conda-forge.org/conda-forge/iris,netcdf4,numpy,numba,pandas,cartopy
```
which in theory should reduce the downloaded `repodata.json` and thus reduce the used memory, but I probably did not use the correct url: the used memory increased instead.

Another way to reduce it could be using `mamba` which reimplements the solver in C++, but it's experimental right now and it's probably also something that has to be installed by RTD, not us (it needs to be installed before creating the environment).

PR #3753, opened 2020-02-05
Add new h5netcdf backend phony_dims kwarg

kmuehlbauer:
This adds support for the new `phony_dims` keyword for the h5netcdf backend which is available from `h5netcdf` version 0.8.0.

 - [ ] Closes #xxxx
 - [ ] Tests added

### DIFF
--- a/xarray/backends/h5netcdf_.py
+++ b/xarray/backends/h5netcdf_.py
@@ -1,4 +1,5 @@
 import functools
+from distutils.version import LooseVersion
 
 import numpy as np
 
@@ -117,6 +118,7 @@ class H5NetCDFStore(WritableCFDataStore):
         lock=None,
         autoclose=False,
         invalid_netcdf=None,
+        phony_dims=None,
     ):
         import h5netcdf
 
@@ -124,6 +126,14 @@ class H5NetCDFStore(WritableCFDataStore):
             raise ValueError("invalid format for h5netcdf backend")
 
         kwargs = {"invalid_netcdf": invalid_netcdf}
+        if phony_dims is not None:
+            if LooseVersion(h5netcdf.__version__) >= LooseVersion("0.8.0"):
+                kwargs["phony_dims"] = phony_dims
+            else:
+                raise (
+                    "h5netcdf backend keyword argument 'phony_dims' needs "
+                    "h5netcdf >= 0.8.0."
+                )
 
         if lock is None:
             if mode == "r":
--- a/xarray/backends/h5netcdf_.py
+++ b/xarray/backends/h5netcdf_.py
@@ -130,7 +130,7 @@ class H5NetCDFStore(WritableCFDataStore):
             if LooseVersion(h5netcdf.__version__) >= LooseVersion("0.8.0"):
                 kwargs["phony_dims"] = phony_dims
             else:
-                raise (
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


xarray/backends/h5netcdf_.py
```diff
@@ -117,13 +118,16 @@ def open(
...
         invalid_netcdf=None,
+        phony_dims=None,
     ):
         import h5netcdf
 
         if format not in [None, "NETCDF4"]:
             raise ValueError("invalid format for h5netcdf backend")
 
         kwargs = {"invalid_netcdf": invalid_netcdf}
+        if LooseVersion(h5netcdf.__version__) >= LooseVersion("0.8.0"):
```
dcherian: Let's raise an error otherwise when this condition is not satisfied

xarray/backends/h5netcdf_.py
```diff
@@ -117,13 +118,16 @@ def open(
...
+            kwargs["phony_dims"]: phony_dims
```
mathause: Should this be
```suggestion
            kwargs["phony_dims"] = phony_dims
```
?

xarray/backends/h5netcdf_.py
kmuehlbauer: @dcherian We can do that, but that means users have to use h5netcf >=0.8.0 (which should be OK, I think). Not sure if we should force them. What's the xarray policy for that, BTW? 

Anyway, the whole thing is implemented in h5netcf to be fully backwards compatible. So it should not interfere with users using older versions of h5netcdf.

I'll add the suggested raise first thing tomorrow, if no one stops me.

xarray/backends/h5netcdf_.py
kmuehlbauer: @mathause thanks for spotting, you're right.

xarray/backends/h5netcdf_.py
dcherian: Sorry I wasn't clear enough. What I meant was if a user passes `phony_dims` we should raise if the `h5netcdf` version is not >= 0.8.0

kmuehlbauer: @dcherian Ah, OK, that makes absolutely sense. Will do.

kmuehlbauer:
Changes added as suggested in the review comments. Sorry for excessive use of force-pushing.

xarray/backends/h5netcdf_.py
```diff
@@ -117,13 +118,22 @@ def open(
...
 
         if format not in [None, "NETCDF4"]:
             raise ValueError("invalid format for h5netcdf backend")
 
         kwargs = {"invalid_netcdf": invalid_netcdf}
+        if phony_dims is not None:
+            if LooseVersion(h5netcdf.__version__) >= LooseVersion("0.8.0"):
+                kwargs["phony_dims"] = phony_dims
+            else:
+                raise (
```
dcherian: raise `ValueError`?

kmuehlbauer: @dcherian Sure. I should take more time and be more careful...

kmuehlbauer:
@dcherian This is ready for another review. Please let me know, if I can be of any further assistance.

dcherian:
LGTM. I'll merge in a couple of days if no one else has any comments.

dcherian:
Thanks @kmuehlbauer

Issue #2723, opened 2019-01-29
Implementation of CRS storage in rasterio with PROJ.4 & WKT


snowman2:
Continuation of the discussion from #2722 to move onto implementation details.

snowman2:
Tag: @fmaussion @djhoese

fmaussion:
Sorry for being late on this one. Yes, I agree that the "coordinate attribute" is worth a try.

To avoid redundancy in the name one could name the coord variable attributes simply `wkt` and `proj4`, like:

```
<xarray.DataArray 'crs' ()>
array(0)
Coordinates:
    crs          int64 0
Attributes:
    wkt:  PROJCS["UTM Zone 15, Northern Hemisphere",GEOGCS["WGS 84",D...
    proj4: +proj=...
```

snowman2:
@fmaussion just a ping for what are your thoughts are for next steps on this?

snowman2:
Going to close this as it is over a year old.

For reference, WKT support is in `rioxarray`.
https://corteva.github.io/rioxarray/stable/examples/crs_management.html

dcherian:
@snowman2 @djhoese It'd be nice to update our docs to link to rioxarray and other relevant packages. (e.g. here https://xarray.pydata.org/en/stable/io.html#rasterio).

Adding an example on doing these manipulations would also be very welcome.

snowman2:
> Adding an example on doing these manipulations would also be very welcome.

What types of manipulations are you thinking about? Reprojection? Export to geotiff? Clip? Reading in files?

fmaussion:
I think we should seriously consider removing all rasterio logic in xarray and make rioxarray an optional dependency to read geotiff files with xarray.

dcherian:
@snowman2 Any/all of those "common operations" sounds good. We don't have a single geoTIFF example: https://xarray.pydata.org/en/stable/examples.html so at this point anything would be a big improvement

> I think we should seriously consider removing all rasterio logic in xarray and make rioxarray an optional dependency to read geotiff files with xarray.

:+1: Though maybe we should wait till the backends refactor is done so that things are a little more easy.

snowman2:
@dcherian Maybe a link to these examples in that section then?
https://corteva.github.io/rioxarray/stable/examples/examples.html

PR #3757, opened 2020-02-06
DOC: Add rioxarray and other external examples

snowman2:
- [x] Addresses https://github.com/pydata/xarray/issues/2723#issuecomment-582961469

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -10,6 +10,7 @@ Examples
     examples/visualization_gallery
     examples/ROMS_ocean_model
     examples/ERA5-GRIB-example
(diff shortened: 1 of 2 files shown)


snowman2:
> We might want to think about adding rioxarray as documentation build dependency

That brings along with it the fun of GDAL as a dependency.

dcherian:
Yeah lets avoid GDAL ;)

The I/O section looks good now

image

dcherian:
Docs build passed. Merging. Thanks @snowman2

snowman2:
Thanks for the fixes 👍

PR #3769, opened 2020-02-13
concat now handles non-dim coordinates only present in one dataset

dcherian:
- [x] Tests added

```
da1 = xr.DataArray([1, 2, 3], dims="x", coords={"x": [1, 2, 3], "y": 1})
da2 = xr.DataArray([4, 5, 6], dims="x", coords={"x": [4, 5, 6]})
xr.concat([da1, da2], dim="x")
```
This use case is quite common since you can get `da1` from something like `bigger_da1.sel(y=1)`

On master this raises an uninformative  `KeyError` because `'y'` is not present in all datasets. This is because `coords="different"` by default which means that we are checking for equality. However `coords='different'`(and the equality check) is meaningless when the variable is only present in one of the objects to be concatenated.

This PR skips equality checking when a variable is only present in one dataset and raises a nicer error message when it is present in more than one but not all datasets.

### DIFF
--- a/xarray/tests/test_combine.py
+++ b/xarray/tests/test_combine.py
@@ -365,9 +365,9 @@ class TestNestedCombine:
         expected = Dataset({"x": ("a", [0, 1]), "y": ("a", [0, 1])})
         assert_identical(expected, actual)
 
-        objs = [Dataset({"x": [0], "y": [0]}), Dataset({"x": [0]})]
+        objs = [Dataset({"x": [0], "y": [0]}), Dataset({"x": [1]})]
         actual = combine_nested(objs, concat_dim="x")
-        expected = Dataset({"x": [0, 0], "y": [0]})
+        expected = Dataset({"x": [0, 1], "y": [0]})
         assert_identical(expected, actual)
 
     @pytest.mark.parametrize(
--- a/xarray/tests/test_combine.py
+++ b/xarray/tests/test_combine.py
@@ -366,8 +366,9 @@ class TestNestedCombine:
         assert_identical(expected, actual)
 
         objs = [Dataset({"x": [0], "y": [0]}), Dataset({"x": [0]})]
-        with pytest.raises(KeyError):
-            combine_nested(objs, concat_dim="x")
+        actual = combine_nested(objs, concat_dim="x")
+        expected = Dataset({"x": [0, 0], "y": [0]})
+        assert_identical(expected, actual)
 
     @pytest.mark.parametrize(
         "join, expected",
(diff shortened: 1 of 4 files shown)


xarray/tests/test_combine.py
```diff
@@ -366,8 +366,9 @@ def test_nested_concat(self):
         assert_identical(expected, actual)
 
         objs = [Dataset({"x": [0], "y": [0]}), Dataset({"x": [0]})]
-        with pytest.raises(KeyError):
-            combine_nested(objs, concat_dim="x")
+        actual = combine_nested(objs, concat_dim="x")
+        expected = Dataset({"x": [0, 0], "y": [0]})
```
dcherian: seems sensible to me.

xarray/tests/test_concat.py
```diff
@@ -475,3 +475,24 @@ def test_concat_attrs_first_variable(attr1, attr2):
...
+
+    with raises_regex(ValueError, "'y' is not present in all datasets."):
+        concat([da1, da2], dim="x", coords="all")
+
+    da1 = DataArray([1, 2, 3], dims="x", coords={"x": [1, 2, 3], "y": 1})
+    da2 = DataArray([4, 5, 6], dims="x", coords={"x": [4, 5, 6]})
+    da3 = DataArray([7, 8, 9], dims="x", coords={"x": [7, 8, 9], "y": 1})
+    for coords in ["different", "all"]:
+        with raises_regex(ValueError, "'y' not present in all datasets"):
+            concat([da1, da2, da3], dim="x")
```
TomNicholas: Is there a reason why one of these regexes is looking for `"'y' is not present"` and the other looking for `"'y' not present"`? Shouldn't the latter one be failing?

xarray/tests/test_combine.py
TomNicholas: In general if you concatenate two different numbers then you're also verifying the order was retained, but it's not really important.

dcherian: good point. Fixed.

xarray/tests/test_concat.py
dcherian: Nice catch!

It's failing in two different places. Since we eventually want to get rid of this limitation (#508), shall we just leave this as is?

TomNicholas: So the tests are catching the intended errors, and will need to be rewritten as part of #508 anyway? In that case then yeah this seems fine to leave like this.

dcherian:
Thanks for the review @TomNicholas

PR #3792, opened 2020-02-22
Enable pandas-style rounding of cftime.datetime objects

spencerkclark:
- [x] Tests added

This is particularly useful for removing microsecond noise that can sometimes be added from decoding times via `cftime.num2date`, though also applies more generally.  The methods used here for rounding dates in the integer domain are copied from pandas.

On a somewhat more internal note, this adds an `asi8` property to `CFTimeIndex`, which encodes the dates as integer values representing microseconds since 1970-01-01; this encoding is made exact via the `exact_cftime_datetime_difference` function.  It's possible this could be useful in other contexts.

Some examples:

```
In [1]: import xarray as xr

In [2]: times = xr.cftime_range("2000", periods=5, freq="17D")

In [3]: time = xr.DataArray(times, dims=["time"], name="time")

In [4]: time.dt.floor("11D")
Out[4]:
<xarray.DataArray 'floor' (time: 5)>
array([cftime.DatetimeGregorian(1999-12-31 00:00:00),
       cftime.DatetimeGregorian(2000-01-11 00:00:00),
       cftime.DatetimeGregorian(2000-02-02 00:00:00),
       cftime.DatetimeGregorian(2000-02-13 00:00:00),
       cftime.DatetimeGregorian(2000-03-06 00:00:00)], dtype=object)
Coordinates:
  * time     (time) object 2000-01-01 00:00:00 ... 2000-03-09 00:00:00

In [5]: time.dt.ceil("11D")
Out[5]:
<xarray.DataArray 'ceil' (time: 5)>
array([cftime.DatetimeGregorian(2000-01-11 00:00:00),
       cftime.DatetimeGregorian(2000-01-22 00:00:00),
       cftime.DatetimeGregorian(2000-02-13 00:00:00),
       cftime.DatetimeGregorian(2000-02-24 00:00:00),
       cftime.DatetimeGregorian(2000-03-17 00:00:00)], dtype=object)
Coordinates:
  * time     (time) object 2000-01-01 00:00:00 ... 2000-03-09 00:00:00

In [6]: time.dt.round("11D")
Out[6]:
<xarray.DataArray 'round' (time: 5)>
array([cftime.DatetimeGregorian(1999-12-31 00:00:00),
       cftime.DatetimeGregorian(2000-01-22 00:00:00),
       cftime.DatetimeGregorian(2000-02-02 00:00:00),
       cftime.DatetimeGregorian(2000-02-24 00:00:00),
       cftime.DatetimeGregorian(2000-03-06 00:00:00)], dtype=object)
Coordinates:
  * time     (time) object 2000-01-01 00:00:00 ... 2000-03-09 00:00:00
```

### DIFF
--- a/xarray/coding/cftimeindex.py
+++ b/xarray/coding/cftimeindex.py
@@ -666,7 +666,7 @@ def _total_microseconds(delta):
     -------
     int
     """
-    return int(delta.total_seconds() * 1e6)
+    return delta / timedelta(microseconds=1)
 
 
 def _floor_int(values, unit):
--- a/xarray/coding/cftimeindex.py
+++ b/xarray/coding/cftimeindex.py
@@ -528,6 +528,45 @@ class CFTimeIndex(pd.Index):
         """
         return pd.Index([date.strftime(date_format) for date in self._data])
 
+    @property
+    def asi8(self):
+        """Convert to integers with units of microseconds since 1970-01-01"""
+        from ..core.resample_cftime import exact_cftime_datetime_difference
+
+        epoch = self.date_type(1970, 1, 1)
+        return np.array(
+            [
+                _total_microseconds(exact_cftime_datetime_difference(epoch, date)) for date
+                in self.values
+            ]
+        )
+
+    def _round(self, freq, method):
+        from .cftime_offsets import CFTIME_TICKS, to_offset
+
+        offset = to_offset(freq)
+        if not isinstance(offset, CFTIME_TICKS):
+            raise ValueError(
+                f"{offset} is a non-fixed frequency"
+            )
+
+        unit = _total_microseconds(offset.as_timedelta())
+        values = self.asi8
+        rounded = method(values, unit)
+        return _cftimeindex_from_i8(rounded, self.date_type, self.name)
+
+    def floor(self, freq):
+        """Round dates down to fixed frequency"""
+        return self._round(freq, _floor_int)
+
+    def ceil(self, freq):
+        """Round dates up to fixed frequency"""
+        return self._round(freq, _ceil_int)
+
+    def round(self, freq):
+        """Round dates to a fixed frequency"""
+        return self._round(freq, _round_to_nearest_half_even)
+
 
 def _parse_iso8601_without_reso(date_type, datetime_str):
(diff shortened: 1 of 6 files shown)


xarray/coding/cftimeindex.py
```diff
@@ -528,6 +528,83 @@ def strftime(self, date_format):
         """
         return pd.Index([date.strftime(date_format) for date in self._data])
 
+    @property
+    def asi8(self):
```
dcherian: This does seem useful. Shall we list it as public API?

xarray/tests/test_accessor_dt.py
```diff
@@ -435,3 +436,94 @@ def test_seasons(cftime_date_type):
...
+        [
+            [cftime_date_type(1, 1, 1, 0), cftime_date_type(1, 1, 1, 0)],
+            [cftime_date_type(1, 1, 1, 0), cftime_date_type(1, 1, 2, 0)],
+        ],
+        name="floor",
+    )
+
+    if use_dask:
+        chunks = {"dim_0": 1}
+        result = cftime_rounding_dataarray.chunk(chunks).dt.floor(freq)
```
dcherian: Can we surround this with `raise_if_dask_computes` imported from `test_dask.py` just to be sure no computes happen indirectly?

xarray/coding/cftimeindex.py
spencerkclark: Yeah, that sounds like a good idea.  It looks like it is [already listed in the docs for `CFTimeIndex`](http://xarray.pydata.org/en/stable/generated/xarray.CFTimeIndex.html) (despite the fact that it currently returns `None`); is there anything else that I would need to do?  

I added a couple tests for good measure (including one that turned up a floating point issue that I fixed).

xarray/tests/test_accessor_dt.py
spencerkclark: Good idea.  It happens that we do one minor compute to check if the array contains `cftime.datetime` objects before proceeding through the `dt` accessor; I added a comment noting this in the tests.

dcherian:
Thanks @spencerkclark

PR #3797, opened 2020-02-25
install most documentation requirements using pip

### LINKED ISSUE #3796: RTD failing yet again
memory consumption errors as usual.

@keewis I remember you had an idea for using pip instead of conda?

keewis:
RTD seems to have quite a lot of memory issues. In theory, they could upgrade to allow up to 2GB of RAM per project, but since `conda-forge` will probably continue to grow in the future, that would only be a short-term fix. `pip` does not have this issue (at least not as much), but we cannot solely rely on it since we also depend on packages that either don't provide wheels (`cartopy` and `cf-units`) or are not available on PyPI (`hdf5`?).

At the moment, the builds still fail, we probably need to add non-python dependencies such as `hdf5`.

 - [x] Closes #3796

### DIFF
--- a/ci/requirements/doc.yml
+++ b/ci/requirements/doc.yml
@@ -4,23 +4,25 @@ channels:
   - conda-forge
 dependencies:
   - python=3.8
-  - bottleneck
+  - pip
   - cartopy
-  - cfgrib
-  - h5netcdf
-  - ipykernel
-  - ipython
-  - iris
-  - jupyter_client
-  - nbsphinx
-  - netcdf4
-  - numba
-  - numpy
-  - numpydoc
-  - pandas
-  - rasterio
-  - seaborn
-  - setuptools
-  - sphinx
-  - sphinx_rtd_theme
-  - zarr
+  - pip:
+    - bottleneck
+    - cfgrib
+    - h5netcdf
+    - ipykernel
+    - ipython
+    - scitools-iris
+    - jupyter_client
+    - nbsphinx
+    - netcdf4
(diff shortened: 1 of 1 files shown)


max-sixty:
LGTM. Thanks for working through this @keewis ...

keewis:
this turns out to be really difficult: we need a few dependencies to be installed via `conda` (`eccodes`, `hdf5` and `cf-units`, in addition to `pip` and `cartopy`), which already consumes too much memory. Also, the pip-install step hits the memory issue, too. I assume that's because it is trying to compile something?

It seems the memory issues (pure `conda`) occur because the `latest` image was upgraded and now uses the latest version of `conda`, which apparently requires more memory. The official fixes for that are to request more memory via email or to switch to using the `stable` image (in that order, apparently).

max-sixty:
Let's definitely email before spending more time on this. Does anyone have an existing relationship with RTD? Otherwise @keewis feel free to reach out...

keewis:
I contacted them, let's see what they say

keewis:
The answer is
> We upgraded conda and it seems to be having some issues. I've tried to increase your memory all the way up to 1.8GB, and it still is failing. We're looking at a deploy to roll back the conda version, which should hopefully fix this.

so we probably have to wait until they finished the downgrading. If we're impatient, we can temporarily fix it by setting `build.image: stable`.

keewis:
It seems our builds are passing again. I am closing this PR since I think it would make the `doc` environment too complicated.

PR #3802, opened 2020-02-27
Raise on multiple string args to groupby

max-sixty:
- [x] Tests added

From the comment:

While we don't generally check the type of every arg, passing
multiple dimensions as multiple arguments is common enough, and the
consequences hidden enough (strings evaluate as true) to warrant
checking here.
A future version could make squeeze kwarg only, but would face
backward-compat issues.


max-sixty:
Ready for review!

dcherian:
LGTM

PR #3804, opened 2020-02-27
Allow chunk_store argument when opening Zarr datasets

ajelenak:
This is a simple modification to pass `chunk_store` argument to the zarr package when opening Zarr datasets.

I don't think this requires any documentation changes and new tests.


rsignell-usgs:
This PR is motivated by the work described in this [Medium blog post](https://medium.com/pangeo/cloud-performant-reading-of-netcdf4-hdf5-data-using-the-zarr-library-1a95c5c92314)

rabernat:
Fantastic, thanks @ajelenak so much for this contribution!

Two items come to mind:
- [ ] this will definitely need a test, just something simple to verify that the `chunk_store` argument works as expected
- [ ] curious what will happen if you pass an actual Zarr [storage](https://zarr.readthedocs.io/en/stable/api/storage.html#module-zarr.storage) object. I assume it will work fine, since these are all mutable mappings.

ajelenak:
@rabernat Added a test that uses an actual Zarr store.

xarray/tests/test_backends.py
```diff
@@ -1539,6 +1539,15 @@ def test_roundtrip_consolidated(self):
             self.check_dtypes_roundtripped(expected, actual)
             assert_identical(expected, actual)
 
+    def test_with_chunkstore(self):
+        save_kwargs = {}
+        expected = create_test_data()
+        with self.create_zarr_target() as store_target:
+            self.save(expected, store_target, **save_kwargs)
+            open_kwargs = {'chunk_store': store_target}
+            with self.open(store_target, **open_kwargs) as ds:
```
rabernat: Here it looks like the same underlying store is being used for both the main store and the chunk store. My concern here is that this does not actually test the functionality of the `chunk_store` keyword. Imagine if `chunk_store` were just a dummy argument that did nothing? The test would pass anyway, since all the required keys would be find in `store_target`.

This test also drives home the need for a `chunk_store` argument in `to_zarr` as well.

ajelenak: I added (9060925) `chunk_store` argument to the `to_zarr()` method so the test now uses separate Zarr stores.

rabernat:
Sorry for letting this hang so long. Now that https://github.com/zarr-developers/zarr-python/pull/557 has been merged and zarr 2.4.0 has been released, we should be able to finish this off.

The easiest thing would be to just require zarr > 2.4.0. However, we currently don't require zarr at all...it's an optional dependency. The code in this PR will fail for older versions of zarr. What's the best way to handle this? Having a bunch of internal checks for zarr versions feels clunky, but we've done it before.

xarray/backends/zarr.py
```diff
@@ -276,10 +276,12 @@ def open_group(
         group=None,
         consolidated=False,
         consolidate_on_close=False,
+        chunk_store=None,
     ):
         import zarr
 
-        open_kwargs = dict(mode=mode, synchronizer=synchronizer, path=group)
+        open_kwargs = dict(mode=mode, synchronizer=synchronizer, path=group,
+                           chunk_store=chunk_store)
```
jhamman, line 284: Regarding @rabernat's comment above, I think the best way forward might be to check the zarr version somewhere around here:

```python
if chunk_store is not None:
    if LooseVersion(zarr.__version__) < LooseVersion('2.4.0'):
        raise ValueError('...version not supported...')
```

dcherian: `zarr          2.3     (2019-03-25) 2.4     (2020-01-11) <`

I think we can just bump zarr to 2.4? It's almost 6 months old and so it would be in line with our "6 months for optional dependencies policy": https://xarray.pydata.org/en/stable/installing.html#minimum-dependency-versions

dcherian:
It'd be nice to add some documentation to https://xarray.pydata.org/en/stable/io.html but that can happen in a future PR.

xarray/backends/zarr.py
rabernat: > It's almost 6 months old and so it would be in line with our "6 months for optional dependencies policy":

According to [github](https://github.com/zarr-developers/zarr-python/releases), Zarr 2.4.0 was released on Apr. 28. In any case, the relevant commit (https://github.com/zarr-developers/zarr-python/commit/c18537d3695c7ad6fb582e880110749b5f761a77) was merged on Apr. 30. So I think we do need a new Zarr release.

rabernat: Or maybe we don't need a Zarr release. We might be able to test against Zarr master and use `LooseVersion(zarr.__version__) > LooseVersion('2.4.0')`.

shoyer: Indeed, our min deps script is currently broken. When I run it with https://github.com/pydata/xarray/pull/4178, I see:
`zarr          2.3     (2019-03-25) 2.3     (2019-03-25) =`

So we're supporting the right min version of zarr.

shoyer: The easiest way to do this without explicit version checks would probably be to only insert the argument it isn't the default value:
```python
        open_kwargs = dict(mode=mode, synchronizer=synchronizer, path=group)
        if chunk_size is not None:
            open_kwargs['chunk_store'] = chunk_store
```

ajelenak: I made the suggested change. (Thanks @rabernat and @shoyer!)

rsignell-usgs:
@dcherian , are we just waiting for one more "+1" here?

xarray/tests/test_sparse.py
```diff
@@ -1,889 +0,0 @@
-import pickle
```
jhamman, line 1: I don't think the removal of this file was intentional.

xarray/backends/zarr.py
```diff
@@ -563,6 +568,8 @@ def open_zarr(
     consolidated : bool, optional
         Whether to open the store using zarr's consolidated metadata
         capability. Only works for stores that have already been consolidated.
+    chunk_store : MutableMapping, optional
+        A separate Zarr store only for chunk data.
```
jhamman, line 572: Since we didn't explicitly handle the version requirement discussed [here](https://github.com/pydata/xarray/pull/3804#discussion_r445156318). I think we should indicate that this kwarg has a minimum zarr version requirement in the docstring.

xarray/tests/test_sparse.py
ajelenak: I restored this file when merging master into my branch.

xarray/backends/zarr.py
ajelenak: Done, in de8ee28e.

ajelenak:
Is there anything else to be done in this PR, or is it ready for merge? All tests pass.

rsignell-usgs:
Drumroll.... @dcherian....

xarray/tests/test_backends.py
```diff
@@ -1580,6 +1589,7 @@ def test_auto_chunk(self):
                 # only index variables should be in memory
                 assert v._in_memory == (k in actual.dims)
                 # chunk size should be the same as original
+
```
dcherian, line 1592:
```suggestion
```

dcherian:
Thanks for being patient @ajelenak I am excited to see this go in. welcome to xarray!

ajelenak:
Thanks @dcherian . And thanks to all who provided suggestions and comments in this PR.

PR #3807, opened 2020-02-28
Fix recombination in groupby when changing size along the grouped dimension

ej81:
When a a Dataset or DataArray is recombined after a groupby operation, xarray tries to restore the original coordinate. This causes problems if the group operation changes the size of the groups along the grouped dimension (e.g. some filtering operation that removes entries).

Restoring the original coordinate appears to be necessary only if it was squeezed from the groups, otherwise it is already restored by concatenation of the groups. This fix: 
1. avoids reordering if the size of the dimension has changed;
2. avoids restoring a coordinate that has not been squeezed out.

 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -35,6 +35,10 @@ New Features
 Bug fixes
 ~~~~~~~~~
 
+- Fix recombination of groups in :py:meth:`Dataset.groupby` and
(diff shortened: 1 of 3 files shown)


max-sixty:
(Also you can merge master for the tests to pass IYL)

max-sixty:
Thanks @ej81 !

PR #3816, opened 2020-03-02
Add template xarray object kwarg to map_blocks

### LINKED ISSUE #3575: map_blocks output inference problems
I am excited about using `map_blocks` to overcome a long-standing challenge related to calculating climatologies / anomalies with dask arrays. However, I hit what feels like a bug. I don't love how the new `map_blocks` function does this:

> The function will be first run on mocked-up data, that looks like ‘obj’ but has sizes 0, to determine properties of the returned object such as dtype, variable names, new dimensions and new indexes (if any).

The problem is that many functions will simply error on size 0 data. As in the example below

#### MCVE Code Sample

```python
import xarray as xr
ds = xr.tutorial.load_dataset('rasm').chunk({'y': 20})


def calculate_anomaly(ds):
    # needed to workaround xarray's check with zero dimensions
    #if len(ds['time']) == 0:
    #    return ds
    gb = ds.groupby("time.month")
    clim = gb.mean(dim='T')
    return gb - clim

xr.map_blocks(calculate_anomaly, ds)
```

Raises

```
/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/dataset.py in _construct_dataarray(self, name)
   1145         try:
-> 1146             variable = self._variables[name]
   1147         except KeyError:

KeyError: 'time.month'

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/parallel.py in infer_template(func, obj, *args, **kwargs)
     77     try:
---> 78         template = func(*meta_args, **kwargs)
     79     except Exception as e:

<ipython-input-40-d7b2b2978c29> in calculate_anomaly(ds)
      5     #    return ds
----> 6     gb = ds.groupby("time.month")
      7     clim = gb.mean(dim='T')

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/common.py in groupby(self, group, squeeze, restore_coord_dims)
    656         return self._groupby_cls(
--> 657             self, group, squeeze=squeeze, restore_coord_dims=restore_coord_dims
    658         )

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/groupby.py in __init__(self, obj, group, squeeze, grouper, bins, restore_coord_dims, cut_kwargs)
    298                 )
--> 299             group = obj[group]
    300             if len(group) == 0:

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/dataset.py in __getitem__(self, key)
   1235         if hashable(key):
-> 1236             return self._construct_dataarray(key)
   1237         else:

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/dataset.py in _construct_dataarray(self, name)
   1148             _, name, variable = _get_virtual_variable(
-> 1149                 self._variables, name, self._level_coords, self.dims
   1150             )

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/dataset.py in _get_virtual_variable(variables, key, level_vars, dim_sizes)
    157         else:
--> 158             data = getattr(ref_var, var_name).data
    159         virtual_var = Variable(ref_var.dims, data)

AttributeError: 'IndexVariable' object has no attribute 'month'

The above exception was the direct cause of the following exception:

Exception                                 Traceback (most recent call last)
<ipython-input-40-d7b2b2978c29> in <module>
      8     return gb - clim
      9 
---> 10 xr.map_blocks(calculate_anomaly, ds)

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/parallel.py in map_blocks(func, obj, args, kwargs)
    203     input_chunks = dataset.chunks
    204 
--> 205     template: Union[DataArray, Dataset] = infer_template(func, obj, *args, **kwargs)
    206     if isinstance(template, DataArray):
    207         result_is_array = True

/srv/conda/envs/notebook/lib/python3.7/site-packages/xarray/core/parallel.py in infer_template(func, obj, *args, **kwargs)
     80         raise Exception(
     81             "Cannot infer object returned from running user provided function."
---> 82         ) from e
     83 
     84     if not isinstance(template, (Dataset, DataArray)):

Exception: Cannot infer object returned from running user provided function.
```

#### Problem Description

We should try to imitate what dask does in `map_blocks`: https://docs.dask.org/en/latest/array-api.html#dask.array.map_blocks

Specifically:
- We should allow the user to override the checks by explicitly specifying output dtype and shape
- Maybe the check should be on small, rather than zero size, test data

<details>

INSTALLED VERSIONS
------------------
commit: None
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.14.138+
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.6.2

xarray: 0.14.0
pandas: 0.25.3
numpy: 1.17.3
scipy: 1.3.2
netCDF4: 1.5.1.2
pydap: installed
h5netcdf: 0.7.4
h5py: 2.10.0
Nio: None
zarr: 2.3.2
cftime: 1.0.4.2
nc_time_axis: 1.2.0
PseudoNetCDF: None
rasterio: 1.0.25
cfgrib: None
iris: 2.2.0
bottleneck: 1.3.0
dask: 2.7.0
distributed: 2.7.0
matplotlib: 3.1.2
cartopy: 0.17.0
seaborn: 0.9.0
numbagg: None
setuptools: 41.6.0.post20191101
pip: 19.3.1
conda: None
pytest: 5.3.1
IPython: 7.9.0
sphinx: None
​

</details>

dcherian:
- [x] Closes #3575
 - [x] Tests added

This PR adds a `template` kwarg to `map_blocks` so that we can do more complicated things where automated inference of the template fails. `template` is expected to be an xarray object that looks like the result of the `map_blocks` computation.

@jhamman To me, this seems a lot easier than defining a `dict` based schema. With dask variables, the memory cost shouldn't be high. It's easy to use standard xarray operations to make something that looks like the result dataset. Here's a notebook prototyping a `to_schema`/`from_schema` approach: https://gist.github.com/dcherian/130ba22d0fbadb616837deb914eaa67e#file-map_blocks_for_metsim_test-ipynb

```python
ds = xr.tutorial.load_dataset("air_temperature").chunk({"lat": 5})
template = ds.isel(lat=slice(1, None, 5))
result = xr.map_blocks(lambda x: x.isel(lat=[1]), ds) # fails
result = xr.map_blocks(lambda x: x.isel(lat=[1]), ds, template=template) # yay
xr.testing.assert_equal(result, template)  # passes
```
Todo:
- [x] add tests
- [x] what if the function returns a variable not in the template?
- [x] should we do anything with `template.attrs`?


xarray/core/parallel.py
```diff
@@ -379,10 +409,13 @@ def _wrapper(func, obj, to_array, args, kwargs):
         dims = template[name].dims
         var_chunks = []
         for dim in dims:
-            if dim in input_chunks:
-                var_chunks.append(input_chunks[dim])
+            if dim in output_chunks:
+                var_chunks.append(output_chunks[dim])
             elif dim in indexes:
                 var_chunks.append((len(indexes[dim]),))
+            elif dim in template.dims:
```
dcherian: this is an independent change. I can break this out in to a new PR.

dcherian: see #3817

xarray/tests/test_dask.py
```diff
@@ -1178,6 +1176,32 @@ def test_map_blocks_ds_transformations(func, map_ds):
...
+        xr.map_blocks(lambda x: x.isel(x=1), obj, template=obj).compute()
+    with raises_regex(ValueError, "Received dimension 'x' of length 1"):
+        xr.map_blocks(lambda x: x.isel(x=[1]), obj, template=obj).compute()
+
+
+def test_map_blocks_errors_bad_template_2(map_ds):
+    with raises_regex(ValueError, "unexpected data variables {'xyz'}"):
+        xr.map_blocks(lambda x: x.assign(xyz=1), map_ds, template=map_ds).compute()
+
+
```
jhamman: Can we also raise an error when a block with unexpected shapes (and no template) is returned?

xarray/core/parallel.py
```diff
@@ -107,6 +131,7 @@ def map_blocks(
     obj: Union[DataArray, Dataset],
     args: Sequence[Any] = (),
     kwargs: Mapping[str, Any] = None,
+    template: Union[DataArray, Dataset] = None,
```
jhamman: TODO: Add this to `DataArray.map_blocks()` and `Dataset.map_blocks()`

xarray/tests/test_dask.py
```diff
@@ -1039,7 +1039,7 @@ def test_map_blocks_error(map_da, map_ds):
     def bad_func(darray):
         return (darray * darray.x + 5 * darray.y)[:1, :1]
 
-    with raises_regex(ValueError, "Length of the.* has changed."):
+    with raises_regex(ValueError, "Received dimension.*"):
```
dcherian: @jhamman this test raises an error for unexpected shapes with no template present.

doc/dask.rst
```diff
@@ -382,6 +391,95 @@ application.
     structure of a problem, unlike the generic speedups offered by
     ``dask='parallelized'``.
 
+
```
dcherian: @aaronspring @bradyrx @andersy005 As likely downstream users of map_blocks, can you read through this doc file and see if it is clear and understandable?

doc/dask.rst
```diff
@@ -382,6 +391,95 @@ application.
...
+the returned result is identical to ``ds.time`` as expected.
+
+
+Here is a common example where automated inference will not work.
+
+.. ipython:: python
+    :okexcept:
+
+    def func(da):
+	print(da.sizes)
```
aaronspring: indent?

doc/dask.rst
```diff
@@ -382,6 +391,95 @@ application.
...
+        return da.isel(time=[1])
+
+    mapped = xr.map_blocks(func, ds.temperature)
+
+``func`` cannot be run on 0-shaped inputs because it is not possible to extract element 1 along a
+dimension of size 0. In this case we need to tell :py:func:`map_blocks` what the returned result looks
+like using the ``template`` kwarg. ``template`` must be an xarray Dataset or DataArray (depending on
+what the function returns) with dimensions, shapes, chunk sizes, coordinate variables *and* data
```
aaronspring: I imagine getting the correct coordinates in the template to be a bit tricky. does this need to be? in the example it is obvious and often there will be no coords. probably the users will get this.

doc/dask.rst
```diff
@@ -274,12 +274,21 @@ loaded into Dask or not:
...
+Automatic parallelization with ``apply_ufunc`` and ``map_blocks``
+-----------------------------------------------------------------
 
 Almost all of xarray's built-in operations work on Dask arrays. If you want to
-use a function that isn't wrapped by xarray, one option is to extract Dask
-arrays from xarray objects (``.data``) and use Dask directly.
+use a function that isn't wrapped by xarray, and have it applied in parallel on
+each block of your xarray object, you have three options:
+
+1. One option is to extract Dask arrays from xarray objects (``.data``) and use Dask directly.
```
jhamman:
```suggestion
1. Extract Dask arrays from xarray objects (``.data``) and use Dask directly.
```

doc/dask.rst
dcherian: seems to work OK.

doc/dask.rst
dcherian: yes it is unfortunately necessary. to be lazy, `map_blocks` needs to know what the output looks like.

dcherian:
Thanks for the review @aaronspring

xarray/core/parallel.py
```diff
@@ -248,8 +292,29 @@ def _wrapper(func, obj, to_array, args, kwargs):
...
+        preserved_indexes = template_indexes & dataset_indexes
+        new_indexes = template_indexes - dataset_indexes
+        indexes = {dim: dataset.indexes[dim] for dim in preserved_indexes}
+        indexes.update({k: template.indexes[k] for k in new_indexes})
+        output_chunks = {
+            dim: input_chunks[dim] for dim in template.dims if dim in input_chunks
+        }
+
+    else:
+        # template xarray object has been provided with proper sizes and chunk shapes
```
shoyer, line 309: Can we add a type check to verify that `template` is an xarray object? I can imagine users trying something like a tuple of DataArray objects and getting a strange error.

xarray/tests/test_dask.py
```diff
@@ -1180,6 +1180,36 @@ def test_map_blocks_ds_transformations(func, map_ds):
     assert_identical(actual, func(map_ds))
 
 
+@pytest.mark.parametrize("obj", [make_da(), make_ds()])
```
shoyer, line 1183: How about adding a cross test, where a Dataset is converted to a DataArray or the other way around?

xarray/core/parallel.py
```diff
@@ -201,20 +231,34 @@ def map_blocks(
...
                     raise ValueError(
-                        "Length of the %r dimension has changed. This is not allowed."
-                        % name
+                        f"Received dimension {name!r} of length {len(index)}. Expected length {check_shapes[name]}."
                     )
 
+        check_result_variables(result, expected, "coords")
+        if isinstance(result, Dataset):
+            check_result_variables(result, expected, "data_vars")
+
```
shoyer, line 261: I think the last thing that probably needs to be checked here are indexes on the returned object. We should verify those are as expected.

xarray/core/dataarray.py
```diff
@@ -3223,34 +3223,38 @@ def map_blocks(
...
             not allowed.
         kwargs: Mapping
             Passed verbatim to func after unpacking. xarray objects, if any, will not be
             split by chunks. Passing dask collections is not allowed.
+        template: (optional) DataArray, Dataset
+            xarray object representing the final result after compute is called. If not provided,
+            the function will be first run on mocked-up data, that looks like 'obj' but
+            has sizes 0, to determine properties of the returned object such as dtype,
+            variable names, new dimensions and new indexes (if any).
+            'template' must be provided if the function changes the size of existing dimensions.
```
shoyer, line 3257: I assume attrs are also copied from the template and ignored from the computed chunks?

shoyer:
Looks really nice! I like this design, just a few minor concerns.

doc/dask.rst
```diff
@@ -382,6 +391,95 @@ application.
...
+``func`` is received 0-sized blocks! :py:meth:`map_blocks` needs to know what the final result
+looks like in terms of dimensions, shapes etc. It does so by running the provided function on 0-shaped
+inputs (*automated inference*). This works in many cases, but not all. If automatic inference does not
+work for your function, provide the ``template`` kwarg (see below).
+
+In this case, automatic inference has worked so let's check that the result is as expected.
+
+.. ipython:: python
+
+    mapped.compute(scheduler="single-threaded")
```
TomAugspurger, line 429: Do you want to assign this to `mapped = ...`? Otherwise `mapped.identical` will compute it again?

dcherian:
Thanks for the review @TomAugspurger 


> Question on the name `template`. I think in dask.dataframe and dask.array we might call this `meta`. Is that keyword already used elsewhere in xarray? `template` is also a fine name though.

I added the `meta` kwarg to `apply_ufunc` so that users could pass that down to dask i.e. that meta = dask's meta = np.ndarray or something like that. So I'd like to avoid reusing `meta` here where it would exclusively be an xarray object ≠ dask's meta

 BUT it seems to me like there's a better name than `template`. Any ideas?

TomAugspurger:
Makes sense. template seems fine.

shoyer:
I think it's also a good idea to use a different name from dask's `meta` because `template` is used differently. It's an object with the full shape of the desired output, not size 0 dimensions.

xarray/core/dataarray.py
dcherian: Yes. But I don't know that this is right.

With auto-inferred templates, attrs are set by the user function. It= think it would be nice to preserve that behaviour. What do you think?

xarray/core/parallel.py
```diff
@@ -248,8 +310,38 @@ def _wrapper(func, obj, to_array, args, kwargs):
...
+        # template xarray object has been provided with proper sizes and chunk shapes
+        template_indexes = set(template.indexes)
+        indexes = {dim: dataset.indexes[dim] for dim in dataset_indexes}
+        indexes.update({k: template.indexes[k] for k in template_indexes})
+        if isinstance(template, DataArray):
+            output_chunks = dict(zip(template.dims, template.chunks))  # type: ignore
+        else:
+            output_chunks = template.chunks  # type: ignore
+
+    for dim in output_chunks:
```
dcherian, line 336: new error message

xarray/core/parallel.py
```diff
@@ -201,22 +238,47 @@ def map_blocks(
...
-            if name in obj.indexes:
-                if len(index) != len(obj.indexes[name]):
+            if name in check_shapes:
+                if len(index) != check_shapes[name]:
                     raise ValueError(
-                        "Length of the %r dimension has changed. This is not allowed."
-                        % name
+                        f"Received dimension {name!r} of length {len(index)}. Expected length {check_shapes[name]}."
+                    )
+            if name in expected["indexes"]:
```
dcherian, line 264: now checking indexes

xarray/core/parallel.py
```diff
@@ -102,14 +127,24 @@ def make_dict(x: Union[DataArray, Dataset]) -> Dict[Hashable, Any]:
     return {k: v.data for k, v in x.variables.items()}
 
 
+def _get_chunk_slicer(dim: Hashable, chunk_index: Mapping, chunk_bounds: Mapping):
```
dcherian, line 130: tiny refactor.

xarray/tests/test_dask.py
```diff
@@ -1180,6 +1180,64 @@ def test_map_blocks_ds_transformations(func, map_ds):
...
+    with raise_if_dask_computes():
+        actual = xr.map_blocks(func, obj, template=template)
+    assert_identical(actual, template)
+
+    with raise_if_dask_computes():
+        actual = obj.map_blocks(func, template=template)
+    assert_identical(actual, template)
+
+
+def test_map_blocks_template_convert_object():
```
dcherian, line 1196: tests array → dataset and vice versa

xarray/core/dataarray.py
dcherian: Thinking about this some more... there's no way to update `attrs` after computation is there?

shoyer: Correct, `attrs` needs to be set when the resulting Dataset is created, before we do the the computation.

bradyrx:
I missed this originally @dcherian, but thanks for the great work here. The docs changes are a great help.

dcherian:
Added docs on attrs issue. This should be good to go.

Thanks for the kind words, @bradyrx

dcherian:
Thanks for the reviews everyone.

PR #3800, opened 2020-02-26
Add note on diff's n differing from pandas

max-sixty:
This surprised me so I thought worth calling out. I need to add to Dataset's docstring too.

PR #3822, opened 2020-03-03
Add note on diff's n differing from pandas

max-sixty:
This surprised me so I thought worth calling out. 

Ref https://github.com/pydata/xarray/pull/3800

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -2694,8 +2694,8 @@ class DataArray(AbstractArray, DataWithCoords):
 
         .. note::
 
-            `n` matches numpy's behavior and is different from pandas' similarly
-            located `periods` argument
+            `n` matches numpy's behavior and is different from pandas' first
+            argument named `periods`.
 
 
         Examples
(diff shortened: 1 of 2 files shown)


dcherian:
I guess pandas now calls this `periods`? https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.diff.html

max-sixty:
Oh good spot @dcherian . I'll add that to the note

max-sixty:
Updated

xarray/core/dataarray.py
```diff
@@ -2692,6 +2692,12 @@ def diff(self, dim: Hashable, n: int = 1, label: Hashable = "upper") -> "DataArr
         difference : same type as caller
             The n-th order finite difference of this object.
 
+        .. note::
+
+            `n` matches numpy's behavior and is different from pandas' similarly
```
dcherian: I think "similarly located" is confusing.

How about 

```
The first argument `n` matches numpy's behavior and is different from pandas' first argument named `periods`.
```

max-sixty: Yes much better

dcherian:
Thanks @max-sixty

max-sixty:
Ended up being more from you than me @dcherian :)

Issue #3814, opened 2020-02-28
Allow an ellipsis in stack?


max-sixty:
Could we add the ability to use an ellipsis to represent all dims in more places? For example, `stack`:

#### MCVE Code Sample

```python                                                                                                                                            

In [14]: data = np.arange(15, 301, 15).reshape(2, 10)  
    ...: da = xr.DataArray(data, dims=('y', 'x'), attrs={'test': 'test'})  
    ...:                                                                                                                                                            

In [15]: da.stack(z=['x','y'])                                                                                                                                      
Out[15]: 
<xarray.DataArray (z: 20)>
array([ 15, 165,  30, 180,  45, 195,  60, 210,  75, 225,  90, 240, 105,
       255, 120, 270, 135, 285, 150, 300])
Coordinates:
  * z        (z) MultiIndex
  - x        (z) int64 0 0 1 1 2 2 3 3 4 4 5 5 6 6 7 7 8 8 9 9
  - y        (z) int64 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0 1
Attributes:
    test:     test

In [16]: da.stack(z=[...])                                                                                                                                          
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-16-a92d0ffe931a> in <module>
----> 1 da.stack(z=[...])

~/workspace/./.venv/lib64/python3.6/site-packages/xarray/core/dataarray.py in stack(self, dimensions, **dimensions_kwargs)
   1739         DataArray.unstack
   1740         """
-> 1741         ds = self._to_temp_dataset().stack(dimensions, **dimensions_kwargs)
   1742         return self._from_temp_dataset(ds)
   1743 

~/workspace/./.venv/lib64/python3.6/site-packages/xarray/core/dataset.py in stack(self, dimensions, **dimensions_kwargs)
   3291         result = self
   3292         for new_dim, dims in dimensions.items():
-> 3293             result = result._stack_once(dims, new_dim)
   3294         return result
   3295 

~/workspace/./.venv/lib64/python3.6/site-packages/xarray/core/dataset.py in _stack_once(self, dims, new_dim)
   3246 
   3247         # consider dropping levels that are unused?
-> 3248         levels = [self.get_index(dim) for dim in dims]
   3249         idx = utils.multiindex_from_product_levels(levels, names=dims)
   3250         variables[new_dim] = IndexVariable(new_dim, idx)

[...]

PR #3826, opened 2020-03-04
Allow ellipsis to be used in stack

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/3814
 - [x] Tests added

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -1705,8 +1705,8 @@ class DataArray(AbstractArray, DataWithCoords):
         ----------
         dimensions : Mapping of the form new_name=(dim1, dim2, ...)
             Names of new dimensions, and the existing dimensions that they
-            replace. Passing a list with an ellipsis (`[...]`) will stack over all
-            dimensions.
+            replace. Passing a list with an ellipsis (`[...]`) as existing
+            dimensions will stack over all dimensions.
         **dimensions_kwargs:
             The keyword arguments form of ``dimensions``.
             One of dimensions or dimensions_kwargs must be provided.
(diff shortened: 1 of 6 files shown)


max-sixty:
Should we support 
- `.stack(x=[...])`
- `.stack(x=...)`
- both

Currently implemented for both but leaning towards the first, partly because it makes the typing correct

max-sixty:
Changed to supporting the first (`.stack(x=[...])`) only

shoyer:
This is very cool!

It seems like it would be natural to support a combination of ellipsis and some explicit dimensions, like what can be done in transpose? E.g., `.stack(x=[‘y’, ...])` or `.stack(x=[..., ‘y’])`?

max-sixty:
Yes great!

max-sixty:
> It seems like it would be natural to support a combination of ellipsis and some explicit dimensions, like what can be done in transpose? E.g., `.stack(x=[‘y’, ...])` or `.stack(x=[..., ‘y’])`?

Yes great!

max-sixty:
Updated!

xarray/core/dataarray.py
```diff
@@ -1705,7 +1705,8 @@ def stack(
         ----------
         dimensions : Mapping of the form new_name=(dim1, dim2, ...)
             Names of new dimensions, and the existing dimensions that they
-            replace.
+            replace. Passing a list containing an ellipsis (`[...]`)
+            dimensions will stack over all dimensions.
```
dcherian: `Passing a partial list of dimensions with ellipsis is also supported ([..., dim2]).`

max-sixty: What do you think of this? I wanted to say what the ellipsis does. Or do you think unneeded / people can try it and will work it out.

dcherian: I think we should always keep the docstrings exhaustive since it makes it easier for people.

Also think both DataArray and Dataset should have the same two sentences:

 An ellipsis (`...`) will be replaced by all unlisted dimensions.Passing a list containing an ellipsis (`stacked_dim=[...]`)  will stack over all dimensions.

max-sixty: Very good!

dcherian:
Thanks @max-sixty

Issue #3770, opened 2020-02-15
`where` ignores incorrectly typed arguments


max-sixty:
#### MCVE Code Sample

I optimistically tried passing a lambda to `where`* to avoid another `.pipe`. To my surprise it ran! But unfortunately it just ignored the lambda, as though all values were True.

```python

In [1]: import xarray as xr                                                                                                                                         

In [2]: import numpy as np                                                                                                                                          

In [3]: da = xr.DataArray(np.random.rand(2,3))                                                                                                                      

In [4]: da.where(da > 0.5)                                                                                                                                          
Out[4]: 
<xarray.DataArray (dim_0: 2, dim_1: 3)>
array([[       nan, 0.71442406,        nan],
       [0.55748705,        nan,        nan]])
Dimensions without coordinates: dim_0, dim_1

# this should fail!

In [5]: da.where(lambda x: x > 0.5)                                                                                                                                 
Out[5]: 
<xarray.DataArray (dim_0: 2, dim_1: 3)>
array([[0.26085668, 0.71442406, 0.05816167],
       [0.55748705, 0.15293073, 0.12766453]])
Dimensions without coordinates: dim_0, dim_1
```

#### Expected Output

Raise a `TypeError` when passed the lambda

\* (maybe that could work, but separate discussion)

#### Output of ``xr.show_versions()``
<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.8 (default, Aug  7 2019, 17:28:10) 
[...]
python-bits: 64
OS: Linux
OS-release: [...]
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.utf8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: None

xarray: 0.14.1
pandas: 0.25.3
numpy: 1.18.1
scipy: 1.4.1
netCDF4: None
pydap: None
h5netcdf: None
h5py: 2.10.0
Nio: None
zarr: None
cftime: None
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: None
distributed: None
matplotlib: 3.1.2
cartopy: None
seaborn: 0.10.0
numbagg: None
setuptools: 45.0.0
pip: 20.0.2
conda: None
pytest: 5.3.2
IPython: 7.11.1
sphinx: 2.3.1

</details>

max-sixty:
[...]

Issue #3825, opened 2020-03-04
Accept lambda in methods with one obvious xarray argument


max-sixty:
Branching from https://github.com/pydata/xarray/issues/3770

Here's the proposal: allow lambdas on methods where the primary argument is a single xarray object, and interpret lambas as though they'd be supplied in a `pipe` method followed by the current method. 
Taking the example from the linked issue:

```python

In [1]: import xarray as xr                                                                                                                                         

In [2]: import numpy as np                                                                                                                                          

In [3]: da = xr.DataArray(np.random.rand(2,3))                                                                                                                      

In [4]: da.where(da > 0.5)                                                                                                                                          
Out[4]: 
<xarray.DataArray (dim_0: 2, dim_1: 3)>
array([[       nan, 0.71442406,        nan],
       [0.55748705,        nan,        nan]])
Dimensions without coordinates: dim_0, dim_1

# this should be equivalent (currently not valid)

In [5]: da.where(lambda x: x > 0.5)

# the longer version (currently works)

In [5]: da.pipe(lambda x: x.where(x > 0.5))
                                                                                                                                 
```

Others I miss from pandas: `assign`, and `loc`.
I haven't gone through the list though assume there are others; we don't have to agree 100% on the list before starting with the most obvious ones, assuming we're in agreement with the principle.

max-sixty:
`assign` is already supported! 🤦‍♂

max-sixty:
`where` is complete

I think the others that could be worthwhile are the selection methods:
- `sel` & `isel`
- `loc`
- `reindex`?
...but much of their use case is bool arrays, which is mostly covered by `where(..., drop=True)`. 

Interested in what others think. I'm in no rush to implement those; `where` seemed much clearer to me.

max-sixty:
Closing; can do others with the benefit of more experience

PR #3827, opened 2020-03-04
Allow `where` to receive a callable

max-sixty:
- [x] Towards https://github.com/pydata/xarray/issues/3825, closes https://github.com/pydata/xarray/issues/3770
 - [x] Tests added

The first one; I'll await feedback before adding to docs or doing other methods

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -1153,7 +1153,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         from .dataset import Dataset
 
         if callable(cond):
-            return self.where(cond(self), other=other, drop=drop)
+            cond = cond(self)
 
         if drop:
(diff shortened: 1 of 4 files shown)


xarray/core/common.py
```diff
@@ -1152,6 +1152,9 @@ def where(self, cond, other=dtypes.NA, drop: bool = False):
         from .dataarray import DataArray
         from .dataset import Dataset
 
+        if callable(cond):
+            return self.where(cond(self), other=other, drop=drop)
```
keewis: this is not really a shortcut so the `return` might be misleading (we're calling the method again with different arguments). What about using
```suggestion
            cond = cond(self)
```
instead?

max-sixty: Great, that's much cleaner

PR #3836, opened 2020-03-06
Improve where docstring

max-sixty:

The `x, y ` doesn't carry over to the online docs: http://xarray.pydata.org/en/stable/generated/xarray.where.html#xarray.where, lmk if there's a hack for representing that in a better way

### DIFF
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -1225,14 +1225,18 @@ def where(cond, x, y):
     ----------
     cond : scalar, array, Variable, DataArray or Dataset with boolean dtype
         When True, return values from `x`, otherwise returns values from `y`.
-    x, y : scalar, array, Variable, DataArray or Dataset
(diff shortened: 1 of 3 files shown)


xarray/core/computation.py
```diff
@@ -1225,14 +1225,18 @@ def where(cond, x, y):
...
+        values to choose from where `cond` is False
+
+    All dimension coordinates on these objects must be aligned with each
+    other and with `cond`.
 
     Returns
     -------
-    In priority order: Dataset, DataArray, Variable or array, whichever
-    type appears as an input argument.
+    Dataset, DataArray, Variable or array, whichever type appears as an input
```
dcherian: don't we need the "in priority order" bit?

max-sixty: Yes, you're right. It was confusing to me at first, but in lieu of better wording, I'll add it back

max-sixty:
Ready

PR #3847, opened 2020-03-08
provide a error summary for assert_allclose

### LINKED ISSUE #3617: diff formatting for assert_allclose
At the moment, the output of a failing `assert_allclose` is not really useful, it simply prints the values it compared:
```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: from xarray.testing import assert_allclose, assert_identical

In [2]: da1 = xr.DataArray(data=[2, 5, 8], coords={"x": [0, 1, 2]}, dims="x")
   ...: da2 = xr.DataArray(data=[1, 5, 8], coords={"x": [0, 1, 3]}, dims="x")

In [3]: assert_allclose(da1, da2)
<ipython-input-6-55cbcc1eeb58> in <module>
----> 1 assert_allclose(da1, da2)

~/Documents/Programming/xarray/xarray/testing.py in assert_allclose(a, b, rtol, atol, decode_bytes)
    123         assert allclose, f"{a.values}\n{b.values}"
    124     elif isinstance(a, DataArray):
--> 125         assert_allclose(a.variable, b.variable, **kwargs)
    126         assert set(a.coords) == set(b.coords)
    127         for v in a.coords.variables:

~/Documents/Programming/xarray/xarray/testing.py in assert_allclose(a, b, rtol, atol, decode_bytes)
    121         assert a.dims == b.dims
    122         allclose = _data_allclose_or_equiv(a.values, b.values, **kwargs)
--> 123         assert allclose, f"{a.values}\n{b.values}"
    124     elif isinstance(a, DataArray):
    125         assert_allclose(a.variable, b.variable, **kwargs)

AssertionError: [2 5 8]
[1 5 8]

In [4]: assert_identical(da1, da2)
<ipython-input-7-127e2b1ca0dc> in <module>
----> 1 assert_identical(da1, da2)

~/Documents/Programming/xarray/xarray/testing.py in assert_identical(a, b)
     83     elif isinstance(a, DataArray):
     84         assert a.name == b.name
---> 85         assert a.identical(b), formatting.diff_array_repr(a, b, "identical")
     86     elif isinstance(a, (Dataset, Variable)):
     87         assert a.identical(b), formatting.diff_dataset_repr(a, b, "identical")

AssertionError: Left and right DataArray objects are not identical

Differing values:
L
    array([2, 5, 8])
R
    array([1, 5, 8])
Differing coordinates:
L * x        (x) int64 0 1 2
R * x        (x) int64 0 1 3
```
It would be good to somehow extend `formatting.diff_array_repr` and `formatting.diff_dataset_repr` and to then use those to print a diff for `assert_allclose`.

I found a reference to this: https://github.com/pydata/xarray/pull/1690#issuecomment-455540813, but it's not being actively worked on.

keewis:
right now, unlike `assert_identical` and `assert_equals`, `assert_allclose` simply prints the mismatching data without providing a summary about what failed. This tries to fix that by allowing the `compat` parameter to `formatting.diff_array_repr` and `formatting.diff_dataset_repr` to be a callable and by rewriting `assert_allclose` to only have one assert statement per type.

 - [x] Closes #3617, closes #1690?
 - [x] Tests added

There are still more tests to write but I would like to get early feedback about this approach vs. adding a `"allclose"` compat type in addition to the existing `"equals"` and `"identical"` to the formatting functions.


xarray/tests/test_testing.py
```diff
@@ -1,7 +1,31 @@
...
+        ),
+        pytest.param(
+            xr.Dataset({"a": ("x", [1e-17, 2]), "b": ("y", [-2e-18, 2])}),
+            xr.Dataset({"a": ("x", [0, 2]), "b": ("y", [0, 1])}),
+            id="Dataset",
+        ),
+    ),
+)
+def test_assert_allclose(obj1, obj2):
+    xr.testing.assert_allclose(obj1, obj2)
```
max-sixty: Do we need to test that it correctly fails?

max-sixty:
Looks great!

keewis:
the failures should be a bug in `pint` and hopefully fixed by hgrecco/pint#1044

dcherian:
Let's mark those as xfail then?

keewis:
now only the `min-all-deps` and `min-nep18` CI fail. I suspect that is due to the pinned dependencies.

keewis:
to me the failing tests on the `min-all-deps` and `min-nep18` CI seem like a `dask` issue (in `var`?) that has been resolved in later versions: if we compute `actual` before passing it to `assert_allclose` the tests don't fail. How should I resolve this?

dcherian:
Our minimum versions policy allows us to bump dask. So let's do that?
```
dask          2.2     (2019-08-01) 2.5     (2019-09-27) <
distributed   2.2     (2019-08-01) 2.5     (2019-09-27) <
```

> Also, where should I put the whats-new.rst entry?

New features? It is public API and is a very nice enhancement.

keewis:
unfortunately, we need `dask>=2.9.1` for that. I could try debugging a bit more to find out exactly why it fails (or someone helps me with that?), but that would take me a bit more time than just skipping / xfailing those tests if on `dask<2.9.1`.

keewis:
The issue is with how `dask<2.9.1` handles dtypes on compute (in `nanvar`?) when the data is an array with dtype `object` filled with `False` and some missing values represented by `np.nan`.

I really lack experience with `dask`, though, so I'm clueless as to what to do to fix that (besides calling `compute` before passing to `allclose_or_equiv`) and would appreciate help with this.

keewis:
@pydata/xarray, this is really close but I'm not familiar enough with `dask` to get it to work with `dask<2.9.1`.

Once we get the `py36-min-all-deps` and `py36-min-nep18` CI to pass, this should be ready for a final review and merging.

dcherian:
I'm confused here. What did you change to trigger this error?

We could call `arr1.compute(), arr2.compute()` in 
https://github.com/pydata/xarray/blob/69548df9826cde9df6cbdae9c033c9fb1e62d493/xarray/core/duck_array_ops.py#L203-L204

At that point, we need to compare actual values. I'm not sure that dask exits `all()` early if one chunk evaluates to False.

dcherian:
example failure:
```
___________________ test_reduce[None-True-var-True-bool_-1] ____________________

dim_num = 1, dtype = <class 'numpy.bool_'>, dask = True, func = 'var'
skipna = True, aggdim = None

    @pytest.mark.parametrize("dim_num", [1, 2])
    @pytest.mark.parametrize("dtype", [float, int, np.float32, np.bool_])
    @pytest.mark.parametrize("func", ["sum", "min", "max", "mean", "var"])
    # TODO test cumsum, cumprod
    @pytest.mark.parametrize("skipna", [False, True])
    @pytest.mark.parametrize("aggdim", [None, "x"])
    def test_reduce(dim_num, dtype, dask, func, skipna, aggdim):
    
        if aggdim == "y" and dim_num < 2:
            pytest.skip("dim not in this test")
    
        if dtype == np.bool_ and func == "mean":
            pytest.skip("numpy does not support this")
    
        if dask and not has_dask:
            pytest.skip("requires dask")
    
        if dask and skipna is False and dtype in [np.bool_]:
            pytest.skip("dask does not compute object-typed array")
    
        rtol = 1e-04 if dtype == np.float32 else 1e-05
    
        da = construct_dataarray(dim_num, dtype, contains_nan=True, dask=dask)
        axis = None if aggdim is None else da.get_axis_num(aggdim)
    
        # TODO: remove these after resolving
        # https://github.com/dask/dask/issues/3245
            warnings.filterwarnings("ignore", "Mean of empty slice")
            warnings.filterwarnings("ignore", "All-NaN slice")
            warnings.filterwarnings("ignore", "invalid value encountered in")
    
            if da.dtype.kind == "O" and skipna:
                # Numpy < 1.13 does not handle object-type array.
                try:
                    if skipna:
                        expected = getattr(np, f"nan{func}")(da.values, axis=axis)
                    else:
                        expected = getattr(np, func)(da.values, axis=axis)
    
                    actual = getattr(da, func)(skipna=skipna, dim=aggdim)
                    assert_dask_array(actual, dask)
                    assert np.allclose(
                        actual.values, np.array(expected), rtol=1.0e-4, equal_nan=True
                    )
                except (TypeError, AttributeError, ZeroDivisionError):
                    # TODO currently, numpy does not support some methods such as
                    "casting='same_kind'"
>                   % (funcname(function), str(dtype), str(result.dtype))
                )
E               ValueError: Inferred dtype from function 'sub' was 'int64' but got 'float64', which can't be cast using casting='same_kind'
```

shoyer:
It looks like there might be some sort of issue with dask's numpy compatibility layer? e.g., in np.allclose?

These fallbacks are catching quite a few errors....

One idea would be to try plumbing this logic into `np.testing.assert_allclose` (which doesn't do dispatching) rather than `np.allclose`. The former comes pre-packaged with better error messages, which might be handy.

dcherian:
this comment: `# Numpy < 1.13 does not handle object-type array.`  suggests that this code needs to be updated for our new `numpy>=1.15` requirement

keewis:
Unfortunately, simply replacing `assert np.allclose(...)` with `np.testing.assert_allclose(...)` does not fix this. Also, I undid the swapping of arguments to `allclose_or_equiv` which makes the `min-all-deps` CI fail, too. Does anyone know why that works?

Would it make sense to wait until we can bump the dask version to `2.9`?

dcherian:
> Would it make sense to wait until we can bump the dask version to 2.9?

Sounds good to me. It's only 1.5 months away

keewis:
I needed to use `assert_allclose` in the `pint` tests in #3975, so I modified it to use `.data` instead of `.values` and now the exact same tests are failing. So it seems that `assert_allclose` never worked with `dask` but always converted to `numpy`. Would it make sense to investigate further, or is it better to just wait until we can merge this?

dcherian:
How about we call `compute` in `assert_allclose` for boolean dask arrays when `dask < 2.9.1`? This bit can then be removed in a couple of months.

This compute was happening in previous versions anyway (because `assert_allclose` was using `.values`) so this  would not be a regression.

keewis:
let's see if that works

xarray/core/duck_array_ops.py
```diff
@@ -199,6 +200,17 @@ def allclose_or_equiv(arr1, arr2, rtol=1e-5, atol=1e-8):
     """
     arr1 = asarray(arr1)
     arr2 = asarray(arr2)
+    # TODO: remove after we require dask > 2.9.1
```
dcherian, line 203: let's wait to check the tests but this bit should go under `if lazy_equiv is None:`

keewis: true. There was also a mistake in the way I checked the version

shoyer:
> So it seems that `assert_allclose` never worked with `dask` but always converted to `numpy`. Would it make sense to investigate further, or is it better to just wait until we can merge this?

This was intentional, I think. `np.testing.assert_allclose()` doesn't support dispatching -- it always converts into NumPy arrays. I'm not sure that `assert_allclose` would  even be well defined on dask arrays otherwise, because it doesn't have any output (it only raises an error).

keewis:
hmm... using `arr.compute()` fails while `np.array(arr)` works. For now, I'm converting all dask arrays to numpy if the dask version is not high enough (I can't do that only for `bool_` arrays since they have been converted to float somewhere).

@shoyer: I'm confused, I thought that `xr.testing.assert_allclose` explicitly supported duck arrays (it calls `duck_array_ops.allclose_or_equiv`). TBC, what I was talking about was these lines:
https://github.com/pydata/xarray/blob/e5cc19cd8f8a69e0743f230f5bf51b7778a0ff96/xarray/testing.py#L124
https://github.com/pydata/xarray/blob/e5cc19cd8f8a69e0743f230f5bf51b7778a0ff96/xarray/testing.py#L132-L135
where the `.values` always convert to `numpy`

shoyer:
> I'm confused, I thought that `xr.testing.assert_allclose` explicitly supported duck arrays (it calls `duck_array_ops.allclose_or_equiv`).

Right, it explicitly supports duck arrays -- by always converting them into NumPy!

keewis:
then it seems that this PR removes this intentional restriction. I'm not sure it is still needed, though: we don't use `np.testing.assert_allclose` internally because we want to provide our own error messages, and the actual comparison is done with https://github.com/pydata/xarray/blob/e5cc19cd8f8a69e0743f230f5bf51b7778a0ff96/xarray/core/duck_array_ops.py#L204
which should allow dispatching

keewis:
anyway, with this the tests finally pass :tada: so this should be ready for review and merging.

keewis:
gentle ping, @shoyer

xarray/tests/test_units.py
```diff
@@ -1413,6 +1428,13 @@ def example_1d_objects(self):
         ]:
             yield (self.cls("x", data), data)
 
+    # TODO: remove once pint==0.12 has been released
+    @pytest.mark.xfail(
```
dcherian, line 1432: Looks like pint 0.12 is out. Shall we remove these xfails now or do you want to do that later?

keewis: I think something went wrong with the release of `0.12` so I'd like to wait until I got a response from the maintainer

dcherian: :+1: Sounds good. IMO we should merge this.

keewis:
unless there are any objections (@shoyer?) I'm going to merge this tomorrow.

mathause:
Do you need to add a note to what's new that the dask version is bumped?

keewis:
I think we could bump the version even further (at least to `2.8`, maybe also `2.9`) so I think that should be a different PR

PR #3854, opened 2020-03-10
Update doc env pinned versions

### LINKED ISSUE #3796: RTD failing yet again

dcherian:
Attempt to fix RTD build by updating pinned version
 - [ ] Closes #3796

### DIFF
--- a/ci/requirements/doc.yml
+++ b/ci/requirements/doc.yml
@@ -4,23 +4,23 @@ channels:
   - conda-forge
 dependencies:
   - python=3.8
-  - bottleneck
-  - cartopy
+  - bottleneck>=1.3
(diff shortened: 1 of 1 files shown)


keewis:
In the issue tracker of RTD they state that they are testing an upgrade to a bigger build server, which potentially could resolve the `conda` / `conda-forge` memory issues.

Should we have this close #3797? I think the `conda` / `pip` hybrid environment consumed even more memory.

fmaussion:
If you want to test builds on RTD before merging you can use the fix-docs branch for that and check its status here: https://readthedocs.org/projects/xray/builds/

PR #3857, opened 2020-03-11
Blacken the doctest code in docstrings

keewis:
With a bit of preparation this is the result of running [black-doctest](https://github.com/keewis/black-doctest) on the repository. The tool itself is experimental but I didn't find any issues after manually reviewing the changes. The exact call was:
```bash
python -m blackdoc .
```
I'm not too sure what line length we use for docstrings (I used the default of `88` for this run), but if we'd like a different number we can just rerun the tool.

`python -m pytest --doctest-modules xarray/core` still fails but from what I saw these failures are not due to changes in this PR.


xarray/coding/cftimeindex.py
```diff
@@ -268,29 +268,32 @@ def _partial_date_slice(self, resolution, parsed):
...
         <xarray.DataArray (time: 1)>
         array([1])
         Coordinates:
           * time     (time) object 2001-01-01 00:00:00
-        >>> da = xr.DataArray([1, 2],
-                              coords=[[pd.Timestamp(2001, 1, 1),
-                                       pd.Timestamp(2001, 2, 1)]],
-                              dims=['time'])
-        >>> da.sel(time='2001-01-01')
+        >>> da = xr.DataArray(
```
max-sixty: I had thought the standard was a blank line after each result; is that a general standard or was I holding that incorrectly?
(no need to do it this round though)

max-sixty:
Looks great! Congrats on such a good tool!

xarray/coding/cftimeindex.py
keewis: I had the same impression, but the [docs](https://docs.python.org/3/library/doctest.html#how-are-docstring-examples-recognized) say otherwise:

> Expected output cannot contain an all-whitespace line, since such a line is taken to signal the end of expected output. If expected output does contain a blank line, put `<BLANKLINE>` in your doctest example each place a blank line is expected.

I think that means that we only need the newline if we want to continue with normal text.

xarray/coding/cftimeindex.py
```diff
@@ -268,29 +268,32 @@ def _partial_date_slice(self, resolution, parsed):
...
         >>> import pandas as pd
         >>> import xarray as xr
-        >>> da = xr.DataArray([1, 2],
-                              coords=[[DatetimeNoLeap(2001, 1, 1),
-                                       DatetimeNoLeap(2001, 2, 1)]],
-                              dims=['time'])
-        >>> da.sel(time='2001-01-01')
+        >>> da = xr.DataArray(
+        ...     [1, 2],
+        ...     coords=[[DatetimeNoLeap(2001, 1, 1), DatetimeNoLeap(2001, 2, 1),]],
```
keewis: the trailing comma (`,]],`) at the end of the line happens with `black`, too: calling `black` on a file containing
```python
da = xr.DataArray(
    [1, 2],
    coords=[[
        DatetimeNoLeap(2001, 1, 1),
        DatetimeNoLeap(2001, 2, 1),
    ]],
    dims=["time"],
)
```
gives us
```python
da = xr.DataArray(
    [1, 2],
    coords=[[DatetimeNoLeap(2001, 1, 1), DatetimeNoLeap(2001, 2, 1),]],
    dims=["time"],
)
```

while having
```python
da = xr.DataArray(
    [1, 2],
    coords=[[
        DatetimeNoLeap(2001, 1, 1),
        DatetimeNoLeap(2001, 2, 1)
    ]],
    dims=["time"],
)
```
correctly returns
```python
da = xr.DataArray(
    [1, 2],
    coords=[[DatetimeNoLeap(2001, 1, 1), DatetimeNoLeap(2001, 2, 1)]],
    dims=["time"],
)
```


I'm not sure why, though.

xarray/coding/cftimeindex.py
max-sixty: OK. IIUC we _can_ still have blank lines, but it becomes a preference rather than enforced. 
FWIW I do find they make it much easier to read the examples!

keewis: sounds good. Let's do that in a new PR, though.

dcherian:
Thanks @keewis

PR #3862, opened 2020-03-17
Raise error when assigning to IndexVariable.values & IndexVariable.data

### LINKED ISSUE #3470: warn when updating coord.values : indexes are not updated
Hi, I've recently come across what it seems to be a bit annoying bug. After adding an offset to the time dimension using pd.add_offset(), the wrong times are selected when using .sel(). I've come across this behavior in recent versions of xarray, starting from 0.12. I've now worked around it by using ds.assign_coords(time = ...). I'm not sure however if this is a possible bug in Xarray or Pandas. Below a minimum working example.

#### MCVE Code Sample

```python
import xarray as xr
import pandas as pd
import numpy as np

ds = xr.Dataset({'temperature': (['x', 'y', 'time'],  temp)},
                         'lat': (['x', 'y'], lat),
                         'time': pd.date_range('2014-09-01', periods=3,freq='MS'),
                         })

seltime = ds.time[2]
ds.time.values = pd.to_datetime(ds.time.values) + pd.DateOffset(months=2)
```
```python
print(ds.sel(time=seltime).time)
```
```python
<xarray.DataArray 'time' ()>
array('2015-01-01T00:00:00.000000000', dtype='datetime64[ns]')
Coordinates:
    time     datetime64[ns] 2015-01-01
```
```python
print(seltime)
```
```python
<xarray.DataArray 'time' ()>
array('2014-11-01T00:00:00.000000000', dtype='datetime64[ns]')
Coordinates:
    time     datetime64[ns] 2014-11-01
```

#### Expected Output
Expected output is that seltime and ds.sel(time=seltime).time are the same

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.3 (default, Apr  3 2019, 19:16:38) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
python-bits: 64
OS: Linux
OS-release: 4.15.0-58-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.0
libnetcdf: 4.6.0

xarray: 0.14.0
pandas: 0.25.2
numpy: 1.17.3
scipy: 1.3.1
netCDF4: 1.4.0
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: 1.1.0
PseudoNetCDF: None
rasterio: 1.0.26
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 2.3.0
distributed: None
matplotlib: 3.1.1
cartopy: 0.17.0
seaborn: 0.9.0
numbagg: None
setuptools: 41.2.0
pip: 9.0.1
conda: None
pytest: None
IPython: 7.7.0
sphinx: None
</details>

dcherian:
- [x] Closes #3470
 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,7 +24,8 @@ Breaking changes
 
 - Raise an error when assigning to the ``.values`` or ``.data`` attribute of
   dimension coordinates i.e. ``IndexVariable`` objects. This has been broken since
-  v0.12.0. (:issue:`3470`, :pull:`3862`)
+  v0.12.0. Please use :py:meth:`DataArray.assign_coords` or :py:meth:`Dataset.assign_coords`
+  instead. (:issue:`3470`, :pull:`3862`)
   By `Deepak Cherian <https://github.com/dcherian>`_
 
 New Features
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -22,6 +22,11 @@ v0.15.1 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- Raise an error when assigning to the ``.values`` or ``.data`` attribute of
+  dimension coordinates i.e. ``IndexVariable`` objects. This has been broken since
+  v0.13.0. (:issue:`3470`)
+  By `Deepak Cherian <https://github.com/dcherian>`_
+
 New Features
 ~~~~~~~~~~~~
 
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,7 +24,7 @@ Breaking changes
 
 - Raise an error when assigning to the ``.values`` or ``.data`` attribute of
   dimension coordinates i.e. ``IndexVariable`` objects. This has been broken since
-  v0.13.0. (:issue:`3470`)
+  v0.12.0. (:issue:`3470`, :pull:`3862`)
   By `Deepak Cherian <https://github.com/dcherian>`_
 
 New Features
--- a/doc/plotting.rst
+++ b/doc/plotting.rst
@@ -657,7 +657,7 @@ Additionally, the boolean kwarg ``add_guide`` can be used to prevent the display
 
 .. ipython:: python
 
-    ds.w.values = [1, 2, 3, 5]
+    ds = ds.assign(w=[1, 2, 3, 5])
     @savefig ds_discrete_legend_hue_scatter.png
     ds.plot.scatter(x='A', y='B', hue='w', hue_style='discrete')
 
--- a/xarray/core/variable.py
+++ b/xarray/core/variable.py
@@ -2061,9 +2061,17 @@ class IndexVariable(Variable):
     # https://github.com/python/mypy/issues/1465
     @Variable.data.setter  # type: ignore
     def data(self, data):
-        Variable.data.fset(self, data)
-        if not isinstance(self._data, PandasIndexAdapter):
-            self._data = PandasIndexAdapter(self._data)
+        raise ValueError(
+            f"Cannot assign to the .data attribute of dimension coordinate a.k.a IndexVariable {self.name!r}. "
+            f"Please use DataArray.assign_coords, Dataset.assign_coords or Dataset.assign as appropriate."
+        )
+
+    @Variable.values.setter  # type: ignore
+    def values(self, values):
+        raise ValueError(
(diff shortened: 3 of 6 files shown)


max-sixty:
Great, good spot @dcherian

max-sixty:
OK to merge @dcherian ? We can almost release 0.15.1 then

dcherian:
I think @shoyer should confirm that this is what we want to do

doc/whats-new.rst
```diff
@@ -22,6 +22,11 @@ v0.15.1 (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- Raise an error when assigning to the ``.values`` or ``.data`` attribute of
+  dimension coordinates i.e. ``IndexVariable`` objects. This has been broken since
+  v0.13.0. (:issue:`3470`)
```
```suggestion
  v0.12.0. (:issue:`3470`, :pull:`3862`)
```

keewis:
this causes the docs to fail: https://github.com/pydata/xarray/blob/0722b424b547a362b5f4678483d0a42812d611ee/doc/plotting.rst#L660

dcherian:
I think the failed tests are unrelated since they only fail on the `upstream-dev`

```
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[365_day]
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[360_day]
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[julian]
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[all_leap]
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[366_day]
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[gregorian]
FAILED xarray/tests/test_cftimeindex.py::test_indexing_in_series_getitem[proleptic_gregorian]
```

spencerkclark:
I think now that you merged #3874, if you re-merge master they should pass.

dcherian:
Great. Thanks all.

Now to find out how many downstream projects test against master on CI :)

rsokl:
@dcherian @max-sixty is there a reason why this change wasn't incorporated into `values.setter`? This change blocks the ability to update coordinate values in-place. This breaks code in a more fundamental way than just requiring users to rely on `assign_coords` - it can necessitate a substantial change in code design for people who were relying on the ability to mutate coordinates. This also leads me to ask: was it really appropriate to call such a substantial breaking change only a patch?

shoyer:
> is there a reason why this change wasn't incorporated into `values.setter`?

I'm not quite sure what you mean here. That's exactly what this change does, making assignment to `IndexVariable.values` an error.

If you were mutating `.values` on an indexed variable, then you already had a likely bug, because only part of xarray's data structures were being updated (the Variable, but not indexes). That's why we put this in as a bug fix -- see the linked issue https://github.com/pydata/xarray/issues/3470 for examples.

Unfortunately we couldn't figure out another way to make this work -- see https://github.com/pydata/xarray/issues/3470#issuecomment-566214653 for discussion. We're definitely open to alternatives if you can come up with them, though!

rsokl:
> That's exactly what this change does, making assignment to IndexVariable.values an error.

Sorry, I got a bit criss-crossed between this PR and the thread in #3470. 

In light of https://github.com/pydata/xarray/issues/3470#issuecomment-566214653 it seems like the next best thing would be  to given `assign_coords`  the optional ability to update the `DataArray`/`DataSet` in-place. Thoughts?

shoyer:
> In light of [#3470 (comment)](https://github.com/pydata/xarray/issues/3470#issuecomment-566214653) it seems like the next best thing would be to given `assign_coords` the optional ability to update the `DataArray`/`DataSet` in-place. Thoughts?

There's no need to use `assign_coords`. You can also just use normal assignment to a DataArray/Dataset or `coords`:
```
ds['time'] = new_values  # good
ds.coords['time'] = new_values  # good
ds.coords['time'].values = new_values  # broken, disabled by this PR
```

This would probably be worth clarifying in the error message.

(Under the hood, `assign_coords` basically does the exactly same thing as assignment to `coords`, it just copies the `Dataset`/DataArray` first)

rsokl:
Ah! Nice! Yeah the error message led me to believe that there was no path for in-place updates.

PR #3871, opened 2020-03-20
Implement idxmax and idxmin functions

### LINKED ISSUE #60: Implement DataArray.idxmax()
Should match the pandas function: http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.idxmax.html

toddrjen:
This implements `idxmax` and `idxmin` functions similar to thier pandas equivalents.

This is my first time contributing to the project so I am not certain the structure or approach is the best.  Please let me know if there is a better way to implement this.

This also includes two other changes.

First, it drops some code for backwards-compatibility with numpy 1.12, which isn't supported.  This code was hiding an error I needed to have access to in order to get the function working.

Second, it adds an option to `Dataset.map` to let you map `DataArray` methods by name.  I used this to implement the `Dataset` versions of `idxmax` and `idxmin`.

 - [X] Closes #60
 - [X] Tests added
 - [X] Passes `isort -rc . && black . && mypy . && flake8`

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -38,6 +38,9 @@ New Features
 - Limited the length of array items with long string reprs to a
   reasonable width (:pull:`3900`)
   By `Maximilian Roos <https://github.com/max-sixty>`_
+- Implement :py:meth:`DataArray.idxmax`, :py:meth:`DataArray.idxmin`,
+  :py:meth:`Dataset.idxmax`, :py:meth:`Dataset.idxmin`.  (:issue:`60`, :pull:`3871`)
+  By `Todd Jennings <https://github.com/toddrjen>`_
 
 
 Bug fixes
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -42,6 +42,9 @@ New Features
 
 Bug fixes
 ~~~~~~~~~
+- Fix a regression where deleting a coordinate from a copied :py:class:`DataArray`
+  can affect the original :py:class:`Dataarray`.  (:issue:`3899`, :pull:`3871`)
+  By `Todd Jennings <https://github.com/toddrjen>`_
 
 
 Documentation
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -23,9 +23,10 @@ from typing import (
 
 import numpy as np
 
-from . import duck_array_ops, utils
+from . import dtypes, duck_array_ops, utils
 from .alignment import deep_align
 from .merge import merge_coordinates_without_align
+from .nanops import dask_array
 from .options import OPTIONS
 from .pycompat import dask_array_type
 from .utils import is_dict_like
@@ -1338,3 +1339,66 @@ def polyval(coord, coeffs, degree_dim="degree"):
         coords={coord.name: coord, degree_dim: np.arange(deg_coord.max() + 1)[::-1]},
     )
     return (lhs * coeffs).sum(degree_dim)
+
+
+def _calc_idxminmax(
+    *,
+    array,
+    func: Callable,
+    dim: Hashable = None,
+    skipna: bool = None,
+    fill_value: Any = dtypes.NA,
+    keep_attrs: bool = None,
+):
+    """Apply common operations for idxmin and idxmax."""
+    # This function doesn't make sense for scalars so don't try
+    if not array.ndim:
+        raise ValueError("This function does not apply for scalars")
+
+    if dim is not None:
+        pass  # Use the dim if available
+    elif array.ndim == 1:
+        # it is okay to guess the dim if there is only 1
+        dim = array.dims[0]
+    else:
+        # The dim is not specified and ambiguous.  Don't guess.
+        raise ValueError("Must supply 'dim' argument for multidimensional arrays")
+
+    if dim not in array.dims:
+        raise KeyError(f'Dimension "{dim}" not in dimension')
+    if dim not in array.coords:
+        raise KeyError(f'Dimension "{dim}" does not have coordinates')
+
+    # These are dtypes with NaN values argmin and argmax can handle
+    na_dtypes = "cfO"
+
+    if skipna or (skipna is None and array.dtype.kind in na_dtypes):
+        # Need to skip NaN values since argmin and argmax can't handle them
+        allna = array.isnull().all(dim)
+        array = array.where(~allna, 0)
+
+    # This will run argmin or argmax.
+    indx = func(array, dim=dim, axis=None, keep_attrs=keep_attrs, skipna=skipna)
+
+    # Get the coordinate we want.
+    coordarray = array[dim]
+
+    # Handle dask arrays.
+    if isinstance(array, dask_array_type):
+        res = dask_array.map_blocks(coordarray, indx, dtype=indx.dtype)
+    else:
+        res = coordarray[
+            indx,
+        ]
+
+    if skipna or (skipna is None and array.dtype.kind in na_dtypes):
+        # Put the NaN values back in after removing them
+        res = res.where(~allna, fill_value)
+
+    # The dim is gone but we need to remove the corresponding coordinate.
+    del res.coords[dim]
+
+    # Copy attributes from argmin/argmax, if any
+    res.attrs = indx.attrs
+
+    return res
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -3508,6 +3508,194 @@ class DataArray(AbstractArray, DataWithCoords):
         )
         return self._from_temp_dataset(ds)
 
+    def idxmin(
+        self,
+        dim: Hashable = None,
+        skipna: bool = None,
+        fill_value: Any = dtypes.NA,
+        keep_attrs: bool = None,
+    ) -> "DataArray":
+        """Return the coordinate label of the minimum value along a dimension.
+
+        Returns a new `DataArray` named after the dimension with the values of
+        the coordinate labels along that dimension corresponding to minimum
+        values along that dimension.
+
+        In comparison to :py:meth:`~DataArray.argmin`, this returns the
+        coordinate label while :py:meth:`~DataArray.argmin` returns the index.
+
+        Parameters
+        ----------
+        dim : str, optional
+            Dimension over which to apply `idxmin`.  This is optional for 1D
+            arrays, but required for arrays with 2 or more dimensions.
+        skipna : bool or None, default None
+            If True, skip missing values (as marked by NaN). By default, only
+            skips missing values for ``float``, ``complex``, and ``object``
+            dtypes; other dtypes either do not have a sentinel missing value
+            (``int``) or ``skipna=True`` has not been implemented
+            (``datetime64`` or ``timedelta64``).
+        fill_value : Any, default NaN
+            Value to be filled in case all of the values along a dimension are
+            null.  By default this is NaN.  The fill value and result are
+            automatically converted to a compatible dtype if possible.
+            Ignored if ``skipna`` is False.
+        keep_attrs : bool, default False
+            If True, the attributes (``attrs``) will be copied from the
+            original object to the new one.  If False (default), the new object
+            will be returned without attributes.
+
+        Returns
+        -------
+        reduced : DataArray
+            New `DataArray` object with `idxmin` applied to its data and the
+            indicated dimension removed.
+
+        See also
+        --------
+        Dataset.idxmin, DataArray.idxmax, DataArray.min, DataArray.argmin
+
+        Examples
+        --------
+
+        >>> array = xr.DataArray([0, 2, 1, 0, -2], dims="x",
+        ...                      coords={"x": ['a', 'b', 'c', 'd', 'e']})
+        >>> array.min()
+        <xarray.DataArray ()>
+        array(-2)
+        >>> array.argmin()
+        <xarray.DataArray ()>
+        array(4)
+        >>> array.idxmin()
+        <xarray.DataArray 'x' ()>
+        array('e', dtype='<U1')
+
+        >>> array = xr.DataArray([[2.0, 1.0, 2.0, 0.0, -2.0],
+        ...                       [-4.0, np.NaN, 2.0, np.NaN, -2.0],
+        ...                       [np.NaN, np.NaN, 1., np.NaN, np.NaN]],
+        ...                      dims=["y", "x"],
+        ...                      coords={"y": [-1, 0, 1],
+        ...                              "x": np.arange(5.)**2}
+        ...                      )
+        >>> array.min(dim="x")
+        <xarray.DataArray (y: 3)>
+        array([-2., -4.,  1.])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        >>> array.argmin(dim="x")
+        <xarray.DataArray (y: 3)>
+        array([4, 0, 2])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        >>> array.idxmin(dim="x")
+        <xarray.DataArray 'x' (y: 3)>
+        array([16.,  0.,  4.])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        """
+        return computation._calc_idxminmax(
+            array=self,
+            func=lambda x, *args, **kwargs: x.argmin(*args, **kwargs),
+            dim=dim,
+            skipna=skipna,
+            fill_value=fill_value,
+            keep_attrs=keep_attrs,
+        )
+
+    def idxmax(
+        self,
+        dim: Hashable = None,
+        skipna: bool = None,
+        fill_value: Any = dtypes.NA,
+        keep_attrs: bool = None,
+    ) -> "DataArray":
+        """Return the coordinate label of the maximum value along a dimension.
+
+        Returns a new `DataArray` named after the dimension with the values of
+        the coordinate labels along that dimension corresponding to maximum
+        values along that dimension.
+
+        In comparison to :py:meth:`~DataArray.argmax`, this returns the
+        coordinate label while :py:meth:`~DataArray.argmax` returns the index.
+
+        Parameters
+        ----------
+        dim : str, optional
+            Dimension over which to apply `idxmax`.  This is optional for 1D
+            arrays, but required for arrays with 2 or more dimensions.
+        skipna : bool or None, default None
+            If True, skip missing values (as marked by NaN). By default, only
+            skips missing values for ``float``, ``complex``, and ``object``
+            dtypes; other dtypes either do not have a sentinel missing value
+            (``int``) or ``skipna=True`` has not been implemented
+            (``datetime64`` or ``timedelta64``).
+        fill_value : Any, default NaN
+            Value to be filled in case all of the values along a dimension are
+            null.  By default this is NaN.  The fill value and result are
+            automatically converted to a compatible dtype if possible.
+            Ignored if ``skipna`` is False.
+        keep_attrs : bool, default False
+            If True, the attributes (``attrs``) will be copied from the
+            original object to the new one.  If False (default), the new object
+            will be returned without attributes.
+
+        Returns
+        -------
+        reduced : DataArray
+            New `DataArray` object with `idxmax` applied to its data and the
+            indicated dimension removed.
+
+        See also
+        --------
+        Dataset.idxmax, DataArray.idxmin, DataArray.max, DataArray.argmax
+
+        Examples
+        --------
+
+        >>> array = xr.DataArray([0, 2, 1, 0, -2], dims="x",
+        ...                      coords={"x": ['a', 'b', 'c', 'd', 'e']})
+        >>> array.max()
+        <xarray.DataArray ()>
+        array(2)
+        >>> array.argmax()
+        <xarray.DataArray ()>
+        array(1)
+        >>> array.idxmax()
+        <xarray.DataArray 'x' ()>
+        array('b', dtype='<U1')
+
+        >>> array = xr.DataArray([[2.0, 1.0, 2.0, 0.0, -2.0],
+        ...                       [-4.0, np.NaN, 2.0, np.NaN, -2.0],
+        ...                       [np.NaN, np.NaN, 1., np.NaN, np.NaN]],
+        ...                      dims=["y", "x"],
+        ...                      coords={"y": [-1, 0, 1],
+        ...                              "x": np.arange(5.)**2}
+        ...                      )
+        >>> array.max(dim="x")
+        <xarray.DataArray (y: 3)>
+        array([2., 2., 1.])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        >>> array.argmax(dim="x")
+        <xarray.DataArray (y: 3)>
+        array([0, 2, 2])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        >>> array.idxmax(dim="x")
+        <xarray.DataArray 'x' (y: 3)>
+        array([0., 4., 4.])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        """
+        return computation._calc_idxminmax(
+            array=self,
+            func=lambda x, *args, **kwargs: x.argmax(*args, **kwargs),
+            dim=dim,
+            skipna=skipna,
+            fill_value=fill_value,
+            keep_attrs=keep_attrs,
+        )
+
     # this needs to be at the end, or mypy will confuse with `str`
     # https://mypy.readthedocs.io/en/latest/common_issues.html#dealing-with-conflicting-names
     str = property(StringAccessor)
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -930,7 +930,10 @@ class DataArray(AbstractArray, DataWithCoords):
         """
         variable = self.variable.copy(deep=deep, data=data)
         coords = {k: v.copy(deep=deep) for k, v in self._coords.items()}
-        indexes = self._indexes
+        if self._indexes is None:
+            indexes = self._indexes
+        else:
+            indexes = {k: v.copy(deep=deep) for k, v in self._indexes.items()}
         return self._replace(variable, coords, indexes=indexes)
 
     def __copy__(self) -> "DataArray":
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -6,6 +6,7 @@ import warnings
 from collections import defaultdict
 from html import escape
 from numbers import Number
+from operator import methodcaller
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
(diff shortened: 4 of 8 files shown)


max-sixty:
This looks extremely good and thorough! Thanks @toddrjen !

I'll have a proper look through later. I see a couple of minor questions I'll add in too. Others feel free to get ahead of me!

xarray/core/dataarray.py
```diff
@@ -3430,6 +3432,225 @@ def pad(
...
+            res.attrs = self.attrs
+
+        return res
+
+    def idxmin(
+        self,
+        dim: Optional[Hashable] = None,
+        axis: Optional[int] = None,
+        skipna: Optional[bool] = None,
+        promote: Optional[bool] = None,
```
max-sixty: Nit: no need for `Optional` if the default is `None`

xarray/core/dataarray.py
```diff
@@ -3430,6 +3432,225 @@ def pad(
...
+        returns the index.
+
+        Parameters
+        ----------
+        dim : str (optional)
+            Dimension over which to apply `idxmin`.
+        axis : int (optional)
+            Axis(es) over which to repeatedly apply `idxmin`. Exactly one of
+            the 'dim' and 'axis' arguments must be supplied.
+        skipna : bool, optional
```
```suggestion
        skipna : bool, default True
```

(and elsewhere, unless you / anyone has a strong view on the current state)

xarray/core/dataarray.py
```diff
@@ -3430,6 +3432,225 @@ def pad(
...
+        Returns
+        -------
+        reduced : DataArray
+            New DataArray object with `idxmin` applied to its data and the
+            indicated dimension removed.
+
+        See also
+        --------
+        Dataset.idxmin, DataArray.idxmax, DataArray.min, DataArray.argmin
+        """
```
max-sixty: No obligation, but to the extent you're up for it, a couple of examples in the docstrings would be awesome.

xarray/core/dataset.py
```diff
@@ -4284,7 +4284,7 @@ def reduce(
 
     def map(
         self,
-        func: Callable,
+        func: Union[Callable, str],
```
max-sixty: I'm not opposed to this; though is there an advantage relative to passing the function directly?

toddrjen:
@keewis @max-sixty The map thing is purely a convenience function.  I know there are other ways to do it, but since I thought this would be a useful feature for users in its own right, I did it that way.  But of course I can do it another way if you disagree.

The one complication is that using `DataArray.idxmax` and `DataArray.idxmin` assumes that the `Dataset` would only ever contain `DataArray` objects.  That may be mostly the case now, but I didn't want to bake that into the code.  I could do it using a lambda or nested function, but as I said I thought this approach had other benefits to users.

I will address the other comments inline.

xarray/core/dataarray.py
```diff
@@ -3430,6 +3432,225 @@ def pad(
...
+        return self._calc_idxminmax(
+            func="argmin",
```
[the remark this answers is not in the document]
toddrjen: Done, I will submit a second commit (to be squashed later) with the changes all together.

xarray/core/dataset.py
```diff
@@ -5914,5 +5921,131 @@ def pad(
...
+        reduced : Dataset
+            New Dataset object with `idxmin` applied to its data and the
+            indicated dimension removed.
+
+        See also
+        --------
+        DataArray.idxmin, Dataset.idxmax, Dataset.min, Dataset.argmin
+        """
+        return self.map(
+            "idxmin",
```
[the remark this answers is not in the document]
toddrjen: Please see my separate comment.

xarray/core/dataarray.py
toddrjen: Fixed, will submit a new pull request with all changes

keewis: you could also push to this PR? I don't think you need to worry about the commit history, we always do a squash-merge so in master this whole PR ends up as a single commit.

xarray/core/dataarray.py
toddrjen: Fixed, again the fix will come in another commit.

xarray/core/dataset.py
toddrjen: Please see my separate comment on the subject.

xarray/core/dataarray.py
max-sixty: Yes fine either way; no need to get everything in here, but no need to do a separate PR if you've already done the code!

xarray/core/dataarray.py
toddrjen: Done

xarray/core/dataarray.py
toddrjen: Sorry, I misspoke.  I meant new commit, not new pull request.  The new commit has already been pushed to this branch, please take a look.

toddrjen:
@keewis @max-sixty The new commit with the requested changes has been pushed to this branch (except for the `map` one, pending ongoing discussion).  Please take a look.

toddrjen:
@max-sixty 

> To what extent should this support non-index coordinates?

I am not familiar with non-index coordinates, what are those?  

Do you mean non-dimension coordinates?  Does that even make sense in a general way?  If they are 1D and tied to just one dimension coordinate that could be done, but if they are not tied to any dimension or tied to multiple dimensions or otherwise not 1D I am not sure what it would mean to take the idxmin/idxmax of them.

xarray/core/dataarray.py
```diff
@@ -3545,41 +3545,79 @@ def idxmin(
...
+        <xarray.DataArray (y: 3)>
+        array([4, 0, 2])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+        >>> array.idxmin(dim="x")
+        <xarray.DataArray 'x' (y: 3)>
+        array([16.,  0.,  4.])
+        Coordinates:
+          * y        (y) int64 -1 0 1
+
```
max-sixty: These are perfect, thanks!

toddrjen: Great!

max-sixty:
> The one complication is that using `DataArray.idxmax` and `DataArray.idxmin` assumes that the `Dataset` would only ever contain `DataArray` objects

I hear you and share the impulse that baking this in seems not ideal. Though I think it's a reasonable compromise to make, and there are no plans to deviate from it.

(Ideally maybe we have a `._contained_class` attribute on the dataset which is almost always `DataArray`)

I think having a lambda is fine too.

xarray/core/dataset.py
```diff
@@ -5914,5 +5921,169 @@ def pad(
 
         return self._replace_vars_and_dims(variables)
 
+    def idxmin(
+        self,
+        dim: Hashable = None,
+        axis: int = None,
```
max-sixty: Generally I don't think we offer `axis` as a arg except to maintain compatibility, but no harm in including

xarray/core/dataset.py
```diff
@@ -5914,5 +5921,169 @@ def pad(
...
+            Dimension over which to apply `idxmin`.
+        axis : int, optional
+            Axis(es) over which to repeatedly apply `idxmin`. Exactly one of
+            the 'dim' and 'axis' arguments must be supplied.
+        skipna : bool or None, default None
+            If True, skip missing values (as marked by NaN). By default, only
+            skips missing values for float dtypes; other dtypes either do not
+            have a sentinel missing value (int) or skipna=True has not been
+            implemented (object, datetime64 or timedelta64).
+        promote :  bool or None, default True
```
```suggestion
        promote : bool or None, default True
```

max-sixty:
> Do you mean non-dimension coordinates?

Yes, thanks for clarifying

> Does that even make sense in a general way? If they are 1D and tied to just one dimension coordinate that could be done, but if they are not tied to any dimension or tied to multiple dimensions or otherwise not 1D I am not sure what it would mean to take the idxmin/idxmax of them.

Yes, it wouldn't work in all cases, fair point. There are some cases in which it would work though, I'm unsure if it would be too complicated an interface to return them depending on whether it would work. 

(and completely fine to contemplate these later)

xarray/core/dataset.py
```diff
@@ -5914,5 +5921,169 @@ def pad(
 
         return self._replace_vars_and_dims(variables)
 
+    def idxmin(
+        self,
+        dim: Hashable = None,
+        axis: int = None,
+        skipna: bool = None,
+        promote: bool = None,
```
max-sixty: And generally I don't think we've given a `promote` option but, again, no harm in including

xarray/core/dataarray.py
```diff
@@ -3430,6 +3432,301 @@ def pad(
         )
         return self._from_temp_dataset(ds)
 
+    def _calc_idxminmax(
```
max-sixty: We've generally put these sorts of functions in `core/computation.py`, but I actually think it's fine here assuming others do

xarray/core/dataset.py
toddrjen: I was trying to keep the arguments similar to`min`, `max`, `argmin`, and `argmax`

xarray/core/dataset.py
toddrjen: This was the part I was most uncertain about, both the presence of the argument and its default.  I thought this was a bit of a special case since the data being tested is not the same as the data being returned, which can lead to non-intuitive results.

The options I thought of were:

   1. Always raisr an exception on operations that would insert an NaN.  This is the safest option, and the same as what `argmin` and `argmax` do, but would prevent an enormous number of valid operations and is overall not a good option.  It is the easiest to implement, though.
   2. Raise an exception if it would insert an NaN value in a data type that doesn't support it.  This is safe, but can have some unintuitive results since valid data for analysis can lead to invalid results.
   3. Only promote if NaN values are actually encountered.  This is safe, but makes the results hard to predict since they data type of the result depends on the specific values in the input.
   4. Always promote.  This works in the widest variety of situations, but can change data types where it isn't wanted.
  5. Ask

I settled on 5 with a default of 4 because I thought it worked most generally while still allowing people to restrict it only to save operations.

xarray/core/dataset.py
toddrjen: I'll fix this

xarray/core/dataset.py
max-sixty: Cool, I think that's very reasonable, and well laid-out options. 

Any other thoughts? @pydata/xarray on the general policy? I know we've had discussions on this before.

xarray/core/dataarray.py
dcherian: Yeah let's move it to `computation.py`. `dataarray.py` is quite long

xarray/core/dataset.py
dcherian: `axis` is really confusing and doesn't really fit in with xarray. Let's not add it?

toddrjen: @dcherian: it is already present in `min`, `max`, `argmin`, and `argmax`.  Since these all serve similar purposes, shouldn't we keep them as consistent as possible?

max-sixty: I think it's in those functions for compatibility with numpy, but I don't think it's in there intentionally. I'm not adamant either way, probably -0.2

xarray/core/dataarray.py
toddrjen: Okay

toddrjen:
I fixed the extra space in the docstring and moved the business logic to `computation.py`.

xarray/core/dataset.py
toddrjen: Okay, I removed it.

max-sixty:
LGTM! Any other thoughts before we merge?

xarray/core/computation.py
```diff
@@ -1306,3 +1307,86 @@ def where(cond, x, y):
...
+    dim: Optional[Hashable],
+    skipna: Optional[bool],
+    promote: Optional[bool],
+    keep_attrs: Optional[bool],
+    **kwargs: Any,
+):
+    """Apply common operations for idxmin and idxmax."""
+    # This function doesn't make sense for scalars so don't try
+    if not array.ndim:
+        ValueError("This function does not apply for scalars")
```
shoyer:
```suggestion
        raise ValueError("This function does not apply for scalars")
```

It would probably also be a good idea to add a test for these errors just in case :)

xarray/core/computation.py
```diff
@@ -1306,3 +1307,86 @@ def where(cond, x, y):
...
+    elif array.ndim == 1:
+        # it is okay to guess the dim if there is only 1
+        dim = array.dims[0]
+    else:
+        # The dim is not specified and ambiguous.  Don't guess.
+        raise ValueError("Must supply 'dim' argument for multidimensional arrays")
+
+    if dim in array.coords:
+        pass  # This is okay
+    else:
```
shoyer:
```suggestion
    if dim not in array.coords:
```

xarray/core/dataarray.py
```diff
@@ -3430,6 +3430,202 @@ def pad(
...
+        Coordinates:
+          * y        (y) int64 -1 0 1
+
+        See also
+        --------
+        Dataset.idxmin, DataArray.idxmax, DataArray.min, DataArray.argmin
+        """
+        return computation._calc_idxminmax(
+            array=self,
+            func=lambda x, *args, **kwargs: x.argmin(*args, **kwargs),
```
shoyer: Optional: could use `operator.methodcaller('argmin')` here.

xarray/core/dataset.py
shoyer: I think (4) is a reasonable default choice. One other reason why it is nice is that Dask won't work great with the other options, because that would require doing a computational pass over the data to figure out the dtype.

xarray/core/dataset.py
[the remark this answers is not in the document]
shoyer: I would rather use `methodcaller` here.

xarray/core/dataset.py
shoyer: I'm not a big fan of this as currently written. The main reason is consistency: we don't support strings instead of functions in other xarray methods, like `Dataset.reduce`.

We could consider this more broadly, but I would rather decouple this from this change.

xarray/core/computation.py
```diff
@@ -1306,3 +1307,86 @@ def where(cond, x, y):
...
+        raise KeyError(f'Dimension "{dim}" does not have coordinates')
+
+    # These are dtypes with NaN values argmin and argmax can handle
+    na_dtypes = "cf0"
+
+    if skipna or (skipna is None and array.dtype.kind in na_dtypes):
```
shoyer: I think the right check here is probably:
```suggestion
    if skipna or (skipna is None and issubclass(array.dtype.type, np.inexact):
```
This will handle all complex/float types: https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.scalars.html#arrays-scalars-built-in

xarray/core/computation.py
```diff
@@ -1306,3 +1307,86 @@ def where(cond, x, y):
...
+        # The dim is not specified and ambiguous.  Don't guess.
+        raise ValueError("Must supply 'dim' argument for multidimensional arrays")
+
+    if dim in array.coords:
+        pass  # This is okay
+    else:
+        raise KeyError(f'Dimension "{dim}" does not have coordinates')
+
+    # These are dtypes with NaN values argmin and argmax can handle
+    na_dtypes = "cf0"
```
shoyer: What is the `0` dtype?

xarray/core/dataset.py
shoyer: It occurs to me that there is at least one other possibly defensible behavior in the case of all NaN values: returning an arbitrarily selected coordinate value. This would still satisfy the invariant `array.sel({dim: ds.idxmin(dim)}) == array.min(dim)`.

shoyer:
In general this is really nicely put together. My main asks:

1. Remove the unrelated API change to `map`
2. Think about if the alternative of returning an arbitrary value rather than promotion or raising an error.

For details see the comments above. If we want  to support (2), then it might make sense to use a string for selecting values of `promote` rather than `True/False/None` (e.g., so we can include the option to return an arbitrary coordinate value).

xarray/core/computation.py
```diff
@@ -1306,3 +1307,86 @@ def where(cond, x, y):
...
+    if dim is not None:
+        pass  # Use the dim if available
+    elif array.ndim == 1:
+        # it is okay to guess the dim if there is only 1
+        dim = array.dims[0]
+    else:
+        # The dim is not specified and ambiguous.  Don't guess.
+        raise ValueError("Must supply 'dim' argument for multidimensional arrays")
+
+    if dim in array.coords:
```
shoyer: Let's also add a check that `dim` is in `array.dims`. Otherwise some other weird error could be triggered.

xarray/core/dataset.py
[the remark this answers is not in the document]
toddrjen: Since everyone seems to be against the `map` changes I have removed that.

xarray/core/dataset.py
toddrjen: This has been removed.

xarray/core/computation.py
toddrjen: Fixed, this was a leftover from when I also had an `axis` argument.

xarray/core/computation.py
toddrjen: Done

xarray/core/dataarray.py
toddrjen: This doesn't seem to work since as far as I can tell I need to specify all the arguments when setting up the methodcaller, but I don't know them at this point.

keewis: that's true. You could define your own version, though:
```python
def methodcaller(name):
    def caller(obj, *args, **kwargs):
        return getattr(obj, name)(*args, **kwargs)
    return caller
```
and then:
```python
return computation._calc_idxminmax(
    array=self,
    func=methodcaller("argmin"),
    ...,
)
```
which is much more readable than the lambda

shoyer: OK no worries, I think either the lambda or the new function is fine then.

xarray/core/dataset.py
toddrjen: @shoyer I should have mentioned that approach.  I didn't include that approach since it would be very easy to do manually at the end.  But it wouldn't tell anyone that they are dealing with an unusual situation.

xarray/core/computation.py
toddrjen: This is supposed to very specifically test the types that have NA types xarray can handle.  This particular test is used throughout xarray.

xarray/core/computation.py
toddrjen: A typo, it should be `'cfO'`.  I have added a test for this.

xarray/core/computation.py
shoyer: Looks good!

xarray/core/dataset.py
shoyer: How could you do "return an arbitrary value" without type promotion as a post-processing step?

toddrjen: @shoyer It would still be promoted.  Trying to let people specify an arbitrary value would considerably increase the complexity of the code.  There would need to be an additional argument, and more complicated promotion handling.  As it stands now, the automatic promotion is decided right at the beginning, so we only need to handle two cases for the most part.  Adding a fourth option would make that more complicated, and would require additional checks for the argument, and possibly still require promotion if the dtype of the fill value is different than the coordinate dtype.  It just opens up an enormous can of worms and corner cases that need to be resolved and tested that I think aren't really worth it.

shoyer: Just to throw out another API option: what about having a `fill_value` argument instead of `promote`? The default (`fill_value=dtypes.NA`) would do type promotion for integer dtypes and always fill with NA. Other values (e.g., `fill_value=0`) could be used to avoid type promotion with an integer coordinate.

Advantages:

- No special cases to keep track of.
- Consistent with other xarray methods that take a `fill_value` argument.

Disadvantages:

- No built-in way to raise an error instead of promotion (but users could do this themselves pretty easily)
- No built-in way to "only promote if necessary" (but this is a weird non-type stable API that doesn't work great with Dask, anyways)

toddrjen:
That could work.

The corner case we would need to decide on is again promotion.

What happens if the fill value is a "higher" type in the numeric tower than
the original type?  What if it is lower?

   1. We could try to always convert to the fill dtype (or more often the
dtype equivalent to the Python native type), and raise and exception of it
doesn't work.
   2. We could promote the fill value or original data, whichever is
"lower".

What if someone tries to use a string type for numeric data or vice
versus?  If we do option 1 that is easy.  Otherwise we probably need to use
numpy casting rules?


What about an object dtype fill value?

What about a date/time regard dtype?

xarray/core/dataset.py
toddrjen: That could work.

The corner case we would need to decide on is again promotion.

What happens if the fill value is a "higher" type in the numerical tower than
the original type?  What if it is lower?

   1. We could try to always convert to the fill dtype (or more often the
dtype equivalent to the Python native type), and raise and exception of it
doesn't work.
   2. We could promote the fill value or original data, whichever is
"lower".

What if someone tries to use a string type for numeric data or vice
versus?  If we do option 1 that is easy.  Otherwise we probably need to use
numpy casting rules?

What about an object dtype fill value?

What about a date/time related dtypes?

We would need to figure out how to handle these cases before I try to implement too much of this.

shoyer: I _think_ the right answer is to cast to the common dtype between the array and `fill_value`. This is what should automatically happen in xarray when you combine arrays of different types, e.g., using `duck_array_ops.where`:
https://github.com/pydata/xarray/blob/6378a711d50ba7f1ba9b2a451d4d1f5e1fb37353/xarray/core/duck_array_ops.py#L246

toddrjen: Okay, I have done that.

toddrjen:
Please see the newest version with the `promote` argument changed to `fill_value`.

toddrjen:
I am not sure why the tests are suddenly failing.  The tests were all working, then I rebased on the latest master and they are failing and I can't figure out why.

xarray/core/dataset.py
toddrjen: This is done, please check the lastest  commit.

toddrjen:
I figured out what is going wrong.  I will make a commit with a fix and include it in this pull request later today.

xarray/core/dataarray.py
```diff
@@ -3505,6 +3505,202 @@ def pad(
         )
         return self._from_temp_dataset(ds)
 
+    def idxmin(
+        self,
+        dim: Hashable = None,
+        skipna: bool = None,
+        fill_value: Any = np.NaN,
+        keep_attrs: Optional[bool] = False,
```
max-sixty: @toddrjen I should have spotted this earlier, and I appreciate you've gone through a lot of iteration on this PR already, thank you.

Could we conform to the global config on `keep_attrs`? It should only be setting this default to `None` and adding a couple of extra lines of code, copied from any (I hope...) of the other methods that take `keep_attrs`. Let me know any issues at all.

xarray/core/dataarray.py
```diff
@@ -3505,6 +3505,202 @@ def pad(
         )
         return self._from_temp_dataset(ds)
 
+    def idxmin(
+        self,
+        dim: Hashable = None,
+        skipna: bool = None,
+        fill_value: Any = np.NaN,
```
shoyer: Use `fill_value=dtypes.NA` here. Then use the utilities from `dtypes` to replace it on individual `Variable` objects :
https://github.com/pydata/xarray/blob/6378a711d50ba7f1ba9b2a451d4d1f5e1fb37353/xarray/core/variable.py#L1007-L1010

I think this should make it work on arbitrary dtypes, without any need for dtype specific checks elsewhere.

(This handling of `dtypes.NA` could probably be moved into a separate utility functions in `dtypes.py`, if you like)

xarray/core/dataarray.py
toddrjen: I will just delegate this to `argmin` and `argmax` and use whatever attributes they come up with.

xarray/core/dataarray.py
toddrjen: I don't currently do any dtype-specific checks, I am just letting `where` handle that.  But I will use `dtypes.NA`.

toddrjen:
I think I have implemented all the requested changes and all tests are passing.  Please take a look.

xarray/core/dataarray.py
```diff
@@ -930,7 +930,10 @@ def copy(self, deep: bool = True, data: Any = None) -> "DataArray":
         """
         variable = self.variable.copy(deep=deep, data=data)
         coords = {k: v.copy(deep=deep) for k, v in self._coords.items()}
-        indexes = self._indexes
+        if self._indexes is None:
+            indexes = self._indexes
+        else:
+            indexes = {k: v.copy(deep=deep) for k, v in self._indexes.items()}
```
max-sixty: I had thought that because indexes were immutable, we wouldn't need to deep copy, though the linked issue shows there's a flaw in that argument. 

This may have some performance implications. 

Do we need this change for this PR? We could merge the rest otherwise.

dcherian: Should deepcopy the `indexes` object rather than the `index` itself?

@toddrjen can you open a separate PR with this fix please?

toddrjen: Yes, the change is required for this PR.  My tests depend on it.  I guess I could conceivably reorganize them to work without it, but that is quite a bit of work for a problem that is already fixed.

xarray/core/duck_array_ops.py
```diff
@@ -303,19 +303,15 @@ def f(values, axis=None, skipna=None, **kwargs):
...
-                    "installed version of numpy; upgrade to numpy 1.12 "
-                    "or newer to use skipna=True or skipna=None" % name
+            if not isinstance(values, dask_array_type):
+                raise
+            try:  # dask/dask#3133 dask sometimes needs dtype argument
+                # if func does not accept dtype, then raises TypeError
+                return func(values, axis=axis, dtype=values.dtype, **kwargs)
+            except (AttributeError, TypeError):
+                raise NotImplementedError(
+                    f"{name} is not yet implemented on dask arrays"
```
shoyer: This is just removing compatibility code for old versions of NumPy, right?

xarray/core/dataarray.py
shoyer: I think this change _should_ be safe. It only does a deep copy of indexes if someone explicitly writes `array.copy(deep=True)`, in which I would guess this is desired behavior.

Ideally `pandas.Index` objects would have immutable values, but that isn't the case at the moment: https://github.com/pandas-dev/pandas/issues/33001

xarray/core/computation.py
```diff
@@ -1338,3 +1339,69 @@ def polyval(coord, coeffs, degree_dim="degree"):
...
     return (lhs * coeffs).sum(degree_dim)
+
+
+def _calc_idxminmax(
+    *,
+    array,
+    func: Callable,
+    dim: Optional[Hashable],
+    skipna: bool = None,
+    fill_value: Any = None,
```
shoyer: Should this be `dtypes.NA`? Or maybe we don't need default values at all, given that this is only called internally?

xarray/core/dataarray.py
```diff
@@ -3505,6 +3508,202 @@ def pad(
         )
         return self._from_temp_dataset(ds)
 
+    def idxmin(
+        self,
+        dim: Hashable = None,
+        skipna: bool = None,
+        fill_value: Any = dtypes.NA,
+        keep_attrs: bool = None,
+        **kwargs: Any,
```
shoyer: What other arguments would be valid to be passed as `kwargs`?

I would be inclined to list these specifically rather than handling anything, which could unintentionally expose details of xarray's internal implementation as a public API.

xarray/core/dataarray.py
toddrjen: @dcherian  As far as I can tell the `indexes` object is just a `dict`, which doesn't have a `deepcopy` option.  That is why I do it manually.

xarray/core/computation.py
```diff
@@ -1338,3 +1339,69 @@ def polyval(coord, coeffs, degree_dim="degree"):
...
+        # The dim is not specified and ambiguous.  Don't guess.
+        raise ValueError("Must supply 'dim' argument for multidimensional arrays")
+
+    if dim not in array.dims:
+        raise KeyError(f'Dimension "{dim}" not in dimension')
+    if dim not in array.coords:
+        raise KeyError(f'Dimension "{dim}" does not have coordinates')
+
+    # These are dtypes with NaN values argmin and argmax can handle
+    na_dtypes = "cfO"
```
shoyer: `datetime64` and `timedelta64` also have missing values. Those have type codes `m` and `M`.

To encourage consistency, would you mind moving this into a constant in `dtypes.py`? e.g.,
```python
NA_DTYPE_KINDS = 'cfOmM'
```

xarray/core/duck_array_ops.py
toddrjen: Sort of.  This code was catching *every* `AttributeError` or `TypeError`, under the assumption that the only ones that could happen were ones caused by this issue in numpy.  This made it impossible for me to find the source of an error I ran into while I was developing the main patch.

I strictly speaking could have reverted this after I got everything working, but since the check is no longer relevant, and this sort of error hiding can easily cause problems for others down the road, I figured I might as well submit the change,

xarray/core/computation.py
toddrjen: I'll fix that.

xarray/core/computation.py
toddrjen: Those aren't supported by `argmin` and `argmax` yet so I can't support them here, either.

xarray/core/dataarray.py
toddrjen: This is here because `argmin` and `argmax` have it and I wanted to keep as close to them as possible.  These arguments are passed directly to `argmin` and `argmax`.  As long as our `argmin` and `argmax` support it I would be inclined to keep it here.

xarray/core/computation.py
shoyer: Are timedelta/datetime dtypes just broken?

xarray/core/dataarray.py
shoyer: argmin/argmax have this because they use a generic template for reduction functions, but if we wrote more specific wrappers for them I would also drop `**kwargs`.

xarray/core/computation.py
toddrjen: They work without NaT values, I specifically test them.  But NaT values don't work.

xarray/core/dataarray.py
toddrjen: I can remove it if you really want.

shoyer: Yes, I'd prefer to keep it out for now. It's always easier to expand API later rather shrink them :)

toddrjen: Okay

toddrjen:
Here is a new commit with the discussed changes.

max-sixty:
Thank you again @toddrjen, both for the content and the iteration.

@shoyer any final thoughts? Otherwise I'll merge tomorrow?

xarray/core/duck_array_ops.py
shoyer: Sounds good, thanks!

doc/whats-new.rst
```diff
@@ -35,13 +35,19 @@ New Features
   :py:func:`combine_by_coords` and :py:func:`combine_nested` using
   combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
   By `John Omotani <https://github.com/johnomotani>`_
+- Implement :py:meth:`~xarray.DataArray.idxmax`, :py:meth:`~xarray.DataArray.idxmin`,
```
```suggestion
- Implement :py:meth:`DataArray.idxmax`, :py:meth:`DataArray.idxmin`,
```

doc/whats-new.rst
```diff
@@ -35,13 +35,19 @@ New Features
...
+  :py:meth:`~xarray.Dataset.idxmax`, :py:meth:`~xarray.Dataset.idxmin`.  (:issue:`60`, :pull:`3871`)
```
```suggestion
  :py:meth:`Dataset.idxmax`, :py:meth:`Dataset.idxmin`.  (:issue:`60`, :pull:`3871`)
```

doc/whats-new.rst
```diff
@@ -35,13 +35,19 @@ New Features
...
+  By `Todd Jennings <https://github.com/toddrjen>`_
 - Limited the length of array items with long string reprs to a
   reasonable width (:pull:`3900`)
   By `Maximilian Roos <https://github.com/max-sixty>`_
 
 
 Bug fixes
 ~~~~~~~~~
+- Fix regression where deleting a coordinate from a copied `DataArray` can affect
```
```suggestion
- Fix regression where deleting a coordinate from a copied :py:class:`DataArray` can affect
```

doc/whats-new.rst
```diff
@@ -35,13 +35,19 @@ New Features
...
+  the original `Dataarray`.  (:issue:`3899`, :pull:`3871`)
```
```suggestion
  the original :py:class:`DataArray`.  (:issue:`3899`, :pull:`3871`)
```

xarray/core/dataarray.py
```diff
@@ -3505,6 +3508,191 @@ def pad(
...
+        fill_value: Any = dtypes.NA,
+        keep_attrs: bool = None,
+    ) -> "DataArray":
+        """Return the coordinate of the minimum value along a dimension.
+
+        Returns a new DataArray named after the dimension with the values of
+        the coordinate along that dimension corresponding to minimum value
+        along that dimension.
+
+        In comparison to `argmin`, this returns the coordinate while `argmin`
```
dcherian:
```suggestion
        In comparison to `argmin`, this returns the coordinate label  while `argmin`
```

xarray/core/dataarray.py
```diff
@@ -3505,6 +3508,191 @@ def pad(
...
+        fill_value: Any = dtypes.NA,
+        keep_attrs: bool = None,
+    ) -> "DataArray":
+        """Return the coordinate of the maximum value along a dimension.
+
+        Returns a new DataArray named after the dimension with the values of
+        the coordinate along that dimension corresponding to maximum value
+        along that dimension.
+
+        In comparison to `argmax`, this returns the coordinate while `argmax`
```
dcherian:
```suggestion
        In comparison to `argmax`, this returns the coordinate label while `argmax`
```

xarray/core/dataset.py
```diff
@@ -6093,5 +6094,155 @@ def pad(
...
+        fill_value: Any = dtypes.NA,
+        keep_attrs: bool = None,
+    ) -> "Dataset":
+        """Return the coordinate of the minimum value along a dimension.
+
+        Returns a new Dataset named after the dimension with the values of
+        the coordinate along that dimension corresponding to minimum value
+        along that dimension.
+
+        In comparison to `argmin`, this returns the coordinate while `argmin`
```
dcherian:
```suggestion
        In comparison to `argmin`, this returns the coordinate label while `argmin`
```

max-sixty:
Amazing — let's merge on green!

toddrjen:
I have gone over it one more time and made a few documentation fixes.  Please take one more look before merging.

dcherian:
Thanks @toddrjen that's a great contribution!

max-sixty:
@toddrjen thank you again! This ended up being quite an adventure, really appreciate you pushing all the way.

xarray/core/computation.py
```diff
@@ -1338,3 +1339,66 @@ def polyval(coord, coeffs, degree_dim="degree"):
...
+
+    # This will run argmin or argmax.
+    indx = func(array, dim=dim, axis=None, keep_attrs=keep_attrs, skipna=skipna)
+
+    # Get the coordinate we want.
+    coordarray = array[dim]
+
+    # Handle dask arrays.
+    if isinstance(array, dask_array_type):
+        res = dask_array.map_blocks(coordarray, indx, dtype=indx.dtype)
```
kmuehlbauer: @toddrjen @dcherian 

Sorry, I might be wrong, but it seems, that the `func` is missing as argument to `map_blocks`. I tried  with `lambda a, b: a[b]` which seems to work.

kmuehlbauer: Two more things:

- `isinstance`-check need to use `array.data`. 
- `res` need to be computed, otherwise subsequent actions with `res` will fail.

shoyer: You’re right, this is definitely broken. Anyone up for putting together a fix in a follow up PR?

kmuehlbauer: @shoyer See #3922

Issue #3867, opened 2020-03-19
macos py38 CI failing


dcherian:
`import matplotlib` is failing when it imports `PIL`

```python

E       ImportError: dlopen(/usr/local/miniconda/envs/xarray-tests/lib/python3.8/site-packages/PIL/_imaging.cpython-38-darwin.so, 2): Library not loaded: @rpath/libwebp.7.dylib
E         Referenced from: /usr/local/miniconda/envs/xarray-tests/lib/libtiff.5.dylib
E         Reason: Incompatible library version: libtiff.5.dylib requires version 9.0.0 or later, but libwebp.7.dylib provides version 8.0.0

/usr/local/miniconda/envs/xarray-tests/lib/python3.8/site-packages/PIL/Image.py:69: ImportError
```

max-sixty:
The only reference I could find is https://github.com/BVLC/caffe/issues/1599#issuecomment-144854014, from 2015...

Does anyone have more experience with these?

keewis:
this should be closer: https://github.com/conda-forge/libwebp-feedstock/issues/26 which is a work in progress. If we don't want to wait on a fix, we could temporarily pin `libtiff` on OSX.

keewis:
the libwebp issue has been closed, so I guess we can try reactivating the OSX CI?

PR #3875, opened 2020-03-21
remove macos build while waiting for libwebp fix

max-sixty:
- [x] Closes https://github.com/pydata/xarray/issues/3867

I think fine to pause the macos build, given the relative cost of fixing vs chance of finding a bug only in macos


keewis:
I agree, it should be fine not to run the macos CI until the packages are fixed.

You might also be able to temporarily bypass the issue by adding `- libtiff<4.1.0 # [osx]` to `py38.yml`.

PR #3877, opened 2020-03-23
Control attrs of result in `merge()`, `concat()`, `combine_by_coords()` and `combine_nested()`

### LINKED ISSUE #3865: `merge` drops attributes
`xarray.merge()` drops the `attrs` of `Dataset`s being merged. They should be kept, at least if they are compatible

#### MCVE Code Sample

```python
# Your code here
import xarray as xr

ds1 = xr.Dataset()
ds1.attrs['a'] = 42
ds2 = xr.Dataset()
ds2.attrs['a'] = 42

merged = xr.merge([ds1, ds2])

print(merged)
```
the result is
```
<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*
```

#### Expected Output
```
<xarray.Dataset>
Dimensions:  ()
Data variables:
    *empty*
Attributes:
    a:        42
```

#### Problem Description

The current behaviour means I have to check and copy `attrs` to the result of `merge` by hand, even if the `attrs` of the inputs were identical or not conflicting.

I'm happy to attempt a PR to fix this.
Proposal (following pattern of `compat` arguments):
* add a `combine_attrs` argument to `xarray.merge`
* `combine_attrs = 'drop'` do not copy `attrs` (current behaviour)
* `combine_attrs = 'identical'` if `attrs` of all inputs are identical (using `dict_equiv`) then copy the `attrs` to the result, otherwise raise an exception
* `combine_attrs = 'no_conflicts'` merge the `attrs` of all inputs, as long as any keys shared by more than one input have the same value (if not raise an exception) [I propose this is the default behaviour]
* `override` copy the `attrs` from the first input, to the result

This proposal should also allow `combine_by_coords`, etc. to preserve attributes. These should probably also take a `combine_attrs` argument, which would be passed through to `merge`.

#### Versions

Current `master` of `pydata/xarray` on 17/3/2020
<details><summary>Output of `xr.show_versions()`</summary>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
python-bits: 64
OS: Linux
OS-release: 5.3.0-40-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8
libhdf5: 1.10.2
libnetcdf: 4.6.3

xarray: 0.15.0
pandas: 1.0.2
numpy: 1.18.1
scipy: 1.3.0
netCDF4: 1.5.1.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.12.0
distributed: None
matplotlib: 3.1.1
cartopy: None
seaborn: None
numbagg: None
setuptools: 45.2.0
pip: 9.0.1
conda: None
pytest: 4.4.1
IPython: 7.8.0
sphinx: 1.8.3
</details>

johnomotani:
`combine_attrs` argument for `merge()`, `concat()`, `combine_by_coords()` and `combine_nested()` controls what attributes the result is given. Defaults maintain the current behaviour. Possible values (named following `compat` arguments) are:
- 'drop': empty attrs on returned Dataset.
- 'identical': all attrs must be the same on every object.
- 'no_conflicts': attrs from all objects are combined, any that have the same name must also have the same value.
- 'override': skip comparing and copy attrs from the first dataset to the result.

 - [x] Closes #3865 
 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -25,6 +25,10 @@ Breaking changes
 New Features
 ~~~~~~~~~~~~
 
+- Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
+  :py:func:`combine_by_coords` and :py:func:`combine_nested` using
+  combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
+  By `John Omotani <https://github.com/johnomotani>`_
 - Weighted array reductions are now supported via the new :py:meth:`DataArray.weighted`
   and :py:meth:`Dataset.weighted` methods. See :ref:`comput.weighted`. (:issue:`422`, :pull:`2922`).
   By `Mathias Hauser <https://github.com/mathause>`_
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -15,6 +15,27 @@ What's New
 
 .. _whats-new.0.15.1:
 
+v0.16.0 (unreleased)
+---------------------
+
+Breaking changes
+~~~~~~~~~~~~~~~~
+
+New Features
+~~~~~~~~~~~~
+- Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
+  :py:func:`combine_by_coords` and :py:func:`combine_nested` using
+  combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
+  By `John Omotani <https://github.com/johnomotani>`_
+
+Bug fixes
+~~~~~~~~~
+
+Documentation
+~~~~~~~~~~~~~
+
+Internal Changes
+~~~~~~~~~~~~~~~~
 
 v0.15.1 (unreleased)
 ---------------------
@@ -25,10 +46,6 @@ Breaking changes
 New Features
 ~~~~~~~~~~~~
 
-- Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
-  :py:func:`combine_by_coords` and :py:func:`combine_nested` using
-  combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
-  By `John Omotani <https://github.com/johnomotani>`_
 - Weighted array reductions are now supported via the new :py:meth:`DataArray.weighted`
   and :py:meth:`Dataset.weighted` methods. See :ref:`comput.weighted`. (:issue:`422`, :pull:`2922`).
   By `Mathias Hauser <https://github.com/mathause>`_
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -24,32 +24,15 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
-
-
-Bug fixes
-~~~~~~~~~
-
-v0.16.0 (unreleased)
----------------------
-
-Breaking changes
-~~~~~~~~~~~~~~~~
-
-New Features
-~~~~~~~~~~~~
 - Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
   :py:func:`combine_by_coords` and :py:func:`combine_nested` using
   combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
   By `John Omotani <https://github.com/johnomotani>`_
 
+
 Bug fixes
 ~~~~~~~~~
 
-Documentation
-~~~~~~~~~~~~~
-
-Internal Changes
-~~~~~~~~~~~~~~~~
 
 Documentation
 ~~~~~~~~~~~~~
--- a/xarray/core/combine.py
+++ b/xarray/core/combine.py
@@ -155,6 +155,7 @@ def _combine_nd(
     compat="no_conflicts",
     fill_value=dtypes.NA,
     join="outer",
+    combine_attrs="drop",
 ):
     """
     Combines an N-dimensional structure of datasets into one by applying a
@@ -202,13 +203,21 @@ def _combine_nd(
             compat=compat,
             fill_value=fill_value,
             join=join,
+            combine_attrs=combine_attrs,
         )
     (combined_ds,) = combined_ids.values()
     return combined_ds
 
 
 def _combine_all_along_first_dim(
-    combined_ids, dim, data_vars, coords, compat, fill_value=dtypes.NA, join="outer"
+    combined_ids,
+    dim,
+    data_vars,
+    coords,
+    compat,
+    fill_value=dtypes.NA,
+    join="outer",
+    combine_attrs="drop",
 ):
 
     # Group into lines of datasets which must be combined along dim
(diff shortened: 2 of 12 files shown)


xarray/core/dataarray.py
```diff
@@ -487,6 +493,8 @@ def to_dataset(self, dim: Hashable = None, *, name: Hashable = None) -> Dataset:
         name : hashable, optional
             Name to substitute for this array's name. Only valid if ``dim`` is
             not provided.
+        promote_attrs : bool, optional
```
```suggestion
        promote_attrs : bool, default False
```

xarray/core/utils.py
```diff
@@ -371,6 +371,35 @@ def ordered_dict_intersection(
     return new_dict
 
 
+def ordered_dict_union(
+    first_dict: MutableMapping[K, V],
```
max-sixty: Why does this need to be mutable?

xarray/core/concat.py
```diff
@@ -92,15 +93,20 @@ def concat(
         - 'override': if indexes are of same size, rewrite indexes to be
           those of the first object with that dimension. Indexes for the same
           dimension must have the same size in all objects.
+    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'}, optional
```
max-sixty:
```suggestion
    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'}, default 'override'
```

xarray/core/utils.py
johnomotani: `update_safety_check()` expects a `MutableMapping` as its first argument. I can't see why that needs to be mutable though... Should I change both `update_safety_check()` and `ordered_dict_union()` to have their first arg be `Mapping[K, V]`?

xarray/core/combine.py
```diff
@@ -390,6 +410,15 @@ def combine_nested(
         - 'override': if indexes are of same size, rewrite indexes to be
           those of the first object with that dimension. Indexes for the same
           dimension must have the same size in all objects.
+    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'}, optional
```
johnomotani: This could note the default too.

```suggestion
    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'},
                    default 'drop'
```

xarray/core/merge.py
```diff
@@ -614,6 +666,15 @@ def merge(
           dimension must have the same size in all objects.
     fill_value : scalar, optional
         Value to use for newly missing values
+    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'}, optional
```
johnomotani: note default

```suggestion
    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'},
                    default 'drop'
```

xarray/core/combine.py
```diff
@@ -557,6 +588,15 @@ def combine_by_coords(
...
```
johnomotani: note default

```suggestion
    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'},
                    default 'drop'
```

xarray/core/concat.py
johnomotani: need a line break to avoid going over 79 characters
```suggestion
    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'},
                    default 'override
```

xarray/core/utils.py
max-sixty: I think so, yes

max-sixty: Should this method be called `ordered`?

johnomotani: Not sure... If the comparison defined by the `compat` argument means the dict values are identical, then the order of first and second argument doesn't matter, but if something is passed as the `compat` argument that returns `True` when they are 'compatible' but not identical, then the value from the first dict will be used in the result. I think that's the meaning of 'ordered' here (I copied the naming from `ordered_dict_intersection` just above).

max-sixty: Ah I reckon the above method is left over from when we used `OrderedDict`s. Shall we rename this to something like `compat_dict_intersection` (or whatever you think is better)? 

Could also add a TODO in the above method (but let's change it in a different PR, no need to slow down this)

max-sixty:
Any other thoughts from anyone before we hit the big green button?

TomNicholas:
Oh sweet!

Should there perhaps be another option to specify which object to get the attrs from? I'm just thinking by analogy to how `open_mfdataset` now lets you [specify which file](https://github.com/pydata/xarray/pull/3498) you want the attrs from.

johnomotani:
> Should there perhaps be another option to specify which object to get the attrs from? I'm just thinking by analogy to how open_mfdataset now lets you specify which file you want the attrs from.

That would actually be nice to have in `concat()` for a use-case I have. It's not immediately obvious to me how to implement it though. For `merge()` or `concat()` you could give an integer index. I think `open_mfdataset` used the file-name (?), but there's no equivalent for `combine_by_coords` or `combine_nested` is there?

johnomotani:
For specifying which object, one possibility would be to pass an `int` to `combine_attrs` in `merge()`, `concat()` or `combine_by_coords()`, or a tuple of `int` to `combine_nested`, giving the index of the object to use attributes from. This feature would need new tests writing though, so I'd suggest implementing it in a new PR.

TomNicholas:
> For specifying which object, one possibility would be to pass an int

I'm not in favour of an integer argument, for the reasons discussed for `open_mfdataset` [here](https://github.com/pydata/xarray/pull/3498#discussion_r358358171).

> This feature would need new tests writing though, so I'd suggest implementing it in a new PR.

Fine to discuss this on a separate PR though.

doc/whats-new.rst
```diff
@@ -25,6 +25,10 @@ Breaking changes
 New Features
 ~~~~~~~~~~~~
 
+- Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
+  :py:func:`combine_by_coords` and :py:func:`combine_nested` using
+  combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
+  By `John Omotani <https://github.com/johnomotani>`_
```
max-sixty: Could we merge master and move this to 0.16.0?

max-sixty:
Let's merge later unless other comments?

max-sixty:
Merging — any other feedback please post and we can follow-up on.

Thanks a lot @johnomotani , great PR!

PR #3882, opened 2020-03-23
Tweaks to "how_to_release"

max-sixty:


keewis:
while you're at it, I think this should be `should include`, not `show include`:
https://github.com/pydata/xarray/blob/6979a978f220090de7b597b1c580cb515508484e/HOW_TO_RELEASE.md#L123-L124

PR #3885, opened 2020-03-24
xfail test_uamiv_format_write

max-sixty:
- [x] Triages https://github.com/pydata/xarray/issues/3711

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -3548,7 +3548,7 @@ class TestPseudoNetCDFFormat:
         assert_allclose(expected, actual)
         camxfile.close()
 
-    @pytest.mark.xfail("Flaky, GH3711")
+    @pytest.mark.xfail(reason="Flaky; see GH3711")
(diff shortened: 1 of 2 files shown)


spencerkclark:
FYI you'll get an upstream-dev failure due to some ongoing work in cftime: https://github.com/Unidata/cftime/issues/158#issuecomment-603242434.  You can temporarily remove it from the build if you'd like, but hopefully it will be resolved soon.

max-sixty:
Cheers @spencerkclark . Do you know which tests we should xfail? We can do it ex-post if you're not sure.

I think we should really try and keep the build passing, partly as contributor-friendliness, partly as signalling, partly as internalizing the externalities of flaky tests

spencerkclark:
Absolutely; it's this one:
https://github.com/pydata/xarray/blob/c707b337a0f75224ee3b3e3b65a08da792df2fa6/xarray/tests/test_cftime_offsets.py#L1179-L1184

Feel free to add an xfail here.

PR #3886, opened 2020-03-24
Use `fixes` in PR template

max-sixty:
`Closes` doesn't seem to add the issue to "Linked Issues" and auto-close. `Fixes` works elsewhere.


keewis:
FYI `Closes` does work, but only with `#<issue-number>` and not with full links.

max-sixty:
Ah! Thanks. 

Does Fixes work with links? Or should we revert?

keewis:
I don't think you can use full links with either `Fixes` or `Closes`. I've seen the use of `gh-<issue-number>`, though.

PR #3887, opened 2020-03-24
Rename ordered_dict_intersection -> compat_dict_intersection

johnomotani:
`xarray` does not use `OrderedDicts` any more, so name did not make sense. As suggested here https://github.com/pydata/xarray/pull/3877#discussion_r396620551.

 - [x] Test~~s~~ ~~added~~ updated

### DIFF
--- a/xarray/core/utils.py
+++ b/xarray/core/utils.py
@@ -343,7 +343,7 @@ def dict_equiv(
     return True
 
 
-def ordered_dict_intersection(
+def compat_dict_intersection(
     first_dict: Mapping[K, V],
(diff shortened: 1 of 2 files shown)


max-sixty:
Erm, is this function not used anywhere apart from its tests? 😄

johnomotani:
Doesn't look like it, but it seems like a thing that could be useful, e.g. for handling combining `attrs` dicts, so why delete it?

max-sixty:
Agree!

max-sixty:
Thanks @johnomotani

PR #3897, opened 2020-03-26
expose a few zarr backend functions as semi-public api

### LINKED ISSUE #3851: Exposing Zarr backend internals as semi-public API
We recently built a prototype REST API for serving xarray datasets via a Fast-API application (see #3850  for more details). In the process of doing this, we needed to use [a few internal functions in Xarray's Zarr backend](https://github.com/jhamman/xpublish/blob/aff49ec09136a29b56167a1d627fcb3a13fa4d01/xpublish/rest.py#L13-L20):

```python
from xarray.backends.zarr import (
    _DIMENSION_KEY,
    _encode_zarr_attr_value,
    _extract_zarr_variable_encoding,
    encode_zarr_variable,
)
from xarray.core.pycompat import dask_array_type
from xarray.util.print_versions import get_sys_info, netcdf_and_hdf5_versions
```

Obviously, none of these imports are really meant for use outside of Xarray's backends so I'd like to discuss how we may go about exposing these functions (or variables) as semi-public (advanced use) API features. Thoughts?

cc @rabernat

jhamman:
- [x] Fixes #3851
 - [x] Tests added

### DIFF
--- a/xarray/backends/zarr.py
+++ b/xarray/backends/zarr.py
@@ -10,13 +10,20 @@ from ..core.variable import Variable
 from .common import AbstractWritableDataStore, BackendArray, _encode_variable_name
 
 # need some special secret attributes to tell us the dimensions
-_DIMENSION_KEY = "_ARRAY_DIMENSIONS"
+DIMENSION_KEY = "_ARRAY_DIMENSIONS"
 
 
-# zarr attributes have to be serializable as json
-# many xarray datasets / variables have numpy arrays and values
-# these functions handle encoding / decoding of such items
-def _encode_zarr_attr_value(value):
+def encode_zarr_attr_value(value):
+    """
+    Encode a attribute value as something that can be serialized as json
+
+    Many xarray datasets / variables have numpy arrays and values. This
+    function handles encoding / decoding of such items.
+
+    ndarray -> list
+    scalar array -> scalar
+    other -> other (no change)
+    """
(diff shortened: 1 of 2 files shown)


xarray/backends/zarr.py
```diff
@@ -10,13 +10,20 @@
 from .common import AbstractWritableDataStore, BackendArray, _encode_variable_name
 
 # need some special secret attributes to tell us the dimensions
-_DIMENSION_KEY = "_ARRAY_DIMENSIONS"
+DIMENSION_KEY = "_ARRAY_DIMENSIONS"
```
max-sixty: We've successfully used `...` to represent all dimensions in `transpose` & `stack`, to the extent we could replicate that here (maybe another time though)

jhamman: This string is used as a special key in Zarr's array metadata so I think this are mostly unrelated to the selection of specific (all) dimensions.

max-sixty: Ah I see, cheers

dcherian:
Add to api.rst?

jhamman:
> Add to api.rst?

Personally, I'm not sure this is required for these functions. For my particular applications, just having some basic tests and doc strings is enough.

dcherian:
Yes I'm not sure either. Just wanted to bring it up. Perhaps  we should add it once the backends third-party interface is ready.

Issue #3759, opened 2020-02-06
Truncate long lines in repr of coords


max-sixty:
#### MCVE Code Sample

```python
xr.DataArray(coords=dict(a=' '.join(['hello world' for _ in range(100)])))

<xarray.DataArray ()>
array(nan)
Coordinates:
    a        <U5999 'hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello  world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world hello world'
```

#### Expected Output

```
<xarray.DataArray ()>
array(nan)
Coordinates:
    a        <U5999 'hello world ... hello world'
```

#### Problem Description

I think mostly the same as https://github.com/pydata/xarray/issues/1319 but for coords

#### Output of ``xr.show_versions()``
<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: [...]
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.utf8
LOCALE: en_US.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.7.1

xarray: 0.15.0
pandas: 0.25.3
numpy: 1.17.3
scipy: 1.3.2
netCDF4: 1.5.3
pydap: None
h5netcdf: 0.7.4
h5py: 2.10.0
Nio: None
zarr: None
cftime: 1.0.4.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
[...]

PR #3900, opened 2020-03-27
Limit repr of arrays containing long strings

max-sixty:
- [x] Fixes https://github.com/pydata/xarray/issues/3759
 - [x] Tests added

This also more strictly limits the length of array reprs where the max_width is small, such that it doesn't always display the first and last item. Lmk if that could have a adverse impact anywhere.

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -4,6 +4,7 @@ import contextlib
 import functools
 from datetime import datetime, timedelta
 from itertools import zip_longest
+from typing import Hashable
 
 import numpy as np
 import pandas as pd
@@ -14,7 +15,7 @@ from .options import OPTIONS
 from .pycompat import dask_array_type, sparse_array_type
 
 
-def pretty_print(x, numchars):
+def pretty_print(x, numchars: int):
     """Given an object `x`, call `str(x)` and format the returned string so
     that it is numchars long, padding with trailing spaces or truncating with
     ellipses as necessary
@@ -163,7 +164,7 @@ def format_items(x):
     return formatted
 
 
-def format_array_flat(array, max_width):
+def format_array_flat(array, max_width: int):
     """Return a formatted string for as many items in the flattened version of
     array that will fit within max_width characters.
     """
@@ -198,11 +199,20 @@ def format_array_flat(array, max_width):
     num_back = count - num_front
     # note that num_back is 0 <--> array.size is 0 or 1
     #                         <--> relevant_back_items is []
-    pprint_str = (
-        " ".join(relevant_front_items[:num_front])
-        + padding
-        + " ".join(relevant_back_items[-num_back:])
+    pprint_str = "".join(
+        [
+            " ".join(relevant_front_items[:num_front]),
+            padding,
+            " ".join(relevant_back_items[-num_back:]),
+        ]
     )
+
+    # As a final check, if it's still too long even with the limit in values,
+    # replace the end with an ellipsis
+    # NB: this will still returns a full 3-character ellipsis when max_width < 3
+    if len(pprint_str) > max_width:
+        pprint_str = pprint_str[: max(max_width - 3, 0)] + "..."
+
     return pprint_str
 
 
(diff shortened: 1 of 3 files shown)


xarray/core/formatting.py
```diff
@@ -258,10 +268,16 @@ def inline_variable_array_repr(var, max_width):
...
 
-def summarize_variable(name, var, col_width, marker=" ", max_width=None):
+def summarize_variable(
+    name: Hashable, var, col_width: int, marker: str = " ", max_width: int = None
+):
     """Summarize a variable in one line, e.g., for the Dataset.__repr__."""
     if max_width is None:
-        max_width = OPTIONS["display_width"]
+        max_width_options = OPTIONS["display_width"]
+        if not isinstance(max_width_options, int):
```
max-sixty: Is this a reasonable way to enforce types for objects coming from dicts? Or too much cruft and we should just use type hints?

dcherian:
LGTM. thanks

xarray/tests/test_formatting.py
```diff
@@ -134,11 +134,13 @@ def test_format_array_flat(self):
         expected = "0 1 2 ... 98 99"
         assert expected == actual
 
+        # NB: Probably not ideal; an alternative would be cutting after the
+        # first ellipsis
         actual = formatting.format_array_flat(np.arange(100.0), 11)
-        expected = "0.0 ... 99.0"
+        expected = "0.0 ... ..."
```
keewis: I'd cut after the first ellipsis, two of them in a row look strange to me.

max-sixty: Yeah, but what's the rule? Backtrack and see whether the previous values were `...`?

xarray/core/formatting.py
```diff
@@ -198,11 +199,20 @@ def format_array_flat(array, max_width):
...
+            padding,
+            " ".join(relevant_back_items[-num_back:]),
+        ]
     )
+
+    # As a final check, if it's still too long even with the limit in values,
+    # replace the end with an ellipsis
+    # NB: this will still returns a full 3-character ellipsis when max_width < 3
+    if len(pprint_str) > max_width:
+        pprint_str = pprint_str[: max(max_width - 3, 0)] + "..."
```
keewis:
```suggestion
        pprint_str = pprint_str[: max(max_width - 3, 0)] + padding
```

xarray/tests/test_formatting.py
keewis: something like that, yes:
```python
    if len(pprint_str) > max_width:
        pprint_str = pprint_str[: max(max_width - 3, 0)].rstrip()
        if not pprint_str.endswith("..."):
            pprint_str = pprint_str + "..."
```
Does that seem reasonable?

xarray/core/formatting.py
max-sixty: I can't seem to accept this, maybe because the PR is closed...

xarray/tests/test_formatting.py
max-sixty: Something like this is very reasonable, for sure. Just a tradeoff of complexity vs. results. 

For example, this case wouldn't change as a result of the function's change, because it ends with a space... We could add that logic in though

xarray/core/formatting.py
keewis: yes, we'd need a new PR for that.

xarray/tests/test_formatting.py
keewis: true, but we could also run `rstrip` first, truncate the string and run `rstrip` again (apologies if that's what you meant, I'm not really sure).

My suggestion might result in something like `783759.270 ... 592000...` but avoids `0.0 ... ...`. If we'd also like to avoid the former, we could not use `"".join()` immediately but instead save the iterable, check if the result of `"".join()` is short enough and remove the last element of the iterable if it isn't.

keewis: true, but we could also run `rstrip` first, truncate the string and run `rstrip` again (apologies if that's what you meant, I'm not really sure).

My suggestion might result in something like `783759.270 ... 592000...` but avoids `0.0 ... ...`. If we'd also like to avoid the former, we could not use `"".join()` immediately but instead save the iterable, check if the result of `"".join()` is short enough and remove the last element of the iterable if it isn't.

max-sixty: Yes 100%, I agree that's better. I'm probably a bit less motivated to make these stat as I feel like the egregious tails have been limited, which was the pressing issue. But can come back to it in the future if no one else takes it up.

PR #3904, opened 2020-03-27
Modify code in "Grouped arithmetic"

hoseinmadadi:
`results.append(group - alt.sel(x=label))` must be changed to:
`results.append(group - alt.sel(letters=label))`

### DIFF
--- a/doc/groupby.rst
+++ b/doc/groupby.rst
@@ -157,7 +157,7 @@ This last line is roughly equivalent to the following::
 
     results = []
     for label, group in ds.groupby('letters'):
-        results.append(group - alt.sel(x=label))
+        results.append(group - alt.sel(letters=label))
     xr.concat(results, dim='x')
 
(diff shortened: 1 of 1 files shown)


max-sixty:
Thanks @hoseinmadadi , good spot.

Want to add to `whatsnew` and give yourself credit?

hoseinmadadi:
@max-sixty, yes thanks, It will be my honor :)

max-sixty:
Great, ping when added and tests pass, thanks!

max-sixty:
@hoseinmadadi do you want to add a line to the whatsnew file?

shoyer:
Let's just merge. I would also encourage contributors to give themselves credit, but in this case that's almost more of a burden than the fix itself :)

@hoseinmadadi thanks for the fix here! if you want to add a credit to yourself in whatsnew in a later PR, feel free!

PR #3906, opened 2020-03-28
Fix for stack+groupby+apply w/ non-increasing coord

### LINKED ISSUE #3287: GroupBy of stacked dim with strings renames underlying dims
Names for dimensions are lost (renamed) when they are stacked and grouped, if one of the dimensions has string coordinates.

```python
data = np.zeros((2,1,1))
dims = ['c', 'y', 'x']

d1 = xr.DataArray(data, dims=dims)
g1 = d1.stack(f=['c', 'x']).groupby('f').first()
print('Expected dim names:')
print(g1.coords)
print()

d2 = xr.DataArray(data, dims=dims, coords={'c': ['R', 'G']})
g2 = d2.stack(f=['c', 'x']).groupby('f').first()
print('Unexpected dim names:')
print(g2.coords)
```

#### Output
It is expected the 'f_level_0' and 'f_level_1' be 'c' and 'x', respectively in the second part below.
```
Expected dim names:
Coordinates:
  * f        (f) MultiIndex
  - c        (f) int64 0 1
  - x        (f) int64 0 0

Unexpected dim names:
Coordinates:
  * f          (f) MultiIndex
  - f_level_0  (f) object 'G' 'R'
  - f_level_1  (f) int64 0 0
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.4 (default, Jul  9 2019, 18:13:23) 
[Clang 10.0.1 (clang-1001.0.46.4)]
python-bits: 64
OS: Darwin
OS-release: 18.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.2
libnetcdf: 4.6.3

xarray: 0.12.3
pandas: 0.25.1
numpy: 1.17.1
scipy: 1.3.1
netCDF4: 1.5.2
pydap: None
h5netcdf: None
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.3.4
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: None
distributed: None
matplotlib: 3.1.1
cartopy: None
seaborn: None
numbagg: None
setuptools: 41.2.0
pip: 19.2.3
conda: None
pytest: None
IPython: 7.8.0
sphinx: None
</details>

spencerahill:
- [x] Fixes #3287
 - [x] Tests added

I've added a check within `groupby.__init__` for monotonicity of the coords, which makes my test pass, but it causes three existing tests to fail.  Getting the logic right that properly distinguishes among these different cases is escaping me for now, but it does seem like this call to `unique_value_groups` within `groupby.__init__` is where the re-sorting is occurring.  Feedback welcome.  TIA!

### DIFF
--- a/xarray/core/groupby.py
+++ b/xarray/core/groupby.py
@@ -371,9 +371,7 @@ class GroupBy(SupportsArithmetic):
 
             # look through group to find the unique values
             group_as_index = safe_cast_to_index(group)
-            sort = bins is None and (
-                not isinstance(group_as_index, pd.MultiIndex)
-            )
+            sort = bins is None and (not isinstance(group_as_index, pd.MultiIndex))
             unique_values, group_indices = unique_value_groups(
                 group_as_index, sort=sort
             )
--- a/xarray/core/groupby.py
+++ b/xarray/core/groupby.py
@@ -370,10 +370,11 @@ class GroupBy(SupportsArithmetic):
                 group = group.dropna(group_dim)
 
             # look through group to find the unique values
-            all_sorted = all([safe_cast_to_index(c).is_monotonic_increasing
-                              for c in group.coords.values()])
+            sort = bins is None and (
+                not isinstance(safe_cast_to_index(group), pd.MultiIndex)
+            )
             unique_values, group_indices = unique_value_groups(
-                safe_cast_to_index(group), sort=(bins is None and all_sorted)
+                safe_cast_to_index(group), sort=sort
             )
             unique_coord = IndexVariable(group.name, unique_values)
 
--- a/xarray/core/groupby.py
+++ b/xarray/core/groupby.py
@@ -370,11 +370,12 @@ class GroupBy(SupportsArithmetic):
                 group = group.dropna(group_dim)
 
             # look through group to find the unique values
+            group_as_index = safe_cast_to_index(group)
             sort = bins is None and (
-                not isinstance(safe_cast_to_index(group), pd.MultiIndex)
+                not isinstance(group_as_index, pd.MultiIndex)
             )
             unique_values, group_indices = unique_value_groups(
-                safe_cast_to_index(group), sort=sort
+                group_as_index, sort=sort
             )
             unique_coord = IndexVariable(group.name, unique_values)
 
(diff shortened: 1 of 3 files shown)


spencerahill:
OK @max-sixty I've gotten logic that works in the sense that it works for this use case and doesn't break any existing tests.  But I could use a look over by somebody more familiar with groupby + multi-indexing, as it's totally possible that this quick fix causes other problems.

max-sixty:
This looks great @spencerahill ! Tests are very clear. The code fix looks good; I'll leave this open for a day in case those who know the code section better have any comments.

Thank you!

xarray/core/groupby.py
```diff
@@ -370,8 +370,11 @@ def __init__(
                 group = group.dropna(group_dim)
 
             # look through group to find the unique values
+            sort = bins is None and (
+                not isinstance(safe_cast_to_index(group), pd.MultiIndex)
```
max-sixty: One performance opt is to only do `safe_cast_to_index(group)` once; though I don't think it's huge

xarray/core/groupby.py
```diff
@@ -370,8 +370,11 @@ def __init__(
...
+            )
             unique_values, group_indices = unique_value_groups(
-                safe_cast_to_index(group), sort=(bins is None)
+                safe_cast_to_index(group), sort=sort
```
max-sixty: I'm actually not sure why we generally sort; lmk if you happened to have found the reason when testing...

spencerahill: So that was my first try, to simply remove the sorting, but then there were three test failures in `test_dataarray`, see e.g. [here](https://dev.azure.com/xarray/xarray/_build/results?buildId=2476&view=logs&j=ecaa9258-9d16-5560-78f9-8d2abe9f100f&t=df1f0ef6-5536-50dc-6554-aba88dd9b288&l=154).

They all involve grouping on 1D Indexes rather than MultiIndexes, which is what led me to the fix I settled on.  But that's as far as my understanding goes!

xarray/core/groupby.py
spencerahill: Sure, done in 4407f40.

spencerahill:
Great, thanks @max-sixty and happy to do it.

max-sixty:
Great, thanks again @spencerahill !

PR #3909, opened 2020-03-28
update the docstring of diff

### LINKED ISSUE #1040: DataArray.diff dim argument should be optional as is in docstring

keewis:
As per #1040, don't document the `dim` parameter of `Dataset.diff` and `DataArray.diff` as optional.

 - [x] Fixes #1040
 - [ ] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -46,6 +46,10 @@ Bug fixes
 
 Documentation
 ~~~~~~~~~~~~~
+- update the docstring of :py:method:`Dataset.diff` and
+  :py:method:`DataArray.diff` so it does document the ``dim``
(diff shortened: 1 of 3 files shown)


max-sixty:
LGTM, thanks @keewis

keewis:
Thanks, I'll merge on green.

PR #3913, opened 2020-03-29
Use divergent colormap if lowest and highest level span 0

### LINKED ISSUE #3524: xr.plot infers sequential colormap on diverging levels
#### MCVE Code Sample

```python
import numpy as np
import xarray as xr

data = np.random.randn(10, 10)
data = np.abs(data)

da = xr.DataArray(data)

# returns a diverging colormap
da.plot(vmax=2, center=0, extend="both")

# returns a sequential colormap
da.plot(levels=[-2, -1, 0, 1, 2], extend="both")



```

tst

#### Expected Output

A diverging colormap, maybe?

#### Problem Description
I was surprised by getting the viridis colormap until I realised that my data must all be positive and the colormap is infered from the data and not from levels. However, when specifying the range via `vmax=2, center=0` it is not inferred from the data.

<details>
INSTALLED VERSIONS
------------------
commit: None
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.12.14-lp151.28.25-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.2

xarray: 0.14.0+44.g4dce93f1
pandas: 0.25.2
numpy: 1.17.3
scipy: 1.3.1
netCDF4: 1.5.0.1
pydap: None
h5netcdf: 0.7.4
h5py: 2.9.0
Nio: None
zarr: None
cftime: 1.0.4.2
nc_time_axis: 1.2.0
PseudoNetCDF: None
rasterio: 1.0.22
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 2.6.0
distributed: 2.6.0
matplotlib: 3.1.1
cartopy: 0.17.0
seaborn: 0.9.0
numbagg: None
setuptools: 41.4.0
pip: 19.3.1
conda: None
pytest: 5.2.2
IPython: 7.9.0
sphinx: 2.2.1

</details>

dcherian:
- [x] Closes #3524
 - [x] Tests added

### DIFF
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -216,8 +216,13 @@ def _determine_cmap_params(
         vlim = abs(vmax - center)
 
     if possibly_divergent:
+        levels_are_divergent = (
+            isinstance(levels, Iterable) and levels[0] * levels[-1] < 0
+        )
         # kwargs not specific about divergent or not: infer defaults from data
-        divergent = ((vmin < 0) and (vmax > 0)) or not center_is_none
+        divergent = (
+            ((vmin < 0) and (vmax > 0)) or not center_is_none or levels_are_divergent
+        )
     else:
         divergent = False
 
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -169,6 +169,9 @@ def _determine_cmap_params(
     """
     import matplotlib as mpl
 
+    if levels is not None:
+        levels = sorted(levels)
+
     calc_data = np.ravel(plot_data[np.isfinite(plot_data)])
 
     # Handle all-NaN input data gracefully
--- a/xarray/plot/utils.py
+++ b/xarray/plot/utils.py
@@ -169,7 +169,7 @@ def _determine_cmap_params(
     """
     import matplotlib as mpl
 
-    if levels is not None:
+    if isinstance(levels, Iterable):
         levels = sorted(levels)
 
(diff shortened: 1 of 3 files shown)


xarray/plot/utils.py
```diff
@@ -216,8 +216,13 @@ def _determine_cmap_params(
         vlim = abs(vmax - center)
 
     if possibly_divergent:
+        levels_are_divergent = (
+            isinstance(levels, Iterable) and levels[0] * levels[-1] < 0
```
max-sixty: I don't know this well but do we know levels are monotonic or do we need to sort first?

dcherian: nice catch. matplotlib raises an error if `levels` is not sorted using this check: `np.min(np.diff(self.levels)) <= 0.0`. Right now, I'm doing `levels = sorted(levels)`. Should we raise an error instead?

max-sixty: No strong view, either is reasonable imo...

stefraynaud:
Two words.

It may be more appropriate to not only check that vmin and vmax are of opposite signe, but also one of them is not to close from zero, like [-0.01, 100].

This may also be extended to used positive and negative colomaps when vmin and vmax are of the same sign and one of them is close to zeros.

PR #3919, opened 2020-03-29
add a CI that tests xarray with all optional dependencies but dask

### LINKED ISSUE #3794: all-dependencies-but-dask CI
We recently got a few reports about tests failing with most optional dependencies installed but not `dask` (#3777, #3778, #3779).

Since `dask` is pretty tightly coupled in our code, I think it might be worth to add a CI job that runs the tests in a environment with all optional dependencies except from `dask`. Thoughts?

keewis:
Was using the `py38` environment the correct choice? I'm also not sure if adding a new CI was as simple as adding two lines, but that's something I will see once the CI runs :grin:

Edit: `dask`, `dask-core` and `distributed` get pulled in by other packages, so it seems there is a bit more investigation necessary.

 - [x] Closes #3794
 - [ ] Tests added


max-sixty:
Doesn't `py3.6-bare-minimum` cover this? Or you're thinking something that has more optional dependencies but not dask?

keewis:
yes, exactly. There were a few issues we did not detect because there are code paths that only activate if `dask` is not available at all. See #3794 for a list of these issues.

keewis:
seems `iris` was the culprit that pulled in `dask`. I removed it.

I also renamed the CI to `py38-all-but-dask`, but I'm not really confident in my naming abilities so if someone has a better idea I'd be happy to use that instead.

max-sixty:
Ha, the naming is good!

max-sixty:
This may be related: https://github.com/pydata/xarray/issues/1923

keewis:
it is related, but this PR has a different set of failures: there are a lot of `zarr` related failures, two `rasterio` and one `pseudonetcdf` failure and also a `sparse` chunking failure. The `sparse` error obviously needs to be decorated with `requires_dask`, but I don't know about the others.

keewis:
the non-serializable lock from the failing `rasterio` test is from https://github.com/pydata/xarray/blob/280a14ff5298c83fcee23e1e18e3a37397856ea9/xarray/backends/locks.py#L6-L10
which means the assumption that if we don't use `dask` we don't need to worry about serialization is incorrect (or the `rasterio` code uses the serializable lock in unintended ways)?

Also, the `chunks` parameter of `open_zarr` defaults to `"auto"`, meaning it tries to chunk by default, even if `dask` is not available. Should we check the availability of `dask` and overwrite `chunks="auto"` with `chunks=None` if we can't import `dask`? This would reduce the number of failing zarr tests from 98 to 2.

keewis:
a bit more progress: https://github.com/pydata/xarray/blob/fa37ee6d5a3c53f50dff61caebb342b89e2e35d5/xarray/backends/zarr.py#L58-L60
`_arrayize_vectorized_indexer` accepts an indexer, so we should pass `key` directly. This exposes a different error: plain `zarr` does not seem to support negative step sizes (`zarr+dask` seem fine with it, though).

keewis:
it seems the slices get transformed differently depending on the data: for numpy arrays, `slice(-1, 1, -1)` is transformed into `slice(9, 1, -1)` where for dask arrays `slice(-1, 1, -1)` becomes something without a negative step (not sure how that looks, though).

I'd xfail these tests since I think they're bugs and open issues for them.

keewis:
I don't think I can figure out how to fix the remaining three tests. Is it okay to have the fixes for `open_zarr` and `ZarrArrayWrapper.__getitem__` in this PR?

max-sixty:
Perfect, LGTM! Thank you @keewis

dcherian:
Thanks @keewis. Clearly this is a useful test config to add!

PR #3923, opened 2020-03-31
Add missing_dims argument allowing isel() to ignore missing dimensions

### LINKED ISSUE #3866: Allow `isel` to ignore missing dimensions?
Sometimes it would be nice for `isel()` to be able to ignore a dimension if it is missing in the Dataset/DataArray. E.g.
```
ds = Dataset()

ds.isel(t=0) # currently raises an exception

ds.isel(t=0, ignore_missing=True) # would be nice if this was allowed, just returning ds
```
For example, when writing a function can be called on variables with different combinations of dimensions.

I think it should be fairly easy to implement, just add the argument to the condition here
https://github.com/pydata/xarray/blob/65a5bff79479c4b56d6f733236fe544b7f4120a8/xarray/core/variable.py#L1059-L1062
the only downside would be increased complexity of adding another argument to the API for an issue where a workaround is not hard (at least in the case I have at the moment), just a bit clumsy.

johnomotani:
Note: only added to `DataArray.isel()` and `Variable.isel()`. A `Dataset` should include all dimensions, so presumably it should always be an error to pass a non-existent dimension when slicing a `Dataset`?

 - [x] Closes #3866
 - [x] Tests added

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -35,10 +35,10 @@ New Features
   :py:func:`combine_by_coords` and :py:func:`combine_nested` using
   combine_attrs keyword argument. (:issue:`3865`, :pull:`3877`)
   By `John Omotani <https://github.com/johnomotani>`_
-- 'missing_dims' argument to :py:meth:`DataArray.isel` and
-  :py:meth:`Variable.isel` to allow replacing the exception when a dimension
-  passed to ``isel`` is not present with a warning, or just ignore the
-  dimension. (:issue:`3866`, :pull:`3923`)
+- 'missing_dims' argument to :py:meth:`Dataset.isel`,
+  `:py:meth:`DataArray.isel` and :py:meth:`Variable.isel` to allow replacing
+  the exception when a dimension passed to ``isel`` is not present with a
+  warning, or just ignore the dimension. (:issue:`3866`, :pull:`3923`)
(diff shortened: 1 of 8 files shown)


max-sixty:
Thanks @johnomotani ; appreciate the PR.

I'm not strongly opposed to this. I haven't personally had an instance where I've wanted this keyword, but that's not overly representative of others' cases. 
What do people think, @pydata/xarray ?

Separately, if we do merge, I think we should align with the keywords on `drop_vars`: `raise` / `ignore`

dcherian:
I don't see why `Dataset.isel` is different. 

+1 for using the `drop_vars` kwarg

johnomotani:
The Linux-py38-upstream-dev test is failing in cftime tests at various places similar to this
```
        with create_tmp_file() as tmp_file:
            original.to_netcdf(tmp_file)
            with pytest.warns(None) as record:
                with open_dataset(tmp_file, use_cftime=False) as ds:
                    assert_identical(expected_x, ds.x)
                    assert_identical(expected_time, ds.time)
>               assert not record
E               assert not WarningsChecker(record=True)
```
Don't think it's related to this PR...

keewis:
you're right, that's #3928

max-sixty:
Great, thanks a lot @johnomotani !

Issue #934, opened 2016-08-02
Should indexing be possible on 1D coords, even if not dims?


max-sixty:
``` python
In [1]: arr = xr.DataArray(np.random.rand(4, 3),
    ...:    ...:                    [('time', pd.date_range('2000-01-01', periods=4)),
    ...:    ...:                     ('space', ['IA', 'IL', 'IN'])])
    ...:    ...: 

In [17]: arr.coords['space2'] = ('space', ['A','B','C'])

In [18]: arr
Out[18]: 
<xarray.DataArray (time: 4, space: 3)>
array([[ 0.05187049,  0.04743067,  0.90329666],
       [ 0.59482538,  0.71014366,  0.86588207],
       [ 0.51893157,  0.49442107,  0.10697737],
       [ 0.16068189,  0.60756757,  0.31935279]])
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 2000-01-04
  * space    (space) |S2 'IA' 'IL' 'IN'
    space2   (space) |S1 'A' 'B' 'C'
```

Now try to select on the space2 coord:

``` python
In [19]: arr.sel(space2='A')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-19-eae5e4b64758> in <module>()
----> 1 arr.sel(space2='A')

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/xarray/core/dataarray.pyc in sel(self, method, tolerance, **indexers)
    601         """
    602         return self.isel(**indexing.remap_label_indexers(
--> 603             self, indexers, method=method, tolerance=tolerance))
    604 
    605     def isel_points(self, dim='points', **indexers):

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/xarray/core/dataarray.pyc in isel(self, **indexers)
    588         DataArray.sel
    589         """
--> 590         ds = self._to_temp_dataset().isel(**indexers)
    591         return self._from_temp_dataset(ds)
    592 

/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/xarray/core/dataset.pyc in isel(self, **indexers)
    908         invalid = [k for k in indexers if k not in self.dims]
    909         if invalid:
--> 910             raise ValueError("dimensions %r do not exist" % invalid)
    911 
    912         # all indexers should be int, slice or np.ndarrays

[...]

Issue #3774, opened 2020-02-17
Allow xr.combine_by_coords to work with point coordinates?


kazimuth:
#### MCVE Code Sample

```python
# Your code here

data_0 = xr.Dataset({'temperature': ('time', [10,20,30])}, coords={'time': [0,1,2]})
data_0.coords['trial'] = 0

data_1 = xr.Dataset({'temperature': ('time', [50,60,70])}, coords={'time': [0,1,2]})
data_1.coords['trial'] = 1

# the following fails with a ValueError:
# all_trials = xr.combine_by_coords([data_0, data_1])

# but this works:
all_trials = xr.combine_by_coords([data_0.expand_dims('trial'), data_1.expand_dims('trial')]) # works
```
#### Expected output
Both work and produce the same result.

#### Problem Description

I found this behavior somewhat unintuitive -- I didn't understand that I needed to call `expand_dims` to get `combine_by_coords` to work. It might make things a little easier to understand if both behaved the same way -- maybe the code that concatenates dimensions should look for full dimensions first, and point coordinates second.

This would also make code that selects dimensions with `sel` and then calls `combine_by_coords` more easy to write.

On the other hand, this could be seen as "magic", and it might also be a breaking change to `combine_by_coords`; I'm not sure. So I'd understand if isn't implemented.

#### Output of ``xr.show_versions()``
<details>
<pre>
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.6 (default, Jan  8 2020, 19:13:59) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.14.0-115.10.1.el7a.ppc64le
machine: ppc64le
processor: ppc64le
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.2
libnetcdf: None

xarray: 0.15.0
pandas: 0.25.3
numpy: 1.16.5
scipy: 1.3.0
netCDF4: None
pydap: None
h5netcdf: None
h5py: 2.8.0
Nio: None
zarr: None
cftime: None
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.10.0
distributed: 2.10.0
matplotlib: 3.1.1
cartopy: None
seaborn: None
numbagg: None
setuptools: 44.0.0.post20200106
pip: 19.3.1
conda: None
pytest: 5.3.2
[...]

PR #3925, opened 2020-04-02
sel along 1D non-index coordinates

### LINKED ISSUE #2028: slice using non-index coordinates
It should be relatively straightforward to allow slicing on coordinates that are not backed by an IndexVariable, or in other words coordinates that are on a dimension with a different name, as long as they are 1-dimensional (unsure about the multidimensional case).

E.g. given this array:
```
a = xarray.DataArray(
    [10, 20, 30],
    dims=['country'],
    coords={
        'country': ['US', 'Germany', 'France'],
        'currency': ('country', ['USD', 'EUR', 'EUR'])
    })

<xarray.DataArray (country: 3)>
array([10, 20, 30])
Coordinates:
  * country   (country) <U7 'US' 'Germany' 'France'
    currency  (country) <U3 'USD' 'EUR' 'EUR'
```

This is currently not possible:
```
a.sel(currency='EUR')


ValueError: dimensions or multi-index levels ['currency'] do not exist
```

It should be interpreted as a shorthand for:
```
a.sel(country=a.currency == 'EUR')

<xarray.DataArray (country: 2)>
array([20, 30])
Coordinates:
  * country   (country) <U7 'Germany' 'France'
    currency  (country) <U3 'EUR' 'EUR'
```

TomNicholas:
As a user, I find not being able to select along one-dimensional non-dimensional coordinates actually comes up fairly often. I think it's quite common to use multiple coordinates to be able to choose between plotting in different coordinate systems (or units) easily.

I've tried to close #2028 in the simplest (but also least efficient) way which was suggested by @shoyer ([suggestion 1 here](https://github.com/pydata/xarray/issues/934#issuecomment-236960237)).

This should be temporary anyway: it will get superseded by the [explicit indexes refactor](https://github.com/pydata/xarray/issues/1603). If there is another approach which would achieve the same functionality as this PR but actually bring us closer to #1603 then I would be happy to take a stab at that instead.

I don't really know what to do about the [failing test](https://github.com/pydata/xarray/blob/b3bafeefbd6e6d70bce505ae1f0d9d5a2b015089/xarray/tests/test_dataset.py#L3632) in groupby arithmetic - I think it's [caused here](https://github.com/pydata/xarray/blob/b3bafeefbd6e6d70bce505ae1f0d9d5a2b015089/xarray/core/groupby.py#L497) but I'm not sure what to replace the triple error type catching (?!) with.

 - [x] Closes #2028
 - [x] Tests added

### DIFF
--- a/xarray/core/indexing.py
+++ b/xarray/core/indexing.py
@@ -205,27 +205,35 @@ def convert_label_indexer(index, label, index_name="", method=None, tolerance=No
 
 def get_dim_indexers(data_obj, indexers):
     """Given a xarray data object and label based indexers, return a mapping
-    of label indexers with only dimension names as keys.
+    of label indexers with only dimension names (or 1D non-dimensional coord
+    names) as keys.
 
     It groups multiple level indexers given on a multi-index dimension
     into a single, dictionary indexer for that dimension (Raise a ValueError
     if it is not possible).
     """
+    non_dim_1d_coords = [coord for coord in data_obj.coords
+                         if len(data_obj[coord].dims) == 1]
     invalid = [
-        k
-        for k in indexers
-        if k not in data_obj.dims and k not in data_obj._level_coords
+        k for k in indexers
+           if k not in data_obj.dims and k not in data_obj._level_coords
+             and k not in non_dim_1d_coords
     ]
-    if invalid:
-        raise ValueError(f"dimensions or multi-index levels {invalid!r} do not exist")
+    if invalid:  # TODO This was never covered by testing
+         raise ValueError(f"dimensions, 1D coordinates, or  multi-index levels"
+                          f" {invalid!r} do not exist")
 
     level_indexers = defaultdict(dict)
     dim_indexers = {}
     for key, label in indexers.items():
         (dim,) = data_obj[key].dims
         if key != dim:
-            # assume here multi-index level indexer
-            level_indexers[dim][key] = label
+            # If key is 1D non-dimension coordinate let it pass through
+            if key in data_obj.coords and data_obj.coords[key].dims == (dim,):
+                dim_indexers[key] = label
+            else:
+                # assume here multi-index level indexer
+                level_indexers[dim][key] = label
         else:
             dim_indexers[key] = label
 
@@ -253,9 +261,31 @@ def remap_label_indexers(data_obj, indexers, method=None, tolerance=None):
 
     dim_indexers = get_dim_indexers(data_obj, indexers)
     for dim, label in dim_indexers.items():
-        try:
+        if dim in data_obj.indexes:
             index = data_obj.indexes[dim]
-        except KeyError:
+
+            coords_dtype = data_obj.coords[dim].dtype
+            label = maybe_cast_to_coords_dtype(label, coords_dtype)
+            idxr, new_idx = convert_label_indexer(index, label, dim,
+                                                  method, tolerance)
+            pos_indexers[dim] = idxr
+            if new_idx is not None:
+                new_indexes[dim] = new_idx
+
+        elif dim in data_obj.coords and len(data_obj.coords[dim] == 1):
+            # 1D non-dimension coord
+            index = data_obj.coords[dim].to_index()
+            (dim,) = data_obj[dim].dims
+
+            coords_dtype = data_obj.coords[dim].dtype
+            label = maybe_cast_to_coords_dtype(label, coords_dtype)
+            idxr, new_idx = convert_label_indexer(index, label, dim,
+                                                  method, tolerance)
+            pos_indexers[dim] = idxr
+            if new_idx is not None:
+                new_indexes[dim] = new_idx
+
+        else:
             # no index for this dimension: reuse the provided labels
             if method is not None or tolerance is not None:
                 raise ValueError(
@@ -264,13 +294,6 @@ def remap_label_indexers(data_obj, indexers, method=None, tolerance=None):
                     "an associated coordinate."
                 )
             pos_indexers[dim] = label
-        else:
-            coords_dtype = data_obj.coords[dim].dtype
-            label = maybe_cast_to_coords_dtype(label, coords_dtype)
-            idxr, new_idx = convert_label_indexer(index, label, dim, method, tolerance)
-            pos_indexers[dim] = idxr
-            if new_idx is not None:
-                new_indexes[dim] = new_idx
 
     return pos_indexers, new_indexes
 
--- a/xarray/tests/test_dataset.py
+++ b/xarray/tests/test_dataset.py
@@ -56,7 +56,7 @@ except ImportError:
     pass
 
 
-def create_test_data(seed=None):
+def create_test_data(seed=None, dim_coords=True):
     rs = np.random.RandomState(seed)
     _vars = {
         "var1": ["dim1", "dim2"],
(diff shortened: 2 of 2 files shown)


max-sixty:
I think the functionality is useful, but I'm concerned it would make the API more confusing. In particular, one maxim we've generally held is that changing the number of dimensions doesn't change functionality. (ofc it changes whether it's possible to _create_ an index, though the indexing API doesn't change)

What are others' thoughts?

TomNicholas:
> In particular, one maxim we've generally held is that changing the number of dimensions doesn't change functionality. 

Sorry, could you explain what you mean here? How would this PR violate that?

max-sixty:
> Sorry, could you explain what you mean here? How would this PR violate that?

Of course. I mean that if someone changes the dimensionality of a non-index coord from 1D to 2D, then running `.sel` over it will stop working. 

Whereas now, neither would work. Which in some ways is worse, but also less surprising. I think these are always difficult trade-offs...

TomNicholas:
Gotcha, thanks.

Hmm, this seems like a grey area... I think selecting along 1D non-dimension coords is probably way more common than along 2D coords, but I'm biased.

> changing the number of dimensions doesn't change functionality

BTW I can immediately think of at least one other place where this rule is broken (#3774) - where going from 0D coords to 1D coords of length 1 changes whether `combine_by_coords` will accept them.

xarray/core/indexing.py
```diff
@@ -253,9 +261,31 @@ def remap_label_indexers(data_obj, indexers, method=None, tolerance=None):
...
+
+            coords_dtype = data_obj.coords[dim].dtype
+            label = maybe_cast_to_coords_dtype(label, coords_dtype)
+            idxr, new_idx = convert_label_indexer(index, label, dim,
+                                                  method, tolerance)
+            pos_indexers[dim] = idxr
+            if new_idx is not None:
+                new_indexes[dim] = new_idx
+
+        elif dim in data_obj.coords and len(data_obj.coords[dim] == 1):
```
shoyer, line 275: I think this should be `data_obj.coords[dim].ndim == 1` instead of `len(data_obj.coords[dim] == 1)`? The later is checking that the coordinate's first dimension has length 1.

xarray/core/indexing.py
```diff
@@ -253,9 +261,31 @@ def remap_label_indexers(data_obj, indexers, method=None, tolerance=None):
...
+            index = data_obj.coords[dim].to_index()
+            (dim,) = data_obj[dim].dims
+
+            coords_dtype = data_obj.coords[dim].dtype
+            label = maybe_cast_to_coords_dtype(label, coords_dtype)
+            idxr, new_idx = convert_label_indexer(index, label, dim,
+                                                  method, tolerance)
+            pos_indexers[dim] = idxr
+            if new_idx is not None:
+                new_indexes[dim] = new_idx
```
shoyer, line 286: Can you refactor so this code block doesn't need to be repeated?

Every index is a 1D coordinate, so I think you could do this by switching around the order of these `if` checks, e.g.,
```python
    if dim in data_obj.coords and data_obj.coords[dim].ndim == 1:
        if dim in data_obj.indexes:
            index = data_obj.indexes[dim]
        else:
            index = data_obj.coords[dim].to_index()
            (dim,) = data_obj[dim].dims

        coords_dtype = data_obj.coords[dim].dtype
        label = maybe_cast_to_coords_dtype(label, coords_dtype)
        idxr, new_idx = convert_label_indexer(index, label, dim,
                                              method, tolerance)
        pos_indexers[dim] = idxr
        if new_idx is not None:
            new_indexes[dim] = new_idx
```

shoyer:
I think this is generally a good idea!

In the future, creating an `index` explicit would just mean:
1. Repeated lookups are more efficient, due to caching in a hash-table.
2. The coordinate values are immutable, to ensure that the index values can be cached.

One minor concern I have here is about efficiency: building an `pd.Index` and its hash-table from scratch can be quite expensive. If we're only doing a single lookup this is fine, but if it's being done in a loop this could be surprisingly slow, and we would do far better sticking with pure NumPy operations.

Here's an microbenchmark that hopefully illustrates the issue:
```
import pandas as pd
import numpy as np

def lookup_preindexed(needle, index):
  return index.get_loc(needle)

def lookup_newindex(needle, haystack):
  return lookup_preindexed(needle, pd.Index(haystack))

def lookup_numpy(needle, haystack):
  return (haystack == needle).argmax()

haystack = np.random.permutation(np.arange(1000000))
index = pd.Index(haystack)
%timeit lookup_newindex(0, haystack)  # 56.1 ms per loop
%timeit lookup_preindexed(0, index)  # 696 ns per loop
%timeit lookup_numpy(0, haystack)  # 517 µs per loop
```

pandas is 1000x faster than NumPy if the index is pre-existing, but 100x slower if the index is new. That's a 1e5 fold slow-down!

I think users will appreciate the flexibility, but if there's some way we warn users that they really should set the index ahead of time when they are doing repeating indexing that could also be welcome. Figuring out how to save the state for counting the number of times a new index is created could be pretty messy, though. I guess we could stuff it into `Variable.encoding` and issue a warning whenever the same variable has been converted into an index at least 100 times.

shoyer:
Related to my microbenchmark, it might also be worth considering pure NumPy versions of common indexing operations, to avoid the need to repeatedly create hash-tables. But that could be quite a bit of work to do comprehensively.

max-sixty:
OK! May be a sizeable change but I update on @shoyer 's view, let's do it.

benbovy:
> pandas is 1000x faster than NumPy if the index is pre-existing, but 100x slower if the index is new. That's a 1e5 fold slow-down!

> I think users will appreciate the flexibility, but if there's some way we warn users that they really should set the index ahead of time when they are doing repeating indexing that could also be welcome.

I think it's a good use case for some kind of `EphemeralIndex` (or `BasicIndex` or `NumpyIndex`) once the explicit index refactoring is done, along with some good documentation on which index to choose for which purpose.

benbovy:
In #5692 it is possible to perform selection using non-dimension coordinates with an index, although there's no easy way yet to set an index for such coordinates (this will be done in a follow-up PR by updating the API of `set_index`).

TomNicholas:
FYI I probably am not going to work on this PR again - especially as I remember getting quite confused by the indexing internals, which have now changed. If we can just wait for this feature to be enabled via the indexing refactor then I would rather just do that.

ivanakcheurov:
> In #5692 it is possible to perform selection using non-dimension coordinates with an index, although there's no easy way yet to set an index for such coordinates (this will be done in a follow-up PR by updating the API of `set_index`).

@benbovy, please could you give an example how it is possible? 
I would like `sel based on a non-dim coordinate` to be as fast as `sel based on the dim itself` as per the following timings:
```python
# sel based on a non-dim coordinate 
# (using this coordinate directly .sel(product_id=26) would result in error "'no index found for coordinate product_id")
%timeit xds.sel(product=xds.product_id==26)
1.54 ms ± 64.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# sel based on the dim itself
%timeit xds.sel(product='GN91 Glove Medium')
499 µs ± 16.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit xds.where(xds.product_id==26, drop=True)
4.17 ms ± 39 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

mathause:
With #6971 on the way I guess we can close this.

Issue #3830, opened 2020-03-05
Add examples for carrying out "ncecat" to xarray documentation


lukelbd:
In climate science, a very common task involves concatenating NetCDF files with identical variables, dimensions, and coordinates along a brand new "ensemble member" or "record" dimension. With the NetCDF Operators, this is accomplished using [`ncecat`](http://nco.sourceforge.net/nco.html#ncecat-netCDF-Ensemble-Concatenator
). 

#### MCVE Code Sample

Currently, it seems the correct way to do this in xarray is with [`xarray.combine_nested`](http://xarray.pydata.org/en/stable/generated/xarray.combine_nested.html) as follows:

```python
import xarray as xr
files = ['member1.nc', 'member2.nc', ...]
ds = xr.open_mfdataset(
    files,
    combine='nested',
    concat_dim='record',
)
```

#### Problem Description

While this works, there does not seem to be any mention of this approach in the [`combine_nested`](http://xarray.pydata.org/en/stable/generated/xarray.combine_nested.html) or [`open_mfdataset`](http://xarray.pydata.org/en/stable/generated/xarray.combine_nested.html) docs... and using `combine_nested` as it is currently described to concatenate along a brand new dimension feels quite unintuitive to me.

It would be nice if you guys could add examples to `combine_nested` and/or `open_mfdataset` with this special usage and/or mention the possibility of creating *brand new* dimensions with `concat_dim`. For example:

```python
In [1]: import xarray as xr 
   ...: datasets = []
   ...: for i in range(3):
   ...:     dataset = xr.Dataset({'temp': (('x', 'y'), np.random.rand(10, 20)
   ...:     datasets.append(dataset)
   ...: dataset = xr.combine_nested(datasets, concat_dim='record')
   ...: dataset
Out[1]:
<xarray.Dataset>
Dimensions:  (record: 3, x: 10, y: 20)
Dimensions without coordinates: record, x, y
Data variables:
    temp     (record, x, y) float64 0.32 0.4897 0.2659 ... 0.3485 0.0251 0.399
```

#### Output of ``xr.show_versions()``

n/a

TomNicholas:
Thanks @lukelbd for pointing this, this is certainly a common operation, and there should be a clear example in the docs. 

The method you suggest (`open_mfdataset(files, combine='nested', concat_dim='record')`) is the intended way to do it.

[...]

PR #3926, opened 2020-04-02
Remove old auto combine

TomNicholas:
- [x] Finishes deprecation cycle started in #2616 (was supposed to have been done in 0.15)

I've set `combine='by_coords'` as the default argument to `open_mfdataset`. Technically we could go for either, as the deprecation warning just told users to make it explicit from now on, but going for `by_coords` rather than `nested` means that:
- The `concat_dim` argument is not needed by default,
- The default behaviour of the function is the "magic" one - users have to opt-in to the more explicit behaviour.


TomNicholas:
It would also be good to include any improvement to the relevant docs (e.g. #3830) in the same release.

max-sixty:
I think `FutureWarning` and `DeprecationWarning` are pretty similar — mostly focused on whether the behavior will change or the API is going away. 
`FutureDeprecationWarning` is generally lower visibility IIRC.

So maybe this should have been `DeprecationWarning` but no need to go through another cycle imo.

dcherian:
There was a case I ran in to where the old `auto_combine` was needed and none of the newer options worked.

Let me see if I can dig it up.

TomNicholas:
> To make the docs CI pass

Thanks @keewis .

> Shouldn't we change the FutureWarning to a DeprecationWarning before removing auto_combine? 

I wasn't even aware that there were two types of warnings I could have used for this. I was taking "deprecating" to mean "removed" and "starting the deprecation cycle" to mean "start warning people it will soon be removed". The warning messages said "will no longer accept" and "will require" in "version 0.15" - that's probably authoritative enough isn't it?

> There was a case I ran in to where the old auto_combine was needed and none of the newer options worked.

I think that was always a possible edge case - I would very much like to see it though, if only so that we can put a "what to do if" section in the docs.

keewis:
> The warning messages said "will no longer accept" and "will require" in "version 0.15" - that's probably authoritative enough isn't it?

I'd think so, yes. Rereading what the warnings are intended for, [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) is indeed the same as [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning), the only difference is the intended target audience. I seem to have confused `FutureWarning` with `PendingDeprecationWarning`.

Then my only issue is with the `whats-new.rst` entry. I'd explicitly state that `auto_combine` was removed (but maybe that's what "fully deprecated" means?).

keewis:
thanks, @TomNicholas, looks good to me.

I opened #3928 for the `upstream-dev` failures.

TomNicholas:
Just noticed there's a [section](http://xarray.pydata.org/en/stable/api.html#deprecated-pending-deprecation) of the API reference for `Deprecated/Pending Deprecation`, but `auto_combine` was never added to it...

Issue #3917, opened 2020-03-29
running numpy functions on xarray objects


keewis:
In the `pint` integration tests I tried to also test calling numpy functions on xarray objects (we provide methods for all of them).

Some of these functions, like `numpy.median`, `numpy.searchsorted` and `numpy.clip`, depend on `__array_function__` (i.e. not `__array_ufunc__`) to dispatch. However, neither `Dataset` nor `DataArray` (nor `Variable`, I think?) define these protocols (see #3643).

Should we define `__array_function__` on xarray objects?

shoyer:
I think this is probably a good idea, assuming we figure out the type casting hierarchy challenge. Right now people often check for the existence of `__array_function__`, which clearly isn't quite right.

Issue #3160, opened 2019-07-24
xarray: How to get the coordinates of the maximum and minimum value over a spatial domain in xarray?


anubhavchoudhary:
I have a netcdf file containing a surface variable over a latlon area. I want to get the coordinate (as lat, lon value) of maximum as well as minimum value of that variable over a certain area which I can define with latlon box. How to do it in xarray ?

TomNicholas:
There might be a better way, but I think this is one way:

First mask out all the data that isn't in your lat/long box. If your latitude and longitude are 1D then you could use `.sel()`, otherwise use `.where()`. You have two conditions (one for latitude and one for longitude), so combine them with `np.logical_and()`:

```python
condition = np.logical_and(data.lat > 10.0, data.lon < 50.0)
box = data.where(condition)
```

Then use `DataArray.argmax()` across all the dimensions of the result to find the indices of the maximum values. 

```python
inds_of_max = box.argmax(dims=['lat', 'lon'])
```

Pass those indices to your latitude/longitude coordinate arrays to get the (lat, lon) pair you want.

```python
lat, lon = data.lat[inds_of_max], data.lon[inds_of_max]
```

(This would probably be better placed on stackoverflow than here but doesn't matter)

shoyer:
In fact, this was already asked on StackOverflow :)
https://stackoverflow.com/questions/40179593/how-to-get-the-coordinates-of-the-maximum-in-xarray

I'm going to close this, but feel free to comment or reopen if you have any further questions.

JohnWatsonFd:
I don't think the answer in stackoverflow is the best idea for all conditions. Factly, it can only be used in xarray.dataarray not suitble for xarray.dataset.

PR #3922, opened 2020-03-31
FIX: correct dask array handling in _calc_idxminmax

kmuehlbauer:
Fixes dask handling for implementation in #3871.

 - [ ] Closes #xxxx
 - [x] Tests added
 - [x] Passes `isort -rc . && black . && mypy . && flake8`
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

PR #3936, opened 2020-04-05
Support multiple dimensions in DataArray.argmin() and DataArray.argmax() methods

johnomotani:
These return dicts of the indices of the minimum or maximum of a DataArray over several dimensions. Inspired by @fujiisoup's work in #1469. With #3871, replaces #1469. Provides a simpler solution to #3160.

Implemented so that
```
da.isel(da.argmin(list_of_dim)) == da.min(list_of_dim)
da.isel(da.argmax(list_of_dim)) == da.max(list_of_dim)
```

 - [x] Closes #1469
 - [x] Tests added


shoyer:
This looks really comprehensive, thank you!

Before doing a really careful review here, I'd like to try to work out the full API design we want. I'll write out some of my thoughts here, but your thoughts would also be very welcome!

Here's my summary of the current situation:
1. We have two pairs of methods, `argmin`/`argmax` and the new `idxmin`/`idxmax` for calculating integer positions or coordinate labels of the minimum/maximum along _one_ dimension.
2. We would like to have a way to do a similar calculation in multiple dimensions, which should return a dictionary of DataArray objects suitable for feeding into `isel`/`sel`.

This PR implements the multidimensional equivalent of `argmin`/`argmax`. It's fine to save the multidimensional equivalent of `idxmin`/`idxmax` for later, but we'll want to copy whatever design we pick here.

My first concern is about the name: it isn't obvious to me whether `indices_min` refers to integer positions or coordinate labels. I think a name based on `argmin`/`argmax` could help, e.g., perhaps `argmin_nd` or `argmin_dict`.

Another option would be to overload `argmin`/`argmax` to _also_ support return a dictionary, depending on the type of the arguments. If you pass a `list` of dimensions over which to calculate the min or max locations, you would get a dictionary back:
- `array.argmin('x')` would return a single `DataArray`
- `array.argmin(['x', 'y'])` would return a dict of `DataArray` objects with keys `{'x', 'y'}`
- `array.argmin(['x'])` would also return a dict `{'x': ...}`

I think I like this last option best but I would be curious what others think! @pydata/xarray any thoughts on this?

johnomotani:
@shoyer I think your last option sounds good. Questions:
* What should `da.argmin()` with no arguments do?
    * Currently returns the flattened index of the global minimum.
    * I think returning a `dict` of indices would be much more useful, but it does change existing behaviour (more useful because you can then do `da.isel(da.argmin())`).
    * Could anyway do `da.argmin(da.dims)` to get the `dict` result, but that's a bit ugly
    * Could have something like `da.argmin(...)` - maybe as a temporary workaround while we deprecate the current behaviour of `da.argmin()`?
* If we overload `argmin`, what's the cleanest way to get the existing `argmin()` method within the new one? Would something like `from .duck_array_ops import argmin as base_argmin` work?

johnomotani:
Maybe worth noting, at the moment if you try to call `argmin(dim=("x", "y"))` with multiple dimensions, there's a not-very-helpful exception
```
>>> array.argmin(dim=("x", "y"))
  File "/home/john/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 61, in _wrapfunc
    return bound(*args, **kwds)
TypeError: 'tuple' object cannot be interpreted as an integer

  File "/<...>/xarray/xarray/core/common.py", line 46, in wrapped_func
    return self.reduce(func, dim, axis, skipna=skipna, **kwargs)
  File "/<...>/xarray/xarray/core/dataarray.py", line 2288, in reduce
    var = self.variable.reduce(func, dim, axis, keep_attrs, keepdims, **kwargs)
  File "/<...>/xarray/xarray/core/variable.py", line 1579, in reduce
    data = func(input_data, axis=axis, **kwargs)
  File "/<...>/xarray/xarray/core/duck_array_ops.py", line 304, in f
    return func(values, axis=axis, **kwargs)
  File "/<...>/xarray/xarray/core/duck_array_ops.py", line 47, in f
    return wrapped(*args, **kwargs)
  File "<__array_function__ internals>", line 6, in argmin
  File "/<...>/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 1267, in argmin
    return _wrapfunc(a, 'argmin', axis=axis, out=out)
  File "/<...>/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 70, in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
  File "/<...>/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 47, in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
TypeError: 'tuple' object cannot be interpreted as an integer
```

TomNicholas:
> Another option would be to overload `argmin`

+1 for overloading `argmin` (and later `idxmin`). IMO we should never have one function for a 1D operation and one for an N-D operation if we can avoid it, everything should be N-dimensional.

I also really like how neat this resultant property is
```python
da.isel(da.argmin(list_of_dim)) == da.min(list_of_dim)
```
we could even use a [hypothesis](https://github.com/HypothesisWorks/hypothesis) test to check it...

> I think returning a dict of indices would be much more useful, but it does change existing behaviour (more useful because you can then do da.isel(da.argmin())).

Although it's breaking and would require a deprecation cycle, I think this is what we should aim for.

> there's a not-very-helpful exception

Yes let's take the time to make that clearer for users - this will be a commonly-used function.

kmuehlbauer:
+1 for making `argmin`/`argmax` (and `idxmin`/idxmax`) work, when given multiple dimensions.

According to the current docstring it should already work that way for `argmin`/'argmax`:

**Reduce this DataArray’s data by applying argmin along some dimension(s).**

**Returns: New DataArray/Dataset object with `argmin` applied to it's data and the indicated dimension(s) removed**

But this behaviour is broken currently (works only for one given dim).

My main concern for changing the API as suggested above is, how should we discern (at least for `argmin`/`argmax`), if the user want's to:

1. reduce **over all** given dimensions? (current behaviour according to docstring)
2. reduce **along all** given dimensions? (suggested behaviour by @johnomotani [please correct if I'm wrong])

TomNicholas:
> how should we discern (at least for argmin/argmax), if the user wants to...

That's a good question @kmuehlbauer, and the distinction probably needs to be clearer in the docs in general.

> reduce over all given dimensions?

By this do you mean find the minimum as if the array were first (partially or totally) flattened along the given dims somehow? I'm not sure we provide that kind of behaviour anywhere in the current API.

kmuehlbauer:
> > reduce over all given dimensions?
> 
> By this do you mean find the minimum as if the array were first (partially or totally) flattened along the given dims somehow? I'm not sure we provide that kind of behaviour anywhere in the current API.

@TomNicholas Probably I was bit confused by the current docstring. I think I understand now and there should be no problem at all.

kmuehlbauer:
@johnomotani FYI: For #3871 (merged) there is #3922 (yet unmerged) to fix dask-handling.

johnomotani:
I've updated so the new functionality is provided by `argmin()` and `argmax()`, when they are passed a multiple dimensions.
* I've added `Dataset.argmin()` and `Dataset.argmax()`, which only work for a single dimension, but give informative exceptions for unsupported cases. If anyone has some behaviour that they would like for `Dataset.argmin()` with multiple dimensions, I'm happy to implement it, but returning something like a `dict` of `Dataset`s didn't seem nice, and as far as I understand a `Dataset` cannot contain something like a `dict` of `DataArray`s.
* I renamed `argmin` and `argmax` in `ops.py` and `duck_array_ops.py` to `_argmin_base` and `_argmax_base` so that I could still use them inside the `argmin` and `argmax` methods. Any alternatives, or better suggestions for naming them?
* `Variable` no longer has `argmin` or `argmax` methods. Is that a problem? I just updated `test_variable.py` and `test_dask.py` to use `_argmin_base` and `_argmax_base`.

johnomotani:
The rename to `_argmin_base` is also causing `test_aggregation` (which is a test of units with `pint`) to fail, because the numpy function and the `Variable`  method do not have the same name. Any suggestions how to fix this? It seems tricky because I can't pass either `argmin` (no `Variable` method) or `_argmin_base` (no numpy method) to the test.

johnomotani:
These test failures seem to have uncovered a larger issue: overriding a method injected by `ops.inject_all_ops_and_reduce_methods(DataArray, priority=60)` is tricky. I think there may be a way around it by mangling the name of the injected operation if that method is already defined on the class; I tried this and it nearly works, but I think I need to implement an override for argmin/argmax on `Variable`, because `DataArray.__array_wrap__` uses the `Variable` version of the method... Work in progress!

shoyer:
Also, feel free to rewrite to avoid using inject_all_ops_and_reduce_methods() for argmin/argmax at all. Method injection is pretty hacky, and generally not worthwhile, e.g., see [this note](https://github.com/pydata/xarray/blob/c97e7998b69c5b41e863dfb0a43450491fb8de45/xarray/core/ops.py#L3) about removing it.

johnomotani:
I eventually found that the cause of the errors I was getting was that the `argmin` and `argmax` methods did not have an `out` argument. In order for the methods to be wrapped by `numpy` (and I guess `dask` is the same), the call signature of `np.argmin()` has to be supported, which means `axis` and `out` arguments are needed.

@shoyer I've kept the use of injected functions, because removing the injected `argmin` and `argmax` completely meant re-implementing handling of `skipna` and `only` within `_unravel_argminmax`, which seemed less nice to me. If/when method injection is refactored, it would be nice to include a mechanism to override the core operation with an explicitly implemented, extended version like `argmin`/`argmax`.

I think this PR is ready for review now :smile:

johnomotani:
@pydata/xarray - I think this PR is ready to be merged, are there any changes I should make? Thanks!

xarray/core/dataset.py
```diff
@@ -6294,5 +6294,55 @@ def idxmax(
...
         )
 
+    def argmin(self, dim=None, axis=None, **kwargs):
+        if dim is None and axis is None:
+            warnings.warn(
+                "Behaviour of DataArray.argmin() with neither dim nor axis argument "
+                "will change to return a dict of indices of each dimension, and then it "
+                "will be an error to call Dataset.argmin() with no argument. To get a "
+                "single, flat index, please use np.argmin(ds) instead of ds.argmin().",
+                DeprecationWarning,
```
shoyer, line 6304: When issuing warnings, please add the `stacklevel` keyword argument, which ensures that the warning ends up in user code where they call the xarray method, rather than inside the method. Usually `stacklevel=2` is right:
https://docs.python.org/3/library/warnings.html#warnings.warn

Sometimes you need a higher `stacklevel` if the warning is issued from a helper function

xarray/core/ops.py
```diff
@@ -260,6 +260,9 @@ def inject_reduce_methods(cls):
         + [("count", duck_array_ops.count, False)]
     )
     for name, f, include_skipna in methods:
+        if hasattr(cls, name):
+            name = "_injected_" + name
```
shoyer, line 264: We're only using `_injected_*` for two methods in this file (argmin/argmax), but you added it this check everywhere methods get added. This makes this already complicated modules more complicated :/.

I would strongly prefer to avoid adding these `_injected_argmin` methods if possible. Let me see if I can work-up an example for you...

shoyer:
> I eventually found that the cause of the errors I was getting was that the `argmin` and `argmax` methods did not have an `out` argument. In order for the methods to be wrapped by `numpy` (and I guess `dask` is the same), the call signature of `np.argmin()` has to be supported, which means `axis` and `out` arguments are needed.

I guess this would happen if somebody calls `np.argmin(xarray_object)`? Does this happen from inside xarray somewhere, or in external code?

If it's the former, then we should just fix xarray not to do this. If it's only in external code, I would still consider breaking this behavior.

The override mechanism that NumPy uses in argmin/argmax is really old/fragile, and it's not something we ever intended to support. There's no way we could write an `argmin`/`argmax` method with an incompatible signature that would work with these functions.

The better way to do overrides in NumPy is probably with `__array_function__` now, but I'm not sure we want the trouble...

xarray/core/dataset.py
```diff
@@ -6294,5 +6294,55 @@ def idxmax(
             ),
         )
 
+    def argmin(self, dim=None, axis=None, **kwargs):
+        if dim is None and axis is None:
+            warnings.warn(
+                "Behaviour of DataArray.argmin() with neither dim nor axis argument "
```
shoyer, line 6300: This warning mentions DataArray, but this is on the Dataset method.

xarray/core/variable.py
```diff
@@ -2069,6 +2079,171 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        stacked = self.stack({newdimname: dim})
+
+        result_dims = stacked.dims[:-1]
+        reduce_shape = tuple(self.sizes[d] for d in dim)
+
+        result_flat_indices = getattr(stacked, "_injected_" + str(argminmax))(
+            axis=-1, skipna=skipna, out=out
+        )
+
+        result_unravelled_indices = np.unravel_index(result_flat_indices, reduce_shape)
```
shoyer, line 2134: Could you please add `unravel_index` to `xarray/core/duck_array_ops.py` and use it here? That way this method will work with dask arrays:
```python
unravel_index = _dask_or_eager_func("unravel_index")
```

shoyer:
I notice that some deprecation warnings get issued when running your tests. Please silence these methods, e.g., using `pytest.mark.filterwarnings`.

xarray/core/ops.py
shoyer: OK, I just pushed a commit to your branch that shows how you can call `Variable.reduce()` instead of using these injected methods. I think it should be possible to remove all of these injected methods now.

xarray/core/variable.py
```diff
@@ -2069,6 +2079,174 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        if dim is None and axis is None:
+            warnings.warn(
+                "Behaviour of argmin/argmax with neither dim nor axis argument will "
+                "change to return a dict of indices of each dimension. To get a "
+                "single, flat index, please use np.argmin(da.data) or "
+                "np.argmax(da.data) instead of da.argmin() or da.argmax().",
+                DeprecationWarning,
+            )
+
+        argminax_func = getattr(duck_array_ops, argminmax)
```
keewis, line 2103:
```suggestion
        argminmax_func = getattr(duck_array_ops, argminmax)
```

xarray/core/variable.py
```diff
@@ -2069,6 +2079,174 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        if (
+            dim is None
+            or axis is not None
+            or not isinstance(dim, Sequence)
+            or isinstance(dim, str)
+        ):
+            # Return int index if single dimension is passed, and is not part of a
+            # sequence
+            return self.reduce(
+                argminax_func, dim=dim, axis=axis, keep_attrs=keep_attrs, skipna=skipna
```
keewis, line 2117:
```suggestion
                argminmax_func, dim=dim, axis=axis, keep_attrs=keep_attrs, skipna=skipna
```

xarray/core/variable.py
```diff
@@ -2069,6 +2079,174 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
         )
         return type(self)(self.dims, numeric_array, self._attrs)
 
+    def _unravel_argminmax(
+        self,
+        argminmax: Hashable,
```
keewis, line 2084: `mypy` complains that `getattr` only accepts `str`, not `Hashable`. Since we only use it there:
```suggestion
        argminmax: str,
```

xarray/core/ops.py
```diff
@@ -325,24 +331,35 @@ def inject_all_ops_and_reduce_methods(cls, priority=50, array_only=True):
...
     for name in PANDAS_UNARY_FUNCTIONS:
         f = _func_slash_method_wrapper(getattr(duck_array_ops, name), name=name)
+        if hasattr(cls, op_str(name)):
+            name = "_injected_" + name
         setattr(cls, name, cls._unary_op(f))
 
     f = _func_slash_method_wrapper(duck_array_ops.around, name="round")
-    setattr(cls, "round", cls._unary_op(f))
+    if hasattr(cls, "round"):
+        setattr(cls, "_injected_round", cls._unary_op(f))
```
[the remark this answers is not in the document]
johnomotani: The idea was that `round` should behave the same as all the other injected methods, in that if `round` was (at some point in the future) explicitly defined as a method of say `DataArray`, it would override the currently used `duck_array_ops.around`. This is what would happen for all the other injected methods, so I thought it would be more consistent for `round` to behave the same way, rather than be skipped because it's already treated specially here. Happy to revert this change if you prefer.

xarray/core/dataarray.py
```diff
@@ -3722,6 +3722,215 @@ def idxmax(
...
+            object to the new one.  If False (default), the new object will be
+            returned without attributes.
+        skipna : bool, optional
+            If True, skip missing values (as marked by NaN). By default, only
+            skips missing values for float dtypes; other dtypes either do not
+            have a sentinel missing value (int) or skipna=True has not been
+            implemented (object, datetime64 or timedelta64).
+        out : None
+            'out' should not be passed - provided for compatibility with numpy function
+            signature
```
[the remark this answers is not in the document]
johnomotani: Without the `out` argument `np.argmin(da)` went wrong somewhere in the internals of `numpy`. Is there another workaround? Is it OK to just throw an exception if `out` is passed?

xarray/core/variable.py
```diff
@@ -2069,6 +2079,174 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        while newdimname in self.dims:
+            newdimname = "_unravel_argminmax_dim_{}".format(count)
+            count += 1
+
+        stacked = self.stack({newdimname: dim})
+
+        result_dims = stacked.dims[:-1]
+        reduce_shape = tuple(self.sizes[d] for d in dim)
+
+        result_flat_indices = stacked.reduce(argminax_func, axis=-1, skipna=skipna)
```
johnomotani, line 2133:
```suggestion
        result_flat_indices = stacked.reduce(argminmax_func, axis=-1, skipna=skipna)
```

johnomotani:
> I guess this would happen if somebody calls np.argmin(xarray_object)? Does this happen from inside xarray somewhere, or in external code?
>
> If it's the former, then we should just fix xarray not to do this. If it's only in external code, I would still consider breaking this behavior.

`test_units.py` has a couple of tests that apply `numpy` functions to `xr.Variable`s, `xr.DataArray`s and `xr.Dataset`s. Those tests break for `argmin` and `argmax` without the `out` argument. Should I just remove the `argmin` and `argmax` function calls? The tests would still call them as methods.

keewis:
you could mark the `function("argmin")` / `function("argmax")` with `xfail` (or `skip`?). Then we can decide on what to do with them when we close #3917

xarray/core/dataset.py
johnomotani: That is intentional, although maybe it should mention both `DataArray.argmin()` and `Variable.argmin()` - the current behaviour of `DataArray.argmin()` with no `dim` argument is deprecated - it will change in future to return a `dict`. At that point it wouldn't make sense for `Dataset.argmin()` with no dim argument to return a result, because it doesn't make much sense to return a `dict` of `Dataset`s (and if the method returned a result using `reduce` to wrap `duck_array_ops.argmin`, it would return a single `Dataset` with the flattened index value for each member), so I think `Dataset.argmin()` with no `dim` argument should raise an error once the change is made to `DataArray.argmin()` and `Variable.argmin()`. I'll reword the message to make this clearer...

xarray/core/dataarray.py
```diff
@@ -3819,6 +3819,209 @@ def idxmax(
             keep_attrs=keep_attrs,
         )
 
+    def argmin(
+        self,
+        dim: Union[Hashable, Sequence[Hashable]] = None,
+        axis: Union[int, None] = None,
+        keep_attrs: bool = None,
+        skipna: bool = None,
+    ) -> Union["DataArray", Dict[Hashable, "DataArray"]]:
```
shoyer, line 3828: In theory, it would be nice to use either `TypeVar` or `@overload` so tools like mypy can find out the type of the the return value from `argmin()` based on the type of `dim`.

But definitely don't worry about that now, we can save that for a follow-up :)

shoyer:
I'm going to wait a little while before merging in case anyone else has comment, but otherwise will merge in a day or two (definitely before the next release)

xarray/core/dataarray.py
```diff
@@ -3819,6 +3819,209 @@ def idxmax(
...
+        """Index or indices of the minimum of the DataArray over one or more dimensions.
+
+        If a sequence is passed to 'dim', then result returned as dict of DataArrays,
+        which can be passed directly to isel(). If a single str is passed to 'dim' then
+        returns an int.
```
dcherian, line 3833: minor comment:
```suggestion
        returns a DataArray with dtype int.
```

xarray/core/dataarray.py
```diff
@@ -3819,6 +3819,209 @@ def idxmax(
             keep_attrs=keep_attrs,
         )
 
+    def argmin(
+        self,
+        dim: Union[Hashable, Sequence[Hashable]] = None,
+        axis: Union[int, None] = None,
```
dcherian, line 3825: nit: IIUC this could be
```suggestion
        axis: int = None,
```
like `keep_attrs` and `skipna` below

xarray/core/dataset.py
```diff
@@ -6294,5 +6294,55 @@ def idxmax(
             ),
         )
 
+    def argmin(self, dim=None, axis=None, **kwargs):
```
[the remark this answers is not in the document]
dcherian: Can we copy over the docstring and signature (missing `keep_attrs` and `skipna`) from `DataArray.argmin` with suitable modifications please? That would make it quite convenient for users.

xarray/core/dataset.py
```diff
@@ -6368,5 +6368,65 @@ def idxmax(
...
+            )
+        if (
+            dim is None
+            or axis is not None
+            or not isinstance(dim, Sequence)
+            or isinstance(dim, str)
+        ):
+            # Return int index if single dimension is passed, and is not part of a
+            # sequence
+            argmin_func = getattr(duck_array_ops, "argmin")
```
dcherian, line 6392: we'll need to preserve `Dataset.attrs` if `keep_attrs=True`

xarray/tests/test_units.py
```diff
@@ -2283,6 +2298,9 @@ def test_repr(self, func, variant, dtype):
         ),
         ids=repr,
     )
+    @pytest.mark.filterwarnings(
+        "ignore:Behaviour of argmin/argmax with neither dim nor :DeprecationWarning"
+    )
```
keewis, line 2303: same here: change to `method("argmin", dim="x")` / `method("argmax", dim="x")`

xarray/tests/test_units.py
```diff
@@ -3841,6 +3871,12 @@ def test_repr(self, func, variant, dtype):
         ),
         ids=repr,
     )
+    @pytest.mark.filterwarnings(
+        "ignore:Once the behaviour of DataArray:DeprecationWarning"
+    )
+    @pytest.mark.filterwarnings(
+        "ignore:Behaviour of argmin/argmax with neither dim nor :DeprecationWarning"
+    )
```
keewis, line 3879: same here: change to `method("argmin", dim="x")` / `method("argmax", dim="x")`

xarray/core/dataset.py
johnomotani: I think this is taken care of by `self.reduce()`?

xarray/tests/test_units.py
```diff
@@ -1427,6 +1427,9 @@ def test_real_and_imag(self):
         ),
         ids=repr,
     )
+    @pytest.mark.filterwarnings(
+        "ignore:Behaviour of argmin/argmax with neither dim nor :DeprecationWarning"
+    )
```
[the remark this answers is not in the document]
johnomotani: Unfortunately the numpy function won't accept a `dim` argument, but we can work around in a similar way by passing `axis=0`.

xarray/core/dataarray.py
```diff
@@ -3819,6 +3819,209 @@ def idxmax(
...
+        result = self.variable.argmin(dim, axis, keep_attrs, skipna)
+        if isinstance(result, dict):
+            return {k: self._replace_maybe_drop_dims(v) for k, v in result.items()}
+        else:
+            return self._replace_maybe_drop_dims(result)
+
+    def argmax(
+        self,
+        dim: Union[Hashable, Sequence[Hashable]] = None,
+        axis: Union[int, None] = None,
```
johnomotani, line 3926: change as suggested for argmin
```suggestion
        axis: int = None,
```

xarray/core/dataarray.py
```diff
@@ -3819,6 +3819,209 @@ def idxmax(
...
+        keep_attrs: bool = None,
+        skipna: bool = None,
+    ) -> Union["DataArray", Dict[Hashable, "DataArray"]]:
+        """Index or indices of the maximum of the DataArray over one or more dimensions.
+
+        If a sequence is passed to 'dim', then result returned as dict of DataArrays,
+        which can be passed directly to isel(). If a single str is passed to 'dim' then
+        returns an int.
```
johnomotani, line 3934: change as for argmin
```suggestion
        returns a DataArray with dtype int.
```

xarray/core/variable.py
```diff
@@ -2069,6 +2079,166 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        self,
+        dim: Union[Hashable, Sequence[Hashable]] = None,
+        axis: Union[int, None] = None,
+        keep_attrs: bool = None,
+        skipna: bool = None,
+    ) -> Union["Variable", Dict[Hashable, "Variable"]]:
+        """Index or indices of the minimum of the Variable over one or more dimensions.
+        If a sequence is passed to 'dim', then result returned as dict of Variables,
+        which can be passed directly to isel(). If a single str is passed to 'dim' then
+        returns an int.
```
johnomotani, line 2162: change similarly to `DataArray.argmin`
```suggestion
        returns a Variable with dtype int.
```

xarray/core/variable.py
```diff
@@ -2069,6 +2079,166 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        if keep_attrs:
+            for v in result.values():
+                v.attrs = self.attrs
+
+        return result
+
+    def argmin(
+        self,
+        dim: Union[Hashable, Sequence[Hashable]] = None,
+        axis: Union[int, None] = None,
```
johnomotani, line 2155: change similarly to `DataArray.argmin`
```suggestion
        axis: int = None,
```

xarray/core/variable.py
```diff
@@ -2069,6 +2079,166 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        See also
+        --------
+        DataArray.argmin, DataArray.idxmin
+        """
+        return self._unravel_argminmax("argmin", dim, axis, keep_attrs, skipna)
+
+    def argmax(
+        self,
+        dim: Union[Hashable, Sequence[Hashable]] = None,
+        axis: Union[int, None] = None,
```
johnomotani, line 2200: change similarly to `DataArray.argmin`
```suggestion
        axis: int = None,
```

xarray/core/variable.py
```diff
@@ -2069,6 +2079,166 @@ def _to_numeric(self, offset=None, datetime_unit=None, dtype=float):
...
+        keep_attrs: bool = None,
+        skipna: bool = None,
+    ) -> Union["Variable", Dict[Hashable, "Variable"]]:
+        """Index or indices of the maximum of the Variable over one or more dimensions.
+        If a sequence is passed to 'dim', then result returned as dict of Variables,
+        which can be passed directly to isel(). If a single str is passed to 'dim' then
+        returns an int.
```
johnomotani, line 2207: change similarly to `DataArray.argmin`
```suggestion
        returns a Variable with dtype int.
```

xarray/core/dataset.py
```diff
@@ -4295,15 +4295,15 @@ def reduce(
...
-                    elif len(reduce_dims) == var.ndim:
+                    if len(reduce_dims) == var.ndim:
                         # prefer to aggregate over axis=None rather than
                         # axis=(0, 1) if they will be equivalent, because
                         # the former is often more efficient
                         reduce_dims = None  # type: ignore
+                    elif len(reduce_dims) == 1:
+                        # unpack dimensions for the benefit of functions
+                        # like np.argmin which can't handle tuple arguments
+                        (reduce_dims,) = reduce_dims
```
johnomotani, line 4306: I needed to swap these conditionals around to allow passing `axis=0` argument when calling `argmax` on a `Dataset` in `test_units.py`. The test only has one dimension, so it was going through the `len(reduce_dims) == 1` branch so `reduce_dims` was not `None`, causing an error passing poth `dim` and `axis` arguments to `var.reduce`. I can't see any harm to the change, but could you check please @shoyer, @dcherian?

johnomotani: This change broke `TestDataset.test_resample_and_first` and `TestDataset.test_resample_by_last_discarding_attrs`, so not as harmless as I'd hoped. Is there a simple fix for this? Otherwise I guess I'll revert and put the warning filter back on `test_units.TestDataArray.test_aggregation`.

keewis:
I noticed a few issues while debugging:
- calling on a array with missing values raises `ValueError: All-NaN slice encountered` (try `ds = xr.tutorial.open_dataset("rasm"); ds.argmin(dim="y")`)
- `ds.argmin(dim=...)` returns a single index (the same result as `ds.argmin()`) but `ds.Tair.argmin(dim=...)` returns something different from `ds.Tair.argmin()`. Is that intentional?

xarray/core/dataset.py
keewis: I modified the `method` class to always remove certain kwargs if the object is not a xarray object (e.g. `dim`) and also made sure duck arrays get `axis` instead of `dim`. That doesn't work for `Dataset` so I used `dim=...` instead (not sure if that should work, though).

xarray/core/dataset.py
dcherian: yes you are right. sorry for the noise.

johnomotani:
> ds.argmin(dim=...) returns a single index (the same result as ds.argmin()) but ds.Tair.argmin(dim=...) returns something different from ds.Tair.argmin(). Is that intentional?

I think this is a bug. `dim=...` to `Dataset.argmin` should be an error because `dim=...` returns a `dict` of results, and it's not clear how to do that consistently for a Dataset that might have several members with different combinations of dimensions.

keewis:
well, we'd need to somehow be able to use `sel` to index every variable in a `Dataset` differently. While that would probably be possible it would make `sel` way too complicated, so I agree that an error would be good here.

I think no further changes to the `pint` tests should be required (except refactoring, but I can do that in a new PR after this one has been merged).

johnomotani:
> calling on a array with missing values raises ValueError: All-NaN slice encountered (try ds = xr.tutorial.open_dataset("rasm"); ds.argmin(dim="y"))

Some missing values should be OK. In your example though for example
```
In [27]: ds.Tair.isel(time=0, x=0).values                                                     
Out[27]: 
array([nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
       nan, nan, nan, nan, nan, nan, nan, nan, nan, nan])
```
So there are no values in the array slice that argmin should be applied over, which is an error.

I guess we could add some special handling for this (not sure what though, because we can't set a variable with type `int` to `nan` or `None`), but I think that is a separate issue that would need a new PR.

johnomotani:
Thanks @keewis! I think we've addressed all the review comments now.

doc/whats-new.rst
```diff
@@ -52,6 +52,14 @@ Enhancements
...
 New Features
 ~~~~~~~~~~~~
+- :py:meth:`DataArray.argmin` and :py:meth:`DataArray.argmax` now support
+  sequences of 'dim' arguments, and if a sequence is passed return a dict
+  (which can be passed to :py:meth:`isel` to get the value of the minimum) of
+  the indices for each dimension of the minimum or maximum of a DataArray.
+  (:pull:`3936`)
+  By `John Omotani <https://github.com/johnomotani>`_, thanks to `Keisuke Fujii
+  <https://github.com/fujiisoup>`_ for work in :pull:`1469`.
+- Added :py:meth:`DataArray.polyfit` and :py:func:`xarray.polyval` for fitting polynomials. (:issue:`3349`)
```
keewis, line 62: is this due to an error when merging?
```suggestion
```

johnomotani: Some sort of merge error, thanks!

dcherian:
Thanks @johnomotani . This is a significant contribution.

johnomotani:
Thanks @dcherian :smile: You're very welcome!

PR #3947, opened 2020-04-07
Fix typo

dmey:


keewis:
thanks @dmey, this is indeed a typo. We really should try to run doctest (or something similar for `rst` files) on our examples to catch these issues.

PR #3956, opened 2020-04-08
Add multi-dimensional extrapolation example and mention different behavior of kwargs in interp

risebell:
This corrects an error in the documentation of using `interp` to extrapolate. Apparently the argument `fill_value` changed at some point and now uses `None` instead of `'extrapolate'` to indicate that extrapolation should be done.

### DIFF
--- a/doc/interpolation.rst
+++ b/doc/interpolation.rst
@@ -150,8 +150,14 @@ Additional keyword arguments can be passed to scipy's functions.
 
     # fill 0 for the outside of the original coordinates.
     da.interp(x=np.linspace(-0.5, 1.5, 10), kwargs={'fill_value': 0.0})
-    # extrapolation
+    # 1-dimensional extrapolation
     da.interp(x=np.linspace(-0.5, 1.5, 10), kwargs={'fill_value': 'extrapolate'})
+    # multi-dimensional extrapolation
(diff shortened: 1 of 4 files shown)


dcherian:
Can you add a note explicitly mentioning that change in scipys API?

risebell:
Ok, I should have looked into this some more, my bad. The problem here is that scipy's interp1d and interpn accept different values for `fill_value`, and I blindly assumed they would use the same. The option `'extrapolate'` does work for interp1d, while interpn emits a ValueError in that case and needs `None` to extrapolate. On the other hand, interp1d with `fill_value=None` does not extrapolate.

Maybe a note on the `kwargs` parameter of interp mentioning that scipy's interpolation functions accept different arguments and a example of multi-dimensional extrapolation would be the way to go here?

dcherian:
Sure

dcherian:
Thanks @risebell . It'd be nice to open a scipy issue asking them to make their API consistent.

dcherian:
Also please add a note to `whats-new` taking credit for this contribution

risebell:
Sorry for the delay. I added a note to `whats-new` and updated the title of the PR to reflect what's actually done with this. I also opened an issue with scipy over at scipy/scipy#11853.

dcherian:
Thanks @risebell

max-sixty:
Thanks @risebell !

PR #3958, opened 2020-04-09
New coords to existing dim (doc)

### LINKED ISSUE #3952: Doc improvement for DataArray.assign_coords
Hi, first I wanted to say that xarray is a really fantastic library, it saves so many manipulations in my work.

I was looking for a way to add a coordinate to a DataArray and attribute it to a specific dimension. I found that the following syntax was exactly what I was looking for :

```python
da.assign_coords(new_coordinate=('dim_to_use', values))
```

However, the [documentation](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.assign_coords.html#xarray.DataArray.assign_coords) of the method was not referencing this syntax. What do you think?

EtienneCmb:
- [x] Closes #3952 

### DIFF
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -444,7 +444,7 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         Coordinates:
           * lon      (lon) int64 -2 -1 0 1
 
-        New coordinate can also be attached to an existing dimension :
+        New coordinate can also be attached to an existing dimension:
 
         >>> lon_2 = np.array([300, 289, 0, 1])
         >>> da.assign_coords(lon_2=('lon', lon_2))
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -453,8 +453,9 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
         Coordinates:
           * lon      (lon) int64 358 359 0 1
             lon_2    (lon) int64 300 289 0 1
-        >>> # Note that the same result can also be obtained with a dict e.g.
-        >>> # `da.assign_coords(dict(lon_2=('lon', lon_2)))`
+        
+        Note that the same result can also be obtained with a dict e.g.
+        >>> _ = da.assign_coords(dict(lon_2=('lon', lon_2)))
 
         Notes
         -----
--- a/xarray/core/common.py
+++ b/xarray/core/common.py
@@ -398,11 +398,15 @@ class DataWithCoords(SupportsArithmetic, AttrAccessMixin):
 
         Parameters
         ----------
-        coords : dict, optional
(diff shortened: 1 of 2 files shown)


xarray/core/common.py
```diff
@@ -440,10 +444,22 @@ def assign_coords(self, coords=None, **coords_kwargs):
...
+
+        >>> lon_2 = np.array([300, 289, 0, 1])
+        >>> da.assign_coords(lon_2=('lon', lon_2))
+        <xarray.DataArray (lon: 4)>
+        array([0.28298 , 0.667347, 0.657938, 0.177683])
+        Coordinates:
+          * lon      (lon) int64 358 359 0 1
+            lon_2    (lon) int64 300 289 0 1
+        >>> # Note that the same result can also be obtained with a dict e.g.
+        >>> # `da.assign_coords(dict(lon_2=('lon', lon_2)))`
```
max-sixty, line 457:
```suggestion
        
        Note that the same result can also be obtained with a dict e.g.
        >>> _ = da.assign_coords(dict(lon_2=('lon', lon_2)))
```

xarray/core/common.py
```diff
@@ -398,11 +398,15 @@ def assign_coords(self, coords=None, **coords_kwargs):
...
         Parameters
         ----------
-        coords : dict, optional
-            A dict with keys which are variables names. If the values are
-            callable, they are computed on this object and assigned to new
-            coordinate variables. If the values are not callable,
-            (e.g. a ``DataArray``, scalar, or array), they are simply assigned.
+        coords : (dict, tuple, optional)
+            If a dict is provided, the keys are the names of the coordinates
+            with the new values to assign. If the values are callable, they are
```
keewis, line 403:
```suggestion
        coords : dict, optional
            A dict where the keys are the names of the coordinates
            with the new values to assign. If the values are callable, they are
```
I think you confused this parameter with the `**coords_kwargs` below: the usage with `coords` is
```python
da.assign_coords({"lon_2": ('lon', lon_2)})
```
whereas the `coords_kwargs` form is:
```python
da.assign_coords(lon_2=('lon', lon_2))
```

EtienneCmb:
@keewis thanks for the suggestions

dcherian:
Thanks @EtienneCmb

EtienneCmb:
@dcherian @keewis The final line is not properly formatted in the latest [version of the doc](http://xarray.pydata.org/en/latest/generated/xarray.Dataset.assign_coords.html#xarray.Dataset.assign_coords). Might be caused by the `_` symbol?

keewis:
thanks for checking, @EtienneCmb. I don't think the `_` matters, this mostly happens because of a missing newline before the doctest line (the one that starts with `>>>`). Would you like to open a new PR fixing that?

PR #3960, opened 2020-04-09
parametrized accessor examples

### LINKED ISSUE #3829: {DataArray,Dataset} accessors with parameters
I would like to be able to create an DataArray accessor that takes parameters, e.g. `obj.weighted(w).sum(dim)`. This appears to be impossible using the existing `@xr.register_dataarray_accessor`, which supports only accessors of the form `obj.weighted.sum(w, dim)`.

To support the desired syntax, one could simply change https://github.com/pydata/xarray/blob/master/xarray/core/extensions.py#L36
from
```
            accessor_obj = self._accessor(obj)
```
 to
```
            accessor_obj = partial(self._accessor, obj)
```
But that would break the current syntax (i.e. would require `obj.accessor().foo()`, so is clearly not acceptable.

So any suggestions (short of simply creating a slightly modified copies of `register_{dataarray,dataset}_accessor`) for supporting both the existing `obj.accessor.foo()` syntax as well as my desired `obj.accessor(*args, **kwargs).foo()` syntax?

keewis:
This starts adding the parametrized accessor examples from #3829 to the accessor documentation as suggested by @jhamman. Since then the `weighted` methods have been added, though, so I'd like to use a different example instead (ideas welcome).

Also, this feature can be abused to add functions to the main `DataArray` / `Dataset` namespace (by registering a function with the `register_*_accessor` decorators, see the second example). Is this something we want to explicitly discourage?

(When trying to build the docs locally, sphinx keeps complaining about a code block without code. Not sure what that is about)

 - [x] Closes #3829

### DIFF
--- a/doc/internals.rst
+++ b/doc/internals.rst
@@ -113,6 +113,44 @@ reasons:
    a DataArray from a Dataset (ex. ``ds[var_name]``), will have new
    accessors created.
 
+Parametrizing an accessor is possible, too, by defining ``__call__``:
+
+.. ipython:: python
+
+    @xr.register_dataarray_accessor("weighted")
+    class Weighted:
+        def __init__(self, xarray_obj):
+            self._obj = xarray_obj
+            self._weight = None
+
+        def __call__(self, weight):
+            self._weight = weight
+
+        def sum(self, dim):
+            return "weighted sum"
+
+If we need to require the parameter, the easiest way to do so is using
+a wrapper function:
+
+.. ipython:: python
+
+    class Weighted:
+        def __init__(self, obj, weight):
+            self._obj = obj
+            self._weight = weight
+
+        def sum(self, dim):
+            return f"weighted sum over {dim} and with weight {self._weight}"
+
+    @xr.register_dataarray_accessor("weighted")
+    def weighted(obj):
+        def wrapped(weight):
+            return Weighted(obj, weight)
+        return wrapped
+
+    da = xr.DataArray(data=range(5), dims="x")
+    da.weighted(5).sum(dim="x")
+
 Back in an interactive IPython session, we can use these properties:
 
 .. ipython:: python


jhamman:
Thanks @keewis!

keewis:
I tried to polish the example (see [here](http://xarray.pydata.org/en/accessors-example/internals.html#extending-xarray)) and also added a note trying to explicitly discourage using this to add new methods to the main namespace.

This makes it clear, though, that the error message for the required parameter implementation could be better. For replacement, I think we could use either a new keyword-only parameter to the `register_*_functions`, a class decorator or some sort of mixin / base class / metaclass. The decorators would generate some sort of wrapper that enforces the `__call__` while the other options would need to do some kind of magic. Considering that, I think I'd prefer using a decorator (but I'm not sure how important this use-case is).

mathause:
+1 on not using `weighted` as an example - although I don't have a good idea, either... So maybe just add a note that it is included in xarray > 0.15.1?

doc/internals.rst
```diff
@@ -129,9 +129,77 @@ Back in an interactive IPython session, we can use these properties:
...
+
+    In [4]: da.weighted.sum()
+
+    In [5]: da.weighted(weights).sum()
+
+.. note::
+
+   Keep in mind, though, that accessors are designed to add new
+   namespaces to the :py:class:`Dataset` and :py:class:`DataArray`
+   objects and should not be used to add methods.
```
kmuehlbauer, line 199: @keewis Reading this as a user. Can you explain what  you mean by **"_should not be used to add new methods_"**?

keewis: you could use the `register_*_accessor` functions to add new methods to the main namespace using
```python
In [2]: @xr.register_dataset_accessor("my_method") 
   ...: def my_method(self): 
   ...:     def wrapped(*args, **kwargs): 
   ...:         print(
   ...:             "called my_method on",
   ...:             self,
   ...:             f"with {args}, {kwargs}",
   ...:             sep="\n -- ",
   ...:         )
   ...:     return wrapped 
   ...:  
   ...: ds = xr.Dataset({"a": ("x", [0, 1])}) 
   ...: ds.my_method()
called my_method on
 -- <xarray.Dataset>
Dimensions:  (x: 2)
Data variables:
    a        (x) int64 0 1
 -- with (), {}
```
and I thought we should discourage that. The note probably needs a rewrite to make that a bit easier to understand (or maybe remove it?).

kmuehlbauer: Thanks, makes sense now. But adding methods to my new accessor is the way proposed then, right? (like shown for the `geo` accessor)

keewis: yes, you'd use the accessor as a namespace object so your methods would be called with `ds.my_accessor.my_method()`

keewis:
@mathause I played around with extending the example `GeoAccessor` to parametrize on `crs` but decided that the current version about reimplementing the API of `weighted` was easier.

doc/internals.rst
```diff
@@ -129,9 +129,77 @@ Back in an interactive IPython session, we can use these properties:
...
+       ...:         self._weights = weights
+       ...:
+       ...:     def sum(self, *args, **kwargs):
+       ...:         return np.sum(self._obj * self._weights, *args, **kwargs)
+
+    In [2]: @xr.register_dataarray_accessor("weighted")
+       ...: def weighted(obj):
+       ...:     def wrapped(weights):
+       ...:         return Weighted(obj, weights)
+       ...:     return wrapped
```
shoyer, line 183: Using `partial` from functools could look a little cleaner here?
```suggestion
       ...:     return partial(Weighted, obj)
```

I guess the downside is that introspection on the argument args/signature probably wouldn't work quite as well.

doc/internals.rst
```diff
@@ -129,9 +129,77 @@ Back in an interactive IPython session, we can use these properties:
...
+
+    In [1]: @xr.register_dataarray_accessor("weighted")
+       ...: class WeightedAccessor:
+       ...:     def __init__(self, xarray_obj):
+       ...:         self._obj = xarray_obj
+       ...:         self._weights = np.ones_like(xarray_obj) / xarray_obj.size
+       ...:
+       ...:     def __call__(self, weights):
+       ...:         self._weights = weights
+       ...:         return self
```
shoyer, line 149: This approach seems potentially a little error prone because it involves mutation. Could we possibly mention it less prominently? or maybe not at all?

keewis: That's true, but I don't really understand why this would be an issue if you only ever call it with `da.accessor()` or `da.accessor(data)`. We could also do
```python
class Accessor:
    def __init__(self, obj, data=None):
        self._obj = obj
        self._data = data

    def __call__(self, data=None):
        return type(self)(self._obj, data)
```
which does not rely on the internal state but is a little bit more complex.

doc/internals.rst
keewis: This looks somewhat like a decorator, but I'm not really sure if that's good or bad. The missing introspection is not that much of an issue: contrary to properties, autocompleters usually don't complete the attributes of return values of function calls. However, the signature and docstring is lost on `partial` objects, so `help()` becomes useless. Also, the error message when forgetting to call the accessor could be better (a `AttributeError` complaining about `function` or the `partial` object not having the attribute).

One way to work around that could be to extend the accessor registration functions to automatically (or only if a keyword-only argument was set to `True`) use a wrapper class if the accessor defines `__call__` or some sort of hook method. That wrapper would then do the same as the function, but provide more helpful error messages.

I'm thinking of something like
```python
class ParametrizedAccessorWrapper:
    def __init__(self, accessor_cls):
        self._accessor_cls = accessor_cls

    def __call__(self, *args, **kwargs):
        return self._accessor_cls(*args, **kwargs)

    def __getattr__(self, name):
        raise AttributeError(
            f"cannot access attribute {name!r} without parametrizing the accessor"
        )
```

keewis:
revisiting this: does it make sense to require calling if a `__call__` is defined? I.e. always require
`da.accessor().attribute` instead of `da.accessor.attribute`?

This would probably need changes to the `register_*_accessor` decorators that adds the checking, but that way we could also make sure people don't use this to add new functions: if `__call__` may only return `self` or a new object of the same type, abusing the accessor decorators would be a lot harder. The frequent creation of new objects would require a cheap `Accessor.__init__`, though.

Another way to achieve that would be to patch out `__call__` and put it on a wrapper class, as mentioned above.

Does that make sense or is this a bit too complicated?

doc/internals/extending-xarray.rst
```diff
@@ -88,9 +88,21 @@ Back in an interactive IPython session, we can use these properties:
...
 The intent here is that libraries that extend xarray could add such an accessor
 to implement subclass specific functionality rather than using actual subclasses
-or patching in a large number of domain specific methods. For further reading
-on ways to write new accessors and the philosophy behind the approach, see
-:issue:`1080`.
+or patching in a large number of domain specific methods.
+
+.. note::
+
+   With the exception of ``__init__``, ``__del__`` and the descriptor protocols it is
```
keewis, line 95: is that right? I assumed that the descriptor protocols would not work because the accessor object is constructed by another descriptor, but `__init__` is only restricted in the signature (this could probably use more careful wording) and I don't know if it makes sense to discourage using `__del__`?

keewis (commented):
I just noticed that it is better not to focus on `__call__` too much, especially since the builtin `.str` and `cf-xarray`'s `.cf` accessor use other protocols like `__getitem__` or `__add__`.

PR #3973, opened 2020-04-15
ensure Variable._repr_html_ works

### LINKED ISSUE #3972: Variable objects not displayable in notebook
class `Variable` subclasses` AbstractArray` which has a `_repr_html_` method so variable objects
should display in a notebook. 

But this fails with a 
`AttributeError: 'Variable' object has no attribute 'variable'`:

```python
from os.path import join,dirname
import xarray as xr
ds=join(dirname(xr.__file__), 'tests/data/bears.nc')
xr.open_dataset(ds).variables['cross']
```

### Fix:
change line 186 in `xarray.core.formatting_html.py:: array_section`
from
```python
preview = escape(inline_variable_array_repr(obj.variable, max_width=70))
```
to (because obj *is* a variable):

```python
preview = escape(inline_variable_array_repr(obj.variable if hasattr(obj, "variable") else obj, max_width=70))
```
I can make a PR if you like.

```
INSTALLED VERSIONS
------------------
commit: None
python: 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 23:03:20) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 5.3.0-46-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_AU.UTF-8
LOCALE: en_AU.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.7.4

xarray: 0.15.1
pandas: 1.0.3
numpy: 1.18.1
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.1.1.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.14.0
distributed: 2.6.0
matplotlib: 3.2.1
cartopy: None
seaborn: 0.10.0
numbagg: None
setuptools: 46.1.3.post20200325
pip: 20.0.2
conda: 4.8.3
pytest: 5.4.1
IPython: 7.13.0
sphinx: 2.2.1
```

arabidopsis:
Thanks for all the good work guys :)

 - [x] Closes #3972
 - [x] Test added

### DIFF
--- a/xarray/core/formatting_html.py
+++ b/xarray/core/formatting_html.py
@@ -183,7 +183,7 @@ def array_section(obj):
     # "unique" id to expand/collapse the section
     data_id = "section-" + str(uuid.uuid4())
     collapsed = ""
-    variable = obj.variable if hasattr(obj, "variable") else obj
+    variable = getattr(obj, "variable", obj)
     preview = escape(inline_variable_array_repr(variable, max_width=70))
     data_repr = short_data_repr_html(obj)
     data_icon = _icon("icon-database")
--- a/xarray/core/formatting_html.py
+++ b/xarray/core/formatting_html.py
@@ -183,7 +183,8 @@ def array_section(obj):
     # "unique" id to expand/collapse the section
     data_id = "section-" + str(uuid.uuid4())
     collapsed = ""
-    preview = escape(inline_variable_array_repr(obj.variable, max_width=70))
+    variable = obj.variable if hasattr(obj, "variable") else obj
+    preview = escape(inline_variable_array_repr(variable, max_width=70))
     data_repr = short_data_repr_html(obj)
     data_icon = _icon("icon-database")
 
--- a/xarray/tests/test_variable.py
+++ b/xarray/tests/test_variable.py
@@ -1077,17 +1077,6 @@ class TestVariable(VariableSubclassobjects):
         ).strip()
         assert expected == repr(v)
 
-    def test_repr_html(self):
-        v = Variable(["time", "x"], [[1, 2, 3], [4, 5, 6]], {"foo": "bar"})
-        assert hasattr(v, "_repr_html_")
-        with set_options(display_style="html"):
-            html = v._repr_html_().strip()
-        # We don't do a complete string identity since
-        # html output is probably subject to change, is long and... reasons.
-        # Just test that something reasonable was produced.
-        assert html.startswith("<div") and html.endswith("</div>")
-        assert "xarray.Variable" in html
-
     def test_repr_lazy_data(self):
         v = Variable("x", LazilyOuterIndexedArray(np.arange(2e5)))
         assert "200000 values with dtype" in repr(v)
--- a/xarray/tests/test_variable.py
+++ b/xarray/tests/test_variable.py
@@ -1077,6 +1077,17 @@ class TestVariable(VariableSubclassobjects):
         ).strip()
         assert expected == repr(v)
 
+    def test_repr_html(self):
+        v = Variable(["time", "x"], [[1, 2, 3], [4, 5, 6]], {"foo": "bar"})
+        assert hasattr(v, "_repr_html_")
+        with set_options(display_style="html"):
+            html = v._repr_html_().strip()
+        # We don't do a complete string identity since
+        # html output is probably subject to change, is long and... reasons.
+        # Just test that something reasonable was produced.
+        assert html.startswith("<div") and html.endswith("</div>")
+        assert "xarray.Variable" in html
+
     def test_repr_lazy_data(self):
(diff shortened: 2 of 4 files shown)


xarray/core/formatting_html.py
```diff
@@ -183,7 +183,8 @@ def array_section(obj):
     # "unique" id to expand/collapse the section
     data_id = "section-" + str(uuid.uuid4())
     collapsed = ""
-    preview = escape(inline_variable_array_repr(obj.variable, max_width=70))
+    variable = obj.variable if hasattr(obj, "variable") else obj
+    preview = escape(inline_variable_array_repr(variable, max_width=70))
```
keewis, line 187:
```suggestion
    preview = escape(inline_variable_array_repr(getattr(obj, "variable", obj), max_width=70))
```
you should probably run `black` after that (or assign the result of `getattr` to `variable`)

xarray/tests/test_variable.py
```diff
@@ -1077,6 +1077,17 @@ def test_repr(self):
...
+        v = Variable(["time", "x"], [[1, 2, 3], [4, 5, 6]], {"foo": "bar"})
+        assert hasattr(v, "_repr_html_")
+        with set_options(display_style="html"):
+            html = v._repr_html_().strip()
+        # We don't do a complete string identity since
+        # html output is probably subject to change, is long and... reasons.
+        # Just test that something reasonable was produced.
+        assert html.startswith("<div") and html.endswith("</div>")
+        assert "xarray.Variable" in html
+
```
keewis, line 1090: should this be moved into `test_formatting_html.py`?

xarray/core/formatting_html.py
arabidopsis: I believe I have black formatting on save (but I've been doing PRs on other repos and
black just reformats everything -- because they don't use it -- so I have to
turn it off)

arabidopsis: OK `whats-new.rst` updated

keewis: that's fine, it's just that I'm not sure if my suggestion would need to be reformatted. So what I meant was that if you commit via github's interface, you might have to run `black`.

xarray/tests/test_variable.py
```diff
@@ -1077,6 +1077,17 @@ def test_repr(self):
         ).strip()
         assert expected == repr(v)
 
+    def test_repr_html(self):
+        v = Variable(["time", "x"], [[1, 2, 3], [4, 5, 6]], {"foo": "bar"})
+        assert hasattr(v, "_repr_html_")
+        with set_options(display_style="html"):
```
shoyer, line 1083: There's no need to set the display option here if you're calling `_repr_html_` direclty.

xarray/core/formatting_html.py
keewis: TBC the `black` issue is not really that important, I wanted to propose replacing the conditional assignment
```python
variable = ... if hasattr(obj, ...) else ...
```
with a `getattr` call with a default:
```python
variable = getattr(obj, ..., obj)
```

arabidopsis: yep that's better

xarray/tests/test_variable.py
arabidopsis: Are you sure? There's an OPTIONS switch in AbstractArray._repr_html_() which
might break the test if it's not set to "html".

keewis: I think what @shoyer meant was that you should directly use `formatting_html.array_repr` instead of `_html_repr_` so there's no need for the options switch (you can keep the `hasattr` check, though).

xarray/tests/test_variable.py
arabidopsis: I thought about that... but I've got the Variable class in scope here so I don't have
to mess with any imports.

xarray/tests/test_variable.py
shoyer: OK, good catch. This is fine then.

xarray/tests/test_variable.py
shoyer: You can just use `xr.Variable` instead. I do think this test would make more test in `test_formatting_html.py` along with all the other HTML tests. Otherwise it's likely that someone reading through those tests would not  find this one.

arabidopsis: OK moved to `test_formatting_html.py`

shoyer:
Looks good, thank you!

I'll merge after verifying tests pass.

PR #3975, opened 2020-04-15
pint support for Dataset

keewis:
This is part of the effort to add support for `pint` (see #3594) to `Dataset` objects (although it will probably be a test-only PR, just like #3643).

 - [x] Tests added

The list of failing tests from #3594:
* Dataset methods
  - `__init__`: Needs unit support in IndexVariable, and merge does not work yet (test bug is also possible)
  - aggregation: `xarray` does not implement `__array_function__` (see #3917)
  - `rank`: depends on `bottleneck` and thus only works with `numpy.array`
  - `ffill`, `bfill`: uses `bottleneck`
  - `interpolate_na`: uses `numpy.vectorize`, which does not support NEP-18, yet
  - `equals`, `identical`: works (but no units / unit checking in `IndexVariable`)
  - `broadcast_like`: works (but no units / unit checking in `IndexVariable`)
  - `to_stacked_array`: no units in `IndexVariable`
  - `sel`, `loc`: no units in `IndexVariable`
  - `interp`, `reindex`: partially blocked by `IndexVariable`. `reindex` works with units in `data`, but `interp` uses `scipy`
  - `interp_like`, `reindex_like`: same as `interp` / `reindex`
  - `quantile`: works, but needs `pint` >= 0.12
  - `groupby_bins`: needs `pint` >= 0.12 (for `isclose`)
  - `rolling`: uses `numpy.lib.stride_tricks.as_strided`
  - `rolling_exp`: uses `numbagg` (supports NEP-18, but `pint` doesn't support its functions)


keewis:
it seems `assert_allclose` was one of the sources of `UnitStrippedWarning`s, but since there's a bug in `pint`'s `isclose` (fixed on `master`) the tests fail now.

Also, because it casted to `numpy`, there were a few bugs that were hidden.

dcherian:
Thanks for working on this @keewis . Since there are no changes outside `test_units.py`, I think you should merge whenever you think this is ready.

keewis:
I just found another issue: `pint` implements `prod` (but not yet `nanprod`) so the `prod` tests could be un-xfailed. However, we define a custom `nanprod` function that uses `where` to replace `nan` with `1`. This won't work on quantities since, unlike `nan` and `0`, a bare `1` cannot be put into quantities with a dimension (i.e. with a unit other than `dimensionless`).

I don't really understand the purpose of `nanprod`'s `min_count` parameter (and `_maybe_null_out`) so I'm not sure how to fix that.

For now, I think we can merge this PR and I'll add that issue to the list in #3594.

dcherian:
Looks like an incompatibility with latest pandas

```
____________________ TestDataset.test_resample[int-coords] _____________________

self = <xarray.tests.test_units.TestDataset object at 0x7fa84fb9e490>
variant = 'coords', dtype = <class 'int'>
    return func(*all_args, **all_kwargs)
xarray/core/common.py:1123: in resample
    grouper = pd.Grouper(

cls = <class 'pandas.core.groupby.grouper.Grouper'>, args = ()
kwargs = {'base': 0, 'closed': None, 'freq': '6m', 'label': None, ...}
TimeGrouper = <class 'pandas.core.resample.TimeGrouper'>, stacklevel = 2

    def __new__(cls, *args, **kwargs):
        if kwargs.get("freq") is not None:
            from pandas.core.resample import TimeGrouper
    
            # Deprecation warning of `base` and `loffset` since v1.1.0:
            # we are raising the warning here to be able to set the `stacklevel`
            # properly since we need to raise the `base` and `loffset` deprecation
            # warning from three different cases:
            #   core/generic.py::NDFrame.resample
            #   core/groupby/groupby.py::GroupBy.resample
            #   core/groupby/grouper.py::Grouper
            # raising these warnings from TimeGrouper directly would fail the test:
            #   tests/resample/test_deprecated.py::test_deprecating_on_loffset_and_base
    
            # hacky way to set the stacklevel: if cls is TimeGrouper it means
            # that the call comes from a pandas internal call of resample,
            # otherwise it comes from pd.Grouper
            stacklevel = 4 if cls is TimeGrouper else 2
            if kwargs.get("base", None) is not None:
>               warnings.warn(
                    "'base' in .resample() and in Grouper() is deprecated.\n"
                    "The new arguments that you should use are 'offset' or 'origin'.\n"
                    '\n>>> df.resample(freq="3s", base=2)\n'
                    "\nbecomes:\n"
                    '\n>>> df.resample(freq="3s", offset="2s")\n',
                    FutureWarning,
                    stacklevel=stacklevel,
                )
E               FutureWarning: 'base' in .resample() and in Grouper() is deprecated.
E               The new arguments that you should use are 'offset' or 'origin'.
E               
E               >>> df.resample(freq="3s", base=2)
E               
E               becomes:
E               
E               >>> df.resample(freq="3s", offset="2s")
```

keewis:
yeah, I don't know how to only filter for `pint` warnings, I tried `pytest.mark.filterwarnings("error:::pint[.*]")` but that doesn't work

PR #3978, opened 2020-04-16
use preformatted text for the code in the issue template

keewis:
Using backticks does not work in the summary, so we need to use html tags (i.e. `<tt>xr.show_versions()</tt>`) instead.

A demo:

```
...
```

</details>


shoyer:
thanks!

Issue #3361, opened 2019-10-01
Recommendations for domain-specific accessor documentation


gmaze:
Hi,

I'm currently working on an ocean domain specific accessor for a machine learning technique (https://github.com/gmaze/pyxpcm/tree/nfeatures).  
I implemented an xarray.DataSet accessor (name space ``pyxpcm``) that I'd like to document.

I thus wonder whether the xarray/pangeo community has recommendations about how to do this appropriately ?

Right now, I simply have an API auto-generated page, see:
https://pyxpcm-dev.readthedocs.io/en/latest/api.html#xarray-dataset-accessor-the-pyxpcm-name-space

jthielen:
@gmaze Just as an example, here is what we recently added for MetPy: https://unidata.github.io/MetPy/latest/api/generated/metpy.xarray.html. Previously, we just had a narrative-form tutorial.

If there are other recommendations, it would be great to incorporate those into MetPy as well!

gmaze:
Thanks for the example @jthielen, this looks great !

May be it would be better to not show to readers the accessor class name, since it will never be seen on the API frontend, only the scope name will.  

So it would be great to be able to have the documentation to read something like:  

`xarray.DataSet.<scope_name>.<accessor_method_or_property>`  

instead of:  

`<package_name>.xarray.<accessor_class_name>.<accessor_method_or_property>`
(in the case where the accessor class is in xarray.py file on the package)

But I don't know how to manage that with sphinx documentation

keewis:
comment from @benbovy in #3625:

> I re-used pandas templates for xarray accessors here: https://github.com/benbovy/xarray-simlab/blob/45359e99cbf6341464b02cb937618c051a58a31c/doc/conf.py#L233
> 
> It works pretty well: https://xarray-simlab.readthedocs.io/en/latest/api.html#dataset-xsimlab-xarray-accessor

For reference, here's pandas' version: https://github.com/pandas-dev/pandas/blob/837daf18d480cce18c25844c591c39da19437252/doc/source/conf.py#L441-L591

I think to make this a bit easier for everyone we should add this to the documentation on accessors.

Also, we could create a sphinx extension module that contains all the code in `conf.py` so defining the custom directives would be as simple as adding something like `"xarray.doc.accessors"` to `extensions`. The module doesn't have to be in `xarray`, though, it could also become a separate package.

benbovy:
> I think to make this a bit easier for everyone we should add this to the documentation on accessors.

[...]

PR #3988, opened 2020-04-21
built-in accessor documentation

### LINKED ISSUE #3625: Documentation for injected methods
At the moment, the documentation for methods introduced by accessors is broken. We work around this issue by directly referencing the corresponding accessor method or the underlying function. Of course, this is not how we recommend calling them, so we should try to fix this.

It seems this is mostly a sphinx issue since this is known (see #3333 and #3602), but we didn't find a way to resolve this. I'd like use this issue to collect ideas for solutions.

keewis:
We currently use `property` to attach our built-in accessors (`plot`, `dt` and `str`) to the `DataArray` and `Dataset` classes. However, because `property` returns itself when trying to access it as a class attribute, this does not work when generating the documentation (`sphinx` inspects classes, not class instances).

This adds a `property`-like descriptor that works on both classes and instance objects (the name could be more descriptive) and uses that to document the `Dataset.plot.*` and `DataArray.plot.*` methods (see the [rendered documentation](https://xarray-keewis.readthedocs.io/en/accessor-documentation/api.html)). I have not been able to get `sphinx` to work with `_PlotMethods.__slots__`, though.

A few questions / comments on this:
1. I noticed we have `DataArray.plot.__call__` and `xarray.plot.plot` but not `DataArray.plot.plot`. Is that something that is worth adding?

2. The functions decorated with the custom property define docstrings which currently are lost. Should we patch them on their return values?

3. Right now, the error message when accidentally trying to call e.g. `xr.DataArray.plot.line()` is not very helpful:
```pytb
AttributeError: 'NoneType' object has no attribute 'dims'
```

4. Now that we can document the accessors, we need to think about how to structure `api.rst`. For `plot` and `str`, we could just list methods / attributes in subsections of `DataArray` / `Dataset` (or keep the `Plotting` / `str` / ... sections and add class subsections?), but `dt` is more complicated since it dispatches for `datetime` and `timedelta`

 - [x] Closes #3625

Edit: err, it seems I used the branch I pushed to the main repository for a documentation preview for this...


xarray/core/utils.py
```diff
@@ -787,6 +787,14 @@ def drop_dims_from_indexers(
         )
 
 
+class property_:
+    def __init__(self, func):
```
dcherian, line 791: let's add a docstring describing why this is necessary and what its used for.

keewis: sure. Do you have a suggestion for a more descriptive name?

keewis:
there are a few of my questions / comments we can resolve by using something like
```python
class UncachedAccessor:
    def __init__(self, accessor):
        self._accessor = accessor

    def __get__(self, obj, cls):
        if obj is None:
            return self._accessor

        return self._accessor(obj)
```

The only disadvantage is that now `pydoc` (or `help()`) show the accessor as a class. That was already the case for `str` and `dt`, but `plot` was a function.

I think we could try to trick `inspect.getmembers` (which is used to decide which kind of output should be done), but that would require a lot of monkeypatching I'd rather not do.

`pandas` simply uses the same docstring for the class and `__call__`, which is not ideal but a lot simpler.

xarray/core/utils.py
dcherian: Not really...

But lets mention in the docstring that we do this specifically to make Sphinx work.

xarray/plot/plot.py
```diff
@@ -438,7 +438,7 @@ class _PlotMethods:
     For example, DataArray.plot.imshow
     """
 
-    __slots__ = ("_da",)
+    # __slots__ = ("_da",)
```
dcherian, line 441: cc @crusaderky

xarray/core/utils.py
keewis: that's not the only thing it enables: without it, tab-completion on the class does not work either

shoyer:
I think pandas might do something similar to do this? It might be worth checking, though it looks like you've figured most of it out already :)

shoyer:
This is a really clever yet simple fix. I like it!

A few ideas for testing:

- Verify that `DataArray.plot` returns the PlotMethods object.
- Verify that `data_array.plot.__doc__` is still there.

shoyer:
> there are a few of my questions / comments we can resolve by using something like
> 
> ```python
> class UncachedAccessor:
>     def __init__(self, accessor):
>         self._accessor = accessor
> 
>     def __get__(self, obj, cls):
>         if obj is None:
>             return self._accessor
> 
>         return self._accessor(obj)
> ```
> 
> The only disadvantage is that now `pydoc` (or `help()`) show the accessor as a class. That was already the case for `str` and `dt`, but `plot` was a function.
> 
> I think we could try to trick `inspect.getmembers` (which is used to decide which kind of output should be done), but that would require a lot of monkeypatching I'd rather not do.
> 
> `pandas` simply uses the same docstring for the class and `__call__`, which is not ideal but a lot simpler.

👍  for a simple solution here like pandas rather than using inspect. I think it's probably actually slightly more accurate to show the class.

keewis:
where should the docstring of the accessor come from? Right now, this is the docstring of the return value which seems reasonable to me.

I added tests verifying that `cls.property` returns the accessor class and `obj.property` the accessor instance. Do we need additional checks for `__doc__`?.

keewis:
I added the accessor templates from `pandas` (we should eventually outsource these to a [separate package](https://github.com/pydata/xarray/issues/3361#issuecomment-618338469) used by both `pandas` and `xarray`) and modified the [API page](http://xarray.pydata.org/en/accessor-documentation/api.html#plotting) to display the accessor version of the plotting functions.

I think we should add a accessor section to each of the main objects (`Dataset` and `DataArray`), where we either copy what `pandas` does – listing every method / attribute of the accessor as a subsection of the class documentation – or we could show them as attributes (just like what we do for the injected numpy functions) and generate separate pages for the accessors.

doc/api.rst
```diff
@@ -403,6 +403,56 @@ Computation
...
 
+
+String manipulation
+-------------------
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor_method.rst
+
+   DataArray.str.capitalize
```
dcherian, line 414: can we do this for the `.dt` accessor too?

keewis: sure, that's what I was planning to (and plotting, too, if that makes sense), and to add those to `Dataset`, too

doc/api.rst
```diff
@@ -403,6 +418,122 @@ Computation
 :py:attr:`~core.groupby.DataArrayGroupBy.where`
 :py:attr:`~core.groupby.DataArrayGroupBy.quantile`
 
+
+String manipulation
+-------------------
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor_method.rst
```
keewis, line 427: not sure if it's the template or the `UncachedAccessor` class, but somehow the signature loses the `self`

doc/api.rst
```diff
@@ -403,6 +418,122 @@ Computation
...
+   DataArray.str.zfill
+
+Datetimelike properties
+-----------------------
+
+**Datetime properties**:
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor_attribute.rst
```
keewis, line 477: not sure if we should use the template or if using `:py:attr:` as for the other attributes would be better

doc/api.rst
```diff
@@ -419,6 +550,33 @@ Reshaping and reorganizing
...
+   :toctree: generated/
+   :template: autosummary/accessor.rst
+
+   DataArray.plot
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor_callable.rst
+
+   DataArray.plot
```
keewis, line 566: I need to investigate this a bit more, I can't seem to find out how to get sphinx to document `DataArray.plot()` properly. Also, we should probably only have either `accessor_callable` or `accessor`, not both.

doc/api.rst
```diff
@@ -419,6 +550,33 @@ Reshaping and reorganizing
...
+   :toctree: generated/
+   :template: autosummary/accessor_method.rst
+
+   DataArray.plot.contourf
+   DataArray.plot.contour
+   DataArray.plot.hist
+   DataArray.plot.imshow
+   DataArray.plot.line
+   DataArray.plot.pcolormesh
+   DataArray.plot.step
```
keewis, line 578: here, too, the `self` is not mentioned anywhere which means it's not immediately obvious that these are methods

doc/api.rst
keewis: it seems that's the case for all methods, not just the accessors, so nothing to do here

doc/api.rst
keewis: adding `:hidden:` (or maybe moving this to `api-hidden.rst`) and using `:py:attr:` should work, so this becomes a matter of taste. Thoughts?

doc/api.rst
```diff
@@ -233,6 +233,21 @@ Reshaping and reorganizing
...
    Dataset.broadcast_like
 
+Plotting
+--------
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor.rst
+
+   Dataset.plot
```
dcherian, line 243: This should be removed. The `__call__` method just raises ValueError.
```suggestion
```

doc/api.rst
dcherian: `DataArray.plot` is duplicated above.

To me, the summary is a minor regression compared to the major improvement --- showing that the plotting methods can be used through the accessor interface.

doc/api.rst
dcherian: I like that you can see the methods in the left sidebar. AFAIK that's not possible with `:py:attr:`? The current version looks great to me.

doc/api.rst
keewis: yeah, sorry about that, I had those removed locally but forgot to push. The summary works for `pandas.Series.plot` so I've been trying to find out what they do differently

keewis: should we merge this and leave the summary (and possibly the conversion to `:py:attr:` if someone complains) to a new PR?

PR #3989, opened 2020-04-21
Fix distributed tests on upstream-dev

### LINKED ISSUE #3987: failing distributed tests on upstream-dev
Since dask/distributed#3706 `gen_cluster` requires functions decorated with it to be coroutines. We use it twice in `test_distributed.py`: on `test_async` and on `test_serializable_locks`, which both are pre-3.5 style coroutines (`yield` / `yield from` instead of `async` / `await`) but not marked as such (I'm not sure what `pytest` does with coroutines).

I think this should be a pretty strait-forward fix: replace `def` with `async def` and `yield` / `yield from` with `await`.

keewis:
- [x] Closes #3987

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -111,7 +111,9 @@ Internal Changes
 - Add a CI job that runs the tests with every optional dependency
   except ``dask``. (:issue:`3794`, :pull:`3919`)
   By `Justus Magin <https://github.com/keewis>`_.
-
+- Use ``async`` / ``await`` for the asynchronous distributed
+  tests. (:issue:`3987`, :pull:`3989`)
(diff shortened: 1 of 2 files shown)


crusaderky:
LGTM. I'm the author of the upstream change - apologies, I did not expect the gen_cluster decorator to be used in downstream projects.

keewis:
hmm... this was added in #1674 by @mrocklin. Maybe it should be part of a public "testing" API in `distributed`?

mrocklin:
Yeah, my sense here is that it probably makes sense to relax the assertion that only async `def functions` are supported.  Maybe we should warn for a few releases?

PR #3998, opened 2020-04-23
Fix handling of abbreviated units like msec

dopplershift:
By default, xarray tries to decode times with pandas and falls back to
cftime. This fixes the exception handler to fallback properly in the
cases an unhandled abbreviated unit is passed in.

An additional item here would be to add support for msec, etc. to xarray's handling, but I wasn't sure the best way to handle that. I'm happy just if things properly fall back to cftime.

 - [ ] Closes #xxxx
 - [x] Tests added


dopplershift:
cc @dcamron

dopplershift:
Ah, didn't realize bug fixes went in there too. And I *thought* I could get away without black, but I missed the quote style. All done.

shoyer:
Thanks!

Issue #3995, opened 2020-04-23
open_mfzarr files + intake-xarray


Mikejmnez:
This is related to a previous issue (#3668), although the actual problem on that issue is a bit more technically involved and is related with clusters... I decided to open this related issue so that the discussion in #3668 remains visible for other users. 

This issue is about the need to implement code that allows to read multiple zarr files. This can be particularly helpful when reading data through an Intake Catalog entry (Intake-xarray plugin), which can allow for a compact way to introduce parallelism when working with multiple zarr files. There are two steps for this, one to work at the xarray level (write a fuction that does that) and then write an option on the Intake-xarray plugin that can use such xarray functionality.

I am more than willing to work on such problem, if nobody's already working on this.

Issue #3668, opened 2020-01-07
open_mfdataset: support for multiple zarr datasets


dmedv:
I am running calculations on a remote Dask cluster. Some of the data is only available on the workers, not on the client. It is already possible to have an xarray dataset that "points" to a remote NetCDF data collection by using the `parallel` option with `xarray.open_mfdataset()` like this:

```python
from dask.distributed import Client
import xarray as xr

client = Client('<dask_scheduler_ip>:<port>')
ds = xr.open_mfdataset(remote_nc_file_paths, combine='by_coords', parallel=True)
```

Then it will use `dask.delayed` and, for example, the following simple mean calculation will be distributed between the workers, the result returned to the client:

```python
ds['Temp'].mean().compute()
``` 

Unfortunately, I cannot do the same thing with zarr, because `open_mfdataset()` does not support it, and `open_zarr()` does not have an option to utilize `dask.delayed`. Would it be possible to add `dask.delayed` support to the zarr backend? Or, perhaps, I am missing something, and there is another better way to work with zarr data on a remote Dask cluster? 

Output of `xr.show_versions()`:
```
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.7 |Anaconda custom (64-bit)| (default, Oct 23 2018, 19:16:44) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 3.10.0-862.2.3.el7.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None
libhdf5: 1.10.4
libnetcdf: 4.6.3

xarray: 0.14.1
pandas: 0.25.3
numpy: 1.17.3
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: None
h5py: 2.8.0
Nio: None
zarr: 2.3.2
cftime: 1.0.4.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.2.1
dask: 2.9.1
distributed: 2.9.1
matplotlib: 3.1.2
cartopy: None
seaborn: 0.9.0
numbagg: None
setuptools: 40.4.3
pip: 18.1
conda: 4.8.0
pytest: 3.8.2
IPython: 7.0.1
sphinx: 1.8.1
```

dcherian:
[...]

PR #4003, opened 2020-04-25
xarray.open_mzar: open multiple zarr files (in parallel)

Mikejmnez:
This is associated with #3995  and somewhat mentioned in #3668. This is, emulating ```xarray.open_mfdataset```, it allows xarray to read multiple zarr files from a glob or a list of files pointing to a zarr store. It follows the code for ```xarray.open_mfdataset``` and the description to the function is the same, as they do the same thing but for respective files type.


dcherian:
I think the better way to do this would be to add a kwarg to `open_dataset` that specifies the backend to use. For e.g. `xr.open_dataset(..., format="zarr")`. 

This would then delegate to `open_zarr` or a new `open_netcdf` or `open_rasterio` as appropriate. Then `open_mfdataset` would just work for all these formats without requiring duplicate code.

cc @pydata/xarray for thoughts.

TomNicholas:
+1 for having `open_dataset` and `open_mfdataset` as the main (ideally only) points of entry for users, which then delegate to different backend openers. That will keep the API neater, avoid duplicate code, and be easier to make into a completely general and extensible solution eventually.

Mikejmnez:
I like this approach (add capability to open_mfdataset to open multiple zarr files), as it is the easiest and cleanest. I considered it, and I am glad this is coming up because I wanted to know different opinions. Two things influenced my decision to have ```open_mzarr``` separate from ```open_mfdataset```:

1. Zarr stores are inherently different from netcdf-files, which becomes more evident when openning multiple files given a glob-path  (```paths='directory*/subdirectory*/*'```). zarr stores can potentially be recognized as directories rather than files (e.g. as opposed to ```paths='directory*/subdirectory*/*.nc'```). This distinction comes into play when, for example, trying to open files (```zarr``` vs ```netcdf```) through ```intake-xarray```. I know this an upstream behavior, but I think it needs to be considered and it is my end goal by allowing ```xarray``` to read multiple ```zarr``` files (in parallel) - To use ```intake-xarray``` to read them. The way to open files on ```intake-xarray``` (```zarr``` vs others) is again kept separate, and uses different functions. This is, 

**For netcdf**-files (```intake-xarray/netcdf.py```):
```python
url_path = fsspec.open_local(paths, *kwargs) 
```
which can interpret a glob path. Then ```url``` is then passed to ```xarray.open_mfdataset```

**zarr** files (```intake-xarray/xzarr```):
```python
url_path = fsspec.mapper(paths, *kwargs) 
```
```fsspec.mapper``` does not recognize glob-paths, and ```fspec.open_local```, which does recognize globs, cannot detect zarr-stores (as these are recognized as directories rather than files with a known extension). See an issue I created about such behavior https://github.com/intake/filesystem_spec/issues/286#issue-606019293 (apologizes, I am new at github and don't know if this is the correct way to link issues across repositories)


2. Zarr continues to be under development, and the behavior of zarr it appears will rely heavily on ```fsspec``` more in the future. I wonder if such future development is the reason why even on ```xarray```, ```open_zarr``` is contained in a different file from ```open_mfdataset```, a similar behavior also happening in ```intake-xarray```.


I am extremely interested what people think about ```xarray``` and ```intake-xarray``` compatibility/development, when it comes with ```zarr``` files being read in parallel...

**Comment by TomNicholas** [comment]
@rabernat and @jhamman I expect you will want to weigh in on how best to handle this for zarr

**Comment by martindurant** [comment]
>  the behavior of zarr it appears will rely heavily on fsspec more in the future.

**IF** we can push on https://github.com/zarr-developers/zarr-python/pull/546 ; but here is also an opportunity to get the behaviour out of the zarr/fsspec interaction most convenient for this work.

File: xarray/backends/zarr.py [hunk 1]
```diff
@@ -692,3 +700,233 @@ def maybe_chunk(name, var, chunks):
...
+    open_dataset
+    References
+    ----------
+    .. [1] http://xarray.pydata.org/en/stable/dask.html
+    .. [2] http://xarray.pydata.org/en/stable/dask.html#chunking-and-performance
+    """
+    if isinstance(paths, str):
+        if is_remote_uri(paths):
+            raise ValueError(
+                "cannot do wild-card matching for paths that are remote URLs: "
```
martindurant, line 844: In theory, you should be able to do this. The trouble is matching directories? It may be worthwhile writing out how this ought to work and making a test case for fsspec. 

The code here could maybe also be gathered into some "open_mappers" function in fsspec, which parses the glob and opens the mappers. I'm not sure how useful that idea is, maybe its better to keep the code here, likely the only place it will get used.

jhamman:
+1 on deprecating `open_zarr` and moving to `open_dataset(..., engine='zarr')

I also agree that globing zarr stores is a tricky nut to crack. For the sake of simplicity, I'd like to suggest handling this functionaly as separate PRs. Given the heterogeneity in zarr storage options, I'm not sure its practical to support this behavior within Xarray but I'd be happy to discuss that in a separate issue.

rabernat:
I agree with everything Joe said. I'm fine with getting a `NotImplemented` error if I try to glob with zarr and open_mfdataset.

xarray/backends/zarr.py
Mikejmnez: An idea is to have (on ```intake-xarray```) a conditional that **if** paths is glob, ```intake-xarray``` passes the glob path as is to ```open_mzarr``` (i.e. not use ```open_mapper```). As written here, ```xarray.open_mzarr``` can sort the glob and open the zarr stores. However, we could make use of ```open_mappers``` corresponding to entries on the sorted ```paths``` list (line 851), to open the zarr stores. Potentially modifying how ```open_kwargs`` (line 870) and ```datasets``` (886) are defined in order to make use of ```fsspec.open_mappers``` .

A question that comes to my mind (as I am new at this), is what is the advantage of employing ```open_mapper``` to open a zarr store? That is, as opposed to provide the location of the store to ```open_zarr``` the way it is currently defined on ```datasets``` (line 886), which does work... performance?

dcherian:
IMO we should support
```
zarr-store-1/
zarr-store-2/
zarr-file-store-3
```
but raise `NotImplementedError` for

```
zarr-store-1/
subdir/zarr-store-2
```

I don't know whether it would be easy to detect the `glob` pattern for the second example in all cases though.

Mikejmnez:
Following your advise, ```open_dataset``` can now open ``zarr`` files. This is done:
```python
ds = xarray.open_dataset(store, engine="zarr", chunks="auto")
```

NOTE:  ``xr.open_dataset`` has ``chunks=None`` by default,  whereas it used to be ``chunks="auto"`` on ``xarray.open_zarr``. 

**Additional feature**: As a result of these changes, ```open_mfdataset`` can now (automatically) open multiple zarr files (e.g. in parallel) when given a glob. This is, 
```python
paths='directory*/*/subdirectory/*'
ds = xarray.open_mfdataset(paths, enginne="zarr", chunks="auto", concat_dim="time", combine="nested")
```
does yield the desired behavior. 

This is different from  ```fsspec.open_local``` vs ```fsspec.mapper``` on ``intake-xarray`` when opening files with a glob. But agreed, that can be addressed in a different PR.

xarray/backends/api.py
```diff
@@ -383,6 +384,10 @@ def open_dataset(
         represented using ``np.datetime64[ns]`` objects.  If False, always
         decode times to ``np.datetime64[ns]`` objects; if this is not possible
         raise an error.
+    overwrite_encoded_chunks: bool, optional
+        Whether to drop the zarr chunks encoded for each variable when a
+        dataset is loaded with specified chunk sizes (default: False)
```
jhamman, line 389: Since this only applies to the zarr backend, would it make sense to pass this through `backend_kwargs`?

xarray/backends/api.py
```diff
@@ -542,7 +527,99 @@ def maybe_decode_store(store, lock=False):
...
+                                    "dimension %r. This significantly "
+                                    "degrades performance. Consider "
+                                    "rechunking after loading instead."
+                                    % (chunks[dim], chunk_spec[dim], dim),
+                                    stacklevel=2,
+                                )
+                        chunk_spec[dim] = chunks[dim]
+                return chunk_spec
+
+            def maybe_chunk(name, var, chunks):
```
jhamman, line 602: I think I have a slight preference to relocate `maybe_chunk` and `get_chunk` to the `zarr.py` module. Just skimming here but it doesn't seem like they are using local function state.

xarray/backends/api.py
```diff
@@ -542,7 +527,99 @@ def maybe_decode_store(store, lock=False):
...
+                if chunks != "auto":
+                    raise ValueError(
+                        "chunks must be an int, dict, 'auto', or None. "
+                        "Instead found %s. " % chunks
+                    )
+            if isinstance(chunks, int):
+                chunks = dict.fromkeys(ds.dims, chunks)
+
+            if isinstance(chunks, tuple) and len(chunks) == len(ds.dims):
+                chunks = dict(zip(ds.dims, chunks))
```
jhamman, line 571: I don't think this line is reachable. How could chunks ever be a tuple when line 561 checks against `(int, dict)`?

xarray/backends/api.py
```diff
@@ -542,7 +527,99 @@ def maybe_decode_store(store, lock=False):
...
+                else:
+                    return var
+
+            variables = {k: maybe_chunk(k, v, chunks) for k, v in ds.variables.items()}
+            ds2 = ds._replace_vars_and_dims(variables)
+        return ds2
+    else:
+        ds2 = ds
+
+    return ds2
```
jhamman, line 622: since we return in the block above, I think it would be okay to also return in line 620.

xarray/backends/api.py
Mikejmnez: You are right, it does make sense to not make it an argument for ``open_dataset``, and pass it to ``open_dataset`` through ``backend_kwargs`` from depricated ``open_zarr`` (i.e. zarr.py). However, it is not an argument for ``ZarrStore.open_group``, instead it is used within ``def maybe_chunk``. I had to pop-it and use it before passing ``backend_kwargs`` to ``open_group`` on zarr.py

xarray/backends/api.py
Mikejmnez: It does make sense, and would make the code cleaner/easier to read and/or debug. You think it should be implemented within ``ZarrStore.open_group``?

xarray/backends/api.py
Mikejmnez: yeah, doesn't look like is reachable. I'll remove it.

xarray/backends/api.py
Mikejmnez: This part I am not so sure I completely understand. ``ds2`` gets created if ``chunk`` is not ``None``. If ``chunk is None`` then ``ds`` is passed, so we need ``ds2=ds``.

This brings to a question: Can we just have
```python
ds = ds.chunk(chunks, ...)
```
instead of 
```python 
ds2 = ds.chunk(chunks,...)
ds2._file_obj = ds._file_obj
```
that way I could just return ``ds`` on line ``620``

xarray/backends/zarr.py
```diff
@@ -356,6 +358,48 @@ def encode_variable(self, variable):
     def encode_attribute(self, a):
         return encode_zarr_attr_value(a)
 
+
+    def get_chunk(name, var, chunks):
```
weiji14, line 362:
```suggestion
    def get_chunk(self, name, var, chunks):
```

xarray/backends/zarr.py
```diff
@@ -356,6 +358,48 @@ def encode_variable(self, variable):
...
+                            "dimension %r. This significantly "
+                            "degrades performance. Consider "
+                            "rechunking after loading instead."
+                            % (chunks[dim], chunk_spec[dim], dim),
+                            stacklevel=2,
+                        )
+                chunk_spec[dim] = chunks[dim]
+        return chunk_spec
+
+    def maybe_chunk(name, var, chunks):
```
weiji14, line 391:
```suggestion
    def maybe_chunk(self, name, var, chunks, overwrite_encoded_chunks):
```

xarray/backends/zarr.py
```diff
@@ -356,6 +358,48 @@ def encode_variable(self, variable):
...
+        chunk_spec = get_chunk(name, var, chunks)
+
+        if (var.ndim > 0) and (chunk_spec is not None):
+            # does this cause any data to be read?
+            token2 = tokenize(name, var._data)
```
weiji14, line 396:
```suggestion
        if (var.ndim > 0) and (chunk_spec is not None):
            from dask.base import tokenize

            # does this cause any data to be read?
            token2 = tokenize(name, var._data)
```

xarray/backends/api.py
```diff
@@ -455,43 +462,25 @@ def maybe_decode_store(store, lock=False):
...
+        return ds
 
     if isinstance(filename_or_obj, Path):
         filename_or_obj = str(filename_or_obj)
 
     if isinstance(filename_or_obj, AbstractDataStore):
         store = filename_or_obj
 
+    if isinstance(filename_or_obj, MutableMapping):
+        if engine == 'zarr':
```
shoyer, line 474: These conditions should be combined into a single `if` -- otherwise somebody could provided `filename_or_obj` as a mutable mapping with a different engine, which could result in undefined behavior.

xarray/backends/api.py
```diff
@@ -455,43 +462,25 @@ def maybe_decode_store(store, lock=False):
...
+            # on ZarrStore, mode='r', synchronizer=None, group=None,
+            # consolidated=False.
+            overwrite_encoded_chunks = backend_kwargs.pop("overwrite_encoded_chunks", None)
```
shoyer, line 477: `pop` modifies a dict inplace. It's best to make a copy of `backend_kwargs` first to ensure that objects users supply as arguments into xarray functions aren't unexpected modified.

xarray/backends/api.py
shoyer: Could we add `overwrite_encoded_chunks` as a new argument in `ZarrStore.open_group`? That seems preferable to adding another highly backend specific method on to `open_dataset`.

(To be honest, I'm not entirely sure why we need `overwrite_encoded_chunks`... it's not obvious from the description why it's useful)

weiji14:
@Mikejmnez, do you mind if I pick up working on this branch? I'd be really keen to see it get into xarray 0.16, and then it will be possible to resolve the intake-xarray issue at https://github.com/intake/intake-xarray/issues/70. Not sure if it's possible to get commit access here, or if I should just submit a PR to your fork, or maybe there's a better way?

xarray/backends/api.py
weiji14: Quoting from https://github.com/pydata/xarray/pull/2530#issue-227407536, where `overwrite_encoded_chunks` was initially added.

>  It seems reasonable that anyone manually specifying chunks may want to rewrite the dataset in those chunks, and the error that arises when the encoded Zarr chunks mismatch the variable Dask chunks may quickly get annoying. `overwrite_encoded_chunks=True` sets the encoded chunks to None so there is no clash.

As mentioned by @Mikejmnez , `overwrite_encoded_chunks` is not actually used by `ZarrStore.open_group`, it is used at the `open_dataset` level. So it's not strictly a 'backend_kwarg', but we can still put it in there I suppose? Just looking through the code now to see how best to do it.

shoyer:
@weiji14 could you kindly reopen your new pull request against the main xarray repository? Your pull request is currently in Mikejmnez/xarray

weiji14:
Sure, I can move it, but I just wanted to make sure @Mikejmnez gets the credit for this PR.

shoyer:
> Sure, I can move it, but I just wanted to make sure @Mikejmnez gets the credit for this PR

Yes, absolutely! As long as you preserve his original commits and add yours on top of them, both of you will be credited in the Git history. If you're writing a release note in `whats-new.rst` about the feature, please include both of your names in the credits.

Mikejmnez:
@weiji14 @shoyer Thanks you guys! Sorry it has taken me long to come back to this PR -  I really mean to come back to this but I got stuck with another bigger PR that is actually part of my main research project. Anyways, much appreciated for the help, cheers!!

- Since I am a novice at this, on my end, should I close this PR?

keewis:
> Since I am a novice at this, on my end, should I close this PR?

don't worry about that: we can close this PR when we merge #4187

PR #4006, opened 2020-04-25
fix dangerous default arguments

### LINKED ISSUE #4002: Remove dangerous default argument
Hi,
It is good not to use the default arguments while calling a function.

#### MCVE Code Sample

In file `xarray/tests/test_conventions.py` 
```python
# Your code here
329    @contextlib.contextmanager
330    def roundtrip(
331        self, data, save_kwargs={}, open_kwargs={}, allow_cleanup_failure=False332    ):
333        store = CFEncodedInMemoryStore()
334        data.dump_to_store(store, **save_kwargs)
```

#### Problem Description

Default arguments in Python are evaluated once when the function definition is executed. This means that the expression is evaluated only once when the function is defined and that the same value is used for each subsequent call. So if you are modifying the mutable default argument — a list, dict, etc., it will be modified for all future calls.

pnijhara:
- [x] Closes #4002 


xarray/tests/test_backends.py
```diff
@@ -3621,11 +3621,15 @@ def create_tmp_geotiff(
     ny=3,
     nz=3,
     transform=None,
-    transform_args=[5000, 80000, 1000, 2000.0],
-    crs={"units": "m", "no_defs": True, "ellps": "WGS84", "proj": "utm", "zone": 18},
+    transform_args=None,
+    crs=None,
```
keewis, line 3625: unfortunately, since the default values for parameters to `rasterio.open` are mostly `None` we can't use it as default value: otherwise, how would we explicitly request `rasterio.open`'s default?

We might fix this by using a different value, e.g. a `object` instance:
```python
_default_value = object()
```

pnijhara: you mean by replacing None with object() should fix this?

pnijhara: @keewis Could you please take a look at other values also and suggest changes?

keewis: no, what I meant was that we define `default_value` somewhere outside the function (at module level):
```python
default_value = object()
```
and then use that everywhere you have been using `None`:
```python
    transform_args=default_value,
    crs=default_value,
```
and later:
```python
if transform_args is default_value:
    transform_args = [...]

if crs is default_value:
    crs = {...}
```

the parameters of `roundtrip` are only used as kwargs, so they're fine

pnijhara:
@keewis are there any other changes required?

max-sixty:
Sorry for the delay @pnijhara ! Merging now.

Thanks again, we're happy to have you as a contributor!

pnijhara:
Thank you!

PR #4007, opened 2020-04-25
Allow DataArray.to_series() without invoking sparse.COO.todense()

khaeru:
**Superseded by #11528**

This adds some code (from iiasa/ixmp#317) that allows DataArray.to_series() to be called without invoking sparse.COO.todense() when that is the backing data type.

I'm aware this needs some improvement to meet the standard of the existing codebase, so I hope I could ask for some guidance on how to address the following points (including whom to ask about them):
- [ ] Make the same improvement in {DataArray,Dataset}.to_dataframe().
- [ ] Possibly move the code out of dataarray.py to a more appropriate location (where?).
- [ ] Possibly check for sparse.COO explicitly instead of xarray.core.pycompat.sparse_array_type. Other SparseArray subclasses, e.g. DOK, may not have the same attributes.

Standard items:
 - [ ] Tests added.
   (Sort of: these wanted to modify 7 files beyond the one I touched; didn't commit these changes.)
 - [ ] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API.

### DIFF
--- a/xarray/core/dataarray.py
+++ b/xarray/core/dataarray.py
@@ -56,6 +56,7 @@ from .indexes import Indexes, default_indexes, propagate_indexes
 from .indexing import is_fancy_indexer
 from .merge import PANDAS_TYPES, _extract_indexes_from_coords
 from .options import OPTIONS
+from .pycompat import sparse_array_type
 from .utils import Default, ReprObject, _check_inplace, _default, either_dict_or_kwargs
 from .variable import (
     IndexVariable,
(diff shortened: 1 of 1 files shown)


dcherian:
Thanks @khaeru 

> Make the same improvement in {DataArray,Dataset}.to_dataframe().

I think only `Dataset._to_dataframe` is needed since `DataArray.to_dataframe()` calls that. 

> Possibly move the code out of dataarray.py to a more appropriate location (where?).

Perhaps `to_series` could be refactored to use `_to_dataframe` too?

> Possibly check for sparse.COO explicitly instead of xarray.core.pycompat.sparse_array_type. Other SparseArray subclasses, e.g. DOK, may not have the same attributes.

:+1: Should add something to the docstring once you figure this out.

This PR also needs tests.

patnr:
I believe there is an error in the proposed code: `set_levels()` should not be given the label values of each data point, but rather [unique](https://pandas.pydata.org/docs/reference/api/pandas.MultiIndex.levels.html) labels, with their order determined by the preceding `from_arrays()` call. This can be achieved using

```python
.set_levels(
        [
            [labels.values[i] for i in ints]
            for (labels, ints) in zip(array.coords.values(), index.levels)
        ]
    )
 ```

khaeru:
Closing, as this was superseded by #11528.

PR #4012, opened 2020-04-27
Apply blackdoc to the documentation

keewis:
This applies `black` to the code in the documentation using [`blackdoc`](https://github.com/keewis/blackdoc). The tool itself is not stable or released yet, but this demonstrate the upcoming support for `rst` code blocks, including the ipython sphinx directive. There are a few issues with the latter, e.g. the semi-colon to suppress output will be removed by `black`. In a few cases this changes the other outputs, mostly plots. I think we might fix/work around that by moving the line decorator to the last line that should be plotted (which is what I did), but we could also use the ipython syntax with cell decorators:
```python
@savefig ...
In [1]: # code
   ...: # with multiple lines
```

Also, while I checked the changes manually, please don't take my word for it: I'm really not sure I didn't miss anything. I know this is a rather large PR so I'm fine with waiting.

### DIFF
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -195,7 +195,8 @@ We can also manually iterate through ``Rolling`` objects:
 .. code:: python
 
    for label, arr_window in r:
-      # arr_window is a view of x
+       # arr_window is a view of x
+       ...
 
 .. _comput.rolling_exp:
 
--- a/doc/computation.rst
+++ b/doc/computation.rst
@@ -18,17 +18,19 @@ Arithmetic operations with a single DataArray automatically vectorize (like
 numpy) over all array values:
 
 .. ipython:: python
-   :suppress:
+    :suppress:
 
     import numpy as np
     import pandas as pd
     import xarray as xr
+
     np.random.seed(123456)
 
 .. ipython:: python
 
-    arr = xr.DataArray(np.random.RandomState(0).randn(2, 3),
-                       [('x', ['a', 'b']), ('y', [10, 20, 30])])
+    arr = xr.DataArray(
+        np.random.RandomState(0).randn(2, 3), [("x", ["a", "b"]), ("y", [10, 20, 30])]
+    )
     arr - 3
     abs(arr)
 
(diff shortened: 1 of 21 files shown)


crusaderky:
LGTM - can't wait for the tool to mature!

doc/computation.rst
```diff
@@ -254,52 +259,52 @@ support weighted ``sum`` and weighted ``mean``.
...
-  weighted_prec
+    weighted_prec = prec.weighted(weights)
+    weighted_prec
 
 Calculate the weighted sum:
 
 .. ipython:: python
 
-  weighted_prec.sum()
+    weighted_prec.sum()
```
crusaderky, line 278: The RST reference (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html) dictates a 3-spaces indentation for multi-line .. blocks. However I appreciate the frustration in not being to use the Tab key. Was it a conscious decision?

crusaderky:
I went through everything and it seems all fine.
I'm happy to merge as soon as you add a line to the What's New

doc/computation.rst
keewis: no, not really. I was using 4 spaces since that's the indent in the python code. Actually, the tool only makes sure there is a consistent indentation depth (it enforces the indentation of the code).

If you look at the reference, though, the 3-spaces indentation is only implicitly mentioned in the syntax diagram:
```
+-------+-------------------------------+
| ".. " | directive type "::" directive |
+-------+ block                         |
        |                               |
        +-------------------------------+
```
but not in the text, it only says:
> The directive block [...] consists of any text on the first line of the directive after the directive marker, and any subsequent indented text.

So I'd assume 3 spaces is not mandatory.

keewis:
done

dcherian:
Thanks @keewis & @crusaderky 

> I went through everything and it seems all fine.

Wow!

DocOtak:
I really like what it did to some of the long `xr.Dataset()` and `xr.DataArray()` calls. Very readable!

max-sixty:
Great job @keewis !

PR #4017, opened 2020-04-30
Combining attrs of member DataArrays of Datasets

### LINKED ISSUE #4009: Incoherencies between docs in open_mfdataset and combine_by_coords and its behaviour.
PR #3877 adds nice control over the attrs of the ouput, but there are some incoherencies in the docs and the behaviour that break previously fine code.

#### MCVE Code Sample

```python
import xarray as xr
out = xr.open_mfdataset('/files/with/*_conflicting_attrs.nc', combine='by_coords')
```
#### Expected Output
`out` having the attributes from the first file in the sorted glob list.

#### Problem Description

Fails with a `MergeError` .

In the doc of  `open_mfdataset` it is said:
```
    attrs_file : str or pathlib.Path, optional
        Path of the file used to read global attributes from.
        By default global attributes are read from the first file provided,
        with wildcard matches sorted by filename.
```
But in the code, `open_mfdataset` calls `combine_by_coords` without specifying its `combine_attrs` argument, which defaults to 'no_conflicts', instead of the expected 'override' or 'drop'. The attributes are anyway managed by `open_mfdataset` further down, but in the case of conflicts the code never reaches that point.

Also, in the doc of `combine_by_coords` the wrong default is specified:
```
    combine_attrs : {'drop', 'identical', 'no_conflicts', 'override'},
                    default 'drop'
        String indicating how to combine attrs of the objects being merged:

        - 'drop': empty attrs on returned Dataset.
        - 'identical': all attrs must be the same on every object.
        - 'no_conflicts': attrs from all objects are combined, any that have
          the same name must also have the same value.
        - 'override': skip comparing and copy attrs from the first dataset to
          the result.
```

I think we expect either `combine_by_coords` to have 'drop' as the default or `open_mfdataset` to pass `combine_attrs='drop'`.

#### Versions

INSTALLED VERSIONS
------------------
commit: None
python: 3.8.2 | packaged by conda-forge | (default, Apr 24 2020, 08:20:52) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 5.6.7-arch1-1
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: fr_CA.utf8
LOCALE: fr_CA.UTF-8
libhdf5: 1.10.6
libnetcdf: 4.7.4

xarray: 0.15.2.dev29+g7eeba59f
pandas: 1.0.3
numpy: 1.18.1
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.1.1.2
nc_time_axis: 1.2.0
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.3.2
dask: 2.14.0
distributed: 2.14.0
matplotlib: 3.2.1
cartopy: None
seaborn: None
numbagg: None
setuptools: 46.1.3.post20200325
pip: 20.0.2
conda: None
pytest: 5.4.1
IPython: 7.13.0
sphinx: 3.0.2

</details>

johnomotani:
While looking at #4009, I noticed that the `combine_attrs` kwarg to `concat` or `merge` only affects the global attrs of the `Dataset`, not the attrs of `DataArray` or `Variable` members of the `Dataset`s being combined. I think this is a bug.

So far this PR adds tests that reproduce the issue in #4009, and the issue described above. Fixing should be fairly simple: for #4009 pass `combine_attrs="drop"` to `combine_by_coords` and `combine_nested` in `open_mfdataset`; for this issue insert the `combine_attrs` handling in an appropriate place - possibly in `merge.unique_variable`. I'll update with fixes when I get a chance.

 - [ ] Closes #4009
 - [x] Tests added

### DIFF
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -2654,7 +2654,7 @@ class TestOpenMFDatasetWithDataVarsAndCoordsKw:
 
     def test_open_mfdataset_dataset_attr_by_coords(self):
         """
-        Case when an attribute of type list differs across the multiple files
+        Case when an attribute differs across the multiple files
         """
         with self.setup_files_and_datasets() as (files, [ds1, ds2]):
             # Give the files an inconsistent attribute
@@ -2669,7 +2669,7 @@ class TestOpenMFDatasetWithDataVarsAndCoordsKw:
 
     def test_open_mfdataset_dataarray_attr_by_coords(self):
         """
-        Case when an attribute of type list differs across the multiple files
+        Case when an attribute of a member DataArray differs across the multiple files
         """
         with self.setup_files_and_datasets() as (files, [ds1, ds2]):
             # Give the files an inconsistent attribute
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -2652,6 +2652,36 @@ class TestOpenMFDatasetWithDataVarsAndCoordsKw:
                 ds_expect = xr.concat([ds1, ds2], data_vars=opt, dim="t", join=join)
                 assert_identical(ds, ds_expect)
 
+    def test_open_mfdataset_dataset_attr_by_coords(self):
+        """
+        Case when an attribute of type list differs across the multiple files
+        """
+        with self.setup_files_and_datasets() as (files, [ds1, ds2]):
+            # Give the files an inconsistent attribute
+            for i, f in enumerate(files):
+                ds = open_dataset(f).load()
+                ds.attrs["test_dataset_attr"] = 10 + i
+                ds.close()
+                ds.to_netcdf(f)
+
+            with xr.open_mfdataset(files, combine="by_coords", concat_dim="t") as ds:
+                assert ds.test_dataset_attr == 10
+
+    def test_open_mfdataset_dataarray_attr_by_coords(self):
+        """
+        Case when an attribute of type list differs across the multiple files
+        """
+        with self.setup_files_and_datasets() as (files, [ds1, ds2]):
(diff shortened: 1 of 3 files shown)


xarray/tests/test_backends.py
```diff
@@ -2652,6 +2652,36 @@ def test_open_mfdataset_does_same_as_concat(self, combine, opt, join):
...
+        with self.setup_files_and_datasets() as (files, [ds1, ds2]):
+            # Give the files an inconsistent attribute
+            for i, f in enumerate(files):
+                ds = open_dataset(f).load()
+                ds["v1"].attrs["test_dataarray_attr"] = i
+                ds.close()
+                ds.to_netcdf(f)
+
+            with xr.open_mfdataset(files, combine="by_coords", concat_dim="t") as ds:
+                assert ds["v1"].test_dataarray_attr == 0
```
johnomotani, line 2683: @TomNicholas do you think we should make this test pass (i.e. support combining `attrs` on DataArrays contained in the Dataset being loaded)?

Proposal for a fix:
* Support `open_mfdataset`-style attribute combining by upgrading the `combine_attrs` argument to `combine_nested`/`combine_by_coords` (and probably the other functions that have a `combine_attrs` argument) so it can be passed one of the `Dataset`s or `DataArray`s being combined/merged, and will then take the `attrs` from that.
* In `open_mfdataset` use the `attrs_file` argument to get the `Dataset` to take the `attrs` from, and pass that to the `combine_attrs` argument of `combine_nested`/`combine_by_coords`.

PR #4018, opened 2020-05-01
Add NetCDF3 dtype coercion for unsigned integer types

### LINKED ISSUE #4014: Add NetCDF3 dtype coercion for unsigned integer types
`xr.Dataset.to_netcdf` does not seem to support writing data with unsigned integer dtypes, `uint32`, `uint64` etc.. This seems to be the case for both scipy-based output formats, `NETCDF3_CLASSIC` and `NETCDF3_64BIT`.

It seems like dtype coercions for `int64` and `bool` are done automatically for NetCDF3 in the [`xarray.netcdf3`](https://github.com/pydata/xarray/blob/33a66d6380c26a59923922ee11e8ffcf0b4f379f/xarray/backends/netcdf3.py#L31) module.
Shouldn't data of signed dtypes then *also* be coerced, i.e. to their signed equivalent?

#### MCVE Code Sample

```python
import numpy as np
import xarray as xr

da = xr.DataArray(np.array([1,2,3], dtype='uint64'))

# The following all fail:
da.to_netcdf("foo")  # default format: scipy NETCDF3_CLASSIC
da.to_netcdf("bar", format='NETCDF3_64BIT')
da.astype('uint32').to_netcdf("baz")
da.astype('uint16').to_netcdf("spam")

# This works:
da.astype('int64').to_netcdf("working64")  # is coerced
da.astype('int32').to_netcdf("working32")  # works natively
```

*Importantly,* this is with the netcdf4 python package **not** being installed, in which case that package would be used for writing rather than scipy's netcdf.

#### Expected Output
NetCDF3 file is written with an appropriately coerced data format, e.g. as done with `int64`.

Alternatively, writing data fails for _all_ dtypes that would natively be unsupported, including `int64` and `bool`.

#### Problem Description
Given that the infrastructure for coercion is [already in place](https://github.com/pydata/xarray/blob/33a66d6380c26a59923922ee11e8ffcf0b4f379f/xarray/backends/netcdf3.py#L37-L57), it seems more consistent to me to apply coercion to all cases where it would lead to `to_netcdf` method calls _succeeding_ rather than failing, not only to `int64` and `bool`.

Ideally, coercion would happen towards another unsigned integer type.
However, writing `uint32` seems not to be possible, so it's not a 64bit/32bit issue.
While the [NetCDF Format Specification](https://www.unidata.ucar.edu/software/netcdf/docs/file_format_specifications.html#classic_format_spec) declares as only unsigned integer type `NON_NEG`, which I presume to be equivalent to `uint32`, writing unsigned integers seems not possible via scipy's NetCDF3 writer. Thus, the only viable coercion, as far as I see, would be to signed equivalents.

I see no new cast safety implications here, because the existing `coerce_nc3_dtype` function already checks if the original and cast arrays compare equivalently.

#### Versions

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.6 (default, Dec 30 2019, 19:38:26) 
[Clang 11.0.0 (clang-1100.0.33.16)]
python-bits: 64
OS: Darwin
OS-release: 19.3.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: None

xarray: 0.15.1
pandas: 1.0.1
numpy: 1.18.3
scipy: 1.4.1
netCDF4: None
pydap: None
h5netcdf: None
h5py: 2.10.0
Nio: None
zarr: None
cftime: 1.0.4.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.10.1
distributed: 2.10.0
matplotlib: 3.1.3
cartopy: None
seaborn: None
numbagg: None
setuptools: 45.1.0
pip: 20.0.2
conda: None
pytest: 5.3.5
IPython: 7.12.0
sphinx: 2.4.1

</details>

blsqr:
This PR addresses #4014. 

 - [x] Closes #4014 
 - [x] Tests added or adjusted
   - [x] Adjusted a test case where `int64` coercion played a role
   - [x] Explicit test for failing coercion (seems to be missing currently -- should this be added?)
 - [x] Fully documented
 - [x] Adjusted `whats-new.rst` for all changes ~~and `api.rst` for new API~~

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -59,6 +59,9 @@ New Features
   the :py:class:`~core.accessor_dt.DatetimeAccessor` (:pull:`3935`).  This
   feature requires cftime version 1.1.0 or greater.  By
   `Spencer Clark <https://github.com/spencerkclark>`_.
+- For the netCDF3 backend, added dtype coercions for unsigned integer types.
+  (:issue:`4014`, :pull:`4018`)
+  By `Yunus Sevinchan <https://github.com/blsqr>`_
 
 Bug fixes
 ~~~~~~~~~
--- a/xarray/backends/netcdf3.py
+++ b/xarray/backends/netcdf3.py
@@ -28,7 +28,14 @@ _reserved_names = {
 
 # These data-types aren't supported by netCDF3, so they are automatically
 # coerced instead as indicated by the "coerce_nc3_dtype" function
-_nc3_dtype_coercions = {"int64": "int32", "bool": "int8"}
+_nc3_dtype_coercions = {
+    "int64": "int32",
+    "uint64": "int32",
+    "uint32": "int32",
+    "uint16": "int16",
+    "uint8": "int8",
+    "bool": "int8",
+}
 
 # encode all strings as UTF-8
 STRING_ENCODING = "utf-8"
@@ -37,12 +44,17 @@ STRING_ENCODING = "utf-8"
 def coerce_nc3_dtype(arr):
     """Coerce an array to a data type that can be stored in a netCDF-3 file
 
-    This function performs the following dtype conversions:
-        int64 -> int32
-        bool -> int8
+    This function performs the dtype conversions as specified by the
+    ``_nc3_dtype_coercions`` mapping:
+        int64  -> int32
+        uint64 -> int32
+        uint32 -> int32
+        uint16 -> int16
+        uint8  -> int8
+        bool   -> int8
 
-    Data is checked for equality, or equivalence (non-NaN values) with
-    `np.allclose` with the default keyword arguments.
+    Data is checked for equality, or equivalence (non-NaN values) using the
+    ``(cast_array == original_array).all()``.
     """
     dtype = str(arr.dtype)
     if dtype in _nc3_dtype_coercions:
--- a/xarray/tests/test_backends.py
+++ b/xarray/tests/test_backends.py
@@ -228,7 +228,27 @@ class TestCommon:
 
 
 class NetCDF3Only:
-    pass
+    netcdf3_formats = ("NETCDF3_CLASSIC", "NETCDF3_64BIT")
+
+    @requires_scipy
+    def test_dtype_coercion_error(self):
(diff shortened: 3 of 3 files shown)


blsqr:
Currently, the error that is raised by `coerce_nc3_dtype` is not covered by tests.

A test could look something like this:

```python
    def test_dtype_coercion_error(self):
        for dtype in _nc3_dtype_coercions:
            if dtype == "bool":
                continue

            # Create some data that cannot be coerced (cause it's a downcast)
            dtype = np.dtype(dtype)
            maxval = np.iinfo(dtype).max
            x = np.array([0, 1, 2, maxval], dtype=dtype)
            test_data = Dataset({"x": ("t", x, {})})

            # Make sure a roundtrip fails
            with pytest.raises(ValueError, match="could not safely cast"):
                self.roundtrip(test_data)
```

... but I'm not sure ...
* whether this would be desirable to test
* whether this PR should take care of this or not
* where in `test_backend.py` I would have to put the test case such that it's invoked only for netcdf3. 
  * I already found the `NetCDF3Only` class and `requires_scipy` decorator, but don't know which test class is responsible here ...

blsqr:
@dcherian I think this would be ready for review now. I'd appreciate some guidance on the implementation of the tests; see above.

dcherian:
If you are up for it, we'd love to add testing for that error in this PR.

 I think `NetCDF3Only` is the right place to put it. It gets called by a number of test classes that inherit from `NetCDF3Only`. Add it there and see if things pass?

blsqr:
Alright, I added the test. From my side, this would be ready. :)

Let me know if further changes to the tests or What's New are necessary.

dcherian:
LGTM. Thanks @blsqr

dcherian:
Yay. Thanks @blsqr. I see this is your first PR. welcome!

blsqr:
Thanks, @dcherian! I'm happy to contribute 😊

Issue #3169, opened 2019-07-30
Plotting inconsistencies with Cartopy


andreas-h:
I'm on xarray 0.11.3, and currently do not have the possibility to test on master.  However, the commit history suggests that this issue still exists.  Please accept my apologies if I missed something.

I'm following the plotting documentation at https://xarray.pydata.org/en/stable/plotting.html#maps

When plotting two maps in the same call (exactly as written in the documentation),

    p = air.isel(time=[0, 1]).plot(transform=ccrs.PlateCarree(), col='time',
                                   subplot_kws={'projection': ccrs.Orthographic(-80, 35)})
    
    for ax in p.axes.flat:
        ax.coastlines()
        ax.gridlines()

everything works fine.

However, when I only want to plot one map,

    p = air.isel(time=[0]).plot(transform=ccrs.PlateCarree(),
                                subplot_kws={'projection': ccrs.Orthographic(-80, 35)})
    
    for ax in p.axes.flat:
        ax.coastlines()
        ax.gridlines()

I get the following error:

    ---------------------------------------------------------------------------
    ValueError                                Traceback (most recent call last)
    <ipython-input-63-2c5a2a533878> in <module>
          1 p = air.isel(time=[0]).plot(transform=ccrs.PlateCarree(),
    ----> 2                             subplot_kws={'projection': ccrs.Orthographic(-80, 35)})
          3 
          4 for ax in p.axes.flat:
          5     ax.coastlines()
    
    /home/jupyterhub/conda/envs/prod/lib/python3.6/site-packages/xarray/plot/plot.py in __call__(self, **kwargs)
        550 
        551     def __call__(self, **kwargs):
    --> 552         return plot(self._da, **kwargs)
        553 
        554     @functools.wraps(hist)
    
    /home/jupyterhub/conda/envs/prod/lib/python3.6/site-packages/xarray/plot/plot.py in plot(darray, row, col, col_wrap, ax, hue, rtol, subplot_kws, **kwargs)
        185     kwargs['ax'] = ax
        186 
    --> 187     return plotfunc(darray, **kwargs)
        188 
        189 
    
[...]

PR #4020, opened 2020-05-02
Allow passing axis kwargs to plot

raphaeldussin:
facecolor can be passed in plot kwargs but has no effect and does not raise exception.
PR fixes the problem.


mathause:
`facecolor` is a valid argument for several of the plotting functions (e.g. `pcolormesh`). The better approach here would be to thread `subplot_kws`

https://github.com/pydata/xarray/blob/3820fb77256682d909c1e41d962e29bec0edd62d/xarray/plot/plot.py#L125

through to `get_axis`

https://github.com/pydata/xarray/blob/3820fb77256682d909c1e41d962e29bec0edd62d/xarray/plot/utils.py#L384

in all the right places (+ tests). This would also solve #3169.

raphaeldussin:
apologies for the many commits, testing the test was a bit difficult.
I'm 99% there with the PR but I need help with mypy not happy with importing cartopy.

raphaeldussin:
an example of using cartopy and a custom background:

Screenshot 2020-05-16 20 26 52

raphaeldussin:
@mathause is there anything I can do to make this PR move forward?
I fixed the conflict that happened since my last commit. failed tests here seem independent from PR.

thanks!

raphaeldussin:
all tests passed!

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,61 @@ def test_facetgrid_single_contour():
...
+            ax = get_axis(figsize, size, aspect, ax)
+
+    ax = None
+    if figsize is not None and size is not None:
+        # cannot provide both size and figsize
+        with pytest.raises(ValueError):
+            ax = get_axis(figsize, size, aspect, ax)
+
+    if figsize is None:
+        if aspect is not None and size is None:
```
keewis, line 2418: Unless you're planning to add more code for the case where `figsize is None`, can you combine these conditions?

xarray/plot/utils.py
```diff
@@ -406,9 +406,12 @@ def _assert_valid_xy(darray, xy, name):
         raise ValueError(f"{name} must be one of None, '{valid_xy_str}'")
 
 
-def get_axis(figsize, size, aspect, ax):
-    import matplotlib as mpl
-    import matplotlib.pyplot as plt
+def get_axis(figsize, size, aspect, ax, **kwargs):
```
keewis, line 409: each of these parameters should probably get a default (`None`?):
```suggestion
def get_axis(figsize=None, size=None, aspect=None, ax=None, **kwargs):
```
that means if we just want to set e.g. `ax`, we can use
```python
get_axis(ax=ax)
```

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,61 @@ def test_facetgrid_single_contour():
...
+            # cannot provide aspect and size
+            with pytest.raises(ValueError):
+                ax = get_axis(figsize, size, aspect, ax)
+
+    try:
+        ax = plt.axes()
+        axtype = type(ax)
+        ax = get_axis(None, None, None, ax)
```
keewis, line 2426: with the addition of default values, these calls could become `get_axis(ax=ax)`, `get_axis()` and `get_axis(**kwargs)`, which I think are a lot easier to read.

raphaeldussin:
@keewis thank you for the review. I have implemented the proposed changes.

keewis:
thanks, looks good to me.

raphaeldussin:
would it make sense to update the documentation here: http://xarray.pydata.org/en/stable/plotting.html#maps

the update would be:

```diff
-    ax = plt.axes(projection=ccrs.Orthographic(-80, 35))
-    air.isel(time=0).plot.contourf(ax=ax, transform=ccrs.PlateCarree())
-    ax.set_global()
+    p = air.isel(time=0).plot(projection=ccrs.Orthographic(-80, 35),
+                              transform=ccrs.PlateCarree())
+    p.axes.set_global()

     @savefig plotting_maps_cartopy.png width=100%
-    ax.coastlines()
+    p.axes.coastlines()
```

xarray/plot/plot.py
```diff
@@ -177,10 +176,10 @@ def plot(
```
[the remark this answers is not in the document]
raphaeldussin: This does not work because line plots do not accept these kwargs

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,60 @@ def test_facetgrid_single_contour():
...
+@pytest.mark.parametrize("aspect", [4 / 3, None])
+@pytest.mark.parametrize("size", [200, None])
+def test_get_axis(figsize, aspect, size):
+    # test get_axis works with different args combinations
+    # and return the right type
+    try:
+        import matplotlib.pyplot as plt
+    except ImportError:
+        with pytest.raises(ImportError):
+            ax = get_axis(figsize, size, aspect, None)
```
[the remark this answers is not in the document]
raphaeldussin: this tests is not necessary anymore with the decorator

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,60 @@ def test_facetgrid_single_contour():
...
+        print("matplotlib not available")
+        return
+
+    ax = "something"
+    if figsize is not None:
+        # cannot provide both ax and figsize
+        with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
raphaeldussin: done

doc/plotting.rst
```diff
@@ -743,12 +743,13 @@ This script will plot the air temperature on a map.
 
     air = xr.tutorial.open_dataset("air_temperature").air
 
-    ax = plt.axes(projection=ccrs.Orthographic(-80, 35))
-    air.isel(time=0).plot.contourf(ax=ax, transform=ccrs.PlateCarree())
-    ax.set_global()
+    p = air.isel(time=0).plot(subplot_kws=dict(projection=ccrs.Orthographic(-80, 35),
+                                               facecolor="gray"),
+                              transform=ccrs.PlateCarree())
```
mathause, line 748:
```suggestion
    p = air.isel(time=0).plot(
        subplot_kws=dict(projection=ccrs.Orthographic(-80, 35), facecolor="gray"),
        transform=ccrs.PlateCarree(),
    )
```

nit: formatting

xarray/plot/plot.py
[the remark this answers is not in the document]
mathause: You are right. Could make sense to add them - but this can also be left for a future PR.

xarray/plot/plot.py
```diff
@@ -592,7 +591,7 @@ def newplotfunc(
         levels=None,
         infer_intervals=None,
         colors=None,
-        subplot_kws=None,
+        subplot_kws=dict(),
```
mathause, line 594:
```suggestion
        subplot_kws=None,
```

raphaeldussin:
> Looks good to me - thanks.
> 
> The "resolve conversation" button seems to have disappeared... so I cannot close the comments that are no longer relevant.
> 
> pinging @keewis @dcherian

I can see the "resolve conversation" buttons, do you want me to close them?

doc/plotting.rst
keewis: in the future, these kinds of formatting issues in documentation can be automatically fixed using `blackdoc`, see #4177.

xarray/plot/plot.py
```diff
@@ -548,8 +548,7 @@ def _plot2d(plotfunc):
         always infer intervals, unless the mesh is irregular and plotted on
         a map projection.
     subplot_kws : dict, optional
-        Dictionary of keyword arguments for matplotlib subplots. Only applies
-        to FacetGrid plotting.
+        Dictionary of keyword arguments for matplotlib subplots.
```
dcherian, line 551: Lets also add "only used for 2D and FacetGrid plots" (if I'm, understanding this correctly)

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,50 @@ def test_facetgrid_single_contour():
...
+
+    if figsize is None and aspect is not None and size is None:
+        # cannot provide aspect and size
+        with pytest.raises(ValueError, match="`aspect` argument without `size`"):
+            ax = get_axis(figsize, size, aspect, ax)
+
+    ax = plt.axes()
+    axtype = type(ax)
+    ax = get_axis(ax=ax)
+    assert isinstance(ax, axtype)
```
dcherian, line 2419: Can we change to
```suggestion
    assert isinstance(ax, mpl.axes.Axes)
```

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,50 @@ def test_facetgrid_single_contour():
...
+@requires_matplotlib
+@pytest.mark.parametrize("figsize", [[4, 4], None])
+@pytest.mark.parametrize("aspect", [4 / 3, None])
+@pytest.mark.parametrize("size", [200, None])
+def test_get_axis(figsize, aspect, size):
+    # test get_axis works with different args combinations
+    # and return the right type
+
+    ax = "something"
+    if figsize is not None:
```
dcherian, line 2396: I think this will be more readable if we didn't use `parametrize` so the tests look like

``` python
with pytest.raises(ValueError, match="both `figsize` and `ax`"):
        get_axis(figsize=[4,4], size=200, aspect=None, ax=ax)
```
and so on...

xarray/tests/test_plot.py
```diff
@@ -2381,3 +2382,50 @@ def test_facetgrid_single_contour():
...
+
+    ax = plt.axes()
+    axtype = type(ax)
+    ax = get_axis(ax=ax)
+    assert isinstance(ax, axtype)
+    ax = get_axis()
+    assert isinstance(ax, axtype)
+
+    try:
+        import cartopy as ctpy  # type: ignore
```
dcherian, line 2424: Let's move the cartopy bit to its own test. It's now running 8 times though it isn't using any of the parameters in the `parametrize` statements :)

I think you can use `ctpy = pytest.importorskip("cartopy")` to skip the test if `cartopy` is not installed.

keewis:
@mathause: yep, that seems to be restricted to maintainers and pull request authors, so if someone reviewed without being a maintainer they won't be able to change the resolve status of their own comments. Also, there doesn't seem to be a setting to change this.

xarray/tests/test_plot.py
keewis: either that or add a line here: https://github.com/pydata/xarray/blob/7ae74feb9a96e0ad0ca12f33fdb3b625ff58f924/xarray/tests/__init__.py#L57 and then use `requires_cartopy` as a decorator. I actually would prefer that since then reading the test body is not required to know why the test got skipped.

xarray/tests/test_plot.py
```diff
@@ -2421,6 +2422,8 @@ def test_get_axis(figsize, aspect, size):
 
 @requires_cartopy
 def test_get_axis_cartopy():
+    import cartopy as ctpy
+
```
keewis, line 2426: you could also add this to the top of the module:
```python
try:
    import cartopy as ctpy
except ImportError:
    ctpy = None
```

raphaeldussin: done

xarray/tests/test_plot.py
raphaeldussin: done

xarray/tests/test_plot.py
raphaeldussin: that's 8 tests and would make the test pretty long so I think there's no perfect solution

dcherian:
this should be good to go when tests pass.

raphaeldussin:
@dcherian thanks for the final fixes!

dcherian:
@raphaeldussin thanks for your patience here.

PR #4022, opened 2020-05-03
Fix/apply ufunc meta dtype

### LINKED ISSUE #4015: apply_ufunc gives wrong dtype with dask=parallelized and vectorized=True
Applying a function to a data array with `dtype = complex` returns one with `dtype = float`. It seems to work before commit 17b70caa6eafa062fd31e7f39334b3de922ff422.

#### MCVE Code Sample
```python
import numpy as np
import xarray as xr

def func(x):
    return np.sum(x ** 2)

da = xr.DataArray(np.arange(2*3*4).reshape(2,3,4))
da = da + 1j * da
da = da.chunk(dict(dim_1=1))

da2 = xr.apply_ufunc(
    func,
    da,
    vectorize=True,
    dask="parallelized",
    output_dtypes=[da.dtype],
)

assert da2.dtype == da.dtype, "wrong dtype"
```

#### Expected Output
`da` and `da2` should both have the same `dtype=complex`.

#### Problem Description
To me it seems to me that the kwarg `meta=None` somehow causes dask to allocate a float array and ignore the dtype kwargs (which seems to be carried through correctly down to `dask.array.blockwise.blockwise()`) . I'm not familiar with the apply_ufing and the dask code, so I can't tell on which end the bug sits.

#### Versions

INSTALLED VERSIONS
------------------
commit: None
python: 3.8.2 (default, Apr  8 2020, 14:31:25) 
[GCC 9.3.0]
python-bits: 64
OS: Linux
OS-release: 5.6.5-arch3-1
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: de_DE.utf8
LOCALE: de_DE.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.7.3

xarray: 0.15.2.dev47+g33a66d63
pandas: 1.0.3
numpy: 1.18.3
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: 0.7.4
h5py: 2.10.0
Nio: None
zarr: None
cftime: 1.1.1.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.3.2
dask: 2.12.0
distributed: 2.14.0
matplotlib: 3.2.1
cartopy: 0.17.0
seaborn: 0.10.0
numbagg: None
pint: None
setuptools: 46.1.3
pip: 20.0.2
conda: None
pytest: 5.4.1
IPython: 7.13.0
sphinx: 3.0.2

</details>

mathause:
- [x] Closes #4015
 - [x] Tests added

### DIFF
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -864,7 +864,9 @@ def apply_ufunc(
           inputs are a dask array. If used, the ``output_dtypes`` argument must
           also be provided. Multiple output arguments are not yet supported.
     output_dtypes : list of dtypes, optional
-        Optional list of output dtypes. Only used if dask='parallelized'.
+        Optional list of output dtypes. Used in ``np.vectorize`` and required if
+        dask='parallelized'. Note that the dtype of meta takes precedence over
+        output_dtypes in ``dask.array.blockwise``.
     output_sizes : dict, optional
         Optional mapping from dimension names to sizes for outputs. Only used
         if dask='parallelized' and new dimensions (not found on inputs) appear
@@ -1005,10 +1007,12 @@ def apply_ufunc(
         func = functools.partial(func, **kwargs)
 
     if vectorize:
-        if meta is None:
-            # set meta=np.ndarray by default for numpy vectorized functions
-            # work around dask bug computing meta with vectorized functions: GH5642
-            meta = np.ndarray
+        if dask == "parallelized" and meta is None:
+            # only basic checks: _apply_blockwise will raise the appropriate errors
+            if output_dtypes is not None and isinstance(output_dtypes, list):
+                # set meta=np.ndarray by default for numpy vectorized functions
+                # work around dask bug computing meta with vectorized functions: GH5642
+                meta = np.ndarray((0, 0), dtype=output_dtypes[0])
 
         if signature.all_core_dims:
             func = np.vectorize(
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -1016,12 +1016,16 @@ def apply_ufunc(
         func = functools.partial(func, **kwargs)
 
     if vectorize:
-        if dask == "parallelized" and meta is None:
-            # only basic checks: _apply_blockwise will raise the appropriate errors
-            if output_dtypes is not None and isinstance(output_dtypes, list):
-                # set meta=np.ndarray by default for numpy vectorized functions
-                # work around dask bug computing meta with vectorized functions: GH5642
-                meta = np.ndarray((0, 0), dtype=output_dtypes[0])
+        if (
+            dask == "parallelized"
+            and meta is None
+            and output_dtypes is not None
+            and isinstance(output_dtypes, list)
+        ):
+            # set meta=np.ndarray by default for numpy vectorized functions
+            # work around dask bug computing meta with vectorized functions: GH5642
+            # defer raising errors to _apply_blockwise (e.g. if output_dtypes is None)
+            meta = np.ndarray((0, 0), dtype=output_dtypes[0])
 
         if signature.all_core_dims:
             func = np.vectorize(
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -4,6 +4,7 @@ Functions for applying functions that act on arrays to xarray's labeled data.
 import functools
 import itertools
 import operator
+import warnings
 from collections import Counter
 from typing import (
     TYPE_CHECKING,
@@ -686,6 +687,14 @@ def _apply_blockwise(
             )
         )
     (dtype,) = output_dtypes
+
+    if (meta is not None) and hasattr(meta, "dtype") and (meta.dtype != dtype):
+        warnings.warn(
+            f"dtype of meta ({meta.dtype}) takes precedence over"
+            f" output_dtypes ({dtype})",
+            UserWarning,
+            stacklevel=3,
+        )
 
     if output_sizes is None:
         output_sizes = {}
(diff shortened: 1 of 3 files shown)


xarray/core/computation.py
```diff
@@ -1005,10 +1007,12 @@ def earth_mover_distance(first_samples,
         func = functools.partial(func, **kwargs)
 
     if vectorize:
-        if meta is None:
-            # set meta=np.ndarray by default for numpy vectorized functions
-            # work around dask bug computing meta with vectorized functions: GH5642
-            meta = np.ndarray
+        if dask == "parallelized" and meta is None:
+            # only basic checks: _apply_blockwise will raise the appropriate errors
+            if output_dtypes is not None and isinstance(output_dtypes, list):
```
mathause, line 1012: Is this ok? `_apply_blockwise` will raise informative errors if `output_dtypes` is None, if it is not a list or if it has more than one element. I don't necessarily want to repeat them here.

xarray/core/computation.py
```diff
@@ -1005,10 +1007,12 @@ def earth_mover_distance(first_samples,
...
+                # set meta=np.ndarray by default for numpy vectorized functions
+                # work around dask bug computing meta with vectorized functions: GH5642
+                meta = np.ndarray((0, 0), dtype=output_dtypes[0])
```
ulijh, line 1015: What should meta look like if there are multiple outputs with different dtypes?

ulijh:
Thanks @mathause , this works for me. Also with a change of dtype in func, as long as "output_dtypes" is set correctly.

xarray/core/computation.py
mathause: Currently multiple outputs are not supported with `dask="parallelized"` - this is checked in `_apply_blockwise`. I'm nor super happy with the way this is handled in my PR, but didn't want to copy the error check here, either.

@dcherian: should `meta` be created in `_apply_blockwise` instead (even if `vectorize=False`)?

dcherian: I thought about this but decided against it. We only want to automatically specify `meta` when `vectorize=True` because `np.vectorize` will return a numpy array and that too only to work around that dask bug.
I think we should leave the automatic `meta` inference to dask.

xarray/core/computation.py
dcherian: fine by me.

xarray/tests/test_computation.py
```diff
@@ -817,6 +817,58 @@ def test_vectorize_dask():
...
+
+    data_array = xr.DataArray([[0, 1, 2], [1, 2, 3]], dims=("x", "y"))
+    expected = xr.DataArray([1, 2], dims=["x"])
+    actual = apply_ufunc(
+        pandas_median,
+        data_array.chunk({"x": 1}),
+        input_core_dims=[["y"]],
+        vectorize=True,
+        dask="parallelized",
+        output_dtypes=[int],
```
dcherian, line 865: I think we should raise a warning in this case and add some text to the docstring describing this behaviour.

xarray/core/computation.py
```diff
@@ -1005,10 +1016,12 @@ def earth_mover_distance(first_samples,
         func = functools.partial(func, **kwargs)
 
     if vectorize:
-        if meta is None:
-            # set meta=np.ndarray by default for numpy vectorized functions
-            # work around dask bug computing meta with vectorized functions: GH5642
-            meta = np.ndarray
+        if dask == "parallelized" and meta is None:
+            # only basic checks: _apply_blockwise will raise the appropriate errors
+            if output_dtypes is not None and isinstance(output_dtypes, list):
```
dcherian, line 1021: combine this with the `if` above?

xarray/core/computation.py
```diff
@@ -1005,10 +1016,16 @@ def earth_mover_distance(first_samples,
...
+            dask == "parallelized"
+            and meta is None
+            and output_dtypes is not None
+            and isinstance(output_dtypes, list)
+        ):
             # set meta=np.ndarray by default for numpy vectorized functions
             # work around dask bug computing meta with vectorized functions: GH5642
-            meta = np.ndarray
+            # defer raising errors to _apply_blockwise (e.g. if output_dtypes is None)
+            meta = np.ndarray((0, 0), dtype=output_dtypes[0])
```
shoyer, line 1028: shouldn't we still set `meta = np.ndarray` if no output dtype is specified?

mathause: `output_dtypes` is required for `dask="parallelized"` and will error if it is missing:

https://github.com/pydata/xarray/blob/8051c47686ae23062f9008785563d62327735002/xarray/core/computation.py#L670-L674

so this wont take effect. I am also not very happy with my approach, but didn't want to copy the checks from `apply_blockwise` up here - suggestions?

dcherian: Maybe the cleaner workaround is to move this down in to `_apply_blockwise`? Would it be enough to pass `vectorize` down to that level and then set `meta` as you are doing here? 

Also, it seems like we should raise that error about `output_dtypes` only if `meta.dtype` has not been set?

mathause: Yes, I agree it would be cleaner to thread `vectorize` through to `_apply_blockwise`.

> Also, it seems like we should raise that error about output_dtypes only if meta.dtype has not been set?

Depends how important `output_dtypes` is for `np.vectorize`.

I am happy to work more on this, but I think it would be good to discuss #4060 first, which might make this obsolete.

mathause:
I'll close this in favor of #4060

PR #4029, opened 2020-05-05
Support overriding existing variables in to_zarr() without appending

shoyer:
This is nice for consistency with `to_netcdf`. It should be useful for cases where users want to update values in existing Zarr datasets.

 - [x] Tests added


rabernat:
This is great Stephan. Perhaps you could also update the `to_zarr` docstring to make it more consistent with `to_netcdf` and explicitly describe this behavior. For `to_netcdf`, we say
>  If mode=’a’, existing variables will be overwritten.

shoyer:
Thanks for the reviews!

The doc build failure is something else (see https://github.com/pydata/xarray/issues/4030), so I'm going to submit.

PR #4033, opened 2020-05-05
xr.infer_freq

aulemahal:
- [x] Tests added

This PR adds a `xr.infer_freq`  method to copy pandas `infer_freq` but on `CFTimeIndex` objects. I tried to subclass pandas `_FrequencyInferer` and to only override as little as possible.

Two things are problematic right now and I would like to get feedback on how to implement them if this PR gets the dev's approval.

1) `pd.DatetimeIndex.asi8` returns integers representing _nanoseconds_ since 1970-1-1, while `xr.CFTimeIndex.asi8` returns _microseconds_. In order not to break the API, I patched the  `_CFTimeFrequencyInferer` to store 1000x the values. Not sure if this is the best, but it works.

2) As of now, `xr.infer_freq` will fail on weekly indexes. This is because pandas is using `datetime.weekday()` at some point but cftime objects do not implement that (they use `dayofwk` instead).  I'm not sure what to do?   Cftime could implement it to completly mirror python's datetime or pandas could use `dayofwk` since it's available on the `TimeStamp` objects.

Another option, cleaner but longer, would be to reimplement `_FrequencyInferer` from scratch. I may have time for this, cause I really think  a `xr.infer_freq` method would be useful.


aulemahal:
@spencerkclark Thanks for the feedback!

I rewrote the `_CFTimeFrequencyInferer` class independently of pandas private objects. I removed outputs that wouldn't match offsets defined in xarray: weekly frequencies, week-of-month (WOM) and all business-related anchors. It wouldn't be difficult to add them later on if need be and if `cftime_offsets.py` defines them.  Note that `infer_freq` is quite independent of the rest of xarray, this choice was purely so that the behavior matches the rest of xr. I did leave the sub-seconds frequencies (L for ms and U for µs).

I expect the performance of `xr.infer_freq` to be slightly worst than pandas' since they use optimized and compiled code whereas I only used common numpy and xarray stuff. However, I assume that this part of the workflow is far from being the bottleneck of most xarray usecases...

Will write up some docs soon.

xarray/coding/frequencies.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        if not self.is_monotonic or not self.index.is_unique:
+            return None
+
+        delta = self.deltas[0]  # Smallest delta
+        if _is_multiple(delta, _ONE_DAY):
+            return self._infer_daily_rule()
+        # There is not other possible intraday frequency
+        # Different from pandas: we don't need to manage DST and business offsets in cftime
+        elif not len(self.deltas) == 1:
```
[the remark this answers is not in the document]
aulemahal: It is right!

xarray/tests/test_cftimeindex.py
```diff
@@ -1046,3 +1046,28 @@ def test_asi8_distant_date():
     result = index.asi8
     expected = np.array([1000000 * 86400 * 400 * 8000 + 12345 * 1000000 + 123456])
     np.testing.assert_array_equal(result, expected)
+
```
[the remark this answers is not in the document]
aulemahal: I added almost all those tests, except the non-datetime-or-timedelta-like index.  When passing an object that is not a DataArray, a pd.Series or a CFTimeIndex, `xr.infer_freq` directly calls `pd.infer_freq`, so I felt this test wasn't in the scope of xarray.

doc/weather-climate.rst
```diff
@@ -74,6 +74,15 @@ instance, we can create the same dates and DataArray we created above using:
     dates = xr.cftime_range(start="0001", periods=24, freq="MS", calendar="noleap")
     da = xr.DataArray(np.arange(24), coords=[dates], dims=["time"], name="foo")
 
+Mirroring pandas' method with the same name, :py:meth:`~xarray.infer_freq` allows to
+infer the sampling frequency of a :py:class:`~xarray.CFTimeIndex` or a 1-D
+:py:class:`~xarray.DataArray` containing cftime objects. It also works tranparently with
```
mathause, line 79:
```suggestion
:py:class:`~xarray.DataArray` containing cftime objects. It also works transparently with
```

aulemahal:
@spencerkclark  Here you go! Tests added for all cases you mentioned.

xarray/coding/frequencies.py
```diff
@@ -0,0 +1,272 @@
...
+        If the index is not datetime-like.
+    ValueError
+        If there are fewer than three values or the index is not 1D.
+    """
+    from xarray.core.dataarray import DataArray
+
+    if isinstance(index, (DataArray, pd.Series)):
+        if index.ndim != 1:
+            raise ValueError("'index' must be 1D")
+        elif not _contains_datetime_like_objects(DataArray(index)):
```
spencerkclark, line 85: Oh sorry, one more minor thing here, I suppose we should check for the `timedelta64[ns]` dtype here too.  `_contains_datetime_like_objects` just checks for dates.

aulemahal: I think it does check for timedelta64? Its docstring says : 
``` 
"""Check if a variable contains datetime like objects (either
    np.datetime64, np.timedelta64, or cftime.datetime)
"""
```
Short tests with timedelta64 did work on my end.

spencerkclark: Oops you're right -- I was mistaken.  Sorry for the noise.

spencerkclark:
Thanks @aulemahal!  Nice work.

PR #4038, opened 2020-05-06
Use literal syntax instead of function calls to create the data structure

pnijhara:
---
It is slower to call e.g. `list()` than using the empty literal, because the name list must be looked up in the global scope in case it has been rebound. Same for the other two types like `dict()` and `tuple()`.

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -146,6 +146,10 @@ Internal Changes
   ``all``, ``any``, ``enumerate``, ``sum``, ``tuple`` etc. can work directly with a
   generator expression. (:pull:`4026`)
   By `Prajjwal Nijhara <https://github.com/pnijhara>`_.
+- Use literal syntax instead of function calls to create the data structure
+  It is slower to call e.g. list() than using the empty literal, because the name list
+  must be looked up in the global scope in case it has been rebound. (:pull:`4038`)
+  By `Prajjwal Nijhara <https://github.com/pnijhara>`_.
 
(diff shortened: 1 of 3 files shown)


doc/whats-new.rst
```diff
@@ -146,6 +146,10 @@ Internal Changes
   ``all``, ``any``, ``enumerate``, ``sum``, ``tuple`` etc. can work directly with a
   generator expression. (:pull:`4026`)
   By `Prajjwal Nijhara <https://github.com/pnijhara>`_.
+- Use literal syntax instead of function calls to create the data structure
+  It is slower to call e.g. list() than using the empty literal, because the name list
+  must be looked up in the global scope in case it has been rebound. (:pull:`4038`)
+  By `Prajjwal Nijhara <https://github.com/pnijhara>`_.
```
shoyer, line 152: I would suggest consolidating all of these notes into a single entry, e.g.,
```
- Various internal code clean-ups (:pull:`4026`,  :pull:`4038`).
  By `Prajjwal Nijhara <https://github.com/pnijhara>`_.
```

The details of these refactorings are not worthy of extended notes.

pnijhara: @shoyer made the changes. Please review it again.

shoyer:
Thanks @pnijhara !

(For what's worth, the speed difference between `list()` and `[]` is almost certainly unmeasurable in these functions, but I agree that using a literal does look a little cleaner.)

PR #4039, opened 2020-05-06
Revise pull request template

shoyer:
See below for the new language, to clarify that documentation is only necessary
for "user visible changes."

I added "including notable bug fixes" to indicate that minor bug fixes may not
be worth noting (I was thinking of test-suite only fixes in this category) but
perhaps that is too confusing.

cc @pydata/xarray for opinions!

 - [ ] Closes #xxxx
 - [ ] Tests added
 - [ ] Fully documented, including `whats-new.rst` for user visible changes
       (including notable bug fixes) and `api.rst` for new API

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,4 +3,5 @@
  - [ ] Closes #xxxx
  - [ ] Tests added
  - [ ] Passes `isort -rc . && black . && mypy . && flake8`
- - [ ] Fully documented, including `whats-new.rst` for user visible changes (including notable bug fixes) and `api.rst` for new API
+ - [ ] User visible changes (including notable bug fixes) are documented in `whats-new.rst`
+ - [ ] New functions/methods are listed in `api.rst`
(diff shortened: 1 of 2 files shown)


max-sixty:
I think that's reasonable

I was leaning to "everyone add their name", as it's a way of giving contributors credit, and some visible ownership — even for small changes which might be someone's first PR. 

But I agree that's not always consistent with making the whatsnew better for readers.

pnijhara:
Just a small suggestion: You can make a new ``.rst`` or ``.md`` file for "everyone add their name".in order to give contributors credit (If you want to).

shoyer:
> I was leaning to "everyone add their name", as it's a way of giving contributors credit, and some visible ownership — even for small changes which might be someone's first PR.

I agree, this is a nice idea.

My goals were to make things simpler for new contributors and also to make whats-new a little bit for readers.

Two ways we might still achieve this:

- We could still have a section in the release notes for each version for contributors to add their name
- We could update this manually as part of the release process (but this adds one more step for core developers)

max-sixty:
>     * We could update this manually as part of the release process (but this adds one more step for core developers)

Actually we do this already so maybe that covers it?

dcherian:
> So maybe we paste that into the whatsnew too, and we're covered?

Seems like this was the consensus at this meeting. So let's add another step to `HOW_TO_RELEASE` and then merge this?

shoyer:
OK, revised.

Issue #4046, opened 2020-05-08
automatic chunking of zarr archive


apatlpo:
I store data in a zarr archive that is not chunked and the resulting zarr archive is chunked.
This may be as simple usage question.
I don't know how to turn this behavior off.

#### Code sample

Here is minimal example that reproduces the issue:

```python
ds = xr.DataArray(np.ones((200,800))).rename('foo').to_dataset()
print('Initial chunks = {}'.format(ds.foo.chunks))
ds.to_zarr('test.zarr', mode='w')

print('zarr archives contains: {}'.format(os.listdir('test.zarr/foo')))

ds = xr.open_zarr('test.zarr')
print('Final chunks = {}'.format(ds.foo.chunks))
```

#### Expected Output

I would expect the archive to not to be chunked.

#### Versions

<details><summary>Output of <tt>xr.show_versions()</tt></summary>

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 23:03:20) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 3.12.53-60.30-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.7.4

xarray: 0.15.2.dev29+g6048356
pandas: 1.0.3
numpy: 1.18.1
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: 2.4.0
cftime: 1.1.1.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.13.0
distributed: 2.13.0
matplotlib: 3.2.1
cartopy: 0.17.0
seaborn: 0.10.0
numbagg: None
pint: None
setuptools: 46.1.3.post20200325
pip: 20.0.2
conda: None
pytest: None
IPython: 7.13.0
sphinx: None

</details>

rabernat:
Thanks for raising this useful issue.

There are two ways to control Zarr chunks:
- Specify chunks in `encoding` (always takes precedence)
- Determine chunks based on Dask chunks

[...]

PR #4048, opened 2020-05-08
improve to_zarr doc about chunking

apatlpo:
- [X] follows #4046
 - [X] Passes `isort -rc . && black . && mypy . && flake8`

I'm not sure the last point is really necessary for this PR, is it?


xarray/core/dataset.py
```diff
@@ -1604,6 +1604,12 @@ def to_zarr(
         References
         ----------
         https://zarr.readthedocs.io/
+
+        Notes
+        -----
+        zarr may automatically chunk a DataArray if it is not chunked or
+        if `chunks` is not set to -1 in its `enconding` attribute or in the
+        argument of the `encoding` parameter.
```
shoyer, line 1612: Can we instead state what _does_ happen if a variable is already chunked or `chunks` can be found in `encoding`?

Something like

- If `chunks` are found in the `encoding` argument or attribute corresponding to any DataArray, those chunks are used.
- Otherwise, if the DataArray is already a dask array, it is written with those chunks.
- Finally, if not other chunks are found, Zarr uses its own heuristics to choose automatic chunk sizes.

apatlpo: I closely followed your suggestion.
Note that I am not sure about the docstring style for a list-like note like this one.

apatlpo:
if anybody has a clue on how to fix tests this would be welcome.
thx

rabernat:
There are two sphinx warnings in building the docs
https://dev.azure.com/xarray/xarray/_build/results?buildId=2819&view=logs&j=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36&t=68484831-0a19-5145-bfe9-6309e5f7691d&l=280

```
xarray/core/dataset.py:docstring of xarray.Dataset.to_zarr:44: WARNING: Definition list ends without a blank line; unexpected unindent.
xarray/core/dataset.py:docstring of xarray.Dataset.to_zarr:46: WARNING: Unexpected indentation.
```

These seem related to your changes. I would try to get the docs building locally without any warnings.

keewis:
the `flake8` is fixed in #4057, so you can merge `master` to fix it.

apatlpo:
ok, tests are passing.

keewis:
they do, but could you remove the `saved_on_disk.nc` file that was accidentally added?

Also, the items are not a list despite being formatted like one:
to_zarr
is that intentional?

apatlpo:
argh ... `saved_on_disk.nc` has been removed.

I was unfortunately not able to create a proper list without breaking tests and have exhausted my time trying to figure it out.
If you know how to do that I am more than happy to follow your advice.

shoyer:
I think it's OK that it's not formatted as a list. It's still pretty readable

shoyer:
thanks @apatlpo !

PR #4053, opened 2020-05-12
Fix html repr in untrusted notebooks (plain text fallback)

### LINKED ISSUE #4041: expanded HTML repr when opening notebook
When I open a notebook, the new HTML repr is "expanded": 

image

I'm running
```
jupyter core     : 4.6.3
jupyter-notebook : 6.0.3
qtconsole        : 4.7.2
ipython          : 7.13.0
ipykernel        : 5.2.0
jupyter client   : 6.1.2
jupyter lab      : 2.1.0
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.4
traitlets        : 4.3.3
```
and see this behaviour with both the notebook and jupyterlab interface; on both chrome and firefox (both on linux).

Is anyone else seeing this?

benbovy:
- [x] Closes #4041
 - [x] Tests added

This is not very elegant (actually plain text repr is already included in the notebook as `text/plain` mime type but it is ignored when `text/html` mime type is present), but it seems to work. I haven't found a better workaround.

I don't really know if this can be properly tested (I only added a basic test).

Steps to test this fix:

- To "untrust" a notebook: open an existing notebook with a simple editor, manually edit one output cell with a xarray object repr, and save the ipynb file.
- Open this notebook with the Notebook app, you should see the plain text repr.


shoyer:
I tested this out exporting a notebook to Github:
https://gist.github.com/shoyer/6ee36acd164116fee4681f6237ca5249

Here's what I see:
image

Somehow we get both text and malformed HTML? :(

benbovy:
Mmm it looks like Github's notebook renderer removes the `hidden` attribute of the `xr-wrap` div in addition to removing CSS. That's unfortunate.

I tested this workaround in jupyterlab, classic notebook and nbviewer. It works as expected. I haven't tested other front-ends (Colab, vscode, etc.).

shoyer:
I sent a support request to GitHub asking for supporting the `hidden` attribute. We'll see if we get a response!

dcherian:
Thanks @benbovy

dcherian:
just confirmed that the fallback works perfectly. Thanks @benbovy !

Issue #4031, opened 2020-05-05
0.16.0 release


dcherian:
It'd be nice to issue a release soon. We should decide if this is a minor 0.15.2 or major 0.16.0

Please edit this list as you wish.

**Must-have**

- [x] #3936 multiple dims argmin, argmax

- [x] #4019, #4088; MultiIndex to sparse DataArray bug
- [x] #3922 fix dask array handling in idxmax, idxmin
- [x] #4009, #4173 combine_attrs with open_mfdataset
- [x] #3824 transpose coords (if major release)
- [x] #3926 remove old autocombine
- [x] #4135 nD dask arrays in idxmax/min

**Nice to have**

- [x] #4049, #4094 unstacking merge bug
- [ ] #4022 apply_ufunc meta vectorize
- [ ] #3924 interpolating to coordinates with nans
- [ ] #3925 sel along unindexed 1d coordinates
- [ ] #3993 replace dim with coord in integrate
- [ ] #3594 unit support with pint
- [ ] #4003 mfdataset with zarr
- [x] #3905 length of dataarray reprs
- [x] #3847 error summary for assert_allclose
- [x] #4033 infer_freq
- [x] #3938 plotting with multiindex
- [x] #3816 map_blocks template
- [x] #3976 in-place replacement error message
- [x] #3975 pint dataset
- [x] #4018 netdf3 dtype coercion
- [ ] #3985 groupby multi-index bug (ref discussion below & in the issue)

mathause:
- [ ]  #4075 passing boolean weights to weighted mean

would be nice

max-sixty:
As discussed on the dev call, I'll take this one.

I think we need to do 0.16.0 as we had some backward-incompat changes (I can't remember now what they were; but can hunt them down if anyone otherwise has a preference for 0.15.2?)

dcherian:
Yes we have some breaking changes so definitely 0.16.0

IMO we should do #4184 #4175 and #4163 and then release. #3985  is still waiting for a minimal example so perhaps we should leave that for the next release.

max-sixty:
Agree! I'll keep an eye on those this week.

max-sixty:
Thanks for updating @dcherian . Is it worth waiting for #3985? I'm still not sure exactly what the bug is there. I'd vote to leave it until we have a repro.

I was thinking of sneaking in https://github.com/pydata/xarray/issues/3891 but didn't want to bring it up as probably won't be able to; definitely not worth waiting on me for!

[...]

PR #4060, opened 2020-05-14
ENH: use `dask.array.apply_gufunc` in `xr.apply_ufunc` 

### LINKED ISSUE #1815: apply_ufunc(dask='parallelized') with multiple outputs
I have an application where I'd like to use `apply_ufunc` with dask on a function that requires multiple inputs and outputs. This was left as a TODO item in the #1517. However, its not clear to me looking at the code how this can be done given the current form of dask's atop. I'm hoping @shoyer has already thought of a clever solution here...

```python
def func(foo, bar):
    
    assert foo.shape == bar.shape
    spam = np.zeros_like(bar)
    spam2 = np.full_like(bar, 2)

    
    return spam, spam2

foo = xr.DataArray(np.zeros((10, 10))).chunk()
bar = xr.DataArray(np.zeros((10, 10))).chunk() + 5

xrfunc = xr.apply_ufunc(func, foo, bar,
                        output_core_dims=[[], []],
                        dask='parallelized')
```
#### Problem description

This currently raises a `NotImplementedError`.

#### Expected Output

Multiple dask arrays. In my example above, two dask arrays.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.86+
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

xarray: 0.10.0+dev.c92020a
pandas: 0.22.0
numpy: 1.13.3
scipy: 1.0.0
netCDF4: 1.3.1
h5netcdf: 0.5.0
Nio: None
zarr: 2.2.0a2.dev176
bottleneck: 1.2.1
cyordereddict: None
dask: 0.16.0
distributed: 1.20.2+36.g7387410
matplotlib: 2.1.1
cartopy: None
seaborn: None
setuptools: 38.4.0
pip: 9.0.1
conda: 4.3.29
pytest: 3.3.2
IPython: 6.2.1
sphinx: None
</details>

cc @mrocklin, @arbennett

### LINKED ISSUE #4015: apply_ufunc gives wrong dtype with dask=parallelized and vectorized=True

kmuehlbauer:
use `dask.array.apply_gufunc` in `xr.apply_ufunc` for multiple outputs when `dask='parallelized'`, add/fix tests

 - [x] Closes #1815, closes #4015
 - [x] Tests added

Remaining Issues:

- [ ] fitting name for current `dask_gufunc_kwargs`
- [ ] rephrase [dask docs](https://github.com/pydata/xarray/blob/master/doc/dask.rst) to fit new behaviour
- [ ] combine `output_core_dims` and `output_sizes`, eg. `xr.apply_ufunc(..., output_core_dims=[{"abc": 2]])`


kmuehlbauer:
This would need some docstring changing too. But I first want to check, if I've missed anything vital in the implementation.

kmuehlbauer:
This is ready for review from my side.

mathause:
- It might be good to add a test with a reduction and one with `vectorize=True`.
- Would it be possible to replace the call to `dask.array.blockwise` (for one output variable) with `dask.array.apply_gufunc`? Do you know why `blockwise` is used further below and not `dask.array.apply_gufunc`? I assume it's due to historical reasons but I am not sure.
- `dask.array.apply_gufunc` does all sorts of stuff - e.g. infer meta. This could potentially solve #4015 (pull #4022) and simplify the call signature of `apply_ufunc`?

https://github.com/dask/dask/blob/3573b2ddca81aeb41a7def6dd4194020f853ab18/dask/array/gufunc.py#L175

kmuehlbauer:
Thanks @mathause for your comments and raising those questions. JFTR, I was taking the road from #1815, so my explicit use-case was the multiple (dask) outputs.

> * It might be good to add a test with a reduction and one with `vectorize=True`.

I'll try to add some tests for the multiple output using dask.

> * Would it be possible to replace the call to `dask.array.blockwise` (for one output variable) with `dask.array.apply_gufunc`? Do you know why `blockwise` is used further below and not `dask.array.apply_gufunc`? I assume it's due to historical reasons but I am not sure.

AFAIK, `apply_gufunc` wasn't available at the time these functions were introduced. Good chance, that `apply_gufunc` can be used for handling single output dask too.

> * `dask.array.apply_gufunc` does all sorts of stuff - e.g. infer meta. This could potentially solve #4015 (pull #4022) and simplify the call signature of `apply_ufunc`?

That's a good question. If you want me to go the long way, please be aware, that I'm a novice in xarray as well as in dask. A complete refactor of `apply_ufunc` would be quite some challenge.

mathause:
Ah yes I see (https://github.com/pydata/xarray/issues/1815#issuecomment-440089606). `dask.array.apply_gufunc` should also be able to handle one output only.

> A complete refactor of `apply_ufunc` would be quite some challenge.

Indeed - I think it could simplify `_apply_blockwise` (and might make the `meta` keword obsolete) but it would be good if someone with more experience of dask could weigh in.

@dcherian @shoyer

shoyer:
> > * Would it be possible to replace the call to `dask.array.blockwise` (for one output variable) with `dask.array.apply_gufunc`? Do you know why `blockwise` is used further below and not `dask.array.apply_gufunc`? I assume it's due to historical reasons but I am not sure.
> 
> AFAIK, `apply_gufunc` wasn't available at the time these functions were introduced. Good chance, that `apply_gufunc` can be used for handling single output dask too.

Exactly. It would be nice remove the use of `blockwise` entirely in favor of `apply_gufunc`.

kmuehlbauer:
I've given this a try, but this will need some design decisions.

- currently `vectorize` is handled in any case if requested, before falling through the if/else.
  `dask.array.apply_gufunc` takes `vectorize` as parameter, so for dask we do not need to apply   vectorization. We would need to apply `vectorize` only for non-dask cases (maybe just before calling the final function).
-currently dask is only handled, if `dask` keyword is issued ('allowed' and `parallelized`). From my perspective the `dask` keyword is not needed any more. We could just divert to the `apply_gufunc` when dask backed arrays are detected.

This will really have much impact on the code/tests. I'll come up with a updated PullRequest in short time, but any thoughts /remarks whatsoever are very much appreciated.

mathause:
Thanks! That would be quite cool!

`dask.array.apply_gufunc` does some fancy stuff with `np.vectorize` (to determine the `output_dtypes`) so I would not vectorize the function ourselves. For the other I don't have a qualified opinion.

kmuehlbauer:
@mathause @shoyer

First serve of trying to use `dask.array.apply_gufunc` in `xr.apply_ufunc`. I've added a list with problems in the topmost comment of this PR, to not loose track of this. Please enhance that list, if needed.

Most problematic issue now: `xr.dot` doesn't work well with `apply_gufunc` with regard to core dimensions and chunking.

kmuehlbauer:
From looking at the tests, it seems that setting `allow_rechunk=True` would solve many issues. I've no idea about the implications on memory usage (docstring: "Warning: enabling this can increase memory usage significantly"). Might `apply_gufunc` not be suitable for processing of eg. `dot`?

mathause:
Should you keep the `dask="allowed"` branch? That might solve some of the issues (instead of using `allow_rechunk=True`) . For example dask has it's own `einsum` implementation (used in `xr.dot`) so it may not be necessary to pipe that through `dask.array.apply_gufunc` (from the [docstring](https://docs.dask.org/en/latest/array-api.html#dask.array.gufunc.apply_gufunc): this function is like np.vectorize, but for the blocks of dask arrays). However, I am only speculating.

kmuehlbauer:
@mathause Great! Seems like this works better, thanks! Will update the PR after some more tests etc.

kmuehlbauer:
@mathause Only one breaking test which is connected to the meta stuff you handle in #4022. Any suggestions on that topic? I've removed meta completely since it was only needed in `dask.array.blockwise` but not in `dask.array.apply_gufunc`.

kmuehlbauer:
@mathause All tests green, good starting point for review. Please notify other people who should have a look at this.

There are still things to discuss: 
- keywords of `dask.array.apply_gufunc`
- howto properly handle deprecation (eg. meta)
- doscstring
- adding/revision of tests

shoyer:
The original motivation for requiring `dask='allowed'` is that I was concerned that users would put a function that coerces its arguments into NumPy arrays into `apply_ufunc` (e.g., like many functions from SciPy), which could have surprisingly bad performance when called on dask arrays due to automatic coercion.

Maybe this is too defensive/surprising, and could be relaxed. We don't really have any guard-rails like this elsewhere in xarray.

mathause:
> Maybe this is too defensive/surprising, and could be relaxed.

You would remove the `daks="forbidden"` branch and not the `dask="parallelized"`?

For the functions that don't handle dask arrays gracefully, `dask="parallelized"` would be the better option?

---

Very cool - good progress.

- I guess you'll have to properly deprecate meta, something along the lines of: `meta is no longer necessary and has no effect. it will be removed in a future version`
- I think it would be good to pass `allow_rechunk`.

I'll only be able to look at it properly next week.

shoyer:
> > Maybe this is too defensive/surprising, and could be relaxed.
> 
> You would remove the `daks="forbidden"` branch and not the `dask="parallelized"`?
> 
> For the functions that don't handle dask arrays gracefully, `dask="parallelized"` would be the better option?

This is probably another good motivation: defaulting to `dask='forbidden'` forces users to make an explicit choice about whether or not use `dask='parallelized'`.

The problem is that we don't have any way to detect ahead of time whether the applied function already supports dask arrays (e.g., if it is built-up out of functions from dask.array). If it does, we don't want to set `dask='parallelized'` but rather let the function handle dask arrays itself.

dcherian:
> If it does, we don't want to set dask='parallelized' but rather let the function handle dask arrays itself.

I think we still need all the current options for the `dask` kwarg. There can be disastrous consequences so it's good to make users explicitly choose the behaviour they want.

> howto properly handle deprecation (eg. meta)

I don't think we should deprecate meta. Not all user functions can deal with zero shaped inputs, so automatically inferring meta need not always work. We've had to add a similar feature for map_blocks (#3575) so I think `meta` should stay.

> keywords of dask.array.apply_gufunc

Shall we add a new `dask_gufunc_kwargs` and pass that down to `appy_gufunc`?

mathause:
I don't see `meta` listed in the [docs](
https://docs.dask.org/en/latest/array-api.html#dask.array.gufunc.apply_gufunc) which is why it thought it's not needed. But if it is handled in `dask.array.apply_gufunc` it can of course stay.

I only realised the exact distinction between `"allowed"` and `"parallelized"` today - i.e. that `"parallelized"` is kind of the dask equivalent of `np.vectorize`. I can suggest something for the docstring (e.g. `prefer "allowed" if ``func`` natively handles dask arrays` or so)

dcherian:
good point @mathause. Looks like `apply_gufunc` tries to make blockwise infer meta and then does its own thing when that fails:
https://github.com/dask/dask/blob/3b92efff2e779f59e95e05af9b8f371d56227d02/dask/array/gufunc.py#L417-L440 . I don't understand how this can work in all possible cases.

kmuehlbauer:
> I'll only be able to look at it properly next week.

@mathause I'll leave the PR unchanged and catch up with you next week.

@shoyer @dcherian Thanks for your comments. Please let me know, which tests should be added to check for any possible surprises with this change to `apply_gufunc`.

> The problem is that we don't have any way to detect ahead of time whether the applied function already supports dask arrays (e.g., if it is built-up out of functions from dask.array). If it does, we don't want to set dask='parallelized' but rather let the function handle dask arrays itself.

*(Att: no native english speaker here, so bear with me, if something sounds clunky or not exactly matching)*
Then we would have to keep the `dask='forbidden'` as default, as well as `parallelized` and `allowed` to force the decision to the user. Maybe the keyword settings itself could be a bit more clear.  In the `allowed`-case the function in question has to natively support dask-arrays. So I would use `dask='native'` in that case. For the `parallelized`-case this PR proposes to use `dask.array.apply_gufunc` (generalized ufunc). So either we stick to `parallelized` or we try to find a better fitting name (eg. `dask='gufunc'`). 

For the keywords I think @dcherian 's proposal of something like `dask_gufunc_kwargs` (or `gufunc_kwargs`) is useful (would match with `dask='gufunc'`), although only two keywords seem to be worth feeding through (`keepdims`, `allow_rechunk`).

kmuehlbauer:
> although only two keywords seem to be worth feeding through (`keepdims`, `allow_rechunk`).

Just noticed, that `output_dtypes` and `output_sizes` could also be fed as `gufunc_kwargs`. So this would really simplify the calling signature of `apply_ufunc` if those kwargs could be handled that way.

kmuehlbauer:
@mathause Thanks for the elaborate review. But, I'm unsure how to move on with this. 

I'll reply to some of your comments below.

> Note that `output_dtypes` is used in `vectorize` so this may not be a candidate for `dask_gufunc_kwargs`. 

Hmm, the docstring indicates that it is only used with `dask='parallelized'`. But in the code it can be used in any case. This should be clarified.

> The second is not so trivial. I see three possibilities (1) just error, (2) try `dask.array.apply_gufunc` if that fails issue a warning and use the old `apply_blockwise` (3) figure out ourselves if non-core dimensions (called loop dimensions in dask) are not-equally chunked, issue a warning and re-chunk them ourselves. Maybe @shoyer and @dcherian can weight in here.

When looking at the code of [dask.array.apply_gufunc](https://github.com/dask/dask/blob/1ee16ca4d8b26e0a75b68a93ae7e7e10318b61fb/dask/array/gufunc.py#L394-L424) it seems, that any possible re-chunking is done in `dask.array.blockwise`. I also do not see any special treatment of warning or re-chunking of non-core dimensions in the current [_apply_blockwise](https://github.com/pydata/xarray/blob/e5cc19cd8f8a69e0743f230f5bf51b7778a0ff96/xarray/core/computation.py#L652). So from my reasoning we should be safe to use (4) try `dask.array.apply_gufunc` since `dask.array.apply_gufunc` will issue the appropriate ValueError if something is wrong with the chunking. (Only thing is that the ValueError concerning the non-core dimension doesn't mention `allow_rechunk`). If the ValueError is issued, the user can the just set `allow_rechunk=True` keyword.

xarray/tests/test_computation.py
```diff
@@ -670,7 +685,11 @@ def check(x, y):
     check(data_array, 0 * data_array)
     check(data_array, 0 * data_array[0])
     check(data_array[:, 0], 0 * data_array[0])
-    check(data_array, 0 * data_array.compute())
+    # todo: this raises ValueError since chunks do not match inside apply_gufunc due to
+    #  computing, could be fixed by setting `allow_rechunk=True` in the call to
+    #  apply_gufunc (or rechunking before feeding to apply_ufunc)
+    with raises_regex(ValueError, "with different chunksize present"):
```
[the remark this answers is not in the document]
kmuehlbauer: It seems it does allow it if `allow_rechunk=True`.

mathause:
`output_dtypes`: yes I think the docs should be updated

What I was trying to say - this is a change that can break existing code and it is not easy to do a proper deprecation cycle (i.e. only throw a warning). (I think `allow_rechunk=True` is not a good default.)

---

Ok an easier idea (could be brittle, though):

```python
try:
    res = da.apply_gufunc(...)
# catch error that was not raised in earlier versions of `xr.apply_ufunc`
except ValueError as e:
    if "with different chunksize present" in str(e):
        warnings.warn(str(e) + " This will throw an error in the future.")
        res = da.apply_gufunc(..., allow_rechunk=True)
    else:
        raise

```
However, I am only throwing around ideas - this is certainly not up to me. Let's see what the others think. Maybe they have an opinion or a better idea.

kmuehlbauer:
> What I was trying to say - this is a change that can break existing code and it is not easy to do a proper deprecation cycle (i.e. only throw a warning). (I think `allow_rechunk=True` is not a good default.)

Yes, sure, I didn't think about the effects on existing code. We might need to capture both ValueErrors and handle differently. I'll try to come up with a proposal for this.

mathause:
No need to worry about the second one:

https://github.com/pydata/xarray/blob/e5cc19cd8f8a69e0743f230f5bf51b7778a0ff96/xarray/core/computation.py#L704

shoyer:
Looking at the internals of `dask.array.apply_gufunc`, it looks like it passes on `**kwargs` to `blockwise`. `blockwise` gets called twice, first with only `**kwargs` and then with both `meta` and `**kwargs`:
https://github.com/dask/dask/blob/77628f2d5248cc61cc366cac3e400b6df5c654c1/dask/array/gufunc.py#L422
https://github.com/dask/dask/blob/77628f2d5248cc61cc366cac3e400b6df5c654c1/dask/array/gufunc.py#L439

So it seems we can actually still pass on a `meta` argument explicitly, at least in the first case.

Here's my suggestion for moving forwarding:

- If `meta` was explicitly set (`meta is not None`), pass it into `apply_gufunc` in `**kwargs`
- Otherwise, omit `meta`.

shoyer:
> * un-equal chunking along non-core dimensions
> 
> The second is not so trivial. I see three possibilities (1) just error, (2) try `dask.array.apply_gufunc` if that fails issue a warning and use the old `apply_blockwise` (3) figure out ourselves if non-core dimensions (called loop dimensions in dask) are not-equally chunked, issue a warning and re-chunk them ourselves. Maybe @shoyer and @dcherian can weight in here.

I am pretty confident that existing behavior of `xarray.apply_ufunc` with un-equal chunks along non-core dimensions is entirely broken. I am OK with just erroring for now.

kmuehlbauer:
Thanks @shoyer and @mathause. I've tried to adapt the PR to your comments.

- The `meta` and `output_sizes` parameters **should be** and the `vectorize` and `output_dtypes` parameters **can be** given within the `dask_gufunc_kwargs` (which have priority). This would also be the place for `allow_rechunk`. Everything is mentioned in the docstring and warnings are issued to use the new `dask_gufunc_kwargs`. The `dask_gufunc_kwargs` itself is used to hold all keyword arguments for `dask.array.apply_gufunc`.

- vectorization is done inside `apply_ufunc` only for `dask` not `parallelized`. In that case it is deferred to `dask.array.apply_gufunc`.

- for chunking problems (core dimension with multiple chunks and non-core dimensions with different chunksizes) we let dask raise the errors.

kmuehlbauer:
Currently tests are failing for some `test_missing`.  Any hints appreciated.

kmuehlbauer:
OK, vectorization has to be inside `apply_variable_ufunc` to be applied properly on dask and non-dask configurations.

For the time being I use `dask_gufunc_kwargs` to distribute `vectorize` and `output_dtypes` into `apply_variable_ufunc`. Maybe we have to decide on more sensible name for that dict.

kmuehlbauer:
@mathause I've added your tests from #4022 to this PR (dtype/meta).

kmuehlbauer:
@mathause Any idea why the meta test fails in `Linux py36-min-all-deps`?

mathause:
Could be "Avoid adjusting gufunc’s meta dtype twice" in https://docs.dask.org/en/latest/changelog.html#id84

Not sure if it is better to xfail or bump the version of dask (which could be done `ipython ci/min_deps_check.py requirements/py36-min-all-deps.yml`)

kmuehlbauer:
@mathause Thanks, I had just approached that particular line while debugging, when your message came in :grin:.  I'll add an xfail for that dask version first.

kmuehlbauer:
This is ready for another review cycle. :smiley:

xarray/tests/test_computation.py
```diff
@@ -818,6 +820,60 @@ def test_vectorize_dask():
...
+    actual = apply_ufunc(
+        identity,
+        data_array.chunk({"x": 1}),
+        vectorize=True,
+        dask="parallelized",
+        output_dtypes=[complex],
+    )
+    assert_identical(expected, actual)
+    assert expected.dtype == actual.dtype
+
```
mathause, line 854: Could you add two tests without `output_dtypes` and check the `dtype` of the result; something along the lines of:
```python
data_array = xr.DataArray([[0, 1, 2], [1, 2, 3]], dims=("x", "y"))
    actual = apply_ufunc(
        identity,
        data_array.chunk({"x": 1}),
        dask="parallelized",
    )

    actual = apply_ufunc(
        pandas_median,
        data_array.chunk({"x": 1}),
        input_core_dims=[["y"]],
        vectorize=True,
        dask="parallelized",
    )

```

xarray/core/computation.py
```diff
@@ -579,28 +590,39 @@ def apply_variable_ufunc(
...
         elif dask == "allowed":
             pass
         else:
             raise ValueError(
                 "unknown setting for dask array handling in "
                 "apply_ufunc: {}".format(dask)
             )
+    else:
+        if dask_gufunc_kwargs is not None:
+            if dask_gufunc_kwargs.setdefault("vectorize", False):
```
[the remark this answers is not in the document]
kmuehlbauer: I've changed the behaviour to simple .get, with default value. Although, `dask_gufunc_kwargs` should be already prefilled with the default values.

xarray/core/computation.py
```diff
@@ -579,28 +590,39 @@ def apply_variable_ufunc(
...
+                func = _vectorize(
+                    func, signature, dask_gufunc_kwargs.get("output_dtypes")
```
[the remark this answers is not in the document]
kmuehlbauer: I've changed the behaviour to simple .get, with default value. Although, `dask_gufunc_kwargs` should be already prefilled with the default values.

xarray/tests/test_computation.py
kmuehlbauer: I've added the suggested tests, but the second changes dtype, so it would fail. I just changed the assert. Please let me know, if this is what you had in mind.

kmuehlbauer:
@mathause Sorry for the delay. I have currently no other idea how to circumvent the use of `dask_gufunc_kwargs` also for the non-dask vectorize. I'm happy to adapt the code to a more convenient and readable solution.

xarray/tests/test_computation.py
```diff
@@ -1102,8 +1159,10 @@ def test_dot(use_dask):
         da_a = da_a.chunk({"a": 3})
         da_b = da_b.chunk({"a": 3})
         da_c = da_c.chunk({"c": 3})
-
+    # dask dot processing via apply_ufunc-> apply_gufunc is broken, due to chunking
+    # of core dimensions
```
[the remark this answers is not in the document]
kmuehlbauer: :+1:

xarray/tests/test_computation.py
```diff
@@ -818,6 +819,82 @@ def test_vectorize_dask():
...
+        identity, data_array.chunk({"x": 1}), vectorize=True, dask="parallelized",
+    )
+
+    assert_identical(expected, actual)
+    assert expected.dtype == actual.dtype
+
+    data_array = xr.DataArray([[0, 1, 2], [1, 2, 3]], dims=("x", "y"))
+    expected = xr.DataArray([1, 2], dims=["x"])
+    actual = apply_ufunc(
+        pandas_median,
```
mathause, line 866:
```suggestion
        identity,
```

xarray/tests/test_computation.py
```diff
@@ -818,6 +819,82 @@ def test_vectorize_dask():
...
+        data_array.chunk({"x": 1}),
+        input_core_dims=[["y"]],
+        vectorize=True,
+        dask="parallelized",
+    )
+
+    assert_identical(expected, actual)
+    assert actual.dtype == "float64"
```
mathause, line 874:
```suggestion
    assert actual.dtype == expected.dtype
```

xarray/tests/test_computation.py
```diff
@@ -818,6 +819,82 @@ def test_vectorize_dask():
...
+    expected = data_array.copy()
+    actual = apply_ufunc(
+        identity, data_array.chunk({"x": 1}), vectorize=True, dask="parallelized",
+    )
+
+    assert_identical(expected, actual)
+    assert expected.dtype == actual.dtype
+
+    data_array = xr.DataArray([[0, 1, 2], [1, 2, 3]], dims=("x", "y"))
+    expected = xr.DataArray([1, 2], dims=["x"])
```
mathause, line 864:
```suggestion
    expected = data_array.copy()
```

xarray/tests/test_computation.py
mathause: Ahh, indeed: `pd.Series([1, 2, 3]).median()` returns a float. The idea is to test if it can work without `output_dtypes`, but if `func` changes the dtype it won't work. Sorry my mistake - suggestions below.

xarray/core/computation.py
```diff
@@ -579,20 +589,40 @@ def apply_variable_ufunc(
...
+                        )
+                        res = da.apply_gufunc(
+                            numpy_func,
+                            str(signature),
+                            *arrays,
+                            **dask_gufunc_kwargs,
+                            allow_rechunk=True,
+                        )
+                    else:
+                        raise
```
mathause, line 618:
```suggestion
                res = da.apply_gufunc(
                    numpy_func, str(signature), *arrays, **dask_gufunc_kwargs,
                )
```

mathause: I think @shoyer suggested to just error here. (Also my mistake for leading you down this path.)

kmuehlbauer:
@mathause I reverted the mentioned error catching. I created a dedicated test for testing the dask vectorize without given dtype. Hope that fits your requirements.

kmuehlbauer:
@dcherian @shoyer Only the upstream-dev is failing with the sparse-tests (#4146). This is ready for another review cycle. Please let me know, if there is more to work on this PR.

xarray/core/computation.py
```diff
@@ -579,28 +589,41 @@ def apply_variable_ufunc(
...
+                import dask.array as da
+
+                res = da.apply_gufunc(
+                    numpy_func, str(signature), *arrays, **dask_gufunc_kwargs,
                 )
 
+                # todo: covers for https://github.com/dask/dask/pull/6207
+                #  remove when minimal dask version >= 2.17.0
+                if signature.num_outputs > 1:
+                    res = (*res,)
```
shoyer, line 607: Does this work any differently from `tuple(res)`? If not, then I would slightly lean towards using `tuple()` explicitly. This newer syntax is a little unfamiliar to me, but maybe I'll get used to it eventually :)

xarray/core/computation.py
```diff
@@ -579,28 +589,41 @@ def apply_variable_ufunc(
...
-                    input_dims,
-                    output_dims,
-                    signature,
-                    output_dtypes,
-                    output_sizes,
-                    meta,
+                import dask.array as da
+
+                res = da.apply_gufunc(
+                    numpy_func, str(signature), *arrays, **dask_gufunc_kwargs,
```
shoyer, line 601: I think it's important to use `to_gufunc_string()`:
```suggestion
                    numpy_func, signature.to_gufunc_string(), *arrays, **dask_gufunc_kwargs,
```

Otherwise, apply_ufunc would break when called on arrays with dimension names that are not valid Python identifiers.

xarray/core/computation.py
```diff
@@ -991,43 +941,49 @@ def earth_mover_distance(first_samples,
...
+                " It will be removed as direct parameter in a future version."
+            )
+            dask_gufunc_kwargs.setdefault("meta", meta)
+        if output_sizes is not None:
+            warnings.warn(
+                "``output_sizes`` should be given in the ``dask_gufunc_kwargs`` "
+                "parameter. It will be removed as direct parameter in a future "
+                "version."
+            )
+            dask_gufunc_kwargs.setdefault("output_sizes", output_sizes)
```
shoyer, line 971: If we use `to_gufunc_string()` in generating the signature for passing into dask, then we won't be able to entirely defer handling of `output_sizes`. So I think we should probably keep it as our own argument?

xarray/tests/test_computation.py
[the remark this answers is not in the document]
shoyer: Hmm, I don't think I fully understood this before.

My original thinking was that passing a NumPy array into `apply_gufunc` shouldn't really require `allow_rechunk=True`, because even though the NumPy array has different chunks, there is a difference between "no chunks" and "different chunks".

It seems like it should be OK for `apply_gufunc` to handle "no chunks"? In that case, the NumPy array could be given the same chunks as the other dask arrays.

Not a huge deal to include this change (there is an easy work around), but I could be worth fixing upstream in dask.

xarray/core/computation.py
kmuehlbauer: According to some resources at [SO](https://stackoverflow.com/questions/30216000/why-is-faster-than-list) use of literal syntax is faster than using  `tuple()`. I've no strong opinion about that and can change it to `tuple()`.

xarray/core/computation.py
kmuehlbauer: Hmm, here I would need a bit more assistance, example taken from `test_apply_dask_parallelized_errors()`

```python
import xarray as xr

import dask.array as da

array = da.ones((2, 2), chunks=(1, 1))
data_array = xr.DataArray(array, dims=("x", "y"))

def identity(x):
    return x

xr.apply_ufunc(
        identity,
        data_array,
        dask="parallelized",
        output_dtypes=[float],
        input_core_dims=[("y",)],
        output_core_dims=[("y",)],
    )
```
The error raised using `signature.to_gufunc_string()`

```
ValueError: Core dimension `'dim0'` consists of multiple chunks. To fix, rechunk into a single chunk along this dimension or set `allow_rechunk=True`, but beware that this may increase memory usage significantly.
```

The error raised using `str(signature)`

```
ValueError: Core dimension `'y'` consists of multiple chunks. To fix, rechunk into a single chunk along this dimension or set `allow_rechunk=True`, but beware that this may increase memory usage significantly.
```

For the latter I can create a working test, for the former not. I've no idea how to work around this.

kmuehlbauer: At least I would have my problems with the error message, because there is no `dim0` in the source data.

kmuehlbauer: @shoyer This test breaks within dask using `signature.to_gufunc_string()`:

```python
    import dask.array as da

    array = da.ones((2, 2), chunks=(1, 1))
    data_array = xr.DataArray(array, dims=("x", "y"))

    def stack_negative(obj):
        def func(x):
            return np.stack([x, -x], axis=-1)

        return apply_ufunc(
            func,
            obj,
            output_core_dims=[["sign"]],
            dask="parallelized",
            output_dtypes=[obj.dtype],
            output_sizes={"sign": 2},
        )
    stack_negative(data_array)
```

```
tests/test_computation.py:776: in stack_negative
    return apply_ufunc(
core/computation.py:1020: in apply_ufunc
    return apply_dataarray_vfunc(
core/computation.py:234: in apply_dataarray_vfunc
    result_var = func(*data_vars)
core/computation.py:630: in apply_variable_ufunc
    result_data = func(*input_data)
core/computation.py:600: in func
    res = da.apply_gufunc(
/home/kai/miniconda/envs/wradlib_38_01/lib/python3.8/site-packages/dask/array/gufunc.py:461: in apply_gufunc
    core_output_shape = tuple(core_shapes[d] for d in ocd)

.0 = <tuple_iterator object at 0x7f46d9d3e910>

>   core_output_shape = tuple(core_shapes[d] for d in ocd)
E   KeyError: 'dim0'

/home/kai/miniconda/envs/wradlib_38_01/lib/python3.8/site-packages/dask/array/gufunc.py:461: KeyError

```

kmuehlbauer: How can this be resolved properly?

xarray/core/computation.py
kmuehlbauer: Sorry, I'm not sure I understand.

xarray/core/computation.py
kmuehlbauer: Any ideas here?

mathause: According to the docstring `to_gufunc_string` "handles dimensions that don't map to Python identifiers". I am honestly not entirely sure what kind of dimension names these are? Non-string? 

Anyway - what it does is to map the dimension names like so: 
```
    "(x),(x,y)->(z)" -> "(dim0),(dim0,dim1)->(dim2)"
```

Now in `stack_negative` the name of `output_sizes` (`"sign"`) and the `signature` (`"()->(dim0)"`) are inconsistent. A possible way forward is:

1. re-map the names in `output_sizes` to the new names
2. Open a new issue to use the original names in `to_gufunc_string` as much as possible and fix this in a follow up PR. (I also find this confusing in an error message). For the `ValueError: Core dimension 'dim0'` I.e. I would leave this for the moment.

(Alternatively just use `str` and also defer 1. to a later PR. The proposed solution for #4130 would require using `to_gufunc_string` (i.e. I would probably do this))

---

To achieve 2, you would need to expose [dims_map](https://github.com/pydata/xarray/blob/a64cf2d5476e7bbda099b34c40b7be1880dbd39a/xarray/core/computation.py#L130) as a property in `_UFuncSignature` and then translate the `output_sizes` like so
```python
output_sizes = {"dim{}".format(signature.dims_map[key]): value for key, value in output_sizes.items()}
```

xarray/core/computation.py
mathause: I think Stephan refers to the problem you ran into with `stack_negative`.

xarray/core/computation.py
keewis: > I am honestly not entirely sure what kind of dimension names these are? Non-string?

In addition to non-string dimension names I think that also means all strings you can't use as a python variable (e.g. starting with a digit or containing a `-` or `.`)

kmuehlbauer: > (Alternatively just use `str` and defer 1. to a later PR. The proposed solution for #4130 would require using `to_gufunc_string` (i.e. I would probably do this))
> 
> To achieve 2, you would need to expose [dims_map](https://github.com/pydata/xarray/blob/a64cf2d5476e7bbda099b34c40b7be1880dbd39a/xarray/core/computation.py#L130) as a property in `_UFuncSignature` and then translate the `output_sizes` like so
> 
> ```python
> output_sizes = {"dim{}".format(signature.dims_map[key]): value for key, value in output_sizes.items()}
> ```

@mathause Your solution 2 looks better from my perspective. I'll try to apply to this.

xarray/core/computation.py
kmuehlbauer: changed to tuple, adapt dask version check

xarray/core/computation.py
kmuehlbauer: adapted to @mathause's suggestion (2)

xarray/core/computation.py
kmuehlbauer: Would it be worth to combine `output_core_dims` and `output_sizes` into one?

Like

```python
output_core_dims=[[("sign", 2)]]
```
or

```python
output_core_dims=[{"sign": 2}]
```

xarray/core/computation.py
mathause: Just came to my mind: it would be good to throw a nice error message `if key not in dims_map`. Or use `dims_map. get(key, key)` and let dask throw the error.

kmuehlbauer:
@mathause @shoyer @dcherian @max-sixty 

This PR has gone a long way from first enabling multiple outputs to implementation of `dask.apply_gufunc`. Thanks for all your help, comments and suggestions. As xarray 0.16.0 is handled in the next days (or so I understood from reading the comments here #4031), I'm asking myself, if this can be finalized for 0.16.0?

I appreciate if we could work on any remaining problems to get this PR merged.

xarray/core/computation.py
```diff
@@ -90,6 +92,10 @@ def all_core_dims(self):
             self._all_core_dims = self.all_input_core_dims | self.all_output_core_dims
         return self._all_core_dims
 
+    @property
+    def dims_map(self):
+        return dict(zip(sorted(self.all_core_dims), range(len(self.all_core_dims))))
```
mathause, line 97:
```suggestion
        return dict(zip(sorted(self.all_core_dims), "dim{}".format(range(len(self.all_core_dims)))))
```

xarray/core/computation.py
```diff
@@ -126,14 +132,12 @@ def to_gufunc_string(self):
         Unlike __str__, handles dimensions that don't map to Python
         identifiers.
         """
-        all_dims = self.all_core_dims
-        dims_map = dict(zip(sorted(all_dims), range(len(all_dims))))
         input_core_dims = [
-            ["dim%d" % dims_map[dim] for dim in core_dims]
+            ["dim%d" % self.dims_map[dim] for dim in core_dims]
```
mathause, line 136:
```suggestion
            [self.dims_map[dim] for dim in core_dims]
```

xarray/core/computation.py
```diff
@@ -126,14 +132,12 @@ def to_gufunc_string(self):
...
             for core_dims in self.input_core_dims
         ]
         output_core_dims = [
-            ["dim%d" % dims_map[dim] for dim in core_dims]
+            ["dim%d" % self.dims_map[dim] for dim in core_dims]
```
mathause, line 140:
```suggestion
            [self.dims_map[dim] for dim in core_dims]
```

xarray/core/computation.py
```diff
@@ -579,28 +592,55 @@ def apply_variable_ufunc(
...
-            input_dims = [broadcast_dims + dims for dims in signature.input_core_dims]
             numpy_func = func
 
+            if dask_gufunc_kwargs is None:
+                dask_gufunc_kwargs = {}
+
+            output_sizes = dask_gufunc_kwargs.pop("output_sizes", None)
+            if output_sizes is not None:
+                output_sizes = {
+                    "dim{}".format(signature.dims_map[key]): value
```
mathause, line 603:
```suggestion
                    signature.dims_map.get(key, key): value
```

If there is a `output_sizes` that is missing from the core_dims this will raise a informative error in dask (instead of a non-informative error here)

kmuehlbauer:
> I made one more suggestion to catch the case
> 
> ```python
> xr.apply_ufunc(..., output_core_dims=[["abc"]], output_sizes={"xyz": 2})
> ```

Yes, I've got the email, but did not find it here. Thanks! Would it not be useful to combine `output_core_dims` and `output_sizes`, eg.

```python
xr.apply_ufunc(..., output_core_dims=[{"abc": 2]])
```

mathause:
Yes, I think this got lost in a resolved comment.

`output_core_dims=[{"abc": 2}]` looks very elegant. But it is passed to `_UFuncSignature(input_core_dims, output_core_dims)` so you would probably need to take it apart again in the code (or adapt `_UFuncSignature` to handle this case). Not sure how involved this is. Could potentially be left for a future PR?

kmuehlbauer:
@mathause I've tested your suggestion but I do not know, what better error message you expect from the dask side. 

- xr.apply_ufunc(..., output_core_dims=[["z"]], output_sizes={"z": 1})
`signature.to_gufunc_string()` will output `()->(dim0)` (and `output_sizes` will be `{'dim0': 1}`
- xr.apply_ufunc(..., output_core_dims=[["z"]], output_sizes={"y": 1})
`signature.to_gufunc_string()` will output `()->(dim0)` (and `output_sizes` will be `{'y': 1}`

So the second will fail with `KeyError: 'dim0'`. 

I think we can't get proper error messages from dask side, because dask doesn't know anything about the new dimension name ( at least not from `signature.to_gufunc_string()`). Should we catch the error and add a meaningful output?

mathause:
Ok, yes I am happy with erroring ourselves.

kmuehlbauer:
@mathause It seems I can't find a nice way of checking/catching errors. 

For now, this fails with `KeyError` inside xarray when `output_sizes` contains a wrong dimension name and this fails with `KeyError` inside dask, but with the not helpful `dim0` message. Maybe this can be fixed in a follow up PR? 

I would also leave combining of `output_core_dims` and `output_sizes` for another PR.

mathause:
You could go for something like this:

```python
output_sizes_renamed = dict()
for key, value in output_sizes.items():

    if key not in signature.output_core_dims:
        raise ValueError("dimension name in 'output_sizes' must correspond to output_core_dims")
    output_sizes_renamed[signature.dims_map[key]] = value
```

Instead of
```python
output_sizes = {
    signature.dims_map[key]: value
    for key, value in output_sizes.items()
}
```

kmuehlbauer:
> ```python
> if key not in signature.output_core_dims:
> ```

This does not work, because signature.output_core_dims is a list of tuples. So this would raise in any case. But I've no idea how to properly extract this into a single list/tuple...

mathause:
Could you use `signature.all_output_core_dims` ?

kmuehlbauer:
@mathause This should now raise for both type of 'output_sizes' errors, wrong dim name in 'output_sizes' and missing dim name in 'output_sizes'. I'll catch up with any review tomorrow morning CET.

kmuehlbauer:
How should we proceed with this PR?

mathause:
I had another look and seems fine to me. So let's ping @shoyer & @dcherian for a review.

mathause:
gentle ping @shoyer @dcherian I think this is not too far away

xarray/core/computation.py
```diff
@@ -579,28 +597,65 @@ def apply_variable_ufunc(
...
+                    if key not in signature.all_output_core_dims:
+                        raise ValueError(
+                            f"dimension name '{key}' in 'output_sizes' must correspond to output_core_dims"
+                        )
+                    output_sizes_renamed[signature.dims_map[key]] = value
+                dask_gufunc_kwargs["output_sizes"] = output_sizes_renamed
+
+            for key in signature.all_output_core_dims:
+                if key not in signature.all_input_core_dims:
+                    if key not in output_sizes:
```
dcherian, line 618: combine with above `if` using `and`?

xarray/core/computation.py
```diff
@@ -579,28 +597,65 @@ def apply_variable_ufunc(
...
+
         elif dask == "allowed":
             pass
         else:
             raise ValueError(
                 "unknown setting for dask array handling in "
                 "apply_ufunc: {}".format(dask)
             )
+    else:
+        if dask_gufunc_kwargs is not None:
```
dcherian, line 651: I think this is part of Stephan's earlier point.

It's confusing that we are looking for `vectorize` and `output_dtypes` in `dask_gufunc_kwargs` when we are not using `apply_gufunc`. I think we should  pass these two kwargs explicitly in for clarity.

kmuehlbauer:
@dcherian Thanks for your latest review comments. I'll try to come up with a solution after merging with latest master (and resolving conflicts).

xarray/core/computation.py
kmuehlbauer: done

xarray/core/computation.py
kmuehlbauer: done

kmuehlbauer:
@mathause I've merged with master ( + resolved conflicts) and addressed the latest review comments by @dcherian.

doc/whats-new.rst
```diff
@@ -128,6 +128,8 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+- Support multiple outputs in :py:func:`xarray.apply_ufunc` when using ``dask='parallelized'``. (:issue:`1815`, :pull:`4060`)
+  By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
```
keewis, line 132: can you move this to the new section (0.16.1)?

doc/whats-new.rst
```diff
@@ -296,6 +298,8 @@ Documentation
 
 Internal Changes
 ~~~~~~~~~~~~~~~~
+- Use :py:func:`dask.array.apply_gufunc` instead of :py:func:`dask.array.blockwise` in
+  :py:func:`xarray.apply_ufunc` when using ``dask='parallelized'``. (:pull:`4060`)
```
keewis, line 302: same here, please move this to 0.16.1

xarray/core/computation.py
```diff
@@ -90,6 +92,15 @@ def all_core_dims(self):
...
         return self._all_core_dims
 
+    @property
+    def dims_map(self):
+        return dict(
+            zip(
+                sorted(self.all_core_dims),
+                ["dim{}".format(d) for d in range(len(self.all_core_dims))],
+            )
+        )
```
keewis, line 102:
```suggestion
        return {
            f"dim{n}": core_dim
            for n, core_dim in enumerate(sorted(self.all_core_dims))
        }
```

xarray/core/computation.py
```diff
@@ -579,28 +599,61 @@ def apply_variable_ufunc(
                 "``.load()`` or ``.compute()``"
             )
         elif dask == "parallelized":
-            input_dims = [broadcast_dims + dims for dims in signature.input_core_dims]
             numpy_func = func
 
+            if dask_gufunc_kwargs is None:
+                dask_gufunc_kwargs = {}
+
+            output_sizes = dask_gufunc_kwargs.pop("output_sizes", dict())
```
keewis, line 607:
```suggestion
            output_sizes = dask_gufunc_kwargs.pop("output_sizes", {})
```

xarray/core/computation.py
```diff
@@ -579,28 +599,61 @@ def apply_variable_ufunc(
...
+            if output_sizes:
+                output_sizes_renamed = dict()
+                for key, value in output_sizes.items():
+                    if key not in signature.all_output_core_dims:
+                        raise ValueError(
+                            f"dimension name '{key}' in 'output_sizes' must correspond to output_core_dims"
+                        )
+                    output_sizes_renamed[signature.dims_map[key]] = value
```
keewis, line 615: this part could be moved into a helper function and called in a dict comprehension, but I guess it's also fine to leave it as is (IIRC dictionary comprehensions tend to be faster than assigning lots of keys to a existing dictionary, but in this case there's usually not enough keys for this to really make a difference).

xarray/core/computation.py
```diff
@@ -579,28 +599,61 @@ def apply_variable_ufunc(
...
+                    **dask_gufunc_kwargs,
                 )
 
+                # todo: covers for https://github.com/dask/dask/pull/6207
+                #  remove when minimal dask version >= 2.17.0
+                from dask import __version__ as dask_version
+
+                if LooseVersion(dask_version) < LooseVersion("2.17.0"):
+                    if signature.num_outputs > 1:
+                        res = tuple(r for r in res)
```
keewis, line 642: nit: this could be just
```suggestion
                        res = tuple(res)
```

xarray/core/computation.py
```diff
@@ -857,19 +827,29 @@ def apply_ufunc(
...
+          inputs are a dask array by using `dask.array.apply_gufunc`. Multiple output
+          arguments are supported. Only use this option if ``func`` does not natively
+          support dask arrays (e.g. converts them to numpy arrays).
+    dask_gufunc_kwargs : dict, optional
+        Optional keyword arguments passed to ``dask.array.apply_gufunc`` if
+        dask='parallelized'. Possible keywords are ``output_sizes``, ``allow_rechunk``
+        and ``meta``.
     output_dtypes : list of dtypes, optional
-        Optional list of output dtypes. Only used if dask='parallelized'.
+        Optional list of output dtypes. Only used if dask='parallelized' or
```
keewis, line 841:
```suggestion
        Optional list of output dtypes. Only used if ``dask='parallelized'`` or
```

xarray/core/computation.py
mathause: I copy my comment over for completeness:

> `output_core_dims=[{"abc": 2}]` looks very elegant. But it is passed to `_UFuncSignature(input_core_dims, output_core_dims)` so you would probably need to take it apart again in the code (or adapt `_UFuncSignature` to handle this case). Not sure how involved this is. Could potentially be left for a future PR?

xarray/tests/test_computation.py
```diff
@@ -818,6 +820,85 @@ def test_vectorize_dask():
...
+    assert expected.dtype == actual.dtype
+
+    # complex
+    data_array = xr.DataArray([[0 + 0j, 1 + 2j, 2 + 1j]], dims=("x", "y"))
+    expected = data_array.copy()
+    actual = apply_ufunc(
+        identity, data_array.chunk({"x": 1}), vectorize=True, dask="parallelized",
+    )
+    assert_identical(expected, actual)
+    assert expected.dtype == actual.dtype
```
keewis, line 878: you could use `pytest.mark.parametrize` to pass different `DataArray` instances. That way, we have separate tests and the `int` test cannot shadow the `complex` test.

xarray/core/computation.py
```diff
@@ -579,28 +599,61 @@ def apply_variable_ufunc(
...
+                        raise ValueError(
+                            f"dimension name '{key}' in 'output_sizes' must correspond to output_core_dims"
+                        )
+                    output_sizes_renamed[signature.dims_map[key]] = value
+                dask_gufunc_kwargs["output_sizes"] = output_sizes_renamed
+
+            for key in signature.all_output_core_dims:
+                if key not in signature.all_input_core_dims and key not in output_sizes:
+                    raise ValueError(
+                        f"dimension name '{key}' in 'output_core_dims' needs corresponding (key,value) in 'output_sizes'"
```
mathause, line 621:
```suggestion
                        f"dimension '{key}' in 'output_core_dims' needs corresponding (dim, size) in 'output_sizes'"
```
?
Or just a space
```suggestion
                        f"dimension '{key}' in 'output_core_dims' needs corresponding (key, value) in 'output_sizes'"
```

kmuehlbauer: @mathause Maybe?

```python
f"dimension '{key}' in 'output_core_dims' needs corresponding (dim, size) in 'output_sizes'"
```

kmuehlbauer:
@keewis @mathause Thanks for the review. I've added a checklist in the first post with "open issues" with this PR, which might be solved in a follow up PR. Would be good to know which need to go in here, so I can add this.

keewis:
I think all of these can be done in a new PR, we just have to make sure to include these in the next release (which might need to be soon so we regain compatibility with the most recent `pandas`).

kmuehlbauer:
Great, than it looks like it's finally done. :smiley:

kmuehlbauer:
While having a last review I've found another small glitch. I'll come back the next days to see, if anything needs to be done from reviewers side.

kmuehlbauer:
@mathause I've merged latest master into this PR to hopefully get all tests green. The former had some problems with a conda error in MinimumVersions job.

Please let me know, if there is anything for me to do, to get this merged.

xarray/core/computation.py
```diff
@@ -1006,34 +983,41 @@ def earth_mover_distance(first_samples,
...
+    # handle dask_gufunc_kwargs
+    if dask == "parallelized":
+        if dask_gufunc_kwargs is None:
+            dask_gufunc_kwargs = {}
+        # todo: remove warnings after deprecation cycle
+        if meta is not None:
+            warnings.warn(
+                "``meta`` should be given in the ``dask_gufunc_kwargs`` parameter."
+                " It will be removed as direct parameter in a future version."
+            )
```
shoyer, line 995: Could you please set a class (DeprecationWarning) and stacklevel=2 on these warnings? That results in better messages for users.

mathause: Sorry to nitpick - shouldn't that be a `FutureWarning` so that users actually get to see it?

kmuehlbauer: @mathause At least in the tests the [warnings are issued ](https://dev.azure.com/xarray/xarray/_build/results?buildId=3580&view=logs&j=ba13898e-1dfb-5ace-9966-8b7af3677790&t=2bb5a181-aa33-506a-9515-1f09b3e5c016&l=395). 

What's the actual difference between `DeprecationWarning` and `FutureWarning`? And when should they be used? Just to know for future contributions.

shoyer: FutureWarning would be fine, too. We should probably try to come to consensus on a general policy for xarray.

The Python docs have some guidance but the overall recommendation is not really clear to me: https://docs.python.org/3/library/warnings.html#warning-categories

mathause: `FutureWarning` is for users and `DeprecationWarning` for library authors (https://docs.python.org/3/library/warnings.html#warning-categories). Which is why you see `DeprecationWarning` in the test but won't when you execute the code. Took me a while to figure this out when I wanted to deprecate some stuff in my package.

```python
import warnings

def test():
    warnings.warn("DeprecationWarning", DeprecationWarning)
    warnings.warn("FutureWarning", FutureWarning)
```

If you try this in ipython `test()` will raise both warnings. But if you save to a file and try

```python
from test_warnings import test
test()
```
only `FutureWarning` will appear (I did not know this detail either https://www.python.org/dev/peps/pep-0565/).

kmuehlbauer: @mathause @shoyer I'll switch to `FutureWarning` since this seems to be the only user-visible warning, See https://www.python.org/dev/peps/pep-0565/#id13

kmuehlbauer: And, thanks for the pointers and explanations.

mathause:
Nice! Unless @dcherian has any additional comments I'll merge in a few days

xarray/tests/test_sparse.py
```diff
@@ -877,13 +877,13 @@ def test_dask_token():
...
-def test_apply_ufunc_meta_to_blockwise():
-    da = xr.DataArray(np.zeros((2, 3)), dims=["x", "y"]).chunk({"x": 2, "y": 1})
-    sparse_meta = sparse.COO.from_numpy(np.zeros((0, 0)))
+def test_apply_ufunc_check_meta_coherence():
+    s = sparse.COO.from_numpy(np.array([0, 0, 1, 2]))
+    a = DataArray(s)
+    ac = a.chunk(2)
+    sparse_meta = ac.data._meta
 
     # if dask computed meta, it would be np.ndarray
```
dcherian, line 886:
```suggestion

```

dcherian (approved):
Looks great! Thanks @kmuehlbauer this is a great improvement!

mathause:
ok then - let's do this. Thanks a lot @kmuehlbauer

kmuehlbauer:
Thanks to all reviewers! Great job!

PR #4064, opened 2020-05-15
Auto chunk

### LINKED ISSUE #4055: Automatic chunking of arrays ?
Hi there,

Hopefully this turns out to be a basic issue, but I was wondering why the `chunks='auto'` that dask seems to provide (https://docs.dask.org/en/latest/array-chunks.html#automatic-chunking) isn't an option for xarray? I'm not 100% sure of how dask decides how to automatically chunk its arrays, so maybe there's a compatibility issue? 

I get the impression that the dask method automatically tries to prevent the issues of "too many chunks" or "too few chunks" which can sometimes happen when choosing chunk sizes automatically. If so, it would maybe be a useful thing to include in future versions?

Happy to be corrected if I've misunderstood something here though, still getting my head around how the dask/xarray compatibility really works...

Cheers!

AndrewWilliams3142:
Adding `chunks='auto'` option to `Dataset.chunk()`. 

 - [x] Closes #4055
 - [x] Tests added in `test_dask.py`
 - [x] Updated `whats-new.rst` for changes


AndrewWilliams3142:
In my git clone, when I run the `flake8` and `black .` tests, I get the following messages.

```
(xarray-tests) Andrews-MacBook-Pro-2:xarray andrewwilliams$ black .
All done! ✨ 🍰 ✨
143 files left unchanged.
(xarray-tests) Andrews-MacBook-Pro-2:xarray andrewwilliams$ flake8
./xarray/backends/memory.py:43:32: E741 ambiguous variable name 'l'
./xarray/backends/common.py:244:32: E741 ambiguous variable name 'l'
./xarray/backends/.ipynb_checkpoints/memory-checkpoint.py:43:32: E741 ambiguous variable name 'l'
```

I'm not sure why something has changed in these files (I haven't touched them), I also can't work out what the `l` variable is meant to be doing there. Any tips appreciated!

kmuehlbauer:
> I'm not sure why something has changed in these files (I haven't touched them), I also can't work out what the `l` variable is meant to be doing there. Any tips appreciated!

@AndrewWilliams3142 This is due to flake8 applying new changes from pycodestyle. xarray-devs already dived into this, see https://github.com/pydata/xarray/pull/4057.

You might just need to rebase with current master, to make the error go away.

keewis:
no need to rebase, simply merge `master`. Also, you'd then have to remove the checkpoint.

AndrewWilliams3142:
Okay cheers both! I'll have a look at these now.

keewis:
to merge `master`, what I usually do is:
```bash
$ git checkout master
$ git pull  # synchronize master with `origin/master`
$ git checkout <feature-branch>
$ git merge master
```
and if there are merge conflicts (i.e. the merge was interrupted), I follow the advice given by `git status`

AndrewWilliams3142:
Okay, that makes sense. Though, it seems that I forked the master branch before @kmuehlbauer's commit, which fixed this flake8 issue? So I think I need to make a new fork?

keewis:
no, the `git pull` or `git pull origin master` (`git pull <remote> <branch>` in general) should do the trick. "Forking" just means to setup your own repository on github as a mirror of the repository you're forking. After that, `git` can handle the synchronization between both repositories.

If you need more explanations on `git`, how to use it and also how it works under the hood, check out the [git book](https://git-scm.com/book/en/v2).

keewis:
there's something wrong with the merge. Are you able to resolve that by yourself, or should I fix it for you?

AndrewWilliams3142:
Do you mean the master merge? If that's wrong would you be able to fix it for me? My bad, hopefully i'll be able to do it more cleanly in future

keewis:
if you have any unpushed commits: could you push them now? While they would not be lost, it's way easier to handle them now than after the force-push

AndrewWilliams3142:
No unpushed commits

keewis:
then you'll have to run `git pull -f` once to get the new state

For future reference, github does not work terribly well with rebases (or merges without a merge commit) so it would be good to avoid those.

AndrewWilliams3142:
Okay so I've traced the error back to the `map_blocks()` function. I don't fully understand the code for this function in `xarray/core/parallel.py`, but here's a quick report on the different behaviours.

Normally, when using the `make_ds()` and `make_da()` functions in `test_dask.py`, **without any changes to `ds.chunk()`** we have:

```python
>>> def func(obj):
...         result = obj + obj.x + 5 * obj.y
...         return result
...
>>> xr.map_blocks(func, ds).unify_chunks().chunks
Frozen(SortedKeysDict({'x': (4, 4, 2), 'y': (5, 5, 5, 5), 'z': (4,)}))
>>> func(ds).chunk().unify_chunks().chunks
Frozen(SortedKeysDict({'x': (4, 4, 2), 'y': (5, 5, 5, 5), 'z': (4,)}))
```

However, when I use the changes I've made to `dataset.py` (changing `isinstance(chunks, Number)` to `is_scalar(chunks)`), the behaviour becomes:

```python
>>> xr.map_blocks(func, ds).unify_chunks().chunks
Frozen(SortedKeysDict({'x': (4, 4, 2), 'y': (5, 5, 5, 5), 'z': (4,)}))
>>> func(ds).chunk().unify_chunks().chunks
Frozen(SortedKeysDict({'x': (10,), 'y': (20,), 'z': (4,)}))
```

Which means that it now fails the `test_map_blocks()` call in `test_dask.py` line 1077. 

I've tried to follow through the code and see what is actually happening when this change is made, but I'm out of my depth here. My guess is that `is_scalar(chunks)` is giving the wrong behaviour when `chunks=None` ?

dcherian:
ah right. this is now rechunking chunked objects to a single chunk when `chunks=None`. This was not happening previously

AndrewWilliams3142:
@dcherian do you have any idea about this `mypy` Type error? I can't find much (accessible) documentation on how the `Union[]` is working in this context. 

```
xarray/core/dataset.py:1737: error: Argument 2 to "fromkeys" of "dict" has incompatible type "Union[Number, Mapping[Hashable, Union[None, Number, Tuple[Number, ...]]]]"; expected "Union[None, Number, Tuple[Number, ...]]"
xarray/core/dataset.py:1740: error: Item "Number" of "Union[Number, Mapping[Hashable, Union[None, Number, Tuple[Number, ...]]]]" has no attribute "keys"
```

xarray/core/dataset.py
```diff
@@ -1733,7 +1733,7 @@ def chunk(
         """
         from dask.base import tokenize
 
-        if isinstance(chunks, Number):
+        if is_scalar(chunks) and chunks is not None:
```
mathause, line 1736: Should this be
```suggestion
        if isinstance(chunks, (Number, str)):
```

or

```suggestion
        if isinstance(chunks, Number) or chunks == "auto":
```
?

AndrewWilliams3142: Thanks! I initially had used `if isinstance(chunks, Number) or chunks == "auto":` but then changed it after suggestions in #4055 , however I think it makes sense to revert now.

AndrewWilliams3142: I'll try using `None, Number, str, Mapping[Hashable, Union[None, Number, str, Tuple[Number, ...]]]` tomorrow :) thanks

mathause: The mypy error probably stems from the fact that
```python
xr.core.utils.is_scalar(Mapping[Hashable, Union[None, Number, Tuple[Number, ...]]])
```
returns `True` (although I am not sure why).

Maybe you'll need `Literal["auto"]` instead of `str`.

AndrewWilliams3142: It seem that `isinstance(chunks, (Number, str))` is the more robust solution, because otherwise if only checking `chunks="auto"` then generic strings can still get through. 

I've also added `None, Number, str, Mapping[Hashable, Union[None, Number, str, Tuple[Number, ...]]]` and this appears to have fixed the `mypy` argument.

Edit: Sorry for all the extra commits, I keep forgetting to run the code format checks!

keewis: don't worry about the commit history: only a single commit will end up in `master` since we squash all commits when merging a PR (as an example, you can take a look at #4012 which became 8834afa9f617bd201eba00374bb55d96dccec96b)

dcherian:
Thanks @AndrewWilliams3142 . We should add a test for `chunk("auto")` . This could test that `dataarray.chunk("auto").data` is the same as `dataarray.data.rechunk("auto")` (or something like that).

AndrewWilliams3142:
Cheers! I forgot about the tests, will add them this week or next hopefully

AndrewWilliams3142:
> This could test that `dataarray.chunk("auto").data` is the same as `dataarray.data.rechunk("auto")` (or something like that).

@dcherian Thanks for the tip:) Quick question: Is there a reason why you're specifying the `.data` here? Also I think I'm missing something because I don't get what the difference between `.chunk()` and `.rechunk()` would be in this case.

keewis:
`chunk` is a method of `DataArray` while `rechunk` is a method of `dask.array.Array`

`dataarray.chunk("auto").data` calls `DataArray.chunk` and then fetches the underlying data structure (a `dask` array) while `dataarray.data.rechunk("auto")` fetches the underlying data structure and then calls `dask.array.Array.rechunk`.

AndrewWilliams3142:
@keewis thanks for this! I've added what I think is a suitable test for `DataArray`s, do you think it's also a good idea to have a `DataSet` test?

xarray/tests/test_dask.py
```diff
@@ -1035,6 +1035,13 @@ def test_unify_chunks_shallow_copy(obj, transform):
     assert_identical(obj, unified) and obj is not obj.unify_chunks()
 
 
+@pytest.mark.parametrize("obj", [make_da()])
+def test_auto_chunk_da(obj):
+    expected = obj.chunk("auto").data
+    actual = obj.data.rechunk("auto")
```
keewis, line 1041: this should be the other way around :wink:

the result of `dask`'s `rechunk` is what we expect, actual should be what `chunk` returns. That doesn't change the result of the tests, but it makes identifying the test subject a bit easier.

xarray/tests/test_dask.py
```diff
@@ -1035,6 +1035,13 @@ def test_unify_chunks_shallow_copy(obj, transform):
...
+    np.testing.assert_array_equal(actual, expected)
```
dcherian, line 1042: we should check that the chunk sizes are equal so `assert actual.chunks == expected.chunks`?  Not sure a dataset test is really necessary since both dataarrays and datasets go through the same code path.

xarray/tests/test_dask.py
AndrewWilliams3142: aaaaah jeez haha 

noted

xarray/tests/test_dask.py
AndrewWilliams3142: Added!

dcherian:
Thanks @AndrewWilliams3142

AndrewWilliams3142:
No problem ! Thanks everyone for helping me get up to speed :)

PR #4067, opened 2020-05-15
make text wrap width an argument in label_from_attrs

aspitarl:
label_from_attrs used textwrap.wrap with a default wrap width of 30,
this commit changes label_from_attrs to instead take an argument
wrap_width that specifies the wrap width.

context: I find the label_from_attrs function useful for plots that I make using pyplot, but needed to change the text wrap with. 

I am new to contributing to projects like this so bear with me.  I only had the mypy and flake8 linters and figured that was good enough for a minor change like this. And I didn't see label_from_attrs in the api document so just added a line to whats-new.rst

 - [x] Passes  mypy . && flake8`

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -36,6 +36,7 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+- 'wrap_amount' argument added to :py:func:'label_from_attrs' to specify text wrap amount for labels
 - Added :py:meth:`DataArray.polyfit` and :py:func:`xarray.polyval` for fitting polynomials. (:issue:`3349`)
   By `Pascal Bourgault <https://github.com/aulemahal>`_.
 - Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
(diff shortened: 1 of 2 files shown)


TomNicholas:
Hi Lee, thanks for this.

> I only had the mypy and flake8 linters and figured that was good enough for a minor change like this.

As the tests all passed then that means the changes fit those style requirements :)

> And I didn't see label_from_attrs in the api document so just added a line to whats-new.rst

So `label_from_attrs` isn't public API, which we could change, but should be considered. If we make it public API, then we would need to not change/delete it in future between versions without warning, otherwise some users' code (such as yours) would break unexpectedly.

This additional maintenance needs to be weighed against the advantages of making the function public API. Are you planning to change something else within xarray that will later use this change?

aspitarl:
I have no plans to alter xarray further in ways that make use of this function. I have simply been importing the function for use on its own in scripts/notebooks and have needed to change the wrap width. I don't think it necessarily needs to be added to the public API.

aspitarl:
@TomNicholas Hi, any update on this being able to get implemented? Should I resolve the merge conflicts?

dcherian:
Sorry but this function isn't public API. I recommend copying it over to your code .

PR #4070, opened 2020-05-16
remove the backslash escapes and typehint fragments in the API docs

### LINKED ISSUE #3178: type annotations make docs confusing

keewis:
The reason for those fragments is that `napoleon` and `numpydoc` somehow conflict with each other. From a quick glance I couldn't find anything that breaks without `numpydoc`, but I might be missing something.

For reference, here's the new [API page](https://xarray-keewis.readthedocs.io/en/fix-signature-escapes/api.html).
 - [x] Closes #3178

### DIFF
--- a/ci/requirements/doc.yml
+++ b/ci/requirements/doc.yml
@@ -17,7 +17,6 @@ dependencies:
   - netcdf4>=1.5
   - numba
   - numpy>=1.17
-  - numpydoc
   - pandas>=1.0
   - rasterio>=1.1
   - seaborn
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -79,7 +79,6 @@ extensions = [
     "sphinx.ext.extlinks",
     "sphinx.ext.mathjax",
     "sphinx.ext.napoleon",
-    "numpydoc",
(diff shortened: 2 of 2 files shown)


dcherian:
Nice! the apply-ufunc signature is readable again. Should close #3178

image

keewis:
is the (seemingly random) `distributed` failure a known issue?

keewis:
there is a difference between the versions with and without `numpydoc`: the reference pages for the ufuncs doesn't show the signature. Since that would always have been `(*args, **kwargs)` we don't lose anything, I guess?

max-sixty:
Great! Thanks @keewis !

dcherian:
> : the reference pages for the ufuncs don't show the signature.

I think this is a relatively minor issue for a really large improvement in signatures for other functions

keewis:
if there are no issues without `numpydoc`, should we merge?

dcherian:
SOudns good to me!

PR #4071, opened 2020-05-16
#1621 optional decode timedelta

aurghs:
Releated to ticket #1621. Add `decode_timedelta` kwargs in open_dataset, `xr.open_datarray`, `xr.open_zarr`, `xr.decode_cf`. If not passed explicitly the behaviour is not changed.
 - [x] Tests added for `xr.decode_cf`. 

### DIFF
--- a/xarray/backends/api.py
+++ b/xarray/backends/api.py
@@ -303,6 +303,7 @@ def open_dataset(
     drop_variables=None,
     backend_kwargs=None,
     use_cftime=None,
+    decode_timedelta=None,
 ):
     """Open and decode a dataset from a file or file-like object.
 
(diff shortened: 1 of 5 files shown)


xarray/backends/api.py
```diff
@@ -477,6 +485,7 @@ def maybe_decode_store(store, lock=False):
                 chunks,
                 drop_variables,
                 use_cftime,
+                decode_timedelta,
```
shoyer, line 488: In theory, it would be better organized to add this argument after `decode_times`, but for now putting it at the end is a good choice to preserve backwards compatibility. I made a note about making most `open_dataset` arguments keyword only in https://github.com/pydata/xarray/issues/4080.

shoyer:
Thank you @aurghs, I'm looking forward to more great PRs!

PR #4089, opened 2020-05-23
xr.cov() and xr.corr()

### LINKED ISSUE #3784: Function for regressing/correlating multiple fields?
I came across this StackOverflow thread about applying linear regression across multi-dimensional arrays and there were some very efficient, xarray-based solutions suggested to the OP which leveraged vectorized operations. I was wondering if there was any interest in adding a function like this to future releases of xarray? It seems like a very neat and useful functionality to have, and not just for meteorological applications!

https://stackoverflow.com/questions/52108417/how-to-apply-linear-regression-to-every-pixel-in-a-large-multi-dimensional-array

Most answers draw from this blog post: 
https://hrishichandanpurkar.blogspot.com/2017/09/vectorized-functions-for-correlation.html

AndrewWilliams3142:
**PR** for the `xr.cov()` and `xr.corr()` functionality which others have been working on. Most code adapted from @r-beer in PR #3550. 

TODO:
 - [x] ~Write a reasonable set of tests, maybe not using `pandas` as a benchmark? (See https://github.com/pydata/xarray/issues/3784#issuecomment-633145887) Will probably need some help with this~

CHECKLIST:
 - [x] Closes #3784, #3550, #2652, #1115 
 - [x] Tests added
 - [x] Passes `isort -rc . && black . && mypy . && flake8` (something wrong with docs though??)

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -36,6 +36,8 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+- Added :py:func:`xarray.cov` and :py:func:`xarray.corr` (:issue:`3784`, :pull:``).
+  By `Andrew Williams <https://github.com/AndrewWilliams3142>`_ and `Robin Beer <https://github.com/r-beer>`_.
 - Added :py:meth:`DataArray.polyfit` and :py:func:`xarray.polyval` for fitting polynomials. (:issue:`3349`)
   By `Pascal Bourgault <https://github.com/aulemahal>`_.
 - Control over attributes of result in :py:func:`merge`, :py:func:`concat`,
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -36,7 +36,7 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
-- Added :py:func:`xarray.cov` and :py:func:`xarray.corr` (:issue:`3784`, :pull:``).
+- Added :py:func:`xarray.cov` and :py:func:`xarray.corr` (:issue:`3784`, :pull:`3550`, :pull:`4089`).
   By `Andrew Williams <https://github.com/AndrewWilliams3142>`_ and `Robin Beer <https://github.com/r-beer>`_.
 - Added :py:meth:`DataArray.polyfit` and :py:func:`xarray.polyval` for fitting polynomials. (:issue:`3349`)
   By `Pascal Bourgault <https://github.com/aulemahal>`_.
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -24,7 +24,7 @@ from typing import (
 import numpy as np
 
 from . import dtypes, duck_array_ops, utils
-from .alignment import broadcast, deep_align
+from .alignment import align, deep_align
 from .merge import merge_coordinates_without_align
 from .options import OPTIONS
 from .pycompat import dask_array_type
@@ -1072,6 +1072,7 @@ def apply_ufunc(
 def cov(da_a, da_b, dim=None, ddof=1):
     """
     Compute covariance between two DataArray objects along a shared dimension.
+
     Parameters
     ----------
     da_a: DataArray (or Variable) object
@@ -1082,6 +1083,7 @@ def cov(da_a, da_b, dim=None, ddof=1):
         The dimension along which the covariance will be computed
     ddof: int
         If ddof=1, covariance is normalized by N-1, giving an unbiased estimate.
+        
     Returns
     -------
     covariance: DataArray
@@ -1089,6 +1091,7 @@ def cov(da_a, da_b, dim=None, ddof=1):
     --------
     pandas.Series.cov: corresponding pandas function
     xr.corr: respective function to calculate correlation
+
     Examples
     --------
     >>> da_a = DataArray(np.random.random((3, 5)),
@@ -1126,19 +1129,20 @@ def cov(da_a, da_b, dim=None, ddof=1):
     """
     from .dataarray import DataArray
 
-    if any(not isinstance(arr, (Variable, DataArray)) for arr in [da_a, da_b]):
+    if any(not isinstance(arr, DataArray) for arr in [da_a, da_b]):
         raise TypeError(
-            "Only xr.DataArray and xr.Variable are supported."
+            "Only xr.DataArray is supported."
             "Given {}.".format([type(arr) for arr in [da_a, da_b]])
         )
 
     return _cov_corr(da_a, da_b, dim=dim, ddof=ddof, method="cov")
 
 
-def corr(da_a, da_b, dim=None, ddof=0):
+def corr(da_a, da_b, dim=None):
     """
     Compute the Pearson correlation coefficient between
     two DataArray objects along a shared dimension.
+
     Parameters
     ----------
     da_a: DataArray (or Variable) object
@@ -1147,6 +1151,7 @@ def corr(da_a, da_b, dim=None, ddof=0):
         Array to compute.
     dim: str, optional
         The dimension along which the correlation will be computed
+
     Returns
     -------
     correlation: DataArray
@@ -1154,6 +1159,7 @@ def corr(da_a, da_b, dim=None, ddof=0):
     --------
     pandas.Series.corr: corresponding pandas function
     xr.cov: underlying covariance function
+
     Examples
     --------
     >>> da_a = DataArray(np.random.random((3, 5)),
@@ -1191,9 +1197,9 @@ def corr(da_a, da_b, dim=None, ddof=0):
     """
     from .dataarray import DataArray
 
-    if any(not isinstance(arr, (Variable, DataArray)) for arr in [da_a, da_b]):
+    if any(not isinstance(arr, DataArray) for arr in [da_a, da_b]):
         raise TypeError(
-            "Only xr.DataArray and xr.Variable are supported."
+            "Only xr.DataArray is supported."
             "Given {}.".format([type(arr) for arr in [da_a, da_b]])
         )
 
@@ -1208,7 +1214,7 @@ def _cov_corr(da_a, da_b, dim=None, ddof=0, method=None):
     from .dataarray import DataArray
 
     # 1. Broadcast the two arrays
-    da_a, da_b = broadcast(da_a, da_b)
+    da_a, da_b = align(da_a, da_b, join="inner", copy=False)
 
     # 2. Ignore the nans
     valid_values = da_a.notnull() & da_b.notnull()
@@ -1229,12 +1235,12 @@ def _cov_corr(da_a, da_b, dim=None, ddof=0, method=None):
     cov = (demeaned_da_a * demeaned_da_b).sum(dim=dim, skipna=False) / (valid_count)
 
     if method == "cov":
-        return DataArray(cov)
+        return cov
 
     else:
         # compute corr
         corr = cov / (da_a_std * da_b_std)
-        return DataArray(corr)
+        return corr
 
 
 def dot(*arrays, dims=None, **kwargs):
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -1129,7 +1129,6 @@ def cov(da_a, da_b, dim=None, ddof=1):
 
     # 2. Ignore the nans
     valid_values = da_a.notnull() & da_b.notnull()
-    # TODO: avoid drop
     da_a = da_a.where(valid_values)
     da_b = da_b.where(valid_values)
     valid_count = valid_values.sum(dim) - ddof
@@ -1211,8 +1210,6 @@ def corr(da_a, da_b, dim=None, ddof=0):
 
     # 2. Ignore the nans
     valid_values = da_a.notnull() & da_b.notnull()
-    # TODO: avoid drop https://github.com/pydata/xarray/pull/2652#discussion_r245492002
-    # FIXED?: I think @shoyer convinced that you can just remove drop=True
     da_a = da_a.where(valid_values)
     da_b = da_b.where(valid_values)
 
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -1134,11 +1134,6 @@ def cov(da_a, da_b, dim=None, ddof=1):
     da_b = da_b.where(valid_values)
     valid_count = valid_values.sum(dim) - ddof
 
-    # if dim is not None:
-    #     valid_count = da_a[dim].size - ddof
-    # else:
-    #     valid_count = da_a.size
-
     # 3. Compute mean and standard deviation along the given dim
     demeaned_da_a = da_a - da_a.mean(dim=dim)
     demeaned_da_b = da_b - da_b.mean(dim=dim)
--- a/xarray/core/computation.py
+++ b/xarray/core/computation.py
@@ -1094,38 +1094,38 @@ def cov(da_a, da_b, dim=None, ddof=1):
 
     Examples
     --------
-    >>> da_a = DataArray(np.random.random((3, 5)),
+    >>> da_a = DataArray(np.array([[1, 2, 3], [0.1, 0.2, 0.3], [3.2, 0.6, 1.8]]),
     ...                  dims=("space", "time"),
     ...                  coords=[('space', ['IA', 'IL', 'IN']),
-    ...                          ('time', pd.date_range("2000-01-01", freq="1D", periods=5))])
+    ...                          ('time', pd.date_range("2000-01-01", freq="1D", periods=3))])
     >>> da_a
-    <xarray.DataArray (space: 3, time: 5)>
-    array([[0.04356841, 0.11479286, 0.70359101, 0.59072561, 0.16601438],
-            [0.81552383, 0.72304926, 0.77644406, 0.05788198, 0.74065536],
-            [0.96252519, 0.36877741, 0.22248412, 0.55185954, 0.23547536]])
+    <xarray.DataArray (space: 3, time: 3)>
+    array([[1. , 2. , 3. ],
+           [0.1, 0.2, 0.3],
+           [3.2, 0.6, 1.8]])
     Coordinates:
-    * space    (space) <U2 'IA' 'IL' 'IN'
-    * time     (time) datetime64[ns] 2000-01-01 2000-01-02 ... 2000-01-05
-    >>> da_b = DataArray(np.random.random((3, 5)),
+      * space    (space) <U2 'IA' 'IL' 'IN'
+      * time     (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03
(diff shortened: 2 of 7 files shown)


AndrewWilliams3142:
The current problem is that we can't use Pandas to test `xr.cov()` or `xr.corr()` because once you convert the `DataArrays` to a `series` or a `dataframe` for testing, you can't easily index them with a `dim` parameter. See @r-beer 's comment here https://github.com/pydata/xarray/pull/3550#issuecomment-557895005.

As such, I think it maybe just makes sense to test a few low-dimensional cases? Eg

```python3
>>> da_a = xr.DataArray(
          np.random.random((3, 21, 4)),
          coords={"time": pd.date_range("2000-01-01", freq="1D", periods=21)},
          dims=("a", "time", "x"),
      )

>>> da_b = xr.DataArray(
          np.random.random((3, 21, 4)),
          coords={"time": pd.date_range("2000-01-01", freq="1D", periods=21)},
          dims=("a", "time", "x"),
      )

>>> xr.cov(da_a, da_b, 'time')
<xarray.DataArray (a: 3, x: 4)>
array([[-0.01824046,  0.00373796, -0.00601642, -0.00108818],
       [ 0.00686132, -0.02680119, -0.00639433, -0.00868691],
       [-0.00889806,  0.02622817, -0.01022208, -0.00101257]])
Dimensions without coordinates: a, x
>>> xr.cov(da_a, da_b, 'time').sel(a=0,x=0)
<xarray.DataArray ()>
array(-0.01824046)
>>> da_a.sel(a=0,x=0).to_series().cov(da_b.sel(a=0,x=0).to_series())
-0.018240458880158048
```

So, while it's easy to check that a few individual points from `xr.cov()` agree with the pandas implementation, it would require a loop over `(a,x)` in order to check all of the points for this example. Do people have thoughts about this? 

I think it would also make sense to have some test cases where we don't use Pandas at all, but we specify the output manually?

```python3
>>> da_a = xr.DataArray([[1, 2], [1, np.nan]], dims=["x", "time"])
>>> expected = [1, np.nan]
>>> actual = xr.corr(da_a, da_a, dim='time')
>>> assert_allclose(actual, expected)
```

Does this seem like a good way forward?

keewis:
If you want to test individual values without reimplementing the function in the tests (which is what I fear comparing with the result of `np.cov` would require), that might be the only way.

If not, you could also check properties of covariance / correlation matrices, e.g. that `assert_allclose(xr.cov(a, b) / (a.std() * b.std()), xr.corr(a, b))` or that the diagonal of the auto-covariance matrix is the same as the variance of the array (with a 1D vector, not sure about more dimensions).
If you decide to test using properties, you could also extend our small collection of tests using `hypothesis` (see #1846).

AndrewWilliams3142:
One problem I came across here is that `pandas` automatically ignores 'np.nan' values in any `corr` or `cov` calculation. This is hard-coded into the package and there's no `skipna=False` option sadly, so what I've done in the tests is to use the `numpy` implementation which pandas is built on (see, for example [here](https://github.com/pandas-dev/pandas/blob/cb35d8a938c9222d903482d2f66c62fece5a7aae/pandas/core/nanops.py#L1325)). 

Current tests implemented are:
 - [x] `assert_allclose(xr.cov(a, b) / (a.std() * b.std()), xr.corr(a, b))`
 - [x] `assert_allclose(xr.cov(a,a)*(N-1), ((a - a.mean())**2).sum())`
 - [x] For the example in my previous comment, I now have a loop over all values of `(a,x)` to reconstruct the covariance / correlation matrix, and check it with an `assert_allclose(...)`.
    - [x] Add more test arrays, with/without `np.nans` -- **done**

mathause:
Currently `corr` needs to sanitize the inputs twice, which will be inefficient. One way around this is to define an internal method which can do both, depending on a `method` keyword (no need to write extra tests for this IMHO): 

```python

def corr(da_a, da_b, dim=None, ddof=0):

        return _cov_corr(da_a, da_b, dim=None, ddof=0, method="corr")

def cov(da_a, da_b, dim=None, ddof=0):

    return _cov_corr(da_a, da_b, dim=None, ddof=0, method="cov")


def _cov_corr(da_a, da_b, dim=None, ddof=0, method=None):

    # compute cov

    if method = "cov":
        return cov

    # compute corr

    return corr
```

Maybe you could use `xr.apply_ufunc` instead of looping in the tests (might be overkill).

doc/whats-new.rst
```diff
@@ -36,6 +36,8 @@ Breaking changes
 
 New Features
 ~~~~~~~~~~~~
+- Added :py:func:`xarray.cov` and :py:func:`xarray.corr` (:issue:`3784`, :pull:``).
```
keewis, line 39: The docs fail because of the mission PR number. So:
```suggestion
- Added :py:func:`xarray.cov` and :py:func:`xarray.corr` (:issue:`3784`, :pull:`3550`, :pull:`4089`).
```

keewis:
no worries about `hypothesis`, that's something we can add in a new PR.

Also, I don't think there is a `hypothesis.extra.xarray` module, yet. Any comments on that, @Zac-HD?

Zac-HD:
> Also, I don't think there is a hypothesis.extra.xarray module, yet. Any comments on that, @Zac-HD?

Not yet, though I'm interested in collaborations to write one!

Our existing Numpy + Pandas integrations, along with the `st.builds()` and `@st.composite` strategies, make it reasonably straightforward-if-tedious to compose the objects you want in exquisite detail... the real challenge is to design a clean high-level API which lets you control only the parts you care about.  Plausibly we have to let go of that last part to make this work, and just point people at the low-level tools if high-level isn't working.

[This third-party module](https://github.com/uber/hypothesis-gufunc/blob/master/hypothesis_gufunc/extra/xr.py) is unidiomatic in that it doesn't have the API design above, but I believe it works.  rdturnermtl has a history of great features; we eventually got idiomatic high-performance gufunc shapes upstream and I'm confident we'll get Xarray support eventually too... and sooner if there are people who can help design it :smile:

max-sixty:
This looks great! I think it's fine to have a simple implementation like this, even if it's not perfectly efficient. Well done for getting something working @AndrewWilliams3142 .

For the future: let's move away from using `np.random` in docstrings, since they're not reproducible, and will never pass a doctest.

Anything remaining before merging?

AndrewWilliams3142:
Cheers ! I've got a day off today so I'll do another pass through the changes and see if there's any low-hanging fruit I can improve (in addition to `np.random`, `_cov_corr` internal methods and maybe `apply_ufunc()`) :)

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    # 4. Compute covariance along the given dim
+    # N.B. `skipna=False` is required or there is a bug when computing
+    # auto-covariance. E.g. Try xr.cov(da,da) for
+    # da = xr.DataArray([[1, 2], [1, np.nan]], dims=["x", "time"])
+    cov = (demeaned_da_a * demeaned_da_b).sum(dim=dim, skipna=False) / (valid_count)
+
+    return DataArray(cov)
+
+
+def corr(da_a, da_b, dim=None, ddof=0):
```
mathause, line 1149:
```suggestion
def corr(da_a, da_b, dim=None):
```
`ddof` will cancel out for corr (but need to make sure to use the same `ddof` for `cov` and `std`)

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    """
+    Compute covariance between two DataArray objects along a shared dimension.
+    Parameters
+    ----------
+    da_a: DataArray (or Variable) object
+        Array to compute.
+    da_b: DataArray (or Variable) object
+        Array to compute.
+    dim : str, optional
+        The dimension along which the covariance will be computed
```
mathause, line 1082: Could you document `ddof`?

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    # 2. Ignore the nans
+    valid_values = da_a.notnull() & da_b.notnull()
+    da_a = da_a.where(valid_values)
+    da_b = da_b.where(valid_values)
+
+    # 3. Compute correlation based on standard deviations and cov()
+    da_a_std = da_a.std(dim=dim)
+    da_b_std = da_b.std(dim=dim)
+
+    corr = cov(da_a, da_b, dim=dim, ddof=ddof) / (da_a_std * da_b_std)
```
mathause, line 1223:
```suggestion
    corr = cov(da_a, da_b, dim=dim, ddof=0) / (da_a_std * da_b_std)
```
The default `ddof` for `std` in numpy is `0`.

mathause:
Could you also add a test for the `TypeError`?

```python
with raises_regex(TypeError, "Only xr.DataArray is supported"):
    xr.corr(xr.Dataset(), xr.Dataset())
```

AndrewWilliams3142:
> Could you also add a test for the `TypeError`?
> 
> ```python
> with raises_regex(TypeError, "Only xr.DataArray is supported"):
>     xr.corr(xr.Dataset(), xr.Dataset())
> ```

Where do you mean sorry? Isn't this already there in corr() and cov()?

```python3
if any(not isinstance(arr, (Variable, DataArray)) for arr in [da_a, da_b]):
        raise TypeError(
            "Only xr.DataArray and xr.Variable are supported."
            "Given {}.".format([type(arr) for arr in [da_a, da_b]])
        )
```

xarray/core/computation.py
AndrewWilliams3142: Done!

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    >>> xr.cov(da_a, da_b, dim='time')
+    <xarray.DataArray (space: 3)>
+    array([0.00207952, 0.01024296, 0.08214707])
+    Coordinates:
+    * space    (space) <U2 'IA' 'IL' 'IN'
+    """
+    from .dataarray import DataArray
+
+    # 1. Broadcast the two arrays
+    da_a, da_b = broadcast(da_a, da_b)
```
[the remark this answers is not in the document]
AndrewWilliams3142: Gotcha 👍

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    demeaned_da_a = da_a - da_a.mean(dim=dim)
+    demeaned_da_b = da_b - da_b.mean(dim=dim)
+
+    # 4. Compute covariance along the given dim
+    # N.B. `skipna=False` is required or there is a bug when computing
+    # auto-covariance. E.g. Try xr.cov(da,da) for
+    # da = xr.DataArray([[1, 2], [1, np.nan]], dims=["x", "time"])
+    cov = (demeaned_da_a * demeaned_da_b).sum(dim=dim, skipna=False) / (valid_count)
+
+    return DataArray(cov)
```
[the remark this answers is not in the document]
AndrewWilliams3142: Yep, you're right! I inherited this from the previous PR and didn't notice

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    Coordinates:
+    * space    (space) <U2 'IA' 'IL' 'IN'
+    """
+    from .dataarray import DataArray
+
+    if any(not isinstance(arr, (Variable, DataArray)) for arr in [da_a, da_b]):
+        raise TypeError(
+            "Only xr.DataArray and xr.Variable are supported."
+            "Given {}.".format([type(arr) for arr in [da_a, da_b]])
+        )
```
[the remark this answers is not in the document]
AndrewWilliams3142: Good point, I'll update the docstring as well

xarray/tests/test_computation.py
```diff
@@ -817,6 +818,174 @@ def test_vectorize_dask():
...
+            return np.cov(ts1.data.flatten(), ts2.data.flatten(), ddof=1)[0, 1]
+
+        expected = np_cov(da_a, da_b)
+        actual = xr.cov(da_a, da_b, dim)
+        assert_allclose(actual, expected)
+
+
+@pytest.mark.parametrize(
+    "da_a, da_b",
+    [arrays_w_tuples()[1][0], arrays_w_tuples()[1][1], arrays_w_tuples()[1][2]],
```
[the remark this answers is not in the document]
AndrewWilliams3142: I didn't use them all here because the middle two arrays just have dim `"time"` and in this test function it loops over `a` and `x` dims. I'm thinking of just removing these unused test arrays because the  capability of handling `nans` is well-covered by the other arrays?

xarray/tests/test_computation.py
```diff
@@ -817,6 +818,174 @@ def test_vectorize_dask():
...
+        (arrays[2], arrays[3]),
+        (arrays[3], arrays[3]),
+        (arrays[4], arrays[4]),
+        (arrays[4], arrays[5]),
+        (arrays[5], arrays[5]),
+    ]
+
+    return arrays, array_tuples
+
+
```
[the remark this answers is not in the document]
AndrewWilliams3142: Added

xarray/tests/test_computation.py
```diff
@@ -817,6 +818,174 @@ def test_vectorize_dask():
...
+        np.random.random((3, 21, 4)),
+        coords={"time": pd.date_range("2000-01-01", freq="1D", periods=21)},
+        dims=("a", "time", "x"),
+    )
+
+    arrays = [
+        da.isel(time=range(0, 18)),
+        da.isel(time=range(2, 20)).rolling(time=3, center=True).mean(),
+        xr.DataArray([0, np.nan, 1, 2, np.nan, 3, 4, 5, np.nan, 6, 7], dims="time"),
+        xr.DataArray([1, 1, np.nan, 2, np.nan, 3, 5, 4, 6, np.nan, 7], dims="time"),
```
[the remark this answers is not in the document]
AndrewWilliams3142: Removed these middle test arrays, now using all

xarray/core/computation.py
```diff
@@ -1069,6 +1069,162 @@ def earth_mover_distance(first_samples,
...
+    <xarray.DataArray ()>
+    array(0.03823054)
+    >>> xr.cov(da_a, da_b, dim='time')
+    <xarray.DataArray (space: 3)>
+    array([0.00207952, 0.01024296, 0.08214707])
+    Coordinates:
+    * space    (space) <U2 'IA' 'IL' 'IN'
+    """
+    from .dataarray import DataArray
+
```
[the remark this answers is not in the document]
AndrewWilliams3142: done

xarray/core/computation.py
```diff
@@ -1069,6 +1069,179 @@ def earth_mover_distance(first_samples,
...
+    dim : str, optional
+        The dimension along which the covariance will be computed
+    ddof: int, optional
+        If ddof=1, covariance is normalized by N-1, giving an unbiased estimate,
+        else normalization is by N.
+
+    Returns
+    -------
+    covariance: DataArray
+    See also
```
mathause, line 1091:
```suggestion

    See also
```

xarray/core/computation.py
```diff
@@ -1069,6 +1069,179 @@ def earth_mover_distance(first_samples,
...
+        Array to compute.
+    da_b: DataArray object
+        Array to compute.
+    dim: str, optional
+        The dimension along which the correlation will be computed
+
+    Returns
+    -------
+    correlation: DataArray
+    See also
```
mathause, line 1159:
```suggestion

    See also
```

xarray/core/computation.py
```diff
@@ -1069,6 +1069,179 @@ def earth_mover_distance(first_samples,
...
+    Internal method for xr.cov() and xr.corr() so only have to
+    sanitize the input arrays once and we don't repeat code.
+    """
+    # 1. Broadcast the two arrays
+    da_a, da_b = align(da_a, da_b, join="inner", copy=False)
+
+    # 2. Ignore the nans
+    valid_values = da_a.notnull() & da_b.notnull()
+    da_a = da_a.where(valid_values)
+    da_b = da_b.where(valid_values)
```
mathause, line 1221:
```suggestion
    if not valid_values.all():
        da_a = da_a.where(valid_values)
        da_b = da_b.where(valid_values)
```

xarray/core/computation.py
```diff
@@ -1069,6 +1069,179 @@ def earth_mover_distance(first_samples,
...
+    valid_count = valid_values.sum(dim) - ddof
+
+    # 3. Compute mean and standard deviation along the given dim
+    demeaned_da_a = da_a - da_a.mean(dim=dim)
+    demeaned_da_b = da_b - da_b.mean(dim=dim)
+    da_a_std = da_a.std(dim=dim)
+    da_b_std = da_b.std(dim=dim)
```
mathause, line 1228: Could you move these below the `else`?

AndrewWilliams3142: No ;)

AndrewWilliams3142: (...okay)

AndrewWilliams3142:
One more thing actually, is there an argument for not defining `da_a_std` and `demeaned_da_a` and just performing the operations in place? Defining these variables makes the code more readable but in https://github.com/pydata/xarray/pull/3550#discussion_r355157809 and https://github.com/pydata/xarray/pull/3550#discussion_r355157888 the reviewer suggests this is inefficient?

mathause:
If you insist ;)

```
da_a -= da_a.mean(dim=dim)
```
is indeed marginally faster. As they are already aligned, we don't have to worry about this.

AndrewWilliams3142:
> If you insist ;)
> 
> ```
> da_a -= da_a.mean(dim=dim)
> ```
> 
> is indeed marginally faster. As they are already aligned, we don't have to worry about this.

Sweet! On second thought, I might leave it for now...the sun is too nice today. Can always have it as a future PR or something. :)

max-sixty:
Awesome @AndrewWilliams3142 ! Very excited we have this.

Thanks for the review @mathause
 
Hitting merge; any other feedback is welcome and we can iterate.

kefirbandi:
Just a small comment: in the docs (http://xarray.pydata.org/en/latest/generated/xarray.cov.html#xarray.cov) there is a typo: da_a is declared twice, the second should really be da_b.

keewis:
thanks. Do you want to put in a PR fixing that?

kefirbandi:
Well, actually I was thinking, that correcting it for someone who is working on the code on a daily basis is ~30 seconds.  For me, I think, it would be quite a bit of overhead for a single character...

AndrewWilliams3142:
@kefirbandi I didn't want to step on your toes, but I'm happy to put in a PR to fix the typo. :)

kefirbandi:
@AndrewWilliams3142 I see. Thanks.

PR #4099, opened 2020-05-27
Allow non-unique and non-monotonic coordinates in get_clean_interp_index and polyfit

aulemahal:
- [ ] Closes #xxxx
 - [x] Tests added

 Pull #3733 added `da.polyfit` and `xr.polyval` and is using `xr.core.missing.get_clean_interp_index` in order to get the fitting coordinate. However, this method is stricter than what polyfit needs: as in `numpy.polyfit`, non-unique and non-monotonic indexes are acceptable. This PR adds a `strict` keyword argument to `get_clean_interp_index` so we can skip the uniqueness and monotony tests.

`ds.polyfit` and  `xr.polyval` were modified to use that keyword. I only added tests for  `get_clean_interp_index`, could add more for `polyfit` if requested.

### DIFF
--- a/xarray/tests/test_missing.py
+++ b/xarray/tests/test_missing.py
@@ -541,7 +541,9 @@ def test_get_clean_interp_index_strict(index):
     with pytest.raises(ValueError):
         get_clean_interp_index(da, "x")
 
-    get_clean_interp_index(da, "x", strict=False)
+    clean = get_clean_interp_index(da, "x", strict=False)
+    np.testing.assert_array_equal(index, clean)
+    assert clean.dtype == np.float64
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


xarray/tests/test_missing.py
```diff
@@ -534,6 +534,16 @@ def test_get_clean_interp_index_potential_overflow():
...
 
 
+@pytest.mark.parametrize("index", ([0, 2, 1], [0, 1, 1]))
+def test_get_clean_interp_index_strict(index):
+    da = xr.DataArray([0, 1, 2], dims=("x",), coords={"x": index})
+
+    with pytest.raises(ValueError):
+        get_clean_interp_index(da, "x")
+
+    get_clean_interp_index(da, "x", strict=False)
```
dcherian, line 544: Can we check the returned values against `index`?

Issue #2815, opened 2019-03-16
Cache datasets in XDG directories


QuLogic:
#### Problem description

The default directory for caching datasets is currently `~/.xarray_tutorial_data`:
https://github.com/pydata/xarray/blob/225868d232219440b188956531d5764ff4cd1b53/xarray/tutorial.py#L15

This doesn't follow the XDG standard on Linux, and AFAIK isn't actually invisible on Windows.

#### Expected Output

It would be better if this followed the XDG standard, i.e., `~/.cache/xarray_tutorial_data` (or some other places if you've set some environment variables) and something like `%LOCALAPPDATA%` on Windows. There are packages like [appdirs](https://pypi.org/project/appdirs/) that handle all this for you.

shoyer:
Contributions would be very welcome here.

I would be happy to vendor a copy of appdirs to handle this.

shoyer:
Or even just picking some better, platform specific defaults would be an improvement.

dcherian:
I think `pooch` handles this for us now.

PR #4102, opened 2020-05-27
cache rasterio example files

### LINKED ISSUE #3986: building the visualization gallery is slow
When running sphinx to build the documentation, it frequently times out when trying to build the visualization gallery. Running
```bash
/usr/bin/time -v python -c 'import xarray as xr; xr.open_rasterio("https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif")'
```
reports that it takes at least 5 minutes (or time out after 10 minutes) if opened from the url. Subsequent calls use the cache, so the second `rasterio` example is fast.

If instead I download the file manually and then load from disk, the whole notebook completes in about 10 seconds. Also, directly calling `rasterio.open` completes in a few seconds, so the bug should be in `open_rasterio`.

I do think we should try to fix this in the backend, but maybe we could also cache `RGB.byte.tif` in the same directory as the `xarray.tutorial` data and open the cached file in the gallery?

Edit: this is really flaky, I can't reliably reproduce this.

Edit2: for now, I'm using a extra cell containing
```python
import pathlib
import shutil
import requests

cache_dir = pathlib.Path.home() / ".xarray_tutorial_data"
path = cache_dir / "RGB.byte.tif"
url = "https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif"

if not path.exists() or path.stat().st_size == 0:
    with requests.get(url) as r, path.open(mode="wb") as f:
        if r.status_code == requests.codes.ok:
            shutil.copyfileobj(r.raw, f)
        else:
            print("download failed: {r.status_code}")
            r.raise_for_status()

url = path
```
and modify both examples to use the new url

keewis:
To make building the documentation a bit faster (by 5-10 minutes for me), this adds a `xr.tutorial.open_rasterio` function with a signature and behavior that is almost identical to `tutorial.open_dataset` (when we replace the `open_*` functions with `open_dataset(..., format="...")` we should do that for the tutorial functions, too).

It uses `requests.get` instead of [`urllib.request.urlretrieve`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve) so that would be a new dependency. I'm not sure if that's an issue since it's installed in the `bare-minimum` CI's environment.

The `tutorial.open_dataset` code could be rewritten to use the same structure but I wanted to get feedback on `open_rasterio` first.

Edit: I also changed the default cache directory to `~/.cache/xarray_tutorial_data` with a fallback to the old default if `~/.cache` does not exist

 - [x] Closes #3986
 - [x] Tests added (but that's just a comparison between cached and uncached)


jthielen:
For these kind of cached files, would it be worth implementing something like [Pooch](https://www.fatiando.org/pooch/latest/) for xarray?

keewis:
using a existing library for that instead of creating our own caching / downloading code does seem useful (the code there seems to be much more sophisticated than anything in the `tutorial` module), but we'd be introducing a new dependency: not sure how much of an issue that would be.

We could also make it a optional dependency, but that would mean that the `tutorial` module is not always available.

xarray/tutorial.py
```diff
@@ -6,23 +6,78 @@
...
 """
 import hashlib
-import os as _os
-from urllib.request import urlretrieve
+import pathlib
+import shutil
+import tempfile
 
 import numpy as np
+import requests
```
dcherian, line 14: don't we need to add `requests` to a list of dependencies somewhere?

keewis: yes, that's what I was thinking, too. Should that be a mandatory or an optional dependency?

dcherian: i think it's nice to have tutorial functions that always work. So i guess a mandatory dependency.

cc @pydata/xarray

shoyer: +1 for a mandatory dependency, just so it gets installed by default. I think it's fine to add a hard dependency on `requests`. It's already distributed on nearly every platform, and if it's hard to install for some reason it's easy for a redistributer to simply drop `xarray.tutorial`.

keewis: done

shoyer:
> Edit: I also changed the default cache directory to `~/.cache/xarray_tutorial_data` with a fallback to the old default if `~/.cache` does not exist

I would suggest just creating the full directory tree, e.g., with `os.makedirs(path, exist_ok=True)`. I don't think we want to let this be inconsistent.

xarray/tutorial.py
```diff
@@ -6,23 +6,78 @@
...
+    return md5sum == checksum
+
+
+def download_to(url, path):
+    with requests.get(url, stream=True) as r, path.open("wb") as f:
+        if r.status_code != 200:
+            raise OSError(f"download failed: {r.reason}")
+
+        r.raw.decode_content = True
+        shutil.copyfileobj(r.raw, f)
```
shoyer, line 44: could you kindly add comment to clarify the origin of this snippet? e.g., is it based on this https://stackoverflow.com/a/39217788/809705?

xarray/tutorial.py
```diff
@@ -6,23 +6,78 @@
...
+def check_md5sum(content, checksum):
+    md5 = hashlib.md5()
+    md5.update(content)
+    md5sum = md5.hexdigest()
+
+    return md5sum == checksum
+
+
+def download_to(url, path):
+    with requests.get(url, stream=True) as r, path.open("wb") as f:
```
shoyer, line 39: nit: I usually find it a little easier to read when you nest context managers:
```suggestion
    with requests.get(url, stream=True) as r
        with path.open("wb") as f:
            ...
```

xarray/tutorial.py
```diff
@@ -6,23 +6,78 @@
...
+    if not path.suffix:
+        path = path.with_suffix(default_extension)
+    elif path.suffix == ".byte":
+        path = path.with_name(name + default_extension)
+
+    if cache and path.is_file():
+        return _open_rasterio(path, **kws)
+
+    url = f"{github_url}/raw/{branch}/tests/data/{path.name}"
+    # make sure the directory is deleted afterwards
```
shoyer, line 77: I don't think this is currently done? should this comment be removed then?

xarray/tutorial.py
```diff
@@ -6,23 +6,78 @@
...
+    with cache_dir:
+        download_to(url, path)
```
shoyer, line 79: The reason why the MD5 hash is checked below is to guard against incomplete / failed downloads. This is actually a pretty common scenario, e.g., if a user hits Ctrl+C because downloading is taking too long (e.g., if they have a poor internet connection). So I think this would be important to include for rasterio as well.

That said, I'm not sure we really need to use MD5 verification, which is rather annoying to setup and requires downloading extra files. Instead, I would suggest ensuring that `download_to` always deletes the downloaded file if an exception is raised, e.g., based on https://stackoverflow.com/a/27045091/809705:
```python
def download_to(url, path):
    try:
        ...  # current contents of `download_to`
    except Exception:
        # switch to path.unlink(missing_ok=True) for Python 3.8+
        with contextlib.suppress(FileNotFoundError):
            path.unlink()
        raise
```

This would probably work in about 90% of cases.

There may be a few others where a computer is shut down suddenly in the middle to writing a file. To handle this, it's probably a good idea to try to do an _atomic write_:
https://stackoverflow.com/a/2333979/809705

keewis:
> I would suggest just creating the full directory tree, e.g., with `os.makedirs(path, exist_ok=True)`. I don't think we want to let this be inconsistent.

When writing this I wanted to make sure we didn't force a `~/.cache` folder on everyone, regardless of the platform.

After looking into this a bit more, it seems we would probably need to have different cache folders for different platforms:
- on windows, the path is given by `APPDATA` (or `LOCALAPPDATA`, not sure) *(I can't find a official resource for this)*
- on [macos](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFileSystem/Articles/LibraryDirectory.html#//apple_ref/doc/uid/20002282-BAJHCHJI) it can be either `~/Library/Caches` (or `~/Library/Application Support`?)
- on unix (following the [freedesktop basedir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)) it is indicated by `XDG_CACHE_HOME` with a default of `~/.cache`.

So a currently released version of `xarray` will always put it into `~/.xarray_tutorial_data`, and my changes will partially align with the freedesktop spec.

How closely should we follow those specifications?

xarray/tutorial.py
keewis: I implemented the "atomic" write with a deterministic temporary name, which (as long as we properly detect something like disconnected sockets / broken pipes) should guard against incomplete writes and also avoid stale files. The only disadvantage I know of is that concurrent calls will interfere with each other.

dcherian:
> Edit: appdirs solves that (but: it's a new dependency)

See #2815 :)

keewis:
thanks for pointing me to that issue, @dcherian

shoyer:
> > Edit: appdirs solves that (but: it's a new dependency)
> 
> See #2815 :)

As mentioned in that issue, we could definitely vendor a copy of appdirs if desired. Or we could pick some simple heuristic. In practice I think `~/.cache` is fine. It's only standard on Linux, but I have a handful of directories there on my Mac laptop already.

xarray/tutorial.py
shoyer: My main concern here would that if two invocations of `tutorial.open_dataset` start at the same time (e.g., in different processes). Then it's possible that both could partially write to the same temporary file, which could get moved into place with an inconsistent state.

I think it's probably a better idea to make the temporary file with a random name (like the [atomicwrites](https://github.com/untitaker/python-atomicwrites) package). In the worst case you end up with an extra file, but that's better than an bad state that requires manual intervention to resolve.

Another option is to vendor a copy of [atomicwrites](https://github.com/untitaker/python-atomicwrites) (e.g., as well as appdirs). It's a single file project so this would be relatively straightforward.

keewis: sure. I changed it to use `NamedTemporaryFile` to generate a unique file within the same directory and rename once it's done downloading. If the downloading succeeded, the rename will always work (not sure if that's the case on windows, though)

keewis:
I vendored `appdirs.user_cache_dir`, but I need help with `mypy`: the windows code uses `ctypes.windll` which is only available on windows

shoyer:
> I vendored `appdirs.user_cache_dir`, but I need help with `mypy`: the windows code uses `ctypes.windll` which is only available on windows

Let's skip this module for mypy. You can add it to the list in `setup.cfg`.

xarray/tutorial.py
```diff
@@ -19,13 +21,175 @@
 from .core.dataarray import DataArray
 from .core.dataset import Dataset
 
-_cache_name = "xarray_tutorial_data"
-_cache_dir = pathlib.Path.home() / ".cache"
-# TODO: I/O on import. Might not be a good idea.
-if _cache_dir.is_dir():
-    _default_cache_dir = _cache_dir / _cache_name
+# vendored from appdirs
```
shoyer, line 24: We still need to keep the appdirs copyright notice. The simplest way to do this is to keep appdirs in a separate, self-contained file, e.g., like what's done for Numba: https://github.com/numba/numba/blob/13ece9b97e6f01f750e870347f231282325f60c3/numba/misc/appdirs.py

xarray/tutorial.py
```diff
@@ -6,23 +6,267 @@
...
+
+    try:
+        f = tempfile.NamedTemporaryFile(dir=folder, prefix=prefix, delete=False)
+        temporary_path = pathlib.Path(f.name)
+
+        yield f
+
+        temporary_path.rename(path)
+    except Exception:
+        temporary_path.unlink()
```
shoyer, line 217: can we add this bit to ignore non-existing files?
```python
        # switch to path.unlink(missing_ok=True) for Python 3.8+
        with contextlib.suppress(FileNotFoundError):
            path.unlink()
        raise
```

xarray/tutorial.py
```diff
@@ -6,23 +6,267 @@
...
+        raise
+    finally:
+        pass
+
```
shoyer, line 221: There's no need to include an empty `finally` clause :)

keewis: that was from before I added the clause to delete the temporary file if an exception was raised. I'll delete it.

xarray/tutorial.py
keewis: sure

xarray/tutorial.py
keewis: done, although the directory could have a better name

xarray/tutorial.py
```diff
@@ -6,23 +6,129 @@
...
+    elif path.suffix == ".byte":
+        path = path.with_name(name + default_extension)
+
+    if cache and path.is_file():
+        return _open_rasterio(path, **kws)
+
+    url = f"{github_url}/raw/{branch}/tests/data/{path.name}"
+    # if cache_dir points to a temporary directory, make sure it is
+    # deleted afterwards
+    with cache_dir:
```
shoyer, line 129: I feel like I'm missing something -- what does using a pathlib.Path object as a context manager do?

I don't think it automatically deletes the files after using it?

xarray/tutorial.py
```diff
@@ -59,46 +165,52 @@ def open_dataset(
...
+        # verify the checksum (md5 guards only against transport corruption)
+        md5_path = path.with_name(path.name + ".md5")
+        download_to(construct_url(md5_path.name), md5_path)
+        if not check_md5sum(path.read_bytes(), md5_path.read_text()):
+            path.unlink()
+            md5_path.unlink()
             msg = """
             MD5 checksum does not match, try downloading dataset again.
             """
             raise OSError(msg)
```
shoyer, line 211: if the new machinery works, I think we could probably safely remove the MD5 checks?

xarray/tutorial.py
```diff
@@ -6,23 +6,129 @@
...
+        raise
+
+
+def download_to(url, path):
+    # based on https://stackoverflow.com/a/39217788
+    with open_atomic(path, mode="wb") as f:
+        with requests.get(url, stream=True) as r:
+            if r.status_code != 200:
+                raise OSError(f"download failed: {r.reason}")
+
```
shoyer, line 60: would this suffice instead?
```suggestion
           r.raise_for_status()
```

xarray/tutorial.py
keewis: no, of course not. `cache_dir` can be either a `pathlib.Path` instance or a `NamedTemporaryDirectory` instance. Using a `Path` object as a context manager will "close" it so you can't use it for I/O anymore (`unlink`, `read_text`, etc. won't work anymore) while a `NamedTemporaryDirectory` instance will delete the temporary directory it created when `__exit__` is called.

Not sure if there's a way to make that easier to read.

xarray/tutorial.py
keewis: I don't think so? The [docs](https://requests.readthedocs.io/en/master/user/quickstart/#response-status-codes) say that would only raise for `4xx` or `5xx`, so this is not strictly the same. We could replace the `200` with `requests.codes.ok`.

keewis:
what should we do with this? Maybe it's easier to decide if we split this PR into one that introduces the caching and a new one that tries to find the correct "appdirs" path (which I think is nice to have but not really important)?

keewis:
I switched the tutorial code to use `pooch` (the tests still need to be updated), which means that we get the appdirs, download, and caching for free. It is a optional dependency for now, thus requiring `pooch` to be able to use `xr.tutorial.open_*`.

However, `pooch` seems to be mainly written for repositories that contain both the data and the code (like `scikit-image`'s `data` module). If I understand it correctly, this means that we can't take full advantage of the features `pooch` has (such as the file registry, or the `version` mechanism). Some of the options we have (`github_url` and `branch`) make this even more difficult. I wonder if we should deprecate / remove them?

So for now, instead of using `pooch.create(base_url=url, registry=registry).fetch(name)` we use `pooch.retrieve(url, hash=None, ...)`. Not sure if there's a better way, though.

dcherian:
Ping @andersy005 who has experience with pooch

shoyer:
I don't think it's important to preserve the GitHub URL and branch options. I suspect those options exist strictly for testing new datasets.

I'm surprised that Pooch seems to be focused on data that is bundled in code repositories. It seems like that could easily lead to bloated repositories (though I guess xarray data is bigger than image data, for example).

shoyer:
I suspect we could make Pooch work OK by tagging releases in the xarray-data repository. Then we can manually update the version number inside xarray whenever we add or update a dataset.

keewis (commented):
I decided to combine `tutorial.open_dataset` and `tutorial.open_rasterio` because that results in less duplicated code, but right now `open_dataset` does not support `engine="rasterio"` so this might be confusing.

shoyer (approved):
LGTM. I don't think we have a strong need for data versioning here, as long as we don't remove existing datasets.

The main use-case for hashing was verifying that a file downloaded successfully even if the network connection was interrupted, but pooch writes to a temp file and then moves it into place, which is probably a sufficiently robust solution for that.

keewis:
data versioning might be useful if we ever decide to change datasets: once we do, the examples would change as well, so people might not be able to exactly reproduce them. Not sure how much of an issue that is, though.

I fixed the failing tests, but the trick I used (point `$XDG_CACHE_DIR` to a temporary directory) only works on linux.

andersy005:
>  pooch writes to a temp file and then moves it into place, which is probably a sufficiently robust solution for that.

I concur. 👍🏽

andersy005 (approved):

andersy005:
Thank you for this refactor, @keewis!

PR #4108, opened 2020-05-28
Convert Variable to IndexVariable or vice versa when renamed

### LINKED ISSUE #4417: Inconsistency in whether index is created with new dimension coordinate?
It seems like `set_coords()` doesn't create an index variable. Is there a reason for this? I was surprised that the following code snippets produce different Datasets (first one has empty `indexes`, second one has `x` in `indexes`), even though both Datasets have a 'dimension coordinate' `x`:

(1)
```
import numpy as np
import xarray as xr

ds = xr.Dataset()
ds['a'] = ('x', np.linspace(0,1))
ds['b'] = ('x', np.linspace(3,4))
ds = ds.rename(b='x')
ds = ds.set_coords('x')

print(ds)
print('indexes', ds.indexes)
```

(2)
```
import numpy as np
import xarray as xr

ds = xr.Dataset()
ds['a'] = ('x', np.linspace(0,1))
ds['x'] = ('x', np.linspace(3,4))

print(ds)
print('indexes', ds.indexes)
```

**Environment**:

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.6 | packaged by conda-forge | (default, Jun  1 2020, 18:57:50) 
[GCC 7.5.0]
python-bits: 64
OS: Linux
OS-release: 5.4.0-47-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.7.4

xarray: 0.16.0
pandas: 1.1.1
numpy: 1.18.5
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: None
h5py: 2.10.0
Nio: None
zarr: None
cftime: 1.2.1
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.23.0
distributed: 2.25.0
matplotlib: 3.2.2
cartopy: None
seaborn: None
numbagg: None
pint: 0.13
setuptools: 49.6.0.post20200814
pip: 20.2.3
conda: 4.8.4
pytest: 5.4.3
IPython: 7.15.0
sphinx: 3.2.1

</details>

dcherian:
- [x] Closes #4107, closes #4417
 - [x] Tests added


xarray/tests/test_dataset.py
```diff
@@ -2558,6 +2558,21 @@ def test_rename_vars(self):
...
+        ds["a"] = xr.DataArray([1, 0], [coord_1])
+        obj = ds.reset_index("coord_1").rename({"coord_1_": "coord_1"})
+        assert_equal(ds, obj)
+
+        newds = (
+            ds.reset_index("coord_1")
+            .reset_coords("coord_1_")
+            .rename({"coord_1_": "coord_1"})
+        )
+        assert "coord_1" not in newds.data_vars
```
dcherian, line 2574: Apparently we can create a dataset where a data variable and dimension share the same name. Should this be allowed? Should we raise an error instead of automatically promote `coord_1` to a coordinate variable?

```
<xarray.Dataset>
Dimensions:  (coord_1: 2)
Dimensions without coordinates: coord_1
Data variables:
    a        (coord_1) int64 1 0
    coord_1  (coord_1) int64 1 2
```

xarray/core/dataset.py
```diff
@@ -2782,6 +2782,17 @@ def _rename_all(self, name_dict, dims_dict):
         variables, coord_names = self._rename_vars(name_dict, dims_dict)
         dims = self._rename_dims(dims_dict)
         indexes = self._rename_indexes(name_dict, dims.keys())
+
+        # Variable could be renamed to an existing dimension name
+        # in this case, convert to IndexVariable and set indexes
+        # GH4107
+        for name in coord_names:
```
dcherian, line 2789: This condition tests whether the new name is an existing coord name and then promotes to IndexVariable. See next comment for a case not caught by this condition.

xarray/tests/test_dataset.py
dcherian: now promotes to coords.

keewis:
I wonder if it would be better to first reorganize / document all of the existing functions: we currently have `rename` (and `Dataset.rename_dims` / `Dataset.rename_vars`), `set_coords`, `reset_coords`, `set_index`, `reset_index` and `swap_dims`, which partially overlap. For example, the code sample from #4417 works if instead of
```python
ds = ds.rename(b='x')
ds = ds.set_coords('x')
```
we use
```python
ds = ds.set_index(x="b")
```
and something similar for the code sample in #4107.

I believe we currently have these use cases (not sure if that list is complete, though):
- rename a `DataArray` → `rename`
- rename a existing variable to a name that is not yet in the object → `rename` / `Dataset.rename_vars` / `Dataset.rename_dirs`
- convert a data variable to a coordinate (not a dimension coordinate) → `set_coords`
- convert a coordinate (not a dimension coordinate) to a data variable → `reset_coords`
- swap a existing dimension coordinate with a coordinate (which may not exist) and rename the dimension → `swap_dims`
- use a existing coordinate / data variable as a dimension coordinate (do not rename the dimension) → `set_index`
- stop using a coordinate as dimension coordinate and append `_` to its name (do not rename the dimension) → `reset_index`
- use two existing coordinates / data variables as a MultiIndex → `set_index`
- stop using a MultiIndex as a dimension coordinate and use its levels as coordinates → `reset_index`

Sometimes, some of these can be emulated by combinations of others, for example:
```python
# x is a dimension without coordinates
assert_identical(ds.set_index({"x": "b"}), ds.swap_dims({"x": "b"}).rename({"b": "x"}))
assert_identical(ds.swap_dims({"x": "b"}), ds.set_index({"x": "b"}).rename({"x": "b"}))
```
and, with this PR:
```python
assert_identical(ds.set_index({"x": "b"}), ds.set_coords("b").rename({"b": "x"}))
assert_identical(ds.swap_dims({"x": "b"}), ds.rename({"b": "x"}))
```
which means that it would increase the overlap of `rename`, `set_index`, and `swap_dims`.

In any case I think we should add a guide which explains which method to pick in which situation (or extend `howdoi`).

dcherian:
> I wonder if it would be better to first "reorganize" all of the existing functions

I fully agree. However the bug in #4107 is that an invalid xarray object is created (`_assert_internal_invariants` fails), so I think we should merge a solution even before cleaning up the API.

xarray/core/dataset.py
```diff
@@ -2995,7 +2995,7 @@ def _rename_dims(self, name_dict):
 
     def _rename_indexes(self, name_dict, dims_set):
         if self._indexes is None:
-            return None
+            return {}
```
dcherian, line 2998:
```suggestion
            return None
```

PR #4130, opened 2020-06-08
Fix: apply_ufunc with exclude_dims and vectorize

### LINKED ISSUE #3890: error for apply_ufunc with exclude_dims and vectorize
I tried to use `apply_ufunc` for a function that takes input of unequal length and requires `vectorize=True` which resulted in a `ValueError`. I think the problem stems from the way `np.vectorize` is called.

#### MCVE Code Sample

```python
import xarray as xr
import scipy as sp
import scipy.stats
import numpy as np

# create dataarrays of unequal length
ds = xr.tutorial.open_dataset("air_temperature")

da1 = ds.air
da2 = ds.air.isel(time=slice(None, 50))

# function that takes arguments of unequal length and requires vectorizing
def mannwhitneyu(x, y):
    _, p = sp.stats.mannwhitneyu(x, y)
    return p

# test that the function takes arguments of unequal length
mannwhitneyu(da1.isel(lat=0, lon=0), da2.isel(lat=0, lon=0))

xr.apply_ufunc(
    mannwhitneyu,
    da1,
    da2,
    input_core_dims=[["time"], ["time"]],
    exclude_dims=set(["time"]),
    vectorize=True,
)

```

Returns

```python
ValueError: inconsistent size for core dimension 'n': 50 vs 2920
```
Note: the error stems from numpy.

#### Expected Output
A DataArray.

#### Problem Description

I can reproduce the problem in pure numpy:

``` python
vec_wrong = np.vectorize(mannwhitneyu, signature="(n),(n)->()", otypes=[np.float])
vec_wrong(da1.values.T, da2.values.T)
```

The correct result is returned when the `signature` is changed:
``` python
vec_correct = np.vectorize(mannwhitneyu, signature="(m),(n)->()", otypes=[np.float])
vec_correct(da1.values.T, da2.values.T)
```
So I think the signature needs to be changed when `exclude_dims` are present.

#### Versions

<details><summary>Output of `xr.show_versions()`</summary>

This is my development environment, so i think xarray should be 'master'.

**PNC:/home/mathause/conda/envs/xarray_devel/lib/python3.7/site-packages/PseudoNetCDF/pncwarn.py:24:UserWarning:
  pyproj could not be found, so IO/API coordinates cannot be converted to lat/lon; to fix, install pyproj or basemap (e.g., `pip install pyproj)`

INSTALLED VERSIONS
------------------
commit: None
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 4.15.0-91-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.5
libnetcdf: 4.7.1

xarray: 0.11.1+335.gb0c336f6
pandas: 0.25.3
numpy: 1.17.3
scipy: 1.3.1
netCDF4: 1.5.3
pydap: installed
h5netcdf: 0.7.4
h5py: 2.10.0
Nio: None
zarr: 2.3.2
cftime: 1.0.4.2
nc_time_axis: None
PseudoNetCDF: installed
rasterio: 1.1.0
cfgrib: 0.9.5.4
iris: None
bottleneck: 1.2.1
dask: 2.6.0
distributed: 2.6.0
matplotlib: 3.1.2
cartopy: None
seaborn: 0.9.0
numbagg: None
setuptools: 41.6.0.post20191101
pip: 19.3.1
conda: installed
pytest: 5.2.2
IPython: 7.9.0
sphinx: None

</details>

mathause:
- [x] Closes #3890
 - [x] Tests added

This would be pretty useful when using `apply_ufunc` for e.g. statistical tests  that require `vectorize=True` and can consume data of unequal length (i.e. uses `exclude_dims`). This should also extend to dask arrays with `dask="parallelized"` once #4060 is implemented.


xarray/core/computation.py
```diff
@@ -992,7 +1019,7 @@ def earth_mover_distance(first_samples,
     if kwargs is None:
         kwargs = {}
 
-    signature = _UFuncSignature(input_core_dims, output_core_dims)
+    signature = _UFuncSignature(input_core_dims, output_core_dims, exclude_dims)
```
shoyer, line 1022: As far as I can tell, we only need `exclude_dims` inside the `to_gufunc_string` method. What about just adding it as an optional argument there?

mathause:
Sounds good. I can move `_enumerate` into `to_gufunc_string`. I think it makes sense to wait for #4060.

mathause:
Inplemented @shoyer's suggestion to add `exclude_dims` as optional argument to `to_gufunc_string`. The readthedocs error is a timeout so should be fine to review.

keewis:
> The readthedocs error is a timeout

Unfortunately not. The timeout happened because of an *error* and the traceback is visible. The timeout doesn't happen on my own setup, so I asked support to help fix that. They removed stale memory constraints so now the traceback is visible, but somehow the build still times out.

In this case `pandas` has a minimum version requirement for `xarray`, but the detected version is `0.1`. I'm not sure why that works for me but not for you, maybe because you didn't update the tags on your repo in a long time?

mathause:
Thanks keewis! I don't think I would have figured that out. I pushed the tags to my fork.

mathause:
But unfortunately this doesn't help :( If you have another idea I am all ears. But anyways it should not stop anyone from reviewing.

keewis:
on my own setup the build completes: https://readthedocs.org/projects/xarray-keewis/builds/11696307/ So this is definitely an issue with the setup, but I don't know how to fix that...

kmuehlbauer:
@keewis In the xarray setup there is `--prune-tags`. In your setup this is not there. Not sure, if this might be connected.

keewis:
thanks, @kmuehlbauer, good spot. However, that seems to come from the PR build (see also https://readthedocs.org/projects/xray/builds/11695402/, which doesn't fail), so I guess there has to be something in the code that changes the version.

dcherian (approved):
Very clean. Great work!

keewis:
the merge seems to have fixed it. Not sure why it didn't work before.

mathause:
Thanks @keewis!

PR #4144, opened 2020-06-10
Improve typehints of xr.Dataset.__getitem__

### LINKED ISSUE #4125: Improving typing of `xr.Dataset.__getitem__`
First, I'd like the thank the xarray dev's for adding type hints to this library, not many libraries have this feature!

That said, the indexing notation of `xr.Dataset` does not currently play well wit mypy since it returns a Union type. This results in a lot of mypy errors like this: 
```
workflows/fine_res_budget/budget/budgets.py:284: error: Argument 6 to "compute_recoarsened_budget_field" has incompatible type "Union[DataArray, Dataset]"; expected "DataArray"
workflows/fine_res_budget/budget/budgets.py:285: error: Argument 1 to "storage" has incompatible type "Union[DataArray, Dataset]"; expected "DataArray"
workflows/fine_res_budget/budget/budgets.py:286: error: Argument "unresolved_flux" to "compute_recoarsened_budget_field" has incompatible type "Union[DataArray, Dataset]"; expected "DataArray"
workflows/fine_res_budget/budget/budgets.py:287: error: Argument "saturation_adjustment" to "compute_recoarsened_budget_field" has incompatible type "Union[DataArray, Dataset]"; expected "DataArray"
```

#### MCVE Code Sample

```
def func(ds: xr.Dataset):
    pass

dataset: xr.Dataset = ...

# error:
# this line will give type error because mypy doesn't know 
# if ds[['a', 'b]] is Dataset or a DataArray
func(ds[['a', 'b']])
```

#### Expected Output

Mypy should be able to infer that `ds[['a', b']]` is a Dataset, and that `ds['a']` is a DataArray.

#### Problem Description

This requires any routine with type hints that consume an output of `xr.Dataset.__getitem__` to require a `Union[DataArray, Dataset]` even if it really intends to  be used with either `DataArray` or `DataArray`. Because `ds[something]` is a ubiquitous syntax,  this behavior accounts for approximately 50% of mypy errors in my xarray heavy code.

#### Versions

In [1]: import xarray as xr                                                                                                                                                                          
xr.
In [2]: xr.show_versions()                                                                                                                                                                           

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.7 (default, May  7 2020, 21:25:33) 
[GCC 7.3.0]
python-bits: 64
OS: Linux
OS-release: 5.3.0-1020-gcp
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: C.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.7.3

xarray: 0.15.1
pandas: 1.0.1
numpy: 1.18.1
scipy: 1.4.1
netCDF4: 1.5.3
pydap: None
h5netcdf: 0.8.0
h5py: 2.10.0
Nio: None
zarr: 2.4.0
cftime: 1.1.2
nc_time_axis: 1.2.0
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.17.2
distributed: 2.17.0
matplotlib: 3.1.3
cartopy: 0.17.0
seaborn: 0.10.1
numbagg: None
setuptools: 46.4.0.post20200518
pip: 20.0.2
conda: 4.8.3
pytest: 5.4.2
IPython: 7.13.0
sphinx: None
</details>

# Potential solution

I think we can fix this with [typing.overload](https://docs.python.org/3/library/typing.html#typing.overload). I am not too familiar with that librariy, but I think something like the following might work:

```
from typing import overload

class Dataset
    @overload
    def __getitem__(self, key: Hashable) -> DataArray: ...
    
     @overload
    def __getitem__(self, key: List[Hashable]) -> "Dataset": ...

     # actual implementation
    def __getitem__
```

nbren12:
To resolve some common type-related errors, this PR adds some overload type hints to `Dataset.__getitem__`. Now mypy can correctly infer that hashable inputs return DataArrays.

 - [x] Closes #4125

### DIFF
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1243,7 +1243,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         return _LocIndexer(self)
 
     @overload  # noqa: F811
-    def __getitem__(self, key: Hashable) -> DataArray:
+    def __getitem__(self, key: Hashable) -> "DataArray":
         ...
 
     @overload  # noqa: F811
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1242,15 +1242,15 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         """
         return _LocIndexer(self)
 
-    @overload
+    @overload  # noqa: F811
     def __getitem__(self, key: Hashable) -> DataArray:
         ...
 
-    @overload
+    @overload  # noqa: F811
     def __getitem__(self, key: Any) -> "Union[DataArray, Dataset]":
         ...
 
-    def __getitem__(self, key):
+    def __getitem__(self, key):  # noqa: F811
         """Access variables or coordinates this dataset as a
         :py:class:`~xarray.DataArray`.
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1241,13 +1241,18 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         and only when the key is a dict of the form {dim: labels}.
         """
         return _LocIndexer(self)
-    
 
     @overload
-    def __getitem__(self, key: Hashable) -> DataArray: ...
+    def __getitem__(self, key: Hashable) -> DataArray:
+        ...
 
     @overload
-    def __getitem__(self, key: Any) -> Dataset: ...
+    def __getitem__(self, key: Mapping) -> "Dataset":
+        ...
+
+    @overload
+    def __getitem__(self, key: List) -> "Dataset":
+        ...
 
     def __getitem__(self, key):
         """Access variables or coordinates this dataset as a
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1242,21 +1242,25 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         """
         return _LocIndexer(self)
 
-    @overload  # noqa: F811
-    def __getitem__(self, key: Hashable) -> "DataArray":
+    # FIXME https://github.com/python/mypy/issues/7328
+    @overload
+    def __getitem__(self, key: Mapping) -> "Dataset":  # type: ignore
         ...
 
-    @overload  # noqa: F811
-    def __getitem__(self, key: Any) -> "Union[DataArray, Dataset]":
+    @overload
+    def __getitem__(self, key: Hashable) -> "DataArray":  # type: ignore
         ...
 
-    def __getitem__(self, key):  # noqa: F811
+    @overload
+    def __getitem__(self, key: Any) -> "Dataset":
+        ...
+
+    def __getitem__(self, key):
         """Access variables or coordinates this dataset as a
         :py:class:`~xarray.DataArray`.
 
         Indexing with a list of names will return a new ``Dataset`` object.
         """
-        # TODO(shoyer): type this properly: https://github.com/python/mypy/issues/7328
         if utils.is_dict_like(key):
             return self.isel(**cast(Mapping, key))
 
--- a/xarray/core/dataset.py
+++ b/xarray/core/dataset.py
@@ -1247,11 +1247,7 @@ class Dataset(Mapping, ImplementsDatasetReduce, DataWithCoords):
         ...
 
     @overload
-    def __getitem__(self, key: Mapping) -> "Dataset":
-        ...
-
-    @overload
-    def __getitem__(self, key: List) -> "Dataset":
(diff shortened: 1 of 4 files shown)


mathause:
The mypy check throws an error: `xarray/core/dataset.py:1250: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same` I think you can ignore the other failures. Should that be:

```python
    @overload
    def __getitem__(self, key: Hashable) -> DataArray:
        ...

    @overload
    def __getitem__(self, key: Iterable[Hashable]) -> "Dataset":
        ...

```
? Only guessing, though - it was `Any` so there may be more options. @crusaderky

nbren12:
@mathause On further consideration, I think it might not be possible to get this to work. This method has three behaviors:
- Mapping -> Dataset
- Hashable -> DataArray
- else (List): -> Dataset

With my limited understanding of `mypy`, I think that any two of these is supported by `overload`, but I'm not sure it's possible to support all 3 with overload. I tried several different options.

nbren12:
Okay. Assuming the tests pass, I think this is ready for review. I tried adding a test, but mypy didn't seem to find problems even with code that I *know* doesn't work (e.g. `'a'+ 1`). Is there some strategy for testing tricky type hints like this? 

In any case, this code does work:
```
$ cat test_mypy.py                                                                                                                                                                                                    (fv3net) 
import xarray as xr
ds = xr.Dataset({"a": ()})


arr = ds['a']
union_obj = ds[['a']]

reveal_locals()
$ mypy test_mypy.py                                                                                                                                                                                                   (fv3net) 
test_mypy.py:8: note: Revealed local types are:
test_mypy.py:8: note:     arr: xarray.core.dataarray.DataArray
test_mypy.py:8: note:     ds: xarray.core.dataset.Dataset
test_mypy.py:8: note:     union_obj: Union[xarray.core.dataarray.DataArray, xarray.core.dataset.Dataset]
```

mathause:
Seems this was already discussed in [GH3210 (comment)](https://github.com/pydata/xarray/pull/3210/files/3cfd9a0c8c09de28e71abf8c861a8599c93bfd29#r317969251) and see also the TODO:

https://github.com/pydata/xarray/blob/8f688ea92ae8416ecc3e18f6e060dad16960e9ac/xarray/core/dataset.py#L1244-L1250

(although it is not entirely clear to me whether this is actually fixed or not)

max-sixty:
Nice find @mathause ; I remember that discussion now.

As @nbren12 said, this doesn't go all the way given mypy's restrictions, but seems like a dominant improvement. 

Test failures are unrelated.

Thanks @nbren12 !

nbren12:
No problem! I think I am done with this one unless you think its important that I document or test this somehow. Can someone review it?

crusaderky:
I took the liberty to rework it, please have a look
Test script:

```python
from typing import Hashable, Mapping
import xarray
ds: xarray.Dataset


class D(Hashable, Mapping):
    def __hash__(self): ...
    def __getitem__(self, item): ...
    def __iter__(self): ...
    def __len__(self): ...

reveal_type(ds["foo"])
reveal_type(ds[["foo", "bar"]])
reveal_type(ds[{}])
reveal_type(ds[D()])
```
mypy output:
```
t1.py:12: note: Revealed type is 'xarray.core.dataarray.DataArray'
t1.py:13: note: Revealed type is 'xarray.core.dataset.Dataset'
t1.py:14: note: Revealed type is 'xarray.core.dataset.Dataset'
t1.py:15: note: Revealed type is 'xarray.core.dataset.Dataset'
```

xarray/core/dataset.py
```diff
@@ -1241,13 +1242,25 @@ def loc(self) -> _LocIndexer:
         """
         return _LocIndexer(self)
 
-    def __getitem__(self, key: Any) -> "Union[DataArray, Dataset]":
+    # FIXME https://github.com/python/mypy/issues/7328
```
shoyer, line 1245: It looks like this is fixed now and can be removed? Or perhaps we more it below above the third `@overload` and add a comment that `Any` means `list`?

crusaderky: I don't think it's fixed. Specifically, mypy can't deal in the signature with an overload of Mapping and Hashable. Curiously however, once you add ``#type: ignore`` to the overloaded signature, the actual type inspection works just fine (see my test script above).

nbren12:
@crusaderky Thanks for the re-work. For my own benefit, could you explain why that code worked? I remember writing something very similar, and running into mypy errors. <y understanding of how mypy intreprets overload seems incomplete.

crusaderky:
@nbren12 it seems to me that mypy is being overly aggressive when parsing the hinted code (hence why I had to put ``# type: ignore`` on it) but it is being more lax when the same code is invoked somewhere else like in my test script. Overall I suspect it may be fragile and break in future mypy versions...

PR #4149, opened 2020-06-12
speed up map_blocks

### LINKED ISSUE #4147: readthedocs build / documentation build time
It seems that after the merge of #3818, the RTD builds started to time out while the [`docs` CI](https://dev.azure.com/xarray/xarray/_build/results?buildId=3028&view=logs&jobId=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36) take about 37 minutes.

As a reference, before #3818 our [`docs` CI](https://dev.azure.com/xarray/xarray/_build/results?buildId=3000&view=logs&jobId=7e620c85-24a8-5ffa-8b1f-642bc9b1fc36) completed in about 6 minutes.

I'm not sure if that is due to #3818 or because of updated dependencies (`dask`?), but I think we should try not to take these 30 minutes.

keewis:
I'm not sure if I broke something, but this is what I would suggest to use instead of putting xarray objects into an object array. The list of arguments is normally far less than 100 elements, so using python for that should not be a problem.

Edit: we're back to a more reasonable 6 minutes for the CI (the build itself takes about 3 minutes)

 - [x] might close #4147
 - [ ] Tests added

### DIFF
--- a/xarray/core/parallel.py
+++ b/xarray/core/parallel.py
@@ -34,11 +34,8 @@ from .dataset import Dataset
 T_DSorDA = TypeVar("T_DSorDA", DataArray, Dataset)
 
 
-def to_object_array(iterable):
-    # using empty_like calls compute
-    npargs = np.empty((len(iterable),), dtype=np.object)
-    npargs[:] = iterable
-    return npargs
+def unzip(iterable):
+    return zip(*iterable)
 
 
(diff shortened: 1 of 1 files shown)


dcherian:
LGTM and tests pass.

I'm merging early so that our CI finishes quicker.

Thanks @keewis

PR #4150, opened 2020-06-12
parameter documentation for DataArray.sel

keewis:
I noticed yesterday that `DataArray.sel` doesn't document its parameters. Instead, it points to `Dataset.sel` (using the `See Also` section).


dcherian:
LGTM. Thanks @keewis!

PR #4154, opened 2020-06-15
Update issue templates inspired/based on dask

shoyer:
See https://github.com/dask/dask/issues/new/choose for an approximate example of what this looks like.


shoyer:
Assuming no objections, I'll go ahead and merge this in a day or two

PR #4155, opened 2020-06-15
Implement interp for interpolating between chunks of data (dask)

### LINKED ISSUE #4078: Feature request: Implement interp for interpolating between chunks of data (dask)
In a project of mine I need to interpolate a dask-based xarray between chunk of data.

I made it work using monkey patching. I'm pretty sure that you can write it better but I made it as good as I could.

I hope that what I wrote can help you implement it properly.

```python
from typing import Union, Tuple, Callable, Any, List

import dask.array as da
import numpy as np
import xarray as xr
import xarray.core.missing as m

def interp_func(var: Union[np.ndarray, da.Array],
                x: Tuple[xr.DataArray, ...],
                new_x: Tuple[xr.DataArray, ...],
                method: str,
                kwargs: Any) -> da.Array:
    """
    multi-dimensional interpolation for array-like. Interpolated axes should be
    located in the last position.

    Parameters
    ----------
    var: np.ndarray or dask.array.Array
        Array to be interpolated. The final dimension is interpolated.
    x: a list of 1d array.
        Original coordinates. Should not contain NaN.
    new_x: a list of 1d array
        New coordinates. Should not contain NaN.
    method: string
        {'linear', 'nearest', 'zero', 'slinear', 'quadratic', 'cubic'} for
        1-dimensional itnterpolation.
        {'linear', 'nearest'} for multidimensional interpolation
    **kwargs:
        Optional keyword arguments to be passed to scipy.interpolator

    Returns
    -------
    interpolated: array
        Interpolated array

    Note
    ----
    This requiers scipy installed.

    See Also
    --------
    scipy.interpolate.interp1d
    """

    try:
        # try the official interp_func first
        res = official_interp_func(var, x, new_x, method, kwargs)
        return res
    except NotImplementedError:
        # may fail if interpolating between chunks
        pass

    if len(x) == 1:
        func, _kwargs = m._get_interpolator(method, vectorizeable_only=True,
                                            **kwargs)
    else:
        func, _kwargs = m._get_interpolator_nd(method, **kwargs)

    # reshape new_x (TODO REMOVE ?)
    current_dims = [_x.name for _x in x]
    new_x = tuple([_x.set_dims(current_dims) for _x in new_x])

    # number of non interpolated dimensions
    nconst = var.ndim - len(x)

    # duplicate the ghost cells of the array
    bnd = {i: "none" for i in range(len(var.shape))}
    depth = {i: 0 if i < nconst else 1 for i in range(len(var.shape))}
    var_with_ghost = da.overlap.overlap(var, depth=depth, boundary=bnd)

    # chunks x and duplicate the ghost cells of x
    x = tuple(da.from_array(_x, chunks=chunks) for _x, chunks in zip(x, var.chunks[nconst:]))
    x_with_ghost = tuple(da.overlap.overlap(_x, depth={0: 1}, boundary={0: "none"})
                         for _x in x)

    # compute final chunks
    chunks_end = [np.cumsum(sizes) - 1 for _x in x
                                       for sizes in _x.chunks]
    chunks_end_with_ghost = [np.cumsum(sizes) - 1 for _x in x_with_ghost
                                                  for sizes in _x.chunks]
    total_chunks = []
    for dim, ce in enumerate(zip(chunks_end, chunks_end_with_ghost)):
        l_new_x_ends: List[np.ndarray] = []
        for iend, iend_with_ghost in zip(*ce):

            arr = np.moveaxis(new_x[dim].data, dim, -1)
            arr = arr[tuple([0] * (len(arr.shape) - 1))]

            n_no_ghost = (arr <= x[dim][iend]).sum()
            n_ghost = (arr <= x_with_ghost[dim][iend_with_ghost]).sum()

            equil = np.ceil(0.5 * (n_no_ghost + n_ghost)).astype(int)

            l_new_x_ends.append(equil)

        new_x_ends = np.array(l_new_x_ends)
        chunks = new_x_ends[0], *(new_x_ends[1:] - new_x_ends[:-1])
        total_chunks.append(tuple(chunks))
    final_chunks = var.chunks[:-len(x)] + tuple(total_chunks)

    # chunks new_x
    new_x = tuple(da.from_array(_x, chunks=total_chunks) for _x in new_x)

    # reshape x_with_ghost (TODO REMOVE ?)
    x_with_ghost = da.meshgrid(*x_with_ghost, indexing='ij')

    # compute on chunks (TODO use drop_axis and new_axis ?)
    res = da.map_blocks(_myinterpnd, var_with_ghost, func, _kwargs, len(x_with_ghost), *x_with_ghost, *new_x,
                        dtype=var.dtype, chunks=final_chunks)

    # reshape res and remove empty chunks (TODO REMOVE ?)
    res = res.squeeze()
    new_chunks = tuple([tuple([chunk for chunk in chunks if chunk > 0]) for chunks in res.chunks])
    res = res.rechunk(new_chunks)
    return res


def _myinterpnd(var: da.Array,
                func: Callable[..., Any],
                kwargs: Any,
                nx: int,
                *arrs: da.Array) -> da.Array:
    _old_x, _new_x = arrs[:nx], arrs[nx:]

    # reshape x (TODO REMOVE ?)
    old_x = tuple([np.moveaxis(tmp, dim, -1)[tuple([0] * (len(tmp.shape) - 1))]
                   for dim, tmp in enumerate(_old_x)])

    new_x = tuple([xr.DataArray(_x) for _x in _new_x])

    return m._interpnd(var, old_x, new_x, func, kwargs)


official_interp_func = m.interp_func
m.interp_func = interp_func
```

pums974:
In a project of mine I need to interpolate a dask-based xarray between chunk of data.

When using the current official `interp` function (xarray v0.15.1), the code:

```python
    datax = xr.DataArray(data=da.from_array(np.arange(0, 4), chunks=2),
                         coords={"x": np.linspace(0, 1, 4)},
                         dims="x")
    datay = xr.DataArray(data=da.from_array(np.arange(0, 4), chunks=2),
                         coords={"y": np.linspace(0, 1, 4)},
                         dims="y")
    data = datax * datay

    # both of these interp call fails
    res = datax.interp(x=np.linspace(0, 1))
    print(res.load())

    res = data.interp(x=np.linspace(0, 1), y=0.5)
    print(res.load())
```
fails with `NotImplementedError: Chunking along the dimension to be interpolated (0) is not yet supported.`, but succeed with this version

I also want to alert that my version does not work with "advanced interpolation" (as shown in the xarray documentation)
Also, my version cannot be used to make `interpolate_na` work with chunked data

 - [x] Closes #4078
 - [x] Tests added


pums974:
On my computer it passes pytest:
```
$> pytest . 
======================= test session starts =================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.8.2, pluggy-0.13.1
[...]
===== 3822 passed, 2710 skipped, 77 xfailed, 24 xpassed, 32 warnings in 48.25s ========

$> pip freeze
appdirs==1.4.4
attrs==19.3.0
black==19.10b0
click==7.1.2
dask==2.18.1
flake8==3.8.3
isort==4.3.21
mccabe==0.6.1
more-itertools==8.4.0
numpy==1.18.5
packaging==20.4
pandas==1.0.4
pathspec==0.8.0
pluggy==0.13.1
py==1.8.2
pycodestyle==2.6.0
pyflakes==2.2.0
pyparsing==2.4.7
pytest==5.4.3
python-dateutil==2.8.1
pytz==2020.1
PyYAML==5.3.1
regex==2020.6.8
scipy==1.4.1
six==1.15.0
toml==0.10.1
toolz==0.10.0
typed-ast==1.4.1
wcwidth==0.2.4
-e git+git@github.com:pums974/xarray.git@c47a1d5d8fd7ca401a0dddea67574af00c4d8e3b#egg=xarray
```

rabernat:
Thanks for this contribution @pums974! We appreciate your patience in awaiting a review of your PR.

pums974:
No problem, we are all very busy. But thanks for your message.

fujiisoup:
Hi @pums974 

Thanks for sending the PR.
I'm working to review it, but it may take more time.

A few comments;
Does it work with an unsorted destination?
e.g.,
```python
da.interp(y=[0, -1, 2])
```

I'm feeling that the basic algorithm, such as `np.interp`-equivalence, should be interpreted in upstream.
I'm sure Dask community welcomes this addition.
Do you have an interest on it?

fujiisoup:
Also in my local environment, it gives 
```AttributeError: 'memoryview' object has no attribute 'dtype'```

The full stack trace is
```
_______________________________________________________ test_interpolate_1d[1-y-cubic] ________________________________________________________

method = 'cubic', dim = 'y', case = 1

    @pytest.mark.parametrize("method", ["linear", "cubic"])
    @pytest.mark.parametrize("dim", ["x", "y"])
    @pytest.mark.parametrize("case", [0, 1])
    def test_interpolate_1d(method, dim, case):
        if not has_scipy:
            pytest.skip("scipy is not installed.")
    
        if not has_dask and case in [1]:
            pytest.skip("dask is not installed in the environment.")
    
        da = get_example_data(case)
        xdest = np.linspace(0.0, 0.9, 80)
    
        actual = da.interp(method=method, **{dim: xdest})
    
        # scipy interpolation for the reference
        def func(obj, new_x):
            return scipy.interpolate.interp1d(
                da[dim],
                obj.data,
                axis=obj.get_axis_num(dim),
                bounds_error=False,
                fill_value=np.nan,
                kind=method,
            )(new_x)
    
        if dim == "x":
            coords = {"x": xdest, "y": da["y"], "x2": ("x", func(da["x2"], xdest))}
        else:  # y
            coords = {"x": da["x"], "y": xdest, "x2": da["x2"]}
    
        expected = xr.DataArray(func(da, xdest), dims=["x", "y"], coords=coords)
>       assert_allclose(actual, expected)

xarray/tests/test_interp.py:86: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
xarray/testing.py:132: in compat_variable
    return a.dims == b.dims and (a._data is b._data or equiv(a.data, b.data))
xarray/testing.py:31: in _data_allclose_or_equiv
    return duck_array_ops.allclose_or_equiv(arr1, arr2, rtol=rtol, atol=atol)
xarray/core/duck_array_ops.py:221: in allclose_or_equiv
    arr1 = np.array(arr1)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/array/core.py:1314: in __array__
    x = self.compute()
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/base.py:165: in compute
    (result,) = compute(self, traverse=False, **kwargs)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/base.py:436: in compute
    results = schedule(dsk, keys, **kwargs)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/local.py:527: in get_sync
    return get_async(apply_sync, 1, dsk, keys, **kwargs)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/local.py:494: in get_async
    fire_task()
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/local.py:466: in fire_task
    callback=queue.put,
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/local.py:516: in apply_sync
    res = func(*args, **kwds)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/local.py:227: in execute_task
    result = pack_exception(e, dumps)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/local.py:222: in execute_task
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/core.py:119: in _execute_task
    return func(*args2)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/optimization.py:982: in __call__
    return core.get(self.dsk, self.outkey, dict(zip(self.inkeys, args)))
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/core.py:149: in get
    result = _execute_task(task, cache)
../../../anaconda3/envs/xarray/lib/python3.7/site-packages/dask/core.py:119: in _execute_task
    return func(*args2)
xarray/core/missing.py:830: in _dask_aware_interpnd
    return _interpnd(var, old_x, new_x, func, kwargs)
xarray/core/missing.py:793: in _interpnd
    x, new_x = _floatize_x(x, new_x)
xarray/core/missing.py:577: in _floatize_x
    if _contains_datetime_like_objects(x[i]):
xarray/core/common.py:1595: in _contains_datetime_like_objects
    return is_np_datetime_like(var.dtype) or contains_cftime_datetimes(var)
xarray/core/common.py:1588: in contains_cftime_datetimes
    return _contains_cftime_datetimes(var.data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

array = <memory at 0x7f771d6daef0>

    def _contains_cftime_datetimes(array) -> bool:
        """Check if an array contains cftime.datetime objects
        """
        try:
            from cftime import datetime as cftime_datetime
        except ImportError:
            return False
        else:
>           if array.dtype == np.dtype("O") and array.size > 0:
E           AttributeError: 'memoryview' object has no attribute 'dtype'

xarray/core/common.py:1574: AttributeError
```

**Comment by pums974** [comment]
Thanks, 
That's weird, I have no problem in mine...
What are your versions of dask and numpy ?

As for implementing this in dask, you may be right, it probably belong there, 
But I am even less use to their code base, and have no clue where to put it.

And for unsorted destination, that's something I didn't think about.
maybe we can add an `argsort` at the begging.

**Comment by keewis** [comment]
@pums974, the CI gets the same error (e.g. [here](https://dev.azure.com/xarray/xarray/_build/results?buildId=3066&view=logs&j=ba13898e-1dfb-5ace-9966-8b7af3677790&t=2bb5a181-aa33-506a-9515-1f09b3e5c016&l=717)) so you should be able to reproduce this by setting up an environment with something like
```sh
conda env create -f ci/requirements/py38.yml -n xarray-py38
conda activate xarray-py38
```

fujiisoup:
> As for implementing this in dask, you may be right, it probably belong there,
> But I am even less use to their code base, and have no clue where to put it.

OK. 
Even so, I would suggest restructuring the code base;
maybe we can add an `interp1d` equivalence into
core.dask_array_ops.interp1d
which works with dask-arrays (non-xarray object).
It'll be easier to test.
The API should be the as same with `scipy.interp.interp1d` as possible.

In missing.py, we can call this function.

pums974:
Hum, ok, but I don't see how it would work if all points are between chunks (see my second example)

fujiisoup:
> Hum, ok, but I don't see how it would work if all points are between chunks (see my second example)

Maybe we can support sequential interpolation only at this moment.
In this case,
```python
res = data.interp(x=np.linspace(0, 1), y=0.5)
```
can be interpreted as
```python
res = data.interp(x=np.linspace(0, 1)).interp(y=0.5)
```
which might not be too difficult.

pums974:
ok, but what about 
```python
res = data.interp(y=0.5)
```

pums974:
I mean, in this case you have to interpolate in another direction. You cannot consider having a 1d function.

xarray/core/missing.py
```diff
@@ -751,3 +803,76 @@ def _interpnd(var, x, new_x, func, kwargs):
...
+            for dim, tmp in enumerate(_old_x)
+        ]
+    )
+
+    new_x = tuple([DataArray(_x) for _x in _new_x])
+
+    return _interpnd(var, old_x, new_x, func, kwargs)
+
+
+def _add_interp_ghost(var, x, nconst: int):
```
dcherian, line 833: can we use `map_overlap` instead of `_add_interp_ghost + map_blocks`?

pums974: Not directly because it doesn't pass `*args` through to `map_blocks`

dcherian: I think it does. See https://github.com/dask/dask/blob/137d039250eadf757b8941d6e1c5cfb6c1e61679/dask/array/overlap.py#L714

Is there a specific problem you were running into? Maybe pass them as kwargs?

pums974: OK, sorry, I answered too fast...
In fact the problem is that I need `var` with overlap, `old_x` with overlap and `new_x` **without** overlap...
And I don't think I can do this without `map_blocks`

pums974:
@fujiisoup I managed to solve the issues you raised about ``` AttributeError: 'memoryview' object has no attribute 'dtype'```
This was due to datetime index. Using an `IndexVariable` seems to solve it

Also I realize that for 1d interpolation `cubic` and `quadratic` method are allowed which may not give the same result with chunked data (or even crash if there is not enough data in the chunked direction). Now they are forbidden

xarray/core/missing.py
```diff
@@ -751,3 +808,83 @@ def _interpnd(var, x, new_x, func, kwargs):
...
+    Currently this need original coordinate to be full arrays (meshgrid)
+
+    TODO: find a way to use 1d coordinates
+    """
+    _old_x, _new_x = coords[:n_coords], coords[n_coords:]
+
+    # reshape x (TODO REMOVE)
+    old_x = tuple(
+        [
+            IndexVariable(
```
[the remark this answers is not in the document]
pums974: It seems necessary with a `datetime` index. This was the root of the memoryview error.
The problem arise later on when `floatize_x` is called. The method `_to_numeric` does not exist for `numpy` array

xarray/core/missing.py
```diff
@@ -751,3 +808,83 @@ def _interpnd(var, x, new_x, func, kwargs):
...
+                str(dim), np.moveaxis(tmp, dim, -1)[tuple([0] * (len(tmp.shape) - 1))]
```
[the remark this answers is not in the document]
pums974: I agree, hence the Todo.
But I'm afraid that using 1d coordinates prevent me to use `map_blocks` and I don't know how iterate efficiently over all the blocks with the block index for any number of dimension.

Something like:
```python
for index, block in var.indexed_blocks():
    x = [ _x.block[i] for i, _x in zip(index, _old_x)]
    xx = [ _x.block[i] for i, _x in zip(index, _new_x)]
    res.block[index] = interp_func(var.block[index], x, xx, interp_kwarg)
```

xarray/core/missing.py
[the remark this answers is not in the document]
fujiisoup: I see. Then, let's use `Variable`, which might be cheaper.

xarray/core/missing.py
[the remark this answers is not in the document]
fujiisoup: Does `        x_with_ghost = da.meshgrid(*x_with_ghost, indexing="ij")` make the destination n-dimensional? I've no answer yet how to improve, but I understood the situation. Let's keep it as is:)

xarray/core/missing.py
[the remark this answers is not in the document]
pums974: ok

xarray/core/missing.py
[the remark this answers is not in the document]
pums974: That is exactly the purpose the `meshgrid`.

From the numpy documentation:
>  `numpy.meshgrid(*xi, copy=True, sparse=False, indexing='xy')`
>  Return coordinate matrices from coordinate vectors.
>  Make N-D coordinate arrays for vectorized evaluations of N-D scalar/vector fields over N-D grids, given one-dimensional coordinate arrays x1, x2,…, xn.

xarray/core/missing.py
```diff
@@ -751,3 +808,83 @@ def _interpnd(var, x, new_x, func, kwargs):
...
+    chunks_end_with_ghost = [
+        np.cumsum(sizes) - 1 for _x in x_with_ghost for sizes in _x.chunks
+    ]
+    total_chunks = []
+    for dim, ce in enumerate(zip(chunks_end, chunks_end_with_ghost)):
+        l_new_x_ends: List[np.ndarray] = []
+        for iend, iend_with_ghost in zip(*ce):
+
+            arr = np.moveaxis(new_x[dim].data, dim, -1)
+            arr = arr[tuple([0] * (len(arr.shape) - 1))]
```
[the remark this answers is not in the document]
pums974: ok

xarray/core/missing.py
```diff
@@ -751,3 +808,83 @@ def _interpnd(var, x, new_x, func, kwargs):
...
+        da.overlap.overlap(_x, depth={0: 1}, boundary={0: "none"}) for _x in x
+    )
+    return var_with_ghost, x_with_ghost
+
+
+def _compute_chunks(x, x_with_ghost, new_x):
+    """Compute equilibrated chunks of new_x
+
+    TODO: This only works if new_x is a set of 1d coordinate
+          more general function is needed for advanced interpolation with chunked dimension
```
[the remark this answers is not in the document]
pums974: is it better now ?

pums974:
> Thanks, @pums974 for this update.
> I left some comments.
> 
> Can you add some tests for more edge cases?
> Something we may want to check would be
> 
>     * scalar interpolation
>     * interpolation into an unsorted dimension (e.g., `da.interp(x=[0, 3, 2])`)

Your welcome, thanks for the feedback

 - **scalar interpolation**: you mean a test like `test_interpolate_nd_scalar` but between chunks ?
 - **Unsorted interpolation:** As I said, I did not looked into it, This need some work, presumably an `argsort` at the begining in order to interpolate in a sorted dimension and reorder the result into the requested order.
I'm trying to implement it, but this seems a bit more challenging than I thought.

pums974:
@fujiisoup I managed to work it through the support for unsorted interpolation.

Also, I reworked the tests, I now test for much more situations.

fujiisoup:
Thanks @pums974 for this update and sorry for my late response.
It looks good but I'll take a deeper look in the next week.

pums974:
While at it, I added the missing bit to make it work with cubic or quadratic method.
I'm not touching the code anymore, waiting for review.

xarray/core/missing.py
```diff
@@ -706,22 +721,140 @@ def interp_func(var, x, new_x, method, kwargs):
...
             dtype=var.dtype,
-            chunks=chunks,
-            new_axis=new_axis,
-            drop_axis=drop_axis,
+            chunks=final_chunks,
+        )
+
+        # reshape res and remove empty chunks
+        # TODO: remove it by using drop_axis and new_axis in map_blocks
+        res = res.squeeze()
```
fujiisoup, line 838: Is it safe to give the squeezing axis explicitly?
What happens if the original array already has a size-one dimension?

xarray/core/missing.py
```diff
@@ -706,22 +721,140 @@ def interp_func(var, x, new_x, method, kwargs):
...
+        x = tuple(
+            da.from_array(_x, chunks=chunks)
+            for _x, chunks in zip(x, var.chunks[nconst:])
+        )
+
+        # duplicate the ghost cells of the array in the interpolated dimensions
+        depth = {"quadratic": 2, "cubic": 3}.get(method, 1)
+        var, x, var_with_ghost, x_with_ghost = _add_interp_ghost(
+            var, x, nconst, depth=depth
+        )
```
fujiisoup, line 810: Are you sure if cubic and quadratic interpolations should work with a chunked array if we add a few edge points?
As far as I know, at least cubic spline is not a local algorithm and in principle values at a very far point can affect the interpolated result.

pums974: In my modest research I read that the support of bspline are compact. But maybe I'm mistaken, I'll have to check more thoroughly.

xarray/core/missing.py
pums974: No you're right, it's probably not safe, hence the TODO. 
But at the time I didn't manage to use `drop_axis` and `new_axis`...
I'll give an another try tomorrow.

xarray/core/missing.py
fujiisoup: Yes, B-spline is compact, but I don't think cubic spline is.

pums974: A quick test shows that indeed quadratic and cubic spline are not local algorithms in `scipy.interpolate.interp1d`
```python
import numpy as np
from scipy.interpolate import interp1d

x = np.linspace(0,6,20)
y = np.sin(x)
y2 = np.sin(x)
y2[10] += 1

x2 = np.linspace(0,6,100)
y3 = interp1d(x,y, kind="cubic")(x2)
y4 = interp1d(x,y2, kind="cubic")(x2)
y3-y4
```

Should I refuse to interpolate, or should I emit a warning ?

fujiisoup: I lean not to support inexact methods.

pums974: ok

pums974:
Guys, I got it.
I managed to use `da.blockwise` which allows me to overcome all the previous limitations.

The result is much more simple, much more reliable.

pums974:
Since I was on it, I extended the decomposition of orthogonal interpolation.
If you want I can break this into two PR.

xarray/core/missing.py
```diff
@@ -751,3 +753,61 @@ def _interpnd(var, x, new_x, func, kwargs):
...
+
+    # simple speed up for the local interpolation
+    if method in ["linear", "nearest"]:
+        var, indexes_coords = _localize(var, indexes_coords)
+    localized_x, localized_new_x = zip(*[indexes_coords[d] for d in indexes_coords])
+
+    return _interpnd(var.data, localized_x, localized_new_x, interp_func, interp_kwargs)
+
+
+def decompose_interp(indexes_coords):
```
fujiisoup, line 784: Nice improvement!

pums974:
FYI, don't merge yet. I fixed a bug today, but did not push it.
And there is some work to do on the testing side.

fujiisoup:
OK.
If you have additional time, it would be nicer if you could add more comments on tests, like what is being tested there ;)

pums974:
OK, I'm happy with the results now (better than my first submission of course).

I did not add so much tests since the result replace what was done before,
thus the previous tests applies.

I'm going for some holidays so I won't work that much for the time being.
But I'll be able to answer any questions.

Thanks for the reviewing and pushing me into doing a much better job.

chrisroat:
Hi!  This work is interesting to me, as I was implementing in dask an image processing algo which needs an intermediate 1-d linear interpolation step.  This bottlenecks the calculation through a single node.  Your work here on distributed interpolation is intriguing, and I'm wondering if it would be useful in my work and if it could possibly become part of dask itself.

[Here](https://github.com/deisseroth-lab/dask-image/blob/b943ed067a4f51d6756c00a076a158b132a36a6c/dask_image/skexposure/histogram_matching.py#L6) is the particular function, which you'll note has a dask.delayed wrapper around np.interp.

pums974:
Hi.

I agree, part of this work might belong in dask. But I don't know dask internals enough to go there.
In this case, everything was already in place.

Moreover I do think that there is room for optimization.
In particular, in this implementation, the work is distributed along chunks corresponding to *destination*. This means that one may have big intermediate array.
For example interpolating one value in a chunked vector will load the full vector in memory (first localization aside).
In my previous implementation (and uglier), the interpolation was done with the chunks of the starting array. This might be a better choice sometimes.

fujiisoup:
This PR looks good for me.
Maybe we can wait for a few days in case anyone has some comments on it.
If no comments, I'll merge this then.

fujiisoup:
Thanks @pums974 :)

pums974:
You're welcome :)

cyhsu:
Hi 
Just curious about this.  I followed the discussion since this issue addressed.  Is this chunk interpolation solved already?

fujiisoup:
@cyhsu 
Yes, in the current master.

cyhsu:
@fujiisoup Thanks for letting me know. But I am still unable to do even though I have updated my xarray via "conda update xarray".

fujiisoup:
@cyhsu Yes, because it is not yet released.
(I'm not sure when the next release will be, but maybe a few months later)
If you do `pip install git+https://github.com/pydata/xarray`, the current master will be installed in your system and interpolation over the chunks can be used.
But note that this means you will install (a kind of) beta version.

cyhsu:
@fujiisoup  Thanks for the response. Since I have not updated my xarray package through this beta version. I hope you can answer my additional question for me.  By considering the interpolation, which way is faster? a. chunk the dataset, and then do interpolation or b. chunk the interpolation list and then do interpolation?

a.     

    datax = xr.DataArray(data=da.from_array(np.arange(0, 4), chunks=2),
                         coords={"x": np.linspace(0, 1, 4)},
                         dims="x")
    datay = xr.DataArray(data=da.from_array(np.arange(0, 4), chunks=2),
                         coords={"y": np.linspace(0, 1, 4)},
                         dims="y")
    data = datax * datay

    # both of these interp call fails
    res = datax.interp(x=np.linspace(0, 1))
    print(res.load())

    res = data.interp(x=np.linspace(0, 1), y=0.5)
    print(res.load())

b.     

    datax = xr.DataArray(data=np.arange(0, 4),
                         coords={"x": np.linspace(0, 1, 4)},
                         dims="x")
    datay = xr.DataArray(data=np.arange(0, 4),
                         coords={"y": np.linspace(0, 1, 4)},
                         dims="y")
    data = datax * datay

   x = xr.DataArray(data = da.from_array(np.linspace(0,1), chunks=2), dims='x')
   res = data.interp(x=x)

pums974:
@cyhsu I can answer this question.

For best performance you should chunk the input array on the non interpolated dimensions and chunk the destination. Aka :

```
datax = xr.DataArray(data=np.arange(0, 4),
                     coords={"x": np.linspace(0, 1, 4)},
                     dims="x")
datay = xr.DataArray(data=da.from_array(np.arange(0, 4), chunks=2),
                     coords={"y": np.linspace(0, 1, 4)},
                     dims="y")
data = datax * datay

x = xr.DataArray(data = da.from_array(np.linspace(0,1), chunks=2), dims='x')

res = data.interp(x=x)
```

cyhsu:
@pums974  then how about if we do the interpolation by using chunk input array to the chunk interpolated dimension?

pums974:
If the input array is chunked in the interpolated dimension, the chunks will be merge during the interpolation.

This may induce a large memory cost at some point, but I do not know how to avoid it...

pums974:
Do this answer your question?

cyhsu:
Gotcha! Yes it is. If I have many points in lat, lon, depth, and time, I should better chunk my input arrays at this stage to speed up the performance.

pums974:
In your case, each task (20 000) will have the entire input (100²), and interpolate a few points (5²). 

Maybe the overhead comes with duplicating the input array 20 000 times

pums974:
I forgot to take into account that the interpolating are orthogonal
So in sequential we are doing 2 interpolation first x then y
In parallel we do the same:
The fist interpolation will have 20 000 tasks, each task will have the totality of the input array, and compute an interpolation of 5 point of the output (x) producing an array of 5x100 per task or 100 000x100 full result as an intermediate array.
The second interpolation will have 20 000² tasks each task will have a block of 5x100 point of the intermediate array and compute an interpolation on 5 point of the output (y) resulting in a 5² array per task and the 100 000² full result.

So plenty of room for overhead...

pums974:
And I forgot to take into account that your interpolation only need 47² points of the input array, so the input array will be reduced at the start of the process (you can replace every 100 by 47 in my previous answers)

lazyoracle:
@max-sixty Is there a timeline on when we can expect this feature in a stable release? Is it scheduled for the next minor release and to be made available on `conda` and `pip`?

PR #4163, opened 2020-06-17
silence UnitStrippedWarnings

keewis:
Now that the tests are cleaned up, it is time to fix the numerous "unit stripped" warnings. We'll get there by promoting all `UnitStrippedWarning`s to errors (using `filterwarnings`) and fixing / xfailing the failing tests.


keewis:
except from the failing `operator.eq` tests (which are due to `pint` upstream and will be fixed soon) and the `pad` tests that need to be rewritten only the tests from the `VariableSubclassobjects` are left.

Since they use `assert_array_equal` and `np.testing.assert_allclose` (which don't dispatch), I'm not sure how to deal with them: If I wanted to fix them in `TestVariable` I would probably end up rewriting all of these. Other than that, I think we could also write our own version of `assert_array_equal` / `assert_allclose` that works on duck arrays. We already have the functionality in `duck_array_ops.py`, we'd just need to write the summary code. That would probably have to be a new PR, though.

dcherian:
>  I think we could also write our own version of assert_array_equal / assert_allclose that works on duck arrays. 

This sounds right to me.

keewis:
unfortunately, this doesn't seem to really work: `VariableSubtests` was written to use the same tests for `Variable` and `IndexVariable`, and using it to test duckarrays (`pint` in particular) is difficult since the data that is often used to compare the result is not converted. I'm not sure if rewriting `VariableSubtests` to optionally allow converting the data or adding specific Variable-with-units tests is easier...

max-sixty:
I just saw your message from a couple of days ago @keewis . 
To the extent you've solved that, great. 
If you haven't — one option would be to change `assert_array_equal` in those tests to xarray's `assert_equal`, and ensure xarray's functions handle the duck arrays reasonably.

keewis:
I tried that by writing some sort of `assert_duckarray_equal` and `assert_duckarray_allclose` function. That should work for most duck arrays, but not `pint`: the tests often compare the data with the variable (or the variable's data), and since the data is mostly a bare numpy array this fails for `pint`: bare numpy arrays are by definition dimensionless and so the comparison either returns `False` (the assertion fails) or a `DimensionalityError` is raised.

So to make this work I think we'd need to rewrite `VariableSubtests` to have another factory method (`array_type`? `data_type`?) which is called on all data before it is passed to `cls`. That, however, has its own set of issues and in the end I concluded that `VariableSubtests` was just not written with duck arrays in mind, and that pytest doesn't encourage reusing tests by using inheritance (marking or removing tests from base classes is really clunky). With that in mind I simply removed the use of `VariableSubtests` for now (this shouldn't decrease the coverage too much) and we can add it back if / when we notice that we actually need them.

keewis:
Once the CI passes this should be ready for review and merging.

xarray/core/common.py
```diff
@@ -1434,7 +1434,7 @@ def _full_like_variable(other, fill_value, dtype: DTypeLike = None):
             other.shape, fill_value, dtype=dtype, chunks=other.data.chunks
         )
     else:
-        data = np.full_like(other, fill_value, dtype=dtype)
+        data = np.full_like(other.data, fill_value, dtype=dtype)
```
dcherian, line 1437: This LGTM.

The tests are a little hard to review. Is there anything specific you want us to look at?

keewis: well, the idea is that the tests mostly failed due to the indexes, so I split each test into variants so units in data, dimension coordinates and non-dimension coordinates are tested separately (and we can xfail the dimension coordinates tests).

Other than that, there's the new `weighted` and `pad` tests and I just replaced some `assert np.allclose(...)` statements with a new `assert_duckarray_allclose(...)` function. I think a review could be restricted to just reviewing those new tests / functions.

xarray/testing.py
```diff
@@ -148,6 +155,68 @@ def compat_variable(a, b):
...
+
+
+def assert_duckarray_equal(x, y, err_msg="", verbose=True):
+    """ Like `np.testing.assert_array_equal`, but for duckarrays """
+    __tracebackhide__ = True
+
+    def array_like(x):
+        return hasattr(x, "ndim") and hasattr(x, "shape") and hasattr(x, "dtype")
+
+    def scalar(x):
```
dcherian, line 202: minor comment: can we use `is_scalar` from `/core/utils.py` instead?

xarray/testing.py
```diff
@@ -148,6 +155,68 @@ def compat_variable(a, b):
...
+        )
+
+    if not array_like(x) and not scalar(x):
+        x = np.asarray(x)
+
+    if not array_like(y) and not scalar(y):
+        y = np.asarray(y)
+
+    if (array_like(x) and scalar(y)) or (scalar(x) and array_like(y)):
+        equiv = (x == y).all()
```
dcherian, line 214: I'm surprised we want `assert_duckarray_equal(np.array([1, 1]), 1) == True`.

xarray/core/common.py
dcherian: Just two minor comments. The added tests look great!

xarray/testing.py
keewis: sure, and I guess I can also move `array_like` there, if we don't have one already.

xarray/testing.py
keewis: well, that's what `numpy.testing.assert_array_equal` does. Even `numpy.testing.assert_equal` works that way (I'm not sure what the difference between both is, though)

dcherian: :+1:

dcherian:
Thanks @keewis. I'm really excited to see this go in.

PR #4170, opened 2020-06-22
use builtin python types instead of the numpy alias

### LINKED ISSUE #4168: upstream-dev failures due to numpy dtype deprecations
It seems `numpy` recently changed their dtype system and we now get [warnings](https://dev.azure.com/xarray/xarray/_build/results?buildId=3092&view=logs&j=2280efed-fda1-53bd-9213-1fa8ec9b4fa8&t=175181ee-1928-5a6b-f537-168f7a8b7c2d) for `np.bool`, `np.int` and `np.long`. Instead, the warnings suggest we should use the python types or `np.compat.long`.

Since we have a few tests that require zero warnings, this means our test suite fails (and more than 13000 warnings are not a good thing, anyways).

keewis:
Not sure if this enough, but that's something the CI will tell me, right?

 - [x] might close #4168
 - [ ] Tests added

### DIFF
--- a/xarray/core/formatting.py
+++ b/xarray/core/formatting.py
@@ -140,7 +140,7 @@ def format_item(x, timedelta_format=None, quote_strings=True):
         return format_timedelta(x, timedelta_format=timedelta_format)
     elif isinstance(x, (str, bytes)):
         return repr(x) if quote_strings else x
-    elif isinstance(x, (float, np.float)):
+    elif isinstance(x, (float, np.float_)):
         return f"{x:.4}"
     else:
         return str(x)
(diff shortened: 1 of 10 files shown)


keewis:
It seems that was enough to fix the CI. Ready for review and merge, I guess?

xarray/core/formatting.py
```diff
@@ -140,7 +140,7 @@ def format_item(x, timedelta_format=None, quote_strings=True):
         return format_timedelta(x, timedelta_format=timedelta_format)
     elif isinstance(x, (str, bytes)):
         return repr(x) if quote_strings else x
-    elif isinstance(x, (float, np.float)):
+    elif isinstance(x, (float, np.float_)):
```
keewis, line 143: we might change this to just
```suggestion
    elif isinstance(x, float):
```
since
```python
In [1]: import numpy as np 
   ...:  
   ...: a = np.float64(0) 
   ...: a.dtype == float, isinstance(a, float)
Out[1]: (True, True)
```
but I'm not sure if this also works with the minimum supported numpy version (1.15 / 1.16)

PR #4175, opened 2020-06-24
Bump minimum versions for 0.16 release

dcherian:

PR #3713, opened 2020-01-21
bump min deps for 0.15

dcherian:
- [x] Closes #1679
 - [x] Passes `black . && mypy . && flake8`
 - [x] Fully documented, including `whats-new.rst` for all changes and `api.rst` for new API

PR #4178, opened 2020-06-25
Fix min_deps_check; revert to support numpy=1.14 and pandas=0.24

shoyer:
Fixes the issue noticed in:
https://github.com/pydata/xarray/pull/4175#issuecomment-649135372

Let's see if this passes CI...

### DIFF
--- a/ci/min_deps_check.py
+++ b/ci/min_deps_check.py
@@ -141,11 +141,9 @@ def process_pkg(
     policy_months = POLICY_MONTHS.get(pkg, POLICY_MONTHS_DEFAULT)
     policy_published = datetime.now() - timedelta(days=policy_months * 30)
 
-    policy_versions = {
-        version: date for version, date in versions.items() if date < policy_published
-    }
-    version_and_date = lambda version: (version, policy_versions[version])
-    policy_major, policy_minor = max(policy_versions, key=version_and_date)
+    policy_major, policy_minor = max(
+        version for version, date in versions.items() if date < policy_published
+    )
     policy_published_actual = versions[policy_major, policy_minor]
 
     if (req_major, req_minor) < (policy_major, policy_minor):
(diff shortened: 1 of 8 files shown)


xarray/core/nputils.py
```diff
@@ -14,6 +13,17 @@
     _USE_BOTTLENECK = False
 
 
+def normalize_axis_index(data, axis):
```
shoyer, line 16: I added this because I noticed it when manually reverting the quantile change from https://github.com/pydata/xarray/pull/3713/files

ci/min_deps_check.py
```diff
@@ -141,21 +141,20 @@ def process_pkg(
     policy_months = POLICY_MONTHS.get(pkg, POLICY_MONTHS_DEFAULT)
     policy_published = datetime.now() - timedelta(days=policy_months * 30)
 
-    policy_major = req_major
-    policy_minor = req_minor
-    policy_published_actual = req_published
-    for (major, minor), published in reversed(sorted(versions.items())):
-        if published < policy_published:
-            break
```
shoyer, line 149: I'm not *entirely* sure why this was broken, but the new logic is significantly simpler so I'm more confident it's correct. In particular, it determines the policy versions without using the currently required versions as defaults.

shoyer:
Marking this as a draft -- as noted in https://github.com/pydata/xarray/pull/4175#issuecomment-649209534, I just misread our current policy.

ci/min_deps_check.py
shoyer: An alternative implement of the original policy that is possibly clearer:
```suggestion
    # "the minor version (X.Y) initially published no more than N months ago"
    policy_major, policy_minor = min(
        (version for version, date in versions.items() if date > policy_published),
        default=(req_major, req_minor),
    )
```

andersy005:
With the three releases (v0.16.0, v0.16.1, v0.16.2) since this PR, is this PR still relevant?

dcherian:
I think we still want the updates to `min_deps_check.py`, and then update our min dependencies if still appropriate.

keewis:
with the merge of #4907 and #4942 I think we can close this

Issue #4142, opened 2020-06-10
Should we make "rasterio" an engine option?


WeatherGod:
In a similar vein to how #4003 is going for zarr files, I would like to see if a rasterio engine could be created so that geotiff files could get opened through open_mfdataset() and friends. I am willing to put some cycles to putting this together. It has been a *long* time since I did the initial prototype for the pynio backend back in the xray days. Does anyone see any immediate pitfalls or gotchas for doing this?

jhamman:
Thanks @WeatherGod for opening this issue. It is good to have you back around these parts. 

The main thing to consider at the moment is that we are about to start a major backend refactor (#1970). So I'd suggest holding on off on any major work that builds on our current DataStore implementation.

From an API design, I think it makes sense to revisit the rasterio->DataArray convention we currently have and instead treat rasterio like most other drivers and return a Dataset by default. For most applications, the following would be true:

```python
open_rasterio(...) == open_dataarray(..., engine='rasterio') == open_dataset(..., engine='rasterio')[default_var]
```

There's also the question of where the rasterio backend should live once #1970 is implemented. I think we can make a pretty strong argument that it should move out of xarray and into a 3rd party package. Perhaps rasterio itself, or, more likely something like rio-xarray (cc @snowman2). See #3166 for a demo of how this may work. 

As you can see, there are a few big questions up in the air right now. Work on #1970 will begin this summer so now is a great time to game out the various options for the rasterio backend.

fmaussion:
As a not-so-active-but-still-interested xarray dev my opinion doesn't count much, but I would be a proponent of having the rasterio backend live outside of xarray proper. 

At the time we wrote the  rasterio->DataArray conversion we already noticed a lot of issues regarding differences between the two datamodels, and  rio-xarray shows that there is a lot of logic and dev work necessary for this to go smoothly, which would be better handled outside of xarray.

snowman2:
In `rioxarray` we updated `open_rasterio` so it opens a `tif` file like a DataArray and netcdf-like files as a dataset. It also loads in the metadata/tags as well:

### NetCDF file as Dataset

```python
import rioxarray
xds = rioxarray.open_rasterio("PLANET_SCOPE_3D.nc")
```
```
<xarray.Dataset>
Dimensions:      (time: 2, x: 10, y: 10)
Coordinates:
  * y            (y) float64 8.085e+06 8.085e+06 ... 8.085e+06 8.085e+06
[...]

PR #4187, opened 2020-06-30
Xarray open_mfdataset with engine Zarr

### LINKED ISSUE #3668: open_mfdataset: support for multiple zarr datasets

weiji14:
Work on enabling `xr.open_dataset(..., engine="zarr")`, to ~~replace~~ complement `xr.open_zarr`. This also allows `xr.open_mfdataset(..., engine="zarr") to be used.

Note: Credit should be given to @Mikejmnez, I'm just continuing this on from #4003.

 - [x] Closes #3668, closes #4003
 - [x] Tests added


weiji14:
This is the one test failure on Linux py36-bare-minimum:

```python-traceback
__________________________ TestDataset.test_lazy_load __________________________

self = <xarray.tests.test_dataset.TestDataset object at 0x7fa80b2b7be0>

   def test_lazy_load(self):
       store = InaccessibleVariableDataStore()
       create_test_data().dump_to_store(store)
   
       for decode_cf in [True, False]:
>           ds = open_dataset(store, decode_cf=decode_cf)

xarray/tests/test_dataset.py:4188: 
xarray/backends/api.py:578: in open_dataset
   engine = _get_engine_from_magic_number(filename_or_obj)

filename_or_obj = <xarray.tests.test_dataset.InaccessibleVariableDataStore object at 0x7fa80b2b7d30>

   def _get_engine_from_magic_number(filename_or_obj):
       # check byte header to determine file type
       if isinstance(filename_or_obj, bytes):
           magic_number = filename_or_obj[:8]
       else:
>           if filename_or_obj.tell() != 0:
E           AttributeError: 'InaccessibleVariableDataStore' object has no attribute 'tell'

xarray/backends/api.py:116: AttributeError
```

Been scratching my head debugging this one. There doesn't seem to be an obvious reason why this test is failing, since 1) this test isn't for Zarr and 2) this test shouldn't be affected by the new `if` blocks checking if `engine=="zarr"`. Will need to double check the logic here.

weiji14:
Nevermind, I found it. There was an `if` that should have been an `elif`. Onward to the next error...

weiji14:
Is it ok to drop the deprecated `auto_chunk` tests here in this PR (or leave it to another PR)? The deprecation warning was first added in https://github.com/pydata/xarray/pull/2530/commits/ae4cf0ab19b3e563bde90a48b3e6ee615930d4a1, and I see that `auto_chunk` was used back in v0.12.1 at http://xarray.pydata.org/en/v0.12.1/generated/xarray.open_zarr.html.

xarray/backends/zarr.py
```diff
@@ -582,128 +626,36 @@ def open_zarr(
...
-            concat_characters=concat_characters,
-            decode_coords=decode_coords,
-            drop_variables=drop_variables,
-            decode_timedelta=decode_timedelta,
-        )
-
-        # TODO: this is where we would apply caching
-
-        return ds
+    backend_kwargs = {
```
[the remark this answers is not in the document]
weiji14: Ok, I've updated the io.rst documentation to use `open_dataset(..., engine="zarr")` instead of `open_zarr`. Besides `consolidated`, we'll need to document these two Zarr-specific `backend_kwargs` too as mentioned in #4189:

- `synchronizer` (object, optional) – Array synchronizer provided to zarr
- `overwrite_encoded_chunks` (bool, optional) – Whether to drop the zarr chunks encoded for each variable when a dataset is loaded with specified chunk sizes (default: False)

weiji14:
Four more failures, something to do with dask?

- `TestZarrDictStore.test_vectorized_indexing` - IndexError: only slices with step >= 1 are supported
- `TestZarrDictStore.test_manual_chunk` - ZeroDivisionError: integer division or modulo by zero
- `TestZarrDirectoryStore.test_vectorized_indexing` - IndexError: only slices with step >= 1 are supported
- `TestZarrDirectoryStore.test_manual_chunk` - ZeroDivisionError: integer division or modulo by zero

<details>

__________________ TestZarrDictStore.test_vectorized_indexing __________________

self = <xarray.tests.test_backends.TestZarrDictStore object at 0x7f5832433940>

    @pytest.mark.xfail(
        not has_dask,
        reason="the code for indexing without dask handles negative steps in slices incorrectly",
    )
    def test_vectorized_indexing(self):
        in_memory = create_test_data()
        with self.roundtrip(in_memory) as on_disk:
            indexers = {
                "dim1": DataArray([0, 2, 0], dims="a"),
                "dim2": DataArray([0, 2, 3], dims="a"),
            }
            expected = in_memory.isel(**indexers)
            actual = on_disk.isel(**indexers)
            # make sure the array is not yet loaded into memory
            assert not actual["var1"].variable._in_memory
            assert_identical(expected, actual.load())
            # do it twice, to make sure we're switched from
            # vectorized -> numpy when we cached the values
            actual = on_disk.isel(**indexers)
    
        def multiple_indexing(indexers):
            # make sure a sequence of lazy indexings certainly works.
            with self.roundtrip(in_memory) as on_disk:
                actual = on_disk["var3"]
                expected = in_memory["var3"]
                for ind in indexers:
                    actual = actual.isel(**ind)
                    expected = expected.isel(**ind)
                    # make sure the array is not yet loaded into memory
                    assert not actual.variable._in_memory
                assert_identical(expected, actual.load())
    
        # two-staged vectorized-indexing
        indexers = [
            {
                "dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"]),
                "dim3": DataArray([[0, 4], [1, 3], [2, 2]], dims=["a", "b"]),
            },
            {"a": DataArray([0, 1], dims=["c"]), "b": DataArray([0, 1], dims=["c"])},
        ]
        multiple_indexing(indexers)
    
        # vectorized-slice mixed
        indexers = [
            {
                "dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"]),
                "dim3": slice(None, 10),
            }
        ]
        multiple_indexing(indexers)
    
        # vectorized-integer mixed
        indexers = [
            {"dim3": 0},
            {"dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"])},
            {"a": slice(None, None, 2)},
        ]
        multiple_indexing(indexers)
    
        # vectorized-integer mixed
        indexers = [
            {"dim3": 0},
            {"dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"])},
            {"a": 1, "b": 0},
        ]
        multiple_indexing(indexers)
    
        # with negative step slice.
        indexers = [
            {
                "dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"]),
                "dim3": slice(-1, 1, -1),
            }
        ]
>       multiple_indexing(indexers)

xarray/tests/test_backends.py:686: 
xarray/tests/test_backends.py:642: in multiple_indexing
    assert_identical(expected, actual.load())
xarray/core/dataarray.py:814: in load
    ds = self._to_temp_dataset().load(**kwargs)
xarray/core/dataset.py:666: in load
    v.load()
xarray/core/variable.py:381: in load
    self._data = np.asarray(self._data)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/numpy/core/numeric.py:501: in asarray
xarray/core/indexing.py:677: in __array__
    self._ensure_cached()
xarray/core/indexing.py:674: in _ensure_cached
    self.array = NumpyIndexingAdapter(np.asarray(self.array))
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/numpy/core/numeric.py:501: in asarray
xarray/core/indexing.py:653: in __array__
    return np.asarray(self.array, dtype=dtype)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/numpy/core/numeric.py:501: in asarray
xarray/core/indexing.py:557: in __array__
    return np.asarray(array[self.key], dtype=None)
xarray/backends/zarr.py:57: in __getitem__
    return array[key.tuple]
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/core.py:572: in __getitem__
    return self.get_basic_selection(selection, fields=fields)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/core.py:698: in get_basic_selection
    fields=fields)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/core.py:738: in _get_basic_selection_nd
    indexer = BasicIndexer(selection, self)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/indexing.py:279: in __init__
    dim_indexer = SliceDimIndexer(dim_sel, dim_len, dim_chunk_len)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/indexing.py:107: in __init__
    err_negative_step()

    def err_negative_step():
>       raise IndexError('only slices with step >= 1 are supported')
E       IndexError: only slices with step >= 1 are supported

/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/errors.py:55: IndexError
_____________________ TestZarrDictStore.test_manual_chunk ______________________

self = <xarray.tests.test_backends.TestZarrDictStore object at 0x7f5832b80cf8>

    @requires_dask
    @pytest.mark.filterwarnings("ignore:Specified Dask chunks")
    def test_manual_chunk(self):
        original = create_test_data().chunk({"dim1": 3, "dim2": 4, "dim3": 3})
    
        # All of these should return non-chunked arrays
        NO_CHUNKS = (None, 0, {})
        for no_chunk in NO_CHUNKS:
            open_kwargs = {"chunks": no_chunk}
>           with self.roundtrip(original, open_kwargs=open_kwargs) as actual:

xarray/tests/test_backends.py:1594: 
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/contextlib.py:81: in __enter__
    return next(self.gen)
xarray/tests/test_backends.py:1553: in roundtrip
    with self.open(store_target, **open_kwargs) as ds:
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/contextlib.py:81: in __enter__
    return next(self.gen)
xarray/tests/test_backends.py:1540: in open
    with xr.open_dataset(store_target, engine="zarr", **kwargs) as ds:
xarray/backends/api.py:587: in open_dataset
    ds = maybe_decode_store(store, chunks)
xarray/backends/api.py:511: in maybe_decode_store
    for k, v in ds.variables.items()
xarray/backends/api.py:511: in <dictcomp>
    for k, v in ds.variables.items()
xarray/backends/zarr.py:398: in maybe_chunk
    var = var.chunk(chunk_spec, name=name2, lock=None)
xarray/core/variable.py:1007: in chunk
    data = da.from_array(data, chunks, name=name, lock=lock, **kwargs)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:2712: in from_array
    chunks, x.shape, dtype=x.dtype, previous_chunks=previous_chunks
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:2447: in normalize_chunks
    (),
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:2445: in <genexpr>
    for s, c in zip(shape, chunks)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:954: in blockdims_from_blockshape
    for d, bd in zip(shape, chunks)

.0 = <zip object at 0x7f58332d9d48>

        ((bd,) * (d // bd) + ((d % bd,) if d % bd else ()) if d else (0,))
>       for d, bd in zip(shape, chunks)
    )
E   ZeroDivisionError: integer division or modulo by zero

/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:954: ZeroDivisionError
_______________ TestZarrDirectoryStore.test_vectorized_indexing ________________

self = <xarray.tests.test_backends.TestZarrDirectoryStore object at 0x7f5832a08a20>

    @pytest.mark.xfail(
        not has_dask,
        reason="the code for indexing without dask handles negative steps in slices incorrectly",
    )
    def test_vectorized_indexing(self):
        in_memory = create_test_data()
        with self.roundtrip(in_memory) as on_disk:
            indexers = {
                "dim1": DataArray([0, 2, 0], dims="a"),
                "dim2": DataArray([0, 2, 3], dims="a"),
            }
            expected = in_memory.isel(**indexers)
            actual = on_disk.isel(**indexers)
            # make sure the array is not yet loaded into memory
            assert not actual["var1"].variable._in_memory
            assert_identical(expected, actual.load())
            # do it twice, to make sure we're switched from
            # vectorized -> numpy when we cached the values
            actual = on_disk.isel(**indexers)
    
        def multiple_indexing(indexers):
            # make sure a sequence of lazy indexings certainly works.
            with self.roundtrip(in_memory) as on_disk:
                actual = on_disk["var3"]
                expected = in_memory["var3"]
                for ind in indexers:
                    actual = actual.isel(**ind)
                    expected = expected.isel(**ind)
                    # make sure the array is not yet loaded into memory
                    assert not actual.variable._in_memory
                assert_identical(expected, actual.load())
    
        # two-staged vectorized-indexing
        indexers = [
            {
                "dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"]),
                "dim3": DataArray([[0, 4], [1, 3], [2, 2]], dims=["a", "b"]),
            },
            {"a": DataArray([0, 1], dims=["c"]), "b": DataArray([0, 1], dims=["c"])},
        ]
        multiple_indexing(indexers)
    
        # vectorized-slice mixed
        indexers = [
            {
                "dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"]),
                "dim3": slice(None, 10),
            }
        ]
        multiple_indexing(indexers)
    
        # vectorized-integer mixed
        indexers = [
            {"dim3": 0},
            {"dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"])},
            {"a": slice(None, None, 2)},
        ]
        multiple_indexing(indexers)
    
        # vectorized-integer mixed
        indexers = [
            {"dim3": 0},
            {"dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"])},
            {"a": 1, "b": 0},
        ]
        multiple_indexing(indexers)
    
        # with negative step slice.
        indexers = [
            {
                "dim1": DataArray([[0, 7], [2, 6], [3, 5]], dims=["a", "b"]),
                "dim3": slice(-1, 1, -1),
            }
        ]
>       multiple_indexing(indexers)

xarray/tests/test_backends.py:686: 
xarray/tests/test_backends.py:642: in multiple_indexing
    assert_identical(expected, actual.load())
xarray/core/dataarray.py:814: in load
    ds = self._to_temp_dataset().load(**kwargs)
xarray/core/dataset.py:666: in load
    v.load()
xarray/core/variable.py:381: in load
    self._data = np.asarray(self._data)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/numpy/core/numeric.py:501: in asarray
xarray/core/indexing.py:677: in __array__
    self._ensure_cached()
xarray/core/indexing.py:674: in _ensure_cached
    self.array = NumpyIndexingAdapter(np.asarray(self.array))
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/numpy/core/numeric.py:501: in asarray
xarray/core/indexing.py:653: in __array__
    return np.asarray(self.array, dtype=dtype)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/numpy/core/numeric.py:501: in asarray
xarray/core/indexing.py:557: in __array__
    return np.asarray(array[self.key], dtype=None)
xarray/backends/zarr.py:57: in __getitem__
    return array[key.tuple]
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/core.py:572: in __getitem__
    return self.get_basic_selection(selection, fields=fields)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/core.py:698: in get_basic_selection
    fields=fields)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/core.py:738: in _get_basic_selection_nd
    indexer = BasicIndexer(selection, self)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/indexing.py:279: in __init__
    dim_indexer = SliceDimIndexer(dim_sel, dim_len, dim_chunk_len)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/indexing.py:107: in __init__
    err_negative_step()

    def err_negative_step():
>       raise IndexError('only slices with step >= 1 are supported')
E       IndexError: only slices with step >= 1 are supported

/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/zarr/errors.py:55: IndexError
___________________ TestZarrDirectoryStore.test_manual_chunk ___________________

self = <xarray.tests.test_backends.TestZarrDirectoryStore object at 0x7f5831763ef0>

    @requires_dask
    @pytest.mark.filterwarnings("ignore:Specified Dask chunks")
    def test_manual_chunk(self):
        original = create_test_data().chunk({"dim1": 3, "dim2": 4, "dim3": 3})
    
        # All of these should return non-chunked arrays
        NO_CHUNKS = (None, 0, {})
        for no_chunk in NO_CHUNKS:
            open_kwargs = {"chunks": no_chunk}
>           with self.roundtrip(original, open_kwargs=open_kwargs) as actual:

xarray/tests/test_backends.py:1594: 
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/contextlib.py:81: in __enter__
    return next(self.gen)
xarray/tests/test_backends.py:1553: in roundtrip
    with self.open(store_target, **open_kwargs) as ds:
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/contextlib.py:81: in __enter__
    return next(self.gen)
xarray/tests/test_backends.py:1540: in open
    with xr.open_dataset(store_target, engine="zarr", **kwargs) as ds:
xarray/backends/api.py:587: in open_dataset
    ds = maybe_decode_store(store, chunks)
xarray/backends/api.py:511: in maybe_decode_store
    for k, v in ds.variables.items()
xarray/backends/api.py:511: in <dictcomp>
    for k, v in ds.variables.items()
xarray/backends/zarr.py:398: in maybe_chunk
    var = var.chunk(chunk_spec, name=name2, lock=None)
xarray/core/variable.py:1007: in chunk
    data = da.from_array(data, chunks, name=name, lock=lock, **kwargs)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:2712: in from_array
    chunks, x.shape, dtype=x.dtype, previous_chunks=previous_chunks
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:2447: in normalize_chunks
    (),
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:2445: in <genexpr>
    for s, c in zip(shape, chunks)
/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:954: in blockdims_from_blockshape
    for d, bd in zip(shape, chunks)

.0 = <zip object at 0x7f58324b7f48>

        ((bd,) * (d // bd) + ((d % bd,) if d % bd else ()) if d else (0,))
>       for d, bd in zip(shape, chunks)
    )
E   ZeroDivisionError: integer division or modulo by zero

/usr/share/miniconda/envs/xarray-tests/lib/python3.6/site-packages/dask/array/core.py:954: ZeroDivisionError

</details>

xarray/tests/test_backends.py
```diff
@@ -609,8 +609,8 @@ def test_orthogonal_indexing(self):
             assert_identical(expected, actual)
 
     @pytest.mark.xfail(
-        not has_dask,
-        reason="the code for indexing without dask handles negative steps in slices incorrectly",
+        has_dask,
+        reason="the code for indexing with dask handles negative steps in slices incorrectly",
```
weiji14, line 613: This is strange, and goes opposite to the statement at https://github.com/pydata/xarray/pull/3919#issuecomment-606267473 :confused:

- Before: `test_vectorized_indexing` would *fail* when there is **no dask**.
- Now: the test *fails* **with dask**, but* passes* when there is **no dask**.

weiji14, line 613: This is strange, and goes opposite to the statement at https://github.com/pydata/xarray/pull/3919#issuecomment-606267473 :confused:

- Before: `test_vectorized_indexing` would *fail* when there is **no dask**.
- Now: the test *fails* **with dask**, but* passes* when there is **no dask**.

xarray/backends/api.py
```diff
@@ -344,12 +345,12 @@ def open_dataset(
...
-        'pseudonetcdf'}, optional
+        'pseudonetcdf', 'zarr'}, optional
         Engine to use when reading files. If not provided, the default engine
         is chosen based on available dependencies, with a preference for
         'netcdf4'.
     chunks : int or dict, optional
-        If chunks is provided, it used to load the new dataset into dask
+        If chunks is provided, it is used to load the new dataset into dask
         arrays. ``chunks={}`` loads the dataset with dask using a single
         chunk for all arrays.
```
weiji14, line 355: Wonder if it's a good idea to put backend specific notes here:

```suggestion
        If chunks is provided, it is used to load the new dataset into dask
        arrays. ``chunks={}`` loads the dataset with dask using a single
        chunk for all arrays. When using ``engine="zarr"`, if `chunks='auto'`,
        dask chunks are created based on the variable's zarr chunks, and if
        `chunks=None`, zarr array data will lazily convert to numpy arrays upon
        access. This accepts all the chunk specifications as Dask does.
```

shoyer:
I wonder if it's really worth deprecating `open_zarr()`. `open_dataset(..., engine='zarr')` is a bit more verbose, especially with `backend_kwargs` to pass optional arguments. It seems pretty harmless to keep `open_zarr()` around, especially if it's just an alias for `open_datraset(engine='zarr')`.

shoyer:
We could also automatically detect zarr stores in `open_dataset` if:

1. the argument inherits from `collections.abc.Mapping`, and
2. it contains a key `'.zgroup'`, corresponding to zarr metadata.

shoyer:
As for the annoyance of needing to write `backend_kwargs={"consolidated": True}`, I wonder if we could detect this automatically by checking for the existence of a `.zmetadata` key?

weiji14:
> I wonder if it's really worth deprecating `open_zarr()`. `open_dataset(..., engine='zarr')` is a bit more verbose, especially with `backend_kwargs` to pass optional arguments. It seems pretty harmless to keep `open_zarr()` around, especially if it's just an alias for `open_datraset(engine='zarr')`.

Depends on which line in the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) you want to follow - "Simple is better than complex", or "There should be one-- and preferably only one --obvious way to do it". From a maintenance perspective, it's balancing the cost of a deprecation cycle vs writing code that tests both instances I guess.

> We could also automatically detect zarr stores in `open_dataset` without requiring `engine='zarr'` if:
> 
> 1. the argument inherits from `collections.abc.Mapping`, and
> 2. it contains a key `'.zgroup'`, corresponding to zarr metadata.
>
> As for the annoyance of needing to write `backend_kwargs={"consolidated": True}`, I wonder if we could detect this automatically by checking for the existence of a `.zmetadata` key? This would add a small amount of overhead (one file access) but this probably would not be prohibitively expensive.

These are some pretty good ideas. I also wonder if there's a way to mimic the [dataset identifiers like in rasterio](https://rasterio.readthedocs.io/en/latest/topics/datasets.html#dataset-identifiers), something like `xr.open_dataset("zarr:some_zarrfile.zarr")`. Feels a lot more like [fsspec's url chaining](https://filesystem-spec.readthedocs.io/en/latest/features.html#url-chaining) too.

Counter-argument would be that the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of `open_dataset` is already too high, and it really should be refactored. Especially if new backend engines come online (e.g. #4142).

dcherian:
> I wonder if it's really worth deprecating open_zarr(). open_dataset(..., engine='zarr') is a bit more verbose, especially with backend_kwargs to pass optional arguments. It seems pretty harmless to keep open_zarr() around, especially if it's just an alias for open_datraset(engine='zarr').

I agree. I think we should keep `open_zarr` around.

As reminder (because it took me a while to remember!), one goal with this refactor is to have `open_mfdataset` work with all backends (including zarr and rasterio) by specifying the `engine` kwarg.

Carreau:
As a note I'm working on implementing [zarr spec v3 in zarr-python](https://github.com/zarr-developers/zarr-python/pull/568), still deciding how we want to handle the new spec/API.

If there are any changes that you would like or dislike in an API, feedback is welcome.

weiji14:
> I agree. I think we should keep `open_zarr` around.

Just wanted to mention that two of the reviewers in the last PR (see https://github.com/pydata/xarray/pull/4003#issuecomment-619644606 and https://github.com/pydata/xarray/pull/4003#issuecomment-620169860) seemed **in favour** of deprecating `open_zarr`. If I'm counting the votes correctly (did I miss anyone?), that's 2 for, and 2 against. We'll need a tiebreaker :laughing:

> As reminder (because it took me a while to remember!), one goal with this refactor is to have `open_mfdataset` work with all backends (including zarr and rasterio) by specifying the `engine` kwarg.

Yes exactly, time does fly (half a year has gone by already!).

Currently I'm trying to piggyback Zarr into `test_openmfdataset_manyfiles`, and am currently having trouble finding out why opening Zarr stores via `open_mfdataset` doesn't return a `dask` backed array like the other engines. Might need to spend another day digging through the code to see if this is expected behaviour.

> As a note I'm working on implementing [zarr spec v3 in zarr-python](https://github.com/zarr-developers/zarr-python/pull/568), still deciding how we want to handle the new spec/API.
> 
> If there are any changes that you would like or dislike in an API, feedback is welcome.

Thanks for chipping in @Carreau! I'm sure the community will have some useful suggestions. Just cross-referencing https://zarr-developers.github.io/zarr/specs/2019/06/19/zarr-v3-update.html so others can get a better feel for where things are at.

xarray/tests/test_backends.py
```diff
@@ -2572,8 +2582,14 @@ def test_open_mfdataset_manyfiles(
...
-            # check that using open_mfdataset returns dask arrays for variables
-            assert isinstance(actual["foo"].data, dask_array_type)
+            try:
+                # check that using open_mfdataset returns dask arrays for variables
+                assert isinstance(actual["foo"].data, dask_array_type)
+            except AssertionError:
+                # A numpy array is returned instead of a dask array
+                # when reading with Zarr and chunks is None
+                if readengine == "zarr" and chunks is None:
+                    assert isinstance(actual["foo"].data, np.ndarray)
```
weiji14, line 2592: Will need to figure out here why a dask array isn't returned when using `xr.open_mfdataset` with `chunks=None`.

weiji14: Resolved with commit 02f86ca51a5b4182576ac1a66b73fddd011b4d57. Specifically, using `xr.open_mfdataset` with `chunks=None` and `engine="zarr"` should return dask arrays now instead of numpy arrays.

xarray/tests/test_backends.py
```diff
@@ -609,8 +609,8 @@ def test_orthogonal_indexing(self):
             assert_identical(expected, actual)
 
     @pytest.mark.xfail(
-        not has_dask,
-        reason="the code for indexing without dask handles negative steps in slices incorrectly",
+        has_dask or not has_dask,
+        reason="the code for indexing with or without dask handles negative steps in slices incorrectly",
```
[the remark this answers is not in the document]
weiji14: `open_zarr` uses `chunks="auto"` by default, whereas `open_dataset` defaults to `chunks=None` (they have quite different behaviour in the Zarr backend). I think this is a problem because using `chunks=None` with `engine="zarr"` will automatically convert the arrays to numpy arrays upon access (see http://xarray.pydata.org/en/v0.15.1/generated/xarray.open_zarr.html ), instead of keeping them as dask arrays, and that might explain why this trivial test fails when refactoring from `open_zarr` to `open_dataset(engine="zarr")`.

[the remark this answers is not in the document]
weiji14: I've updated the Zarr tests to use `chunks="auto"` as was the default in `open_zarr`, and so `test_vectorized_indexing` works properly for negative slices when `dask` is installed (see commit dce4e7cd1fcf35fb7d3293bf6cc410646b588c64 and 31ce87d75834ddab6466a013c1aae4d55a450b3a). Happy to resolve this, but there should be a discussion around whether `open_mfdataset` with `engine="zarr"` should use `chunks="auto"` as in `open_zarr`, or `chunks=None` as with the other backends (e.g. netcdf, h5netcdf, etc).

shoyer:
> Just to note down a few things:
> 
> 1. The deprecated "auto_chunk" kwarg was removed
> 2. `open_zarr` uses `chunks="auto"` by default, whereas `open_dataset` uses `chunks=None` (see my comment inline)
> 
> The different default chunk behaviour (point 2) is worth raising, and it might be best to postpone the deprecation of `open_zarr` until the next release, so that there's time to discuss what we want the default setting to be (`None` or `auto`).

These all sound good to me!

I agree that we shouldn't change the default behavior for `open_dataset`, and should keep `open_zarr` around for now -- there is no urgent need to deprecate it.

xarray/backends/zarr.py
```diff
@@ -591,129 +637,37 @@ def open_zarr(
...
-        if auto_chunk:
-            chunks = "auto"  # maintain backwards compatibility
-        else:
-            chunks = None
 
-        warnings.warn(
-            "auto_chunk is deprecated. Use chunks='auto' instead.",
-            FutureWarning,
-            stacklevel=2,
-        )
```
shoyer, line 605: For now I would suggest keeping this compatibility code and not (yet) marking `open_zarr` as deprecated.

doc/io.rst
```diff
@@ -970,12 +970,12 @@ with consolidated metadata. To write consolidated metadata, pass the
...
-    ds.to_zarr('foo.zarr', consolidated=True)
+    ds.to_zarr("foo.zarr", consolidated=True)
 
-To read a consolidated store, pass the ``consolidated=True`` option to
-:py:func:`open_zarr`::
+To read a consolidated store, pass ``{"consolidated": True}`` to the
+backend_kwargs option when using :py:func:`open_dataset` with ``engine="zarr"``::
 
-    ds = xr.open_zarr('foo.zarr', consolidated=True)
+    ds = xr.open_dataset("foo.zarr", engine="zarr", backend_kwargs={"consolidated": True})
```
shoyer, line 978: I appreciate that it's nice to be able to use `open_dataset` and `open_mfdataset` with zarr, but these examples really seem a bit less readable to me! This feels like an argument in favor of keeping around `open_zarr`, as a version of `open_dataset` with better defaults (and keyword arguments) for zarr. Potentially it would eventually make sense to have dedicated `open_*` functions for each file-format xarray supports?

xarray/backends/api.py
```diff
@@ -487,10 +489,40 @@ def maybe_decode_store(store, lock=False):
             )
             name_prefix = "open_dataset-%s" % token
             ds2 = ds.chunk(chunks, name_prefix=name_prefix, token=token)
-            ds2._file_obj = ds._file_obj
+
+        elif engine == "zarr":
```
shoyer, line 493: My main concern with this code is that introducing an entirely separate code-path inside `open_dataset()` for chunking zarr in particular feels strange and a little unexpected. Any time we use totally separate code branches for some logic, the odds of introducing inconsistencies/bugs increases greatly.

I wonder if we could consolidate this logic somehow in order to avoid adding a separate branch for the code here? For example, we could put a `get_chunk` method on all xarray backends classes, even if it currently only returns a filler value and/or raises an error for `chunks='auto'`? Chunking is not unique to zarr, e.g., netCDF4 files also have chunks, although the default "auto" chunking logic should probably be different.

I would be OK holding this off for a later clean-up, but this really would be worth doing eventually. CC @alexamici RE: the backends refactor.

xarray/backends/zarr.py
weiji14: Ok I can revert the open_zarr deprecation warning. Should I also not remove the "auto_chunk" compatibility here?

doc/io.rst
weiji14: Yep, I can revert the changes to `io.rst` here (i.e. keep `open_zarr` as the documented way of opening Zarr files instead of using `open_dataset`).

xarray/backends/api.py
```diff
@@ -487,10 +489,40 @@ def maybe_decode_store(store, lock=False):
...
+                try:
+                    import dask.array  # noqa
+                except ImportError:
+                    chunks = None
+
+            # auto chunking needs to be here and not in ZarrStore because
+            # the variable chunks does not survive decode_cf
+            # return trivial case
+            if not chunks:  # e.g. chunks is 0, None or {}
+                return ds
```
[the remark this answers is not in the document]
dcherian: open_zarr defaults to chunks="auto"
    open_dataset default to chunks=None

At one of our past meetings, the consensus was to keep `open_zarr` around with `chunks="auto"` as default for backward compatibility. 

`open_dataset(engine="zarr")` should use the current default `chunks` kwarg.

With that in mind can we avoid adding this stuff here?

xarray/backends/api.py
weiji14: > My main concern with this code is that introducing an entirely separate code-path inside open_dataset() for chunking zarr in particular feels strange and a little unexpected. Any time we use totally separate code branches for some logic, the odds of introducing inconsistencies/bugs increases greatly.
> 
> I wonder if we could consolidate this logic somehow in order to avoid adding a separate branch for the code here? For example, we could put a get_chunk method on all xarray backends classes, even if it currently only returns a filler value and/or raises an error for chunks='auto'? Chunking is not unique to zarr, e.g., netCDF4 files also have chunks, although the default "auto" chunking logic should probably be different.

Thanks for pointing this out! I agree completely that the `open_dataset()` function is overdue for a refactor, it was a nightmare to go through all the if-then branches, but the comprehensive test suite helped to catch most of the bugs and I've tested it on my own real world dataset so the logic *should* be ok :crossed_fingers:.

> I would be OK holding this off for a later clean-up, but this really would be worth doing eventually. CC @alexamici RE: the backends refactor.

Personally I would prefer to hold this off, since this `open_mfdataset` PR (and the previous one at #4003) has been sitting around for months, and I've had to resolve quite a few merge conflicts to keep up. No point in contaminating this complex PR by refactoring the NetCDF logic either.

xarray/backends/api.py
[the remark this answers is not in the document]
weiji14: > ```
> open_zarr defaults to chunks="auto"
> open_dataset default to chunks=None
> ```

Happy with this :ok_hand:

> With that in mind can we avoid adding this stuff here?

Do you mean removing the `if chunks == "auto": ...` code block? I think it should be kept here in `open_dataset` (instead of in the `open_zarr` function) because users might want to do `open_mfdataset(engine="zarr", chunks-"auto")`, and so the `chunks="auto"` logic needs to be in `open_dataset`.

doc/whats-new.rst
```diff
@@ -98,6 +101,14 @@ New Features
...
   document the new method in :doc:`internals`. (:pull:`4248`).
   By `Justus Magin <https://github.com/keewis>`_.
+- Add support for parsing datetime strings formatted following the default
+  string representation of cftime objects, i.e. YYYY-MM-DD hh:mm:ss, in
+  partial datetime string indexing, as well as :py:meth:`~xarray.cftime_range`
+  (:issue:`4337`). By `Spencer Clark <https://github.com/spencerkclark>`_.
+- :py:meth:`~xarray.DataArray.to_dataframe` and :py:meth:`~xarray.Dataset.to_dataframe`
+  now accept a ``dim_order`` parameter allowing to specify the resulting dataframe's
+  dimensions order (:issue:`4331`, :pull:`4333`).
+  By `Thomas Zilio <https://github.com/thomas-z>`_.
```
shoyer, line 111: it looks like this got added in a bad merge?

xarray/backends/api.py
```diff
@@ -487,10 +489,40 @@ def maybe_decode_store(store, lock=False):
...
+            if chunks == "auto":
+                try:
+                    import dask.array  # noqa
+                except ImportError:
+                    chunks = None
+
+            # auto chunking needs to be here and not in ZarrStore because
+            # the variable chunks does not survive decode_cf
+            # return trivial case
+            if not chunks:  # e.g. chunks is 0, None or {}
```
shoyer, line 511: Cam we make this `if chunks is None` instead?

I know this is a discrepancy from how `open_zarr()` works today, but currently `open_dataset(..., chunks={})` is a way to open a dataset with dask chunks equal to the full size of any arrays.

I doubt the (different) behavior of `open_zarr` in this case was intentional....

shoyer (approved):

doc/whats-new.rst
weiji14: Oops yes, you got ahead of me. I've removed this now.

xarray/backends/api.py
weiji14: I'll try it locally and see if the tests break, felt like there was a reason it had to be `not chunks` but can't remember the context now.

shoyer: Thanks! Not a big deal if we have to push off this clean-up.

weiji14: Yep, 2 tests would break with a `ZeroDivisionError` if we switch to `if chunks is None`. Specifically:

- TestZarrDictStore.test_manual_chunk
- TestZarrDirectoryStore.test_manual_chunk

Related to my comment hidden in the mess above at https://github.com/pydata/xarray/pull/4187#issuecomment-652104356 :smile:

shoyer: Well, I would expect `test_manual_chunk` to fail here: it is explicitly verifying that `chunks=0` and `chunks={}` result in-memory numpy arrays. Does it work if you remove those cases, e.g., by setting `NO_CHUNKS = (None,)` ?

weiji14: Yes, setting `NO_CHUNKS = (None,)` works with `if chunks is None`. I'll make the change?

weiji14: If it helps by the way, `test_manual_chunk` with `NO_CHUNKS = (None, 0, {})` was added in #2530.

shoyer: Yes, I think we can go ahead and change that. It doesn't look like that was carefully evaluated in #2530.

xarray/backends/api.py
```diff
@@ -344,14 +345,17 @@ def open_dataset(
...
         is chosen based on available dependencies, with a preference for
         "netcdf4".
     chunks : int or dict, optional
-        If chunks is provided, it used to load the new dataset into dask
+        If chunks is provided, it is used to load the new dataset into dask
         arrays. ``chunks={}`` loads the dataset with dask using a single
-        chunk for all arrays.
+        chunk for all arrays. When using ``engine="zarr"`, if `chunks='auto'`,
+        dask chunks are created based on the variable's zarr chunks, and if
+        `chunks=None`, zarr array data will lazily convert to numpy arrays upon
```
shoyer, line 357: This behavior for `chunks=None` is the same for all backends. The only special behavior for `zarr` is `chunks='auto'`.

xarray/tests/test_backends.py
```diff
@@ -1614,8 +1614,8 @@ def test_auto_chunk(self):
     def test_manual_chunk(self):
         original = create_test_data().chunk({"dim1": 3, "dim2": 4, "dim3": 3})
 
-        # All of these should return non-chunked arrays
-        NO_CHUNKS = (None, 0, {})
+        # Using chunks = None should return non-chunked arrays
+        NO_CHUNKS = (None,)
         for no_chunk in NO_CHUNKS:
```
shoyer, line 1619: could you remove the loop here now?

shoyer (approved):

xarray/tests/test_backends.py
weiji14: Sure, I did think of that actually!

xarray/backends/api.py
weiji14: Ok, will update the docs.

shoyer:
Thanks @weiji14 and @Mikejmnez for your contribution!

martindurant:
Note that zarr.open* now works with fsspec URLs (in master)

dcherian:
Thanks @weiji14  and @Mikejmnez . This is a great contribution.

PR #4191, opened 2020-07-01
use the latest image of RTD

keewis:
This switches back to using the `latest` image on RTD (we tried using `stable` when we had build issues a few months ago and forgot to switch back after they were resolved).


keewis:
I don't think there's much to review here, so I'm merging

PR #4193, opened 2020-07-02
Fix polyfit fail on deficient rank

### LINKED ISSUE #4190: Polyfit fails with few non-NaN values
**What happened**:
A linear `DataArray.polyfit` seems to fail if there are less than 3 non-NaN elements along the fitting dimension.

<details><summary>Traceback</summary>

```
TypeError: only size-1 arrays can be converted to Python scalars

The above exception was the direct cause of the following exception:

  File "polyfit.py", line 6, in <module>
    out = arr.polyfit(dim='x', deg=1)
  File "/home/stephan/venv/variogram/lib/python3.8/site-packages/xarray/core/dataarray.py", line 3455, in polyfit
    return self._to_temp_dataset().polyfit(
  File "/home/stephan/venv/variogram/lib/python3.8/site-packages/xarray/core/dataset.py", line 5962, in polyfit
    coeffs, residuals = duck_array_ops.least_squares(
  File "/home/stephan/venv/variogram/lib/python3.8/site-packages/xarray/core/duck_array_ops.py", line 625, in least_squares
    return nputils.least_squares(lhs, rhs, rcond=rcond, skipna=skipna)
  File "/home/stephan/venv/variogram/lib/python3.8/site-packages/xarray/core/nputils.py", line 239, in least_squares
    out[:, nan_cols] = np.apply_along_axis(
  File "<__array_function__ internals>", line 5, in apply_along_axis
  File "/home/stephan/venv/variogram/lib/python3.8/site-packages/numpy/lib/shape_base.py", line 379, in apply_along_axis
    res = asanyarray(func1d(inarr_view[ind0], *args, **kwargs))
  File "/home/stephan/venv/variogram/lib/python3.8/site-packages/xarray/core/nputils.py", line 227, in _nanpolyfit_1d
    out[:-1], out[-1], _, _ = np.linalg.lstsq(x[~mask, :], arr[~mask], rcond=rcond)
ValueError: setting an array element with a sequence.
```

</details>

I've played around with the degree a little bit and the error seems to occur as soon as `(# of non-NaN values - degree) < 2`

**What you expected to happen**:

The fit to succeed - I think two non-NaN values should be enough for a linear fit. I also noticed that there is no `RankWarning: Polyfit may be poorly conditioned` if the degree is larger than the number of non-NaN values.

**Minimal Complete Verifiable Example**:

```python
import xarray as xr
import numpy as np

arr = xr.DataArray([np.nan, 1, 2], dims='x', coords={'x': [0, 1, 2]})
arr.polyfit(dim='x', deg=1)
```

**Anything else we need to know?**:

**Environment**:

INSTALLED VERSIONS
------------------
commit: None
python: 3.8.2 (default, Apr 27 2020, 15:53:34) 
[GCC 9.3.0]
python-bits: 64
OS: Linux
OS-release: 5.4.0-39-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8
libhdf5: 1.10.4
libnetcdf: 4.6.3

xarray: 0.15.2.dev112+g54b9450b
pandas: 1.0.5
numpy: 1.19.0
scipy: 1.5.0
netCDF4: 1.5.3
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: 2.4.0
cftime: 1.1.3
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2.19.0
distributed: 2.19.0
matplotlib: 3.2.2
cartopy: None
seaborn: None
numbagg: None
pint: None
setuptools: 44.0.0
pip: 20.0.2
conda: None
pytest: None
IPython: 7.16.1
sphinx: None

</details>

aulemahal:
- [x] Closes #4190
 - [x] Tests added

Fixes #4190. In cases where the input matrix had a deficient rank (matrix rank != order) because of the number of NaN values, polyfit would fail, simply because numpy's lstsq returned an empty array for the residuals (instead of a size 1 array). This fixes the problem by catching the case and returning `np.nan` instead.

The other point in the issue was that `RankWarning` is also not raised in that case. That was due to the fact that `da.polyfit` was computing the rank from the coordinate (Vandermonde) matrix, instead of the masked data. Thus, is a given line has too many NaN values, its deficient rank was not detected. I added a test and warning at all places where a rank is computed (5 different lines). Also, to match np.polyfit behaviour of no warning when `full=True`, I changed the warning filters using a context manager, ignoring the `RankWarning` in that case. Overall, it feels a bi ugly because of the duplicated code and it will print the warning for every line of an array that has a deficient rank, which can be a lot...

### DIFF
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -55,7 +55,7 @@ Bug fixes
 ~~~~~~~~~
 - Fixed a bug in backend caused by basic installation of Dask (:issue:`4164`, :pull:`4318`)
   `Sam Morley <https://github.com/inakleinbottle>`_.
-- Fixed a few bugs with :py:meth:`Dataset.polyfit` when encountering deficient matrix ranks (:issue:`4190`, :pull:`4193`) `Pascal Bourgault <https://github.com/aulemahal>`_.
+- Fixed a few bugs with :py:meth:`Dataset.polyfit` when encountering deficient matrix ranks (:issue:`4190`, :pull:`4193`). By `Pascal Bourgault <https://github.com/aulemahal>`_.
 - Fixed inconsistencies between docstring and functionality for :py:meth:`DataArray.str.get`
   and :py:meth:`DataArray.str.wrap` (:issue:`4334`). By `Mathias Hauser <https://github.com/mathause>`_.
 - Fixed overflow issue causing incorrect results in computing means of :py:class:`cftime.datetime`
--- a/doc/whats-new.rst
+++ b/doc/whats-new.rst
@@ -51,6 +51,7 @@ Bug fixes
 ~~~~~~~~~
 - Fixed a bug in backend caused by basic installation of Dask (:issue:`4164`, :pull:`4318`)
   `Sam Morley <https://github.com/inakleinbottle>`_.
+- Fixed a few bugs with :py:meth:`Dataset.polyfit` when encountering deficient matrix ranks (:issue:`4190`, :pull:`4193`) `Pascal Bourgault <https://github.com/aulemahal>`_.
 
 
 Documentation
--- a/xarray/core/nputils.py
+++ b/xarray/core/nputils.py
@@ -224,9 +224,15 @@ def _nanpolyfit_1d(arr, x, rcond=None):
     out = np.full((x.shape[1] + 1,), np.nan)
     mask = np.isnan(arr)
     if not np.all(mask):
-        out[:-1], resid, _, _ = np.linalg.lstsq(x[~mask, :], arr[~mask], rcond=rcond)
+        out[:-1], resid, rank, _ = np.linalg.lstsq(x[~mask, :], arr[~mask], rcond=rcond)
         out[-1] = resid or np.nan
+        warn_on_deficient_rank(rank, x.shape[1])
     return out
+
+
+def warn_on_deficient_rank(rank, order):
+    if rank != order:
+        warnings.warn("Polyfit may be poorly conditioned", np.RankWarning, stacklevel=5)
 
 
 def least_squares(lhs, rhs, rcond=None, skipna=False):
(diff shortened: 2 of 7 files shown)


mathause:
> Overall, it feels a bi ugly because of the duplicated code

Looks fine to me.

> and it will print the warning for every line of an array that has a deficient rank, which can be a lot...

That could be annoying as a user. This happens in `results = da.apply_along_axis(nputils._nanpolyfit_1d, ...)`, right? Can you use a `warnings.simplefilter("once")` so the error is only issued once - something along the lines of:

```python
    warnings.simplefilter("once", np.RankWarning)
    results = da.apply_along_axis(
        nputils._nanpolyfit_1d,
        ...
    )
```

I think this also warrants tests and a whats-new entry.

aulemahal:
So while adding tests, I realized there were more bugs concerning deficient rank matrices and `full=True` output. I believe I fixed all I could find. 
However, I was not able to reduce the number of warnings for the case using dask : the computation occurs outside of the `catch_warnings()` context...
Also, there is a bug in the `dask.array.linalg.lstsq` output (see issue dask/dask#6192), so if `full=True` we must use  `skipna=True` with dask.  Same when rank != order (deficient rank), `dask.array.linalg.lstsq` will fail, so we work around by using our (slower) nan-skipping method.

xarray/core/nputils.py
```diff
@@ -232,10 +232,17 @@ def _nanpolyfit_1d(arr, x, rcond=None):
     out = np.full((x.shape[1] + 1,), np.nan)
     mask = np.isnan(arr)
     if not np.all(mask):
-        out[:-1], out[-1], _, _ = np.linalg.lstsq(x[~mask, :], arr[~mask], rcond=rcond)
+        out[:-1], resid, rank, _ = np.linalg.lstsq(x[~mask, :], arr[~mask], rcond=rcond)
+        out[-1] = resid or np.nan
```
mathause, line 236:
```suggestion
        out[-1] = resid if resid.size > 0 else np.nan
```



This raises a DeprecationWarning:
```
/home/vsts/work/1/s/xarray/core/nputils.py:236: DeprecationWarning:
The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error.
Use `array.size > 0` to check that an array is not empty.
    out[-1] = resid or np.nan
```

xarray/tests/test_dataarray.py
```diff
@@ -4282,6 +4282,10 @@ def test_polyfit(self, use_dask, use_datetime):
         ).T
         assert_allclose(out.polyfit_coefficients, expected, rtol=1e-3)
 
+        # Full output and deficient rank
+        out = da.polyfit("x", 12, full=True)
```
mathause, line 4286: Can you catch this warning as well - it ends up in the test output.

https://dev.azure.com/xarray/xarray/_build/results?buildId=3540&view=logs&j=acc1347f-8235-55b0-95c7-0e2189e61659&t=7284452c-b1fa-561e-b473-09e477cf0d79&l=442

mathause:
Pity that the dask warnings are not caught but I guess there's not much we can do about that.

Does this now work for the originally posted issue? Could you add it as a test?

aulemahal:
@mathause Fixed the tests to catch the warning and added a case for the specific of the issue : too many nans causing a rank deficiency.

doc/whats-new.rst
```diff
@@ -55,6 +55,7 @@ Bug fixes
 ~~~~~~~~~
 - Fixed a bug in backend caused by basic installation of Dask (:issue:`4164`, :pull:`4318`)
   `Sam Morley <https://github.com/inakleinbottle>`_.
+- Fixed a few bugs with :py:meth:`Dataset.polyfit` when encountering deficient matrix ranks (:issue:`4190`, :pull:`4193`) `Pascal Bourgault <https://github.com/aulemahal>`_.
```
keewis, line 58:
```suggestion
- Fixed a few bugs with :py:meth:`Dataset.polyfit` when encountering deficient matrix ranks (:issue:`4190`, :pull:`4193`). By `Pascal Bourgault <https://github.com/aulemahal>`_.
```

mathause:
thanks @aulemahal!

PR #4199, opened 2020-07-04
fix sphinx warnings

keewis:
This adds back the fail-on-warnings flag we lost during the move the RTDs PR preview, and also fixes the warnings that have accumulated since then.

 - [ ] Tests added

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -360,7 +360,6 @@ Computation
    DataArray.rolling_exp
    DataArray.weighted
    DataArray.coarsen
-   DataArray.dt
    DataArray.resample
    DataArray.get_axis_num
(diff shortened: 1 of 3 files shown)


keewis:
it seems we can't enable `fail_on_warning` since now RTD tries to build with in parallel, and `readthedocs-sphinx-ext` doesn't declare itself as `parallel_read_safe`. So we have to wait until the next release of `readthedocs-sphinx-ext` until we can actually enable `fail_on_warning` (or could we try installing from github? Might be risky).

dcherian:
Thanks @keewis

max-sixty:
Thanks @keewis !

PR #4201, opened 2020-07-04
fix the failing matplotlib upstream-dev tests

### LINKED ISSUE #4200: Upstream-dev matplotlib failure
The upstream-dev CI currently fails because `matplotlib` upstream seems to have removed / renamed `Colorbar._label`. We definitely need to change that since it's a private attribute, but I'm not quite sure what we can use instead.

keewis:
These tests were failing because a private attribute was removed.

 - [x] Closes #4200
 - [ ] Tests added

### DIFF
--- a/xarray/tests/test_plot.py
+++ b/xarray/tests/test_plot.py
@@ -88,6 +88,13 @@ def easy_array(shape, start=0, stop=1):
     return a.reshape(shape)
 
 
+def get_colorbar_label(colorbar):
+    if colorbar.orientation == "vertical":
+        return colorbar.ax.get_ylabel()
+    else:
+        return colorbar.ax.get_xlabel()
+
+
 @requires_matplotlib
 class PlotTestCase:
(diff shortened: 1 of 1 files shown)


xarray/tests/test_plot.py
```diff
@@ -88,6 +88,13 @@ def easy_array(shape, start=0, stop=1):
     return a.reshape(shape)
 
 
+def get_colorbar_label(colorbar):
+    if colorbar.orientation == "vertical":
+        return colorbar.ax.get_ylabel()
+    else:
+        return colorbar.ax.get_xlabel()
```
keewis, line 95: it might make sense to push this upstream as `colorbar.get_label()`

shoyer:
Thanks!

xarray/tests/test_plot.py
shoyer: Agreed!

Issue #3750, opened 2020-02-04
isort pre-commit hook does not skip text files


mathause:
#### MCVE Code Sample

Add arbitrary change to the file `doc/pandas.rst`

``` bash
git add doc/pandas.rst
git commit -m "test"
```

The pre-commit hook will fail.

#### Expected Output
the pre-commit hook to pass

#### Problem Description

running `isort -rc doc/*` will change the following files:
``` bash
        modified:   contributing.rst
        modified:   howdoi.rst
        modified:   internals.rst
        modified:   io.rst
        modified:   pandas.rst
        modified:   quick-overview.rst
```
unfortunately it does not behave properly and deletes/ changes arbitrary lines. Can the pre-commit hook be told to only run on *.py files? On the command line this would be `isort -rc *.py`

keewis:
`pre-commit` allows filtering the files passed to a hook using regular expressions:
```diff
   - repo: https://github.com/timothycrosley/isort
     rev: 4.3.21-2
     hooks:
       - id: isort
+        files: ".+[.]py"
   # https://github.com/python/black#version-control-integration
   - repo: https://github.com/python/black
```
Not sure if this regex is the one we should use, but it does seem to work.

mathause:
You can tell isort to skip files in the `setup.cfg` but this seems to be ignored when doing `isort file` (which is what happens in the pre-commit hook) - see https://github.com/pre-commit/mirrors-isort/issues/9#issuecomment-511434587

So I think adapting `.pre-commit-config.yaml` is the way to go at the moment. (Maybe adapt `setup.cfg` anyway to ensure it does not happen when isort is called with `isort -rc .`).

keewis:
actually, `isort` does support filtering using `--filter-files -sg '*.rst'` which would blacklist `rst` files. Since we don't have just `rst` files, I think that whitelisting is a better approach. Maybe we should use `files: .\.py$`?

asottile:
fwiw, the problem here is the `-2` tag of isort is broken -- you want 4.3.21 which [properly specifies `types: [python]`](https://github.com/timothycrosley/isort/commit/30e15ab8251b682e48a1afe76dd65f7bdf39257b)

PR #4204, opened 2020-07-06
update isort CI and pre-commit hook

### LINKED ISSUE #4202: isort flags
Because I've been hit by this elsewhere: `isort` has released a new version today that removes the `-rc` / `--recursive` flag (without deprecation, I think). Once `conda-forge` has been updated, we will start seeing a failing `isort` CI.

Not sure if we should pin `isort` for now or simply remove the flag once the builds are failing (or we remove the flag now and have a pretty much useless `isort` CI until the feedstock has been merged).

Update: `isort` relaxed the error into a warning, so this is not that urgent

keewis:
Now that the conda-forge feedstock has been updated, we would see a failing `isort` CI because of the removed `-rc` (short for `--recursive`). I also updated the version used by `pre-commit` to the `5.0.4` version released about an hour ago, which allows us to remove the file regex as recommended by @asottile in https://github.com/pydata/xarray/issues/3750#issuecomment-624402070 (we couldn't follow that advice before since we were using the `use_parentheses` setting which was broken on `4.3.21`)

 - [x] Closes #4202
 - [ ] Tests added
 - [x] Passes `isort . && black . && mypy . && flake8`

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,6 +2,6 @@
 
  - [ ] Closes #xxxx
  - [ ] Tests added
- - [ ] Passes `isort -rc . && black . && mypy . && flake8`
+ - [ ] Passes `isort . && black . && mypy . && flake8`
  - [ ] User visible changes (including notable bug fixes) are documented in `whats-new.rst`
  - [ ] New functions/methods are listed in `api.rst`
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,10 +2,9 @@
 repos:
   # isort should run before black as black sometimes tweaks the isort output
   - repo: https://github.com/timothycrosley/isort
-    rev: 4.3.21-2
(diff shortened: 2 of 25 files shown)


keewis:
it seems `isort` changed the way it treats imports like `import dask.array as da`: it wants to rewrite them as `from dask import array as da`.

While having such a conflict indicates poor design, this is actually not the same: `import dask.array` will always import a module but `from dask import array` will import anything with that name. That means that if we had a function named `array` in `dask`, we wouldn't get the module but the function.

max-sixty:
Thanks @keewis . Interesting re the difference in import functionality; I didn't know that.

I'm fine either way; to the extent the tests pass on the revised format, I'd be fine moving to that.

keewis:
The failing tests are #4225 and #4226, so this should be ready for review and merge.

I also updated the `flake8` pre-commit hook since `flake8<3.8` seems to not like `mypy` overloads.

max-sixty:
Great! Thanks @keewis !

PR #4206, opened 2020-07-07
pin isort

keewis:
Since the new versions of `isort` want to reformat `import dask.array as da` styled imports, this pins the `isort` version for now.

We can undo this in #4204 and – if we decide they're harmless – also apply the changes `isort` suggests.

 - [ ] Tests added


dcherian:
Thanks @keewis

Issue #4212, opened 2020-07-09
Add cupy support


jacobtomlinson:
I'm intending on working on cupy support in xarray along with @quasiben. Thanks for the warm welcome in the xarray dev meeting yesterday!

I'd like to use this issue to track cupy support and discuss certain design decisions. I appreciate there are also issues such as #4208, #3484 and #3232 which are related to cupy support, but maybe this could represent an umbrella issue for cupy specifically.

The main goal here is to improve support for array types other than numpy and dask in general. However, it is likely there will need to be some cupy specific compatibility code in xarray. (@andersy005 raised issues with calling `__array__` on cupy in #3232 for example).

I would love to hear from folks wanting to use cupy with xarray to help build up some use cases for us to develop against. We have some ideas but more are welcome.

My first steps here will be to add some tests which use cupy. These will skip in the main CI but we will also look at running xarray tests on some GPU CI too as we develop. A few limited experiments that I've run seem to work, so I'll start with tests which reproduce those.

andersy005:
@jacobtomlinson, thank you for getting this started. I'll be monitoring closely this issue. Let me know if I can help in any way.

jacobtomlinson:
This PR for adding pint support is a useful reference. https://github.com/pydata/xarray/pull/3238

dopplershift:
@jacobtomlinson Any idea how this would play with the work that's been going on for units here; I'm specifically wondering if xarray ( pint ( cupy )) would/could work.

jthielen:
> @jacobtomlinson Any idea how this would play with the work that's been going on for units here; I'm specifically wondering if xarray ( pint ( cupy )) would/could work.

As far as I'd see it, the pieces to get this working are

- xarray( pint ) [practicly done, just working out corner cases]
- pint( cupy ) [https://github.com/hgrecco/pint/issues/964]
- xarray( cupy ) [this PR]

and then finally testing xarray( pint( cupy )) works automatically from there. https://github.com/hgrecco/pint/issues/964 was deferred due to CI/testing concerns, so it will be great to see what @jacobtomlinson can come up with here for xarray, since hopefully at some point it would be transferable over to pint as well.

jacobtomlinson:
_I've written this comment a few times to try and not come across as confrontational. I'm not intending to be at all, so please don't take it that way 😅. Tone is hard in comments! I'm just trying to figure out how to proceed quickly._

[...]

PR #4214, opened 2020-07-10
Add initial cupy tests

jacobtomlinson:
Added some initial unit tests for cupy. Mainly to create a place for cupy tests to go and to check some basic functionality.

I've created a fixture which constructs the dataset from the Toy weather data example and converts the underlying arrays to cupy. Then I've added a test which checks that after calling operations such as `mean` and `groupby` the resulting DataArray is still backed by a cupy array.

The main penalty with working on GPUs is accidentally shunting data back and forth between the GPU and system memory. Copying data over the PCI bus is slow compared to the rest of the work so should be avoided. So this first test is checking that we are leaving things on the GPU.

Because this data copying is so expensive cupy have intentionally broken the `__array__` method and introduced a `.get` method instead. This means that users have to be explicit in converting back to numpy and copying back to the main memory. Therefore we will need to add some logic to xarray to use `.get` in appropriate situations such as plotting.

 - [x] Releated to #4212 
 - [x] Tests added

### DIFF
--- a/xarray/tests/test_cupy.py
+++ b/xarray/tests/test_cupy.py
@@ -1,7 +1,6 @@
-import pytest
-
 import numpy as np
 import pandas as pd
+import pytest
 
 import xarray as xr
 
--- a/xarray/tests/test_cupy.py
+++ b/xarray/tests/test_cupy.py
@@ -1,0 +1,41 @@
+import pytest
+
+import numpy as np
+import pandas as pd
+
+import xarray as xr
+
+cp = pytest.importorskip("cupy")
+
+
+@pytest.fixture
+def toy_weather_data():
+    np.random.seed(123)
+    times = pd.date_range("2000-01-01", "2001-12-31", name="time")
+    annual_cycle = np.sin(2 * np.pi * (times.dayofyear.values / 365.25 - 0.28))
+
+    base = 10 + 15 * annual_cycle.reshape(-1, 1)
+    tmin_values = base + 3 * np.random.randn(annual_cycle.size, 3)
+    tmax_values = base + 10 + 3 * np.random.randn(annual_cycle.size, 3)
+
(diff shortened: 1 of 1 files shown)


TomNicholas:
Hi Jacob, would it make sense to add a new CI job that has cupy so that these tests actually get run automatically on future PRs? At the moment they're just going to get skipped, but if you're planning to work on this it might be a good idea to automate the cupy integration tests early on?

jacobtomlinson:
@TomNicholas I've been trying to find a way to do this. In order for the cupy tests to be run the CI machine will need an NVIDIA GPU. But setting up a GPU CI would be a burden on the xarray team both technically and financially. This is why I'm just skipping for now.

For projects over in [RAPIDS](https://github.com/rapidsai) we have have a private GPU CI. We aren't in a position to integrate this here but we may be able to run nightlies. I'll update when I have more info.

jacobtomlinson:
I think what we can offer at the moment is to run a nightly against the default branch on some GPU hardware. If the nightly fails we will open an issue and look into resolving it.

Not as neat as hooking into PRs like the rest of the CI though. Hopefully this will improve over time.

dcherian:
> run a nightly against the default branch on some GPU hardware. If the nightly fails we will open an issue and look into resolving it.

This is great! Thanks @jacobtomlinson

jacobtomlinson:
Ok great! Once this is merged we will get that set up on our end.

dcherian:
I'm merging so that @jacobtomlinson can keep moving forward. 

It isn't breaking anything too :)

jacobtomlinson:
Many thanks! We will start running nightlies from today.

jacobtomlinson:
Nightlies are running. Sorry we can't do more than this right now.

image

PR #4217, opened 2020-07-11
Add release summary, some touch-ups

max-sixty:
This PR:
- Proposes adding a Release Summary with a PR. This is something that maybe more people read than anything else we send, and yet it's the least reviewed text of anything I write; and I'm making guesses on what's important.
- Proposes pasting the release summary into the GH Release page; to the extent people follow xarray by following our releases, it's nice to have something there.
- Some touch-ups

### DIFF
--- a/HOW_TO_RELEASE.md
+++ b/HOW_TO_RELEASE.md
@@ -10,11 +10,6 @@ upstream        https://github.com/pydata/xarray (fetch)
 upstream        https://github.com/pydata/xarray (push)
 ```
 
- 0. Write a release summary: ~50 words describing the high level features. This
-    will be used in the release emails, tweets, GitHub release notes, etc. 
-    I (@max) find it helpful to use a PR given I'm not always familiar with all the
-    features of the release, and the summary is distributed widely.
-
  1. Ensure your master branch is synced to upstream:
      ```sh
      git pull upstream master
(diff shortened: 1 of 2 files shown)


HOW_TO_RELEASE.md
```diff
@@ -2,78 +2,90 @@
...
+$ git remote -v
+{...}
+upstream        https://github.com/pydata/xarray (fetch)
+upstream        https://github.com/pydata/xarray (push)
+```
+
+ 0. Write a release summary: ~50 words describing the high level features. This
+    will be used in the release emails, tweets, GitHub release notes, etc. 
+    I (@max) find it helpful to use a PR given I'm not always familiar with all the
+    features of the release, and the summary is distributed widely.
```
keewis, line 16: I think it definitely makes sense to get a review of the summary. Also, I'd remove this: there's no need to justify yourself for that suggestion :smile:

In any case, I'd like to suggest writing the summary as item 2, then opening a PR with the contributors, the summary and the `whats-new.rst` fixes (new item 4) as item 5 to get a chance for a review and once that is merged continue with the current item 4 (after syncing again).

So now we'd have:
1. sync branch
2. write summary
3. get the contributors (also: what should happen in the rare case that the one releasing is not included in that list?)
4. check `whats-new.rst` and the docs
5. open a PR to add items 2 to 4 to `whats-new.rst`
6. resync (should that be a new step or should that also be in item 5?)
7. continue with the current item 4

Does that make sense?

max-sixty:
Updated based on @keewis good suggestions. I haven't updated the numbers until we agree it's correct. (Maybe there's some plugin to do this? Lucky we don't have a 100 item list...)

dcherian:
Thanks @max-sixty

PR #4219, opened 2020-07-12
nd-rolling

### LINKED ISSUE #4196: Convolution operation
I would like to perform 2d convolution operation with dask-data stored in xarray, defining the kernel.

Related issue are:
https://github.com/pydata/xarray/issues/1142
https://github.com/pydata/xarray/issues/2010
https://github.com/pydata/xarray/issues/1323

Referring to Dask, they have a specific image-oriented API which allows for convolutions:
http://image.dask.org/en/latest/dask_image.ndfilters.html

And this guy already implemented convolution operation based on xarray and dask:
https://github.com/serazing
https://github.com/serazing/xscale
https://xscale.readthedocs.io/en/latest/api.html

Are there plans to implement this feature?
In the meanwhile I'll try dask-image and xscale.

fujiisoup:
- [x] Closes #4196
 - [x] Tests added

I noticed that the implementation of nd-rolling is straightforward.
The core part is implemented but I am wondering what the best API is, with keeping it backward-compatible.

Obviously, it is basically should look like
```python
da.rolling(x=3, y=3).mean()
```

A problem is other parameters, `centers` and `min_periods`.
In principle, they can depend on dimension. 
For example, we can have `center=True` only for `x` but not for `y`.

So, maybe we allow dictionary for them?
```python
da.rolling(x=3, y=3, center={'x': True, 'y': False}, min_periods={'x': 1, 'y': None}).mean()
```

The same thing happens for `.construct` method.
```python
da.rolling(x=3, y=3).construct(x='x_window', y='y_window', stride={'x': 2, 'y': 1})
```
I'm afraid if this dictionary argument was a bit too redundant.

Does anyone have another idea?

### DIFF
--- a/xarray/core/dask_array_ops.py
+++ b/xarray/core/dask_array_ops.py
@@ -32,10 +32,10 @@ def rolling_window(a, axis, window, center, fill_value):
     """
     import dask.array as da
 
-    # for nd-rolling. 
+    # for nd-rolling.
     # TODO It can be more efficient. Currently, the chunks at the boundaries
     # will be copied, but it might be OK for many-chunked-arrays.
-    if hasattr(axis, '__len__'):
+    if hasattr(axis, "__len__"):
         for ax, win, cen in zip(axis, window, center):
             a = rolling_window(a, ax, win, cen, fill_value)
         return a
--- a/xarray/core/dask_array_ops.py
+++ b/xarray/core/dask_array_ops.py
@@ -32,6 +32,14 @@ def rolling_window(a, axis, window, center, fill_value):
     """
     import dask.array as da
 
+    # for nd-rolling. 
+    # TODO It can be more efficient. Currently, the chunks at the boundaries
+    # will be copied, but it might be OK for many-chunked-arrays.
+    if hasattr(axis, '__len__'):
+        for ax, win, cen in zip(axis, window, center):
+            a = rolling_window(a, ax, win, cen, fill_value)
+        return a
+
     orig_shape = a.shape
     if axis < 0:
         axis = a.ndim + axis
--- a/xarray/core/dask_array_ops.py
+++ b/xarray/core/dask_array_ops.py
@@ -86,7 +86,7 @@ def rolling_window(a, axis, window, center, fill_value):
 
     # create overlap arrays
     ag = da.overlap.overlap(a, depth=depth, boundary=boundary)
-    
+
     def func(x, window, axis):
         x = np.asarray(x)
         index = [slice(None)] * x.ndim
--- a/xarray/core/dask_array_ops.py
+++ b/xarray/core/dask_array_ops.py
@@ -86,7 +86,7 @@ def rolling_window(a, axis, window, center, fill_value):
 
     # create overlap arrays
     ag = da.overlap.overlap(a, depth=depth, boundary=boundary)
-    # apply rolling func
+    
     def func(x, window, axis):
         x = np.asarray(x)
         index = [slice(None)] * x.ndim
--- a/xarray/core/dask_array_ops.py
+++ b/xarray/core/dask_array_ops.py
@@ -32,77 +32,80 @@ def rolling_window(a, axis, window, center, fill_value):
     """
     import dask.array as da
 
-    # for nd-rolling.
-    # TODO It can be more efficient. Currently, the chunks at the boundaries
-    # will be copied, but it might be OK for many-chunked-arrays.
-    if hasattr(axis, "__len__"):
-        for ax, win, cen in zip(axis, window, center):
-            a = rolling_window(a, ax, win, cen, fill_value)
-        return a
+    if not hasattr(axis, "__len__"):
+        axis = [axis]
+        window = [window]
+        center = [center]
 
     orig_shape = a.shape
-    if axis < 0:
-        axis = a.ndim + axis
     depth = {d: 0 for d in range(a.ndim)}
-    depth[axis] = int(window / 2)
-    # For evenly sized window, we need to crop the first point of each block.
-    offset = 1 if window % 2 == 0 else 0
+    offset = [0] * a.ndim
+    drop_size = [0] * a.ndim
+    pad_size = [0] * a.ndim
+    for ax, win, cent in zip(axis, window, center):
+        if ax < 0:
+            ax = a.ndim + ax
+        depth[ax] = int(win / 2)
+        # For evenly sized window, we need to crop the first point of each block.
+        offset[ax] = 1 if win % 2 == 0 else 0
 
-    if depth[axis] > min(a.chunks[axis]):
-        raise ValueError(
-            "For window size %d, every chunk should be larger than %d, "
-            "but the smallest chunk size is %d. Rechunk your array\n"
-            "with a larger chunk size or a chunk size that\n"
-            "more evenly divides the shape of your array."
-            % (window, depth[axis], min(a.chunks[axis]))
-        )
+        if depth[ax] > min(a.chunks[ax]):
+            raise ValueError(
+                "For window size %d, every chunk should be larger than %d, "
+                "but the smallest chunk size is %d. Rechunk your array\n"
+                "with a larger chunk size or a chunk size that\n"
+                "more evenly divides the shape of your array."
+                % (win, depth[ax], min(a.chunks[ax]))
+            )
 
-    # Although da.overlap pads values to boundaries of the array,
-    # the size of the generated array is smaller than what we want
-    # if center == False.
-    if center:
-        start = int(window / 2)  # 10 -> 5,  9 -> 4
-        end = window - 1 - start
-    else:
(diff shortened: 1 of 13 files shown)


xarray/tests/test_dataarray.py
```diff
@@ -6399,6 +6397,19 @@ def test_rolling_count_correct():
         assert_equal(result, expected)
 
 
+@pytest.mark.parametrize("da", (1,), indirect=True)
+@pytest.mark.parametrize("center", (True, False))
+@pytest.mark.parametrize("min_periods", (None, 1))
+def test_ndrolling_reduce(da, center, min_periods):
+    rolling_obj = da.rolling(time=3, a=2, center=center, min_periods=min_periods)
```
fujiisoup, line 6404: This is a basic test for the new nd-rolling

max-sixty:
This looks very promising; I'm surprised it was possible without more code. I'm being slow, but where is the nd-rolling algo? I had thought bottleneck didn't support more than one dimension? https://bottleneck.readthedocs.io/en/latest/bottleneck.move.html, and that we'd have to implement our own in numbagg (which would be very possible)

max-sixty:
Re the API, I think the dict is probably the best option, although it does complicate as the arguments become differently typed depending on one vs multiple dimensions.

One alternative is to allow fluent args, like:
```
(
    da
    .rolling(x=3, center=True)
    .rolling(y=5, min_periods=2)
    .mean()
)
```
...but does that then seem like the second `rolling` is operating on the _result_ of the first?

fujiisoup:
Hi @max-sixty 

> One alternative is to allow fluent args, like:
> ...but does that then seem like the second rolling is operating on the result of the first?

I couldn't think of it until just now. But yes, it sounds to me like a repeated rolling operation.

> I'm being slow, but where is the nd-rolling algo? I had thought bottleneck didn't support more than one dimension? 

No. With nd-rolling, we need to use numpy reductions.
Its `skipna=True` operation is currently slow, but it can be improved replacing nan *before* the stride-trick.

fujiisoup:
Another API concern. We now use `min_periods`, in which we implicitly assume one-dimension cases.

With nd-dimension, I think `min_counts` argument is more appropriate like bottleneck, which will limit the lower bound of the number of missing entries in the n-dimensional window.

Even if we leave it, we may disallow nd-argument of `min_periods` but keep it a scalar.

xarray/core/dask_array_ops.py
```diff
@@ -32,6 +32,14 @@ def rolling_window(a, axis, window, center, fill_value):
     """
     import dask.array as da
 
+    # for nd-rolling.
+    # TODO It can be more efficient. Currently, the chunks at the boundaries
+    # will be copied, but it might be OK for many-chunked-arrays.
+    if hasattr(axis, "__len__"):
+        for ax, win, cen in zip(axis, window, center):
+            a = rolling_window(a, ax, win, cen, fill_value)
+        return a
```
fujiisoup, line 41: I hope it is true, though I didn't check it.
A suspicious part is `da.concatenate`, which I expect does not copy the original (strided-)array.

dcherian:
I think the dictionary is OK. We can allow scalars for `center`, `stride` etc. in which case the scalar applies to all dimensions. This would preserve backward compatibility.

scalar for `min_counts` also seems OK to me. We can update it if someone requests it.

fujiisoup:
I think now it is ready for review, though I'm sure tests miss a lot of edge cases.
Maybe we can fix them if pointed out.

fujiisoup:
A possible improvement will be nan-reduction methods for nd-rolling.
Currently, we just use numpy nan-reductions, which is memory consuming for strided arrays.

This issue can be solved by replacing nan by appropriate values and applying nonnan-reduction methods, 
e.g.,
```python
da.rolling(x=2, y=3).construct(x='xw', y='yw').sum(['xw', 'yw'])
```
should be the same with
```python
da.rolling(x=2, y=3).construct(x='xw', y='yw', fill_value=0).sum(['xw', 'yw'], skipna=False)
```
and the latter is much more memory efficient.

I'd like to leave this improvement to future PR.

fujiisoup:
I got an error for typechecking, only in CI but not in local, from the code that I didn't change.

fujiisoup:
Could anyone kindly review this?

xarray/tests/test_dataset.py
```diff
@@ -6007,6 +6005,48 @@ def test_rolling_reduce(ds, center, min_periods, window, name):
         assert src_var.dims == actual[key].dims
 
 
+@pytest.mark.parametrize("ds", (1,), indirect=True)
+@pytest.mark.parametrize("center", (True, False))
+@pytest.mark.parametrize("min_periods", (None, 1))
+@pytest.mark.parametrize("name", ("sum", "mean", "max"))
```
max-sixty, line 6011: Will this test will fail for something like `std`? 
IIUC iteratively applying the function to the array for each dim only works for some functions.

Which is probably fine, no need to write more tests imo, though maybe worth making a note

fujiisoup:
Thanks @max-sixty for the review ;)
I'll work for the update in a few days.

xarray/core/dask_array_ops.py
fujiisoup: I noticed that ghosting breaks my expectation. I need to update the algo here.

xarray/tests/test_dataset.py
fujiisoup: In this case, many other functions should have the same result.
If there are some nans in random positions, the result should be different.
I added a few more tests and include function `std`, but I don't think it is necessary to test all nanops.

xarray/core/rolling.py
```diff
@@ -251,13 +272,27 @@ def construct(self, window_dim, stride=1, fill_value=dtypes.NA):
 
         from .dataarray import DataArray
 
+        if window_dim is None:
+            if len(window_dim_kwargs) == 0:
+                raise ValueError(
+                    "Either window_dim or window_dim_kwargs need to be specified."
```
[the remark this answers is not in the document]
fujiisoup: For the backward-compatibility, we need to support a `str` for window_dim for 1d-rolling.
`either_dict_or_kwargs` does not support this situation.

xarray/tests/test_dataset.py
```diff
@@ -6008,15 +6008,25 @@ def test_rolling_reduce(ds, center, min_periods, window, name):
...
-            ds.rolling(time=3, center=center, min_periods=min_periods), name
-        )().rolling(x=2, center=center, min_periods=min_periods),
+            ds.rolling(time=4, center=center, min_periods=min_periods), name
+        )().rolling(x=3, center=center, min_periods=min_periods),
+        name,
+    )()
+    assert_allclose(actual, expected)
+    assert actual.dims == expected.dims
+
+    # Do it in the oposite order
```
max-sixty, line 6025:
```suggestion
    # Do it in the opposite order
```

max-sixty:
I'm still a bit confused. How does the "roll over each dimension in turn" approach equal the "roll over both dimension together" approach with a function like std? Here's a proposed counter example:

```python
import xarray as xr
import numpy as np

da = xr.DataArray(np.asarray([[0,10,0],[0,10,0], [0,10,0]]), dims=list('xy'))
print(da)

<xarray.DataArray (x: 3, y: 3)>
array([[ 0, 10,  0],
       [ 0, 10,  0],
       [ 0, 10,  0]])

x_std = da.rolling(dict(x=2)).std()
print(x_std)

<xarray.DataArray (x: 3, y: 3)>
array([[nan, nan, nan],
       [ 0.,  0.,  0.],
       [ 0.,  0.,  0.]])

x_then_y_std = x_std.rolling(dict(y=2)).std()
print(x_then_y_std)

<xarray.DataArray (x: 3, y: 3)>
array([[nan, nan, nan],
       [nan,  0.,  0.],
       [nan,  0.,  0.]])

combined_std = da.rolling(dict(x=2, y=2)).std()
print(combined_std)

<xarray.DataArray (x: 3, y: 3)>
array([[nan, nan, nan],
       [nan,  5.,  5.],
       [nan,  5.,  5.]])

```

dcherian:
I think it's more like
```
da.rolling(x=2).construct("xr").rolling(y=2).construct("yr").std(["xr", "yr"])
```

but that gives
```
array([[0., 5., 5.],
       [0., 5., 5.],
       [0., 5., 5.]])
```

fujiisoup:
Thanks @max-sixty .

You are completely correct.
As the test pass, I was fooling myself. 

The reason was that the dataset I was using for the test does not have `time` and `x` simultaneously.
So I was not testing the 2d-rolling but just 1d-rolling.

Fixed. Now it correctly fails for `mean` and `std`, but passes with `max` and `sum`

max-sixty:
Great! Thanks @fujiisoup ! 

Ready to go from my POV

fujiisoup:
@max-sixty 
thanks for the review.
merged

PR #4221, opened 2020-07-13
Change isinstance checks to duck Dask Array checks #4208

### LINKED ISSUE #4208: Support for duck Dask Arrays
https://github.com/pydata/xarray/issues/525#issuecomment-531603357 raised the idea of adding "duck Dask Array" support to xarray as a way to handle xarray > Pint Quantity > Dask Array wrapping in a way that still allowed for most of xarray's Dask integration to work properly. With @rpmanser working on implementing the Dask collection interface in Pint (https://github.com/hgrecco/pint/pull/1129), I thought it best to elevate this to its own issue to track progress and discuss implementation on xarray's side (since hopefully @rpmanser or I can get started on it soon).

Two initial (and intertwined) discussion points that I'd like to bring up (xref https://github.com/hgrecco/pint/pull/1129#issuecomment-655197079):

- How should xarray check for a duck Dask Array?
- Is it acceptable for a Pint Quantity to always have the Dask collection interface defined (i.e., be a duck Dask array), even when its magnitude (what it wraps) is not a Dask Array?

cc @keewis, @shoyer, @crusaderky

rpmanser:
- [X] Closes #4208
 - [x] Tests added

I added the discussed `is_duck_dask_array(x)` function and replaced `isinstance(x, dask_array_type)` checks with `is_duck_dask_array(x)`. Existing tests are passing. I did not change `isinstance(x, dask_array_type)` checks in the testing modules since I am unsure if that is appropriate.

As for additional tests, I am leaning towards testing Xarray ( Pint ( Dask ) ) ) objects as implemented in upcoming Pint v0.15 rather than a mock class, but I am open to discussion. I am unfamiliar with the release schedules of Pint and Xarray but I imagine that would influence which direction the testing should go.


xarray/core/nanops.py
```diff
@@ -77,15 +77,15 @@ def nanmin(a, axis=None, out=None):
     if a.dtype.kind == "O":
         return _nan_minmax_object("min", dtypes.get_pos_infinity(a.dtype), a, axis)
 
-    module = dask_array if isinstance(a, dask_array_type) else nputils
+    module = dask_array if is_duck_dask_array(a) else nputils
```
jthielen, line 80: These nanops will likely require some careful testing, since I'm not sure if using the dask.array module or xarray's nputils is better for non-Dask duck Dask Arrays.

xarray/core/rolling_exp.py
```diff
@@ -13,7 +13,7 @@ def _get_alpha(com=None, span=None, halflife=None, alpha=None):
 
 
 def move_exp_nanmean(array, *, axis, alpha):
-    if isinstance(array, dask_array_type):
+    if is_duck_dask_array(array):
         raise TypeError("rolling_exp is not currently support for dask arrays")
```
jthielen, line 17:
```suggestion
        raise TypeError("rolling_exp is not currently support for dask-like arrays")
```

xarray/core/variable.py
```diff
@@ -341,9 +342,7 @@ def _in_memory(self):
 
     @property
     def data(self):
-        if hasattr(self._data, "__array_function__") or isinstance(
-            self._data, dask_array_type
-        ):
+        if hasattr(self._data, "__array_function__") or is_duck_dask_array(self._data):
```
jthielen, line 345: Could this also become just `is_array_like`?

xarray/core/variable.py
```diff
@@ -885,9 +884,7 @@ def copy(self, deep=True, data=None):
                 data = indexing.MemoryCachedArray(data.array)
 
             if deep:
-                if hasattr(data, "__array_function__") or isinstance(
-                    data, dask_array_type
-                ):
+                if hasattr(data, "__array_function__") or is_duck_dask_array(data):
```
jthielen, line 887: Likewise just `is_array_like`?

jthielen:
Also, a broader discussion that's I've seen hinted to in the past, but is brought to the forefront by this PR: how should xarray be checking for general duck array types it can wrap?

Right now, it looks like there is a mix of 

```python
is_array_like(data)
```
and
```python
hasattr(data, "__array_function__") or isinstance(data, dask_array_type)
```

and it would be nice to bring consistency. However, both these checks seem like they'd let too many types through. For example, `is_array_like` as currently implemented would still let through too much, such as `MemoryCachedArray` and xarray Variables/DataArrays, and `__array_function__` doesn't truly indicate a NumPy duck array on its own (and it really only works to only capture downcast types right now since xarray does not yet implement `__array_function__` itself, #3917).

keewis:
Revisiting it I now realize that `is_array_like` is actually not a good name for that function: officially, "array_like" means that it can be casted using `np.asarray`. Maybe we should rename that to `is_duck_array`? But then we'd also need to check for `__array_ufunc__` and `__array_function__` because we also need those to properly wrap duck arrays.

jthielen:
> Revisiting it I now realize that `is_array_like` is actually not a good name for that function: officially, "array_like" means that it can be casted using `np.asarray`. Maybe we should rename that to `is_duck_array`? But then we'd also need to check for `__array_ufunc__` and `__array_function__` because we also need those to properly wrap duck arrays.

As long as that doesn't break any of the current uses, I think that would be the best way forwards. This would require xarray to be on NumPy 1.16+ (in order to ensure `hasattr(np.ndarray, "__array_function__")`, but that's only 9 days away by NEP 29, so hopefully that isn't too much of an issue. Then, most of the checks for "can this be wrapped by xarray" should be able to become something like

```python
is_duck_array(x) and not isinstance(x, (DataArray, Variable)),
```
with `is_duck_dask_array` only being needed for checking Dask Array-like support. 

Although, while we're at it, do we also need more careful handling of upcast types? I'm not sure if there even are any out there right now (not sure if a HoloViews Dataset counts here), but that doesn't necessarily mean there never will be any.

xarray/core/dask_array_compat.py
```diff
@@ -130,6 +131,10 @@ def _validate_pad_output_shape(input_shape, pad_width, output_shape):
         )
 
 
+def is_duck_dask_array(x):
+    return is_array_like(x) and isinstance(x, dask_array_type)
```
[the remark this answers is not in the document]
rpmanser: Yes, it should be `is_dask_collection(x)`. It would be okay for `is_dask_collection` to be defined as `None` upon an `ImportError` exception, right?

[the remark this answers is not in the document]
keewis: I'd make that either `lambda _: True` or `lambda _: False` (not sure which)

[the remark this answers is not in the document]
jthielen: Since `x` cannot be a Dask collection if Dask is not importable, I think `is_dask_collection` should be
```python
def is_dask_collection(x):
    return False
```
when the `ImportError` occurs.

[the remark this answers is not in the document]
rpmanser: It feels a bit awkward to put a function definition in a try/except, so I think `lambda _: False` works better.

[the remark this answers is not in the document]
jthielen: > It feels a bit awkward to put a function definition in a try/except, so I think `lambda _: False` works better.

I didn't think
```
is_dask_collection = lambda _: False
```
was allowed in flake8, hence the suggestion on regular function definition. But, if you can get away with it, then sure!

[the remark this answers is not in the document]
rpmanser: Flake8 allows it.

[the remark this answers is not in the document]
jthielen: > Flake8 allows it.

Good to know!

rpmanser:
I'm getting a bit lost within the testing suite while trying to figure out the implications of this change and what kinds of tests are necessary for it. Would a test module for this
- Individually test all functions that include the newly implemented `is_duck_dask_array()`?
- Test that a duck dask Array has attributes of a dask Array and redirects the appropriate methods to those defined by `Variable`, `DataArray`, and `Dataset`?
- Be similar to `test_dask.py`?
- All, some, or none of the above?

Any suggestions would be greatly appreciated.

jthielen:
@rpmanser For what it's worth, I'd think doing tests like `test_dask.py` but with a post-https://github.com/hgrecco/pint/pull/1129 Pint Quantity or suitably mocked wrapper class between xarray and Dask would be best. But, I think this is more of a maintainer-level question.

dcherian:
Thanks for working on this @rpmanser.

Since you've changed every `isinstance` check, if `test_dask` passes, you're doing fine in general.

> Test that a duck dask Array has attributes of a dask Array and redirects the appropriate methods to those defined by Variable, DataArray, and Dataset?

Seems like this is covered by `test_dask` already but I could be wrong.

We could add specific pint-as-dask tests to `test_units` with appropriate xfails (they should run on the `upstream-dev` test config).

This could be a discussion topic for our next dev meeting on wednesday.

rpmanser:
I have erred on the side of leaving the `isinstance` checks in place where @jthielen reviewed them. If it is decided that no further tests are needed, then this should be ready for review.

On another note, if it is appropriate to apply these changes:

> > Revisiting it I now realize that `is_array_like` is actually not a good name for that function: officially, "array_like" means that it can be casted using `np.asarray`. Maybe we should rename that to `is_duck_array`? But then we'd also need to check for `__array_ufunc__` and `__array_function__` because we also need those to properly wrap duck arrays.
> 
> As long as that doesn't break any of the current uses, I think that would be the best way forwards. This would require xarray to be on NumPy 1.16+ (in order to ensure `hasattr(np.ndarray, "__array_function__")`, but that's only 9 days away by NEP 29, so hopefully that isn't too much of an issue. Then, most of the checks for "can this be wrapped by xarray" should be able to become something like
> 
> ```python
> is_duck_array(x) and not isinstance(x, (DataArray, Variable))
> ```
> 
> with `is_duck_dask_array` only being needed for checking Dask Array-like support.

I can do that as well. If not, should a new issue be opened for this?

xarray/backends/common.py
```diff
@@ -177,7 +177,7 @@ def __init__(self, lock=None):
         self.lock = lock
 
     def add(self, source, target, region=None):
-        if isinstance(source, dask_array_type):
```
dcherian, line 180: Not sure about this one. cc @jhamman

xarray/backends/common.py
```diff
@@ -8,7 +8,7 @@
 
 from ..conventions import cf_encoder
 from ..core import indexing
-from ..core.pycompat import dask_array_type
+from ..core.dask_array_compat import is_duck_dask_array
```
dcherian, line 11: let's keep this in pycompat as earlier. `dask_array_compat` contains backports of newer dask functions and it has been deleted in the past (!).

xarray/core/duck_array_ops.py
```diff
@@ -161,7 +161,7 @@ def trapz(y, x, axis):
 def asarray(data):
     return (
         data
-        if (isinstance(data, dask_array_type) or hasattr(data, "__array_function__"))
+        if (is_duck_dask_array(data) or hasattr(data, "__array_function__"))
```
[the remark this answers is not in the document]
dcherian: I *think* this is right

xarray/core/duck_array_ops.py
```diff
@@ -190,11 +191,7 @@ def lazy_array_equiv(arr1, arr2):
     arr2 = asarray(arr2)
     if arr1.shape != arr2.shape:
         return False
-    if (
-        dask_array
-        and isinstance(arr1, dask_array_type)
-        and isinstance(arr2, dask_array_type)
-    ):
+    if dask_array and is_duck_dask_array(arr1) and is_duck_dask_array(arr2):
```
dcherian, line 194: Do duck dask arrays have a unique token saved in `name`? Doesn't seem so in https://github.com/hgrecco/pint/pull/1129/files

xarray/core/nanops.py
```diff
@@ -1,8 +1,8 @@
 import numpy as np
 
 from . import dtypes, nputils, utils
-from .duck_array_ops import _dask_or_eager_func, count, fillna, isnull, where_method
 from .pycompat import dask_array_type
+from .duck_array_ops import _dask_or_eager_func, count, fillna, isnull, where_method
```
dcherian, line 5: lets take this out. pretty sure isort doesn't like it.

xarray/core/variable.py
```diff
@@ -977,7 +974,7 @@ def chunk(self, chunks=None, name=None, lock=False):
             chunks = self.chunks or self.shape
 
         data = self._data
-        if isinstance(data, da.Array):
+        if is_duck_dask_array(data):
             data = data.rechunk(chunks)
```
dcherian, line 978: is `pint` forwarding `.rechunk` to `dask.array`? Similarly below in `_shift_one_dim`

xarray/core/duck_array_ops.py
rpmanser: No, they don't. Dask recommends implementing deterministic hashing with `__dask_tokenize__` when [defining custom collections](https://docs.dask.org/en/latest/custom-collections.html#deterministic-hashing), but I refrained from doing so. Pint seems to use tokens, but I'm not familiar with how they work or how they would interact with Dask. Could this cause problems?

xarray/core/variable.py
rpmanser: Yes, but not explicitly. Pint checks the wrapped dask array for the `rechunk` method and finds it, which (seems to) correctly call `dask.array.Array.rechunk`.

xarray/core/duck_array_ops.py
dcherian: Yes the `arr1.name == arr2.name` check depends on the `.name` property having a unique token. Then we can test equality cheaply. This is necessary because xarray likes to compare "non-dimensional coordinate variables" for equality in many places. This can lead to major slowdowns if we can't do this check cheaply.

One options is:
1. update this check to use `tokenize` instead of the `name` property for duck dask arrays.
2. for actual dask arrays we could use `.name` as an optimization, instead of calculating the token

dcherian: Also it would be a good idea to define `__dask_tokenize__` in pint. It helps prevent unnecessary duplicate computations.

> Pint seems to use tokens,

What are these tokens?

xarray/core/variable.py
dcherian: :+1:

xarray/core/duck_array_ops.py
rpmanser: > Also it would be a good idea to define `__dask_tokenize__` in pint. It helps prevent unnecessary duplicate computations.

Okay, I can open an issue and work on that.

> What are these tokens?

They're from the `token` library and if I understand correctly they're used to resolve unit operations. I think this is different than how Dask uses tokens and `.name`. If someone knows more about this please feel free to correct me.

xarray/core/duck_array_ops.py
```diff
@@ -190,11 +191,7 @@ def lazy_array_equiv(arr1, arr2):
...
         # GH3068
         if arr1.name == arr2.name:
```
dcherian, line 196:
```suggestion
        if dask.base.tokenize(arr1) == dask.base.tokenize(arr2):
```

This should work as long as `pint` implements `__dask_tokenize__`

It seems like `return (dask_array.name, self.units)` or something would work.

xarray/core/duck_array_ops.py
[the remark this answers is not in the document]
shoyer: I think we should define another helper function, `is_duck_array()`, to use in situations like this:
```python
def is_duck_array(data):
    # no need to check for dask explicitly now that it implements __array_function__
    # depending on backwards compat, potentially we still need to explicitly include
    # np.ndarray here?
    return hasattr(data, "__array_function__")
```

The reason for `is_duck_array()` rather than `is_array_like()` is that a "duck array" is a strong condition than "array like", at least as [is commonly used by NumPy](https://numpy.org/neps/nep-0022-ndarray-duck-typing-overview.html):

- "Array like"  means "convertible to a NumPy array", i.e., with `np.asarray()`.
- "Duck array" means "can be computed on like a NumPy array".

Examples:

- Python lists, zarr.Array and netCDF4.Variable are "array like" but not "duck arrays"
- CuPy and sparse arrays are "duck arrays" but not "array like" (`np.asarray` fails)
- NumPy, dask and pint arrays are both "array like" _and_ "duck arrays"

It is valid to pass either a "duck array" or an "array like" into an xarray data structure, but `.data` should always be a duck array.

(`._data` can sometimes include "arrays likes", especially xarray's own internal "lazy arrays" that support lazy indexing. This is worth cleaning up eventually, but at least it isn't exposed to users who use `.data`.)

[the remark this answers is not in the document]
dcherian: @rpmaner can you make this change too please?

xarray/core/variable.py
dcherian: yes i think we want the proposed `is_duck_array`

xarray/tests/test_units.py
```diff
@@ -2231,6 +2237,21 @@ def test_pad_unit_constant_value(self, unit, error, dtype):
...
+
+        array = dask.array.linspace(0, 5, 3 * 10).reshape(3, 10).astype(dtype)
+        q = unit_registry.Quantity(array, unit_registry.m)
+        variable = xr.Variable(("x", "y"), q)
+
+        token = dask.base.tokenize(variable)
+        post_op = variable + 5 * unit_registry.m
+
+        assert dask.base.tokenize(variable) != dask.base.tokenize(post_op)
+        assert dask.base.tokenize(variable) == token
```
dcherian, line 2253: this last line seems unnecessary.

xarray/core/duck_array_ops.py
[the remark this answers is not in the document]
rpmanser: @dcherian Yes, I can do this as well. Should *all* `hasattr(data, "__array_function__")` checks be replaced with the above `is_duck_array`, or just in the places discussed in this PR? And should all `is_duck_dask_array` checks also be replaced with `is_duck_array` (assuming dask has implemented `__array_function__`)?

xarray/tests/test_units.py
rpmanser: Yes, it seems redundant, but as I was working on [hgrecco/pint#1151](https://github.com/hgrecco/pint/pull/1151) and testing this I found that if `__dask_tokenize__` is not implemented by the object passed to `dask.base.tokenize` the token for the original object changes after an operation. Many tests in pint have similar immutability checks.

xarray/core/duck_array_ops.py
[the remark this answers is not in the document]
keewis: just to make sure, I think we should rename `is_array_like` to `is_duck_array`, and also have it check for `__array_ufunc__` and `__array_function__`. Not sure if I'm missing anything, though.

`is_duck_dask_array` should be fine as is?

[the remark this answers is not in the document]
rpmanser: > just to make sure, I think we should rename `is_array_like` to `is_duck_array`, and also have it check for `__array_ufunc__` and `__array_function__`. Not sure if I'm missing anything, though.

Makes sense to me.

> `is_duck_dask_array` should be fine as is?

You're right, `is_duck_dask_array` should be okay since it checks if the item is a dask collection.

xarray/tests/test_units.py
rpmanser: I can confirm that the original token does indeed change after an operation is performed if `__dask_tokenize__` is not implemented, so I left the immutability assertion in place.

xarray/core/variable.py
```diff
@@ -933,9 +937,7 @@ def copy(self, deep=True, data=None):
                 data = indexing.MemoryCachedArray(data.array)
 
             if deep:
-                if hasattr(data, "__array_function__") or isinstance(
-                    data, dask_array_type
-                ):
+                if is_duck_array(data):
```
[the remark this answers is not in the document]
rpmanser: Since dask arrays are duck arrays I don't see a need to explicitly check for dask here, unless I'm overlooking some details. Additionally, I believe this was discussed earlier and you suggested

> yes i think we want the proposed `is_duck_array`

So it should be okay with `is_duck_array`, right?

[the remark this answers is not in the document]
keewis: I think the old check was from before `dask` had `__array_function__` (or because we had be backwards compatible with `dask` versions that didn't define `__array_function__`), so this should be fine.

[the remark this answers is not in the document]
dcherian: >  I believe this was discussed earlier and you suggested

haha sorry! thanks for clarifying

xarray/core/dask_array_compat.py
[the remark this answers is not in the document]
keewis: that's because we've configured `flake8` to suppress E731, otherwise `flake8` would complain

xarray/convert.py
```diff
@@ -291,7 +291,7 @@ def from_iris(cube):
     cube_data = cube.core_data() if hasattr(cube, "core_data") else cube.data
 
     # Deal with dask and numpy masked arrays
-    if isinstance(cube_data, dask_array_type):
+    if is_duck_dask_array(cube_data):
```
keewis, line 294: is it possible to have `iris` cubes with `duck dask arrays`?

xarray/coding/variables.py
```diff
@@ -93,7 +93,7 @@ def lazy_elemwise_func(array, func, dtype):
     -------
     Either a dask.array.Array or _ElementwiseFunctionArray.
     """
-    if isinstance(array, dask_array_type):
+    if is_duck_dask_array(array):
         return array.map_blocks(func, dtype=dtype)
```
keewis, line 97: does every `duck dask array` define `map_blocks`, or should we use `dask.array.map_blocks(func, array, ...)` instead?

xarray/core/common.py
```diff
@@ -1510,7 +1510,7 @@ def _full_like_variable(other, fill_value, dtype: DTypeLike = None):
     if fill_value is dtypes.NA:
         fill_value = dtypes.get_fill_value(dtype if dtype is not None else other.dtype)
 
-    if isinstance(other.data, dask_array_type):
+    if is_duck_dask_array(other.data):
```
keewis, line 1513: does `dask.array.Array` define `numpy.full_like`? If it does, we could just remove this code path (doesn't have to be this PR, though).

xarray/core/dask_array_compat.py
```diff
@@ -39,7 +39,7 @@ def meta_from_array(x, ndim=None, dtype=None):
         """
         # If using x._meta, x must be a Dask Array, some libraries (e.g. zarr)
         # implement a _meta attribute that are incompatible with Dask Array._meta
-        if hasattr(x, "_meta") and isinstance(x, dask_array_type):
+        if hasattr(x, "_meta") and is_duck_dask_array(x):
```
keewis, line 42: is `zarr` a `duck_dask_array`? If it is, we might have to switch back to `isinstance(x, dask_array_type)`, see the comment directly above

xarray/core/duck_array_ops.py
```diff
@@ -129,7 +136,7 @@ def notnull(data):
 
 
 def gradient(x, coord, axis, edge_order):
-    if isinstance(x, dask_array_type):
+    if is_duck_dask_array(x):
         return dask_array.gradient(x, coord, axis=axis, edge_order=edge_order)
```
keewis, line 140: is this still necessary, or does `dask` define `numpy.gradient` via `__array_function__`?

xarray/core/formatting.py
```diff
@@ -457,8 +457,8 @@ def short_data_repr(array):
     internal_data = getattr(array, "variable", array)._data
     if isinstance(array, np.ndarray):
         return short_numpy_repr(array)
-    elif hasattr(internal_data, "__array_function__") or isinstance(
-        internal_data, dask_array_type
+    elif hasattr(internal_data, "__array_function__") or is_duck_dask_array(
+        internal_data
     ):
```
keewis, line 462: should that be
```suggestion
    elif is_duck_array(internal_data):
```
instead?

xarray/core/formatting.py
```diff
@@ -527,7 +527,7 @@ def diff_dim_summary(a, b):
 
 
 def _diff_mapping_repr(a_mapping, b_mapping, compat, title, summarizer, col_width=None):
-    def is_array_like(value):
+    def is_duck_array(value):
         return (
             hasattr(value, "ndim")
             and hasattr(value, "shape")
```
keewis, line 533: should we replace the local function with `utils.is_duck_array`?

keewis (commented):
looks good to me, but I have a few questions. I'm not sure if what I'm asking for makes sense, so if they don't feel free to ignore them.

xarray/core/formatting.py
```diff
@@ -457,8 +457,8 @@ def short_data_repr(array):
     internal_data = getattr(array, "variable", array)._data
     if isinstance(array, np.ndarray):
         return short_numpy_repr(array)
-    elif hasattr(internal_data, "__array_function__") or isinstance(
-        internal_data, dask_array_type
+    elif hasattr(internal_data, "__array_function__") or is_duck_dask_array(
+        internal_data
```
keewis, line 461: should that be
```suggestion
    elif is_duck_array(internal_data):
```
instead?

xarray/core/formatting.py
keewis, line 462: should that be
```suggestion
    elif is_duck_array(internal_data):
```
instead?

xarray/core/duck_array_ops.py
dcherian: Our minimum numpy version does not support NEP-18 so this (and all of duck_array_ops) is still necessary?

keewis: ah, right, I forgot about that.

xarray/core/dask_array_compat.py
dcherian: oooh nice catch. I think we can delete this function now that our minimum dask version is 2.9

xarray/core/common.py
keewis: then this also depends on `numpy>=1.16`

xarray/convert.py
rpmanser: I do not know, but it's a good point. If `duck dask arrays` are not specifically supported by `iris`, maybe this should be changed back to the `isinstance` check?

xarray/coding/variables.py
rpmanser: I *think* that methods like this are defined by `duck dask arrays` and I say that only based on `rechunk` being defined by them.

xarray/core/formatting.py
rpmanser: Seems like a good idea to me.

xarray/coding/variables.py
dcherian: It's probably safer to use `dask.array.map_blocks`

keewis: if we do that, it might be good to change all other instances of `array.map_blocks(func, ...)` to `map_blocks(func, array, ...)`, too. I think there are only 4 calls: in `xarray.core.dask_array_ops` and `xarray.core.dask_array_compat`, with different names for `array`.

This is not really related to this PR though, so you could also leave that change to a new PR.

dcherian: :+1:

xarray/core/dask_array_compat.py
rpmanser: > oooh nice catch. I think we can delete this function now that our minimum dask version is 2.9

Should I remove the entire `meta_from_array` function as part of this PR?

dcherian: Lets do that in another PR.

rpmanser:
Tests are failing in `test_formatting.py` for `test_diff_attrs_repr_with_array`. This looks very similar to the previous problem that was occurring for duck array equality due to NEP18. Is a similar xfail appropriate here?

I'm not sure why FormattingBlack failed. It's claiming that 45 files would be reformatted. I used the pre-commit hook, which passed.

keewis:
yes, the failing test is definitely due to NEP18: there's a numpy array in an attribute, but since numpy 1.15 doesn't define `__array_function__`, `is_duck_array` returns `False`. We could extend `is_duck_array` with an `isinstance` check for `numpy.ndarray`, or we could take your suggestion and `xfail` the test. Not sure which is better.

`black` recently released a new version, which changed the way it treats a trailing comma, and it also reformats docstrings (at least as much as it can without breaking formats), but neither `conda` nor the `pre-commit` hook install the new version (yet). To make the diff of this PR as small as possible, I'm opening a new PR with just the automatic changes, once that is merged, it should be fixed here, too.

shoyer:
> yes, the failing test is definitely due to NEP18: there's a numpy array in an attribute, but since numpy 1.15 doesn't define `__array_function__`, `is_duck_array` returns `False`. We could extend `is_duck_array` with an `isinstance` check for `numpy.ndarray`, or we could take your suggestion and `xfail` the test. Not sure which is better.

I would suggest adding an explicit check for NumPy arrays. Otherwise we would need to bump our minimum supported NumPy version.

keewis:
we went with https://github.com/pydata/xarray/blob/edd5c1e96204d59d30496e3f85e2f805a911ac9b/xarray/core/variable.py#L938 in #4379, you could use something like that here: https://github.com/pydata/xarray/blob/3337b6fc4024dcaa2c11831e6d34551adaee4a8b/xarray/core/formatting.py#L554 too, e.g. with:
```python
def is_duck_array_or_ndarray(array):
    return is_duck_array(array) or (not IS_NEP18_ACTIVE and isinstance(data, np.ndarray))
```

xarray/core/utils.py
```diff
@@ -247,9 +249,19 @@ def is_list_like(value: Any) -> bool:
...
+        and hasattr(value, "dtype")
+        and hasattr(value, "__array_function__")
+        and hasattr(value, "__array_ufunc__")
+    )
+
+
+def is_duck_array_or_ndarray(value: Any) -> bool:
+    return is_duck_array(value) or (
+        not IS_NEP18_ACTIVE and isinstance(value, np.ndarray)
     )
```
shoyer, line 265: Is there any circumstance under which we would *not* want to consider NumPy arrays duck arrays? Even if they don't define `__array_function__`, it is always fine to use functions from the `numpy` namespace on them directly. It doesn't matter if NEP-18 is active or not.

Given that we expect _most_ arguments to this function to be NumPy arrays, there's another reason include a check for NumPy arrays explicitly as well: it's much faster, e.g., this version checks a NumPy array in 170 ns vs 550 ns:
```python
def is_duck_array(value):
    if isinstance(value, np.ndarray):
        # much faster than checking for attributes
        # also, NumPy prior to 1.17 does not have an __array_function__ method
        return True
    return (
        hasattr(value, "ndim")
        and hasattr(value, "shape")
        and hasattr(value, "dtype")
        and hasattr(value, "__array_function__")
        and hasattr(value, "__array_ufunc__")
    )
```

rpmanser: I can't think of a situation where NumPy arrays shouldn't be considered duck arrays. Your solution is also more simple than having another function just for these cases. I'll go ahead and make this change.

rpmanser:
I meant `is_duck_array_or_ndarray`, not `is_duck_dask_array_or_ndarray`. I don't suppose there's a way to fix that at this point?

keewis:
you mean the commit message? If so, you can use `git commit --amend` to change it (make sure you don't have anything staged because that would be added to the commit) and force-push. It's not that important, though, I doubt anyone really reads the commit messages.

dcherian (approved):
Thanks @rpmanser

xarray/core/variable.py
```diff
@@ -932,11 +936,7 @@ def copy(self, deep=True, data=None):
                 # don't share caching between copies
                 data = indexing.MemoryCachedArray(data.array)
 
-            if deep and (
-                hasattr(data, "__array_function__")
-                or isinstance(data, dask_array_type)
-                or (not IS_NEP18_ACTIVE and isinstance(data, np.ndarray))
-            ):
+            if deep and is_duck_array(data):
                 data = copy.deepcopy(data)
```
keewis, line 940: I realize that it was removed in #4379 and not here, but do we need the special casing for `PandasIndexingAdapter`?

rpmanser: Is that covered by this test?

https://github.com/pydata/xarray/blob/13caf96efb3f121e232a35aafceed80c832a9876/xarray/tests/test_variable.py#L512-L521

keewis: unfortunately not, the special case is for when `data` is neither a duck array (or a duck dask array / a numpy array) nor a `PandasIndexAdapter`, but I'm not sure what the data could be instead. Maybe a `MemoryCachedArray`?

keewis: actually, after changing
```python
            if deep and is_duck_array(data):
                data = copy.deepcopy(data)
```
to
```python
            if deep:
                if is_duck_array(data):
                    data = copy.deepcopy(data)
                elif not isinstance(data, PandasIndexAdapter):
                    raise ValueError(f"not a index adapter: {data}, {type(data)}")
```
does not cause any test in the test suite fail, which suggests that either the test suite is lacking (or the error is suppressed somehow) or this is a piece of obsolete code. Any insights on this, @shoyer?

shoyer: The main case where non-duck-arrays can be found in `_data` is when the data was loaded from a file on disk.

My inclination is that we should simplify this and probably use a deep copy even for non-duckarray objects. This might do a few extra copies sometimes, but it's better to be overly aggressive when doing a "deep" copy.

For example, even pandas.Index itself is only "mostly" immutable. You can't change the data values, but you can modify  `.name`.

keewis, line 940: then we'd have something like this?
```suggestion
            if deep:
                data = copy.deepcopy(data)
```

shoyer: Yes, exactly! This is definitely much closer to what I would expect `copy(deep=True)` to do without looking at the implementation.

xarray/tests/test_units.py
```diff
@@ -2237,6 +2243,21 @@ def test_pad_unit_constant_value(self, unit, error, dtype):
         assert_units_equal(expected, actual)
         assert_identical(expected, actual)
 
+    @requires_dask
```
keewis, line 2246: these tests are about `dask` and duck dask arrays (`pint` is just an example of a duck dask array), so I think it makes more sense to move them to `test_dask.py`

xarray/core/variable.py
```diff
@@ -427,7 +431,7 @@ def load(self, **kwargs):
         --------
         dask.array.compute
         """
-        if isinstance(self._data, dask_array_type):
+        if is_duck_dask_array(self._data):
             self._data = as_compatible_data(self._data.compute(**kwargs))
         elif not hasattr(self._data, "__array_function__"):
```
keewis, line 436: should this be
```suggestion
        elif not is_duck_array(self._data):
```

xarray/core/rolling.py
```diff
@@ -389,7 +389,7 @@ def _bottleneck_reduce(self, func, **kwargs):
                 valid = (slice(None),) * axis + (slice(-shift, None),)
             padded = padded.pad({self.dim[0]: (0, -shift)}, mode="constant")
 
-        if isinstance(padded.data, dask_array_type):
+        if is_duck_dask_array(padded.data):
             raise AssertionError("should not be reachable")
```
keewis, line 393: @pydata/xarray: why is that here? If it is never reachable, it should be fine to just not do the branching (and if it is necessary, it should be fine to use `assert not is_duck_dask_array(padded.data), "should not be reachable"`).

Also, if we raise the code below will never be reached, so we should be able to remove it.

keewis (commented):
I have two comments, otherwise looks good to me.

xarray/core/rolling.py
dcherian: I think it's from https://github.com/pydata/xarray/pull/3040 . We would like to use this code path but there are bugs IIRC

keewis: then I guess we should raise a `NotImplementedError` or a `ValueError` instead of a `AssertionError` and improve the error message?

dcherian: We could comment it out and add a TODO note. The `if` condition in Line 420 makes sure that this is never reached.

xarray/tests/test_units.py
rpmanser: Okay, that makes sense. What is the best way to import and set up `pint` in that case? A `pytest.importorskip("pint")` wouldn't be good. I'm thinking a guarded import that contains `pint` related setup, then a `skipif` for `pint` < 0.15.

keewis: you could define a `requires_pint` decorator in `xarray/tests/__init__.py`. For the setup, I guess it's fine to add:
```python
import pint
unit_registry = pint.UnitRegistry(force_ndarray_like=True)
```
at the top of both test functions.

xarray/tests/test_dask.py
```diff
@@ -292,6 +293,26 @@ def test_persist(self):
...
         self.assertLazyAndAllClose(u + 1, v2)
 
+    @requires_pint
+    def test_tokenize_duck_dask_array(self):
+        import pint
+
+        pytest.mark.skipif(
+            pint.__version__ < "0.15",
+            reason="Dask tokenize is not implemented for Pint < 0.15",
+        )
```
keewis, line 303: does this actually work (same also for the other test)? If not, we might want to switch to `pytest.skip` (possibly in combination with a normal branch if `pytest.skipif` does not exist). Also, I'm not sure the string comparison is guaranteed to work; we might want to pass both to `LooseVersion` before comparing (even better would be `packaging.version.parse` but that would be a new dependency).

rpmanser: No, it doesn't, good catch. What about requiring the minimum version to be 0.15 for `has_pint` and `requires_pint`? Currently these tests are the only ones using the `requires_pint` wrapper, but it could be a source of confusion if it is used elsewhere in the future.  Thoughts?

keewis: If you need a specific minimum version, we could create a new set, kind of like what already exists for `cftime`: https://github.com/pydata/xarray/blob/55480de69096cc5ae003f639c2c953066e829120/xarray/tests/__init__.py#L68-L69

rpmanser: Sounds good, I'll do something similar for `pint`.

xarray/tests/test_dask.py
```diff
@@ -292,6 +293,22 @@ def test_persist(self):
         self.assertLazyAndAllClose(u + 1, v)
         self.assertLazyAndAllClose(u + 1, v2)
 
+    @requires_pint_0_15
```
keewis, line 296: could you add a comment stating that this is necessary because we need `__dask_tokenize__`? That way deciding when to remove this will be easier.

rpmanser: Yes, in `__init__.py`, `test_dask.py`, or both?

keewis: in `test_dask.py` should be enough

xarray/tests/test_dask.py
```diff
@@ -292,6 +293,23 @@ def test_persist(self):
         self.assertLazyAndAllClose(u + 1, v)
         self.assertLazyAndAllClose(u + 1, v2)
 
+    # Need Pint 0.15 for __dask_tokenize__ implementation
+    @requires_pint_0_15
```
keewis, line 297: I was about to propose extending `_importorskip` when I realized that you can, in fact, do
```suggestion
    @requires_pint_0_15(reason="needs __dask_tokenize__")
```
and the test will still be skipped if the required package is not available.

keewis (approved):
LGTM. The `reason` for `requires_pint_0_15` would be nice but I don't feel strongly about it so it's up to you.

Thanks for being patient, @rpmanser.

rpmanser:
> LGTM. The `reason` for `requires_pint_0_15` would be nice but I don't feel strongly about it so it's up to you.
> 
> Thanks for being patient, @rpmanser.

No worries, thanks for the reviews!

dcherian:
Things seem to have died down here so let's test it out.

Thanks @rpmanser This is an amazing first PR. Thanks for your efforts and welcome to xarray!

rpmanser:
@dcherian No problem! Thanks for all your help!

